PR #118, opened 2016-07-13
Youtube timestamp embed

thomasdenney:
Previously it wasn't possible to start a YouTube video at a particular timestamp. This pull request modifies a regular expression server side to pick out the timestamp if necessary, and some client side JS that overrides some of Semantic UI's embedding code in order to correctly produce the URL.


thomasdenney:
This pull request makes a whitespace change to .gitignore, sorry about that!

PR #149, opened 2016-07-25
Testing generated documentation snippets

thomasdenney:
This pull request adds a new command to the CLI for testing whether TypeScript can successfully parse and type check the snippet, and then whether the compiler can generate blocks for it. It doesn't check that the generated blocks then compile back to the same JS, as this would require modification to Blockly to get it to run server side.

You need to build the documentation snippets directory first, e.g. by running `pxt serve` before `pxt testsnippets`.

Currently this is restricted to micro:bit snippets only.


pelikhan:
Please fix linting issue (see travis build).

thomasdenney:
@pelikhan Should now be fixed, just waiting on Travis

Issue #147, opened 2016-07-24
add `for ... of` statements


tballmsft:
another good one to add for students.

thomasdenney:
I'm experimentally adding support for this, see work in progress https://github.com/Microsoft/pxt/compare/for-of-statements?expand=1

thomasdenney:
The branch linked above now supports for...of statements, although I need to do much more testing (e.g. for memory leaks)

mmoskal:
Good work! A couple comments:

* `refSuff` is for `refSuffix` not `refStuff` ;)
* for the uniqueName() use a global counter, something like:

```
+        constructor(public index: number, public owningProc: Procedure) {
+            super(index, null, null)
+            this.uid = globalCounter++
+        }
+
+        getName() {
+            return "tmp" + this.uid
+        }
+
+        uniqueName() {
+             // 'T' makes sure it doesn't clash with real cells
+            return this.getName() + "___T" + this.uid
+        }
```

thomasdenney:
@mmoskal Thanks :). I'll make these changes shortly.

PR #170, opened 2016-07-27
For of statements

thomasdenney:
This pull request adds support for `for...of` statements in JavaScript. Both arrays and strings are supported. Initially I planned to support this by transforming the AST for the `for...of` loop into the AST for a `for` loop, however generating new AST nodes with the TypeScript API was more awkward than I anticipated, so I instead went the route of just emitting the appropriate IR.

It would be nice to have support in blocks for these loops, I can add that if necessary.

The ACE editor doesn't currently highlight the TypeScript keyword `of`. I've [created an issue for this](https://github.com/ajaxorg/ace/issues/3035).

This pull request fixes https://github.com/Microsoft/pxt/issues/147.


pelikhan:
Could you add more test cases?

pxtlib/emitter/ir.ts
```diff
@@ -310,6 +312,38 @@ namespace ts.pxt.ir {
...
+            super(index, null, null)
+            this.uid = UnnamedCell.unnamedCellCounter++
+        }
+
+        getName() {
+            return "unnamed" + this.uid
+        }
+
+        uniqueName() {
+            return this.getName() + "___" + this.index
```
mmoskal: This needs to be something like "___U", otherwise there is some (small) risk it will clash with normal variables

pxtlib/emitter/ir.ts
```diff
@@ -310,6 +312,38 @@ namespace ts.pxt.ir {
...
+        }
+
+        isByRefLocal() {
+            return this.isLocal()
```
mmoskal: By-ref local means it is captured by an inner function and written subsequently, which means it has to be compiled as a one-field record (feel free to add some comments to that effect). These are never like that, so just return 'false'.

mmoskal:
I fixed ref-counting of the collection variable. You can also add a few more tests to lang-test0.ts - it actually runs, so you can test code with asserts. To run it, go to `pxt/libs/lang-test0` and do `pxt run`. Make sure it doesn't print any dangling references at the end (they look like `RefRecord...`).

thomasdenney:
@mmoskal @pelikhan I've added some more tests. None of them produce dangling references.

mmoskal:
Thank you! Good work!

PR #185, opened 2016-08-01
Fixes (pxt-microbit) issue 132: For loop does not have associated help

microsoftsam:
PR Fixes Microsoft/pxt-microbit#132


microsoftsam:
image

Issue #141, opened 2016-07-21
Opening an existing project or dragging a HEX file to the editor should open it in Java if fails to open in Blocks


darrenge:
If you have code that is not represented in Blocks or fails to convert to Blocks, and you take that associated HEX file, it still should open in JavaScript view.

Repro:
* Create a hex file with this code: input.onGesture(Gesture.Shake, () => {let item = Math.random(3)})
* Drag that hex file to the editor
** Notice - it fails to open it with code view - just gives blank Blocks view

PR #209, opened 2016-08-08
Fix for issue 141: Opening an existing project or dragging a HEX file…

microsoftsam:
Fix for issue 141: Opening an existing project or dragging a HEX file to the editor should open it in Java if fails to open in Blocks

+ Change to make the default option of the Js->Blocks dialog be to stay in JS.

fixes issue #141


microsoftsam:
image

PR #215, opened 2016-08-10
Reduce false negative rate when testing documentation snippets

thomasdenney:
- Renamed`pxt testsnippets` to `pxt snippets`
- Added `-i` flag to ignore snippets previously believed to compile correctly
- Added `-re REGEX` option to only test snippets in files whose name matches the regex
- Ignore certain snippet types that are never meant to be compiled to blocks, such as those that are strictly TypeScript or are meant to demonstrate syntax errors. You can force the test to ignore a snippet by appending `-ignore` after its type
- Use the `package` snippet at the end of a file to determine whether to use extra packages when compiling (e.g. `microbit-radio`, `microbit-bluetooth`, etc)
- Search the original documentation files for snippets, rather than the snippets directory. This means that `pxt snippets` can be run without running `pxt checkdocs` or `pxt serve`

This pull request also includes a minor change to the output of `pxt help` that produces wider columns when necessary.

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -2412,7 +2412,7 @@ function testDirAsync(dir: string) {
 }
 
 function testSnippetsAsync(...args: string[]): Promise<void> {
-    let filenameMatch = new RegExp('.*');
+    let filenameMatch = new RegExp('.*')
     if (args.length > 0) {
         try {
             filenameMatch = new RegExp(args[0])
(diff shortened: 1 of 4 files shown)


thomasdenney:
After this has been merged, [this pull request in Microsoft/pxt-microbit](https://github.com/Microsoft/pxt-microbit/pull/199) can also be merged.

Issue #218, opened 2016-08-11
cleanup docs.js


pelikhan:

pelikhan:
https://github.com/Microsoft/pxt/commit/a2acfe17ee22914acaa90699fca2f1f36865bca4

thomasdenney:
@pelikhan This was added to resolve https://github.com/Microsoft/pxt/issues/92. Do you want the functionality removed completely or moved to a separate script?

thomasdenney:
Leaving this for now.

PR #224, opened 2016-08-12
Move video timestamp functionality to a separate file

thomasdenney:
Fixes https://github.com/Microsoft/pxt/issues/218


docfiles/videos.js
```diff
@@ -0,0 +1,61 @@
+$(document).ready(function () {
+    renderSnippets();
```
pelikhan: this belongs in docs.js.

pelikhan:
I was hoping to avoid having to customize semantic UI, it does not seem that splitting the file is helping in any way. So it's probably best to simply drop this PR. Sorry for the hassle.

thomasdenney:
@pelikhan Don't worry, it was annoying that customizing Semantic UI was the most straightforward way to do this without dropping it completely but hopefully they'll fix this in the future.

PR #298, opened 2016-09-03
Do not emit null initializers

### LINKED ISSUE #283: Decompilation issue when converting Game.LedSprite back to blocks
Open Javascript Editor and type: 
let item: game.LedSprite = null
    item = game.createSprite(2, 2)
})

This is the same as: 
image

However when converting back to blocks, it fails to convert.

riknoll:
Fixes #283 

References to variables with object types, like those used by the "game" blocks, generate declarations that initialize the variables to `null`. Because we emit nothing for `null`, assignments to `null` decompile to invalid assignment blocks that recompile down to invalid typescript (type annotations are removed and you get a "no implicit any" error).

This change reports an error when trying to decompile statements that assign `null` to a variable unless they are variable declarations in the global scope (e.g. like the ones the compiler generates).

### DIFF
--- a/pxtlib/emitter/decompiler.ts
+++ b/pxtlib/emitter/decompiler.ts
@@ -304,9 +304,29 @@ ${output}</xml>`;
         }
 
         function emitVariableSetOrChange(name: string, value: ts.Expression | number, changed = false) {
+            const isNumber = typeof value === 'number';
+            if (!isNumber) {
+                const valueNode = value as ts.Expression;
+
+                // Assignments to null are not supported because there is no block value equivalent to null.
+                // However, declarations that initialize to null are fine so long as they occur in the global
+                // scope since blocks that have object types will regenerate the initializations in compilation
+                if (valueNode.kind === SyntaxKind.NullKeyword) {
+                    if (valueNode.parent.kind !== SyntaxKind.VariableDeclaration) {
+                        error(valueNode, "Assigning null to a variable is not supported in blocks");
+                    }
+                    else if (getEnclosingBlockScopeContainer(valueNode).kind !== SyntaxKind.SourceFile) {
+                        error(valueNode, "Only globally scoped initializations to null are supported in blocks");
+                    }
+
+                    // Do not emit null initializers
+                    return;
+                }
+            }
+
             writeBeginBlock(changed ? "variables_change" : "variables_set")
             write(`<field name="VAR">${Util.htmlEscape(name)}</field>`)
-            if (typeof value == 'number')
+            if (isNumber)
                 write(`<value name="VALUE"><block type="math_number"><field name="NUM">${value}</field></block></value>`);
             else {
                 write('<value name="VALUE">')


pxtlib/emitter/decompiler.ts
```diff
@@ -304,9 +304,29 @@ ${output}</xml>`;
...
+            const isNumber = typeof value === 'number';
+            if (!isNumber) {
+                const valueNode = value as ts.Expression;
+
+                // Assignments to null are not supported because there is no block value equivalent to null.
+                // However, declarations that initialize to null are fine so long as they occur in the global
+                // scope since blocks that have object types will regenerate the initializations in compilation
+                if (valueNode.kind === SyntaxKind.NullKeyword) {
+                    if (valueNode.parent.kind !== SyntaxKind.VariableDeclaration) {
+                        error(valueNode, "Assigning null to a variable is not supported in blocks");
```
pelikhan: use lf macro to make string localizable

pxtlib/emitter/decompiler.ts
```diff
@@ -304,9 +304,29 @@ ${output}</xml>`;
...
+                    }
+                    else if (getEnclosingBlockScopeContainer(valueNode).kind !== SyntaxKind.SourceFile) {
+                        error(valueNode, "Only globally scoped initializations to null are supported in blocks");
```
pelikhan: lf keyword

pelikhan: Is there any way to explain this in an easier way? :)

pelikhan:
We need to start building a test suite failing decompilation tests.

pxtlib/emitter/decompiler.ts
riknoll: How about this:
> Variable declarations that initialize to null must be in the global scope

Now that I think about it, this might not be the most helpful error message to give. Even if you move the initialization to the global scope, your program might still fail the roundtrip check if the initialization isn't in the same place as where the compiler puts it. Similarly, if the initialization is placed correctly but the variable is never referenced by a block, then the compiler won't regenerate the variable declaration and the roundtrip will fail again. Might be better off just deleting this error message altogether rather than giving false hope.

Does there need to be a scheme for declaring your variables so that you can be sure they pass the roundtrip check? This change fixes the initializers that the compiler generates, but if users want to be able to write TS files that reliably convert to blocks then it probably isn't sufficient.

pelikhan: Arguably if the user is able to successfully write variables in TS, he's not coming back to blocks. We need to have a better story to simply and explain error messages. Not just this one. Let's keep this one.

Talking about this issue: does this mean that I cannot do something like this:

```
function f() {
    let s : Sprite = null;
}
```
This is quite a drastic limitation.

riknoll: It means we can't convert that to blocks (you won't get an error typing that in the editor). That isn't a break from the current behavior, which already doesn't support decompiling anything declared in blocks. Here are some example of what declarations work and what don't:
```typescript
// This code does not decompile, because item is declared in a block
for (let i = 0; i < 4; i++) {
    let item = 0
    item = 1
}

// This code also does not decompile, because item is declared but never referenced
let item = 0
for (let i = 0; i < 4; i++) {
}

// This code does not decompile, because item is not declared in the same place
// as where the compiler would declare it:
input.onButtonPressed(Button.A, () => { })
let item = 0
for (let i = 0; i < 4; i++) {
    item = 1
}

// This code decompiles fine
let item = 0
for (let i = 0; i < 4; i++) {
    item = 1
}
```

Whitespace can mess up decompilation as well. If you place a newline between the declaration and for-loop in my last example, it wouldn't decompile.

pelikhan: Ok sounds good.

As far as decompilation, we should not rely on the rountrip forever. Instead, we should detect unsupported features and bail out. The rountrip was mostly there to catch a lot of the issues. It brings its own set of challenges (whitespace, etc...) as well. But that's beyond the point of this PR.

PR #303, opened 2016-09-05
Translate default blocks and category names

sgk:
We do not want to trust the quality of the translation of Blockly itself.
We have to prepare the translation ourselves.

For this purpose, this commit is to add blocks with the same name (id)
as the default/builtin blocks.

Do not think that the strings can be translated "word by word". Short words
like "do", "add", "if", "and" etc have multiple meanings and should be
translated into different words in different contexts. To solve this
issue, I have marked up these words as like "_repeat_do" and "_while_do".
Please be aware you have to add translations for these special strings
**even for English**.


sgk:
Signed the CLA.

pxtblocks/blocklyloader.ts
```diff
@@ -612,7 +649,7 @@ namespace pxt.blocks {
                 if (!this.isCollapsed()) {
                     let option: any = { enabled: true };
                     let name = this.getFieldValue('VAR');
-                    option.text = Blockly.Msg.VARIABLES_SET_CREATE_GET.replace('%1', name);
+                    option.text = lf("Create 'get %1'").replace('%1', name);
```
pelikhan: ``lf`` supports arguments which will be substitued to {0}, {1}, ...

   lf("Create 'get {0}', name)

See example at https://github.com/Microsoft/pxt/blob/c3f64ddcd6533ff8338346d12a7e291b03c7a7f1/webapp/src/appcache.ts#L13

pxtblocks/blocklyloader.ts
```diff
@@ -540,48 +512,113 @@ namespace pxt.blocks {
...
                 let thisBlock = this;
-                this.setTooltip(function () {
-                    return "Have the variable '%1' take on the values from 0 to the end number, counting by 1, and do the specified blocks.".replace('%1',
-                        thisBlock.getFieldValue('VAR'));
-                });
+		setHelpResources(this,
+		    'controls_simple_for',
+		    lf("a loop that repeats the number of times you say"),
+		    function () {
+			return lf("Have the variable '%1' take on the values from 0 to the end number, counting by 1, and do the specified blocks.").replace('%1', thisBlock.getFieldValue('VAR'));
```
pelikhan: ``lf`` supports arguments which will be substitued to {0}, {1}, ...

See example at https://github.com/Microsoft/pxt/blob/c3f64ddcd6533ff8338346d12a7e291b03c7a7f1/webapp/src/appcache.ts#L13

pxtblocks/blocklyloader.ts
```diff
@@ -755,12 +792,12 @@ namespace pxt.blocks {
...
             let deleteOption = {
-                text: deleteList.length == 1 ? Blockly.Msg.DELETE_BLOCK :
-                    Blockly.Msg.DELETE_X_BLOCKS.replace('%1', String(deleteList.length)),
+                text: deleteList.length == 1 ? lf("Delete Block") :
+                    lf("Delete %1 Blocks").replace('%1', String(deleteList.length)),
                 enabled: deleteList.length > 0,
                 callback: function () {
                     if (deleteList.length < 2 ||
-                        window.confirm(Blockly.Msg.DELETE_ALL_BLOCKS.replace('%1',
+                        window.confirm(lf("Delete all %1 blocks?").replace('%1',
```
pelikhan: ``lf`` supports arguments which will be substitued to {0}, {1}, ...

See example at https://github.com/Microsoft/pxt/blob/c3f64ddcd6533ff8338346d12a7e291b03c7a7f1/webapp/src/appcache.ts#L13

pxtblocks/blocklyloader.ts
```diff
@@ -842,8 +969,8 @@ namespace pxt.blocks {
             // In addition to the user's variables, we also want to display the default
             // variable name at the top.  We also don't want this duplicated if the
             // user has created a variable of the same name.
-            goog.array.remove(variableList, Blockly.Msg.VARIABLES_DEFAULT_NAME);
-            variableList.unshift(Blockly.Msg.VARIABLES_DEFAULT_NAME);
+            goog.array.remove(variableList, "item");
```
pelikhan: should we localize ``item`` as well?

pxtblocks/blocklyloader.ts
```diff
@@ -842,8 +969,8 @@ namespace pxt.blocks {
...
+            variableList.unshift("item");
```
pelikhan: Should we localize ``item`` as well?

pxtblocks/blocklyloader.ts
```diff
@@ -913,21 +1040,168 @@ namespace pxt.blocks {
...
+            'controls_if',
+            lf("a conditional statement"),
+            undefined,
+            "blocks/logic/if"
+        );
+
+        // builtin logic_compare
+	Blockly.Blocks['logic_compare'].init = function() {
+	    let OPERATORS = [
+		[lf("_op_eq"),  'EQ'],  // =
```
pelikhan: Why aren't we simply using ``lf("==")`` ?

pxtblocks/blocklyloader.ts
```diff
@@ -913,21 +1040,168 @@ namespace pxt.blocks {
...
+		lf("comparing two numbers"),
+		function() {
+		    return TOOLTIPS[thisBlock.getFieldValue('OP')];
+		},
+		'/blocks/logic/boolean'
+	    );
+	};
+
+	// builtin logic_operation
+	msg.LOGIC_OPERATION_AND = lf("_op_and");
```
pelikhan: Why not simply use lf("&&") ?

pxtblocks/blocklyloader.ts
```diff
@@ -913,21 +1040,168 @@ namespace pxt.blocks {
...
+	msg.LOGIC_OPERATION_OR = lf("_op_or");
```
pelikhan: Why not simply use ``lf("||")`` ?

pxtblocks/blocklyloader.ts
```diff
@@ -913,21 +1040,168 @@ namespace pxt.blocks {
...
+	msg.LOGIC_NEGATE_TITLE = lf("not %1");
+	installHelpResources(
+	    'logic_negate',
+	    lf("logical negation"),
+	    lf("Returns true if the input is false. Returns false if the input is true."),
+	    '/blocks/logic/boolean'
+	);
+
+	// builtin logic_boolean
+	msg.LOGIC_BOOLEAN_TRUE = lf("_logic_true");
```
pelikhan: why not simply use ``lf("true")``  and ``lf("false")``?

pxtblocks/blocklyloader.ts
sgk: I hope the word "item" to be translated, however, multi-byte character in variable name does not work currently. Trye "あ" (it is the first alphabet in Japanese) for variable name.

pelikhan:
@sgk i've added a few comments.

pxtblocks/blocklyloader.ts
sgk: As I wrote in the commit message, the words like "true" or "false" is too short and has several meanings when translated into non-English languages. To give only one translated word for such words is harmful.

sgk:
@pelikhan Thank you. I replied to two comments, but I will check remaining comments tomorrow because it's midnight here.

sgk:
For `lf()` parameter, I forgot about it even I read `lf()` and `lf_va()` source code. I will add another commit to use this feature soon.

pxtblocks/blocklyloader.ts
sgk: I changed my mind. But, it should be `lf("_default_var")` because the word "item" is short and has different meanings in different contexts. Please add a translation to "item" in English. For non-European languages, do not translate it and choose alphabet only name for "_default_var".

sgk:
The above commit is to translate the default variable name "item". However, I would like to use `_default_var` instead of `item` for translation source string. I tried to explain the reason in the commit log.

pxtblocks/blocklyloader.ts
sgk: The original was "=" (not the C-like "=="). There are some possibilities of problem if we translate "=" because of the reason I wrote several times here. So, we have to choose right translation string with single meaning instead of "=". "==" is acceptable, but we have to remember to translate it to "=" in most languages. I think "_op_eq" could be better because we all will not leave this ugly string untranslated even without any note.

Which do you like "==" or "_op_eq"?

pxtblocks/blocklyloader.ts
sgk: Same as "==" above.

pxtblocks/blocklyloader.ts
sgk: Same as "==" above.

pxtblocks/blocklyloader.ts
pelikhan: IMO, the single `=` is something we wanted to fix to make the TypeScript conversion more consistent. So let's go with ``==`` instead of ``_op_eq``.

Note that we currently don't have any EN translation files since we assume that we always use that locale in the code. Adding ids in the ``lf`` (e.g. ``_op_eq`` is not currently supported).

pxtblocks/blocklyloader.ts
```diff
@@ -11,7 +11,9 @@ namespace pxt.blocks {
         images: 45,
         variables: 330,
         text: 160,
-        lists: 260
+        lists: 260,
+	math: 230,
```
pelikhan: If you run [tslint](https://marketplace.visualstudio.com/items?itemName=eg2.tslint) in VS code, this indentation will be fixed automatically by pressing Shift+Alt+F

pxtblocks/blocklyloader.ts
```diff
@@ -684,6 +684,21 @@ namespace pxt.blocks {
...
+        msg.ADD_COMMENT = lf("Add Comment");
+        msg.EXTERNAL_INPUTS = lf("External Inputs");
+        msg.INLINE_INPUTS = lf("Inline Inputs");
+        msg.EXPAND_BLOCK = lf("Expand Block");
+        msg.COLLAPSE_BLOCK = lf("Collapse Block");
+        msg.ENABLE_BLOCK = lf("Enable Block");
+        msg.DISABLE_BLOCK = lf("Disable Block");
+        msg.DELETE_BLOCK = lf("Delete Block");
+        msg.DELETE_X_BLOCKS = lf("Delete %1 Blocks");
+        msg.HELP = "Help";
```
pelikhan: Not localized?

pxtblocks/blocklyloader.ts
sgk: Oh, sorry. I will add another commit to change tabs to spaces. I am using vim on Linux.

pxtblocks/blocklyloader.ts
sgk: Context menu is provided for both workspace and blocks. The original `initContextMenu()` provides the context menu for workspace, and the translation is included in my first commit in this PR. However, the context menu for blocks is provided by "core/block_svg.js" in Blockly's repository.

The above lines are required to give translation for context menu for blocks.

pxtblocks/blocklyloader.ts
pelikhan: If you install the (vim)[https://marketplace.visualstudio.com/items?itemName=vscodevim.vim] extension, you should not be too lost. You'll get the best TypeScript experience in VSCode.

sgk: I did not know that. I will try it.

I have just added a commit to replace tabs with spaces.

pxtblocks/blocklyloader.ts
```diff
@@ -540,48 +512,117 @@ namespace pxt.blocks {
...
+            setHelpResources(this, id, name, goog.isFunction(tooltip) ? function() { return tooltip(block); } : tooltip, url);
+        }
+    }
+
     function initLoops() {
+        let msg:any = Blockly.Msg;
+
+        // builtin controls_repeat_ext
+        msg.CONTROLS_REPEAT_TITLE = lf("repeat %1 times");
+        msg.CONTROLS_REPEAT_INPUT_DO = lf("_repeat_do"); // "do" in English
```
pelikhan: for now, do not use ids (e.g. ``_repeat_do``) in the localized as it requires more changes on our end to support. Instead, please file a bug to support those ids. Let's decouple those from this PR. So this shoud stay as ``lf("do")``

pxtblocks/blocklyloader.ts
```diff
@@ -540,48 +512,117 @@ namespace pxt.blocks {
...
+                  "previousStatement": null,
+                  "nextStatement": null,
+                  "colour": blockColors['loops']
+                });
                 this.appendStatementInput("DO")
-                    .appendField("do");
-                this.setPreviousStatement(true);
-                this.setNextStatement(true);
-                this.setTooltip(lf("Run the same sequence of actions while the condition is met."));
+                    .appendField(lf("_while_do")); // "do" in English
```
pelikhan: for now, do not use ids (e.g. ``_repeat_do``) in the localized as it requires more changes on our end to support. Instead, please file a bug to support those ids. Let's decouple those from this PR. So this should stay as ``lf("do")``

pxtblocks/blocklyloader.ts
```diff
@@ -540,48 +512,117 @@ namespace pxt.blocks {
...
-                this.appendValueInput("TO")
-                    .setCheck("Number")
-                    .setAlign(Blockly.ALIGN_RIGHT);
+                this.jsonInit({
+                  "message0": lf("for %1 from 0 to %2"),
+                  "args0": [
+                    {
+                      "type": "field_variable",
+                      "name": "VAR",
+                      "variable": lf("_default_var") // "item" in English
```
pelikhan: Same as ``_for_do`` and ``_while_do``, please open an issue and we'll fix this in a future commit.

pxtblocks/blocklyloader.ts
```diff
@@ -540,48 +512,117 @@ namespace pxt.blocks {
...
+                  "colour": blockColors['loops'],
+                  "inputsInline": true
+                });
                 this.appendStatementInput('DO')
-                    .appendField(Blockly.Msg.CONTROLS_FOR_INPUT_DO);
-                this.setPreviousStatement(true);
-                this.setNextStatement(true);
-                this.setInputsInline(true);
-                // Assign 'this' to a variable for use in the tooltip closure below.
+                    .appendField(lf("_for_do")); // "do" in English
```
pelikhan: Same as ``_for_do`` and ``_while_do``, please open an issue and we'll fix this in a future commit.

pxtblocks/blocklyloader.ts
sgk: I added a commit to use `==`, `&&` and `||` instead of `_op_`?? strings.

How do you think about other `_op_`?? strings like `_op_add` and `_op_sub` etc? I think these strings *may* be better than bare symbols like `+` or `-`. However, I don't insist on this.

pxtblocks/blocklyloader.ts
pelikhan: Are you localizing ``Help`` as  well?

pxtblocks/blocklyloader.ts
```diff
@@ -788,51 +843,141 @@ namespace pxt.blocks {
...
+        installHelpResources(
+            'math_number',
+            lf("number"),
+            (pxt.appTarget.compile && pxt.appTarget.compile.floatingPoint) ? lf("a decimal number") : lf("an integer number"),
+            '/blocks/math/random'
+        );
+
+        // builtin math_arithmetic
+        let msg:any = Blockly.Msg;
+        msg.MATH_ADDITION_SYMBOL = lf("_op_add");
```
pelikhan: Same as ``_for_do`` and ``_while_do``, please open an issue and we'll fix this in a future commit.

pxtblocks/blocklyloader.ts
```diff
@@ -842,8 +987,8 @@ namespace pxt.blocks {
             // In addition to the user's variables, we also want to display the default
             // variable name at the top.  We also don't want this duplicated if the
             // user has created a variable of the same name.
-            goog.array.remove(variableList, Blockly.Msg.VARIABLES_DEFAULT_NAME);
-            variableList.unshift(Blockly.Msg.VARIABLES_DEFAULT_NAME);
+            goog.array.remove(variableList, lf("_default_var"));
```
pelikhan: Same as ``_for_do`` and ``_while_do``, please open an issue and we'll fix this in a future commit.

pxtblocks/blocklyloader.ts
```diff
@@ -913,21 +1058,168 @@ namespace pxt.blocks {
...
+        installHelpResources(
+            'variables_get',
+            lf("get the value of a variable"),
+            lf("Returns the value of this variable."),
+            '/blocks/variables'
+        );
+
+        // builtin variables_set
+        msg.VARIABLES_SET = lf("set %1 to %2");
+        msg.VARIABLES_DEFAULT_NAME = lf("_default_var");
```
pelikhan: Same as ``_for_do`` and ``_while_do``, please open an issue and we'll fix this in a future commit.

pxtblocks/blocklyloader.ts
```diff
@@ -913,21 +1058,168 @@ namespace pxt.blocks {
...
-                this.setTooltip(lf("Changes the value of the variable by this amount"));
-                this.setHelpUrl('/reference/assign');
-                this.setColour(blockColors['variables']);
+                this.jsonInit({
+                  "message0": lf("change %1 by %2"),
+                  "args0": [
+                    {
+                      "type": "field_variable",
+                      "name": "VAR",
+                      "variable": lf("_default_var")
```
pelikhan: Same as ``_for_do`` and ``_while_do``, please open an issue and we'll fix this in a future commit.

pxtblocks/blocklyloader.ts
```diff
@@ -913,21 +1058,168 @@ namespace pxt.blocks {
...
+                );
             }
         };
     }
+
+    function initLogic() {
+        let msg:any = Blockly.Msg;
+
+        // builtin controls_if
+        msg.CONTROLS_IF_MSG_IF = lf("_logic_if");
```
pelikhan: Same as ``_for_do`` and ``_while_do``, please open an issue and we'll fix this in a future commit.

pxtblocks/blocklyloader.ts
```diff
@@ -913,21 +1058,168 @@ namespace pxt.blocks {
...
+            lf("a conditional statement"),
+            undefined,
+            "blocks/logic/if"
+        );
+
+        // builtin logic_compare
+        Blockly.Blocks['logic_compare'].init = function() {
+            let OPERATORS = [
+                [lf("=="),  'EQ'],  // =
+                [lf("_op_neq"), 'NEQ'], // ≠
```
pelikhan: Same as ``_for_do`` and ``_while_do``, please open an issue and we'll fix this in a future commit.

pxtblocks/blocklyloader.ts
```diff
@@ -913,21 +1058,168 @@ namespace pxt.blocks {
...
+                lf("comparing two numbers"),
+                function() {
+                    return TOOLTIPS[thisBlock.getFieldValue('OP')];
+                },
+                '/blocks/logic/boolean'
+            );
+        };
+
+        // builtin logic_operation
+        msg.LOGIC_OPERATION_AND = lf("&&");
```
pelikhan: On a second thought, ``&&`` and ``||`` are cryptic for kids. We should stick to the natural language description: ``lf("and")`` and ``lf("or")``. Sorry for the confusion.

pxtblocks/blocklyloader.ts
```diff
@@ -913,21 +1058,168 @@ namespace pxt.blocks {
...
+        msg.LOGIC_NEGATE_TITLE = lf("not %1");
+        installHelpResources(
+            'logic_negate',
+            lf("logical negation"),
+            lf("Returns true if the input is false. Returns false if the input is true."),
+            '/blocks/logic/boolean'
+        );
+
+        // builtin logic_boolean
+        msg.LOGIC_BOOLEAN_TRUE = lf("_logic_true");
```
pelikhan: Same as ``_for_do`` and ``_while_do``, please open an issue and we'll fix this in a future commit.

pelikhan:
Supporting "ids" in LF requires a bit more work on our end. I would suggest to split those into a new PR if possible -- or file an issue.

sgk:
@pelikhan Thanks and I understand. We have to "decouple" the multiple-meanings-problem from this PR. I will add another commit to use natural (in English) words instead of ids like "_op_??" or "_repeat_do".

pelikhan:
I'm working on adding support for ids in ``lf``. The syntax will be as follow:

    lf("do{id:_for_do}")

pelikhan:
See https://github.com/Microsoft/pxt/commit/7088e29a297de67d8a501bc62e3faf25036710ad

pelikhan:
I’ve just added support for “ids” in the lf function. https://github.com/Microsoft/pxt/commit/7088e29a297de67d8a501bc62e3faf25036710ad

sgk:
@pelikhan Seems to be nice feature!

sgk:
I did not add `{id:???}` except the example line you added in 7088e29. Do you want to include them in this PR?

pelikhan:
Yes you can now include them in this PR

Sent from Outlook Mobile<https://aka.ms/djubve>

pelikhan:
This PR looks good. We could merge it and bring the rest of the changes in a new PR.

sgk:
Thank you! I will make the new PR later.

Issue #143, opened 2016-07-22
documentation pages should have an "Edit" button


pelikhan:
that opens the corresponding github page.

PR #309, opened 2016-09-07
Edit button for documentation pages

thomasdenney:
Resolves https://github.com/Microsoft/pxt/issues/143

For any documentation page this adds a link to the corresponding Markdown on GitHub - it is necessary to set the URL to the GitHub repository in `pxtarget.json` ([I have already done this for pxt-microbit](https://github.com/Microsoft/pxt-microbit/commit/6deb0683b6bd4122ff6cf9057c1f056501cfe517)).

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -1279,7 +1279,7 @@ function renderDocs(localDir: string) {
                     name: e
                 }
             })
-            let html = pxt.docs.renderMarkdown(docsTemplate, str, pxt.appTarget.appTheme, null, bc)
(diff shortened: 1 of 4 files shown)


pelikhan:
Can we automatically read the github URL from package.json if available?

thomasdenney:
@pelikhan Reading from package.json was an option (pxt-microbit already had the GitHub URL there), but the advantage of adding it in pxtarget.json is that the contents of that file were already being passed to the Markdown rendering function, so it was the smallest change I could make. I did also consider scraping the URL from `git remote -v`, but I didn't want to risk exposing private repository URLs in the future.

PR #320, opened 2016-09-08
Add `{id:}` to short translation source strings

sgk:
Short words may have multiple meanings, and such words should be translated into different words or sentences in different contexts. The `{id:...}` part in translation source strings will not appear in web pages even when the locale is English or when no translation is available. We can use `{id:...}` to identify the context.

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -405,12 +405,12 @@ namespace pxt.blocks {
 
         // Do not remove this comment.
         // These are used for category names.
-        // lf("Loops")
-        // lf("Logic")
-        // lf("Variables")
-        // lf("Lists")
-        // lf("Text")
-        // lf("Math")
+        // lf("{id:cat}Loops")
+        // lf("{id:cat}Logic")
+        // lf("{id:cat}Variables")
+        // lf("{id:cat}Lists")
+        // lf("{id:cat}Text")
+        // lf("{id:cat}Math")
 
         // add extra blocks
         if (tb && pxt.appTarget.runtime && pxt.appTarget.runtime.extraBlocks) {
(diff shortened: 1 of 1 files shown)


pxtblocks/blocklyloader.ts
```diff
@@ -405,12 +405,12 @@ namespace pxt.blocks {
 
         // Do not remove this comment.
         // These are used for category names.
-        // lf("Loops")
-        // lf("Logic")
-        // lf("Variables")
-        // lf("Lists")
-        // lf("Text")
-        // lf("Math")
+        // lf("{id:cat}Loops")
```
pelikhan: cat or dog? :) Let's use ``{id:blocks category}``

pxtblocks/blocklyloader.ts
```diff
@@ -676,7 +676,7 @@ namespace pxt.blocks {
     function initContextMenu() {
         // Translate the context menu for blocks.
         let msg: any = Blockly.Msg;
-        msg.DUPLICATE_BLOCK = lf("Duplicate");
+        msg.DUPLICATE_BLOCK = lf("Duplicate{id:block}");
```
pelikhan: Let's keep the id in front ``{id:block}Duplicate"

pxtblocks/blocklyloader.ts
sgk: OK ;)

pxtblocks/blocklyloader.ts
sgk: OK, if you like it.

The reason I choose "duplicate block" was that it's natural word order of English.

sgk:
@pelikhan Thank you for merging. Will the translation source strings be reflected to Crowdin soon? I'd like to start translation once the new strings available on Crowdin.

pelikhan:
I just uploaded the new strings.json -- so you should see the new identifiers.

Issue #238, opened 2016-08-18
per-browser "download" instructions


pelikhan:
When user hits download and see download screen, add per-browser pictures of the "SAVE" bar.

* [ ] IE 11
* [ ] Edge
* [ ] Firefox
* [ ] Safari
* [ ] Chrome

* [x] fix doc page https://codethemicrobit.com/device/usb

thomasdenney:
The screenshots on this page are also of microbit.co.uk opposed to codethemicrobit.com

thomasdenney:
https://github.com/Microsoft/pxt-microbit/pull/239 fixes the documentation page.

PR #330, opened 2016-09-12
Download instructions

thomasdenney:
This pull request resolves https://github.com/Microsoft/pxt/issues/238.

The instructions for downloading are now a little shorter, with an optional link to a help guide (the URL of which is specified in `pxtarget.json`). If you click on a step then you get more detailed instructions. For example, these instructions are show in Chrome on OS X:

screen shot 2016-09-12 at 16 31 03
screen shot 2016-09-12 at 16 31 06
screen shot 2016-09-12 at 16 31 09

The download dialog now shows until it is dismissed, to give people enough time to click on the instructions.

The URLs to the screenshots are also specified in `pxtarget.json`, and can be specialised per browser and OS. I have verified that the browser and OS detection works correctly in Edge, IE11, Chrome, and Firefox on Windows and Safari, Chrome, and Firefox on OS X. Each image should be specified as:

```js
{
    "name": "connection" //Currently one of connection, copy, or save
    "os": "windows" //A comma separated list of windows, mac, or linux, or * for any
    "browser": "*" //A comma separated list of chrome, firefox, safari, ie11, or edge, or * for any
    "path": "" //The URL for the image to use for this particular platform
}
```

The most specific image will be used for a platform, and then browser. So if there is an image defined for `mac/*` and an image for `*/chrome`, and the user is running `mac/chrome`, then the first image is picked. 

There is a separate pull request in the pxt-microbit repository that adds the appropriate URLs.

### DIFF
--- a/webapp/src/cmds.ts
+++ b/webapp/src/cmds.ts
@@ -36,8 +36,7 @@ function browserDownloadDeployCoreAsync(resp: pxtc.CompileResult): Promise<void>
         return showUploadInstructionsAsync(fn, url);
 }
 
-enum MatchLevel
-{
+enum MatchLevel {
     None,
     Any,
     Exact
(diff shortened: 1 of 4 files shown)


webapp/src/cmds.ts
```diff
@@ -36,44 +36,105 @@ function browserDownloadDeployCoreAsync(resp: pxtc.CompileResult): Promise<void>
...
+        <p><a href="${pxt.appTarget.appTheme.usbDocs}" target="_blank">${lf("For more information on how to transfer the program to your {0} click here", boardName)}</a></p>
+    </div>` : ""}
 ${pxt.BrowserUtils.isWindows() ? `
     <div class="ui info message landscape only">
         ${lf("Tired of copying the .hex file?")}
         <a href="/uploader" target="_blank">${lf("Install the Uploader!")}</a>
     </div>
     ` : ""}
-`,
+<script type="text/javascript">$(".ui.accordion").accordion();</script>`, //This extra call needs to get fired otherwise the accordion isn't interactive
```
pelikhan: Can you automatically open the first accordion step?

thomasdenney: Yes, this can be done with the CSS. I'll change it in tomorrow.

Issue #331, opened 2016-09-12
handle unsupported browsers


pelikhan:
when a user is trying to open PXT on an older browser, the web site looks completely broken. We need to document the list of supported browser and redirect the users as needed.

- [x] build list of supported browser
- [x] document alternative to each old browser
- [x] detect browser in client and redirect to "/browsers" page

thomasdenney:
- [ ] Determine which browsers on the Raspberry Pi actually work. The default is Epiphany, although Firefox can be installed
- [ ] Verify installation instructions for Firefox on Raspberry Pi (IceWeasel has been replaced by Firefox)

thomasdenney:
Currently being worked on in https://github.com/Microsoft/pxt/tree/browser-redirect and https://github.com/Microsoft/pxt-microbit/tree/browser-redirect

PR #336, opened 2016-09-13
Detection of unsupported browsers

thomasdenney:
This pull request improves the browser detection code, adds a function [`isBrowserSupported`](https://github.com/Microsoft/pxt/blob/browser-redirect/pxtlib/browserutils.ts#L136) for determining which browsers we allow, and adds rules to `pxtarget.json` for a redirection policy depending on which unsupported OS or browser is being used (it uses the system introduced in https://github.com/Microsoft/pxt/pull/330).

Currently I'm enforcing the following policy:

* If a browser is *actually* one of Edge, Safari, Chrome, or Firefox I allow it. These browsers are all frequently updated, and I have not run into any major issues using any of them with PXT. I also allow Opera, if it is one of the more recent versions that is essentially a skinned Chromium
* Older versions of Opera are not allowed
* IE11 is allowed (the last version of IE, which runs on Windows 7 and 8). Other versions of IE are not allowed. Due to the common-place use of IE in schools, it may be necessary to allow earlier versions (although nothing earlier than 9)
* Midori, which used to be the default browser on Raspbian, is banned on all OSes. When I tested it on other OSes I found PXT crashed the browser whilst the loading animation was still playing
* Epiphany (GNOME Web) running on Raspberry Pi is not supported
* Mobile browsers (iOS Safari, Android Chrome) are allowed by this policy, although I've not tested them

Resolves https://github.com/Microsoft/pxt/issues/331

### DIFF
--- a/pxtlib/browserutils.ts
+++ b/pxtlib/browserutils.ts
@@ -82,6 +82,7 @@ namespace pxt.BrowserUtils {
     export function os(): string {
         if (isWindows()) return "windows";
         else if (isMac()) return "mac";
+        else if (isLinux() && isARM()) return "rpi";
         else if (isLinux()) return "linux";
         else return "unknown";
     }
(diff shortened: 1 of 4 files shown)


pxtlib/browserutils.ts
```diff
@@ -61,22 +63,147 @@ namespace pxt.BrowserUtils {
...
+        }
+        return matches[matches.length - 1];
+    }
+
+    export function isBrowserSupported(): boolean {
+        if (!!navigator) {
+            return true; //All browsers define this, but we can't make any predictions if it isn't defined, so assume the best
+        }
+        const v = browserVersion();
+        const isModernUpdatedBrowser = isChrome() || isFirefox() || isEdge() || isSafari();
```
pelikhan: Firefox is not updated in schools per admin policy

pelikhan:
School system typically disable auto-update of browsers so the assumption that FF/Chrome are auto-updated is false. We need to check the version number.

thomasdenney:
@pelikhan Do you know of any specific versions of Chrome/FF that schools are likely to be restricted to using, or is the best approach going to be to just figure out what the minimum version we can support?

pelikhan:
Could be anything based on what the it department would decide.

Issue #355, opened 2016-09-21
Editor doesn't launch in private browsing mode in Safari


thomasdenney:
When you are in private browsing mode in Safari (OS X/iOS), the [localStorage API is unavailable](http://stackoverflow.com/questions/21159301/quotaexceedederror-dom-exception-22-an-attempt-was-made-to-add-something-to-st), which means the editor doesn't launch.

I will add `typeof localStorage === 'object'` as a requirement in browser support, and document it in the micro:bit docs.

pelikhan:
In such case, we can simply use the memoryworkspace. I think that s should be already handled. It is important to support that scenario.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #356, opened 2016-09-21
Fixes to browser detection

thomasdenney:
Per https://github.com/Microsoft/pxt/issues/355, Safari throws an exception when trying to access local storage in private browsing mode. This pull request therefore prevents access to the editor when local storage is not available, as well as a few other minor corrections to the browser check.

### DIFF
--- a/pxtlib/browserutils.ts
+++ b/pxtlib/browserutils.ts
@@ -136,7 +136,7 @@ namespace pxt.BrowserUtils {
     let hasLoggedBrowser = false
 
     export function isBrowserSupported(): boolean {
-        if (!!navigator) {
+        if (!navigator) {
             return true; //All browsers define this, but we can't make any predictions if it isn't defined, so assume the best
         }
         const versionString = browserVersion();
--- a/pxtlib/browserutils.ts
+++ b/pxtlib/browserutils.ts
@@ -113,7 +113,7 @@ namespace pxt.BrowserUtils {
             matches = /Midori\/([0-9\.]+)/i.exec(navigator.userAgent);
         }
         else if (isSafari()) {
-            matches = /Safari\/([0-9\.]+)/i.exec(navigator.userAgent);
+            matches = /Version\/([0-9\.]+)/i.exec(navigator.userAgent);
         }
         else if (isChrome()) {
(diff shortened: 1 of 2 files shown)


pelikhan:
If local storage is unavailable, we need to use an in memory version.

thomasdenney:
OK - the changes to the browser detection code (Safari version, navigator check) should be separated out, as these are bug fixes that should be merged.

thomasdenney:
I've now removed the requirement for local storage in the browser check, but this pull request does include other fixes to the browser detection code (Safari version number). I've also left the function that checks whether local storage is available, as this will be useful for determining whether to use local storage or the in memory workspace.

webapp/src/app.tsx
```diff
@@ -1629,8 +1629,26 @@ function initHashchange() {
     });
 }
 
+function unsupportedBrowserRedirect() {
+    Util.httpGetJsonAsync(pxt.webConfig.targetCdnUrl + "target.json")
```
pelikhan: why are we reloading the target.json here?

pxtlib/browserutils.ts
```diff
@@ -135,8 +135,20 @@ namespace pxt.BrowserUtils {
 
     let hasLoggedBrowser = false
 
+    export function isLocalStorageSupported(): boolean {
```
pelikhan: This helper should be used in pxt.storage to test if localStorage can be used. Should be moved in new PR.

pxtlib/browserutils.ts
```diff
@@ -113,7 +113,7 @@ namespace pxt.BrowserUtils {
             matches = /Midori\/([0-9\.]+)/i.exec(navigator.userAgent);
         }
         else if (isSafari()) {
-            matches = /Safari\/([0-9\.]+)/i.exec(navigator.userAgent);
+            matches = /Version\/([0-9\.]+)/i.exec(navigator.userAgent);
```
pelikhan: This fix could be bundled in separate PR.

thomasdenney:
@pelikhan I'm going to separate this out into two separate pull requests for the updated browser checking code and the function for checking the availability of local storage.

Issue #333, opened 2016-09-12
We don't check variable names when decompiling for loops


riknoll:
For loops must be of this form to be decompiled:
```typescript
for (let var_name = 0; var_name < some_expression; var_name++  ) {}
```
However, we don't actually check to make sure the `var_name` is the same across all parts. The following will decompile:
```typescript
let item = 0
for (let i = 0; item < 5; i++) {}
for (let i = 0; i < 5; item++) {}
```
This gets caught by the roundtrip check, but we should error out before that

Issue #332, opened 2016-09-12
Trouble emitting arithmetic in for loops during decompilation


riknoll:
These blocks:
![arithmetic_for](https://cloud.githubusercontent.com/assets/13754588/18453478/57a064d0-78f4-11e6-801a-6e636f5151b8.PNG)

Compile to this code:
```typescript
for (let i = 0; i < 0 + 0; i++) {}
```

Which decompiles to invalid XML. One of the tags generated is mismatched

Issue #316, opened 2016-09-07
Remove roundtrip check for decompilation


riknoll:
We should instead flag errors where decompilation is not supported and fail immediately.
- [x] Build test suite of programs that should decompile
- [x] Build test suite of programs that should not decompile
- [x] Integrate test suite in build
- [x] Integrate docs test suite in build
- [ ] Remove roundtrip

abchatra:
I am closing this as we don't have work planned here.

abchatra:
Reopening after discussing with Richard. We still need to remove the roundtrip.

abchatra:
Fixed by https://github.com/Microsoft/pxt/pull/635

PR #365, opened 2016-09-22
Decompiler rewrite

### LINKED ISSUE #344: Output statements don't decompile correctly
If you have an output statement in a non-top-level scope, decompiling produces garbage xml. For example:
```typescript
if (!!true) {
    0 + 0
}
```
We should error out in this case.
Related to #343

riknoll:
This PR rewrites the decompiler to make it a bit more manageable. The things I focused on improving:

* Removing duplicated code
* Removing the multi-line template literals
* Removing the global state used for matching tags
* Fixing a lot of bugs

I also added a barebones test runner for the decompiler. The test runner will decompile a directory of `.ts` files and check the output against a baseline directory of `.blocks` file to see if it matches. To make generating test cases a little easier, I've added a `decompile` command to the CLI and modified the webapp so that you can load a project by dragging a `.blocks` file into the editor. The workflow for creating a test case looks something like this:

1. Write a test case in TypeScript
2. Run `pxt decompile testCase.ts`
3. Take the `.blocks` file that was written and drag it into the web app
4. Check the blocks in editor to make sure everything looks good
5. Move the `.blocks` file into the `tests\decompile-test\baselines` folder

To run the tests, run `pxt testdecompiler tests\decompile-test`. The plan is to eventually add more tests for code that should roundtrip, but that is not in this PR.

Fixes #344, #333, and #332

Related to #316


cli/cli.ts
```diff
@@ -2591,6 +2591,113 @@ function testDirAsync(dir: string) {
...
+        return decompileAsyncWorker(f)
+            .then(result => {
+                fs.writeFileSync(outFile, result)
+                console.log("Wrote " + outFile)
+            })
+    })
+        .then(() => {
+            console.log("Done")
+        }, error => {
+            console.log("Error: " + error)
```
pelikhan: if we want to automate the test suite, an error should fail the execution of the CLI

cli/cli.ts
```diff
@@ -2655,6 +2762,7 @@ function testSnippetsAsync(...args: string[]): Promise<void> {
         let snippets = uploader.getSnippets(source)
         // [].concat.apply([], ...) takes an array of arrays and flattens it
         let extraDeps: string[] = [].concat.apply([], snippets.filter(s => s.type == "package").map(s => s.code.split('\n')))
+        extraDeps.push("microbit")
```
pelikhan: pxt.appTarget should contain the "default library"... can we get rid of this hardcoded dependency?

pxtlib/emitter/decompiler.ts
```diff
@@ -49,293 +44,508 @@ ${output}</xml>`;
...
                 case SK.PropertyAccessExpression:
-                    emitPropertyAccessExpression(n as ts.PropertyAccessExpression); break;
+                    emitPropertyAccessExpression(n as ts.PropertyAccessExpression);
+                    break;
+                case SK.CallExpression:
+                    emitStatementBlock(n);
+                    break;
                 default:
-                    error(n);
+                    error(n, Util.lf("Unsupported syntax kind for output expression block: " + SK[n.kind]));
```
pelikhan: For localized strings, use numbered paramteers ``lf("bla bla bla {0}", param)``

pxtlib/emitter/decompiler.ts
```diff
@@ -49,293 +44,508 @@ ${output}</xml>`;
...
-                code: 1001
-            }])
-            U.pushRange(result.diagnostics, diags)
-            result.success = false;
-        }
+            function emitBinaryExpression(n: ts.BinaryExpression): void {
+                const op = n.operatorToken.getText();
+                const npp = ops[op];
+                if (!npp) {
+                    return error(n, Util.lf("Could not find operator " + op))
```
pelikhan: no concat with ``lf``.

pelikhan:
We should run the new test suite as part of our travis build... and it should fail the build if a test fails.

cli/cli.ts
riknoll: I can't seem to get the snippet command to work, so I can't really test this. This isn't a change related to my PR; I just needed to take the dependency out of the `SnippetHost` so that I could reuse that code for my decompilation stuff.

cli/cli.ts
riknoll: Fixed!

pxtlib/emitter/decompiler.ts
riknoll: Fixed!

riknoll:
I've added my tests to the Travis build, but they aren't running right now because #369 needs to be merged (none of our tests are currently running)

pelikhan:
Looks good.

Issue #68, opened 2016-06-02
Importing code from a .hex file is not showing the proper name


darrenge:
The name for code that has been imported is always showing as "untitled". It should show the name of the file when it was compiled.

Repro:
* Create script
* Change the name
* Compile
* Drag that .hex file into editor
** Notice - the name is back to "untitled"

abchatra:
Fixed by #373

PR #373, opened 2016-09-26
Fixing project saving in .hex file

abchatra:
Fixes issue #68 (Importing code from a .hex file is not showing the proper name)

Issue is the name change from the EditorPackage is not being propagated to the MainPackage config. MainPackage is used to save the details into the hex files. See [here](https://github.com/Microsoft/pxt/blob/master/pxtlib/main.ts#L450)

@pelikhan, Can you please review? I am not sure if the fix is right. Also if you could explain relationship between MainPackage and EditorPackage that would be awesome. 

Tested in local host using [forceHexDownload](http://localhost:3232/index.html#?dbg=1?forceHexDownload) tag


pelikhan:
Looks good. This area should receive a full review to see if we can simplify and still keep react happy.

PR #375, opened 2016-09-26
Fix #370 (loading dialog hides the conversion error dialog)

### LINKED ISSUE #370: Decompiler errors freeze editor into "switching to blocks..." screen
If you have a program that does not decompile, switching to blocks will freeze the site on a greyed out screen that says "switching to blocks..."

capture

Once it freezes, there is no way to get out of it except by refreshing the site. Previously, this dialog would appear:

decompiler_window

guillaumejenkins:


pelikhan:
Make sure to run tslint (or 'jake lint' from pxt)

PR #377, opened 2016-09-26
Adding a flag to enable source maps in the web app

riknoll:
This adds an optional flag to enable source maps for debugging purposes when building and serving the web app. The source maps only seem to work with Chrome right now.

With Jake:
```
jake sourceMaps=true
```

With the CLI:
```
pxt serve -include-source-maps
```

I also removed the `rootDir` option we were specifying in a few of our `tsconfig.json` files because it wasn't doing anything (it only has an effect when used with `outDir` http://www.typescriptlang.org/docs/handbook/compiler-options.html)

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -1033,7 +1033,7 @@ function addCmd(name: string) {
     return name + (/^win/.test(process.platform) ? ".cmd" : "")
 }
 
-function buildPxtAsync(includeSourceMaps: boolean): Promise<string[]> {
+function buildPxtAsync(includeSourceMaps = false): Promise<string[]> {
     let ksd = "node_modules/pxt-core"
     if (!fs.existsSync(ksd + "/pxtlib/main.ts")) return Promise.resolve([]);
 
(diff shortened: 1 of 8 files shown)


cli/cli.ts
```diff
@@ -1033,14 +1033,14 @@ function addCmd(name: string) {
     return name + (/^win/.test(process.platform) ? ".cmd" : "")
 }
 
-function buildPxtAsync(): Promise<string[]> {
+function buildPxtAsync(includeSourceMaps: boolean): Promise<string[]> {
```
guillaumejenkins: Do we want to default this to false?

cli/cli.ts
```diff
@@ -1256,7 +1256,7 @@ function buildFailed(msg: string, e: any) {
     console.log("")
 }
 
-function buildAndWatchTargetAsync() {
+function buildAndWatchTargetAsync(includeSourceMaps: boolean) {
```
guillaumejenkins: same

cli/cli.ts
guillaumejenkins: Do we want to default this to false?

riknoll: Sure. It does default to false right now, but I'll make it explicit in the declaration.

abchatra:
@riknoll Thank you!

riknoll:
No problem

PR #392, opened 2016-09-28
More decompiler test cases

riknoll:
This PR adds a test package to the repo for use in decompiler test cases. We don't have good coverage on features like call expressions and enums, so I've added some test cases that use the package to this end.

Unfortunately, `.blocks` files generated with blocks defined in the test package cannot be directly dragged into the web app like those that use pure JS can. You can still test them out to make sure the baselines you get are correct, but to do so you will need to manually edit your project's `pxt.json` to include a dependency that looks like this: `"testBlocks": "libs\testBlocks"`. Then, navigate to where your project is saved on disk (for me, it's `pxt-microbit/projects`) and replace the `main.blocks` file with the baseline you want to look at. After that, just open the project in the browser.

Also, I've added a new test runner for testing pieces of code that are not supported by the decompiler to make sure we error out when we should. Test files are defined like this:
```typescript
// @case: Some test descriptor
let x = "test case code goes here"

// @case: Another test case
let y = 1 + 1 
```
The test runner interprets this by reading the file and then separating it into the cases marked by an `@case` comment. It then tries to decompile the code in each case separately and makes sure that each generates an error.

The idea behind this isn't to define a test case for every single TS and JS feature we don't have coverage for, but more to check the edge cases in things we do support. For example, we support decompiling for-loops, but only if the following is true:
* The for-loop's initializer declares one variable
* The for-loop's condition is a `<` or `<=` operation with the declared variable on the left side
* The for-loop's incrementor is of the form `var++`

I added some test cases for what I could think of, but I'm sure there are more.


riknoll:
This is part of https://github.com/Microsoft/pxt/issues/316

PR #394, opened 2016-09-28
Change location of local projects folder

guillaumejenkins:


cli/server.ts
```diff
@@ -17,7 +17,7 @@ let root = ""
 let dirs = [""]
 let simdirs = [""]
 let docfilesdirs = [""]
-let fileDir = process.cwd()
+let userProjectsDir = path.join(os.homedir(), U.lf("My Coding Projects"));
```
pelikhan: What about just "Projects" ?

webapp/src/fileworkspace.ts
```diff
@@ -176,7 +176,7 @@ function saveAsync(h: Header, text: ScriptText) {
 
 function installAsync(h0: InstallHeader, text: ScriptText) {
     let h = <Header>h0
-    let path = "projects/" + h.name.replace(/[^a-zA-Z0-9]+/g, " ").trim().replace(/ /g, "-")
+    let path = h.name.replace(/[^a-zA-Z0-9]+/g, " ").trim().replace(/ /g, "-")
```
pelikhan: where is it install projects now?

cli/server.ts
```diff
@@ -98,7 +98,8 @@ type FsFile = pxt.FsFile;
 type FsPkg = pxt.FsPkg;
 
 function readPkgAsync(logicalDirname: string, fileContents = false): Promise<FsPkg> {
-    let dirname = path.join(fileDir, logicalDirname)
+    nodeutil.mkdirP(userProjectsDir);
```
pelikhan: don't we have a centralized helper to create those root folder?

cli/server.ts
pelikhan: Does this folder under the "My Documents" folder? I beleive homedir() points to ".pxt".

pelikhan: The "user folder" document should be enabled for electron only.

cli/cli.ts
```diff
@@ -1341,20 +1341,24 @@ function renderDocs(localDir: string) {
     console.log("Docs written.")
 }
 
-export function serveAsync(arg?: string) {
+export function serveAsync(...args: string[]) {
     forceCloudBuild = !globalConfig.localBuild
+    let hasArg = (arg: string): boolean => {
```
pelikhan: shouldn't we trim the leading `-` ? --electron or -electron should work.

webapp/src/fileworkspace.ts
```diff
@@ -176,7 +176,7 @@ function saveAsync(h: Header, text: ScriptText) {
 
 function installAsync(h0: InstallHeader, text: ScriptText) {
     let h = <Header>h0
-    let path = "projects/" + h.name.replace(/[^a-zA-Z0-9]+/g, " ").trim().replace(/ /g, "-")
+    let path = pxt.appTarget.appTheme.id + "/" + h.name.replace(/[^a-zA-Z0-9]+/g, " ").trim().replace(/ /g, "-")
```
pelikhan: this should stay "projects" in non-electron mode

cli/server.ts
```diff
@@ -13,11 +13,13 @@ import * as util from 'util';
 import U = pxt.Util;
 import Cloud = pxt.Cloud;
 
+const userProjectsDirName = U.lf("Projects");
```
pelikhan: lower case, unlocalied "projects"

PR #400, opened 2016-09-29
Surface deploy error when no board is connected to computer

guillaumejenkins:


cli/server.ts
```diff
@@ -218,7 +218,10 @@ function handleApiAsync(req: http.IncomingMessage, res: http.ServerResponse, elt
             .then(d => writePkgAsync(innerPath, d))
     else if (cmd == "POST deploy" && deployCoreAsync)
         return readJsonAsync()
-            .then(d => deployCoreAsync(d));
+            .then(d => deployCoreAsync(d))
+            .catch((e) => {
+                throwError(500, U.lf("No board connected to computer"));
```
pelikhan: could you use another error code? This usually means that the server crashed.

webapp/src/cmds.ts
```diff
@@ -111,8 +111,18 @@ function localhostDeployCoreAsync(resp: pxtc.CompileResult): Promise<void> {
...
-        data: resp
-    }).then(r => { });
+        data: resp,
+        allowHttpErrors: true
+    }).then(r => {
+        if (r.statusCode >= 400) {
+            if (r.text) {
+                core.errorNotification(r.text);
+            } else {
+                core.errorNotification(lf("Error during deployment"));
```
pelikhan: The local server should notify that no boards were connected and ask the user to connect their micro:bit.

cli/server.ts
guillaumejenkins: How about 403?

PR #408, opened 2016-10-03
Comments

### LINKED ISSUE #374: Create a comment block in Blockly that translates to comments in code
- [ ] Create a comment block in Blockly
- [ ] Compile block into a comment
- [ ] Decompile comments into a comment block

riknoll:
Fixes #374 

Rather than creating a comment block, we decided to use Blockly comments and slightly change their expand/collapse behavior. Instead of having to click the blue icon, you now just have to click the block anywhere for the comment to expand.

Comments are preserved across decompilation and compilation. Decompilation supports all comment styles (single line, multi-line), and will group the comments to whatever statement is immediately after the comment. Comments need to be on there own lines to be decompiled. For example,

```typescript
// This comment will decompile and be a part of the "set" block
let x = 1 + 2; // This comment will be ignored
x = x * x;
```

Blockly allows all blocks to be commented, so for blocks that can't really be commented in the typescript code (i.e. expressions), comments are grouped and placed above the parent statement. All comments that are generated by the compiler are single-line style and are word wrapped at an arbitrary character count


riknoll:
There is a graphical issue with comments on Chrome right now where a white box gets created behind the text that overlaps part of the comment frame. I'm not sure how to fix it, butI think it is a Blockly issue. Comments work fine in Edge.

riknoll:
Fixed the Chrome issue in 3f11cb8. The issue was a min-width being set by Semantic-UI clashing with Blockly. Not sure why it didn't show up in Edge.

webapp/style.less
```diff
@@ -1,3 +1,6 @@
 /* Import all components */
 @import 'semantic';
- 
\ No newline at end of file
+
+body.blocklyMinimalBody {
```
pelikhan: why this rule? We typically add rules to custom.css.

riknoll: Moved to custom.css

PR #411, opened 2016-10-04
Display an error for floating point numbers if not supported

### LINKED ISSUE #2: prevent decimal numbers in blocks
When not supported (like microbit), floating point numbers should not be allowed.

image

riknoll:
Fixes #2 (mostly)

Makes using floating point numbers when targeting a board that does not support them an error. It looks like this:

decimal_errors

As part of this, I had to move the max-int check from the block compiler to the emitter, so typing in a huge number will give a similar error now. I also removed the max-float check because it didn't work (not sure if we want to add in a different working one?)

### DIFF
--- a/pxtblocks/blocklycompiler.ts
+++ b/pxtblocks/blocklycompiler.ts
@@ -498,23 +498,8 @@ namespace pxt.blocks {
     ///////////////////////////////////////////////////////////////////////////////
 
     function extractNumber(b: B.Block): number {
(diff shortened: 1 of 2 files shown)


riknoll:
Oh, I also just chose arbitrary numbers for the error codes. Is there a process for choosing those?

pelikhan:
Yes there should be a "nextError" variable?

riknoll:
Ahhh, it's in a comment. We should define these in a JSON file or something.

riknoll:
Fixed

PR #421, opened 2016-10-05
Remove top-level shadow blocks when created by editor

### LINKED ISSUE #112: shadow blocks cannot be moved
Repro: get the 0=0 block, drop `receive string` on `0`. `0` jumps out and stays on the board where it cannot be moved.

riknoll:
Fixes #112 

Walks the top blocks and removes any shadow blocks whenever a shadow block is created. Unfortunately, you can still see a flicker of the shadow block appearing and disappearing. Not sure if there is a better solution that doesn't involve altering Blockly.

### DIFF
--- a/webapp/src/blocks.tsx
+++ b/webapp/src/blocks.tsx
@@ -217,6 +217,8 @@ export class Editor extends srceditor.Editor {
         this.editor.addChangeListener((ev) => {
             if (ev.recordUndo)
                 this.changeCallback();
+            if (ev.type == 'create' && ev.xml.tagName == 'SHADOW')
+                this.cleanUpShadowBlocks();
             if (ev.type == 'ui') {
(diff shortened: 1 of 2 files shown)


webapp/src/blocks.tsx
```diff
@@ -336,4 +338,13 @@ export class Editor extends srceditor.Editor {
             <sui.Button text={lf("JavaScript") } textClass="ui landscape only" icon="keyboard" onClick={() => this.openTypeScript() } />
         )
     }
+
+    cleanUpShadowBlocks() {
+        const blocks = this.editor.getTopBlocks(false);
+        for (const block of blocks) {
```
pelikhan: We tend to use a more functional approach to arrays around the code base:

    blocks.filter(b => b.isShadow_).forEach(b => b.dispose(false))

PR #423, opened 2016-10-05
Catch deploy error for missing board in CLI

guillaumejenkins:


cli/cli.ts
```diff
@@ -3066,7 +3066,15 @@ function buildCoreAsync(mode: BuildOption) {
                     console.log("no deploy functionality defined by this target")
                     return null;
                 }
-                return pxt.commands.deployCoreAsync(res);
+                return pxt.commands.deployCoreAsync(res)
```
pelikhan: I think deploy should not fail when board are not available. having to add catch statements is just going to lead to messy code.

cli/server.ts
```diff
@@ -63,7 +63,7 @@ let readFileAsync = Promise.promisify(fs.readFile)
 let writeFileAsync: any = Promise.promisify(fs.writeFile)
 
 // provided by target
-let deployCoreAsync: (r: pxtc.CompileResult) => void = undefined;
+let deployCoreAsync: (r: pxtc.CompileResult) => Promise<void | number> = undefined;
```
pelikhan: aren't we always returning a number?

cli/server.ts
```diff
@@ -219,8 +219,10 @@ function handleApiAsync(req: http.IncomingMessage, res: http.ServerResponse, elt
     else if (cmd == "POST deploy" && deployCoreAsync)
         return readJsonAsync()
             .then(d => deployCoreAsync(d))
-            .catch((e) => {
-                throwError(404, "Board not found");
+            .then((boardCount) => {
+                if (boardCount === 0) {
+                    throwError(404);    // Board not found
```
pelikhan: why are we throwing? The web client could deal with the result instead.

PR #425, opened 2016-10-05
Sharing and publishing feature configuration

microsoftsam:
- limit sharing and publishing features
- handing #project and #sandboxproject hash
- only re-render the share editor when necessary
- render screenshot svg and encoded file contents on button click

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -357,8 +357,12 @@ enum ShareMode {
 
 interface ShareEditorState {
     mode?: ShareMode;
+    publishingEnabled?: boolean;
+    currentPubId?: string;
     screenshotId?: string;
     screenshotUri?: string;
+    fileContents?: string;
+    publishing?: boolean;
 }
 
 class ShareEditor extends data.Component<ISettingsProps, ShareEditorState> {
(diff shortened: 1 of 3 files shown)


webapp/src/app.tsx
```diff
@@ -357,36 +357,76 @@ enum ShareMode {
...
+        this.modal.show();
+    }
+
+    hide() {
+        this.modal.hide();
+    }
+
+    shouldComponentUpdate(nextProps: ISettingsProps, nextState: ShareEditorState, nextContext: any): boolean {
+        if (this.state.mode != nextState.mode) return true;
+        if (this.modal && this.modal.state.visible == false) return false;
```
pelikhan: use

       !this.modal.state.visible

webapp/src/app.tsx
```diff
@@ -357,36 +357,76 @@ enum ShareMode {
...
+                this.state.currentPubId = header.pubId;
+            } else if (!this.state.publishing) {
+                this.state.publishing = true;
+                this.props.parent.exportAsync()
+                    .then(filedata => {
+                        header.pubCurrent = true;
+                        this.setState({ currentPubId: filedata, screenshotId: undefined })
+                    })
+                    .finally(() =>{
+                        this.state.publishing = false;
```
pelikhan: state needs to be change via "setState".

webapp/src/app.tsx
```diff
@@ -357,36 +357,76 @@ enum ShareMode {
...
+        if (!this.state.publishingEnabled && this.state.fileContents != this.state.currentPubId) return true;
+        return false;
+    }
+
     renderCore() {
         const header = this.props.parent.state.header;
         if (!header) return <div></div>
+        if (!header.pubCurrent ) {
+            if (this.state.publishingEnabled) {
+                this.state.currentPubId = header.pubId;
```
pelikhan: state needs to be changed via "setState".

webapp/src/app.tsx
```diff
@@ -357,36 +357,76 @@ enum ShareMode {
...
+            } else if (!this.state.publishing) {
+                this.state.publishing = true;
```
pelikhan: state needs to be changed via "setState".

webapp/src/app.tsx
pelikhan: should not change state from a render function

webapp/src/app.tsx
```diff
@@ -400,39 +440,44 @@ class ShareEditor extends data.Component<ISettingsProps, ShareEditorState> {
...
                     }
                     break;
-            }
+                }
         }
 
         const publish = () => {
             pxt.tickEvent("menu.embed.publish");
-            this.props.parent.publishAsync().done();
+            this.state.publishing = true;
```
pelikhan: state changes need to be done via "setState"

pelikhan:
IMO, using react is becoming an overkill. We should simply use ``core.confirmAsync`` and use jquery to build a dialog.

PR #427, opened 2016-10-05
sharing: adding sharing and publishing config in pxttarget

microsoftsam:


localtypings/pxtarget.d.ts
```diff
@@ -58,6 +58,8 @@ declare namespace pxt {
     interface AppCloud {
         workspaces?: boolean;
         packages?: boolean;
+        sharing?: boolean;
+        publishing?: boolean;
```
pelikhan: where is publishing used?

PR #429, opened 2016-10-05
Handle name collisions in the decompiler

### LINKED ISSUE #151: Blocks cannot be used for variables declared inside lambdas
Related to https://github.com/Microsoft/pxt-microbit/issues/149. Several examples in the documentation cannot be represented in blocks because local variables are declared inside a lambda, however are then moved to global scope by the blockly -> JS compiler.

riknoll:
Fixes #151 

This PR attempts  to handle name collisions in the Decompiler. Because all variable names in Blockly are global, we need to make sure that if variables are declared with the same name (e.g. in different scopes) then we rename them. Variables that are renamed simply have a number added to the end of their name.

The only place that this gets annoying is with For loops, because the initializers that the Blockly compiler emits by default will get renamed if you have more than one in your program. If we want to prevent that, we can always treat For loop initializers as special and allow them to share variable names.


pxtlib/emitter/decompiler.ts
```diff
@@ -817,5 +852,70 @@ ${output}</xml>`;
...
+            return typeof n === "string"
+        }
+
+        function getNewName(name: string) {
+            if (!takenNames[name]) {
+                takenNames[name] = true;
+                return name;
+            }
+
+            for (let i = 0; ; i++) {
```
pelikhan: if the variable name is "i", we could be smart and try to pick "j" and so forth, then fallback to numbers.

pelikhan: probably want to start counting at 1 or 2 (so we get i, i1, i2, i3, i4, ....)

riknoll: Yeah, that sounds good to me. I'll make that change.

PR #431, opened 2016-10-05
sharing: file encoded uri sharing by default, cloud sharing if publishing is enabled

microsoftsam:
sharing: by default allow file encoded sharing for screenshots and editor. if publishing is enabled, allow cloud sharing.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -395,26 +395,27 @@ class ShareEditor extends data.Component<ISettingsProps, ShareEditorState> {
     }
 
     shouldComponentUpdate(nextProps: ISettingsProps, nextState: ShareEditorState, nextContext: any): boolean {
-        if (this.modal && this.modal.state.visible == false) return false;
+        if (this.modal && !this.modal.state.visible) return false;
         return true;
     }
 
     renderCore() {
         const header = this.props.parent.state.header;
+        let currentPubId = this.state.currentPubId;
         if (!header) return <div></div>
 
         let rootUrl = pxt.appTarget.appTheme.embedUrl
         if (!/\/$/.test(rootUrl)) rootUrl += '/';
 
         const mode = this.state.mode;
-        const ready = (!!this.state.currentPubId && header.pubCurrent);
+        const ready = (!!currentPubId && header.pubCurrent);
         let url = '';
         let embed = '';
         let help = lf("Copy this HTML to your website or blog.");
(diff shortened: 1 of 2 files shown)


webapp/src/app.tsx
```diff
@@ -359,16 +359,44 @@ interface ShareEditorState {
...
+                        header.pubCurrent = true;
+                        this.setState({ currentPubId: filedata, screenshotId: undefined })
+                    });
+            }
         }
+        this.modal.show();
+    }
+
+    shouldComponentUpdate(nextProps: ISettingsProps, nextState: ShareEditorState, nextContext: any): boolean {
+        if (this.modal && this.modal.state.visible == false) return false;
```
pelikhan: use
    !this.modal.state.visible

visible might be undefined

webapp/src/app.tsx
```diff
@@ -400,22 +428,22 @@ class ShareEditor extends data.Component<ISettingsProps, ShareEditorState> {
...
                             embed = `<a href="${editUrl}"><img src="${this.state.screenshotUri}" /></a>`
                         else embed = lf("Ooops, no screenshot available.");
-                    }
-                    else {
+                    } else {
+                        pxt.debug("rendering share-editor screenshot png");
                         embed = lf("rendering...");
                         pxt.blocks.layout.toPngAsync(this.props.parent.blocksEditor.editor)
-                            .done(uri => this.setState({ screenshotId: header.pubId, screenshotUri: uri }));
+                            .done(uri => this.setState({ screenshotId: this.state.currentPubId, screenshotUri: uri }));
```
pelikhan: race condition here. capture this.state.currentPubId in local variable

webapp/src/app.tsx
```diff
@@ -442,11 +471,15 @@ class ShareEditor extends data.Component<ISettingsProps, ShareEditorState> {
...
+                                { mode: ShareMode.Editor, label: lf("Editor") }]
+                                .concat(
+                                    this.state.publishingEnabled ? [
+                                        { mode: ShareMode.Simulator, label: lf("Simulator") },
+                                        { mode: ShareMode.Cli, label: lf("Command line") }
+                                    ] : []
+                                )
                                 .map(f =>
-                                    <div className="field">
+                                    <div key={f.mode.toString()} className="field">
```
pelikhan: toString() implicit I think

PR #434, opened 2016-10-06
"Open in full editor" button in sandbox mode, fixes #426

### LINKED ISSUE #426: "Open in full editor" button in sandbox mode
In sandbox mode, we need a new button "Open in Full Editor" that reopens the current script using the #project: hash

microsoftsam:
If in sandboxOn mode (sandbox=1)
- launch full editor without sandbox
If in sandboxPub mode (sandbox:PUBID) 
- ensure published project is the latest, if not, publish the project, then go to full editor with new published project id (pub:PUBID in the path)
If in sandboxProject mode (sandbox:FILECONTENTS)
- export project to get the latest file contents, then go to full editor with project:FILECONTENTS in the path

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1169,39 +1169,17 @@ export class ProjectView extends data.Component<IAppProps, IAppState> {
     }
 
     launchFullEditor() {
-        let sandbox = /sandbox=1/i.test(window.location.href)
-        let sandboxPub = /#sandbox:/i.test(window.location.href)
-        let sandboxProject = /#sandboxproject:/i.test(window.location.href)
-        Util.assert(sandbox || sandboxPub || sandboxProject);
+        Util.assert(sandbox);
 
         let rootUrl = pxt.appTarget.appTheme.embedUrl;
         if (!/\/$/.test(rootUrl)) rootUrl += '/';
 
-        const launchEditor = (url: string) => {
-            pxt.tickEvent("sandbox.launchexternaleditor");
-            let editUrl = `${rootUrl}${url}`;
-            window.open(editUrl, '_blank')
-        }
-
-        if (sandboxPub) {
-            let mpkg = pkg.mainPkg
-            let epkg = pkg.getEditorPkg(mpkg)
-            if (epkg.header.pubCurrent) {
-                launchEditor(`#pub:${epkg.header.pubId}`);
-            } else {
-                this.publishAsync()
-                    .done(() => {
-                        launchEditor(`#pub:${epkg.header.pubId}`);
-                    })
-            }
-        } else if (sandboxProject) {
-            this.exportAsync()
-                .then(fileContent => {
-                    launchEditor(`#project:${fileContent}`);
-                })
-        } else {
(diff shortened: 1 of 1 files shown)


microsoftsam:
Note: opening in a new tab, although we might want to consider opening in current tab to avoid ad block.

webapp/src/app.tsx
```diff
@@ -1168,6 +1168,42 @@ export class ProjectView extends data.Component<IAppProps, IAppState> {
             })
     }
 
+    launchFullEditor() {
+        let sandbox = /sandbox=1/i.test(window.location.href)
+        let sandboxPub = /#sandbox:/i.test(window.location.href)
+        let sandboxProject = /#sandboxproject:/i.test(window.location.href)
+        Util.assert(sandbox || sandboxPub || sandboxProject);
```
pelikhan: this assertion is invalid. we do enter the sandbox mode automatically when hosted in iframe. See global sandbox flag.

webapp/src/app.tsx
```diff
@@ -1168,6 +1168,42 @@ export class ProjectView extends data.Component<IAppProps, IAppState> {
             })
     }
 
+    launchFullEditor() {
+        let sandbox = /sandbox=1/i.test(window.location.href)
```
pelikhan: there is already a global sandbox variable computed when launching the app

webapp/src/app.tsx
```diff
@@ -1168,6 +1168,42 @@ export class ProjectView extends data.Component<IAppProps, IAppState> {
...
+        let sandboxProject = /#sandboxproject:/i.test(window.location.href)
+        Util.assert(sandbox || sandboxPub || sandboxProject);
+
+        let rootUrl = pxt.appTarget.appTheme.embedUrl;
+        if (!/\/$/.test(rootUrl)) rootUrl += '/';
+
+        const launchEditor = (url: string) => {
+            pxt.tickEvent("sandbox.launchexternaleditor");
+            let editUrl = `${rootUrl}${url}`;
+            window.open(editUrl, '_blank')
```
pelikhan: popup blockers might be an issue. can we change the top location?

webapp/src/app.tsx
```diff
@@ -1168,6 +1168,42 @@ export class ProjectView extends data.Component<IAppProps, IAppState> {
...
+        }
+
+        if (sandboxPub) {
+            let mpkg = pkg.mainPkg
+            let epkg = pkg.getEditorPkg(mpkg)
+            if (epkg.header.pubCurrent) {
+                launchEditor(`#pub:${epkg.header.pubId}`);
+            } else {
+                this.publishAsync()
```
pelikhan: IMO, we should not publish to swap out of the sandbox. Always encode as project and pass along

webapp/src/app.tsx
microsoftsam: yeah that was my concern as well. lets keep it in the current window

webapp/src/app.tsx
microsoftsam: Okay cool.

PR #453, opened 2016-10-08
Fix incomplete translation

sgk:
#303 was incomplete because `_localize()` does not know the `{id:...}` notation.

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -398,8 +398,8 @@ namespace pxt.blocks {
         if (tb) {
             let cats = tb.querySelectorAll('category');
             for (let i = 0; i < cats.length; i++) {
-                cats[i].setAttribute('name',
-                    Util._localize(cats[i].getAttribute('name')));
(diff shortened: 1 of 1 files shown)


pelikhan:
@riknoll is this still relevant with the refactoring on the categories?

riknoll:
@pelikhan yeah, now that the toolbox has been moved to `toolbox.ts`, we can just do the localization for category names like we would any other string.. We should do the localization directly in that file

sgk:
@pelikhan @riknoll The translation of category names is not just working at this point even after you updated the translation strings by a5f1b6c. This was my mistake. It should be best to move the translation feature to `toolbox.ts`. If it takes days or weeks to do so, the better way is to merge this PR to translate category names at this point.

The second commit is translate the word "Help" which is forgotten to be marked up with `lf()`.

pelikhan:
Working on translations. In progress.

pelikhan:
Fixing these issues in separate checkins. Stay tuned.

sgk:
Thank you. I will try to build.

pelikhan:
You should be able to provide translations for all categories in microbit/core-strings.json and related files.

PR #467, opened 2016-10-11
Subcategories

### LINKED ISSUE #436: Limit number of blocks that appear in a category by default
X Number of blocks can appear with a "More" sub-category that reveals the rest.

riknoll:
Fixes #436 

Adds support for the "advanced" comment attribute for blocks and namespaces. If a block is marked as advanced, it will be placed in a "More..." subcategory of its parent category like so:

toolbox_1

Here is the subcategory expanded:

toolbox_2

Namespaces that are marked as advanced will be placed in a special "Advanced" subcategory at the end of the toolbox

toolbox_3

And here is what that category looks like when expanded

toolbox_4

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -897,50 +897,52 @@ namespace pxt.blocks {
         // We override Blockly's category mouse event handler so that only one
         // category can be expanded at a time. Also prevent categories from toggling
         // once openend.
-        (<any>Blockly).Toolbox.TreeNode.prototype.onMouseDown = function(a: any) {
+        Blockly.Toolbox.TreeNode.prototype.onMouseDown = function(a: Event) {
+            const that = <Blockly.Toolbox.TreeNode>this;
 
             // Collapse the currently selected node and its parent nodes
-            if (!this.isSelected()) {
-                collapseMoreCategory(this.getTree().getSelectedItem(), this);
+            if (!that.isSelected()) {
+                collapseMoreCategory(that.getTree().getSelectedItem(), that);
             }
 
-            if (this.hasChildren() && this.isUserCollapsible_) {
+            if (that.hasChildren() && that.isUserCollapsible_) {
                 // If this is a category of categories, we want to toggle when clicked
-                if (this.getChildCount() > 1) {
-                    this.toggle();
-                    if (this.isSelected()) {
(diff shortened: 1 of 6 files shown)


riknoll:
The corresponding pxt-microbit PR is here: https://github.com/Microsoft/pxt-microbit/pull/273

pxtblocks/blocklyloader.ts
```diff
@@ -841,7 +894,69 @@ namespace pxt.blocks {
...
+            // Only collapse categories that have a single child (e.g. "More...")
+            if (cat.getChildCount() === 1 && cat.isUserCollapsible_ && (!child || !isChild(child, cat))) {
+                cat.setExpanded(false);
+                cat.updateRow();
+            }
+            cat = cat.getParent();
+        }
+    }
+
+    function isChild(child: any, parent: any): boolean {
```
pelikhan: Can we avoid "any"?

pxtblocks/blocklyloader.ts
```diff
@@ -841,7 +894,69 @@ namespace pxt.blocks {
...
+
+            if (a === null) {
+                collapseMoreCategory(this.selectedItem_);
+            }
+
+            oldSetSelectedItem.call(this, a);
+        }
+    }
+
+    function collapseMoreCategory(cat: any, child?: any) {
```
pelikhan: Can we avoid "any"?

pxtblocks/blocklyloader.ts
```diff
@@ -841,7 +894,69 @@ namespace pxt.blocks {
             Blockly.ContextMenu.show(e, menuOptions, this.RTL);
         };
 
+        // We override Blockly's category mouse event handler so that only one
+        // category can be expanded at a time. Also prevent categories from toggling
+        // once openend.
+        (<any>Blockly).Toolbox.TreeNode.prototype.onMouseDown = function(a: any) {
```
pelikhan: can we avoid "any"?

riknoll:
@pelikhan sure, I can do that. The reason I didn't initially was that all of those methods I am calling are internal to Blockly and not really supposed to be exposed.

riknoll:
I've updated `Blockly.d.ts`

pelikhan:
Looks good.

PR #503, opened 2016-10-15
Theming

pelikhan:
- [x] changed javascript language icon
- [x] move project name next to logo
- [x] using items instead of button in top nav menu
- [x] support for inverted menu in settings

image


abchatra:
:shipit:

PR #512, opened 2016-10-18
API autoupdate support

microsoftsam:
Support for API updates configured in the pxttarget.json under upgrades. 
- Defining a new upgrade policy type: API upgrade
- Upgrade a file if it matches any of the upgrade regexes
- Forcing a write file when there is a match 
- Updating the editor when a force update happens

### DIFF
--- a/pxtlib/main.ts
+++ b/pxtlib/main.ts
@@ -113,7 +113,7 @@ namespace pxt {
 
     export interface Host {
         readFile(pkg: Package, filename: string): string;
-        writeFile(pkg: Package, filename: string, contents: string): void;
+        writeFile(pkg: Package, filename: string, contents: string, force?: boolean): void;
         downloadPackageAsync(pkg: Package): Promise<void>;
         getHexInfoAsync(extInfo: pxtc.ExtensionInfo): Promise<any>;
         cacheStoreAsync(id: string, val: string): Promise<void>;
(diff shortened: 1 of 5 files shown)


pxtlib/main.ts
```diff
@@ -262,9 +262,10 @@ namespace pxt {
             if (upgrades) {
                 upgrades.forEach((rule) => {
                     if (rule.type == "package") {
```
pelikhan:
```
.filter(rule => rule.type == "package")
```

pxtlib/main.ts
```diff
@@ -262,9 +262,10 @@ namespace pxt {
...
-                        for (let match in rule.map) {
+                        let pkgRule = rule as ts.pxtc.PackageUpgradePolicy;
+                        for (let match in pkgRule.map) {
                             if (newPackage == match) {
-                                newPackage = rule.map[match];
+                                newPackage = pkgRule.map[match];
```
pelikhan: should you break here? or is it cascading?

pxtlib/main.ts
```diff
@@ -273,6 +274,23 @@ namespace pxt {
             return newPackage;
         }
 
+        upgradeAPI(fileContents: string): string {
+            const upgrades = appTarget.compile ? appTarget.compile.upgrades : undefined;
+            let updatedContents = fileContents;
+            if (upgrades) {
+                upgrades.forEach((rule) => {
```
pelikhan: .filter(rule => rule.type == "api")

PR #514, opened 2016-10-18
Setting the default editor font size to be smaller on mobile devices.

microsoftsam:
Mobile default: 
screen shot 2016-10-18 at 1 38 04 pm

Desktop default:
screen shot 2016-10-18 at 1 38 21 pm


microsoftsam:
Reducing the line number column to reserve for 3 digits wide. (most kids aren't going to have more than 999 lines of code).

Issue #465, opened 2016-10-11
Can't easily get serial number that data was received from


jaustin:
Referring to using the radio module on the micro:bit.

I can set the serial number:
![image](https://cloud.githubusercontent.com/assets/147113/19281476/6474fe3c-8fe2-11e6-848a-04581e085440.png)

But it looks like I don't have anything to "get" that at the receiving end.

I'd expect a block, perhaps like "radio receive number" or signal strenght that works inside the "on radio receive" event that gives the serial number as a number.
![image](https://cloud.githubusercontent.com/assets/147113/19281561/b356ed94-8fe2-11e6-9919-531a4060ae66.png)
^ lIke this, but "radio received serial number"

Ideally, in the event handler, the number, serial number, signal strength are all atomic for that event. (IE even if more radio data is received, while in *that* flow, the serial number is not updated, so if I store a serial number at the beginning of an event handler and a number at the end, I'm sure they are from the same packet)

riknoll:
Having separate output blocks for each part of the packet is confusing right now. It isn't clear if using one of them "consumes" the packet or not. I think that maybe something like this might work better and be more in line with the flow bit you mentioned:

![packet](https://cloud.githubusercontent.com/assets/13754588/19325903/18e39852-907d-11e6-9b7b-90f5bc2722cc.PNG)

In that block, the variable names would work like the index in the for-loop block and be declared locally:

```typescript
radio.onPacketReceived((serial: number, signal: number, data: number) => {

});
```

We could also give the block a mutation handler to let you add/remove the parts of the packet

pelikhan:
Yes! This is much better.

jaustin:
Yea, I like this too - would it not be a breaking change to the radio module? That's well and good and better to make it sooner rather than later if so - might be a nice test of how we handle breaking changes to scripts.

Is there a nice trivial migration from the old mechanism to the new one?

riknoll:
I don't know if there is an easy way to rewrite the ts code, but I spoke with @pelikhan and he suggested one option would be to keep around the old APIs and just not surface them in the block editor. That way you could still import old files and we can deprecate the APIs for eventual removal.

riknoll:
Alright, I've been looking into this and trying to come up with a clean API for receiving packets. Here are our existing APIs for sending packets and the data that each of their packets comes with:

API                         | 1         | 2     | 3                             | 4             | 5
----------------------------|-----------|-------|-------------------------------|---------------|-------
[...]

PR #522, opened 2016-10-19
Adding support for callbacks that take parameters

riknoll:
In support of #465 

The purpose of this PR is to allow callback blocks to have parameters that can be adjusted using Blockly mutators. If you have a callback that takes in an object parameter, you can annotate the block so that it creates variable declarations for each property of the object. It looks like this:

radio_variables

Note that variable blocks appear in the Toolbox for each of the callback parameters. To configure what parameters show up in the block, you can open the mutator dialog like so:

customizing_radio

The parameters can be dragged into the block in the mini-workspace and will appear in the actual block. Blockly doesn't really let you control if duplicate blocks are dragged in, so users are free to drag in the same parameter multiple times and duplicates are just ignored.

You can also declare the default parameter sets you want to appear in the toolbox, like I did here:

radio_flyout

The idea is that you can define the most common scenarios and hide any advanced features behind the mutator dialog since most people just ignore that blue gear.

I still need to add documentation and a few tests, but wanted to get the review going now since it is a big PR


riknoll:
For the record, the blocks compile to look something like this:

```typescript
radio.ondataPacketReceived(({sentNumber, name}) => {

});
```

Variable renaming is supported as well.

PR #547, opened 2016-10-22
Fixing reference counting of event handlers in the simulator runtime

### LINKED ISSUE #546: Adding an event listener in the simulator runtime does not increment its reference count
Causes the simulator to throw an exception if the event handler captures any variables. Only reproduces if the variables are declared in a local scope.

```typescript
someFunction();

function someFunction() {
   let a = 5;
   input.onButtonPressed(Button.A () => {
        basic.showNumber(a)
    })
}
```

riknoll:
Fixes #546

### DIFF
--- a/pxtsim/runtime.ts
+++ b/pxtsim/runtime.ts
@@ -172,7 +172,7 @@ namespace pxsim {
     export class EventQueue<T> {
         max: number = 5;
         events: T[] = [];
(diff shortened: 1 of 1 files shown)


mmoskal:
Looks good!

abchatra:
Looks good to me as well.

PR #564, opened 2016-10-24
Modifying simulator interface and fixing mutation property name

riknoll:
This is a breaking change for the simulator. In support of https://github.com/Microsoft/pxt-microbit/pull/287

### DIFF
--- a/pxtlib/emitter/decompiler.ts
+++ b/pxtlib/emitter/decompiler.ts
@@ -636,7 +636,7 @@ ${output}</xml>`;
                             return "";
                         }
                     });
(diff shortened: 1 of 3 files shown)


pxtlib/emitter/decompiler.ts
```diff
@@ -636,7 +636,7 @@ ${output}</xml>`;
                             return "";
                         }
                     });
-                    write(`<mutation callbackProperties="${names.join(",")}"></mutation>`)
+                    write(`<mutation callbackproperties="${names.join(",")}"></mutation>`)
```
pelikhan: should you use savedMutationAttribute here?

PR #578, opened 2016-10-25
Move hex cache

guillaumejenkins:
Offline HEX cache is now under built/, and gets populated by `pxt serve`

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -90,23 +90,6 @@ function configPath() {
     return path.join(homePxtDir(), "config.json")
 }
 
-function getTemplates() {
-    let templates: string[] = [];
-    let templatesRoot = path.join("libs", "templates");
-
-    if (fs.existsSync(templatesRoot)) {
-        fs.readdirSync(templatesRoot).forEach((f) => {
-            let fullPath = path.join(templatesRoot, f);
-
-            if (fs.statSync(fullPath).isDirectory()) {
-                templates.push(fullPath);
-            }
-        });
-    }
-
-    return templates;
-}
-
 let homeDirsMade = false
 function mkHomeDirs() {
     if (homeDirsMade) return
@@ -1203,12 +1186,13 @@ function readLocalPxTarget() {
     return cfg
 }
 
-function forEachBundledPkgAsync(f: (pkg: pxt.MainPackage, dirname: string) => Promise<void>, includeTemplates: boolean = false) {
+function forEachBundledPkgAsync(f: (pkg: pxt.MainPackage, dirname: string) => Promise<void>, includeProjects: boolean = false) {
     let prev = process.cwd()
     let folders = pxt.appTarget.bundleddirs;
 
-    if (includeTemplates) {
-        folders = folders.concat(getTemplates());
+    if (includeProjects) {
+        let projects = nodeutil.allFiles("libs", 1, /*allowMissing*/ false, /*includeDirs*/ true).filter(f => /prj$/.test(f));
+        folders = folders.concat(projects);
     }
 
     return Promise.mapSeries(folders, (dirname) => {
@@ -1333,7 +1317,7 @@ export function buildTargetAsync(): Promise<void> {
         .then(buildTargetCoreAsync)
         .then(buildSemanticUIAsync)
         .then(() => buildFolderAsync('cmds', true))
-        .then(() => buildFolderAsync('server', true));
+        .then(() => buildFolderAsync('server', true))
 }
 
 function buildFolderAsync(p: string, optional?: boolean): Promise<void> {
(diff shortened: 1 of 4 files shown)


cli/server.ts
```diff
@@ -23,7 +23,7 @@ let userProjectsDir = path.join(process.cwd(), userProjectsDirName);
 let docsDir = ""
 let tempDir = ""
 let packagedDir = ""
-let localHexDir = path.join("hexcache");
+let localHexDir = path.join("built", "hexcache");
```
pelikhan: didn't you define a const variable for "built"?

pelikhan:
I can still repro the issue:

- link pxt in pxt-microbit
- delete built folder
- add a dummy function in basic.cpp to trigger a C++ compilation
- run "pxt buildtarget"

```
----- Offline HEX cache
Extracting target's default project
Installing project's packages (pxt install)
Building the project (pxt build --cloud)
INTERNAL ERROR: Error: dal.d.ts not found in core
    at Object.assert (C:\gh\pxt\built\pxt.js:54542:27)
    at Object.getExtensionInfo (C:\gh\pxt\built\pxt.js:55643:30)
    at C:\gh\pxt\built\pxt.js:57448:35
    at tryCatcher (C:\gh\pxt\node_modules\bluebird\js\release\util.js:16:23)
    at Promise._settlePromiseFromHandler (C:\gh\pxt\node_modules\bluebird\js\release\promise.js:509:31)
    at Promise._settlePromise (C:\gh\pxt\node_modules\bluebird\js\release\promise.js:566:18)
    at Promise._settlePromiseCtx (C:\gh\pxt\node_modules\bluebird\js\release\promise.js:603:10)
    at Async._drainQueue (C:\gh\pxt\node_modules\bluebird\js\release\async.js:143:12)
    at Async._drainQueues (C:\gh\pxt\node_modules\bluebird\js\release\async.js:148:10)
    at Immediate.Async.drainQueues (C:\gh\pxt\node_modules\bluebird\js\release\async.js:17:14)
    at runCallback (timers.js:574:20)
    at tryOnImmediate (timers.js:554:5)
    at processImmediate [as _immediateCallback] (timers.js:533:5)
```

cli/cli.ts
```diff
@@ -1166,6 +1165,10 @@ function uploadCoreAsync(opts: UploadOptions) {
         })
 }
 
+function saveLocalPxTarget(cfg: pxt.TargetBundle) {
+    fs.writeFileSync("pxtarget-temp.json", JSON.stringify(cfg, null, 4));
```
pelikhan: why do we need this?

localtypings/pxtarget.d.ts
```diff
@@ -24,6 +24,7 @@ declare namespace pxt {
     }
 
     interface ProjectTemplate {
+        _warning?: string;
```
pelikhan: why do we need this?

cli/cli.ts
```diff
@@ -1494,8 +1498,53 @@ function buildSemanticUIAsync() {
...
+    let templatesRoot = path.join("libs", "templates");
+    let defaultProjects = [
+        "blocksprj",
+        "tsprj"
+    ];
+
+    defaultProjects.forEach((p) => {
+        let projectPath = path.join(templatesRoot, p);
+        let newProject: pxt.ProjectTemplate = {
+            _warning: `DO NOT MODIFY. This default project is generated on every target build. To change, modify the template at ${projectPath.replace(/\\\\/g, "/")}`,
```
pelikhan: not needed as the generated target.js files are in the built folder

cli/cli.ts
guillaumejenkins: Woops this "-temp" is a leftover from my testing. I actually meant to overwrite the existing pxtarget.json.

pelikhan: Don't do that. The pxtarget.json is already rewritte as part of the build and dump into the built folder as "target.json" and "target.js". Rewriting should occur there.

cli/cli.ts
```diff
@@ -3769,59 +3812,51 @@ export function extractAsync(...args: string[]) {
 }
 
 export function preCacheHexAsync() {
-    let trgInfo = readLocalPxTarget();
+    pxt.log("----- Offline HEX cache");
```
pelikhan: Move caching logic to separate PR

cli/cli.ts
```diff
@@ -1525,16 +1540,20 @@ function buildSemanticUIAsync() {
...
-    let defaultProjects = [
+    let defaultTemplates = [
         "blocksprj",
         "tsprj"
     ];
 
-    defaultProjects.forEach((p) => {
-        let projectPath = path.join(templatesRoot, p);
+    getTemplates().forEach((templatePath) => {
+        if (defaultTemplates.indexOf(path.basename(templatePath)) === -1) {
```
pelikhan: Isn't this an overkill? If we have more templates in the future, we'll deal with that.

cli/cli.ts
```diff
@@ -1543,30 +1562,36 @@ function updateDefaultProjects(cfg: pxt.TargetBundle) {
             files: {}
         };
 
-        if (!fs.existsSync(projectPath)) {
+        if (!fs.existsSync(templatePath)) {
```
pelikhan: isn't templatePath guaranteed to exist at this point? getTemplates() returns existing paths.

cli/cli.ts
```diff
@@ -1543,30 +1562,36 @@ function updateDefaultProjects(cfg: pxt.TargetBundle) {
...
+        fs.readdirSync(templatePath).forEach((f) => {
+            let file = path.join(templatePath, f);
             if (fs.statSync(file).isDirectory()) {
                 return;
             }
 
             if (f === "pxt.json") {
                 newProject.config = nodeutil.readJson(file);
+                U.iterMap(newProject.config.dependencies, (k, v) => {
+                    if (v.indexOf("file:") === 0) {
```
pelikhan: /^file:/.test(v)

cli/cli.ts
```diff
@@ -1588,13 +1613,52 @@ function buildTargetCoreAsync() {
...
+
+                    // Place the base HEX image in the hex cache if necessary
+                    let sha = compileOpts.extinfo.sha;
+                    let hex: string[] = compileOpts.hexinfo.hex;
+                    let hexFile = path.join(hexCachePath, sha + ".hex");
+
+                    if (fs.existsSync(hexFile)) {
+                        pxt.log(`HEX image already in offline cache for template ${dirname}`);
+                    } else {
+                        fs.writeFileSync(hexFile, hex.join(os.EOL));
```
pelikhan: is os.EOL right? Will that create issues on Windows vs Unix?

cli/cli.ts
guillaumejenkins: My idea was to support more offline HEX images than just the default core + radio. With this in place, it's super easy to create more HEX images: just add a new project under `libs/templates/`, and that will result automatically in a 2nd HEX file in the hexcache. If we remove this check, then we can't do that, since any project under `libs/templates/` will end up in target.json.

What do you think?

cli/cli.ts
guillaumejenkins: You are correct, this is unnecessary

cli/server.ts
guillaumejenkins: No, in this file, "built" is a hard-coded string everywhere, see the 2 following functions
https://github.com/Microsoft/pxt/blob/master/cli/server.ts#L53
https://github.com/Microsoft/pxt/blob/master/cli/server.ts#L114

cli/cli.ts
guillaumejenkins: When the HEX info is returned, we are already intelligent about Windows vs Unix line endings, see
https://github.com/Microsoft/pxt/blob/master/pxtlib/cpp.ts#L832

cli/cli.ts
guillaumejenkins: Done

cli/cli.ts
```diff
@@ -1519,30 +1544,32 @@ function updateDefaultProjects(cfg: pxt.TargetBundle) {
...
+        fs.readdirSync(templatePath).forEach((f) => {
+            let file = path.join(templatePath, f);
             if (fs.statSync(file).isDirectory()) {
                 return;
             }
 
             if (f === "pxt.json") {
                 newProject.config = nodeutil.readJson(file);
+                U.iterMap(newProject.config.dependencies, (k, v) => {
+                    if (/^file:/.test(v)) {
```
pelikhan: not needed anymore

cli/cli.ts
```diff
@@ -1312,7 +1333,7 @@ export function buildTargetAsync(): Promise<void> {
         .then(buildTargetCoreAsync)
         .then(buildSemanticUIAsync)
         .then(() => buildFolderAsync('cmds', true))
-        .then(() => buildFolderAsync('server', true))
+        .then(() => buildFolderAsync('server', true));
```
guillaumejenkins: Ah yes this is a leftover from when I had more stuff there

cli/cli.ts
```diff
@@ -1588,13 +1613,52 @@ function buildTargetCoreAsync() {
                 .map(p => path.join("sim", p))
                 .filter(p => fs.statSync(p).isDirectory()));
     }
+
+    let hexCachePath = path.resolve(process.cwd(), "built", "hexcache");
+    nodeutil.deleteFolderRecursive(hexCachePath);
```
guillaumejenkins: Makes sense

PR #579, opened 2016-10-25
Adding mixpanel event for loading projects that use deprecated blocks

### LINKED ISSUE #571: Add tick event for deprecated blocks
See if they ever get used in the editor

riknoll:
Fixes #571 

This event will fire whenever a Blockly project is loaded, including when you switch from TypeScript to Blockly after making a change in the TypeScript. The payload for the event is a map of deprecated block IDs to the number "1" because the API of `pxt.tickEvent` will only take a number or string map. Does `mixpanel.track()` support arrays? If so, I'll change it to just be an array of block IDs.

### DIFF
--- a/pxtlib/main.ts
+++ b/pxtlib/main.ts
@@ -58,7 +58,7 @@ namespace pxt {
     /**
      * Track an event.
      */
-    export var tickEvent: (id: string, data?: Map<string | number> | string[]) => void = function (id) { }
+    export var tickEvent: (id: string, data?: Map<string | number>) => void = function (id) { }
 
     let activityEvents: Map<number> = {};
     const tickActivityDebounced = Util.debounce(() => {
--- a/pxtlib/main.ts
+++ b/pxtlib/main.ts
@@ -51,14 +51,14 @@ namespace pxt {
     }
 
     /**
-     * Time an event by including the time between this call 
+     * Time an event by including the time between this call
      * and a later 'tickEvent' call for the same event in the properties sent with the event.
      */
     export var timeEvent: (id: string) => void = function (id) { }
     /**
      * Track an event.
      */
-    export var tickEvent: (id: string, data?: Map<string | number>) => void = function (id) { }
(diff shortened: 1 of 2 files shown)


pelikhan:
mix panel does not support batched events, at least not in the JS API

riknoll:
@pelikhan but you can send data in the events, right? I just want to know if I can send an array of strings as data.

abchatra:
This looks good to me.  [Mixpanel help](https://mixpanel.com/help/reference/javascript) says track does support the array of data (track is what our tickEvent maps to)

riknoll:
Updated to send an array

pelikhan:
Mixpanel takes properties (https://mixpanel.com/help/reference/javascript-full-api-reference#mixpanel.track). Maybe it just works with arrays, needs to be tested.
```
A set of properties to include with the event you're sending. These describe the user who did the event or details about the event itself.
```

pelikhan:
I've pushed a dummy event from the console to test out how mixpanel handles it.

image

which leads to

image

I think we want to expose the keys and a count of the occurence of each deprecated property to that we can tally the numbers later on.

pelikhan:
Mixpanel allows to pivot using properties. In fact, we could probably send a lot of properties. Maybe we could compute some rough statistics about code snippets: # statements, # loops, # conditionals, # deepest nesting

riknoll:
Updated to instead send a map of deprecated blocks

pelikhan:
Looks good.

pxtlib/main.ts
```diff
@@ -51,14 +51,14 @@ namespace pxt {
...
-     * Time an event by including the time between this call 
+     * Time an event by including the time between this call
      * and a later 'tickEvent' call for the same event in the properties sent with the event.
      */
     export var timeEvent: (id: string) => void = function (id) { }
     /**
      * Track an event.
      */
-    export var tickEvent: (id: string, data?: Map<string | number>) => void = function (id) { }
+    export var tickEvent: (id: string, data?: Map<string | number> | string[]) => void = function (id) { }
```
pelikhan: string[] not supported.

PR #580, opened 2016-10-25
/--run?... URL is now /--run#...

### LINKED ISSUE #535: /--run using '?' not '#'
Repro:
* With the network debugger on, go to https://pxt.microbit.org/about
* Observe it hit https://pxt.microbit.org/--run?nofooter=1&code=basic.fore...

It should be https://pxt.microbit.org/--run#... to let the browser cache it, and to make sure we don't log it.

guillaumejenkins:
Fixes #535


pelikhan:
Does run.html handle this change?

Issue #601, opened 2016-10-27
Duplicate More category appears in localizations


riknoll:
Only happens if the More... string has been translated. Shows up in German, but not Japanese

PR #602, opened 2016-10-27
Fixing duplicate More category caused by over-zealous localization

riknoll:
Fixes https://github.com/Microsoft/pxt/issues/601

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -246,7 +246,7 @@ namespace pxt.blocks {
                     parentCategoryList.appendChild(category);
             }
             if (fn.attributes.advanced) {
-                category = getOrAddSubcategory(category, Util.lf("More\u2026"), 1, category.getAttribute("colour"))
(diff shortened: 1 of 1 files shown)


pxtblocks/blocklyloader.ts
```diff
@@ -246,7 +246,7 @@ namespace pxt.blocks {
                     parentCategoryList.appendChild(category);
             }
             if (fn.attributes.advanced) {
-                category = getOrAddSubcategory(category, Util.lf("More\u2026"), 1, category.getAttribute("colour"))
+                category = getOrAddSubcategory(category, "More\u2026", 1, category.getAttribute("colour"))
```
microsoftsam: So the More category would not be localized?

riknoll: Localization of the categories happens after this initialization step, so they will still get located. I tried it out to make sure.

microsoftsam: Okay, do we need to do the same with the Advanced category name above?

Issue #585, opened 2016-10-27
Blockly "grab" hover icon broken on Edge


microsoftsam:
![img_20161026_192544](https://cloud.githubusercontent.com/assets/16690124/19752460/61abcf38-9bb2-11e6-9aca-e04fdb34a35e.jpg)

PR #604, opened 2016-10-28
Upgrading to the latest version of Blockly

microsoftsam:
- Upgrades to the latest version of Blockly, 10/25/2016 snap 
- Adds an inverted toolbox 
- Adding blockly upgrade instructions and testing
- Fixes bug #585


pelikhan:
We do not use the blockly translations so you can leave those files out.

PR #612, opened 2016-10-28
Generate default projects from target templates

guillaumejenkins:


cli/cli.ts
```diff
@@ -1524,8 +1524,51 @@ function buildSemanticUIAsync() {
...
+            id: p,
+            config: {
+                name: "",
+                dependencies: {},
+                files: []
+            },
+            files: {}
+        };
+
+        if (!fs.existsSync(projectPath)) {
```
pelikhan: move before newProject declaration

PR #631, opened 2016-11-01
Fix not being able to create new variables in blockly.

### LINKED ISSUE #630: Cannot create a new variable in Blockly
There's no way to create a new variable in Blockly right now. 
It looks like Blockly has removed the ability to create a variable from the variable dropdown.
I believe we need a way to get this block in our editor: 
screen shot 2016-10-31 at 5 12 52 pm

microsoftsam:
Blockly removed the option to create a new variable from the variables dropdown. 
Adding option in the Variables menu to create a new variable, like so: 
 
screen shot 2016-10-31 at 5 56 26 pm

fixes #630

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -1426,6 +1426,11 @@ namespace pxt.blocks {
     function initVariables() {
         let varname = lf("{id:var}item");
         Blockly.Variables.flyoutCategory = function (workspace) {
+            let xmlList: HTMLElement[] = [];
+            let button = goog.dom.createDom('button');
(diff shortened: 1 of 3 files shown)


pxtblocks/blocklyloader.ts
```diff
@@ -1426,6 +1426,11 @@ namespace pxt.blocks {
     function initVariables() {
         let varname = lf("{id:var}item");
         Blockly.Variables.flyoutCategory = function (workspace) {
+            let xmlList: HTMLElement[] = [];
+            let button = goog.dom.createDom('button');
+            button.setAttribute('text', (<any>Blockly).Msg.NEW_VARIABLE);
```
pelikhan: NEW_VARIABLE needs to be localized.

PR #644, opened 2016-11-02
UI Changes for inverted theme, and supporting extensibility of CSS in custom.css

microsoftsam:
- Moving some CSS code to pxt.less (so that targets can overwrite it)
- Monaco editor error view zone changes to support inverted theme
- Moving some layout CSS into a pxt.less so that variables can be overwritten in a target to control the layout, eg: menu height, simulator width
- Minor UI fixes to support inverted theme 
- Removing unnecessary borders


theme/pxt.less
```diff
@@ -0,0 +1,259 @@
...
+  position: absolute;
+  top:@mainMenuHeight;
+  bottom: 0rem;
+}
+#sidedocs {
+  display:none;
+  overflow-y: auto;
+  right: 0;
+  border:none;
+  height: calc(100% - 4.5rem);
```
pelikhan: Mainmenuheight?

theme/pxt.less
```diff
@@ -0,0 +1,259 @@
...
+    #maineditor {
+        top:@mobileMenuHeight !important;
+    }
+}
+
+
+/* extra-large desktop screen */
+@media only screen and (min-width: 92em) and (min-height:30em) {
+    .sideDocs #maineditor {
+        right:@sideBarWidth;
```
pelikhan: Nice!

theme/pxt.less
```diff
@@ -0,0 +1,259 @@
...
+    }
+    .rtl .sideDocs #maineditor {
+        left:@sideBarWidth;
+        right:@simulatorWidth;
```
pelikhan: Finally!

theme/pxt.less
microsoftsam: Menubar height.

PR #692, opened 2016-11-04
Fixing the Blockly Monaco resize issue.

### LINKED ISSUE #691: Blockly and Monaco resize issue
Currently onComponentChanged is firing a resize event on the entire window, which is making Blockly reload, causing lags if you are in the middle of editing something. 

Simple repro, make a change that calls the simulator to update, then immediately start editing the text in the showString block, continue typing until the simulator reloads. When that happens, blockly will kick you out of editing that textbox, causing a jarring experience. 

We should be resizing the blockly window, in the same way recommended by Blockly. here: 
https://developers.google.com/blockly/guides/configure/web/resizable

microsoftsam:
No more global resize event firing when the Project component is updated. 

- Blockly resizing itself (as recommended: https://developers.google.com/blockly/guides/configure/web/resizable)
- Monaco using Auto layout

Tested Main and Sandbox view on Chrome and Safari. 
This needs to be tested on IE and Edge.

Fixes #691

### DIFF
--- a/localtypings/blockly.d.ts
+++ b/localtypings/blockly.d.ts
@@ -33,6 +33,7 @@ declare namespace Blockly {
     function genUid(): string;
     function terminateDrag_(): void;
     function mouseToSvg(e: Event, svg: Element): any;
+    function svgResize(workspace: Blockly.Workspace): void;
 
     let ALIGN_RIGHT: number;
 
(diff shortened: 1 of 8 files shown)


microsoftsam:
Things to test: 
- Main editor loads if the last editor used is either blocks or monaco. 
- Resizing the window resizes both blocks and monaco 
- Sandbox mode (same as above)

pelikhan:
Tested on Edge, IE11 and sandbox IE11. Looks good.

Issue #328, opened 2016-09-09
stopAnimation and timing


mmoskal:
Go to https://codethemicrobit.com/about

Click the A or B button on the simulator a couple of times. The text starts to scroll really fast. I don't see why.

pelikhan:
We need to push events on the bus to restart the various waiting fibers

pelikhan:
```
basic.forever(() => {
  basic.showString("Hi!");
})
input.onButtonPressed(Button.A, () => {
    led.stopAnimation();
    basic.showLeds(`
. . . . .
. # . # .
. . . . .
# . . . #
. # # # .`);
});
input.onButtonPressed(Button.B, () => {
    led.stopAnimation();
    basic.showLeds(`
. # . # .
# . # . #
# . . . #
. # . # .
. . # . .`);
});
```

PR #723, opened 2016-11-08
Stopanimation issue

abchatra:
Queued up setTimeout were causing the screen animation timing to haywire. When the AnimationOptions are cancelled all the queued up timeouts needs to be cleared as well. 
This change fixes https://github.com/Microsoft/pxt/issues/328


abchatra:
@pelikhan Can you take a quick peek?

PR #725, opened 2016-11-08
Update pxtCoreDir when targetDir is updated

guillaumejenkins:
Needed for Electron.
Also removed a couple `else` in `serveAsync()`, because some arguments should not be mutually exclusive (i.e. we should be able to specify --cloud and --just-serve)

### DIFF
--- a/cli/nodeutil.ts
+++ b/cli/nodeutil.ts
@@ -17,6 +17,11 @@ export var targetDir: string = process.cwd();
 //When running the Electron app, this will be based on the initial value
 export var pxtCoreDir: string = path.join(targetDir, "node_modules/pxt-core")
 
+export function setTargetDir(dir: string) {
+    targetDir = dir;
+    pxtCoreDir = path.join(targetDir, "node_modules/pxt-core");
+}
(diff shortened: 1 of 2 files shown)


cli/nodeutil.ts
```diff
@@ -17,6 +17,11 @@ export var targetDir: string = process.cwd();
 //When running the Electron app, this will be based on the initial value
 export var pxtCoreDir: string = path.join(targetDir, "node_modules/pxt-core")
 
+export function setTargetDir(dir: string) {
+    targetDir = dir;
+    pxtCoreDir = path.join(targetDir, "node_modules/pxt-core");
```
pelikhan: maybe add some logging here to help diagnostics

OwenBrotherwood: yes please to logging.
this area of code is giving me headaches last night trying to find out what was breaking electron

OwenBrotherwood: @guillaumejenkins any chance of a push before the 14/11 : or else I have to keep working on my fork

PR #732, opened 2016-11-09
Fix for pxtCoreDir path on npm 3+

guillaumejenkins:
Our CLI does not account for NPM 3 vs NPM 2, and assumes that pxt-core is under pxt-microbit/node_modules/pxt-core (which is not always true in NPM 3).

This is a band-aid fix until we can come up with a way for the target to expose its pxt-core dependency, so the CLI won't need to guess where pxt-core is anymore.

### DIFF
--- a/cli/nodeutil.ts
+++ b/cli/nodeutil.ts
@@ -19,7 +19,16 @@ export var pxtCoreDir: string = path.join(targetDir, "node_modules/pxt-core")
 
 export function setTargetDir(dir: string) {
     targetDir = dir;
-    pxtCoreDir = path.join(targetDir, "node_modules/pxt-core");
+    pxtCoreDir = path.join(targetDir, "node_modules", "pxt-core");
+
+    if (!fs.existsSync(pxtCoreDir)) {
+        // Fix for NPM 3: dependencies are flattened, which means pxt-core is not under [target]/node_modules/pxt-core.
+        // In that case, use the pxt-core that nodeutil.js script is running in. It's in pxt-core/built/nodeutil.js, so
+        // go up once to reach root of pxt-core.
+        pxtCoreDir = path.join(__dirname, "..");
+    }
+
+    console.log("Set pxtCoreDir to:" + pxtCoreDir);
 }
 
 export function readResAsync(g: events.EventEmitter) {
--- a/cli/nodeutil.ts
+++ b/cli/nodeutil.ts
@@ -19,16 +19,24 @@ export var pxtCoreDir: string = path.join(targetDir, "node_modules/pxt-core")
 
 export function setTargetDir(dir: string) {
     targetDir = dir;
-    pxtCoreDir = path.join(targetDir, "node_modules", "pxt-core");
+    let newPxtCoreDir = path.join(targetDir, "node_modules", "pxt-core");
 
-    if (!fs.existsSync(pxtCoreDir)) {
+    if (!fs.existsSync(newPxtCoreDir)) {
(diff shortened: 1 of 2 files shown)


cli/nodeutil.ts
```diff
@@ -19,7 +19,16 @@ export var pxtCoreDir: string = path.join(targetDir, "node_modules/pxt-core")
...
 export function setTargetDir(dir: string) {
     targetDir = dir;
-    pxtCoreDir = path.join(targetDir, "node_modules/pxt-core");
+    pxtCoreDir = path.join(targetDir, "node_modules", "pxt-core");
+
+    if (!fs.existsSync(pxtCoreDir)) {
+        // Fix for NPM 3: dependencies are flattened, which means pxt-core is not under [target]/node_modules/pxt-core.
+        // In that case, use the pxt-core that nodeutil.js script is running in. It's in pxt-core/built/nodeutil.js, so
+        // go up once to reach root of pxt-core.
+        pxtCoreDir = path.join(__dirname, "..");
```
pelikhan: Should you test that pxtCoreDir is pointing at the proper directory?

cli/nodeutil.ts
```diff
@@ -19,7 +19,16 @@ export var pxtCoreDir: string = path.join(targetDir, "node_modules/pxt-core")
...
+    }
+
+    console.log("Set pxtCoreDir to:" + pxtCoreDir);
```
pelikhan: Do we use a pxt.log?

OwenBrotherwood:
For want of a better place to post this comment.
packages , like electron package, should probably not have direct dependency on pxt-core but have a dependency on pxt-microbit if only using pxt-microbit and nothing from pxt-core (or you get double pxt-cores)
The node_modules structure with the npm v3 I use is therefore. with no dependancy on pxt-core from electron:
```
C:\Users\owen\Source\Repos\pxt-microbit-electron>rmdir /s node_modules\pxt-core
node_modules\pxt-core, Are you sure (Y/N)? y
C:\Users\owen\Source\Repos\pxt-microbit-electron>rmdir /s node_modules\pxt-microbit
node_modules\pxt-microbit, Are you sure (Y/N)? y
C:\Users\owen\Source\Repos\pxt-microbit-electron>npm --version
3.10.8

C:\Users\owen\Source\Repos\pxt-microbit-electron>npm install
code-the-microbit@1.0.0 C:\Users\owen\Source\Repos\pxt-microbit-electron
`-- pxt-microbit@0.5.35
  `-- pxt-core@0.5.51
```

cli/nodeutil.ts
guillaumejenkins: Wasn't sure about the best way to test that. Ended up looking for package.json and making sure "name" is "pxt-core".

cli/nodeutil.ts
guillaumejenkins: changed it

OwenBrotherwood:
I cannot confirm that the pull works. 
In fact, it made the situation worse as I now have "Not found" when using npm start in the pxt-microbit-electron cloned directory.
As I had to npm link the pxt and pxt-microbit, there could be side effects there.
node_modules checked for version numbers so that I know I have the versions available in the git repos for the two packages.
All work based on the Microsoft repos.

OwenBrotherwood:
Please note, that the fail is for npm start, not package.
All work based on the current, as of yesterday, MS repos (as stated).
I will check this evening from todays commits as to if there is progress.

PR #736, opened 2016-11-09
A number of patches as I find them

OwenBrotherwood:
consistent path join and add semi-colon

### DIFF
--- a/cli/nodeutil.ts
+++ b/cli/nodeutil.ts
@@ -18,8 +18,7 @@ export var targetDir: string = process.cwd();
 export var pxtCoreDir: string = path.join(targetDir, "node_modules", "pxt-core");
 
 export function setTargetDir(dir: string) {
-    targetDir = dir;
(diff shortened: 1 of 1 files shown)


cli/nodeutil.ts
```diff
@@ -15,11 +15,10 @@ import Util = pxt.Util;
 //When running inside Electron it gets updated to the correct path
 export var targetDir: string = process.cwd();
 //When running the Electron app, this will be based on the initial value
-export var pxtCoreDir: string = path.join(targetDir, "node_modules/pxt-core")
+export var pxtCoreDir: string = path.join(targetDir, "node_modules", "pxt-core");
 
 export function setTargetDir(dir: string) {
-    targetDir = dir;
```
pelikhan: why are you removing this line?

OwenBrotherwood: Does the line not overwrite targetDir before check and if check is ok overwrites again?

PR #737, opened 2016-11-10
Use target's exposed pxtCoreDir

guillaumejenkins:


cli/nodeutil.ts
```diff
@@ -19,22 +19,14 @@ export var pxtCoreDir: string = path.join(targetDir, "node_modules/pxt-core")
...
+    // The target should expose the path to its bundled pxt-core
+    try {
+        pxtCoreDir = require(targetDir).pxtCoreDir();
     }
-
-    if (!!newPxtCoreDir && newPxtCoreDir !== pxtCoreDir) {
-        pxtCoreDir = newPxtCoreDir;
+    catch (e) {
+        // If not, fallback to default location
+        pxtCoreDir = path.join(targetDir, "node_modules", "pxt-core");
```
pelikhan: Add logging

cli/nodeutil.ts
```diff
@@ -19,22 +19,14 @@ export var pxtCoreDir: string = path.join(targetDir, "node_modules/pxt-core")
...
-        // go up once to reach root of pxt-core.
-        newPxtCoreDir = path.join(__dirname, "..");
-
-        let packageJsonFile = path.join(newPxtCoreDir, "package.json");
-        if (!fs.existsSync(packageJsonFile) || JSON.parse(fs.readFileSync(packageJsonFile, "utf8")).name !== "pxt-core") {
-            newPxtCoreDir = null;
-        }
+    // The target should expose the path to its bundled pxt-core
+    try {
+        pxtCoreDir = require(targetDir).pxtCoreDir();
```
pelikhan: Why should this fail?

pelikhan: You should test if pxtCoreDir is defined before calling it.

pelikhan:
maincli should return the package instead of the directory. Let's minimize the code needed on the targetside. Also, we already have a file build in the target (cmd.js), can we merge both?

Issue #711, opened 2016-11-07
Blocks/JavaScript button missing when opening project file


pelikhan:
Go to explorer and click on pxt.json, blocks/javascript button is gone.

![image](https://cloud.githubusercontent.com/assets/4175913/20069405/42d7763e-a4d1-11e6-99db-f4dcaff553e9.png)

PR #738, opened 2016-11-10
Back to Code button appears in all the files

abchatra:
Simple fix to address https://github.com/Microsoft/pxt/issues/711

### DIFF
--- a/webapp/src/monaco.tsx
+++ b/webapp/src/monaco.tsx
@@ -143,10 +143,15 @@ export class Editor extends srceditor.Editor {
     }
 
     menu(): JSX.Element {
-        if (!this.hasBlocks()) return null
-        return <sui.Item class="blocks-menuitem" textClass="landscape only" text={lf("Blocks") } icon="puzzle" onClick={() => this.openBlocks() }
-                tooltip={lf("Convert code to Blocks")} tooltipPosition="bottom left"
-         />
+        let editor = pkg.mainEditorPkg(); 
+        if (this.currFile != editor.files["main.ts"]) {
+            return (<sui.Item text={lf("Back to Code") } icon={"align left"} onClick={() => this.parent.setFile(editor.files["main.ts"]) } />);
+        }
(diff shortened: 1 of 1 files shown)


webapp/src/monaco.tsx
```diff
@@ -143,10 +143,15 @@ export class Editor extends srceditor.Editor {
...
     menu(): JSX.Element {
-        if (!this.hasBlocks()) return null
-        return <sui.Item class="blocks-menuitem" textClass="landscape only" text={lf("Blocks") } icon="puzzle" onClick={() => this.openBlocks() }
-                tooltip={lf("Convert code to Blocks")} tooltipPosition="bottom left"
-         />
+        let editor = pkg.mainEditorPkg(); 
+        if (this.currFile != editor.files["main.ts"]) {
+            return (<sui.Item text={lf("Back to Code") } icon={"align left"} onClick={() => this.parent.setFile(editor.files["main.ts"]) } />);
+        }
+        else if (editor.files["main.blocks"]) { //if main.blocks file present
```
pelikhan: You should go back to blocks by default.

webapp/src/monaco.tsx
```diff
@@ -143,10 +143,15 @@ export class Editor extends srceditor.Editor {
...
+            return ( <sui.Item class="blocks-menuitem" textClass="landscape only" text={lf("Blocks") } icon="puzzle" onClick={() => this.openBlocks() }
```
pelikhan: Why not call it "Back to Code"?

abchatra: Not sure I get this comment. It is called "Back to Code". Else part deals with case where you are in typescript view and now want to go back to Blocks, if there is a main.blocks file.

webapp/src/monaco.tsx
pelikhan: Good point!

webapp/src/monaco.tsx
pelikhan: Looks good

PR #743, opened 2016-11-10
Gallery

pelikhan:
This change integrate the project gallery into "Open Project..." dialog.

### DIFF
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -145,6 +145,7 @@ declare namespace pxt {
         blocklyOptions?: Blockly.Options;
         simAnimationEnter?: string;
         simAnimationExit?: string;
+        projectGallery?: string;
     }
 
     interface DocMenuEntry {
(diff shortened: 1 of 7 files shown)


abchatra:
@pelikhan  Can I understand the motivation for this change? @Jaqster - FYI

pelikhan:
We had a lengthy discussion about fixing the getting started page and also about the project page being not adequate. The current rendering of the project list in the side docs does not work well and it is somehow hidden under the ? menu.
This is a little tweak that exposes the cards rendered in the project in the “Open Project…”.


-        It makes those projects easier to discover. “Open Project” is a top level menu item.

-        Clicking on a project tile, starts a new project (with the right name) and opens up the docs. (We have analytics for that too).

-        The project list is still available under /projects which is nice – can be referenced outside the app.

-        We also used a lot of templates in Touch Developer, students liked to go through them. Also similar to Arduino IDE that has a ton of annotated examples to start from.

Still adjusting. I’ll consult next time.

PR #753, opened 2016-11-11
out of date node version check

OwenBrotherwood:
the check should possibly be done elsewhere as to many checks create a problem with maintenance

### DIFF
--- a/pxt-cli/cli.js
+++ b/pxt-cli/cli.js
@@ -1,10 +1,5 @@
 "use strict";
 
-if (/^v[0-3]\./.test(process.version)) {
-    console.error("Please upgrade your node.js to at least v4.x.x.")
-    process.exit(1)
-}
-
 var fs = require("fs")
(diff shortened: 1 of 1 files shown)


pxt-cli/cli.js
```diff
@@ -1,10 +1,5 @@
 "use strict";
 
-if (/^v[0-3]\./.test(process.version)) {
-    console.error("Please upgrade your node.js to at least v4.x.x.")
-    process.exit(1)
-}
-
 var fs = require("fs")
```
OwenBrotherwood: These var's should probably be constants as they are not changed and should not be changed.

pelikhan: You can declare minimum version of node and npm in the package.json.

OwenBrotherwood:
engines min node/npm is already defined in package.json so the check is no longer needed.
added the const instead of var for return of require.
Good to go?

PR #760, opened 2016-11-11
Adding getting started button

microsoftsam:
Adding getting started button for the widescreen view (in the top right corner). See screenshot:
screen shot 2016-11-11 at 1 33 24 pm

+ Ui layout cleanups
+ Fix for Explorer / Sim buttons showing up before the editor has fully loaded.

The getting started button disappears after the user has loaded any content in the side docs.

### DIFF
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -146,6 +146,7 @@ declare namespace pxt {
         simAnimationEnter?: string;
         simAnimationExit?: string;
         projectGallery?: string;
+        gettingStarted?: string;
     }
 
     interface DocMenuEntry {
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -124,12 +124,12 @@ declare namespace pxt {
         embedUrl?: string;
         docMenu?: DocMenuEntry[];
         hideSideDocs?: boolean;
+        sideDoc?: string; // if set: show the getting started button, clicking on getting started button links to that page
         boardName?: string;
         privacyUrl?: string;
         termsOfUseUrl?: string;
         contactUrl?: string;
         accentColor?: string;
-        invertedMenu?: boolean;
         locales?: Map<AppTheme>;
         cardLogo?: string;
         appLogo?: string;
@@ -140,13 +140,13 @@ declare namespace pxt {
         usbDocs?: string;
         exportVsCode?: boolean;
         browserSupport?: SpecializedResource[];
-        invertedToolbox?: boolean;
-        invertedMonaco?: boolean;
-        blocklyOptions?: Blockly.Options;
-        simAnimationEnter?: string;
-        simAnimationExit?: string;
+        invertedMenu?: boolean; // if true: apply the inverted class to the menu
+        invertedToolbox?: boolean; // if true: use the blockly inverted toolbox
+        invertedMonaco?: boolean; // if true: use the vs-dark monaco theme
+        blocklyOptions?: Blockly.Options; // Blockly options, see Configuration: https://developers.google.com/blockly/guides/get-started/web
+        simAnimationEnter?: string; // Simulator enter animation
+        simAnimationExit?: string; // Simulator exit animation
         projectGallery?: string;
-        gettingStarted?: string;
     }
 
     interface DocMenuEntry {
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1668,8 +1668,8 @@ export class ProjectView extends data.Component<IAppProps, IAppState> {
     gettingStarted() {
         pxt.tickEvent("btn.gettingstarted");
         const targetTheme = pxt.appTarget.appTheme;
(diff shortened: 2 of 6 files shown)


webapp/src/app.tsx
```diff
@@ -1665,6 +1665,14 @@ export class ProjectView extends data.Component<IAppProps, IAppState> {
         this.shareEditor.show(header);
     }
 
+    gettingStarted() {
+        Util.assert(!this.state.sideDocsLoadUrl);
+        const targetTheme = pxt.appTarget.appTheme;
+        const gettingStartedUrl = targetTheme.gettingStarted || 'getting-started';
```
pelikhan: not guaranteed that a getting-started page is available.

webapp/src/app.tsx
```diff
@@ -1754,6 +1762,11 @@ export class ProjectView extends data.Component<IAppProps, IAppState> {
                         </div> : undefined }
                     </div>
                 </div>
+                {this.state.sideDocsLoadUrl && targetTheme.gettingStarted ? undefined : 
+                    <div id="getting-started-btn">
+                        <div className="ui bottom attached button green" onClick={() => this.gettingStarted() }>Getting Started</div>
```
pelikhan: lf("Getting started")

pelikhan: should we use "primary" color?

webapp/src/app.tsx
```diff
@@ -1754,6 +1762,11 @@ export class ProjectView extends data.Component<IAppProps, IAppState> {
                         </div> : undefined }
                     </div>
                 </div>
+                {this.state.sideDocsLoadUrl && targetTheme.gettingStarted ? undefined :
```
pelikhan: if targetTheme is not defined, this should not be visible.

webapp/src/app.tsx
```diff
@@ -1665,6 +1665,14 @@ export class ProjectView extends data.Component<IAppProps, IAppState> {
         this.shareEditor.show(header);
     }
 
+    gettingStarted() {
+        pxt.tickEvent("btn.gettingstarted");
+        const targetTheme = pxt.appTarget.appTheme;
+        Util.assert(!this.state.sideDocsLoadUrl && targetTheme && targetTheme.gettingStarted != undefined);
```
pelikhan: or simply

```
Util.assert(!this.state.sideDocsLoadUrl && targetTheme && targetTheme.gettingStarted);
```

webapp/src/app.tsx
```diff
@@ -1754,6 +1762,11 @@ export class ProjectView extends data.Component<IAppProps, IAppState> {
                         </div> : undefined }
                     </div>
                 </div>
+                {!this.state.sideDocsLoadUrl && targetTheme && targetTheme.gettingStarted != undefined ?
```
pelikhan:
```
{!this.state.sideDocsLoadUrl && targetTheme && targetTheme.gettingStarted
```

webapp/src/app.tsx
```diff
@@ -1754,6 +1762,11 @@ export class ProjectView extends data.Component<IAppProps, IAppState> {
                         </div> : undefined }
                     </div>
                 </div>
+                {!this.state.sideDocsLoadUrl && targetTheme && targetTheme.gettingStarted != undefined ? 
+                    <div id="getting-started-btn">
+                        <sui.Button class="bottom attached green" title={lf("Getting Started")} text={lf("Getting Started")} onClick={() => this.gettingStarted() } />
```
pelikhan: button text and title should not be the same. Title can be more expressive and use the semantic UI css (see tooltip property)

localtypings/pxtarget.d.ts
```diff
@@ -146,6 +146,7 @@ declare namespace pxt {
         simAnimationEnter?: string;
         simAnimationExit?: string;
         projectGallery?: string;
+        gettingStarted?: string;
```
pelikhan: There is a "sideDocs" property already. Don't we want to reuse or remove that one?

microsoftsam: I thought the sideDocs property enabled/disabled showing content in the side docs. 
My thoughts on the gettingStarted property is to be a flag to enable / disable whether or not the button shows up.

webapp/src/app.tsx
microsoftsam: Without the != undefined, I get a type error: 
Argument of type 'string' is not assignable to parameter of type 'boolean'

I'd say it has something to do with the assert parameter type

pelikhan: Try !! Operator

localtypings/pxtarget.d.ts
pelikhan: It was meant exactly for that :) there is hideSideDocs to remove sidedocs.

pelikhan: Need to document schema.

microsoftsam: Agree with documenting schema. I'll start with the ones I know.

microsoftsam: Using sideDoc instead of gettingStarted.

PR #774, opened 2016-11-15
move "new project" tile under "open project"

pelikhan:
Add tile in "Open Project" dialog instead of menu item under hamburger.


abchatra:
👍 @Jaqster to take a look

pelikhan:
@Jaqster is ok.

Issue #744, opened 2016-11-10
shared loop variable not reinitialized properly


pelikhan:
Here is a repo. The assert triggers
```
let k = 0;
for (var i = 0; i < 1; ++i) {
    k = i;
}
for (var i = 0; i < 1; ++i) {
    k = i + 5;
}
basic.showNumber(k);
control.assert(k == 5);
```

abchatra:
Seems like the initialization in the second loop is peeped out. I will take a look. Works fine with 'let' instead of var though as it introduces a separate scoped variable for each let instance.

mmoskal:
The problem is that all `i` share node-id except for the second `var` which is different. If TS introduces a new node id for the second `var` they should also resolve the following references to it.  Maybe this is fixed in more recent TypeScript. Should we try to merge 2.1?

abchatra:
In JavaScript var is not block scoped and both var refer to the same var. Redeclaration is no-op.

mmoskal:
Sure. Which is why all the node-ids should be the same.

abchatra:
Yes makes sense.. Though I do want to remove 'var' and just allow 'let' for language semantics perspective.

pelikhan:
As long as we have a nice error message saying "change 'var' to 'let'"

pelikhan:
https://github.com/Microsoft/pxt/pull/780

PR #780, opened 2016-11-16
Disabling var in the typescript editor

abchatra:
We don't fully adhere to javascript var semantics. This change disables var as the way to define a variable and will throw an error requiring let or const.

Motivated by https://github.com/Microsoft/pxt/issues/744
The real issue in #744 is typescript binds to one global variable, but we implement let semantics for var which contradicts.


pxtlib/emitter/emitter.ts
```diff
@@ -2525,9 +2525,16 @@ ${lbl}: .short 0xffff
         function emitBlock(node: Block) {
             node.statements.forEach(emit)
         }
+        function checkForLetOrConst(declList: VariableDeclarationList): boolean  {
+            if ((declList.flags & NodeFlags.Let) || (declList.flags & NodeFlags.Const)) {
+                return true;
+            }
+            throw userError(9203, lf("variable needs to be defined using 'let' or 'const'"));
```
riknoll: I think we are already using the 9203 error code elsewhere.

pelikhan:
We should automatically fix these issues by using "let" and notify the users that we did so. 

* A learner won't be able to really understand why "var" is not valid 
* we are giving a choice (let vs const) which is also confusing. Some students won't be able to decide between both.

abchatra:
I am going to experiment with first one and see how it goes. I will update the PR after that
Agree with 2nd point. I am going to update with just a choice for let.

pelikhan:
What abou issuing an error saying "use let instead of var"?

abchatra:
Yes will fix this asap.

abchatra:
Going to reopen an another one. This one is stale

PR #784, opened 2016-11-16
report tick instead of crash for user error

pelikhan:
Reporting "error" popup as a tick -- as the error should have been handled upstream. Also sending message as property instead of string to avoid incorrect hashing.


abchatra:
👍

pelikhan:
I've dug deeper into other error reports and found more issues. Complete review of the reportError/reportException usage and updates.

pelikhan:
I'm done with this one. This PR changes various bugs in our crash reporting.

PR #785, opened 2016-11-16
Avoiding LZMA compression if the compilation is for simulator

abchatra:
Richard noticed LZMA compression algorithm kicking in and consuming lot of CPU cycles in Chromebook. We don't need to do compression and generate embedblob for the simulator.

### DIFF
--- a/pxtlib/main.ts
+++ b/pxtlib/main.ts
@@ -524,7 +524,7 @@ namespace pxt {
                 })
                 .then(() => this.config.binaryonly || appTarget.compile.shortPointers ? null : this.filesToBePublishedAsync(true))
                 .then(files => {
-                    if (files) {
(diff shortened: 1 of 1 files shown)


abchatra:
Also improvement in 37% in CPU time on edge for blockly edits.

pxtlib/main.ts
```diff
@@ -524,7 +524,7 @@ namespace pxt {
                 })
                 .then(() => this.config.binaryonly || appTarget.compile.shortPointers ? null : this.filesToBePublishedAsync(true))
                 .then(files => {
-                    if (files) {
+                    if (files && opts.target.isNative) {
```
abchatra: Makes sense. Did that and pushed.

PR #792, opened 2016-11-17
Add local server support for external messaging

guillaumejenkins:
Needed by Electron to communicate with the WebApp. The built-in Electron messaging channels are a security issue for us, because they would let the web app require Node core modules such as fs.

### DIFF
--- a/cli/server.ts
+++ b/cli/server.ts
@@ -23,6 +23,7 @@ let userProjectsDir = path.join(process.cwd(), userProjectsDirName);
 let docsDir = ""
 let packagedDir = ""
 let localHexDir = path.join("built", "hexcache");
+let externalMessageHandlers: ExternalMessageHandlerDictionary = {};
 
 export function forkPref() {
     if (pxt.appTarget.forkof)
@@ -305,6 +306,31 @@ function handleApiAsync(req: http.IncomingMessage, res: http.ServerResponse, elt
 
                 return res;
             });
+    else if (cmd == "POST externalmsg") {
+        return readJsonAsync()
+            .then((data: ExternalMessageData) => {
+                if (!data || !data.messageType) {
+                    throw throwError(400);
+                }
+
+                let resultDeferred = Promise.defer<ExternalMessageResult>();
+
+                if (!externalMessageHandlers[data.messageType]) {
+                    resultDeferred.resolve({
+                        error: "noHandler"
+                    });
+                } else {
+                    externalMessageHandlers[data.messageType]((res) => {
+                        resultDeferred.resolve({
+                            error: res.error,
+                            result: res.result
+                        });
+                    }, data.args);
+                }
+
+                return resultDeferred.promise;
+            });
+    }
     else throw throwError(400)
 }
 
@@ -642,12 +668,25 @@ function getBrowserLocation(browser: string) {
     return browser;
 }
 
+export interface ExternalMessageData {
+    messageType: string;
+    args: any
+}
+export interface ExternalMessageResult {
+    error?: any;
+    result?: any;
+}
+export interface ExternalCallback { (res: ExternalMessageResult): void }
+export interface ExternalMessageHandler { (cb: ExternalCallback, data?: ExternalMessageData): void }
+export interface ExternalMessageHandlerDictionary { [messageType: string]: ExternalMessageHandler }
+
 export interface ServeOptions {
     localToken: string;
     autoStart: boolean;
     packaged?: boolean;
     electron?: boolean;
     browser?: string;
+    externalHandlers?: ExternalMessageHandlerDictionary;
 }
 
 let serveOptions: ServeOptions;
@@ -656,6 +695,10 @@ export function serveAsync(options: ServeOptions) {
     setupRootDir();
     initTargetCommands();
     initSerialMonitor();
+
+    if (serveOptions.externalHandlers) {
+        externalMessageHandlers = serveOptions.externalHandlers;
+    }
 
     let server = http.createServer((req, res) => {
         let error = (code: number, msg: string = null) => {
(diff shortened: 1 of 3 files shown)


pelikhan:
Why aren't we checking for update in the web app directly?

pelikhan:
Ok

cli/server.ts
```diff
@@ -642,12 +668,25 @@ function getBrowserLocation(browser: string) {
...
+    messageType: string;
+    args: any
+}
+export interface ExternalMessageResult {
+    error?: any;
+    result?: any;
+}
+export interface ExternalCallback { (res: ExternalMessageResult): void }
+export interface ExternalMessageHandler { (cb: ExternalCallback, data?: ExternalMessageData): void }
+export interface ExternalMessageHandlerDictionary { [messageType: string]: ExternalMessageHandler }
```
pelikhan: use ``pxt.Map<ExternalMessageHandler>``

webapp/src/app.tsx
```diff
@@ -67,7 +67,16 @@ interface IAppState {
...
+interface ExternalMessageData {
+    messageType: string;
+    args: any
+}
+interface ExternalMessageResult {
+    error?: any;
+    result?: any;
+}
 
+let isElectron = /[?&]electron=1/.test(window.location.href);
```
OwenBrotherwood: Please consider a more generic type until a specific need for the label isElectron.
For example notBrowser to indicate another type of usage.
Perhaps coupled with a --not-browser option instead of a --electron

pelikhan: This flag is used in line 1798 for Electron specific updates

OwenBrotherwood: The comment concerning the usage of flags that include the name of a specific implementation of non-browser solution of presenting pxt(-microbit).

- It is conceivable that names of specific implementations can change leaving historical names in the code: been there, refactored it.
- It causes a specific implementation focus in code that should not be concerned with what is using it's services: one begins to forget that other Open Source users of the code would like to avoid explaining that it is not, for example, electron that is being used.

Please consider using  a type, example nonBrowser, as the facility can be useful to other implementations, be they electron or other specific implementations of encapsulating the pxt(-microbit)

OwenBrotherwood:
pelican wrote:

> Why aren't we checking for update in the web app directly?

I have taken a second look at the question.

The non-browser app has it's own update process for own code change, ie not dependant on dependencies?
-  "other updates" are dictated by dependency versions that are probably not "latest"
- any change in the non-browser app code, also dependency version changes, would dictate a version change, if only in the very minor version of  version X.Y.Z
- the non-browser app would then have to have an update request to the user.

Given the above, that dependencies are not "lattest", why does the non-browser app require access to the pxt(-microbit) web page to ask for update thereby blotting the pxt(-microbit) code with functions that are possible outside of pxt(-microbit)?

I will re.read all the above comments/PR and will come with bowed head if I have overseen something in my analyasis

webapp/src/app.tsx
OwenBrotherwood: A menu item, specific for an implementation of non-browser, would be undesirable if the menu is called electron with no means of giving the name of the requester in the general case of non-browser

pelikhan:
It is useful for the web app to know that it is hosted in a native app. Electron has implementation specific issue, we might address the naming in future PRs if we support different hosts.

OwenBrotherwood:
guillaumejenkins  wrote
> The built-in Electron messaging channels are a security issue for us, because they would let the web app require Node core modules such as fs.

The two cases of a modal window are:
- file based html: possible security issue due to fs access?
- web based html: requires on-line

As the non-browser app has to be on-line in order to flag update needed/available, the case for using file based html is "limited".

There is perhaps a third method in creating an update needed/available, and this is not actually enabling the webview and display a suitable html that fits the use case.
https://github.com/Microsoft/pxt-microbit-electron/blob/40710af55f26efb5f7aac574deffa95a0bef77d0/src/index.html#L10

As the index.html that contains the webview, is on the file system, I do not understand what security issue is present with fs.

OwenBrotherwood:
pelikan wrote
> It is useful for the web app to know that it is hosted in a native app. 
Yes: in the case of the cookies accept, I totally agree that the pxt(-microbit) has to know the environment as to which policies to enforce
> Electron has implementation specific issue
Maybe :)
> we might address the naming in future PRs if we support different hosts.
As always, the master has the last call :)

OwenBrotherwood:
My take on implementing an 
- Implement some upToDate() when connection to the update server is possible (on-line scenario)
 - https://github.com/OwenBrotherwood/microbit-drill/blob/6505e239f147cd13979133f185cf1dbd6d6d4e68/src/microbit_drill/microbit_drill.js#L50 
- Some appalling html
  - https://github.com/OwenBrotherwood/microbit-drill/blob/6505e239f147cd13979133f185cf1dbd6d6d4e68/app/app.html#L22-L25 Some appalling html
- No new dependancies needed.
 - html file to use localhost is already on file system

Note:
- App start
 - https://github.com/OwenBrotherwood/microbit-drill/blob/6505e239f147cd13979133f185cf1dbd6d6d4e68/src/app.js#L11-L13

cli/server.ts
guillaumejenkins: Will do

OwenBrotherwood:
Thanks guillaumejenkins 
As to if there is a need to communicate with pxt(-microbit) to achieve an update process: I leave this to the master of the git :) 
I will probably not use the possibility, as I do not see that I need it "just now" for any update process.
But please, make sure isElectron is not overloaded for other non-browser cases.

As for VS Code, I once had a crazy though of pxt(-microbit) as an extension (and I still feel the itch to investigate if this is possible).
I will examine the update mechanism in VS Code to see if I can also use the same code (I am lazy).

pelikhan:
LGTM

PR #794, opened 2016-11-18
Fixing modal dimmer jquery selector

### LINKED ISSUE #793: Semantic modal clashes with Blockly comment box
If you open a comment on a block and then press download, the screen will dim and freeze.

For some reason the dimmer in the Semantic UI modal is accidentally trying to dim the Blockly comment box along with the modal dimmer. Ends up causing an exception and freezing the site.

riknoll:
Fixes #793 

By default, semantic just uses a selector that gets all `<body />` elements when it tries to dim the screen for a modal (it assumes there is only one in the page). Blockly comments use `<body />` elements for their bubbles, so when a bubble is present Semantic throws an exception. This fix just overrides that selector.

### DIFF
--- a/webapp/src/core.ts
+++ b/webapp/src/core.ts
@@ -165,7 +165,7 @@ export function dialogAsync(options: DialogOptions): Promise<void> {
     let html = `
   <div class="ui ${options.size || "small"} modal">
     <div class="header">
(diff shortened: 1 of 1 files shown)


pelikhan:
Tested, looks go.

PR #797, opened 2016-11-21
Publish anonymous gist from command line

microsoftsam:
Adding feature to publish a project to an anonymous gist.
To test: 
```
mkdir testproject; cd testproject
pxt init
pxt gist
```

Command line output: 
Publishing project to gist anonymously.
Uploading....
Success, your gist url is https://gist.github.com/527cb76153bb38b5c86aac5697619a6e
You can load your published project at https://pxt.microbit.org/#pub:gh/gists/527cb76153bb38b5c86aac5697619a6e

Also fixed issue with package install, see github.ts

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -3425,15 +3425,12 @@ function checkDocsAsync(...args: string[]): Promise<void> {
 
 function publishGistCoreAsync(): Promise<void> {
     console.log("Publishing project to gist anonymously.");
-    return Promise.resolve()
-        .then(() => {
-            return mainPkg.loadAsync()
-        })
+    return mainPkg.loadAsync()
         .then(() => {
             console.log("Uploading....")
             let files: string[] = mainPkg.getFiles()
             let pxtConfig = mainPkg.config;
-            let filesMap: Map<{}> = {};
+            let filesMap: Map<{content: string;}> = {};
 
             files.forEach((fn) => {
                 let fileContent = fs.readFileSync(fn, "utf8");
@@ -3442,7 +3439,7 @@ function publishGistCoreAsync(): Promise<void> {
                         "content": fileContent
                     }
                 } else {
-                    // Cannot publish empty files, go through and remove empty file references form pxt.json
+                    // Cannot publish empty files, go through and remove empty file references from pxt.json
                     if (pxtConfig.files && pxtConfig.files.indexOf(fn) > -1) {
                         pxtConfig.files.splice(pxtConfig.files.indexOf(fn), 1);
                     } else if (pxtConfig.testFiles && pxtConfig.testFiles.indexOf(fn) > -1) {
@@ -3452,7 +3449,7 @@ function publishGistCoreAsync(): Promise<void> {
             })
             // Add pxt.json
             filesMap['pxt.json'] = {
-                "content": JSON.stringify(pxtConfig)
+                "content": JSON.stringify(pxtConfig, null, 4)
             }
             return pxt.github.publishGist(filesMap, pxtConfig.description, false)
         })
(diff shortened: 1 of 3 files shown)


cli/cli.ts
```diff
@@ -3423,6 +3423,52 @@ function checkDocsAsync(...args: string[]): Promise<void> {
     return Promise.resolve();
 }
 
+function publishGistCoreAsync(): Promise<void> {
+    console.log("Publishing project to gist anonymously.");
+    return Promise.resolve()
```
pelikhan: Do we need this line?

cli/cli.ts
```diff
@@ -3423,6 +3423,52 @@ function checkDocsAsync(...args: string[]): Promise<void> {
...
+        .then(() => {
+            return mainPkg.loadAsync()
+        })
+        .then(() => {
+            console.log("Uploading....")
+            let files: string[] = mainPkg.getFiles()
+            let pxtConfig = mainPkg.config;
+            let filesMap: Map<{}> = {};
```
pelikhan: ... ``Map<{ content:string;}>ˋ`

cli/cli.ts
```diff
@@ -3423,6 +3423,52 @@ function checkDocsAsync(...args: string[]): Promise<void> {
...
+
+            files.forEach((fn) => {
+                let fileContent = fs.readFileSync(fn, "utf8");
+                if (fileContent) {
+                    filesMap[fn] = {
+                        "content": fileContent
+                    }
+                } else {
+                    // Cannot publish empty files, go through and remove empty file references form pxt.json
```
pelikhan: form -> from

cli/cli.ts
```diff
@@ -3423,6 +3423,52 @@ function checkDocsAsync(...args: string[]): Promise<void> {
...
+                    if (pxtConfig.files && pxtConfig.files.indexOf(fn) > -1) {
+                        pxtConfig.files.splice(pxtConfig.files.indexOf(fn), 1);
+                    } else if (pxtConfig.testFiles && pxtConfig.testFiles.indexOf(fn) > -1) {
+                        pxtConfig.testFiles.splice(pxtConfig.testFiles.indexOf(fn), 1);
+                    }
+                }
+            })
+            // Add pxt.json
+            filesMap['pxt.json'] = {
+                "content": JSON.stringify(pxtConfig)
```
pelikhan: Make it 4 spaces to make it more readable

pelikhan:
Could we support non-anonymous gists and updates?

cli/cli.ts
microsoftsam: Nope, mainPkg.loadAsync returns our promise.

cli/cli.ts
microsoftsam: Sure.

microsoftsam:
Yes, but we might want to consider pulling in a dependency on a tool that already does this. 
For example: https://github.com/defunkt/gist (MIT).

Reason for this, is we'd need to deal with github login and two factor authentication.

pelikhan:
We should also support the scenario of a signed in user and allowing to patch an existing gist.

pelikhan:
We don't need to deal with oauth if the use provides a token (which can be generated from the GitHub UI). Seems that this tool is written in ruby so we wouldn't want to take on that dependency.

microsoftsam:
Okay, so we should
- use the git auth token if it's there and pass it with the request
- if there's not git auth token, ask the user for their username and password (just like any private git command). 

It just seems like we're reinventing the wheel a little here, I'll keep digging. 
In the meantime, looks like this PR is good to go.

pelikhan:
Let's not ask the user/password (with 2FA things are more complicated). Uses gives a token a nothing.

Need to store gist id in PXT.json to be able to update it.

pelikhan:
PR good to go

microsoftsam:
Great. Thanks!

PR #801, opened 2016-11-23
Publish a gist under a user account using a github TOKEN.

microsoftsam:
Publish a gist under a user using a github TOKEN.
Saving current gist in pxt.json
cli options to force create a new gist, and publishing a public gist (default to secret).

**Create an anonymous gist**
```
pxt gist
```

**Create an gist under a user account**
```
pxt gist --token TOKEN
```

**Create a new gist (if preexisting)**
```
pxt gist --new
```

**Create a public gist (instead of secret)*
```
pxt gist --public
```

valid example: pxt gist --token TOKEN --new --public

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -3438,7 +3438,7 @@ function publishGistCoreAsync(forceNewGist: boolean = false, publishPublicly: bo
     const token = globalConfig.githubAccessToken;
     return mainPkg.loadAsync()
         .then(() => {
-            let pxtConfig = U.clone(mainPkg.config);
+            const pxtConfig = U.clone(mainPkg.config);
             if (publishPublicly && pxtConfig.gistId && !pxtConfig.publicGist) {
                 console.warn("You are trying to update an existing project but the current project is secret, publishing a new public project instead.")
                 forceNewGist = true;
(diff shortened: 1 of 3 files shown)


cli/cli.ts
```diff
@@ -3451,19 +3457,47 @@ function publishGistCoreAsync(): Promise<void> {
...
+            console.log("Uploading....")
+            return pxt.github.publishGist(token, forceNewGist, filesMap, pxtConfig.description, pxtConfig.gistId, publishPublicly)
         })
         .then((published_id) => {
-            console.log(`Success, your gist url is https://gist.github.com/${published_id}`);
+            console.log(`Success, view your gist at https://gist.github.com/${published_id}`);
             console.log(`You can load your published project at ${pxt.appTarget.appTheme.homeUrl}#pub:gh/gists/${published_id}`)
+
+            // Save gist id to pxt.json
+            mainPkg.config.gistId = published_id;
```
pelikhan: If someone opens the PXT.json and tries to publish the gist, you will probably end up with some http error code like 403. We need to handle that.

microsoftsam: Okay, so the way to handle that would be to try, and if it fails, publish a new one?

microsoftsam: The gist id, should only be a local pxt thing. Whenever we publish to gist, I don't think we should send in the gistId. 
If we make sure to strip those out, then any fork has no attached gist id.

pelikhan:
Yes stripping out sounds good.

microsoftsam:
Are we happy with this?

PR #820, opened 2016-11-29
Adding an "Add package" button to the blockly toolbox

microsoftsam:
Adding an "Add package" button to the blockly toolbox. Clicking on the button opens up the Add package dialog.

Button is the same size in all viewports.

See examples below:

screen shot 2016-11-28 at 11 32 45 pm

screen shot 2016-11-28 at 11 11 55 pm


abchatra:
Does Add package needs to be more options as well?  @Jaqster FYI.

Issue #811, opened 2016-11-28
microbit.pxtCore.mainCli(targetDir, mainCliOpt); exits


OwenBrotherwood:
If for some reason it is not possible to bind to the ip/port for the serve, there is no error for the import(-ing)/require(-ing) program.
A simple exit routine is called:
\node_modules\pxt-core\built\pxt.js:77188 INTERNAL ERROR: Error: listen EADDRINUSE 127.0.0.1:3233

It would be desirable if there was a an error that the import/require-ing program could use so that the necessary user feedback could be created.

OwenBrotherwood:
It could also be desirable for some return from pxt-microbit to when the service is fully initialised.
At the present time, obtaining the local token is not sufficient to indicate that the pxt-microbit is fully initialised

Background:
In order to test that pxt-microbit is working in an encapsulated app, I have a test that creates a connection to the pxt-microbit url.
In some os's (found in Mac and Ubuntu), there has to be an artificial delay (or just repeat the connection attempt more times than strictly necessary) in order to ensure the encapsulated app works

OwenBrotherwood:
@guillaumejenkins @pelikhan 
https://github.com/microbit-drill/microbit-drill-epxt goes to v1.0.0 today with the fact that it can be run on Mac.
This issue is probably the biggest issue I have met concerning the use of pxt-microbit.
Now I go over to seeing if I can get something to run on Windows Core.

OwenBrotherwood:
I try and use asap : thanks

guillaumejenkins:
This still isn't fixed, because [`server.serveAsync()`](https://github.com/Microsoft/pxt/blob/master/cli/server.ts#L686) returns a promise that is never resolved. [This line](https://github.com/Microsoft/pxt/blob/master/cli/server.ts#L873) should be `return Promise.resolve()`; I'll fix that as part of my electron PR

guillaumejenkins:
PR is in, it should now be possible to use the promise returned by `mainCli()` , at least for the `serve` command

abchatra:
This is fixed.

OwenBrotherwood:
@guillaumejenkins I do not seem to be able to confirm that a INTERNAL ERROR: Error: listen EADDRINUSE 127.0.0.1:3233 is solved if the pr is finally pulled.

https://github.com/microbit-drill/epxt/issues/10#issuecomment-267658701

guillaumejenkins:
[...]

PR #827, opened 2016-11-30
fix for #811, mainCli returns promise

pelikhan:


OwenBrotherwood:
I am not having any luck with mainCli returns promise.
I attach my test code here: I may have made a simple mistake
Also, it is wished that the mainCli returns when server.js is finished, and if server.js has a problem, throw the err so it can be caught
```
/* 
test of Promise for mainCli
*/

"use strict";
var pxtMicrobit = require('pxt-microbit');
var path = require('path');
var reqResPxtMicrobit = require.resolve('pxt-microbit'); //TODO: is this "// Use new ES6 modules syntax for everything."
var targetDir = path.resolve(reqResPxtMicrobit, "..", ".."); //TODO: try and get rid of path
var mainCliOpt =  ["serve", "--no-browser"];

function mainCliNoPromise(){
    pxtMicrobit.pxtCore.mainCli(targetDir, mainCliOpt);
    var pxtLocalToken = pxtMicrobit.pxtCore.globalConfig.localToken;
    console.log("==========>>NO Promise:",pxtLocalToken);    
};

function mainCliPromise(){
    try{
        pxtMicrobit.pxtCore.mainCli(targetDir, mainCliOpt).then(function(){
//This is never called
            var pxtLocalToken = pxtMicrobit.pxtCore.globalConfig.localToken;
            console.log("==========>>Promise:", pxtLocalToken);    
        });
    }catch(ex){
// it would be nice if this would throw ex if one has enabled both functions (listen error)
        console.error(ex);
    }
};
```

PR #839, opened 2016-12-02
Developing with own instance of PXT

OwenBrotherwood:
I am using the enclosed information

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,13 +24,24 @@ More info:
 
 Please follow the [instructions here](https://www.pxt.io/cli).
 
-## Linking a target to PXT
+## Developing with own instance of PXT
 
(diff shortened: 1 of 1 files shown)


pelikhan:
Instead of one command (``npm link ../pxt``) the user has to issue 2 commands. What's the benefit?

OwenBrotherwood:
The link is then "global" there are made symbolic links from pxt to npm directory.
Then any reference to pxt-core is taken from the "global", ie own development.
It has it's pro's and con's I grant but saves me time in development.

OwenBrotherwood:
There seems to be an issue with electron in that it may not support links.
I am currently investigating what is best practise for electron in connection with links.
I close the pull request and create another when the whole truth is found:
- best practise links or npm install path_to_package which seems to be have been introduced after about version 2 of npm
- electron and links

PR #843, opened 2016-12-04
option --interface -i

OwenBrotherwood:
This pull request adds an option, --interface or -i to pxt serve.

Note that the automatic start of browser is checked against an interface of 0.0.0.0, in which case the automatic start is with 127.0.0.1

It is at the moment not known the consequences of multiple access to 0.0.0.0 for "localhost" type uses.
There may be a need for a separate token (uuid) per user in what is effectively a "localhost" working as a server.

### DIFF
--- a/cli/server.ts
+++ b/cli/server.ts
@@ -861,9 +861,9 @@ export function serveAsync(options: ServeOptions) {
     }
 
     server.listen(serveOptions.port, serveOptions.interface);
-    
+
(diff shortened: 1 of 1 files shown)


pelikhan:
``pxt serve`` is not meant to be served outside a developer box.

OwenBrotherwood:
https://github.com/Microsoft/pxt/pull/1395 ?

Issue #804, opened 2016-11-23
Reset crashes application 


microsoftsam:
Resetting the project crashes application: 

main.js:7809 Unhandled rejection TypeError: workspace.resetAsync(...).catch(...).done is not a function
    at https://pxt.azureedge.net/blob/50e21afb1073e20bb00a08161a3379dccda41a26/main.js:1179:18
From previous event:
    at ProjectView.reset (https://pxt.azureedge.net/blob/50e21afb1073e20bb00a08161a3379dccda41a26/main.js:1174:12)
    at onClick (https://pxt.azureedge.net/blob/50e21afb1073e20bb00a08161a3379dccda41a26/main.js:1438:3435)
    at Object.ReactErrorUtils.invokeGuardedCallback (https://pxt.azureedge.net/blob/50e21afb1073e20bb00a08161a3379dccda41a26/main.js:48022:16)
    at executeDispatch (https://pxt.azureedge.net/blob/50e21afb1073e20bb00a08161a3379dccda41a26/main.js:41038:21)
    at Object.executeDispatchesInOrder (https://pxt.azureedge.net/blob/50e21afb1073e20bb00a08161a3379dccda41a26/main.js:41061:5)
    at executeDispatchesAndRelease (https://pxt.azureedge.net/blob/50e21afb1073e20bb00a08161a3379dccda41a26/main.js:40497:22)
    at executeDispatchesAndReleaseTopLevel (https://pxt.azureedge.net/blob/50e21afb1073e20bb00a08161a3379dccda41a26/main.js:40508:10)
    at Array.forEach (native)
    at forEachAccumulated (https://pxt.azureedge.net/blob/50e21afb1073e20bb00a08161a3379dccda41a26/main.js:54508:9)
    at Object.processEventQueue (https://pxt.azureedge.net/blob/50e21afb1073e20bb00a08161a3379dccda41a26/main.js:40713:7)
    at runEventQueueInBatch (https://pxt.azureedge.net/blob/50e21afb1073e20bb00a08161a3379dccda41a26/main.js:48048:18)
    at Object.handleTopLevel [as _handleTopLevel] (https://pxt.azureedge.net/blob/50e21afb1073e20bb00a08161a3379dccda41a26/main.js:48064:5)
    at handleTopLevelWithoutPath (https://pxt.azureedge.net/blob/50e21afb1073e20bb00a08161a3379dccda41a26/main.js:48162:24)
[...]

PR #844, opened 2016-12-05
fixes for reset issue 

abchatra:
Couple of issues
- Pouchdb  promises were native promises instead of bluebird based. This is due to the fact we initialize window.Promise after requiring pouchdb. Reordered initialization to generate the right promise
- Unload event would trigger saving the scripts again. Guard against the reload. 

Fixes https://github.com/Microsoft/pxt/issues/804


OwenBrotherwood:
May have effected reset functionality https://github.com/Microsoft/pxt/issues/902

PR #865, opened 2016-12-06
Support for different docs font then main page font.

microsoftsam:
Adding docsPageFont and docsHeaderFont to set a different font for the docs page and sidedocs.

### DIFF
--- a/webapp/public/docs.html
+++ b/webapp/public/docs.html
@@ -66,7 +66,7 @@
     </style>
 </head>
 
-<body>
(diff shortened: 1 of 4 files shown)


webapp/public/docs.html
```diff
@@ -66,7 +66,7 @@
     </style>
 </head>
 
-<body>
+<body id="docs">
```
pelikhan: should this theme be applied to the "docs.html" under /docsfile as well?

PR #872, opened 2016-12-07
Electron communication and update dialogs

guillaumejenkins:


cli/cli.ts
```diff
@@ -3838,6 +3835,10 @@ export function mainCli(targetDir: string, args: string[] = process.argv.slice(2
         });
 }
 
+export function sendElectronMessage(message: server.ElectronMessage) {
```
pelikhan: do we need this function?

cli/server.ts
```diff
@@ -411,7 +387,10 @@ interface SerialPortInfo {
 }
 
 let wsSerialClients: WebSocket[] = [];
-let serialPorts: pxt.Map<SerialPortInfo> = {}
+let serialPorts: pxt.Map<SerialPortInfo> = {};
+let electronSocket: any = null;
```
pelikhan: type?

webapp/src/app.tsx
```diff
@@ -1936,6 +1918,166 @@ function initLogin() {
     }
 }
 
+let electronSocket: WebSocket = null;
```
pelikhan: could we move this into a new file?

webapp/src/app.tsx
```diff
@@ -1936,6 +1918,166 @@ function initLogin() {
...
+    }
+
+    function hideDownloading() {
+        core.hideLoading();
+    }
+
+    function onCriticalUpdate(args: any) {
+        core.confirmAsync({
+            header: lf("Critical update required"),
+            body: lf("To continue using {0}, you must update to the latest version.", args.appName || lf("this app")),
```
pelikhan: "you must update your application." (might not be the latest).

pelikhan: "this app" -> "this application"

webapp/src/app.tsx
```diff
@@ -1936,6 +1918,166 @@ function initLogin() {
...
+            agreeLbl: lf("Update"),
+            disagreeLbl: lf("Quit app"),
```
pelikhan: Just "Quit"

webapp/src/app.tsx
```diff
@@ -1936,6 +1918,166 @@ function initLogin() {
...
+                showDownloading();
+                sendElectronMessage("update", {
+                    targetVersion: args.targetVersion,
+                    isCritical: true
+                });
+            }
+        });
+    }
+
+    function onUpdateAvailable(args: any) {
```
pelikhan: type of "args" ?

webapp/src/app.tsx
```diff
@@ -1936,6 +1918,166 @@ function initLogin() {
...
+            agreeLbl: lf("Ok"),
+            hideCancel: true
+        });
+    }
+
+    function onUpdateCheckError() {
+        displayUpdateError(lf("Unable to check for update"), lf("Ok"));
+    }
+
+    function onUpdateDownloadError(args: any) {
```
pelikhan: type of "args" ? You can use inline types: ``args?: { isCritical?: boolean }``

webapp/src/app.tsx
```diff
@@ -1936,6 +1918,166 @@ function initLogin() {
...
+        let isCritical = args && args.isCritical;
+
+        hideDownloading();
+        displayUpdateError(lf("There was an error downloading the update"), isCritical ? lf("Quit app") : lf("Ok"))
```
pelikhan: "Quit app" => "Quit".
Is it "Quit" or "Exit" the correct terminology?

webapp/src/app.tsx
```diff
@@ -1936,6 +1918,166 @@ function initLogin() {
...
+            body: lf("Please ensure you are connected to the Internet and try again later."),
+            agreeClass: "red",
+            agreeIcon: "cancel",
+            agreeLbl: btnLabel,
+            hideCancel: true
+        });
+    }
+
+    pxt.log('initializing electron socket');
+    electronSocket = new WebSocket('ws://localhost:3233/' + Cloud.localToken + '/electron');
```
pelikhan: what happens for clients who do not have a serial terminal open, like Minecraft?

webapp/src/app.tsx
```diff
@@ -1936,6 +1918,166 @@ function initLogin() {
...
+}
+
+function sendElectronMessage(type: string, args?: any) {
+    if (!electronSocket) {
+        return;
+    }
+
+    let message: ElectronMessage = {
+        type: type,
+        args: args || void 0
```
pelikhan: can we simply use "undefined" instead of "void 0"? TypeScript or obfuscator can do the "void 0" stuff.

webapp/src/app.tsx
guillaumejenkins: Hmm, according to this [Stack Exchange question](http://ux.stackexchange.com/questions/50893/do-we-exit-quit-or-close-an-application), both seem fine, but Mac tends to use "Quit" and Windows, "Exit". We should just pick one and stick to it, which one do you prefer?

webapp/src/app.tsx
guillaumejenkins: Good point, I just need to move the `initSocketServer()` call outside of the `initSerialMonitor()` function in server.ts

webapp/src/app.tsx
pelikhan: "Quit"

webapp/src/app.tsx
guillaumejenkins: I have an interface on the Electron side, I'll reuse it

webapp/src/app.tsx
```diff
@@ -2320,7 +2156,11 @@ $(document).ready(() => {
         .then(() => {
             initSerial()
             initHashchange();
-            initElectron();
+
+            if (isElectron && Cloud.isLocalHost() && Cloud.localToken) {
```
pelikhan: should isElectron logic be moved in the electron module?

webapp/src/electron.ts
```diff
@@ -0,0 +1,174 @@
...
+
+interface ElectronMessage {
+    type: string;
+    args?: UpdateEventInfo;
+}
+
+let electronSocket: WebSocket = null;
+
+export function init() {
+    function showDownloading() {
```
pelikhan: how about putting the logic to init electron here?

pelikhan:
Looks good

cli/cli.ts
guillaumejenkins: Oops no it wasn't called, removed

PR #877, opened 2016-12-07
Opening the getting started docs in new tab when sidedocs is not supported

### LINKED ISSUE #841: Getting Started button does nothing in IE11
This is going to be blocker for beta->release. IE11 is our second most user base.

microsoftsam:
Showing the getting started button for the computer and tablet form factors, as well as on browsers that don't support sidedocs. Opening the getting started docs in a new tab in that case.

Fixes #841

### DIFF
--- a/theme/pxt.less
+++ b/theme/pxt.less
@@ -87,6 +87,9 @@
     top: @mainMenuHeight;
     right: 2rem;
 }
+#getting-started-btn:not(.sideDocs) {
+    display: block;
+}
 
 /* Side Docs layout */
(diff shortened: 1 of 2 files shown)


microsoftsam:
@pelikhan just want to make sure we're good to use window.open in gettingStartedLink. 
It's an onclick function call so browsers shouldn't have a problem with opening it in a new tab / window. 

Tested on IE, Chrome, Safari and Firefox.

pelikhan:
Why not just a <a/>?

PR #890, opened 2016-12-08
resurect  "save .pxt file" support

pelikhan:
Add a "Save Project" button under "Project" that generates a .pxt file (LZMA project contents).

### DIFF
--- a/pxtlib/emitter/driver.ts
+++ b/pxtlib/emitter/driver.ts
@@ -57,7 +57,9 @@ namespace ts.pxtc {
         quickFlash?: {
             words: number[];
             startAddr: number;
-        }
+        };
+        // client options
+        saveOnly?: boolean;
     }
 
     export function computeUsedParts(resp: CompileResult, ignoreBuiltin = false): string[] {
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -269,7 +269,7 @@ class ScriptSearch extends data.Component<ISettingsProps, ScriptSearchState> {
         const saveProject = () => {
             pxt.tickEvent("projects.save");
             this.hide();
-            this.props.parent.saveProjectToFile();
+            this.props.parent.compile(true);
         }
         const isEmpty = () => {
             if (this.state.searchFor) {
@@ -1460,7 +1460,7 @@ export class ProjectView extends data.Component<IAppProps, IAppState> {
         });
     }
 
-    compile() {
+    compile(saveOnly = false) {
         pxt.tickEvent("compile");
         pxt.debug('compiling...');
         if (this.state.compiling) {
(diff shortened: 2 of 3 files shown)


jaustin:
What's the advantage of the .pxt file over, say, .hex ? Does it contain more than the hex does?

pelikhan:
The .pxt file only contains the project sources (LZMA-ed) so it should be much smaller.

microsoftsam:
and human readable, unlike a .hex file (which is cryptic to students).

pelikhan:
It contains the project contains encoded in a JSON payload, encoded using LZMA. Not really human readable...

jaustin:
Not very human readable if LZMA'd thought, right? Also do we care about
512K versus something smaller?

pelikhan:
@jausin, what are you trying to imply? That the save button should produce a .hex file?

microsoftsam:
It might be worth considering making it human readable. That way there's a clear distinction between "Downloading" for the purpose of flashing it and "Saving" for the purpose of sharing it. 

If we are going to go down the path of making compressing it, then I think the above link (.hex for downloading and .pxt for saving / sharing), but we might as well make it a .hex, thus allowing students to share the file OR flash their device / any device with it AND at the same time educating them in some way that a .hex file can be used for both (sharing and downloading).

pelikhan:
> It might be worth considering making it human readable

The majority of the users are blockly users, blockly is encoded in cryptic XML. I don't see how we can make this human readable. We also have exotic files in JSON, Make, .gitignore.

>  but we might as well make it a .hex, thus allowing students to share the file OR flash their device / any device with it AND at the same time educating them in some way that a .hex file can be used for both (sharing and downloading).

Sounds like a reasonable argument for generating .hex file.

jaustin:
I think there are two logical things: hex-including-json or javascript (if you're editing JS). I've copy/pasted out javascript a number of times, but I need to remember I'm not necessarily representative!

pelikhan:
Copying javascript will not work if you have package dependencies or blocks+js projects. You can copy paste JS without a UI menu item.

abchatra:
I really like human readable format. We already have a hex format which is not readable and I don't see value in create another format .pxt.

Zip format which contains blocks, javascript & hex file will be awesome here. Ideally users should be able to drag to pxt editor or flash the device using just the zip file.

pelikhan:
Human readble for a pro developer, but is it really helpful for kids? We already have a file format to share file., ``.hex``, that contains all we need. Do we really expect our users to unzip a file, edit the JS by hand, rezip and share?

* ``.json``: not humand readable
* ``.blocks``: not human readable
* ``.ts``: not really human friendly for blocks users
* ``.gitignore``, ``makefile``, ...: not humand friendly

abchatra:
For some one who can only understand blockly, all formats are black box. Then the format you save doesn't matter (.pxt or .zip). But for someone who is progressing from blockly to typescript, .ts is definitely readable. This is same reasoning as we show explorer bar in the typescript view but not in blockly view.

pelikhan:
No hard feelings, we can also go with ``.zip`` files to please the advanced crowd.

jaustin:
I agree .ts is useful but I think Peli makes a good point that it might
actually be more confusing than not once people start adding packages, etc.
If you have a simple .ts file and it doesn't work when you import it to PXT
that's annoying.

.zip with everything you need for building would be ideal IF going advanced
(might even be a chance some advanced people choose to do plugins for other
IDEs, etc?!).

I still like .hex for blocks as it keeps things nice and simple.
J

PR #903, opened 2016-12-09
nav menu refactoring

pelikhan:
- project input box moved to settings page
- project name displayed in window title
- window title gets updated with name
- projects and /blocksjavascript buttons swapped
- blocks/javascript have an active state
- remove editor specific menu item
- renamed "Project Settings..." to "Rename..."
- hamburger menu to the right, using "setting" icon

bugreport5

Updated icons.

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,13 @@
 language: node_js
 node_js:
-  - "5.7.0"
+  - "6.9.2"
 before_script:
   - "npm install"
   - "tsd reinstall"
 sudo: false
 notifications:
   email:
-    - touchdevelop-build@microsoft.com
+    - yelm-eng@microsoft.com
 cache:
   directories:
     - node_modules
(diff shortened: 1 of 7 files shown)


jaustin:
I like the tabs here, although not sure how clear it is to a user that it’s a tab until they click it - we can see I guess whether the number of clicks of Javascript increases. I think the main question is "is it clear that they can see the *same* project with a different view, or will that just take them to a fresh slate with Javascript". Someone showed a toggle somewhere, which I thought was nice.

On the project settings - I like name in title, and I think it's good no longer being on the main bar, but I'd call the option in the hamburger "rename" just to make rename a bit more discoverable for the future.

pelikhan:
This PR + https://github.com/Microsoft/pxt/pull/900 (removing sidedocs) gets us close to the changes discussed.

PR #921, opened 2016-12-12
Fix error when very large numbers are typed into blocks

riknoll:
Before, they just compiled to "Infinity" and you got a "name not recognized" error.


pelikhan:
Could we add tests?

PR #925, opened 2016-12-13
Hats

pelikhan:
- [x] render events with hats
- [x] compile to javascript
- [x] decompile from javascript
- [x] layout blocks as needed
- [x] fix test suite
- [x] when importing legacy script, bail out to javascript and decompile from there (to recover blocks)
- [x] use flow layout (squarish) instead of just stacking blocks in docs

Note that we are not enabling hats currently; Blockly does not properly compute the bounding box of blocks with hats which messes up our docs rendering. hats can be enabled via pxtarget.json flag.

image

Flow layout works better for docs

image


pelikhan:
Blocks rendered as hats:
image

PR #926, opened 2016-12-13
Merge electron repo

guillaumejenkins:


cli/cli.ts
```diff
@@ -17,6 +17,7 @@ import Map = pxt.Map;
```
guillaumejenkins: The PXT CLI commands to init / run / build the Electron app

cli/electron.ts
```diff
@@ -0,0 +1,280 @@
+"use strict";
```
guillaumejenkins: The PXT CLI commands to init / run / build the Electron app

electron/build/win32/pxt.iss
```diff
@@ -0,0 +1,71 @@
+#define LocalizedLanguageFile(Language = "") \
```
guillaumejenkins: The Inno-Setup compiler configuration file for the Windows installer

electron/build/win32/strings/installer.en.isl
```diff
@@ -0,0 +1,3 @@
+[CustomMessages]
```
guillaumejenkins: Installer strings that will need to be localized. Localization of this will be annoying, we'll need to figure it out.

electron/dev_tools/rebuild-native.js
```diff
@@ -0,0 +1,139 @@
+/**
```
guillaumejenkins: Dev-only script to rebuild native modules for Electron consumption. Can mostly be ignored.

electron/dev_tools/test_data/release.json
```diff
@@ -0,0 +1,7 @@
+{
```
guillaumejenkins: Fake release manifest for testing app update scenarios.

electron/package.json
```diff
@@ -0,0 +1,40 @@
+{
```
guillaumejenkins: "Dev" package.json for the Electron app. Only contains dependencies needed to **work** on the app, but nothing that the app needs to actually run.

electron/src/main.ts
```diff
@@ -0,0 +1,201 @@
+/// <reference path="typings/windows-mutex.d.ts" />
```
guillaumejenkins: The entry point for the app shell. see `main()`.

cli/electron.ts
```diff
@@ -0,0 +1,280 @@
...
+
+import * as p from "./commandparser";
+import * as fs from "fs";
+import * as nodeutil from './nodeutil';
+import * as path from "path";
+
+const gulpCmd = process.platform === "win32" ? "gulp.cmd" : "gulp";
```
pelikhan: Why do we need a gulp dependency in our CLI?

cli/electron.ts
```diff
@@ -0,0 +1,280 @@
...
+export function electronAsync(parsed: p.ParsedCommand): Promise<void> {
+    // Ensure there is a subcommand
+    let subcommand = parsed.arguments[0];
+
+    if (!subcommand) {
+        errorOut("Please specify a subcommand");
+    }
+
+    // Validate current target
+    let needsCurrentTarget = (subcommand !== "build" && subcommand !== "dist") || !parsed.flags["release"];
```
pelikhan: What is the "dist" subcommand?

electron/src/updater/updateServer.ts
```diff
@@ -0,0 +1,82 @@
...
+import product from "../util/productInfoLoader";
+import * as qs from "querystring";
+import * as url from "url";
+import * as Utils from "../util/electronUtils";
+
+/**
+ * This server is needed due to how Squirrel.Mac works. Squirrel.Mac expects JSON responses in a specific format.
+ * Because we host update binaries on Github, we need a "proxy" server that will translate responses from PXT API to a
+ * format Squirrel understands / expects.
+ */
```
guillaumejenkins: See the in-file comment just above this for a description of this file.

cli/electron.ts
```diff
@@ -0,0 +1,280 @@
...
+
+    if (needsCurrentTarget && !fs.existsSync("pxtarget.json") && !fs.existsSync("package.json")) {
```
pelikhan: shouldn't it be
```
needsCurrentTarget && (!fs.existsSync("pxtarget.json") || !fs.existsSync("package.json"))
```

electron/src/updater/updateService.ts
```diff
@@ -0,0 +1,206 @@
+"use strict";
```
guillaumejenkins: The gut of the update logic. It is responsible for downloading and caching the release manifest, interpreting it, and emitting update events to the shell (which in turns pipes these events to the webapp, which can then display dialogs to the user).

cli/electron.ts
```diff
@@ -0,0 +1,280 @@
...
+        errorOut("This command requires to be in a valid target directory (pxtarget.json and package.json required)");
+    }
+
+    targetId = JSON.parse(fs.readFileSync("package.json", "utf8")).name;
```
pelikhan: we have boiler plate code to load pxtarget.json which contains the targetid, then you can use ``pxt.appTarget.id``.

electron/src/updater/windowsUpdater.ts
```diff
@@ -0,0 +1,125 @@
+"use strict";
```
guillaumejenkins: The updater implementation for Windows. Pretty much taken from VS Code. Downloads the new installer when told to, saves it to temp, then launches it as a detached process and exits the app. The installer takes care of overwriting all necessary files, rolling back in case of error, etc.

cli/electron.ts
```diff
@@ -0,0 +1,280 @@
...
+
+    // Invoke subcommand
+    switch (subcommand) {
+        case "init":
+            return initAsync();
+        case "run":
+            return runAsync();
+        case "build":
+            return buildAsync(parsed);
+        case "dist":
```
pelikhan: dist -> package

cli/electron.ts
```diff
@@ -0,0 +1,280 @@
...
+        default:
+            return errorOut("Unknown subcommand: " + subcommand);
+    }
+}
+
+function initAsync(): Promise<void> {
+    return npmVersion()
+        .then((v) => {
+            if (/^3\./.test(v)) {
+                return errorOut("'pxt electron init/run' is not supported in NPM 3+ due to a package linking bug. You can still use 'pxt electron build', however.");
```
pelikhan: which version should we be using then?

electron/src/util/electronUtils.ts
```diff
@@ -0,0 +1,191 @@
+"use strict";
```
guillaumejenkins: Some utils that I copied (and in some cases modified) from our PXT.Util module. I can't use PXT.Util directly because I don't want to have to depend on files outside the electron repo, which would make it rather annoying when building and packaging the app.

electron/src/util/productInfoLoader.ts
```diff
@@ -0,0 +1,16 @@
+"use strict";
```
guillaumejenkins: As name suggests. A loader for the product.json file of the app, which contains target-specific info (just like pxtarget.json), such as the name of the app, the app GUID, etc

cli/electron.ts
```diff
@@ -0,0 +1,280 @@
...
+    return nodeutil.spawnAsync({
+        cmd: npmCmd,
+        args: [
+            "link",
+            moduleToLink],
+        cwd
+    });
+}
+
+function npmInstall(cwd: string, npmPackage?: string): Promise<void> {
```
pelikhan: why not using ``runNpmAsync`` ?

cli/electron.ts
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    return nodeutil.spawnAsync({
+        cmd: npmCmd,
+        args,
+        cwd
+    });
+}
+
+function npmUninstall(cwd: string, npmPackage: string): Promise<void> {
```
pelikhan: why not using ``runNpmAsync``?

cli/electron.ts
```diff
@@ -0,0 +1,280 @@
...
+        cmd: npmCmd,
+        args: [
+            "uninstall",
+            npmPackage
+        ],
+        cwd
+    });
+}
+
+function npmPrune(cwd: string): Promise<void> {
```
pelikhan: why not using ``runNpmAsync``?

cli/electron.ts
```diff
@@ -0,0 +1,280 @@
...
+    return nodeutil.spawnAsync({
+        cmd: npmCmd,
+        args: ["prune"],
+        cwd
+    });
+}
+
+function npmRunElectronScript(script: string, scriptArgs?: string[]): Promise<void> {
```
pelikhan: why not running ``runNpmAsync``?

electron/src/webview/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
```
guillaumejenkins: The entry point of the web content that runs inside Chromium. Just spins up a webview, which gets navigated to our local server that is running `pxt serve`.

electron/build/win32/pxt.iss
```diff
@@ -0,0 +1,71 @@
...
+    DirExists(RepoDir + "\licenses") && Language != "" \
+      ? ('; LicenseFile: "' + RepoDir + '\licenses\LICENSE-' + Language + '.txt"') \
+      : '; LicenseFile: "' + RepoDir + '\LICENSE.txt"'
+
+[Setup]
+AppId={#AppId}
+AppName={#NameLong}
+AppVerName={#NameVersion}
+AppPublisher=Microsoft Corporation
+AppPublisherURL=https://pxt.microbit.io/
```
pelikhan: invalid URI

electron/src/webview/index.ts
```diff
@@ -0,0 +1,18 @@
+"use strict";
```
guillaumejenkins: Script that is included by the main HTML page of the web content.

electron/src/updater/updateServer.ts
```diff
@@ -0,0 +1,82 @@
+"use strict";
```
pelikhan: do we need this file?

electron/src/util/electronUtils.ts
```diff
@@ -0,0 +1,191 @@
...
+        .then((content: string) => {
+            return JSON.parse(content) as T;
+        });
+}
+
+export function writeJsonFileAsync<T>(content: T, thePath: string): Promise<void> {
+    return fsWriteFilePromise(thePath, JSON.stringify(content));
+}
+
+export function request(options: I.RequestOptions): Promise<I.RequestContext> {
```
pelikhan: can't we use helpers from pxt CLI?

electron/src/webview/index.ts
```diff
@@ -0,0 +1,18 @@
...
+import { WebviewStartMessage } from "../typings/interfaces";
+
+function start(event: Electron.IpcRendererEvent, message: WebviewStartMessage) {
+    if (message.devtools) {
+        webview.addEventListener("did-stop-loading", () => {
+            webview.openDevTools();
+        });
+    }
+
+    webview.src = `http://localhost:3232/#local_token=${message.localtoken}&electron=1`;
```
pelikhan: what if port is blocked?

cli/electron.ts
guillaumejenkins: yes good catch

cli/electron.ts
guillaumejenkins: I use "package" and "build" interchangeably in comments, code and readme, so what about "buildInstaller" or something?

cli/electron.ts
guillaumejenkins: Clarified the message

cli/electron.ts
pelikhan: let's use "package", "build" has a different mean with "pxt build"

cli/electron.ts
pelikhan: It is not enforced. Simply use the helper function to load the target json on start and you'll have access to pxt.appTarget.id

electron/src/updater/updateServer.ts
guillaumejenkins: Yes, for the Mac updater. Electron.autoUpdater expects a JSON response containing a url to the downloadable asset, but our backend api downloads the asset directly. This server simply immediately returns the necessary json response:
```
{
    url: "url to downloadable update form pxt backend"
}
```

If we remove this file, we need a new route in our backend to return the above response. I think an extra file in the app is much better than another query to the backend (which can fail, takes some time, etc)

cli/electron.ts
guillaumejenkins: Hadn't noticed that one :) moving it to nodeutil so I can use it in this file

pelikhan:
Should be replaced with the target homeUrl field or electronUrl field?

electron/src/webview/index.ts
guillaumejenkins: Now that the CLI supports custom port I'll add logic to find a free port in the Electron main process, and then it can communicate the port to the webview

cli/cli.ts
```diff
@@ -3648,6 +3596,32 @@ function initCommands() {
         }
     }, publishGistAsync)
 
+    p.defineCommand({
+        name: "electron",
+        help: "SUBCOMMANDS: 'init': prepare target for running inside Electron app; 'run': runs current target inside Electron app; 'package': generates a packaged Electron app for current target",
```
pelikhan: I don't see "dist" here.

cli/electron.ts
```diff
@@ -0,0 +1,214 @@
+"use strict";
```
pelikhan: do we need this annotation?

electron/src/updater/updateServer.ts
guillaumejenkins: Here's how Electron AutoUpdater works (this is not my stuff, this is how Electron is implemented):

1. We give it a URL
2. We tell it to check for an update
3. It queries the URL from 1. to check for an update
4. a. If there is no update available, it expects status code 204
    b. If there is an update available, it expects a 200 with the following JSON in the body:
```
        {
            url: "download URL"
        }
```
5. If an update was available, it queries the download URL from the JSON response to download the update in .zip format

There's no way to tell the Electron updater to skip 1., 2. and 3. The thing is, we have our own way of deciding if there's an update or not, by using our release manifest. So, we _already_ know there's an update available. There's absolutely no reason to let Electron AutoUpdater do an unnecessary query to our backend to get the JSON response. Instead, we use this local server to satisfy the query at 3. above, and the only time the AutoUpdater module queries the PXT backend is to download the actual binary.

If there was a way to tell Electron to skip the check for update, we wouldn't need this, but there isn't.

electron/src/updater/windowsUpdater.ts
```diff
@@ -0,0 +1,125 @@
...
+    constructor(private cacheInfo: I.UpdateCacheInfo) {
+        super();
+    }
+
+    public setFeedURL(url: string): void {
+        this.url = url;
+    }
+
+    public checkForUpdates(): void {
+        if (!this.url) {
```
abchatra: This file is lacking in comments. Not a blocker for merging the PR. But please add it after merging.

PR #940, opened 2016-12-16
Detect and disable duplicate events

pelikhan:
- [x] detect when duplicate events are registered in blocks taking into account event arguments
- [x] disable insert duplicate event
- [x] add warning on disabled event
- [x] remove warning when compilation update

dupdetect

Issue not addressed in this PR: in general, we don't store disabled blocks in Javascript. This creates a rountrip issue where we potentially loose code when the user modifies the javascript. We need to support this scenario for the "on start" event functionality any so it will be addressed in a future PR.

### DIFF
--- a/pxtblocks/blocklycompiler.ts
+++ b/pxtblocks/blocklycompiler.ts
@@ -210,29 +210,6 @@ namespace pxt.blocks {
         throw e;
     }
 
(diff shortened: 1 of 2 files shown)


abchatra:
:shipit:

Issue #941, opened 2016-12-16
Uncaught exception in server.serveAsync() when port in use


guillaumejenkins:
If the port is in use when serving, an exception is thrown outside the server Promise chain, which results in an ugly crash. We should catch that exception and return a rejected promise instead.

guillaumejenkins:
#942 ensures the `server.serveAsync()` promise gets rejected properly if the specified port is in use. Consumers of `mainCli()` will need to catch that rejected promise and try again manually with a different port.

PR #942, opened 2016-12-16
Fix for #941

guillaumejenkins:


OwenBrotherwood:
Confirmed, tested and working and thanks

PR #979, opened 2016-12-21
Help URLs now open in the default browser

### LINKED ISSUE #767: access to menu items requires "new window"
The access to menu items such as those in "?" , "Privacy and cookies", "Terms of Use" require a new window to present the contents.
In a --not-browser scenario, such as is the case if using electron, there is probably need for more work in order to support presentation of the information that the menu items refer to as no new window is presented when the menu item is clicked

---

Edit

Clarification.

"?" menu items are shown in the right hand side of the webpage if the size of the window is above some limit.
On a small window size, the "new window" is not created.

guillaumejenkins:
Fixes #767

### DIFF
--- a/electron/src/webview/index.ts
+++ b/electron/src/webview/index.ts
@@ -1,7 +1,8 @@
 "use strict";
 
-import { ipcRenderer } from "electron";
+import { ipcRenderer, shell } from "electron";
 import { WebviewStartMessage } from "../typings/interfaces";
+import * as url from "url";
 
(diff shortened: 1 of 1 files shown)


OwenBrotherwood:
I have to do some more work to double check, but at the present time, it doesn't seem to work on https://github.com/microbit-drill/epxt 
The most important is that MS's electron encapsulation works.

OwenBrotherwood:
oops: it is not committed: I apologise and await the commit

Issue #965, opened 2016-12-20
help text inhibits access to block


OwenBrotherwood:
Unable to access block due to help text.
The child has to start moving blocks around or use +/-
![image](https://cloud.githubusercontent.com/assets/866122/21367440/77361e22-c6fe-11e6-9e27-9d1a9bec216e.png)

pelikhan:
What is the resolution on this screen?

OwenBrotherwood:
Same type of issue with "Getting Started" both with full screen 1920x1080
![image](https://cloud.githubusercontent.com/assets/866122/21368580/7139c7d0-c703-11e6-9556-36a3f9d46d4f.png)

pelikhan:
PR https://github.com/Microsoft/pxt/issues/965

microsoftsam:
PR #1033

PR #987, opened 2016-12-22
remove codecard from blocks editor

pelikhan:
This PR removes the automatic codecard when clicking on a block.
- Codecard gets in the way
- Introduces a small delay to render the card
- Description of the blocks is available in the tooltip
- Full help still available on the context menu


abchatra:
Is it better to have an option to close the codecard instead of removing it?

pelikhan:
Currently it opens up on each selected block, so the user would end up closing the codecard constantly.

pelikhan:
https://github.com/Microsoft/pxt/issues/965

pelikhan:
Replaced by https://github.com/Microsoft/pxt/pull/1033

PR #988, opened 2016-12-22
Add support for website updates

guillaumejenkins:
Enables us to mark certain versions for "url updates", which causes the update prompts to open a URL in the default browser instead of downloading the update directly.

### DIFF
--- a/electron/src/typings/interfaces.d.ts
+++ b/electron/src/typings/interfaces.d.ts
@@ -54,7 +54,7 @@ export interface VersionInfo {
     banned: string[];
     latest: string;
     prompt: string;
-    url: { [versionRange: string]: string };
+    urls: { [versionRange: string]: string };
 }
 
 export interface ReleaseManifest {
--- a/electron/src/typings/interfaces.d.ts
+++ b/electron/src/typings/interfaces.d.ts
@@ -54,6 +54,7 @@ export interface VersionInfo {
     banned: string[];
     latest: string;
     prompt: string;
+    url: { [versionRange: string]: string };
 }
 
 export interface ReleaseManifest {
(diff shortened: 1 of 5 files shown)


electron/src/updater/updateservice.ts
```diff
@@ -115,11 +116,20 @@ export class UpdateService extends EventEmitter {
...
 
     /**
-     * Performs the update to the specified target version by using the AutoUpdater implementation for the current
-     * platform. By using the AutoUpdater.checkForUpdates(), any available update is downloaded automatically. When the
-     * download is complete, we install the update in the update-downloaded handler.
+     * Performs an update. If the target version is a URL, the URL is opened in the default browser. If the target
+     * version is a tag, the associated installer is downloaded and run.
      */
     public update(targetVersion: string, isCritical: boolean = false): void {
+        if (/^https?:\/\//.test(targetVersion)) {
```
pelikhan: should always be https.

electron/src/typings/interfaces.d.ts
```diff
@@ -54,6 +54,7 @@ export interface VersionInfo {
     banned: string[];
     latest: string;
     prompt: string;
+    url: { [versionRange: string]: string };
```
pelikhan: make it plural ("urls")

webapp/src/electron.ts
```diff
@@ -48,12 +57,21 @@ export function init() {
     }
 
     function onUpdateAvailable(args: UpdateEventInfo) {
-        const header = lf("Version {0} available", args.targetVersion);
+        const isUrl = /^https?:\/\//.test(args.targetVersion);
```
pelikhan: always https ?

PR #994, opened 2016-12-23
App telemetry

guillaumejenkins:


electron/src/updater/updateservice.ts
```diff
@@ -82,15 +85,27 @@ export class UpdateService extends EventEmitter {
 
                 if (targetVersion) {
                     if (criticalPrompt) {
+                        Telemetry.tickEvent("electron.update.available", {
+                            initial: "true",
```
pelikhan: bool not supported ?

electron/src/util/telemetry.ts
```diff
@@ -0,0 +1,27 @@
+"use strict";
+
+import * as I from "../typings/interfaces";
+
+let pxtCore: I.PxtCore = null;
+let pendingEvents: I.TelemetryEventInfo[] = [];
+
+export function init(pxt: I.PxtCore): void {
+    pxtCore = pxt;
+    pendingEvents.forEach((e) => tickEvent(e.event, e.data));
```
pelikhan: Clear array?

electron/src/updater/updateservice.ts
guillaumejenkins: It doesn't look like it according to [appInsights docs](https://docs.microsoft.com/en-us/azure/application-insights/app-insights-api-custom-events-metrics#properties)

electron/src/util/telemetry.ts
guillaumejenkins: Will do

PR #996, opened 2016-12-26
Aligning categories with Scratch

pelikhan:
This PR is realigning various aspects of the editor with Scratch.
- [ ] refactoring built-in categories (loops, Math, logic, Variables) into Scratch-like categories (Control, Data, Operators). 
- [ ] full for loop support
- [ ] random range support

Not ready for merge.

scratchcat

### DIFF
--- a/libs/pxt-common/pxt-core.d.ts
+++ b/libs/pxt-common/pxt-core.d.ts
@@ -214,7 +214,7 @@ declare namespace String {
     //% help=math/string-from-char-code
     //% shim=String_::fromCharCode
     //% advanced=true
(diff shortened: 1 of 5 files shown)


Jaqster:
I like the idea of reorganizing, but don't want to copy Scratch too much...  For example, I don't like "Control" - that isn't really helpful in aligning to modern programming practices for example.  Loops and Variables seem more consistent names/categories.  We can discuss "Data"/"Operators" versus "Logic"/"Math" - I always thought those were a bit confusing...  Let's ask some middle school technology teachers their opinion.

pelikhan:
Control encompasses control flow structure, including loops, if statements. Also, it is a great place to put "on start". I don't think we want a single "events" category as that one is more API-specific.

PR #1009, opened 2016-12-30
Handle string concatenation in the blocks

### LINKED ISSUE #678: String concatenation not caught by decompiler
This used to be caught by the roundtrip, but hasn't been handled in the decompiler. Repro:

```typescript
let item = "hello" + 1
```

Switching to blocks will show no blocks (or really messed up blocks) and switching back to JS gives you this:

```typescript
let item = 0
0 + 0
"hello"
```

riknoll:
Fixes #678 

This makes it so that string concatenation works when going from typescript to blocks. I think the future plan is to have a whole string package that gets added when you use strings, but I don't think that story is 100% thought out yet. In the blocks, I am using Blockly's built in `text_join` block. For example,

With this typescript:

```typescript
let x = "hello" + (5 + 6) + "okay";
```

You get these blocks:

text_join

I don't do type checking in the blocks for the inputs. Rather, I ensure that the block always outputs a string by checking if the first input is a string and adding an empty string if it is not. That ensures that these blocks:

text_concat2

Produce this TS:

```typescript
//  A string 
let x = "" + 5 + 6
```

And not this TS:

```typescript
// A number
let x = 5 + 6
```

Blockly solves that issue by generating code that looks like this:

```typescript
let x = new String("hello") + new String(5 + 6) + new String("okay");
```

...but that code is much uglier.

### DIFF
--- a/pxtblocks/blocklycompiler.ts
+++ b/pxtblocks/blocklycompiler.ts
@@ -557,6 +557,28 @@ namespace pxt.blocks {
         return H.mkStringLiteral(b.getFieldValue("TEXT"));
     }
 
+    function compileTextJoin(e: Environment, b: B.Block, comments: string[]): JsNode {
+        let last: JsNode;
+        let i = 0;
+        while (true) {
+            const val = b.getInputTargetBlock("ADD" + i);
+            i++;
+
+            if (!val) {
+                break;
+            }
+
+            const compiled = compileExpression(e, val, comments);
+            if (!last) {
+                last = compiled
+            }
+            else {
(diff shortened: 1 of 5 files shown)


rharbird:
Hello,

New text blocks look great. Just adding a wish for concatenating strings and numbers or two numbers with the result being a string.  We have a use case for this: when run, our program will create a new file with a unique name. The filename is formed from a string with a number suffix. The program should check whether the filename already exists and, if it does, increment the number and try again....


Regards,



Rae

pelikhan:
@rharbird could you open a new issue?

PR #1017, opened 2016-12-30
Error out correctly on reserved or invalid variable names

### LINKED ISSUE #1001: "new" as name for variable create javascript error
If you create a variable with the name "new" the javascript line becomes for example `new = 5` and that goes wrong because new is a special word in javascript. But when you are inside the blockeditor you won't understand what goes wrong if you don't know this.

So it should prevent you from using reserved words as variable names?

### LINKED ISSUE #1005: Blockly variable name encoding causes error
This Blockly:
screen shot 2016-12-29 at 2 14 28 pm

Converts to this Typescript:
screen shot 2016-12-29 at 2 12 48 pm

And my program won't build or run.

riknoll:
Fixes #1001 
Fixes #1005 

This PR adds friendlier errors for invalid variable names in the blocks; currently handles variable names that start with numbers and reserved keywords. As part of the PR, I removed a call to `setDiagnostics()` on the current editor when the simulator is run because it seemed redundant (we already set the diagnostics as part of our type checking). Let me know if that was in error.

### DIFF
--- a/localtypings/blockly.d.ts
+++ b/localtypings/blockly.d.ts
@@ -302,6 +302,7 @@ declare namespace Blockly {
             viewTop: number;
             viewWidth: number;
         }
+        variableIndexOf(name: string): number;
     }
 
     class WorkspaceSvg {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pxt-core",
-  "version": "0.7.11",
+  "version": "0.7.12",
   "description": "Programming Experience Toolkit provides Blocks / JavaScript tools and editors",
   "keywords": [
     "TypeScript",
--- a/pxtblocks/blocklycompiler.ts
+++ b/pxtblocks/blocklycompiler.ts
@@ -307,7 +307,7 @@ namespace pxt.blocks {
             return find((<any>b).p);
 
         if (b.type == "variables_get")
-            return find(lookup(e, escapeVarName(b.getFieldValue("VAR"), b)).type);
+            return find(lookup(e, escapeVarName(b.getFieldValue("VAR"), e)).type);
 
         assert(!b.outputConnection || b.outputConnection.check_ && b.outputConnection.check_.length > 0);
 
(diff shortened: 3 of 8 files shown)


pelikhan:
Could we automatically rename the variables?

riknoll:
@pelikhan I think automatic renaming in this case is more confusing than helpful. I can't think of a good renaming scheme that would make sense for a user.

pelikhan:
In blocks, users should be able to use whatever variable name they like. We can add a _ to numbers or reserved keywords. Ugly but better than an error IMO.

riknoll:
It's a pretty easy error for them to fix, though. Blockly says we should allow anything as a name, but Blockly also doesn't really have our roundtrip model. Should we re-convert variable names with underscores when going back to the blocks?

pelikhan:
It is not obvious that they will understand the issue. We should auto rename blocks to js and do nothing the other way.

riknoll:
Okay, underscore doesn't look great in the blocks:

underscore

Any other ideas?

riknoll:
Also, how should we handle name collisions?

pelikhan:
Maybe add trailing numbers and also avoid duplicates like this?

riknoll:
Done! Adds underscores to reserved names and handles name collisions by numbering (including the name collisions we were not handling before).

darzu:
One idea would be to have variable name just be keys on some object.

`
VARS["1st variable"] = 5
if (VARS["2nd variable"]) { ... }
`

This would only be necessary for troublesome variable names but should allow perfect round tripping.

Scope/shadowing would need special consideration but it might not be that hard.

pelikhan:
We have perfect round tripping if the JS is not modified.

If kids are going to edit JS, they need to learn proper variable naming habits.

riknoll:
@pelikhan we don't have any tests for the Blockly compiler right now

pelikhan:
Oopsy, LGTM

PR #1022, opened 2017-01-02
Array builtins - pop, unshift, shift, reverse & sort

abchatra:
Implements various array built-ins such as pop, shift, unshift, reverse and sort. Also implements simulator to mimic target behavior.

More built-ins can be easily added if needed.


abchatra:
This is dependent on https://github.com/Microsoft/pxt-microbit/pull/328. Need to merge that first.

PR #1028, opened 2017-01-03
Fix flyout highlighting bug in Edge

### LINKED ISSUE #696: Blocks highlighting in Edge
Repro: Click on a toolbox category, move mouse over flyout blocks - all of them get highlighted

riknoll:
Fixes #696 

This is a blockly bug and so the fix involves overwriting some code in Blockly. I didn't know where to put the fix, so I went with the Blockly loader because we already have some other Blockly fixes for the toolbox there. 

This changes the behavior of the block highlighting slightly so that blocks that are highlighted no longer un-highlight when you mouse away from them. The change isn't very noticeable, but I can limit this fix to Edge if people find it annoying.


microsoftsam:
LGTM. Tested on Chrome, Firefox and Safari for Mac.

PR #1030, opened 2017-01-04
JavaScript code card

pelikhan:
This change renders the selected block as javascript in the codecard -- instead of the jsdocs/block.
- jsdoc is already available in the tooltip
- subtle way to show the JavaScript code related to a block
Not addressed: code card getting in the way.
image

- height is limited so that long snippets don't cover the entire screen


OwenBrotherwood:
Not addressed: code card getting in the way.
On block movement, suppress card visible on screen.

pelikhan:
Replace by https://github.com/Microsoft/pxt/pull/1033

PR #1032, opened 2017-01-05
Namespace icons

pelikhan:
Block icons are applied for an entire namespace, rather than individual blocks.

- [x] support icon namespace annotation
- [x] render icon in toolbox

image


microsoftsam:
screen shot 2017-01-05 at 9 40 39 am

screen shot 2017-01-05 at 9 40 50 am

PR #1050, opened 2017-01-07
moving import/new/save as buttons

pelikhan:
A small reorg of the projects dialog until cleanup. Note that save and rename should eventually move away to the new toolbar.

image

### DIFF
--- a/webapp/public/custom.css
+++ b/webapp/public/custom.css
@@ -54,11 +54,11 @@ body {
 
 
 .searchdialog {
-    margin-top: -15rem !important;
+    margin-top: -20rem !important;
 }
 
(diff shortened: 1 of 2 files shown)


abchatra:
Looks good me. But why is import next to search bar?

pelikhan:
Import is related to "opening a script". The above buttons are related to the current project.

Save, rename should move out to bottom toolbar soon.

microsoftsam:
Looks good. Minor comment, could make the import button attached to the search box. (similar to the search icon at the end)

pelikhan:
Merged in https://github.com/Microsoft/pxt/pull/1052

Issue #1053, opened 2017-01-08
crash on array access to `(1<<31)`


mmoskal:
The following program crashes on the microbit, not in the simulator (when compiled in beta):
```
let arr: number[] =[]
for (let i = 0; i < 32; ++i) {
    arr.push(1 << i)
}
basic.showNumber(arr[31])
```

Issue #1054, opened 2017-01-08
crash on OOB array access


mmoskal:
The following program crashes on microbit and in the simulator, even though it's perfectly legal JavaScript:

```
let arr: number[] =[]
for (let i = 0; i < 32; ++i) {
    if (!arr[i])
        arr[i] = i
}
```

cc @pelikhan @tballmsft - comments?

Related issue: #1053

abchatra:
Adding default values to arrays should solve these issues.

abchatra:
Also a[I] access above is beyond the length of the array. Do we want to support default values for these as well?

pelikhan:
Yes we should have a runtime that is resilient to invalid input.

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

mmoskal:
I think the primary goal should be to stay true to JavaScript semantics, which also happens to be resilient to invalid input.

Of course we're still gonna crash (OOM) if someone does `arr[100000] = 1`...

abchatra:
Fixed

PR #1055, opened 2017-01-09
Default values for array

abchatra:
This adds default values for 
- Missing values are no more error instead default value is returned.
- Out of bound access for array are no more error, default value is returned.

Default values are 
- number[]->0
- string[]->null
- boolean[]->false
- object[]->null 

Related change in pxt-microbit: https://github.com/Microsoft/pxt-microbit/pull/335

Fixes https://github.com/Microsoft/pxt/issues/1053 & https://github.com/Microsoft/pxt/issues/1054

### DIFF
--- a/pxtsim/libgeneric.ts
+++ b/pxtsim/libgeneric.ts
@@ -2,10 +2,14 @@ namespace pxsim {
     // A ref-counted collection of either primitive or ref-counted objects (String, Image,
     // user-defined record, another collection)
     export class RefCollection extends RefObject {
-        data: any[] = [];
+        private data: any[] = []; 
+        //undefiend or null values need to be handled specially to support default values
+        //default values of boolean, string, number & object arrays are respectively, false, null, 0, null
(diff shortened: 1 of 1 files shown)


pxtsim/libgeneric.ts
```diff
@@ -21,7 +25,64 @@ namespace pxsim {
...
+            return ret[0]; //return the deleted element.
+        }
+
+        indexOf(x: number, start: number) {
+            if (x != 0) {
+                return this.data.indexOf(x, start);
+            }
+            //As we treat undefined same as 0 which is default value for all the arrays, will need to search both.
+            let defaultValueIndex = this.data.indexOf(x, start);
+            let undefinedIndex = this.data.indexOf(undefined, start);
```
mmoskal: `indexOf(undefined)` doesn't work:

```
> let b = [1]; b[7]= 13; b.indexOf(undefined)
< -1
```

I think it's best to just do a loop.

Issue #542, opened 2016-10-21
Deleting a package in pxt.json does not remove it from the project


riknoll:
If you edit the text of pxt.json, package changes do not show up until you refresh the page. Deleting a package using the trash button immediately causes the package to disappear without a refresh.

PR #1071, opened 2017-01-10
Project settings

microsoftsam:
Changing rename to project settings in "More" dropdown menu 
- adding save button in settings
- adding delete button in settings
- disabling auto save

screen shot 2017-01-10 at 2 51 19 pm

### DIFF
--- a/webapp/src/pxtjson.tsx
+++ b/webapp/src/pxtjson.tsx
@@ -13,6 +13,7 @@ const lf = Util.lf
 export class Editor extends srceditor.Editor {
     config: pxt.PackageConfig = {} as any;
     isSaving: boolean;
+    changeMade: boolean = false;
 
     prepare() {
         this.isReady = true
@@ -24,10 +25,6 @@ export class Editor extends srceditor.Editor {
 
     display() {
         const c = this.config
-        const updateName = (v: any) => {
-            this.config.name = v;
-            this.parent.forceUpdate()
-        }
         const save = () => {
             this.isSaving = true;
             const f = pkg.mainEditorPkg().lookupFile("this/" + pxt.CONFIG_NAME);
@@ -37,7 +34,12 @@ export class Editor extends srceditor.Editor {
                 this.parent.forceUpdate()
                 Util.nextTick(this.changeCallback)
                 this.isSaving = false;
+                this.changeMade = true;
             })
+        }
+        const setFileName = (v: string) => {
+            c.name = v;
+            this.parent.forceUpdate();
         }
         const deleteProject = () => {
             this.parent.removeProject();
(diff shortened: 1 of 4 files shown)


webapp/src/pxtjson.tsx
```diff
@@ -68,12 +77,12 @@ export class Editor extends srceditor.Editor {
...
             }
             // trigger update            
-            update(uc);
+            save();
         }
         return (
             <div className="ui content">
                 <div className="ui segment form text" style={{ backgroundColor: "white" }}>
-                    <sui.Input label={lf("Name")} value={c.name} onChange={v => update(c.name = v)} />
+                    <sui.Input label={lf("Name")} value={c.name} onChange={(v) => updateName(c.name = v)} />
```
pelikhan: we should not use the onChange, it is broken for script languages. Can we do without it?

webapp/src/pxtjson.tsx
```diff
@@ -82,7 +91,9 @@ export class Editor extends srceditor.Editor {
                             onChange={() => applyUserConfig(uc) } />
                     ) }
                     <sui.Field>
+                        <sui.Button text={lf("Save")} class={`green ${this.isSaving ? 'disabled' : ''}`} onClick={() => save()} />
                         <sui.Button text={lf("Edit Settings As text") } onClick={() => this.parent.editText() } />
+                        <sui.Button text={lf("Delete")} class={`red right floated`} onClick={() => deleteProject()} />
```
pelikhan: We do have a "delete project" button in the menu, why are we adding it again?

webapp/src/pxtjson.tsx
```diff
@@ -23,15 +24,23 @@ export class Editor extends srceditor.Editor {
...
+        }
+        const save = () => {
+            this.isSaving = true;
             const f = pkg.mainEditorPkg().lookupFile("this/" + pxt.CONFIG_NAME);
             f.setContentAsync(JSON.stringify(this.config, null, 4) + "\n").then(() => {
                 pkg.mainPkg.config.name = c.name;
                 this.parent.setState({projectName: c.name});
                 this.parent.forceUpdate()
                 Util.nextTick(this.changeCallback)
+                this.isSaving = false;
```
pelikhan: when done saving, we should reload the project

webapp/src/pxtjson.tsx
microsoftsam: I think we can use ref, and extract the value of the input field on save.

webapp/src/pxtjson.tsx
microsoftsam: Thought it made sense in the Settings page. Up for discussion.

pelikhan: We should avoid having multiple ways of doing the same thing. IMO.

abchatra: May be remove from the menu? Delete makes more sense Project settings view.

pelikhan: Really hard to discover where the "delete project" button is when only in settings.

microsoftsam: Removing from settings.

microsoftsam:
Fixes #542

microsoftsam:
screen shot 2017-01-13 at 10 21 03 am

PR #1081, opened 2017-01-11
Fix for music jack image in firefox

### LINKED ISSUE #1039: jack icon doesn't show in Firefox/OS X
On both /beta and pxt.microbit.org at the moment, I don't get a jack icon when using the audio blocks

microsoftsam:
fixes #1039


pelikhan:
XML namespace, computer scientist self inflicted hell

PR #1087, opened 2017-01-12
Control event value and event timestamp

### LINKED ISSUE #395: implement pxsim.control.eventValue and pxsim.control.eventTimestamp in the simulator
Currently the simulator build complains about it, and we have blocks using this.

chmora:
Helper methods to enable: 
pxsim.control.eventValue and pxsim.control.eventTimestamp

Note: using performance.now() in order to get microsecond precision since start time

Fixes #395

@mmoskal please review

### DIFF
--- a/pxtsim/runtime.ts
+++ b/pxtsim/runtime.ts
@@ -34,16 +34,14 @@ namespace pxsim {
         }
 
         export function now(): number {
-            return Date.now();
-        }
-
-        export function nowUs(): number {
-            return (performance as any).now() ||
-                   (performance as any).mozNow    ||
-                   (performance as any).msNow     ||
-                   (performance as any).oNow      ||
-                   (performance as any).webkitNow ||
-                    Date.now();
+            if(typeof performance != "undefined")
+              return (performance as any).now() ||
+                     (performance as any).mozNow()    ||
+                     (performance as any).msNow()     ||
+                     (performance as any).oNow()      ||
+                     (performance as any).webkitNow() ||
+                     Date.now();
+            return process.hrtime();
         }
 
         export function nextTick(f: () => void) {
@@ -234,7 +232,6 @@ namespace pxsim {
         dead = false;
         running = false;
         startTime = 0;
-        startTimeUs = 0;
         id: string;
         globals: any = {};
         currFrame: StackFrame;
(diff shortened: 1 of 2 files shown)


pxtsim/runtime.ts
```diff
@@ -37,6 +37,15 @@ namespace pxsim {
             return Date.now();
         }
 
+        export function nowUs(): number {
+            return (performance as any).now() ||
```
mmoskal: You need to do `if (typeof performance != "undefined")` ... for node.js

For node you could also do `process.hrtime()`

pxtsim/runtime.ts
```diff
@@ -37,6 +37,15 @@ namespace pxsim {
...
+                   (performance as any).mozNow    ||
```
mmoskal: Don't you need mozNow() with ()?

pxtsim/runtime.ts
```diff
@@ -225,6 +234,7 @@ namespace pxsim {
         dead = false;
         running = false;
         startTime = 0;
+        startTimeUs = 0;
```
mmoskal: Do we need a separate field or can just reuse startTime?

pxtsim/runtime.ts
chmora: When I try to use process.hrtime() it says  " Cannot find name 'process' ", What should I do?

pxtsim/runtime.ts
```diff
@@ -34,16 +34,14 @@ namespace pxsim {
         }
 
         export function now(): number {
-            return Date.now();
```
mmoskal: This is bad idea. performance.now() returns time since web navigation.

Keep a separate function, and maybe name it `perfNow()`.

microsoftsam: Makes sense. But is it okay if we override startTime's value with perfNow instead or is something else using it?

pelikhan: Rename it to startTimeUs and you'll know if anyone uses it.

pxtsim/runtime.ts
```diff
@@ -38,12 +38,13 @@ namespace pxsim {
...
-                       (performance as any).msNow      ||
-                       (performance as any).oNow       ||
-                       (performance as any).webkitNow  ||
-                       Date.now;
+            const perf = typeof performance != "undefined" ?
+                        performance.now ||
+                        (performance as any).mozNow    ||
+                        (performance as any).msNow     ||
+                        (performance as any).oNow      ||
+                        (performance as any).webkitNow :
```
microsoftsam: It is still possible (although unlikely) that performance is defined and non of the methods we listed work, ie: now, mozNow, etc all undefined. Which means right at the end we should still have Date.now as a fall back there.

PR #1088, opened 2017-01-12
Asmoutput

mmoskal:
keep the assembly file in built/ in explorer - it is only generated when user hits "Download" and is usually overwritten by the autorun very quickly, so it's impossible to see it


pelikhan:
Linking issue...

pelikhan:
Linting

PR #1102, opened 2017-01-13
Support for default classes in OOBlocks

riknoll:
Adds the ability to specify a default instance for object-oriented blocks so that the ability to change the instance being referenced in the blocks is hidden behind a mutator. Right now the input is tacked onto the end of the block due to limitations in Blockly.


pelikhan:
LGTM.

PR #1145, opened 2017-01-14
Undo / Redo buttons grey out if the undo / redo stack is empty.

### LINKED ISSUE #1139: grey out undo/redo buttons
When you get to the end of the history buffer (for undo/redo), the appropriate button should be greyed out.

microsoftsam:
screen shot 2017-01-13 at 4 27 06 pm

Fixes #1139


microsoftsam:
Note, only in Blockly. Monaco might be a little more complicated.

PR #1150, opened 2017-01-14
Fixing perfnow issue

microsoftsam:


pxtsim/runtime.ts
```diff
@@ -38,14 +38,17 @@ namespace pxsim {
         }
 
         export function perfNow(): number {
-            const perf = typeof performance != "undefined" ?
-                        performance.now                ||
+            if (typeof window === 'undefined') return Date.now();
+            let perf = window.performance || {now};
+            perf.now = () => {
```
abchatra: Not sure this is right. You are overriding the window.performance.now here?

Issue #976, opened 2016-12-21
undo do not undo shuffle blocks


OwenBrotherwood:
undo do not undo shuffle blocks

![shuffle-undue gif](https://cloud.githubusercontent.com/assets/866122/21399859/babd6b00-c7ad-11e6-8c30-ee3dbb0235ee.gif)

OwenBrotherwood:
https://github.com/Microsoft/pxt/pull/1168#issuecomment-273987024

OwenBrotherwood:
The solution seems to be a warning for cannot undo which is ok for me

microsoftsam:
Cutting the feature altogether

OwenBrotherwood:
Yep: let's get RC done
You are up early :) or late dependent on time zone

PR #1168, opened 2017-01-17
Fix "Undo do not undo shuffle blocks"

chmora:
fixes Microsoft/pxt#976


OwenBrotherwood:
I go for Peli's simple: just need a big warning as the placement of shuffle blocks and format code are just to darn close

microsoftsam:
Closing PR. Let's stick with a simpler solution that just warns the user that it's an undoable operation.

PR #1174, opened 2017-01-19
Lazy load the monaco editor

### LINKED ISSUE #242: delay load Monaco editor (or Blockly)
Only one editor needs to be loaded. Delay load the other (especially Monaco, which is quite large).

microsoftsam:
Only load the monaco editor when pxt is trying to open the file

Before (Memory usage): 
screen shot 2017-01-18 at 4 46 12 pm

After (Memory usage):
screen shot 2017-01-18 at 4 46 19 pm

Fixes #242

### DIFF
--- a/webapp/src/blocks.tsx
+++ b/webapp/src/blocks.tsx
@@ -407,34 +407,32 @@ export class Editor extends srceditor.Editor {
     }
 
     hasUndo() {
-        Util.assert(this.editor != undefined); // Guarded
-        return this.editor.undoStack_.length != 0;
+        return this.editor ? this.editor.undoStack_.length != 0 : false;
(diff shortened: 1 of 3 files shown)


webapp/src/blocks.tsx
```diff
@@ -402,34 +399,42 @@ export class Editor extends srceditor.Editor {
...
     }
 
     undo() {
+        Util.assert(this.editor != undefined); // Guarded
         this.editor.undo();
         this.parent.forceUpdate();
     }
 
     hasRedo() {
+        Util.assert(this.editor != undefined); // Guarded
```
pelikhan: just bail out when not available.

PR #1195, opened 2017-01-20
Fix index error when counting default instance args

riknoll:
Not sure if this should go to master or develop.


pelikhan:
Looks great. Is that something we want in master?

PR #1200, opened 2017-01-20
Don't auto add packages if they cause conflicts; Prompt for conflicts BEFORE adding a package

### LINKED ISSUE #441: Removing conflicting packages results in the possibility of stray blocks
If we try to add the bluetooth package, when radio is installed, we are prompted with the need to remove one of them. 
If we have existing radio blocks and remove the radio package. We're left with blocks that don't work, and don't compile to typescript. 
Refreshing the page is also an issue, as it results in black unknown blocks (that look like a bug to the user). 
image

Maybe we should remove related blocks when removing a (conflicting) package.

### LINKED ISSUE #905: Addition of incompatible libraries gives non-intuitive/not-expected dialog "suggestion"
In connection with addition of an incompatible library to a project, the requester has a suggestion that the correct answer is remove the library one has requested be added.

"Suggestion" is that the library to be added has a greater weight than the library that would be needed to be removed.
image

### LINKED ISSUE #1119: Incompatible packages warning is unhelpful and recurring when blocks from a package are already used
Using pxt.microbit.org/beta I see the following issue:

Add some blocks that use radio to a program
try to add the bluetooth package
You get the following dialogue
image
Choosing 'Remove Radio' doesn't work, it just respawns the dialogue. Choosing 'Remove Bluetooth' ends the respawn loop but doesn't help the user understand the problem
remove radio

Instead I think it would be better to explain to the user after the first 'Remove Radio' that this cannot be done because there are radio blocks in the program. The user can then remove them and try again
(This works fine if there are no radio/bluetooth blocks already there)

Also the little message that mentions yotta settings being incompatible seems a bit too much detail? It is only there a little while, though..

### LINKED ISSUE #1159: Incompatible packages dialog is unclear about repercussions
Its unclear from the incompatible packages dialog text that you'll lose any references to radio. I think it needs to be clear that if you are using any radio blocks / references you will lose those by removing 'Radio'. 

screen shot 2017-01-17 at 9 40 46 am

Otherwise you can find yourself with empty blocks, not knowing why.

### LINKED ISSUE #1232: Add Package Feature does not allow Blue Tooth to be substituted by Radio
microbit.org-ticket: 815

When trying the "Add Package" Feature I had Radio on the primary list and asked for Blue Tooth. I was told that either Blue Tooth of Radio had to be removed. I selected Blue Tooth but this would not be accepted and the Remove window came up again.  The only way I could get out of this loop was to accept Blue Tooth.  

I then tried to "Add Package" - Radio but as before, this was not allowed.  In fact the Remove Blue Tooth option is a lighter colour as though it had been greyed out.  

I signed out and back in, using microbit org  PTX but got a conflict message that was too fast to record but was along the lines   " Conflict on yotta settings microbit.bluetooth enabled etc ...."

The screen I was left with was mostly blank and unusable except I tried Getting Started which gives me the getting started editor which may work.

### LINKED ISSUE #1236: yotta conflict on aborted add package
yottaconflict

guillaumejenkins:
Fixes #441
Fixes #905
Fixes #1119
Fixes #1159
Fixes #1232
Fixes #1236

### DIFF
--- a/pxtlib/main.ts
+++ b/pxtlib/main.ts
@@ -189,15 +189,33 @@ namespace pxt {
     }
 
     export class Package {
+        static getConfigAsync(id: string, fullVers: string): Promise<pxt.PackageConfig> {
+            return Promise.resolve().then(() => {
+                if (pxt.github.isGithubId(fullVers)) {
+                    const repoInfo = pxt.github.parseRepoId(fullVers);
+                    return pxt.packagesConfigAsync()
+                        .then(config => pxt.github.repoAsync(repoInfo.fullName, config))    // Make sure repo exists and is whitelisted
+                        .then(gitRepo => gitRepo ? pxt.github.pkgConfigAsync(repoInfo.fullName, repoInfo.tag) : null);
+                } else {
+                    // If it's not from GH, assume it's a bundled package
+                    // TODO: Add logic for shared packages if we enable that
+                    return JSON.parse(pxt.appTarget.bundledpkgs[id][CONFIG_NAME]) as pxt.PackageConfig;
+                }
+            });
+        }
+
+        public addedBy: Package[];
         public config: PackageConfig;
         public level = -1;
         public isLoaded = false;
         private resolvedVersion: string;
 
-        constructor(public id: string, public _verspec: string, public parent: MainPackage) {
+        constructor(public id: string, public _verspec: string, public parent: MainPackage, addedBy: Package) {
             if (parent) {
                 this.level = this.parent.level + 1
             }
+
+            this.addedBy = [addedBy];
         }
 
         version() {
@@ -270,10 +288,9 @@ namespace pxt {
                     return this.host().downloadPackageAsync(this)
                         .then(() => {
                             const confStr = this.readFile(pxt.CONFIG_NAME)
-                            const mainTs = this.readFile("main.ts")
                             if (!confStr)
                                 U.userError(`package ${this.id} is missing ${pxt.CONFIG_NAME}`)
-                            this.parseConfig(confStr, mainTs);
+                            this.parseConfig(confStr);
                             if (this.level != 0)
                                 this.config.installedVersion = this.version()
                             this.saveConfig()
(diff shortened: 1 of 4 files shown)


pelikhan:
Could we try to get some test cases for this logic? It's getting quite intricate.

webapp/src/app.tsx
```diff
@@ -260,14 +258,71 @@ class ScriptSearch extends data.Component<ISettingsProps, ScriptSearchState> {
...
+                                return false;
+                            });
+                    } else if (conflicts.length) {
+                        const conflictNames = conflicts.map((c) => c.pkg0.id).join(",<br/>");
+                        addDependencyPromise = addDependencyPromise
+                            .then(() => core.confirmAsync({
+                                header: lf("Packages will be removed"),
+                                agreeLbl: lf("Remove packages and add {0}", config.name),
+                                agreeClass: "pink",
+                                htmlBody: `<p>${lf("The following package(s) are incompatible with {0} and will be removed. Continue?", config.name, conflictNames)}<br/><br/>${conflictNames}`
```
pelikhan: Use "body" or escape all user provided strings.

pxtlib/main.ts
```diff
@@ -189,15 +189,33 @@ namespace pxt {
     }
 
     export class Package {
+        static getConfigAsync(id: string, fullVers: string): Promise<pxt.PackageConfig> {
+            const ghConfig = /^github:(.*)#(.*)$/.exec(fullVers);
+            return Promise.resolve().then(() => {
+                if (ghConfig) {
+                    return pxt.packagesConfigAsync()
+                        .then(config => pxt.github.latestVersionAsync(ghConfig[1], config))
```
pelikhan: not sure why we call lastestVersionAsync since we do not use ``tag`` afterwards

pxtlib/main.ts
```diff
@@ -310,13 +348,84 @@ namespace pxt {
...
+                                ancestors.push.apply(allAncestors(a));
+                                ancestors.push(a);
+                            }
+                        });
+                        return ancestors;
+                    }
+                    const additionalConflicts: cpp.PkgConflictError[] = [];
+                    conflicts.forEach((c) => {
+                        additionalConflicts.push.apply(additionalConflicts, allAncestors(c.pkg0).map((anc) => {
+                            const confl = new cpp.PkgConflictError("");
```
pelikhan: no details?

pxtlib/main.ts
```diff
@@ -310,13 +348,84 @@ namespace pxt {
...
+                    conflicts = conflicts.filter((c, index) => {
+                        for (let i = 0; i < index; ++i) {
+                            if (c.pkg0.id === conflicts[i].pkg0.id) {
+                                return false;
+                            }
+                        }
+                        return true;
+                    });
+                })
+                .then(() => conflicts);
```
pelikhan: when `.then(() `` here?

webapp/src/app.tsx
```diff
@@ -260,14 +258,71 @@ class ScriptSearch extends data.Component<ISettingsProps, ScriptSearchState> {
...
+
+                    let addDependencyPromise = Promise.resolve(true);
+
+                    if (inUse) {
+                        addDependencyPromise = addDependencyPromise
+                            .then(() => core.confirmAsync({
+                                header: lf("Cannot add {0} package", config.name),
+                                hideCancel: true,
+                                agreeLbl: lf("Ok"),
+                                htmlBody: lf("Remove all the blocks from the {0} package and try again.", inUse.pkg0.id)
```
pelikhan: Use "body" instead.

pelikhan:
image

Make it a single sentence: Packages ... are incompatiable with blah blah blah...

pelikhan:
The simulator does not seem to restart after adding a package.

pelikhan:
We should try to add end-to-end integration. Maybe you can add a CLI command that allows to add a package (eg "pxt install --save") so that we can automate test cases for the logic.

PR #1201, opened 2017-01-22
Mute simulator using the simulator toolbar

microsoftsam:
screen shot 2017-01-21 at 4 57 30 pm

screen shot 2017-01-21 at 4 57 42 pm

screen shot 2017-01-21 at 4 57 48 pm

Split the sim toolbar into two parts (as you can see in the screenshot)

### DIFF
--- a/pxtsim/simlib.ts
+++ b/pxtsim/simlib.ts
@@ -128,6 +128,8 @@ namespace pxsim {
         let _vco: any; // OscillatorNode;
         let _vca: any; // GainNode;
 
+        let _mute = false; //mute audio
+
         function context(): any {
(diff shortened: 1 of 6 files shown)


pxtsim/simlib.ts
```diff
@@ -145,6 +147,10 @@ namespace pxsim {
             return undefined;
         }
 
+        export function mute(mute: boolean) {
+            _mute = mute;
```
pelikhan: call 'stop' ?

pelikhan:
Looks great. When receiving a mute message, should you call stop() in case the AudioManager is already ringing a tone?

PR #1206, opened 2017-01-24
Support for specifiying images for enums in blockly

pelikhan:
Requires blockly update that supports images.


microsoftsam:
I'll update blockly before merging this in.

Issue #1190, opened 2017-01-20
IE11 opening one of the example projects does not launch a new tab or open drawer


riknoll:
When opening from the projects modal. It should launch a new window with the project instructions.

PR #1211, opened 2017-01-25
Don't load sample projects in IE11

riknoll:
Fixes https://github.com/Microsoft/pxt/issues/1190

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -166,6 +166,7 @@ class ScriptSearch extends data.Component<ISettingsProps, ScriptSearchState> {
         if (this.state.mode != ScriptSearchMode.Projects
             || sandbox
             || this.state.searchFor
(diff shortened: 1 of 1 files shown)


webapp/src/app.tsx
```diff
@@ -350,7 +350,7 @@ class ScriptSearch extends data.Component<ISettingsProps, ScriptSearchState> {
                             onClick={() => chgHeader(scr) }
                             />
                     ) }
-                    {galleries.map(scr => <codecard.CodeCardView
+                    {pxt.options.light ? undefined : galleries.map(scr => <codecard.CodeCardView
```
pelikhan: You are still paying the price of fetching the galleries form the web in "fetchGalleries". You should shortcut there.
Projects refactoring is long overdue...

riknoll: Fixed

PR #1213, opened 2017-01-25
Upgrading the Monaco editor to version 0.8.1.

microsoftsam:
Bringing in dragging and dropping in monaco toolbox.

monacoeditordrag

Do not merge until the following are complete:
- [ ] Change is tested on IE, Edge and Chrome on Windows
- [ ] Editor not loading in Safari 9.0, fix or workaround or don't support. monaco-editor/issues/327


microsoftsam:
We should wait for 8.1.0 which should be releasing by the end of the week according to Alex.

microsoftsam:
With 8.1, we've fixed the issue with the context menu showing all the buttons for all editors. If it's read only, save, and run don't show. 
Same thing with help. You need to be selecting a word for help (reference) to show in the context menu.

Issue #652, opened 2016-11-02
clearer error condition when non-ascii characters are used as variables


OwenBrotherwood:
Enhancement: clearer error condition when non-ascii characters are used as variables
Very much "up" is the translation process of pxt.
And some are even using non-ascii for variables: why not?
Of course, the "backend" doesn't like it.

This is a true example from the real world of teaching: the old block editor did not protest and the sim ran. The program did not run on the micro:bit.

Example screen dump of todays pxt (possibly a few comitts behind.
![variables](https://cloud.githubusercontent.com/assets/866122/19936262/35f2efe6-a11d-11e6-9d43-b1eca71668d6.jpg)

pelikhan:
A triangle poped up on the set ... til block, it should have opened up when selecting the block.

OwenBrotherwood:
It gets worse, sorry
I continued with show string (where there was no triangle) 
and added value, then triangle came on show string
<img width="310" alt="value ae" src="https://cloud.githubusercontent.com/assets/866122/19942729/0495497c-a135-11e6-8df6-3362bbcb562c.PNG">

OwenBrotherwood:
The error: text is too long

![image](https://cloud.githubusercontent.com/assets/866122/22137858/e11dabba-dedd-11e6-9c19-0309b5211ff9.png)

OwenBrotherwood:
Add show string to canvas
Add variable "æøå" to show string
"Unable to convert" dialog
There is no blue triangle
![image](https://cloud.githubusercontent.com/assets/866122/22137919/45d7b870-dede-11e6-80e5-0cdacd7f5dfc.png)

OwenBrotherwood:
heads up to me when need for tests so that the issue can be closed

OwenBrotherwood:
I have some test ready when "this" is available in /beta

PR #1216, opened 2017-01-26
Rename variables with entirely non-ascii names

riknoll:
Fixes https://github.com/Microsoft/pxt/issues/652

Rename variables with completely non-ascii names to `myVariable` (or `myVariable2`, `myVariable3`, etc.). This matches our current strategy of throwing out non-ascii characters when going to TS but allowing anything in the blocks.

### DIFF
--- a/pxtblocks/blocklycompiler.ts
+++ b/pxtblocks/blocklycompiler.ts
@@ -890,7 +890,10 @@ namespace pxt.blocks {
             //.map((c, i) => (i ? c[0].toUpperCase() : c[0].toLowerCase()) + c.substr(1)) breaks roundtrip...
             .join('');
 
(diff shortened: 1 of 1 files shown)


pelikhan:
Could we take this opportunity to address the issue of non-ascii languages? For example, this won't work for kids writing in Japanese/Arabic. We should escape to a subset of letters that produces valid JavaScript identifiers.

riknoll:
@pelikhan escape how?

pelikhan:
TypeScript supports unicode characters. Is there a specification of identifiers in TypeScript, we should escape according to that spec.

image

riknoll:
@pelikhan sure, I'll take a look.

riknoll:
Closing for now

PR #1234, opened 2017-01-28
Download flags

pelikhan:
URL flags to force various download strategies for mobile testing.


microsoftsam:
What do the flags do?

pelikhan:
- downloadWindowOpen uses window.open to try to generate a download
- downloadProtocol provides custom protocols

PR #1241, opened 2017-01-30
Adding support for subcategories

riknoll:
In addition to the common "More" category.


pelikhan:
Why do we need those?

pelikhan:
We should guarantee a constant ordering in sub-categories.

pelikhan:
Feels like we should simply use the namespace information.

riknoll:
@pelikhan I agree on having a constant ordering, but using the namespace information makes the signature pretty ugly:

```typescript
light.builtin.showAnimationFrame(light.animation.sparkleAnimation());
```

Kids might find that somewhat confusing considering `light.builtin` is an object and `light.animation` is a namespace.

pelikhan:
```light.animation.sparkle()``` does not sound too bad. we can support that later.

File: pxtlib/emitter/emitter.ts [hunk 1]
```diff
@@ -301,6 +301,7 @@ namespace ts.pxtc {
         parts?: string;
         trackArgs?: number[];
         advanced?: boolean;
+        subcategory?: string;
```
pelikhan: field for this interface typically use a camelized notation.

pelikhan: otherwise LGTM

riknoll: subcategory is one word

riknoll:
I'm adding support for specifying subcategory weights, so hold off on merging

pxtlib/emitter/emitter.ts
pelikhan: goodToGoThen!

riknoll:
@pelikhan how about this:

1. Subcategories specified by the parent package always come first and are weighted by the package
2. Subcategories from other packages come next and are sorted alphabetically
3. "More" always goes last

pelikhan:
Could we simply get rid of the built in ones? We just need to have a sufficient weight. This is how all the other categories work...

riknoll:
You mean sort them all alphabetically? "More" should remain as is because it should be consistent across parent categories.

PR #1273, opened 2017-02-01
Disabling var in the typescript editor 

abchatra:
We don't fully adhere to javascript var semantics. This change disables var as the way to define a variable and will throw an error requiring let or const.
Motivated by #744
The real issue in #744 is typescript binds to one global variable, but we implement let semantics for var which contradicts.


abchatra:
Also see https://github.com/Microsoft/pxt/issues/744

PR #1292, opened 2017-02-04
Hide the menu bar using a flag

microsoftsam:
hide the menu bar under a flag 
don't show the add package button when packages are not supported

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -633,7 +633,7 @@ namespace pxt.blocks {
         }
 
         // Add the "Add package" category
-        if (tb && showCategories && pxt.appTarget.cloud.packages) {
+        if (tb && showCategories && pxt.appTarget.cloud && pxt.appTarget.cloud.packages) {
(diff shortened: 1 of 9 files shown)


pxtblocks/blocklyloader.ts
```diff
@@ -633,7 +633,7 @@ namespace pxt.blocks {
         }
 
         // Add the "Add package" category
-        if (tb && showCategories) {
+        if (tb && showCategories && pxt.appTarget.cloud.packages) {
```
pelikhan: cloud might be undefined

webapp/src/app.tsx
```diff
@@ -1213,6 +1225,8 @@ export class ProjectView
         const run = true; // !compileBtn || !pxt.appTarget.simulator.autoRun || !isBlocks;
         const restart = run && !simOpts.hideRestart;
         const fullscreen = run && !simOpts.hideFullscreen;
+        const showMenuBar = !targetTheme.layoutOptions.hideMenuBar;
```
pelikhan: layoutOptions might be undefined

PR #1327, opened 2017-02-09
Storecontinuations

tballmsft:
I've generated, for each function, an array of the continuations in that function (JS only). I've also opened up the runtime to allow the simulator code in a target to overwrite the continuation. This is needed for my thumb simulator.


pxtlib/emitter/backjs.ts
```diff
@@ -127,7 +128,7 @@ switch (step) {
         let info = nodeLocationInfo(proc.action) as FunctionLocationInfo
         info.functionName = proc.getName()
         writeRaw(`${proc.label()}.info = ${JSON.stringify(info)}`)
-
+        writeRaw(`${proc.label()}.continuations = [ ${async_continuations.join(",")} ]`)
```
pelikhan: do we need to write this if ``async_continuations`` is empty? (note: js typeically uses camelized variable names)

pxtsim/runtime.ts
```diff
@@ -222,13 +222,16 @@ namespace pxsim {
...
+
         dead = false;
         running = false;
         startTime = 0;
         id: string;
         globals: any = {};
         currFrame: StackFrame;
+        entry: LabelFn;
 
+        overwriteResume: (retPC: number) => void;
```
pelikhan: is this used anywhere?

tballmsft: yes, it is used in my target!

pxtlib/emitter/backjs.ts
tballmsft: yes, but I only need to generate the array for the main function. I will revise

tballmsft: I tested it. No issues. pxt-microbit does not use my changes.

Issue #951, opened 2016-12-19
Error with code generated for ternary expressions in function parameters


riknoll:
The following code will show the number `1` when run in the simulator or on a micro:bit:

```typescript
let x = 0;

function someFunction(arg1: boolean, arg2: number) {
    basic.showNumber(arg2)
}

someFunction(true, x === 1 ? 5 : -5)
```

If you remove the first argument (`arg1`) from the function, then the code runs correctly and displays `-5`.

I'm going to mark this as P2 because it doesn't affect the blocks.

riknoll:
Passing false returns will make it display the number `0`.

riknoll:
Here is a better example of the issue:
```
let x = 0;

basic.showNumber((x === 1 ? 5: 5) + (x === 1 ? -3: 3))
```

This will show the number `6` when it should show `8`.

abchatra:
`basic.showNumber((2 + 2) + (true ? -3: 3))` expected result here is 1

abchatra:
fixed

PR #1329, opened 2017-02-10
Conditional operator was not saving the temporary value

abchatra:
Simple fix to the issue https://github.com/Microsoft/pxt/issues/951


riknoll:
LGTM

PR #1343, opened 2017-02-13
avoiding 1rem in Edge to avoid blockly bug

pelikhan:
Fixes font overlap issues in Edge.

Tested in Windows Chrome, FF, Edge, IE11. Looking good.


pelikhan:
@microsoftsam please take a look.

PR #1346, opened 2017-02-14
Bundled npm modules

pelikhan:
Allow to specify a NPM module folder to be bundled in the target build process (as opposed to having it under /libs)


pelikhan:
Actually, we can simply use "bundleddirs".

PR #1362, opened 2017-02-17
Always decompile!

### LINKED ISSUE #1271: When a typescript construct is not understand, bundle in read-only block
When the decompiler detects that a typescript construct cannot be represented in blocks, generate special block that contains raw code.

- [ ] support for encapsulating statements

Support for sub-expression is not planned in this version.

### LINKED ISSUE #1345: Support for ArrayLiteralExpression
image

riknoll:
Fixes #1345
Fixes #1271

This PR makes it so that we always decompile from javascript to blocks, except when there is a compile error in the javascript. As a result, I've removed the decompiler error test suite and added them as positive test cases.

We decompile at the statement level and the expression level where possible. Another result of this is that decompilation no longer allows duplicate variable names in the code (because all variables in Blockly are global). In other words, this code:

``` typescript
let x = 0 & 9;

while ((x & 7) === 0) {
    x++;
}

for (;;) {
    let x = 0;
    x ++;
}

```

Will decompile to this:

always_decompile

Note that the second `x`'s name was changed to `y`. We already did some renaming in the past, this just makes it more extensive.

The grey blocks work just like any other blocks. They can be moved, deleted, copy-pasted, etc. 

Please test this extensively, because it is a big rewrite.


abchatra:
I am doing testing and quick comment on the UI. Please add some spaces at the right end of the column. It seems cut off for me. 
image
image

abchatra:
Also please add help to these blocks.

pxtblocks/blocklycompiler.ts
```diff
@@ -1179,6 +1192,50 @@ namespace pxt.blocks {
         return mkBlock(stmts);
     }
 
+    // FIXME: Whenever we add unicode support, handle that here
+    const variableStatementRegex = /\s*(?:(?:let)|(?:const)|(?:var))\s+(.+)/
```
abchatra: >var [](start = 63, length = 3)

We don't support war anymore.

pxtblocks/blocklycompiler.ts
```diff
@@ -1179,6 +1192,50 @@ namespace pxt.blocks {
...
+    const declarationRegex = /\s*()/
```
abchatra: Where are the above two const variables used?

riknoll: Should have deleted those. I'll fix it!

abchatra:
Rest of the changes look good to me (Took a long time to go through decompiler.ts)

PR #1378, opened 2017-02-21
Make a variable button

### LINKED ISSUE #915: "Make a variable" item would be better as something other than a block
The "Make a Variable" block is not really a block, in that you can'd drag it into your program

So instead, it would make more sense as a button

image

Or at least, something else different from the blocks.
image

microsoftsam:
Removing color from the make a variable block to make it appear different to the other blocks under Variables, and highlight the fact that it's not a block (but a button). 

Before: 
screen shot 2017-02-21 at 10 31 05 am

After: 
screen shot 2017-02-21 at 10 37 50 am

Fixes #915

### DIFF
--- a/theme/pxt.less
+++ b/theme/pxt.less
@@ -498,9 +498,6 @@ span.blocklyTreeLabel {
     font-size:1.25rem;
 }
 
-.blocklyFlyoutButton {
-    fill: rgb(165, 91, 128) !important;
-}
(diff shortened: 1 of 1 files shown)


microsoftsam:
Button changes state on mouse hover too.

abchatra:
Looks good to me. @jaustin  FYI

PR #1380, opened 2017-02-21
Blocklyintegrate PXT

chmora:


pxtblocks/blocklyloader.ts
```diff
@@ -378,7 +378,11 @@ namespace pxt.blocks {
         const ns = (fn.attributes.blockNamespace || fn.namespace).split('.')[0];
         const instance = fn.kind == pxtc.SymbolKind.Method || fn.kind == pxtc.SymbolKind.Property;
         const nsinfo = info.apis.byQName[ns];
-
+        const color =
+            fn.attributes.color
+            || (nsinfo ? nsinfo.attributes.color : undefined)
+            || blockColors[ns]
+            || 255;
```
microsoftsam: Is this being used in multiple places? is that why you're setting it to a field 'color'? If so, use block.setColor(color)

pxtlib/emitter/emitter.ts
```diff
@@ -302,6 +303,7 @@ namespace ts.pxtc {
         trackArgs?: number[];
         advanced?: boolean;
         deprecated?: boolean;
+        noteValue?: string; //  value for enum Note
```
microsoftsam: Do we need this?

PR #1419, opened 2017-02-27
implement resume without new stack frame

tballmsft:


pxtsim/runtime.ts
```diff
@@ -551,7 +555,12 @@ namespace pxsim {
             this.setupTop = setupTop
             this.handleDebuggerMsg = handleDebuggerMsg
             this.entry = entryPoint
-            this.overwriteResume = (retPC: number) => { currResume = null; setupResume(this.currFrame, retPC) }
+            this.overwriteResume = (retPC: number) => {
```
mmoskal: this seems never used, what is it for? @tballmsft

PR #1420, opened 2017-02-27
Blocks only workspace

microsoftsam:
Simple flag to remove the blocks / javascript toggle (for blocks only workspaces). 

To test: 

in pxtarget set ``appTheme.blocksOnly`` to ``true``


tballmsft:
Hmm... I was thinking of having no UI element at all. Some reason for keeping the button?

microsoftsam:
No reason to keep the button. I can change that.

pelikhan:
There's probably some arcane ways to still end up on the javascript side but it's probably ok as it is.

tballmsft:
Tested. Thanks, Sam!

PR #1423, opened 2017-02-27
Decompiler should handle property access expressions that aren't enums

riknoll:
No issue open for this one, but the problem was that Enums are treated as a special case for the decompiler and that meant that property access expressions on class members were not decompiled properly.


abchatra:
:shipit:

PR #1444, opened 2017-02-28
Number Slider integration

chmora:
image

Now we can use min and max range value to number fields just adding the range "[min-max]" after the parameter name in the declaration of the function.
IMPORTANT: it is necessary to set a shadow value because that's how the new container is created in blocklyloader

example:

image

In blocklyloader it appends the new blockly math_number_minmax block instead of math_number

image


pelikhan:
Screenshots?

microsoftsam:
Tested in monaco, that the min-max range don't affect the param jsdoc?

pxtblocks/blocklyloader.ts
```diff
@@ -119,15 +124,26 @@ namespace pxt.blocks {
...
             fn.parameters.filter(pr => !!attrNames[pr.name].name &&
                 (/^(string|number|boolean)$/.test(attrNames[pr.name].type)
                     || !!attrNames[pr.name].shadowType
                     || !!attrNames[pr.name].shadowValue))
                 .forEach(pr => {
                     let attr = attrNames[pr.name];
-                    block.appendChild(createShadowValue(attr.name, attr.type, attr.shadowValue, attr.shadowType));
+                    let shadowValue = createShadowValue(attr.name, attr.type, attr.shadowValue, attr.shadowType);
+                    let container: any;
+                    if (fn.attributes.min && fn.attributes.max) {
```
microsoftsam: Why is the container outside the if statement?

pxtlib/emitter/emitter.ts
```diff
@@ -321,6 +321,9 @@ namespace ts.pxtc {
         paramHelp?: pxt.Map<string>;
         // foo.defl=12 -> paramDefl: { foo: "12" }
         paramDefl: pxt.Map<string>;
+
+        min?: Map<string>; // min range
```
pelikhan: why is this a map?

chmora: I store the minimum and maximum values by name using a Map because when I create the shadow value I don't have the explicit position of the parameter in the parameters array of the block.

image

pelikhan: why not storing it on the parameter attributes?

pxtblocks/blocklyloader.ts
```diff
@@ -566,6 +583,17 @@ namespace pxt.blocks {
...
                     let pnames = parameterNames(fn);
+                    // use slider for min and max number
+                    if (fn.attributes.min && fn.attributes.max) {
+                        // change to numberMinMax only if min and max are defined
+                        for (let i = 0; i < fn.parameters.length; i++) {
+                            let pr = fn.parameters[i];
+                            if (fn.attributes.min[pr.name] && fn.attributes.max[pr.name]) {
+                                fn.parameters[i].type = "numberMinMax";
+                                pnames[pr.name].type = "numberMinMax";
+                            }
```
chmora: Here was necessary to change the block type, because in PXT a variable of type number automatically is linked with a math_number block.

pxtlib/emitter/emitter.ts
```diff
@@ -423,8 +426,17 @@ namespace ts.pxtc {
...
-            doccmt = doccmt.replace(/^\s*@param\s+(\w+)\s+(.*)$/mg, (full: string, name: string, desc: string) => {
+            doccmt = doccmt.replace(/^\s*@param\s+(\w+)\s+(\[.*\])?\s+(.*)$/mg, (full: string, name: string, type: string, desc: string) => {
                 res.paramHelp[name] = desc
+                if (type) {
+                    type.replace(/^\[([0-9]+)-([0-9]+)\]$/i, (full: string, min: string, max: string) => {
+                        if (!res.min) res.min = {}
+                        if (!res.max) res.max = {}
+                        res.min[name] = min
+                        res.max[name] = max
+                        return ""
```
chmora: I created a regex to identify the range values in the parameter declaration and store those values

pxtblocks/blocklyloader.ts
```diff
@@ -119,15 +124,25 @@ namespace pxt.blocks {
...
                 .forEach(pr => {
                     let attr = attrNames[pr.name];
-                    block.appendChild(createShadowValue(attr.name, attr.type, attr.shadowValue, attr.shadowType));
+                    let shadowValue = createShadowValue(attr.name, attr.type, attr.shadowValue, attr.shadowType);
+                    if (fn.attributes.min && fn.attributes.max) {
+                        if (fn.attributes.min[pr.name] && fn.attributes.max[pr.name]) {
+                            let container = document.createElement('mutation');
+                            container.setAttribute('min', fn.attributes.min[pr.name]);
+                            container.setAttribute('max', fn.attributes.max[pr.name]);
+                            shadowValue.firstChild.appendChild(container);
```
chmora: created a new container called 'mutation' where I put the min and max values for set the new constrains in the math_number_minmax Block. Those changes are applied in mutationToDom and domToMutation functions in the (pxt-blockly) math.js file in the math_number_minmax block creation.

pelikhan: couldn't just just do ``pr.attributes.min`` ?

chmora: I'm going to try that

chmora: The problem with that solution was that when I do the regex to extract Min and Max I don't have access to the parameters, I just know their names and description.

pxtlib/emitter/service.ts
```diff
@@ -8,6 +8,8 @@ namespace ts.pxtc {
         initializer?: string;
         defaults?: string[];
         properties?: PropertyDesc[];
+        min?: string; // min range value to use slider
```
microsoftsam: Is this still being used?

pelikhan:
Update docs at https://www.pxt.io/defining-blocks

PR #1477, opened 2017-03-01
Make sure globals initialized with `...` are treated just like "..."

mmoskal:
Otherwise, the ones with `...` were always treated as used


abchatra:
:shipit:

PR #1493, opened 2017-03-02
handle partial target definition

pelikhan:


abchatra:
Can you please add some description?

pelikhan:
Allows to build a dummy target like pxt-common-packages that is missing the "theme" part.

PR #1513, opened 2017-03-03
added documentation to include slider use

chmora:


pelikhan:
Why is it a separate PR?

PR #1531, opened 2017-03-04
Destructuring mutators should handle extra arguments

riknoll:
Deals with a bug where if you used the destructuring mutator (like in the radio packet) and you include extra arguments besides the object being destructured, the mutator would fail to load.


pxtblocks/blocklymutators.ts
```diff
@@ -66,9 +66,25 @@ namespace pxt.blocks {
 
         switch (mutationType) {
             case MutatorTypes.ObjectDestructuringMutator:
-                if (!info.parameters || info.parameters.length !== 1 || info.parameters[0].properties.length === 0) {
-                    console.error("Mutating blocks only supported for functions with one parameter that has multiple properties");
-                    return;
+                if (!info.parameters || info.parameters.length < 1) {
+                    console.error("Destructuring mutations require at least one parameter")
```
pelikhan: PXT error

pelikhan: Localize error message

pxtblocks/blocklymutators.ts
```diff
@@ -66,9 +66,25 @@ namespace pxt.blocks {
...
+                }
+                else {
+                    let found = false;
+                    for (const param of info.parameters) {
+                        if (param.type.indexOf("=>") !== -1) {
+                            if (!param.properties || param.properties.length === 0) {
+                                console.error("Destructuring mutations only supported for functions with an event parameter that has multiple properties");
```
pelikhan: PXT.error, loc message

pxtblocks/blocklymutators.ts
```diff
@@ -66,9 +66,25 @@ namespace pxt.blocks {
...
+                                return;
+                            }
+                            found = true;
```
pelikhan: break?

pxtblocks/blocklymutators.ts
riknoll: These are dev errors, not user errors. Only for package authors.

riknoll:
I can add a decompiler test case at least...

riknoll:
@pelikhan  expanding this PR to also include a new feature. This lets APIs set a flag so that when compiling to TS the destructured properties are also passed to the API in an array of enum values. For example, if I chose `arg1` and `arg2`, it would compile to something like this:

```typescript
ns.someCallback([MyArgs.arg1, MyArgs.arg2], ({arg1, arg2}) => {

})
```

That way, blocks can change runtime behavior based on the arguments selected by the user. This involves some complex decompile checking which I have not added a test case for because it is being blocked by an unrelated compiler bug. I have tested it locally and it works fine.

pelikhan:
Could you add docs about that new feature?

pelikhan:
LGTM

riknoll:
Adding docs now

pelikhan:
Do you need to cherry pick into v0?

riknoll:
No cherry-picking required

PR #1549, opened 2017-03-06
Apply faded colour to disabled blocks. 

### LINKED ISSUE #1537: Disabled blocks to keep their color
Use a transparency layer over the original block color for disabled blocks.

microsoftsam:
Fixes #1537

screen shot 2017-03-06 at 8 59 46 am

### DIFF
--- a/theme/pxt.less
+++ b/theme/pxt.less
@@ -621,6 +621,12 @@ span.blocklyTreeIcon {
     border-radius: 5px;
 }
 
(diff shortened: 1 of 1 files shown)


pelikhan:
Isn't this a fix for pxt-blockly?

microsoftsam:
Yeah I was tossing up where to put it. I guess this should be applied to the "general" case. Will move it.

PR #1550, opened 2017-03-06
Upload a board visual to the cdn

microsoftsam:


microsoftsam:
Setting sim manifest in the target instead.

Issue #1496, opened 2017-03-02
min/max roundtripping not working properly


DavidWhaleMEF:
microbit.org-ticket: 1203

The min/max in this code gets damaged (both change to min or both change to max) when roundtripping back to blocks:
```
let posn = 0
let init = 0
let level = 0
input.onButtonPressed(Button.A, () => {
    if (init == 1) {
        level = Math.max(1, level - 1)
        basic.showNumber(level)
    } else {
        led.unplot(posn, 4)
        posn = Math.min(0, posn - 1)
        led.plot(posn, 4)
    }
})
```

DavidWhaleMEF:
I tried this in /beta too and the problem is still there

DavidWhaleMEF:
beta:
A Blocks / JavaScript code editor for the micro:bit.
pxt.microbit.org version: 0.9.37
PXT version: 0.11.61
C++ runtime version: v2.0.0-rc8

live:
A Blocks / JavaScript code editor for the micro:bit.
pxt.microbit.org version: 0.9.37
PXT version: 0.11.61
C++ runtime version: v2.0.0-rc8

PR #1555, opened 2017-03-07
Fixing max roundtrip

abchatra:
Fixes https://github.com/Microsoft/pxt/issues/1496


pxtlib/emitter/emitter.ts
```diff
@@ -306,7 +306,7 @@ namespace ts.pxtc {
         advanced?: boolean;
         deprecated?: boolean;
         useEnumVal?: boolean; // for conversion from typescript to blocks with enumVal
-
+        field? :  Object;
```
riknoll: This interface is for the attributes that are parsed from the comments on ts and cpp files. Instead of storing the field attribute here, can we instead just check for it on line 1061?

pelikhan:
Décompilation tests

abchatra:
Implemented the review comments.

PR #1579, opened 2017-03-07
Subcategories redo

riknoll:
Resurrecting this old PR, updated to master. Allows namespace to have subcategories. By default subcategories are sorted by name, but namespaces may also specify an ordering as an array of strings. The ordering is always deterministic. Not really sure how to do localization, @pelikhan thoughts?


pelikhan:
Do we really need the additional array of strings? Aren’t weight sufficient?

riknoll:
We can't do weight because then you would have to copy the weight for each API. This way the weights are in one place.

PR #1595, opened 2017-03-09
support for docs sharing between targets with per-target augmentation

mmoskal:
@pelikhan @microsoftsam @Jaqster comments?

Simple example:
Base: https://github.com/Microsoft/pxt-common-packages/blob/master/libs/neopixel/docs/reference/light/default-pin.md
Target: https://github.com/Microsoft/pxt-adafruit/blob/master/docs/reference/light/default-pin.md
Result: https://makecode.adafruit.com/reference/light/default-pin


docs/writing-docs.md
```diff
@@ -164,6 +164,114 @@ Append `-ignore` to any of the above to ignore a snippet in automated testing:
...
+version will be used as base and sections in it overridden.
+
+Sections are overridden based on matching their ids. For example, a target's version of
+the document above can look as follows:
+
+```markdown
+# @extends
+### #ex2
+My example
+## See Also These! #also
```
microsoftsam: where is the end of ex2 in this case?

microsoftsam: What if we wanted to add something after ex2 that is part of the main template?

pelikhan: You would have to modify the main template then and percolate the changes

pelikhan: ex2 ends at ### See Also

microsoftsam: not clear that the See also ends that section? Do you always need a See also? it's a little confusing. 
My expectation would be that we would have something like this:
### #ex2

### #endex2

It's clear where it starts and where that section ends

pelikhan: It stops on the next same-level or higher header. So any #, ##, ### or end of file.

mmoskal: I think the `#endex` would be too much work. Think of section headers in the document as forming a tree, following the nesting levels of headers. Then an ID on the section header applies to this section and all its children.

mmoskal: I mean too much work for people writing docs, for me it would be easier ;)

mmoskal: Added following:

The sections in the upstream document are treated as a tree, following the nesting level of headers. In the target document, sections stretch from the beginning of a section with id, until the next section with id starts, that is the tree structure is not required.

PR #1602, opened 2017-03-10
Functions as custom blocks

pelikhan:
customblocks

A prototype experience to enable easily defining custom blocks from JavaScript functions.

- user goes to JS
- click on + in explorer view,
- custom.ts gets added to project with boiler plate function/block
image
- user added more custom functions that show up as blocks
image
- shares scripts and students use the custom blocks!


OwenBrotherwood:
Placeholder for Owen : will be edited at some time with more or better information.

Is the namespace "free style coding" ie observe normal programming (Typescript) with no constraints as there is with the "javascript" that is used in the "javascript canvas (the non-blocks canvas)"
The requirement of "advanced usage" creating requirement of "advanced knowledge" to achieve the most from this great feature.

pelikhan:
> Is the namespace "free style coding" ie observe normal programming (Typescript) with no constraints as there is with the "javascript" that is used in the "javascript canvas (the non-blocks canvas)"

There is no limitation on the namespace name, as long as it is accepted by typescript. You can also define multiple namespaces or inject functions in existing namespaces.

> The requirement of "advanced usage" creating requirement of "advanced knowledge" to achieve the most from this great feature.

Please explain.

OwenBrotherwood:
I have sometimes seen the words advanced usage with some issues. There could be some delaying factors with adding this feature if one uses too much time with making editor features that help with syntax etc: sometimes one has to disallow features as syntax etc can not follow. If the feature is advanced usage just now, it can be made easier with syntax checks etc latter when it is found how/if the feature is used

pxtlib/emitter/emitter.ts
```diff
@@ -488,6 +488,15 @@ namespace ts.pxtc {
             return cached
         let res = parseCommentString(getComments(node))
         res._name = getName(node)
+        if (node.kind == SK.FunctionDeclaration && res.block === "true" && !res.blockId) {
+            const fn = (node as any) as ts.FunctionDeclaration;
```
riknoll: You don't need to cast to any here

pxtlib/emitter/emitter.ts
```diff
@@ -488,6 +488,15 @@ namespace ts.pxtc {
...
+            if ((fn.symbol as any).parent) {
+                res.blockId = `${(fn.symbol as any).parent.name}_${getDeclName(fn)}`;
+                res.block = `${node.symbol.name}${fn.parameters.length ? '|' + fn.parameters
+                    .filter(p => !p.questionToken)
+                    .map(p => `${p.symbol.name} %${p.symbol.name}`).join('|') : ''}`;
```
riknoll: Instead of `p.symbol.name` you should use `(p.name as ts.Identifier).text` as `symbol` might not be defined. Also, error out here on a binding pattern.

webapp/src/filelist.tsx
```diff
@@ -100,13 +101,68 @@ export class FileList extends data.Component<ISettingsProps, FileListState> {
         this.props.parent.setState({ showFiles: !this.props.parent.state.showFiles });
     }
 
+    private addFile() {
```
riknoll: `addCustomBlocksFile`?

pxtlib/emitter/emitter.ts
pelikhan: cast did not work with node, worked out with node0.

PR #1633, opened 2017-03-14
adding basic blog entry

pelikhan:
Basic blog structure


docs/blog/hello-world.md
```diff
@@ -0,0 +1,8 @@
+# Hello Microsoft MakeCode!
+
+### Posted on 3/14/2017
```
mmoskal: I guess ``**Posted on ...**`` would make more sense from document structuring standpoint.

Also, March 14, 2017 or 2017-03-14 would be less ambiguous (for future dates).

microsoftsam:
Link to the blog from the MakeCode site? Perhaps in the footer

PR #1649, opened 2017-03-15
Static TypeScript design

tballmsft:
creating a pull request to discuss STS design. Edit doc at

https://github.com/Microsoft/pxt/blob/STS/common-docs/javascript/sts.md


tballmsft:
@mmoskal said:

We do in fact support type assertions for class types. We could support union and intersection types between object values. What we cannot do is distinguish between strings, numbers, booleans and classes, since they all share representation. This is similar to Java, where `int` is different from `Int`, and you can cast `Int` to `Object` but not `int` to `Object`. C# hides that, while C++ doesn't really have the concept.

tballmsft:
@tball said:

We just emit the underlying expression of the type assert. There is no runtime check of class, as we don't have a runtime tag.  We should make a decision on whether we want to add a runtime tag for classes. This will further distinguish them from records.   This would also allow us to check safety of downcasts.  Regarding union and intersection, once we have a runtime tag, they are easier to check.

tballmsft:
@pelikhan  said:

Reads great :)

Need to address exceptions, floating point vs integer semantics, regex, debugging, async!!!.

We do support "any" for built in functions like string concat.

It would be good to give the historical context behind sts, it helps putting design decision in the context of microcontrollers.

tballmsft:
@mmoskal said:

@tballmsft Sorry, that's right! I thought I implemented the tags. It shouldn't be too hard though. I think we should have it. instanceof expression is only for classes, not for records, so we should be good there.

tballmsft:
@pelikhan, I don't understand this:

> We do support "any" for built in functions like string concat.

riknoll:
@tballmsft I believe he is referring to the `+` operator

mmoskal:
The `+` operator supports a few types, based on static information. It's much like static overloading.

mmoskal:
One idea for floating point would be:
* compile all numbers to 32 bit floating point (JS uses 64 bit)
* (optional) unless a function has some `int` annotation - then compile everything in there to integers

The compiler would then need to be smart about casting floats to ints and back when calling `int` functions and C++.

This is seriously ugly though.

mmoskal:
Another idea: compile everything to 32 bit float, and introduce a new int32 type. Then define functions like int32.add etc - OCaml style https://caml.inria.fr/pub/docs/manual-ocaml/libref/Int32.html

Again the compile will need to know about C++.

mmoskal:
Of course optimal for float would be to have some sort of range types in TypeScript. But I guess it's unlikely they would have the right semantics anyway.

abchatra:
Thanks for the starting this thread. First I looked at this specification from a javascript perspective. 

We need to make up our mind on the following ES6\ES7 features:
- Spread, rest 
- Destructuring 
- Generators 
- Async
- Built-in constructors such as Proxy, Promise, Reflect etc. 
- Temporal dead zone for let\const.

And many other tiny ones such as Subclassing. We support Classes, lamdas, let\const, template strings. 

We could limit our implementation to support ES5.1 (which is what IE9\IE10 supports) + some hand picked ES6 features such as classes, let\const etc. Even then we are missing following features: 
- Built-in constructors such as Object, Array.
- Getter\Setter properties on an object. 
- Function constructor & apply & call. 
- Built-in API surface. We are very adhoc in our implementation here. We support some ES5 (push) & some ES6 (map) on Array not others.  Monaco intelligence helps, but it is not clear why we chose some and why not others. 

Floating point is bit sticky as Javascript\Typescript everything is 64 bit floating point number. May be we can play tricks with tagged integers and floats on supported processors. 

We can take one of the following approaches: 

1.  Support ES5.1+(Classes & lamda & let\const) : Document every unsupported feature. 
2.  Define our own specification with a set of features inspired from Javascript.  

I am leaning towards (2) here. It is still a subset of typescript, but clearly small set of features which are supported.

tballmsft:
@abchatra, thanks for the thoughts. My comments:

Already present.
- Subclassing
- Getter\Setter properties on an object (we have get/set for classes, I think this enough )

Yes
- Destructuring (a core concept)
- Built-in API surface. We are very adhoc in our implementation here. We support some ES5 (push) & 
  some ES6 (map) on Array not others. Monaco intelligence helps, but it is not clear why we chose some    
  and why not others.

No (too complex, or please explain why we need for CS Education, physical computing)
- Spread, rest
- Async (too much complexity, plus we already support async our own way)
- Generators
- Built-in constructors such as Proxy, Promise, Reflect etc.
- Function constructor & apply & call.

Need more information:
- Temporal dead zone for let\const.
- Built-in constructors such as Object, Array.

tballmsft:
Also, I am leaning towards (2):

2. Define our own specification with a set of features inspired from Javascript.

One issue we will have with (2) is that if we decided to later go with a JavaScript VM, we will have a back compat issue.

mmoskal:
I think we could implement spread and rest. One can do variadic functions with them, and otherwise they are mostly syntactic sugar. I guess it is not something you teach at first, but variadic functions are quite common in programming languages.

We could also add `delete` on object literals for completeness.

`new Object()` should I guess compile to `{}`, similarly to `new Array()` (`[]`) - should be easy. We can support them, as long as TypeScript knows the correct types.

As for promises and async - this is I guess for explicit manipulation of threads and is not really supported by Codal.

I've updated the language.md accordingly in https://github.com/Microsoft/pxt/commit/d69ee49c846a55b8f1682566ccc2371f7819948c

(Also compared our compilation to JS engines.)

tballmsft:
Thanks, @mmoskal. Aren't object literals a reference counted type?

pelikhan:
rest would be very helpfull for STEM logging api (write multiple values to serial for example). 1st class support in blocks would be great (gearwheel where the user can add parameters)

mmoskal:
Object literals are reference counted - they are just hash maps. Did I suggest somewhere they weren't?

abchatra:
@mmoskal note _new Array()_ has various semantics like 
_new Array(100)_  creates an empty array but with a length 100.
_new Array(1,2,3,4)_ creates an array with those elements.
Also you could override the Array as it is mutable. 

We would have to translate all of them. My proposal is keep it minimal but complete for the first cut.

abchatra:
@tballmsft 

I would skip generators. Correct semantics for yield, yield* are pretty much nightmare. Generators are not very popular either.

_Need more information:_
Temporal dead zone for let\const.: 
It is runtime reference error to use let before the declaration (not compile time). Example:
```js
function foo() {
	function nested () {
 		console.log (number); //closure captures number
 	}
	//nested(); //throws a reference error for accessing number before declaration.
	let number = 10;
	nested();  // call to nested is fine here.
 }
```
Built-in constructors such as Object, Array. : new Array() & new Object()

mmoskal:
@abchatra wow, so `new Array(10)` is very different from `new Array(10,1)`? I don't think this is something we should inflict on kids.

As per semantics in general, I would like to be a subset of JS also semantically - that is we might reject the program (runtime or compile error) more often, but shouldn't let through code that doesn't work in JS. For example, for these temporal dead zones, I would rather replace them with a compiler check, and for arrays only support Array(100).

Unfortunately, this won't be so easy for null/undefined/0 and floats :// 

We could try to catch (statically) conversions from null to numbers/booleans, and ban undefined altogether. 

For numbers, I guess we should go with floats everywhere by default, but they would still be 32 bit and not 64 bit ... 

Also, in general if we could catch most of semantic differences in the compiler that would be perfect.

mmoskal:
Thinking about it a bit more, there is an alternative design for numbers, undefined and null:
* for numbers, use 31 bit signed tagged integers (lowest bit is one) or reference-counted, boxed 64 bit floats – depending on the value
* for `null` use 0
* for `undefined` use a specific object in flash
* for `true` and `false` also use specific objects in flash

The disadvantages of this are:
* it will not use floating point unit on Cortex-M4F, since it only supports 32 bit floats
* it will be somewhat slower than our current code and consume more memory in some cases
* the code will be somewhat bigger

I would estimate typical impact be under 50% in all three areas (RAM, flash, speed). It seems like a price worth paying for JS compatibility and floating point number support.

This also brings us closer to being able to dynamically examine type of values. We could determine the type of:
* numbers
* booleans
* `undefined`, `null`
* user-defined classes and dynamic objects (provided we add some RTTI to `RefObject` class, but this is relatively easy and cheap)

What we cannot determine type of is Codal/Dal objects based on `RefCounted` struct, since they have no RTTI:
* strings
* buffers
* images

I guess we would add RTTI to them by turning them into something akin to our user-code classes, with vtables. I would add a compile-time option in Codal to supply vtable pointers for these three types. This will break word-alignment of data in strings and buffers, but hopefully no-one depends on this.

Once this is done, we would no longer necessarily need to know statically which values are primitive and which need reference counting, as this could be determined dynamically.

tballmsft:
I think we should stick with the runtime distinction between RefRecord (for classes) and Map (for object literals).   We can use the itable to indirectly access field/vtable/property so that structural subtyping is respected. We pay no more in memory usage than now (we already create the itables), but pay an indirection cost for every field/vtable/property access. Some simple optimizations can be done to allow direct access via fields/vtables.

tballmsft:
I've been thinking more about classes in TypeScript. Classes do come with behavior, but structural subtyping doesn't care about inheritance of specific behavior. This is really odd for intro CS.  Here's a proposal: 
1. only allow casts between class types within an inheritance chain (no structural comparison);
2. don't allow an interface type to be cast to a class type
The first one follows C#/Java, etc (nominal typing).  The second one is needed to prevent object literals from being interpreted via a class type. They can only be interpreted via interface type.

mmoskal:
I'm not against redoing the field access to always go via iface-table. As @tballmsft says the overhead wouldn't be big and we align with JS semantics. It's also simpler (in the sense of being uniform).

Now, what you want to allow in the source language is another question. I always thought class types were distinct from interfaces (having programmed in TypeScript pretty much since day one it was available), so I guess it's not a very essential feature to be able to cast interfaces to classes. Also it makes the semantic of `instanceof` confusing to say the least. But maybe people used to JavaScript think otherwise.

pelikhan:
I am for restricting confusing aspects of the language. Confusing language features do not help teachers/learners.

tballmsft:
>  I guess it's not a very essential feature to be able to cast interfaces to classes.

I agree. Plus, it's pretty confusing since classes have associated code with them and interfaces don't.  By the same reasoning, we should NOT permit casts between classes that are related by structural subtyping but not related by extends clause.  

That is, we use nominal typing for classes, and structural subtyping for  relating classes to interfaces and interfaces to interfaces.

tballmsft:
I've started to implement our more restrictive type checker here. Want to keep code and spec in sync.

tballmsft:
@mmoskal - can you please look at the failure with respect to lookupTypeParameter. To repro, sync to this branch, build, and do "pxt run" in libs/lang-test0.

tballmsft:
I've traced the problem with TypeParameters some more. Here's the code:
```
function swap<T>(arr: T[], i: number, j: number): void {
    let temp: T = arr[i];
    arr[i] = arr[j];
    arr[j] = temp;
}


function sortHelper<T>(arr: T[], callbackfn?: (value1: T, value2: T) => number): T[] {
    if (arr.length <= 0 || !callbackfn) {
        return arr;
    }
    let len = arr.length;
    // simple selection sort.     
    for (let i = 0; i < len - 1; ++i) {
        for (let j = i + 1; j < len; ++j) {
            if (callbackfn(arr[i], arr[j]) > 0) {
                swap(arr, i, j);
            }
        }
    }
    return arr;
}
```

tballmsft:
OK. I've narrowed it down to emitCallCore. The problem (I think) is that the mapping of TypeParameters doesn't consider the mapping of type arguments at the call site:
```
            // TODO: this mapping should take into account the potentially generic function
            // TODO: being called and the types inferred for its typearguments
            if (sig) {
                // NOTE: we are playing with TypeScript internals here
                let trg: Signature = (sig as any).target
                let typeParams = sig.typeParameters || (trg ? trg.typeParameters : null) || []
                // NOTE: mapper also a TypeScript internal
                bindings = getTypeBindingsCore(typeParams, typeParams.map(x => (sig as any).mapper(x)))
            }
```

tballmsft:
The problem, I think, is that in emitCallCore we map the typeargument at the call site to the typeparameter in the signature,  and insert this into the bindings in scope, but then later do a lookup on the typeargument, which fails to resolve, because the mapping is not applied to it.
```
            if (sig) {
                // NOTE: we are playing with TypeScript internals here
                let trg: Signature = (sig as any).target
                let typeParams = sig.typeParameters || (trg ? trg.typeParameters : null) || []
                // NOTE: mapper also a TypeScript internal
                bindings = getTypeBindingsCore(typeParams, typeParams.map(x => (sig as any).mapper(x)))
            }
```

tballmsft:
OK. I have reached a stable point with the more restrictive type checker. All library code from pxt-microbit and pxt-adafruit passes the type checker. I'll push to master at the end of the week in case you would like to test in this branch first.

PR #1662, opened 2017-03-16
Enum decompilation bugs

### LINKED ISSUE #1434: showIcon decompiles into the first icon
The following code: 
```
for (let index = 0; index <= 4; index++) {
    basic.showIcon(index);
}
```

decompiles to 
screen shot 2017-02-27 at 3 13 38 pm

### LINKED ISSUE #1656: Blocks<-->JS: Melody change gets lost
When switching between blocks and JS changes to the melody in a 'play melody' block seem to get lost
melody break

Project to try where it breaks
https://pxt.microbit.org/39743-48834-70525-48858

riknoll:
Fixes #1656 
Fixes #1434 

Fixing some decompiler enum bugs. The issue for both is that we special-case enum values in the blocks as dropdown fields rather than block inputs. As a result we have to check for enum values and decompile them separately.


pelikhan:
LGTM.

PR #1702, opened 2017-03-21
Global toolbox filters from controller

microsoftsam:
Adding support for global toolbox filters for both blocks and javascript.

Interface:
```
   export interface ProjectFilters {
        namespaces?: { [index: string]: FilterState; }; // Disabled = 2, Hidden = 0, Visible = 1
        blocks?: { [index: string]: FilterState; }; // Disabled = 2, Hidden = 0, Visible = 1
        fns?: { [index: string]: FilterState; }; // Disabled = 2, Hidden = 0, Visible = 1
        defaultState?: FilterState; // hide, show or disable all by default
    }

    export enum FilterState {
        Hidden = 0,
        Visible = 1,
        Disabled = 2
    }
```

### DIFF
--- a/pxteditor/editor.ts
+++ b/pxteditor/editor.ts
@@ -29,6 +29,7 @@ namespace pxt.editor {
     export interface IAppState {
         active?: boolean; // is this tab visible at all
         header?: pxt.workspace.Header;
+        filters?: pxt.editor.ProjectFilters;
         currFile?: IFile;
         fileState?: string;
         showFiles?: boolean;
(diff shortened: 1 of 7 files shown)


pxteditor/editor.ts
```diff
@@ -61,9 +62,23 @@ namespace pxt.editor {
...
+    export interface ProjectFilters {
+        namespaces?: { [index: string]: FilterState; }; // Disabled = 2, Hidden = 0, Visible = 1
+        blocks?: { [index: string]: FilterState; }; // Disabled = 2, Hidden = 0, Visible = 1
+        fns?: { [index: string]: FilterState; }; // Disabled = 2, Hidden = 0, Visible = 1
+        defaultState?: FilterState; // hide, show or disable all by default
+    }
+
+    export enum FilterState {
+        Hidden = 0,
+        Visible = 1,
```
pelikhan: shouldn't visible by the default state?

webapp/src/app.tsx
```diff
@@ -684,7 +685,7 @@ export class ProjectView
             const files = JSON.parse(data.source) as pxt.Map<string>;
             // we cannot load the workspace until we've loaded the project
             workspace.installAsync(h, files)
-                .done(hd => this.loadHeaderAsync(hd));
+                .done(hd => this.loadHeaderAsync(hd, null));
```
pelikhan: filter optional?

microsoftsam: yeah filter is optional. probably dont need this.

pxteditor/editor.ts
microsoftsam: Nah I'd say most of the time, you're gonna want to hide everything and only show certain blocks, rather than the other way around.

PR #1712, opened 2017-03-22
always open help urls in new window

pelikhan:
With this change, all links from the help menu open in new window (- tutorial)

* Provide a consistent experience accross all browser sizes
* Documentation is browsable/searchable from the full page view
* side doc is really meant to see the help of a particular block (right click help still opens sidedocs) or follow a project (starting package from "projects" dialog still opens sidedocs)

### DIFF
--- a/webapp/src/container.tsx
+++ b/webapp/src/container.tsx
@@ -15,11 +15,6 @@ export class DocsMenuItem extends data.Component<ISettingsProps, {}> {
         super(props);
     }
 
-    openDoc(path: string) {
-        pxt.tickEvent(`docs`, { path });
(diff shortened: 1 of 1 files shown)


abchatra:
What is the motivation for the change? @Jaqster - FYI

pelikhan:
@abchatra updated description

Jaqster:
Yup, I'm okay with this change.  But did we fix the issue where in a small screen, nothing happens when they click a Make project from the Project dialog?

pelikhan:
> But did we fix the issue where in a small screen, nothing happens when they click a Make project from the Project dialog?

Please file an issue. It is unrelated.

Issue #1713, opened 2017-03-22
pxt init does not appear to include line for PXT/TARGET


OwenBrotherwood:
https://pxt.microbit.org/packages

> Make sure you keep the line for PXT/TARGET (where TARGET is the target id) in README.md. Otherwise the package will not show up in search.

after following doc, pxt init README.md is empty

pelikhan:
Oh, that's silly. Need to fix this.

OwenBrotherwood:
I cannot confirm that the fix works: I need to double check my installation 
(I pulled latest pxt from github and followed doc for github prep and usage)

pelikhan:
Fixed in beta. Run "pxt update" and try again

OwenBrotherwood:
(aarg: I was on exploreraddfile branch)

PR #1714, opened 2017-03-22
fix generation of readme.md in pxt init

pelikhan:
When running ``pxt init`` make sure README.md content is properly written.


pelikhan:
https://github.com/Microsoft/pxt/issues/1713

Issue #1657, opened 2017-03-16
packages: add package by URL?


jaustin:
This is more a point for discussion than a direct feature request.

In a conversation with @carlosperate this week we talked about getting access to packages that weren't yet whitelisted.

Is there any mechanism to add a package directly to a project by putting the URL to the package.json in somewhere? This would help people test packages before they're whitelisted. 

It certainly subverts the point of a whitelist to some extent, but it also adds flexibility.

What do people think?

pelikhan:
If you can circumvent whitelisting, then whitelisting is useless. We should revisit this feature. I agree, it is darn too hard to do packages right now.

carlosperate:
Thanks for posting this Jonny, I almost forgot.
From my point of view, the whitelisting is most useful for discoverability (only whitelisted packages show in the "add package" menu/search) and as "stamp of approval" for the package quality.
If there was a way to add external packages by pointing to the json file directly, and perhaps block sharing the project, would then be any real disadvantages?

abchatra:
I do want to solve this problem for package developer. 
One proposed solution:
 - User types the entire GitHub URL in the search box (which is not easy for kids to type in, but provides an workaround for package developers)
 - Display a warning message that it is unverified package.
 - Load the package and let the user try out the program. Though disallow publishing projects with the unverified package.

@jaustin, @pelikhan  any comments

pelikhan:
"Though disallow publishing projects with the unverified package." Any script is potentially an unverified package. I don't think that makes sense.

abchatra:
@pelikhan, yes sorry we can skip the last step.

abchatra:
Fixed. We have add package by URL

Issue #1573, opened 2017-03-07
When use the “search” box to find existing package, no results.


WendyFeng01:
•	Navigate to https://pxt.microbit.org/beta  
•	Click “Add Package” 
•	Such as taping “radio” on search box
•	Press “Search”
![4](https://cloud.githubusercontent.com/assets/23497626/23652023/3a0c4868-02dc-11e7-8b31-36a5d212308c.PNG)

**Expect result** : 
The list should have ‘radio’ package.

**Actual result** :
No corresponding results.
![5](https://cloud.githubusercontent.com/assets/23497626/23652106/80cd554e-02dc-11e7-82d4-a38f0e530d84.PNG)

pelikhan:
Search does not seem to match built-in packages.

pelikhan:
Fixed.

WendyFeng01:
This bug still repro. When I search the "radio", no corresponding results.
![capture](https://cloud.githubusercontent.com/assets/23497626/25609654/a260142e-2f52-11e7-80be-a83ce2bfecff.PNG)

pelikhan:
If radio is already added to your project, it will not show up.

Issue #1146, opened 2017-01-14
Add Package dialog


Jaqster:
Let's only show packages which aren't already loaded.

pelikhan:
We should tackle that "projects" dialog in the next design meetup.

abchatra:
Fixed

PR #1718, opened 2017-03-22
Easier packages

pelikhan:
Follow up of #1657 

- [x] gracefully handle missing target config in offline mode
- [x] treat packages as "unknown" when target config cannot be loaded
- [x] NOT IMPLEMENTED prompt the user that the project references a non-approved package
- [x] support for referencing shared projects (https://github.com/Microsoft/pxt/issues/1657)
- [x] search bundled packages (#1573)
- [x] do not show referenced bundled packages (#1146)
- [x] allow reference github repo by deep url


abchatra:
:shipit:

PR #1736, opened 2017-03-24
Add optional comment when compiling on start event

riknoll:
Configurable in the target. Prevents you from having comments in your code that say "On start" but we don't really have a better way of managing that in the decompiler and it's an unlikely scenario.


pxtlib/emitter/emitter.ts
```diff
@@ -7,6 +7,7 @@ namespace ts.pxtc {
     export import U = pxtc.Util;
 
     export const ON_START_TYPE = "pxt-on-start";
+    export const ON_START_COMMENT = U.lf("On Start");
```
pelikhan: We should use the same capitalization as the block... "on start"

PR #1738, opened 2017-03-24
ignore version number for safari pinned web sites

pelikhan:
No version available when the web site is pinned.


pelikhan:
Pinned pages only.

Issue #1728, opened 2017-03-23
Download to micro:bit - Usability


tracygardner:
Related issues have been raised before:
https://github.com/Microsoft/pxt/issues/1686 
 https://github.com/Microsoft/pxt/issues/888 

But downloading to the micro:bit from PXT is our biggest pain point. 

I'm using it in a Code Club and with upper Key Stage 2. Downloading becomes the main focus of the lesson!

Would it be possible to use a 'Save As' approach like most online editors do? Are there any issues with saving direct to the micro:bit?

I know you also download to the Downloads directory. But that's our other pain point. Downloads and local browser storage are wiped every session. Children have to copy their files somewhere else to keep them. A "Save As" dialog would mean they could just save somewhere else.

pelikhan:
The only fix available in IE/Edge is to remove that button.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

tracygardner:
@pelikhan Check the full issue. (My fault, randomly did something that caused a premature submit and then editted.)

Is 'Save As' an option?

pelikhan:
We tried that as well to no avail. IE/Egde will not allow to open or save as a data:... file. At least in our testing it didn't.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

tracygardner:
Not sure I understand. Are you saying you can't use a standard browser 'Save As' dialog (like you would in Scratch) with the hex file? 

In Edge things aren't too bad, you can just click 'Save As' in the dialog that pops up. The issue here is that the instructions tell you to save and then send the file: https://pxt.microbit.org/device/usb/windows-edge which is much more complex!

In Explorer you can change the option from 'Save' to 'Save As' - a bit less straightforward. But again the instructions make it seem much harder. 

In Chrome there's no option to choose where you save (well you can click 'Show All' and then right-click on the link there.) It works nicely with the uploader once you have it running. But the children find this tricky, they forget to run it, or they cancel the Agreement Window, or close the Window once they've gone through process, or don't see the message telling them the board can't be detected ...

The overall issue is that behaviour is different in every browser! Having a 'Save As' option just for Chrome (and Firefox?) might make the overall behaviour a bit more consistent.

pelikhan:
The "download again" button is there on the case where the first download got blocked.

Unfortunately (or fortunately), a web page cannot control how browsers handle downloads.

Get Outlook for iOS<https://aka.ms/o0ukef>
[...]

PR #1746, opened 2017-03-24
Save as msg

pelikhan:
- [x] When the browser supports it, suggest to "save as" directly into the board drive
- [x] display .hex file name instead of "download again"


pelikhan:
https://github.com/Microsoft/pxt/issues/1728

PR #1751, opened 2017-03-25
Better advanced category and subcategory support

### LINKED ISSUE #1725: Minor subcategories pain points
- Clicking on a category when inside a subcategory collapses the category instead of opening the top-level fly-out
    - This is very annoying, as you need to click the category twice to get back to the top-level blocks
    - This is inconsistent with the behavior of the "More..." subcategory: going from "More..." to top-level works, instead of collapsing the category
- Categories with subcategories do not collapse when another category is opened, leading to a toolbox that is way too populated after a while
    - I understand this behavior is by design because of the "Advanced" category (we don't want to collapse Advanced every time user opens a new category), but we should collapse other top-level categories

riknoll:
Fixes #1725 

This is a minor UI breaking change! This PR fixes the subcategory behavior and removes our hack for advanced categories. Instead, advanced categories and the "Add Package" category are now hidden underneath a toggle.

Regular toolbox:

advanced_expand_collapse_microbit

Inverted toolbox:

advanced_expand_collapse_adafruit

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -809,11 +809,9 @@ namespace pxt.blocks {
             insertTopLevelCategory(document.createElement("sep"), tb, 1.5, false);
             insertTopLevelCategory(cat, tb, 1, false);
 
-            if (showCategories === ShowCategoryMode.All) {
+            if (showCategories === ShowCategoryMode.All && pxt.appTarget.cloud && pxt.appTarget.cloud.packages) {
                 // Add the "Add package" category
-                if (tb && showCategories !== ShowCategoryMode.None && pxt.appTarget.cloud && pxt.appTarget.cloud.packages) {
-                    getOrAddSubcategoryByWeight(tb, Util.lf("{id:category}Add Package"), "Add Package", 1, "#717171", 'blocklyTreeIconaddpackage')
-                }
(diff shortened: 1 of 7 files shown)


webapp/src/app.tsx
```diff
@@ -501,7 +502,7 @@ export class ProjectView
                 switch (t.subtype) {
                     case 'steploaded':
                         let tt = msg as pxsim.TutorialStepLoadedMessage;
-                        let showCategories = tt.showCategories ? tt.showCategories : Object.keys(tt.data).length > 7;
+                        let showCategories = (tt.showCategories ? tt.showCategories : Object.keys(tt.data).length > 7) ? ShowCategoryMode.All : ShowCategoryMode.None;
                         this.editor.filterToolbox({ blocks: tt.data, defaultState: pxt.editor.FilterState.Hidden }, showCategories);
```
microsoftsam: Is this right? Setting showcateogrymode to all for tutorial mode? Should probably be set to basic here

pxtblocks/blocklyloader.ts
```diff
@@ -15,6 +15,12 @@ namespace pxt.blocks {
         lists: '#D83B01'
     }
 
+    export enum ShowCategoryMode {
+        All,
```
microsoftsam: More appropriate name would just be CategoryMode or ToolboxCategoryMode

pelikhan:
Can we prevent text selection?

microsoftsam:
Should be able to with some CSS


riknoll:
Fixed the category selection CSS

PR #1759, opened 2017-03-27
Icon urls need to be prefixed with CDN url

riknoll:


pxtblocks/blocklyloader.ts
```diff
@@ -269,7 +269,7 @@ namespace pxt.blocks {
             toolboxStyleBuffer += `
                 .blocklyTreeIcon.${className} {
                     display: inline-block !important;
-                    background-image: url("${encodeURI(i)}")!important;
+                    background-image: url("${pxt.webConfig.commitCdnUrl + encodeURI(i)}")!important;
```
guillaumejenkins: Don't you also need to get rid of "/cdn/" that will be at the start of the URL?

riknoll: I was just thinking they shouldn't put it. The icons have to be in the CDN anyway.

PR #1762, opened 2017-03-28
Make the monaco toolbox categories match the Blockly ones

riknoll:
Adds an advanced toggle to the Monaco toolbox, more built-in categories with TS snippets, and fixes the sorting of the categories.

Still to do: subcategories


microsoftsam:
Screenshots please

PR #1763, opened 2017-03-28
smaller buttons for inline blocks

pelikhan:
More compact rendering of inline buttons / blocks
image


microsoftsam:
Looks good :)

PR #1772, opened 2017-03-28
mine first image as image for twitter card

pelikhan:
Automatically mine the first image in a doc page for the twitter card


mmoskal:
Might be a bit overzealous - I would go for // or something similar. Not sure when does the rewriting of URLs happen too.

pelikhan:
@mmoskal restricted regex

PR #1791, opened 2017-03-30
Allow advanced to be set to false.

microsoftsam:
Allow advanced namespace flag to be set to false, not just true.
This allows us to override the advanced field

### DIFF
--- a/pxtlib/emitter/emitter.ts
+++ b/pxtlib/emitter/emitter.ts
@@ -439,7 +439,7 @@ namespace ts.pxtc {
 
         for (let n of booleanAttributes) {
             if (typeof (res as any)[n] == "string")
-                (res as any)[n] = (res as any)[n] == 'true' ? true : false;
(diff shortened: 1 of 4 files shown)


pelikhan:
We should parse the string into a Boolean rather than having to spread those error prone === 'true' all over the press. Also sometimes we use =1 as well.

microsoftsam:
Fixed

pxtlib/emitter/emitter.ts
```diff
@@ -436,6 +437,11 @@ namespace ts.pxtc {
                 (res as any)[n] = parseInt((res as any)[n])
         }
 
+        for (let n of booleanAttributes) {
+            if (typeof (res as any)[n] == "string")
+                (res as any)[n] = (res as any)[n] == 'true' ? true : false;
```
pelikhan: Handle "1" as well

PR #1834, opened 2017-04-04
Moving Monaco toolbox to use the same blockinfo as Blockly loader

### LINKED ISSUE #1792: Monaco toolbox doesn't respect namespace annotations in order
We should render the monaco toolbox in the same way as the blockly one, and this means respecting the order in which namespace annotations are written. 

In order to allow targets to override specific namespace attributes, we need to do this in the right order, and the blockly toolbox does that today. Need to also do that for the monaco toolbox.

### LINKED ISSUE #1809: Monaco toolbox titles should be lowercase
To match the API namespace in the editor

riknoll:
Fixes #1809 
Fixes #1792 

This PR moves the Monaco toolbox to match the Blockly toolbox much more closely by using the same blockinfo instead of parsing it separately. Now, the same blocks/functions show up in the Monaco toolbox as in the Blockly toolbox and in the same order. Additionally, we no longer query the language service a ton before loading the toolbox so load time for both the syntax highlighting and the toolbox has improved significantly. Finally, namespaces in the JS are no longer capitalized. Here's a screenshot of what that looks like:

lowercase_toolbox

One caveat to this is that blocks that are class instance methods (like those in the "Game" category), do not show up in the Monaco toolbox because we don't really have a good snippet to insert for them. This problem does not affect `fixedInstance` or `defaultInstance` APIs so Adafruit works fine, but it might be something to look into in the future.


microsoftsam:
@Jaqster okay with merging this? Are you okay with some categories being uppercase and others being lowercase. @jaustin is this UI change okay to merge for the micro:bit target?

Jaqster:
yeah, this is a good change.  How hard would it be to change the name on the Toolbox to be all upper case? (blocks themselves would still be lower case).  Just looks odd have the mix of upper/lower case category names on the Toolbox...

microsoftsam:
Wouldn't be very hard. That is actually what it is right now. @tballmsft and @pelikhan were fans of dropping it to lower case. see #1809

pelikhan:
Hummm. Not strong opinion here. It is a UI breaking change for microbit.

riknoll:
@Jaqster @pelikhan switched back to capitalized. I'll merge this when the build finishes

PR #1836, opened 2017-04-04
Mkcd

pelikhan:
Rename file extension to .mkcd . .pxt still supported to be loaded.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1581,7 +1581,7 @@ function isTypescriptFile(filename: string): boolean {
 }
 
 function isProjectFile(filename: string): boolean {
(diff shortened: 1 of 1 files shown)


webapp/src/app.tsx
```diff
@@ -1129,7 +1129,7 @@ export class ProjectView
             },
             htmlBody: `<div class="ui form">
   <div class="ui field">
-    <label>${lf("Select a .hex file to open.")}</label>
+    <label>${lf("Select a .hex or .mkcd file to open.")}</label>
```
abchatra: This should just be .mkcd right?

pelikhan: right...

PR #1862, opened 2017-04-06
Better rtl support using rtlcss

microsoftsam:
@pelikhan need to do some sort of server re-write to the semantic.css to switch it over to rtlsemantic.css when using an rtl language.


microsoftsam:
screen shot 2017-04-05 at 11 37 14 pm

PR #1863, opened 2017-04-07
Don't decompile output call expressions as statements

riknoll:
Also fixing some incorrect test cases.


tests/decompile-test/baselines/functions_outputs_as_statements.blocks
```diff
@@ -1,46 +1,26 @@
 <xml xmlns="http://www.w3.org/1999/xhtml">
 <block type="pxt-on-start">
 <statement name="HANDLER">
-<block type="test_no_argument_output">
+<block type="typescript_statement">
```
pelikhan: do we expect to always resort to ``typescript_statement`` in this case?

riknoll: Yep

PR #1869, opened 2017-04-07
hot swap to rtl css on demand

pelikhan:
Change the ``href`` on semantic.css to the rtlsemantic.css file


pelikhan:
I'm still looking into making sure that we swap to the correct version

PR #1882, opened 2017-04-11
Code execution trace

riknoll:
The start of our debugging experience. Currently hidden behind a target flag because we still need to figure out a nice UI. Here's a demo:

block_trace
ts_trace


tballmsft:
Single step support?

riknoll:
@tballmsft Nope, not part of this change. The code is there, but we need to figure out what that debugging experience looks like.

PR #1887, opened 2017-04-12
Remove typescript dependency in the editor

microsoftsam:
Removing typescript dependency in the pxtlib. 

- All interface pulled out of the emitter are in pxtlib/service.ts. This is mostly parsing logic.
- Blockly and Monaco use some typescript lib functions, only had to copy a couple of methods: 
-- flattenDiagnosticMessageText, isIdentifierStart and isIdentifierPart
-- These are in their own file for update reasons (typescript.ts in pxtlib)
- Formatting code is now done in the worker
- Stripped out worker from webapp, as pxtblocks needs to format code as well. (pxtblocks takes a dependency on the web worker)
- Since the formatting in the blocklycompiler's compile method is done async, I've had to make the compile methods of blockly async
- CLI had a couple of helper methods in pxtlib (buildAsync and serviceAsync), that are not used by anything but the CLI, and since they depended on typescript, I've pulled them into the CLI.


pxtblocks/blocklycompiler.ts
```diff
@@ -1341,7 +1354,7 @@ namespace pxt.blocks {
         return e;
     }
 
-    export function compileBlock(b: B.Block, blockInfo: pxtc.BlocksInfo): BlockCompilationResult {
+    export function compileBlock(b: B.Block, blockInfo: pxtc.BlocksInfo): Promise<BlockCompilationResult> {
```
pelikhan: rename to ``compileBlockAsync``

pxtblocks/blocklycompiler.ts
```diff
@@ -1479,7 +1492,7 @@ namespace pxt.blocks {
         return undefined;
     }
 
-    export function compile(b: B.Workspace, blockInfo: pxtc.BlocksInfo): BlockCompilationResult {
+    export function compile(b: B.Workspace, blockInfo: pxtc.BlocksInfo): Promise<BlockCompilationResult> {
```
pelikhan: rename to ``compileAsync``

pxteditor/editor.ts
```diff
@@ -95,7 +95,7 @@ namespace pxt.editor {
 
         switchTypeScript(): void;
         openTypeScriptAsync(): Promise<void>;
-        saveBlocksToTypeScript(): string;
+        saveBlocksToTypeScript(): Promise<string>;
```
pelikhan: Add ``Async`` suffix

pelikhan:
image
image

cli/cli.ts
```diff
@@ -2335,7 +2336,12 @@ enum BuildOption {
 
 export function serviceAsync(parsed: commandParser.ParsedCommand) {
     let fn = "built/response.json"
-    return mainPkg.serviceAsync(parsed.arguments[0])
```
mmoskal: I guess this one could go altogether. It was for testing the webworker interface. Not critical either way.

abchatra:
Looks good to me. Only thing I would do is rename pxttypescript directory to pxtcompiler as it is actually a compiler which uses typescript

pelikhan:
Compiler, decompiler, formater...

PR #1894, opened 2017-04-12
Explicitly save the project name when saving a file.. 

### LINKED ISSUE #1821: save button does not always update the filename
microbit.org-ticket: 1431

If you change the name of a script in the script name box next to the save button, then press save or download, the name of the script is not updated the first time round. You have to do it twice for it to work.

I have seen this myself too, it looks like focus from the text entry box has to be lost for the internal script name to be updated.

This has caused much confusion with users saving a script that is then the wrong name.

microsoftsam:
Fixes #1821

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -898,6 +898,7 @@ export class ProjectView
     }
 
     saveAndCompile() {
+        this.saveProjectName();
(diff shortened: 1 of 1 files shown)


abchatra:
:shipit:

PR #1898, opened 2017-04-13
support for rendering referenced images

pelikhan:
Images reference by URL must be loaded and inlined as data uris.


pelikhan:
microbit-screenshot 6

PR #1899, opened 2017-04-13
tagged execution strategy

mmoskal:
This PR implements boxed floating point (double) and tagged 31-bit integer compilation strategy.

The semantics of double->int conversions etc probably doesn't exactly match the JS semantics yet, but is still closer than what we had (ie just ints). I think we should fix these after merging this one. The actual semantics is in `pxt-common-packages` for the most part.

Depends on: 
https://github.com/Microsoft/pxt-common-packages/pull/9
https://github.com/lancaster-university/codal/pull/12 (merged)
https://github.com/Microsoft/pxt-microbit/pull/395


cli/hid.ts
```diff
@@ -38,6 +38,16 @@ export function serialAsync() {
         })
 }
 
+export function dmesgAsync() {
+    return hf2DeviceAsync()
+        .then(d => d.talkAsync(0x0010)
```
abchatra: >0x0010 [](start = 31, length = 6)

Add comment on these magic numbers?

pxtcompiler/emitter/backthumb.ts
```diff
@@ -123,6 +143,73 @@ ${lbl}:`
...
+
+            for (let op of ["adds", "subs", "ands", "orrs", "eors"]) {
+                r +=
+                    `
+_numops_${op}:
+    @scope _numops_${op}
+    lsls r2, r0, #31
+    beq .boxed
+    lsls r2, r1, #31
+    beq .boxed
```
abchatra: This can be done in 3 instructions

and r2, r0, 1
tst r2, r1
beq .boxed

pxtcompiler/emitter/backthumb.ts
```diff
@@ -123,6 +143,73 @@ ${lbl}:`
...
+    push {lr}
+    bl pxt::toInt
+    pop {pc}
+
+_numops_fromInt:
+    lsls r2, r0, #1
+    asrs r1, r2, #1
+    cmp r0, r1
+    bne .over2
+    adds r0, r2, #1
```
abchatra: 4 instructions instead of 5?

teq r0, r0, LSL#1
bmi .over2
lsl r0, r0, #1
add r0, r0, #1

abchatra: XOR the src with itself shifted left one. If there's no overflow, the result should be positive (top bit clear). Similar logic for negative numbers.

---
In reply to: [111836937](https://github.com/Microsoft/pxt/pull/1899#discussion_r111836937) [](ancestors = 111836937)

abchatra:
:shipit:

pxtcompiler/emitter/backthumb.ts
mmoskal: Thumb doesn't have `and r2, r0, #1`, you can only `and rX, rX, rY`

pxtcompiler/emitter/backthumb.ts
mmoskal: No `teq` on Thumb (see http://infocenter.arm.com/help/topic/com.arm.doc.qrc0006e/QRC0006_UAL16.pdf ), and also no shifts/offsets applied to arguments of other instructions.

I don't quite get the second comment.

pxtcompiler/emitter/backthumb.ts
abchatra: Thanks Yep makes sense.

pxtcompiler/emitter/backthumb.ts
abchatra: Got it. My bad! I was explaining the teq. Should have checked the instruction set.

abchatra: Looks good for me to ship.

mmoskal:
Shall we merge?

As for codal, we can use the `vtable` branch for now - we are using `master` now, as there are no tagged releases anyways.

mmoskal:
The codal PR is now merged. I would like to have this on master, especially if you want rebase or squash, since my `dbg` branch is hanging off this one.

pelikhan:
We are doing a last round of testing on the microbit branch then we can merge. Wait another day ?

mmoskal:
n/p. digging into codal now anyway.

tballmsft:
Good to get it in this week.

tballmsft:
@pelikhan - are we ready to go. Don't want this to become too stale.

pelikhan:
@abchatra we need to do a test pass on the micro:bit before getting this in.

abchatra:
We are done with the test on microbit. This is good to merge.

PR #1900, opened 2017-04-14
Simulator modals now show up, but only outside tutorials

guillaumejenkins:


pxtsim/embed.ts
```diff
@@ -64,6 +64,7 @@ namespace pxsim {
         command: "modal" | "restart"
         header?: string;
         body?: string;
+        htmlBody?: string;
```
pelikhan: This is unsafe. The simulator should not be trusted.

webapp/src/simulator.ts
```diff
@@ -98,15 +99,23 @@ export function init(root: HTMLElement, cfg: SimulatorConfig) {
...
                             copyable: msg.copyable,
                             hideAgree: true,
                             disagreeLbl: lf("Close")
-                        }).done();
+                        };
+
+                        if (msg.body) {
+                            modalOptions.body = msg.body;
+                        } else if (msg.htmlBody) {
+                            modalOptions.htmlBody = msg.htmlBody;
```
pelikhan: This allows the target to inject JS into the editor.

webapp/src/core.ts
```diff
@@ -455,3 +455,7 @@ export function apiAsync(path: string, data?: any) {
             console.log(err.message)
         })
 }
+
+export function getSimulatorModalUrls() {
```
pelikhan: Not needed.

webapp/src/simulator.ts
```diff
@@ -98,15 +99,30 @@ export function init(root: HTMLElement, cfg: SimulatorConfig) {
...
+                        const modalOpts: core.ConfirmOptions = {
                             header: msg.header,
                             body: msg.body,
                             size: "large",
                             copyable: msg.copyable,
-                            hideAgree: true,
                             disagreeLbl: lf("Close")
-                        }).done();
+                        };
+                        const hasTrustedLink = msg.linkButtonHref && core.getSimulatorModalUrls().indexOf(msg.linkButtonHref) !== -1;
```
pelikhan: trusted urls may be undefined

PR #1901, opened 2017-04-14
Execution trace ui

riknoll:
Also slows the execution speed and only pauses for user code (i.e. code in main or `main.ts`)

trace_ui


microsoftsam:
Looks good

PR #1951, opened 2017-04-23
Add runtime check to generated JS code

tballmsft:
Added check in generated JS code to make sure we don't try to dereference a property of null/undefined.

### DIFF
--- a/pxtcompiler/emitter/backjs.ts
+++ b/pxtcompiler/emitter/backjs.ts
@@ -332,11 +332,16 @@ switch (step) {
                     args.splice(1, 0, procid.mapIdx.toString())
                     write(`  s.retval = ${shimToJs(procid.mapMethod)}(${args.join(", ")});`)
                     write(`  ${frameRef}.fn = doNothing;`)
-                    write(`} else`)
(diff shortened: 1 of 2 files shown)


pxtcompiler/emitter/backjs.ts
```diff
@@ -332,11 +332,16 @@ switch (step) {
                     args.splice(1, 0, procid.mapIdx.toString())
                     write(`  s.retval = ${shimToJs(procid.mapMethod)}(${args.join(", ")});`)
                     write(`  ${frameRef}.fn = doNothing;`)
-                    write(`} else`)
+                    write(`} else {`)
                 }
+                write (`pxsim.check(typeof ${frameRef}.arg0  != "number", "Can't access property of null/undefined.")`)
```
mmoskal: Why is `arg0` a number? I guess it's `0`? If so, we should have `check(!!${frameRef}.arg0, ...)` - as we align with JS semantics we should switch to `null` being represented by `null` etc.

tballmsft: Currently, we are using numbers. I agree we should align (later).

PR #1978, opened 2017-04-26
Decompiling literals and Custom Field editors in targets

microsoftsam:
Support for decompiling literals and custom Blockly field editors in pxt and pxt targets. 

Decompiling literals:
We now decompile literals that have a shadow type into a shadow block with that type. 
eg: 
```
music.playTone(277, music.beat(BeatFraction.Whole))
```
now decompiles to a note field editor: 
screen shot 2017-04-25 at 6 00 07 pm
- Added decompiler tests

Blockly field editors now live in PXT. 
- Moved the note editor and the grid picker into pxt. 

We now have support for PXT targets to author custom field editors. This can be done using the extension model like so: 
```
/// <reference path="../node_modules/pxt-core/built/pxteditor.d.ts" />

namespace pxt.editor {
    initExtensionsAsync = function(opts: pxt.editor.ExtensionOptions): Promise<pxt.editor.ExtensionResult> {
        pxt.debug('loading pxt-sample target extensions...')
        const res: pxt.editor.ExtensionResult = {
            fieldEditors: [{
                selector: "test",
                editor: SampleFieldTest
            },{
                selector: "angle",
                editor: SampleFieldAngle
            }]
        };
        return Promise.resolve<pxt.editor.ExtensionResult>(res);
    }
}
```

Breaking change for Authoring field editors:
Field editors are now authored like so:
- param.fieldEditor to indicate a parameters field editor
- param.fieldOptions to indicate options that should be passed to the field editor
In cases where you want to pass options to the shadow block generated, use: 
- param.shadowOptions

eg: 
```
    //% i.fieldEditor="gridpicker"
    //% i.fieldOptions.width="400" i.fieldOptions.columns="5"
    //% i.fieldOptions.itemColour="black" i.fieldOptions.tooltips="true"
```


abchatra:
We need to a final fix to one of the our targets, don't merge until then.

pxtlib/service.ts
```diff
@@ -149,7 +148,9 @@ namespace ts.pxtc {
         paramMin?: pxt.Map<string>; // min range
         paramMax?: pxt.Map<string>; // max range
         // Map for custom field editor parameters
-        blockFieldEditorParams?: pxt.Map<string>;
+        paramFieldEditor?: pxt.Map<string>; //.fieldEditor
+        paramShadowOptions?: pxt.Map<pxt.Map<any>>; //.shadowOptions.
```
riknoll: Do we have to use `any` here?

pxtblocks/fields/field_gridpicker.ts
```diff
@@ -0,0 +1,321 @@
+/// <reference path="../../localtypings/blockly.d.ts" />
+
+namespace pxtblockly {
```
microsoftsam: Wanting to keep field editors separate. They shouldn't have to share much of the logic with pxt.blocks. That way we don't have to worry about conflicting util functions etc.

PR #1981, opened 2017-04-26
Fix source interval mapping for blocks with multiple handlers

### LINKED ISSUE #1958: Slomo mode looks backwards!
In the following program
https://pxt.microbit.org/_W6ECcWK1TD7c

On Chrome 57, I get the order of steps appearing weirdly when using slowmo mode:
step through

riknoll:
Fixes #1958


pelikhan:
Tested and worked for me.

PR #1990, opened 2017-04-27
Friendly namespace

tballmsft:
use block attribute on namespace as friendly category name:
```
//% color=#B4009E weight=90
//% block="Shift Instructions"
namespace shift { }
```
Yields

image

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -176,9 +176,10 @@ namespace pxt.blocks {
             let isAdvanced = nsn && nsn.attributes.advanced;
 
             if (nsn) ns = nsn.attributes.block || ns;
-            let catName = ns
-            if (!catName)
-                catName = ts.pxtc.blocksCategory(fn);
+            let catName = ts.pxtc.blocksCategory(fn);
+            if (nsn && nsn.attributes.block)
+                catName = nsn.attributes.block
+
             let category = categoryElement(tb, catName);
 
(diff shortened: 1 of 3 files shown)


pelikhan:
Could you describe the change?

tballmsft:
It computes into the same variable as before, so whatever we did to localize that holds.

pelikhan:
We get the namespace localization via the blocks loc files.

tballmsft:
So, is there something I need to do? I am just providing a different way for the dev to provide the category name.

pelikhan:
Use the "block" attribute instead and it should just work (TM).

tballmsft:
What is the "block" attribute? There are many "block" attributes.

pelikhan:
There is one called “block” in the metadata. It is currently not used for namespaces.

pxtblocks/blocklyloader.ts
```diff
@@ -176,7 +176,9 @@ namespace pxt.blocks {
             let isAdvanced = nsn && nsn.attributes.advanced;
 
             if (nsn) ns = nsn.attributes.block || ns;
-            let catName = ts.pxtc.blocksCategory(fn);
+            let catName = ns
```
pelikhan: nitpick: ``let catName = ns || ts.pxtc.blocksCategory(fn)``

tballmsft:
This is failing on pxt-microbit. I think block attribute on namespace is being used to some other effect.

tballmsft:
Working fine for Blockly. Need to add code to do the same for Monaco toolbox: createCategoryElement

tballmsft:
All done.

PR #1993, opened 2017-04-28
Lang picker

guillaumejenkins:


svgicons/globe.svg
```diff
@@ -0,0 +1,62 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
pelikhan: Use svg online optimizer to clean up

PR #2014, opened 2017-05-02
Support target experiments

microsoftsam:
This would be useful for switching certain target flags for key experiments. 

For example: If I wanted to run an experiment where the following flags in pxt-microbit are switched to false:
```
        "invertedMenu": false,
        "coloredToolbox": false,
        "monacoToolbox": false,
        "hasAudio": false,
```

In the target, under experiments, author it like this: 
```
    "experiments": {
        "coolexperiment": {
            "name": "My Cool experiment",
            "description": "Cool experiment description ",
            "appTheme": {
                "invertedMenu": false,
                "coloredToolbox": false,
                 "monacoToolbox": false,
                 "hasAudio": false,
             }
        }
    }
```

To get to this experiment, I just need to go to pxt.microbit.org?coolexperiment=1

Note: I added the name and description field, so we're able to pop this up in some way to our beta users and give them the option to look at these experiment, and check them out :)

### DIFF
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -35,6 +35,13 @@ declare namespace pxt {
         appTheme: AppTheme;
         compileService?: TargetCompileService;
         analytics?: AppAnalytics;
+        experiments?: pxt.Map<AppExperiment>;
+    }
+
+    interface AppExperiment {
+        name: string;
+        description?: string;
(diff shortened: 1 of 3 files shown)


pelikhan:
- when an experiment is active, it should be traced in the analytics
- in experiment mode, we should have a proeminent feedback mechanism

PR #2020, opened 2017-05-02
Fix missing blocks and duplicated categories when language is not EN

### LINKED ISSUE #2001: On beta site, languages other than English have two Maths entries on the toolbox
Go to https://pxt.microbit.org/beta and choose French from the Language menu.

capture

guillaumejenkins:
`nsn.attributes.block` is already translated; for the category's `nameid`, we want the non-translated name

Fixes #2001


pelikhan:
Do you see them show up In the generated strings.json?

pelikhan:
Good to go then.

PR #2021, opened 2017-05-03
Now clears language in the URL hash when selecting a new language in the UI

### LINKED ISSUE #1998: Changing language using settings doesn't work if the URL already has a language
1. Goto https://pxt.microbit.org/beta?lang=fr  
2. Change language using settings.
3. Language change options don't reflect.

Looks like one can change language only in English version of the website.

guillaumejenkins:
Fixes #1998

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -1763,11 +1763,10 @@ export function serveAsync(parsed: commandParser.ParsedCommand) {
     } else if (parsed.flags["pkg"]) {
         justServe = true
         packaged = true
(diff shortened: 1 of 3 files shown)


pelikhan:
I do not understand this change. ``https://pxt.microbit.org/beta#lang:fr`` does not see to be handled in beta.

pelikhan:
The hash handler should always handle "lang" and automatically swallow it. Then you do not need to do this patching.

PR #2031, opened 2017-05-03
Add definitions for more math functions

mmoskal:
Need to think if we want these functions in integer targets at all

CC @tballmsft @pelikhan @abchatra @riknoll @microsoftsam


mmoskal:
OK, no integer targets, so this is easy - ready to merge

PR #2037, opened 2017-05-04
Fix missing blocks and duplicated categories when language is not EN …

guillaumejenkins:
…(#2020)

* Fix missing blocks and duplicated categories when language is not EN

* We now localize friendly namespace names


pelikhan:
Is this PR still relevant?

PR #2047, opened 2017-05-09
Audio modulator support

microsoftsam:
Audio modulator support

Fix issues with rtlcss and autoprefixer. 
Added cleancss to the css pipeline (OSS request approved)
Support for custom download text and download icon

In target: 
Support for custom deploy async method
Support for specifying a custom before compile method
Support for building editor extensions with module loaders and browserify
UI components for audio modulator


webapp/src/cmds.ts
```diff
@@ -1,4 +1,5 @@
 /// <reference path="../../built/pxtlib.d.ts"/>
+/// <reference path="../../localtypings/modulator.d.ts"/>
```
mmoskal: Should this file be checked in?

cli/cli.ts
```diff
@@ -1213,7 +1213,15 @@ export function buildTargetAsync(): Promise<void> {
         .then(buildTargetCoreAsync)
         .then(buildSemanticUIAsync)
         .then(() => buildFolderAsync('cmds', true))
-        .then(() => buildFolderAsync('editor', true, 'editor'))
+        .then(() => {
+            if (fs.existsSync(p + "/tsconfig.json")) {
```
pelikhan: use ``path.join(`` in future

webapp/src/app.tsx
```diff
@@ -1549,6 +1552,7 @@ ${compileService ? `<p>${lf("{0} version:", "C++ runtime")} <a href="${Util.html
         const trace = run && simOpts.enableTrace;
         const fullscreen = run && !inTutorial && !simOpts.hideFullscreen
         const audio = run && !inTutorial && targetTheme.hasAudio;
+        const useModulator = pxt.appTarget.compile.useModulator;
```
pelikhan: compile might be undefined

theme/common.less
```diff
@@ -716,6 +716,12 @@ Avatar
     width: 0;
 }
 
+/* Modulator */
```
pelikhan: why do we need this?

webapp/src/app.tsx
mmoskal: no, it cannot https://github.com/Microsoft/pxt/blob/master/pxtlib/main.ts#L18

theme/common.less
microsoftsam: UI elements for the modulator are defined in pxt. and we should hide them for all targets that don't need them.

webapp/src/cmds.ts
microsoftsam: Moved to target.

microsoftsam: removing

PR #2065, opened 2017-05-09
Support for decompiling field editors when they are on the block

microsoftsam:
If a block has a field editor inline, we need a way to decompile literals into a field with that value passed to the field editor.

### DIFF
--- a/pxtcompiler/emitter/decompiler.ts
+++ b/pxtcompiler/emitter/decompiler.ts
@@ -1178,8 +1178,8 @@ ${output}</xml>`;
                             } else if (info.attrs && info.attrs.paramFieldEditor && info.attrs.paramFieldEditorOptions) {
                                 if (info.attrs.paramFieldEditorOptions[vName] && info.attrs.paramFieldEditorOptions[vName]['onParentBlock']) {
                                     (r.fields || (r.fields = [])).push(getField(vName, e.getText()));
+                                    return;
                                 }
-                                return;
                             }
                         }
                         if (defaultV) {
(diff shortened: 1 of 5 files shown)


pxtcompiler/emitter/decompiler.ts
```diff
@@ -1150,29 +1150,42 @@ ${output}</xml>`;
...
+                                        v = {
+                                            kind: "value",
+                                            name: vName,
+                                            value: fieldBlock
+                                        };
+                                        defaultV = false;
+                                    }
+                                }
+                            } else if (info.attrs && info.attrs.paramFieldEditor && info.attrs.paramFieldEditorOptions) {
+                                let onBlockFieldName = '';
```
riknoll: Shouldn't this be the same as vName?

microsoftsam: Yep.

pxtcompiler/emitter/decompiler.ts
```diff
@@ -1150,30 +1150,36 @@ ${output}</xml>`;
...
-                                        name: vName,
-                                        value: fieldBlock
-                                    };
-                                    defaultV = false;
                                 }
+                            } else if (info.attrs && info.attrs.paramFieldEditor && info.attrs.paramFieldEditorOptions) {
+                                if (info.attrs.paramFieldEditorOptions[vName] && info.attrs.paramFieldEditorOptions[vName]['onParentBlock']) {
+                                    (r.fields || (r.fields = [])).push(getField(vName, e.getText()));
+                                }
+                                return;
```
riknoll: I think this return should move up a line

microsoftsam: yep my bad.

PR #2070, opened 2017-05-10
run "py" on windows

pelikhan:
When installing python3 on windows, the "py" command is registered.

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -4374,7 +4374,7 @@ function initCommands() {
         argString: "<package-directory> <support-directory>...",
         anyArgs: true,
         advanced: true,
-    }, c => pyconv.convertAsync(c.arguments, /^win/i.test(os.platform()) ? "py" : "python"))
+    }, c => pyconv.convertAsync(c.arguments, /^win/i.test(os.platform()) ? "py" : "python3"))
 
     advancedCommand("thirdpartynotices", "refresh third party notices", thirdPartyNoticesAsync);
     p.defineCommand({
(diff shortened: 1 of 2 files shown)


cli/cli.ts
```diff
@@ -4374,7 +4374,7 @@ function initCommands() {
         argString: "<package-directory> <support-directory>...",
         anyArgs: true,
         advanced: true,
-    }, c => pyconv.convertAsync(c.arguments))
+    }, c => pyconv.convertAsync(c.arguments, /^win/i.test(os.platform()) ? "py" : "python"))
```
pelikhan: that should be python3... fixing

cli/pyconv.ts
```diff
@@ -1739,15 +1739,19 @@ export function convertAsync(fns: string[]) {
         }
     }
 
+    const pkgFilesKeys = Object.keys(pkgFiles);
+    pxt.log(`files (${pkgFilesKeys.length}):\n   ${pkgFilesKeys.join('    \n')}`);
+
     return nodeutil.spawnWithPipeAsync({
-        cmd: "python3",
+        cmd: /^win/i.test(process.platform) ? "py" : "python3",
```
carlosperate: If you have Python2 installed as well `py` will most likely resolve to the python2 installation. So if you definitely need to run a 3.x version perhaps `py -3` would be better. I would assume the `-3` flag will work even if you only have a single version of python 3.x installed, but perhaps it would be better to confirm that.

mmoskal: We definitely need python3, python2 cannot parse MicroPython modules.

PR #2075, opened 2017-05-10
Customize built in colors in target

microsoftsam:
Support ability for targets to alter built in namespace colors in pxtarget.


microsoftsam:
Mind looking into that @guillaumejenkins, we just need a way to get the direct category children there.

PR #2087, opened 2017-05-12
Clean up some random type casting

mmoskal:
@pelikhan any idea why this stuff was here? it is just wrong


pelikhan:
I moved this from Daryl's prototype a while ago.

PR #2090, opened 2017-05-12
Arrays/Lists support in blocks

riknoll:
Adds support for arrays and some placeholder docs for all of the blocks. Arrays are still hidden behind a flag, so they will only show up if enabled in the target. However, the decompiler will decompile arrays regardless of if they are enabled or not (that matches the behavior for the other builtin categories). This also adds a block for iterating over arrays in the loops category (compiles to `for...of`) that only shows up if lists are enabled.

This is a pretty big change to the Blockly type inference, so I would appreciate it if people try it out. This should be considered a beta feature for now. I have some tests for the Blockly compiler as well but I'm going to open those as a separate PR. 

Once this PR is merged, I'll work on porting it over to v0.


libs/pxt-common/pxt-helpers.ts
```diff
@@ -94,6 +94,29 @@ namespace helpers {
         }
         return initialValue
     }
+
+    export function arraySlice<T>(arr: T[], start: number, end: number): T[] {
+        const res: T[] = [];
+
+        if (start < 0) {
```
pelikhan: this does not match the JS slice semantics

pxtlib/blocks.ts
```diff
@@ -96,6 +96,11 @@ namespace pxt.blocks {
                 url: 'blocks/loops/for',
                 category: 'loops'
             },
+            'controls_for_of': {
+                name: Util.lf("a loop that repeats for each value in an array"),
+                url: 'blocks/loops/for_of',
```
pelikhan: use `-` instead of `_` in doc links

pelikhan: ``list`` or ``array``s?

mmoskal: I would stick with JS convention - arrays.

riknoll: I figured I would leave that up to whoever writes the proper docs for this.

cli/cli.ts
```diff
@@ -1888,9 +1888,14 @@ class SnippetHost implements pxt.Host {
                     "name": this.name,
                     "dependencies": this.dependencies,
                     "description": "",
-                    "files": [
+                    "files": this.includeCommon ? [
                         "main.blocks", //TODO: Probably don't want this
-                        "main.ts"
+                        "main.ts",
+                        "pxt-core.d.ts",
```
pelikhan: is this for testing? aren't those files auto-added?

common-docs/blocks/lists.md
```diff
@@ -0,0 +1,14 @@
+# Lists
```
pelikhan: arrays?

pxtcompiler/emitter/decompiler.ts
```diff
@@ -1481,20 +1544,37 @@ ${output}</xml>`;
...
-                case SK.EqualsToken:
-                    return checkExpression(n.right, blocksInfo);
-                case SK.PlusEqualsToken:
-                case SK.MinusEqualsToken:
-                    return undefined;
-                default:
-                    return Util.lf("Unsupported operator token in statement {0}", SK[n.operatorToken.kind]);
+            if (n.left.kind === SK.ElementAccessExpression) {
+                if (n.operatorToken.kind !== SK.EqualsToken) {
+                    return Util.lf("Element access expresions may only be assigned to using =");
```
pelikhan: expressions

pelikhan: `=`

PR #2091, opened 2017-05-12
File download in extensions

guillaumejenkins:
These changes let targets that override `deployCoreAsync()` download files and show confirm dialogs. This is required for targets that use a modulator deployment.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -965,6 +965,7 @@ export class ProjectView
                 resp.saveOnly = saveOnly
                 resp.userContextWindow = userContextWindow;
                 resp.downloadFileBaseName = pkg.genFileName("");
+                resp.confirmAsync = core.confirmAsync;
                 return pxt.commands.deployCoreAsync(resp)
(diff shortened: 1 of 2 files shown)


webapp/src/app.tsx
```diff
@@ -964,9 +964,11 @@ export class ProjectView
                 }
                 resp.saveOnly = saveOnly
                 resp.userContextWindow = userContextWindow;
+                resp.downloadFileBaseName = pkg.genFileName("");
+                resp.confirmAsync = core.confirmAsync;
                 return pxt.commands.deployCoreAsync(resp)
                     .catch(e => {
-                        core.warningNotification(lf(".hex file upload failed, please try again."));
+                        core.warningNotification(pxt.appTarget.compile.useModulator ? lf("Upload failed, please try again.") : lf(".hex file upload failed, please try again."));
```
pelikhan: Can we use a single message string here? Drop the ".hex file"

pelikhan:
Minor editor otherwise LGTM

Issue #2067, opened 2017-05-09
P1 is not listed under P2


DavidWhaleMEF:
Teacher reported that children were finding this really confusing, that P1 did not appear before P2

![screen shot 2017-05-10 at 00 08 29](https://cloud.githubusercontent.com/assets/25079586/25876489/76b74e46-3515-11e7-9368-128541d8b6fc.png)

![screen shot 2017-05-10 at 00 12 24](https://cloud.githubusercontent.com/assets/25079586/25876491/7996c20e-3515-11e7-8077-0d6a49dbd20c.png)

tballmsft:
Several problems need to be addressed. Analog pins are in strange order:

![image](https://cloud.githubusercontent.com/assets/10767146/26018029/72787784-3721-11e7-8de8-a50917597100.png)

tballmsft:
And digital pins not in correct order:

![image](https://cloud.githubusercontent.com/assets/10767146/26018050/8e99172a-3721-11e7-9987-ec06ec7a400e.png)

tballmsft:
the weird sort that is causing the above is:
```
                    // if a value is provided, move it first
                    if (pr.shadowValue)
                        dd.sort((v1, v2) => v1[1] == pr.shadowValue ? -1 : v2[1] == pr.shadowValue ? 1 : 0);
```
in blocklyloader.ts

tballmsft:
@mmoskal - the sort you added above back in December is causing this problem. Can you please take a look.  For some reason, the two enums generate shadowValue and then get sorted by the above code , resulting in a bad order.

pelikhan:
@tballmsft let's display the pins in a grid. It's much easier to scan them. There are too many elements and the drop down continuously falls out of the screen.

mmoskal:
I'm not sure what this is supposed to do - @pelikhan added it in https://github.com/Microsoft/pxt/commit/f86d55824c132bf7f5375650ba2ee8cb9f3cff39 , I only moved it a few lines

guillaumejenkins:
Fixed in pxt-core@0.12.70 and pxt-microbit@0.12.10

Fix is now in /beta website

PR #2096, opened 2017-05-15
Fix logic for showing shadowvalue at the top of dropdown

guillaumejenkins:
Fixes https://github.com/Microsoft/pxt/issues/2067

Idea is to simply extract the shadow value and insert it back to the front


pelikhan:
Does this fix this issue?

pelikhan:
We should use a grid display for gesture.

PR #2105, opened 2017-05-16
Change grid picker looks for text items

guillaumejenkins:
- Item background color is now slightly faded instead of pure white
- Column widths are now uniform

Before:
image

After:
image


abchatra:
@pelikhan I requested the changes. White background was not blending well with the rest of the editor. New background color is same as the one selected. 

Can we change now and fly this through the UI design meeting tomorrow?

pelikhan:
sure

PR #2109, opened 2017-05-16
Setup field editors in docs

microsoftsam:
This fixes the docs loading issue.

### DIFF
--- a/docfiles/docs.js
+++ b/docfiles/docs.js
@@ -135,6 +135,20 @@ function setupSemantic() {
         $('.ui.footer').append($('<div class="ui center aligned small container"/>').text('user agent: ' + navigator.userAgent))
 }
 
+function setupBlockly() {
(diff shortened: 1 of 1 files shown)


docfiles/docs.js
```diff
@@ -146,6 +160,7 @@ function renderSnippets() {
     ksRunnerReady(function () {
         setupSidebar();
         setupSemantic();
+        setupBlockly();
```
pelikhan: there's a race condition here. "setBlockly" should return a promise.

pelikhan: and renderAsync should start after it is done.

docfiles/docs.js
```diff
@@ -136,9 +136,10 @@ function setupSemantic() {
 }
 
 function setupBlockly() {
```
pelikhan: Functions that return promises should end with Async

PR #2110, opened 2017-05-17
Add beta tag if target has a beta URL.

microsoftsam:
screen shot 2017-05-16 at 5 37 06 pm


abchatra:
:shipit:

Jaqster:
It's okay.  I like the Minecraft Beta tag better - more subtle, and this looks a bit squished up in the corner there...  Don't worry about changing for now, but let's see if we can do this more elegantly in the future.

PR #2135, opened 2017-05-22
Moving all tests to one folder and framework

riknoll:
Refactoring all of our tests to use Mocha and Chai. I also split up `lang-test0.ts` into 39 files and made all the tests run individually instead of altogether. I left a `pxt.json` file for running them all at once if you want to test them on hardware. I also deleted the `tests-ts/` folder because it seemed like those tests were never used and it only contained two files. The only tests I did not touch were the package conflict tests because we don't run them.

I'm adding a do not merge label to this because right now the lang tests take ~3 minutes to complete.


abchatra:
:shipit:

PR #2136, opened 2017-05-22
Dont decompile any whitespace other than space

### LINKED ISSUE #2114: control characters in serial.writestring don't round trip properly
microbit.org-ticket: 1815

User is trying to send control codes via serial to control a receiving end terminal emulator.

I tried this, and it seems that every time I switch from javascript to blocks mode, PXT rewrites the code to something different.

Here are successive switches back and forth between javascript and blocks and back again, see how the round-tripping rewrites it to something different every time. This is non terminating I think (i.e. every time I switch back and forth, it rewrites it to something different)

```
    serial.writeString("\r")
})


    serial.writeString("\n")
})


    serial.writeString("\n\n            ")
})


    serial.writeString(`

            
            `)
})


    serial.writeString(`

            
            
            `)
})
```

riknoll:
Fixes #2114


riknoll:
Note that this does affect template strings but should not affect `showLeds()` because we handle that as a special case.

pelikhan:
What about ``hex`` template like ?

``` hex `abc
de` ```

is a valid literal buffer

riknoll:
@pelikhan do hex templates decompile right now?

This change should only affect strings with whitespace characters other than space (e.g. \n, \r, \t, \v, etc). We currently don't support those in the blocks and you get weird behavior if you use them. For example, using \v will delete your code altogether (it's an illegal character in XML). Using \n or \r round trips to garbage.

pelikhan:
Get it in, hex templates do not roundtrip i think.

PR #2138, opened 2017-05-23
Emit breakpoints for loop conditionals

### LINKED ISSUE #2130: slow-mo mode doesn't highlight conditional in loop
For the following code:
```
let x = 0
    basic.showIcon(IconNames.SmallHeart)
})
x = 1
while (x > 0) {
    x = x + 1
    basic.pause(100)
}
```
You'll see that the two statements in the while loop are highlighted during slow-mode (snail) execution but the loop expression never is:

image

riknoll:
Fixes #2130 

trace_loop_conditionals

### DIFF
--- a/pxtcompiler/emitter/emitter.ts
+++ b/pxtcompiler/emitter/emitter.ts
@@ -3,7 +3,7 @@
 
 namespace ts.pxtc {
 
-    // in tagged mode, 
+    // in tagged mode,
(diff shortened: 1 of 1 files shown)


mmoskal:
LGTM.

In general there's is an issue with breakpoints when the stack is not empty, which could be the case inside of an expression, but here it's right before the jump so the stack should be empty.

PR #2143, opened 2017-05-23
Loc search

### LINKED ISSUE #2064: Search with localized text
The search functionality doesn't allow non-English users to search with localized text.

guillaumejenkins:
Addresses the following:

- Advanced categories were not searchable unless Advanced was opened
- Built-in blocks were not searchable by their block text, making it impossible to find most built-in blocks
- Search was not localized (fixes #2064)

Overview of the changes:

- The search worker did not have access to localized strings and localized APIs, so the webapp now passes that info in the search options
- When displaying the search results, if the current toolbox has Advanced collapsed, we now create a separate, full toolbox to find blocks that would otherwise be missed because they're in collapsed categories (which are not loaded in the current TB definition)
- Block text for built-in Blockly blocks were hardcoded in blocklyloader.ts, which means the search didn't have access to them. These texts were moved to pxtlib/blocks.ts so the search can query them


pxtblocks/blocklyloader.ts
```diff
@@ -1202,17 +1204,17 @@ namespace pxt.blocks {
...
     function installBuiltinHelpInfo(id: string) {
-        const info = pxt.blocks.helpResources()[id];
-        installHelpResources(id, info.name, info.tooltip, info.url, getNamespaceColor(info.category))
+        const info = pxt.blocks.getBlockDefinition(id);
+        installHelpResources(id, info.name, info.tooltip, info.url, getNamespaceColor(info.category));
     }
 
     function setHelpResources(block: any, id: string, name: string, tooltip: any, url: string, colour: string) {
-        if (tooltip) block.setTooltip(tooltip);
+        if (tooltip && typeof tooltip === "string") block.setTooltip(tooltip);
```
pelikhan: why?

guillaumejenkins: Blockly expects a string or a function that returns a string, so for our tooltips that are a string map, we should not pass those to Blockly (instead, we handle them in our code card)

I'll update this check to include functions

PR #2145, opened 2017-05-23
Fix race condition in tutorial when serving locally in ie11

riknoll:


pxtrunner/runner.ts
```diff
@@ -332,6 +332,7 @@ namespace pxt.runner {
                 let dm = m as pxsim.SimulatorDocMessage;
                 if (dm && dm.localToken) {
                     Cloud.localToken = dm.localToken;
+                    pendingLocalToken.forEach(p => p());
```
pelikhan: clear array?

riknoll: Oh yeah, whoops 🙂

PR #2159, opened 2017-05-25
Support undetectable blocks

microsoftsam:
Support for undeletable blocks through annotations: 
//% undeletable=true

Additionally: Added onStartUnDeletable: true in the runtime to set the undeletable state of onstart.

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -573,6 +573,8 @@ namespace pxt.blocks {
 
         block.setTooltip(fn.attributes.jsDoc);
         block.setColour(color);
+        if (fn.attributes.undeletable)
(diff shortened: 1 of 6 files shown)


pxtblocks/blocklyloader.ts
```diff
@@ -573,6 +573,8 @@ namespace pxt.blocks {
 
         block.setTooltip(fn.attributes.jsDoc);
         block.setColour(color);
+        if (fn.attributes.undeletable)
+            block.setDeletable(true);
```
riknoll: Shouldn't this be false?

microsoftsam: if undeletable is true. deletable should be false. Yes.

PR #2182, opened 2017-05-30
Build common sim as part of the target build

riknoll:
See https://github.com/Microsoft/pxt-common-packages/pull/63

This builds the simulator code in pxt-common-packages and copies it into the target's built directory. If you have a better idea for how the build should work, let me know.

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -1210,7 +1210,8 @@ function maxMTimeAsync(dirs: string[]) {
 export function buildTargetAsync(): Promise<void> {
     if (pxt.appTarget.id == "core")
         return buildTargetCoreAsync()
-    return simshimAsync()
+    return buildCommonSim()
+        .then(simshimAsync)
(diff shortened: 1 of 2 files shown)


cli/cli.ts
```diff
@@ -1252,6 +1253,31 @@ function buildFolderAsync(p: string, optional?: boolean, outputName?: string): P
...
+        pxt.log(`building common-sim...`)
+        dirsToWatch.push(p)
+        return nodeutil.spawnAsync({
+            cmd: "node",
+            args: ["../typescript/bin/tsc"],
+            cwd: p
+        })
+        .then(() => {
+            if (fs.existsSync("built")) {
+                nodeutil.cpR(path.join(p, "built"), "built");
```
pelikhan: we should only be copying 2 files: ``pxt-common-sim.d.ts`` and ``pxt-common-sim.js``

PR #2190, opened 2017-05-30
Adding simulator editor events

riknoll:
Adds simulator events for the editor including the usual start, stop, resume, exception, and a new one for when the top level code has finished executing.


microsoftsam:
As discussed, honour the allowParentController flag when sending these messages

PR #2210, opened 2017-06-02
Preserve embedded pages across reloads

riknoll:
Fixes the bug where if we publish a release and you navigate to a page with our editor embedded, then the embedded editor will update and the embedded project will be replaced with the template project.


pelikhan:
Can we cherry-pick?

riknoll:
Sure thing

PR #2235, opened 2017-06-06
Don't decompile literals in custom field editors by default

riknoll:
Don't merge this just yet because it requires updating the authoring of a few APIs.


pelikhan:
@riknoll conflicts

riknoll:
This is good to merge. Just need to make sure to bump the pxt-common-packages in a target if we bump to this version of PXT.

pelikhan:
Can you merge and bump?

PR #2236, opened 2017-06-07
Fix execution trace for statements with comments

### LINKED ISSUE #2224: After adding comment for a block, in "Slow-Mo" mode this block will not be selected
**Repro steps:**
• Navigate to https://makecode.adafruit.com/
• Drag two "show string" and put them under "forever"
• Click "Slow-Mo" and observe the contemt pane
• Add comment for one block "show string" and observe the content pane again

**Actual result:**

riknoll:
Fixes #2224 

This changes the behavior in Monaco so that multi-line statements (e.g. if-statements) get highlighted in their entirety instead of just the first line during slow-mo. It looks like this:

slow_mo_comments

This also fixes the highlighting in blocks for code with comments, which was previously broken.

### DIFF
--- a/pxtblocks/blocklycompiler.ts
+++ b/pxtblocks/blocklycompiler.ts
@@ -1262,15 +1262,15 @@ namespace pxt.blocks {
         }
         let l = r[r.length - 1]; if (l) l.id = b.id;
 
-        r.forEach(l => {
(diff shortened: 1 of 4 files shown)


pelikhan:
@riknoll  Cherry-pick?

PR #2237, opened 2017-06-07
Hide the sidedocs when we popout

microsoftsam:
Hide the sidedocs when we choose to popout the content into a new tab.


pelikhan:
@sam cherry-pick?

microsoftsam:
Will cherrypick

PR #2257, opened 2017-06-09
Fix tutorial rendering. 

### LINKED ISSUE #2189: Tutorial - instruction clarification
Step 5 on the Getting Started Tutorial - add "in an On Shake block" to the instruction text.

gettingstarted

microsoftsam:
Fix bug in tutorial rendering logic. Fixes #2189

The issue was after the full page was rendered by markdown, I had HTML in string format, and I needed the first element of each step. I was using regex to parse the HTML. I am now using the contentdiv passed in and temporarily setting the HTML of each step, and extracting the first elements HTML. Using the DOM to do the parsing.

### DIFF
--- a/pxtrunner/runner.ts
+++ b/pxtrunner/runner.ts
@@ -597,7 +597,7 @@ ${files["main.ts"]}
                         let stepcontent = content.innerHTML.split(/<h3.*\/h3>/gi);
                         for (let i = 0; i < stepcontent.length - 1; i++) {
                             content.innerHTML = stepcontent[i + 1];
-                            stepInfo[i].headerContent = content.firstElementChild.innerHTML;
+                            stepInfo[i].headerContent = `<p>` + content.firstElementChild.innerHTML + `</p>`;
                             stepInfo[i].content = stepcontent[i + 1];
                         }
                         content.innerHTML = '';
--- a/pxtrunner/runner.ts
+++ b/pxtrunner/runner.ts
@@ -596,9 +596,11 @@ ${files["main.ts"]}
                         // Split the steps
                         let stepcontent = content.innerHTML.split(/<h3.*\/h3>/gi);
                         for (let i = 0; i < stepcontent.length - 1; i++) {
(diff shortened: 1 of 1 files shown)


microsoftsam:
Needs to be cherrypicked into v0

pxtrunner/runner.ts
```diff
@@ -596,9 +596,11 @@ ${files["main.ts"]}
                         // Split the steps
                         let stepcontent = content.innerHTML.split(/<h3.*\/h3>/gi);
                         for (let i = 0; i < stepcontent.length - 1; i++) {
-                            stepInfo[i].headerContent = stepcontent[i + 1].split(/(<.*?>.*<\/.*?>)/i)[1];
+                            content.innerHTML = stepcontent[i + 1];
```
pelikhan: Isn't there a dom parser to do this kind of stuff?

microsoftsam: There is most definitely a library, but I don't see a need to add another dependency. 
I have figured out a way to use the content element (ie: the passed in DOM) to do the parsing.

pelikhan:
The browser has a DOMParser right? Does your trick influence the HTML layout?

microsoftsam:
Nope my trick doesn't influence the HTML layout, as in the case of the tutorial project the content div is never returned, instead the content is returned in messages. The tutorial iframe is 1px by 1px iframe.
Yes I'm using the browser's DOM parser with this.

microsoftsam:
This needs to be tested on firefox, etc. but I think the reason this trick works is because the content div is already added to the tutorial iframe's DOM. Otherwise you get weird rendering issues, because the font sizes and such aren't passed in.

pelikhan:
Sounds good then.

pelikhan:
Oh font size computation. Makes sense now. Yes it has to be in the DOM.

microsoftsam:
Fails in FF, back to the drawing board.

microsoftsam:
Fixed in FF.

microsoftsam:
Need to test on IE / Edge and Windows.

PR #2261, opened 2017-06-09
Don't always filter out the variables category in tutorials

riknoll:
In tutorial mode, we always remove the variables category even if you use variable blocks in your code because the category itself doesn't have any child blocks in its definition. This fixes that bug.


riknoll:
Whoops, forgot to handle the namespace filters

riknoll:
Fixed

PR #2263, opened 2017-06-11
Fix HID serial on mac

### LINKED ISSUE #2259: SerialPort (for pxt-microbit) not working on Mac
@pelikhan mentioned that it might be something related to the regex for understanding which device it is.

Note that I was serving pxt from the master branch.

microsoftsam:
Fixing hid serial on mac for detecting the micro:bit. 
Fixes #2259

### DIFF
--- a/cli/hid.ts
+++ b/cli/hid.ts
@@ -50,7 +50,7 @@ export function getHF2Devices() {
     const hid = getHID();
     if (!hid) return [];
     let devices = hid.devices() as HidDevice[]
-    return devices.filter(d => (d.release & 0xff00) == 0x4200)
+    return devices.filter(d => (d.release & 0xff00) == 0x4200 || (d.release & 0xff00) == 0x1000);
 }
 
 export function hf2ConnectAsync(path: string) {


microsoftsam:
Will cherry pick to master once this goes in.

mmoskal:
@pelikhan @microsoftsam  The HID serial uses our custom HF2 protocol. The microbit HID device uses DAPLink protocol. Also I don't believe it can run serial over it (but I'm not sure).

Merging this  will cause all sort of random devices being detected as HF2 devices.

microsoftsam:
@mmoskal is there somewhere else we test the device to see if its a microbit? in order to get serial working?
I thought the filter for release numbers was filtering for microbits on windows..

pelikhan:
microbit does not use hid, it uses serial-over-usb. See https://github.com/Microsoft/pxt/blob/master/cli/serial.ts#L66

microsoftsam:
Okay so did some debugging and the serial info for the microbit on windows is: 
```
comName: 'COM18',
  manufacturer: 'mbed',
  serialNumber: undefined,
  pnpId: 'USB\\VID_0D28&PID_0204&MI_01\\9900000040304E4500431003000000150000000097969901',
  locationId: 'mbed Serial Port',
  vendorId: '0D28',
  productId: '0204'
```

and on MAC it's: 

```
comName: '/dev/cu.usbmodem1422',
  manufacturer: 'ARM',
  serialNumber: '9900000040304e4500431003000000150000000097969901',
  pnpId: undefined,
  locationId: '0x14200000',
  vendorId: '0x0d28',
  productId: '0x0204' 
```

we are currently filtering for manufacturer: 
```
        "manufacturerFilter": "^mbed$",
```

so a simple fix would be to change pxtarget filter to: 
```
        "manufacturerFilter": "^(mbed|ARM)$",
```

which I've tested works, @mmoskal do you see any issues with that?
I wonder if this filters for all ARM devices, and not just micro:bits. The product ID is the same on both, should we switch to filtering based on the product id instead of manufacturer?

pelikhan:
Yes, we should use the vendor/product id.

mmoskal:
Yeah, the USB numeric ID is probably the way to go. OTOH maybe capturing any and all USB Serial ports wouldn't be that bad either.

PR #2293, opened 2017-06-15
Staticpkg appcache

dragonx:
Re-enable appcache for staticpkg

Seems like this was disabled for some reason, but I'm not sure why.


webapp/public/release.manifest
```diff
@@ -52,6 +52,8 @@ https://az416426.vo.msecnd.net/scripts/a/ai.0.js
 @workerjs@
 @monacoworkerjs@
 
+@timestamp@
```
pelikhan: What about the nom-static case? Does it get this macro get expanded as well?

dragonx:
At least on 'pxt serve', these macros are ignored, so the manifest isn't
included at all.

What other situations should I test for. So far those are the only two I've
used.

microsoftsam:
Thanks!

PR #2296, opened 2017-06-15
Fix parameter defaults in blocks and monaco

riknoll:
This fixes two issues:

1. We were parsing the default values for parameters in two places which led to the Monaco toolbox not picking up the ones set with the `eg.` syntax
2. The name of the field for the slider block changed at some point and we didn't patch blocklyloader (so blocks with slider inputs didn't have default values set)


pelikhan:
Is this ready to be merged?

PR #2307, opened 2017-06-16
Remove stale C++ files from build directories (as they confuse pio and yt)

mmoskal:
It should work but would be good to test on Windows - check if it says it's removing lots of files like `/source/core/core.cpp` etc.

### DIFF
--- a/cli/buildengine.ts
+++ b/cli/buildengine.ts
@@ -118,6 +118,15 @@ export function buildHexAsync(buildEngine: BuildEngine, mainPkg: pxt.MainPackage
     let allFiles = U.clone(extInfo.generatedFiles)
     U.jsonCopyFrom(allFiles, extInfo.extensionFiles)
 
+    for (let f of nodeutil.allFiles(buildEngine.buildPath)) {
+        let bn = f.slice(buildEngine.buildPath.length)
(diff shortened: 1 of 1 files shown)


cli/buildengine.ts
```diff
@@ -118,6 +118,15 @@ export function buildHexAsync(buildEngine: BuildEngine, mainPkg: pxt.MainPackage
     let allFiles = U.clone(extInfo.generatedFiles)
     U.jsonCopyFrom(allFiles, extInfo.extensionFiles)
 
+    for (let f of nodeutil.allFiles(buildEngine.buildPath)) {
+        let bn = f.slice(buildEngine.buildPath.length)
+        bn = bn.replace(/\\/g, "/").replace(/^\//, "/")
+        if (U.startsWith(bn, "/source/") && !allFiles[bn]) {
+            pxt.log("removing stale " + bn)
+            fs.unlinkSync(f)
```
pelikhan: you can also use rimraf to nuke a folder

mmoskal: I really only want to nuke files, and also not update timestamps of the ones that didn't change (to avoid excessive rebuilds)

PR #2315, opened 2017-06-18
Remove cookie notification from sandbox mode

abchatra:
It is the responsibility of the embed editor to show the cookie notification.


pelikhan:
Needs cherry picking

PR #2316, opened 2017-06-18
Functions!

microsoftsam:
Adding function support. 
- "Make a function" button creates a function block and adds it close to the top blockly block in the workspace. It also closes the flyout and highlights the newly created function block. 
- Functions flyout lists all the available functions to call (similar to variables). The list is also populated any call function block in a special procedure field editor (dropdown).
- Call function text:  'call function @name@'

functions

Blockly toolbox:
screen shot 2017-06-18 at 12 46 37 pm

Monaco toolbox:
screen shot 2017-06-18 at 12 48 16 pm

Defining: (Click "Make a function" in the flyout)
screen shot 2017-06-18 at 12 48 58 pm

screen shot 2017-06-18 at 12 47 03 pm

screen shot 2017-06-18 at 12 47 07 pm

Calling:
screen shot 2017-06-18 at 12 47 20 pm


microsoftsam:
Need a better icon, but we currently don't support custom icons in the toolbox, so I had to use one from font-awesome.

pelikhan:
How do you rename the function?

microsoftsam:
The function definition name is a text field, you rename by changing the text in there.

pelikhan:
LGTM

pxtblocks/blocklyloader.ts
```diff
@@ -2269,6 +2273,268 @@ namespace pxt.blocks {
...
+            /**
+             * Notification that a procedure is renaming.
+             * If the name matches this block's procedure, rename it.
+             * @param {string} oldName Previous name of procedure.
+             * @param {string} newName Renamed procedure.
+             * @this Blockly.Block
+             */
+            renameProcedure: function(oldName: string, newName: string) {
+                if (Blockly.Names.equals(oldName, this.getProcedureCall())) {
+                    this.setFieldValue(newName, 'NAME');
```
pelikhan: what if newName is empty?

pxtblocks/blocklyloader.ts
```diff
@@ -2269,6 +2273,268 @@ namespace pxt.blocks {
...
+            }
+
+            workspace.registerButtonCallback('CREATE_FUNCTION', function(button) {
+                let promptAndCheckWithAlert = (defaultName: string) => {
+                    Blockly.prompt(newFunctionTitle, defaultName, function(newFunc) {
+                        // Merge runs of whitespace.  Strip leading and trailing whitespace.
+                        // Beyond this, all names are legal.
+                        if (newFunc) {
+                            newFunc = newFunc.replace(/[\s\xa0]+/g, ' ').replace(/^ | $/g, '');
+                            if (newFunc == newFunction) {
```
pelikhan: should you check against the dictionary of reserved keywords? (can you name a function "var" or "function" ?)

pxtblocks/blocklyloader.ts
```diff
@@ -2269,6 +2273,268 @@ namespace pxt.blocks {
...
+                                        promptAndCheckWithAlert(newFunc);  // Recurse
+                                    });
+                            }
+                            else {
+                                createFunction(newFunc);
+                            }
+                        }
+                    });
+                };
+                promptAndCheckWithAlert('');
```
pelikhan: default to "foo" ? :)

abchatra:
In my opinion, functions should go into the advanced category. Here is my argument. Any new block category goes into advanced. Only after we see the usage and believe it needs a bump, move it up to the top level. We prioritized Arrays (advanced category) to Functions, which itself is an indicator that this shouldn't be at the top level. 

I do understand, the move to advanced might be a hindrance to adoption.  But anything new needs to prove its worth. 

@Jaqster @tballmsft ?

pxtblocks/blocklyloader.ts
riknoll: Invalid names will get automatically fixed just like variables when compiling to TS

pxtblocks/blocklyloader.ts
microsoftsam: newName is already validated to be not null and legal by blockly before entering this function. 
see: https://github.com/Microsoft/pxt-blockly/blob/master/core/procedures.js#L144

pxtblocks/blocklyloader.ts
microsoftsam: Okay. I chose to default to doSomething in monaco? should I update both to foo? thoughts on which is a better default?

riknoll: +1 for `doSomething()`

microsoftsam: went with doSomething

Issue #2308, opened 2017-06-16
Strange scoping behaviour in functions 


rharbird:
Hello,

There is a set of functions in the custom.ts in this snippet illustrating what I think might be broken variable scoping. Is this a bug or can someone explain what's going on here? Variable scoping behaviour seems to have changed in the last day or so.
 
https://makecode.microbit.org/_YdiVFx1vu8TV
 
@whaleygeek

Rgds,

Rae

tballmsft:
@mmoskal, can you look at this. Related to https://github.com/Microsoft/pxt/issues/2251 ?

tballmsft:
```

/**
 * Use this file to define custom functions and blocks.
 * Read more at https://makecode.microbit.org/blocks/custom
 */

const val = 0x40;

/**
 * Custom blocks
 */
//% weight=100 color=#0fbc11 icon=""
namespace custom {
    const val2 = 0x40;
    
    //% block
    export function foo(): number {
        return 0x40;
    }

    //% block
    export function bar(): number {
        return val;
    }

    //% block
    export function bar2(): number {
        return val2;
    }

    //% block
    export function bar3(): number {
        const val3 = 0x40;
        return val3;
    }



}
```

tballmsft:
Inlining the custom code into main makes the thing work:

https://makecode.microbit.org/_LW4PWRCYPXCu

tballmsft:
Something odd is going on with respect to code gen in custom.cs
```
/**
 * Custom blocks
 */
//% weight=100 color=#0fbc11 icon=""
namespace custom {  
    let life42 = 42
    
    //% block
    export function life(): number {
        return life42;
    }
```
yields
```
var __main__1 = entryPoint = function (s) {
var r0 = s.r0, step = s.pc;
s.pc = -1;
while (true) {
if (yieldSteps-- < 0 && maybeYield(s, step, r0)) return null;
switch (step) {
  case 0:

    s.lastBrkId = 1;
    r0 = pxsim.Array_.mk(0);
    s.tmp_0 = r0;
    r0 = globals.freqTable___110;
    pxtrt.decr(r0);
    r0 = s.tmp_0;
    globals.freqTable___110 = (r0);
    s.lastBrkId = 2;
    s.tmp_0 = { fn: _life_146, parent: s };
    s.pc = 2;
    return actionCall(s.tmp_0)
  case 2:
    r0 = s.retval;
    s.tmp_1 = r0;
    setupResume(s, 3);
[...]

PR #2318, opened 2017-06-19
fix compile order problem

tballmsft:
The introduction of the custom,.ts file for blocks created a problem, as the custom.ts file was ordered after main.ts, resulting in issues like:

https://github.com/Microsoft/pxt/issues/2308

This PR fixes the problem by ensuring that main.ts is the last file in the compilation order.

### DIFF
--- a/pxtcompiler/emitter/driver.ts
+++ b/pxtcompiler/emitter/driver.ts
@@ -137,7 +137,7 @@ namespace ts.pxtc {
 
         let tsFiles = opts.sourceFiles.filter(f => U.endsWith(f, ".ts"))
         // ensure that main.ts is last of TS files
-        let tsFilesNoMain = tsFiles.filter(f => !f.match("^main.ts$"))
+        let tsFilesNoMain = tsFiles.filter(f => f != "main.ts")
         if (tsFiles.length > tsFilesNoMain.length) {
             tsFiles = tsFilesNoMain
             tsFiles.push("main.ts")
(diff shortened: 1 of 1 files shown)


pxtcompiler/emitter/driver.ts
```diff
@@ -136,6 +136,13 @@ namespace ts.pxtc {
             opts.sourceFiles = Object.keys(opts.fileSystem)
 
         let tsFiles = opts.sourceFiles.filter(f => U.endsWith(f, ".ts"))
+        // ensure that main.ts is last of TS files
+        let tsFilesNoMain = tsFiles.filter(f => !f.match("^main.ts$"))
```
pelikhan: . Will match any character. Just equality here?

tballmsft: right

tballmsft: The insertion is done into a map, so there is no "first".

pxtcompiler/emitter/driver.ts
```diff
@@ -136,6 +136,13 @@ namespace ts.pxtc {
             opts.sourceFiles = Object.keys(opts.fileSystem)
 
         let tsFiles = opts.sourceFiles.filter(f => U.endsWith(f, ".ts"))
+        // ensure that main.ts is last of TS files
+        let tsFilesNoMain = tsFiles.filter(f => f != "main.ts")
```
pelikhan: shorter: https://jsfiddle.net/cwygddr6/

PR #2334, opened 2017-06-22
Adding new function icon

microsoftsam:
see svg.


svgicons/function.svg
```diff
@@ -0,0 +1,75 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
pelikhan: Can you minify it?

microsoftsam:
This should be cherrypicked to v0.

PR #2335, opened 2017-06-22
Fix Blockly variable prompt dialog bug

microsoftsam:
Issue with the Blockly prompt dialog when the defaultValue is set. 
We always return the defaultValue whereas we should always return "" if cancel is pressed. 
If okay pressed, we should return the value in the field, and if defaultValue is set, we should just set the initial value to be the default value (instead of what we do now which is set it as the placeholder). 

this is all in sync with what a window.prompt would do.


webapp/src/core.ts
```diff
@@ -341,13 +341,14 @@ export function promptAsync(options: PromptOptions): Promise<string> {
...
     options.htmlBody = `<div class="ui fluid icon input">
-                            <input type="text" id="promptDialogInput" placeholder="${options.defaultValue}">
+                            <input type="text" id="promptDialogInput" value="${options.defaultValue}">
                         </div>`;
 
     options.onLoaded = () => {
         let dialogInput = document.getElementById('promptDialogInput') as HTMLInputElement;
         if (dialogInput) {
             dialogInput.focus();
+            dialogInput.selectionStart = dialogInput.selectionEnd = 10000;
```
pelikhan: ?

microsoftsam: Hack to set the selection to the end, apparently works on all browsers.

pelikhan:
Isn't it better to select the entire text? Makes it easier to rename.

microsoftsam:
Yeah probably, let me take a look.

PR #2336, opened 2017-06-22
Add support for Flyout headings and grouping labels

microsoftsam:
Authoring changes: 
In block, we can author a group. 
In namespace, we can author groups list, which is just the way the editor should sort them. Otherwise the editor will sort them in alphabetical order. 

I made sure we're localizing the groups. 

Icons are currently not supported in our version of blockly in master, but that will come in the freshcoat changes.


localtypings/pxtarget.d.ts
```diff
@@ -164,6 +164,7 @@ declare namespace pxt {
         invertedMonaco?: boolean; // if true: use the vs-dark monaco theme
         blocklyOptions?: Blockly.Options; // Blockly options, see Configuration: https://developers.google.com/blockly/guides/get-started/web
         disableBlockIcons?: boolean; // Disable icons in blocks
+        flyoutHeadings?: boolean; // Show headings at the top of the flyout
```
pelikhan: this should be always on -- with option to turn it off ( ``flyoutHeadings`` -> ``hideFlyoutHeadings``)

pxtblocks/blocklyloader.ts
```diff
@@ -957,8 +997,69 @@ namespace pxt.blocks {
...
+                    })
+
+                    // Add the blocks to the xmlList
+                    let xmlList: Element[] = [];
+                    for (let bg = 0; bg < sortedGroups.length; ++bg) {
+                        let group = sortedGroups[bg];
+                        // Add the group label
+                        if (group != 'other') {
+                            let groupLabel = goog.dom.createDom('label');
+                            groupLabel.setAttribute('text', pxt.Util.rlf(`{id:group}${group}`));
```
riknoll: Do we have a process for manually adding strings to be localized in the target? Otherwise, this call won't do anything.

pelikhan: The CLI should parse the metadata and insert the ``{group}...`` strings in the genreated strings.json files.

microsoftsam: Yeah the CLI is parsing this and adding it to strings.json. see: https://github.com/Microsoft/pxt/pull/2336/files/0b1f98e636fc0789088dd5e93a0328f4c36ccd26#diff-8e4cc26c3bfd145eed3f0b841f212ec3R252

PR #2337, opened 2017-06-22
Moving to JavaScript random

pelikhan:
- [x] Math.random returns double between 0 and 1 (exclusive)
- [x] Math.randomRange(min, max) follows scratch semantics
- [x] remove special blocks decompiler support for random
- [x] docs

### DIFF
--- a/common-docs/reference/math/random-range.md
+++ b/common-docs/reference/math/random-range.md
@@ -1,0 +1,15 @@
+# Random Range
+
+Returns a pseudo-random number in the range ``[min, max];`` that is, from ``0`` (inclusive) up to including ``1`` (inclusive), which you can then scale to your desired range. 
+
+```sig
+Math.randomRange(0, 10)
+```
+
+## Returns
+
+* a pseudo-random [number](types/number) between ``max`` (inclusive) and ``max`` (inclusive).
+
+## See Also
+
+[``||random||``](/reference/math/random)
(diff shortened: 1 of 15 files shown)


pxtsim/libgeneric.ts
```diff
@@ -196,13 +196,20 @@ namespace pxsim {
...
+            if (min == max) return min;
+            if (min > max) {
+                let t = min;
+                min = max;
+                max = t;
+            }
+            if (Math.floor(min) == min && Math.floor(max) == max)
+                return min + Math.round(Math.random() * (max - min));
+            else
+                return min + Math.random() * (max - min);
```
pelikhan: max cannot be hit in this case

mmoskal: but you can get arbitrarly close to it

common-docs/reference/math/random-range.md
```diff
@@ -0,0 +1,15 @@
+# Random Range
+
+Returns a pseudo-random number in the range ``[min, max];`` that is, from ``0`` (inclusive) up to including ``1`` (inclusive), which you can then scale to your desired range.
```
mmoskal: We need to decide if this is a floating point number or integer. I think switching based on the bounds being integer isn't the best design. Maybe rename it randomInt ?

pelikhan: I was trying to map scratch: https://wiki.scratch.mit.edu/wiki/Pick_Random_()_to_()_(block)

pelikhan: I agree that integers would probably be better in our case.

PR #2343, opened 2017-06-23
Fix monaco editor error markers, line number's off by 1.

microsoftsam:
Not sure if this is the right place for this fix.
I'm able to add 1 to the line number to avoid the off by 1 issue, but the monaco editor is just surfacing what is returned inside file.diagnostics. 

@mmoskal should this fix be done in the compiler?


mmoskal:
As far as I understand, TypeScript `ts.getLineAndCharacterOfPosition()` function returns zero-based line numbers. We could change how we use it, but I would rather leave it, since it's also used in some other places. So I think this PR is good.

riknoll:
Yeah, generally those conversions are done in the editor.

PR #2377, opened 2017-06-29
[Accessibility] Tabbing

veler:
The current PR include fixed on the Tab index for the main page of makecode. It includes the navigation by the Enter key into the menus. It does not include the tab index into the modals.

- [x] Context menu on the main page focus tabindex managed
- [x] JQuery Modal focus tabindex loop
- [x] Modal focus tabindex loop
- [x] Modal close on escape
- [x] SideDoc close on escape
- [x] Explorer menu not tabbable (in JS mode)
- [x] Tabbing through the Project items (to open a previous project)
- [x] Projects dialog (projects.tsx)
- [x] Add package dialog (scriptsearch.tsx)
- [x] Share dialog (share.tsx)
- [x] SideDoc focus tabindex
- [x] Project settings
- [x] Tabs
- [x] Do not restart the focus from the beginning when a modal refresh or change it's content
- [x] IE will not fire keypress events for non-alphanumeric keys. Use onkeydown instead.
- [x] Side docs toggle is not part of the tabindex
- [x] After escaping from a modal, return focus to last element.


microsoftsam:
It seems the "Support" link in micro:bit has a tabIndex, and so as you tab through, it selects it, but then hides the Help menu. so we're losing a single tab that way. Going left to right, it should go to Projects -> Share -> Blocks -> JS -> Help (You can use up and down arrow keys here), then if you continue tabbing it should jump to the More dropdown menu.

microsoftsam:
Since <a links have implicit tabindex, you will need to set their tabindex to -1 in the dropdown menu, since we use up and down arrow keys in there to get to them.

webapp/src/sui.tsx
```diff
@@ -52,6 +54,12 @@ function removeClass(el: HTMLElement, cls: string) {
     else if (el.className.indexOf(cls) >= 0) el.className.replace(new RegExp(`(?:^|\\s)${cls}(?:\\s|$)`), ' ');
 }
 
+export function fireClickOnEnter(e: React.KeyboardEvent): void {
+    if (e.charCode == 13) {
```
microsoftsam: Can we handle Space as well? 
Space = 32

Not all browsers store the character value in charCode. use .which instead. 

let charCode = (typeof e.which == "number") ? e.which : e.keyCode

microsoftsam:
Tab index for share options: 
screen shot 2017-06-29 at 10 03 55 pm

webapp/src/sui.tsx
```diff
@@ -82,13 +90,29 @@ export class DropdownMenuItem extends UiElement<DropdownProps> {
     componentDidMount() {
         this.popup()
         this.child("").dropdown({
-            action: "hide",
+            action: (text: string, value: any, element: any) => {
```
microsoftsam: Why not have the element by of type JQuery? no need to cast down there.

veler: The reason why is that depending of if we click on a href or a div, the behavior of .click() is different, and we call "action" two times in some cases. One time with a Jquery element, and the other time with the contained HTMLElement of this Jquery value.

microsoftsam:
Okay makes sense, please add comments describing this behaviour.

microsoftsam:
I think the focus first interactive element shouldn't be just the first button (in our case it's the close icon). I think it should go somewhere logic. As a matter of fact, I would say the close should be the last one to get to, but the tabindex trap loop works well :)

microsoftsam:
In any modal dialog, as it get's re-rendered (ie: in projects dialog, you switch a tab. Or in share dialog, you hit publish). The focus interactive item method returns it to the start. It makes sense to me that if I wish to browse to a different tab in the projects dialog, that focus doesn't just switch to the start every time.

microsoftsam:
If I open anything in the side docs. I'm stuck in a tab loop inside of there? Why isn't the tabbing shared with the rest of the page. This is not a modal.

webapp/src/sui.tsx
```diff
@@ -846,6 +884,15 @@ export class Portal extends data.Component<PortalProps, PortalState> {
         }
     }
 
+    handleDocumentKeyPress = (e: KeyboardEvent) => {
```
microsoftsam: rename to handleEscape

webapp/src/sui.tsx
```diff
@@ -905,6 +954,7 @@ export class Portal extends data.Component<PortalProps, PortalState> {
         )
 
         this.portalNode = this.rootNode.firstElementChild;
+        core.giveFocusFirstInteractiveElement($(this.portalNode));
     }
```
microsoftsam: Can we avoid the Jquery call here?

webapp/src/container.tsx
```diff
@@ -71,6 +73,15 @@ export class SideDocs extends data.Component<ISettingsProps, {}> {
         this.firstLoad = false;
     }
 
+    private handleDocumentKeyPress = (e: KeyboardEvent) => {
+            if (e.keyCode !== 27) {
+                return;
```
microsoftsam: Use .which same as above.  
    let charCode = (typeof e.which == "number") ? e.which : e.keyCode

veler: Oops, copy past from another part of makecode I used. O:)

webapp/src/sui.tsx
veler: Done.

webapp/src/sui.tsx
veler: Done.

veler:
For the side docs, in my opinion, it's true that it's not a modal, but the blind user doesn't know that because he doesn't see how the UI looks like. I think it could be less confusing for him to be handled into a set of "sections"/"modals", so he will not have to take care of a "open the documentation in a new tab" when he's in the middle of something else for example.

webapp/src/sui.tsx
microsoftsam: Pushed?

webapp/src/monaco.tsx
```diff
@@ -536,7 +536,7 @@ export class Editor extends srceditor.Editor {
         root.className = 'blocklyTreeRoot';
         toolbox.appendChild(root);
         let group = document.createElement('div');
-        group.setAttribute('role', 'group');
+        group.setAttribute('role', 'tree');
         root.appendChild(group);
```
microsoftsam: Unrelated change. Let's do roles in a separate PR.

microsoftsam:
The projects page has tabs, and tabs follow a different paradigm. Once focus is within the tab menu, only the current tab has tabIndex=0 the other's have -1, and you use left / right arrow keys to switch between the tabs. 
see https://www.w3.org/TR/2010/WD-wai-aria-practices-20100916/#tabpanel
and http://simplyaccessible.com/article/danger-aria-tabs/

webapp/src/sui.tsx
veler: Not yet sorry, it was a note to myself.

microsoftsam:
Can we remove the outline from the tabPanel menu, keep the focus. 
screen shot 2017-07-06 at 1 22 13 am

You can achieve this by adding something like this to the css:
```
.ui.modal.projectsdialog .ui.secondary.inverted.menu:focus {
    outline: none;
}
```

microsoftsam:
Side docs toggle is not part of the tabindex: 
screen shot 2017-07-06 at 1 38 42 am

webapp/src/core.ts
```diff
@@ -449,6 +450,72 @@ export function scrollIntoView(item: JQuery, margin = 0) {
...
+    let charCode = (typeof e.which == "number") ? e.which : e.keyCode
+    if (charCode === 9 && e.shiftKey) {
+        e.preventDefault();
+        (e.data as HTMLElement).focus()
+    }
+};
+
+export function initializeFocusTabIndex(element: Element, trapLoop: boolean) {
+    if (jQuery.contains(element, document.activeElement)) {
+        return
```
microsoftsam: Can we avoid using Jquery here?

microsoftsam:
Issue: Privacy & Cookies link and Terms of use open two tabs on Chrome (Mac) when clicked / enter hit

microsoftsam:
When a side docs is open, if any page refresh is made (eg: simulator reload or fullscreen mode), focus is switched into the first sidedoc item. 
This is problematic. 
I understand the desire to focus the sidedocs when we first load it, but it might be best to not do that, and instead use aria-live fields to indicate to a user that a part of the UI has changed.

microsoftsam:
Exception in initializeFocusTabIndex, not sure how I got there, but you should probably null check the element: 
```
Uncaught TypeError: Cannot read property 'contains' of null
    at Object.initializeFocusTabIndex (main.js:4033)
    at SideDocs.unmountSideDocs (main.js:3615)
    at SideDocs.componentDidUpdate (main.js:3671)
    at CallbackQueue.notifyAll (main.js:43151)
    at ReactReconcileTransaction.close (main.js:55270)
    at ReactReconcileTransaction.closeAll (main.js:58110)
    at ReactReconcileTransaction.perform (main.js:58057)
    at ReactUpdatesFlushTransaction.perform (main.js:58044)
    at ReactUpdatesFlushTransaction.perform (main.js:56103)
    at Object.flushBatchedUpdates (main.js:56164)
    at Object.wrapper [as flushBatchedUpdates] (main.js:54750)
    at ReactDefaultBatchingStrategyTransaction.closeAll (main.js:58110)
    at ReactDefaultBatchingStrategyTransaction.perform (main.js:58057)
    at Object.batchedUpdates (main.js:50830)
    at Object.enqueueUpdate (main.js:56193)
    at enqueueUpdate (main.js:55781)
```

microsoftsam:
^^ Create new project, repro's the previous null exception.

microsoftsam:
I'm still seeing odd behavior with the focus jumping around or getting reset. eg: when I tab to the full screen button. If I hit on it, it resets. Previously, this wasn't a problem and the focus would stay on that button, and you'd be able to undo that action (right away).
Same behaviour with the simulator collapse / uncollapse button. It seems that after any re-render of the UI focus is switching to the <body > element.

webapp/src/container.tsx
```diff
@@ -71,6 +73,16 @@ export class SideDocs extends data.Component<ISettingsProps, {}> {
         this.firstLoad = false;
     }
 
+    private handleEscape = (e: KeyboardEvent) => {
+        let charCode = (typeof e.which == "number") ? e.which : e.keyCode
```
microsoftsam: I don't think we should be doing this, there's an assumption of where the focus is for this to work. When you first focus is switched to inside the side docs, this doesn't work. The side docs is not like a modal, doesn't need an escape to get out of it. and in the case of the modal, usually when you escape you get put back to where you were last (which isn't what you're doing in this case).

webapp/src/container.tsx
```diff
@@ -84,10 +86,34 @@ export class SideDocs extends data.Component<ISettingsProps, {}> {
...
+            this.unmountSideDocs();
+        }
+    }
+
+    componentWillUnmount() {
+        this.unmountSideDocs();
+    }
+
+    unmountSideDocs = () => {
+        (document.activeElement as HTMLElement).blur();
```
microsoftsam: The issue I've been seeing is due to this line of code. 
Why are we doing this? 

You should be able to repro this on Windows Browsers, have a side docs open, and then try to tab to any of the simulator buttons and click on them to cause a re-render. You'll lost focus as componentDidUpdate will call unmountSideDocs and cause the blur.
You should really not be doing a blur, what's the reason for it?

webapp/src/core.ts
```diff
@@ -449,6 +456,67 @@ export function scrollIntoView(item: JQuery, margin = 0) {
...
+    let jFirstTag: JQuery
+    let jLastTag: JQuery
+
+    if (firstFocused.length > 1) {
+        console.error("There must be maximum one element with the class firstFocused in the scope.")
+        return
+    } else if (firstFocused.length === 1) {
+        firstTag = firstFocused.item(0) as HTMLElement
+        jFirstTag = $(firstTag)
+        jFirstTag.off("keydown", giveFocusToLastTag)
```
microsoftsam: Refrain from using Jquery. 
el.removeEventListener(eventName, eventHandler);

Use this site: http://youmightnotneedjquery.com/

webapp/src/sui.tsx
```diff
@@ -553,6 +637,7 @@ export interface ModalProps {
     closeIcon?: any;
     closeOnDimmerClick?: boolean;
     closeOnDocumentClick?: boolean;
```
microsoftsam: Add option for closeOnEscape

webapp/src/core.ts
```diff
@@ -449,6 +456,67 @@ export function scrollIntoView(item: JQuery, margin = 0) {
...
+        jFirstTag.off("keyup", giveFocusToLastTag)
+    } else {
+        return
+    }
+
+    if (lastFocused.length > 1) {
+        console.error("There must be maximum one element with the class lastFocused in the scope.")
```
microsoftsam: No need to throw an error, just take the first element that you see.

webapp/src/srceditor.tsx
```diff
@@ -20,6 +21,13 @@ export class Editor implements pxt.editor.IEditor {
     changeCallback = () => { };
     setVisible(v: boolean) {
         this.isVisible = v;
+
+        if (this.isVisible && this.getId() === "pxtJsonEditor") {
+            // There is probably a better way, like waiting that React update the style itself.
```
microsoftsam: No need to trap focus in the settings editor. It's just like any other editor.

webapp/src/core.ts
```diff
@@ -449,6 +456,67 @@ export function scrollIntoView(item: JQuery, margin = 0) {
...
+        return
+    } else if (lastFocused.length === 1) {
+        lastTag = lastFocused.item(0) as HTMLElement
+        jLastTag = $(lastTag)
+        jLastTag.off("keydown", giveFocusToFirstTag)
+        jLastTag.off("keyup", giveFocusToFirstTag)
```
microsoftsam: Why are we doing a keydown and a keyup.

PR #2382, opened 2017-06-30
Separate the two length blocks

### LINKED ISSUE #2374: Array type inference issue
Type decl emitted for array variable incorrect. 

Repro program: https://makecode.microbit.org/_cjt1b0hdC22h

riknoll:
Fixes #2374 

We have two length blocks (one for strings and one for arrays) that up until now have been interchangeable. Unfortunately, our compiler doesn't support that and it caused type inference issues where arrays would occasionally be declared as strings. This PR makes them explicitly separate and alters the checks in the block definition so that they only accept either strings or arrays but not both.

Before:

length_before

After:

length_after

I'll need to cherry-pick this into V0 once merged. This might be a minor breaking change...


riknoll:
Yep, both of them show up in the results but the array length doesn't show up first when I search "of array".

riknoll:
@guillaumejenkins is this good to merge?

PR #2387, opened 2017-07-03
Surface toggle trace event to the controller

microsoftsam:
Surface toggle trace event to the controller. 
- optional intervalSpeed parameter to alter execution trace speed.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1013,13 +1013,13 @@ export class ProjectView
         this.startSimulator();
     }
 
-    toggleTrace() {
+    toggleTrace(intervalSpeed?: number) {
         if (this.state.tracing) {
             this.editor.clearHighlightedStatements();
             simulator.setTraceInterval(0);
         }
         else {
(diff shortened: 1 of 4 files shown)


webapp/src/app.tsx
```diff
@@ -1013,13 +1013,13 @@ export class ProjectView
...
 
-    toggleTrace() {
+    toggleTrace(intervalSpeed?: number) {
         if (this.state.tracing) {
             this.editor.clearHighlightedStatements();
             simulator.setTraceInterval(0);
         }
         else {
-            simulator.setTraceInterval(simulator.SLOW_TRACE_INTERVAL);
+            simulator.setTraceInterval(intervalSpeed || simulator.SLOW_TRACE_INTERVAL);
```
riknoll: Is 0 valid?

microsoftsam: It seems to be the case, going off line 1019.

riknoll: Maybe change the check to `intervalSpeed == undefined`

microsoftsam: Done.

PR #2401, opened 2017-07-06
Use function keyword for blocks instead of arrow functions

### LINKED ISSUE #1676: replace ()=>{} with function(){} in event handlers
when translating from JS to blocks. Also handle decompilation.

riknoll:
Fixes #1676 

This is a big change to our generated TypeScript, so let's make sure we're all on the same page before merging. Instead of compiling event handlers to arrow functions, we will instead compile to function expressions because the keyword makes it easier to understand.

In other words, this:

simple_button

Will now compile to this:

```typescript
input.buttonA.onEvent(ButtonEvent.Click, function () {
})
```

Instead of this:

```typescript
input.buttonA.onEvent(ButtonEvent.Click, () => {
})
```

### DIFF
--- a/pxtblocks/blocklymutators.ts
+++ b/pxtblocks/blocklymutators.ts
@@ -282,7 +282,7 @@ namespace pxt.blocks {
                 return escapedParam;
             }).join(", ");
 
-            const lambdaString = ` ({ ${declarationString} }) => `;
+            const lambdaString = `function ({ ${declarationString} })`;
 
             if (this.info.attributes.mutatePropertyEnum) {
                 return mkText(` [${this.parameters.map(p => `${this.info.attributes.mutatePropertyEnum}.${p}`).join(", ")}],${lambdaString}`)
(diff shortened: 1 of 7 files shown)


riknoll:
BTW, this shouldn't break anyone

pxtblocks/blocklymutators.ts
```diff
@@ -282,7 +282,7 @@ namespace pxt.blocks {
                 return escapedParam;
             }).join(", ");
 
-            const lambdaString = ` ({ ${declarationString} }) => `;
+            const lambdaString = `function ({ ${declarationString} })`;
```
abchatra: Rename the variable?

abchatra:
👍

pxtblocks/blocklymutators.ts
riknoll: Haha, fair enough

microsoftsam:
In addition to this change, you'll need to update the monaco language service that does the autocompletion: 
https://github.com/Microsoft/pxt-monaco-typescript/blob/master/src/worker.ts#L300

mmoskal:
Well, the only downside I see is that people will learn to use `function()` and get bitten when they start writing classes.

tballmsft:
I think this is an improvement for beginners.  Teachers should like it.

PR #2412, opened 2017-07-07
[Accessibility] Hidden tab menu

veler:
Adding a hidden menu displayed only when the user starts to use tab in the page.

- [ ] Align the icon in the middle with the text


microsoftsam:
Menu should be the first focus, it's the second. The micro:bit logo is the first tab, this should be before it.

microsoftsam:
Can we not include "Switch to Javascript" if we are already in the JS view (on file: main.ts).

webapp/public/locales/fr/strings.json
```diff
@@ -21,6 +21,7 @@
   "Cancel": "Annuler",
   "Cannot add {0} package": "Impossible d’ajouter le progiciel {0}",
   "Cannot load package: {0}": "Impossible de charger le paquet : {0}",
+  "Change Language": "Changer de langue",
   "Changes the value of the variable by this amount": "Modifie la valeur de la variable par ce montant",
```
microsoftsam: Did you edit these manually? That's not how we translate things in pxt. If you're interested in translating these labels in makecode, you'll need to translate them on crowdin. Ask me for details

webapp/src/app.tsx
```diff
@@ -1531,6 +1532,24 @@ ${compileService ? `<p>${lf("{0} version:", "C++ runtime")} <a href="${Util.html
         pxt.tickEvent(`tutorial.showhint`, { tutorial: options.tutorial, step: options.tutorialStep });
     }
 
+    showHiddenMenu = (e: React.KeyboardEvent) => {
+        let charCode = (typeof e.which == "number") ? e.which : e.keyCode
```
microsoftsam: This isn't the right way to do this. The best way to do this is with CSS. I see why the menu only shows up once you've already focused on the logo. 

What you want to do is do what Mozilla does on mdn, and add the menu in the DOM, if it's the first one it'll be the first to get focus. and then you hide the menu, and only show it on focus with CSS.

```
#accessibilityMenu {
    display:none;
}
#accessibilityMenu:focus {
   display: block;
}
```

webapp/src/app.tsx
```diff
@@ -1667,6 +1686,12 @@ ${compileService ? `<p>${lf("{0} version:", "C++ runtime")} <a href="${Util.html
                                 </a> : undefined }
                             </div>
                         </div>
+
+                        <sui.HiddenMenu ref={(self) => { this.hiddenMenu = self; }}>
```
microsoftsam: This doesn't need to be a sui component..

microsoftsam:
Also, this change is independent of accessibility-tab-jseditor. Would be best if we keep PR's separate and base it off the accessibility branch

webapp/public/locales/fr/strings.json
veler: Yes I did it manually. The text actually doesn't render when the website is in French. I asked to Richard but he didn't know why. I would appreciate some explanations please.

pelikhan:
Localized strings are pulled from crowdin via the cloud backend.

microsoftsam:
Closing this PR. rebase of the accessibility branch and create a new PR.

PR #2413, opened 2017-07-07
Fixing nested toc accordions in the docs

riknoll:
You can see the issue this fixes [here](https://makecode.microbit.org/docs). Try to expand the nested accordions under the courses heading and they collapse immediately. The fix is that only top-level accordions should have the `ui` class added to them


riknoll:
Do I need to cherry-pick this to V0? Or will the docs pick this up?

microsoftsam:
Yeah I think you'll need to cherry pick.

riknoll:
Okay

PR #2419, opened 2017-07-07
Fixing syntax error in IE 11

riknoll:
This was causing the docs to not render because IE11 doesn't support arrow functions. The regression happened here: https://github.com/Microsoft/pxt/commit/262d7e30a706e2a61ef15fa626d3e58d399a8c98


microsoftsam:
Woops, my bad. Good catch. I must've thought I was editing a TS file.

PR #2426, opened 2017-07-10
[Accessibility] Hidden tab menu

veler:
Adding a hidden menu displayed only when the user starts to use tab in the page.

One point : I inverted the "High Contrast On" with "Off" (compared to the contextual menu). I think this menu must display the current state of this option, not the state that is will take once clicked on it. It's like in the Windows settings in fact.

- [ ] Give the focus to the JS editor when using the "Switch to JS" menu.


microsoftsam:
Instead of "Switch to Javascript", we could change it to "Skip to JavaScript editor". and have it show up always, the point of this item is to skip having to tab through the menu and other components to get to the main content (in this case it's the Javascript editor).
so always show, and change text to "Skip to Javascript editor"

microsoftsam:
I don't get the note about the High contrast. When I hover over to that menu item, it says "High contrast off" (when it's not on in the first place, why would I be turning it off)??

microsoftsam:
Minor bug: This only repro's the first time, but when I select "Change language" the modal pops up, but it doesn't focus on the first interactable element. All subsequent tries work through..

theme/common.less
```diff
@@ -181,6 +181,44 @@ div.simframe > iframe {
...
+        top:20em;
+    }
+
+    .ui.item.link:hover {
+        top:20em;
+    }
+}
+
+@media only screen and (max-width: @largestTabletScreen) {
+    #menubar .ui.menu {
```
microsoftsam: What's this? Why is it here. 

If there is a reason to add this media query, it should be with the rest of the media queries in the bottom of the file. and please add a comment above it.

veler: I forgot to remove it, it was while I was testing the mobile size. I removed it, it's deprecated.

microsoftsam:
Match menu item text and capitalisation: 
screen shot 2017-07-12 at 5 25 27 pm

"Set High Contrast On", "Set High Contrast Off"

Change "Change Language" to "Select Language" to match Lang modal header: 
screen shot 2017-07-12 at 5 26 41 pm

microsoftsam:
If already on the Javascript view, focusing on "Switch to text" should jump the focus onto the Monaco editor. It doesn't do that at the moment

webapp/src/core.ts
```diff
@@ -456,14 +456,15 @@ export function scrollIntoView(item: JQuery, margin = 0) {
     }
 }
 
-export function initializeFocusTabIndex(element: Element) {
-    if (element !== document.activeElement && element.contains(document.activeElement)) {
+export function initializeFocusTabIndex(element: Element, allowResetFocus = false) {
+    if (!allowResetFocus && element !== document.activeElement && element.contains(document.activeElement)) {
```
microsoftsam: What's the allowResetFocus for?

veler: It is part of the "Fixed a bug where the focus was not given to the correct element in a modal in case of the UI was empty at a moment because of a "loading in progress" behavior." commit [https://github.com/Microsoft/pxt/pull/2426/commits/e3c88971f81dc62072cc245a47656eb495f33843](https://github.com/Microsoft/pxt/pull/2426/commits/e3c88971f81dc62072cc245a47656eb495f33843).

For the Select Language modal, when we open it for the first time, a "Loading..." splash screen appear (visible only with a breakpoint or with a very slow internet connection), and only then the languages list is displayed. It means that the modal content changed at a moment.

But in this method, we made a condition that avoid to change reset the focus to the first interactive element. It is useful for the Share modal, but it generate this problem for the Select Language.
The idea of the allowResetFocus is to allow to move the focus to another element if a modal's content change. This argument is true for Select Language, but false for Share modal for example.

PR #2435, opened 2017-07-11
Updates for pxt electron

### LINKED ISSUE #2434: Running pxt electron * throws errors on Windows
Attempting to run ```pxt electron run``` or ```pxt electron package``` throws errors:

1. when the cli tries to run the local gulp command: ```'C:\Users\Adam' is not recognized as an internal or external command,
operable program or batch file.```

   Workaround:
   a. install gulp globally
   b. change the cli to call the global gulp command

2. when npm 5.x is installed: ```[...]/pxt-microbit/node_modules/caniuse-db is not a child of [...]/pxt/electron/src```

   Workaround:
   a. use npm 4.6.1

adamthegameguy:
Fixes #2434

- ```pxt electron run``` and ```pxt electron package``` now use a global installation of gulp
- updated documentation to reflect global gulp and npm version prerequisites

> **NOTE:** ```pxt electron package``` doesn't work on non-Windows machines

### DIFF
--- a/cli/electron.ts
+++ b/cli/electron.ts
@@ -372,7 +372,7 @@ function electronGulpTask(taskName: string): Promise<void> {
     const gulpPath = path.join(pxtElectronPath, "node_modules", ".bin", "gulp");
 
     return nodeutil.spawnAsync({
-        cmd: nodeutil.addCmd(gulpPath),
+        cmd: "gulp.cmd",
         args: [
             taskName,
             "--" + targetProductJson,
--- a/electron/README.md
+++ b/electron/README.md
@@ -3,9 +3,25 @@
 Electron app wrapper for PXT targets.
 
 ## Prerequisites
+
+### Native Components
+
 This app has dependencies that have native components. These native components will need to be recompiled when you install them. To recompile them, you need the following: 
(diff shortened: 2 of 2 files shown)


riknoll:
@guillaumejenkins

cli/electron.ts
```diff
@@ -372,7 +372,7 @@ function electronGulpTask(taskName: string): Promise<void> {
     const gulpPath = path.join(pxtElectronPath, "node_modules", ".bin", "gulp");
 
     return nodeutil.spawnAsync({
-        cmd: nodeutil.addCmd(gulpPath),
+        cmd: "gulp.cmd",
```
guillaumejenkins: Before shifting to global gulp, can you try something like:
```
cmd: `"${nodeutil.addCmd(gulpPath)}"`
```
Thanks for looking into this!

adamthegameguy: Adding quotes did not solve the issue

adamthegameguy: Shifting to global is a bit of a workaround but it seamed like the simplest/fastest solution.

adamthegameguy:
Creating a user without spaces (...\adamc\... instead of ...\Adam Carriere\...) and running pxt electron * avoids the global gulp issue.

I'm going to close this PR and issue #2434 

npm issues still remain and I'll open a new issue for that.

PR #2442, opened 2017-07-12
Restoring the mutation to the procedure block

riknoll:
This does not make a cogwheel appear, it just saves the name of the function in a mutation as well as a field. Our XML still won't match the Blockly XML exactly, but this is compatible at least.

Blockly xml (and what the decompiler produces):
```xml
<block type="procedure_callnoreturn">
    <mutation name="foo" />
</block>
```

Our XML after this change (note the redundant field):
```xml
<block type="procedure_callnoreturn">
    <mutation name="foo" />
    <field name="NAME">foo</field>
</block>
```

We can't get rid of that extra field because a dropdown is editable and all editable fields are automatically saved.

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -2545,33 +2545,33 @@ namespace pxt.blocks {
                         JSON.stringify((def as any).arguments_) != JSON.stringify(this.arguments_))) {
                         // The signatures don't match.
                         def = null;
-                }
-                if (!def) {
-                    Blockly.Events.setGroup(event.group);
-                    /**
-                     * Create matching definition block.
-                     * <xml>
-                     *   <block type="procedures_defreturn" x="10" y="20">
-                     *     <field name="NAME">test</field>
-                     *   </block>
-                     * </xml>
-                     */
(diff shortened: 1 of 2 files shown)


pxtblocks/blocklyloader.ts
```diff
@@ -2545,33 +2545,33 @@ namespace pxt.blocks {
...
+                    }
+                    if (!def) {
+                        Blockly.Events.setGroup(event.group);
+                        /**
+                         * Create matching definition block.
+                         * <xml>
+                         *   <block type="procedures_defreturn" x="10" y="20">
+                         *     <field name="NAME">test</field>
+                         *   </block>
+                         * </xml>
```
microsoftsam: Can you update the comment here with what a sample XML would look like.

microsoftsam:
Needs to be cherry picked to V0

pxtblocks/blocklyloader.ts
riknoll: That's the definition block's xml (`procedures_defreturn` not `procedures_callreturn`). That xml hasn't changed.

microsoftsam:
Cherry-picked into v0 and freshcoat

PR #2444, opened 2017-07-13
Update Blockly (June release)

microsoftsam:
This release come's with a lot of refactoring in many Blockly structures including: 
- Variables. Now are stored in a VariableModel instead of a string, which includes a type
- Gestures. Now handled by a central gesture class, instead of each component handling the same gestures. 
- Flyouts. Refactored into Vertical and Horizontal Flyout classes. 

This release will make the freshcoat (scratch-blocks) changes much more aligned with the master version.

In pxt, we override the Variable flyout code, and so I've updated that to deal with the refactor.
I haven't noticed any other changes from our end, but let's do some testing on this branch before we merge.

Respective Blockly branch: https://github.com/Microsoft/pxt-blockly/tree/blocklyjune


riknoll:
I'm getting "Field editor lights not registered" when I try and open the light category in Adafruit

riknoll:
Ah, looks like the issue is that you removed the `playAudio` API from the workspace in blockly.d.ts

microsoftsam:
Yeah audio API has been refactored. Will have to reflect these changes in the ring field editor.

riknoll:
Tested IE, Firefox, and Edge. Seems to be working fine.

PR #2446, opened 2017-07-13
[Accessibility] Support of the screen readers

veler:
- Support of the screen readers. It's mainly supported thanks to the ARIA attributes.
- Fixed a bug where in Safari, IE and Edge, the (Terms & Conditions) and (Privacy) menu was opening two tabs.

Note : On Mac, VoiceOver is totally not supported by Firefox. The screen reader doesn't detect the navigation in a page.


webapp/src/sui.tsx
```diff
@@ -59,7 +60,7 @@ export function fireClickOnEnter(e: React.KeyboardEvent): void {
     let charCode = (typeof e.which == "number") ? e.which : e.keyCode
     if (charCode === 13 || charCode === 32) {
         e.preventDefault();
-        (document.activeElement as HTMLElement).click();
+        (e.currentTarget as HTMLElement).click();
     }
```
microsoftsam: What's this? seems unrelated.

webapp/src/sui.tsx
```diff
@@ -90,55 +91,87 @@ export class UiElement<T extends WithPopupProps> extends data.Component<T, {}> {
...
-            hrefTags.item(0).onkeypress = () => { return false }
-        }
     }
 
     renderCore() {
         return (
             <div className={genericClassName("ui dropdown item", this.props) }
-                role={this.props.role}
+                role="menuitem"
                 title={this.props.title}
```
microsoftsam: Always a menuitem?

microsoftsam:
Mostly looks good. I don't quite understand the changes you made to the dropdown menu component. If you could describe those changes a little, that'd be great.

webapp/src/sui.tsx
veler: Hum, we are in a DropdownMenuItem, except if we have a special case, I don't think we will need something else. At least for the moment I don't think we need.

webapp/src/sui.tsx
veler: This was a part of the changes I had to do to support the focus in the dropdown menu items. The fact is that previously, I used "document.activeElement" because I didn't founded the "target" or "currentTarget" of React.KeyboardEvent, probably because I didn't searched enough.
Now, in the behavior of an Item of a DropDownMenu, when we hit Enter, we give the focus back to the DropDownMenu, when we close the modal, the focus will be given to the menu again. But the problem is that this function here is executed on a keyDown on each item, it will look for "document.activeElement" and do a click on it... In our case, the activeElement is now the DropDownMenu, not the item. That's why I do this change.
I checked the behavior of the rest of the website, and I didn't found any regression with this change.

webapp/src/sui.tsx
```diff
@@ -90,55 +91,87 @@ export class UiElement<T extends WithPopupProps> extends data.Component<T, {}> {
 }
 
 export class DropdownMenuItem extends UiElement<DropdownProps> {
```
veler: The idea of all those changes in the DropdownMenuItem is to be give the focus to the selected element of the menu without closing the dropdown (because by default, it close if we give the focus to an item of the menu).

webapp/src/sui.tsx
```diff
@@ -90,55 +91,87 @@ export class UiElement<T extends WithPopupProps> extends data.Component<T, {}> {
...
+        this.preventHide = false
+        this.child("").dropdown("hide")
+    }
+
     componentDidMount() {
         this.popup()
         this.child("").dropdown({
-            action: (text: string, value: any, element: any) => {
-                let htmlElement: HTMLElement
+            action: (text: string, value: any, element: JQuery) => {
```
veler: As now, when we navigate with the keyboard, the items have the focus, we don't need to simulate a click on them as by default they have the "fireClickOnEnter". But there was still a small problem with the "a href" tags. With this new implementation, it works with Enter key and Click, and the Privacy menu is not opened 2 times on Edge and Safari anymore.

webapp/src/sui.tsx
```diff
@@ -90,55 +91,87 @@ export class UiElement<T extends WithPopupProps> extends data.Component<T, {}> {
...
-                if (htmlElement.tagName.toLowerCase() === 'a') {
-                    window.open((htmlElement as HTMLLinkElement).href, '_blank')
+                    if (element.get(0).tagName.toLowerCase() === 'a') {
+                        window.open((element.get(0) as HTMLLinkElement).href, '_blank')
+                    }
                 }
             },
             fullTextSearch: true,
-            onChange: (v: string) => {
+            onChange: (v: string, text: string, item: JQuery) => {
```
veler: When we select an item of the menu, we give the focus to it, and prevent the dropdown to be closed.

webapp/src/sui.tsx
```diff
@@ -90,55 +91,87 @@ export class UiElement<T extends WithPopupProps> extends data.Component<T, {}> {
...
+                this.preventHide = true
+                item.get(0).focus()
+
                 if (this.props.onChange && v != this.props.value) {
                     this.props.onChange(v)
                 }
+            },
+            onShow: () => {
+                this.isOpened = true
+                this.forceUpdate()
```
veler: The forceUpdate is here to force a renderCore, so the aria-hidden and expanded are updated.

theme/common.less
```diff
@@ -40,6 +40,19 @@ html, body {
...
+  position: absolute !important;
+  display: block;
+  visibility: visible;
+  overflow: hidden;
+  width: 1px;
+  height: 1px;
+  margin: -1px;
+  border: 0;
+  padding: 0;
+  clip: rect(0 0 0 0);
```
microsoftsam: nit: add tab (4 spaces before) like below.

webapp/src/sui.tsx
```diff
@@ -434,6 +482,7 @@ export class Segment extends data.Component<SegmentProps, SegmentState> {
 }
 
 export interface MenuProps {
+    activeMenuItem?: any;
     activeIndex?: number;
```
microsoftsam: Change to ariaDescribedBy as it's being used, or maybe it's obsolete?

webapp/src/sui.tsx
```diff
@@ -544,13 +595,12 @@ export class Menu extends data.Component<MenuProps, MenuState> {
             return
         }
 
-        let rootNode = ReactDOM.findDOMNode(this)
-        let items = rootNode.getElementsByClassName("link item")
+        var menuItems = this.child(".link")
```
microsoftsam: use let

PR #2448, opened 2017-07-13
Fire stopped event when the simulator is stopped by the editor

riknoll:


microsoftsam:
Wonderful, cherrypick to freshcoat @riknoll ?

riknoll:
Will do

PR #2453, opened 2017-07-14
Fix modulus op in monaco flyout

dragonx:
Looks like if there's two blocks with the same jsDoc, they get merged with some odd behavior.  There should be a modulus operator, but it just shows up as a block with just "0".  I've always assumed that this was a simple number block, but it's the broken modulus block.

image

Fixing the jsDoc makes this turn into the modulus block.


abchatra:
@riknoll I think Richard has already fixed this

riknoll:
Nope, this is a good change

PR #2457, opened 2017-07-14
Fixing array inference with the create_empty block

### LINKED ISSUE #2451: micro:bit blocks - can't create empty sprite array
It seems this isn't allowed when using blocks:

empty_array_1

Having to add a sprite at the start can make some messy code, e.g. if the sprites appear and disappear in the "forever" loop.

We can write in JavaScript:
`let list: game.LedSprite[] = []`
but converting to blocks and back breaks it.

riknoll:
Fixes #2451

### DIFF
--- a/pxtblocks/blocklycompiler.ts
+++ b/pxtblocks/blocklycompiler.ts
@@ -203,7 +203,7 @@ namespace pxt.blocks {
                     }
                 }
             }
-            return tp || ground("number[]");
+            return tp || ground("Array");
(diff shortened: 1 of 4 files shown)


riknoll:
Basically, if an array block is empty (has no fields with blocks attached), we shouldn't infer anything from it. Before we were trying to assign a random type (`number[]`) when really we should have just returned `Array` which will get turned into a concrete type later.

PR #2469, opened 2017-07-17
WinRT now supports both 1.22 and 1.23 UF2 bootloader versions simulta…

guillaumejenkins:
…neously

Fixes https://github.com/Microsoft/pxt-adafruit/issues/232

### DIFF
--- a/pxtwinrt/hiddeploy.ts
+++ b/pxtwinrt/hiddeploy.ts
@@ -109,15 +109,7 @@ namespace pxt.winrt {
 
         private initUf2Selectors() {
             const whid = Windows.Devices.HumanInterfaceDevice.HidDevice;
-            // Selectors common to all UF2 bootloaders
-            WindowsRuntimeIO.uf2Selectors = [
-                // UF2 Bootloader v1.23.0
-                whid.getDeviceSelector(0xFF97, 0x0001, 0x239A, 0x0018),
-                // UF2 Bootloader v1.22.0
-                whid.getDeviceSelector(0xFF00, 0x0001, 0x239A, 0x0019)
-            ];
-
-            // Target-specified selectors (for User Space)
+            WindowsRuntimeIO.uf2Selectors = [];
             if (pxt.appTarget && pxt.appTarget.compile && pxt.appTarget.compile.hidSelectors) {
                 pxt.appTarget.compile.hidSelectors.forEach((s) => {
                     const sel = whid.getDeviceSelector(parseInt(s.usagePage), parseInt(s.usageId), parseInt(s.vid), parseInt(s.pid));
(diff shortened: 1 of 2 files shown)


pxtwinrt/hiddeploy.ts
```diff
@@ -49,44 +53,78 @@ namespace pxt.winrt {
...
         }
+
+        private initUf2Selectors() {
+            const whid = Windows.Devices.HumanInterfaceDevice.HidDevice;
+            // Selectors common to all UF2 bootloaders
+            WindowsRuntimeIO.uf2Selectors = [
+                // UF2 Bootloader v1.23.0
+                whid.getDeviceSelector(0xFF97, 0x0001, 0x239A, 0x0018),
+                // UF2 Bootloader v1.22.0
+                whid.getDeviceSelector(0xFF00, 0x0001, 0x239A, 0x0019)
```
abchatra: Shouldn't vendorId etc be in the target?

pxtwinrt/hiddeploy.ts
```diff
@@ -49,44 +53,78 @@ namespace pxt.winrt {
...
-                .then(devices => {
-                    pxt.debug(`hid enumerate ${devices.length} devices`)
+            const wd = Windows.Devices;
+            const whid = wd.HumanInterfaceDevice.HidDevice;
+            if (!WindowsRuntimeIO.uf2Selectors) {
+                this.initUf2Selectors();
+            }
+            const getDevicesPromise = WindowsRuntimeIO.uf2Selectors.reduce((soFar, currentSelector) => {
+                // Try all selectors, in order, until some devices are found
+                return soFar.then((devices) => {
```
abchatra: Does the new code try to query all the HID devices matching the selectors? If yes can we postpone that change until we test that scenario. Old code seem to query just the first device.

pxtwinrt/hiddeploy.ts
guillaumejenkins: These 2 specific IDs are for the UF2 bootloader, which is target-independent (any target using UF2 will use these IDs)

pxtwinrt/hiddeploy.ts
guillaumejenkins: New code works just like before: Get a list of all devices matching criteria, and connect to the 1st one. The difference is that before, we were not using VID / PID in the selectors, so we had a higher chance of attempting to connect to a wrong device (which would fail and say "Device not found").

Now that we use VID / PID to enumerate devices, only expected devices should be returned, and we will connect to the first one

pxtwinrt/hiddeploy.ts
abchatra: 0x239A is the vendor ID for Adafruit AFAIK

pxtwinrt/hiddeploy.ts
abchatra: Makes sense. Thanks

pxtwinrt/hiddeploy.ts
guillaumejenkins: Ah, you're right

I'll put those back in the Adafruit target

Issue #2445, opened 2017-07-13
micro:bit sim - not firing all button events


HelenFoster:
I was looking for a "button down" event (to make games more responsive). It seemed like it should be possible with the Control blocks, but it's not on the list.

On further searching, MICROBIT_BUTTON_EVT_DOWN=1 gets fired. If I switch into JavaScript and write the number 1, it does actually work on the real micro:bit (though not on the simulator). This code moves the sprite to the left on A "pressed", and moves it to the right on B "button down".

```
let item: game.LedSprite = null
input.onButtonPressed(Button.A, () => {
    item.change(LedSpriteProperty.X, -1)
})
control.onEvent(EventBusSource.MICROBIT_ID_BUTTON_B, 1, () => {
    item.change(LedSpriteProperty.X, 1)
})
item = game.createSprite(2, 2)

```
~~There might be other really useful things that also haven't made it onto the list of values.~~

~~On the other hand, it's pretty long already, so if the menu could somehow tell which were relevant from the "from" field, that would be even better.~~

riknoll:
The simulator thing is definitely a bug. We probably skip a lot of runtime events that we should fire.

HelenFoster:
I put too many things in this ticket. If we say it's about simulator handing of button events:

MICROBIT_BUTTON_EVT_DOWN - not simulated
MICROBIT_BUTTON_EVT_UP - simulated
MICROBIT_BUTTON_EVT_CLICK - simulated, but takes LONG_CLICK as well
MICROBIT_BUTTON_EVT_LONG_CLICK - not simulated - becomes CLICK in simulator
MICROBIT_BUTTON_EVT_HOLD - not simulated
MICROBIT_BUTTON_EVT_DOUBLE_CLICK - doesn't seem to get fired on real micro:bit either

PR #2472, opened 2017-07-17
Add support for more button events

riknoll:
This is part of the fix for #2445 

Note that this is for V0, I'm not event sure it's worth porting this to master since we don't have any targets that use the ButtonPairState class. Also requires a PR in pxt-microbit.


pxtsim/state/buttonpair.ts
```diff
@@ -19,11 +75,63 @@ namespace pxsim {
...
+        private fireEventHandler(id: number, event: ButtonEvent) {
+            switch (event) {
+                case ButtonEvent.UP:
+                    this.fireEvent(id, this.props.BUTTON_EVT_UP);
+                    break;
+                case ButtonEvent.CLICK:
+                    this.fireEvent(id, this.props.BUTTON_EVT_CLICK);
+                    break;
+                case ButtonEvent.DOWN:
+                    if (this.props.BUTTON_EVT_DOWN != undefined) {
```
guillaumejenkins: Use `!==` when comparing with `undefined`?

riknoll: Using `!=` catches `null` as well

riknoll:
I'm adding the do not merge label until I get a PR open on the microbit side since this will cause some weird sim problems

pelikhan:
@riknoll what's the status on this?

riknoll:
Don't have time to redo this.

PR #2473, opened 2017-07-17
Give better control over inline inputs

### LINKED ISSUE #2471: Add way to override forced external block inputs
When there are 4+ (4 or more) parameters the block automatically becomes vertical and disables the use of inline blocks even when you specify `//% blockExternalInputs="false"` in the block definition. Ideally there should be a way to override this.

riknoll:
Fixes #2471 

I'm soft-deprecating the old comment attribute in favor of this new one. Gives three modes "inline", "external" and "auto".

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -761,7 +761,15 @@ namespace pxt.blocks {
             }
         }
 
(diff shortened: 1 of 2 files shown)


riknoll:
Fixed in 5ed650c

Issue #2102, opened 2017-05-16
Error icon disappear quickly which leads to user can't see the error message


Mofell:
Repro steps:
• Navigate to https://pxt.microbit.org/beta
• Choose blocks like the gif.

Expect result: Error icon should display when there is an error
Actual result: Error icon disappear quickly
![111](https://cloud.githubusercontent.com/assets/23447632/26095788/79e62c0e-3a51-11e7-9903-3e5b99835794.gif)

PR #2476, opened 2017-07-17
Fix warning labels for blocks

riknoll:
Fixes https://github.com/Microsoft/pxt/issues/2102

The issue was that `diag.endline` is undefined on compiler diagnostics.

### DIFF
--- a/webapp/src/blocks.tsx
+++ b/webapp/src/blocks.tsx
@@ -576,7 +576,7 @@ export class Editor extends srceditor.Editor {
         let sourceMap = this.compilationResult.sourceMap;
 
         diags.filter(diag => diag.category == ts.pxtc.DiagnosticCategory.Error).forEach(diag => {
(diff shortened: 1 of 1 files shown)


webapp/src/blocks.tsx
```diff
@@ -576,7 +576,7 @@ export class Editor extends srceditor.Editor {
         let sourceMap = this.compilationResult.sourceMap;
 
         diags.filter(diag => diag.category == ts.pxtc.DiagnosticCategory.Error).forEach(diag => {
-            let bid = pxt.blocks.findBlockId(sourceMap, { start: diag.line, length: diag.endLine - diag.line });
+            let bid = pxt.blocks.findBlockId(sourceMap, { start: diag.line, length: 0 });
```
guillaumejenkins: Does using a line length of 0 mess with functions that take callbacks as argument, which are pretty much always longer than 1 line?

```
loops.forever(() => {
	
})
```

riknoll: It shouldn't but I can double check

PR #2478, opened 2017-07-18
Add rendering of the 'for of' statement.

ganicke:
* add 'ForOfStatement' to renderer.ts
* add card for the 'for of' statement.


pelikhan:
This would need cherry picking into v0 as well

PR #2482, opened 2017-07-19
Put in 'See also' links for base block pages.

ganicke:
Fixes pxt#2248. 

Also, are we eventually moving *Math* into _/reference_?


common-docs/reference/math.md
```diff
@@ -0,0 +1,166 @@
...
+Numbers can have their fractional part too. The decimal point is between the digits of the number.
+But, _floating point_ numbers have the decimal point at any spot between digits, like: 3.14159 or 651.75.
+
+```block
+let num = 0
+num = 6.7
+num = 10.083
+```
+
+## Arithmetic binary operation (+, -, *, /)
```
pelikhan: * needs to be escaped

ganicke: It's been working all along but I'll add the escape.

pelikhan:
Merge issues.

PR #2488, opened 2017-07-20
Friendlier layout of blocks in portrait mode

riknoll:
Every time we decompile, we lose the block location information and so we format the user's block code. If the Blockly workspace's height is greater than its width, then we should format the blocks so that they are arranged vertically rather than horizontally. This fixes that.


dragonx:
@riknoll Sam, this works pretty well.  Can you use the same flow mechanism for the right click menu "Format code" command?

pelikhan:
We should also use this heuristic in docs.

Issue #2461, opened 2017-07-16
micro:bit array pop not working


HelenFoster:
This code shows 0 on the simulator and the real thing. It should show 1, right? (Changing it to "shift" shows 1.)

```
let list = [1]
basic.showNumber(list.pop())
```

riknoll:
This should be fixed in the next release. Thanks for catching it!

PR #2489, opened 2017-07-20
Fix array pop logic

riknoll:
Part of the fix for #2461


riknoll:
This needs to be cherry-picked to v0

abchatra:
:shipit:

PR #2496, opened 2017-07-21
Fix array pop logic V0

riknoll:


abchatra:
@riknoll  merge this?

Issue #2459, opened 2017-07-15
[Accessibility/Usability] Handle onEnter in core.prompt dialogs


microsoftsam:
Dialogs used for Blockly.Prompt (eg: Create a variable) don't handle the onEnter key pressed. This means the user has to click the OK button everytime after entering their input.

veler:
Can't reproduce this problem. Typing Enter on the OK button make the modal close. But I realized that the tabbing trap doesn't work for this modal. It's another issue

microsoftsam:
@veler we want to handle enter on the input field. eg: a user can type the var name and hit enter, rather than having to tab first to the ok button and hit enter. (enter should do the approve action of a modal prompt dialog)

veler:
Oh OK ! I though you were talking about the Enter key on the OK button. I will work on it later.

veler:
Bug fixed and included to [https://github.com/Microsoft/pxt/pull/2440](https://github.com/Microsoft/pxt/pull/2440)

veler:
Reopened to push the fix directly into Master.

PR #2503, opened 2017-07-24
Fixed a bug where pressing Enter was not validating a Modal.

veler:
Fixed a bug where pressing Enter in the input field of Make a Variable and Make a function modals was not doing the behavior of a click on the OK button.

Original issue : [https://github.com/Microsoft/pxt/issues/2459](https://github.com/Microsoft/pxt/issues/2459)

### DIFF
--- a/webapp/src/core.ts
+++ b/webapp/src/core.ts
@@ -347,8 +347,14 @@ export function promptAsync(options: PromptOptions): Promise<string> {
     options.onLoaded = () => {
         let dialogInput = document.getElementById('promptDialogInput') as HTMLInputElement;
         if (dialogInput) {
-            dialogInput.focus();
             dialogInput.setSelectionRange(0, 9999);
+            dialogInput.onkeyup = (e: KeyboardEvent) => {
+                let charCode = (typeof e.which == "number") ? e.which : e.keyCode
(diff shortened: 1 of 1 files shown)


webapp/src/core.ts
```diff
@@ -347,8 +347,14 @@ export function promptAsync(options: PromptOptions): Promise<string> {
     options.onLoaded = () => {
         let dialogInput = document.getElementById('promptDialogInput') as HTMLInputElement;
         if (dialogInput) {
-            dialogInput.focus();
```
microsoftsam: Why remove the focus? It's used to highlight the entire default argument, making it easy to select all and delete.

veler: Weirdly, without this line, the focus is still given to the input field (tested on Chrome, Edge, Firefox, Safari).

Furthermore, once the accessibility branch will be merged, the focus will still be given thanks to the initializeFocusTabIndex method. It works fine with or without this line, with or without the accessibility changes.

microsoftsam: Reason this works is because setSelectionRange does the same thing. This might be legacy. Let's remove for now, but keep an eye out for any regressions in selection.

Issue #2516, opened 2017-07-26
Hack Your Headphones Error


gbowman85:
On the final image at https://makecode.microbit.org/projects/hack-your-headphones/code all of the "ring tone" blocks are showing Middle C, where in fact the code view shows C,D,E,F,G,A.

This could be confusing for children who are recreating the code for themselves.

abchatra:
Looks like de-compilation is failing. @riknoll  any clue?
```
music.ringTone(music.noteFrequency(Note.D));
```

abchatra:
Thanks @gbowman85 for the issue.

riknoll:
I believe this is fixed in master and V0, but I'll double check

riknoll:
Nope, looks like this regressed at some point

caitlin-hennessy:
Fixed by https://github.com/Microsoft/pxt/pull/2592 and https://github.com/Microsoft/pxt/pull/2590.

PR #2590, opened 2017-07-31
Bug 2516 fix v0

caitlin-hennessy:
Allows blocks to be built from Javascript syntax like `music.ringTone(music.noteFrequency(Note.D));` 

See https://github.com/Microsoft/pxt/issues/2516

### DIFF
--- a/pxtblocks/fields/field_note.ts
+++ b/pxtblocks/fields/field_note.ts
@@ -237,7 +237,7 @@ namespace pxtblockly {
         setValue(note: string) {
             let regex: RegExp = new RegExp("Note.(.*)");
             let match = regex.exec(note);
-            let noteName: any = (match && match.length) > 1 ? match[1] : null;
+            let noteName: any = (match && match.length > 1) ? match[1] : null;
             note = Note[noteName] ? String(parseFloat(Note[noteName])) : String(parseFloat(note || "0"));
             if (isNaN(Number(note)) || Number(note) < 0)
                 return;
--- a/pxtblocks/fields/field_note.ts
+++ b/pxtblocks/fields/field_note.ts
@@ -1,7 +1,7 @@
 /// <reference path="../../localtypings/blockly.d.ts" />
 
 namespace pxtblockly {
-        enum Note {
+    enum Note {
         C = 262,
         CSharp = 277,
         D = 294,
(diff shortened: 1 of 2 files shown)


pxtblocks/fields/field_note.ts
```diff
@@ -185,7 +235,10 @@ namespace pxtblockly {
          * @param {string} note The new note in string format.
          */
         setValue(note: string) {
-            note = String(parseFloat(note || "0"));
+            let regex: RegExp = new RegExp("Note.(.*)");
```
riknoll: Move this regex declaration outside of the `setValue`

pxtblocks/fields/field_note.ts
```diff
@@ -185,7 +235,10 @@ namespace pxtblockly {
...
+            let match = regex.exec(note);
+            let noteName: any = (match && match.length > 1) ? match[1] : null;
```
riknoll: We typically avoid using the `any` type anywhere in the code. I think you can just declare `string` here.

pxtblocks/fields/field_note.ts
```diff
@@ -185,7 +235,10 @@ namespace pxtblockly {
...
+            note = Note[noteName] ? String(parseFloat(Note[noteName])) : String(parseFloat(note || "0"));
```
riknoll: The first `parseFloat` call here isn't needed. I think it should already be a number

pxtblocks/fields/field_note.ts
```diff
@@ -1,6 +1,56 @@
 /// <reference path="../../localtypings/blockly.d.ts" />
 
 namespace pxtblockly {
+        enum Note {
```
riknoll: The tabbing looks to be a little off here

caitlin-hennessy: Fixed

pxtblocks/fields/field_note.ts
caitlin-hennessy: Fixed

pxtblocks/fields/field_note.ts
riknoll: Oh, also, try `$Note\.(.+)^` instead

caitlin-hennessy: Moved regex but left "any" as per our discussion.

PR #2620, opened 2017-08-02
browserify bundle ignores non-js files

bioball:
This allows us to generate type declaration files. Otherwise, browserify will error because `.d.ts` files fail to compile.


bioball:
Thanks!

PR #2640, opened 2017-08-02
Normalizing and validating loc strings

pelikhan:
Don't load loc strings is invalid spaces or invalid parameter names.


ChrisDorna:
Thanks for fixing this issue!

PR #2657, opened 2017-08-04
Move built-in function topics from 'blocks' to 'reference'.

ganicke:
All non-statement blocks are set to have the help root as _/reference_ (see: blocklyloader). Moving the built-ins to live under _/reference_ to path in properly.

- [x] Move 'Arrays' topics from _/blocks/arrays_ to _/reference/arrays_.
- [x] Move 'Text' topics from _/blocks/text_ to _/reference/text_.

Changes in rendering to accommodate the built-ins coming in a follow-on PR.


riknoll:
This seems fine to me. @pelikhan are there any consequences in changing these URLs?

ganicke:
See related PR https://github.com/Microsoft/pxt/pull/2658 for aliasing namespace names.

pelikhan:
Should be fine.

PR #2658, opened 2017-08-04
Render changes for namespace aliasing.

ganicke:
Namespaces for some built-ins don't match block naming, such as **Array** to **arrays** and **String** to **text**. So, the help urls are formed like _/reference/String/char-at_ instead of _/reference/text/char-at_. These don't match editor block groups or the doc tree. Here's a proposed fix:

- [x] Use the **blockNamespace** attribute in the namespace jsDoc to alias/map the namespace name (renderer.ts).
- [x] Add a ```declare namespace Array {}``` to inject jsDoc for the **arrays** alias (pxt-core.d.ts ).
- [x] Add descriptions and ```//% blockNamespace="xxx"``` for aliases on **Array** and **String** namespaces (pxt-core.d.ts).
- [x] Include a modified _blocks.md_ for context.

Be careful of the new ```declare namespace Array{}``` since I broke **String** doing this once before :-(

This is related to https://github.com/Microsoft/pxt/pull/2657.

### DIFF
--- a/common-docs/blocks.md
+++ b/common-docs/blocks.md
@@ -7,13 +7,25 @@ Blocks snap into each other to define the program that your @boardname@ will run
 Blocks can be event (buttons, shake, ...) or need to be snapped into an event to run.
 The [on-start](/blocks/on-start) event runs first.
 
+## Blocks
+
 ```namespaces
 for (let i = 0;i<5;++i) {}
 if (true){}
 let x = 0;
+```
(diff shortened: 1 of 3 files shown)


ganicke:
This is how it looks with this change:

blocksmd

riknoll:
I was able to clean and build locally with no problem, so I think it's fine.

Issue #2553, opened 2017-07-27
nested handlers doesn't convert


tballmsft:
See https://makecode.microbit.org/_6hJetPWAy2rz

The code is
```
input.onButtonPressed(Button.A, () => {
    control.inBackground(() => {
        basic.showIcon(IconNames.SmallHeart)
    })
    control.inBackground(() => {
        basic.showString("Hello!")
    })
}) 
```
Note that control.inBackground is a one-shot function with no guarding event, so it is fine to call it multiple times to queue work to run in parallel.

PR #2660, opened 2017-08-05
Support statements with handlers

riknoll:
Fixes https://github.com/Microsoft/pxt/issues/2553

Adds support for statements that take in a callback. Previously, any block with a callback was automatically treated as an event and could not be nested. These blocks get disabled when top level, just like other statements:

run_in_background2

@cdiddy77


pelikhan:
Yes!

Issue #2582, opened 2017-07-30
Toolbox hides menu items in JavaScript mode


sgk:
See the picture below. The toolbox hides the left most part of the menu items. This happens to the translated languages which the toolbox width changes if "Advanced" is open.

Try as follows:
1. Change the language to Japanese or Sinhala.
1. Click "Advanced" to open it.
1. Click "Basic" or any other toolbox item.

![image](https://user-images.githubusercontent.com/485416/28752239-696d4fe4-7555-11e7-85f8-3c79a7191a1c.png)

microsoftsam:
Fixed in /beta

PR #2663, opened 2017-08-07
Resize the monaco editor after clicking Advanced 

microsoftsam:
In case the Advanced section introduces categories with greater length like in #2582

Fixes 2582


microsoftsam:
Needs to be cherry-picked to master.

Issue #2651, opened 2017-08-03
Tutorial authoring needs better documentation


riknoll:
Pretty slim right now: https://makecode.com/writing-docs/tutorials

We need to document all of the changes that have been made and the "magic" things like how we populate the toolbox

ganicke:
Updated in https://github.com/Microsoft/pxt/pull/2665.

PR #2665, opened 2017-08-07
Expand the 'tutorials' doc page.

ganicke:
Ok, here's a stab at updating the _writing_tutorials_ page. Let me know if I should say something more or if there's something incorrect.

Addresses https://github.com/Microsoft/pxt/issues/2651.


pelikhan:
A tutorial needs to be referenced from the "tutorials.md" page in order to show up there as well.

Issue #2648, opened 2017-08-03
[Screen Reader-Home Page-Blog]: Alt Attribute is not defined for the image showing near Documention link on blog page.


v-mesala:
**User Experience:** 
Users who depends on Screen Reader will get confuse if Atl attribute is not defined for the interactive images,then information about the image will get missed from users.

**Test Environment:**
OS: RS2 Version 1703(OS Build 15063.483)
Platform: Edge.
Tool Used: IE11 Developer (F12)

**Repro Steps:-**
1. Navigate to https://makecode.com/ 
2. Navigate to the Blog in header section and select it.
3. Navigate to the image showing near Documentation link in left panel.
6. Verify that alt attribute is defined or not for the image.
 
**Actual Result:-**
Alt Attribute is not defined for the image showing near Documention link on blog page.

**Expected Result:-**
Alt Attribute should be defined for the image showing near Documention link on blog page.

**Note:** This issue exists with NVDA +FF.
 
**MAS Reference -**
https://microsoft.sharepoint.com/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={d2d2051f-bdc8-4af7-8e18-38aae867e216}

**Suggested Fix:**
Using a text alternative on one item within a group of images that describes all items in the group 
e.g.
<p>Rating: 
         <img src="star1" alt="3 out of 5 stars">
          <img src="star1" alt="">
          < img src="star1" alt="">
           < img src="star2" alt="">
           < img src="star2" alt="">
  </p>

**Please refer the attachment.**
![mas1 1 1_homepage_blog](https://user-images.githubusercontent.com/30434417/28909353-c570203a-7845-11e7-9199-db39d460b6c8.jpg)

v-mesala:
Bug is verified on URL : https://makecode.com/ and bug still exists as Still alt is not defined for the image showing near Documentation.Hence Reactivated the bug and added the Active tag.

v-mesala:
Verified on URL: https://makecode.com/ and issue is fixed now. Hence closing the bug

Issue #2626, opened 2017-08-02
[Screen Reader-Home Page-Blog] Name Property and Control type is not defined for the 'Search Icon'  showing near Search edit box.


v-mesala:
**User Experience:**
Users with Visual disabilities might not able to understand the controls if Name Property is not defined. Like in this scenario Name Property is not defined for the Print button. 

**Test Environment**
OS: RS2 Version 1703(OS Build 15063.2)
Platform: Edge.
Screen reader: Narrator

**Repro Steps**
1-Navigate to url:- https://makecode.com/
2-Navigate to the Blog control lying in the header section on the page and select it.
3-Navigate to the Search icon showing near Search editable box  on blog page.
4-Verify that Name Property and Control type is defined for  'Search Icon' or not.

**Actual Result**
Name Property and Control type is not defined for the 'Search Icon'  showing near Search edit box  as Screen Reader reads it as Text on scan mode and blank in Items mode. 

**Expected Result**
Name Property and Control type should be defined for the 'Search Icon' showing near Search edit box.

**MAS Reference**
https://microsoft.sharepoint.com/teams/msenable/Documents/LCA/MAS%20-%20July%201,%202017%20Release/MAS%204.1.2%20%E2%80%93%20Name,%20Role,%20Value.docx?d=w248054a65e6847719e1e242fb5025730&csf=1

**Suggested fix:**
1- Using aria-label to give an invisible name to element, where name/title property is not used 
e.g.  
'Close' button can have only ‘X’, to make it more meaningful aria-label attribute is used with ‘Close’ text to help AT’s without showing the content on UI 
 
<div id="box"> 
   This is a pop-up box. 
     <button aria-label="Close" onclick="document.getElementById('box').style.display='none';" class="closebutton"> X </button>                                          
</div> 
 
2-   Using ARIA role to expose the control type of user interface component 
e.g. 
<div role="toolbar" tabindex="0" id="customToolbar" >  
<img src="img/btn1.gif" role="button" tabindex="-1" alt="Home" id="b1" title="Home"> 
 
3-   Using ARIA attributes to expose the element’s state  
e.g. 
<div role="button" id="customToolbar" aria-checked=”” aria-disabled=”” aria-haspopup=”” aria-expanded=”” aria-selected=””> 

**Refer Attachment:**
[...]

Issue #2625, opened 2017-08-02
[Keyboard Navigation-Home Page-Blog]: Search icon is not accessible using keyboard showing near Search edit box.


v-mesala:
**User Experience:**
User who depends on Keyboard with Screen Reader may get restricted from accessing all the functionalities if that element is not accessible using keyboard.

**Test Environment**
OS: RS2 Version 1703(OS Build 15063.2)
Platform: Edge.

**Repro Steps**
1-Navigate to url:- https://makecode.com/
2-Navigate to the Blog showing in header section and press enter on it.
3-Navigate to the Search editable box.
4-Verify that Search icon showing near Search box is accessible using keyboard or not.

**Actual Result**
Search icon is not accessible using keyboard showing near Searcg edit box on Blog Page.

**Expected Result**
Search icon should be accessible using keyboard showing near Searcg edit box on Blog Page.

**MAS Reference**
https://microsoft.sharepoint.com/teams/msenable/Documents/LCA/MAS%20-%20July%201,%202017%20Release/MAS%202.1.1%20%E2%80%93%20Keyboard.docx?d=w8492c4ebc17940ae8777cd044ed725a2&csf=1

**Suggested fix:**
Applicability: Web  
1- Always ensure each functionality of a page is keyboard accessible. 
E.g. A page with images used as links changes when the user hovers over the image with a mouse. To provide keyboard users with a similar experience, the image is also changed when a user tabs to it 
2-Always try to use standard HTML form controls and link elements to provide keyboard operation and assistive technology interoperability of interactive user interface elements. 
3-Use of both keyboard-specific and mouse-specific events with code that has a scripting function associated with an event. 
Use mousedown with  keydown   
Use mouseup  with keyup   
Use click   with  keypress    
Use mouseover with  focus   
Use mouseout with blur   
e.g. 
<img onclick="nextPage();" onkeypress="nextPage();" tabindex="0" src="arrow.gif"  
alt="Go to next page">  
 
Key References for Fix: 
http://www.w3.org/TR/UNDERSTANDING-WCAG20/keyboard-operation-keyboard-operable.html.

**Refer Attachment**
[MAS2.1.1_Home Page_Blog.zip](https://github.com/Microsoft/pxt/files/1193588/MAS2.1.1_Home.Page_Blog.zip)

veler:
[...]

Issue #2623, opened 2017-08-02
[Screen Reader-Home Page-Blog] Name Property is not defined for the 'Print ' button showing on Blog Page.


v-mesala:
**User Experience:**
Users with Visual disabilities might not able to understand the controls if Name Property is not defined. Like in the scenario Name Property is not defined for the Print button. 

**Test Environment**
OS: RS2 Version 1703(OS Build 15063.2)
Platform: Edge.
Screen reader: Narrator

**Repro Steps**
1-Navigate to url:- https://makecode.com/
2-Navigate to the Blog control lying in the header section on the page and select it.
3-Navigate to the Print button showing on blog page.
4-Verify that Name Property is defined for the Print bbutton or not.

**Actual Result**
Name Property is not defined for the 'Print ' button showing on Blog Page.

**Expected Result**
Name Property should be defined for the 'Print ' button showing on Blog Page.

**MAS Reference**
https://microsoft.sharepoint.com/teams/msenable/Documents/LCA/MAS%20-%20July%201,%202017%20Release/MAS%204.1.2%20%E2%80%93%20Name,%20Role,%20Value.docx?d=w248054a65e6847719e1e242fb5025730&csf=1

**Suggested fix:**
1- Using aria-label to give an invisible name to element, where name/title property is not used 
e.g.  
'Close' button can have only ‘X’, to make it more meaningful aria-label attribute is used with ‘Close’ text to help AT’s without showing the content on UI 
 
<div id="box"> 
   This is a pop-up box. 
     <button aria-label="Close" onclick="document.getElementById('box').style.display='none';" class="closebutton"> X </button>                                          
</div> 
 
2-   Using ARIA role to expose the control type of user interface component 
e.g. 
<div role="toolbar" tabindex="0" id="customToolbar" >  
<img src="img/btn1.gif" role="button" tabindex="-1" alt="Home" id="b1" title="Home"> 
 
3-   Using ARIA attributes to expose the element’s state  
e.g. 
<div role="button" id="customToolbar" aria-checked=”” aria-disabled=”” aria-haspopup=”” aria-expanded=”” aria-selected=””> 

**Refer Attachment:**
![mas4 1 2_home page_blog](https://user-images.githubusercontent.com/30434417/28869616-f68701da-779a-11e7-840f-9199c0460a40.jpg)

v-mesala:
Bug is verified on URL : https://makecode.com/ bug still exists .Hence reactivated the bug and added the Active tag.

[...]

PR #2666, opened 2017-08-07
[Accessibility] Documentation menu accessible with keyboard and screen reader

veler:
- Made the tree view in the side bar accessible with the keyboard and screen reader.
- Made the breadcrumb accessible by the screen reader.
- Added a hidden menu to skip the side menu.
- Changed the behavior of the hamburger menu to give the focus to the side menu once opened.
- Made the menu descriptive for the screen reader in general.

Related issues : [https://github.com/Microsoft/pxt/issues/2625](https://github.com/Microsoft/pxt/issues/2625), [https://github.com/Microsoft/pxt/issues/2623](https://github.com/Microsoft/pxt/issues/2623), [https://github.com/Microsoft/pxt/issues/2648](https://github.com/Microsoft/pxt/issues/2648), [https://github.com/Microsoft/pxt/issues/2626](https://github.com/Microsoft/pxt/issues/2626)

### DIFF
--- a/docfiles/footer.html
+++ b/docfiles/footer.html
@@ -10,6 +10,6 @@
                 <!-- we need to force the browser to load this font -->
                 <div style='font-family: Icons; color: #010101;' aria-hidden="true">.</div>
             </div>
-            <a class="item" href="https://www.microsoft.com/" target="_blank" rel="noopener" aria-label="Microsoft website"><img class="ui centered image" src="https://az851932.vo.msecnd.net/pub/pmapoirq" /></a>
+            <a class="item" href="https://www.microsoft.com/" target="_blank" rel="noopener" aria-label="Microsoft Home Page"><img class="ui centered image" src="https://az851932.vo.msecnd.net/pub/pmapoirq" /></a>
         </div>
     </div>
--- a/docfiles/footer.html
+++ b/docfiles/footer.html
@@ -10,6 +10,6 @@
                 <!-- we need to force the browser to load this font -->
                 <div style='font-family: Icons; color: #010101;' aria-hidden="true">.</div>
             </div>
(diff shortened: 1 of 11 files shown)


docfiles/footer.html
```diff
@@ -10,6 +10,6 @@
                 <!-- we need to force the browser to load this font -->
                 <div style='font-family: Icons; color: #010101;' aria-hidden="true">.</div>
             </div>
-            <a class="item" href="https://www.microsoft.com/" target="_blank" rel="noopener"><img class="ui centered image" src="https://az851932.vo.msecnd.net/pub/pmapoirq" /></a>
+            <a class="item" href="https://www.microsoft.com/" target="_blank" rel="noopener" aria-label="Microsoft website"><img class="ui centered image" src="https://az851932.vo.msecnd.net/pub/pmapoirq" /></a>
```
microsoftsam: Microsoft Home Page?

docfiles/style.css
```diff
@@ -186,6 +186,30 @@ svg {
...
+    font-size: 1em;
+    text-align: center;
+    color: white;
+}
+
+#accessibleMenu a:focus,
+#accessibleMenu a:hover {
+    top:20em;
+}
+
```
microsoftsam: Why don't we use the same css for both the docs and the editor? 
semantic.css is loaded in the docs.

docfiles/toc.html
```diff
@@ -1,6 +1,6 @@
 <div class="toc ui accent inverted hideprint">
     <div class="ui vertical accent inverted sticky menu tocmenu">
-        <a href="/" class="header item">
+        <a href="/" class="header item focused" tabindex="0" title="Go back to the editor">
```
microsoftsam: Not necessarily going back. Maybe "Go to editor"

docfiles/style.css
veler: I though it would be better to separate it in case we wanted something different. It's fixed but I still need a small part in style.css.

veler:
Adding some changes regarding the Keros validation

PR #2701, opened 2017-08-11
Adding minify option to static package

riknoll:
I already got approval for uglify-js. This runs minify on all JS files, including blockly and Monaco because doing so doesn't hurt anything. By default, uglify-js mangles and compresses the code. The other options are [here](https://www.npmjs.com/package/uglify-js#minify-options), let me know if there is anything I should be turning on or off.


cli/cli.ts
```diff
@@ -988,6 +992,16 @@ function uploadCoreAsync(opts: UploadOptions) {
                     content = content.replace(/\/(cdn|sim)\//g, "/blb/")
                 }
 
+                if (opts.minify && mime == "application/javascript" && fileName !== "target.js") {
```
pelikhan: minified?

cli/cli.ts
```diff
@@ -3552,24 +3566,26 @@ export function staticpkgAsync(parsed: commandParser.ParsedCommand) {
     const route = parsed.flags["route"] as string || "";
     const ghpages = parsed.flags["githubpages"];
     const builtPackaged = parsed.flags["output"] as string || "built/packaged";
+    const minify = parsed.flags["minify"] as boolean;
```
pelikhan: !!parse.flags["minify"]

PR #2710, opened 2017-08-14
[Accessibility] Multiple minor fixes

veler:
- Fixed an issue where the hidden menu was staying opened when it was loosing the focus but has the mouse hover.
- Fixed an issue that was generating a warning in the console after closing a modal with the Escape key because of a bad state in React.
- Fixed an issue where the focus was not trapped in the Import modal when we closed another modal before the Project modal (for i.e : the issue was appearing after using the Language modal, and then the Projects/Import).
- Fixed an issue where, in High contrast mode, the Open documentation in a new tab button was not clearly visible (yellow color on white background when the High Contrast of the system was not enabled). I changed the color and put the background to black so it's visible even if the background color of the documentation is white (when the High Contrast of the operating system is disabled).
- Fixed an issue where the Open documentation in a new tab button was not working with the Space/Enter key.

Tested on IE, Chrome, FireFox, Safari with Narrator, Jaws and Voice Over. NVDA is buggy on my machine right now, I will reinstall it.

Without the High Contrast OS's setting enabled.

With the High Contrast OS's setting enabled.

### DIFF
--- a/theme/common.less
+++ b/theme/common.less
@@ -211,10 +211,6 @@ div.simframe > iframe {
     .ui.item.link:focus {
         top:20em;
     }
-
-    .ui.item.link:hover {
-        top:20em;
-    }
 }
 
 #menubar .ui.menu .item.editor-menuitem {
--- a/theme/highcontrast.less
+++ b/theme/highcontrast.less
@@ -93,13 +93,32 @@
         }
     }
 
-    .ui.button.download-button:hover > span, i {
+    .ui.button.download-button:hover > span, 
+    .ui.button.download-button:hover > i {
         color: white !important;
     }
 
     /* Hyperlinks */
     a {
         color: yellow !important;
+    }
+
+    #sidedocsbar {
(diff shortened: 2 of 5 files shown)


microsoftsam:
Screenshots please

veler:
Added :-)

theme/common.less
```diff
@@ -211,10 +211,6 @@ div.simframe > iframe {
     .ui.item.link:focus {
         top:20em;
     }
-
-    .ui.item.link:hover {
-        top:20em;
```
microsoftsam: What's this top css do?

veler: The hidden menu is in fact visible (otherwise we could not tab on it) but has a negative top value, so it's hidden from the user on the screen. By removing this CSS, we prevent to keep the hidden menu showed when it loose the focus and that the mouse is hover the menu bar. It happened to me a lot of time in fact. It's better like this.

PR #2725, opened 2017-08-15
Use target semantic and blockly css in the docs.

microsoftsam:
Use target semantic and blockly css in the docs.
Use minified Jquery version.

### DIFF
--- a/docfiles/head.html
+++ b/docfiles/head.html
@@ -1,8 +1,8 @@
 <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
 <meta name="viewport" content="width=device-width,height=device-height,user-scalable=no,initial-scale=1.0,maximum-scale=1.0,minimum-scale=1.0">
 
-<link rel="stylesheet" href="/doccdn/semantic.css" />
-<link rel="stylesheet" href="/doccdn/blockly.css" />
+<link rel="stylesheet" href="/blb/semantic.css" />
+<link rel="stylesheet" href="/blb/blockly.css" />
 <link rel="stylesheet" href="/docfiles/style.css" />
 <link rel="stylesheet" href="/docfiles/vs.css" />
 <link rel="stylesheet" href="/docfiles/target.css" />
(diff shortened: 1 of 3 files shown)


docfiles/head.html
```diff
@@ -1,8 +1,8 @@
 <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
 <meta name="viewport" content="width=device-width,height=device-height,user-scalable=no,initial-scale=1.0,maximum-scale=1.0,minimum-scale=1.0">
 
-<link rel="stylesheet" href="/doccdn/semantic.css" />
-<link rel="stylesheet" href="/doccdn/blockly.css" />
+<link rel="stylesheet" href="/blb/semantic.css" />
+<link rel="stylesheet" href="/blb/blockly.css" />
```
mmoskal: I don't think that `/blb/` paths are currently expanded in docs. I think the issue is that it's unclear against which version of the `-built` repo they should be expanded.

microsoftsam: Oh I see. so we'd have to use a standard template of CSS for all docs.

microsoftsam: How does this work for index.html? 

Do we just need to add it to this list in cli.ts: 
```

    let replFiles = [
        "index.html",
        "embed.js",
        "run.html",
        "docs.html",
        "siminstructions.html",
        "release.manifest",
        "worker.js",
        "tdworker.js",
        "monacoworker.js",
        "simulator.html",
        "sim.manifest",
        "sim.webmanifest",
    ]

```

microsoftsam: Also if it's not in that replFiles list, how is it replacing doccdn in the first place for head.html?

mmoskal: `doccdn` is replaced on the server - see https://github.com/Microsoft/pxt-backend/blob/master/src/routing.ts#L181

The `docfiles/*` are not uploaded with release - they are taken by the server straight from the `pxt-microbit` repo (and possibly from the appropriate tag of `pxt` repo).

You could check in `doccdn/blockly.css` or whatever in `pxt-microbit` and update it manually when needed - the server will take the `HEAD` version. Or I could look in the `HEAD` of the `-built` repo, but this will have versioning issues.

microsoftsam: Okay makes sense.

PR #2737, opened 2017-08-16
Minify CSS

dragonx:
This saves about 1MB


cli/cli.ts
```diff
@@ -993,12 +994,15 @@ function uploadCoreAsync(opts: UploadOptions) {
...
 
                 if (minified) {
-                    const res = uglify.minify(content);
-                    if (!res.error) {
-                        content = res.code;
-                    }
-                    else {
-                        pxt.log(`        Could not minify ${fileName} ${res.error}`)
+                    if (mime == "application/javascript") {
+                        const res = uglify.minify(content);
```
microsoftsam: Are there options we can pass here. @dragonx Do you know if uglifycss removes comments as well?

dragonx:
I don't see an option to get rid of comments completely.  The closest I see is this:
`                        content = uglifycss.processString(content, { uglyComments: true });`
That gets rid of newlines in comments, but it doesn't make any significant difference in the file size.  Would you like me to add that?  I have no preference.

microsoftsam:
I think I'd prefer to use the clean-css package. Let me look into this. 
https://github.com/jakubpawlowicz/clean-css

That has support for minifying and removing comments.

microsoftsam:
Closing in favour of #2763. Ended up going with cssnano.

PR #2739, opened 2017-08-17
Adding target initialization on static packages to ge the telemetry

abchatra:
Static packages didn't add the target information in the pxtConfig. This prevented the telemetry from them adding it.

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -3584,6 +3584,7 @@ function internalStaticPkgAsync(builtPackaged: string, label: string, minify: bo
         pkgversion: "0.0.0",
         fileList: pxtFileList("node_modules/pxt-core/").concat(targetFileList()),
         localDir,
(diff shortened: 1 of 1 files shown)


cli/cli.ts
```diff
@@ -3584,6 +3584,7 @@ function internalStaticPkgAsync(builtPackaged: string, label: string, minify: bo
         pkgversion: "0.0.0",
         fileList: pxtFileList("node_modules/pxt-core/").concat(targetFileList()),
         localDir,
+        target: pxt.appTarget.id,
```
pelikhan: - [ ] make "target" non optional in the interface definition
- [ ] provide default is appTarget.id is missing ( || "unknown")

abchatra: Target is used only when localDir is valid. For example it is not valid for ghPages. 
I am  fixing the second comment

PR #2744, opened 2017-08-17
[Accessibility] Fixed Close button alignment

veler:
Fixed an issue where the close button on the modals was not positioned correctly on Safari and iOS.

On left, Safari, on right Google Chrome.

screen shot 2017-08-17 at 3 45 55 pm
screen shot 2017-08-17 at 3 46 07 pm
screen shot 2017-08-17 at 3 46 24 pm
screen shot 2017-08-17 at 3 46 43 pm

### DIFF
--- a/webapp/src/sui.tsx
+++ b/webapp/src/sui.tsx
@@ -887,6 +887,12 @@ export class Modal extends data.Component<ModalProps, ModalState> {
 
         const modalJSX = (
             <div className={classes} style={{ marginTop }} ref={this.handleRef} role="dialog" aria-labelledby={this.props.header ? this.id + 'title' : undefined} aria-describedby={this.props.description ? this.id + 'description' : this.id + 'desc'} >
+                {this.props.closeIcon ? <Button
+                    icon={closeIconName}
+                    class={`huge clear right floated closeIcon focused`}
+                    onClick={() => this.handleClose(null) }
+                    tabIndex={0}
+                    ariaLabel={lf("Close dialog")} /> : undefined }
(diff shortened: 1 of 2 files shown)


webapp/src/sui.tsx
```diff
@@ -887,6 +887,12 @@ export class Modal extends data.Component<ModalProps, ModalState> {
 
         const modalJSX = (
             <div className={classes} style={{ marginTop }} ref={this.handleRef} role="dialog" aria-labelledby={this.props.header ? this.id + 'title' : undefined} aria-describedby={this.props.description ? this.id + 'description' : this.id + 'desc'} >
+                {this.props.closeIcon ? <Button
```
microsoftsam: Why are you moving this to the top? Won't it interfere with the tab order?

veler: So here is the thing : 
- The Position and Float properties on Safari have a weird behavior, from what I've found, in particular with Absolute and Fixed. It's a known recurrent issue.
- The workaround I've found for the moment is to put the position to relative. But because of this, if I keep the button at this place, it appears at the bottom of the screen. By putting it at this place, it keeps the right position on all web browsers, Safari included, whatever the size of the screen is.

But yes, it interferes with the tab order, and even break it in the Share modal (partially because of the initializeFocus method, once again...). I'll continue on it tomorrow morning. Sorry about it.
My guess is that it is possible to fix it entirely with CSS, but I didn't found the solution yet even by mixing classical things like display, position, margin, float...
Any idea on your side maybe?

veler: Disclaimer : I guess I have a solution, I will check it soon :)

Issue #2749, opened 2017-08-18
Image literal parser does not handle CRLF new lines


pelikhan:
Easily happens when users copy/paste code between editors.

pelikhan:
```
basic.showLeds(`
   . # # # .
   . . . # .
   . . # # .
   . . . . .
   . . # . .
   `)
```

pelikhan:
All the code snippets using ``showLeds`` in the microbit CS Intro course are subject to this issue.

pelikhan:
Actually, it is broken because of ``00A0`` NON-BREAKING SPACES characters

PR #2754, opened 2017-08-18
V0 Don't break compilation on whitespace in image literals

riknoll:
Fixes https://github.com/Microsoft/pxt/issues/2749
This is a V0 PR, so don't merge.


pelikhan:
Sounds good. Do we have a test for those?

riknoll:
I don't think I can add a test for this; I'd have to add a C++ file to our test project and I'm not sure how easy it would be to get that to work in the CI.

PR #2755, opened 2017-08-19
Mostly cleanup, adding focus outline

microsoftsam:
Mostly cleaning up CSS. 

Adding a focus outline like so: 
```
*:focus{
    outline: 2px dotted black !important;
    z-index: 1000;
}
```

TODO:
- [ ] Need to make sure the blocklySVG doesn't capture focus
- [x] Need to make sure the search field doesn't capture tabs


microsoftsam:
@veler this should help resolve some of these button focus bugs.

veler:
The current changes looks good to me. But in the TODO, why do you want to make sure the search field doesn't capture tabs ?

Issue #2756, opened 2017-08-19
Flooding the server with requests when gallery page is not found


microsoftsam:
When one of the gallery pages is not found "Projects", "Examples" in docs. We flood the server with requests, and don't stop causing a large number of failures on the backend. 

<img width="736" alt="screen shot 2017-08-18 at 11 42 17 pm" src="https://user-images.githubusercontent.com/16690124/29484231-189f0b38-846f-11e7-997e-ad2ccc45718d.png">

Marking as P2 as workaround is to resolve the docs issue. Either remove gallery item, or add docs page.

pelikhan:
What's the frequency of requests here?

pelikhan:
``pxt checkdocs`` should validate galleries and fail if needed

pelikhan:
We do check for broken gallery pages.

PR #2758, opened 2017-08-19
Fixing ``pxt checkdocs``

pelikhan:
Properly check urls and snippets, fail build when there is a documentation error


microsoftsam:
This would avoid #2756 in the future?

pelikhan:
@microsoftsam no, it does not check for missing doc pages used in pxtarget

pelikhan:
@microsoftsam needs cherrypick in v2

PR #2768, opened 2017-08-22
[Accessibility] Fixed an issue related to the focus in the tutorial.

veler:
Fixed an issue where the focus was not initialized correctly in the tutorial, and set the first focused item to the OK button.

### DIFF
--- a/webapp/src/tutorial.tsx
+++ b/webapp/src/tutorial.tsx
@@ -207,7 +207,7 @@ export class TutorialCard extends data.Component<ISettingsProps, {}> {
                     let root = document.getElementById('root');
                     if (root.classList.contains('dimmed')) {
                         this.focusInitialized = true;
-                        document.getElementById('okButton').focus();
+                        document.getElementById('tutorialOkButton').focus();
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


webapp/src/tutorial.tsx
```diff
@@ -232,7 +245,7 @@ export class TutorialCard extends data.Component<ISettingsProps, {}> {
                     <div className={`tutorialmessage ${hasHint ? 'focused' : undefined}`} role="alert" aria-label={tutorialAriaLabel} tabIndex={hasHint ? 0 : -1} onClick={() => {if (hasHint) this.showHint();}} onKeyDown={sui.fireClickOnEnter}>
                         <div className="content" dangerouslySetInnerHTML={{__html: tutorialHeaderContent}} />
                     </div>
-                    <sui.Button class="large green okbutton showlightbox focused" text={lf("Ok") } onClick={() => this.closeLightbox() } onKeyDown={sui.fireClickOnEnter} />
+                    <sui.Button id="okButton" class="large green okbutton showlightbox focused" text={lf("Ok") } onClick={() => this.closeLightbox() } onKeyDown={sui.fireClickOnEnter} />
```
microsoftsam: Be more specific with your ID. okButton could mean any OK button.

veler: Right, furthermore we might have several okButton too in the future which would be problematic. I'll edit it.

PR #2769, opened 2017-08-22
Fixing appcache for target editor extensions using staticpkg

riknoll:
editor.js was not being appended to the release manifest when building with staticpkg.

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -925,7 +925,7 @@ function uploadCoreAsync(opts: UploadOptions) {
             "var pxtConfig = null": "var pxtConfig = " + JSON.stringify(cfg, null, 4),
             "@defaultLocaleStrings@": "",
             "@cachedHexFiles@": "",
-            "@targetEditorJs@": ""
+            "@targetEditorJs@": targetEditorJs ? opts.localDir + "editor.js" : ""
         }
     }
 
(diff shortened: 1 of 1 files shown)


cli/cli.ts
```diff
@@ -925,7 +925,7 @@ function uploadCoreAsync(opts: UploadOptions) {
             "var pxtConfig = null": "var pxtConfig = " + JSON.stringify(cfg, null, 4),
             "@defaultLocaleStrings@": "",
             "@cachedHexFiles@": "",
-            "@targetEditorJs@": ""
+            "@targetEditorJs@": targetEditorJs ? opts.localDir + "editor.js" : ""
         }
```
microsoftsam: Don't we want targetEditorJS here?

cli/cli.ts
```diff
@@ -925,7 +925,7 @@ function uploadCoreAsync(opts: UploadOptions) {
             "var pxtConfig = null": "var pxtConfig = " + JSON.stringify(cfg, null, 4),
             "@defaultLocaleStrings@": "",
             "@cachedHexFiles@": "",
-            "@targetEditorJs@": ""
+            "@targetEditorJs@": targetEditorJs ? opts.localDir + "editor.js" : ""
```
microsoftsam: use `${opts.localDir}editor.js` instead

microsoftsam:
@pelikhan please make sure this is the right fix here.

pelikhan:
LGTM

PR #2770, opened 2017-08-22
Accessibility!!!

microsoftsam:
This is a massive PR! Most of this code has been reviewed by myself (@microsoftsam), but I'd like everyone to have a pass on it, and do a final code review before we check this into v0. 

To try it, go to https://makecode.microbit.org/acc


pelikhan:
Is https://MakeCode.microbit.org/acc running the latest bits from here?

microsoftsam:
Bumping as we speak, if you wait a few minutes, it should.

pelikhan:
And I just answered myself reading the description...

docfiles/docs.js
```diff
@@ -47,8 +54,21 @@ function searchSubmit(form) {
 }
 
 function setupSidebar() {
+    let togglesidebar = document.getElementById("togglesidebar");
```
pelikhan: I don't think "let" works in older browsers

docfiles/docs.js
```diff
@@ -47,8 +54,21 @@ function searchSubmit(form) {
...
+    togglesidebar.onkeydown = handleEnterKey;
```
pelikhan: We have jquery loaded so should just use it here: ``$("#togglesidebar").keydown(handleEnterKey)``

docfiles/docs.js
```diff
@@ -47,8 +54,21 @@ function searchSubmit(form) {
...
+
     $('.ui.sidebar')
-        .sidebar({ dimPage: false })
+        .sidebar({ 
+            dimPage: false,
+            onShow: () => {
+                togglesidebar.setAttribute("aria-expanded", "true");
+                document.getElementsByClassName("sidebar").item(0).getElementsByClassName("focused").item(0).focus();
+            },
+            onHidden: () => {
```
pelikhan: Lambdas not supported in older browsers

docfiles/docs.js
```diff
@@ -47,8 +54,21 @@ function searchSubmit(form) {
...
+    togglesidebar.onkeydown = handleEnterKey;
+
+
     $('.ui.sidebar')
-        .sidebar({ dimPage: false })
+        .sidebar({ 
+            dimPage: false,
+            onShow: () => {
+                togglesidebar.setAttribute("aria-expanded", "true");
+                document.getElementsByClassName("sidebar").item(0).getElementsByClassName("focused").item(0).focus();
```
pelikhan: is it more reliable to use jquery here? ``$('.sidebar > focused').focus()``

docfiles/docs.html
```diff
@@ -21,12 +25,12 @@
...
     <!-- @include toc.html -->
     
-    <div class="article">
+    <div class="article" id="maincontent">
         
         <!-- @include header.html -->
 
         <div id="docs" class="main ui container mainbody">
-            <button id="printbtn" class="circular ui icon right floated button hideprint">
+            <button id="printbtn" class="circular ui icon right floated button hideprint" title="Print this page">
```
pelikhan: are these strings picked up in "webstrings.json"? (we can fix that in a future PR if needed)

pxtblocks/blocklyloader.ts
```diff
@@ -1183,6 +1196,13 @@ namespace pxt.blocks {
                     pxt.log("searching for: " + searchFor);
                     updateUsedBlocks = false;
                     if (!blocks) return;
+
+                    if (blocks.length == 0) {
+                        blocklySearchLabel.innerText = lf("No search results...");
+                    } else {
+                        blocklySearchLabel.innerText = lf("{0} results matching '{1}'", blocks.length, blocklySearchInputField.value.toLowerCase());
```
pelikhan: - Need to handle singular form of "results" using ``{0:s}`` I think.
- what if the value is huuuuuuuuuuuuuuuuuuuuuuge, will it overflow the UI?

pxtlib/docsrender.ts
```diff
@@ -275,10 +278,17 @@ namespace pxt.docs {
             params["github"] = "";
         }
 
+        // Add accessiblity menu 
+        const accMenuHtml = `
+            <a href="#maincontent" class="ui item link" tabindex="0" role="menuitem">${lf("Skip to main content")}</a>
```
pelikhan: What is "main content" supposed to be? Aren't they skipping to the JavaScript editor?

pxtsim/accessibility.ts
```diff
@@ -0,0 +1,59 @@
...
+namespace pxsim.accessibility {
+    let liveRegion: HTMLDivElement;
+
+    export function makeFocusable(elem: SVGElement): void {
+        elem.setAttribute("focusable", "true");
+        elem.setAttribute("tabindex", "0");
+    }
+
+    export function enableKeyboardInteraction(elem: Element, handlerKeyDown?: () => void, handlerKeyUp?: () => void): void {
+        elem.addEventListener('keydown', (e: KeyboardEvent) => {
```
pelikhan: don't attach handler is handlerKeyDown is not defined

pxtsim/accessibility.ts
```diff
@@ -0,0 +1,59 @@
...
+            let charCode = (typeof e.which == "number") ? e.which : e.keyCode
+            if ((charCode === 32 || charCode === 13) && handlerKeyDown) {
+                handlerKeyDown();
+            }
+        });
+        elem.addEventListener('keyup', (e: KeyboardEvent) => {
```
pelikhan: do not attach handler if handlerKeyUp is not defined

webapp/src/app.tsx
```diff
@@ -1519,6 +1533,7 @@ ${compileService ? `<p>${lf("{0} version:", "C++ runtime")} <a href="${Util.html
         const hc = !this.state.highContrast;
         pxt.tickEvent("menu.highcontrast", { on: hc ? 1 : 0 });
         this.setState({ highContrast: hc }, () => this.restartSimulator());
+        core.highContrast = hc;
```
pelikhan: the meaning of "hc" is weird. is on or off?

webapp/src/app.tsx
```diff
@@ -1616,75 +1632,85 @@ ${compileService ? `<p>${lf("{0} version:", "C++ runtime")} <a href="${Util.html
...
+
+                                    </sui.DropdownMenuItem> }
 
                                 {sandbox && !targetTheme.hideEmbedEdit ? <sui.Item role="menuitem" icon="external" textClass="mobile hide" text={lf("Edit") } onClick={() => this.launchFullEditor() } /> : undefined}
                                 {inTutorial ? <sui.ButtonMenuItem class="exit-tutorial-btn" role="menuitem" icon="external" text={lf("Exit tutorial") } textClass="landscape only" onClick={() => this.exitTutorial(true) } /> : undefined}
-
-                                {!sandbox ? <a id="organization" href={targetTheme.organizationUrl} target="blank" rel="noopener" className="ui item logo" onClick={() => pxt.tickEvent("menu.org") }>
+                                {!sandbox ? <a id="organization" href={targetTheme.organizationUrl} aria-label={`${targetTheme.organization} Logo`} role="menuitem" target="blank" rel="noopener" className="ui item logo" onClick={() => pxt.tickEvent("menu.org") }>
```
pelikhan: label needs to be localizeable

webapp/src/core.ts
```diff
@@ -11,6 +11,8 @@ import * as pkg from "./package";
 import Cloud = pxt.Cloud;
 import Util = pxt.Util;
 
+export let highContrast: boolean;
```
pelikhan: what is this variable for? Seems that we are leaking state here.

webapp/src/core.ts
```diff
@@ -342,14 +367,20 @@ export function promptAsync(options: PromptOptions): Promise<string> {
...
                         </div>`;
 
     options.onLoaded = () => {
         let dialogInput = document.getElementById('promptDialogInput') as HTMLInputElement;
         if (dialogInput) {
-            dialogInput.focus();
             dialogInput.setSelectionRange(0, 9999);
+            dialogInput.onkeyup = (e: KeyboardEvent) => {
+                let charCode = (typeof e.which == "number") ? e.which : e.keyCode
+                if (charCode === 13 || charCode === 32) {
```
pelikhan: the "enter" key detection is repeated in various places, help?

webapp/src/core.ts
```diff
@@ -450,6 +481,88 @@ export function scrollIntoView(item: JQuery, margin = 0) {
...
+    data.lastTag.removeEventListener('keydown', data.targetArea.focusDataInfo.giveFocusToFirstTagBinding);
+    if (data.firstTag === data.lastTag) {
+        data.firstTag.removeEventListener('keydown', data.targetArea.focusDataInfo.giveFocusToFirstTagBinding);
+        data.lastTag.removeEventListener('keydown', data.targetArea.focusDataInfo.giveFocusToLastTagBinding);
+    }
+}
+
+function giveFocusToFirstTag(e: KeyboardEvent) {
+    let charCode = (typeof e.which == "number") ? e.which : e.keyCode
+    if (charCode === 9 && !e.shiftKey) {
```
pelikhan: 9 is tab?

webapp/src/editortoolbar.tsx
```diff
@@ -122,7 +122,7 @@ export class EditorToolbar extends data.Component<ISettingsProps, {}> {
                     <div className="ui equal width grid">
                         <div className="left aligned column">
                             <div className="ui icon small buttons">
-                                <sui.Button icon={`${collapsed ? 'toggle up' : 'toggle down'}`} class={`collapse-button ${collapsed ? 'collapsed' : ''} ${hideEditorFloats ? 'disabled' : ''}`} title={collapseTooltip} onClick={() => this.toggleCollapse('mobile') } />
+                                <sui.Button icon={`${collapsed ? 'toggle up' : 'toggle down'}`} class={`collapse-button ${collapsed ? 'collapsed' : ''} ${hideEditorFloats ? 'disabled' : ''}`} ariaLabel={lf("{0}, {1}", collapseTooltip, hideEditorFloats ? lf("Disabled") : "")} title={collapseTooltip} onClick={() => this.toggleCollapse('mobile') } />
```
pelikhan: what is the label supposed to show?

webapp/src/scriptsearch.tsx
```diff
@@ -201,17 +201,19 @@ export class ScriptSearch extends data.Component<ISettingsProps, ScriptSearchSta
...
+                description={lf("Add a package to the project") }>
                 <div className="ui vertical segment">
                     <div className="ui search">
                         <div className="ui fluid action input" role="search">
-                            <input ref="searchInput" type="text" placeholder={lf("Search or enter project URL...") } onKeyUp={kupd} />
+                            <input ref="searchInput" className="focused" type="text" placeholder={lf("Search or enter project URL...") } onKeyUp={kupd} />
                             <button title={lf("Search") } className="ui right icon button" onClick={upd}>
                                 <i className="search icon"></i>
                             </button>
+                            <div aria-live="polite" className="accessible-hidden">{lf("{0} results matching '{1}'", bundles.length + ghdata.length + urldata.length, this.state.searchFor)}</div>
```
pelikhan: singular form handling needed

webapp/src/share.tsx
```diff
@@ -172,20 +173,22 @@ pxt extract ${url}`;
                         : undefined }
                     { url && ready ? <div>
                         <p>{lf("Your project is ready! Use the address below to share your projects.") }</p>
-                        <sui.Input class="mini" readOnly={true} lines={1} value={url} copy={true} selectOnClick={true}/>
+                        <sui.Input id="projectUri" class="focused mini" readOnly={true} lines={1} value={url} copy={true} selectOnClick={true} aria-describedby="projectUriLabel"/>
+                        <label htmlFor="projectUri" id="projectUriLabel" className="accessible-hidden">{lf("This is the read-only internet address of your project. You can copy it by pressing Control C.")}</label>
```
pelikhan: Is Ctrl+C supported on all accessible platforms?

webapp/src/share.tsx
```diff
@@ -172,20 +173,22 @@ pxt extract ${url}`;
...
                             <sui.Menu pointing secondary>
                                 {formats.map(f =>
-                                    <sui.MenuItem key={`tab${f.label}`} active={mode == f.mode} name={f.label} onClick={() => this.setState({ mode: f.mode }) } />) }
+                                    <sui.MenuItem key={`tab${f.label}`} id={`tab${f.mode}`} active={mode == f.mode} name={f.label} onClick={() => this.setState({ mode: f.mode }) } />) }
                             </sui.Menu> : undefined }
                         { advancedMenu ?
                             <sui.Field>
-                                <sui.Input class="mini" readOnly={true} lines={4} value={embed} copy={ready} disabled={!ready} selectOnClick={true}/>
+                                <sui.Input id="embedCode" class="mini" readOnly={true} lines={4} value={embed} copy={ready} disabled={!ready} selectOnClick={true}/>
+                                <label htmlFor="embedCode" id="embedCodeLabel" className="accessible-hidden">{lf("This is the read-only code for the selected tab. You can copy it by pressing Control C.")}</label>
```
pelikhan: Is Control C supported on all accessible platforms?

webapp/src/srceditor.tsx
```diff
@@ -3,6 +3,7 @@ import * as core from "./core";
 import * as data from "./data";
 import * as app from "./app";
 import * as React from "react";
+import * as pxtjson from "./pxtjson"
```
pelikhan: not used?

webapp/src/sui.tsx
```diff
@@ -52,6 +56,14 @@ function removeClass(el: HTMLElement, cls: string) {
     else if (el.className.indexOf(cls) >= 0) el.className.replace(new RegExp(`(?:^|\\s)${cls}(?:\\s|$)`), ' ');
 }
 
+export function fireClickOnEnter(e: React.KeyboardEvent): void {
+    let charCode = (typeof e.which == "number") ? e.which : e.keyCode
+    if (charCode === 13 || charCode === 32) {
```
pelikhan: one more enter key detection logic

webapp/src/sui.tsx
```diff
@@ -79,15 +91,78 @@ export class UiElement<T extends WithPopupProps> extends data.Component<T, {}> {
...
 
 export class DropdownMenuItem extends UiElement<DropdownProps> {
+    private isOpened = false
+    private preventHide = false
+
+    private menuItemKeyDown = (e: KeyboardEvent) => {
+        let charCode = (typeof e.which == "number") ? e.which : e.keyCode
+        if (charCode === 9) {
+            this.close()
+        } else if (charCode === 32 || charCode === 13) {
```
pelikhan: another enter key detection logic

webapp/src/sui.tsx
```diff
@@ -79,15 +91,78 @@ export class UiElement<T extends WithPopupProps> extends data.Component<T, {}> {
 }
 
 export class DropdownMenuItem extends UiElement<DropdownProps> {
+    private isOpened = false
+    private preventHide = false
+
+    private menuItemKeyDown = (e: KeyboardEvent) => {
+        let charCode = (typeof e.which == "number") ? e.which : e.keyCode
+        if (charCode === 9) {
```
pelikhan: and tab detection

webapp/src/sui.tsx
```diff
@@ -846,12 +1023,22 @@ export class Portal extends data.Component<PortalProps, PortalState> {
         }
     }
 
-    close = (e: Event) => {
-        const { onClose } = this.props;
-        if (onClose) onClose(e);
+    private handleEscape = (e: KeyboardEvent) => {
+        let charCode = (typeof e.which == "number") ? e.which : e.keyCode
+        if (charCode !== 27) {
```
pelikhan: comment to describe which key it is

webapp/src/sui.tsx
```diff
@@ -871,6 +1058,7 @@ export class Portal extends data.Component<PortalProps, PortalState> {
         mountNode.appendChild(this.rootNode);
 
         document.addEventListener('click', this.handleDocumentClick)
+        document.addEventListener('keydown', this.handleEscape, true)
```
pelikhan: so this one has true, but not the above?

pelikhan:
This is a ton of work! Very well done!

pelikhan:
Bug: When I click on download, the dialog opens and focuses on "done". However, if I keep tabbing, it does not cycle through the other buttons.

pelikhan:
Blockly tooltips are not high contrast
image

pelikhan:
When opening the "Add Package" dialog, focus is on the help icon instead of the search bar
image

pelikhan:
Aside from various small comments, LGTM

pxtblocks/blocklyloader.ts
veler: This text is not displayed to the user. It's a hidden label for the screen reader. Maybe a good idea to rename it to "hiddenBlocklySearchLabel" or something like that.

docfiles/docs.js
veler: It is supported from IE 11 on Windows 8.1. In my country (at least) it is true that a lot of schools are still on Windows 7.

docfiles/docs.js
veler: So, @microsoftsam told me that in long term we want to stop to use JQuery. What is your opinion on that?

docfiles/docs.js
veler: Right. It is supported from IE 11 since 2015 (not 2013) with the release of W10.

pxtlib/docsrender.ts
veler: It's for the documentation. There is a hidden menu that appear when you start to tab into the documentation that let you skip the side menu and tab directly into the content.

pxtsim/accessibility.ts
veler: Well see !

webapp/src/core.ts
veler: Yes, it might be a good idea to use a constant here.

webapp/src/editortoolbar.tsx
veler: This is a hidden text for the screen reader. It must says :
- "Show the simulator"
- or "Show the simulator, disabled" (when the button is disabled)
- or "Hide the simulator"
- or "Hide the simulator, disabled"

webapp/src/scriptsearch.tsx
veler: Thanks =)

docfiles/docs.js
pelikhan: TypeScript compiles away "let" - since doc.js is JavaScript - we need to do it ourselves.

docfiles/docs.js
pelikhan: Lambdas are compiled away by TypeScript. Since this is JS plain, we should use function instead.

webapp/src/core.ts
pelikhan: Or a little comment -- or a helper

docfiles/docs.js
microsoftsam: Yeah let's stay away from JQuery, there's no need to have react and JQuery. This isn't as relevant for v0, but since this PR would be going to master, I want to stay away from JQuery.

microsoftsam: ^^ That being said, we have JQuery today and the reason we have it is because of semantic ui. Once we switch to a fully react semantic, we probably will have to refactor this code anyway.

pelikhan: True for the editor, but the docs do not use React. They use Semantic + JQuery.

microsoftsam: ^^ Also being said, this code here is in docs.js which probably will never get react, in which case we'd have to keep using semantic ui JS, and so maybe Jquery?

pelikhan: The docs will never migrate to react (or very very unlikely). We already use JQuery extensively in this part of the house when rendering snippets.

docfiles/docs.html
microsoftsam: This should be in docsrenderer just like accMenu above.

pxtblocks/blocklyloader.ts
microsoftsam: Yes please rename.

pxtlib/docsrender.ts
microsoftsam: Skips search and sidemenu and takes you directly to content.

webapp/src/share.tsx
microsoftsam: Nope. It's Cmd C on Mac.. Can we assume an accessible user that is using a screenreader knows how to Copy something with their native app.

webapp/src/share.tsx
microsoftsam: Same as above.

webapp/src/share.tsx
microsoftsam: @veler I would remove the last bit.

docfiles/docs.js
veler: Got it !

webapp/src/core.ts
microsoftsam: core.ts doesn't have access to the state, and msg alerts show up at the top level (above content and root), and so the standard trick of applying the hc class on root won't work for them, hence why we need an extra field to communicate that across.

veler: It allow us to display the #msg in high contrast (the temporary banner that says for example "building .hex file"). As this DOM is outside of #root, we had to pass the setting to core when it displays the banners.

veler:
In my last commit : 

- renamed blocklySearchLabel to blocklyHiddenSearchLabel in blocklyloader.ts.
- refactored enableKeyboardInteraction om accessibility.ts.
- refactored the key handlers.
- removed unused code.
- removed the "You can copy it by pressing Control C" in share.tsx
- fixed a grammar mistake.
- fixed an issue where the modal that appear after a download had a bad keyboard navigation.
- Changed the outline on focused items, and changed the render of some components in high contrast mode.

untitled

theme/accessibility.less
```diff
@@ -0,0 +1,61 @@
...
+/* Reference import */
+@import (reference) "semantic.less";
+
+/*******************************
+          Accessibility
+*******************************/
+
+*:focus{
+    outline: 2px solid #4D90FE !important;
+}
```
microsoftsam: Let's specialize this a little more. Doing this for all elements is not a good call, given that some elements could be disabled.
Here's what microsoft.com uses: 

```
a:focus:not([tabindex='-1']), input:not([disabled]):focus:not([tabindex='-1']), select:not([disabled]):focus:not([tabindex='-1']), textarea:not([disabled]):focus:not([tabindex='-1']), button:not([disabled]):focus:not([tabindex='-1']), [tabindex]:focus:not([tabindex='-1']), [contentEditable=true]:focus:not([tabindex='-1'])
```

abchatra:
Is it possible to squash some of the commits and keep what is relevant? 84 commits is too much for anything useful. But I don't want to lose all the history.

microsoftsam:
In order to make merging not a nightmare, we will be squashing all commits here.

abchatra:
ok makes sense

PR #2778, opened 2017-08-23
Param default fixes to v0

microsoftsam:
A number of issues here with the param default parsing: 
- Showing null for optional fields with initializers
- Not showing defaults for when the "eg" field is used. Looks like this issue was fixed in master but never cherry-picked to v0. 
- When the parsed value is a string, not adding quotes. 

Screenshot before:
screen shot 2017-08-22 at 5 32 17 pm

Screenshot after: 
screen shot 2017-08-22 at 5 40 09 pm


microsoftsam:
Only part of this change needs to be cherry-picked to master. "Adding quotes around string literals"

microsoftsam:
master change: https://github.com/Microsoft/pxt/pull/2780

pxtcompiler/emitter/service.ts
```diff
@@ -887,13 +887,16 @@ namespace ts.pxtc.service {
...
-        const args = n.parameters ? n.parameters.filter(param => !param.questionToken).map(param => {
+        const args = n.parameters ? n.parameters.filter(param => !param.initializer && !param.questionToken).map(param => {
             const typeNode = param.type;
             if (!typeNode) return "null";
 
             const name = param.name.kind === SK.Identifier ? (param.name as ts.Identifier).text : undefined;
 
             if (attrs && attrs.paramDefl && attrs.paramDefl[name]) {
+                if (typeNode.kind == SK.StringKeyword) {
+                    return `"${attrs.paramDefl[name]}"`;
```
riknoll: Check if there are quotes first

microsoftsam: I thought we pulled out quotes at that point??

PR #2779, opened 2017-08-23
Tying together UI and extension manager

riknoll:
This is a PR against the packageextension branch. Just connecting my code and Sam's code.


microsoftsam:
Looks good. It's great that we're using the same protocol, but do we want to allow packages to control things like switching to Blocks/JS? proxytosim, etc.

@pelikhan

riknoll:
@microsoftsam they can only do that if the target sets the allowParentController flag. I added a separate flag for package extensions.

PR #2780, opened 2017-08-23
Add quotes around string literals when compiling the default parameter

microsoftsam:
Related to #2778 
(master change)

Also fixing adafruit int32 issue and initializer issue: 
screen shot 2017-08-23 at 11 08 12 am
screen shot 2017-08-23 at 11 08 08 am


microsoftsam:
@riknoll Not relevant for v0, but does your check at line https://github.com/Microsoft/pxt/blob/dab97531b42f2adaec9dd229d0d1a5a34d5e0eb3/pxtcompiler/emitter/service.ts#L908
cover our int32 number. 

I see this in adafruit: 
screen shot 2017-08-23 at 12 33 00 am

and this is it's definition:
```
    function playTone(frequency: int32, ms: int32): void;
```

Issue #2604, opened 2017-08-01
[Supporting the platform-Home Page-Navigation]Focus indicator is not visible on all the Sections element like links on all 4 High contrast modes. 


v-mesala:
**User Experience:**
If a user is not able to see the content properly in any high contrast themes, then this will make it difficult for him to interpret the information properly. This will hinder him from to use the page properly. 

**Test Environment:**
OS: RS2 Version 1703(OS Build 15063.483)
Platform: Edge.

**Pre-Requisite:** Turn on the High Contrast Settings(Settings>Ease of Access>High Contrast)

**Repro Steps:-**
1. Navigate to url:- https://makecode.com/
2. Navigate through all controls (say computing education, about Microsoft makecode, get inspired, resources elements etc)lying on the page.3. Navigate to the Help control lying in the header section on the page and select it.
3. Turn on the High Contrast mode.
4. Verify that focus indicator and all the elements are visible under all 4 HC  modes.

**Actual Result:-**
Focus indicator is not visible on all the Sections element on all 4 High contrast modes. (HC#1,HC#2,HC Black and HC White)

**Expected Result:-**
Focus indicator should be visible on all the Sections element on all 4 High contrast modes. (HC#1,HC#2,HC Black and HC White)

**MAS Reference -**
https://microsoft.sharepoint.com/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={672c8a8b-ed4f-463f-868f-e8131caf9b74}

**Suggested fix:** 
a. Ensure web application adheres to the colors of platform provided high contrast themes 
Ensure the web application content colors are updated as per the provided colors for that specific theme 
Ex: Links/Buttons/Text/Selected text have different set of colors for different themes 
b. Ensure text/icons/images are all visible even with high contrast themes on 
Ex1: In some cases background images are used and they will be invisible due to darker background color of selective theme 
Ex2: white colored icons/images are used which will get merged with white background colored high contrast theme 

**Please refer the Video attached.**

[MAS4.3.1_Home Page.zip](https://github.com/Microsoft/pxt/files/1190140/MAS4.3.1_Home.Page.zip)

microsoftsam:
@veler something you can look into? I can show you how to edit the makecode.com page if you know what the fix is here.

veler:
Fix available on [https://makecode.microbit.org/](https://makecode.microbit.org/)

v-mesala:
[...]

PR #2810, opened 2017-08-25
[Accessibility] Removed focus outline on touch screen

veler:
Removed the focus outline on touch screen.
I tested on Android, it works well.
On Edge on SurfaceBook, is doesn't change anything, probably because it still considerate that I have a mouse.
Does anyone can try with iOS ? The simulator doesn't work on my account on Peli's MacBook.

Also, I fixed some focus related issue in the index page of MakeCode. Related issue : [https://github.com/Microsoft/pxt/issues/2604](https://github.com/Microsoft/pxt/issues/2604)

### DIFF
--- a/theme/accessibility.less
+++ b/theme/accessibility.less
@@ -9,9 +9,16 @@
           Accessibility
 *******************************/
 
-*:focus{
-    outline: 2px solid #4D90FE !important;
+*:focus {
+    outline: 1px solid transparent !important;
 }
+
+@media not all and (pointer:coarse) {
+    *:focus {
+        outline: 2px solid #4D90FE !important;
+    }
+}
+
 .blocklyTreeRoot:focus {
     outline: none !important;
 }
--- a/theme/accessibility.less
+++ b/theme/accessibility.less
@@ -14,7 +14,7 @@
 }
 
 @media not all and (pointer:coarse) {
-    *:focus {
+    *:not([tabindex='-1']):focus {
         outline: 2px solid #4D90FE !important;
     }
 }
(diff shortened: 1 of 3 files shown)


theme/accessibility.less
```diff
@@ -9,9 +9,16 @@
           Accessibility
 *******************************/
 
-*:focus{
-    outline: 2px solid #4D90FE !important;
+*:focus {
+    outline: 1px solid transparent !important;
 }
```
microsoftsam: Is this necessary? Setting it to transparent?? Why not stick with whatever focus outline is set by the component itself.

theme/accessibility.less
```diff
@@ -9,9 +9,16 @@
...
+
+@media not all and (pointer:coarse) {
+    *:focus {
+        outline: 2px solid #4D90FE !important;
+    }
```
microsoftsam: Don't do a blanket focus. Make sure we're excluding tabIndex=-1 and others. Check how microsoft.com does it.

theme/accessibility.less
veler: So the reason why is that on Android, when there is no outline defined, a Red outline appear.

There was a solution couple of year ago but it is not supported anymore. The fixed I've found is to put it to transparent. "none" was not working. [https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-tap-highlight-color](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-tap-highlight-color)

microsoftsam:
All makecode.com files need to be based of master.

PR #2821, opened 2017-08-26
Initial service worker work

microsoftsam:
This PR is by far not ready. 

Beginning initial service worker work to support better caching of editor and doc pages for browsers that support it.
Right now, roughly 63% of our users are on a browser that supports service workers. 

Service workers allow us to do better than just cache static files (like application cache), it allows us to cache page requests, images, and perhaps do more complicated caching with blocks / packages for a better offline scenario.

We will still keep the appcache as a fallback.

Some resources RE service workers: 
- Udacity course: https://classroom.udacity.com/courses/ud899
- https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook
- https://developers.google.com/web/fundamentals/getting-started/primers/service-workers

### DIFF
--- a/docfiles/docs.js
+++ b/docfiles/docs.js
@@ -184,6 +184,6 @@ function registerServiceWorker() {
 }
 
 $(document).ready(function () {
-    registerServiceWorker();
+    //registerServiceWorker();
     renderSnippets();
 });
--- a/docfiles/docs.js
+++ b/docfiles/docs.js
@@ -166,6 +166,24 @@ function renderSnippets() {
     });
 }
 
+function registerServiceWorker() {
+    if (pxt.Cloud.isLocalhost()) return;
+    if ('serviceWorker' in navigator) {
+        window.addEventListener('load', function() {
+            navigator.serviceWorker.register('/sw.js').then(function(registration) {
+                // Registration was successful
+                console.log('ServiceWorker registration successful with scope: ', registration.scope);
+                // Abort app cache if service worker is loaded
+                window.applicationCache.abort();
+            }, function(err) {
+                // registration failed :(
+                console.log('ServiceWorker registration failed: ', err);
+            });
+        });
+    }
+}
+
 $(document).ready(function () {
+    registerServiceWorker();
     renderSnippets();
 });
--- a/webapp/public/serviceworker.js
+++ b/webapp/public/serviceworker.js
@@ -1,6 +1,6 @@
-var CACHE_NAME = 'pxt-cache-v2';
+// Making any chance to the service worker reloads a new one.
 
-
+var CACHE_NAME = 'pxt-cache-v1';
 
 // keep in sync with release.manifest
 var urlsToCache = [
@@ -17,6 +17,7 @@ var urlsToCache = [
   "https://az742082.vo.msecnd.net/pub/psopafpj",
 
   // editor
+  "/index.html",
   "/blb/semantic.js",
   "/blb/main.js",
   "/blb/typescript.js",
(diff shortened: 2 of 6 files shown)


webapp/public/serviceworker.js
```diff
@@ -0,0 +1,149 @@
...
+      event.waitUntil(self.clients.claim());
+    })
+  );
+});
+
+self.addEventListener('fetch', function(event) {
+  event.respondWith(
+    (event.request.url.indexOf('/api/') >= 0) ?
+      fetch(event.request)
+    : caches.match(event.request).then(function (response) {
```
pelikhan: How are the cache entries invalidated?

docfiles/docs.js
```diff
@@ -166,6 +166,24 @@ function renderSnippets() {
     });
 }
 
+function registerServiceWorker() {
+    if (pxt.Cloud.isLocalhost()) return;
+    if ('serviceWorker' in navigator) {
+        window.addEventListener('load', function() {
+            navigator.serviceWorker.register('/sw.js').then(function(registration) {
```
pelikhan: serviceworker.js ?

webapp/public/serviceworker.js
microsoftsam: I think this current cache strategy is no good. 
This is my proposal: 
- static files (index.html, JS and CSS files) are cached on install (like above). 
- doc files in pxtarget menu options (projects, examples), are also cached on install
- docs have a button that is "Available offline" and adds the doc page to the cache
- API calls are never cached and are network only
- App insights is always network only
- General strategy is: for JS / CSS files, query cache first then network 
- For all other files, do a cache and network race.

@mmoskal @pelikhan thoughts?

abchatra:
What is the rationale for this to be in v0?

pelikhan:
https://developers.google.com/web/tools/lighthouse/audits/appcache

abchatra:
I am still not convinced that this needs to change in v0. Our offline support works great with app cache. We need to retest all the scenarios with service workers now. 

I have pinged standards folks on the final word on how long it takes to remove application cache from the standards. If it is going to several years we should stick with our strategy v0.

pelikhan:
Agreed. It'll take a long time. Let's move this to master.

microsoftsam:
This PR is mostly for discussing cache strategies and figuring out what we want the web worker to do. Started this work in v0, just in case we wanted to have it there. Happy to move the official version of this PR to master, for now let's keep this for discussion.

webapp/public/serviceworker.js
mmoskal: Everything that sits on the CDN is uniquely determined by name. The entries have the form `https://pxt.azureedge.net/blob/<SHA>/<PATH>` or `https://pxt.azureedge.net/commit/<SHA>/<PATH>`. I think, we could just use `<PATH>` to see which entries to invalidate - when we get a request for `https://pxt.azureedge.net/blob/abc...def/foo/bar` and we already have `https://pxt.azureedge.net/blob/123...456/foo/bar` then we remove the existing one and add the new one.

mmoskal:
Everything that sits on the CDN is uniquely determined by name. The entries have the form `https://pxt.azureedge.net/blob/<SHA>/<PATH>` or `https://pxt.azureedge.net/commit/<SHA>/<PATH>`. I think, we could just use `<PATH>` to see which entries to invalidate - when we get a request for `https://pxt.azureedge.net/blob/abc...def/foo/bar` and we already have `https://pxt.azureedge.net/blob/123...456/foo/bar` then we remove the existing one and add the new one.

Maybe it's not necessary to list (most of) the files in the service worker - just cache them, as we load the app. I guess this won't work for Monaco since we load it on demand? Maybe the service worker can induce the app to load Monaco (and all other optional components) when the app is first cached?

Also, I guess you could just download the app manifest and get the list of files from there if that would simplify things. Maybe it wouldn't.

microsoftsam:
Closing this, in hope to tackle this someday soon.

PR #2822, opened 2017-08-27
Add custom pathnames for apps and warning banner

guillaumejenkins:
**Add ?w=1 to the github URL to ignore whitespace changes in this PR**

Adds navigation to custom pathnames. Needed to test /beta and other pathnames (/acc, /newblocks, etc) inside released apps (electron, uwp, etc).

Also displays a warning banner when this navigation is used, as a fail-safe for users accidentally navigating:

image


microsoftsam:
UI looks good. Can you not do a Ctrl shift f, at least until we merge the accessibility branch. It makes merging so much harder..

PR #2823, opened 2017-08-27
Editor load telemetry with actual window width and height

microsoftsam:
Fix editor load telemetry to use actual window width and height.

screen.width = width of the screen
screen.availWidth = width of the screen - scrollbars / toolbars
window.innerWidth = width of the browser


pelikhan:
Can you rename to "screen.... to use the name prefix as the other measures?

Issue #2803, opened 2017-08-24
[Screen Reader - Makecode-Documentation] Narrator focus is navigating to hidden element in scan mode and reading it as 'image'


v-mesala:
**User Experience:**
If Narrator in scan mode is navigating to hidden element and reading as image then user with visual disabilities may get misguided to understand the behavior of the controls

**Test Environment:**
OS: RS2 Version 1703(OS Build 15063.483)
Platform: Edge

**Repro Steps**
1-Navigate to url:-https://makecode.microbit.org/device/error-codes
2-Navigate to footer section
3- launch narrator change to scan mode by (caps+ space key)
4-Verify if screen reader moving to any hidden element or not

**Actual Result:**
Narrator in Scan mode is navigating to hidden element and reading as image

**Expected Result:**
In scan mode Narrator focus should not navigate to hidden element

**Note:**
In item mode narrator is navigating to hidden element and reading as image

**MAS Reference**
https://microsoft.sharepoint.com/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={b24201d1-2276-4471-8183-e1e5353031ad}

**Refer Attachment:**
[MAS4.2.10_Footer_FocusMovingHiddenElement.zip](https://github.com/Microsoft/pxt/files/1248729/MAS4.2.10_Footer_FocusMovingHiddenElement.zip)

Latest Video:
[Narrator reads hidden element image.zip](https://github.com/Microsoft/pxt/files/1298769/Narrator.reads.hidden.element.image.zip)

veler:
Can't reproduce this issue. Doing those keyboard combination doesn't work. I'm on Windows 10 RS2.

v-mesala:
We have verified the issue OS: RS2 Version 1703(OS Build 15063.483)Win 10, and the issue still repro in current environment, Launch narrator and please use Caps+ space key for scan mode, and navigate by up\down key. Please refer to the video to get more clear view about the issue.

v-mesala:
Verified on URL:https://makecode.microbit.org/device/error-codes and issue is not fixed now. Hence reactivating the bug.

veler:
I cannot reproduce the issue in the footer anymore. It focus on the footer and describes it as the footer (which is normal and was not the case before), but does not have a hidden narrator focus anymore after the Microsoft logo.
Can you please take a new video?

@abchatra 

[...]

PR #2830, opened 2017-08-28
[Accessibility] Fixed the focus outline of makecode.com

veler:
* Fixed the focus outline of makecode.com
* Fixed an issue where the narrator was outlining a hidden element in the footer.

Related issues : [https://github.com/Microsoft/pxt/issues/2604](https://github.com/Microsoft/pxt/issues/2604), [https://github.com/Microsoft/pxt/issues/2803](https://github.com/Microsoft/pxt/issues/2803)

capture

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -17,7 +17,7 @@
 				outline: auto !important; /* initial or inherit has no render on Chrome */
 			}
 
-			.inverted:not(#makecode-resources-container) .ui.card:not([tabindex='-1']):focus,
+			.ui.inverted:not(#makecode-resources-container) .ui.card:not([tabindex='-1']):focus,
 			.ui.button.green:focus {
 				outline: 2px solid white !important;
 			}
--- a/docs/index.html
+++ b/docs/index.html
@@ -12,11 +12,6 @@
 	<link rel="stylesheet" href="https://static2.sharepointonline.com/files/fabric/office-ui-fabric-core/4.1.0/css/fabric.min.css">
 
 	<style>
-
-		*:focus {
-			outline: 1px solid transparent !important;
-		}
-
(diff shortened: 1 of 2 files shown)


docs/index.html
```diff
@@ -12,6 +12,17 @@
 	<link rel="stylesheet" href="https://static2.sharepointonline.com/files/fabric/office-ui-fabric-core/4.1.0/css/fabric.min.css">
 
 	<style>
+
+		*:focus {
+			outline: 1px solid transparent !important;
+		}
```
microsoftsam: Let's just use browser outline for makecode.com

docs/index.html
```diff
@@ -12,6 +12,12 @@
 	<link rel="stylesheet" href="https://static2.sharepointonline.com/files/fabric/office-ui-fabric-core/4.1.0/css/fabric.min.css">
 
 	<style>
+		@media not all and (pointer:coarse) {
+			*:not([tabindex='-1']):focus {
+				outline: 2px solid #4D90FE !important;
+			}
+		}
```
microsoftsam: Why not stick with browser outline??

veler: I know that we chosen this color because of the feedbacks we had from micro:bit, but I feel it is the occasion to keep the same outline for all targets and our own website.
Would you prefer the native outline anyway?

microsoftsam: Yes native for the website.

veler:
So with the changes I did, Edge, IE and FireFox use the dotted focus, and Chrome select a color automatically depending of the background. Setting initial or inherit to the outline display nothing on Chrome.

As Chrome was selected color that were not visible enough for me, I personalized them for the cards and button, so it's visible with or without the high contrast mode.

untitled

PR #2839, opened 2017-08-29
Adding support multiple session data collection

abchatra:
This enables support for multiple session data collection. This should facilitate
- Correlating the usage of MakeCode website across multiple sessions
- Understanding progression path from blockly to typescript. 
- Amount of time spent in MakeCode editor in a typical classroom environment

Todo:
- [ ] validate that consent was given before enable multi-session logging
- [ ] display cookie banner in docs/html pages
- [ ] display cookie banner in App (and use the MSCC cookie)


abchatra:
- I will test out clear, I am not sure it clears cookie today.
- Local storage is to save the unsent telemetry in offline case. We need both.
- Good point I will work on the Cookie banner for doc pages.

pelikhan:
This latest changes uses the Microsoft cookie service to determine if a consent UI has to be displayed to the user. Hooked up for the docs/landing page, needs work in the app.
This will replace the current cookie disclaimer dialog.

PR #2840, opened 2017-08-29
[Accessibility] Fixed a set of minor issues mainly related to focus.

veler:
- Fixed an issue where the focus was not given to the JavaScript editor when this one was already displayed.
- Fixed an issue where the focus was not being reset after a specific action.
- Fixed an issue where the focus outline was not displayed in the contextual menu.
- Fixed an issue where the color of the notifications was not reset after leaving the high contrast mode.
- Fixed an issue in Rename Your Project where the color of the Save button was not green anymore.
- Fixed an issue where the search button's border in Add Package was not round anymore.
- Removed the focus outline in a dropdown menu in Blockly.

### DIFF
--- a/theme/accessibility.less
+++ b/theme/accessibility.less
@@ -14,7 +14,8 @@
 }
 
 @media not all and (pointer:coarse) {
-    *:not([tabindex='-1']):focus {
+    *:not([tabindex='-1']):not(.blocklyGridPickerMenu):focus,
+    [role='menuitem']:focus {
         outline: 2px solid #4D90FE !important;
     }
 }
--- a/theme/accessibility.less
+++ b/theme/accessibility.less
@@ -14,9 +14,15 @@
 }
 
 @media not all and (pointer:coarse) {
-    *:not([tabindex='-1']):not(.blocklyWidgetDiv):not(.blocklySVG):focus,
+    *:not([tabindex='-1']):focus,
     [role='menuitem']:focus {
         outline: 2px solid #4D90FE !important;
+    }
+
+    .blocklyWidgetDiv *:focus,
+    .blocklySVG *:focus,
+    .monaco-editor *:focus {
+        outline: 1px solid transparent !important;
     }
 }
 
--- a/theme/accessibility.less
+++ b/theme/accessibility.less
@@ -14,7 +14,7 @@
 }
 
 @media not all and (pointer:coarse) {
-    *:not([tabindex='-1']):not(.blocklyGridPickerMenu):focus,
+    *:not([tabindex='-1']):not(.blocklyWidgetDiv):not(.blocklySVG):focus,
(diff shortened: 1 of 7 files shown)


theme/accessibility.less
```diff
@@ -14,7 +14,8 @@
 }
 
 @media not all and (pointer:coarse) {
-    *:not([tabindex='-1']):focus {
+    *:not([tabindex='-1']):not(.blocklyGridPickerMenu):focus,
+    [role='menuitem']:focus {
```
microsoftsam: not just the blocklyGridpickerMenu but all blockly elements. Is there any way to exclude all elements under blocklySVG. We probably are going to want to exclude the monaco-editor too. Perhaps we can instead do this only on semantic ui classes.

webapp/src/core.ts
```diff
@@ -485,6 +485,14 @@ export function scrollIntoView(item: JQuery, margin = 0) {
     }
 }
 
+export function resetFocus() {
+    let accessibleMenu = document.getElementById('accessibleMenu');
+    accessibleMenu.tabIndex = 0;
+    accessibleMenu.focus();
+    accessibleMenu.blur();
+    accessibleMenu.tabIndex = -1;
```
microsoftsam: This doesn't look right? why are we reseting the focus on the accessibleMenu? why not the document / root?

veler: So it's weird but resetting it to the body didn't worked for me on Edge. It was still focusing the simulator.
Also, just blurring the element or giving the focus to an element that doesn't had a tabindex was not working on Edge and Chrome.

veler: I can do it on the element with the 'content' id if you prefer.

theme/accessibility.less
```diff
@@ -14,9 +14,16 @@
...
-    *:not([tabindex='-1']):focus {
+    *:not([tabindex='-1']):focus,
+    [role='menuitem']:focus {
         outline: 2px solid #4D90FE !important;
     }
+
+    .blocklyWidgetDiv *:focus,
+    .blocklySVG *:focus,
+    .monaco-editor *:focus {
+        outline: 1px solid transparent !important;
```
microsoftsam: This won't use the default outline instead. You want to try and exclude these divs from your outline set above. Not the other way around.

microsoftsam:
Approving so we can do some testing, but I'll need you to revert the focus change.

theme/accessibility.less
veler: It's the first thing I tried but the ``:not`` selector doesn't let me exclude the children elements of a specific class, only the class itself.

microsoftsam: Then instead of using * but not, try just including the ones you want added. All semantic ui classes start with .ui 
that could be the solution

PR #2845, opened 2017-08-30
[Accessibility] Fixed a set of issues detected during the team test.

veler:
- Fixed an issue where the focus outline was not displayed anymore on IE.
- Fixed the padding of the JavaScript editor in tutorial mode.
- Fixed an issue where the user was not able to do a Shift+Tab after clicking the OK button of a step of the tutorial very quickly after it was displayed.
- Fixed an issue where the Enter key was not supported on IE in the tutorial.

### DIFF
--- a/theme/accessibility.less
+++ b/theme/accessibility.less
@@ -8,23 +8,26 @@
 /*******************************
           Accessibility
 *******************************/
-
-*:focus {
-    outline: 1px solid transparent !important;
+*[tabindex='0']:focus,
+*[tabindex*='d1']:focus,
+*[tabindex*='d2']:focus, /* Takes items with a defined tabIndex from 0 to 29. */
+*[role='menuitem']:focus {
+    outline: 2px solid #4D90FE !important;
 }
 
-@media not all and (pointer:coarse), /* If not touch screen */
-all and (-ms-high-contrast: none), (-ms-high-contrast: active) { /* Or if IE 10+ */
-    *:not([tabindex='-1']):focus,
-    [role='menuitem']:focus {
-        outline: 2px solid #4D90FE !important;
-    }
-
-    .blocklyWidgetDiv *:focus,
-    .blocklySVG *:focus,
-    .monaco-editor *:focus {
+@media all and (pointer:coarse) { /* If touch screen */
+    *[tabindex='0']:focus,
+    *[tabindex*='d1']:focus,
+    *[tabindex*='d2']:focus,
+    *[role='menuitem']:focus {
         outline: 1px solid transparent !important;
     }
+}
+
+.blocklyWidgetDiv *:focus,
+.blocklySVG *:focus,
+.monaco-editor *:focus {
+    outline: 1px solid transparent !important;
 }
 
 .blocklyTreeRoot:focus {
--- a/theme/accessibility.less
+++ b/theme/accessibility.less
@@ -13,7 +13,8 @@
     outline: 1px solid transparent !important;
 }
 
-@media not all and (pointer:coarse) {
+@media not all and (pointer:coarse), /* If not touch screen */
+all and (-ms-high-contrast: none), (-ms-high-contrast: active) { /* Or if IE 10+ */
     *:not([tabindex='-1']):focus,
     [role='menuitem']:focus {
         outline: 2px solid #4D90FE !important;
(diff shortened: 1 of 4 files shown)


theme/accessibility.less
```diff
@@ -13,7 +13,8 @@
     outline: 1px solid transparent !important;
 }
 
-@media not all and (pointer:coarse) {
+@media not all and (pointer:coarse), /* If not touch screen */
+all and (-ms-high-contrast: none), (-ms-high-contrast: active) { /* Or if IE 10+ */
```
microsoftsam: Tested on Firefox? According to this, pointer:course is not supported on Firefox. I would also test Chrome on IOS, I believe that doesn't work on it either. 
http://caniuse.com/#search=pointer%3Acoarse

webapp/src/tutorial.tsx
```diff
@@ -286,7 +291,7 @@ export class TutorialComplete extends data.Component<ISettingsProps, TutorialCom
                 closeOnDimmerClick closeOnDocumentClick
                 >
                 <div className="ui two stackable cards">
-                    <div className="ui grid centered link card focused" aria-selected="true" aria-label={lf("More Tutorials")} tabIndex={0} onClick={() => this.moreTutorials() } onKeyPress={sui.fireClickOnEnter}>
+                    <div className="ui grid centered link card focused" aria-selected="true" aria-label={lf("More Tutorials")} tabIndex={0} onClick={() => this.moreTutorials() } onKeyDown={sui.fireClickOnEnter}>
```
microsoftsam: Any other instances of this change? Could you verify there aren't any other places where we should be using onKeyDown instead of onKeyPress.

microsoftsam: Please also document in your dev docs, when to use which.

veler: Yes, there was another instance in webapp/src/app.tsx that is already included to this PR. I checked if any other onKeyPress was existing.

It is also already documented in my PR related to dev documentation : https://github.com/Microsoft/pxt/pull/2853/files#diff-32b935a28c228bd912fbdb42b89f1548 .
I quote : _"You will also have to decide on which keyboard event to use: ``KeyDown``, ``KeyUp`` or ``KeyPress``. When possible, always use the ``KeyDown`` event as ``KeyPress`` has compatibility problems with Internet Explorer. "_

theme/accessibility.less
veler: I'm investigating a better solution without breaking IE (which is tricky).

veler:
Last commit : 
- Fixed an issue where the focus outline was not appearing on FireFox.
- Fixed an issue where the focus outline was appearing on IE on element with no tabIndex defined.
- The outline disappear on Chrome for Android but not FireFox, like on microsoft.com in fact.

Tested on Edge, IE, Chrome and FireFox on Windows. Tested on Chrome and Firefox on Android. Tested on Edge on SurfaceBook in tablet mode.

Previously, we were applying to elements that has a focus different from -1 on screen that are not touch. But first, on Firefox on Desktop, the check was not working, there was no focus.
And on IE, when we were clicking on a non-interactive element, the focus was appear. The reason why is that IE consider elements that has no defined tabIndex as 0 instead of -1, even if we still cannot navigation to them with the keyboard.

To fix the issue, I apply now the focus to every element that has a "defined" (weirdly this works on IE) tab index between 0 and 29 (I think 29 is enough but we can go upper if needed), and remove the focus outline on screen that are detected as touch screen. This has the effect to remove the outline on Chrome for Android, but not on Edge and FireFox on Android and in Tablet Mode. I observe the same behavior with Microsoft.com on FireFox and Edge with a touch screen.

theme/accessibility.less
```diff
@@ -8,24 +8,28 @@
...
           Accessibility
 *******************************/
-
-*:focus {
-    outline: 1px solid transparent !important;
+*[tabindex='0']:focus,
+*[tabindex*='d1']:focus,
+*[tabindex*='d2']:focus, /* Takes items with a defined tabIndex from 0 to 29. */
+*[role='menuitem']:focus {
+    outline: 2px solid #4D90FE !important;
```
microsoftsam: Does this affect Blockly or Monaco?

veler: No it doesn't. Below, there is a CSS block that remove the focus outline on Blockly and Monaco.

PR #2875, opened 2017-09-08
[Accessibility] Fixed some CSS issues linked to the focus outline (again...)

veler:
- Removed the focus outline from the blockly text input.
- Added the focus outline to some text input and textarea that lost it.
- Refactored the css.

I changed the "outline: none !important; " to "outline: 1px solid transparent !important; " because on Edge and IE, "none" is considered as the default focus render (which is the dotted grey outline).


pelikhan:
Could you add screenshots?

veler:
Sure but I'm not sure it's very relevant.

Previously, there was a blue outline on the text input of the blocks (I removed it). And the blue outline of the project name of textarea in the share was not there (but it should be there).

This is how it looks now with this PR :

untitled

microsoftsam:
Apply change to your master PR.

PR #2883, opened 2017-09-08
Fixing decompilation of functions as expressions

### LINKED ISSUE #2854: Blockly compiler breaks when calling a function with a return statement
The following breaks blockly compilation: 

```
loops.forever(function () {
    let test = getStuff();
})
function getStuff(): string {
    return "Unknown";
}
```

It's because the function has a return type and we don't currently support that. The above should be a gray block.

riknoll:
Fixes #2854


microsoftsam:
Mind explaining what the fix is?

riknoll:
@microsoftsam the important part of the fix is [here](https://github.com/Microsoft/pxt/pull/2883/commits/1d95b7abf41c7ce280a68f6568da0c76fa26fd29#diff-c69b7c3dbaf11a400a8562549d7cccffR1714).

PR #2889, opened 2017-09-08
Remove block count from delete blocks context menu

dragonx:
How do you guys feel about changing this text?  The block count appears to confuse many users.  i.e., it counts the "on start" block, but we don't delete that block.


microsoftsam:
I believe I've already fixed this. Please double check. 
Note: you should still see a count if you click on a parent block like the forever or event block, as deleting that will delete the blocks inside of it, so it's good to give the user an indication of that.

microsoftsam:
I've added this bug here: https://github.com/Microsoft/pxt-blockly/issues/50

Looks like we should be excluding shadow blocks from the count in the workspace clear all count.

pelikhan:
I like the new phrasing

PR #2891, opened 2017-09-08
Don't decompile if there are TS errors

### LINKED ISSUE #2890: foo compiles back to Blocks
The following code snippet decompiles back to Blocks when it shouldn't:

```
foo
```

Higher pri as a bad code snippet can be converted back to Blocks

riknoll:
Fixes #2890 
The decompiler passes different compile options than the compiler does (it sets the AST flag to true). That was causing some bugs where if there was an error that the TS compiler caught but our compiler didn't, we would still decompile even if the code was bad. This change doesn't alter the options that are passed, but it does make it so that TS compiler diagnostics are always surfaced.


riknoll:
Well, apparently this breaks all the decompiler tests so back to the drawing board

riknoll:
@pelikhan  take a look at the latest commit. I had to do a hack to fix the decompiler test cases but I think it's a pretty safe one.

PR #2893, opened 2017-09-08
Add 'null' and 'undefined' to JS types doc.

ganicke:
'null' and 'undefined' aren't allowed as types so only a small discussion about setting as values.


ganicke:
'null' and 'undefined' aren't allowed as types so only a small discussion about setting as values.

PR #2910, opened 2017-09-11
no hint bubble in tutorial

pelikhan:
Don't show the "hint" image when there is no hint


microsoftsam:
Cherry-pick change to master after?

pelikhan:
@abchatra  ping

PR #2918, opened 2017-09-12
Hide share embed

pelikhan:


abchatra:
@pelikhan can you explain the intent of the change?

pelikhan:
Adds a theme option to hide all advanced "embed" sharing options. Merged from v2.

abchatra:
Why is this getting merged to v0. I don't understand

pelikhan:
Keeping v0 in sync.

abchatra:
Lets keep the churn low in v0 and only merge fixes we need? Sync is not a priority right?

Issue #2795, opened 2017-08-24
[Keyboard Navigation-Project-Makecode Documentation]: Focus is moving to some hidden link showing on left panel on Project documentation page.


v-mesala:
**User Experience:**
User who depends on Keyboard with screen reader may get misguided about the links and controls if focus moves to any hidden link on Project Page.

**Test Environment:**
OS: RS2 Version 1703(OS Build 15063.2)
Platform: Edge.

**Repro Steps:-**
1. Navigate to https://makecode.microbit.org/projects
2. Navigate to left panel controls and verify that focus is navigating to hidden link.

**Actual Result:-**
While navigating the page using tab key Focus is moving to some hidden link showing on left panel on Project page.

**Expected Result:-**
While navigating the page using tab key Focus should not move to any hidden link showing on left panel on Project page

**Note:** 
1-Focus also navigates to hidden links after USB and Streaming links under Hardware in the left panel.
2-This issue exists with Firefox also.

**MAS Reference -**
https://microsoft.sharepoint.com/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={0de7fbe1-ad7e-48e5-bcbb-8d986691e2b9}

**Suggested fix:-**
https://microsoft.sharepoint.com/teams/msenable/mas/pages/fixit-details.aspx?fixitid=10 
 

**Please refer the Attachment:**
[MAS2.4.3_Make code_Projects_Focus hidden links.zip](https://github.com/Microsoft/pxt/files/1248368/MAS2.4.3_Make.code_Projects_Focus.hidden.links.zip)

veler:
Issue duplicated with [https://github.com/Microsoft/pxt/issues/2649](https://github.com/Microsoft/pxt/issues/2649)

veler:
Fix available on [https://makecode.microbit.org/](https://makecode.microbit.org/)

v-mesala:
Verified on URL: https://makecode.microbit.org/projects and issue is not fixed now. Hence reactivating the bug

v-mesala:
Verified on URL: https://makecode.microbit.org/projects and issue is not fixed now. Hence reactivating the bug. Attached latest video
[Issue is not fixed.zip](https://github.com/Microsoft/pxt/files/1320122/Issue.is.not.fixed.zip)

samelhusseini:
Please verify again. issue should be fixed.

sonimishra:
[...]

PR #2923, opened 2017-09-12
Fixed an issue where the focus was moving to some hidden link showing…

veler:
Fixed an issue where the focus was moving to some hidden link showing on left panel on documentation page.

Version for Master : [https://github.com/Microsoft/pxt/pull/2938](https://github.com/Microsoft/pxt/pull/2938)

Related issue : [https://github.com/Microsoft/pxt/issues/2795](https://github.com/Microsoft/pxt/issues/2795)

capture

I managed to support those special cases where there is no link for the header of a menu, but that the markdown parser was still generating a href tag.

capture

### DIFF
--- a/docfiles/macros.html
+++ b/docfiles/macros.html
@@ -112,7 +112,7 @@
     </div>
 </aside>
 
-<aside id=top-dropdown-noHeading class=toc>
+<aside id=top-dropdown-noheading class=toc>
     <div class="item" title="@TITLE@">
         <div class="menu">
             @ITEMS@
--- a/docfiles/macros.html
+++ b/docfiles/macros.html
@@ -112,6 +112,14 @@
     </div>
 </aside>
 
+<aside id=top-dropdown-noHeading class=toc>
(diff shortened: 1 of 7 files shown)


pelikhan:
Merged in master as well?

docfiles/macros.html
```diff
@@ -112,6 +112,14 @@
     </div>
 </aside>
 
+<aside id=top-dropdown-noHeading class=toc>
```
pelikhan: i think we use lower case for css classes

veler:
@pelikhan , I will need to cherry pick to master once approved.

microsoftsam:
@pelikhan will you apply these changes to pxt-backend once we merge?

pelikhan:
Yes

microsoftsam:
@veler create another PR just like this based on master, so we can merge both at the same time.

veler:
PR for master created : [https://github.com/Microsoft/pxt/pull/2938](https://github.com/Microsoft/pxt/pull/2938)

PR #2925, opened 2017-09-12
Don't decompile if a file is too large

riknoll:
Limit the decompilation to 1000 blocks and bail out if we hit that limit.

too_large


pxtcompiler/emitter/decompiler.ts
```diff
@@ -343,6 +372,40 @@ ${output}</xml>`;
             result.success = false;
         }
 
+        function countBlock() {
+            emittedBlocks ++;
+            if (emittedBlocks > MAX_BLOCKS) {
+                let e = new Error(Util.lf("Could not decompile because the script is too large"));
```
pelikhan: Maybe just "There are too many blocks!"

riknoll: Actually, the user won't see that text. I'll remove the call to lf

pxtcompiler/emitter/decompiler.ts
```diff
@@ -975,45 +1011,35 @@ ${output}</xml>`;
...
-                        inputs: [{
-                            kind: "value",
-                            name: "VALUE",
-                            value: negateNumericNode(n.right),
-                            shadowType: numberType
-                        }],
-                        fields: [getField("VAR", getVariableName(n.left as ts.Identifier))]
-                    };
+                    const r = mkStmt("variables_change");
+                    countBlock();
```
pelikhan: why extra countBlock?

webapp/src/monaco.tsx
```diff
@@ -138,7 +138,9 @@ export class Editor extends srceditor.Editor {
                         .then(resp => {
                             if (!resp.success) {
                                 this.currFile.diagnostics = resp.diagnostics;
-                                return failedAsync(blockFile);
+                                let tooLarge = false;
+                                resp.diagnostics.forEach(d => tooLarge = (tooLarge || d.code === 9266 /* error code when script is too large */));
```
pelikhan: define constant for 9266?

PR #2927, opened 2017-09-12
Catch block errors

### LINKED ISSUE #2899: Blocks editor should handle large/small numbers gracefully
Currently, entering a too number too big gets reported as an error in the backend, we should handle this case more gracefully.

riknoll:
Fixes #2899 
This should catch all of those annoying too big or too small errors.


abchatra:
Please merge to v0 as well

PR #2929, opened 2017-09-12
Check docs using blocksprj not core

riknoll:
Was causing all of the docs to fail in Adafruit.


pelikhan:
There may be cases where this would be problematic: for example, blocksprj in microbit imports radio which is incompatible with the bluetooth packages. However, we haven't had such case in the CODAL system.

Issue #2535, opened 2017-07-27
[Visual Requirement-Help-Getting Started]: Luminosity ratio is less than 4.5:1 for the text showing in Orange color and Ok button .


v-swasha:
**User Experience:**
Users who depends on Keyboard with Screen reader will not be able to access the any Textual content if its luminosity ratio is less than 4.5:1.

**Test Environment:**
OS: RS2 Version 1703(OS Build 15063.483)
Platform: Edge.
Tool Used: Color Contrast Analyzer

**Repro Steps:-**
1. Navigate to https://makecode.microbit.org/acc
2. Navigate to the Microbit section element and select Code control given on it.
3. Navigate to the Help control lying in the header section on the page and select it.
4. Navigate to 'Getting started' control and select it.
5. Verify the Luminosity ratio for the all the text showing on orange color using CCA tools.

**Actual Result:-**
Luminosity ratio is less than 4.5:1 for the text showing in Orange color and Ok button on Getting started page.

**Expected Result:-**
Luminosity ratio is less than 4.5:1 for the text showing in Orange color and Ok button on Getting started page.

**Note:** This issue exist throughout app for text showing in Orange color

**MAS Reference -**
https://microsoft.sharepoint.com/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={a73546c7-765f-489c-b18f-afb659fe99e6}

**Suggested Fix:**
1.1-Ensuring that a contrast ratio of at least 4.5:1 exists between text (and images of text) and background behind the text 
1.2- If design allows , Not specify background color, not specify text color, and not using technology features that change those defaults .In this technique the author avoids having to do any contrast measures by simply not specifying the text color and not specifying the background. As a result the colors of both are completely determined by the user agent.
Tips:-
•The easiest fix is to increase the brightness of light colors and decrease the brightness of darker colors. 
•Exaggerate lightness differences between foreground and background colors, and avoid using colors of similar brightness adjacent to one another, even if they differ in saturation or hue. 
•During design Use the Color Contrast Analyzer to determine whether new colors have color contrast ratio of at least 4.5:1. 
 
**Key References for Fix:** 
http://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html
 
**Please refer the attchment:**
![mas1 4 3_help_get started](https://user-images.githubusercontent.com/30459830/28669507-215531a0-72f2-11e7-94ff-c7b3fdb705ea.jpg)

[...]

PR #2931, opened 2017-09-12
[Accessibility] Fixed the color contrast of the tutorial

veler:
Fixed the Orange color contrast of the tutorial by using the Microsoft color.

Related issue : [https://github.com/Microsoft/pxt/issues/2535](https://github.com/Microsoft/pxt/issues/2535)

untitled

### DIFF
--- a/theme/themes/pxt/globals/site.variables
+++ b/theme/themes/pxt/globals/site.variables
@@ -18,7 +18,6 @@
 @grey:#505050;
 
 @greenTextColor: #107C10;
-@microsoftorange:#D83B01;
 
 @lightTextColor              : rgba(0, 0, 0, 0.68);
 @invertedUnselectedTextColor : rgba(255, 255, 255, 0.8); 
--- a/theme/themes/pxt/globals/site.variables
+++ b/theme/themes/pxt/globals/site.variables
@@ -17,7 +17,7 @@
 @brown:#00BCF2;
 @grey:#505050;
 
-@greenTextColor: #107C10;
+@greenTextColor: darken(@green, 5);
 
 @lightTextColor              : rgba(0, 0, 0, 0.68);
 @invertedUnselectedTextColor : rgba(255, 255, 255, 0.8); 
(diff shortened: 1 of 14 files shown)


pelikhan:
It's very very red now. Isn't that too scary?

abchatra:
Can we try a different color?

microsoftsam:
This PR description doesn't make sense. You're updating the block color, yet you're showing the color change for the getting started button. The two should be independent.

veler:
Hum, in fact, this change affect everything in orange : the block, the buttons (Getting started, Finish) and the orange top bar of the tutorial.
Do you want me to do the change only on the blocks? It would makes sense to me to change it on the buttons too as there is the same color contrast issue than on the block even if the accessibility team don't talk about the buttons explicitly.

microsoftsam:
The issue in question here says the "Ok" and the text in orange in the getting started experience, ie the blocks are the issue here. They use an orange color that is non compliant given the text size and the orange color chosen.
You've already fixed the "Ok" button in a separate change, so I think this change should only fix the block text orange to use the MS color in all targets.
We are not able to change the micro:bit orange without Foundation consolation, let's discuss offline if that is necessary.

veler:
I changed the color of the non-categorized inline block to use the MS orange color. I closed the PR related to micro:bit [https://github.com/Microsoft/pxt-microbit/pull/539](https://github.com/Microsoft/pxt-microbit/pull/539).
I updated the screenshot.

theme/themes/pxt/globals/site.variables
```diff
@@ -18,6 +18,7 @@
 @grey:#505050;
 
 @greenTextColor: #107C10;
+@microsoftorange:#D83B01;
```
microsoftsam: No need to specify here. use the color directly. or rename to something more descriptive of it's use and place it at the top of tutorial.less

veler:
TODO once merged : Do the same PR for Master, or cherry-pick to accessibility-master.

theme/tutorial.less
```diff
@@ -5,6 +5,8 @@
 /* Reference import */
 @import (reference) "semantic.less";
 
+@inlineblockmicrosoftorange: #D83B01; // The official Microsoft orange color.
```
microsoftsam: change to inlineBlockColor..

microsoftsam: No need for comment.

PR #2937, opened 2017-09-13
Inverted close button

pelikhan:
When invertedMenu is true, add "inverted" class on modal close button.
image


microsoftsam:
Is this still needed?

pelikhan:
Still think that the button is bused but

Issue #2385, opened 2017-07-02
Pause in "on start" does not delay "forever" loops?


microbit-carlos:
microbit.org-ticket: 2187

Aren't the "forever" loops meant to start running after the "on start" blocks have finished?

If we need to initialise something before it can be used (like neopixels or game sprites), the "on start" is usually the preferred place to do that, but if the "forever" loops are declared before the "on start" code, and then the main fibre sleeps, you could have a race condition like in the following example:

![microbit-screenshot](https://user-images.githubusercontent.com/29712657/27773713-e13ff006-5f77-11e7-8689-3b686d53e81d.png)

This program produces the following error and crashes the emulator:

```
Program Error: r.isRef is not a function
```

Hex file with this minimalist example: [issue.zip](https://github.com/Microsoft/pxt/files/1118125/issue.zip)

guillaumejenkins:
All event handlers and forever loops are registered / started before the `on start` block gets executed. Also, the `on start` block, the `forever` loop and all event handlers are executed in parallel, so pausing execution in `on start` does not affect the execution of `forever` or any event handler.

If you need to delay execution in an event handler or in a forever loop until after something happens in `on start`, a good approach is to use a variable:

![image](https://user-images.githubusercontent.com/14299377/27798740-37685c80-5fc7-11e7-8f6c-fb1a5f753452.png)

abchatra:
Forever is a background fiber. It will run as soon as it gets a chance. Pause in the onstart will give that chance to the forever fiber. This is by design.

microbit-carlos:
Thank you very much for your responses. Adding a flag to indicate "readiness" is certainly a workaround, but it cannot be denied that the current implementation is not really intuitive. The user would expect the application to "pause" for an amount of time, and instead other parts of the program run earlier.

The user has no way to know that when a "pause" block is used within "on start", it will automatically trigger processing of the other fibers (in this case the "for ever" blocks). So they would have no way to know that the would need to create a "ready" flag to avoid this issue. On top of that, the exposed error "Program Error: r.isRef is not a function" is cryptic and gives absolutely no indication of where the issue might lie.

I understand that forever being a background fibre is a design decision,  but is it done "by design" that the pause block on the "on start" will trigger them? This behavior would be very difficult to predict unless the user has a very good knowledge of the internal micro:bit runtime implementation.

abchatra:
[...]

PR #2940, opened 2017-09-13
Add an option to move APIs after on start

riknoll:
The fix discussed in https://github.com/Microsoft/pxt/issues/2385

Needs to be moved to V0 as well.


pelikhan:
Needs cherry-pick in v0

pelikhan:
@ganicke could we update documentation for `on-start`, `forever` to mention the ordering logic?

ganicke:
Ok.

PR #2950, opened 2017-09-15
New Slider

microsoftsam:
Update field slider with better focus and tab support, and a better looking slider, which is more useable on touch.

fieldslider


pelikhan:
Can the label be the parameter name? LGTM

microsoftsam:
Yep it already is, sorry wasn't running the latest code in the GIF.

PR #2958, opened 2017-09-19
Add support for events where the callbacks take parameters

riknoll:
Adds support for event handlers that pass arguments to the callback. There are two modes: one where the block always has all arguments shown and one where the arguments can be dynamically added by clicking a button.

event_parameters

The downside to all this is that in scratch-blocks the variable fields (shown above) look very similar to a variable block, so it's a little unclear that these apply to the callback and aren't inputs to the parent block.


riknoll:
redoing because apparently there isn't a V2 anymore

PR #2959, opened 2017-09-19
Add support in blocks for callbacks with parameters

riknoll:
Redo!

Adds support for event handlers that pass arguments to the callback. There are two modes: one where the block always has all arguments shown and one where the arguments can be dynamically added by clicking a button.

event_parameters

The downside to all this is that in scratch-blocks the variable fields (shown above) look very similar to a variable block, so it's a little unclear that these apply to the callback and aren't inputs to the parent block.


pxtblocks/blocklyloader.ts
```diff
@@ -837,6 +860,68 @@ namespace pxt.blocks {
...
+                for (let j = 0; j < diff; j++) {
+                    const arg = handlerArgs[actuallyVisible - j - 1];
+                    i.removeField("HANDLER_" + arg.name);
+                }
+            }
+
+            actuallyVisible = currentlyVisible;
+        };
+
+        i.appendField(new Blockly.FieldImage("/cdn/blockly/media/add.svg", 24, 24, "*", () => {
```
pelikhan: Is this patched correctly in the cloud?

riknoll: I think I should be using `pxt.webConfig.commitCdnUrl`. I'll fix it

PR #2962, opened 2017-09-19
Minor fixes to the home screen. Passing hero image in pxtarget

microsoftsam:


webapp/src/projects.tsx
```diff
@@ -301,7 +301,7 @@ export class Projects extends data.Component<ProjectsProps, ProjectsState> {
                 </div>
                 {tab == WELCOME ? <div className={tabClasses}>
                     {hasGettingStarted ?
-                        <div className="ui segment getting-started-segment">
+                        <div className="ui segment getting-started-segment" style={{backgroundImage: `url(${pxt.webConfig.commitCdnUrl + encodeURI(targetTheme.homeScreenHero)})`}}>
```
microsoftsam: I'm not sure if this works or not. Testing using uploadTrg didn't work but I think its because uploadTrg doesn't upload the /static dir

@pelikhan does that look right?

PR #2970, opened 2017-09-19
Change to disable application insight cookie

abchatra:
Reverting the change until we have the right cookie banner


abchatra:
Git says it is recommended to check-in

PR #2976, opened 2017-09-19
Publish blockly.d.ts from pxt-blockly instead of using out own

microsoftsam:
Let's keep these files in sync. I've made sure that when we publish pxt-blockly it copies over the blockly.d.ts file into localtypings.


pxtrunner/runner.ts
```diff
@@ -590,8 +590,14 @@ ${files["main.ts"]}
         return initPromise.then(() => pxt.Cloud.downloadMarkdownAsync(tutorialid, editorLocale, pxt.Util.localizeLive))
             .then(tutorialmd => {
                 let steps = tutorialmd.split(/^##[^#].*$/gmi);
+                let newAuthoring = true;
```
pelikhan: sneaked in by accident?

microsoftsam: Yes. Woops

PR #2988, opened 2017-09-20
propagate tutorial title from tutorial menu -- or -- format tutorial id

pelikhan:
Tutorial title was just ignored when creating a new project


microsoftsam:
Looks good.

microsoftsam:
I do want to point out (for reference) that we used to parse the tutorial name from the markdown, but since we've avoided loading the markdown in the editor, this (to me) seems like the best solution.

pelikhan:
Do you need to also apply the fix on the new home page?

PR #3001, opened 2017-09-21
Fix TS (gray) blocks, making them multi-lined

microsoftsam:
screen shot 2017-09-20 at 10 41 56 pm

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -2169,6 +2169,7 @@ namespace pxt.blocks {
                 that.setColour("#717171")
                 that.setPreviousStatement(true);
                 that.setNextStatement(true);
+                that.setInputsInline(false);
 
                 this.domToMutation = (element: Element) => {
                     const n = parseInt(element.getAttribute("numlines"));


pelikhan:
I think it's a mistake to show the entire content of the grey block. It's giberish for the kids anyway. We should just do a single line and ...

There are extra new lines that bloat the size of that block.

microsoftsam:
There are no extra lines.. The new rendering has a larger gap between lines. 
As for how to collapse it. I'm okay coming up with a way to collapse the block. but I wouldn't just do a blatant ellipsis. I think you could collapse functions like: "function test() { ... }" but not some blocks might make sense, if they're just simple function calls with parameters perhaps

pelikhan:
With such new line sizes, these grey blocks are going to be huge.

microsoftsam:
Agreed. For now the it looks much worse on the same line. Let's push this in, and discuss an alternative solution for the gray blocks.

pelikhan:
If you pull in some long piece of code, you'll end up with massive greyblocks. I think it is worse than a single line with a guaranteed reasonable height.

abchatra:
@pelikhan  I don't understand your concern. This is for someone who has written code in javascript and compiling back to blocks. I am fine with showing the same code with the same indentation. Are we adding any extra padding?

pelikhan:
The problem is not passing. It is that Scratch blocks add a huge gap between lines.

PR #3005, opened 2017-09-21
Lots of carousel improvements

riknoll:
* Better animations
* Better click handling
* Touch support
* Removed page length and margin calculations
* Added CSS classes for disabling the arrows (but no CSS)
* Bug fixes


webapp/src/carousel.tsx
```diff
@@ -81,58 +78,82 @@ export class Carousel extends React.Component<ICarouselProps, ICarouselState> {
...
                     event.preventDefault();
                     event.stopPropagation();
                 }
             }
-        });
+        };
 
-        this.dragSurface.addEventListener("mouseleave", event => {
+        let leave = (event: MouseEvent | TouchEvent | PointerEvent) => {
+            console.log("leave");
```
microsoftsam: pxt.log?

PR #3008, opened 2017-09-21
Visibility update only resumes sim if it was running

guillaumejenkins:
Fix: when visibility is updated to true, only run simulator if it was running when visibility became false

`?w=1` in URL to ignore whitespace

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -124,7 +124,10 @@ export class ProjectView
         pxt.debug(`page visibility: ${active}`)
         this.setState({ active: active })
         if (!active) {
-            this.stopSimulator();
+            if (this.state.running) {
+                this.stopSimulator();
+                this.setState({ resumeOnVisibility: true });
+            }
             this.saveFileAsync().done();
         } else {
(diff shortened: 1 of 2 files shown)


webapp/src/app.tsx
```diff
@@ -124,16 +124,21 @@ export class ProjectView
...
+                this.setState({ resumeOnVisibility: true });
+            }
             this.saveFileAsync().done();
         } else {
             if (workspace.isSessionOutdated()) {
                 pxt.debug('workspace changed, reloading...')
                 let id = this.state.header ? this.state.header.id : '';
                 workspace.initAsync()
                     .done(() => id ? this.loadHeaderAsync(workspace.getHeader(id)) : Promise.resolve());
-            } else if (pxt.appTarget.simulator.autoRun && !this.state.running)
```
pelikhan: Missing simulator.autoRun check

guillaumejenkins: Autorun shoudln't matter: if it was running, we resume it, otherwise we don't

Currently if you manually stop the simulator, it resumes when you come back, which is annoying

PR #3011, opened 2017-09-21
Select first block when hit "enter" in gridpicker search bar

caitlin-hennessy:
Demo: no mouse involved! demo

### DIFF
--- a/pxtblocks/fields/field_gridpicker.ts
+++ b/pxtblocks/fields/field_gridpicker.ts
@@ -407,24 +407,6 @@ namespace pxtblockly {
                 }
             });
 
-            // Listen for touch events (why doesn't Closure handle this already?).
-            function callbackTouchStart(e: any) {
-                const control = this.getOwnerControl(/** @type {Node} */(e.target));
-                // Highlight the menu item.
-                control.handleMouseDown(e);
-            }
-
-            function callbackTouchEnd(e: any) {
-                const control = this.getOwnerControl(/** @type {Node} */(e.target));
-                // Activate the menu item.
-                control.performActionInternal(e);
-            }
-
-            menu.getHandler().listen(menu.getElement(), goog.events.EventType.TOUCHSTART,
-                callbackTouchStart);
-            menu.getHandler().listen(menu.getElement(), goog.events.EventType.TOUCHEND,
-                callbackTouchEnd);
-
             this.menu_ = menu;
             return menu;
         }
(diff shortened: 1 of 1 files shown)


pxtblocks/fields/field_gridpicker.ts
```diff
@@ -86,6 +88,11 @@ namespace pxtblockly {
             let tableContainerDom = tableContainer.getElement();
             if (tableContainerDom) {
                 let menuItemsDom = tableContainerDom.childNodes;
+                try {
+                    let firstItem = menuItemsDom[0].childNodes[0] as HTMLElement
+                    firstItem.className += " goog-menuitem-highlight"
+                } catch (e) {
```
guillaumejenkins: What is this trying to catch?

pxtblocks/fields/field_gridpicker.ts
```diff
@@ -374,10 +394,18 @@ namespace pxtblockly {
...
                 menuItem.setChecked(value == this.value_);
                 menu.addChild(menuItem, true);
+                if (i == 0) {
+                    this.firstItem_ = menuItem;
+                }
             }
 
             // Listen for mouse/keyboard events.
-            goog.events.listen(menu, goog.ui.Component.EventType.ACTION, callback);
+            goog.events.listen(menu, goog.ui.Component.EventType.ACTION, (e: any) => {
```
guillaumejenkins: What happens if you scroll the inventory picker with touch? Does it select the item when you stop the scroll? That's the current behavior (and is a bug), I'm wondering if `EventType.Action` fixes this

pxtblocks/fields/field_gridpicker.ts
caitlin-hennessy: Case where menuItemsDom is an empty array.  E.g., when you enter text in the search box such that there are no search results (like "foo").

caitlin-hennessy: Could do `if (menuItemsDom.length > 0 && menuItemsDom[0].childNodes.length > 0)` or something but that seems more cumbersome.

microsoftsam:
Awesome!

pxtblocks/fields/field_gridpicker.ts
guillaumejenkins: I think it's worth to use the `if`, to avoid accidentally swallowing exceptions (e.g. if in the future we add code to the `try`)

You can use `if (menuItemsDom.length && menuItemsDom[0].childNodes)` for slightly more compact code

caitlin-hennessy:
1. Select first item when hit 'enter' in search bar
2. Don't select item when scrolling on touch screen

PR #3017, opened 2017-09-22
V0 codalbuild

mmoskal:
This is backport of the "codal" build engine


pxtrunner/renderer.ts
```diff
@@ -193,6 +193,7 @@ namespace pxt.runner {
             const js = $('<code class="lang-typescript highlight"/>').text(c.text().trim());
             if (options.snippetReplaceParent) c = c.parent();
             const compiled = r.compileJS && r.compileJS.success;
+            // TODO should this use pxt.outputName() and not pxtc.BINARY_HEX
```
pelikhan: should we file a bug on this?

mmoskal: probably not worth it, unless it is actually breaking something (like UF2 download from embeds)

PR #3020, opened 2017-09-22
changing the tick target name from wonderworkshop to wonder

abchatra:
Tick is named as wonder in the target, makes it easier to correlate


docfiles/editorcards.html
```diff
@@ -1,4 +1,4 @@
-<a class="ui card raised link" target="_blank" href="https://makecode.microbit.org" onclick="tickEvent('makecode.target.click', {'target': 'microbit'})"
+<a class="ui card raised link" target="_blank" href="https://makecode.microbit.org" onclick="tickEvent('makecode.target.click', {'target': 'makecode', 'card': 'microbit'})"
```
microsoftsam: Instead of adding target: 'makecode' in all json passed to tickEvent, the tickEvent function here is only used by makecode.com and can be altered to always include target=makecode which would avoid a lot of duplication and make sure any new event added here always has target=makecode

PR #3022, opened 2017-09-23
Hide the plus when we run out of arguments

riknoll:
Simply hide the plus button on event blocks when we run out of arguments to add.


pxtblocks/blocklyloader.ts
```diff
@@ -887,13 +887,17 @@ namespace pxt.blocks {
                 }
             }
 
+            if (currentlyVisible === handlerArgs.length) {
```
pelikhan: `>==`, could it be that blockly source may have been corrupted or the API may have changed?

PR #3034, opened 2017-09-25
Fix sim locale

guillaumejenkins:
Ended up having to duplicate some `lf` logic in the sim. Tried to keep it minimal.

pxtsim does not depend on pxtrunner or any other namespace - which means it needs its own `lf`

### DIFF
--- a/pxtsim/localization.ts
+++ b/pxtsim/localization.ts
@@ -2,85 +2,11 @@
 namespace pxsim.localization {
     let _localizeStrings: Map<string> = {};
 
-    export function _localize(s: string) {
-        return _localizeStrings[s] || s;
-    }
-
     export function setLocalizedStrings(strs: Map<string>) {
-        _localizeStrings = strs;
+        _localizeStrings = strs || {};
     }
 
-    export function htmlEscape(_input: string) {
-        if (!_input) return _input; // null, undefined, empty string test
-        return _input.replace(/([^\w .!?\-$])/g, c => "&#" + c.charCodeAt(0) + ";");
-    }
-
-    export function jsStringQuote(s: string) {
-        return s.replace(/[^\w .!?\-$]/g,
-            (c) => {
-                let h = c.charCodeAt(0).toString(16);
-                return "\\u" + "0000".substr(0, 4 - h.length) + h;
-            });
-    }
-
-    export function fmt_va(f: string, args: any[]): string {
-        if (args.length == 0) return f;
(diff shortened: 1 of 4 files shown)


pxtsim/localization.ts
```diff
@@ -0,0 +1,85 @@
+namespace pxsim.localization {
+    let _localizeStrings: Map<string> = {};
```
microsoftsam: Can we leave a comment here describing where to find the duplicated function. Just so we always update one and the other hand in hand

pxtsim/localization.ts
```diff
@@ -0,0 +1,86 @@
+// Localization functions. Please port any modifications over to pxtlib/util.ts
+namespace pxsim.localization {
+    let _localizeStrings: Map<string> = {};
+
+    export function _localize(s: string) {
+        return _localizeStrings[s] || s;
+    }
+
+    export function setLocalizedStrings(strs: Map<string>) {
+        _localizeStrings = strs;
```
pelikhan: strs may be undefined -> ``strs || {}``

pxtsim/localization.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    export function jsStringQuote(s: string) {
+        return s.replace(/[^\w .!?\-$]/g,
+            (c) => {
+                let h = c.charCodeAt(0).toString(16);
+                return "\\u" + "0000".substr(0, 4 - h.length) + h;
+            });
+    }
+
+    export function fmt_va(f: string, args: any[]): string {
```
pelikhan: I don't think we need to support all that stuff for the sim. Maybe we just simplify it to just swapping the strings.

pxtsim/localization.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    export function _localize(s: string) {
+        return _localizeStrings[s] || s;
+    }
+
+    export function setLocalizedStrings(strs: Map<string>) {
+        _localizeStrings = strs;
+    }
+
+    export function htmlEscape(_input: string) {
```
pelikhan: don't export

pxtsim/localization.ts
```diff
@@ -0,0 +1,86 @@
...
+            } else if (spec == ":url") {
+                r = encodeURI(r).replace(/'/g, "%27").replace(/"/g, "%22");
+            } else if (spec == ":%") {
+                r = (v * 100).toFixed(1).toString() + '%';
+            }
+            return r;
+        });
+    }
+
+    let sForPlural = true;
```
pelikhan: always true

PR #3049, opened 2017-09-26
Bug 387 fix

caitlin-hennessy:
Fixes https://github.com/Microsoft/pxt-minecraft/issues/387

### DIFF
--- a/pxtblocks/fields/field_gridpicker.ts
+++ b/pxtblocks/fields/field_gridpicker.ts
@@ -265,8 +265,6 @@ namespace pxtblockly {
                         let firstItem = this.getFirstItem()
                         if (text && firstItem) {
                             this.selectItem(firstItem)
-                        } else {
-                            this.close()
                         }
                     }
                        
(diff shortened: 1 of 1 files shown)


pxtblocks/fields/field_gridpicker.ts
```diff
@@ -250,10 +260,16 @@ namespace pxtblockly {
...
-                    let firstItem = this.getFirstItem.bind(this)()
-                    if (e.keyCode == 13 && firstItem) {
-                        this.selectItem.bind(this)(firstItem)
+                    if (e.keyCode == 13) {
+                        let text = searchBar.value;
+                        let firstItem = this.getFirstItem()
+                        if (text && firstItem) {
+                            this.selectItem(firstItem)
+                        } else {
+                            this.close()
```
guillaumejenkins: I think "enter" should not close the picker if there are no blocks to select, I think it's ok to just leave the picker open

caitlin-hennessy: Done

PR #3050, opened 2017-09-26
Only show save and exit if the project has not been renamed

riknoll:


riknoll:
What do you want the text to say? Should the affirmative button still say save?

abchatra:
Save is fine as it is shown only once until you rename.

abchatra:
Rename means you have already named it once.

riknoll:
So what should the text say?

pelikhan:
We already have a similar purpose dialog when clicking on the save button: 
image

abchatra:
Sorry Peli it is not rename. Rename means they have named it once. @Jaqster tie breaker please.

abchatra:
Maybe "Name your project"

pelikhan:
We already have a dialog that is used to rename a project before saving, we should reuse it or fix it.

abchatra:
Yes agree "Name your project and Exit"

Jaqster:
Just to be clear, the behavior is:
1.) User is working on a named project already, clicks on the Home button - no dialog, gets taken to Home (his named project is at the front of My Stuff)
2.) User is working on an Untitled project, clicks on the Home button - pop this dialog giving them the opportunity to Name their project before going back to Home.

In the 2nd case, the dialog should say Project Name: <textbox here with default Untitled> with an "Exit" and "Cancel" button.  If the user doesn't want to name their project, they don't have to, they can simply click Exit.

Make sense?

pelikhan:
Exit Minecraft? I find this dialog confusing and I don't understand what the point is.

Jaqster:
No, exit your project.  And as part of exiting your program, we want to give them the opportunity to name their project if they haven't already.

Jaqster:
Can we rename this dialog box?
Would like the title to be "Exit Project", and the green button to be "Exit" (instead of Save)
exit dialog

PR #3051, opened 2017-09-26
handle report abuse on missing script

pelikhan:
Special 404 when user tries to report an missing script id. It seems that we have conflicting tslint versions that add/remove white spaces at the end of lines.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1328,7 +1328,7 @@ export class ProjectView
                         })
                         .catch(e => {
                             if (e.statusCode == 404)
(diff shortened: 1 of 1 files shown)


webapp/src/app.tsx
```diff
@@ -1326,7 +1326,12 @@ export class ProjectView
                         .then(res => {
                             core.infoNotification(lf("Report sent. Thank you!"))
                         })
-                        .catch(core.handleNetworkError);
+                        .catch(e => {
+                            if (e.statusCode == 404)
+                                core.warningNotification("Oops, we could not find this script.")
```
riknoll: lf?

PR #3052, opened 2017-09-26
Create a project before trying to decompile

riknoll:
If you're on the home screen after loading the editor and you select an example, it will fail to load. The error was that we were trying to decompile before a project was loaded and an exception was being thrown. It has been masked up until now because before the home screen we always had a project loaded in the editor.

Scenarios to test:
- [x] edit javascript with syntax errors, reload editor -> it should go back to JS
- [x] edit javascript correct code, reload editor -> goes back to JS
- [x] edit blocks correct code, reload editor -> goes back to blocks
- [x] edit blocks, invalid syntax (like a type error) - goes back to blocks

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -607,10 +607,6 @@ export class ProjectView
                 if (pkg.File.blocksFileNameRx.test(file.getName()) && file.getVirtualFileName()) {
                     if (!file.content) // empty blocks file, open javascript editor
                         file = main.lookupFile("this/" + file.getVirtualFileName()) || file
-                    else this.textEditor.decompileAsync(file.getVirtualFileName()).then(resp => {
-                        if (!resp.success)
-                            file = main.lookupFile("this/" + file.getVirtualFileName()) || file
-                    });
                 }
                 this.setState({
                     header: h,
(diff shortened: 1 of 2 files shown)


riknoll:
We don't decompile when you create a project. That being said, I am seeing it decompile twice for some reason; seems to be that we decompile when you reload the headers for a project if the new file is a blocks file. I'm investigating. The code is really weird.

riknoll:
@microsoftsam please take a look at the latest commit. I removed a check in app.tsx that I'm pretty sure does nothing as it was an async call in the middle of sync code. Not really sure what the original intention was.

microsoftsam:
@riknoll that bit of code is ancient so it's possible that its no longer needed however that code was added in order to make sure were were able to decompile the code at startup. This was back when some projects diverged and could no longer be converted back to blocks. We would want to open those projects in JS. 
In that instance the check would first check that it can decompile the code, before opening the blocks, if we didn't do that we would end up opening the main.blocks file which is not a representation of actual project code (as that no longer decompiles).

pelikhan:
This area is tricky, we should multiple scenarios like editing broken JavaScript, reopening and stuff. Let's do a full review tomorrow.

pelikhan:
Scenarios to test:
- edit javascript with syntax errors, reload editor -> it should go back to JS

riknoll:
testing

riknoll:
@pelikhan all those scenarios work fine.

riknoll:
The one scenario that might break is if the JS code somehow becomes invalid without any interaction for the user. For example, if we drop a language feature or API's change under the hood. I really don't think that's a scenario we need to handle, though, and even if we did I'm not sure going to the JS is the right response.

PR #3060, opened 2017-09-27
UI Fixes

microsoftsam:
- Switch to using textContent instead of innerText as it's the standard, more performant and FF 35-45 doesn't support innerText. 
- Fix block drag opacity to apply just on the block and not the stack
- Fix dropdown text to apply the monaco font instead of the default
- Pull a pxt-blockly release that has further offline fixes (including the dropdown arrow svg)
- Fix card width in side doc


pelikhan:
What is the data of FF version usage?

microsoftsam:
Roughly 50% on FF 55. 

2% on 38
2% on 45
1% on 43
< other <40 account for about 2%
so roughly 5-7% of FF users affected by the innerText issue. 
Granted as described in http://kellegous.com/j/2013/02/27/innertext-vs-textcontent/ the fix here is not only to standard, but also more performant. For something like the monaco toolbox that we continually call this, it makes sense to make that change.

pelikhan:
LGTM

PR #3061, opened 2017-09-27
More robust tutorials. Loading queue.

microsoftsam:
This PR includes changes to bring a loading queue to the core show/hide loading methods. The issue with the current paradigm of just showing the loading spinner at different points and then hiding it at various points is that it's non deterministic and it's possible to be done with one action, but not done with another (namely tutorials). It's common to find yourself still waiting for the tutorial to load but since the new project has finished loading and all the blocks are there, the spinner disappears. 

This introduces an ID associated with each show/hide loading call, and uses a queue to show the messages. If one action is complete, it shows the next until there are no more to show. 

PR also includes a couple of catch clauses and a tutorial error message to send the parent if we fail to decompile the blocks or fail to find the tutorial referenced. 

PR also checks for internet connection before starting to load a tutorial. Since we rely on the iframe loading and requesting the tutorial from the server, we need to check that the iframe is able to connect to the Internet in the first place. 
using navigator.isOnline:
http://caniuse.com/#feat=online-status

Tested on: 
- [x] Edge
- [x] IE
- [x] Chrome Mac
- [x] Firefox Mac and Windows
- [x] Safari

### DIFF
--- a/pxtlib/browserutils.ts
+++ b/pxtlib/browserutils.ts
@@ -20,6 +20,10 @@ namespace pxt.BrowserUtils {
         return hasNavigator() && /mobi/i.test(navigator.userAgent);
     }
 
+    export function isOnline(): boolean {
+        return navigator.onLine;;
+    }
+
     //MacIntel on modern Macs
     export function isMac(): boolean {
         return hasNavigator() && /Mac/i.test(navigator.platform);
--- a/pxtlib/browserutils.ts
+++ b/pxtlib/browserutils.ts
@@ -20,10 +20,6 @@ namespace pxt.BrowserUtils {
         return hasNavigator() && /mobi/i.test(navigator.userAgent);
     }
 
-    export function isOnline(): boolean {
-        return navigator.onLine;;
-    }
-
     //MacIntel on modern Macs
     export function isMac(): boolean {
         return hasNavigator() && /Mac/i.test(navigator.platform);
--- a/pxtlib/emitter/cloud.ts
+++ b/pxtlib/emitter/cloud.ts
@@ -94,7 +94,7 @@ namespace pxt.Cloud {
 
     export function isLoggedIn() { return !!accessToken }
 
-    function isNavigatorOnline() {
+    export function isNavigatorOnline() {
         return navigator && navigator.onLine;
     }
 
--- a/pxtrunner/runner.ts
+++ b/pxtrunner/runner.ts
@@ -351,6 +351,7 @@ namespace pxt.runner {
                         case "tutorial":
                             let body = $('body');
                             body.addClass('tutorial');
+                            $(loading).hide();
                             return renderTutorialAsync(content, src);
                         case "book":
                             return renderBookAsync(content, src);
@@ -617,6 +618,7 @@ ${files["main.ts"]}
                 // Extract toolbox block ids
                 let toolboxSubset: { [index: string]: number } = {};
                 return Promise.resolve()
+                    .then(() => renderMarkdownAsync(content, tutorialmd, { tutorial: true }))
                     .then(() => {
                         let uptoSteps = steps.join();
                         uptoSteps = uptoSteps.replace(/((?!.)\s)+/g, "\n");
@@ -650,7 +652,6 @@ ${files["main.ts"]}
                         }
                         return Promise.resolve();
                     })
-                    .then(() => renderMarkdownAsync(content, tutorialmd, { tutorial: true }))
                     .then(() => {
                         // Split the steps
                         let stepcontent = content.innerHTML.split(newAuthoring ? /<h2.*\/h2>/gi : /<h3.*\/h3>/gi);
--- a/pxtrunner/runner.ts
+++ b/pxtrunner/runner.ts
@@ -597,6 +597,10 @@ ${files["main.ts"]}
                     steps = tutorialmd.split(/^###[^#].*$/gmi);
                     newAuthoring = false;
                 }
+                if (steps[0].indexOf("# Not found") == 0) {
+                    pxt.log(`Tutorial not found: ${tutorialid}`);
+                    throw new Error(`Tutorial not found: ${tutorialid}`);
+                }
                 let stepInfo: editor.TutorialStepInfo[] = [];
                 tutorialmd.replace(newAuthoring ? /^##[^#](.*)$/gmi : /^###[^#](.*)$/gmi, (f, s) => {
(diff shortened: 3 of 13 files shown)


pxtlib/browserutils.ts
```diff
@@ -20,6 +20,10 @@ namespace pxt.BrowserUtils {
         return hasNavigator() && /mobi/i.test(navigator.userAgent);
     }
 
+    export function isOnline(): boolean {
```
pelikhan: use ``Cloud.isOnline()`` instead or unify both

pxtrunner/runner.ts
```diff
@@ -597,6 +597,10 @@ ${files["main.ts"]}
                     steps = tutorialmd.split(/^###[^#].*$/gmi);
                     newAuthoring = false;
                 }
+                if (steps[0].indexOf("# Not found") == 0) {
+                    pxt.log(`Tutorial not found: ${tutorialid}`);
+                    throw new Error(`Tutorial not found: ${tutorialid}`);
```
pelikhan: use ``lf`` if message displayed to user

pxtrunner/runner.ts
```diff
@@ -639,6 +643,9 @@ ${files["main.ts"]}
                                         toolboxSubset[blk.type] = 1;
                                     }
                                 }
+                            }).catch(() => {
+                                pxt.log(`Failed to decompile tutorial: ${tutorialid}`);
+                                throw new Error(`Failed to decompile tutorial: ${tutorialid}`);
```
pelikhan: use ``lf`` if message displayed to user

webapp/src/app.tsx
```diff
@@ -1533,9 +1548,15 @@ ${compileService && compileService.githubCorePackage && compileService.gittag ?
 
     startTutorial(tutorialId: string, tutorialTitle?: string) {
         pxt.tickEvent("tutorial.start");
-        core.showLoading(lf("starting tutorial..."));
-        this.startTutorialAsync(tutorialId, tutorialTitle)
-            .then(() => Promise.delay(500));
+        // Check for Internet access
+        if (!pxt.BrowserUtils.isOnline()) {
```
pelikhan: use Cloud.isOnline

pxtrunner/runner.ts
microsoftsam: Yep. thanks

microsoftsam: Message not displayed to the user. Just goes to our logs

microsoftsam:
@pelikhan addressed PR comments

webapp/src/app.tsx
microsoftsam: Cloud.isOnline maintains a variable remembering if the cloud is online. going to use cloud.isNavigatorOnline for this one, as there's a possibility that the cloud could be online but the user just went offline.

webapp/src/app.tsx
```diff
@@ -1549,7 +1549,7 @@ ${compileService && compileService.githubCorePackage && compileService.gittag ?
     startTutorial(tutorialId: string, tutorialTitle?: string) {
         pxt.tickEvent("tutorial.start");
         // Check for Internet access
-        if (!Cloud.isOnline()) {
+        if (!pxt.Cloud.isNavigatorOnline()) {
```
pelikhan: navigator.isOnline is not reliable. It does not handle hotel proxy, or disconnect internet. Why not using ``Cloud.isOnline``?

microsoftsam: Cloud online only works the second time you try to connect and it fails.

microsoftsam: This method still isn't fool proof, it's possible that you start the iframe, and not receive a response back since it never loaded. I'm trying to mitigate that a little, but there's no way for us to tell if an iframe didnt load (Unless we set a timeout to check, but then we're limiting the max it will take for a tutorial to load).

pelikhan: > Cloud online only works the second time you try to connect and it fails.

I don't see why isNavigatorOnline would change that behavior. How do you repro?

pxtlib/emitter/cloud.ts
```diff
@@ -94,7 +94,7 @@ namespace pxt.Cloud {
 
     export function isLoggedIn() { return !!accessToken }
 
-    function isNavigatorOnline() {
+    export function isNavigatorOnline() {
```
pelikhan: Use ``Cloud.isOnline`` instead.

webapp/src/app.tsx
microsoftsam: Try loading the page. Then disconnecting from the net. Cloud.isOnline maintains an _isOnline variable and so it will return true. 
IsNavigatorOnline asks the browser if it's online at that point in time, and would return false.

pelikhan: We should fix this behavior rather than introducing a new fuction.

riknoll:
I would prefer if instead of passing strings around we did something similar to `setTimeout()` and `cancelTimeout()`. Like this:

```typescript
const mId = core.showLoading();

// Do stuff...

core.hideLoading(mId);
```

But I don't feel super strongly about it.

microsoftsam:
@riknoll the above might work for most scenarios, but as you can see in the code there are scenarios where the loading start and end are in different parts of the code.

webapp/src/app.tsx
microsoftsam: isNavigatorOnline was already there. Cloud.isOnline serves a different purpose as it checks whether or not the backend is online. ie: both the user is connected to the internet AND they are able to access pxt.io/api and like I mentioned earlier, it's not calling the backend to check, it's just keeping a var around for whether or not it's online 
In the case of the tutorial, the backend is not the concern, but whether or not an iframe can connect to: 
/--docs#tutorial:tutorials/getting-started:js:en

I think the proper fix here is perhaps to send a message on load from the iframe telling the parent that we were able to connect. and same on the client side a timeout for say 5 seconds to check whether or not it received that initial onload message.

pelikhan: Agreed on a better fix. Navigator.onLine does not mean the cloud is accessible, might be behind a hotel jail or blocked by the school.

/--docs is served by the cloud and uses the /api to request the .md file. This can fail in many places depending on when the connection drops.

microsoftsam:
Made the tutorial faster by setting the path as soon as we start, rather than waiting to create a project first.

microsoftsam:
I'm just checking a couple more things, let me merge.

PR #3062, opened 2017-09-27
reload in editor if editing on appcache update

pelikhan:
If editor is open with project, reload into that project instead of going to homescreen.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -2156,6 +2156,9 @@ function handleHash(hash: { cmd: string; arg: string }): boolean {
             theEditor.importProjectFromFileAsync(fileContents)
                 .done(() => core.hideLoading());
             return true;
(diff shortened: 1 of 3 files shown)


pelikhan:
Added

abchatra:
Looks good. Thank you Peli for quick fix.

pelikhan:
Good catch

pelikhan:
Not in scope for this fix.

pelikhan:
Still testing do not merge

pelikhan:
Chrome seems to be eating the search section of the URL away when using replace.

pelikhan:
It works because you are changing the URL, but in this case we are changing the hashmark section which does not trigger a page navigation.

pelikhan:
Setting the hash and reloading works for me.

pelikhan:
@guillaumejenkins ready for review

PR #3066, opened 2017-09-27
Mutate defaults for handler parameter blocks

riknoll:
I could have used the same comment attribute that we already have for creating multiple instances of mutator blocks but creating a new one seemed cleaner since the behavior is slightly different.


microsoftsam:
How is this used? Can we add authoring info to the docs, or at least the PR?

PR #3067, opened 2017-09-27
Don't close save and exit on space

riknoll:


microsoftsam:
Hadn't realised you'd already fixed this. There's one other place that needs this fix in core.ts. Otherwise Blockly confirm dialogs are also affected by this. 

See PR: https://github.com/Microsoft/pxt/pull/3071/files

Closing this one as the fix is already in master. Thanks @riknoll

PR #3074, opened 2017-09-28
Better zoom

microsoftsam:
No need to have such a fine grained zoom. Zoom twice as much on each button click

betterzoom


Jaqster:
👍

PR #3078, opened 2017-09-28
Fix decompiling in the gallery

riknoll:


microsoftsam:
Does this need to go to v0?

riknoll:
This is for the home screen. I don't think it needs to be moved over.

microsoftsam:
Projects dialog in microbit uses the same logic. Perhaps we should move over?

microsoftsam:
I guess it's unlikely you'll end up in this situation as we always create a project on start up / load a project in the microbit.

riknoll:
Eh, it might catch a few uncommon bugs. Doesn't hurt to move it over.

microsoftsam:
Ok moving.

PR #3079, opened 2017-09-28
Show cookie on top

microsoftsam:
Always show cookie on top of everything else. 

Cleanup electron logic so that there's one centralized place for the source of truth. Also only showing cookie message on website, not in an app.


pelikhan:
@microsoftsam should be at the top of the page too

PR #3089, opened 2017-09-29
package extension work (formerly v2)

pelikhan:


pelikhan:
@riknoll @microsoftsam please take a peek. I reviewed all lines and it seems to align with master.

PR #3094, opened 2017-09-30
Update text blocks with better defaults

microsoftsam:
- join: "Hello" "World"
- substring had a default length of 1000000 ??
- changed `length of text ""` to `length of ""`
- changed parseInt to have a default number string to make it's function a little more obvious

I would have liked to change the compare block, but I don't think it's possible with our current architecture. @riknoll thoughts?

Before:
screen shot 2017-09-30 at 9 19 50 am

After:
screen shot 2017-09-30 at 9 17 45 am

### DIFF
--- a/webapp/src/toolbox.ts
+++ b/webapp/src/toolbox.ts
@@ -238,7 +238,7 @@ const defaultToolboxString = `<xml id="blocklyToolboxDefinition" style="display:
         <block type="text_length">
             <value name="VALUE">
                 <shadow type="text">
-                    <field name="TEXT">abc</field>
+                    <field name="TEXT">Hello</field>
                 </shadow>
             </value>
         </block>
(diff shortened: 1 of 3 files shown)


webapp/src/toolbox.ts
```diff
@@ -238,20 +238,20 @@ const defaultToolboxString = `<xml id="blocklyToolboxDefinition" style="display:
         <block type="text_length">
             <value name="VALUE">
                 <shadow type="text">
-                    <field name="TEXT">abc</field>
```
pelikhan: use lf

webapp/src/toolbox.ts
```diff
@@ -238,20 +238,20 @@ const defaultToolboxString = `<xml id="blocklyToolboxDefinition" style="display:
...
             <mutation items="2"></mutation>
             <value name="ADD0">
                 <shadow type="text">
-                    <field name="TEXT"></field>
+                    <field name="TEXT">Hello</field>
                 </shadow>
             </value>
             <value name="ADD1">
                 <shadow type="text">
-                    <field name="TEXT"></field>
```
pelikhan: use lf

webapp/src/toolbox.ts
microsoftsam: done

PR #3095, opened 2017-09-30
Add ON / OFF description text onto the field toggle

microsoftsam:
onoff


pelikhan:
Do we have a documented list of field editors?

microsoftsam:
No but I was thinking we should add, plus add the list of parameters that they accept. Even though we have interfaces in the code, they're not exported and hard to find.

pelikhan:
+ those amazing gifs :)

microsoftsam:
Yeah I'll start preparing a docs page

PR #3102, opened 2017-10-02
Updating avatar images in v0

microsoftsam:


abchatra:
Can you add more information? Before & after screen shots? Why this change in v0?

microsoftsam:
I'm pulling these SVGs from master. I can minify in another change. 

This change has to be in v0, because v0 is where we pull the semantic.css and use it for all our docs sites.  Issue I'm trying to fix is the docs having an old avatar image.

Before: 
screen shot 2017-10-01 at 10 17 25 pm

After: 
Hopefully the new MakeCode avatar icon

abchatra:
Ok makes sense

Issue #3104, opened 2017-10-02
Navigation buttons in docs overlap the text


microsoftsam:
Making it hard to select the text. 

It only overlaps in the computer but not wide view. 

<img width="560" alt="screen shot 2017-10-01 at 10 49 58 pm" src="https://user-images.githubusercontent.com/16690124/31062755-0cc6bb68-a6fb-11e7-8655-f0953add0db3.png">

microsoftsam:
Fixed.

PR #3105, opened 2017-10-02
Fix overlapping navigation buttons in docs

microsoftsam:
Fix issue with navigation buttons overlapping the body text making it hard to select the text.
This has to be in v0 as it's a docs fix and we need to publish the doc CSS from v0.

The fix was to make sure the nav buttons had a limited width smaller than the padding of the main body text.
 
After: 
screen shot 2017-10-01 at 10 52 52 pm

Fixes https://github.com/Microsoft/pxt/issues/3104

### DIFF
--- a/docfiles/style.css
+++ b/docfiles/style.css
@@ -275,8 +275,8 @@ svg {
     bottom: 0;
     margin: 0;
     padding: 10px;
-    max-width: 150px;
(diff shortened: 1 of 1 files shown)


Jaqster:
Awesome!  That was quick!  Thanks Sam :-)

PR #3108, opened 2017-10-02
C# code generation

mmoskal:
This implements `appTarget.nativeType == "C#"`
See https://github.com/Microsoft/pxt-samplecs/blob/master/libs/core/Core.cs for accompanying runtime implementation, which should be integrated in the final target.

It currently passes `lang-test0`. 

While we're not using this yet, I would like it merged, to avoid bitrot (especially in `cpp.ts`, which now also handles parsing C#).


pxtcompiler/emitter/hexfile.ts
```diff
@@ -155,6 +155,12 @@ namespace ts.pxtc {
 
             patchSegmentHex(hex)
 
+            if (opts.nativeType == "C#") {
```
pelikhan: Consider having a constant for C#

pelikhan:
@mmoskal did you do a sanity check for other targets?

mmoskal:
Will do tomorrow

mmoskal:
Just checked Adafruit - the exact same binary is generated and no changes in shims.d.ts.

Issue #3117, opened 2017-10-03
Replace serial animation with flashing indicator


cait1inhennessy:

cait1inhennessy:
Done

PR #3127, opened 2017-10-03
don't show animation pill for serial buttons

pelikhan:
The Animation Pill is confusing (showing random data that has nothing to do with the current process). It's since' burning cycles for no good reason, this PR rips it off.

We need to be extra careful about running JS in the background of the editor. It might create glitches in the editing experience (the simulator runs in its own iframe for example).

image


pelikhan:
Fix for https://github.com/Microsoft/pxt/issues/3117

microsoftsam:
@pelikhan generally use sui.fireClickOnEnter for onKeyDown, setting it to the same handler will mean any key down on that element will fire the click. This includes tab and other keys. 

We should really just default to that

microsoftsam:
screen shot 2017-10-03 at 6 39 23 pm

microsoftsam:
Fixes https://github.com/Microsoft/pxt/issues/3111

pelikhan:
LGTM

PR #3134, opened 2017-10-03
Fixing race conditions with simulator start + serial output

pelikhan:
- delay serial cleaning after compilation
- don't typecheck when showing serial editor
- don't show serial stuff if not supported.


caitlin-hennessy:
Awesome!

PR #3139, opened 2017-10-04
Home screen updates

microsoftsam:
Update home screen with better button placement. 
Finishing a tutorial lands a user in the editor.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1622,7 +1622,7 @@ ${compileService && compileService.githubCorePackage && compileService.gittag ? 
         this.exitTutorialAsync()
             .then(() => {
                 let curr = pkg.mainEditorPkg().header;
-                this.loadHeaderAsync(curr);
+                return this.loadHeaderAsync(curr);
             }).done(() => {
                 core.hideLoading("leavingtutorial");
             })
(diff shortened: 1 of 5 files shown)


webapp/src/app.tsx
```diff
@@ -1618,15 +1618,18 @@ ${compileService && compileService.githubCorePackage && compileService.gittag ?
 
     completeTutorial() {
         pxt.tickEvent("tutorial.complete");
-        this.leaveTutorial();
+        core.showLoading("leavingtutorial", lf("leaving tutorial..."));
+        this.exitTutorialAsync()
+            .then(() => {
+                let curr = pkg.mainEditorPkg().header;
+                this.loadHeaderAsync(curr);
```
riknoll: return this promise

webapp/src/projects.tsx
```diff
@@ -328,44 +338,29 @@ export class ProjectsCarousel extends data.Component<ProjectsCarouselProps, Proj
...
-        }
-
         const onClick = (scr: any) => {
             this.props.onClick(scr);
         }
 
-        return <div className="ui dimmable">
-            {this.hasFetchErrors ?
-                <div className="ui carouselouter">
+        let carouselJSX: JSX.Element;
```
riknoll: Why the variable? Why not just have separate return statements?

microsoftsam: Very different code paths whether its the header or parts. If split up, it would be three different chained if elses

webapp/src/app.tsx
microsoftsam: Ok.

webapp/src/projects.tsx
microsoftsam: Ah I see what you mean. Yes. Fixing

PR #3141, opened 2017-10-04
Search unable to find built-in blocks

abchatra:
Fixes https://github.com/Microsoft/pxt-minecraft/issues/409
Blockly object is not defined in here. Access to it throws an exception which results in built-in blocks not being added to cached block definitions.

Remove the unnecessary code.


abchatra:
Yes, as far as I could tell.

microsoftsam:
Hang on.

microsoftsam:
blocklyloader.ts calls mathOp2Def.outputShape

abchatra:
Oops I merged it. I can revert it.

microsoftsam:
The output shape defined when we didn't used to have round as our default. 
Now that round is our default, this can potentially be cleaned up, but we should clean up all references to it as well :)

abchatra:
This looks good then? I want search to be fixed in the final release of minecraft next week. We can do clean up later

microsoftsam:
Well it's fine as in it doesn't cause any issues, but it's because the default is round, so got lucky ha! I'll clean it up now.

microsoftsam:
but actually come to think of it, this isn't the right fix. I think there's an underlying issue here, removing the outputShape setting there removes any flexibility for having a different shape for a built in block. Investigating.

abchatra:
Looks good. Though the enum is duplicate, I don't see a work around it if we need it.

pelikhan:
We don’t need it now so we should cut and clean.

microsoftsam:
Already fixed properly: https://github.com/Microsoft/pxt/pull/3143

pelikhan:
@microsoftsam Proper fix is to remove the feature all together as it is not used (if I understand correctly). If and when we would need such flexibility, it can come back.

microsoftsam:
There you go: https://github.com/Microsoft/pxt/pull/3144

PR #3143, opened 2017-10-04
Fixing search whilst still maintaining outputShape

microsoftsam:
Revert https://github.com/Microsoft/pxt/pull/3141

Blockly is not defined in the web worker which is where search happens. Using out own enum to define Output Shapes (with the exception of initBlock, where Blockly is defined). 

The enum matches Blockly's values for shapes.


abchatra:
Thanks @microsoftsam

pelikhan:
Why do we still need that enum ?

PR #3151, opened 2017-10-04
SVG helper for horizontal gradients

riknoll:


pxtsim/svg.ts
```diff
@@ -152,18 +152,18 @@ namespace pxsim.svg {
         });
     }
 
-    export function mkLinearGradient(id: string): SVGLinearGradientElement {
+    export function mkLinearGradient(id: string, vertical = true): SVGLinearGradientElement {
```
pelikhan: default arguments should always be false IMO. so ``vertical`` -> ``horizontal``

riknoll: Sure

PR #3153, opened 2017-10-04
Square field toggle

microsoftsam:
Fix field toggle to work when not in a shadow block, and can be in a square on block field.


pelikhan:
What scenario does this address?

PR #3154, opened 2017-10-04
Every 5 seconds, drop charts that haven't received data

caitlin-hennessy:


webapp/src/serial.tsx
```diff
@@ -147,6 +149,18 @@ export class Editor extends srceditor.Editor {
         }
     }
 
+    dropStaleCharts() {
+        let now = Util.now()
+        for (let i = 0; i < this.charts.length; ++i) {
```
pelikhan: use ``this.charts.forEach(chart => ...)``

webapp/src/serial.tsx
```diff
@@ -54,6 +55,7 @@ export class Editor extends srceditor.Editor {
 
     constructor(public parent: pxt.editor.IProjectView) {
         super(parent)
+        setInterval(this.dropStaleCharts.bind(this), 5000)
```
pelikhan: should only be started be active while editor is active

caitlin-hennessy: Fixed

webapp/src/serial.tsx
caitlin-hennessy: Fixed

PR #3161, opened 2017-10-05
Support for theming serial chart colors

caitlin-hennessy:


webapp/src/serial.tsx
```diff
@@ -365,22 +359,38 @@ class Chart {
...
+        this.chart.addTimeSeries(this.line, {strokeStyle: lineColor, fillStyle: this.hexToHalfOpacityRgba(lineColor), lineWidth: 3})
 
         if (this.variable)
             this.rootElement.appendChild(this.makeLabel())
         this.rootElement.appendChild(this.makeCanvas())
         this.addPoint(value)
     }
 
+    hexToHalfOpacityRgba(hex: string) {
+        let shorthandRegex = /^#?([a-f\d])([a-f\d])([a-f\d])$/i
```
microsoftsam: This should really be in a color util somewhere.

PR #3168, opened 2017-10-06
Arduino UNO support

mmoskal:
This fixes many things in AVR code gen. In addition:
* make parsing of C++ constants a bit smarter
* adapt to recent codal build system changes (this will be required to build CPX after "merge fest"; once it builds again)
* auto-reconnect in `pxt hidserial`
* allow specifying flash size in pxtarget


pelikhan:
@mmoskal did you do a sanity check with Adafruit?

mmoskal:
Adafruit doesn't build ATM - merge fest

tballmsft:
@mmoskal - did you check that thumb code gen wasn't affected on an ARM target?

mmoskal:
The thing is that Adafruit doesn't build because of codal issues. Joe promised some tags to have a stable version (we were using branches which is WRONG). Will look into that tomorrow.

mmoskal:
Now tested to verify ARM isn't affected.

tballmsft:
Good to go, @mmoskal?

mmoskal:
Good to go!

PR #3169, opened 2017-10-06
load gallery list from targetconfig

pelikhan:
Galleries are specific in the targetconfig which allows to modify them without shipping the entire editor.


microsoftsam:
Good call.

PR #3218, opened 2017-10-12
fixing download screenshot feature

pelikhan:
The batch download screenshot feature looks for a "blocklycss" style item.


pelikhan:
@microsoftsam do we support the ``data-rtl`` stuff in our docs? :)

PR #3223, opened 2017-10-13
Support localized and custom text inside Field toggle

microsoftsam:
Localising the text inside of the field editor, supporting custom text length (dependent on language). 
Adding two new field editors: 
- FieldToggleYesNo
- FieldToggleOnOff

The default toggle is now True / False. 

Creating new fields with custom text is a simple as extending FieldToggle and replacing the getTrueText and getFalseText methods. 

Sample:
screen shot 2017-10-12 at 10 17 37 pm

Tested on: 
- Chrome Mac
- Edge 15
- IE 11

### DIFF
--- a/pxtblocks/blocklycustomeditor.ts
+++ b/pxtblocks/blocklycustomeditor.ts
@@ -32,6 +32,8 @@ namespace pxt.blocks {
         registerFieldEditor('imagedropdown', pxtblockly.FieldImageDropdown);
         registerFieldEditor('colorwheel', pxtblockly.FieldColorWheel);
         registerFieldEditor('toggle', pxtblockly.FieldToggle);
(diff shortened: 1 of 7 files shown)


pxtblocks/blocklycustomeditor.ts
```diff
@@ -32,6 +32,8 @@ namespace pxt.blocks {
         registerFieldEditor('imagedropdown', pxtblockly.FieldImageDropdown);
         registerFieldEditor('colorwheel', pxtblockly.FieldColorWheel);
         registerFieldEditor('toggle', pxtblockly.FieldToggle);
+        registerFieldEditor('toggleonoff', pxtblockly.FieldToggleOnOff);
```
pelikhan: We really need to document these. Lots of great stuff here.

PR #3228, opened 2017-10-13
VM code generation

mmoskal:
This implements codegen for VM running on AVR.

It additionally adds support for some recent codal build changes.

### DIFF
--- a/pxtcompiler/emitter/emitter.ts
+++ b/pxtcompiler/emitter/emitter.ts
@@ -99,6 +99,14 @@ namespace ts.pxtc {
         return target.nativeType == NATIVE_TYPE_CS || (!target.jsRefCounting && !target.isNative)
     }
 
+    export function isStackMachine() {
+        return target.nativeType == NATIVE_TYPE_AVRVM
+    }
+
+    export function isAVR() {
+        return target.nativeType == NATIVE_TYPE_AVRVM || target.nativeType == NATIVE_TYPE_AVR
+    }
+
     function isRefType(t: Type) {
         checkType(t);
         if (noRefCounting())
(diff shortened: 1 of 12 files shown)


mmoskal:
Tested on pxt-adafruit - no effect on ARM code

pxtcompiler/emitter/emitter.ts
```diff
@@ -2155,7 +2165,7 @@ ${lbl}: .short 0xffff
             }
 
             // otherwise we assume a lambda
-            if (args.length > 3)
+            if (args.length > (isStackMachine() ? 2 : 3))
                 userError(9217, lf("lambda functions with more than 3 arguments not supported"))
```
pelikhan: update error message with number of args?

pxtcompiler/emitter/hexfile.ts
```diff
@@ -646,15 +646,17 @@ ${hex.hexPrelude()}
     }
 
     function mkProcessorFile(nativeType: string) {
-        let processor: assembler.AbstractProcessor = null
+        let b: assembler.File
+
         if (nativeType == NATIVE_TYPE_AVR)
```
pelikhan: use isAvr() ?

pxtcompiler/emitter/emitter.ts
mmoskal: didn't want to mess with localization, also I guess we may want to lift this restriction

pxtcompiler/emitter/hexfile.ts
mmoskal: isAvr() == AVR || AVRVM - this is only for native AVR

PR #3256, opened 2017-10-16
UWP save()

caitlin-hennessy:
Goes with https://github.com/Microsoft/pxt-microbit/pull/553

### DIFF
--- a/pxtwinrt/deploy.ts
+++ b/pxtwinrt/deploy.ts
@@ -50,15 +50,18 @@ namespace pxt.winrt {
         return pxt.winrt.promisify(savePicker.pickSaveFileAsync()
             .then((file) => {
                 if (file) {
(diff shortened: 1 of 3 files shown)


pxtwinrt/deploy.ts
```diff
@@ -50,15 +50,18 @@ namespace pxt.winrt {
...
                     bytes.forEach((b) => ar.push(b));
                     return Windows.Storage.FileIO.writeBytesAsync(file, ar)
                         .then(() => true);
                 }
                 // Save cancelled
                 return Promise.resolve(false);
-            }));
+            })
+        ).catch(e => {
+            console.log(e);
```
pelikhan: Use PXT.reportException

caitlin-hennessy: I've made this change, it will be in my next PR.

PR #3259, opened 2017-10-17
Serial tweaks

pelikhan:
- [x] show cursors on status indicator button
- [x] show last value (rounded to 2 sig decimals) in label . The problem is that smoothie only shows min/max not the current value.

### DIFF
--- a/webapp/src/serial.tsx
+++ b/webapp/src/serial.tsx
@@ -225,7 +225,7 @@ export class Editor extends srceditor.Editor {
 
         core.confirmAsync({
             logos: undefined,
-            header: lf("Analyze Data"),
+            header: lf("Export data"),
             hideAgree: true,
             disagreeLbl: lf("Close"),
             onLoaded: (_) => {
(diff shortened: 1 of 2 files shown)


webapp/src/serial.tsx
```diff
@@ -352,7 +353,7 @@ class Chart {
     constructor(source: string, variable: string, value: number, chartIdx: number) {
         const serialTheme = pxt.appTarget.serial && pxt.appTarget.serial.editorTheme
         // Initialize chart
-        const chartConfig = {
+        const chartConfig: IChartOptions = {
```
caitlin-hennessy: Is this interface defined somewhere?

webapp/src/serial.tsx
```diff
@@ -422,7 +422,10 @@ class Chart {
 
     addPoint(value: number) {
         this.line.append(Util.now(), value)
-        this.lastUpdatedTime = Util.now()
+        this.lastUpdatedTime = Util.now();
+        // update label with last value
+        const valueText = Number(Math.round(Number(value + "e+2"))  + "e-2").toString();
```
caitlin-hennessy: What does `Number(Math.round(Number(value + "e+2"))  + "e-2")` do?

webapp/src/serial.tsx
pelikhan: yes in our local typings.

webapp/src/serial.tsx
pelikhan: It's special JS trick to round number to up to 2 decimal numbers.

PR #3277, opened 2017-10-18
Misc blockly fixes

microsoftsam:
- Only show group labels when there are blocks in that group
- Use groups for subcategories
- Show a subcategory heading label such as: 
screen shot 2017-10-18 at 1 51 34 pm
- Fix icondropdown field removing its explicit width when it's hidden


pelikhan:
Looks great!

PR #3288, opened 2017-10-19
More serial

caitlin-hennessy:
Add buffers to ensure that only complete serial messages are sent to the editor.

This will need to go into master as well.

### DIFF
--- a/pxtwinrt/serial.ts
+++ b/pxtwinrt/serial.ts
@@ -56,13 +56,25 @@ namespace pxt.winrt {
         port.device.baudRate = 115200;
         let stream = port.device.inputStream;
         let reader = new Windows.Storage.Streams.DataReader(stream);
+        let serialBuffers: {[source: string]: string} = {};
+        const maxBufferLength = 255;
         let readMore = () => reader.loadAsync(32).done((bytesRead) => {
             let msg = reader.readString(Math.floor(bytesRead / 4) * 4);
-            window.postMessage({
-                type: 'serial',
-                data: msg,
-                id: id
-            }, "*");
+            const data = msg || ""
+            const source = id || "?"
+            for (let i = 0; i < data.length; ++i) {
+                const char = data[i]
+                serialBuffers[source] = serialBuffers[source] ? serialBuffers[source] + char : char
+                if (/.*\n/.test(serialBuffers[source]) || serialBuffers[source].length > maxBufferLength) {
+                    let buffer = serialBuffers[source]
+                    serialBuffers[source] = ""
+                    window.postMessage({
+                        type: "serial",
+                        id: source,
+                        data: buffer
+                    }, "*")
+                }
+            }
             readMore();
         }, (e) => {
             setTimeout(() => startDevice(id), 1000);
--- a/pxtwinrt/serial.ts
+++ b/pxtwinrt/serial.ts
@@ -57,24 +57,9 @@ namespace pxt.winrt {
         let stream = port.device.inputStream;
         let reader = new Windows.Storage.Streams.DataReader(stream);
         let serialBuffers: pxt.Map<string> = {};
-        const maxBufferLength = 255;
         let readMore = () => reader.loadAsync(32).done((bytesRead) => {
             let msg = reader.readString(Math.floor(bytesRead / 4) * 4);
-            const data = msg || ""
-            const source = id || "?"
-            for (let i = 0; i < data.length; ++i) {
-                const char = data[i]
-                serialBuffers[source] = serialBuffers[source] ? serialBuffers[source] + char : char
-                if (char === "\n" || serialBuffers[source].length > maxBufferLength) {
-                    let buffer = serialBuffers[source]
-                    serialBuffers[source] = ""
-                    window.postMessage({
-                        type: "serial",
-                        id: source,
-                        data: buffer
-                    }, "*")
-                }
-            }
+            pxt.Util.bufferSerial(serialBuffers, msg, id)
             readMore();
         }, (e) => {
             setTimeout(() => startDevice(id), 1000);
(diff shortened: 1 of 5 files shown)


pxtwinrt/serial.ts
```diff
@@ -56,13 +56,25 @@ namespace pxt.winrt {
...
-                type: 'serial',
-                data: msg,
-                id: id
-            }, "*");
+            const data = msg || ""
+            const source = id || "?"
+            for (let i = 0; i < data.length; ++i) {
+                const char = data[i]
+                serialBuffers[source] = serialBuffers[source] ? serialBuffers[source] + char : char
+                if (/.*\n/.test(serialBuffers[source]) || serialBuffers[source].length > maxBufferLength) {
```
pelikhan: What is the intent of this regex?

webapp/src/app.tsx
```diff
@@ -1913,6 +1913,8 @@ function initSerial() {
 
     pxt.debug('initializing serial pipe');
     let ws = new WebSocket(`ws://localhost:${pxt.options.wsPort}/${Cloud.localToken}/serial`);
+    let serialBuffers: {[source: string]: string} = {}
```
pelikhan: Use pxt.Map

webapp/src/app.tsx
```diff
@@ -1921,9 +1923,24 @@ function initSerial() {
     }
     ws.onmessage = (ev) => {
         try {
-            let msg = JSON.parse(ev.data) as pxsim.SimulatorMessage;
-            if (msg && msg.type == 'serial')
-                window.postMessage(msg, "*")
+            let msg = JSON.parse(ev.data) as pxsim.SimulatorSerialMessage;
+            if (msg && msg.type == 'serial') {
```
pelikhan: Can we move this logic into a common helper function?

caitlin-hennessy: Sure.  Where can it go, such that it will be accessible by app.tsx and pxtwinrt/serial.ts?

pxtwinrt/serial.ts
caitlin-hennessy: We want to post a chunk of serial data when either (a) we hit a newline of (b) the buffer is full.  This regex tests for case (a).

webapp/src/app.tsx
caitlin-hennessy: Done.

pxtwinrt/serial.ts
pelikhan: Since you are adding the last character, ``char`` is your last letter in the buffer.

pxtwinrt/serial.ts
```diff
@@ -56,13 +56,25 @@ namespace pxt.winrt {
...
+        let serialBuffers: pxt.Map<string> = {};
+        const maxBufferLength = 255;
         let readMore = () => reader.loadAsync(32).done((bytesRead) => {
             let msg = reader.readString(Math.floor(bytesRead / 4) * 4);
-            window.postMessage({
-                type: 'serial',
-                data: msg,
-                id: id
-            }, "*");
+            const data = msg || ""
```
pelikhan: this code should be refactored into a helper

webapp/src/app.tsx
pelikhan: pxtlib is loaded in both cases (see https://github.com/Microsoft/pxt/blob/master/pxtwinrt/deploy.ts#L3). You can have a "serial" object in there. THere might be something already.

webapp/src/serial.tsx
```diff
@@ -75,32 +73,21 @@ export class Editor extends srceditor.Editor {
 
         const data = smsg.data || ""
         const source = smsg.id || "?"
-        let theme = source.split("-")[0] || "black"
```
pelikhan: how do you differentiate messages coming from different micro:bits?

webapp/src/serial.tsx
```diff
@@ -21,8 +21,6 @@ export class Editor extends srceditor.Editor {
     maxConsoleLineLength: number = 500
     maxConsoleEntries: number = 100
     active: boolean = true
-    rawDataBuffer: string = ""
-    maxBufferLength: number = 5000
```
pelikhan: any reason why we are dropping this?

webapp/src/serial.tsx
```diff
@@ -247,11 +230,6 @@ export class Editor extends srceditor.Editor {
                     _.modal('hide')
                     pxt.commands.browserDownloadAsync(this.entriesToCSV(), "data.csv", "text/csv")
                 })
-                _.find('#datasavetxtfile').click(() => {
-                    pxt.tickEvent("serial.dataExported.txt")
-                    _.modal('hide')
-                    pxt.commands.browserDownloadAsync(this.entriesToPlaintext(), "data.txt", "text/plain")
```
pelikhan: no more text output?

webapp/src/serial.tsx
caitlin-hennessy: It was only a placeholder until exportToCSV was implemented.  Now that you've done that, we don't need this.

webapp/src/serial.tsx
caitlin-hennessy: We don't need to support that now, but I assume the message IDs would be different.

webapp/src/serial.tsx
caitlin-hennessy: No, thankfully.

pxtwinrt/serial.ts
caitlin-hennessy: Oh yeah ... fixed :D

webapp/src/app.tsx
caitlin-hennessy: Great, will do.

caitlin-hennessy: Done.

PR #3291, opened 2017-10-20
simpler type signature of rgb2hsl

pelikhan:
A small (breaking) change to simplify neopixel sim.
@microsoftsam this will break chibi.


microsoftsam:
Tested chibi, seems to work fine..

PR #3294, opened 2017-10-20
Multi-chart support

pelikhan:
If a value is named ``aaa.bbb``, we collapse all values starting with ``aaa`` under the same graph.

multichart

- [ ] update docs (I could not find the docs on serial output)


pelikhan:
We need to finish up the CSV output support. I think single chart output goes away but we still need to be able to dump the entire thing.

pelikhan:
Needs a bit of skinning mostkly ready.

pelikhan:
@microsoftsam

pelikhan:
@caitlin-hennessy this PR is ready for mergin

caitlin-hennessy:
@pelikhan This looks great! Can you please make a PR for v0 too?

PR #3302, opened 2017-10-21
Translate the delete variable confirmation dialog

sgk:
This is to make it possible to translate this dialog.
image

Please do not forget to update the `strings.json` file soon.

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -2286,6 +2286,7 @@ namespace pxt.blocks {
         // Dropdown menu of variables_get
         msg.RENAME_VARIABLE = lf("Rename variable...");
         msg.DELETE_VARIABLE = lf("Delete the \"%1\" variable");
+        msg.DELETE_VARIABLE_CONFIRMATION = lf("Delete %1 uses of the \"%2\" variable?");
 
         // builtin variables_set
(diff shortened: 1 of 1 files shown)


pxtblocks/blocklyloader.ts
```diff
@@ -2286,6 +2286,7 @@ namespace pxt.blocks {
         // Dropdown menu of variables_get
         msg.RENAME_VARIABLE = lf("Rename variable...");
         msg.DELETE_VARIABLE = lf("Delete the \"%1\" variable");
+        msg.DELETE_VARIABLE_CONFIRMATION = lf("Delete %1 uses of the \"%2\" variable?");
```
microsoftsam: Thanks @sgk for the change. I know this is how Blockly words the confirmation, could we perhaps use this opportunity to come up with a better way to word this? I feel like the language used here is a little pro dev. @ganicke @pelikhan thoughts?

pelikhan: No opinion on this.

pelikhan: At this I propose to get it in. We need this fix in the next release iteraiton of the micro:bit.

ganicke: Something like: "The \"%2\" variable is used %1 times, delete it?"

PR #3307, opened 2017-10-23
Support populating dropdown values from namespace properties

microsoftsam:
Support properties on namespaces to be used as dropdown values in blocks (instead of Enums)

Reason for this change: 
Enums are limiting and cannot be extended by the target or other packages. An example of this is the showAnimation method in adafruit that accepts an enum as the animation ID. If we wanted to create packages that potentially add new animations, we couldn't extend the Enum to add more attributes as it's defined in one place, and TS doesn't support merging of enums.

This way we can define namespace properties like so: 
```
namespace LightAnimation {
      export const Rainbow = 0;
      export const Sparkle = 1;
      export const ColorCycle = 2;
}
```

and a target or package can add to that list by exporting more properties under that namespace. All the properties will get merged, and the block that used to only include the 5 or so animations that are built in, now can be expanded by the target or the package ecosystem to include more animations built by others. 

Each of these properties can be annotated just like enums with block names and blockImages, etc. 

@riknoll I'm a little unsure how to write tests for this, since the values defined on the namespace are only relevant when populating the dropdown field.

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -671,38 +671,58 @@ namespace pxt.blocks {
 
                 let isEnum = typeInfo && typeInfo.kind == pxtc.SymbolKind.Enum
                 let isFixed = typeInfo && !!typeInfo.attributes.fixedInstances
+                let nsConstants = fn.attributes.blockValuesNs && info.apis.byQName[fn.attributes.blockValuesNs]
+                let constantDropdowns = nsConstants && nsConstants.constants
                 let customField = (fn.attributes.paramFieldEditor && fn.attributes.paramFieldEditor[p]);
                 let fieldLabel = pr.name.charAt(0).toUpperCase() + pr.name.slice(1);
                 let fieldType = pr.type;
 
-                if (isEnum || isFixed) {
-                    const syms = Util.values(info.apis.byQName)
-                        .filter(e =>
-                            isEnum ? e.namespace == pr.type
-                                : (e.kind == pxtc.SymbolKind.Variable
-                                    && e.attributes.fixedInstance
-                                    && isSubtype(info.apis, e.retType, typeInfo.qName)))
-                    if (syms.length == 0) {
-                        console.error(`no instances of ${typeInfo.qName} found`)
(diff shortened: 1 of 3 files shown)


pelikhan:
So in the case of animations, these props would hold an animation instance? 
Where are the images picked from?
Do these properties get tree shaked properly?

microsoftsam:
Yeah that's the idea. Will have to investigate if they get tree shaken properly. Although that's independent, because we could instead store number values in those properties. (just like enums).

For images, we need cloud support, independent effort.

mmoskal:
I'm wondering if you could do the same with `fixedInstances` and `whenUsed` attribute. It will not work for numbers, but will for instances of classes, which is better from TypeScript perspective (you will not have people pass random ints). For example:

```typescript
namespace pins {
    //% fixedInstance whenUsed
    export const A7 = new AnalogPin(7);
    //% fixedInstance whenUsed
    export const A8 = new AnalogPin(8);
}
```

`whenUsed` is needed for proper tree-shaking.

From the sources it seems `fixedInstances` should apply to all arguments, not only `this`.

riknoll:
@microsoftsam we can add this to the Karma tests. I've been wanting to add blockly loader tests anyway.

riknoll:
Wouldn't that be overloading the meaning of `fixedInstance`? It's usually accompanied by a `shim=` annotation. The `whenUsed` should still work regardless.

microsoftsam:
pxt-adafruit change: https://github.com/Microsoft/pxt-common-packages/pull/142

mmoskal:
`fixedInstances` on a class means that when it occurs as parameter, it should have a dropdown, and `fixedInstance` marks the instances that should be in the dropdown.

The `shim` specifies how to access the instance which is orthogonal to how it's accessed in blocks.

pelikhan:
Note: these should be factories not constant. Animations have a state and need to be instatiated on demand.

pelikhan:
@microsoftsam  We are using fixedInstances now. I think we can close this PR.

microsoftsam:
Closing as this is not needed anymore

PR #3309, opened 2017-10-23
Support for many boards in a single target

mmoskal:
* support multiple core packages in target (but only one in a project)
* allow `boardDefnition` (including visuals) to be specified in `pxt.json` of the core package
* add special `namespace config` (and overriding `namespace userconfig`), where constants are compiled into the binary and can be accessed from C++ (or simulator)
* shims with arguments can now take more than 1 argument, and handle constants from `config` and `DAL` namespaces
* when `additionalFilePath` specified in `pxt.json`, the target `pxt.json` needs to only specify what's changed 
* add `dalDTS` config in `pxt.json` to specify what to include and exclude from `dal.d.ts`; make parsing of constants smarter
* fix `level` field in packages and thus remove delete button from packages included as deps (clicking the button had no effect)
* disable HID connection restart after connected once - it interferes with flashing


webapp/src/app.tsx
```diff
@@ -1991,14 +1991,19 @@ function setHidPollerDelay() {
 }
 
 function initSerial() {
+    startHidConnectionPoller();
```
microsoftsam: This causes an infinite loop here where initSerial is calling startHidConnectionpoller, and startHidConnectionpoller is calling initSerial

pelikhan: That seems like the intended behavior: every 5 seconds, scan devices which may have dissapeared or popped up.

We should move this stuff out of app.tsx and clean up to avoid confusion.

microsoftsam: Yeah intended behaviour is to scan every 5 seconds, but here's what happens: 
- initSerial calls startHid which sets an interval to call initSerial every 5 seconds. 
- every time that runs, it calls initSerial and then that sets a new interval which then runs every 5 seconds. 

You get the idea. Basically should only set the interval if hasn't already been set..

PR #3313, opened 2017-10-23
Theming for close-editor button

caitlin-hennessy:
Goes with https://github.com/Microsoft/pxt-adafruit/pull/413

### DIFF
--- a/theme/themes/pxt/globals/site.variables
+++ b/theme/themes/pxt/globals/site.variables
@@ -19,6 +19,7 @@
 @pink:#E2008C;
 @brown:#00BCF2;
 @grey:#505050;
+@lightGrey: #fefefe;
 
 @greenTextColor: darken(@green, 5);
 
@@ -158,6 +159,17 @@
 
 @blocklyToolboxColor: rgba(0, 0, 0, 0.05);
 @trashIconColor: @primaryColor;
+
+/*-------------------
+   Serial editor
+--------------------*/
+@serialBackgroundColor: lighten(@simulatorBackground, 2);
+@serialTextColor: white;
+@serialGraphBackground: @lightGrey;
+@serialGraphColor: black;
+@closeColor: black;
+@closeColorHover: darken(@closeColor, 25%);
+@borderSize: 2px;
 
 /*-------------------
    Blockly
--- a/theme/themes/pxt/globals/site.variables
+++ b/theme/themes/pxt/globals/site.variables
@@ -164,12 +164,12 @@
    Serial editor
 --------------------*/
 @serialBackgroundColor: lighten(@simulatorBackground, 2);
-@serialTextColor: @darkTextColor;
+@serialTextColor: @invertedTextColor;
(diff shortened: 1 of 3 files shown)


theme/themes/pxt/globals/site.variables
```diff
@@ -160,6 +161,17 @@
 @trashIconColor: @primaryColor;
 
 /*-------------------
+   Serial editor
+--------------------*/
+@serialBackgroundColor: lighten(@simulatorBackground, 2);
+@serialTextColor: white;
+@serialGraphBackground: @lightGrey;
+@serialGraphColor: black;
+@closeColor: black;
```
microsoftsam: rename to editorCloseColor. This will conflict with semantic's close color here.

theme/themes/pxt/globals/site.variables
```diff
@@ -160,6 +161,17 @@
...
+@closeColorHover: darken(@closeColor, 25%);
```
microsoftsam: same as above.

theme/themes/pxt/globals/site.variables
```diff
@@ -160,6 +161,17 @@
...
+@borderSize: 2px;
+
```
microsoftsam: same here.

webapp/src/serial.tsx
```diff
@@ -223,10 +223,10 @@ export class Editor extends srceditor.Editor {
                         </div>
                     </div>
                     <div className="rightHeader">
-                        <sui.Button class="ui icon circular small inverted button" ariaLabel={lf("Close")} onClick={this.goBack.bind(this)}>
-                            <sui.Icon icon="close" />
+                        <sui.Button class="ui icon circular small button closeEditor" ariaLabel={lf("Close")} onClick={this.goBack.bind(this)}>
```
microsoftsam: Should this be over to the left? Makes more sense to go back that way. Especially if you're using the arrow left icon.

theme/themes/pxt/globals/site.variables
```diff
@@ -160,6 +161,17 @@
 @trashIconColor: @primaryColor;
 
 /*-------------------
+   Serial editor
+--------------------*/
+@serialBackgroundColor: lighten(@simulatorBackground, 2);
+@serialTextColor: white;
+@serialGraphBackground: @lightGrey;
```
microsoftsam: use @darkTextColor instead of white.

webapp/src/serial.tsx
caitlin-hennessy: Just to the left of the download button?  Or *all* the way to the left, by the play/pause button?

microsoftsam: All the way to the left. Consistent with most back buttons

theme/themes/pxt/globals/site.variables
caitlin-hennessy: Doesn't look like @darktextcolor is defined?

microsoftsam:
Will merge once build passes.

PR #3326, opened 2017-10-24
Fix HID/serial re-connecting

### LINKED ISSUE #3325: Hundres of hidbridge failed to load messages
Hundreds of hidbridge failed to load, Error: No device connected console messages. 
Why is it not getting debounced?

```
hidbridge failed to load, Error: No device connected
51pxtlib.js:1398 hidbridge failed to load, Error: No device connected
pxtlib.js:1398 hidbridge failed to load, Error: No device connected
18pxtlib.js:1398 hidbridge failed to load, Error: No device connected
pxtlib.js:1398 hidbridge failed to load, Error: No device connected
47pxtlib.js:1398 hidbridge failed to load, Error: No device connected
```

mmoskal:
Fixes #3325

It also adds a loop that pings the device every few seconds, and only when the device fails to respond (most likely disconnected), it tries to re-connect.

### DIFF
--- a/pxtlib/hf2.ts
+++ b/pxtlib/hf2.ts
@@ -406,6 +406,13 @@ namespace pxt.HF2 {
             return this.talkAsync(HF2_CMD_READ_WORDS, args)
         }
 
+        pingAsync() {
+            if (this.rawMode)
+                return Promise.resolve()
+            return this.talkAsync(HF2_CMD_BININFO)
+                .then(buf => { })
+        }
+
         flashAsync(blocks: pxtc.UF2.Block[]) {
             let start = Date.now()
             let fstart = 0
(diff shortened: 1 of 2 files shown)


abchatra:
Does this needs to be in v0 as well?

pelikhan:
Probably since those changes have been integrated down.

abchatra:
@caitlin-hennessy can you port these to v0?

abchatra:
@mmoskal Can you port these? Might be easier for you as you did the original change.

caitlin-hennessy:
Thanks @mmoskal.  To clarify: I don't think these changes will affect microbit, since microbit doesn't actually use the initSerial() method in app.tsx.  However, it seems like they should go into v0 anyway for code consistency.

PR #3337, opened 2017-10-25
Remove 'at' directives from header text

ganicke:
- [x] Filter `@` directives just like routing tags `#` so they won't render.
- [x] Prefix heading text in `id=` tags for a complete ID. Will remove this is if it breaks something needed post render.

Fixes https://github.com/Microsoft/pxt-minecraft/issues/486

### DIFF
--- a/pxtlib/docsrender.ts
+++ b/pxtlib/docsrender.ts
@@ -436,11 +436,11 @@ namespace pxt.docs {
                 return '<li>' + text + '</li>\n';
             }
             renderer.heading = function (text: string, level: number, raw: string) {
-                let m = /(.*)#([\w\-]+)\s*$/.exec(text)
+                let m = /(.*)[#@]([\w\-]+)\s*$/.exec(text)
                 let id = ""
                 if (m) {
                     text = m[1]
-                    id = m[2]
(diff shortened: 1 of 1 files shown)


pelikhan:
This change will have to be imported in the backend.

pxtlib/docsrender.ts
```diff
@@ -436,11 +436,11 @@ namespace pxt.docs {
                 return '<li>' + text + '</li>\n';
             }
             renderer.heading = function (text: string, level: number, raw: string) {
-                let m = /(.*)#([\w\-]+)\s*$/.exec(text)
+                let m = /(.*)[#@]([\w\-]+)\s*$/.exec(text)
```
pelikhan: I don't think @fullscreen is meant to be a anchor ID. Can you explain your change?

ganicke: Just want ``@fullscreen`` to not show in doc view of tutorial page.

Currently it's like this:

```
<h2 id="introduction-fullscreen">Introduction @fullscreen</h2>
```
image

Is there a better solution?

ganicke: Oh, you mean the anchor ID stuff, sorry. I just noticed that the ID's of the routing tags are single work but now I get your point. I'll remove that concatenation.

PR #3338, opened 2017-10-25
AVR compiler fixes

mmoskal:


mmoskal:
This should be merged before it bitrots. Review anyone?

PR #3339, opened 2017-10-26
Close HF2 connection before reloading window

caitlin-hennessy:


webapp/src/app.tsx
```diff
@@ -1029,11 +1029,24 @@ export class ProjectView
...
-            if (!r) return;
-            this.reload = true; //Indicate we are goint to reload next.
-            workspace.resetAsync()
-                .done(() => window.location.reload(),
-                () => window.location.reload())
+            if (!r) return Promise.resolve();
+            if (hf2Connection) {
+                return hf2Connection.disconnectAsync()
+                .then(() => {
+                    this.reload = true; //Indicate we are going to reload next.
```
riknoll: Move this code block into a function so that it doesn't need to be duplicated below

abchatra: +1

caitlin-hennessy:
@riknoll Look better now?

riknoll:
What's with the commit from @microsoftsam?

pelikhan:
Isn't this fix required for Adafruit?

caitlin-hennessy:
@peli This PR should be deleted, it was redone here https://github.com/Microsoft/pxt/pull/3353

PR #3351, opened 2017-10-26
Close HF2 connection before reloading window (Cherry-pick pxt/pull/3339)

caitlin-hennessy:


webapp/src/app.tsx
```diff
@@ -939,6 +939,15 @@ export class ProjectView
         }
     }
 
+    resetWorkspace() {
+        this.reload = true;
+        return workspace.resetAsync()
```
pelikhan: you do not have to return anything here

caitlin-hennessy: Oh cool, fixed.

PR #3353, opened 2017-10-26
Clean PR for 'close hf2 before reloading'

caitlin-hennessy:


webapp/src/app.tsx
```diff
@@ -1029,11 +1038,13 @@ export class ProjectView
...
         }).then(r => {
-            if (!r) return;
-            this.reload = true; //Indicate we are goint to reload next.
-            workspace.resetAsync()
-                .done(() => window.location.reload(),
-                () => window.location.reload())
+            if (!r) return Promise.resolve();
+            if (hf2Connection) {
+                return hf2Connection.disconnectAsync()
+                .then(this.resetWorkspace.bind(this))
```
guillaumejenkins: You can use the arrow notation, `.then(() => this.resetWorkspace())`, to let the TypeScript compiler automatically set the correct `this` (notice the parentheses after `resetWorkspace()`)

Not required at all, but I find it more convenient than to bind it myself :)

Issue #3378, opened 2017-10-28
assertion fails (delayLoadXml)


microsoftsam:
The delayLoadXml assertion in loadEditorFile in blocks.tsx fails everytime you load a project that has instructions (ie: not a tutorial or an example). 

@pelikhan any thoughts?

Repro: makecode.adafruit.com/beta
Load the Magic wand project

microsoftsam:
Looks like loadFileAsync in that case is getting called twice..

microsoftsam:
Probably because the sidedocs is causing the render function to get called a bunch of times, we really need to optimize that side doc. I'll look into it.

pelikhan:
Good use of an assert :)

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #3380, opened 2017-10-28
Home screen perf improvements

microsoftsam:
Improve home screen performance by making it not a modal. Shying away from having to handle the home screen state all over the place. The App is now in charge of the home screen state.
The result is fantastic, the home screen loads right away with no weird transition in between.

Two modes. A user is either in home, or in the editor. 
The only way to switch to editor mode is to load a header (loadHeaderAsync). (so for card actions that open urls, or when errors occur, the editor is not loaded). This avoids having to constantly show home when an error occurs anywhere.
When a header is loaded, the url hash changes to #editor (so if a user reloads, they go back to their previous project). 
If you go back to home, the hash is cleared. 

The import dialogs are now part of home, and so they show up as modals on top of the home screen. Rather than on top of the editor (previously)

Other entry points: (Tested)
#pub, loads the project, clears the hash, and then the editor hash is shown since a project is loaded
#tutorial, same as above
#newproject, and #newjavascript, create a new project

Haven't tested: #doc, and #follow. 
I've added comments on other hash shortcut but I'm not familiar with these two.

The simulator, and the various editors are still prepared in the background, ready for a project to be loaded. 
I stop the simulator when returning back to home
Reduces the two setStates in loadHeaderAsync to one, perf reasons
Fixed an issue with the editor loading a file twice, if render is called consecutively (added a lock to only load one file at a time) Fixes https://github.com/Microsoft/pxt/issues/3378
Fixes https://github.com/Microsoft/pxt-chibitronics/issues/112
Cleaned up more leftover code from the previous projects dialog.

Targets that use the css rule .ui.modal.home or .ui.modal.projectsdialog will need to update their CSS to use .ui.home.projectsdialog

@guillaumejenkins @mmoskal @pelikhan please review, this is a big change to how we load things. 

Still needs to be tested: 
- [ ] UWP / Electron app
- [ ] Reload logic
- [ ] #doc and #follow
- [x] session outdated (app::updateVisibility)

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -642,7 +642,7 @@ export class ProjectView
         if (!h)
             return Promise.resolve()
 
-        window.location.hash = "#editor";
+        pxt.BrowserUtils.changeHash("#editor", true);
         this.stopSimulator(true);
         pxt.blocks.cleanBlocks();
         this.clearSerial()
@@ -889,7 +889,7 @@ export class ProjectView
         this.stopSimulator();
         this.setState({ home: true });
         // clear the hash
-        window.location.hash = "";
+        pxt.BrowserUtils.changeHash("", true);
     }
 
     exitAndSave() {
@@ -2255,7 +2255,7 @@ function handleHash(hash: { cmd: string; arg: string }, loading: boolean): boole
         case "newproject": // shortcut to create a new blocks proj
             pxt.tickEvent("hash.newproject")
             editor.newProject();
-            window.location.hash = "";
+            pxt.BrowserUtils.changeHash("");
             return true;
         case "newjavascript": // shortcut to create a new JS proj
             pxt.tickEvent("hash.newjavascript");
@@ -2265,36 +2265,41 @@ function handleHash(hash: { cmd: string; arg: string }, loading: boolean): boole
                     "main.blocks": ""
                 }
             });
-            window.location.hash = "";
+            pxt.BrowserUtils.changeHash("");
+            return true;
+        case "gettingstarted":
+            pxt.tickEvent("hash.gettingstarted");
+            editor.newProject();
+            pxt.BrowserUtils.changeHash("");
             return true;
         case "tutorial": // shortcut to a tutorial. eg: #tutorial:tutorials/getting-started
             pxt.tickEvent("hash.tutorial")
             editor.startTutorial(hash.arg);
-            window.location.hash = "";
+            pxt.BrowserUtils.changeHash("");
             return true;
         case "home": // shortcut to home
             pxt.tickEvent("hash.home");
             editor.openHome();
-            window.location.hash = "";
+            pxt.BrowserUtils.changeHash("");
             return true;
         case "sandbox":
         case "pub":
         case "edit": // load a published proj, eg: #pub:27750-32291-62442-22749
             pxt.tickEvent("hash." + hash.cmd);
-            window.location.hash = "";
+            pxt.BrowserUtils.changeHash("");
             loadHeaderBySharedId(hash.arg);
             return true;
         case "sandboxproject":
         case "project":
             pxt.tickEvent("hash." + hash.cmd);
             const fileContents = Util.stringToUint8Array(atob(hash.arg));
-            window.location.hash = "";
+            pxt.BrowserUtils.changeHash("");
             core.showLoading("loadingproject", lf("loading project..."));
             theEditor.importProjectFromFileAsync(fileContents)
                 .done(() => core.hideLoading("loadingheader"));
             return true;
         case "reload": // need to reload last project - handled later in the load process
-            if (loading) window.location.hash = "";
+            if (loading) pxt.BrowserUtils.changeHash("");
             return false;
     }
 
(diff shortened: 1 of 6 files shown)


microsoftsam:
I wasn't sure what to do about the updateVisibility sessionOutdated state, so just commented it out for now

webapp/src/app.tsx
```diff
@@ -2248,24 +2267,19 @@ function handleHash(hash: { cmd: string; arg: string }, loading: boolean): boole
             });
             window.location.hash = "";
             return true;
-        case "gettingstarted":
```
mmoskal: There might be external links to `#gettingstarted`, and also it has a separate tick, maybe we need it?

webapp/src/app.tsx
```diff
@@ -2248,24 +2267,19 @@ function handleHash(hash: { cmd: string; arg: string }, loading: boolean): boole
...
-            pxt.tickEvent("hash.gettingstarted")
-            editor.newProject();
-            window.location.hash = "";
-            return true;
-        case "tutorial":
+        case "tutorial": // shortcut to a tutorial. eg: #tutorial:tutorials/getting-started
             pxt.tickEvent("hash.tutorial")
             editor.startTutorial(hash.arg);
             window.location.hash = "";
```
mmoskal: I think in all these places we should be using `Location.replace`, so it won't store the entry in browser history.

(except when the user actually clicks on something to go from home to editor or back - then `Location.assign` so history entry is stored)

webapp/src/app.tsx
microsoftsam: Yeah looking at the history, even though this does the same thing as newproject. I'll keep it.

microsoftsam: Done.

webapp/src/app.tsx
microsoftsam: I went with replaceState instead of replace, and kept using location.hash = "" to make sure we keep an entry stored. Any reason to use replace and assign?

microsoftsam:
What actions do we care about the back button. If the action is associated with a hash change then window.location.hash = ... takes care of the history entry. 
Right now, I've made going from editor -> home and home -> editor place a history entry. 
Would we want to include actions like going from JS to blocks and blocks to JS as a history action. Maybe opening the settings editor. 

And do we want to use the hash as an indication of these states like react-router?

pelikhan:
Yes switching editor should be a history action and that’s probably enough.

Back in TD, this stuff had a lots of weird discrepancies across browsers. Maybe things got fixed by now.

microsoftsam:
@pelikhan thoughts on #doc and #follow. how would I test?

microsoftsam:
and what should we when the session is outdated, show home?

pelikhan:
Sounds good.

webapp/src/app.tsx
```diff
@@ -114,15 +114,27 @@ export class ProjectView
...
         if (!this.settings.fileHistory) this.settings.fileHistory = [];
     }
 
+    shouldShowHomeScreen() {
+        const hash = parseHash();
+        const isSandbox = pxt.shell.isSandboxMode() || pxt.shell.isReadOnly();
+        // Only show the start screen if there are no initial projects requested
+        // (e.g. from the URL hash or from WinRT activation arguments)
+        const skipStartScreen = pxt.appTarget.appTheme.allowParentController
+            || !pxt.appTarget.appTheme.showHomeScreen || window.location.hash == "#editor";
```
pelikhan: showHomeScreen should be our default setting now.

webapp/src/app.tsx
```diff
@@ -137,8 +149,8 @@ export class ProjectView
             if (workspace.isSessionOutdated()) {
                 pxt.debug('workspace changed, reloading...')
                 let id = this.state.header ? this.state.header.id : '';
-                workspace.initAsync()
-                    .done(() => id ? this.loadHeaderAsync(workspace.getHeader(id)) : Promise.resolve());
+                // workspace.initAsync()
```
pelikhan: go to home?

webapp/src/app.tsx
```diff
@@ -869,7 +886,10 @@ export class ProjectView
 
     openHome() {
         pxt.tickEvent("menu.open");
-        this.home.showHome();
+        this.stopSimulator();
+        this.setState({ home: true });
+        // clear the hash
+        pxt.BrowserUtils.changeHash("", true);
```
pelikhan: we should change the hash, then the onhashchanged handler should actually do the state transition

webapp/src/app.tsx
```diff
@@ -1736,6 +1751,8 @@ ${compileService && compileService.githubCorePackage && compileService.gittag ?
         const sideDocs = !(sandbox || targetTheme.hideSideDocs);
         const tutorialOptions = this.state.tutorialOptions;
         const inTutorial = !!tutorialOptions && !!tutorialOptions.tutorial;
+        const inHome = this.state.home && !sandbox;
+        const inEditor = this.state.header;
```
pelikhan: ``!!this.state.header``

webapp/src/app.tsx
microsoftsam: Done.

Issue #3271, opened 2017-10-18
Accessibility menu appears on home screen. 


veler:
### Issue

The "Skip to JavaScript editor" hidden menu appears in the Home view of Minecraft.

### Environment

Windows 10 Fall Creator Update, MS Edge.

### Reproduce

1. Go to https://minecraft.makecode.com/
2. "New project" has the focus.
3. Let's imagine that while the page was loading, the user was on another app/window.
4. He clicks back to the web browser window, in the makecode page. Because of this click, the focus on "New project" is lost.
5. Press Tab, and the hidden menu will appear.

![capture](https://user-images.githubusercontent.com/3747805/31706115-237274a2-b3e8-11e7-884c-9811b41bc19d.PNG)

PR #3389, opened 2017-10-31
Refactor app.tsx into smaller React modules

microsoftsam:
Refactoring the app.tsx codebase into many different react modules each in charge of their own state and properties (the way we should be using react).

This PR is tied in with another small feature to add a detail view in the home screen for better browsing. 

There are also a number of small accessibility fixes around the accessibility menu in the home screen (to show a different one that is representative of the actions a user can take on the home screen).

This PR requires a bunch of updates to our targets. I will prepare a number of target PRs.

I've started optimizing each component to only re-render if a state that it cares about has changed, but there's probably more to be optimized.

Fixes https://github.com/Microsoft/pxt/issues/3271, Fix cookie close icon


webapp/src/projects.tsx
```diff
@@ -356,6 +422,70 @@ export class ProjectsCarousel extends data.Component<ProjectsCarouselProps, Proj
...
 }
 
+export interface ProjectsDetailProps extends ISettingsProps {
+    name: string;
+    description?: string;
+    url?: string;
+    imageUrl?: string;
+    largeImageUrl?: string;
+    youTubeId?: string;
+    onClick: Function;
```
riknoll: We typically don't use the `Function` type. Maybe `() => void`?

pelikhan:
Add a warning comment on every state class that implements a custom "shouldUpdate" function that it should also be updated.

webapp/src/accessibility.tsx
```diff
@@ -0,0 +1,118 @@
...
+    componentWillReceiveProps(nextProps: EditorAccessibilityMenuProps) {
+        const newState: EditorAccessibilityMenuState = {};
+        if (nextProps.highContrast != undefined) {
+            newState.highContrast = nextProps.highContrast;
+        }
+        if (Object.keys(newState).length > 0) this.setState(newState)
+    }
+
+    shouldComponentUpdate(nextProps: EditorAccessibilityMenuProps, nextState: EditorAccessibilityMenuState, nextContext: any): boolean {
+        return this.state.highContrast != nextState.highContrast;
```
pelikhan: isn't React supposed to do this already?

webapp/src/accessibility.tsx
```diff
@@ -0,0 +1,118 @@
...
+    componentWillReceiveProps(nextProps: HomeAccessibilityMenuProps) {
+        const newState: HomeAccessibilityMenuState = {};
+        if (nextProps.highContrast != undefined) {
+            newState.highContrast = nextProps.highContrast;
+        }
+        if (Object.keys(newState).length > 0) this.setState(newState)
+    }
+
+    shouldComponentUpdate(nextProps: HomeAccessibilityMenuProps, nextState: HomeAccessibilityMenuState, nextContext: any): boolean {
+        return this.state.highContrast != nextState.highContrast;
```
pelikhan: isn't React handling this already?

webapp/src/accessibility.tsx
microsoftsam: Not the way we're using it. Since we have basically one AppState. Every time we call setState it calls a render on all components and their child components.
The way react is meant to be used is to have separate components each maintaining their own state. 
React still gives you the opportunity to override it's shouldComponentUpdate and have your own logic (as above) to when it should or shouldn't update the state, thus limiting the updates in our case.

Issue #3364, opened 2017-10-27
Headings in CSV don't make sense


Jaqster:
"time(s)" is first heading
"red-30078101486949966480.9176019446108943" is the second heading which doesn't make sense

Default heading should be "data 1", "data 2", etc. or something like that.

PR #3394, opened 2017-10-31
Map each data source name to a more readable name

caitlin-hennessy:
Fixes https://github.com/Microsoft/pxt/issues/3364 - CSV headings will be "source1", "source2", etc.

### DIFF
--- a/webapp/src/serial.tsx
+++ b/webapp/src/serial.tsx
@@ -16,7 +16,6 @@ const lf = Util.lf
 export class Editor extends srceditor.Editor {
     charts: Chart[] = []
     chartIdx: number = 0
-    sourceIdx: number = 1
     sourceMap: pxt.Map<string> = {}
     consoleBuffer: string = ""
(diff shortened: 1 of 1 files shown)


webapp/src/serial.tsx
```diff
@@ -74,12 +76,17 @@ export class Editor extends srceditor.Editor {
         const data = smsg.data || ""
         const source = smsg.id || "?"
 
+        if (!(source in this.sourceMap)) {
+            this.sourceMap[source] = lf("source") + (this.sourceIdx++).toString()
```
pelikhan: the app theme has a "boardName" variable that would read good (e.g. micro:bit ...). Typically, we simply do ``!!this.sourceMap[source]`` for lookups.
sourceIdx = Object.keys(this.sourceMap).length

caitlin-hennessy:
@pelikhan I made the changes we discussed.

PR #3401, opened 2017-11-01
moving refcounting in runtime to recycle between runs

pelikhan:
We were leaking refcounted objects in the simulator. Moving ref counting in the runtime state.

### DIFF
--- a/pxtsim/embed.ts
+++ b/pxtsim/embed.ts
@@ -6,6 +6,7 @@ namespace pxsim {
         id?: string;
         boardDefinition?: BoardDefinition;
         frameCounter?: number;
+        refCountingDebug?: boolean;
         options?: any;
         parts?: string[];
         partDefinitions?: Map<PartDefinition>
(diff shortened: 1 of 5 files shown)


pelikhan:
This fix might need to go in v0 too.

mmoskal:
Looks good!

microsoftsam:
So this breaks the simulators in SIM only targets.

VM51526:1 Uncaught TypeError: pxsim.noRefCounting is not a function
    at eval (eval at Runtime (pxtsim.js:4220), <anonymous>:1:7)
    at new Runtime (pxtsim.js:4220)
    at run (pxtsim.js:1928)
    at receiveMessage (pxtsim.js:1886)

There's a dependency on pxsim.noRefCounting as emitted by the emitter: https://github.com/Microsoft/pxt/blob/a5c9cf0acf8a4d369c1f68d4bc53a86e8605df42/pxtcompiler/emitter/backjs.ts#L67

pelikhan:
Can you file it on me?

microsoftsam:
I can do better, is this the right fix? https://github.com/Microsoft/pxt/pull/3506

PR #3404, opened 2017-11-01
Support for jres resources in blocks

pelikhan:
Support for using iconURL if available in block drop downs.

image


microsoftsam:
Looks good.

PR #3410, opened 2017-11-02
Error message update from upload to download

abchatra:
Fixes https://github.com/Microsoft/pxt-adafruit/issues/423

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1148,7 +1148,7 @@ export class ProjectView
                         if (e.notifyUser) {
                             core.warningNotification(e.message);
                         } else {
-                            core.warningNotification(lf("Upload failed, please try again."));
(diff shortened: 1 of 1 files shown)


webapp/src/app.tsx
```diff
@@ -1148,7 +1148,8 @@ export class ProjectView
                         if (e.notifyUser) {
                             core.warningNotification(e.message);
                         } else {
-                            core.warningNotification(lf("Upload failed, please try again."));
+                            const errorText = pxt.appTarget.appTheme.useUploadMessage ? lf("Upload failed, please try again.") : lf("Download failed, please try again.");
```
microsoftsam: Can appTheme be null?

pelikhan: It can be null in the CLI but not in the app itself.

PR #3419, opened 2017-11-03
Crowdin stats command

pelikhan:
A CLI command to compute Crowdin progress


pelikhan:
It dumps a CSV file. That works for PowerBI?

abchatra:
Good for now. Powerbi will consume from excel, but will need manual update every time we want to run it. I will see if there are other integration options.

PR #3425, opened 2017-11-06
Adding seeed target to the card list

microsoftsam:
screen shot 2017-11-06 at 11 25 38 am

screen shot 2017-11-06 at 11 31 00 am

Tested on: 
- [x] Edge 16
- [x] IE
- [x] Firefox
- [x] Chrome Mac
- [x] Chrome Windows


microsoftsam:
@Jaqster when you get a moment, waiting for your Approval.

Jaqster:
Approved for Beta.  But not a huge fan of the image - maybe we can switch it out for something that looks a bit more polished for RC.

microsoftsam:
That's the image they supplied, let's give them our feedback and ask for another.

Jaqster:
I don't think we can hold off posting the card any longer.  Let's publish, and aim for a nicer image next time.  Also, what's the logic behind the ordering of cards.  We need to establish a principle early, otherwise we'll piss off partners.  I kind of like the simplicity of in the order that they onboarded...

microsoftsam:
Here's a proposal for a process: 
- Three categories of editors: Non-beta (RC editor), Beta targets, app targets (eg: WW)
- Ordered in the order of which they were onboarded, with non-beta categories first, beta targets next, and then finally app targets. 

This means: 
- microbit, adafruit, minecraft, chibi
- sparkfun, grove zero
- WW

microsoftsam:
Merging for now, will discuss in person later.

PR #3427, opened 2017-11-06
Open docs in new tab

microsoftsam:
Open docs in new page rather than the side docs.

### DIFF
--- a/webapp/src/projects.tsx
+++ b/webapp/src/projects.tsx
@@ -152,8 +152,11 @@ export class Projects extends data.Component<ISettingsProps, ProjectsState> {
                             window.open('https://youtu.be/' + scr.youTubeId, 'yt');
                         else if (/^https:\/\//i.test(scr.url)) // External video
                             window.open(scr.url, '_blank');
-                        else if (/^\//i.test(scr.url)) // Docs url, open in new tab
-                            window.open(scr.url, '_blank');
+                        else if (scr.url) // Docs url, open in new tab
+                            if (/^\//i.test(scr.url))
+                                window.open(scr.url, '_blank');
+                            else
+                                core.errorNotification(lf("Sorry, the project url looks invalid."));
                         else
                             this.props.parent.newEmptyProject(scr.name.toLowerCase(), scr.url);
                     }
--- a/webapp/src/projects.tsx
+++ b/webapp/src/projects.tsx
@@ -152,7 +152,7 @@ export class Projects extends data.Component<ISettingsProps, ProjectsState> {
                             window.open('https://youtu.be/' + scr.youTubeId, 'yt');
                         else if (/^https:\/\//i.test(scr.url)) // External video
                             window.open(scr.url, '_blank');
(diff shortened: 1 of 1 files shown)


webapp/src/projects.tsx
```diff
@@ -145,12 +145,14 @@ export class Projects extends data.Component<ISettingsProps, ProjectsState> {
...
                 default:
-                    const m = /^\/#tutorial:([a-z0A-Z0-9\-\/]+)$/.exec(scr.url);
+                    const m = /^\/#tutorial:([a-z0A-Z0-9\-\/]+)$/.exec(scr.url); // Tutorial
                     if (m) this.props.parent.startTutorial(m[1]);
                     else {
-                        if (scr.youTubeId && !scr.url)
+                        if (scr.youTubeId && !scr.url) // Youtube video
                             window.open('https://youtu.be/' + scr.youTubeId, 'yt');
-                        else if (/^https:\/\//i.test(scr.url))
+                        else if (/^https:\/\//i.test(scr.url)) // External video
```
pelikhan: both paths do the same... we should allow local urls (start with / or https urls only)

webapp/src/projects.tsx
```diff
@@ -152,7 +152,7 @@ export class Projects extends data.Component<ISettingsProps, ProjectsState> {
                             window.open('https://youtu.be/' + scr.youTubeId, 'yt');
                         else if (/^https:\/\//i.test(scr.url)) // External video
                             window.open(scr.url, '_blank');
-                        else if (scr.url) // Docs url, open in new tab
+                        else if (/^\//i.test(scr.url)) // Docs url, open in new tab
                             window.open(scr.url, '_blank');
                         else
                             this.props.parent.newEmptyProject(scr.name.toLowerCase(), scr.url);
```
pelikhan: probably just want to show a warning and ignore the URL.

webapp/src/projects.tsx
microsoftsam: Fixed

webapp/src/projects.tsx
microsoftsam: Cool. Done

microsoftsam:
@pelikhan ping for another look

webapp/src/projects.tsx
```diff
@@ -152,8 +152,11 @@ export class Projects extends data.Component<ISettingsProps, ProjectsState> {
...
                             window.open(scr.url, '_blank');
-                        else if (/^\//i.test(scr.url)) // Docs url, open in new tab
-                            window.open(scr.url, '_blank');
+                        else if (scr.url) // Docs url, open in new tab
+                            if (/^\//i.test(scr.url))
+                                window.open(scr.url, '_blank');
+                            else
+                                core.errorNotification(lf("Sorry, the project url looks invalid."));
                         else
                             this.props.parent.newEmptyProject(scr.name.toLowerCase(), scr.url);
```
pelikhan: scr.url should always be undefined here

PR #3433, opened 2017-11-07
Home / Editor tick event

microsoftsam:
Add two new events for when the home is loaded, and when the editor is loaded.


microsoftsam:
All Code paths for the editor go through loadHeaderAsync. I need to double check the openHome.

PR #3436, opened 2017-11-07
Add a block translation tip section

ganicke:
Add a section to the translate page to help the translators test their block strings. This might reduce some of the questions related to the format and wording...maybe.


ganicke:
Probably, yes. Although I just did it on the hosted microbit target:

image

ganicke:
We should include a lightweight dummy target to the PXT, makecode.com so "live" example blocks are included in the dev docs. Also, how about a "Block Tool" page where you can design the look of a block by messing with the attribute fields to see what the block might look like: https://makecode.com/blocktool

microsoftsam:
Yes, a block tool is a great idea! 
We can have something similar to the Blockly developer tools, that creates blocks based on the function definition and block attributes.

PR #3438, opened 2017-11-08
rename data -> console

pelikhan:


webapp/src/serialindicator.tsx
```diff
@@ -39,11 +39,11 @@ export class SerialIndicator extends React.Component<SerialIndicatorProps, Seria
     render() {
         if (!this.state.active) return <div />;
         return(
-            <div title={lf("Open data viewer")} className="ui label circular" tabIndex={0} onClick={this.props.onClick} onKeyDown={sui.fireClickOnEnter}>
+            <div title={lf("Open Console")} className="ui label circular" tabIndex={0} onClick={this.props.onClick} onKeyDown={sui.fireClickOnEnter}>
                 <div className="detail">
                     <sui.Icon icon="bar graph"/>
                 </div>
-                {lf("Show data") }
+                {lf("Show console") }
```
abchatra: C in Console should be in caps?

pelikhan: to caps or not to caps

pelikhan: I've removed all caps, do you want caps?

PR #3440, opened 2017-11-08
improvement to serial view

pelikhan:
A number of improvements to make the charts more useful when looking at data:

- [x] a vertical grid line every 5 sec
- [x] better skinning of data tooltip
- [x] min/max chart values

image

### DIFF
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -91,6 +91,7 @@ declare namespace pxt {
 
     interface SerialTheme {
         graphBackground?: string;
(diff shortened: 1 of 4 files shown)


cait1inhennessy:
Looks good. My only thought is maybe we should use a single font for the variable label, graph labels, and tooltip. Seems a bit weird to have the graph label use sans-serif while the other two use monospace.

abchatra:
Merge to v0 as well.

abchatra:
@pelikhan port to v0 please. These are good changes

PR #3447, opened 2017-11-09
localized strings cannot be used before the language is loaded

pelikhan:
Caching happens too early for exported variables. This impacts localization of "Advanced" and "Extensions" in the editor.


microsoftsam:
Was just about to mention that moving it to the local context would break the monaco toolbox, but looks like you figured that out.

pelikhan:
Need to check if v0 has this problem.

PR #3480, opened 2017-11-15
Support inline block images from jres

microsoftsam:
Looks a bit like this: 

screen shot 2017-11-14 at 3 56 15 pm

Authored like so: 
- Add image base64 to a jres. 
- Reference the jres in a backtick to add the icon in the block. eg: 
```
block="teleport`__icons.agent`to player"
```


pxtblocks/blocklyloader.ts
```diff
@@ -3199,4 +3212,18 @@ namespace pxt.blocks {
...
     }
+
+    let iconCanvasCache: Map<string> = {};
+    function iconToFieldImage(id: string): Blockly.FieldImage {
+        let url = iconCanvasCache[id];
+        return new Blockly.FieldImage(url, 30, 30, false, '');
+    }
+
+    function initJresIcons(blockInfo: pxtc.BlocksInfo) {
+        Object.keys(blockInfo.apis.jres).forEach((jresId) => {
```
pelikhan: Why are you not doing this lazily?

microsoftsam: So I don't have to pass the blockInfo all the way down the stack to the initField function.

pxtblocks/blocklyloader.ts
```diff
@@ -585,7 +585,19 @@ namespace pxt.blocks {
     }
 
     function initField(i: any, ni: number, fn: pxtc.SymbolInfo, ns: pxtc.SymbolInfo, pre: string, right?: boolean, type?: string, nsinfo?: pxtc.SymbolInfo): any {
-        if (pre)
+        if (pre && pre.indexOf('`') > -1) {
+            // parse and create icon fields for every inline icon
+            let regex = /([^`]+|(`([^`]*)`))/gi;
+            let match: RegExpExecArray;
+            while (match = regex.exec(pre)) {
```
pelikhan: tip: replace is easier to go through matches of a regex

pxtblocks/blocklyloader.ts
```diff
@@ -3199,4 +3212,18 @@ namespace pxt.blocks {
             }
         }
     }
+
+    let iconCanvasCache: Map<string> = {};
+    function iconToFieldImage(id: string): Blockly.FieldImage {
+        let url = iconCanvasCache[id];
```
pelikhan: check that URL exists

pxtblocks/blocklyloader.ts
```diff
@@ -3199,4 +3212,18 @@ namespace pxt.blocks {
             }
         }
     }
+
+    let iconCanvasCache: Map<string> = {};
```
pelikhan: this name is quite misleading it should be jresIconCache

pxtblocks/blocklyloader.ts
```diff
@@ -3199,4 +3212,18 @@ namespace pxt.blocks {
...
+    function iconToFieldImage(id: string): Blockly.FieldImage {
+        let url = iconCanvasCache[id];
+        return new Blockly.FieldImage(url, 30, 30, false, '');
+    }
+
+    function initJresIcons(blockInfo: pxtc.BlocksInfo) {
+        Object.keys(blockInfo.jres).forEach((jresId) => {
```
pelikhan: should you clear iconCanvasCache to keep it from leaking icons between project reloads?

pxtblocks/blocklyloader.ts
```diff
@@ -585,7 +585,19 @@ namespace pxt.blocks {
     }
 
     function initField(i: any, ni: number, fn: pxtc.SymbolInfo, ns: pxtc.SymbolInfo, pre: string, right?: boolean, type?: string, nsinfo?: pxtc.SymbolInfo): any {
-        if (pre)
+        if (pre && pre.indexOf('`') > -1) {
+            // parse and create icon fields for every inline icon
+            let regex = /([^`]+|(`([^`]*)`))/gi;
```
pelikhan: what are you trying to match here? https://rise4fun.com/Rex/NY0

microsoftsam: text without a backticks, and text within a pair of backticks

pxtblocks/blocklyloader.ts
microsoftsam: I tired that but for some reason I couldn't use it. I'll look into it again

pxtblocks/blocklyloader.ts
microsoftsam: Yeah, legacy name, that I copied over. Will change

pxtblocks/blocklyloader.ts
microsoftsam: What if it doesn't exist? We need to return a Field image.

pxtblocks/blocklyloader.ts
microsoftsam: Yes.

pxtblocks/blocklyloader.ts
pelikhan: I see that but I don't get the logic why you match text without backtick. Can we simplify this (and document it?)

microsoftsam: Sure

pelikhan: You use + in between the backticks

pxtblocks/blocklyloader.ts
pelikhan: We should emit an error in the console and skip that image or put a default image

pelikhan:
Btw, we should also support this in the rendered docs.

pelikhan:
We are leaking a "cached" compile options that did not get cleaned up properly. If we are serializing apiinfos back and forth, we should fix that instead.

pelikhan:
@microsoftsam tests are passing.

PR #3482, opened 2017-11-15
Don't filter stale charts while recording is paused

cait1inhennessy:
Fixes https://github.com/Microsoft/pxt-adafruit/issues/523

### DIFF
--- a/webapp/src/serial.tsx
+++ b/webapp/src/serial.tsx
@@ -182,7 +182,7 @@ export class Editor extends srceditor.Editor {
         this.active = true
         if (this.startPauseButton) this.startPauseButton.setState({ active: this.active });
         this.charts.forEach(s => s.start())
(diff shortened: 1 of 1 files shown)


webapp/src/serial.tsx
```diff
@@ -177,12 +175,14 @@ export class Editor extends srceditor.Editor {
...
         if (this.startPauseButton) this.startPauseButton.setState({ active: this.active });
         this.charts.forEach(s => s.stop())
+        clearInterval(this.chartDropper)
     }
 
     startRecording() {
         this.active = true
         if (this.startPauseButton) this.startPauseButton.setState({ active: this.active });
         this.charts.forEach(s => s.start())
+        this.chartDropper = setInterval(this.dropStaleCharts.bind(this), 5000)
```
pelikhan: maybe bump it to 20secs too

Issue #3415, opened 2017-11-03
The text of icon didn't display when open the "High Contrast"


yanlxu:
Repro steps:
• Navigate to https://makecode.microbit.org/beta
• Click "Projects" -> "Examples" ->"Plot Acceleration" 
• Switch between "Blocks" and "JavaScript"
• Click "Setting" -> "High Contrast On"

Expected result:
1. The display of icon "show data simulator" should be stable when switch between "Blocks" and "JavaScript"
2. The text of icon "show data simulator" should display when open the "High Contrast"
 
Actual result:
1. The display of icon "show data simulator" is not stable when switch between "Blocks" and "JavaScript"
2. The text of icon "show data simulator" didn't display when open the "High Contrast"

![plot acceleration](https://user-images.githubusercontent.com/11880996/32365940-f93a573c-c0b6-11e7-8f5e-7611beaa4cb5.gif)

PR #3487, opened 2017-11-15
White text on preview button in hc mode

cait1inhennessy:
capture
Fixes https://github.com/Microsoft/pxt/issues/3415


pelikhan:
Needs to go in v0 (#3415 is on micro:bit)

PR #3495, opened 2017-11-16
support for debuggable local url

pelikhan:


pelikhan:
Allow package to specify local URL to allow local debugging.

PR #3503, opened 2017-11-18
Fix emit of boolean expressions

mmoskal:
Fixes compiler bug uncovered in https://github.com/Microsoft/pxt-common-packages/pull/182


pelikhan:
Do we have to port this to v0?

mmoskal:
Probably a good idea

PR #3504, opened 2017-11-18
Update webfonts-generator

microsoftsam:
Update webfonts-generator as it depends on handlebar < 4.0 which has a vulnerability


pelikhan:
Is this needed for v0?

microsoftsam:
Yeah. Will cherry pick.

PR #3506, opened 2017-11-18
Fix undefined pxsim.noRefCounting.

microsoftsam:
Fix undefined pxsim.noRefCounting by adding it back using the new runtime.refCounting to indicate when the refCounting is off.

Issue: https://github.com/Microsoft/pxt/pull/3401#issuecomment-345413821

Verified this fix resolves the error introduced in the SIM.


pelikhan:
LGTM

PR #3520, opened 2017-11-28
Note

cait1inhennessy:
Make piano size configurable by target.  Goes with https://github.com/Microsoft/pxt-common-packages/pull/189 and https://github.com/Microsoft/pxt-ev3/pull/56.

### DIFF
--- a/pxtblocks/fields/field_note.ts
+++ b/pxtblocks/fields/field_note.ts
@@ -2,8 +2,8 @@
 
 namespace pxtblockly {
 
-     enum Note {
-        C = 262,
+    enum Note {
+       C = 262,
         CSharp = 277,
         D = 294,
         Eb = 311,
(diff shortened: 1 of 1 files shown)


pxtblocks/fields/field_note.ts
```diff
@@ -50,19 +50,53 @@ namespace pxtblockly {
...
+    interface NoteData {
+        name: string,
+        prefix: string,
+        altPrefix?: string,
+        freq: number
+    }
+
+    const Notes: {[key: number]: NoteData} = {
+        //TODO fill out the rest of these
+        40: {name: "C", prefix: "Middle", freq: 262},
```
pelikhan: Needs to be localizable

cait1inhennessy: How was it working before? Doesn't the big comment starting on line 237 take care of that?

pxtblocks/fields/field_note.ts
```diff
@@ -114,6 +149,16 @@ namespace pxtblockly {
                 this.colour_ = pxtblockly.parseColour(params.editorColour);
                 this.colourBorder_ = goog.color.rgbArrayToHex(goog.color.darken(goog.color.hexToRgb(this.colour_), 0.2));
             }
+
+            if (params.minNote && params.maxNote) {
+                let minNote = parseInt(params.minNote);
```
microsoftsam: You already have defaults for min and max, why not allow an author to only set one.

guillaumejenkins: Agreed, it would be great if they were independent settings

pxtblocks/fields/field_note.ts
guillaumejenkins: Could we move to a format similar to: `C (middle)`, `D (Tenor)`, etc? Advantages:

- Enables independent localization of prefix and notes, because it should work in most languages
- We can now programmatically generate this Notes dictionary during initialization instead of manually maintaining it manually
- We no longer need the huge localization comment below (which also needs to be maintained manually)

pxtblocks/fields/field_note.ts
cait1inhennessy: Good point, fixed.

PR #3526, opened 2017-11-29
Support for targetconfig in packaged editors

pelikhan:
When an editor is packaged, load targetconfig from local path


pelikhan:
Found the "isStatic" flag, fixing.

Issue #3546, opened 2017-12-01
Cannot escape About page if windows banner shows up in the background


pelikhan:
The dimmer does go away due to interaction with About and windows banner.

- open the editor in anonymous window, on windows
- click download to pop download dialog
- wait for windows banner to pop up
- close while banner is on

![image](https://user-images.githubusercontent.com/4175913/33499353-716a7e5c-d689-11e7-90ea-84491d8e644d.png)

cait1inhennessy:
This happens for Adafruit too.

samelhusseini:
This has been fixed.

PR #3558, opened 2017-12-04
Create separate div for dimmer context

cait1inhennessy:
Fixes https://github.com/Microsoft/pxt/issues/3546.

Problem seems to be that the "about" modal adds class `dimmed` to `#root` when it appears.  The re-render caused by the banner toggle removes this class.  When the `dimmed` class isn't there, modal close doesn't remove the dimmer.  I added a separate element `#dimmerRoot` so that changing the classes on `#root` won't affect the modal/dimmer.

Needs to go into master too.


microsoftsam:
Merge v0, and then looks good.

PR #3559, opened 2017-12-04
Move to TypeScript 2.6.1

riknoll:
This moves us from TypeScript 1.8.7 to Typescript 2.6.1 (both for our compiler and our code). Benefits include:

* No more `typings`; everything is now in `@types` and it should be *much* easier to add new types because we no longer need to get pre-2.0 compatible definitions
* Async await
* String enums
* Lots of other stuff https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-6.html

This is a **very dangerous** change and should not be merged before any big releases. It is also a breaking change. For example, this code is no longer legal:

```typescript
if (2 === 3) {
 // do something
}
```

TypeScript 2.0+ considers that comparison to be between the types `2` and `3` and not two numbers, so this comes up as an error. I'm sure there is other stuff I'm not thinking of.

Things still to be done:
* Not sure how much needs to be done with the Monaco language service
* Update our React version (we were waiting on the new types)
* Use Async await all over the place

## Testing

- [ ] basic test run against Adafruit
- [ ] basic test run against Minecraft
- [ ] basic test run against Chibitronics


package.json
```diff
@@ -108,8 +108,19 @@
     "semantic-ui-less": "^2.2.11",
```
pelikhan: dump the minor version as well

pxtcompiler/emitter/decompiler.ts
```diff
@@ -302,7 +302,7 @@ namespace ts.pxtc.decompiler {
 
     export function decompileToBlocks(blocksInfo: pxtc.BlocksInfo, file: ts.SourceFile, options: DecompileBlocksOptions, renameMap?: RenameMap): pxtc.CompileResult {
         let emittedBlocks = 0;
-        let stmts: ts.Statement[] = file.statements;
+        let stmts: NodeArray<ts.Statement> = file.statements;
```
pelikhan: is this a new type?

riknoll: No, but the type definition in the TS compiler changed so it can no longer be assigned to a normal array. Pretty sure the reason is that the indices are `readonly` in the `NodeArray` (which is a TS 2.0 feature).

pxtcompiler/emitter/service.ts
```diff
@@ -566,7 +564,7 @@ namespace ts.pxtc.service {
         }
     }
 
-    const operations: pxt.Map<(v: OpArg) => any> = {
+    const operations: pxt.Map<(v: any) => any> = {
```
pelikhan: any?

pxtlib/docsrender.ts
```diff
@@ -249,12 +248,12 @@ namespace pxt.docs {
 
         if (currentTocEntry) {
             if (currentTocEntry.prevPath) {
-                params["prev"] = `<a href="${currentTocEntry.prevPath}" class="navigation navigation-prev " title="${'Previous page: {0}', currentTocEntry.prevName}">
+                params["prev"] = `<a href="${currentTocEntry.prevPath}" class="navigation navigation-prev " title="${currentTocEntry.prevName}">
```
pelikhan: why this change?

riknoll: Look at the comma; the left side isn't doing anything so the TS compiler was throwing an error. I'm not sure what the desired behavior is, but my change shouldn't affect the current behavior.

pxtsim/libgeneric.ts
```diff
@@ -513,16 +513,6 @@ namespace pxsim {
             return new RefBuffer(buf.data.slice(offset, offset + length));
         }
 
-        export function toHex(buf: RefBuffer): string {
```
pelikhan: gone?

riknoll: Not used anywhere

pxtsim/svg.ts
```diff
@@ -73,15 +73,15 @@ namespace pxsim.svg {
 
 
     export function fill(el: SVGElement, c: string) {
-        (<SVGStylable><any>el).style.fill = c;
```
pelikhan: No more SVGStyleable?

pxtsim/libgeneric.ts
riknoll: Oh, it's a sim-shim. I'll bring it back

pxtcompiler/emitter/emitter.ts
```diff
@@ -309,7 +325,7 @@ namespace ts.pxtc {
         while (true) {
             node = node.parent
             if (!node)
-                userError(9229, lf("cannot determine parent of {0}", stringKind(node0)))
+            userError(9229, lf("cannot determine parent of {0}", stringKind(node0)))
```
pelikhan: missing tab?

pxtcompiler/emitter/emitter.ts
```diff
@@ -557,6 +599,39 @@ namespace ts.pxtc {
         return checkType(r)
     }
 
+    function checkUnionOfLiterals(t: Type): HasLiteralType {
```
pelikhan: coool! do we have tests on those?

pxtcompiler/emitter/emitter.ts
```diff
@@ -1064,7 +1152,14 @@ namespace ts.pxtc {
                 msg = lf("{0} not currently supported", syntax)
             }
             else {
-                msg = lf("{0} not supported", syntax)
+                msg = lf("{0} not supported", ts.SyntaxKind[n.kind])
+            }
+
+            let t = (n as any) as any[];
```
pelikhan: left over logging

pxtcompiler/emitter/emitter.ts
```diff
@@ -3310,13 +3404,13 @@ ${lbl}: .short 0xffff
                 return r;
             let tp = typeOf(e)
 
-            if (target.floatingPoint && (tp.flags & (TypeFlags.NumberLike | TypeFlags.Boolean)))
+            if (target.floatingPoint && (tp.flags & (TypeFlags.NumberLike | TypeFlags.Boolean | TypeFlags.BooleanLiteral)))
```
pelikhan: at this point, we should really clean out the floating point stuff

pxtlib/docsrender.ts
pelikhan: Oh... it is missing lf("..."). Let's keep this change. simpler.

cli/cli.ts
```diff
@@ -2300,14 +2302,14 @@ class Host
         check(p)
 
         if (U.endsWith(filename, ".uf2"))
-            fs.writeFileSync(p, contents, "base64")
+            fs.writeFileSync(p, contents, { encoding: "base64" })
```
mmoskal: Node.js doesn't seem to deprecate string as encoding usage. Is the @types missing it?

pxtsim/svg.ts
riknoll: Yeah, I couldn't find it in `lib.d.ts`. Either it's gone or the name changed.

pelikhan: sniff :(

cli/cli.ts
riknoll: Yeah, [you're right](https://nodejs.org/api/fs.html#fs_fs_writefile_file_data_options_callback). Not sure why it was removed from the `@types`

pxtcompiler/emitter/emitter.ts
mmoskal: Maybe it needs a general cleanup but I'm convinced we should leave support for integer compilation for tiny targets (AVR, maybe some tiny ARMS)

PR #3560, opened 2017-12-04
Move event listener addition/deletion to React lifecycle methods

cait1inhennessy:
Needs to go into master too.

### DIFF
--- a/webapp/src/serialindicator.tsx
+++ b/webapp/src/serialindicator.tsx
@@ -19,7 +19,14 @@ export class SerialIndicator extends React.Component<SerialIndicatorProps, Seria
     constructor(props: any) {
         super(props)
         this.state = { active: false }
+    }
+
+    componentDidMount() {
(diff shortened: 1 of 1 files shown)


webapp/src/serialindicator.tsx
```diff
@@ -19,9 +19,16 @@ export class SerialIndicator extends React.Component<SerialIndicatorProps, Seria
...
+
+    componentDidMount() {
         window.addEventListener("message", this.setActive.bind(this))
     }
 
+    componentWillUnmount() {
+        window.removeEventListener("message", this.setActive.bind(this))
+    }
+
     setActive(ev: MessageEvent) {
```
cait1inhennessy: This method _already_ verifies that `this.state.active` is false before setting it to true.  So it's not constantly updating the state as we thought.

microsoftsam:
Integrate to master please.

PR #3562, opened 2017-12-04
Add gif and fix alignment

cait1inhennessy:
Replace bar chart icon with animated gif from designers.  Also fix alignment.

Before: 
capture2

After: 
capture


webapp/src/serialindicator.tsx
```diff
@@ -41,7 +41,7 @@ export class SerialIndicator extends React.Component<SerialIndicatorProps, Seria
         return(
             <div title={lf("Open console")} className="ui label circular" tabIndex={0} onClick={this.props.onClick} onKeyDown={sui.fireClickOnEnter}>
                 <div className="detail">
-                    <sui.Icon icon="bar graph"/>
+                    <img className="barcharticon" src={Util.pathJoin(pxt.webConfig.commitCdnUrl, `images/Bars_black.gif`)}></img>
```
pelikhan: This image won't load in offline mode.

cait1inhennessy: How can we fix that?

pelikhan: maybe we just hide it if it fails to load

pelikhan: Open a bug to fix charting support when offline. This gif + smoothie need to be properly cached.

microsoftsam: Need alt text for the img.

microsoftsam: or aria-hidden

cait1inhennessy:
@pelikhan @microsoftsam Ping for another review. I added alt text and offline caching.

webapp/src/serialindicator.tsx
```diff
@@ -41,7 +41,7 @@ export class SerialIndicator extends React.Component<SerialIndicatorProps, Seria
         return(
             <div title={lf("Open console")} className="ui label circular" tabIndex={0} onClick={this.props.onClick} onKeyDown={sui.fireClickOnEnter}>
                 <div className="detail">
-                    <sui.Icon icon="bar graph"/>
+                    <img alt="Animated bar chart" className="barcharticon" src={Util.pathJoin(pxt.webConfig.commitCdnUrl, `images/Bars_black.gif`)}></img>
```
pelikhan: use ``lf`` to localize alt text

PR #3568, opened 2017-12-05
New approach for fixing dimmer issues

cait1inhennessy:
Instead of having a separate "dimmerRoot" element, preserve jQuery classes attached to "root" upon re-render.

Needs to go to master too.


microsoftsam:
Tested in IE?

PR #3606, opened 2017-12-08
 Limit bne+b optimization

mmoskal:
This fixes Microsoft/pxt-adafruit#514 (which was currently masked due to to slightly different code generation, but was indeed a code gen bug)


mmoskal:
Actually there is another related problem - do not merge just yet

mmoskal:
Should be good to go now.

PR #3609, opened 2017-12-08
Adding pause until block

riknoll:
Optional `pauseUntil()` block that can be defined in the target. This block is special, because it looks likes it takes in a Boolean but we compile that down into a lambda. As part of this PR I also added a flag for the decompiler to insert error messages into the XML whenever we produce a grey block (it's only enabled for the decompiler tests).

Note: This uses the *lambda* syntax even though we use function expressions everywhere else. The intention was to make the block a single line.


pelikhan:
The ``pauseUntil`` function has a timeout optional argument. Is this covered in the testsuite?

riknoll:
Oops, forgot to add compiler tests in addition to decompiler tests. I'll add those now.

riknoll:
Tests added

PR #3637, opened 2017-12-13
Fixing equality operators in common docs

riknoll:
In the new TS version you can't compare two different numbers directly because they are considered different literal types (i.e. instead of "number" and "number" the types would be "0" and "42"). This doesn't have much of a practical effect because those expressions will always evaluate the same way anyways.

### DIFF
--- a/common-docs/blocks/logic.md
+++ b/common-docs/blocks/logic.md
@@ -5,7 +5,7 @@ if(true) {}
 true;
 true && false;
 !true;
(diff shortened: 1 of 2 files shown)


mmoskal:
@riknoll is there a way to disable this check in TypeScript? I have a feeling this is one type too many for kids.

riknoll:
Sadly, no. We could maybe filter out the errors though

Issue #3541, opened 2017-12-01
After refreshed the page, the page will change from "High Contrast On" to "High Contrast Off" automatically .


Emma-sunny:
**Repro step:**
- Navigate to [https://makecode.microbit.org/beta](url)
- Click "_setting_" and select "_High Contrast On_ "
- Then refresh the page

**Expected Result:**
It should back to the mode "High Contrast Off" after clicking "_High Contrast Off_".
![c6603071-6f58-4856-a957-3f100983886f](https://user-images.githubusercontent.com/17866936/33475577-de90033c-d6b9-11e7-8e43-caf52d986118.gif)

**Actual result:**
![cc8deb81-0d46-4cf7-b215-cc1b35460231](https://user-images.githubusercontent.com/17866936/33475508-a7ab65e6-d6b9-11e7-8230-3dba90cd1d37.gif)

PR #3639, opened 2017-12-13
Save high contrast state between renders

cait1inhennessy:
Working on https://github.com/Microsoft/pxt/issues/3541.

This solution doesn't work because `componentWillUnmount` is not called when the page refreshes.

Suggestions?


cait1inhennessy:
After discussion with Peli, closing as "won't fix".

PR #3646, opened 2017-12-15
Add 'done' button to 'download completed' dialog

cait1inhennessy:
Same way it's done in v0.


microsoftsam:
There was a decision made a while back to remove the "Done" button, the reason being if I recall correctly is that kids assumed that "Done" means they didn't need to do any of the steps outlined in the description. 

I would recommend instead of adding a "Done" button, add a "close" icon in the modal.

cait1inhennessy:
@microsoftsam OK, fixed

PR #3648, opened 2017-12-15
Fixing offline hexcache and other minor CLI issues

riknoll:
The ternary expression was missing parentheses


pelikhan:
Aaaargggh! Missing parenthesis!

PR #3649, opened 2017-12-15
Serial message queue

cait1inhennessy:
While the serial editor is invisible, it saves messages in a queue and displays them upon opening.

Removed all logic surrounding "source" - now graphs are simply separated by variable name.  Separating graphs by source (which is runtime ID for simulator) led to conflict: queued data and live data from the same stream were going to different graphs because the runtime ID changes when you open the serial editor.

Should go into master too.

serial


webapp/src/serial.tsx
```diff
@@ -14,9 +14,10 @@ import Util = pxt.Util
 const lf = Util.lf
 
 export class Editor extends srceditor.Editor {
+    savedMessageQueue: MessageEvent[] = []
+    maxSavedMessages: number = 10
```
pelikhan: bump it to 100

webapp/src/serial.tsx
```diff
@@ -51,10 +52,12 @@ export class Editor extends srceditor.Editor {
         }
         this.isVisible = b
         if (this.isVisible) {
+            this.processQueuedMessages()
```
pelikhan: I think you want to move this call in ``startRecording``

webapp/src/serial.tsx
```diff
@@ -75,60 +78,76 @@ export class Editor extends srceditor.Editor {
...
-        }
-        let niceSource = this.sourceMap[source]
 
         const m = /^\s*(([^:]+):)?\s*(-?\d+(\.\d*)?)/i.exec(data);
         if (m) {
             const variable = m[2] || '';
             const nvalue = parseFloat(m[3]);
             if (!isNaN(nvalue)) {
-                this.appendGraphEntry(niceSource, variable, nvalue)
+                this.appendGraphEntry(variable, nvalue)
```
pelikhan: are we not differentiating data from different sources? what happens when 2 micro:bit stream the same data?

webapp/src/serial.tsx
cait1inhennessy: The editor is always recording when it's opened, so this should be fine as-is.

cait1inhennessy:
@pelikhan This should be ready to merge.

webapp/src/serial.tsx
```diff
@@ -75,27 +79,51 @@ export class Editor extends srceditor.Editor {
...
+        this.savedMessageQueue.forEach(m => this.processMessage(m));
+        this.savedMessageQueue = [];
+    }
+
+    processEvent(ev: MessageEvent) {
+        let msg = ev.data
+        if (msg.type !== "serial") return;
         const smsg = msg as pxsim.SimulatorSerialMessage
+
+        smsg.receivedTime = Util.now();
```
pelikhan: only assign time if missing

pelikhan:
Needs cherry-pick to master.

PR #3653, opened 2017-12-18
Support icons on group labels.

microsoftsam:
Support icons on group labels 
Minor fix to labelWidth set to 0 issue as well.

### DIFF
--- a/pxtlib/service.ts
+++ b/pxtlib/service.ts
@@ -147,6 +147,7 @@ namespace ts.pxtc {
         // On namepspace
         subcategories?: string[];
         groups?: string[];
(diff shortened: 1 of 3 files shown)


pxtlib/service.ts
```diff
@@ -503,6 +504,14 @@ namespace ts.pxtc {
                 res.groups = undefined;
             }
         }
+        if (res.groupIcons) {
+            try {
+                res.groupIcons = JSON.parse(res.groupIcons as any);
+            }
+            catch (e) {
+                res.groupIcons = undefined;
```
pelikhan: can we emit some kind of error?

PR #3658, opened 2017-12-19
Remove last usage of the long time deprecated tsd

nedseb:


pelikhan:
@riknoll we do not need typings either, don't we? It's all @types now...

riknoll:
Yep, I guess I forgot to change the README

riknoll:
Fixed by #3662

PR #3696, opened 2018-01-04
Correct monaco snippets for fixed instances

cait1inhennessy:
screen shot 2018-01-05 at 11 11 08 am

screen shot 2018-01-05 at 11 10 56 am

Reworks creation of monaco snippets for fixed instances in toolbox and javascript editor.
- Toolbox snippets now look like `fixedInstanceName.method()`, e.g., `A0.analogRead()`.
- Javascript editor snippets now look like `namespace.fixedInstanceName.method()`, e.g., `pins.A0.analogRead()`.
- New logic to map fixed instances w/specialized classes to the correct namespace by creating a lookup array of all classes and subclasses within the namespace.

### DIFF
--- a/webapp/src/monaco.tsx
+++ b/webapp/src/monaco.tsx
@@ -788,7 +788,7 @@ export class Editor extends srceditor.Editor {
             let snippetPrefix = fn.noNamespace ? "" : ns;
             let isInstance = false;
             let addNamespace = false;
-            let namespaceToUse = ""
+            let namespaceToUse = "";
 
             const element = fn as pxtc.SymbolInfo;
             if (element.attributes.block) {
--- a/webapp/src/monaco.tsx
+++ b/webapp/src/monaco.tsx
@@ -801,14 +801,15 @@ export class Editor extends srceditor.Editor {
                         let instances = Util.values(this.blockInfo.apis.byQName)
                         let getExtendsTypesFor = function(name: string) {
                             return instances
-                                    .filter(v => v.extendsTypes && v.extendsTypes.indexOf(name) !== -1)
+                                    .filter(v => v.extendsTypes)
+                                    .filter(v => v.extendsTypes.reduce((x, y) => x || y.indexOf(name) != -1, false))
                                     .reduce((x, y) => x.concat(y.extendsTypes), [])
                         }
                         namespaceToUse = element.attributes.blockNamespace || nsInfo.namespace || "";
                         instances = instances.filter(value =>
                             value.kind === pxtc.SymbolKind.Variable &&
                             value.attributes.fixedInstance &&
-                            getExtendsTypesFor(nsInfo.namespace + "." + nsInfo.name).indexOf(value.retType) !== -1)
+                            getExtendsTypesFor(nsInfo.name).indexOf(value.retType) !== -1)
                             .sort((v1, v2) => v1.name.localeCompare(v2.name));
                         if (instances.length) {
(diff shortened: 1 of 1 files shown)


samelhusseini:
Let's get @riknoll to review this, just to make sure we haven't missed any context.

riknoll:
What is the issue?

webapp/src/monaco.tsx
```diff
@@ -839,6 +842,7 @@ export class Editor extends srceditor.Editor {
                     let currPos = monacoEditor.editor.getPosition();
                     let cursor = model.getOffsetAt(currPos)
                     let insertText = snippetPrefix ? `${snippetPrefix}.${snippet}` : snippet;
+                    insertText = addNamespace ? `${/([^\.]+)/.exec(element.namespace)[0]}.${insertText}` : insertText;
```
riknoll: Let's move this regex (and the one below) into a function. Also, the parentheses seem unnecessary since you are using the full match anyways

samelhusseini:
The code added for adafruit for class functions (strip.showAnimation) broke the fixed instance snippet generation for something like pins.D11.digitalWrite.

screen shot 2018-01-04 at 10 47 28 am

PR in question: https://github.com/Microsoft/pxt/pull/3422

Currently (in adafruit), input -> button.onEvent inserts: 
```
button.onEvent(ButtonEvent.Click, function () {

})
```
instead of input.buttonA.onEvent...

webapp/src/monaco.tsx
```diff
@@ -787,31 +787,34 @@ export class Editor extends srceditor.Editor {
...
-                    if (nsInfo.kind === pxtc.SymbolKind.Class) {
-                        return undefined;
-                    }
-                    else if (nsInfo.attributes.fixedInstances) {
+                    if (nsInfo.attributes.fixedInstances) {
                         const instances = Util.values(this.blockInfo.apis.byQName).filter(value =>
                             value.kind === pxtc.SymbolKind.Variable &&
                             value.attributes.fixedInstance &&
-                            value.retType === nsInfo.name)
+                            value.retType.endsWith(nsInfo.name))
```
riknoll: Don't use `endsWith` (not all browsers support it). I think we have `Util.endsWith()` that you can use instead

samelhusseini:
There's another issue here with the instance search as it doesn't account for specialized classes. ie: in the case of AnalogPin in adafruit, the instance search fails to find any matches when searching for AnalogPin, as all the fixed instance (analog pins) are of type PwmPin and not AnalogPin. AnalogPin inherits from PwmPin. 

We need some way to catch those cases. Perhaps to be fixed in a separate issue. @riknoll thoughts?

riknoll:
Yeah, let's fix this as part of this PR. The `SymbolInfo` has a property called `extendsTypes` that says what classes it extends from; maybe we could build and cache an inheritance map of some sort. @cait1inhennessy thoughts?

cait1inhennessy:
@riknoll That idea sounds good, working on implementing it now

cait1inhennessy:
@samelhusseini @riknoll Ping for review of changes

samelhusseini:
Mostly looks good to me. 
This part of the codebase keeps breaking, would it be possible to add more comments describing the different code paths. Perhaps a small example for each code path, that if (in the future) someone is making a change they could test these different scenarios. 

Also final commit has a build break.

riknoll:
LGTM. I think that build issue is just Travis messing up (network timed out)

PR #3698, opened 2018-01-06
Support for compilation of integer and whole numbers

samelhusseini:
Since latest Blockly pull, I added support for integer and whole numbers. 
These number types limit number inputs in blocks to numbers only (no decimals) and positive numbers only respectively. 

Adding compiler support for these and using them for the repeat and for blocks. 
There's no way you can repeat (negative) values, and no way you can repeat / iterate a fraction of times.

TODO: 
- [x] Decompiler support? Do we want to add that for repeat and for blocks
- [x] Tests passing
- [ ] Authoring support? Do we want to make this something blocks can configure.


riknoll:
Yes to both of the TODOs. I don't want to merge this without decompiler support at least.

riknoll:
What is the difference between this PR and https://github.com/Microsoft/pxt/pull/3699/files ?

samelhusseini:
Sorry there were files that sneaked in this PR. reverted.

samelhusseini:
and makes sense about the Todos.

pxtblocks/blocklyloader.ts
```diff
@@ -2600,31 +2600,21 @@ namespace pxt.blocks {
...
-            (Blockly as any).Colours.textField,
-            (Blockly as any).Colours.textField
-        );
+        const numberBlocks = ['math_number', 'math_integer', 'math_whole_number', 'math_number_minmax']
+        numberBlocks.forEach(num_id => {
+            const mInfo = pxt.blocks.getBlockDefinition(num_id);
+            installHelpResources(
+                num_id,
+                mInfo.name,
+                (pxt.appTarget.compile && pxt.appTarget.compile.floatingPoint) ? lf("a decimal number") : lf("an integer number"),
```
riknoll: Let's just move the description to `pxtlib\blocks.ts` because the logic is more complicated now.

samelhusseini:
Going to skip authoring support for a different PR. @riknoll ready for review.

samelhusseini:
I can't add any new tests @riknoll, as the only new tests cases not covered are decompiling something like this: 
```
for (let i = 0; i <= 10.1; i++) {
	
}
for (let i = 0; i < 10.1; i++) {
	
}
```
and (i tried) adding that into our test framework and it complains that decimals were not supported. 
This is because pxt is integer only (see pxtarget configuration in pxt).

so there must not be any decimal tests cases as of today. 
Going to push this forward since we don't have a story to solve this at the moment.

Peli and I have spoken about the possibility of turning floating points on in pxt (master) but we saw a couple of issues with that today, let's chat about this in person.

PR #3706, opened 2018-01-09
Add support for bold field annotation inside the block text

samelhusseini:
Adds support for bold fields inside of the text label. 
This is annotated inside the block annotation using markdown annotation (*). 

For example: 
```
//% blockId="testFunction" block="test *function*"
export function testFunction() {

}
```
In the above, function will be bold.

### DIFF
--- a/localtypings/blockly.d.ts
+++ b/localtypings/blockly.d.ts
@@ -632,6 +632,10 @@ declare namespace Blockly {
         position_(): void;
     }
 
+    class FieldLabel extends Field {
+        constructor(text: string, opt_class: string);
+    }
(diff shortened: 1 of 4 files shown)


pelikhan:
Can you add a test? E.g. a block that uses this code path?

samelhusseini:
Yes will do.

samelhusseini:
Decompiler tests won't catch this, and there are numerous code paths that need unit tests in blockly loader. Filed a bug for that and we can handle that as a feature item: 
https://github.com/Microsoft/pxt/issues/3708

PR #3710, opened 2018-01-10
Removing share -> screenshot option

pelikhan:
In favor of lightweight codembed option.

- screenshot can be slow
- screenshot does not work on IE
- screenshot generates huge URLs


abchatra:
I would remove command line as well. Does anyone use that? Also add telemetry?

pelikhan:
Sure can. Let’s keep it slim.

Issue #3712, opened 2018-01-10
Math.Pow doesn't decompile correctly


riknoll:
Didn't know we ever supported Math.pow in the blocks, but apparently we do.

pelikhan:
We also support **

________________________________

riknoll:
No we don't

riknoll:
I guess we should though. Nothing stopping us

PR #3713, opened 2018-01-10
Redoing compilation and decompilation of exponents

riknoll:
Fixes https://github.com/Microsoft/pxt/issues/3712

Better exponent support:
* Adds decompiler support for `**` and `Math.pow()`
* Change the exponentiation operator in the blocks from `^` (which is XOR) to `**`
* Change the compiled TS from `Math.pow()` to `**` to better match the other operators


riknoll:
Needs to go to V0

PR #3717, opened 2018-01-11
Adding support for ** in V0

riknoll:
Cherry picking #3713 

This adds support for `**` and changes the compilation of exponents in the blocks to use `**` instead of `Math.pow()`. This isn't a breaking change for code, because `**` and `Math.pow()` both decompile to the same thing. However, it is *technically* a UI change because the operator in the block dropdown is changing from the incorrect `^` (which is actually XOR) to `**`.


riknoll:
I also moved a few test cases around that were in the wrong place. I think they must have been cherry-pick mistakes from past PRs

abchatra:
Can you add this test case `2**3**2` answer should be 512

riknoll:
@abchatra is that to test the parsing? We don't really have tests for that (since we just use the TS compiler's parsing).

abchatra:
I am curious. As we are converting to Math.pow operation, want to make sure left operator associativity is preserved. One time test will do.

riknoll:
Tested on hardware and in the simulator, seems to be working fine!

PR #3764, opened 2018-01-25
adding blog about cs intro japanese

pelikhan:


ganicke:
A few edits in #3765.

PR #3766, opened 2018-01-25
allow inverting the "up" "down" logic

pelikhan:
Allow to invert up/down mapping to true/false for specific case.s


riknoll:
I think we should just have one `field_toggle` with two parameters: `trueText` and `falseText`. Right now we have 4.

pelikhan:
Not for this PR.

PR #3767, opened 2018-01-25
New cookie banner and consent message

riknoll:
__Note: Please review this PR carefully! Especially check the events I have labelled as constituting interactive consent to make sure they are the result of explicit user actions. If you have questions about something, message me.__

This PR enables the application insights cookies along with a new UI for the cookie consent message. A localized banner now appears on top of everything else and disappears automatically when the user interacts with the page. To see the banner in action on another site, visit https://www.microsoft.com/fr-fr in a private-browsing window. When the banner disappears, a cookie is stored indicating that the user consents to cookies and the _next_ time they visit application insights will be allowed to drop cookies (but not in the current session, AI doesn't really allow conditional initialization like that).

As part of this change I also added `docfiles/pxtweb` for holding onto any code that belongs on _every_ page served on the site. If you add any code there, make sure it doesn't have any dependencies. I also removed the pxtarget flag for disabling the cookie message; targets can override it by including their own tracking.html (but they shouldn't).

Known issues:
* In IE, the cookie banner appears over the header in makecode.com (rather than pushing it down). On small screens it can cover up the entire header

Once this is merged it also needs to be ported to V0 and V2.


riknoll:
BTW, the language cookie is unaffected by this.

docfiles/docs.js
```diff
@@ -50,14 +50,14 @@ function dirAuto($el) {
 }
 
 function searchSubmit(form) {
-    if (pxt && pxt.tickEvent) pxt.tickEvent("docs.search", { 'source': form.id })
+    if (pxt && pxt.tickEvent) pxt.tickEvent("docs.search", { 'source': form.id }, /*interactiveConsent*/true)
```
pelikhan: this notation is going to hard to enfore in the future. Instead, you could define an Options interface 
```
interface Options {
    interactiveConsent?: boolean;
}
```
and do
```
tickEvent(...., { interactiveConsent: true })
```
that way, TypeScript forces us to explain the meaning of that boolean value.

docfiles/pxtweb/cookieCompliance.ts
```diff
@@ -0,0 +1,217 @@
...
+            }
+
+            if (resp.status === 200) {
+                try {
+                    const info = JSON.parse(resp.body);
+                    cb(undefined, info as CookieBannerInfo);
+                    return;
+                }
+                catch (e) {
+                    cb(new Error("Bad response from server: " + resp.body))
```
pelikhan: bail out and consider that cookies are not permitted. Maybe the user is behind a hotel front page and the banner returns HTML garbage.

docfiles/pxtweb/cookieCompliance.ts
```diff
@@ -0,0 +1,217 @@
...
+                    return;
+                }
+            }
+            cb(new Error("didn't get 200 response: " + resp.status + " " + resp.body));
```
pelikhan: Handle failure gracefully.

docfiles/pxtweb/cookieCompliance.ts
```diff
@@ -0,0 +1,217 @@
...
+            }
+
+            // The markup is trusted because it's from our backend, so it shouldn't need to be scrubbed
+            bannerDiv.innerHTML = info.Markup;
+
+            if (info.Css && info.Css.length) {
+                info.Css.forEach(injectStylesheet)
+            }
+
+            all(info.Js || [], injectScriptAsync, msccError => {
```
pelikhan: it practices there's only 1 javascript file (and there should never be more than one)

docfiles/pxtweb/cookieCompliance.ts
riknoll: What's un-graceful about this? I'm passing an error, not throwing it.

docfiles/pxtweb/cookieCompliance.ts
riknoll: The API returns an array so I support an array.

docfiles/pxtweb/cookieCompliance.ts
riknoll: That's what I'm doing already. This will pass an error to the callback which is handled above (see `bannerErr` check)

riknoll:
Answers:

1. __Is this tested in all browsers?__ Tested Chrome, Edge, IE, and Firefox on Windows
2. __What happens when the user goes offline?__ The request to the service will fail and app insights will load without cookies. If the cookies are already present, they will not be cleared (but may be the next time the page is loaded).
3. __What happens if the mscc.js file fails to load?__ App insights will be loaded without cookies. I think the banner might also be displayed; I have to check that.
4. __Is this happening in the background or blocking the editor load?__ Everything is happening in the callback to the `XMLHttpRequest`, so it should be async.

docfiles/docs.js
riknoll: Sure

pelikhan:
LGTM

riknoll:
I just found a bug that I'm fixing; I'll comment when I'm finished and do the merge.

abchatra:
Lets port to v0 once merged and tested.

abchatra:
I don't think this needs to be ported to v2.

riknoll:
@pelikhan @abchatra please take a look at the last commit; I added queueing of events to prevent races as the editor loads and fixed a bug that caused analytics not to load in the editor.

docfiles/pxtweb/cookieCompliance.ts
```diff
@@ -126,6 +138,10 @@ namespace pxt {
         const loadAI = (window as any).loadAppInsights;
         if (loadAI) {
             loadAI(includeCookie)
+            analyticsLoaded = true;
+            while (pendingCallbacks.length) {
+                pendingCallbacks.pop()();
```
pelikhan: are you supposed to call these callbacks?

riknoll: I am calling them. Maybe I should add a comment since I had to stare at that line for a bit to figure that out...

pelikhan:
@riknoll why aren't we simply storing the data instead of all the callbacks?

docfiles/pxtweb/cookieCompliance.ts
pelikhan: Please. I'm really confuse about what's happening here.

riknoll: There are two sets of parentheses, one for `pop` and one to call the callback. I'll add a comment.

pelikhan: what happens when the client overrides pxtError / pxtEvent?

riknoll: Not sure what you mean by that

docfiles/pxtweb/cookieCompliance.ts
```diff
@@ -0,0 +1,233 @@
...
+        if (loadAI) {
+            loadAI(includeCookie)
+            analyticsLoaded = true;
+            while (pendingCallbacks.length) {
+                pendingCallbacks.pop()();
+            }
+        }
+    }
+
+    function httpGetAsync(url: string, cb: Callback<HttpResponse>) {
```
pelikhan: don't we have helpers to do this already?

pxtlib/analytics.ts
```diff
@@ -1,13 +1,60 @@
...
-        pxt.tickEvent = function (id: string, data?: Map<string | number>): void {
-            if (te) te(id, data);
+        const teHandled = wrap((ev: TickData) => tickEventCore(ev.id, ev.data, ev.opts));
+        pxt.tickEvent = function (id: string, data?: Map<string | number>, opts?: TelemetryEventOptions): void {
+            teHandled({ id, data, opts });
+        };
+
+        const rexp = pxt.reportException;
+        const rexpHandled = wrap((ev: ExceptionData) => reportExceptionCore(ev.err, ev.data));
+        pxt.reportException = function (err: any, data: pxt.Map<string>): void {
```
pelikhan: this function can be overriden, does it matter? it won't be using wrap most likely.

riknoll: I don't think it matters. Who overrides this BTW? The only place I could find in the codebase is the editor controller business and that shouldn't matter.

riknoll: If someone overrides it and they don't call the original then it's a moot point anyway.

riknoll:
I am storing the data (see the `wrap` function). The reason I also have to do the `pendingCallbacks` stuff is because the analytics code lives in pxtlib and AI gets loaded by pxtweb and I'm trying not to create any hard dependencies between the two. If you want, I could instead move the implementations of `tickEvent`, `reportError`, etc. into `pxtweb` and just leave the declarations in `pxtlib`.

pelikhan:
Sounds simpler.

riknoll:
Will do

riknoll:
@pelikhan check it out. Also, I'm removing the DO NOT MERGE so feel free to merge this if you're happy with it. The backend is already set up.

docfiles/pxtweb/cookieCompliance.ts
```diff
@@ -71,13 +102,18 @@ namespace pxt {
         });
     }
 
-    export function onAnalyticsLoaded(cb: () => void) {
-        if (analyticsLoaded) {
-            cb();
+    export function aiTrackEvent(id: string, data?: any, measures?: any) {
+        if (!eventLogger) {
+            eventLogger = new TelemetryQueue((a, b, c) => (window as any).appInsights.trackEvent(a, b, c));
```
pelikhan: is appInsights always defined here?

docfiles/pxtweb/cookieCompliance.ts
```diff
@@ -71,13 +102,18 @@ namespace pxt {
...
         }
-        else {
-            pendingCallbacks.push(cb);
+        eventLogger.track(id, data, measures);
+    }
+
+    export function aiTrackException(err: any, kind?: string, props?: any) {
+        if (!exceptionLogger) {
+            exceptionLogger = new TelemetryQueue((a, b, c) => (window as any).appInsights.trackException(a, b, c));
```
pelikhan: is appInsights always defined here?

pelikhan:
What happens in statically packaged editors?

pelikhan:
Tested in a number of places and it looks good.

PR #3794, opened 2018-01-30
Support for optional arguments in blocks

riknoll:
It's finally happening!

The purpose of this PR is to surface optional arguments in TS functions up to the blocks. You can do it like this:

```typescript
/**
 * A test api
 */
//% weight=99 inlineInputMode="inline"
//% blockId=whatever block="normal args %str %ang || optional args %opt %enu"
export function whatever(str: string, ang: number, opt?: boolean, enu?: Direction) {
}
```

The double pipe (`||`) is used to specify where the optional arguments begin. All arguments after that will be hidden by default and can be revealed using the + and - buttons at the end of the block.

As part of this change, I had to redo the block parser and it is now new-and-improved! Changes include:

1. You no longer need to use `|` to separate parameters
2. Support for the markdown bold and italics syntax. I think this is technically a breaking change because we were supporting the wrong syntax before (single asterisk for bold instead of double asterisk). You can even nest styles if you're crazy.
3. Support for adding arbitrary CSS classes to text! I copied the markdown link syntax for this (e.g. `[some text](aClassName)`)
3. Very rudimentary escape sequences! I currently do not support any of the special escapes (like newlines), unicode escapes (`\uXXXX`), or latin-1 escapes (`\xXX`). Really it's only useful for escaping the characters that have special meanings in the block string (like `|`)

Things that still need to be done:

1. Proper escape sequence parsing. This could end up being pretty useful in combination with the CSS class feature to do things like embed font icons in blocks
2. Add docs for all of this
3. Better errors for bad block definitions in the console

I'm putting "DO NOT MERGE" on this because I still need to do some tests to make sure everything is backwards compatible.


pelikhan:
Looks awesome.

One comment: I don't think we should allow users to use italics or random CSS classes on blocks. Are are running the risk of turning the editor into a giant WordArt horror?

riknoll:
Ha! I'm fine with removing italics; I just added it because we were already doing the markdown syntax for bold and I figured people would expect us to support both. I refuse to remove CSS classes because it's awesome (font icons!)

pelikhan:
We can’t people take dependencies on internal CSS classes - otherwise we might break blocks whenever we refactor CSS. Can we scope this down to icon fonts only?

riknoll:
@pelikhan they wouldn't use our classes; targets can define their own CSS in `theme/blockly.less`. All I'm doing is surfacing [a Blockly feature](https://developers.google.com/blockly/guides/create-custom-blocks/define-blocks#label)

pelikhan:
Users could also publish packages relying on internal classes. It is unlikely though so push it in.

pelikhan:
The `|` was used to specificy field breakpoint for blocks. Is this still supported? It is important for multi-line blocks.

pelikhan:
I tried \x25 but the parser complained.

pelikhan:
There is an issue with instance methods. The first optional argument become the "this" argument.

riknoll:
I'm going to redo this PR

PR #3800, opened 2018-01-31
upgrading to jquery ^3.0.0

pelikhan:
Vulnerability detected in 2.2.4


abchatra:
I wish we got rid of jquery

pelikhan:
Yes we should get rid of it, but that requires a bit more work.

PR #3806, opened 2018-02-02
Uwp hid improvements

guillaumejenkins:
- Uses the WinRT device watcher to reconnect as little as possible to HID devices
- As a result, the serial data polling logic is no longer needed, since we already reconnect when a device is newly plugged in

Not sure if this will fix the Adafruit crash issue, but it certainly makes the whole HID connection much cleaner for UF2

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1776,7 +1776,7 @@ ${compileService && compileService.githubCorePackage && compileService.gittag ? 
             <div id='root' className={rootClasses}>
                 {hideMenuBar ? undefined :
                     <header className="menubar" role="banner">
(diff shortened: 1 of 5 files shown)


webapp/src/app.tsx
```diff
@@ -1882,72 +1881,34 @@ function initLogin() {
...
-                pxt.log(`hidbridge failed to load, ${e}`);
-                startSerialConnectionPoller();
-            })
-        return
+        hidbridge.configureHidSerial((buf, isErr) => {
+            let data = Util.fromUTF8(Util.uint8ArrayToString(buf))
+            //pxt.debug('serial: ' + data)
+            window.postMessage({
+                type: 'serial',
+                id: 'n/a', // TODO
```
pelikhan: Do we have any kind of id available?

Issue #3650, opened 2017-12-16
Serial editor enhancement ideas


cait1inhennessy:
## Ideas
- [x] Store more data
- [ ] Documentation
- [ ] Add timestamps to csv file
- [ ] Print out graphable data to the console: "x:123", etc
- [ ] Allow user to specify data retention limit - maybe forever
- [ ] (App) Spreadsheet should go to downloads folder
- [ ] Make new project/old data scenario less confusing
- [ ] Raw console output view/no-charts option
- [ ] Restore Azure streaming
- [ ] Minimum 5 data points and 5 ms delay before data is turned into a graph
- [ ] Label x- and y-axes on charts
- [ ] Dynamic x-axis scaling OR predetermined time length -> evenly divide x-axis
- [ ] Allow user to specify sample rate
- [ ] Have charts always take up 100% of chart area - if there are 2, 50% each, etc.
- [ ] Console area only appears if there's text output
- [ ] Permanent data viewer button instead of responsive one (app-only)
    - [ ] Then don't have recording on by default, user should have to click "start recording"
- [ ] Restore small serial preview (https://github.com/Microsoft/pxt/issues/3845)

Jaqster:
I would also add that it's confusing to have Serial Data show up in a New Project when you have the device plugged in and running another program that's writing serial values.  I don't know if we should stop piping the data?  Or somehow make it clear this data is streaming from a live connected device, and not from anything in the current program?
<img width="413" alt="showdata" src="https://user-images.githubusercontent.com/22108954/34956211-01f63a80-f9dd-11e7-99ee-4beda13a2554.PNG">

CarlLyman:
Apparently right now it will only keep datapoints for 20 seconds. It would be really helpful if the datapoints could last for several minutes (15-30 minutes) when you are collecting data over a longer period but only sampling it every 20-30 seconds.

It would also be nice to be able to collect data over a 24 hour period when you are only sampling the data every 10 minutes or so.

If there was a known number of datapoints that can be collected, then if there is a certain duration that is needed to collect the data, then a person could adjust the frequency of datapoint collect with the time needed to collect the sample.

CarlLyman:
[...]

PR #3810, opened 2018-02-04
Serial view improvement

pelikhan:
- [x] don't drop stale graphs. this might interfere with experiment when data comes really slowly. Stale graphs can be dropped by hitting pause/start
- [x] store more data points. Smoothie drops the data as soon as it leaves the screen, storing a copy of the data on the side
- [x] specify the separator in the CSV file (`sep=...`) to avoid locale import issues


pelikhan:
https://github.com/Microsoft/pxt/issues/3650

PR #3818, opened 2018-02-05
Suspend serial when app is suspending

guillaumejenkins:


pxtwinrt/serial.ts
```diff
@@ -1,51 +1,84 @@
...
+        watcher.addEventListener("updated", deviceUpdated);
         watcher.start();
     }
 
+    export function suspendSerial() {
+        if (watcher) {
+            watcher.stop();
+            watcher.removeEventListener("added", deviceAdded);
+            watcher.removeEventListener("removed", deviceRemoved);
+            watcher.removeEventListener("updated", deviceUpdated);
```
pelikhan: set watcher to undefined?

pxtwinrt/serial.ts
```diff
@@ -1,51 +1,84 @@
...
+        }
+
+        Object.keys(activePorts).forEach((deviceId) => {
+            const port = activePorts[deviceId];
+            const device = port.device;
+            delete activePorts[deviceId];
```
pelikhan: not needed since you clear the map later on

pxtwinrt/serial.ts
```diff
@@ -1,51 +1,84 @@
...
+            device.close();
+        });
+        activePorts = {};
+    }
+
+    function deviceAdded(deviceInfo: DeviceInfo) {
+        if (!deviceNameFilter || !deviceNameFilter.test(deviceInfo.name)) {
```
pelikhan: we should use the vendor id instead of the name hack

pxtwinrt/serial.ts
guillaumejenkins: According to the docs, `stop()` resets the state, so I was only constructing it the first time and restarting it on `resume`. But I guess it's safer to just delete it and use a new one, so I'll do that

pelikhan:
Correct, all WinRT object should be destroyed. Their state might be bogus on resume.

pxtwinrt/serial.ts
guillaumejenkins: Those properties are not default on the `DeviceInformation` object, so it's hard to do it in master. For now I'll leave the name filter and when I port these to V0 I'll debug with micro:bits attached, which will let me debug to see if I have access to that info

pxtwinrt/serial.ts
```diff
@@ -1,51 +1,82 @@
...
 
     export function initSerial() {
         if (!pxt.appTarget.serial
             || !pxt.appTarget.serial.log
             || !pxt.appTarget.serial.nameFilter) return;
 
-        const filter = new RegExp(pxt.appTarget.serial.nameFilter);
-        const serialDeviceSelector = (Windows.Devices as any).SerialCommunication.SerialDevice.getDeviceSelector();
+        deviceNameFilter = new RegExp(pxt.appTarget.serial.nameFilter);
+        const serialDeviceSelector = Windows.Devices.SerialCommunication.SerialDevice.getDeviceSelector();
```
pelikhan: can we specify the productId/deviceId here?

guillaumejenkins: might be able to; will take a look on the v0 side and port that change back to master if it works

PR #3819, opened 2018-02-05
Port new cookie consent to V0

riknoll:
Bringing over the changes in master to V0. I did this entire merge by hand, so please take a look to make sure I didn't make any obvious mistakes. I still need to do testing.

I had to do a lot more granular instrumentation in this PR than I did in the previous one. In master, every target enters through the home screen which was pretty well covered by our telemetry (every UI element had a call to `pxt.tickEvent()` that I could instrument). App.tsx is much more complicated than the home screen because most of the methods that are called in `onClick` handlers are also called elsewhere. The cookie can only be set in response to user action so I had to edit every click handler to set the consent separately from the telemetry.

Another note is that I couldn't think of an easy way to instrument Blockly, so interacting with Blockly will not remove the banner. It's not a big deal because just about everything else will get rid of the banner including clicking "new project" and "download". If anyone has any suggestions for how to instrument Blockly let me know. I thought about using their [events](https://developers.google.com/blockly/guides/configure/web/events) but I wasn't confident enough that I could always determine whether those were fired as the result of user action or not.

### DIFF
--- a/Jakefile.js
+++ b/Jakefile.js
@@ -9,7 +9,7 @@ var cmdIn = ju.cmdIn;
 var strpSrcMap = ju.strpSrcMap;
 
 function tscIn(task, dir, builtDir) {
-    let command = 'node ../node_modules/typescript/bin/tsc'
+    let command = 'node ' + path.relative(dir, './node_modules/typescript/bin/tsc')
     if (process.env.sourceMaps === 'true') {
         command += ' --sourceMap --mapRoot file:///' + path.resolve(builtDir)
     }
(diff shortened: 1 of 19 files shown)


riknoll:
Tested across all Windows browsers. I'll merge once @abchatra approves

riknoll:
Oh, also @pelikhan what was the purpose of this commit: https://github.com/Microsoft/pxt/commit/d81073e3a8c3e318a9e915c685792138128a1ba3

I did it here as well but I don't know the context

pelikhan:
The “pxt” target is a bit weird and we have copy of the compiled js/css in the backend. That’s the doccdn . The /blb/ does not resolve in PXT docs

PR #3825, opened 2018-02-06
Serialport 6 - v0

pelikhan:
Upgrading to serialport 6.x requires also to run node 8.x in travis.
v3 - https://github.com/Microsoft/pxt/pull/3826

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "5.7.0"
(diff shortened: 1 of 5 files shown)


pelikhan:
I read on https://github.com/node-serialport/node-serialport/blob/master/UPGRADE_GUIDE.md that node 5,7 had been dropped. We don’t really an issue to move to 8.

PR #3826, opened 2018-02-06
upgrading to serialport 6

pelikhan:
Also requires node.js 8.*


riknoll:
Add the required Node version to the README

PR #3831, opened 2018-02-06
Update hello-world.md

sweedixie:
translated English to Turkish.


pelikhan:
Thank you for this translation. Translations are handled with https://crowdin.com/project/kindscript instead of in the source code. 
See https://makecode.com/translate.

PR #3835, opened 2018-02-07
Expandable blocks redo

riknoll:
This is a redo of https://github.com/Microsoft/pxt/pull/3794

Changes from the initial PR include:

1. Fixed the bugs @pelikhan mentioned
2. Fixed a crash when rendering insertion markers after inputs in the block had been hidden
3. Fixed a rendering bug where shadow blocks were appearing in insertion markers
4. Added tests to the Karma test suite
5. Added a "toggle" mode to this feature, described below

Clicking + and - reveals/hides one parameter at a time by default, but some APIs might want to hide/reveal multiple parameters all at once. I've added a comment attribute called `expandableArgumentMode` to enable this. There are three possible values:

* **normal** - reveal/hide one parameter at a time. This is the default
*  **toggle** - reveal/hide all parameters at once
*  **disabled** - disable the expandable blocks feature entirely. This is useful for if you want to get rid of the behavior but don't want to redo the `block` string to remove the extra parameters (also if we ever add comment attribute inheritance). 

Here is a demo of the three modes:

three_modes

You may ask, "what happens if I'm using toggle mode and I try to decompile a function call where only some of the parameters are given?" Have no fear! In that case, the block will decompile in the same way that it would in normal mode with only some of the parameters visible. If the user hits + or - then the block will return to the normal toggle behavior.

This PR is dependent on a blockly change here: https://github.com/Microsoft/pxt-blockly/pull/95


pelikhan:
LGTM, can we merge it in?

andybru:
When adding "karma.js" module
did you check the compatibility with the "pxt staticpkg" command?
It looks as if "karma.js" is missing in "pxt staticpkg"-directories.
nodehttpservertrace

PR #3838, opened 2018-02-07
[UWP App] Offer to download when device not found

guillaumejenkins:
image

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1068,17 +1068,10 @@ export class ProjectView
             .done();
     }
 
-    showDeviceNotFoundDialogAsync(docPath: string, resp: pxtc.CompileResult): Promise<void> {
+    showDeviceNotFoundDialogAsync(docPath: string, resp?: pxtc.CompileResult): Promise<void> {
         pxt.tickEvent(`compile.devicenotfound`);
-
-        let fileType = ".mkcd";
-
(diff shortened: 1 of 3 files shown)


webapp/src/app.tsx
```diff
@@ -1068,19 +1068,40 @@ export class ProjectView
...
+                    label: lf("Troubleshoot"),
+                    class: "focused",
+                    icon: "help",
+                    url: docPath,
+                    onclick: () => {
+                        pxt.tickEvent(`compile.troubleshoot`);
+                    }
+                },
+                {
+                    label: lf("Download {0} file", fileType),
```
pelikhan: Just put the file name

guillaumejenkins: `Download [filename]`?

pelikhan:
We already only show the file name at https://github.com/Microsoft/pxt/blob/321b7adfb7c36ac9cc60cf7d4e3d6d7e092544c2/webapp/src/cmds.ts#L92

webapp/src/app.tsx
pelikhan: Only filename to be consistent with https://github.com/Microsoft/pxt/blob/321b7adfb7c36ac9cc60cf7d4e3d6d7e092544c2/webapp/src/cmds.ts#L92

PR #3840, opened 2018-02-08
partial translation to indonesian 

raja-rizki:


pelikhan:
This file is not meant to be localized.

PR #3842, opened 2018-02-08
Web USB download dialogs

pelikhan:
- [x] added "webUSB" flag in "compile" section to enable it by default
- [x] restricted to Windows 10+ on Windows

- when download fails, a first dialog is shown
image
- the upgrade firmware dialog is displayed. The URL is board specific and specified in targetconfig.
image
- the pairing dialog is shown
image

### DIFF
--- a/webapp/src/cmds.ts
+++ b/webapp/src/cmds.ts
@@ -59,7 +59,7 @@ export function browserDownloadDeployCoreAsync(resp: pxtc.CompileResult): Promis
 }
 
 function showUploadInstructionsAsync(fn: string, url: string, confirmAsync: (options: any) => Promise<number>): Promise<void> {
-    const boardName = pxt.appTarget.appTheme.boardName || "???";
+    const boardName = pxt.appTarget.appTheme.boardName || lf("device");
     const boardDriveName = pxt.appTarget.appTheme.driveDisplayName || pxt.appTarget.compile.driveName || "???";
 
     // https://msdn.microsoft.com/en-us/library/cc848897.aspx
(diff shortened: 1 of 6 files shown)


webapp/src/cmds.ts
```diff
@@ -102,13 +102,138 @@ function showUploadInstructionsAsync(fn: string, url: string, confirmAsync: (opt
...
 }
 
+function askWebUSBPairAsync(resp: pxtc.CompileResult): Promise<void> {
+    const boardName = pxt.appTarget.appTheme.boardName || lf("device");
+    return core.confirmAsync({
+        header: lf("No device detected..."),
+        htmlBody: `
+<p><strong>${lf("Do you want to pair your {0} to the editor?", boardName)}</strong>
+${lf("You will get instant downloads and data logging.")}</p>
+<p class="ui font small">The pairing experience is a one-time process and available for Chrome version 61 and up.</p>
```
mmoskal: I certainly wouldn't call out specific version here - if people see it, it means they have the right version. The way you phrase it, they should go and check. I would actually not call out Chrome either, I don't see a point.

pelikhan: The point is that they might be confused if they try to use another browser (even on the same machine).

mmoskal:
Again, why do we need 3 dialogs? Maybe just one dialog, that has UF2 download button and pairing button?

pelikhan:
We've discussed the expression with @samelhusseini and we wanted to do more handholding.

webapp/src/cmds.ts
mmoskal: At least skip the version number - Chrome auto-updates and usage of older versions is near zero

mmoskal:
Don't users just always click the green button anyway?

mmoskal:
At least the green buttons should say something like: "Sure, let's do it!", "I got the right firmware", and "Pair"

Issue #3847, opened 2018-02-09
staticpkg command doesn't work on some @extended documentation


nedseb:
When I test to build a static package for pxt-adafruit, the CLI fail with the following error : 
```
INTERNAL ERROR: TypeError: Cannot read property 'replace' of undefined
    at Object.renderMarkdown (/dropbox/Dev/Makecode/PXT/pxt-adafruit/node_modules/pxt-core/built/pxt.js:102198:33)
    at renderDocs (/dropbox/Dev/Makecode/PXT/pxt-adafruit/node_modules/pxt-core/built/pxt.js:129880:33)
    at /dropbox/Dev/Makecode/PXT/pxt-adafruit/node_modules/pxt-core/built/pxt.js:131483:34
    at tryCatcher (/dropbox/Dev/Makecode/PXT/pxt-adafruit/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/dropbox/Dev/Makecode/PXT/pxt-adafruit/node_modules/bluebird/js/release/promise.js:512:31)
    at Promise._settlePromise (/dropbox/Dev/Makecode/PXT/pxt-adafruit/node_modules/bluebird/js/release/promise.js:569:18)
    at Promise._settlePromise0 (/dropbox/Dev/Makecode/PXT/pxt-adafruit/node_modules/bluebird/js/release/promise.js:614:10)
    at Promise._settlePromises (/dropbox/Dev/Makecode/PXT/pxt-adafruit/node_modules/bluebird/js/release/promise.js:693:18)
    at Promise._fulfill (/dropbox/Dev/Makecode/PXT/pxt-adafruit/node_modules/bluebird/js/release/promise.js:638:18)
    at Promise._resolveCallback (/dropbox/Dev/Makecode/PXT/pxt-adafruit/node_modules/bluebird/js/release/promise.js:432:57)
...
```
The problem seems come from the Markdown path resolving for the `@extended` documentation file. The problem is also present on pxt-maker target

PR #3848, opened 2018-02-09
Correct the path resolving of the `@extended` documentation (#3847)

nedseb:
First try to resolve the issue #3847 on the Markdown `@extended` documentation. I have just added an `"/"` at the end of the `dirs` array.

### DIFF
--- a/cli/nodeutil.ts
+++ b/cli/nodeutil.ts
@@ -467,7 +467,7 @@ export function resolveMd(root: string, pathname: string): string {
         return targetMd
 
     let dirs = [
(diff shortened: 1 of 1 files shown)


cli/nodeutil.ts
```diff
@@ -480,11 +480,11 @@ export function resolveMd(root: string, pathname: string): string {
             dirs.push(path.join(d, "..", cfg.additionalFilePath, "docs"))
     }
     for (let d of dirs) {
-        let template = tryRead(d + pathname)
+        let template = tryRead(d +"/"+ pathname)
```
pelikhan: Use path.join if possible

PR #3850, opened 2018-02-09
rearranging array blocks

abchatra:
Just rearranging the array blocks

### DIFF
--- a/libs/pxt-common/pxt-core.d.ts
+++ b/libs/pxt-common/pxt-core.d.ts
@@ -13,7 +13,7 @@ interface Array<T> {
       * @param items New elements of the Array.
       */
     //% help=arrays/push
-    //% shim=Array_::push weight=49
+    //% shim=Array_::push weight=50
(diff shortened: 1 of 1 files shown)


pelikhan:
screenshots?

pelikhan:
How about tweaking the block names to avoid the blank on the left
image

abchatra:
@pelikhan Any suggestions? This is a point fix to go in with Chibi.

pelikhan:
It looks better when there is a word on the left blank section.

abchatra:
image

abchatra:
@pelikhan quick sign off please? As discussed other changes will be in a separate work.

samelhusseini:
Also work that should be done in future PRs is ensure all blocks have shadows (we discussed having shadow blocks based on the array type). ie: if it's an array of numbers, show a number field, if it's an array of strings, show a string shadow block. 
Much better than having an empty slot, and it's actually quite clunky having to drag out a string / number block for each one.

PR #3868, opened 2018-02-13
Transfer include fix

pelikhan:
Back porting changes to the transfer.html page.

### DIFF
--- a/webapp/public/transfer.html
+++ b/webapp/public/transfer.html
@@ -1,13 +1,8 @@
 <html>
 
 <head>
-    <link rel="stylesheet" href="/doccdn/semantic.css" />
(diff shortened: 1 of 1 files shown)


abchatra:
What does this do? I am confused.

pelikhan:
It imports scripts from a previous domain.

abchatra:
What does this change do?

pelikhan:
Rewriting is not enabled fo transfer.html so doccdn is not resolved.

riknoll:
Why did you remove all the CSS? I'm fine with that hardcoded URLs as long as we are planning to kill this page eventually.

pelikhan:
It was already referenced and we are killing this page.

PR #3878, opened 2018-02-14
Adding PXT link command to the pxt cli

riknoll:
We'll have to publish this manually since it's in the JS cli and not `cli.ts`

### DIFF
--- a/pxt-cli/cli.js
+++ b/pxt-cli/cli.js
@@ -82,12 +82,18 @@ function link(dir) {
         process.exit(1);
     }
 
+    var pkgName = "pxt-core";
+    if (fs.existsSync(path.join(absPath, "package.json"))) {
+        var pkg = require(path.join(absPath, "package.json"));
+        pkgName = pkg.name;
+    }
+
     var modulesPath = path.resolve("node_modules");
 
     if (!fs.existsSync(modulesPath))
         fs.mkdirSync(modulesPath)
 
-    var corePath = path.join(modulesPath, "pxt-core");
+    var corePath = path.join(modulesPath, pkgName);
     if (fs.existsSync(corePath)) {
         var stats = fs.statSync(corePath);
         if (stats.isSymbolicLink()) {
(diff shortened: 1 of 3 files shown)


pxt-cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "pxt",
-  "version": "0.3.0",
+  "version": "0.3.1",
```
mmoskal: Should be 0.4.0

pxt-cli/cli.js
```diff
@@ -70,13 +70,60 @@ function target(n) {
...
+        console.log("Could not find " + absPath);
+        process.exit(1);
+    }
+
+    var modulesPath = path.resolve("node_modules");
+
+    if (!fs.existsSync(modulesPath))
+        fs.mkdirSync(modulesPath)
+
+    var corePath = path.join(modulesPath, "pxt-core");
```
mmoskal: What about `pxt-common-packages` ? You could look in `dir/package.json` for `name`

pxt-cli/cli.js
```diff
@@ -70,13 +70,60 @@ function target(n) {
...
+    if (fs.existsSync(corePath)) {
+        var stats = fs.statSync(corePath);
+        if (stats.isSymbolicLink()) {
+            fs.unlinkSync(corePath);
+        }
+        else {
+            try {
+                var rimraf = require("rimraf");
```
mmoskal: The package doesn't require rimraf, but maybe this is fine

riknoll: Yeah, but pxt-core does

riknoll:
If we put it in cli.ts then we'll have to maintain it in V0 and master. What's the issue with keeping it in the JS?

pxt-cli/cli.js
riknoll: That's a great idea

mmoskal:
OK, let's put it here. But it would be nice to support pxt-common-packages as well.

riknoll:
I'll add common-packages support

riknoll:
Done! @mmoskal take a look

mmoskal:
LGTM!

PR #3879, opened 2018-02-15
Dismiss banner on toolbox click and in native apps

riknoll:
This fix comes with a caveat: we need to be vigilant to make sure that we don't accidentally automate something in the future that causes Blockly to fire this event. Right now this works great

### DIFF
--- a/docfiles/pxtweb/cookieCompliance.ts
+++ b/docfiles/pxtweb/cookieCompliance.ts
@@ -1,4 +1,7 @@
 /// <reference path="../../localtypings/mscc.d.ts" />
+/// <reference path="../../pxtwinrt/winrtrefs.d.ts"/>
+
+declare var process: any;
 
 namespace pxt {
     type Map<T> = {[index: string]: T};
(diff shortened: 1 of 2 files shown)


webapp/src/blocks.tsx
```diff
@@ -408,6 +408,9 @@ export class Editor extends srceditor.Editor {
             if (ev.type == 'ui') {
                 if (ev.element == 'category') {
                     let toolboxVisible = !!ev.newValue;
+                    if (toolboxVisible) {
```
pelikhan: Add comments about caveat here

riknoll: Done

docfiles/pxtweb/cookieCompliance.ts
```diff
@@ -240,6 +248,14 @@ namespace pxt {
         }
     }
 
+    /**
+     * Checks for winrt and electron
+     */
+    function isNativeApp(): boolean {
+        return typeof Windows !== "undefined" ||
```
pelikhan: do we already have helper in BrowserUtil to detect that we are in WinRT or Electron?

samelhusseini: There is, see electron.isElectron || pxt.winrt.isWinRT();

pelikhan:
@riknoll does this need cherry picking into v3.3, v3.4?

abchatra:
@riknoll can you cherry pick to stable3.3 and master?

docfiles/pxtweb/cookieCompliance.ts
riknoll: I can't use the existing ones because the pxtweb stuff doesn't have any dependencies (it's embedded in every page). The electron check we use elsewhere is no good anyways because it is just checking the URL for `electron=1`

PR #3888, opened 2018-02-15
Enabling local storage when cookie is enabled

abchatra:
Appisights enabling local storage when cookie is enabled.


abchatra:
I will port to stable3.3 and v0 once merged here

PR #3890, opened 2018-02-16
fix sidedocs z-index issue

pelikhan:
Fix for https://github.com/Microsoft/pxt-microbit/issues/643

### DIFF
--- a/theme/sidedoc.less
+++ b/theme/sidedoc.less
@@ -41,7 +41,7 @@
   top: 40%;
   bottom: 40%;
   height: 20%;
-  z-index: 21; // blockly scrollbar is 20
+  z-index: 95; // blockly scrollbar is 20
   padding: 0;
   margin: 0;
   background: grey;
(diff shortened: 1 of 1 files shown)


pelikhan:
@guillaumejenkins could you text those scenarios?

theme/sidedoc.less
```diff
@@ -41,7 +41,7 @@
   top: 40%;
   bottom: 40%;
   height: 20%;
-  z-index: 10;
+  z-index: 21; // blockly scrollbar is 20
```
samelhusseini: Set it 95. This entire issue has been resolved in master by having relative zindexes in https://github.com/Microsoft/pxt/blob/master/theme/themes/pxt/globals/zindex.variables. 
Obviously this is a patch fix, but based on zindex.variables I think it's safest to set it to 95 (can't tell you exactly why, you'd have to read through that file).

PR #3893, opened 2018-02-16
Fix pointer events double trigger issue on Android

samelhusseini:
PR change: https://github.com/Microsoft/pxt-blockly/pull/98

Patch fix to fix the Android double trigger issue with pointer events conflicting with mouse events.

Note: master fix is slightly different, and is still underway (as it's based of scratch-blocks and not Blockly).


abchatra:
My understanding was we don't have android issue in microbit. As this issue requires iframe. Do we need this in v0?

samelhusseini:
The issue is that we are registering both mouse and pointer events and are "hoping" the browser links the two and prevents the other one from firing when it handles one of them.

Currently I've only seen browsers fail to link the two events on Android + iFrame, but lot's of things are changing in Chrome's pointer event handling code, that I'd rather not be at the mercy of browsers to ensure they do the right thing. 

Pointer events is an extension of mouse events and there's no need to register both events if pointer events are available. 

So in summary, this is somewhat precautionary, and somewhat (the right fix), and somewhat there might be other skews or instances where this bug portrays itself that we don't understand yet.

abchatra:
Lets wait for the release of microbit before merging this in. Without the repro, I would like this change to bake in.

samelhusseini:
To test use: https://makecode.microbit.org/app/f0b1bb3e6c38713c56939487c512b4b95ec7f872-1f92f12c2b

samelhusseini:
Closing these, will create a new PR with latest fixes.

PR #3897, opened 2018-02-17
Adding new parameter field for destructuring mutator

riknoll:
Fixes the confusing dropdown in the radio blocks. This PR is for V0 only.

new_parameter_field_take_two

Changes include:
1. No more dropdown, you now just get a rename dialog when you click on the parameter
2. Parameters cannot be renamed to an existing variable name
3. Parameters cannot be renamed to other parameters on the block (e.g. `receivedNumber` cannot be renamed to `receivedString`). That check ignores whitespace and capitalization (`receivedstring` and `received string` will both be rejected).
4. Attempting to delete the variable from another block will revert the parameter field to its default value. Previously it deleted the entire block and all of its children

This change is backwards compatible and hopefully subtle enough to not break screenshots. BTW, the above GIF doesn't have quotes around the variable name in the error dialog but I actually fixed that and was too lazy to re-record it.

### DIFF
--- a/pxtblocks/fields/field_parameter.ts
+++ b/pxtblocks/fields/field_parameter.ts
@@ -1,0 +1,132 @@
+/// <reference path="../../localtypings/pxtblockly.d.ts" />
+
+declare namespace Blockly {
+    interface Workspace {
+        createVariable(name: string): VariableModel;
+        getVariable(name: string): VariableModel;
+        renameVariable(oldName: string, newName: string): void;
+    }
+}
+
+declare namespace Blockly.Msg {
+    const RENAME_VARIABLE_TITLE: string;
+    const VARIABLE_ALREADY_EXISTS_FOR_ANOTHER_TYPE: string;
+}
+
+declare namespace Blockly.Events {
+    const VAR_DELETE: string;
+    const VAR_RENAME: string;
+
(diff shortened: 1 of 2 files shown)


pxtblocks/fields/field_parameter.ts
```diff
@@ -0,0 +1,132 @@
...
+declare namespace Blockly {
+    interface Workspace {
+        createVariable(name: string): VariableModel;
+        getVariable(name: string): VariableModel;
+        renameVariable(oldName: string, newName: string): void;
+    }
+}
+
+declare namespace Blockly.Msg {
+    const RENAME_VARIABLE_TITLE: string;
```
pelikhan: are we overriding those to make them localizable?

pxtblocks/fields/field_parameter.ts
```diff
@@ -0,0 +1,132 @@
...
+    const VARIABLE_ALREADY_EXISTS_FOR_ANOTHER_TYPE: string;
+}
+
+declare namespace Blockly.Events {
+    const VAR_DELETE: string;
```
pelikhan: localizable?

riknoll: Yep, these are both already localized in blocklyloader

pelikhan:
LGTM

PR #3899, opened 2018-02-20
Cache promise for targetconfig query

pelikhan:
Guarantees single call per client

### DIFF
--- a/pxtlib/main.ts
+++ b/pxtlib/main.ts
@@ -864,14 +864,17 @@ namespace pxt {
 
 
     let _targetConfig: pxt.TargetConfig = undefined;
+    let _targetConfigPromise: Promise<pxt.TargetConfig> = undefined;
     export function targetConfigAsync(): Promise<pxt.TargetConfig> {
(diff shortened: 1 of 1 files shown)


pxtlib/main.ts
```diff
@@ -864,14 +864,18 @@ namespace pxt {
...
+        if (!Cloud.isOnline()) // offline, don't try to download
             return Promise.resolve(undefined);
-        return _targetConfig ? Promise.resolve(_targetConfig)
-        : Cloud.downloadTargetConfigAsync()
-                .then(
-                js => { _targetConfig = js; return _targetConfig; },
-                err => { _targetConfig = undefined; return undefined; });
+        if (_targetConfigPromise) // cached promise
+            return _targetConfigPromise;
+        return _targetConfigPromise = Cloud.downloadTargetConfigAsync()
```
guillaumejenkins: small nit: should we try to avoid this pattern (using the return value of an assignment)? I think splitting this in 2 lines is clearer / less error-prone

PR #3903, opened 2018-02-21
Add support for math functions and constants

riknoll:
This PR adds a math function block. My criteria for which functions were included was mostly about which ones I have ever used in my life, and the list I came up with was:

1. `Math.sin(0)`
2. `Math.cos(1)`
3. `Math.tan(2)`
4.  `Math.sqrt(3)`
5.  `Math.ceil(4)`
6. `Math.floor(5)`
7. `Math.atan2(6, 7)`

And here are the ones that were cut:

1. `Math.asin(0)`
2. `Math.acos(1)`
3. `Math.atan(2)`
4. `Math.log(3)`
5. `Math.exp(4)`
6. `Math.trunc(5)`
6. `Math.round(6)`
7. `Math.imul(7, 8)`
7. `Math.idiv(9, 10)`

In addition to functions, I have also added support for constants on namespaces. It works exactly like enum blocks work; you add a dummy shim function (with the comment attribute `constantsShim`) and then annotate the constants you want in the dropdown with `blockIdentity="qName"`. I don't do any type checking.

@ganicke I filled in some placeholders for the tooltip text so take a look!

### DIFF
--- a/pxtblocks/blocklycompiler.ts
+++ b/pxtblocks/blocklycompiler.ts
@@ -659,6 +659,17 @@ namespace pxt.blocks {
 
     }
 
+    function compileMathJsOp(e: Environment, b: B.Block, comments: string[]): JsNode {
+        const op = b.getFieldValue("OP");
+        const args = [compileExpression(e, getInputTargetBlock(b, "ARG0"), comments)];
+
+        if ((b as any).getInput("ARG1")) {
+            args.push(compileExpression(e, getInputTargetBlock(b, "ARG1"), comments));
+        }
+
+        return H.mathCall(op, args);
+    }
+
     function compileProcedure(e: Environment, b: B.Block, comments: string[]): JsNode[] {
         const name = escapeVarName(b.getFieldValue("NAME"), e, true);
         const stmts = getInputTargetBlock(b, "STACK");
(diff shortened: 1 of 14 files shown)


riknoll:
It looks like this (except I removed a few functions from the list, see above).
math_functions

pelikhan:
Why not add them all?

riknoll:
@pelikhan take a look at the GIF, the dropdown is so long that it scrolls.

pelikhan:
Can we simply use a grid? All good here names are rather short.

pelikhan:
Let ‘s try to be 1-to-1 with JS.

riknoll:
We aren't 1-1 with JS. The functions we support are a subset of the math functions anyway, we are missing a lot of them

pelikhan:
Can I get log/exp? Does not make sense to expose E Without those.

riknoll:
I don't care strongly about what we include, I just want it to look nice and have the useful functions at the top. Never been a fan of grid pickers with text. @samelhusseini any ideas?

samelhusseini:
Most math functions have symbols so a grid of images might work well here

riknoll:
Works for me. If we wanted, we could split up the "math-y" ones from the JS ones. That is, move `ceil`, `floor`, `round`, and `trunc` to a different block since they do not have symbols.

pelikhan:
It is just going to be a long list I am afraid.

pelikhan:
We could split the number manipulation from the scientific operands (trig / exp / log etc..)

pelikhan:
Let's get those in and add more later.

PR #3904, opened 2018-02-22
Release all devices on reset, lang change and /beta

guillaumejenkins:
Ok so here's what was happening:
- `reader.loadAsync(32)` in our serial logic blocks the device until 32 bytes have been read
- Until 32 bytes have been read, we cannot close the device
- If we navigate (/beta, change lang, reset workspace) while there is an active `loadAsync()` call, even if the navigation occurs, the device remains stuck in that `loadAsync()` call; disconnecting the device doesn't work and reconnecting is impossible until it is unplugged
- `loadAsync()` can be cancelled, but the cancellation is not enough to put the device into a good state, we need to wait until the `loadAsync()` call fully returns (the `cancel()` will make it reject the promise)

So, the overview of this fix:
- If we're about to change lang, go to /beta or reset workspace, we stop all ongoing `loadAsync()` operations
- To do that, we call `cancel()`, then wait for the `loadAsync()` to **fully** return
- We achieve that by waiting on deferred promises that are only resolved in the `.done` of `loadAsync`
- Once all deferred have been resolved, it means we can now safely `close()` all serial devices
- After all devices are closed we can safely navigate away

Also, I fixed our `suspending` handler so that it correctly waits on our async operations

### DIFF
--- a/pxtwinrt/hiddeploy.ts
+++ b/pxtwinrt/hiddeploy.ts
@@ -68,6 +68,7 @@ namespace pxt.winrt {
                 });
             }, Promise.resolve<Windows.Devices.Enumeration.DeviceInformationCollection>(null));
 
+            let deviceId: string;
             return getDevicesPromise
                 .then((devices) => {
                     if (!devices || !devices[0]) {
@@ -77,12 +78,15 @@ namespace pxt.winrt {
                     pxt.debug(`hid enumerate ${devices.length} devices`);
                     const device = devices[0];
                     pxt.debug(`hid connect to ${device.name} (${device.id})`);
+                    deviceId = device.id;
                     return whid.fromIdAsync(device.id, Windows.Storage.FileAccessMode.readWrite);
                 })
                 .then((r: WHID) => {
                     this.dev = r;
                     if (!this.dev) {
                         pxt.debug("can't connect to hid device");
+                        let status = Windows.Devices.Enumeration.DeviceAccessInformation.createFromId(deviceId).currentStatus;
+                        pxt.reportError("winrt_device", `could not connect to HID device; device status: ${status}`);
                         return Promise.reject(new Error("can't connect to hid device"));
                     }
                     pxt.debug(`hid device version ${this.dev.version}`);
--- a/pxtwinrt/serial.ts
+++ b/pxtwinrt/serial.ts
@@ -3,6 +3,9 @@
 type DeviceWatcher = Windows.Devices.Enumeration.DeviceWatcher;
 type DeviceInfo = Windows.Devices.Enumeration.DeviceInformation;
 type SerialDevice = Windows.Devices.SerialCommunication.SerialDevice;
+type LoadOperation = Windows.Storage.Streams.DataReaderLoadOperation;
+type IBuffer = Windows.Storage.Streams.IBuffer;
+type AsyncOp<TResult, TProgress> = Windows.Foundation.IAsyncOperationWithProgress<TResult, TProgress>
 
 namespace pxt.winrt {
     let watcher: DeviceWatcher;
@@ -12,6 +15,8 @@ namespace pxt.winrt {
     interface DevicePort {
         info: DeviceInfo;
         device?: SerialDevice;
+        readingOperation?: LoadOperation;
+        cancellingDeferred?: Promise.Resolver<void>;
     }
 
     export function initSerial() {
@@ -43,7 +48,7 @@ namespace pxt.winrt {
         watcher.start();
     }
 
-    export function suspendSerial() {
+    export function suspendSerialAsync(): Promise<void> {
         if (watcher) {
             watcher.stop();
             watcher.removeEventListener("added", deviceAdded);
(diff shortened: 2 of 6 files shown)


pelikhan:
We should consider hosting the editor in a webview and use the standard hid interface to talk to the host (if that helps)

pxtwinrt/serial.ts
```diff
@@ -96,23 +121,40 @@ namespace pxt.winrt {
...
+            port.readingOperation.done((bytesRead) => {
+                port.readingOperation = undefined;
+                if (port.cancellingDeferred) {
+                    port.cancellingDeferred.resolve();
+                }
+
+                let msg = reader.readString(Math.floor(reader.unconsumedBufferLength / 4) * 4);
                 pxt.Util.bufferSerial(serialBuffers, msg, id);
-                readMore();
+                setTimeout(() => readMore(), 10);
```
pelikhan: why not immediately?

pxtwinrt/winrt.ts
```diff
@@ -84,19 +84,44 @@ namespace pxt.winrt {
...
+        return Promise.resolve()
+            .then(() => {
+                if (packetIO) {
+                    console.log(`disconnecting packetIO`);
+                    return packetIO.disconnectAsync();
+                }
+                return Promise.resolve();
+            })
+            .catch((e) => {
+                console.log(`error disconnecting packetIO: ${e.message}`);
```
pelikhan: use pxt.reportError

pxtwinrt/winrt.ts
```diff
@@ -84,19 +84,44 @@ namespace pxt.winrt {
...
+            })
+            .then(() => {
+                console.log("suspending serial");
```
pelikhan: use pxt.log

webapp/src/app.tsx
```diff
@@ -962,7 +962,13 @@ export class ProjectView
             disagreeLbl: lf("Cancel")
         }).then(r => {
             if (!r) return Promise.resolve();
-            return hidbridge.disconnectWrapperAsync()
+            return Promise.resolve()
+                .then(() => {
+                    return pxt.winrt.releaseAllDevicesAsync();
+                })
+                .catch((e) => {
+                    pxt.log(`Error disconnecting devices: ${e.message}`);
```
pelikhan: use pxt.reportException

webapp/src/lang.tsx
```diff
@@ -101,7 +101,11 @@ export class LanguagePicker extends data.Component<ISettingsProps, LanguagesStat
 
         if (langId !== initialLang) {
             pxt.tickEvent(`menu.lang.changelang.${langId}`);
-            window.location.reload();
+            pxt.winrt.releaseAllDevicesAsync()
+                .then(() => {
```
pelikhan: nit: you can use done here

pxtwinrt/serial.ts
guillaumejenkins: Reading from the device should be as much a background task as possible, but I can change the 10 to 1

webapp/src/app.tsx
guillaumejenkins: what's the diff between `reportException` and `reportError`?

webapp/src/lang.tsx
guillaumejenkins: done

;)

pelikhan:
Was think of a JS shell but correct. Does not help here. LGTM.

pelikhan:
Ping when it’s ready to test in beta

pxtwinrt/hiddeploy.ts
```diff
@@ -68,6 +68,7 @@ namespace pxt.winrt {
                 });
             }, Promise.resolve<Windows.Devices.Enumeration.DeviceInformationCollection>(null));
 
+            let deviceId: string;
```
pelikhan: Do you clear this value?

pxtwinrt/hiddeploy.ts
```diff
@@ -77,12 +78,15 @@ namespace pxt.winrt {
...
                     pxt.debug(`hid connect to ${device.name} (${device.id})`);
+                    deviceId = device.id;
                     return whid.fromIdAsync(device.id, Windows.Storage.FileAccessMode.readWrite);
                 })
                 .then((r: WHID) => {
                     this.dev = r;
                     if (!this.dev) {
                         pxt.debug("can't connect to hid device");
+                        let status = Windows.Devices.Enumeration.DeviceAccessInformation.createFromId(deviceId).currentStatus;
+                        pxt.reportError("winrt_device", `could not connect to HID device; device status: ${status}`);
```
pelikhan: Is this shown to the user? Use lf macro

pxtwinrt/winrt.ts
```diff
@@ -84,23 +84,52 @@ namespace pxt.winrt {
...
+        return Promise.resolve()
+            .then(() => {
+                if (packetIO) {
+                    pxt.log(`disconnecting packetIO`);
+                    return packetIO.disconnectAsync();
+                }
+                return Promise.resolve();
+            })
+            .catch((e) => {
+                pxt.reportError("winrt_device", `error disconnecting packetIO: ${e.message}`);
```
pelikhan: Use lf

pxtwinrt/winrt.ts
```diff
@@ -84,23 +84,52 @@ namespace pxt.winrt {
...
+            })
+            .then(() => {
+                pxt.log("suspending serial");
+                return suspendSerialAsync();
+            })
+            .catch((e) => {
+                pxt.reportError("winrt_device", `error suspending serial: ${e.message}`);
```
pelikhan: Use reportExption which logs the stacktrace

pxtwinrt/winrt.ts
pelikhan: reportException logs the stacktrace as well

pxtwinrt/hiddeploy.ts
guillaumejenkins: It's a local variable to the function, only used to get the device status in case of an error, not used outside of this function

pxtwinrt/hiddeploy.ts
guillaumejenkins: This is just for us to maybe help diagnose problems, users are shown a more generic and user friendly error message (see in the `.catch` at the end of this chain)

pxtwinrt/winrt.ts
guillaumejenkins: Not meant to be user facing, this is just for us

PR #3908, opened 2018-02-22
When UF2 device not found, attempt to switch to bootloader using 1200 baud

guillaumejenkins:
A generic way to enable targets to tap into HID connection failures, letting them deal with the failure in a customized way


pxteditor/editor.ts
```diff
@@ -257,6 +257,7 @@ namespace pxt.editor {
         showUploadInstructionsAsync?: (fn: string, url: string, confirmAsync: (options: any) => Promise<number>) => Promise<void>;
         fieldEditors?: IFieldCustomOptions[];
         toolboxOptions?: IToolboxOptions;
+        winrtHandleDeviceNotFoundAsync?: (io: pxt.HF2.PacketIO) => Promise<void>
```
pelikhan: can we make this work for any HID based onnection? (not just winrt)?

guillaumejenkins: Sure but the other scenarios are not user facing and are in independent code paths, so they can be done in a separate change

pelikhan: Why is this WinRT specific? Couldn't we could also face this via webusb as well? Just suggesting to make this callback a bit more generic so that it can be reused in other platforms.

guillaumejenkins: I don't think we can make it too generic, everything inside that callback on the Adafruit needs to use WinRT-specific APIs:
- Reconnect via WinRT serial
- Set the baudrate using WinRT SerialDevice API
- WinRT-specific device disconnect
- Re-init the WindowsRuntimePacketIo (`initAsync` is not standard in our base PacketIo class)

I'm open to suggestions if you can think of how to design a single generic WinRT / WebUSB extension point. Otherwise I suggest we use 2 separate callbacks, and then the target can refactor some common logic between the 2 callbacks if possible / necessary

pelikhan:
I get that the WinRT implementation is WinRT specific. But this callback takes a PacketIO which is our generic HID interface. Wouldn’t there be a WebUSB version of the callback too? We need to try avoiding to special too much for each platform.

localtypings/pxtparts.d.ts
```diff
@@ -44,21 +44,25 @@ declare namespace pxsim {
         i2cPins?: {
             SDA: string,
             SCL: string,
-        }
+        },
+        uf2BootloaderBaudSwitch?: {
```
pelikhan: nit: not specific to uf2

PR #3921, opened 2018-02-27
Remove electron app (was deprecated and broken)

guillaumejenkins:


pelikhan:
That line is the patch. I think you can try to remove it. 
https://github.com/Microsoft/pxt/blob/v0/webapp/public/index.html#L45

PR #3927, opened 2018-02-27
Fix clashses between radio mutator and serial namespace

riknoll:
Fixes https://github.com/Microsoft/pxt-microbit/issues/671


abchatra:
👍

PR #3929, opened 2018-02-28
Preserve sandbox hash in embedded pages

riknoll:
We use `#sandbox`  in the share page instead of `#pub`. I can't test this locally.


pelikhan:
@riknoll  v0 ?

riknoll:
I guess we should see if it works first? But yeah, needs to go to V0

PR #3934, opened 2018-02-28
Port search improvements and fix search crash

guillaumejenkins:
- Port search improvements from stable2.4:
  - Filtered (disabled and hidden) blocks are excluded from search
  - Localized category names are now searched
- Fix search crash caused by `pxt.appTarget` being undefined inside workers


pelikhan:
Any other commits in stable24 that we missed?

PR #3938, opened 2018-03-01
Fixes the leaky key in the telemetry URL

abchatra:
Var scriptIdRegex is not hoisted up and when the call to scrubURL function is invoked it sees scriptIdRegex being undefined and doesn't scrub the URL. 

It is ok to declare it as a local variable as all the browsers cache the regex parse tree. 

I will port to v0.

### DIFF
--- a/docfiles/apptracking.html
+++ b/docfiles/apptracking.html
@@ -24,12 +24,15 @@
                     telemetryItem.properties["WindowsApp"] = 1;
             });
         });
+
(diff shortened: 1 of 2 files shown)


abchatra:
Couple of more changes: 

Removed version from the customDimensions as it is never populated
Adding cookie present information to customDimensions. This will be easier to monitor if the usage is first time or not.

PR #3947, opened 2018-03-05
store key names in refmap for js targest

pelikhan:
In a JS target, also store key names.


pxtsim/langsupport.ts
```diff
@@ -81,6 +81,14 @@ namespace pxsim {
             if (runtime && runtime.refCountingDebug)
                 console.log(`RefObject id:${this.id} refs:${this.refcnt}`)
         }
+
+        static toAny(o: any): any {
+            if (o instanceof RefMap)
```
riknoll: Nitpick: 
```typescript
if (o instanceof RefMap || o instanceof RefCollection) {
    return (<RefMap | RefCollection>o).toAny()
}
```

PR #3953, opened 2018-03-07
Fixes in shims for simulator-only targets

mmoskal:
The following are now properly reflected on sim classes:
* `extends` clauses
* `get`/`set`
* constructors

Add `pxsim.syntheticRefAction` to make sim code look like user code.


pelikhan:
Do you have an example of ``syntheticRefAction``?

mmoskal:
Sure,
```
            // menu button by default switches into edit mode; user can override that
            this.onPressed(ControllerButton.Menu, syntheticRefAction(() => {
                board().toggleEditMode()
            }))
```

The user can then call `controller.onPressed(ControllerButton.Menu, () => somethingElse())`

mmoskal:
build now passes after merging master

PR #3954, opened 2018-03-07
adding array.foreach

pelikhan:
Note that both Array.Map and Array.forEach should leave a sparse array sparse, but they don't in our case.

### DIFF
--- a/libs/pxt-common/pxt-core.d.ts
+++ b/libs/pxt-common/pxt-core.d.ts
@@ -86,7 +86,7 @@ interface Array<T> {
 
     /**
       * Call a defined callback function on each element of an array.
-      * @param callbackfn A function that accepts up to two arguments. The map method calls the callbackfn function one time for each element in the array.
+      * @param callbackfn A function that accepts up to two arguments. The forEach method calls the callbackfn function one time for each element in the array.
       */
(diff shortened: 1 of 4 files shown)


libs/pxt-common/pxt-core.d.ts
```diff
@@ -83,6 +83,14 @@ interface Array<T> {
     //% helper=arrayMap weight=40
     map<U>(callbackfn: (value: T, index: number) => U): U[];
 
+
+    /**
+      * Call a defined callback function on each element of an array.
+      * @param callbackfn A function that accepts up to two arguments. The map method calls the callbackfn function one time for each element in the array.
```
mmoskal: The for each not map method

tests/compile-test/lang-test0/241arrayforeach.ts
```diff
@@ -0,0 +1,12 @@
+function testArrayForEach() {
+    msg("testArrayFoEach");
+    let strs: string[] = [];
+    [1, 2, 3].forEach(x => strs.push("X" + x))
+    let r = "A"
+    for (let s of strs) {
+        r += s
+    }
+    assert(r == "AX1X2X3", "map")
```
mmoskal: For each as well

PR #3963, opened 2018-03-09
Only import PNG files if the target has saveAsPNG turned on

riknoll:
I was copying and pasting a block in the blocks editor and it caused the entire app to freeze. I guess I had a PNG on my clipboard (not sure why copying a block didn't overwrite that) because it was hanging in the decode function. I don't know why it hung, but we shouldn't try to load PNG's in non-PNG supporting editors anyhow.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1911,7 +1911,7 @@ function isProjectFile(filename: string): boolean {
 }
 
 function isPNGFile(filename: string): boolean {
-    return /\.png$/i.test(filename)
+    return pxt.appTarget && pxt.appTarget.compile && pxt.appTarget.compile.saveAsPNG && /\.png$/i.test(filename);
 }
 
 function initLogin() {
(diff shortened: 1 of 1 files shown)


webapp/src/app.tsx
```diff
@@ -1911,7 +1911,7 @@ function isProjectFile(filename: string): boolean {
 }
 
 function isPNGFile(filename: string): boolean {
-    return /\.png$/i.test(filename)
+    return pxt.appTarget && pxt.appTarget.compile && pxt.appTarget.compile.saveAsPNG && /\.png$/i.test(filename);
```
mmoskal: I think you just need `return pxt.appTarget.compile.saveAsPNG && /\.png$/i.test(filename);`

riknoll: You definitely need the compile check, but yeah you're right about checking for appTarget. I was a tad overzealous and forgot this is in the webapp 😄

mmoskal: No you don't. https://github.com/Microsoft/pxt/blob/master/pxtlib/main.ts#L20

riknoll: I stand corrected! We definitely check it all over the place though.

mmoskal: I know... wasting precious cycles...

Issue #3919, opened 2018-02-26
please support minimal annotation (//% block)


tballmsft:
The new parser isn't backwards compatible with the minimal //% block syntax.

Please add default blockId and block string.

samelhusseini:
I think this is fixed? @riknoll

PR #3983, opened 2018-03-17
Support for `//% blockCombine`

mmoskal:
This adds support for a new `//% blockCombine` attribute that can be specified on properties to generate enum-like setter and getter. See changes in docs for docs. This includes compiler and decompiler support.

It also fixes support for `//% block` without arguments and inference of `blockId` - IMHO we should remove `blockId` from the non-shipped APIs for simplicity.

### DIFF
--- a/docs/defining-blocks.md
+++ b/docs/defining-blocks.md
@@ -400,7 +400,7 @@ class Foo {
     //% blockCombine
     y: number;
     // exposed with custom name
-    //% blockCombine="foo bar"
+    //% blockCombine block="foo bar"
     foo_bar: number;
     
     // not exposed
--- a/docs/defining-blocks.md
+++ b/docs/defining-blocks.md
@@ -8,7 +8,7 @@ simulator files.
 
 ## Category
 
-Each top-level javascript namespace is used to populate a category in the Block Editor toolbox. The name will automatically be capitalized in the toolbox.
+Each top-level TypeScript namespace is used to populate a category in the Block Editor toolbox. The name will automatically be capitalized in the toolbox.
 
 ```typescript-ignore
 namespace basic {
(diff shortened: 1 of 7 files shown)


docs/defining-blocks.md
```diff
@@ -379,6 +386,39 @@ have side effects. This happens automatically when there is no initializer,
...
+This can be done automatically with `//% blockCombine` annotation, for example:
+
+```typescript
+class Foo {
+    //% blockCombine
+    x: number;
+    //% blockCombine
+    y: number;
+    // exposed with custom name
+    //% blockCombine="foo bar"
```
pelikhan: I think the localization extraction logic only looks for "block" field so we should be using that instead.

mmoskal: `block=...` will (or should) expose it as a regular block; I guess we could have `block="..." combine`?

pelikhan: Yes that would work. It’ll get localized automatically.

mmoskal: hmm, then I guess we need `//% block combine`? or maybe just `//% combine` ?

riknoll:
Fixes https://github.com/Microsoft/pxt/issues/3919

docs/defining-blocks.md
pelikhan: Either should work but we need that stuff to be localizable.

pxtlib/service.ts
```diff
@@ -544,7 +544,14 @@ namespace ts.pxtc {
...
 
     const numberAttributes = ["weight", "imageLiteral"]
-    const booleanAttributes = ["advanced", "handlerStatement", "afterOnStart", "optionalVariableArgs", "blockHidden", "constantShim"]
+    const booleanAttributes = [
+        "advanced",
+        "handlerStatement",
+        "afterOnStart",
+        "optionalVariableArgs",
+        "blockHidden",
+        "constantShim"
```
mmoskal: Should add `blockCombine` here?

PR #3987, opened 2018-03-19
Snap to a specific version of the less compiler.

samelhusseini:
Latest less compiler (3.x) is not compatible with semantic ui. 
Latest semantic ui (2.3.x) has modal bug: semantic ui bug #6267.

Snapping to less compiler v 2.6 and semantic ui v2.2.


pelikhan:
Looks like we need to update the package.json in each target as well :(

samelhusseini:
Correct :(

Issue #3988, opened 2018-03-19
Regression in getBlocksInfo() w.r.t. handling of qualified names with no block nor block id.


matthewtebbs:
The function

`export function getBlocksInfo(info: ApisInfo): BlocksInfo {...}`

changed substantially between 3.5.4 and 3.5.7 of pxt-core. I believe the change was to support the `%blockCombine` attribute.

I'm finding name information for symbols of kind 'Module' are now no longer being pruned out.

The old code filtered out qualified names which did not have an associated block and blockId is below.

```javascript
const blocks = pxtc.Util.values(info.byQName).
    filter(s => !!s.attributes.block && !!s.attributes.blockId &&
    (s.kind != pxtc.SymbolKind.EnumMember));
```

The effect I'm seeing is that the filtered block information derived from the compiler's qualified name set includes items that are not blocks with categories and therefore fail later on when creating the Blockly toolbox, specifically those of symbol kind 'Module'.

These I believe are compiled from the following kind of declaration in my 'lib/core'.

```typescript
//% color=0 icon="\uf1fc" block="Design" weight=100
declare namespace design { /* paint-brush icon */
}
```

It's unfortunate as the end result is no toolbox shown although of course the loaded project runs just fine.

matthewtebbs:
FWIW I added a check to filter out modules as well as enums.

matthewtebbs:
} else if (!!s.attributes.block
                && !s.attributes.fixedInstance
                && s.kind != pxtc.SymbolKind.EnumMember
                && s.kind != pxtc.SymbolKind.Module) {
                if (!s.attributes.blockId)
                    s.attributes.blockId = s.qName.replace(/\./g, "_")

pelikhan:
Also impacts editor in non-english moe.

pelikhan:
https://github.com/Microsoft/pxt/pull/3989

samelhusseini:
I believe this is fixed. Closing.

matthewtebbs:
FWIW worked for me.

PR #3989, opened 2018-03-20
Fix for #3988

pelikhan:
With the new auto block logic, we try to generate blocks for namespaces, classes, interfaces.


pelikhan:
Probably, maybe another PR.

pelikhan:
Thanks for reporting the issue!

PR #3993, opened 2018-03-21
v0: Fix hex cache generation

guillaumejenkins:
Was a simple order of operation bug. I ended up just removing uf2 caching, since it will never be needed in v0.


pelikhan:
Can we also fix master?

PR #4000, opened 2018-03-22
add ``blockSetVariable`` to auto-wrap factory into set variable

pelikhan:
When function instantiates an object, we most likely want to store it in a variable. the ``blockSetVariable`` does that automatically.

image

TODO: support a formatting in "block" to specify the name of the return variable so that it can be localized.

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -239,7 +239,7 @@ namespace pxt.blocks {
         return result;
     }
 
-    function injectToolbox(tb: Element, info: pxtc.BlocksInfo, fn: pxtc.SymbolInfo, block: HTMLElement, showCategories = CategoryMode.Basic, comp: pxt.blocks.BlockCompileInfo, filters?: BlockFilters) {
+    function injectToolbox(tb: Element, info: pxtc.BlocksInfo, fn: pxtc.SymbolInfo, block: Element, showCategories = CategoryMode.Basic, comp: pxt.blocks.BlockCompileInfo, filters?: BlockFilters) {
         // identity function are just a trick to get an enum drop down in the block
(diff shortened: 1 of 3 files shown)


pxtblocks/blocklyloader.ts
```diff
@@ -362,6 +362,22 @@ namespace pxt.blocks {
                 });
             }
             else {
+                // if requested, wrap block into a "set variable block"
+                if (fn.attributes.blockSetVariable && fn.retType) {
+
+                    const setblock = Blockly.Xml.textToDom(`
```
riknoll: Escape these strings

pxtblocks/blocklyloader.ts
```diff
@@ -362,6 +362,22 @@ namespace pxt.blocks {
...
+<block type="variables_set" gap="${fn.attributes.blockGap || 8}">
+<field name="VAR" variabletype="">${fn.retType.toLowerCase()}</field>
+</block>`);
+                    {
+                        let value = goog.dom.createDom('value');
```
riknoll: We generally use `document.createElement()`. I'd rather not take a dependency on the Google closure stuff

pxtblocks/blocklyloader.ts
riknoll: Also, I don't think you need to specify the variable type field

pelikhan: i use ``fn.retType.toLowerCase()`` later on for the variable name, just ensuring nothing crashes here.

pxtblocks/blocklyloader.ts
pelikhan: We already use good.dom.createDom in various spots. Cleanup for another PR.

PR #4013, opened 2018-03-27
Client-side github repo caching

pelikhan:
Caches pxt.config and /text lookups in memory and/or pouchdb.

* /refs is not cached
* repoes pointing to "master" are not stored offline as they might be changing.

### DIFF
--- a/webapp/src/db.ts
+++ b/webapp/src/db.ts
@@ -137,3 +137,50 @@ class TranslationDb implements ts.pxtc.Util.ITranslationDb {
 }
 
 ts.pxtc.Util.translationDb = new TranslationDb();
+
+class GithubDb implements pxt.github.IGithubDb {
+    // in memory cache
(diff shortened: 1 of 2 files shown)


webapp/src/db.ts
```diff
@@ -129,11 +129,66 @@ class TranslationDb implements ts.pxtc.Util.ITranslationDb {
...
+ts.pxtc.Util.translationDb = new TranslationDb();
+
+class GithubDb implements pxt.github.IGithubDb {
+    // in memory cache
+    private mem = new pxt.github.MemoryGithubDb();
+    private table = new Table("github");
+
+    loadConfigAsync(repopath: string, tag: string): Promise<pxt.PackageConfig> {
+        // don't cache master
+        if (tag == "master")
```
mmoskal: O guess it would be safer to only cache `/v\d+/` and shas.

PR #4014, opened 2018-03-28
Light mode optimizations.

samelhusseini:
- Disable transitions
- Pass light mode to simulator on initialization and through runoptions
- Disable loading and dimmer animations
- Remove blockly workspace and drag filters
- Remove simulator reveal and remove transitions

Still yet to come, dynamically enabling light mode when compile is slow.

### DIFF
--- a/pxtsim/simdriver.ts
+++ b/pxtsim/simdriver.ts
@@ -38,6 +38,7 @@ namespace pxsim {
         partDefinitions?: pxsim.Map<PartDefinition>;
         mute?: boolean;
         highContrast?: boolean;
+        light?: boolean;
         cdnUrl?: string;
         localizedStrings?: pxsim.Map<string>;
         refCountingDebug?: boolean;
@@ -120,7 +121,7 @@ namespace pxsim {
             }
         }
 
-        private createFrame(): HTMLDivElement {
+        private createFrame(light?: boolean): HTMLDivElement {
             const wrapper = document.createElement("div") as HTMLDivElement;
             wrapper.className = 'simframe';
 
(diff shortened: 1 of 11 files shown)


pxtsim/simdriver.ts
```diff
@@ -130,6 +131,7 @@ namespace pxsim {
             frame.setAttribute('sandbox', 'allow-same-origin allow-scripts');
             frame.sandbox.value = "allow-scripts allow-same-origin"
             let simUrl = this.options.simUrl || ((window as any).pxtConfig || {}).simUrl || "/sim/simulator.html"
+            if (light) simUrl+= '?light=1';
```
pelikhan: This breaks offline caching. Pass light in the simulator options

samelhusseini: I do do that, but issue is simulator options are only passed in once loaded. Which means by that time, the SVGs have been constructed. I needed a way to pass in that option on initialization. 
Thoughts?

pelikhan: You'll need to encode that info in the hashtag arguments

samelhusseini: Cool yeah that should be easy. thanks

webapp/public/custom.css
```diff
@@ -220,4 +220,33 @@ div.blocklyWidgetDiv {
...
+    animation: none !important;
+}
+
+/* Remove blockly filter on drag */
+.light svg.blocklyBlockDragSurface > g {
+    filter: none !important;
+}
+
+.light .blocklyMainBackground {
+    fill: transparent !important;
```
pelikhan: should we have a solid background to speed things up?

samelhusseini: Yeah we could use the fill color for blocklySvg.

pelikhan:
Other idea: increase debouncing time for autoRunBlocksSimulator / autoRunSimulator / typecheck

pelikhan:
Let's make sure we track this. It is worth knowing if "light" mode is used a lot or not.

Issue #4016, opened 2018-03-28
Default theme hardcodes need for 'Powered-by-Microsoft-logo[-small].png' files.


matthewtebbs:
It'd be nice to have this style conditionally included based on the existence of the logo file and/or a variable that a client could define say '@beta' or '@external' or '@thirdparty' or some such.

As an example, the **pxt-sample** project will output the following. Any client of pxt-core that doesn't have the Microsoft logo images in their _/docs/static/logo_ folder will.

```
Skipped data-uri embedding of ../docs/static/logo/Powered-By-Microsoft-logo-small.png because file not found
Skipped data-uri embedding of ../docs/static/logo/Powered-By-Microsoft-logo.png because file not found
... repeated many times ...
```

samelhusseini:
Fixed by https://github.com/Microsoft/pxt/commit/65b83939241a07b3a836a355e9dd61033b9cfc28

PR #4017, opened 2018-03-28
Only include poweredby images when flag is set.

samelhusseini:
set @includePoweredByImages: true in target site.variables in order to include these files.

Fixes https://github.com/Microsoft/pxt/issues/4016

Couldn't find any way to test in less if a file exists.


samelhusseini:
Looks like I accidentally checked in an identical commit. Closing.

Issue #4022, opened 2018-03-28
Toolbox categories sorted differently between Blockly and Monaco views


matthewtebbs:
Two repro' cases here.

1) To see this look at the [Adafruit editor](https://makecode.adafruit.com/#editor). The Advanced categories 'Arrays' and 'Text' flip order between Blockly and Moaco editing views.

2) In my target's case I've noticed that if you don't assign a weight to your category it will be sorted differently between the two editing views, specifically a non-weighted category, will weight after built-ins in the Advanced group in the Blockly toolbox view, and before them in the Advanced group in the Monaco view.

Two fixes here. I have a PR that I'll attach.

1) Advanced categories are not assigned weights in the Monaco category view code; unlike their non-advanced counterparts.
Fix: to assign them weights in the snippet.

2) During insertion of the Blockly category an equal weighting will force insertion after an existing category, e.g. lower, in the list while in the Monaco category sort it will cause the new category order to be left as is.
Fix: to explicitly sort equal weightings after in the list as per the Blockly category insertion sort.

Tested with the Adafruit target and my own.

matthewtebbs:
[PR 4023](https://github.com/Microsoft/pxt/pull/4023)

samelhusseini:
Fixed. Thanks @matthewtebbs

PR #4023, opened 2018-03-28
Fix Monaco editor sorting to match Blockly editor sorting.

matthewtebbs:
Fix for issue [4022](https://github.com/Microsoft/pxt/issues/4022).


matthewtebbs:
FWIW not a huge fan of having the sort() predicate return only either -1 or +1 as this makes the return order both sort() algorithm and original array order dependent BUT I suspect both are constants in the code. A truer fix would be to user an insertion sort as per the Blockly namespace category insertion sort but even then the Blockly namespace category order is in part dependent on the order of the categories in the default Toolbox definition which is XML specified.

BUT I do think this improves the situation from what it was imho.

PR #4024, opened 2018-03-28
Various emitted JS optimizations

mmoskal:


pxtsim/runtime.ts
```diff
@@ -610,13 +611,17 @@ namespace pxsim {
             function buildResume(s: StackFrame, retPC: number) {
                 if (currResume) oops("already has resume")
                 s.pc = retPC;
+                let start = Date.now()
```
pelikhan: Util.now()?

mmoskal: We seem to be using `Date.now` much more than `Util.now`

mmoskal:
Tests fixed

PR #4035, opened 2018-03-30
Removing cookie from share page

abchatra:
I am still testing this and will update in case I find issues


pelikhan:
Looks like a good start.

PR #4043, opened 2018-04-02
Let simulator clients send event values other than the number type.

matthewtebbs:
It would have been nice to just convert,

```typescript
export class EventQueue {
```

to 

```typescript
export class EventQueue<T = number> {
```

but TypeScript's obligatory compile time type inferencing didn't allow the generic assignment of a default value such as 0 in,

```typescript
queue(id: number | string, evid: number | string, value: T = ????) {
```

Hence the derived class and method override from `EventQueueGeneric<T>`.

Existing code will have exactly the same behavior, including the 0 default value. Ran unit tests, lint and tested `pxt-adafruit` and my client against the new pxt-core.

Usage is in the client simulator code is like this,

```typescript
export class Mouse {
    constructor(public x: number, public y: number) {}
} 

export class MyBoard {
    private static _eventsMouse: EventBusGeneric<Mouse>;
    ...
    MyBoard._eventsMouse = new EventBusGeneric<Mouse>(runtime, value => [value.x, value.y]);
}
```

Such that we can have callbacks like this,

```typescript
    /**
     * Code to run when the mouse is moved over the render window.
     */
    //% blockId=scene_on_mousemove
    //% block="on mouse move"
    //% shim=scene::onMouseMove
    function onMouseMove(cb: (x: number, y: number) => void): void;
```

And the `EventBus` `listen` and `queue` methods behave as they are now and take a single value, in this example above a object of type `Mouse` and not a `number`.

### DIFF
--- a/pxtsim/runtime.ts
+++ b/pxtsim/runtime.ts
@@ -164,14 +164,16 @@ namespace pxsim {
         }
     }
 
+    export type EventValueToActionArgs<T> = (value: T) => any[];
+
     export class EventQueue<T> {
         max: number = 5;
         events: T[] = [];
         private awaiters: ((v?: any) => void)[] = [];
-        private mHandler: RefAction;
         private lock: boolean;
+        private _handler: RefAction;
 
-        constructor(public runtime: Runtime) { }
+        constructor(public runtime: Runtime, private valueToArgs?: EventValueToActionArgs<T>) { }
 
         public push(e: T, notifyOne: boolean) {
             if (this.awaiters.length > 0) {
@@ -195,8 +197,8 @@ namespace pxsim {
 
         private poke() {
             this.lock = true;
-            let top = this.events.shift()
-            this.runtime.runFiberAsync(this.handler, top)
+            const value = this.events.shift();
+            this.runtime.runFiberAsync(this.handler, ...(this.valueToArgs ? this.valueToArgs(value) : [value]))
                 .done(() => {
                     // we're done processing the current event, if there is still something left to do, do it
                     if (this.events.length > 0) {
@@ -209,18 +211,18 @@ namespace pxsim {
         }
 
         get handler() {
-            return this.mHandler;
+            return this._handler;
         }
 
         set handler(a: RefAction) {
-            if (this.mHandler) {
-                pxtcore.decr(this.mHandler);
+            if (this._handler) {
+                pxtcore.decr(this._handler);
             }
 
-            this.mHandler = a;
+            this._handler = a;
 
-            if (this.mHandler) {
-                pxtcore.incr(this.mHandler);
+            if (this._handler) {
+                pxtcore.incr(this._handler);
             }
         }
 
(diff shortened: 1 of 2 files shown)


matthewtebbs:
You can also imagine a class hierarchy like this below such that you could use a single event bus and queue up events with non-intrinsic values that self-extract to RefAction suitable action arguments.

If this PR is pulled this is the pattern I plan on converting to. :-)

```typescript
export abstract class EventValue {
    public abstract toActionArgs(): any[];
}

export class MouseEventValue extends EventValue {
    constructor(public x: number, public y: number) {}

    public abstract toActionArgs(): any[] {
        return [this.x, this.y];
    }
} 

export class MyBoard {
    private static _events: EventBusGeneric<EventValue>;
    ...
    MyBoard._events = new EventBusGeneric<EventValue>(runtime, value => value.toActionArgs());
}

```

matthewtebbs:
FWIW this is what I came up with given my local fork with this change.

```typescript
    export abstract class EventValue {
        public abstract toActionArgs(): any[];
    }

    export class MouseEventValue extends EventValue {
        constructor(public x: number, public y: number) {
            super();
        }

        public toActionArgs(): any[] {
            return [this.x, this.y];
        }
    }

    export class WorldEventBus extends EventBusGeneric<EventValue | string | number> {
        constructor(runtime: Runtime) {
            super(runtime, value => typeof value === 'object' ? value.toActionArgs() : [value]);
        }
    }
```

pxtsim/runtime.ts
```diff
@@ -195,8 +197,8 @@ namespace pxsim {
 
         private poke() {
             this.lock = true;
-            let top = this.events.shift()
-            this.runtime.runFiberAsync(this.handler, top)
+            const value = this.events.shift();
+            this.runtime.runFiberAsync(this.handler, ...(this.valueToArgs ? this.valueToArgs(value) : [value]))
```
pelikhan: where is the "this.handler" field defined?

matthewtebbs: It's the RefAction associated with the event via EventBus.listen(). That didn't change, well I did rename the variable _handler instead of mHandler but the property is the same.

matthewtebbs: Mis-read the question. It's defined below in the EventQueue class.

matthewtebbs: Class member-ordering is not defined in the lint rules :(

matthewtebbs:
Any concerns with this PR? For at least my client, and apparently the pxt-minecraft, it'd be valuable.

pxtsim/runtime.ts
pelikhan: We are passing ``[value]`` instead of ``value`` to the fiber. Is that expected?

matthewtebbs:
The spread ‘...’ operator will split the args. In the default case just a single arg.

matthewtebbs:
Yes, the rename on mHandler to _handler was gratuitous.

PR #4044, opened 2018-04-02
Adding output modes to the field_colour

riknoll:
This is non-breaking, the default behavior is unchanged. I've added two extra modes to the color picker for returning strings and return indices instead of hex values.


samelhusseini:
Tested on chibi?

riknoll:
Yep, works in Chibi

PR #4045, opened 2018-04-02
cli: git uploadrefs

samelhusseini:
Add cli support to directly upload git refs to the cloud. 

CLI command recurses through all git references (commit, tree, blob) and checks the cloud for their status in chunks (50 per request). It then uploads to the cloud any missing hashes. 

Finally it calls the cloud to update the git refs config file.

### DIFF
--- a/cli/gitfs.ts
+++ b/cli/gitfs.ts
@@ -1,10 +1,14 @@
+
 import * as child_process from "child_process";
+import * as util from 'util';
+
+import U = pxt.Util;
 
 export async function uploadRefs(id: string, repoUrl: string): Promise<void> {
     let gitCatFile: child_process.ChildProcess
-    let gitCatFileBuf = new PromiseBuffer<Buffer>()
+    let gitCatFileBuf = new U.PromiseBuffer<Buffer>()
 
-    let apiLockAsync = promiseQueue()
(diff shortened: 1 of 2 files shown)


cli/gitfs.ts
```diff
@@ -0,0 +1,444 @@
...
+
+export type SMap<T> = { [s: string]: T };
+
+interface QEntry {
+    run: () => Promise<any>;
+    resolve: (v: any) => void;
+    reject: (err: any) => void;
+}
+
+export function promiseQueue() {
```
mmoskal: Better use Util.PromiseQueue

cli/gitfs.ts
```diff
@@ -0,0 +1,444 @@
...
+        this.cache[key] = v
+    }
+
+    flush() {
+        this.size = 0
+        this.cache = {}
+    }
+}
+
+export class PromiseBuffer<T> {
```
mmoskal: Better use Util.PromiseBuffer

cli/gitfs.ts
samelhusseini: makes sense

cli/gitfs.ts
samelhusseini: makes sense

samelhusseini:
Ready.

PR #4083, opened 2018-04-10
Use prod react

samelhusseini:
Use the production version of react on prod builds (travis)
As per https://reactjs.org/docs/optimizing-performance.html#use-the-production-build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,4 +22,5 @@ addons:
       - g++-4.8
       - libusb-1.0-0-dev
 env:
+  - NODE_ENV=production
   - CXX=g++-4.8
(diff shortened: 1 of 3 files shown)


pelikhan:
Cherry pick to v0. Does uglify  destroy stack trace in prod?

samelhusseini:
According to the react docs, the only thing you're losing is "helpful" react warnings that are useful in development but should be taken out of production. so that doesn't include errors.

samelhusseini:
and nothing about a stack trace.

samelhusseini:
Also, note I'm not actually uglifying the code. There's one more step that we could do to actually uglify and minify the code, but that would make debugging the prod site much harder without sourcemaps (maybe something to look into down the track). 
I'm only doing the first two steps recommended by React to strip out unused code and debugging warnings.

PR #4098, opened 2018-04-12
Several changes and bug-fix to allow for combined property blocks that differ in return type.

matthewtebbs:
**pxtcompiler/emitter/decompiler.ts**
- Lookup for setter using qualified name that now includes property type (retType).

**pxtcompiler/emitter/service.ts**
- Fix check for readonly symbolinfo and include property declarations.

**pxtlib/service.ts**
- Support value of retType and not 'number'.
- To do this needed to expand the qualified name for the block to be
```typescript
const mkey = `${s.namespace}.${s.retType}` `
```
- Also added in a %blockCombineShadow attribute that will allow the set accessors to default to a shadow type, e.g. a color picker.

For example,

```typescript
//% blockNamespace=design
declare class Material {
    //% blockCombine
    //% blockCombineShadow=color_picker
    //% group="Material"
    //% shim=.color property
    public color: Color;

    //% blockCombine
    //% blockCombineShadow=color_picker
    //% group="Material"
    //% shim=.emissive property
    public readonly emissive: Color;

    //% blockCombine
    //% group="Material"
    //% shim=.roughness property
    public roughness: number;

    //% blockCombine
    //% group="Material"
    //% shim=.metalness property
    public metalness: number;
}
```

**webapp/src/monaco.tsx**
- Fixed pre-existing bug that meant block combined properties would not be populated correctly in the Monaco toolbox as they _currently_ have no snippets generated. Snippets are only generated for function like symbols.

### DIFF
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -235,7 +235,7 @@ namespace ts.pxtc {
                 snippet: service.getSnippet(decl, attributes)
             }
 
-            if ((stmt.kind === SK.GetAccessor && !getDeclarationOfKind(decl.symbol, SK.SetAccessor)) ||
+            if (stmt.kind === SK.GetAccessor ||
                 (stmt.kind === SK.PropertyDeclaration && isReadonly(stmt as Declaration))) {
                 r.isReadOnly = true
             }
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -113,6 +113,10 @@ namespace ts.pxtc {
         return false
     }
 
+    function isReadonly(decl: Declaration) {
+        return decl.modifiers && decl.modifiers.some(m => m.kind == SK.ReadonlyKeyword)
+    }
+
     function createSymbolInfo(typechecker: TypeChecker, qName: string, stmt: Node): SymbolInfo {
         function typeOf(tn: TypeNode, n: Node, stripParams = false) {
             let t = typechecker.getTypeAtLocation(n)
@@ -174,7 +178,7 @@ namespace ts.pxtc {
                 pkg,
                 extendsTypes,
                 retType: kind == SymbolKind.Module ? "" : typeOf(decl.type, decl, hasParams),
-                parameters: !hasParams ? null : (Util.toArray(decl.parameters)).map((p, i) => {
+                parameters: !hasParams ? null : Util.toArray(decl.parameters).map((p, i) => {
                     let n = getName(p)
                     let desc = attributes.paramHelp[n] || ""
                     let minVal = attributes.paramMin && attributes.paramMin[n];
@@ -231,8 +235,10 @@ namespace ts.pxtc {
                 snippet: service.getSnippet(decl, attributes)
             }
 
-            if (stmt.kind == SK.GetAccessor)
+            if ((stmt.kind === SK.GetAccessor && !getDeclarationOfKind(decl.symbol, SK.SetAccessor)) ||
+                (stmt.kind === SK.PropertyDeclaration && isReadonly(stmt as Declaration))) {
                 r.isReadOnly = true
+            }
 
             return r
         }
@@ -335,8 +341,6 @@ namespace ts.pxtc {
                 vs.declarationList.declarations.forEach(collectDecls)
                 return
             }
-
-            // if (!isExported(stmt as Declaration)) return; ?
 
             if (isExported(stmt as Declaration)) {
                 if (!stmt.symbol) {
--- a/pxtlib/service.ts
+++ b/pxtlib/service.ts
@@ -388,6 +388,7 @@ namespace ts.pxtc {
         function addCombined(rtp: string, s: SymbolInfo) {
             const isGet = rtp == "get"
             const isSet = rtp == "set"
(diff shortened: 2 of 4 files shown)


pxtlib/service.ts
```diff
@@ -384,23 +385,31 @@ namespace ts.pxtc {
...
             if (!ex) {
-                let paramName = s.namespace.toLowerCase()
-                ex = m[s.namespace] = {
+                const tp = `@${rtp}@`
+                const paramName = s.namespace.toLowerCase()
+                const paramValue = `value=${s.attributes.blockCombineShadow || ""}`;
+
+                ex = m[mkey] = {
                     attributes: {
+                        blockId: `${mkey}_blockCombine_${rtp}`,
```
riknoll: Changing the blockId is a breaking change for existing scripts. It's probably fine in this case since we don't have any editors that use this feature right now.

riknoll: If you wanted to avoid the break, you could special case `number` (but it probably doesn't matter)

pelikhan: Only used in Arcade right now

matthewtebbs: Good point. I'll special case generating the blockid re: number type. I'll update the PR.

matthewtebbs: Done. PR updated.

Validated with project scripts that pre-existed this change and that were generated with the old format blockId for number property types.

mmoskal:
This look great, certainly better than the current situation.

I understand the behavior is you get one getter (or setter or changer) block per type and they all look the same? did you try a single block? Were there problems with type inference?

pxtcompiler/emitter/service.ts
```diff
@@ -231,8 +235,10 @@ namespace ts.pxtc {
                 snippet: service.getSnippet(decl, attributes)
             }
 
-            if (stmt.kind == SK.GetAccessor)
+            if ((stmt.kind === SK.GetAccessor && !getDeclarationOfKind(decl.symbol, SK.SetAccessor)) ||
```
mmoskal: Aren't you getting double the setter with this change? I think the `addCombined("set", s)` will be called now once for the getter and once for the setter?

mmoskal: (anyway should be easy to fix, just check if `qName` already exists before `ex.combinedProperties.push(s.qName)`)

matthewtebbs: Don't understand the concern here re: addCombined() being called for setter and getter. The behavior should be the same as the old code prior to this PR. The difference is that I was seeing setter blocks getting generated when the property was readonly. This change fixes that by adding,

```typescript
(stmt.kind === SK.PropertyDeclaration && isReadonly(stmt as Declaration))) 
```

And the above change,

```typescript
if ((stmt.kind === SK.GetAccessor && !getDeclarationOfKind(decl.symbol, SK.SetAccessor)) || 
```

is intended to shore up code that I felt was wrong to begin with. It appeared that symbol info was created and marked as read-only when a statement was of _GetAccessor_ kind. That seemed weird to me. I made this particular change blind as I've been unable to get a case where the emitter is running over non-declared properties. I'm happy to revert, the only change really needed is the fix to check for read-only declared properties, i.e. the check for the _PropertyDeclaration_ kind.

matthewtebbs:
Let me know what more I can do to make this PR smell of roses and get merged to master :)

samelhusseini:
Just waiting for final review from @mmoskal.

pxtcompiler/emitter/service.ts
mmoskal: I think the change to remove isReadOnly from GetAccessors when there is also a setter is a good one. However, previously `addCombined("get", ...)` was called for the getter, and `addCombined("set", ...)` (and `"change"`) was called for the setter. Now, both `addCombined("get", ...)` and `addCombined("set", ...)` will be called for the  getter, leading to duplicate entries in the list (as the behavior for setter will be unchanged). Please just check for duplicates before adding to `ex.combinedProperties`.

matthewtebbs: I'll take a look.

matthewtebbs: Ah, I understand now. You know what I should revert the GetAccessor change.
In looking at the code it assumes a qualified named symbol per getter and per setter in the SymbolInfo. I had not realized this. I assumed it combined the two into one SymbolInfo. Personally I think that'd be better, but that is more code change than just checking for dupes on adding to combinedProperties. In the interest of not churning the code I'll revert part of this.

matthewtebbs: Thanks for clarifying your original review comment.

matthewtebbs:
Reverted part of the original PR given issue noticed by @mmoskal.

pxtcompiler/emitter/service.ts
matthewtebbs: Submitted an updated PR. I feel that if setting the GetAccessor symbol kind to read-only is a good one, and it fixes another issue, than that should be applied as an independent change in the interest of less churn on this PR.

matthewtebbs:
Yay. Thanks.

matthewtebbs:
Fixes https://github.com/Microsoft/pxt/issues/4099.

PR #4100, opened 2018-04-12
Move target logos into app cache

samelhusseini:
Move target images out of target.json and over into the release manifest to reduce the size of target.json and allow us to make use of browser app cache for those images between releases. 

Works locally, need to test on the cloud.

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -924,6 +924,11 @@ function uploadCoreAsync(opts: UploadOptions) {
         pxt.log(`hex cache:\n\t${hexFiles.join('\n\t')}`)
     }
 
+    let logos = (targetConfig.appTheme as any as Map<string>);
+    let targetImages = Object.keys(logos)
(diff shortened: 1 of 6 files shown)


samelhusseini:
I tried it but no, it looks like the hero banner already had an issue with staticpkg. I was going to ping @guillaumejenkins tomorrow to see if he's aware of any of these issues. Basically we reference ./static/... but in static package that has to look like ./docs/static in order to get to the right location. Perhaps we patch the pxtarget that we generate.

samelhusseini:
Fixed staticpkg. See: https://samelhusseini.github.io/test-pxt-sample
and https://samelhusseini.github.io/test-pxt-sample/release.manifest

pelikhan:
Try uploadpkg as well and we should be good!

samelhusseini:
FYI this does create more requests from the client (one for each image), but I think it's worth the compromise of caching between updates.

PR #4101, opened 2018-04-12
No more Jquery and semantic.JS!

samelhusseini:
Quite a few things going on here so let me explain. 

The only reason we're using Jquery in the editor is because Semantic.JS requires it but as it turns out many of our components are using React components and we've moved away from Jquery with the exception of: 
- Modal
- Dropdown
- Dimmer
- Loader

Using Jquery with React is an antipattern also and by removing the need for semantic.js we're able to get rid of Jquery as well which should speed  things up quite a bit. 
This change does just that. Now by removing semantic we also lose some of the work we did to make the semantic dimmer and modal accessible. It turns out however that work wasn't entirely complete and to spec, and there were a few holes in it, so I've gone ahead and re-implemented the Semantic Dropdown in react and made it accessible and to spec. 
The modal however (we were using React already), but the accessibility stuff was an addon and did direct DOM manipulation to achieve that. I've gone ahead instead and pulled in a small library that's fairly widely used called https://github.com/reactjs/react-modal. It's accessible out of the box, and I've adapted the semantic CSS rendering to that. 
For the dimmer and Loader I've implemented custom React components for those too. 

Now to the other part of the story. We use the helpers in core all over the place for three things: notifications, loader, and prompt dialogs (confirm and prompt). I didn't want to make changes all over the place so I've kept those. I've created a ConfirmDialog that utilizes the React Modal (instead of Jquery) but uses the same API to call a confirm. Since it's Async I've put it in its own wrapper and it's managed in its own node, ie: it renders itself when first needed, and maintains that dom element for subsequent dialogs. (Perhaps we can refactor this a little more down the track)
Same thing with notifications and loaders, kept the calling API but the internals have changed to use their own react doms. 

Finally, I've just gotten rid of Jquery dom manipulations all over the place, most of it was pretty straightforward. The simulator is still directly managing it's own DOM (frames). Another one that could be refactored a little more down the track. I've implemented the simulator entry / exit transition by hand (it's still authored in CSS, just executed without a library)

I've also updated React to the latest version and a bunch of other libraries as part of this change.
Change still needs to be tested cross browser, etc.

Test areas: 
- Accessibility of Modal (keyboard navigation)
- Accessibility of Dropdown (keyboard navigation)
- Key components: Home, Share, Add package, Blocks/JS, Sidedocs, Project settings, Full screen
- Less used areas: Extensions, Tracing, Serial editor, tutorial mode, youtube videos?
- Test all targets (looking at any use of core.requestAsync or use of Jquery in editor.js
- Test different views: mobile, tablet, computer
- **Ensure loadings correctly prevent using the editor in the background via keyboard navigation**

Tested: 
- [x] Adafruit extensions

Known issues: 
- [x] Home screen accessibility
- [x] Modal for package extensions

GG


samelhusseini:
Note: I also investigating pulling in https://github.com/Semantic-Org/Semantic-UI-React as part of this change. And I got pretty far using it for most purposes in our editor but it's accessibility story was lacking, and it was difficult to adapt. 
I also ran into issues treeshaking the components that weren't being used out, so it bloated our binary.

package.json
```diff
@@ -76,14 +76,14 @@
     "highlight.js": "^9.7.0",
     "lzma": "^2.3.2",
     "marked": "^0.3.9",
-    "fuse.js": "2.6.1",
+    "fuse.js": "^3.2.0",
```
pelikhan: Any api change?

webapp/src/app.tsx
```diff
@@ -921,41 +841,47 @@ export class ProjectView
...
-        // clear the hash
-        pxt.BrowserUtils.changeHash("", true);
-        this.setState({ home: true });
-        this.allEditors.forEach(e => e.setVisible(false));
-        this.homeLoaded();
-    }
-
-    private homeLoaded() {
-        pxt.tickEvent('app.home');
+    importFile(file: File) {
```
pelikhan: Maybe this should move too to importhelpers

samelhusseini:
Also note, we dont have a react app in our docs so I didnt refactor it out from pxtrunner or our docs, so semantic and jquery are still in our project just not the editor

webapp/src/codecard.tsx
```diff
@@ -16,11 +17,7 @@ export class CodeCardView extends React.Component<pxt.CodeCard, CodeCardState> {
         this.state = {};
     }
 
-    componentDidUpdate() {
-        ($('.ui.embed') as any).embed();
```
pelikhan: Do we support YouTube video embeds?

webapp/src/simulator.ts
```diff
@@ -221,23 +225,23 @@ export function run(pkg: pxt.MainPackage, debug: boolean, res: pxtc.CompileResul
 
 export function mute(mute: boolean) {
     driver.mute(mute);
-    $debugger.empty();
+    while (debuggerDOM.firstChild) debuggerDOM.removeChild(debuggerDOM.firstChild);
```
pelikhan: Add helper

samelhusseini:
Agreed, we need to do a deep test of this.

webapp/src/app.tsx
samelhusseini: It relies on resourceImporters which is attached to the editor (for extensions), so we'd have to pass it around or just keep the method there.

webapp/src/codecard.tsx
samelhusseini: Not sure, can you give an example?

samelhusseini: (also this is for codecard.tsx which is the react equivalent of the docrender one) and only used in the editor. I don't believe we support youtube embeds in the editor.

pelikhan: It’s used in docs but I don’t think we load the YouTube iframe in the editor.

webapp/src/simulator.ts
samelhusseini: Done.

package.json
samelhusseini: Two options that we don't use had been removed. https://github.com/krisk/Fuse/blob/master/CHANGELOG.md

webapp/src/codecard.tsx
samelhusseini: Cool. I think it's fine to get rid of it then.

pelikhan:
Looks like debugger UI is in here too. Probably want to merge that one first.

samelhusseini:
Where do you see the debugger UI. I made sure of not including it in?

samelhusseini:
If you're referring to the changes in simulator.ts (debuggerDOM). Those are unrelated, and basically are the old debugger UI. (the one you access from ?dbg=1). I'll be removing these once things settle down.

matthewtebbs:
FYI I created a local repo with this change merged into it. So far I do see an issue of the System menu on the top-right of the title bar being anchored in the wrong direction, e.g. it is clipped to the right side of the browser window. 

With this PR it happens in Firefox and Chrome (see below).

image

It does not happen with a version of pxt-core without this PR (see below).

image

samelhusseini:
Oh thanks for doing that Matt, I’ll check on that issue tonight.

samelhusseini:
That's been fixed, let me know if you find any other issues @matthewtebbs

matthewtebbs:
FWIW this PR is awesome. I've been running with it on a private fork and the user experience is definitely zippier.

samelhusseini:
Thanks!

abchatra:
Can you share the performance number?

matthewtebbs:
Likewise. Happy to help test on Ubuntu, macOS and Windows 10. Firefox, Chrome, Edge and IE (as installed on Windows 10).

samelhusseini:
Thanks heaps for the enthusiasm to help out with the testing process. I'm still figuring out a couple of known issues (see above list) before getting this in, but in the meantime, here's a link to a build off master with this change that you can use for testing:
https://makecode.adafruit.com/app/fde30de22b95b36daa110531946c11376ded0ff0-80cc42678c

samelhusseini:
Any word on testing efforts here? Anyone find any issues?

matthewtebbs:
Gave it a quick look on macOS on Safari, Chrome, Firefox latest versions. Found some minor bugs what may not be related to this change.

samelhusseini:
CTI found these accessibility regressions: 
- [ ] Can't navigate to 'Privacy & Cookies', 'Terms Of Use' and 'About' buttons
- [ ] The focus returns to the first element of the Editor page after closing the function under Dropdown
- [ ] Entering “Exit Project” modal, first interactable element is not focused
- [ ] Entering “Projects Settings” page, no element is focused

matthewtebbs:
Didn't test for accessibility. Next time I will. You get what you pay for :)

samelhusseini:
Once I fix the accessibility issues, I think this will be ready to merge.

matthewtebbs:
Me too.

PR #4104, opened 2018-04-12
better print support

pelikhan:
- [x] launch print dialog automatically
- [x] don't show edit/play icons under snippets


abchatra:
Awesome. Thanks @pelikhan

PR #4105, opened 2018-04-12
Updates to the variable flyout button UI.

samelhusseini:
screen shot 2018-04-12 at 11 10 24 am
screen shot 2018-04-12 at 11 10 20 am

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -2852,7 +2852,7 @@ namespace pxt.blocks {
             }
 
             let button = goog.dom.createDom('button') as HTMLElement;
-            button.setAttribute('text', lf("Make a Variable"));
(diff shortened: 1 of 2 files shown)


pxtblocks/blocklyloader.ts
```diff
@@ -2852,7 +2852,7 @@ namespace pxt.blocks {
             }
 
             let button = goog.dom.createDom('button') as HTMLElement;
-            button.setAttribute('text', lf("Make a Variable"));
+            button.setAttribute('text', lf("Make a Variable") + "...");
```
pelikhan: Check RTL rendering

samelhusseini: Ahhh yes. I knew this hack was a bad idea. Should I do RTL or should I just add it to the Lf?

pelikhan: just add it to the string.

Issue #3724, opened 2018-01-12
Project names allow seperators


microbit-mark:
micro:bit-support:4086
Makecode allows you to use a ',' seperator in the name of your project (project settings). The resulting .hex file that is downloaded cannot then be flashed to the micro:bit. Could the the use of a separator be prevented in the project name, fixed in the download or can users be advised of this in documentation?

abchatra:
If this is not fixed in alpha. Please reopen in pxt-microbit.

abchatra:
I have kept a tracking bug to validate: https://github.com/Microsoft/pxt-microbit/issues/906

abchatra:
# #

PR #4108, opened 2018-04-12
escape new project name

pelikhan:
Remove funny characters when creating project from template.


samelhusseini:
Cherry pick to v0? 
https://github.com/Microsoft/pxt/issues/3724 complains about a similar behaviour on the microbit.

Issue #3459, opened 2017-11-10
On the home screen selecting a language goes to the editor


riknoll:
If you use the secret tab menu to select a language, the page goes to the editor when it refreshes.

abchatra:
@guillaumejenkins I believe we fixed this. Does this need to be ported to v0?

microsoftsam:
I think @riknoll is referring to the accessibility menu in master. There is no home screen in v0, so nothing to port.

abchatra:
This is with the new language change options at the bottom of home page. Issue makes sense.

PR #4111, opened 2018-04-13
Always reload into home screen.

samelhusseini:
Always reload into the home screen.

If we had to choose between one and the other, home screen is safest.

Fixes https://github.com/Microsoft/pxt/issues/3459

### DIFF
--- a/pxteditor/editor.ts
+++ b/pxteditor/editor.ts
@@ -115,6 +115,8 @@ namespace pxt.editor {
         setState(st: IAppState): void;
         forceUpdate(): void;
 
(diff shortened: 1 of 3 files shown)


samelhusseini:
Reloading into home when in home. I see no reason to use another tag. We only need one for the editor. Otherwise default to home.

pelikhan:
Do we have a tick for editor reloads?

samelhusseini:
Doesn't look like it.

samelhusseini:
@guillaumejenkins?

PR #4113, opened 2018-04-13
merging keytar updates [v0]

pelikhan:
KeyTar APIs are now async. Cherry-pick from master.


abchatra:
No product impact right? Only cli changes?

pelikhan:
CLI only, nothing a user would see. keytar 4+ does provide pre-built binaries so ''npm install'' should not trigger node-gyp anymore.

abchatra:
Looks good.

PR #4114, opened 2018-04-13
a bit of cleanup of docs CSS

pelikhan:
The CSS for docs is a bit scattered around. As a result, the in-editor frame does not get the full CSS for the docs (fonts is smaller etc..). This PR centralizes a lot of the CSS involved in docs but probably missed some.

### DIFF
--- a/theme/docs.less
+++ b/theme/docs.less
@@ -26,6 +26,7 @@
 @docsCardHoverBorderColor: #1dacf4;
 
 #docs {
+    font-size: 12pt;
     background-color: @docsBackgroundColor;
     font-family: @docsPageFont !important;
     color: @docsTextColor;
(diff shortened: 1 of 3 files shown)


theme/docs.less
```diff
@@ -99,6 +115,98 @@
...
+    {
+        background-color: rgba(0,0,0,0.04);
+        color:transparent;   
+    }  
+    /* Blockly */
+    svg {
+        max-width: 100%;
+    }
+    .blocklyText {
+        font-family:'Monaco', 'Menlo', 'Ubuntu Mono', 'Consolas', 'source-code-pro', monospace !important;
```
pelikhan: Not sure whether we should keep this.

samelhusseini: use @blocklyFont

PR #4116, opened 2018-04-13
apptarget not available in webworker

pelikhan:


pxtcompiler/emitter/hexfile.ts
```diff
@@ -132,10 +132,10 @@ namespace ts.pxtc {
             }
         }
 
-        export function encodeVTPtr(ptr: number) {
-            let vv = ptr >> pxt.appTarget.compile.vtableShift
+        export function encodeVTPtr(ptr: number, opts: CompileOptions) {
```
riknoll: Is this function never used?

pelikhan: Maybe used in another branch by Michal.

riknoll: Looks like it's used in emitter/backvm.ts

mmoskal:
Thanks!

BTW, why don't we just make the `appTarget` available in the worker?

pelikhan:
We do modify apptarget in the editor so that would not work anymore.

PR #4117, opened 2018-04-13
Delay loading blockly

pelikhan:
This PR allows to delay load blockly and all dependencies.

* a new file "pxtblockly.js" is generated by the build. It contains blockly_compress.js, blocks_compress.js, en.js and pxtblocks.js. This is the file that gets delay loaded.
* moving toolbox stuff in pxtlib (CSS generation)
* cleanup import of "Util" which had creeped up all over the place.
* editor extensions have to be split. "fieldeditors" now contains any blockly field editors.

**Breaking changes**
* hue color specification not supported anymore. only #rrggbb


samelhusseini:
Yes, this change looks very familiar. Tried to do it a while back. 
A couple things:
- Monaco toolbox relies on Blockly CSS to be loaded, if you delay load Blockly we either load it when we're in Text, or we pull out the CSS we want to use and include it in pxt. 
- Monaco editor also relies on colors and some helper methods from pxtblocks, which can be stripped out (looks like you're doing that).

pelikhan:
Tackling that now.

PR #4119, opened 2018-04-14
Simplify the about dialog

abchatra:
Here are the screenshot before and after

Before: 
image

After:
image

Basically removes 3 reference to the board to just single one. Took cue from Edge about dialog

image


abchatra:
@Jaqster FYI

PR #4121, opened 2018-04-14
Fix delete and reset buttons

samelhusseini:
https://github.com/Microsoft/pxt/commit/79579505b3f8ec17c2a7624e5179649dd488ee47#diff-d8cf7897d76809964c79553d02e3e75c broke the delete and reset buttons. 

Mostly we do something real silly and attach to "approve buttons" in the modal, which I didn't realise at the time as it didn't make sense to add the approve class to all buttons, but anyway here's the fix.


pelikhan:
confirmDelete should use deleteLbl instead indeed.

samelhusseini:
All that's changing anyway. all good

PR #4125, opened 2018-04-14
Lazy blockly.css

pelikhan:
Lazy load blockly.css

### DIFF
--- a/webapp/public/index.html
+++ b/webapp/public/index.html
@@ -8,7 +8,6 @@
 
     <!-- @include appmeta.html -->
 
-    <link id="blocklycss" rel="stylesheet" data-rtl="/blb/rtlblockly.css" href="/blb/blockly.css" type="text/css">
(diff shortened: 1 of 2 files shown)


webapp/public/index.html
```diff
@@ -87,7 +86,9 @@
             "vs/language/typescript/src/worker": "/blb/vs/language/typescript/src/worker.js",
             "lzma/lzma_worker-min.js": "/blb/lzma/lzma_worker-min.js",
             "smoothie/smoothie_compressed.js": "/blb/smoothie/smoothie_compressed.js",
-            "pxtblockly.js": "/blb/pxtblockly.js"
+            "pxtblockly.js": "/blb/pxtblockly.js",
+            "blockly.css": "/blb/blockly.css",
+            "rtlblockly.css": "/blb/rtlblockly.css"
```
samelhusseini: so we load both?

pelikhan: we load the one we need (we used to load both)

samelhusseini: Cool!

Issue #1247, opened 2017-01-31
A short "Getting Started" guide doc to using the user interface


nevilh:
Could you create a short "Getting Started" guide doc to using the user interface, etc pointing out what everything on the screen does and add it as a menu item under the "help" drop down menu. E.g I've only just spotted the Undo/Redo Buttons in the bottom right corner - (I've been after an 'undo' in Block for over a year so I'm really pleased you have it in PXT, including being able to use <Ctrl Z> )

ganicke:
Editor feature overview page added in #4130

PR #4130, opened 2018-04-15
Add an 'editor features' base page

ganicke:
This is in response to #1247 asking for an editor overview page. This is a base page which is overridden in the target, if desired, to add customized feature images and possibly additional sections.


pelikhan:
It need to be reference from other pages (and summary)

ganicke:
Yea, but it hate to dirty up those top links with a child link...

image

PR #4132, opened 2018-04-15
Remove custom.css

samelhusseini:
Remove custom.css file and move CSS to relevant files. 

Split out the Blockly toolbox CSS from blocky.css (making it only contain Blockly specific SVG CSS). 
Blockly toolbox is considered part of the main UI and CSS is re-used for the monaco toolbox, so moving it into semantic.css (toolbox.less). 

Removing need to load blockly.css in monaco view, as all necessary blockly css is in toolbox.less and part of semantic.css

FYI, this requires targets to do the same, move blockly toolbox CSS to style.less. Will prepare PRs for major targets.


pelikhan:
Tedious but worth it :)

samelhusseini:
Ping me for any targets that have funky toolbox CSS when you first load into the monaco editor. Chances are they need this change applied. 

Also run `pxt buildcss --force` on any targets if your CSS looks all funky.

PR #4137, opened 2018-04-16
Don't overuse pxt loader

samelhusseini:
We don't want to show the big pxt loader for all loader, only the main editor one. 

Everything else can use the spinning wheel, or maybe customize later.

### DIFF
--- a/theme/themes/pxt/elements/loader.overrides
+++ b/theme/themes/pxt/elements/loader.overrides
@@ -3,28 +3,27 @@
 *******************************/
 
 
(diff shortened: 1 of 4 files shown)


pelikhan:
Not sure I understand the change. Screenshot?

samelhusseini:
Main loading when we start / switch to JS etc: 
screen shot 2018-04-16 at 9 50 03 am


Other loading like sidebar: 
screen shot 2018-04-16 at 9 49 23 am

PR #4144, opened 2018-04-16
swapping x/y in image format

mmoskal:
* the `shim=@f4` annotation for images now uses column-major, little endian order, used by most screens, improving performance
* image encoding refactored and simplified
* the data in `BoxedBuffer` is now word-aligned for better performance
* add `pxt builddatdts` to force rebuild of dal.d.ts

This needs to be synchronized with a PR in common packages: https://github.com/Microsoft/pxt-common-packages/pull/288

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -75,7 +75,7 @@ namespace pxt.blocks {
             const magic = data.charCodeAt(0)
             const w = data.charCodeAt(1)
             const h = data.charCodeAt(2)
-            if (magic != 0xf1 && magic != 0xf4)
+            if (magic != 0xe1 && magic != 0xe4)
                 return null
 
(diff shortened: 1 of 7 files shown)


pxtblocks/blocklyloader.ts
```diff
@@ -75,7 +75,7 @@ namespace pxt.blocks {
             const magic = data.charCodeAt(0)
             const w = data.charCodeAt(1)
             const h = data.charCodeAt(2)
-            if (magic != 0xf1 && magic != 0xf4)
```
pelikhan: does this matter?

mmoskal: I updated the magic, since it's a different format. We now generate the new magic everywhere. There shouldn't be anything in the wild, only used in 32 really.

PR #4146, opened 2018-04-16
Only load AI when cookie consent given

riknoll:
There are two cases where AI can still be loaded without the cookie:

1. There is an error when trying to contact our backend to get the mscc payload
2. The editor is in sandbox mode

Otherwise, we will only load AI if cookie consent is given. This affects all pages; not just the editor. For docs you will either have to have already given consent (i.e. in the editor) or click a link since those are the only interactive elements on those pages.

### DIFF
--- a/docfiles/pxtweb/cookieCompliance.ts
+++ b/docfiles/pxtweb/cookieCompliance.ts
@@ -110,7 +110,14 @@ namespace pxt {
             }
 
             all(info.Js || [], injectScriptAsync, msccError => {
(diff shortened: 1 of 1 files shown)


abchatra:
Looks good. Please port to v0

pelikhan:
Let's assume that the user is on a super-slow network where there is a large latency around loading the mscc.js file. There will be a window of time where the editor is loaded but that file is still in flight (the editor is cached so it might load faster). Do we cache events until AI is loaded?

riknoll:
@pelikhan yes, we've always done that. See the telemetry queue business in cookiecompliance.ts

PR #4155, opened 2018-04-17
Image dropdown field changes

samelhusseini:
Make onHide a class function rather than a field function for image dropdown.

Easier extensibility


pelikhan:
LGTM

PR #4156, opened 2018-04-17
Stop the audio buffer if one is running.

samelhusseini:
We added a playAudioBuffer to the audio context but never added a stop for it, so calling the context.stop was only stopping any "tone" sounds playing and not any audio buffers. 

Fixing it.


samelhusseini:
Wasn't the case when we had it in common packages. but good question I'll look it up.

samelhusseini:
No exceptions: https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/pause

PR #4157, opened 2018-04-17
Adding a decompiler flag for fixed instance parameters

riknoll:
Fixes a bug where the `Image` type could not be passed as a parameter to an API unless the thing being passed is declared as a fixed instance. This lets shim APIs declare that they can take indirect references to fixed instances as parameters.


pelikhan:
We really need docs for this stuff.

PR #4167, opened 2018-04-19
Mark property signature statements as readonly as well as declarations.

matthewtebbs:
In some work I'm doing with PXT attributed interfaces I've noticed that readonly property signatures in these are not marked read-only. They should be.


matthewtebbs:
There's a file dependency issue on clean builds I believe. Not me. :)

```
[built/webapp/src/app.js] cd webapp; node ../node_modules/typescript/bin/tsc
src/fileworkspace.ts(238,20): error TS2304: Cannot find name 'Util'.
```

pelikhan:
Try merging master again.

matthewtebbs:
Done.

matthewtebbs:
@mmoskal

PR #4168, opened 2018-04-19
Fix debugger move computation

pelikhan:
- [x] move block correctly
- [x] add overlay on top of simulator when paused. When the debugger is paused, it is very tempting to interact with the simulator... But the buttons are not quite paused which generate events and ... assert.

### DIFF
--- a/webapp/src/blocks.tsx
+++ b/webapp/src/blocks.tsx
@@ -663,11 +663,13 @@ export class Editor extends srceditor.Editor {
                 // don't center if block is still on the screen
                 const marginx = 4;
                 const marginy = 4;
-                if (p.x * s < marginx
-                    || (p.x + c.width) * s > m.viewWidth - marginx
-                    || p.y * s < marginy
-                    || (p.y + c.height) * s > m.viewHeight - marginy)
+                if (p.x * s < m.viewLeft + marginx
+                    || (p.x + c.width) * s > m.viewLeft + m.viewWidth - marginx
+                    || p.y * s < m.viewTop + marginy
+                    || (p.y + c.height) * s > m.viewTop + m.viewHeight - marginy) {
+                    // move the block towards the center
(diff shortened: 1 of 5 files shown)


webapp/src/debugger.tsx
```diff
@@ -263,7 +265,7 @@ export class DebuggerToolbar extends data.Component<DebuggerToolbarProps, Debugg
                 <div className={`ui compact borderless menu icon mini`}>
                     <div className={`ui item link dbg-btn dbg-handle`} key={'toolbarhandle'}
                         onMouseDown={this.toolbarHandleDown.bind(this)}>
-                        <sui.Icon key='iconkey' icon={`icon ellipsis vertical`} />
+                        <sui.Icon key='iconkey' icon={`xicon bug`} />
```
samelhusseini: This is for the handle?? Why you changing that.

webapp/src/blocks.tsx
```diff
@@ -663,11 +663,13 @@ export class Editor extends srceditor.Editor {
...
                 const marginy = 4;
-                if (p.x * s < marginx
-                    || (p.x + c.width) * s > m.viewWidth - marginx
-                    || p.y * s < marginy
-                    || (p.y + c.height) * s > m.viewHeight - marginy)
+                if (p.x * s < m.viewLeft + marginx
+                    || (p.x + c.width) * s > m.viewLeft + m.viewWidth - marginx
+                    || p.y * s < m.viewTop + marginy
+                    || (p.y + c.height) * s > m.viewTop + m.viewHeight - marginy) {
+                    // move the block towards the center
```
samelhusseini: Show off

webapp/src/debugger.tsx
pelikhan: not clear that the menu is related to debugging. it's just a bunch of arrows.

PR #4170, opened 2018-04-20
preload blockly on idle

pelikhan:


samelhusseini:
Nice

samelhusseini:
What happens if you're halfway through loading on idle. and you load the editor?

pelikhan:
We could also do this for the doc frame.

Issue #4179, opened 2018-04-20
Emitted code for property change/set value from blocks has extraneous newline.


matthewtebbs:
Minor bug but frankly annoying.

The issue is that the JsNode output for property value change statements will always have a double newline trailer as JsNode `mkStmt()` is called twice, once within the change/set node constructions, and once upon return from the Blockly compiler `compileStdCall()`.

The only time this is not the case is when `compileStdCall()` is called as part of an expression and in that context property change/set statements aren't valid anyway.

I have a PR. I'll submit.

matthewtebbs:
It'd be better if the JSNode helpers never constructed statements as part of another construction. This change will mean `mkComment()` is the only one that does that and I believe it'd be best to lift `mkStmt()` out of that function and into the single caller location. It'd be more consistent that way.

pelikhan:
Keep those PR coming!

matthewtebbs:
@pelikhan what's the best way to test python emit? I have a small change to /cli/pyconv.ts in this PR.

pelikhan:
@mmoskal

mmoskal:
Adafruit has a bunch of python libraries for microcontrollers. The idea was to convert these. However, I don't think anyone every seriously used it, so it never got any honest testing...

PR #4180, opened 2018-04-20
Fix double newline emit for property change/set statements.

matthewtebbs:
See issue https://github.com/Microsoft/pxt/issues/4179.

The JsNode output for property value change statements will always have a double newline trailer as JsNode `mkStmt()` is called twice, once within the change/set node constructions, and once upon return from the Blockly compiler `compileStdCall()`.

The only time this is not the case is when `compileStdCall() `is called as part of an expression and in that context property change/set statements aren't valid anyway.

Also moved out `mkStmt()` from `mkComment().` to make the JSNode helpers consistent. It is up to the clients to make statements as appropriate given their emit context.

The only caller of mkComment() that I could find was pyconv.ts so I fixed that up. Unsure how to test.

### DIFF
--- a/cli/pyconv.ts
+++ b/cli/pyconv.ts
@@ -1848,7 +1848,7 @@ function stmt(e: py.Stmt): B.JsNode {
         currErrs = ""
     }
     if (cmts.length) {
-        r = B.mkGroup(cmts.map(c => B.H.mkComment(c)).concat(r))
+        r = B.mkGroup(cmts.map(c => B.mkStmt(B.H.mkComment(c))).concat(r))
     }
     return r
 }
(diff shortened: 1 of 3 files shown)


samelhusseini:
@riknoll all yours!

riknoll:
Can you give me a repro script? A blocks screenshot is fine. I can't seem to get the behavior you're describing.

matthewtebbs:
Go to https://muddytummy.github.io/pxt-Impetus/ and open up the Test example.
My change fixes the issue with extra blank lines.

FWIW the issue only occurs when the JS node that is being constructed into a statement has expression _depth_. `mkStmt()` checks to see if the last JS node in the root node children is of type newline and does not emit if it is. I suppose this check could be made more robust also/instead.

riknoll:
Looks good. I'd suggest adding a test, but our Blockly compiler test-runner strips out whitespace when comparing against baselines so it wouldn't do much good 😄

PR #4182, opened 2018-04-20
Moving sprite editor into pxt-core 

riknoll:


samelhusseini:
Looks good. We should consider maybe having some way to lazy load certain field editors that matter to the specific editor.

pelikhan:
Field editors could be stored in common packages and selectively pulled in just like the sim.

PR #4189, opened 2018-04-22
more appcompat around docs.js

pelikhan:
- [ ] check the target version and include editor.js for pxt versions < 3.9.
- [ ] don't patch CDN urls that already patched

### DIFF
--- a/pxtlib/browserutils.ts
+++ b/pxtlib/browserutils.ts
@@ -350,8 +350,15 @@ namespace pxt.BrowserUtils {
         if (/^https?:\/\//i.test(path))
             return path;
         const monacoPaths: Map<string> = (window as any).MonacoPaths || {};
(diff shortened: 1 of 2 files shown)


pxtlib/browserutils.ts
```diff
@@ -350,8 +350,15 @@ namespace pxt.BrowserUtils {
...
             return path;
         const monacoPaths: Map<string> = (window as any).MonacoPaths || {};
-        const url = monacoPaths[path] || (pxt.webConfig.commitCdnUrl + path);
-        return url;
+        const blobPath = monacoPaths[path];
+        // find compute blob url
+        if (blobPath)
+            return blobPath;
+        // might have been exanded already
+        if (path.startsWith(pxt.webConfig.commitCdnUrl))
```
samelhusseini: Just FYI, can't use startsWith in IE. 
Fixing in the nojquery PR.

PR #4191, opened 2018-04-22
Debugger 9009

pelikhan:
- [x] unmangle variable names
- [x] render preview
- [x] large menu buttons
- [x] bring back ... icon
- [x] towards deeper exploration of object content


pelikhan:
@riknoll where is the logic that escape blockly variables? I need that logic to map back variables in the debugger.

samelhusseini:
I like that you've kept up the debugger versioning!!

PR #4196, opened 2018-04-22
Use intersection observer on home screen

samelhusseini:
Use intersection observer on home screen images (code card images)

Lazy load home screen images until they're in the view port. 

Current intersection observer support: https://caniuse.com/#feat=intersectionobserver
Basically not IE11 or Safari. 

We load images immediately if on IE11 or Safari. 
Tested on Safari

Here's a build of chibitronics to test: https://makecode.chibitronics.com/app/aa29b81db89ba511ae2424862ef6442d0f5d77c0-c26944f087

Should consider also using intersection observer for: 
- blocks
- docs (images), especially Adafruit projects with a lot of images


samelhusseini:
@abchatra comparison of offscreen images performance numbers: 

Before (chibitronics): 
screen shot 2018-04-23 at 8 15 35 am

After (chibitronics):
screen shot 2018-04-23 at 8 15 42 am

Huge reduction

samelhusseini:
Yeah

samelhusseini:
We should reduce the size of these images in any case

PR #4199, opened 2018-04-23
Debugger Object inspection

pelikhan:
- [ ] ability to expand the fields of an object in the variables view

clip

### DIFF
--- a/pxtsim/simdriver.ts
+++ b/pxtsim/simdriver.ts
@@ -10,7 +10,6 @@ namespace pxsim {
         onStateChanged?: (state: SimulatorState) => void;
         onSimulatorCommand?: (msg: pxsim.SimulatorCommandMessage) => void;
         onTopLevelCodeEnd?: () => void;
-        onVariables?: (msg: pxsim.VariablesMessage) => void;
         simUrl?: string;
         stoppedClass?: string;
     }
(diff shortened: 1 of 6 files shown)


pxtsim/simdriver.ts
```diff
@@ -385,18 +386,35 @@ namespace pxsim {
...
 
         private handleSimulatorCommand(msg: pxsim.SimulatorCommandMessage) {
             if (this.options.onSimulatorCommand) this.options.onSimulatorCommand(msg);
         }
 
+        private debuggerSeq = 1;
+        private debuggerResolvers: Map<(msg: DebuggerMessage | PromiseLike<DebuggerMessage>) => void> = {};
+
+        private clearDebugger() {
+            this.debuggerResolvers = {};
```
pelikhan: Need to cancel resolvers...

pelikhan:
Humm.. we do right?

mmoskal:
Maybe try creating a circular data structure in JS and see if it works (i.e., expands infinitely)

pelikhan:
It should... lemme try.

samelhusseini:
Yeah agreed on the UI, could be good to have an expand icon near the code that can be expanded (kinda like a tree expand / collapse icon)

pelikhan:
No effort work at ALL was done on the expand. Not even clear it is useful for blocks.

Issue #4195, opened 2018-04-22
%blockCombineShadow should allow for shadowing of object as well as value


matthewtebbs:
Mea culpa. I wrote this and in hindsight it'd have been great to let someone provide a shadow block to the object as well as the value. I'll update and submit a PR.

matthewtebbs:
Sweet.

________________________________

PR #4200, opened 2018-04-23
Allow for optional shadowing of object instance of property get/set.

matthewtebbs:
Allowable `%blockCombineShadow` strings:-

'{name shadow},' or '{value shadow}' or ',{value shadow}' or '{name shadow},{value shadow}' 

e.g.
```
    /**
     * Active camera.
     */
    //% blockCombine
    //% blockCombineShadow=world_scene,camera_perspective
    //% shim=.camera property
    camera: Camera;
```

Fixes https://github.com/Microsoft/pxt/issues/4195.

### DIFF
--- a/pxtlib/service.ts
+++ b/pxtlib/service.ts
@@ -403,11 +403,12 @@ namespace ts.pxtc {
 
                     // allowable %blockCombineShadow strings:-
                     //   '{name shadow},' or '{value shadow}' or ',{value shadow}' or '{name shadow},{value shadow}'
-                    const match = s.attributes.blockCombineShadow.match(/^([^,.]*),?([^,.]*)$/);
+                    const attribute = s.attributes.blockCombineShadow;
+                    const match = attribute.match(/^([^,.]*),?([^,.]*)$/);
                     if (match && match.length == 3) {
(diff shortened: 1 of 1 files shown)


matthewtebbs:
@riknoll ping.

pxtlib/service.ts
```diff
@@ -396,8 +396,26 @@ namespace ts.pxtc {
             let ex = U.lookup(m, mkey)
             if (!ex) {
                 const tp = `@${rtp}@`
-                const paramName = s.namespace.toLowerCase()
-                const paramValue = `value=${s.attributes.blockCombineShadow || ""}`;
+
+                let paramNameShadow, paramValueShadow;
```
riknoll: Declare the type

matthewtebbs:
I also noticed that I had switched from using '=' to ',' as the delimiter in the regex as it made more sense and forgot to update the ends with check. Just did. Thx.

matthewtebbs:
@riknoll ok so this change wins the award for issues.
Used `Util.endsWith()`, changed ends with check to "," and actually compare the original matched attribute as intended and not the matched parameter. I ran some tests to verify it does now work as intenetded. Doh!

Issue #4106, opened 2018-04-12
Deleting a variable from the parameter list causes the whole command to be deleted


teenne:
Steps: 
1. Adding a parameter to a command
2. Clicking the dropdown on the parameter-variable
3. Clicking "rename the variable"
4. The whole command disappears

Could renaming a variable be a feature on the "variables" tab?

abchatra:
Fixed

PR #4203, opened 2018-04-24
Fix array blocks, add new for loop blocks

samelhusseini:
Blockly drop: 
- Fixes issue with losing shadow blocks in arrays. (https://github.com/Microsoft/pxt-blockly/pull/102)
- Fixes alignment issue of flyout buttons in Edge and Firefox: (https://github.com/Microsoft/pxt-blockly/pull/101)
- Add data attribute for use in comment decompilation on the workspace comment block (https://github.com/Microsoft/pxt-blockly/pull/103) 
- Duplicate on drag support: https://github.com/Microsoft/pxt-blockly/pull/104

New array blocks: 
- adding a variables setter to make it easier to create a new list
- once you reach more than 5 items, the list becomes vertical. 
- adding a new item to the list adds a shadow block of type number or string depending on index 0's type. Only supports number and string, otherwise empty.

screen shot 2018-04-24 at 4 25 06 pm

newarrayblock

New for iterator blocks: 
- deprecating old blocks 
- fixes issues around deleting field variables that delete the entire block (https://github.com/Microsoft/pxt/issues/4106)
- allows dragging a new variable getter just by dragging the variable out of the block (rather than having to get it from the flyout)
- potential to use this feature further in custom blocks

screen shot 2018-04-24 at 4 21 18 pm

forindexduplicate

Fixes https://github.com/Microsoft/pxt-adafruit/issues/592
Fixes https://github.com/Microsoft/pxt-adafruit/issues/329


pxtblocks/toolbox.ts
```diff
@@ -195,25 +205,45 @@ namespace pxt.blocks {
...
+                        </shadow>
+                    </value>
+                    <value name="ADD1">
+                        <shadow type="text">
+                            <field name="TEXT">${lf("b")}</field>
+                        </shadow>
+                    </value>
+                    <value name="ADD2">
+                        <shadow type="text">
+                            <field name="TEXT">${lf("c")}</field>
```
riknoll: If you use lf here then it needs to be guaranteed to load after the translations. Right now I don't think that's guaranteed.

pxtcompiler/emitter/decompiler.ts
```diff
@@ -1111,10 +1111,12 @@ ${output}</xml>`;
                 r.handlers = [];
             }
             else {
-                r = mkStmt("controls_simple_for");
-                r.fields = [getField("VAR", renamed)];
+                r = mkStmt("pxt_controls_for");
+                r.fields = [];
                 r.inputs = [];
                 r.handlers = [];
+                r.inputs.push(mkValue("VAR",
```
riknoll: Instead of push, just put it in the variable initializer

pxtblocks/blocklycompiler.ts
```diff
@@ -1454,8 +1463,9 @@ namespace pxt.blocks {
         const variableIsScoped = (b: Blockly.Block, name: string): boolean => {
             if (!b)
                 return false;
-            else if ((b.type == "controls_for" || b.type == "controls_simple_for" || b.type == "controls_for_of")
-                && escapeVarName(b.getField("VAR").getText(), e) == name)
+            else if ((b.type == "controls_for" || b.type == "controls_simple_for" || b.type == "controls_for_of"
```
riknoll: We should move all these types into a constant array or something

pxtblocks/blocklycompiler.ts
```diff
@@ -256,6 +256,11 @@ namespace pxt.blocks {
         }
     }
 
+    function getLoopBlockField(b: Blockly.Block) {
```
riknoll: getLoopFieldBlock?

samelhusseini: Helper to get the variable field block depending on type of loop block. The new blocks have it in a shadow block VAR, while older blocks have it in a field on the block. 

Maybe call it getLoopVariableField ?

pxtcompiler/emitter/decompiler.ts
samelhusseini: Like so: 
```r.inputs = [mkValue("VAR",
                    getFieldBlock("variables_get_reporter", "VAR", renamed, true), "variables_get_reporter")];
```

or do you mean something else?

pxtblocks/blocklycompiler.ts
samelhusseini: Sure

samelhusseini:
Checked they decompile in docs :)

Issue #2317, opened 2017-06-18
edits to custom.ts not reflected in toolbox


tballmsft:
If you edit custom.ts and change the signature of functions then switch back to monaco toolbox, the old bindings are still there. Need to rerun the shim compiler upon user edits.

riknoll:
You also get some weird cases where the block definitions change and the old blocks mutate into garbage in the workspace. I don't know how deep we want to go into supporting this scenario but it could certainly use some work.

samelhusseini:
We need to keep refreshing the toolbox whenever the custom.ts file changes. Reason being you could have made a block change, so we must scan that file for any block definition changes.

Issue #2410, opened 2017-07-06
Monaco toolbox will only display categories if they have a color attribute


riknoll:
In the Blockly toolbox, the category will still show up it will just be the default color. This could be an issue for packages that forget to define a color.

PR #4205, opened 2018-04-25
New Blockly toolbox

samelhusseini:
Note: This is a huge change. 

The simple idea here is to move the Blockly toolbox to our web app (written in react). So far we'd been relying on Blockly for creating the toolbox, which meant that for any change to the toolbox we had to regenerate the toolbox XML that Blockly would understand and call it to update the toolbox. It also gave us little flexibility as to what we do in the toolbox. 

This change breaks that pattern and takes complete control of the toolbox. We generate the categories, sub categories, search etc. The only way we communicate to Blockly is by calling the Blockly toolbox flyout show method with an XML list of Blocks. 

This also means we're entirely in charge of caching the Block XML ourselves. 
Another big shift with this change is to move the need to duplicate logic between the Blockly and monaco toolboxes. This merges the two, were Blocks.tsx and Monaco.tsx are still in charge of their respective flyouts but the toolbox is created and managed separately. 

By doing this we've resolved a few issues: 
- Monaco toolbox not having search (now it's consistent with Blockly)
- Monaco toolbox not having subcategories (now it does)
- Search not showing operator or any block mutations (eg: plus, minus, times, etc)
- Accessibility of the Blockly toolbox (navigating around) and moving between toolbox categories and the search box. 

This change also refactors the toolbox extension logic to make it easier to define custom toolboxes. initExtensions is passed a copy of the toolbox schema and is able to return a different schema based on that, without the need to redefine blocks that should be re-used or moved under a separate category. 

Other changes: 
- Built in icons are now defined in one place (in blockIcons in pxtarget) rather than having all sorts of css around the place. We also no longer inject CSS icons, but add them inline.

Issues to look into for this change: 
- [x] (Won't fix in this change) https://github.com/Microsoft/pxt/issues/2410
- [x] Fixes https://github.com/Microsoft/pxt/issues/2317
- [x] Fixes https://github.com/Microsoft/pxt-adafruit/issues/504
- [x] Fixes https://github.com/Microsoft/pxt-adafruit/issues/673
- [ ] https://github.com/Microsoft/pxt-adafruit/issues/584

Known issues: 
- [x] (Fixed) Unrelated by I realised custom.ts is completely broken
- [x] (Fixed) Unrelated by the monaco editor isn't loading in RTL
- [x] (Fixed) There's a small gap when the toolbox is invisible (pxtjson)
- [x] (Fixed) Related to the nojquery change, the render method in blocklyrenderer relies on that. 
- [x] (Fixed) Unrelated a race condition when loading apis where in some cases we end up resetting the host and then calling apiInfo which loads apis with no blocks and you get a toolbox with just built in blocks, old issue but I now checking for empty host options before loading apiinfo

Tested: 
- [x] Toolbox categories
- [x] Toolbox with no categories
- [x] Tutorials (filter toolbox)
- [x] Toolbox extensions (pxt-neoanim)
- [x] Toolbox groups
- [x] Toolbox subcategories
- [x] Extending the toolbox (pxt-chibitronics)
- [x] Adding a custom.ts file
- [x] RTL languages
- [x] pxtJson -> Blocks (directly load workspace)
- [x] PXT json with no categories
- [x] Sandbox mode (no toolbox)
- [x] PXTJson -> Blocks (fresh reload)

Besides some of the features brought to the Monaco toolbox, everything should look just the same with this change. Ultimately this should make future work much easier, cleaner, and avoid bugs.

Build that can be used for testing: 
https://makecode.adafruit.com/app/c3f206fb3e6de20807d8f7f5be2855960cdc4e2a-bed48817a4

Tested on:
- [x] IE
- [x] Edge 16
- [x] Chrome Mac
- [x] Firefox Mac
- [x] Safari


matthewtebbs:
Wow! Can't wait to give it a spin.

samelhusseini:
Updated the build

pxtlib/toolbox.ts
```diff
@@ -54,6 +63,7 @@ namespace pxt.toolbox {
         }
     }
 
+    let namespaceStyleBuffer: string = '';
```
pelikhan: still used?

samelhusseini:
The branch has latest master merged, will look into what's going on there..

samelhusseini:
Yes let's get vendor team to test this build. Already done lots of testing but course more would be good.
I've already split out what I can, I think I could probably factor out a couple of very minor changes regarding RTL languages, but not worth it IMO.

samelhusseini:
Also I'm still testing it thoroughly on my end, no need to rush.

webapp/src/blocks.tsx
```diff
@@ -749,12 +759,13 @@ export class Editor extends srceditor.Editor {
         blocks.filter(b => b.isShadow_).forEach(b => b.dispose(false));
     }
 
-    private getBlocklyOptions(showCategories?: CategoryMode) {
+    private getBlocklyOptions(showCategories?: boolean) {
         let blocklyOptions = this.getDefaultOptions();
         Util.jsonMergeFrom(blocklyOptions, pxt.appTarget.appTheme.blocklyOptions || {});
-        const hasCategories = showCategories ? showCategories !== CategoryMode.None :
-            (blocklyOptions.hasCategories != undefined ? blocklyOptions.hasCategories : this.showToolboxCategories);
+        const hasCategories = showCategories != undefined ? showCategories :
```
riknoll: Checking if a Boolean is undefined?

webapp/src/blocks.tsx
```diff
@@ -763,7 +774,9 @@ export class Editor extends srceditor.Editor {
         return blocklyOptions;
     }
 
+    private blocklyOptionsCache: Blockly.Options;
```
riknoll: Nitpick: move to top of file

pxtlib/toolbox.ts
samelhusseini: Nope getting rid of it, and support to inject CSS by blocklyloader

webapp/src/blocks.tsx
```diff
@@ -789,68 +802,472 @@ export class Editor extends srceditor.Editor {
...
+        if (!this.blockInfo) return;
         // no toolbox when readonly
         if (pxt.shell.isReadOnly()) return;
 
-        pxt.debug('updating toolbox');
+        this.clearCaches();
+
+        const hasCategories = this.shouldShowCategories();
+
         const editor_ = (this.editor as any);
```
riknoll: Instead of casting to any you can declare those internal properties at the type at the top of this file. 
```typescript
declare namespace Blockly {
    declare interface Workspace {
        // This will be merged by TS with Blockly.d.ts
    }
}
```

webapp/src/blocks.tsx
```diff
@@ -789,68 +802,472 @@ export class Editor extends srceditor.Editor {
...
+
+    ///////////////////////////////////////////////////////////
+    ////////////         Flyout methods           /////////////
+    ///////////////////////////////////////////////////////////
+
+    public getBlocksForCategory(ns: string, subns?: string): (toolbox.BlockDefinition | toolbox.ButtonDefinition)[] {
+        let that = this;
+
+        function filterBlocks(blocks: toolbox.BlockDefinition[]) {
+            return blocks.filter((block => !(block.attributes.blockHidden || block.attributes.deprecated)
```
riknoll: Could this be split up?

webapp/src/blocks.tsx
```diff
@@ -789,68 +802,472 @@ export class Editor extends srceditor.Editor {
...
+            return getNonBuiltInBlocks(ns).concat(getExtraBlocks(ns, subns));
+        } else {
+            return getBuiltInBlocks(ns).concat(getExtraBlocks(ns, subns));
+        }
+    }
+
+    private flyoutBlockXmlCache: pxt.Map<Element[]> = {};
+    public showFlyout(treeRow: toolbox.ToolboxCategory) {
+        const { nameid, subns, icon, color, name, labelLineWidth } = treeRow;
+        const ns = nameid;
```
riknoll: Nitpick: `const { nameid: ns, subns, icon, color, name, labelLineWidth } = treeRow;` gets rid of the extra declaration

webapp/src/blocks.tsx
```diff
@@ -789,68 +802,472 @@ export class Editor extends srceditor.Editor {
...
+        if (this.flyoutBlockXmlCache[ns + subns]) {
+            pxt.debug("showing flyout with blocks from flyout blocks xml cache");
+            xmlList = this.flyoutBlockXmlCache[ns + subns];
+            this.showFlyoutInternal_(xmlList);
+            return;
+        }
+
+        function createHeadingLabel() {
+            const categoryName = name ? name :
+                `${subns ? `${Util.capitalize(ns)} > ${Util.capitalize(subns)}` : Util.capitalize(ns)}`;
```
riknoll: The outer template string seems unnecessary

webapp/src/blocks.tsx
```diff
@@ -789,68 +802,472 @@ export class Editor extends srceditor.Editor {
...
+                if (block.type == "button") {
+                    blockXml = that.getButtonXml(block as toolbox.ButtonDefinition);
+                } else {
+                    blockXml = that.getBlockXml(block as toolbox.BlockDefinition);
+                }
+                if (blockXml) xmlList.push(blockXml);
+            })
+        }
+
+        if (this.showFlyoutInternal(treeRow, createHeadingLabel, createGroupLabel, createBlocks)) {
```
riknoll: What is the difference between this method and the one with the underscore and can we eliminate one of them?

samelhusseini: showFlyoutInternal_ is used the one that calls Blockly and shows the blocks (takes in an xml list of blocks). 
showFlyoutInternal_ Calls blockly, defined in blocks.tsx
showFlyoutInternal in the superclass, shared between the two editors
showFlyout defined in each editor, defines the different helper methods needed ie: createHeading, createGroup, etc and calls showFlyoutInternal (in the super class).

webapp/src/blocks.tsx
samelhusseini: True

webapp/src/blocks.tsx
samelhusseini: Cool!

webapp/src/blocks.tsx
samelhusseini: In what way.

webapp/src/blocks.tsx
samelhusseini: TS doesn't like that, something about ambient class.

webapp/src/blocks.tsx
samelhusseini: I wanted to keep the cache next to where it's used. I think that's a better paradigm since it only matters in that context.

webapp/src/blocks.tsx
samelhusseini: differentiating between undefined and false.

webapp/src/blocks.tsx
```diff
@@ -435,15 +414,17 @@ export class Editor extends srceditor.Editor {
     }
 
     resizeToolbox() {
+        const blocklyArea = document.getElementById('blocksArea');
```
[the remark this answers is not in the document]
samelhusseini: Sure

webapp/src/blocks.tsx
```diff
@@ -317,6 +286,22 @@ export class Editor extends srceditor.Editor {
             oldToolboxPosition.call(this);
             editor.resizeToolbox();
         }
+
+        /**
+         * Override blockly methods to support our custom toolbox.
+         */
+        const that = this;
+        (Blockly.WorkspaceSvg as any).prototype.refreshToolboxSelection = function () {
```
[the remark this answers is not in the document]
samelhusseini: I need the parent (editor) object and the local (this) context for ws and flyout, see below.

webapp/src/blocks.tsx
guillaumejenkins: Nit: is using `typeof hasCategories === "undefined"` more explicit?

pelikhan:
Repro for the search issue:
clip

pelikhan:
Why did the screen go blank for a bug in the toolbox? Can we add a guard and recover from future/unknown bugs?

webapp/src/blocks.tsx
samelhusseini: Renamed showFlyoutInternal to abstractShowFlyout perhaps to make it clearer. Let me know if that makes sense.

webapp/src/blocks.tsx
samelhusseini: Resolved.

pelikhan:
We should have a local guard in the toolbox in case it misbehaves.

samelhusseini:
Guards only work for child components, and the idea of guards is to show some alternative UI. or do something to recover from the situation. (as well as log that it happened). 

We can design some guard for the toolbox, but I'd rather do it in a different change, this one is already pretty crammed.

pelikhan:
Considering the bug destroyed the editor DOM, this change should contain some local guards. Search is definitely a difficult one to test so maybe start there.

samelhusseini:
This is just what React does when any react element crashes. I don't have enough requirements to create a guard. what should it do? what should it look like? 

This smells like scope creep. @abchatra to consult.

pelikhan:
When search crashes, it should return an empty result set and report an error in the backend. Not take down the whole editor. It is impossible to test out all possible search queries, unlike opening every categories.

samelhusseini:
Search didn't take down the editor. The flyout did. Adding a react guard won't be guarding against search crashing. Any other flyout using the same block would have resulted in the same issue. 

I understand wanting to have a guard around the toolbox. That makes sense. We probably want it to do something like show a "reload button" and clicking on that will reload the toolbox. That would be my take on that.

samelhusseini:
Anyway reason all this componentDidCatch is a much bigger feature is that you need to have entry points for recovery. If something goes wrong with one component, you want to show an alternate UI but also have a way to reload that component and refresh it's state. For many independent components that a simple task, just render again.. but for something like our editors (which are basically embedded in our main app), they crash the app immediately. 

Something to consider for a future refactor, at this point, the most I'm willing to do in this PR is potentially add a reload button for the toolbox specifically, but that's going to avoid issues like you saw with search before. (as it's not that toolbox that crashed but the Monaco editor, ie: the app)

webapp/src/blocks.tsx
```diff
@@ -401,23 +388,15 @@ export class Editor extends srceditor.Editor {
...
-                        }
-                        else if (this.showToolboxCategories === CategoryMode.Basic) {
-                            this.showToolboxCategories = CategoryMode.All;
-                        }
-                        this.refreshToolbox();
-                    }
                 }
             }
         })
+        if (this.shouldShowCategories()) {
```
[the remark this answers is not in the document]
samelhusseini: Yeah it's different. In the first case, it's trying to force to show / or not show categories. There's a weird circular dependency that I don't want to mess up. We can sit down to try and figure it out when we get a chance but the way I understand it is: 
- !hasCategories flag in Blockly options indicates the target doesn't want categories, so we have to load the flyout toolbox on initial load
- filterToolbox is another entry point for something like the tutorial engine trying to indicate that we should not load in toolbox mode, although in that case the initial load will be with categories, and when we get passed the blocks to filter, we will re-inject blockly (see refreshToolbox) with hasCategories to false. 
- but we also keep a local variable showCategories to indicate the current state, probably can get rid of that, future change maybe
- and then there's the editor state that indicates the desired output by a third party controller (editor controller).

pelikhan:
LGTM. Next we should bump various editors to keep testing it out over the next week

PR #4210, opened 2018-04-26
Improved workspace comment compilation/decompilation

riknoll:
Adds support for grouping workspace comments with event handlers. Any top level comment in Typescript will now be grouped with the nearest event upon decompilation/formatting. Comments are placed to the right of the event block and stacked vertically; I originally tried laying them out above the block but the right looks much more natural in practice. The formatting algorithm is pretty naïve so we could definitely improve upon it.

Likewise, the Blockly compiler now emits the comments above the nearest event handler in the blocks. The grouping from blocks -> TS is done with a heuristic based on distance and bounding boxes of comments and can probably also be improved but it handles the round-trip scenario pretty well.

Other nice fixes:
1. workspace comments no longer compile with too many newlines
2. the decompiler now does a rough estimate of the size of workspace comments
3. onStart is always placed in the top-left of the workspace

### DIFF
--- a/pxtblocks/blocklycompiler.ts
+++ b/pxtblocks/blocklycompiler.ts
@@ -11,6 +11,12 @@ namespace pxt.blocks {
     let placeholders: Map<Map<any>> = {};
     const MAX_COMMENT_LINE_LENGTH = 50;
 
+
+    interface CommentMap {
+        orphans: Blockly.WorkspaceComment[];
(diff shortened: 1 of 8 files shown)


samelhusseini:
We talked about using JSON for the data field to allow us to use the field for other things in the future. Still POR or has something changed?

PR #4212, opened 2018-04-26
Fix reset and import file dialogs

samelhusseini:
Regression from nojquery change. 

Fixes https://github.com/Microsoft/pxt-adafruit/issues/702

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1555,11 +1555,25 @@ export class ProjectView
     }
 
     showImportFileDialog() {
-        dialogs.showImportFileDialog();
+        dialogs.showImportFileDialog().done(res => {
+            if (res) {
+                pxt.tickEvent("app.open.file");
+                this.importFile(res);
+            }
+        });
     }
 
     showResetDialog() {
-        dialogs.showResetDialog();
+        dialogs.showResetDialog().done(r => {
+            if (!r) return Promise.resolve();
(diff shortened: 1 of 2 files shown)


pelikhan:
Looks good.

webapp/src/app.tsx
```diff
@@ -1555,11 +1555,25 @@ export class ProjectView
     }
 
     showImportFileDialog() {
-        dialogs.showImportFileDialog();
+        dialogs.showImportFileDialog().done(res => {
```
pelikhan: rename to showImportFileDialogAsync

webapp/src/app.tsx
```diff
@@ -1555,11 +1555,25 @@ export class ProjectView
...
+            if (res) {
+                pxt.tickEvent("app.open.file");
+                this.importFile(res);
+            }
+        });
     }
 
     showResetDialog() {
-        dialogs.showResetDialog();
+        dialogs.showResetDialog().done(r => {
```
pelikhan: rename to showResetDialogAsync

webapp/src/dialogs.ts
```diff
@@ -135,22 +135,12 @@ export function showReportAbuse(pubId?: string) {
 }
 
 export function showResetDialog() {
```
pelikhan: add Async suffix

PR #4213, opened 2018-04-26
Auto focus inputs on prompts

samelhusseini:
Most browsers support it: https://caniuse.com/#feat=autofocus
IOS doesn't but that's okay in this case.


pelikhan:
Should it be always on for the Input type in sui?

_____________________________

samelhusseini:
Nope. Only in modals. Otherwise it will steal focus all whenever there's an input. (like the project name input)

PR #4215, opened 2018-04-26
Add an error guard around app

samelhusseini:
React no has support for error boundaries: https://reactjs.org/docs/error-boundaries.html

This would mean that if an unhandled error occurs somewhere in the React components, we can display a message like so, and get users to reload the page, log the error etc. 

screen shot 2018-04-26 at 2 02 21 pm

Alternatively, we can just reload the page for them but I feel like that's a little intrusive. 

thoughts?

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -444,6 +444,10 @@ export class ProjectView
         this.setState({ hasError: true });
         // Log critical error
         pxt.tickEvent('pxt.criticalerror', { error, info });
+        // Reload the page in 2 seconds
+        setTimeout(() => {
+            location.reload();
+        }, 2000)
     }
 
     private pickEditorFor(f: pkg.File): srceditor.Editor {
@@ -1774,9 +1778,9 @@ export class ProjectView
             return <div id="root" className="ui middle aligned center aligned grid" style={{ height: '100%', alignItems: 'center' }}>
                 <div className="ui raised segment inverted purple">
                     <h2>{lf("Oops") + "  😞😞😞"}</h2>
-                    {lf("This doesn't happen often, but looks like we messed up..")}
+                    {lf("This doesn't happen often, but looks like we've run into a problem..")}
                     <br /> <br />
-                    {lf("We're going to need you to refresh your page.")}
+                    {lf("Refreshing your page...")}
                 </div>
             </div>
         }
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -439,20 +439,23 @@ export class ProjectView
 
     // Add an error guard for the entire application
     componentDidCatch(error: any, info: any) {
-        core.killLoadingQueue();
-        pxsim.U.remove(document.getElementById('loading'));
-        this.setState({ hasError: true });
-        // Log critical error
-        pxt.tickEvent('pxt.criticalerror', { error, info });
-        // Reload the page in 2 seconds
-        const lastCriticalError = pxt.storage.getLocal("lastcriticalerror") ?
-            Date.parse(pxt.storage.getLocal("lastcriticalerror")) : Date.now();
-        // don't refresh if we refreshed in the last minute
-        if (Date.now() - lastCriticalError > 60 * 1000) {
-            pxt.storage.setLocal("lastcriticalerror", new Date().toISOString());
-            setTimeout(() => {
-                location.reload();
-            }, 2000)
+        try {
+            core.killLoadingQueue();
+            pxsim.U.remove(document.getElementById('loading'));
+            this.setState({ hasError: true });
+            // Log critical error
+            pxt.tickEvent('pxt.criticalerror', { error, info });
+            // Reload the page in 2 seconds
+            const lastCriticalError = pxt.storage.getLocal("lastcriticalerror") ?
+                Date.parse(pxt.storage.getLocal("lastcriticalerror")) : Date.now();
+            // don't refresh if we refreshed in the last minute
+            if (Date.now() - lastCriticalError > 60 * 1000) {
+                pxt.storage.setLocal("lastcriticalerror", new Date().toISOString());
+                setTimeout(() => {
+                    location.reload();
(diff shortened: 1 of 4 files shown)


pelikhan:
We need to reload the page. There is no reload button on iOS/Windows app.

_____________________________

samelhusseini:
Maybe still display the message, and reload after 1-2 seconds?

samelhusseini:
It would be helpedful for our testers / users to report an error by saying, I got that message.

samelhusseini:
Added a reload after 2 seconds. 

Also changed text to: 
"Oops :(:(:("
"This doesn't happen often, but looks like we've run into a problem.."
"Refreshing your page..."

webapp/src/app.tsx
```diff
@@ -437,6 +437,19 @@ export class ProjectView
         this.forceUpdate(); // we now have editors prepared
     }
 
+    // Add an error guard for the entire application
+    componentDidCatch(error: any, info: any) {
+        core.killLoadingQueue();
```
pelikhan: don't do anything if hasError is already set.

webapp/src/app.tsx
```diff
@@ -1760,6 +1773,17 @@ export class ProjectView
             'full-abs'
         ];
         const rootClasses = sui.cx(rootClassList);
+
+        if (this.state.hasError) {
+            return <div id="root" className="ui middle aligned center aligned grid" style={{ height: '100%', alignItems: 'center' }}>
+                <div className="ui raised segment inverted purple">
+                    <h2>{lf("Oops") + "  😞😞😞"}</h2>
+                    {lf("This doesn't happen often, but looks like we've run into a problem..")}
```
pelikhan: Let's keep the wording simpler. Something like "Sorry we detected a problem and we will reload the editor in a few seconds." Or something better but let's not make claims about frequency.

samelhusseini: Haha fair enough, I was trying to be cute

webapp/src/app.tsx
samelhusseini: I'm logging a criticalError event. I believe app insights already logs those anyway but it'd be good to know about the frequency of this event and what component is problematic. 

If react catches a number of exceptions we probably want to log all of them, what's that harm in doing this for all errors? (stop loading is harmless if done more than once, and same with set state) and it's going to all reload in 2 seconds anyway

riknoll:
I would argue that a continuously refreshing-site looks much more broken then an editor that does nothing. If we do the auto-reload I would feel more comfortable if we added logic to limit it to one reload.

samelhusseini:
Yeah I'm on the fence here. I agree one reload is the ideal, as most issues we can recover from one reload.

samelhusseini:
removing auto-reload for now. I'm more interested in the telemetry, and we can choose to do something about it in the future.

pelikhan:
We used this strategy in TouchDevelop very efficiently. You could drop a time stamp in local storage to detect reload loops.

_____________________________

riknoll:
+1 to dropping a timestamp

samelhusseini:
If you use a localstorage flag, say you set it and app reloads (avoiding a loop). When do you reset the flag? ie: say the app is now working right again? 
If you do it at a point where the app hasn't gotten to during it's startup, and then it crashes afterwards, you're still stuck in a reload loop. Let me know if you're not following

pelikhan:
If you don't do auto-refresh, you need to have a button that says "refresh" because there is no refresh button in some of our clients like the Windows app.

samelhusseini:
Yeah will do autorefresh and just set a timestamp.

webapp/src/simulator.ts
```diff
@@ -33,6 +33,7 @@ export function setTranslations(translations: pxt.Map<string>) {
 }
 
 export function init(root: HTMLElement, cfg: SimulatorConfig) {
+    if (!root) return;
```
pelikhan: any reason why we need this?

webapp/src/app.tsx
```diff
@@ -437,6 +437,25 @@ export class ProjectView
...
+    // Add an error guard for the entire application
+    componentDidCatch(error: any, info: any) {
+        core.killLoadingQueue();
+        pxsim.U.remove(document.getElementById('loading'));
+        this.setState({ hasError: true });
+        // Log critical error
+        pxt.tickEvent('pxt.criticalerror', { error, info });
+        // Reload the page in 2 seconds
+        const lastCriticalError = pxt.storage.getLocal("lastcriticalerror") ?
+            Date.parse(pxt.storage.getLocal("lastcriticalerror")) : Date.now();
```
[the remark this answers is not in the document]
samelhusseini: It doesn't but I can guard it anyway.

webapp/src/simulator.ts
samelhusseini: Yes, boardview might not exist. 
It was throwing in cases where checkSession throws (due to an invalid session).

webapp/src/app.tsx
```diff
@@ -1760,6 +1782,15 @@ export class ProjectView
             'full-abs'
         ];
         const rootClasses = sui.cx(rootClassList);
+
+        if (this.state.hasError) {
+            return <div id="root" className="ui middle aligned center aligned grid" style={{ height: '100%', alignItems: 'center' }}>
+                <div className="ui raised segment inverted purple">
+                    <h2>{lf("Oops")}</h2>
+                    {lf("We detected a problem and we will reload the editor in a few seconds..")}
```
riknoll: Show a different message if it isn't going to reload? No idea what that message should say.

samelhusseini: I think that’s over complicating the feature. Users really shouldn’t get into a reload loop, we just added a bit of logic to avoid that.

samelhusseini: I think that’s over complicating the feature. Users really shouldn’t get into a reload loop, we just added a bit of logic to avoid the possibility

PR #4216, opened 2018-04-26
Tweaks to pyconv

pelikhan:
- [ ] Don't mix static and const (generated code is still wrong but in a different way).
- [ ] fix debug output in CLI

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -7,8 +7,8 @@
 			"request": "launch",
 			"program": "${workspaceRoot}/built/pxt.js",
 			"stopOnEntry": false,
-			"args": ["serve", "--cloud"],
-			"cwd": "${workspaceRoot}/../pxt-adafruit",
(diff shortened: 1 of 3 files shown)


.vscode/launch.json
```diff
@@ -7,8 +7,8 @@
 			"request": "launch",
 			"program": "${workspaceRoot}/built/pxt.js",
 			"stopOnEntry": false,
-			"args": ["serve", "--cloud"],
-			"cwd": "${workspaceRoot}/../pxt-adafruit",
+			"args": ["pyconv", "adafruit_seesaw.py"],
+			"cwd": "${workspaceRoot}/../pxt-adafruit/projects/Adafruit_CircuitPython_seesaw",
```
mmoskal: Is this intended?

pelikhan: No but does not hurt really. It was bogus before.

PR #4223, opened 2018-04-30
Fix tutorial perf

samelhusseini:
We used to load the tutorial page in an iframe of the doc pages and then split it up based on some predefined format and just displayed it in the relevant parts of the editor. 

This meant that we had to re-load all editor resources in the iframe (which were already loaded in the editor and could be re-used).
Marked was also already included in the package so I decided to make use of that and just render the blocks on an as needed bases (ie: when the user clicks on the hint). 

Instead, here's what I'm doing: 
- User starts the tutorial engine
- We download the markdown from /docs. We parse it and store it in the state
- We then look for any block / snippet tags used and concat them into one snippet that we decompile in order to find the toolbox blocks used. 
- Update the toolbox with the blocks filtered
- We then render the first step inline (markdown react component). 
- When a user clicks on a hint, we also have an inline react component but this time we have the entire step's markdown in there which could include block tags, any block tags that we see, we call renderBlocks in the editor and get the block SVG. 
(In the meantime, we show a loading snipper as they're being decompiled)

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1668,7 +1668,7 @@ export class ProjectView
             .then(() => pxt.Cloud.downloadMarkdownAsync(tutorialId))
             .then(tutorialmd => {
                 const stepInfo = tutorial.parseTutorialSteps(tutorialId, tutorialmd);
-                return tutorial.getUsedBlocks(tutorialId, tutorialmd)
+                return tutorial.getUsedBlocksAsync(tutorialId, tutorialmd)
                     .then((usedBlocks) => {
                         this.setState({
                             editorState: {
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1670,14 +1670,17 @@ export class ProjectView
                 const stepInfo = tutorial.parseTutorialSteps(tutorialId, tutorialmd);
                 return tutorial.getUsedBlocksAsync(tutorialId, tutorialmd)
                     .then((usedBlocks) => {
+                        let editorState: pxt.editor.EditorState = {
+                            searchBar: false
+                        }
+                        if (usedBlocks && Object.keys(usedBlocks).length > 0) {
+                            editorState.filters = {
+                                blocks: usedBlocks,
+                                defaultState: pxt.editor.FilterState.Hidden
+                            }
+                        }
                         this.setState({
-                            editorState: {
-                                searchBar: false,
-                                filters: {
-                                    blocks: usedBlocks,
-                                    defaultState: pxt.editor.FilterState.Hidden
-                                }
-                            },
+                            editorState: editorState,
                             tutorialOptions: {
                                 tutorial: tutorialId,
                                 tutorialName: title,
@@ -1690,6 +1693,13 @@ export class ProjectView
                         const fullscreen = stepInfo[0].fullscreen;
                         if (fullscreen) this.showTutorialHint();
                         else this.showLightbox();
+                    })
+                    .catch(e => {
+                        // Failed to decompile
+                        pxt.tickEvent('tutorial.faileddecompile', {tutorialId: tutorialId});
+                        core.errorNotification(lf("Oops, an error occured as we were loading the tutorial."));
+                        // Reset state (delete the current project and exit the tutorial)
+                        this.exitTutorial(true);
                     })
             }).catch((e) => {
                 core.handleNetworkError(e);
@@ -1710,18 +1720,19 @@ export class ProjectView
             })
     }
 
-    exitTutorial() {
+    exitTutorial(removeProject?: boolean) {
         pxt.tickEvent("tutorial.exit");
         core.showLoading("leavingtutorial", lf("leaving tutorial..."));
-        this.exitTutorialAsync()
+        this.exitTutorialAsync(removeProject)
             .done(() => {
                 core.hideLoading("leavingtutorial");
                 this.openHome();
             })
     }
 
-    exitTutorialAsync() {
+    exitTutorialAsync(removeProject?: boolean) {
         let curr = pkg.mainEditorPkg().header;
+        curr.isDeleted = removeProject;
         let files = pkg.mainEditorPkg().getAllFiles();
         return workspace.saveAsync(curr, files)
             .then(() => Promise.delay(500))
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -623,10 +623,7 @@ export class ProjectView
             this.setState({ tutorialOptions: tutorialOptions });
             const fullscreen = tutorialOptions.tutorialStepInfo[step].fullscreen;
             if (fullscreen) this.showTutorialHint();
-            else {
-                tutorial.TutorialContent.refresh();
-                this.showLightbox();
-            }
+            else this.showLightbox();
         }
     }
 
@@ -641,9 +638,12 @@ export class ProjectView
                     case 'loaded':
                         let tt = msg as pxsim.TutorialLoadedMessage;
                         if (tt.toolboxSubset && Object.keys(tt.toolboxSubset).length > 0) {
-                            this.setState({editorState: {
-                                searchBar: false,
-                                filters: { blocks: tt.toolboxSubset, defaultState: pxt.editor.FilterState.Hidden } } });
+                            this.setState({
+                                editorState: {
+                                    searchBar: false,
+                                    filters: { blocks: tt.toolboxSubset, defaultState: pxt.editor.FilterState.Hidden }
+                                }
+                            });
                             this.editor.filterToolbox(tt.toolboxSubset, tt.showCategories);
                         }
                         let tutorialOptions = this.state.tutorialOptions;
@@ -653,7 +653,6 @@ export class ProjectView
                         const fullscreen = tutorialOptions.tutorialStepInfo[0].fullscreen;
                         if (fullscreen) this.showTutorialHint();
                         else {
-                            tutorial.TutorialContent.refresh();
                             this.showLightbox();
                         }
                         core.hideLoading("tutorial");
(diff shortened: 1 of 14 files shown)


samelhusseini:
Also in order to improve performance, we have a cache of SVGs used in the tutorial that we maintain in the lifetime of the tutorial. If a user clicks back on a category that we've already rendered the SVG for, it will load from cache rather than re-decompiling the snippet.

webapp/src/app.tsx
```diff
@@ -1652,22 +1651,50 @@ export class ProjectView
...
+                        tutorial: tutorialId,
+                        tutorialName: title
+                    },
+                    tracing: undefined
+                });
+            })
+            .then(() => pxt.Cloud.downloadMarkdownAsync(tutorialId))
+            .then(tutorialmd => {
+                const stepInfo = tutorial.parseTutorialSteps(tutorialId, tutorialmd);
+                return tutorial.getUsedBlocks(tutorialId, tutorialmd)
```
pelikhan: "Async"

webapp/src/marked.tsx
```diff
@@ -0,0 +1,142 @@
...
+            params["drivename"] = this.html2Quote(theme.driveDisplayName);
+        if (theme.homeUrl)
+            params["homeurl"] = this.html2Quote(theme.homeUrl);
+        params["targetid"] = theme.id || "???";
+        params["targetname"] = theme.name || "Microsoft MakeCode";
+        params["targetlogo"] = theme.docsLogo ? `<img aria-hidden="true" role="presentation" class="ui mini image" src="${theme.docsLogo}" />` : "";
+        return params;
+    }
+
+    private renderSnippetsAsync(content: HTMLElement) {
```
pelikhan: Async function should return a promise

webapp/src/marked.tsx
```diff
@@ -0,0 +1,142 @@
...
+        pxt.Util.toArray(document.querySelectorAll(`.lang-blocks`))
+            .forEach((langBlock: HTMLElement) => {
+                const code = langBlock.innerText;
+                langBlock.innerHTML = `<div class="ui segment raised loading"></div>`;
+                if (MarkedContent.blockSnippetCache[code]) {
+                    // Use cache
+                    langBlock.innerHTML = `<div class="ui segment raised">
+                                <img src="${MarkedContent.blockSnippetCache[code]}" /></div>`;
+                } else {
+                    parent.renderBlocksAsync({
```
pelikhan: leaking promise here

webapp/src/marked.tsx
```diff
@@ -0,0 +1,142 @@
...
+        pxt.Util.toArray(document.querySelectorAll(`:not(pre) > code`))
+            .forEach((inlineBlock: HTMLElement) => {
+                const text = inlineBlock.innerText;
+                const mbtn = /^(\|+)([^\|]+)\|+$/.exec(text);
+                if (mbtn) {
+                    const mtxt = /^(([^\:\.]*?)[\:\.])?(.*)$/.exec(mbtn[2]);
+                    const ns = mtxt[2] ? mtxt[2].trim().toLowerCase() : '';
+                    const txt = mtxt[3].trim();
+                    const lev = mbtn[1].length == 1 ? `docs inlinebutton ui button ${txt.toLowerCase()}-button` : `docs inlineblock ${ns}`;
+                    inlineBlock.innerHTML = `<span class="${lev}">${pxt.U.rlf(txt)}</span>`
```
pelikhan: escape txt

webapp/src/marked.tsx
```diff
@@ -0,0 +1,142 @@
...
+                if (MarkedContent.blockSnippetCache[code]) {
+                    // Use cache
+                    langBlock.innerHTML = `<div class="ui segment raised">
+                                <img src="${MarkedContent.blockSnippetCache[code]}" /></div>`;
+                } else {
+                    parent.renderBlocksAsync({
+                        action: "renderblocks", ts: code
+                    } as pxt.editor.EditorMessageRenderBlocksRequest)
+                        .then((xml) => {
+                            langBlock.innerHTML = `<div class="ui segment raised">
```
pelikhan: escape xml

webapp/src/tutorial.tsx
```diff
@@ -6,9 +6,85 @@ import * as data from "./data";
...
+        stepInfo[i].hasHint = contentLines.length > 1;
+    }
+    return stepInfo;
+}
+
+/**
+ * We'll run this step when we first start the tutorial to figure out what blocks are used so we can
+ * filter the toolbox. 
+ */
+export function getUsedBlocks(tutorialId: string, tutorialmd: string): Promise<{ [index: string]: number }> {
```
pelikhan: async

pelikhan: use ``Promise<Map<number>>`` as return type

webapp/src/marked.tsx
samelhusseini: done.

webapp/src/marked.tsx
```diff
@@ -104,7 +110,7 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
                     const ns = mtxt[2] ? mtxt[2].trim().toLowerCase() : '';
                     const txt = mtxt[3].trim();
                     const lev = mbtn[1].length == 1 ? `docs inlinebutton ui button ${txt.toLowerCase()}-button` : `docs inlineblock ${ns}`;
-                    inlineBlock.innerHTML = `<span class="${lev}">${pxt.U.rlf(txt)}</span>`
+                    inlineBlock.innerHTML = `<span class="${lev}">${this.html2Quote(pxt.U.rlf(txt))}</span>`
```
pelikhan: I would expect to use "htmlEscape".

samelhusseini:
Build for testing: https://makecode.adafruit.com/app/56339a6ed5e818de3da7e566c1e4876ad3a3139b-35f1f12714

webapp/src/marked.tsx
```diff
@@ -0,0 +1,148 @@
...
+                                langBlock.innerHTML = `<div class="ui segment raised">
+                                    <span>${lf("Oops, something went wrong trying to render this block snippet.")}</span></div>`;
+                            }
+                        })
+                }
+            })
+    }
+
+    private renderInlineBlocks(content: HTMLElement) {
+        pxt.Util.toArray(document.querySelectorAll(`:not(pre) > code`))
```
pelikhan: query content, not document

webapp/src/marked.tsx
```diff
@@ -0,0 +1,148 @@
...
+        params["targetid"] = theme.id || "???";
+        params["targetname"] = theme.name || "Microsoft MakeCode";
+        params["targetlogo"] = theme.docsLogo ? `<img aria-hidden="true" role="presentation" class="ui mini image" src="${theme.docsLogo}" />` : "";
+        return params;
+    }
+
+    private renderSnippets(content: HTMLElement) {
+        const { parent } = this.props;
+
+        pxt.Util.toArray(document.querySelectorAll(`.lang-blocks`))
```
pelikhan: query content, not document

samelhusseini: Good catch

webapp/src/marked.tsx
```diff
@@ -0,0 +1,148 @@
...
+        s = this.replaceAll(s, "&", "&amp;")
+        s = this.replaceAll(s, "<", "&lt;")
+        s = this.replaceAll(s, ">", "&gt;")
+        s = this.replaceAll(s, "\"", "&quot;")
+        s = this.replaceAll(s, "\'", "&#39;")
+        return s;
+    }
+
+    // the input already should be HTML-quoted but we want to make sure, and also quote quotes
+    private html2Quote(s: string) {
```
pelikhan: do we still need this help? Why not just htmlEscape?

samelhusseini: I was just following what was done in the docs, but perhaps not.

samelhusseini:
PR ready.

webapp/src/marked.tsx
```diff
@@ -0,0 +1,117 @@
...
+        const { parent } = this.props;
+
+        pxt.Util.toArray(content.querySelectorAll(`.lang-blocks`))
+            .forEach((langBlock: HTMLElement) => {
+                const code = langBlock.innerText;
+                langBlock.innerHTML = `<div class="ui segment raised loading"></div>`;
+                if (MarkedContent.blockSnippetCache[code]) {
+                    // Use cache
+                    langBlock.innerHTML = `<div class="ui segment raised">
+                                <img src="${MarkedContent.blockSnippetCache[code]}" /></div>`;
```
pelikhan: cache contains unescaped XML

webapp/src/marked.tsx
```diff
@@ -0,0 +1,117 @@
...
+                } else {
+                    parent.renderBlocksAsync({
+                        action: "renderblocks", ts: code
+                    } as pxt.editor.EditorMessageRenderBlocksRequest)
+                        .done((xml) => {
+                            if (xml) {
+                                langBlock.innerHTML = `<div class="ui segment raised">
+                                    <img src="${pxt.Util.htmlEscape(xml)}" /></div>`;
+                                MarkedContent.blockSnippetCache[code] = xml;
```
pelikhan: escape xml before caching. It would be safer to contruct DOM nodes that rely on innerHTML.

pelikhan:
Adding a second pair of eyes.

webapp/src/marked.tsx
```diff
@@ -58,7 +58,7 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
                             if (xml) {
                                 langBlock.innerHTML = `<div class="ui segment raised">
                                     <img src="${pxt.Util.htmlEscape(xml)}" /></div>`;
```
pelikhan: don't escape twice xml

samelhusseini:
Just using default options. santizie (HTML) is true by default: http://assemble.io/docs/options-marked.html

pelikhan:
We should be explicit about setting sanitize to true regardless of defaults.

samelhusseini:
Done.

webapp/src/marked.tsx
```diff
@@ -0,0 +1,138 @@
...
+                        .done((datauri) => {
+                            if (datauri) {
+                                MarkedContent.blockSnippetCache[code] = datauri;
+                                const imgDiv = document.createElement('img');
+                                imgDiv.src = MarkedContent.blockSnippetCache[code];
+                                wrapperDiv.appendChild(imgDiv);
+                                pxsim.U.removeClass(wrapperDiv, 'loading');
+                            } else {
+                                // An error occured, show alternate message
+                                const textDiv = document.createElement('span');
```
pelikhan: Show original content?

samelhusseini: I guess, just show Javascript.

samelhusseini: that way we know there's an issue, but the user can carry on perhaps?

samelhusseini: unlikely.

PR #4224, opened 2018-04-30
check status before reloading appcache

pelikhan:
Editor gets stuck in infinite reload loop as the appcache updateready event triggers but the status is idle (1)


pelikhan:
@riknoll PR feedback added.

riknoll:
LGTM

Issue #2807, opened 2017-08-24
[Keyboard Navigation-Project -Makecode Documentation] Focus is not naviagating to first interactive element after selecting the any link in the left pane.


v-mesala:
**User Experience:**
User who depends on Keyboard with screen reader may get misguided if the Focus is not navigating to first interactive element after selecting the any link in the left pane..

**Test Environment:**
OS: RS2 Version 1703(OS Build 15063.2)
Platform: Edge.

**Repro Steps:-**
1. Navigate to https://makecode.microbit.org/projects
2. Select any link in the left pane and verify the issue.

**Actual Result:-**
Focus is not navigating to first interactive element after selecting the any link in the left pane.

**Expected Result:-**
Focus should navigate to first interactive element after selecting the any link in the left pane.
If we select Flashing heart link then focus should navigate to first interactive element like Flashing Heart  link after Projects.

**Note:** 
This issue exists with Firefox also.

**MAS Reference -**
https://microsoft.sharepoint.com/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={0de7fbe1-ad7e-48e5-bcbb-8d986691e2b9}

**Suggested Fix:**
In all the applicable forms we need to check below-  
1.Check if 'tabindex' is used 
2.If 'tabindex' is used, check that the tab order specified by the 'tabindex' attributes follows relationships in the content. 
 
tabindex property is applicable to all visual elements on screen.  
<a href="/articles/" tabindex="2">Articles</a> 
<a href="/resources/" tabindex="3">Resources</a> 
 
Tips : 
'tabindex' should only be used in cases where: 
•The default tab order is not ideal, AND 
•The tab order cannot be changed by rearranging items in the content and/or by altering the style sheet to reflect the best visual arrangement.
Additional Note: Consider whether content may be changed, added, or removed frequently on the web page. If so and if content order changes from time to time, then using tabindex="0" may also be an option in order to minimize maintenance. Content that has tabindex="0" set on it will cause the tabbing order to the follow the order of the of the content in the source code.
The links in the snippet below all use the tabindex="0" attribute, which provides a logical tabbing order that matches the visual order and source code order or the links.

**Refer Attachment:**
[MAS 2.4.3_MakeCode_Focus is not navigating to first interactive element.zip](https://github.com/Microsoft/pxt/files/1248891/MAS.2.4.3_MakeCode_Focus.is.not.navigating.to.first.interactive.element.zip)

veler:
[...]

PR #4229, opened 2018-05-01
Port over accessibility menu from master

samelhusseini:
Note: this is a minor UI change for accessible users.

The semantic.css shared in pxt-deployment-config is used by docs.html in both v0 and master. 
A recent change to update that drop with a new semantic.css snap from master broke the accessibility menu in v0. 

Attempt to fix https://github.com/Microsoft/pxt/issues/2807

Porting over the new accessibility menu from master, so they can share the same structure and semantic.css

Here's what the old one looked like: 
screen shot 2018-05-01 at 10 56 56 am

Here's what the new one we're using in master looks like: 
screen shot 2018-05-01 at 10 56 22 am

and in docs: 
screen shot 2018-05-01 at 10 56 36 am

The styling and color used matches the MDN accessibility page.

### DIFF
--- a/docfiles/docs.html
+++ b/docfiles/docs.html
@@ -13,7 +13,7 @@
 
 <body id='root' class='root'>
 
-<div id="accessibleMenu" role="menubar">
+<div id="accessibleMenu" class="ui accessibleMenu borderless fixed menu"  role="menubar">
     @accMenu@
 </div>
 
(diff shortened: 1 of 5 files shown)


samelhusseini:
Change will go live for the docs page. makecode.microbit.org/projects as soon as this PR is merged. 
Also note, the accessibility menu is currently broken in the docs pages.

PR #4234, opened 2018-05-02
Don't throw away comments on auto declarations in decompiler

riknoll:
For some reason, the baseline in the test case was generated without the actual comment content. I expect this has something to do with the decompiler test runner so don't worry about it. I've tested the same code in the browser and it works fine.


riknoll:
Okay, I am very confused as to what is going on with that test case. Guess I'll need to debug it locally.

riknoll:
Windows line endings were the culprit of course

PR #4236, opened 2018-05-02
Update package.json

samelhusseini:
Ensuring dependencies are in the right spot.
Making less a shipped dependency so targets dont have to rely on it.


samelhusseini:
Nah, let's ignore that for now.

PR #4243, opened 2018-05-03
Vertical separators in block authoring

samelhusseini:
With this change we can add vertical separators in our block authoring. 
This allows us to do things like this when referring to objects (OO model): 

screen shot 2018-05-02 at 7 11 45 pm

@Jaqster @abchatra @riknoll @pelikhan @guillaumejenkins this is what I was talking about today. 

Before:
screen shot 2018-05-02 at 7 22 25 pm

After:
screen shot 2018-05-02 at 7 11 45 pm

We're running out of characters to use for our authoring. Since Pipe was already used, I'm using the Hash character (#, Pound key for the Americans) to indicate we want a separator. Open to other suggestions.

### DIFF
--- a/localtypings/blockly.d.ts
+++ b/localtypings/blockly.d.ts
@@ -574,6 +574,10 @@ declare namespace Blockly {
         forceRerender(): void;
     }
 
+    class FieldVerticalSeparator extends Field {
+        constructor();
+    }
+
     class FieldVariable extends Field {
         constructor(d: any);
     }
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -414,9 +414,11 @@ namespace pxt.blocks {
         return true;
     }
 
-    function newLabel(part: pxtc.BlockLabel | pxtc.BlockImage): Blockly.Field {
+    function newLabel(part: pxtc.BlockLabel | pxtc.BlockImage | pxtc.BlockSeparator): Blockly.Field {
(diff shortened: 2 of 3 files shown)


pelikhan:
I don't think this helps. If you want to get closer to OO and JS, you might as well show a '.'.

pelikhan:
(Scratch approach to OO is much better but not quite applicable in MakeCode)

samelhusseini:
@pelikhan any thoughts on the character to use? I'd like this option in authoring to experiment with anyway.

Jaqster:
Agree, I would rather see a '.' for this - maybe we can make it bigger.  I do really like this idea of bringing OOP paradigm into blocks, makes the transition to JavaScript more fluid.

samelhusseini:
I think for a kid, a '.' is just going to be interpreted as an end of sentence, and not as dot notation. Anyway perhaps the discussion for using the vertical separator should be done in person. 

At any rate, I'd like support for Field separators (regardless of what we use them for).

pelikhan:
Let’s discuss this in a design meeting.

samelhusseini:
Agreed lets discuss at the design meeting. 
Once again, I want this in for experimentation. It's just a field in the authoring so doesn't affect anyone that doesn't use it. Any thoughts on what character to use?

pelikhan:
░

samelhusseini:
Closing for now till we decide on the block authoring.

PR #4244, opened 2018-05-03
show warning on orphan exceptions

pelikhan:
Show a warning when an orphan exception (not source available) is found.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -415,9 +415,7 @@ export class ProjectView
             orphanException: brk => {
                 // TODO: start debugging session
                 // TODO: user friendly error message
-                core.warningNotification(lf("Program Error: {0}", brk.exceptionMessage));                const mainPkg = pkg.mainEditorPkg();
-                const f = mainPkg.outputPkg.setFile("output.txt", brk.exceptionMessage)
-                f.numDiagnosticsOverride = 1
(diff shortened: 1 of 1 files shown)


webapp/src/app.tsx
```diff
@@ -413,7 +413,8 @@ export class ProjectView
         this.allEditors.forEach(e => e.prepare())
         simulator.init(document.getElementById("boardview"), {
             orphanException: brk => {
-                // do something!
+                core.warningNotification(brk.exceptionMessage);
```
samelhusseini: Do you want to add a prefix like "Program Error: " ?

webapp/src/app.tsx
```diff
@@ -413,7 +413,11 @@ export class ProjectView
         this.allEditors.forEach(e => e.prepare())
         simulator.init(document.getElementById("boardview"), {
             orphanException: brk => {
-                // do something!
+                // TODO: start debugging session
+                // TODO: user friendly error message
+                core.warningNotification(lf("Program Error: {0}", brk.exceptionMessage));                const mainPkg = pkg.mainEditorPkg();
+                const f = mainPkg.outputPkg.setFile("output.txt", brk.exceptionMessage)
```
samelhusseini: nit: format?

samelhusseini:
Also I wouldn't start debugging unless debugging is enabled.

PR #4245, opened 2018-05-03
Keep track of incomplete typechecks

samelhusseini:
When a change happens, the type checker first checks if the editor is in an incomplete state (ie: the user is editing, or dragging a block, etc, no point completing the typecheck if it's going to be invalidated anyway). 
One such reason for an incomplete state is a field editor is open, but in the chance that the user selects the same option as the previously selected option, you end up in a state where the typecheck is in an incomplete state, but no change event is triggered by Blockly (since nothing changed in the code). 

This change marks when the typechecker has a queued check that is incomplete, and triggers the change handler on any event by blockly (including UI events), so when a user selects anywhere in the workspace that would trigger the last incomplete typecheck to complete. 

Alternatively, we can just not mark typechecks as incomplete when field editors are open.. thoughts?

### DIFF
--- a/webapp/src/blocks.tsx
+++ b/webapp/src/blocks.tsx
@@ -307,7 +307,7 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         const editor = this;
 
         const oldAudioPlay = (Blockly as any).WorkspaceAudio.prototype.play;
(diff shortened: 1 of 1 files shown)


samelhusseini:
Going to go ahead with this change. Let me know if you have any concerns @pelikhan

PR #4246, opened 2018-05-03
Insert debugger breakpoint in the center of the screen

pelikhan:


webapp/src/blocks.tsx
```diff
@@ -551,6 +551,9 @@ export class Editor extends toolboxeditor.ToolboxEditor {
     insertBreakpoint() {
         if (!this.editor) return;
         const b = this.editor.newBlock(pxtc.TS_DEBUGGER_TYPE);
+        // move block roughly to the center of the screen
+        const m = this.editor.getMetrics();
+        b.moveBy(m.viewWidth / 2, m.viewHeight / 3);
```
samelhusseini: You can use b.getHeightWidth() to get the size of the block and incorporate that into the calculation if you like

pelikhan:
I'd like to have some kind of pop up animation and close to the menu but that's trickier.

samelhusseini:
Menu as in the flyout? 

Blockly.hideChaff() to close all flyouts, field editors, etc.

pelikhan:
The debugger menu...

samelhusseini:
Ah okay

PR #4254, opened 2018-05-04
Add electron base data to telemetry

guillaumejenkins:


docfiles/tracking.html
```diff
@@ -22,6 +22,12 @@
                 telemetryItem.properties["cookie"] = includeCookie;
                 if (typeof Windows !== "undefined")
                     telemetryItem.properties["WindowsApp"] = 1;
+                if (typeof pxtElectron !== "undefined")
+                    telemetryItem.properties["PxtElectron"] = 1;
```
pelikhan: could you extract a version?

PR #4258, opened 2018-05-05
Fix tutorials in IE

samelhusseini:
Use blockly XML in the tutorial instead of images. Use innerHTML instead of InnerText so that IE doesn't escape new line characters in <code> blocks

### DIFF
--- a/pxteditor/editor.ts
+++ b/pxteditor/editor.ts
@@ -199,7 +199,7 @@ namespace pxt.editor {
         loadBlocklyAsync(): Promise<void>;
         isBlocksEditor(): boolean;
         isTextEditor(): boolean;
-        renderBlocksAsync(req: EditorMessageRenderBlocksRequest): Promise<string>;
+        renderBlocksAsync(req: EditorMessageRenderBlocksRequest): Promise<any>;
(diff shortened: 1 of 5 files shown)


webapp/src/app.tsx
```diff
@@ -1538,15 +1538,18 @@ export class ProjectView
         this.importDialog.show();
     }
 
-    renderBlocksAsync(req: pxt.editor.EditorMessageRenderBlocksRequest): Promise<string> {
```
pelikhan: can we specify the type in the signature?

pxteditor/editor.ts
```diff
@@ -199,7 +199,7 @@ namespace pxt.editor {
         loadBlocklyAsync(): Promise<void>;
         isBlocksEditor(): boolean;
         isTextEditor(): boolean;
-        renderBlocksAsync(req: EditorMessageRenderBlocksRequest): Promise<string>;
+        renderBlocksAsync(req: EditorMessageRenderBlocksRequest): Promise<any>;
```
pelikhan: can we try to remove any with a proper type here?

PR #4261, opened 2018-05-07
Remove pxt-require

guillaumejenkins:


abchatra:
Can you add details, why we are doing this?

PR #4275, opened 2018-05-09
Image editor performance improvements

riknoll:
No more SVG grids! Those worked well for smaller images but did not scale; this new implementation uses layered canvases and is much much faster with larger sprites. The toolbox flyout should also be considerably faster because the image previews on the block are now PNGs rather than SVGs.

As discussed, I also added a "light mode" for the editor that removes two of the canvases, turns off transparency (in the rendering, not the sprites), and gets rid of the resizing animations. Also fixed a few Edge/IE bugs.

Still to do:
- [ ] Remove old SVG grid code
- [ ] copy/paste support
- [ ] Add an icon for the "fill" tool (and a matching cursor)
- [ ] Better canvas resizing experience


samelhusseini:
There's quite a few DOM utils that would be useful to have in a common utils library, perhaps a pxtutils that's included in all bundles (pxtsim, pxtapp and pxtworker).

pxtblocks/blocklycustomeditor.ts
```diff
@@ -56,6 +56,15 @@ namespace pxt.blocks {
             console.error(`Field editor ${selector} not registered`);
             return null;
         }
+
+        if (!params) {
+            params = {};
+        }
+
+        Util.assert(params.lightMode == undefined, "lightMode is a reserved parameter for custom fields");
```
samelhusseini: Good call, we should add asserts for other reserved parameters.

pxtblocks/fields/sprite/bitmapImage.ts
```diff
@@ -112,7 +112,7 @@ namespace pxtblockly {
                 .at(width / 2, height / 2)
                 .fontSize(30, svg.LengthUnit.px)
                 .fill("white")
-                .alignmentBaseline("middle")
+                // .alignmentBaseline("middle")
```
samelhusseini: not needed?

pelikhan:
@riknoll do you want to fix the remaining work items in this PR? Or should we merge now?

riknoll:
I'll do the other stuff as a separate PR

PR #4285, opened 2018-05-11
Add Microsoft's recommended tslint rules

samelhusseini:
Add tslint rules from tslint-microsoft-contrib and apply all the relevant changes.
All the rules and their descriptions are detailed here: https://github.com/Microsoft/tslint-microsoft-contrib

The idea here is to add a set of Security, Accessibility and common bug problem tslint rules so that we catch / avoid certain types of bugs. Linting will occur on every check in, so you won't be able to check in code that doesn't abide by these rules. I'm going to highlight the most common patterns I found, and the bugs that this change has already caught. 

Obvious bugs already caught: 
https://github.com/Microsoft/pxt/blob/master/webapp/src/hidbridge.ts#L88
https://github.com/Microsoft/pxt/blob/master/pxtsim/visuals/breadboard.ts#L395
https://github.com/Microsoft/pxt/blob/master/webapp/src/debugger.tsx#L68
https://github.com/Microsoft/pxt/blob/master/webapp/src/share.tsx#L137
https://github.com/Microsoft/pxt/blob/master/pxtblocks/blocklymutators.ts#L373

Plus quite a few React issues and accessibility issues.

Here's some general guidance: 
- Don't use innerHTML
- No for in for arrays, use array.forEach instead
- No dangerouslySetInnerHTML (React), same as above
- Always add rel="noopener noreferrer" any a tag with target="_blank" (as a precaution against reverse tabnabbing)
- Always bind this functions in react component constructor

**Security:**
Most of the work here was around removing the use of innerHTML as this is a security risk. There is still work to be done here around the use of jquery html() in the docs, so it's not 100% complete but the entire docs rendering relies on that so it's a significant amount of work to change it.

We had two banned keywords in our code: caller and arguments. I removed caller as it wasn't needed and renamed arguments to args.

I removed no-reserved-keywords from the list that is recommended by Microsoft's SDL because we use quite a few of them since we have an embedded compiler such as type, symbol, package, modules etc.

**React:**
The most common problem in the way we'd been using react has been that we bind functions in the render function. This means that we are creating a new function on every render and mess up any optimizations that react would do to compare states. 
To read more about this issue see the 'react-this-binding-issue' at https://github.com/Microsoft/tslint-microsoft-contrib

Mostly what this means is that all functions are bound in the constructor, and called directly as this.foo when passed in the render function to a different component. This includes the handleRef function. 
No functions are created in the render function.

** Common bugs and correctness:**
I found a number of bugs by disabling assignment in conditionals.
Removing for in array just means we won't mess up in including a for in without a conditional to filter our the prototype. 

There are still some TODOS here around enabling: 
- promise-must-complete
- no-use-before-declare
- no-object-literal-type-assertion
- no-constant-condition

**Accessibility:**
The tslint rules here have been fairly positive so far and have found a number of accessibility bugs around the file explorer and some other miscellaneous items where we hadn't been setting the relevant aria attributes for specific for certain roles.


cli/buildengine.ts
```diff
@@ -411,7 +411,8 @@ export function buildDalConst(buildEngine: BuildEngine, mainPkg: pxt.MainPackage
                 enumVal = -1;
             }
 
-            if (inEnum && (m = /^\s*(\w+)\s*(=\s*(.*?))?,?\s*$/.exec(ln))) {
+            m = /^\s*(\w+)\s*(=\s*(.*?))?,?\s*$/.exec(ln);
+            if (inEnum && m) {
```
pelikhan: why?

cli/commandparser.ts
```diff
@@ -34,7 +36,7 @@ export interface Command {
 
 export interface ParsedCommand {
     name: string;
-    arguments: string[];
+    args: string[];
```
pelikhan: why this renaming?

pxtblocks/blocklyloader.ts
```diff
@@ -390,9 +390,12 @@ namespace pxt.blocks {
         }
 
         let hash = JSON.stringify(fn);
+        /* tslint:disable:possible-timing-attack */
+        // TODO(tslint): look into this possible timing attack
```
pelikhan: this is not a cryptographic related codepath, there is no point on doing an attack here.

pxtblocks/blocklymutators.ts
```diff
@@ -370,7 +370,7 @@ namespace pxt.blocks {
                 properties.forEach(prop => {
                     this.parameters.push(prop.property);
                     if (prop.newName && prop.newName !== prop.property) {
-                        this.parameterRenames[prop.property] === prop.newName;
+                        this.parameterRenames[prop.property] = prop.newName;
```
pelikhan: nice bug, does this break any existing block?

pxtlib/package.ts
```diff
@@ -380,8 +380,10 @@ namespace pxt {
             // no core package? add the first one
             if (corePackages.length == 0) {
                 const allCorePkgs = pxt.Package.corePackages();
+                /* tslint:disable:no-unused-expression TODO(tslint): */
                 if (allCorePkgs.length)
                     this.config.dependencies[allCorePkgs[0].name];
```
pelikhan: nice another bug...

pxtlib/service.ts
```diff
@@ -858,6 +858,7 @@ namespace ts.pxtc {
                 currentLabel = "";
             }
 
+            /* tslint:disable:possible-timing-attack TODO(tslint): look into this security issue */
```
pelikhan: this is not security critical codepath

pxtsim/runtime.ts
```diff
@@ -34,6 +34,10 @@ namespace pxsim {
             while (element.firstChild) element.removeChild(element.firstChild);
         }
 
+        export function clear(element: Element) {
+            this.removeChildren(element);
```
pelikhan: what is "this" in this context?

theme/themes/pxt/views/card.overrides
```diff
@@ -110,7 +110,7 @@
     border-top-right-radius: @borderRadius;
 }
 .ui.cardimage {
-    background-color: @secondaryColor;
+    background-color: @grey;
```
pelikhan: ? unrelated changes?

cli/buildengine.ts
samelhusseini: no conditional assignment

cli/commandparser.ts
samelhusseini: arguments is a banned word.

pxtblocks/blocklyloader.ts
samelhusseini: Okay, will remove TODO.

pxtblocks/blocklymutators.ts
samelhusseini: old chat command blocks used this. No current blocks use this.

pxtlib/package.ts
samelhusseini: wait what's the bug here?

pxtlib/service.ts
samelhusseini: Will remove TODO.

theme/themes/pxt/views/card.overrides
samelhusseini: no, related to Filexplorer changes. (Accessibility)

samelhusseini:
We can include it in tomorrow's release. There's no a huge amount of functional changes here besides all the react stuff, and that's just simple function referencing changes. I think it's easily something that can be caught in a release test pass.

pelikhan:
It did lazy execute the regex.

pxtlib/package.ts
pelikhan: this line does nothing but the comment seems to say otherwise.

cli/buildengine.ts
samelhusseini: Right.. Well no conditional assignment means you need to take it all out. Let me do that.

pelikhan: it did lazy execute the regex so let's bring it back.

cli/commandparser.ts
pelikhan: i don't understand why.

samelhusseini: Do not use banned terms: caller, callee, eval, arguments. These terms refer to functions or properties that should not be used, so it is best practice to simply avoid them.

PR #4286, opened 2018-05-11
Don't include marked.min.js

samelhusseini:
Marked is already required in the main app. No need to add it include it

### DIFF
--- a/webapp/public/embed.js
+++ b/webapp/public/embed.js
@@ -8,7 +8,6 @@
         "/blb/highlight.js/highlight.pack.js",
         "/blb/bluebird.min.js",
         "/blb/semantic.js",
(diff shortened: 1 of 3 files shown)


pelikhan:
Is it bundled in pxtembed.js?

samelhusseini:
Probably not, what's it used for in pxtembed?

pelikhan:
Embed.js is used in docs right? We need to trace all the place where markdown is used.

_____________________________

samelhusseini:
Okay hold off on this PR.

pelikhan:
You could refactor the requireMarked to delay load marked (and still work in node.js). Stil this also used in the backend.

samelhusseini:
Closing for now..

PR #4288, opened 2018-05-11
Use ledmatrix field editor

samelhusseini:
Move ledmatrix from grove-zero into pxt-core. Updated it with better touch support.
Converted imageLiterals to use the field editor instead of the checkbox grid hack. 

dragtopaint

@jaustin thoughts on the new editor?


pxtsim/runtime.ts
```diff
@@ -306,8 +306,8 @@ namespace pxsim {
             return id;
         }
 
-        unregisterLiveObject(object: RefObject) {
-            U.assert(object.refcnt == 0, "ref count is not 0");
+        unregisterLiveObject(object: RefObject, keepAlive?: boolean) {
+            if (!keepAlive) U.assert(object.refcnt == 0, "ref count is not 0");
```
pelikhan: ?

pelikhan:
Would be great to support other image sizes... eventually.

pxtsim/runtime.ts
samelhusseini: Equivalent of noLeakTracking in v0. Sorry it's slightly unrelated change.
https://github.com/Microsoft/pxt/blob/v0/pxtsim/langsupport.ts#L38

PR #4291, opened 2018-05-11
Thou shalt not load a script from an editor 1 major version above

pelikhan:
Following the semver semantic, if the editor detects a script with a major version greater than current, refuse to load.
This PR will go in v0 as well.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -712,7 +712,9 @@ export class ProjectView
                 body: lf("This project was created in a newer version of this editor. Please try again in that editor."),
                 disagreeLbl: lf("Ok"),
                 buttons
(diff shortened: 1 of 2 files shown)


webapp/src/app.tsx
```diff
@@ -695,6 +695,28 @@ export class ProjectView
 
         this.stopSimulator(true);
         this.clearSerial()
+
+        // version check, you should not load a script from 1 major version above.
+        if (h.targetVersion && pxt.semver.majorCmp(h.targetVersion, pxt.appTarget.versions.target) < 0) {
+            // the script is a major version ahead, need to redirect
+            pxt.tickEvent('patch.maxversion');
```
abchatra: Please add the versions to this tick

PR #4295, opened 2018-05-12
Green screen effect using WebCam

pelikhan:
Injects a video elements behind the editor and turns on transparency for various elements.
greenscreen

- [x] Cross platform support of getUserMedia (not supported in IE11, tested in Chrome, Edge, FF)
- [x] webcam selection
- [x] right align of video stream
- [x] monaco
- [x] pxtarget flag


webapp/src/sui.tsx
```diff
@@ -1132,4 +1132,48 @@ export class Loader extends UIElement<LoaderProps, {}> {
             {children}
         </div>
     }
+}
+
+export interface WebCamProps {
```
samelhusseini: WebCam is not an semantic ui component. Could you move it into a different tsx file.

samelhusseini: sui is reserved for semantic ui components.

pelikhan:
Yes cards!

_____________________________

PR #4298, opened 2018-05-12
Emit string literals encoded in UTF8

gedimitr:
When string literals were being emitted, the generated byte strings
would use only the lowest byte per Unicode code point. For ASCII
characters this is not an issue, since they have code points < 256.
For code points that are above 256 (e.g. Greek characters), the
characters would alias to unrelated ones, since only the lowest
byte would be used.

This changeset encodes string literals according to UTF8 before they
get emitted. ASCII characters are not affected by this change, but
characters that would be misrepresented are now encoded into
multiple bytes, allowing their accurate decoding.

### DIFF
--- a/pxtcompiler/emitter/hexfile.ts
+++ b/pxtcompiler/emitter/hexfile.ts
@@ -409,11 +409,12 @@ namespace ts.pxtc {
     function emitStrings(bin: Binary) {
         for (let s of Object.keys(bin.strings)) {
             let lbl = bin.strings[s]
+            let utf8_s = Util.toUTF8(s)
             // string representation of DAL - 0xffff in general for ref-counted objects means it's static and shouldn't be incr/decred
             bin.otherLiterals.push(`
 .balign 4
-${lbl}meta: .short 0xffff, ${s.length}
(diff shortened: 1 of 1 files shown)


mmoskal:
I think we will use 16 bit characters for unicode. This should happen within a month or so. 

With this methods like charAt and length will return invalid values.

mmoskal:
It's a bit complicated.

PXT targets other than microbit use their own string and buffer classes. These classes is what would be updated, so it would not carry over to `microbit-dal` or `codal-core`.

As we update microbit in future we might copy these classes as well, which would give unicode to `pxt-microbit`, but again not `microbit-dal`.

gedimitr:
I understand... It sounds though that Microbit Unicode support is not going to be available in the near future. Would you consider the UTF8 approach as a stopgap / interim solution (especially if the string related methods get adapted)?

mmoskal:
Proper unicode is now implemented - see https://github.com/Microsoft/pxt/pull/5144

It is still optional, so it's unclear it will make it into microbit, given that we only have an ASCII font in there...

PR #4306, opened 2018-05-14
Support custom handler key

samelhusseini:
Support block authoring a custom handler key to use when comparing block event duplicates.

### DIFF
--- a/pxtblocks/blocklycompiler.ts
+++ b/pxtblocks/blocklycompiler.ts
@@ -1263,6 +1263,7 @@ namespace pxt.blocks {
         isExpression?: boolean;
         imageLiteral?: number;
         hasHandler?: boolean;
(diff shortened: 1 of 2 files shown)


pxtblocks/blocklycompiler.ts
```diff
@@ -1460,6 +1461,7 @@ namespace pxt.blocks {
                         isExpression: fn.retType && fn.retType !== "void",
                         imageLiteral: fn.attributes.imageLiteral,
                         hasHandler: !!comp.handlerArgs.length || fn.parameters && fn.parameters.some(p => (p.type == "() => void" || !!p.properties)),
+                        handlerKey: fn.attributes.blockHandlerKey,
```
riknoll: You're already passing all the attributes in this object; no need to change the interface

Issue #4313, opened 2018-05-15
Decompile from instance method access fails when 1st parameter is enum type.


matthewtebbs:
Given this block attribution,

```typescript
//%
declare class Camera extends Object3d {
    /**
     * Attach camera controller.
     * @param type value, eg: CameraController.Orbit
     */
    //% blockId=camera_attachcontroller
    //% block="%camera|attach %type"
    //% group="Camera"
    //% shim=.attachController
    attachController(type: CameraController): void;
}
```

The following code that uses this instance method will emit the correct blocks as the decompiler goes down an incorrect type path for the `attachController` method block.

```typescript
world.scene().camera = design.perspectiveCamera()
world.scene().camera.attachController(CameraController.Orbit)
```

This is because in _pxtcompiler/emitter/decompiler.ts_,

```typescript
    function paramList(info: CallInfo, blocksInfo: BlocksInfo) {
        const res: DecompileArgument[] = [];
        const sym = blocksInfo.apis.byQName[info.qName];

        if (sym) {
            const attributes = blocksInfo.apis.byQName[info.qName].attributes;
            const comp = pxt.blocks.compileInfo(sym);
            const builtin = pxt.blocks.builtinFunctionInfo[info.qName]
            let offset = attributes.imageLiteral ? 1 : 0;

            if (comp.thisParameter) {
                res.push({
                    value: unwrapNode(info.args[0]) as Expression,
                    info: sym.parameters[0],
                    param: comp.thisParameter
                });
            }
            else if (attributes.defaultInstance) {
                res.push({
                    value: unwrapNode(info.args[0]) as Expression,
                    info: sym.parameters[0],
                    param: { definitionName: "__instance__", actualName: "this" }
                });
            }

            const hasThisArgInSymbol = !!(comp.thisParameter || attributes.defaultInstance);
            if (hasThisArgInSymbol) {
                offset++;
            }

            for (let i = offset; i < info.args.length; i++) {
                res.push({
                    value: unwrapNode(info.args[i]) as Expression,
                    info: sym.parameters[hasThisArgInSymbol ? i - 1 : i],
                    param: comp.parameters[i - offset]
                });
            }
        }

        return res;
    }
```

`sym.parameters[0]` is used for the parameter info for the implicit 'this' parameter and also for the 'type' parameter.

[...]

PR #4315, opened 2018-05-15
Do not provide non-sensical symbol information for implicit 'this' parameter.

matthewtebbs:
Fixes https://github.com/Microsoft/pxt/issues/4313.

### DIFF
--- a/pxtcompiler/emitter/decompiler.ts
+++ b/pxtcompiler/emitter/decompiler.ts
@@ -2580,7 +2580,7 @@ ${output}</xml>`;
             if (comp.thisParameter) {
                 res.push({
                     value: unwrapNode(info.args[0]) as Expression,
-                    info: sym.parameters[0],
+                    info: null,
                     param: comp.thisParameter
                 });
             }


matthewtebbs:
FWIW I looked at the test cases that were passing.

The text,

_./pxt/tests/decompile-test/cases/always_enums4.ts_

does have an enum as the first arg to the instance method. I haven't verified this, but I suspect the difference in my case is that the value for the 'this' argument is a combined property and that goes down a different path nowadays for various reasons. 

I still feel pretty good about my change.

matthewtebbs:
@riknoll yup all passing. I hadn't expected the _always_enums4_ to pass PRIOR to my change :) But I think the 'regression' here with enums as parameters happened with support for combined props that then put the code down a different path.

matthewtebbs:
Basically,

- Before my change, enums as 1st parameter to instance method worked.
- After combined props support enums to 1st parameter to instance method that had combined prop attributes didn't work.
- After my change, enums as 1st parameter to instance method works whether or not it has combined property attributes.

matthewtebbs:
(I think)

PR #4329, opened 2018-05-17
always include major version in embeddeable html

pelikhan:
image


samelhusseini:
There are currently issues with the backend returning a manifest for / instead of vN, so you'll run into some funky behaviour right now with this change. I'm fixing that, so shouldn't be long before it all works.

PR #4333, opened 2018-05-17
Use textContent and not innerHTML for getting code from marked

riknoll:
This fixes an issue where any tutorial snippet that had a "<" character was failing to compile because we were getting the escaped text out of the DOM instead of the unescaped text. Tested in Chrome, Firefox, IE, end Edge.

@pelikhan proof that it's impossible to test snippets 100% in the CLI :wink:


pelikhan:
What about https://github.com/Microsoft/pxt/pull/4258 ?

riknoll:
@pelikhan Before we were using `innerText` which didn't work in IE (stripped out the newlines), then we switched to `innerHTML` which fixed that issue but caused this new character escaping issue and now I'm using `textContent` which seems to work in all the browsers

riknoll:
Merged into master and stable3.13

PR #4334, opened 2018-05-18
Add the version into all links in the docs if it exists.

samelhusseini:
Adds vN to all links rendered by the markdown if it exists. We also add the vN prefix for all links in the docs sidemenu.

Reason for changes, if you start browsing on makecode.microbit.org/vN you want to stay browsing on that site, rather than switch to latest.

### DIFF
--- a/pxtlib/docsrender.ts
+++ b/pxtlib/docsrender.ts
@@ -442,7 +442,7 @@ namespace pxt.docs {
                 const relative = href.indexOf('/') == 0;
                 const target = !relative ? '_blank' : '';
                 if (relative && d.versionPath) href = `/${d.versionPath}${href}`;
-                return `<a href="${href}" ${title ? `aria-label="${title}"` : ''} ${target ? `target="${target}"`: ''}>${text}</a>`;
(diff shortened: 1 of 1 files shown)


pxtlib/docsrender.ts
```diff
@@ -434,6 +438,12 @@ namespace pxt.docs {
                 if (m) return `<li class="${m[1] == ' ' ? 'unchecked' : 'checked'}">` + text.slice(m[0].length) + '</li>\n'
                 return '<li>' + text + '</li>\n';
             }
+            renderer.link = function (href: string, title: string, text: string) {
+                const relative = href.indexOf('/') == 0;
+                const target = !relative ? '_blank' : '';
+                if (relative && d.versionPath) href = `/${d.versionPath}${href}`;
+                return `<a href="${href}" ${title ? `aria-label="${title}"` : ''} ${target ? `target="${target}"` : ''}>${text}</a>`;
```
pelikhan: are these strings escaped already?

pelikhan: This comment looks good: https://github.com/markedjs/marked/issues/655#issuecomment-383226346

samelhusseini: Good find. I'll use that comment instead and patch in my fix.

PR #4340, opened 2018-05-18
Show equality comparitor in toolbox

samelhusseini:
Show eq and less than block compare snippets in toolbox

We were showing the less than and greater than. 
Note: these are just defaults it was still possible to get to the eq but this makes more sense. 

Before: 
screen shot 2018-05-18 at 1 26 04 pm

After:
screen shot 2018-05-18 at 1 24 27 pm


pelikhan:
After is same as before?

samelhusseini:
Updated.
but after is the same as before blockly toolbox change

PR #4350, opened 2018-05-21
Pick changes from e2ed641

ganicke:
Merge changes from https://github.com/Microsoft/pxt/pull/4345

### DIFF
--- a/common-docs/javascript/functions.md
+++ b/common-docs/javascript/functions.md
@@ -43,19 +43,19 @@ TypeScript can figure the return type out by looking at the return statements, s
 
 ## Arrow Functions
 
-Arrow functions (also known as "lamba" functions) 
-provide a lightweight syntax for functions. Arrow functions are used extensively to provide event handlers for
-many APIs. For example:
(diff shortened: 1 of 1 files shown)


samelhusseini:
We typically use functions instead of lamda in master. Moved away from the direction in v0, and making the syntax closer to JavaScript.
 
Not sure if this doc page is supplementary, because users can still use it lamda, but that's not the recommendation. 

@pelikhan @riknoll

ganicke:
You mean stay away from both the arrow syntax and any reference to 'lambda'?

ganicke:
What terms are we using: callback, delegate, handler?

samelhusseini:
Yeah I'm not sure on the best thing to do here, but if our snippets and our recommendation paste in a function handler rather than an arrow function, then our docs should be consistent. 

You can still add a reference to a lamda function (arrow function) page since it still works. 

@pelikhan thoughts?

riknoll:
Callback is definitely the most javascript-y of those but I think our codebase and jsdoc usually use handler

pelikhan:
Move the arrow functions to the bottom.

PR #4357, opened 2018-05-23
Test dependencies

pelikhan:
Allow additional dependencies when compiling tests.


pelikhan:
I've been making some adjustments to our python converter. It's not perfect but it'll take a bulk of work out of you.

PR #4367, opened 2018-05-25
Do not inject targetversion when using "pxt build"

pelikhan:
Background: applying upgrade rules to Adafruit bundled projects leads to stack overflow issues. Upgrade rules should not be applied to bundled packages.

- [x] inject targetversion when saving project
- [x] log every file write with size (in debug)

In order to track down what goes on in the build, I've added a lot on every written file. This helps debugging when someone funny happens.

### DIFF
--- a/cli/nodeutil.ts
+++ b/cli/nodeutil.ts
@@ -276,7 +276,9 @@ export function readPkgConfig(dir: string) {
             }
         }
     }
-    if (!js.targetVersions) js.targetVersions = pxt.appTarget.versions;
+    // don't inject version number
+    // as they get serialized later on
+    // if (!js.targetVersions) js.targetVersions = pxt.appTarget.versions;
     return js
 }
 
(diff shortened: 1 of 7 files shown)


pelikhan:
No always on.

riknoll:
Uh, can we put it behind a verbose flag?

cli/nodeutil.ts
```diff
@@ -355,6 +357,12 @@ export function existsDirSync(name: string): boolean {
     }
 }
 
+export function writeFileSync(path: string, data: any, options?: { encoding?: string | null; mode?: number | string; flag?: string; } | string | null) {
+    fs.writeFileSync(path, data, options);
+    const stats = fs.statSync(path);
+    pxt.log(`  + ${path} ${stats.size > 1000000 ? (stats.size / 1000000).toFixed(2) + ' m' : stats.size > 1000 ? (stats.size / 1000).toFixed(2) + 'k' : stats.size}b`)
```
samelhusseini: Can this be pxt debug so it only runs if debug is ON.

PR #4368, opened 2018-05-25
added div warning flag

pelikhan:
When running ``pxt build --compat``, the compiler will do additional changes that might have compatibility issues. The first one is detecting that a package is using /  which needs to be addressed for micro:bit


pelikhan:
@samelhusseini use ``pxt testghpkgs --compat`` to turn on this mode when testing packages.

PR #4373, opened 2018-05-28
microbit v1 support

mmoskal:
* don't flash sources on the device for `.hex` targets (there's no read-back there anyway)
* stop creating `dal.d.ts` except when `pxt builddaldts`
* some ARM peephole improvements
* fixes for size stats


mmoskal:
They should work as is - the sources are still there, but behind the "end of file" marker, so daplink doesn't read them. Need to test though.

PR #4377, opened 2018-05-29
Support for embedding inside of LMS controllers

samelhusseini:
Initial support for embedding editors inside of an LMS controller. 
Remove things like: 
- Home screen 
- Share 
- File re-naming
- Delete project and reset

Don't load our Winrt stack if we're in an iframe. 

you can test using lmscontroller.html (which currently is mostly the same as controller.html but will be different in the future)


pelikhan:
I would say drop "LMS"-- just "controller".

samelhusseini:
We need a tick here too.

samelhusseini:
I renamed to controller, plus I added a tick to differentiate the source (controller) id. Field is optional, so we might get a whole lot of undefined if others use this mode.

PR #4386, opened 2018-05-30
Support for copyright text in the footer

samelhusseini:
Option to add footer / copyright text at the bottom of the Home screen and About dialog.

screen shot 2018-05-30 at 3 33 53 pm

screen shot 2018-05-30 at 3 34 37 pm

### DIFF
--- a/pxtblocks/fields/field_images.ts
+++ b/pxtblocks/fields/field_images.ts
@@ -134,7 +134,7 @@ namespace pxtblockly {
             }
         }
 
-        private createTextNode_(text: string) {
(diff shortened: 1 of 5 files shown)


pxtblocks/fields/field_images.ts
```diff
@@ -134,7 +134,7 @@ namespace pxtblockly {
             }
         }
 
-        private createTextNode_(text: string) {
+        protected createTextNode_(text: string) {
```
samelhusseini: unrelated change

PR #4389, opened 2018-05-31
recorver from unknown definitions in pyconv

pelikhan:
Recover when failing to convert a single FunctionDef or ClassDef.


cli/pyconv.ts
```diff
@@ -908,9 +912,17 @@ function typeAnnot(t: Type) {
     return B.mkText(": " + t2s(t))
 }
 
+function gardedScope(v: py.AST, f: () => B.JsNode) {
```
mmoskal: Guarded?

PR #4396, opened 2018-06-01
Don't reload if already in the selected language

### LINKED ISSUE #4141: [Language dialog] Selecting "English" while already in English does a reload
This is a regression

guillaumejenkins:
Fixes #4141

### DIFF
--- a/pxtlib/util.ts
+++ b/pxtlib/util.ts
@@ -738,9 +738,12 @@ namespace ts.pxtc.Util {
     }
 
     export function normalizeLanguageCode(code: string): string {
-        if (!/^(es|pt|si|sv|zh)/i.test(code))
-            code = code.split("-")[0]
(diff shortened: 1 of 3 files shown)


pxtlib/util.ts
```diff
@@ -737,7 +737,7 @@ namespace ts.pxtc.Util {
 
     }
 
-    function normalizeLanguageCode(code: string): string {
+    export function normalizeLanguageCode(code: string): string {
         if (!/^(es|pt|si|sv|zh)/i.test(code))
```
pelikhan: we should revisit this...

guillaumejenkins: Agreed that it's weird... Changed the normalization logic to instead ensure casing is consistent: `xx` and `xx-XX`

PR #4423, opened 2018-06-05
Fix in editor check for card links (versioning)

samelhusseini:
Fix in editor check for whether or not to append target version to card urls.

Previous PR https://github.com/Microsoft/pxt/pull/4420 didn't correctly check for inEditor.


samelhusseini:
Yeah

PR #4432, opened 2018-06-06
Check shadow type on load

pelikhan:
If type block for a shadowType is not found, just don't use it!!!
@riknoll can we get this in master too?


riknoll:
You'll need to do something in the Decompiler also

pelikhan:
@riknoll added decompiler handling too

abchatra:
@riknoll how can I find out if this was cherry picked to master?

PR #4443, opened 2018-06-07
Fix issue with Safari not showing the right flyout heading labels.

samelhusseini:
When in an iframe and on Safari, 
we're seeing "iframe" or the iframe iD as the name in the flyout. 

This fixes that. 

Screenshot of issue: 
screen shot 2018-06-07 at 3 22 22 pm

### DIFF
--- a/webapp/src/toolbox.tsx
+++ b/webapp/src/toolbox.tsx
@@ -5,8 +5,6 @@ import * as data from "./data"
 import * as editor from "./toolboxeditor"
 import * as sui from "./sui"
 import * as core from "./core"
-import * as snippets from "./monacoSnippets"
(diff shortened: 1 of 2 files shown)


webapp/src/toolbox.tsx
```diff
@@ -5,8 +5,6 @@ import * as data from "./data"
 import * as editor from "./toolboxeditor"
 import * as sui from "./sui"
 import * as core from "./core"
-import * as snippets from "./monacoSnippets"
```
samelhusseini: Cleanup

pelikhan:
Funny bug...

samelhusseini:
Yeah, very.

PR #4444, opened 2018-06-07
patch apptarget when setting variants

pelikhan:
Whenever we assign a new apptarget (this happen once on load and every time we load a new board), we apply the patching logic.

Fixes lost images in maker extensions dialog.


samelhusseini:
Please test the different ways we publish: 
- staticpkg
- staticpkg --gh
- uploadtrg

pelikhan:
This only impacts runtime patching of app target.

PR #4446, opened 2018-06-07
Support for transparent hardware configs

mmoskal:
This adds support for `hw---XYZ` packages, that replace `hw` package with board-specific config. This is currently driven by `?hw=XYZ` query argument, but will use `config.bin` file dropped on the editor later.

Additionally:
* `additionalFilePath` can be recursive
* support embedding source in non-flashed part of UF2


localtypings/pxtpackage.d.ts
```diff
@@ -48,6 +48,7 @@ declare namespace pxt {
         npmDependencies?: Map<string>;
         card?: CodeCard;
         additionalFilePath?: string;
```
pelikhan: do we still need this one?

mmoskal: Yes. User never sets `additionalFilePaths`, they only set `additionalFilePath`.

pelikhan:
Let's wait to push our Adafruit today.

PR #4450, opened 2018-06-08
Lazy load native dependencies

pelikhan:
Most of our native dependencies are not needed for compilation or other tasks. They do create a ton of warnings or fail the installation -- or simply fails in OS like Windows 7 when precompiled images are not available.

- [x] This PR moves those depencencies into a "lazyDependencies" section in package.json
- [x] they are not installed by default, one has to run ``pxt npm-install-native``
- [x] when missing an error message with the instruction above is displayed
- [x] upgrading serialport and node-hid

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -147,32 +147,27 @@ interface KeyTar {
     deletePassword(service: string, account: string): Promise<void>;
 }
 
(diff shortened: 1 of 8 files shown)


pelikhan:
FYI @guillaumejenkins newer versions of node-hid and serialport have better support for Electron
+ need to take into account this field in OSS tool

PR #4454, opened 2018-06-08
More samples + built-in field editors

pelikhan:
Moving a bunch of field editors in common packages.

### DIFF
--- a/docs/static/playground/samples/basic/hello-world/sample.js
+++ b/docs/static/playground/samples/basic/hello-world/sample.js
@@ -8,7 +8,7 @@
 namespace basic {
 
     //% block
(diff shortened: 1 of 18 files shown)


docs/static/playground/samples/basic/hello-world/sample.js
```diff
@@ -1,9 +1,14 @@
...
+ * start with % (//% block).
+ * The most important macro "block" specifies that a block should be generated
+ * for a **exported** function.
+ */
 //% color="#AA278D"
 namespace basic {
 
     //% block
-    export function makeCode() {
+    export function hellWorld() {
```
samelhusseini: Probably fix this.

PR #4464, opened 2018-06-12
Electron drive deploy

guillaumejenkins:


webapp/src/electron.ts
```diff
@@ -1,4 +1,5 @@
 import Cloud = pxt.Cloud;
+import * as cmds from "./cmds";
```
pelikhan: circular dependency, why do you need this here?

webapp/src/electron.ts
```diff
@@ -94,6 +112,26 @@ export function initElectron(projectView: ProjectView): void {
...
+    }
+
+    if (!deployingDeferred) {
+        deployingDeferred = Promise.defer<void>();
+        pxtElectron.sendDriveDeploy(compileResult);
+    }
+
+    return deployingDeferred.promise
+        .catch((e) => {
+            return cmds.browserDownloadDeployCoreAsync(compileResult);
```
pelikhan: add a tick

PR #4469, opened 2018-06-13
Minor fix to doc versioning

samelhusseini:
Minor fix to doc versioning, failing to add a slash when it's not there.

We should probably write some sort of path joining method and use that instead.

### DIFF
--- a/pxtlib/browserutils.ts
+++ b/pxtlib/browserutils.ts
@@ -497,6 +497,8 @@ namespace pxt.BrowserUtils {
      * Simple utility method to join urls.
      */
     export function urlJoin(urlPath1: string, urlPath2: string): string {
+        if (!urlPath1) return urlPath2;
+        if (!urlPath2) return urlPath1;
         const normalizedUrl1 =  (urlPath1.indexOf('/') == urlPath1.length - 1) ?
             urlPath1.substring(0, urlPath1.length - 1) : urlPath1;
         const normalizedUrl2 =  (urlPath2.indexOf('/') == 0) ?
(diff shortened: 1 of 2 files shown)


pxtrunner/runner.ts
```diff
@@ -328,7 +328,8 @@ namespace pxt.runner {
                     const docsUrl = pxt.webConfig.docsUrl || '/--docs';
                     let verPrefix = mp[2] || '';
                     let url = mp[3] == "doc" ? `${mp[4]}` : `${docsUrl}?md=${mp[4]}`;
-                    window.open(verPrefix + url, "_blank");
+                    let fullUrl = (url.indexOf('/') == 0) ? verPrefix + url : verPrefix + "/" + url;
```
pelikhan:
```
verPrefix.replace(/\/$/, '') + '/' + ...
```

samelhusseini: Issue is in some cases the url starts with "/" and in other cases it doesn't. 
verPrefix is authored by us, and so we control that. I don't add a "/" at the end of verPrefix.

pelikhan:
You can do the same “normalization” treatment to url as well. Probably best to have a urlJoin helper to do it right - always

samelhusseini:
Agreed.

pxtlib/browserutils.ts
```diff
@@ -492,4 +492,15 @@ namespace pxt.BrowserUtils {
             window.history.replaceState('', '', hash)
         }
     }
+
+    /**
+     * Simple utility method to join urls.
+     */
+    export function urlJoin(urlPath1: string, urlPath2: string): string {
+        const normalizedUrl1 =  (urlPath1.indexOf('/') == urlPath1.length - 1) ?
```
pelikhan: let's make sure to handle undefiuned

PR #4478, opened 2018-06-14
'release-tests' might be under the wrong doc tree

ganicke:
Was it intended to put this under _common-docs_ such that they are visible in target doc sets?

See: https://makecode.adafruit.com/release-tests

- [x] _release-tests_ doc set moved under _/docs_


pelikhan:
Yes they should be under common-docs so that we can weave in code snippets.

pelikhan:
I've reverted those changes.

ganicke:
Ok

PR #4479, opened 2018-06-14
pxt Gdb docs

pelikhan:
A straw man docs. @mmoskal can you beef up?


ganicke:
Does _markd_ render tabbed blocks the same as "```"? Probably not. GitHub MD does.

PR #4480, opened 2018-06-14
Bump major to 3.20

guillaumejenkins:
Needed to create stable3.19, out of which Minecraft will ship

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pxt-core",
-  "version": "3.20.1",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "pxt-core",
-  "version": "3.19.4",
+  "version": "3.20.1",
```
samelhusseini: Why .1 not 0?

guillaumejenkins: Sure :)

PR #4494, opened 2018-06-18
pxt serve uses cloud by default

pelikhan:
This is a shift from favoring local native to cloud native build. The reason is that local builds often fail because of configuration issue. We do have a cloud which makes the experience much more seamless (and slower). New flag --local to force local build.
Better log output of failed builds.
Fixed some incorrect caching of promises.


pelikhan:
staticpkg looks ok

PR #4502, opened 2018-06-19
Support project tags in home screen.

samelhusseini:
Support for project tags in home screen, like so: 
screen shot 2018-06-19 at 11 37 02 am

- Supports multiple tags
- You can choose a color associated with the tag label (it must be one of the semantic colors). 

Example of the authoring: 
```
  "tags": [{ "label": "Beginner" }],
```
```
  "tags": [{ "label": "Beginner", "color": "blue" }],
```
```
  "tags": [{ "label": "Beginner", "color": "blue" }, { "label": "Fun", "color": "green" }],
```

Note: tags don't appear in the docs, just in the home screen.

### DIFF
--- a/webapp/src/codecard.tsx
+++ b/webapp/src/codecard.tsx
@@ -115,7 +115,7 @@ export class CodeCardView extends data.Component<pxt.CodeCard, CodeCardState> {
             {card.extracontent || card.tags ? <div className="extra content">
                 {card.extracontent}
                 {card.tags ? card.tags.map(tag =>
-                    <span key={`tag${tag.label}`} className={`ui label tiny ${tag.color}`}>{tag.label}</span>
+                    <span key={`tag${tag.label}`} className={`ui label tiny ${tag.color}`}>{pxt.Util.rlf(tag.label)}</span>
                 ) : undefined}
             </div> : undefined}
         </div>;
--- a/webapp/src/codecard.tsx
+++ b/webapp/src/codecard.tsx
@@ -112,7 +112,12 @@ export class CodeCardView extends data.Component<pxt.CodeCard, CodeCardState> {
             {card.time ? <div className="meta">
                 {card.time ? <span key="date" className="date">{pxt.Util.timeSince(card.time)}</span> : null}
             </div> : undefined}
-            {card.extracontent ? <div className="extra content"> {card.extracontent} </div> : undefined}
+            {card.extracontent || card.tags ? <div className="extra content">
+                {card.extracontent}
+                {card.tags ? card.tags.map(tag =>
+                    <span key={`tag${tag.label}`} className={`ui label ${tag.color}`}>{tag.label}</span>
+                ) : undefined}
+            </div> : undefined}
         </div>;
 
         if (!card.onClick && url) {
--- a/webapp/src/codecard.tsx
+++ b/webapp/src/codecard.tsx
@@ -115,7 +115,7 @@ export class CodeCardView extends data.Component<pxt.CodeCard, CodeCardState> {
             {card.extracontent || card.tags ? <div className="extra content">
                 {card.extracontent}
                 {card.tags ? card.tags.map(tag =>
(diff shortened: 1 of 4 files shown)


webapp/src/codecard.tsx
```diff
@@ -112,7 +112,12 @@ export class CodeCardView extends data.Component<pxt.CodeCard, CodeCardState> {
             {card.time ? <div className="meta">
                 {card.time ? <span key="date" className="date">{pxt.Util.timeSince(card.time)}</span> : null}
             </div> : undefined}
-            {card.extracontent ? <div className="extra content"> {card.extracontent} </div> : undefined}
+            {card.extracontent || card.tags ? <div className="extra content">
+                {card.extracontent}
+                {card.tags ? card.tags.map(tag =>
+                    <span key={`tag${tag.label}`} className={`ui label ${tag.color}`}>{tag.label}</span>
```
pelikhan: Can you wrap it in rlf and add a few predefined labels?

samelhusseini: It is localizable like the rest of our docs. Using md. 
I agree it's not easy, but we don't have localization support for any other md meta data. why block on this?

pelikhan:
We should have a known set of tags "beginner", "intermediate", ... for which we have translation available (by adding dummy comments). Then rlf would work. This will make it much easier to translate automatically and keep the tags consistent.

pelikhan:
The problem is that our JSON format defeats Crowdin automated translation sharing. So unlike titles, those tags would have to be tagged over and over.
Scanning for those tags and adding them to target-strings.json sounds like a good idea.

samelhusseini:
Makes sense, other than tags and labels (eg: "Getting started"), is there any other meta data that can be added to this proposed target-json?

pelikhan:
``target-strings.json`` is already generated as part of the build. We need to update the CLI to parse and grab the tags from various tutorials page... Or simple hard code various well know strings like: "beginner", "intermediate", "advanced" to get started.

pelikhan:
https://github.com/Microsoft/pxt/blob/690115cce93b4a16c7efc2734871682b8d616853/cli/cli.ts#L1662

pelikhan:
You can probably do it here: https://github.com/Microsoft/pxt/blob/690115cce93b4a16c7efc2734871682b8d616853/cli/cli.ts#L1654

pelikhan:
I've swam in those water, let me add this

pelikhan:
Added the loc code, how do you test?

abchatra:
Can we just pull the tags & label from the doc file itself https://github.com/Microsoft/pxt-adafruit/blob/master/docs/projects/dandelion.md?  In this way if the doc file is localized these labels & tags are localized, otherwise they are not

abchatra:
I guess @pelikhan just did that?

pelikhan:
We do not load the markdown sources of tutorial until we load them.
Also if we do not auto-localize tags, we’ll end up with a mixed bag of translations. Crowdin does not properly parse our embedded JSON so we don’t get the benefits of shared translations.


samelhusseini:
Tested target-strings.json and it works fine, sample output: 
```
{
  "A Blocks / JavaScript code editor for the Adafruit Circuit Playground Express.": "A Blocks / JavaScript code editor for the Adafruit Circuit Playground Express.",
  "Adafruit Circuit Playground Express": "Adafruit Circuit Playground Express",
  "Adafruit Circuit Playground Express - Blocks / Javascript editor": "Adafruit Circuit Playground Express - Blocks / Javascript editor",
  "Beginner": "Beginner",
  "Blocks": "Blocks",
  "Buy": "Buy",
  "Costumes": "Costumes",
  "Fun": "Fun",
  "Games": "Games",
  "Intermediate": "Intermediate",
  "JavaScript": "JavaScript",
  "Music": "Music",
  "Pranks": "Pranks",
  "Reference": "Reference",
  "Support": "Support",
  "Tutorials": "Tutorials",
  "Wearables": "Wearables",
  "learn.adafruit.com": "learn.adafruit.com"
}
```

PR #4505, opened 2018-06-20
Toolbox namespace casing

samelhusseini:
Match TS behaviour by not converting namespace to lowercase in the toolbox.

Fixes https://github.com/Microsoft/pxt-adafruit/issues/782
Fixes https://github.com/Microsoft/pxt-microbit/issues/879

When I re-did the toolbox, I made the assumption that TS merged all namespace names regardless of case, but that's not true. This change allows two categories with different cases to appear in different toolbox categories (just like TS).

### DIFF
--- a/webapp/src/blocks.tsx
+++ b/webapp/src/blocks.tsx
@@ -988,7 +988,7 @@ export class Editor extends toolboxeditor.ToolboxEditor {
     getNamespaceAttrs(ns: string) {
         const builtin = snippets.getBuiltinCategory(ns);
         if (builtin) {
-            builtin.attributes.color = pxt.toolbox.getNamespaceColor(builtin.nameid.toLowerCase());
+            builtin.attributes.color = pxt.toolbox.getNamespaceColor(builtin.nameid);
             return builtin.attributes;
         }
         if (!this.blockInfo) return undefined;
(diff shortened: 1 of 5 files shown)


webapp/src/blocks.tsx
```diff
@@ -989,7 +988,7 @@ export class Editor extends toolboxeditor.ToolboxEditor {
     getNamespaceAttrs(ns: string) {
         const builtin = snippets.getBuiltinCategory(ns);
         if (builtin) {
-            builtin.attributes.color = pxt.toolbox.getNamespaceColor(builtin.nameid);
+            builtin.attributes.color = pxt.toolbox.getNamespaceColor(builtin.nameid.toLowerCase());
```
pelikhan: move toLowerCase inside getNamespaceColor ?

webapp/src/monaco.tsx
```diff
@@ -881,7 +880,7 @@ export class Editor extends toolboxeditor.ToolboxEditor {
     getNamespaceAttrs(ns: string) {
         const builtin = snippets.getBuiltinCategory(ns);
         if (builtin) {
-            builtin.attributes.color = pxt.toolbox.getNamespaceColor(builtin.nameid);
+            builtin.attributes.color = pxt.toolbox.getNamespaceColor(builtin.nameid.toLowerCase());
```
pelikhan: move to getNamespaceColor

PR #4514, opened 2018-06-21
always show markdown in blocks mode

pelikhan:
Since the only way to reach that file is through Explorer, it always end up to JavaScript.

### DIFF
--- a/webapp/src/container.tsx
+++ b/webapp/src/container.tsx
@@ -439,7 +439,9 @@ export class SideDocs extends data.Component<SideDocsProps, SideDocsState> {
 
     setMarkdown(md: string) {
         const docsUrl = pxt.webConfig.docsUrl || '/--docs';
-        const mode = this.props.parent.isBlocksEditor() ? "blocks" : "js";
+        // always render blocks by default when sending custom markdown
+        // to side bar
+        const mode = "blocks" // this.props.parent.isBlocksEditor() ? "blocks" : "js";
         const url = `${docsUrl}#md:${encodeURIComponent(md)}:${mode}:${pxt.Util.localeInfo()}`;
         this.setUrl(url);
     }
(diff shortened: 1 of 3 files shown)


samelhusseini:
Not sure about this one, let me verify when I get in to work

pelikhan:
I searched and I did not see it being used anywhere else.

webapp/src/container.tsx
```diff
@@ -439,7 +439,9 @@ export class SideDocs extends data.Component<SideDocsProps, SideDocsState> {
 
     setMarkdown(md: string) {
         const docsUrl = pxt.webConfig.docsUrl || '/--docs';
-        const mode = this.props.parent.isBlocksEditor() ? "blocks" : "js";
+        // always render blocks by default when sending custom markdown
+        // to side bar
+        const mode = "blocks" // this.props.parent.isBlocksEditor() ? "blocks" : "js";
```
samelhusseini: How come?

samelhusseini:
Hmmm. It's used in docs.html, error.html, package.html, script.html, stream.html

webapp/src/container.tsx
pelikhan: The way to get through Readme.md is through the explorer, which de-facto sets the editor to javascript. However, the setting markdown feature is mostly useful to preview readme.md... which mostly is about seeing blocks.

pelikhan:
Indeed, somehow search is now poping those up. It seems that the sidedocs does not have this "mainbody" id then.

pelikhan:
I'm reporposing this for editing markdown.

samelhusseini:
It's by design that way. I'll keep digging why the docs.less ui.segment CSS is not being applied in this case.

samelhusseini:
Yep, you're right we need a mainbody in webapp/docs.html

samelhusseini:
(Made the change here)

PR #4521, opened 2018-06-22
Fix duplicate on drag issue when dragging from flyout

samelhusseini:
Gif says it all.

duplicatedragissue

(Notice how two things get duplicated and the whole block gets duplicated when dragging from the flyout)


riknoll:
Closing in favor of https://github.com/Microsoft/pxt/pull/4567

PR #4525, opened 2018-06-25
Subcategory for card types

ganicke:
Isn't ``--snippets`` needed here too?


pelikhan:
It’s ok by default now.

PR #4531, opened 2018-07-04
Fix docs link rendering

guillaumejenkins:
We had a closure variable capturing bug in our logic to render docs links

On [line 461](https://github.com/Microsoft/pxt/blob/master/pxtlib/docsrender.ts#L461) we initialize the `link()` function in a way that captures the state of the `d` object and keeps it forever. Subsequent calls to the `link` function do **not** use the updated `d` object that we set on [line 446](https://github.com/Microsoft/pxt/blob/master/pxtlib/docsrender.ts#L446), but rather the one from the function's closure

Fix is to re-assign the `renderer.link()` function every time, so it always uses the updated `d` object (and the updated `d.versionPath` value to correctly update the `href`s).

### DIFF
--- a/pxtlib/docsrender.ts
+++ b/pxtlib/docsrender.ts
@@ -455,27 +455,29 @@ namespace pxt.docs {
 
         if (!markedInstance) {
             markedInstance = requireMarked();
-            let renderer = new markedInstance.Renderer()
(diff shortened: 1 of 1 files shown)


abchatra:
Nice catch

PR #4538, opened 2018-07-09
Honor the disableLiveTranslations pxtarget setting

guillaumejenkins:
For some reason in v0 we hardcode live localizations to true, even if pxtarget.json disables them. This prevents package authors from testing their package translations.

This change has no impact on production websites, because we never use the `disableLiveTranslations` flag in pxtarget.json. That setting is mostly meant for local serve.

Note that in master, this has already been fixed

### DIFF
--- a/pxtlib/util.ts
+++ b/pxtlib/util.ts
@@ -810,7 +810,9 @@ namespace ts.pxtc.Util {
 
         _localizeLang = code;
         _localizeStrings = {};
-        localizeLive = true;
+        if (live) {
+            localizeLive = true;
+        }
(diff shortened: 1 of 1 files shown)


pxtlib/util.ts
```diff
@@ -810,7 +810,9 @@ namespace ts.pxtc.Util {
 
         _localizeLang = code;
         _localizeStrings = {};
-        localizeLive = true;
+        if (live) {
```
riknoll: `localizeLive = !!live` (feel free to ignore this nitpick)

PR #4542, opened 2018-07-16
Adding editing of packages on github

mmoskal:
This PR adds support for:
* import of existing packages from github for editing
* pushing changes of imported github packages
* upon commit (integrated with push/pull) users can specify check in message and bump
* we use github for merging
* in case automatic merging fails, a PR is created on github and user is moved to current master
* listing all repos in user's account and importing it
* creating new repos
* initializing existing repos (adding `pxt.json` and friends)
* adding packages in local workspace as extensions
* logging in via github - the token is obtained through our cloud, but is only stored locally; all github operations except for login are done without involving our cloud

This should dramatically improve user's ability to author packages. I have so far only tested this with localhost, but should work on cloud instances. Waiting on backend PR for this.

TODO:
* [x] replace `workspace:...` deps with `github:...` upon publishing

screen shot 2018-07-17 at 11 58 32
screen shot 2018-07-17 at 11 58 39
screen shot 2018-07-17 at 11 59 00

screen shot 2018-07-16 at 14 06 36
screen shot 2018-07-16 at 14 06 46
screen shot 2018-07-16 at 14 07 17
screen shot 2018-07-16 at 14 07 22
screen shot 2018-07-16 at 14 07 53
screen shot 2018-07-16 at 14 08 28
screen shot 2018-07-16 at 14 09 00


cli/server.ts
```diff
@@ -99,41 +99,44 @@ function readAssetsAsync(logicalDirname: string): Promise<any> {
...
-                })
-        })
+    let buf = await readFileAsync(path.join(dirname, pxt.CONFIG_NAME))
+    let cfg: pxt.PackageConfig = JSON.parse(buf.toString("utf8"))
+    let r: FsPkg = {
+        path: logicalDirname,
+        config: cfg,
+        files: []
+    };
+
```
guillaumejenkins: Linting error on this line: `ERROR: /home/travis/build/Microsoft/pxt/cli/server.ts[111, 1]: trailing whitespace`

pxtlib/sha1.ts
```diff
@@ -0,0 +1,182 @@
...
+FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+
+// Based on https://github.com/creationix/git-sha1
+// Converted to TypeScript and modified slightly
+
+namespace ts.pxtc.Util {
```
riknoll: We will definitely need to get approval for this file. Not because of the license, but because it implements sha1

abchatra:
This is pretty cool! I haven't looked at code yet. We discussed user experience with Jacqueline and one comment. 

_GitHub Login_ entry point should go to extensions page and not in the import. Login in import indicates user projects will be  saved in that login and it unintiutive that it is only for package writers. We need that space for office 365 login in future. 

Other than that it looks pretty cool.

cli/server.ts
mmoskal: Right, fixed these, thanks!

pxtlib/sha1.ts
mmoskal: @abchatra could you put in OSS req for this one?

The SHA-1 isn't security relevant here, but I'm not sure if that helps. Maybe there is some approved JS SHA-1 we could use?

mmoskal:
@guillaumejenkins does Electron disable all online interactions? If not, I see no reason to disable this. We would need a slightly different login experience (have the app long-poll on a server end point to get the token, which is obtained from the regular browser), but it's certainly doable.

As for the token - indeed our cloud sees it, but only in transit. We probably need to move the app to Microsoft github org.

As for the contributions, I guess the optimal solution would be to fork on import if the user has no write access to the repo, but PR creation is probably best handled by github web interface. I would leave this for later.

mmoskal:
@abchatra I moved the login link to "About" dialog. This is even more obscure than extensions dialogs, and we might keep it there as we refine the feature. I've also changed it to use personal access tokens for now.

However, I don't think this is only for package writers. I can easily imagine high school students collaborating on an Arcade project using GitHub - they would learn it in a quite soft way @eric-camplin 

Also, people can in fact use it for storing their own projects.

abchatra:
Students are not going to have GitHub login and we have plan for o365 this year. Package writers perspective, extensions page is better in my opinion. About is for versions and I am not sure login fits there. @Jaqster what is your opinion?

Is there any limitation in extensions page to add this?

Jaqster:
My vote is to put this feature in the Extensions dialog.  This seems like the most natural entry point for developers who want to write Extensions/Packages.  It's weird in the About page...  I would also say that this feature is most definitely for developers at this stage.  Yes, maybe in 2-3 years it will become more mainstream for students, but not today.  We can talk about how to surface this in the Extensions page... maybe a '+' card or something at the end of the list to prompt people to Add Your Own...

mmoskal:
I guess should be easy to add at the bottom of extensions page. At the top it's problematic since it's common modal stuff.

O365 is not going to cut the collaboration scenario, only storage. I think it might be also typical to teach source control together with a basic programming course. But let's just wait and see.

abchatra:
I registered crypto-js and it was auto approved https://ossmsft.visualstudio.com/_oss?requestId=659384&_a=progress

Please add it to 3rd party notices file

riknoll:
This needs to be behind a flag (the UI part). No need for this in the Cue app. Otherwise looks good!

mmoskal:
@riknoll @abchatra good to merge?

abchatra:
Yes go for it

PR #4552, opened 2018-07-26
Sprite editor overhaul

riknoll:
new_sprite_editor


riknoll:
There are some UI issues in Edge that I'm still working out but it's fine.

PR #4558, opened 2018-07-30
Make groups and subcategories localizable

guillaumejenkins:
- Extract subcategories strings
- Copy groups and subcategories translations to the main editor translations so they can be used
- Fix React duplicate key bug with subcategories
- Fix flyout headers of subcategories by showing only the localized subcategory name
  - Previously, it was showing the non-translated category, followed by the non-translated subcategory

Fixes https://github.com/Microsoft/pxt-microbit/issues/958

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "pxt-core",
-  "version": "3.22.4",
+  "version": "3.22.5",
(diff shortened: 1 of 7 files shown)


riknoll:
Finally, the react keys error is dead!

PR #4562, opened 2018-07-31
Fixing CSS for notification and cookie banners

riknoll:
Note this is a v0 pull request! @abchatra and I made a change to support the notification banner in V0 and it ended up affecting the cookie banner as well. This fixes that issue along with a another I found while testing (the simulator was positioned incorrectly when the notification banner shows on tablet/mobile view)

Fixes https://github.com/Microsoft/pxt-microbit/issues/1000
Fixes https://github.com/Microsoft/pxt-microbit/issues/997

### DIFF
--- a/theme/common.less
+++ b/theme/common.less
@@ -32,6 +32,10 @@ html, body {
     position: relative;
 }
 
+#mainmenu {
+    position: relative !important;
+}
+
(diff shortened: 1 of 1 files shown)


abchatra:
Try on /Prev before bumping live

PR #4565, opened 2018-08-01
Fixing the color of the piano editor when not a shadow block

riknoll:
Fixes https://github.com/Microsoft/pxt-microbit/issues/931

Updating the logic for the piano field editor so that it matches its color with whatever block it is nested in. If a block has only one field then it also doesn't have a color because single-field blocks render just the field edge-to-edge. In that case we need to get the color from the parent block or else it ends up looking arbitrary.


samelhusseini:
What are the scenarios this is trying to address?

I'm still seeing a gray piano when the block is on it's own. 

screen shot 2018-10-14 at 3 01 11 pm

@riknoll

PR #4571, opened 2018-08-02
Handle bool default parameters in shims

riknoll:
Fixes https://github.com/Microsoft/pxt-microbit/issues/1019

Tested on the micro:bit and seems to work fine


riknoll:
Ah, I missed that function call! But yeah that shouldn't make it past the typechecker

PR #4588, opened 2018-08-10
Bring back the C++ runtime version in the about dialog

riknoll:
Fixes https://github.com/Microsoft/pxt-microbit/issues/1063

Looks like it was originally [removed](https://github.com/Microsoft/pxt/pull/4436) because the link was wrong. I've brought it back and fixed that issue. Also, I'm now using the service ID in place of "C++" so instead of reading `C++ runtime version` it now reads `microbit runtime version`. 

@mmoskal can you take a quick look and let me know if I'm using the correct fields from pxtarget?

### DIFF
--- a/webapp/src/dialogs.tsx
+++ b/webapp/src/dialogs.tsx
@@ -137,6 +137,7 @@ export function showAboutDialogAsync() {
     const githubUrl = pxt.appTarget.appTheme.githubUrl;
     const targetTheme = pxt.appTarget.appTheme;
     const versions: pxt.TargetVersions = pxt.appTarget.versions;
+    const showCompile = compileService && compileService.githubCorePackage && compileService.gittag && compileService.serviceId;
 
     core.confirmAsync({
         header: lf("About"),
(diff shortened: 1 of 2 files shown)


webapp/src/dialogs.tsx
```diff
@@ -168,6 +165,38 @@ export function showAboutDialogAsync() {
...
 
+
+function renderCompileLink(cs: pxt.TargetCompileService) {
+    let url: string;
+    let version: string;
+    let name: string;
+
+    if (typeof cs.codalTarget === "object" && typeof cs.codalTarget.url === "string") {
+        // There is also a branch in this object but looks like it is sometimes a branch and sometimes a tag
+        // so can't really generate a good URL
```
mmoskal: Assume `branch` is in fact a tag. Assume it's always there. It can be branch on non-production targets, but we don't care in that case. Don't take the `cs.gittag` - this is a version of the "build system", just default to `"master"` or something.

mmoskal: I mean, build the link as if `branch` was always a tag. If it's a branch it messes up caching, so even during development it's almost always a tag. We don't care if the link is broken otherwise.

mmoskal:
No need for codal version - it's only the "build system", rarely changes and unlikely to cause trouble

riknoll:
Thanks @mmoskal! updated.

mmoskal:
Looks good!

PR #4592, opened 2018-08-10
Adding idiv and imul to the math operation block

riknoll:
We are getting to the point where it might make sense to add some code to filter what appears in this block so it can be controlled per-target

Part of https://github.com/Microsoft/pxt-microbit/issues/1062

### DIFF
--- a/pxtlib/blocks.ts
+++ b/pxtlib/blocks.ts
@@ -7,7 +7,8 @@ namespace pxt.blocks {
     // determines the order of the dropdown in the math_js_op block
     export const MATH_FUNCTIONS = {
         unary: ["sqrt", "sin", "cos", "tan", "round", "ceil", "floor", "trunc"],
-        binary: ["atan2", "idiv", "imul"]
+        binary: ["atan2"],
+        infix: ["idiv", "imul"]
     };
 
     export interface BlockParameter {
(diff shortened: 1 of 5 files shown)


pxtlib/blocks.ts
```diff
@@ -408,6 +410,8 @@ namespace pxt.blocks {
                     "ceil": Util.lf("{id:op}ceiling"),
                     "floor": Util.lf("{id:op}floor"),
                     "trunc": Util.lf("{id:op}truncate"),
+                    "idiv": Util.lf("{id:op}integer ÷"),
+                    "imul": Util.lf("{id:op}integer ×"),
```
mmoskal: We seem to be using `/` and `*` in Math blocks - will this appear as `[10] integer ÷ [3]` ?

riknoll: Where? We use the symbols in the arithmetic blocks:

image

And right now it looks like this:

image

I'll try and move it to the center because I agree that would be much better

mmoskal: Whoops, I was looking in TypeScript toolbox. Never mind.

riknoll:
@guillaumejenkins I made some changes to add infix operations. Can you take a look?

PR #4597, opened 2018-08-14
Fixing monaco snippets for fixed instances, enums, and block definitions

riknoll:
Fixes https://github.com/Microsoft/pxt-microbit/issues/925

This PR fixes three issues:
1. Enum arguments were always being replaced as 0 or null (now we grab the first declared member)
2. Fixed instances of types with qualified names exported from a namespace (e.g. `input.Button` rather than top-level `Button`) were generating weird snippets with the namespace showing up twice

3. The Monaco toolbox was ignoring default values for variable blocks. This was affecting a few of the changes I recently made in pxt-microbit for the image and game blocks

### DIFF
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -874,22 +874,7 @@ namespace ts.pxtc.service {
                 case SK.BooleanKeyword: return "false";
                 case SK.ArrayType: return "[]";
                 case SK.TypeReference:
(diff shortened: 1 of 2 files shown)


riknoll:
BTW I tested every snippet in the micro:bit toolbox and this should fix them all. We still have an issue where object types just get `null` for placeholder values but that's a separate feature.

PR #4600, opened 2018-08-16
 Add 'pxt dumplog' to dump codal's log

mmoskal:
Also:
* use a pipe instead of a socket to invoke gdb; now tested on mac and win
* allow setting `PXT_NODOCKER` to force non-docker build
* bugfix for compilation failure in Adafruit (neopixel init)


mmoskal:
@guillaumejenkins could you take another look? some changes added

PR #4601, opened 2018-08-16
Move rounding operations to their own block

guillaumejenkins:
Fixes https://github.com/Microsoft/pxt-microbit/issues/1014

### DIFF
--- a/pxtlib/blocks.ts
+++ b/pxtlib/blocks.ts
@@ -6,10 +6,13 @@ namespace pxt.blocks {
     // The JS Math functions supported in the blocks. The order of this array
     // determines the order of the dropdown in the math_js_op block
     export const MATH_FUNCTIONS = {
-        unary: ["sqrt", "sin", "cos", "tan", "round", "ceil", "floor", "trunc"],
+        unary: ["sqrt", "sin", "cos", "tan"],
         binary: ["atan2"],
         infix: ["idiv", "imul"]
     };
+
+    // Like MATH_FUNCTIONS, but used only for rounding operations
+    export const ROUNDING_FUNCTIONS = ["round", "ceil", "floor", "trunc"];
 
     export interface BlockParameter {
         // Declared parameter name as it appears in the code. This is the name used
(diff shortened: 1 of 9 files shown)


pxtlib/blocks.ts
```diff
@@ -407,12 +406,28 @@ namespace pxt.blocks {
...
                     "tan": Util.lf("{id:op}tan"),
                     "atan2": Util.lf("{id:op}atan2"),
+                    "idiv": Util.lf("{id:op}integer ÷"),
+                    "imul": Util.lf("{id:op}integer ×"),
+                }
+            },
+            "math_js_round": {
+                name: Util.lf("rounding functions"),
+                tooltip: {
+                    "round": Util.lf("Round the argument half up"),
```
riknoll: @ganicke any thoughts on how to describe these? Seems weird to use the word "round" for all of them but I can't think of anything better.

guillaumejenkins: These were taken from Wikipedia, and I think they are very simple to understand for kids (except maybe the "half up" part). I've seen "round up" or "round down" several times before; isn't it a valid thing to say?

image

ganicke: Sure. Further explanation could happen in docs. Except half away and zero mean what? Zero the number or zero the fraction?

FranklinWhale: The proposed tooltip of `trunc` by @ganicke seems perfect. I think "round up" and "round down" are generally good enough, but if something more descriptive is preferred, I think the descriptions from the .NET docs are good.

I also try to improve the tooltip of `round` below.

Method | Tooltips
------ | -----------
`trunc` | Remove the fractional part of the argument
`ceil` | Round the argument up to an integer or [Return the smallest integral value greater than or equal to argument](https://docs.microsoft.com/en-us/dotnet/api/system.math.ceiling?view=netstandard-2.0)
`floor` | Round the argument down to an integer or [Return the largest integral value less than or equal to the argument](https://docs.microsoft.com/en-us/dotnet/api/system.math.floor?view=netstandard-2.0)
`round` | Round the argument to the nearest integer. If its fractional part is exactly 0.5, round it up.

FranklinWhale: One additional question: Should we use "the specified number" instead of "the argument" to refer to the argument in the tooltips?

riknoll: I think "argument" is more consistent with our other tooltips, but I don't have strong feelings either way.

guillaumejenkins: In the end we went with @ganicke's tooltips, thanks for all the suggestions

PR #4611, opened 2018-08-21
Proposal for compiler changes

mmoskal:
Writing it down to clear my thinking and putting it out for comments

Can read it with nicer formatting here: https://github.com/Microsoft/pxt/blob/morejavascript/docs/morejavascript.md


mmoskal:
This is now all implemented.

PR #4614, opened 2018-08-21
Fixing some bugs with iframe controller scenario and adding settracestate

riknoll:
Adding a pxt controller command for setting the trace state in the editor (the only one that was surfaced was a toggle).

Also fixes a race condition with binding the editor messages and a bug where any target that had `allowParentController` set to true can only be served in an iFrame. Previously you could serve both in and out of an iFrame but the new iFrame workspace provider blocks on a message from the host.

I probably should have done this in separate commits...

@dragonx


samelhusseini:
The call to getEditorAsync makes the entire request / response pair async which breaks a lot of scenarios, basically any scenario that is async and expects a response.

PR #4615, opened 2018-08-21
Allow any two types to be compared in blocks

riknoll:
Fixes https://github.com/Microsoft/pxt-microbit/issues/1050

Removes the type bumping behavior of the Blockly comparison block.

Where once there was this:
old_comparison

Now there is this:

new_comparison

The weird outline bug is unrelated


riknoll:
Fix for weird outline bug: https://github.com/Microsoft/pxt-blockly/pull/113

PR #4628, opened 2018-08-27
Ext search spinner

guillaumejenkins:
Adds a spinner for search. This is especially important with our backend now indexing and caching packages, since it can take up to 8 seconds to perform a search (for the unlucky user that hits our backend cache reset, happens once every 5min).

Current experience:
currentsearch

With this change:
searchspinner

Some notes:
- With this change, the search bar and button are disabled while a search is in progress
- For both the current experience and this change, subsequent searches are very fast

Fixes https://github.com/Microsoft/pxt-microbit/issues/1121

### DIFF
--- a/webapp/src/data.ts
+++ b/webapp/src/data.ts
@@ -17,6 +17,13 @@ interface CacheEntry {
     api: VirtualApi;
 }
 
+export enum FetchStatus { Pending, Complete, Error, Offline };
+
+export interface DataFetchResult<T> {
+    data?: T;
+    status: FetchStatus;
+}
+
 const virtualApis: pxt.Map<VirtualApi> = {}
 let targetConfig: pxt.TargetConfig = undefined;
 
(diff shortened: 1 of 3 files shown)


webapp/src/data.ts
```diff
@@ -17,6 +17,13 @@ interface CacheEntry {
     api: VirtualApi;
 }
 
+export enum FetchStatus { Pending, Complete, Error, Offline };
```
riknoll: Nitpick: multiple lines

webapp/src/scriptsearch.tsx
```diff
@@ -269,6 +280,10 @@ export class ScriptSearch extends data.Component<ISettingsProps, ScriptSearchSta
         const urldata = this.fetchUrlData();
         const local = this.fetchLocal()
 
+        const isSearchingGh = ghdata.status === data.FetchStatus.Pending || ghdata.status === data.FetchStatus.Error;
```
riknoll: So if it errors is it just going to show the spinner forever?

guillaumejenkins: Well, in `data.ts`, errors are re-queued immediately, so that's why I made it that way... but that means if the request keeps getting errors, it will indeed spin forever.

I guess I can change errors to mean "request completed", and if the re-queued request happens to succeed, then the results will be silently added to the search results. Not a big deal, since transient request errors should be very rare

PR #4630, opened 2018-08-27
Make extension errors more visible

riknoll:
If an extension you add to your project contains errors it's pretty difficult to determine where the error preventing compilation is coming from. The current experience just gives a toast that says "compilation failed, please check your code for errors", which is at best unhelpful and at worst misleading (if the errors aren't actually in your code). The only way to determine that a package contains errors is to switch to JavaScript and expand the file explorer.

This PR adds a new dialog that indicates when the errors are not your fault:

ext_dialog_final

Works for both Github packages and for shared URLs added as extensions. The dialog shows when you load a project, attempt to download/save, or attempt to start the simulator.

Currently the "Learn More" link doesn't lead anywhere (hence the DO NOT MERGE). @ganicke @jaustin is there a page describing the migration from v0 to master that I should be linking to?

### DIFF
--- a/webapp/src/dialogs.tsx
+++ b/webapp/src/dialogs.tsx
@@ -4,6 +4,7 @@ import * as ReactDOM from "react-dom";
 import * as sui from "./sui";
 import * as data from "./data";
 import * as core from "./core";
+import * as coretsx from "./coretsx";
 import * as cloudsync from "./cloudsync";
 
 import Cloud = pxt.Cloud;
@@ -194,7 +195,7 @@ function renderVersionLink(name: string, version: string, url: string) {
     </p>;
 }
 
-export function showPackageErrorDialogAsync(badPackages: EditorPackage[]) {
+export function showPackageErrorDialogAsync(badPackages: EditorPackage[], removePackage: (id: string) => Promise<void>) {
     core.confirmAsync({
         header: lf("Extension Errors"),
         hideCancel: true,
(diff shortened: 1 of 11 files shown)


riknoll:
Unfortunately, I don't see any way to detect what editor version a project was created in. We don't have that info in pxt.json

ganicke:
There is no page about migrating from `v0` to `v1` other than a note about it in the beta blog:

https://makecode.com/blog/microbit/v1-beta#moving-programs-around

We could add one to support the "Learn more".

webapp/src/dialogs.tsx
```diff
@@ -193,7 +195,134 @@ function renderVersionLink(name: string, version: string, url: string) {
...
+        return () => {
+            coretsx.hideDialog();
+            return removePackage(id)
+        }
+    }
+}
+
+export function renderEditorVersionMessage(onClick: () => void) {
+    return <div className="ui message">
+        <p>{pxt.Util.lf("This project was made in an older version of the editor and may not be compatible with the latest version. To open the project in the old editor,")} <a onClick={onClick}>{pxt.Util.lf("Click here")}.</a></p>
```
guillaumejenkins: `Click here` -> `click here`?

pelikhan: Concatenating sentence does not always translate properly. Can we have a button instead?

riknoll: The new text is two sentences:

> This project was made in an older version of the editor and may not be compatible with the latest version. Open the project in the old editor


The entire second sentence will be a link, so no string concatenation. @pelikhan is that better for localizing?

riknoll:
ext_errors_again


Now with:

* "Open in old editor"
* "Update All"
* Doesn't show when project is opened, only when downloading or playing

Would "refresh all" perhaps be better? "Update all" implies that an update exists when it might not.

pelikhan:
Overall this is a complicated dialog for kids regardless. We need a help page.

I think we need to discuss the entire flow in a design meeting. What happens when the update fails or introduces errors? Do we create a backup?

riknoll:
@pelikhan in lieu of everyone being around I created this PR as the design review; no worries, I won't merge it until everyone gets a say :smile:

riknoll:
@guillaumejenkins we discussed opening by default and decided that when you click on a project you want to see the project and not an error dialog. Maybe we could show it on first edit?

riknoll:
Closing for a redesign

PR #4635, opened 2018-08-30
Adding back button to the sidedocs

riknoll:
Fixes https://github.com/Microsoft/pxt-microbit/issues/954

sidedocs_back

### DIFF
--- a/theme/docs.less
+++ b/theme/docs.less
@@ -120,6 +120,15 @@
     .ui.primary.button p {
         color: white;
     }
+
+    #sidedocs-back-button {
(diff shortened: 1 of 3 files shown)


theme/docs.less
```diff
@@ -120,6 +120,15 @@
     .ui.primary.button p {
         color: white;
     }
+
+    #sidedocs-back-button {
+        margin-top: 1.25rem;
+        margin-left: 1.25rem;
```
guillaumejenkins: `cursor: pointer;` maybe? From your gif it looks like currently the cursor remains as `text`

PR #4647, opened 2018-09-06
Don't show monaco for package files

pelikhan:
With this change, the monaco toolbox only shows up for main.ts.

monacospecific


pelikhan:
@abchatra says it's fine.

PR #4651, opened 2018-09-06
adding boiler places impl for indexOf, lastIndexOf, includes

pelikhan:
Add some missing string function. Merge after the common packages merge in microbit.
FIx for https://github.com/Microsoft/pxt-microbit/issues/1147
Requires https://github.com/Microsoft/pxt-common-packages/pull/411


pxtsim/libgeneric.ts
```diff
@@ -394,6 +394,24 @@ namespace pxsim {
...
+        export function lastIndexOf(s: string, searchValue: string, start?: number) {
+            pxtrt.nullCheck(s);
+            pxtrt.nullCheck(searchValue);
+            return s.lastIndexOf(searchValue, start);
+        }
+
+        export function includes(s: string, searchValue: string, start?: number) {
+            pxtrt.nullCheck(s);
+            pxtrt.nullCheck(searchValue);
+            return s.includes(searchValue, start);
```
mmoskal: I think these return false when you pass null as second argument

docs/static/playground/pxt-common/pxt-core.d.js
```diff
@@ -237,6 +237,37 @@ declare interface String {
...
+
+    /**
+     * Returns the position of the last occurrence of a specified value in a string.
+     * @param searchValue the text to find
+     * @param start optional start index for the search
+     */
+    //% shim=String_::lastIndexOf
+    //% help=text/last-index-of
+    //% blockId="string_lastindexof" blockNamespace="text"
+    lastIndexOf(searchValue: string, start?: number): number;
```
mmoskal: Doesn't this need to go away?

PR #4660, opened 2018-09-07
Codal listener

tballmsft:
Changes to enable multiple background handlers for an event (simulator). Doesn't affect behavior for user handlers (last handler added wins).  Test program (requires minor pxt-microbit changes) at https://makecode.microbit.org/_M499UaitjCT3


tballmsft:
@riknoll - can you please take a look?

pxtsim/runtime.ts
```diff
@@ -223,7 +223,7 @@ namespace pxsim {
                     aws.forEach(aw => aw());
                 }
             }
-            if (!this.handler || this.events.length > this.max) return;
+            if (this.handlers == [] || this.events.length > this.max) return;
```
mmoskal: should be `this.handlers.length == 0`; `==` is probably ref-equality

pxtsim/runtime.ts
```diff
@@ -245,21 +246,30 @@ namespace pxsim {
                         this.lock = false;
```
mmoskal: You're releasing the lock multiple times it seems

pxtsim/runtime.ts
```diff
@@ -245,21 +246,30 @@ namespace pxsim {
...
-                pxtcore.decr(this._handler);
-            }
+        addHandler(a: RefAction) {
+            this._handlers.push(a);
+            pxtcore.incr(a)
+        }
 
-            this._handler = a;
+        setHandler(a: RefAction) {
+            this._handlers.forEach(old => pxtcore.decr(old))
```
mmoskal: This seems to remove all old handlers, not only the foreground one

pxtsim/simlib.ts
```diff
@@ -37,34 +43,56 @@ namespace pxsim {
...
 
         queue(id: number | string, evid: number | string, value: T = null) {
             // special handling for notify one
             const notifyOne = this.notifyID && this.notifyOneID && id == this.notifyOneID;
             if (notifyOne)
                 id = this.notifyID;
-
-            // grab queue and handle
+            // first, background
+            this.backgroundHandlerFlag = true;
```
mmoskal: This flag seems to be used for two different purposes - for setting handlers and for running them. Two flags would be better, especially when handlers register new handlers.

pxtsim/runtime.ts
tballmsft: thanks.

pxtsim/runtime.ts
tballmsft: Good catch.

tballmsft: but setHandler only called for foreground.

mmoskal: Oh, I think I misunderstood.

pxtsim/simlib.ts
```diff
@@ -37,34 +43,56 @@ namespace pxsim {
...
+            else
+                q.setHandler(handler);
+            this.backgroundHandlerFlag = false;
+        }
+
+        // only for background handlers
+        remove(handler: RefAction) {
+            this.backgroundHandlerFlag = false;
+            Object.keys(this.queues).forEach((k: string) => {
+                if (k.startsWith("back"))
```
mmoskal: should be `"back:"` ?

pelikhan:
Any docs?

tballmsft:
Not yet, this is just plumbing. No user facing APIs surfaced here. That will come in common-packages.

PR #4662, opened 2018-09-10
Fix decompile for tutorial snippets with variables

riknoll:


riknoll:
@pelikhan no, we send new compile options on ever compile request

PR #4670, opened 2018-09-12
don't reverse cards when rendering links

pelikhan:
See http://localhost:3232/reference/radio - cards are in order again.


riknoll:
I figured it was a css thing

PR #4671, opened 2018-09-12
code-gen re-work

mmoskal:
Requires bump of PXT to 4.0.0 - this is not to be integrated in Chibitronics

* drop support for integer-only compilation, and thus also the two AVR backends
* drop support the C# backend (could be resurrected if needed)
* JavaScript reference counting is disabled (at least for now)
* generics used to be compiled for every (used) combination of ref/non-ref parameters; now all parameters are reference counted, so generics are compiled like everything else and can be used as first class values etc
* altogether ~2100 lines of code are gone
* the Thumb code generation now uses more structured scheme, where there is argument computation phase, conversion phase (eg our numbers to ints), call, and cleanup; conversion and cleanup are done using lazily-generated code fragments bringing code size down 20%
* the `commonalize` phase of peep-hole optimization is removed - it was trying hunt down these code fragments, but now they are handled automatically to better effect and without interfering with breakpoints
* the conversion phase can be augmented in future with more dynamic type checks as needed
* the caller, not callee, now cleans up procedure (TS function, as opposed to C++ function) arguments, which are now in reverse order; this means that you can override methods with methods with less arguments
* arguments are now properly converted to string when C++ requires this; fixes eg `[1,2,3].join(",")`
* after every call, the stack is in a "safe" state, where it can be unwound by decrementing everything on the stack by 1 (this is in prep for exceptions)

This requires a related PR in common packages - https://github.com/Microsoft/pxt-common-packages/pull/413


mmoskal:
I guess will happen with the next release of cpx/microbit/ev3/...

riknoll:
Yay 4.0 :tada:

PR #4672, opened 2018-09-13
Ask for nested directories in github; fixes #4667

### LINKED ISSUE #4667: Subdirectory files destroyed with github editing
When editing the pxt-neopixel package, which contains package localization files, we lost all the translations. These files live under _locales/LOCALE/...

This is due to https://github.com/Microsoft/pxt/blob/6e006d120842b2bb8d77818cae16d69438746755/webapp/src/workspace.ts#L531 which deletes all nested files.

mmoskal:


pelikhan:
Don’t you have to patch the call where you filter out files?

mmoskal:
No. You can only edit files that are in `pxt.json` (and in fact the locale files are there). Other files are just ignored - PXT doesn't touch them.

PR #4675, opened 2018-09-13
reduce screenshot size until reasonably small image is generated

pelikhan:
When generating a block screenshot, reduce size of screenshot until it gets < 1Mb.
Fix for https://github.com/Microsoft/pxt-microbit/issues/1203


riknoll:
How big of screenshots were we generating?

pelikhan:
4000x8000

PR #4676, opened 2018-09-13
Various fixes for being embedded inside WKWebview

guillaumejenkins:
- Fixed parsing of UserAgent and browser version to enable WKWebview on Mac OS desktop
- Removed redundant `sandbox` attribute assignment on our iframes, which was illegal in WKWebview
  - **Note:** I tested this in IE, Edge, Chrome, FF and iOS Safari, and the Sim iframe still works fine
- Sim no longer stops / resumes on visibility change if we're in an app

### DIFF
--- a/pxtlib/browserutils.ts
+++ b/pxtlib/browserutils.ts
@@ -154,11 +154,12 @@ namespace pxt.BrowserUtils {
         }
         else if (isSafari()) {
             matches = /Version\/([0-9\.]+)/i.exec(navigator.userAgent);
-            // pinned web site have a different user agent
+            // pinned web sites and WKWebview for embedded browsers have a different user agent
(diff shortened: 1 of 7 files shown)


pelikhan:
Is the sandbox thingy impacting any browser at all?

pxtlib/browserutils.ts
```diff
@@ -81,7 +81,7 @@ namespace pxt.BrowserUtils {
     export function isSafari(): boolean {
         //Could also check isMac but I don't want to risk excluding iOS
         //Checking for iPhone, iPod or iPad as well as Safari in order to detect home screen browsers on iOS
-        return !isChrome() && !isEdge() && !!navigator && /(Safari|iPod|iPhone|iPad)/i.test(navigator.userAgent);
+        return !isChrome() && !isEdge() && !!navigator && /(Macintosh|Safari|iPod|iPhone|iPad)/i.test(navigator.userAgent);
```
riknoll: "Hey did you see the latest Macintosh phone?"

guillaumejenkins: Still a better name than Xr Max

PR #4689, opened 2018-09-17
Fixes for docs in staticpkg

mmoskal:
The `--route` needs to be `/` not `./` as it was to get paths to css right

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -2148,41 +2148,45 @@ function renderDocs(builtPackaged: string, localDir: string) {
     docsTemplate = U.replaceAll(docsTemplate, "/--embed", webpath + "embed.js")
 
     const dirs: Map<boolean> = {}
-    for (const f of nodeutil.allFiles("docs", 8)) {
-        pxt.log(`rendering ${f}`)
-        let dd = path.join(dst, f)
-        let dir = path.dirname(dd)
-        if (!U.lookup(dirs, dir)) {
-            nodeutil.mkdirP(dir)
-            dirs[dir] = true
-        }
-        let buf = fs.readFileSync(f)
-        if (/\.(md|html)$/.test(f)) {
-            let str = buf.toString("utf8")
-            if (/\.md$/.test(f)) {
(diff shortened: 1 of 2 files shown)


mmoskal:
CC @dwaw18

cli/cli.ts
```diff
@@ -3853,7 +3857,7 @@ function stringifyTranslations(strings: pxt.Map<string>): string {
 }
 
 export function staticpkgAsync(parsed: commandParser.ParsedCommand) {
-    const route = parsed.flags["route"] as string || ".";
+    const route = parsed.flags["route"] as string || "/";
```
FranklinWhale: @mmoskal @pelikhan:
According to https://makecode.com/cli/staticpkg, `route`, if missing, defaults to `local`. My understanding of `local` is that it refers to the "current directory", which means that if the output files are put on https://www.example.com/makecode/, all relative paths will be based on https://www.example.com/makecode/. If the value is changed to `/`, the base URL will be changed to https://www.example.com/. Is that intended?

FranklinWhale: Um... Perhaps my understanding was incorrect, probably misled by the previous value `.`. The change to `/` makes the behavior consistent with that described in "Deploying PXT with static files".

I think both `./` in Line 3879 and the `local` on https://makecode.com/cli/staticpkg should be changed to `/`.

Have a question though: Should the use `route` be deprecated in favor of the generation of paths relative to the "current directory"? The advantage is that, when the routing is changed in the future, no re-compilation (or packaging) will be required.

pelikhan:
@guillaumejenkins could you check that this does not break electron?

mmoskal:
You can always do `--route .`

The current default doesn't work - the paths to CSS files are relative and don't work under `docs/`

FranklinWhale:
A quick comment as I am using mobile and don't have the packaged source:

If the css folder and the docs folder are under the same directory, perhaps using `../css/` to access the css files is a better and more compatible fix.

mmoskal:
In general it's more tricky to get the relative paths to work, especially given that many of them come from various random HTML templates. PRs are welcome though.

PR #4691, opened 2018-09-17
Use globs for file list

mmoskal:
Anyone knows why we were not just doing `built/*.js` and `built/*.d.ts` and instead listed files?


pelikhan:
I don't recall any adverse reason.

PR #4692, opened 2018-09-17
Working around some browser exceptions in IE that break tutorials

riknoll:
Fixes https://github.com/Microsoft/pxt-microbit/issues/1249 (kinda)

Fixing a couple of exceptions for IE in tutorials. The first one was caused by a change I made to fix screenshots in Edge. IE throws an exception but it can be safely ignored.

The second bug causes the site to freeze when exiting a tutorial and loading the editor. The browser crashes on a call to `Date.now()` saying that `Date` is undefined. @abchatra and I spent a while debugging but it sure looks like this is an IE bug. Instead of devoting more time to a proper fix we decided to just avoid the codepath in IE and go to the home screen instead of opening the editor.

### DIFF
--- a/pxtblocks/blocklylayout.ts
+++ b/pxtblocks/blocklylayout.ts
@@ -119,7 +119,14 @@ namespace pxt.blocks.layout {
             cvs.width = width * pixelDensity;
             cvs.height = height * pixelDensity;
             img.onload = function () {
-                ctx.drawImage(img, 0, 0, width, height, 0, 0, cvs.width, cvs.height);
+                try {
+                    ctx.drawImage(img, 0, 0, width, height, 0, 0, cvs.width, cvs.height);
+                }
+                catch (e) {
+                    // IE 11 throws an exception
+                    reject(e);
+                    return;
+                }
                 let canvasdata = cvs.toDataURL("image/png");
                 // if the generated image is too big, shrink image
                 while (canvasdata.length > MAX_SCREENSHOT_SIZE) {
@@ -262,6 +269,7 @@ namespace pxt.blocks.layout {
                         imageIconCache[svgUri] = href;
                         image.setAttributeNS(XLINK_NAMESPACE, "href", href);
                     })
+                    .catch(() => { /* Ignore conversion errors from IE */ })
             });
         return Promise.all(p).then(() => { })
     }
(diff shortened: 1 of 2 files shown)


pxtblocks/blocklylayout.ts
```diff
@@ -119,7 +119,14 @@ namespace pxt.blocks.layout {
             cvs.width = width * pixelDensity;
             cvs.height = height * pixelDensity;
             img.onload = function () {
-                ctx.drawImage(img, 0, 0, width, height, 0, 0, cvs.width, cvs.height);
+                try {
+                    ctx.drawImage(img, 0, 0, width, height, 0, 0, cvs.width, cvs.height);
```
pelikhan: is this due to incorrect SVG or security?

pxtblocks/blocklylayout.ts
```diff
@@ -119,7 +119,14 @@ namespace pxt.blocks.layout {
...
+                }
+                catch (e) {
+                    // IE 11 throws an exception
+                    reject(e);
```
pelikhan: log a tick event

webapp/src/app.tsx
```diff
@@ -2111,13 +2111,23 @@ export class ProjectView
...
         core.showLoading("leavingtutorial", lf("leaving tutorial..."));
-        this.exitTutorialAsync()
-            .then(() => {
-                let curr = pkg.mainEditorPkg().header;
-                return this.loadHeaderAsync(curr);
-            }).done(() => {
-                core.hideLoading("leavingtutorial");
-            })
+
+        if (pxt.BrowserUtils.isIE()) {
```
pelikhan: we should consider have a single behavior for all browser. e.g. exit tutorial takes you to the home screen

pelikhan:
Would be worth discussing if we shouldn't just always go back to the home screen.

pxtblocks/blocklylayout.ts
riknoll: It's a browser exception (`IndexSizeError`). I've updated the PR to avoid this codepath in IE entirely since this fix was only meant for "download screenshot" and there's no point in catching it.

webapp/src/app.tsx
riknoll: Add it to the UI meeting discussion list :smile:. I also prefer going straight to the home screen

PR #4695, opened 2018-09-19
Green screen update.

pelikhan:
- [x] Edge
- [x] FF
- [x] Chrome
- [x] IE (disable)
- [x] iOS fix (tested on iOS12)

```
On iPhone, <video playsinline> elements will now be allowed to play inline, and will not automatically enter fullscreen mode when playback begins.
<video> elements without playsinline attributes will continue to require fullscreen mode for playback on iPhone.
When exiting fullscreen with a pinch gesture, <video> elements without playsinline will continue to play inline.
```


pelikhan:
@riknoll @abchatra iOS fixed

PR #4699, opened 2018-09-20
Wide log and light TOC theme

pelikhan:
Required to support LEGO TOC. Probably requires refreshing pxt-backend bits

https://github.com/Microsoft/pxt-ev3/pull/786

### DIFF
--- a/theme/docs.less
+++ b/theme/docs.less
@@ -103,12 +103,21 @@
     }
     /* TOC */
     .toc a.header {
+        color: black;
+        text-transform: uppercase;
+    }
+    .toc .inverted a.header {
         color: white;
     }
     .toc a.item {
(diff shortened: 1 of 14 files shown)


pelikhan:
@guillaumejenkins this change will need to be pushed to the cloud

theme/docs.less
```diff
@@ -102,16 +102,32 @@
...
+    .toc .inverted a.item {
+        color: rgba(255, 255, 255, 1);
+    }
+
     .toc a.item:hover {
         text-decoration: @linkHoverUnderline;
     }
+
+    .toc .inverted .item.search {
+        margin: -.92857143em -1.14285714em;
```
riknoll: Magic numbers?

pelikhan: They were already in there.

PR #4702, opened 2018-09-20
Always load translations

pelikhan:
Docs pages only load sim strings which breaks translations of snippets. Simplify logic. FIx for https://github.com/Microsoft/pxt-ev3/issues/765


pelikhan:
We should talk about bundle support in the api.

PR #4708, opened 2018-09-21
use same domain /api for translations / caching

pelikhan:
- [x] Hit ``/api/translations...`` from client to avoid CORS issues. Also fix "ETag" -> "etag".
- [x] store translations in indexeddb to share them with editor. Translations are stored in PouchDB which is not available in docs.

The translations are stored in a db, along with the ETag. There is a 5 minute expiration in the db after which the ETag might be tested again. ETag is only checked once per browser sessions (once per 24h) to avoid requesting it multiple times during a single sesion.

This should not impact the cloud.


abchatra:
We will pick the new 4.0 for microbit. So no need to port to 3.22. 3.22 is a backup incase we have ton of issues in 4.0, then we can backport.

PR #4710, opened 2018-09-21
Markdown caching

pelikhan:
Uses translation database to cache Markdown sources in indexeddb. This is shared between editor and sidedocs.
- [x] use etag
- [x] background every refresh at most every hour

### DIFF
--- a/pxtlib/emitter/cloud.ts
+++ b/pxtlib/emitter/cloud.ts
@@ -89,7 +89,8 @@ namespace pxt.Cloud {
         })
     }
 
-    const MARKDOWN_EXPIRATION = 60 * 1000;
+    // 1h check on markdown content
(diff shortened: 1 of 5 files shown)


pxtlib/emitter/cloud.ts
```diff
@@ -92,21 +92,22 @@ namespace pxt.Cloud {
     // 1h check on markdown content
     const MARKDOWN_EXPIRATION = 1 * 60 * 60 * 1000;
     export function markdownAsync(docid: string, locale?: string, live?: boolean): Promise<string> {
+        const branch = pxt.appTarget.versions && pxt.appTarget.versions.target || '?';
```
mmoskal: Markdown is not versioned by targetVersion - it's independent, so I'm not sure this helps much

pelikhan: Removed.

PR #4713, opened 2018-09-21
Hit cloud for gh endpoints on localhost

riknoll:
https://github.com/Microsoft/pxt/pull/4708 broke some requests on local serve because we never implemented the endpoints in the development server. Rather than duplicate them in PXT, hit the cloud instead.

### DIFF
--- a/pxtlib/emitter/cloud.ts
+++ b/pxtlib/emitter/cloud.ts
@@ -2,7 +2,7 @@ namespace pxt.Cloud {
     import Util = pxtc.Util;
 
     // hit /api/ to stay on same domain and avoid CORS
-    export let apiRoot = "/api/";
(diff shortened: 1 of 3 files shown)


mmoskal:
Maybe it would be better to set `apiRoot` to `https://makecode.com/api/` on localhost and to `/api/` otherwise? It would also simplify this: `${pxt.Cloud.isLocalHost() ? "https://makecode.com" : ""}/api/translations?lang=...`

riknoll:
But wouldn't that hit the cloud for everything?

mmoskal:
You generally want to do that, since there is very few APIs implemented in the local server. I think that was the behavior before that PR. There are some localhost-specific APIs, but these are accessed not via `apiRoot`.

riknoll:
Ah, I see that now. I was worried about breaking the file workspace but you're right. Updated!

PR #4719, opened 2018-09-24
Update the dropdown selected checkmark from black to white

samelhusseini:
Update the dropdown selected checkmark from black to white to contrast better with coloured dropdowns. 

Here's what it looked like previously: 
screen shot 2018-09-21 at 4 55 23 am

and now: 
screen shot 2018-09-21 at 4 53 03 am

Note: SVG is not used directly and only exported to PNG (no need to optimize)

Fixes https://github.com/Microsoft/pxt-microbit/issues/1276
pxt-blockly PR: https://github.com/Microsoft/pxt-blockly/pull/118


pelikhan:
Can we cleanout the sprites.svg?

samelhusseini:
How so? just like remove the inkscape meta data?
As mentioned in my note, SVG is not used directly and only exported to PNG so I didn't feel the need to optimize size.

pelikhan:
Oops missed the note. Disregard.

PR #4720, opened 2018-09-24
Fix serve and staticpkg

guillaumejenkins:
Right now, `pxt serve` and `pxt staticpkg` are broken for all targets

### DIFF
--- a/pxtlib/util.ts
+++ b/pxtlib/util.ts
@@ -446,7 +446,8 @@ namespace ts.pxtc.Util {
         return a + b;
     }
 
-    export let isNodeJS = false;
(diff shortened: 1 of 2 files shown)


pxtlib/util.ts
```diff
@@ -446,7 +446,8 @@ namespace ts.pxtc.Util {
         return a + b;
     }
 
-    export let isNodeJS = false;
+    // Reliable NodeJS detection is not possible, but the following check should be accurate enough for our needs
+    export let isNodeJS = typeof window === "undefined";
```
pelikhan: Maybe rename to headless

PR #4723, opened 2018-09-24
Try and fix extension errors

riknoll:
Adds a dialog that appears when projects have errors in their dependencies. The user can either ignore them, go to the old editor, or try and fix them. Clicking the X returns you to the home screen.

"Go to old editor" works by duplicating the project from the current editor's workspace to the old editor's workspace. Then, I push a file history entry for the project into localstorage and redirect to the old editor based on the version in the project. The page reloads and drops the user into the editor with the project opened. I rely on the browser prefix stuff so this option only appears if the project was created in the old editor and you are in a browser (no apps or local serve, sorry).

"Try to fix" is more complicated. I create a backup of the project in the pouchdb and save it. Then I update all of the dependencies and check if the project compiles. If it doesn't, I restore the backup and the project is unchanged. The backup logic is such that if the operation is interrupted (eg reloading the page), the backup will be restored next time the project is loaded. Orphaned backups get cleaned up on editor load and no backups show up on the homescreen so it should be transparent to the user.

You can try it here:

https://makecode.microbit.org/app/a36ed3e373d714f00fe05a896ebdd80ddde5d5de-8715ee92cd

And here are some test projects:

[upgrade-tests.zip](https://github.com/Microsoft/pxt/files/2413201/upgrade-tests.zip)


riknoll:
I reversed the filenames in my linked test cases but am too lazy to re-upload.

Here's the success UI:

update_success

And here's the failure UI:

update_fail

theme/common.less
```diff
@@ -538,6 +538,37 @@ Field editors
     background-color: #00aced;
 }
 
+/*******************************
```
samelhusseini: Mind adding this in its own less file.

pelikhan:
Would be good to document this in our versioning docs @ganicke ?

theme/common.less
riknoll: Done!

PR #4728, opened 2018-09-25
Migrating docs from "Package" to "Extension"

pelikhan:
Started migrating the package terminology into extension in the docs. @ganicke can you do another sweep?

- [x] add redirects to all previous page. Specifically, the fondation points to the "packages/..." page in various places.


ganicke:
I ran through the other developer docs and made the changes to 'extensions'.

PR #4735, opened 2018-09-26
idbworkspace

guillaumejenkins:
- Adds an IDBWrapper class, which is a light database that uses IndexedDB directly
- Adds a new idbworkspace that uses the IDBWrapper to store projects
    - This workspace is enabled only on Safari when ipcRenderer is present
- Refactors the translations database to use the IDBWrapper


pelikhan:
What happens when any of those operation fail?

webapp/src/idbworkspace.ts
```diff
@@ -0,0 +1,103 @@
...
+const KEYPATH = "id";
+
+let _db: pxtc.Util.IDBWrapper;
+
+function getDbAsync(): Promise<pxtc.Util.IDBWrapper> {
+    if (_db) {
+        return Promise.resolve(_db);
+    }
+
+    _db = new pxtc.Util.IDBWrapper("__pxt_idb_workspace", 1, (ev, r) => {
```
pelikhan: does "Reset" delete this table?

guillaumejenkins: Reset only clears the stores (texts and headers), the database and both stores will all still exist (but the stores will be empty)

webapp/src/workspace.ts
```diff
@@ -607,6 +609,14 @@ export function syncAsync(): Promise<pxt.editor.EditorSyncState> {
     checkSession();
 
     return impl.listAsync()
+        .catch((e) => {
+            // There might be a problem with the native databases. Switch to memory for this session so the user can at
```
pelikhan: this kind of crashes can happen at any time and this might be too late now? is listAsync() called on load only?

guillaumejenkins: It's called on init, and when going back to home page

I can add a flag `isInit` and only do the workspace switch if we're during initialization

pelikhan: Yes then your fix is good. We should add a notification banner in the future to explain what's going on.

PR #4736, opened 2018-09-26
Add globe icon for language picker

abchatra:
- In home screen language settings need an icon (othewise non-native speaker has no way of knowing where to change the language).
- Deleted old makecode.com site contents.


pelikhan:
Great minds think alike https://github.com/Microsoft/pxt/commit/7273049a026fe06b5387f2ff55a7cb3f796daf63

abchatra:
That is too much of coincidence.

PR #4742, opened 2018-09-26
Extension version upgrade note

ganicke:
A start on adding a note for extension upgrades as related to #4723.

Not really sure what the total scenarios are here so comments welcome.


riknoll:
The scenario the dialog addresses is when there are errors detected in the extension's code and not the user's code (ie errors that are not the user's fault). The "open in old editor" option also only shows up if the project was created in the old editor.

pelikhan:
Is this ready?

ganicke:
Was waiting for 👍 from @riknoll

PR #4747, opened 2018-09-28
Enable support for hiding partner logo from the menubar while coding

guillaumejenkins:
Required for Minecraft


pelikhan:
Aren't you hiding it only when the editor is hosted in the iframe?

PR #4759, opened 2018-10-02
fixed main.cpp

pelikhan:
After discussion with Joe, i have changed the main.cpp template to better use fiber resources + logging various intermediate build files to console.debug

### DIFF
--- a/pxtlib/cpp.ts
+++ b/pxtlib/cpp.ts
@@ -801,6 +801,7 @@ namespace pxt.cpp {
                 cfg[k] = v
             })
             res.generatedFiles["/codal.json"] = JSON.stringify(codalJson, null, 4) + "\n"
+            pxt.debug(`codal.json: ${res.generatedFiles["/codal.json"]}`);
         } else if (isPlatformio) {
             const iniLines = compileService.platformioIni.slice()
             // TODO merge configjson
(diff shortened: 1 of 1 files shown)


pxtlib/cpp.ts
```diff
@@ -836,8 +838,11 @@ namespace pxt.cpp {
...
         res.generatedFiles[sourcePath + "pxtconfig.h"] = pxtConfig
-        if (isYotta)
+        pxt.debug(`pxtconfig.h: ${res.generatedFiles[sourcePath + "pxtconfig.h"]}`)
+        if (isYotta) {
             res.generatedFiles["/config.json"] = JSON.stringify(configJson, null, 4) + "\n"
+            pxt.debug(`yotta config.json: ${res.generatedFiles["/config.json"]}`)
+        }
         res.generatedFiles[sourcePath + "main.cpp"] = `
 #include "pxt.h"
 #ifdef PXT_MAIN
```
mmoskal: I'm not sure this is actually used: https://github.com/Microsoft/pxt-common-packages/blob/master/libs/base/pxtbase.h#L676

pelikhan: we are not releasing the fiber in there?

mmoskal: yes, we are https://github.com/Microsoft/pxt-common-packages/blob/master/libs/base/pxt.cpp#L535

PR #4761, opened 2018-10-02
MIDI in simulator

pelikhan:
Basic support to play MIDI commands on MIDI devices or emulate using AudioContext. Allows code that talk to MIDI device to be "heard" in the simulator.
- [x] test audio context output
- [ ] MIDI device output


pelikhan:
Good point and thanks for helping on getting this ts.config right.

PR #4763, opened 2018-10-03
added css styles for tables

ChaseMor:
Added CSS styles for tables that better defines cells with grid lines. Also, changes the background color of every other row.

Before:
pizza-table-before

After
pizza-table-after

I wasn't sure what the pxt color scheme was, so I just went with what I thought looked best.

### DIFF
--- a/docfiles/style.css
+++ b/docfiles/style.css
@@ -58,19 +58,6 @@ li code:before, li code:after {
     content: "\00a0";
 }
 
-table {
-    border-collapse: collapse;
-}
-
-table tr th, table tr td {
-    border: 1px solid #333;
-    padding: 6px 13px; 
-}
-
-table tr:nth-child(2n) {
-    background-color: #f8f8f8; 
-}
-
 code.hljs {
     padding: 0 !important;
 }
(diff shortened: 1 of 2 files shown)


docfiles/style.css
```diff
@@ -58,6 +58,19 @@ li code:before, li code:after {
     content: "\00a0";
 }
 
+table {
```
pelikhan: should this go in docs.less?

ChaseMor: I moved the styles into theme/docs.less
Out of curiosity, how do you recompile the styles from less in pxt? I ended up having to setup another local server just to test if it worked.

pelikhan:
pxt buildcss —force

pelikhan:
Note that crowdin has parsing issues with URLs in table so we should not mix those.

pelikhan:
Could you add CSS to handle printing? (in print.less)

ChaseMor:
Is there any special styling that should be done for printing? At the moment it looks fine with the current changes.

pelikhan:
No background fills, does not print well on BW printers

ChaseMor:
If I'm not mistaken, print.less already removes all background fills.

Here's what the print preview looks like:
pizza-table-after-printing

pelikhan:
LGTM

PR #4765, opened 2018-10-03
Adding support to the simulator embed for ignoring the aspect ratio

riknoll:
Our current simulator embed forces the aspect ratio of the iframe. This just adds a query parameter to the URL for ignoring the aspect ratio and filling the iframe. Also fixes an unrelated bug where the embedded sim page ignored the reset message.


pelikhan:
Could you test these changes with microbit?

webapp/public/run.html
```diff
@@ -22,7 +22,6 @@
             position: relative;
             background: transparent;
             width: 100%;
-            padding-bottom: 81.96%;
```
pelikhan: does this get injected later on? or is it useless?

riknoll:
Works fine on mbit!

webapp/public/run.html
riknoll: It gets set on the element. No idea why it was also specified in the page because it always gets ignored.

PR #4767, opened 2018-10-03
display github packge version number in explorer

pelikhan:
Helps identifying versioning issues. Only present for github packages.

image


webapp/src/filelist.tsx
```diff
@@ -108,10 +108,11 @@ export class FileList extends data.Component<ISettingsProps, FileListState> {
             && p.getKsPkg().level <= 1;
         const upd = p.getKsPkg() && p.getKsPkg().verProtocol() == "github";
         const meta: pkg.PackageMeta = this.getData("open-pkg-meta:" + p.getPkgId());
+        const version = upd ? p.getKsPkg().verArgument().split('#')[1] : undefined; // extract github tag
```
riknoll: we have a semver package for getting the parts of the string out

pelikhan: I don't particularly care about the version format, just showing what the editor is loading.

Issue #4757, opened 2018-10-02
Proposal: Ignore params and shadowblocks from block definitions translations


guillaumejenkins:
**Is your feature request related to a problem? Please describe.**
We can't make changes to block IDs or param names in block definitions because currently that breaks existing translations. For example, take the existing block definition for Minecraft `mob.spawn()`:

`block="spawn %entity=minecraftAnimal|at %destination"`

Let's say we create a new block that lets you pick animals, monsters and projectiles in the same field editor (right now, all 3 entity types have their own separate picker). Let's say this new block has the ID `minecraftEntity`. To use this new picker with the `mob.spawn()` block, we would need to change the block definition to:

`block="spawn %entity=minecraftEntity|at %destination"`

Doing this breaks all existing translations, because we've changed the English string, even though all we've changed was an internal parameter.

**Describe the solution you'd like**
When extracting strings for translations, we should ignore parameters and shadow block IDs in block definitions by replacing them with a marker of some sort, like `{}` or something similar. Thus, in the above example of `mobs.spawn()`, the extracted block definition would become:

`"spawn {}at {}"`

There are 2 big advantages to this:

1. This greatly reduces the risk of translators breaking the block definition, because they can no longer modify the param names and shadow block IDs (`%entity=minecraftAnimal`)
2. This lets us modify param names and shadow block IDs on block definitions without breaking translations

The only big problem we need to solve with this is migrating existing translations to this new format.

riknoll:
This came up over lunch today and we talked about using the `param.defl=""` syntax for overriding shadow blocks. In other words:

```typescript
//% block="whatever $param=oldShadow"
//% param.shadow="newShadow"
```

And we consider the old inline syntax deprecated just like the `%` vs `$` for parameters.

Does stripping it from the string entirely does that make localization more difficult? I would imagine some languages would want to re-order parameters to match different grammars. Also the parameter name gives a hint about the meaning to the translator.

guillaumejenkins:
That works!

pelikhan:
This issue is fixed.

PR #4768, opened 2018-10-03
New syntax for setting shadow values

riknoll:
Adding some syntax for setting shadow block ids outside of the block string. Hopefully this will make changing shadow blocks without breaking localizations a bit easier. The new syntax overrides the old syntax and I would say the old syntax should be considered "deprecated".

The proposed syntax:

```typescript
//% block="whatever $param=oldShadow"
//% param.shadow="newShadow"
```

Related: https://github.com/Microsoft/pxt/issues/4757

### DIFF
--- a/docs/defining-blocks.md
+++ b/docs/defining-blocks.md
@@ -103,7 +103,7 @@ type = string
 * the function parameter are mapped **in order** to `%parameter` argument. The loader automatically builds
 a mapping between the block field names and the function names.
 * the block will automatically switch to external inputs when enough parameters are detected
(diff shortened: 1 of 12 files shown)


pelikhan:
Update playground?

riknoll:
Right...

riknoll:
Fixed the playground (also fixed the deprecated % syntax in the block string we were still using)

riknoll:
@pelikhan how do I update pxt-core for the playground? Is that a separate step?

PR #4769, opened 2018-10-04
popup boards dialog after create project

pelikhan:
After creating a new empty project, pop up the "change board" dialog to pick which board to use. Small change, good reuse of existing UI. It only happens when the user creates new project, not when starting tutorials or other occurences. Turned on by a flag.

openboards


samelhusseini:
screenshot please?

pelikhan:
Added

PR #4775, opened 2018-10-04
Update the 'blocks-embed' page

ganicke:
Sectioned the process, added more description text, and included an example HTML document.


pelikhan:
Could we link to a jsfiddle example?

ganicke:
Could maybe, where do you suggest _./webfiles_ be located at?

pelikhan:
Let's get this in first. Need to figure out how to do an example properly.

PR #4777, opened 2018-10-04
incorrect type retrieved for new expression

tballmsft:
We were retrieving the contextual type for a new expression... which made the following code fail:
```
    export const P0: PwmPin = new MicrobitPin(DigitalPin.P0);
```
because PwmPin is an interface implemented by the class MicrobitPin. Obviously, you can't new on an interface. The fix is to use the type of the new expression rather than the contextual type.


pelikhan:
Can we have a test for this?

PR #4782, opened 2018-10-05
JS compat, GC and other goodies

mmoskal:
* new shiny garbage collector, which improves perf about 2x
* type of receiver is now validated upon member access and casts between types are allowed (and unchecked)
* add `PXT_COMPILE_SWITCHES=a,b,c pxt ...` and `?compile=a,b,c`
* add `profile` compiler switch, which enables performance counters (other compiler switches are not very useful for a general user at this time)
* `pxt gdb` improvements on samd
* add error code enum with docs
* implement patching of literals in .HEX (in addition to UF2) - this is because of new uniform VT format
* better size stats in `binary.asm`
* add `--noImplicitThis` to most projects
* many perf improvements and bugfixes

Added JS/TS features:
* `any` type support; union/intersection types etc are probably just a matter of lifting restrictions in emitter
* `typeof` expression
* `public`/`private` annotations on constructor arguments (syntactic sugar to make them into fields)
* initializers for class fields
* lambda functions with more than three arguments
* using generic functions as values and nested generic functions
* downcasts of a superclass to a subclass
* `instanceof`
* function parameter bi-variance
* class inheritance for generic classes and methods
* `String.slice`
* `Array.isArray`
* `Object.keys` (only works on dynamic maps at the moment, not on classes)

Requires synchronized change to common packages - https://github.com/Microsoft/pxt-common-packages/pull/423


mmoskal:
It should be fairly safe, I did run the lang-test0 stuff. These are actually rather minor changes (inlining). It's only native.

mmoskal:
I guess I could stay on this branch for JSON changes (which are likely a bit more intrusive), so we can merge it all together

mmoskal:
Will do JSON changes here next week

mmoskal:
@abchatra @pelikhan @tballmsft - updated description with the actual changes; I think it should be good to merge around Monday, especially if we're not planning any releases next week

PR #4798, opened 2018-10-11
Major update to the grid picker to improve it's performance + other bug fixes

samelhusseini:
Improve the performance of the grid picker: (current: ~1400ms to load)
- Add intersection observer support for images (shaved off 200ms)
- Remove closure dependency and calculate positioning more efficiently (~800ms)
- Add tooltips in a more efficient way (working with closure meant that we had to add tooltips as a post process) (~100ms)
It comes down to ~200ms to load

Also in this PR:
- Support a better touch experience by adding a selected bar. Selected bar only appears if there are tooltips, otherwise, we select on click.
- Updated look of grid picker with no images (micro:bit)
- Bug fixes with positioning of grid picker as well as bug fixes with hiding the tooltips after close. 

Selected bar: 
Only shows up on mobile devices and makes it so that the experience of selecting an item is in two steps. A selection which shows you the name of the image, and then a select / cancel button to select. 
screen shot 2018-10-10 at 6 08 26 pm

Updated look for grid pickers without an image: 
screen shot 2018-10-10 at 6 12 49 pm

Tested: 
- micro:bit grid pickers for pins
- Minecraft grid pickers, ones with a set number of rows and ones without, same for ones with searchbar and without
- Devices: Mac, Windows, iPad
- Browsers: Chrome and Edge, Firefox, Safari, Chrome Mobile

Pull in a pxt-blockly build with these two PRs: 
https://github.com/Microsoft/pxt-blockly/pull/125
https://github.com/Microsoft/pxt-blockly/pull/126


pelikhan:
Awesome improvements!

PR #4800, opened 2018-10-11
Fix page visibility check in app.

samelhusseini:
If we're in an app, don't worry about page visibility check.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -121,7 +121,7 @@ export class ProjectView
         this.state = {
             showFiles: false,
             home: shouldShowHomeScreen,
-            active: document.visibilityState == 'visible',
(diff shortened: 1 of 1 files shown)


pelikhan:
What about being hosted in iOS?

samelhusseini:
Good question, I don't know how we know that we're hosted in IOS..

pelikhan:
We are in the micro:bit iOS app (upcoming beta)

samelhusseini:
Yes but how can we identify that? (in code)

samelhusseini:
As in UserAgent won't cut it.

pelikhan:
Forget it for now. Might not matter.

samelhusseini:
Okay.

PR #4801, opened 2018-10-11
Robust WebBLE UART + partial partial flashing service

pelikhan:
Reliability work around WebBLE support. Once the device is paired, the editor will reconnect after reset and/or USB flashing and/or WebUSB flashing.

* refactoring services
* better handling of disconnection
* automatic reconnection for UART
* start of the partial flash service


pelikhan:
@mmoskal ready for review

pxtlib/service.ts
```diff
@@ -1252,6 +1252,7 @@ namespace ts.pxtc {
                 let ptr = i * 512
                 let bl = parseBlock(blocks.slice(ptr, ptr + 512))
                 if (!bl) continue
+                if (endAddr && bl.targetAddr > endAddr) break;
```
mmoskal: I would say `>=` or even `bl.targetAddr + 256 > endAddr`

pelikhan: fixed with 2nd option

PR #4804, opened 2018-10-11
Changing the exit dialog to give just one option

abchatra:
Feedback from the designer.
- If the intent is for name the project, say so in the dialog
- Give only one option Done
- Close (x) should still go back to the previous state

New
image

Old
image

### DIFF
--- a/webapp/src/projects.tsx
+++ b/webapp/src/projects.tsx
@@ -776,21 +776,17 @@ export class ExitAndSaveDialog extends data.Component<ISettingsProps, ExitAndSav
             onclick: this.save,
             icon: 'check',
             className: 'approve positive'
-        }, {
-            label: lf("Cancel"),
-            icon: 'cancel',
-            onclick: this.cancel
         }]
 
         return (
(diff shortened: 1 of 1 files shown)


webapp/src/projects.tsx
```diff
@@ -776,21 +776,17 @@ export class ExitAndSaveDialog extends data.Component<ISettingsProps, ExitAndSav
...
-            label: lf("Cancel"),
-            icon: 'cancel',
-            onclick: this.cancel
         }]
 
         return (
             <sui.Modal isOpen={visible} className="exitandsave" size="tiny"
                 onClose={this.hide} dimmer={true} buttons={actions}
-                closeIcon={true} header={lf("Exit Project")}
+                closeIcon={true} header={lf("Project has no name") + " 😞"}
```
pelikhan: Do ``lf("Project has no name {0}", ":(")`` . String concat does not localize for RTL languages.

abchatra:
Also how it looks on IE
image

samelhusseini:
Fixes https://github.com/Microsoft/pxt-microbit/issues/1365

PR #4806, opened 2018-10-11
Disabling file import dialog on all iOs

abchatra:
There is no support for files and they don't work today. We will reevaluate later if necessary


abchatra:
Fixes https://github.com/Microsoft/pxt-minecraft/issues/672

abchatra:
How is microbit using this?

pelikhan:
You can import and export .hex files via iCloud. We had bugs in the past in that area and this is something that people have reported... so they use it.

riknoll:
@pelikhan what would the authoring look like? Maybe we could take a list of platforms or something? Or maybe we should make this an editor extension and let the target run whatever code they want to detect if they can import files.

pelikhan:
Any editor that can save and import file can use this feature. I don't understand why it has to be disabled on iOS.

PR #4812, opened 2018-10-14
Cancellable deploy

pelikhan:
Click on download button cancels the download operation cleanly. Needs UI review.

This is particularly useful for WebUSB/WebBLE/HID which might run longer than expected.


darzu:
Resurrecting this since we're having an infinite download spinner issue on multiple platforms that's been hard to pin down.

Allowing downloads to be canceled gives the users a work-around and we can instrument the cancellations to hopefully gain more data on what's going wrong.

darzu:
woah, something is goofed since it's showing 837 files changed.

darzu:
Starting a new PR. Somehow this one got messed up.

PR #4813, opened 2018-10-15
Use address in blocks-embed sample

ganicke:
Use literal address string instead of macro.


pelikhan:
Try @embedUrl@ or @homeUrl@

ganicke:
Re-submit in #4845

PR #4840, opened 2018-10-15
don't show more than 7 character of github tag

pelikhan:
Fix for https://github.com/Microsoft/pxt-microbit/issues/1421


webapp/src/filelist.tsx
```diff
@@ -108,7 +108,8 @@ export class FileList extends data.Component<ISettingsProps, FileListState> {
             && p.getKsPkg().level <= 1;
         const upd = p.getKsPkg() && p.getKsPkg().verProtocol() == "github";
         const meta: pkg.PackageMeta = this.getData("open-pkg-meta:" + p.getPkgId());
-        const version = upd ? p.getKsPkg().verArgument().split('#')[1] : undefined; // extract github tag
+        let version = upd ? p.getKsPkg().verArgument().split('#')[1] : undefined; // extract github tag
+        if (version && version.length > 7) version = version.substring(0, 7);
```
mmoskal: maybe `version.length > 20) version = version.substring(0,7)` ? so we don't cut the semvers

PR #4844, opened 2018-10-15
Codal listener

tballmsft:
Handle concurrent modification of handlers correctly and add logic for handling any.


pelikhan:
What is the status on this? Is the CODAL side merged in already?

PR #4861, opened 2018-10-18
Made the emoji change when the user names their project

ChaseMor:
I feel that the emoji that is shown when users are told that their project has no name is a little too sad. It makes me feel really bad when I forget to name my project before going to the homepage.
sad-save-screen

This changes the emoji when the user enters a name other than "" and "Untitled".
sad-save-screen

I also think this could be a fun way to add easter eggs into PXT. Like saving your project as certain names causes certain emojis to appear.

### DIFF
--- a/webapp/src/projects.tsx
+++ b/webapp/src/projects.tsx
@@ -699,6 +699,7 @@ export class ImportDialog extends data.Component<ISettingsProps, ImportDialogSta
 
 export interface ExitAndSaveDialogState {
     visible?: boolean;
+    emoji?: string;
     projectName?: string;
 }
 
@@ -707,7 +708,8 @@ export class ExitAndSaveDialog extends data.Component<ISettingsProps, ExitAndSav
     constructor(props: ISettingsProps) {
         super(props);
         this.state = {
-            visible: false
+            visible: false,
+            emoji: "😞"
         }
 
         this.hide = this.hide.bind(this);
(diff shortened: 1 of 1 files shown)


Jaqster:
+1  What a great idea!  :-)

webapp/src/projects.tsx
```diff
@@ -747,6 +747,12 @@ export class ExitAndSaveDialog extends data.Component<ISettingsProps, ExitAndSav
 
     handleChange(name: string) {
         this.setState({ projectName: name });
+        let header = document.getElementsByClassName("exitandsave").item(0).firstChild as HTMLElement;
```
samelhusseini: Why not do this the react way and update the state to call a render?

samelhusseini: (save's us from having to have a dependency on a classname)

webapp/src/projects.tsx
```diff
@@ -747,6 +747,12 @@ export class ExitAndSaveDialog extends data.Component<ISettingsProps, ExitAndSav
...
+        if (name === "" || name === "Untitled") {
```
samelhusseini: Also, Untitled is not the name of the project in all languages. 
Add lf around that to get the Untitled name in the user's lang

webapp/src/projects.tsx
samelhusseini: Just get rid of this line and you're good!

PR #4862, opened 2018-10-18
Added check so that search bar doesn't crash if no results are returned

ChaseMor:
Fixes Microsoft/pxt-arcade#318

The toolbox disappears if your search for blocks returns no results. Confirmed this on both arcade (issue is on the live site) and adafruit (served locally with latest pxt).

Looks like the issue is related to #4846, but I am unsure if this fixes the the root of the bug or just hides it better.

### DIFF
--- a/webapp/src/blocks.tsx
+++ b/webapp/src/blocks.tsx
@@ -1337,7 +1337,7 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         }
         if (blockXml) {
             if (ignoregap) {
-                blockXml.setAttribute("gap", `${pxt.appTarget.appTheme && pxt.appTarget.appTheme.defaultBlockGap.toString() || 8}`);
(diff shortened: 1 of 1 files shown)


pelikhan:
Are there other locations where we use this field? Maybe we should set it in the patching of the theme

ChaseMor:
It is to my understanding that blocks are allowed to have a blockGap field defined for each block. Each category recently gained the ability to ignore this set block gap in favor of a defaultBlockGap field that is set in pxtarget.json. The code here is intended to set the gap to defaultBlockGap if it exists and set it to "8" if it doesn't. 

[The only other place that uses this field](https://github.com/Microsoft/pxt/blob/64d93b4d47f260621d0c349f598adfce618fcc66/pxtblocks/blocklyloader.ts#L359) performs a similar check to see if defaultBlockGap exists, so this should be a good fix. 

Is this correct @samelhusseini?

ChaseMor:
I have also noticed that the problem extends to all search queries. Not just ones that don't return anything.

samelhusseini:
Woops looks like I fixed this in a different change. My bad. 
https://github.com/Microsoft/pxt/pull/4864 (same change there)

PR #4869, opened 2018-10-18
Load Blockly before trying upgrades

riknoll:
Fixes https://github.com/Microsoft/pxt-microbit/issues/1449


abchatra:
Looks good. Please link pxt to this blockly after the merging.

PR #4872, opened 2018-10-19
Fixing the carousel in RTL languages

riknoll:
Fixes https://github.com/Microsoft/pxt-microbit/issues/1426

Tested in Edge, Firefox, Chrome, and IE


samelhusseini:
Looks good, we should also test this on touch devices.

pelikhan:
FYI iPhone the carousel is too sensible and capture drag gesture as horizontal. Would be nice to address this too.

samelhusseini:
^^ agreed with Peli, but perhaps in a different change ;)

riknoll:
This PR doesn't really change how touch is handled at all, just reverses the direction in RTL

PR #4874, opened 2018-10-19
Valid and invalid code snippets

ChaseMor:
There have been a few times writing course curriculum in which it is helpful to show bad code or code that doesn't work. I fear that students may skim through code examples and mistake bad code for code that they should copy and learn from. 

Some sort of distinction was requested in Microsoft/pxt-arcade#316

This emphasizes what bad and good code is by setting the background color to red or green (I used the same palette as GitHub). 

valid-invalid-code

To use these snippets, the \```typescript-invalid and \```typescript-valid languages are specified in markdown.

The markdown used to create the sample above is as follows:

\# Valid and Invalid code snippets

\## Normal Code Snippet

\```typescript
// Normal Code
function printMessage() {
&nbsp;&nbsp;&nbsp;&nbsp;console.log("This is what a normal code snippet looks like")
}
\```

\## Invalid Code Snippet

\```typescript-invalid
// Invalid Code
function printMessage() {
&nbsp;&nbsp;&nbsp;&nbsp;console.log("This is what a invalid code snippet looks like")
}
\```

\## Valid Code Snippet

\```typescript-valid
// Valid Code
function printMessage() {
&nbsp;&nbsp;&nbsp;&nbsp;console.log("This is what a valid code snippet looks like")
}
\```

Both of these new snippets have the same behavior as ```typescript-ignore in that they are ignored by travis and do not have button to run them (or show/edit). 

Also, when printing, the background colors are removed and they look just like normal code snippets.

### DIFF
--- a/docs/writing-docs/macros.md
+++ b/docs/writing-docs/macros.md
@@ -150,6 +150,26 @@ Renders one or more codecards as JSON into cards
 Append `-ignore` to any of the above to ignore a snippet in automated testing:
 
     ```typescript-ignore
+    // You can include illegal TS in here, e.g. to showcase concepts/psuedocode 
+    for (initialization; check; update) {
+        ...
+    }
+    ```
+
+### invalid
+
+You can use `typescript-invalid` to showcase typescript that is **incorrect**:
+
+    ```typescript-invalid
     // You can include illegal TS in here, e.g. to document syntax errors
     callFunction(;
+    ```
+
+### valid
+
+You can use `typescript-valid` to showcase typescript that is **correct**:
+
+    ```typescript-valid
+    // You can include any TS in here, e.g. to showcase correct syntax
+    callFunction();
     ```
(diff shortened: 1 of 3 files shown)


pelikhan:
This is a great idea. Could we find a better theme for those? I think we should have an icon to depict the problems since red/green is going away at printing. (also red, green == braun for color blind users)

pelikhan:
Also can you update our markdown docs too?

riknoll:
Here's how the material design folks do it:
https://material.io/design/components/buttons.html#anatomy

Color is accompanied by a "do" or "dont" label with explanations.

samelhusseini:
Yeah I love this as well! As Richard suggested perhaps the background isn't what should change? or if so, it should be a much lighter difference, it's too stark today and that introduces contrast issues with the color of the text.

pelikhan:
Can you upgrade the screenshots?

ChaseMor:
valid-invalid-code-updated

This is the current theme.

pelikhan:
How does it look printed?

ChaseMor:
image

pelikhan:
LGTM

PR #4876, opened 2018-10-19
Turning on the no-reserved-keywords tslint rule

riknoll:
I tried my best to not change any external APIs, so its disabled in a few places.


riknoll:
Doesn't seem to be going anywhere

PR #4883, opened 2018-10-22
Added circle and line tools to the sprite editor

ChaseMor:
This adds the ability to use circle and line tools in the sprite editor. By holding down shift, the pencil tool turns into the line tool, and by holding down control, the rectangle tool becomes the circle tool.

new-sprite-editor

Also fixes the following bugs:
* Pencil tool not showing as selected after color is chosen while eraser tool is selected
* Tool width not showing selected on opening the editor
* Circles were previously more pointy. Changed algorithm to better reflect the Midpoint Circle Algorithm

The main bits that I think need to be discussed is which keys should change which tools and also whether or not the first click should set the circles center rather than the upper-left of the circle.

I think it should be the upper-left of the circle because that allows for future implementations of ovals and even-width circles.

### DIFF
--- a/pxtblocks/fields/sprite/sidebar.ts
+++ b/pxtblocks/fields/sprite/sidebar.ts
@@ -184,12 +184,10 @@ namespace pxtblockly {
 
         getButtonForTool(tool: PaintTool) {
             switch (tool) {
-                case PaintTool.Normal: return this.pencilTool;
                 case PaintTool.Normal:
                 case PaintTool.Line: return this.pencilTool;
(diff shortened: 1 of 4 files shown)


pelikhan:
GitHub webhooks is down.

pelikhan:
How are we going to teach this?

ChaseMor:
@pelikhan I am currently working on some tutorials that cover some of the basics of arcade. The first one is on sprites and I was going to mention these tools when showing how to make sprites.

PR #4889, opened 2018-10-22
Don't create a new project when the share page reloads

riknoll:
When we push an editor release, the share page reloads and shows a blank project. I cannot find the issue anymore but you know the one I'm talking about.


abchatra:
Port to stable4.1

PR #4898, opened 2018-10-23
Open editor when migrating browser prefixes

riknoll:
I tested this using the file workspace and it worked fine. Marking as "DO NOT MERGE" until a decision is reached. Also if merged needs to go into stable 4.1


pelikhan:
This seems like a hack that will only apply to the microbit, but we'll forget that it is there.

riknoll:
Closing because the decision was made

PR #4905, opened 2018-10-24
Decompiler shouldn't decompile banned categories

riknoll:
Fixes https://github.com/Microsoft/pxt-arcade/issues/212


samelhusseini:
Nice, didn't realise we had this feature :)

PR #4908, opened 2018-10-24
Pinned releases

pelikhan:
Allow to specify specific package releases for packages per major version of the editor.

In micro:bit, allows us to snap all approved packages to a particular version for v0.


pelikhan:
This will have to go in v0.

PR #4911, opened 2018-10-24
Few edits to freeze note

ganicke:
Is 'freeze' and 'lock to' the same idea?


pelikhan:
Yes freeze / lock is the sma.e

PR #4915, opened 2018-10-25
Tags in card details

pelikhan:
Let's keep discussing this one. 

* do we ever want to show tags on the cards?
* do we need special tags for number values like "duration", "age", etc..

- [x] Very basic tags in the details view. Needs proper skinning. 
- [x] colors specified in pxttarget.
image

### DIFF
--- a/localtypings/pxtpackage.d.ts
+++ b/localtypings/pxtpackage.d.ts
@@ -100,7 +100,7 @@ declare namespace pxt {
         ariaLabel?: string;
         label?: string;
         labelClass?: string;
-        tags?: CodeCardTag[]; // tags shown in home screen, eg: [{ "label": "Beginner", "color": "blue" }]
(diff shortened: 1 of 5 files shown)


localtypings/pxtpackage.d.ts
```diff
@@ -100,7 +100,7 @@ declare namespace pxt {
         ariaLabel?: string;
         label?: string;
         labelClass?: string;
-        tags?: CodeCardTag[]; // tags shown in home screen, eg: [{ "label": "Beginner", "color": "blue" }]
+        tags?: string[]; // tags shown in home screen, colors specified in theme
```
samelhusseini: I presume we're no longer using this anywhere, hence re-purposed.

pelikhan:
Definitely needs love. Probably will happen for LEGO.

PR #4916, opened 2018-10-25
Add tooltip support in React

samelhusseini:
Add support for tooltips on elements. 
Using react-tooltip: https://github.com/wwayne/react-tooltip (MIT)

Note: We need to file an OSS request for this.


pelikhan:
What is the default delay?

pelikhan:
This is very nice. We could/should add a hinting engine to the tutorial that uses those tooltips. Can you trigger them programmatically?

samelhusseini:
Yes you can trigger programmatically and the interface I added here is probably not the best, we'd have to play around and see how we want to use it. Will keep adjusting as we add more features.

samelhusseini:
Filed OSS request and Approved. We're good to go.

samelhusseini:
Default delay is 0. We can change that in our module.

samelhusseini:
This is ready to go in. @pelikhan can I get a final review?

pelikhan:
Looks good. where do we plan to add those?

PR #4920, opened 2018-10-25
Making the editor extension dialog a little less annoying

riknoll:
A few tweaks based on feedback from @pelikhan, @jwunderl, and @ChaseMor 

1. Fixes a bug where the dialog opens if you are editing a file and accidentally create an error. It now only opens on load, simulator start, and compilation.
2. Changed the (x) behavior so that it only takes you to the home screen if you are coming from the home screen. It was very annoying to click the (x) after downloading and have it close the project
3. The dialog should now only show in a blocks file or `main.ts`, so Github package authors won't have it constantly showing up

This should probably go to stable 4.1 once we start merging things again.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -400,10 +400,16 @@ export class ProjectView
     }
 
     private maybeShowPackageErrors(force = false) {
+        // Only show in blocks or main.ts
+        if (this.state.currFile) {
+            const fn = this.state.currFile.name;
+            if (!pkg.File.blocksFileNameRx.test(fn) && fn !== "main.ts") return false;
+        }
+
         if (!this.state.suppressPackageWarning || force) {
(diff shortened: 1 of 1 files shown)


pelikhan:
For (3), you can check header.githubId to check if it is a github repo.

samelhusseini:
Editor extension dialog? What are you referring to. 
editor extensions to me = the blockly buttons that load up an iframe..

pelikhan:
Formerly package dialog.

jwunderl:
@samelhusseini This fixes the bug described in microsoft/pxt-arcade#292

pelikhan:
I didn't realize this triggered on a typo... this is a good catch.

riknoll:
@pelikhan I just added one more check to filter out all core packages from triggering this dialog. They should never have errors after all

PR #4922, opened 2018-10-26
Improved emojis animation

pelikhan:
happy

( also added .header.json in template .gitignore )

### DIFF
--- a/pxtlib/initprj.ts
+++ b/pxtlib/initprj.ts
@@ -61,6 +61,7 @@ yotta_targets
 pxt_modules
 *.db
 *.tgz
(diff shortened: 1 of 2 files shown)


ChaseMor:
This is brilliant!

It looks like this doesn't check if the name is "Untitled". It only checks substrings up until "Untitle".

Also, I bug I noticed from before is that the emoji doesn't reset when a new project is opened. It carries over from whatever it was before until the title is changed. Can we use this as an opportunity to fix this?

pelikhan:
@ChaseMor i think i fixed your issues.

PR #4924, opened 2018-10-26
Make debugging pxt-blockly in PXT easier.

samelhusseini:
I've added a couple of scripts that link pxt-blockly and closure-library under built/web. 
Once blockly is linked, I've created blockly.html which uses the uncompressed blockly files instead of our version that is both compressed and built into pxtblockly.js. 

This means we can debug Blockly in the same way that we do using the Blockly playground but in PXT. 

@riknoll @guillaumejenkins @pelikhan FYI (I know you guys have asked for this over and over again, so thought I'd finally do it ha). 

@pelikhan is there any issue with hosting that blockly.html file, or should I place it somewhere else. This is only for local debugging.

### DIFF
--- a/Jakefile.js
+++ b/Jakefile.js
@@ -307,6 +307,7 @@ file("built/web/pxtlib.js", [
 
     jake.cpR("built/pxtlib.js", "built/web/")
     jake.cpR("built/pxtcompiler.js", "built/web/")
+    jake.cpR("built/pxtblocks.js", "built/web/")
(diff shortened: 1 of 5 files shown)


pelikhan:
Don't see any issue.

pelikhan:
Can you add a markdown docs on how to use this?

samelhusseini:
Added a README file under scripts to describe how to use each of the scripts.

PR #4925, opened 2018-10-26
Support adding base telemetry properties dynamically

guillaumejenkins:


abchatra:
for...in supported, but probably not in conjunction with const.

riknoll:
Best to use `Object.keys` anyway

riknoll:
Yeah

PR #4931, opened 2018-10-27
always center on highlight

pelikhan:
This is too jarring. I think we need something like "scrollIntoView" that push the block just enough to be visible.
tracecenter

Always center in tracing mode.
- [x] fix center API. The example below does not porperly bring the block into the view. https://github.com/Microsoft/pxt/pull/4976
- [ ] tune speed
- [x] we need to consider to make "highlightStatement" async and wait till the transition is done -- otherwise it gets jarring when blocks are far away.

```
basic.forever(function () {
    basic.showLeds(`
        . . . . .
        . . . . .
        . . . . .
        . . . . .
        . . . . .
        `)
    basic.showLeds(`
        . . . . .
        . . . . .
        . . . . .
        . . . . .
        . . . . .
        `)
    basic.showIcon(IconNames.Heart)
    basic.showLeds(`
        . . . . .
        . . . . .
        . . . . .
        . . . . .
        . . . . .
        `)
    basic.showLeds(`
        . . . . .
        . . . . .
        . . . . .
        . . . . .
        . . . . .
        `)
    basic.showIcon(IconNames.Heart)
})
```


pelikhan:
Don’t merge. The centering from blackly has bugs... but it is definitely the way to go for tracing.

Also centerOnScreen should return a promise so that we can wait till the animation is done.

PR #4939, opened 2018-10-29
Grouping serial messages from sim to improve performance

riknoll:
Fixes https://github.com/Microsoft/pxt-microbit/issues/1518

Currently this program will hang our editor:

image

The issue seems to be that we are overloading the UI thread by posting way too many serial messages. I've added some logic to group these messages before sending them and it makes a big improvement.

### DIFF
--- a/pxtsim/embed.ts
+++ b/pxtsim/embed.ts
@@ -78,6 +78,12 @@ namespace pxsim {
         sim?: boolean;
         receivedTime?: number;
     }
+    export interface SimulatorBulkSerialMessage extends SimulatorMessage {
+        type: "bulkserial";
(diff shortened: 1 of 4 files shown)


pelikhan:
The DAL does have a sleep when sending data over serial. We could add a bit of delay in the sim to give a breather too.

riknoll:
200ms still looks pretty instant in the serial view (I was graphing accelerometer data). The other debounce function has different logic and doesn't work in a tight loop; this one debounces with a max timeout (so calling it constantly won't make it wait forever).

riknoll:
I'm investigating the build issue

PR #4949, opened 2018-10-31
Stop the simulator if running when code is invalidated (autorun: false)

samelhusseini:
Reason for this change: 
(Most of our targets have autoRun = true).
If the simulator is running and autoRun is false, then if a user makes a change to their code (invalidates the code running), we should stop the simulator. 
@pelikhan do you agree with this behavior or should I have it behind a flag?

This also affects Wonder workshop so we should chat to them about this change.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -482,6 +482,9 @@ export class ProjectView
                             if (this.editor == this.blocksEditor) this.autoRunBlocksSimulator();
                             else this.autoRunSimulator();
                         }
+                    } else if (this.state.running) {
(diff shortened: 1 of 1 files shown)


pelikhan:
Does it automatically restart if autoRun is true?

samelhusseini:
Yes it does.

abchatra:
@riknoll can you take a look took too?

abchatra:
@dragonx - FYI. Let us know if you are ok with this change.

samelhusseini:
Looks like there's already a flag that allows that: https://github.com/Microsoft/pxt/pull/917

samelhusseini:
Closing.

PR #4950, opened 2018-10-31
Use a cancellationtoken to stop the simulator from running midway through.

samelhusseini:
The purpose of this change is to close up the loop between start and stop. 
- If you start multiple times, only the first should succeed. 
- If you start and then stop, then the start should be cancelled. 

@riknoll @pelikhan PR ready for review.


webapp/src/app.tsx
```diff
@@ -1843,7 +1844,7 @@ export class ProjectView
         return this.state.running;
     }
 
-    runSimulator(opts: compiler.CompileOptions = {}, clickTrigger?: boolean) {
+    runSimulator(opts: compiler.CompileOptions = {}, clickTrigger?: boolean, cancellationToken?: pxt.Util.CancellationToken) {
```
pelikhan: Can we have a run option object that is a superset of compileoptions? This is getting ugly.

webapp/src/app.tsx
```diff
@@ -1713,19 +1713,20 @@ export class ProjectView
...
         }
         this.saveFileAsync()
-            .then(() => this.runSimulator(debug ? { debug: true } : {}, clickTrigger));
+            .then(() => this.runSimulator(debug ? { debug: true } : {}, clickTrigger, cancellationToken));
     }
 
     stopSimulator(unload?: boolean) {
         pxt.tickEvent('simulator.stop')
         simulator.stop(unload)
+        console.log("simulator stopping, setting state false");
```
pelikhan: pxt.log

samelhusseini: Woops, shouldn't be there.

samelhusseini:
I am handling cancellation on the editor side in https://github.com/Microsoft/pxt-minecraft/pull/757 but perhaps you have a good point, to just handle it internally.

webapp/src/app.tsx
samelhusseini: I think I can just move them all into compiler.CompileOptions

samelhusseini:
@pelikhan made the change to handle token internally, ready for review.

webapp/src/app.tsx
```diff
@@ -1719,12 +1721,15 @@ export class ProjectView
             pxt.log("Tried to start simulator when we shouldn't");
             return;
         }
+        if (this.runToken) this.runToken.cancel()
+        this.runToken = new pxt.Util.CancellationToken()
```
pelikhan: I think you need startOperation?

webapp/src/app.tsx
```diff
@@ -1857,14 +1862,19 @@ export class ProjectView
...
                 this.clearSerial();
                 this.editor.setDiagnostics(this.editorFile, state)
+                if (opts.cancellationToken && opts.cancellationToken.isCancelled()) return;
                 if (resp.outfiles[pxtc.BINARY_JS]) {
-                    simulator.run(pkg.mainPkg, opts.debug, resp, this.state.mute, this.state.highContrast, pxt.options.light, clickTrigger)
+                    simulator.run(pkg.mainPkg, opts.debug, resp, this.state.mute, this.state.highContrast, pxt.options.light, opts.clickTrigger)
+                    if (opts.cancellationToken && opts.cancellationToken.isCancelled()) {
+                        this.stopSimulator();
+                    }
                     this.setState({ running: true, showParts: simulator.driver.runOptions.parts.length > 0 })
```
pelikhan: Setting state running after stopping simulator

webapp/src/app.tsx
samelhusseini: I'm only using the token to detect cancel.

webapp/src/app.tsx
samelhusseini: Yeah I'm seeing that too.

samelhusseini:
I'm tossing up whether it's better if we receive a number of simultaneous startSimulator calls to either: 
- start the first and ignore the rest
- cancel each subsequent one and only run during the last one. 

My gut feeling is that it doesn't matter, since the calls are made right after each other, there's no change for saveAsync to be invalidated. @pelikhan does this make sense?

pelikhan:
You could keep a hash of the run option and restart if they changed.

webapp/src/app.tsx
```diff
@@ -1715,16 +1717,20 @@ export class ProjectView
...
+        if (!this.shouldStartSimulator() || (this.runToken && this.runToken.isRunning())) {
+            pxt.log("Ignoring call to start simulator, either already running or we shouldn't start.");
             return;
         }
+        if (this.runToken) this.runToken.cancel()
+        this.runToken = new pxt.Util.CancellationToken()
+        this.runToken.startOperation();
         this.saveFileAsync()
-            .then(() => this.runSimulator(debug ? { debug: true } : {}, clickTrigger));
+            .then(() => this.runSimulator({ debug, clickTrigger }, this.runToken));
```
pelikhan: remove token from signature, it is handled internally

webapp/src/app.tsx
```diff
@@ -1843,7 +1849,7 @@ export class ProjectView
         return this.state.running;
     }
 
-    runSimulator(opts: compiler.CompileOptions = {}, clickTrigger?: boolean) {
+    runSimulator(opts: compiler.CompileOptions = {}, token?: pxt.Util.CancellationToken) {
```
pelikhan: already a member of the class, "this.runToken"

webapp/src/app.tsx
```diff
@@ -1856,19 +1862,27 @@ export class ProjectView
         if (this.state.tracing)
             opts.trace = true;
 
-        this.stopSimulator();
+        this.stopSimulator(undefined, true);
+        if (token && token.isCancelled()) return Promise.resolve();
```
pelikhan: use this.runtToken

webapp/src/app.tsx
```diff
@@ -1856,19 +1862,27 @@ export class ProjectView
...
         const state = this.editor.snapshotState()
         return compiler.compileAsync(opts)
             .then(resp => {
                 this.clearSerial();
                 this.editor.setDiagnostics(this.editorFile, state)
+                if (token && token.isCancelled()) return;
                 if (resp.outfiles[pxtc.BINARY_JS]) {
-                    simulator.run(pkg.mainPkg, opts.debug, resp, this.state.mute, this.state.highContrast, pxt.options.light, clickTrigger)
-                    this.setState({ running: true, showParts: simulator.driver.runOptions.parts.length > 0 })
+                    if (!token || !token.isCancelled()) {
```
pelikhan: this was checked a line ago

samelhusseini: Yes but it takes a while to run the simulator? Should we check after just in case?

webapp/src/app.tsx
samelhusseini: I guess the issue I have with using runToken by default for runSimulator is that we call runSimulator from a number of places and not all of them go through the path of: 
startSimulator -> runSimulator -> stopSimulator.

I'll take another look and see if I can standardize it.

samelhusseini:
How about runSimulator starts the runToken, and stopSimulator cancels. 
StartSimultor just does a save as well. and that's usually pretty quick.

webapp/src/app.tsx
```diff
@@ -1845,11 +1816,49 @@ export class ProjectView
...
+            pxt.log("Ignoring call to start simulator, either already running or we shouldn't start.");
+            return;
+        }
+        this.saveFileAsync()
+            .then(() => this.runSimulator({ debug, clickTrigger }))
+    }
+
+    stopSimulator(unload?: boolean, dontCancel?: boolean) {
+        pxt.tickEvent('simulator.stop')
+        if (!dontCancel && this.runToken) this.runToken.cancel()
```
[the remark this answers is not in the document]
samelhusseini: Because runSimulator calls stopSimulator when it begins, which would cancel the token midway through. We only want to cancel if a different action called stopSimulator.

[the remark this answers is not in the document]
samelhusseini: We could alternately call simulator.stop directly and we can probably avoid the tick event here when its an internal call to stop the simulator.

webapp/src/app.tsx
```diff
@@ -1845,11 +1816,49 @@ export class ProjectView
...
+        simulator.stop(unload)
+        this.setState({ running: false })
+    }
+
+    suspendSimulator() {
+        pxt.tickEvent('simulator.suspend')
+        simulator.suspend()
```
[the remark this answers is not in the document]
samelhusseini: Possibly, let me see.

webapp/src/app.tsx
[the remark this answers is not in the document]
pelikhan: The token is internal sauce, refactor stop into an internal method.

samelhusseini:
Final thoughts on this @pelikhan

samelhusseini:
Yep.

webapp/src/app.tsx
```diff
@@ -1845,37 +1816,90 @@ export class ProjectView
...
-            })
+    suspendSimulator() {
+        pxt.tickEvent('simulator.suspend')
+        if (this.runToken) this.runToken.cancel()
+        simulator.suspend()
+    }
+
+    runSimulator(opts: compiler.CompileOptions = {}) {
+        if (this.runToken) this.runToken.cancel()
+        let runToken = new pxt.Util.CancellationToken();
```
pelikhan: for the sake of clarity, could you use a differnt name for the local token? ``cancellationToken`` ?

webapp/src/app.tsx
```diff
@@ -1845,37 +1816,90 @@ export class ProjectView
...
-                if (resp.outfiles[pxtc.BINARY_JS]) {
-                    simulator.run(pkg.mainPkg, opts.debug, resp, this.state.mute, this.state.highContrast, pxt.options.light, clickTrigger)
-                    this.setState({ running: true, showParts: simulator.driver.runOptions.parts.length > 0 })
-                } else if (!opts.background) {
-                    core.warningNotification(lf("Oops, we could not run this project. Please check your code for errors."))
-                }
-            })
+    suspendSimulator() {
+        pxt.tickEvent('simulator.suspend')
+        if (this.runToken) this.runToken.cancel()
```
pelikhan: clear runtoken aftewards.

webapp/src/app.tsx
```diff
@@ -1845,37 +1816,90 @@ export class ProjectView
...
+            return;
+        }
+        this.saveFileAsync()
+            .then(() => this.runSimulator({ debug, clickTrigger }))
+    }
 
-        this.stopSimulator();
+    stopSimulator(unload?: boolean) {
+        pxt.tickEvent('simulator.stop')
+        if (this.runToken) this.runToken.cancel()
```
pelikhan: clear runtoken afterawards

webapp/src/app.tsx
samelhusseini: Yep, will do.

webapp/src/app.tsx
samelhusseini: Done

Issue #3798, opened 2018-01-31
simulator does not restart after neopixel error


nielsswinkels:
If you create a neopixel, stored in a variable with a custom name, and then drag out a neopixel block, this will crash the simulator. This is because those blocks have the default variable name selected. That's understandable so far. But if you then change the neopixelblock to use the right variable name, it does not restart the simulator, it stays in crashed mode. That's confusing, because at that point the code is correct.

![microbit_neopixel_bug](https://user-images.githubusercontent.com/5718354/35630355-808986ec-06a1-11e8-885b-71d3646a8561.gif)

nielsswinkels:
Any updates about this? I'm writing a guide on this at the moment, and right now I'm considering using the default variable name to not run into this error. But I think giving it a proper variable name is a much better style.

abchatra:
This is fixed in alpha

abchatra:
This is not fixed.

PR #4956, opened 2018-10-31
Suspend simulator

samelhusseini:
This PR does three things: 
- Stop the simulator when we invalidate (no need to run code that is invalidated) @pelikhan FYI
- Adds another state for the simulator called suspended, suspended is more or less the same as stop except it differentiates between a user action stop and a code (compiler / runtime error) stop. ie: if we fix the code, we will run again if the simulator is in the suspended state. Unlike if the user hits the stop button, we will continue to be stopped regardless of what changes.

Explanation:
When we hit an exception, we stop the simulator (which is equivalent to a user hitting the stop button) this means that we don't restart the simulator when a change is made and the "exception" is fixed. 

This change creates a new simulator state "Suspended" which means that the simulator appears as stopped "grayscale class" but is actually in a unique state where it will restart if a change is made. 

Another case that this tries to address is when the page is invisible it stops the simulator and restarts it when we are active again. If there is a compiler error, it won't run, but it also will set the state as "stopped" which is problematic as the user never explicitly stopped the simulator. 

Tested: 
Running -> Compiler error -> Fix error -> Running
Running -> Runtime error -> Fix error -> Running
Running -> Switch page -> Running
Running -> Compiler error -> Switch page -> Running
Running -> Runtime error -> Switch page -> Running
Stopped -> Any change -> Stopped
Stopped -> Switch page -> Stopped

Runtime error test: 
```
let array: number[] = null
array[0]
```

Compiler error test: 
```
let x = 0
x = ""
```

### DIFF
--- a/pxtsim/simdriver.ts
+++ b/pxtsim/simdriver.ts
@@ -182,7 +182,8 @@ namespace pxsim {
             }
         }
 
-        private suspend() {
+        public suspend() {
+            this.postMessage({ type: 'stop' });
             this.setState(SimulatorState.Suspended);
 
             let frames = this.container.getElementsByTagName("iframe");
--- a/pxtsim/simdriver.ts
+++ b/pxtsim/simdriver.ts
@@ -94,7 +94,7 @@ namespace pxsim {
         private setState(state: SimulatorState) {
             if (this.state != state) {
                 this.state = state;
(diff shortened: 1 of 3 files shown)


pelikhan:
Good overdue fix!

samelhusseini:
Fixes https://github.com/Microsoft/pxt/issues/3798

PR #4957, opened 2018-11-01
Default edit button in v0 will point to live

abchatra:
Edit button in the share page is pointing to v0. If I have shared a page a while ago, I still would like to go to the latest instead of v0. There is a link to go to the old editor if needed.  I see most of the traffic to v0 is through this edit button. Changing the default behavior.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1445,7 +1445,7 @@ export class ProjectView
         pxt.tickEvent("sandbox.openfulleditor");
         Util.assert(pxt.shell.isSandboxMode());
 
-        let editUrl = pxt.appTarget.appTheme.embedUrl;
+        let editUrl = pxt.appTarget.appTheme.shareUrl;
         if (!/\/$/.test(editUrl)) editUrl += '/';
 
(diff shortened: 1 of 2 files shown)


abchatra:
Also I added a second change to this. Pop out from the sandbox view also points to live now. Only way to go to v0 is to click on the Edit in old editor link.

webapp/src/app.tsx
```diff
@@ -1445,7 +1445,7 @@ export class ProjectView
         pxt.tickEvent("sandbox.openfulleditor");
         Util.assert(pxt.shell.isSandboxMode());
 
-        let editUrl = pxt.appTarget.appTheme.embedUrl;
+        let editUrl = pxt.appTarget.appTheme.shareUrl;
```
pelikhan: Maybe add || homeUrl in case shareUrl is missing.

Looks good - maybe we should give a round of UI Redesign to the share page :)

abchatra: We have shareURL in v0 which only microbit points. This is specific to v0 and I will keep this way for now.

PR #4964, opened 2018-11-02
Only suspend if we're running.

samelhusseini:
Only suspend the simulator if we're running.
We don't want to end up in a suspended state if we're stopped or paused.

### DIFF
--- a/webapp/src/simulator.ts
+++ b/webapp/src/simulator.ts
@@ -234,7 +234,11 @@ export function makeDirty() { // running outdated code
     pxsim.U.addClass(driver.container, getInvalidatedClass());
     dirty = true;
 
-    // No need to continue running if we're dirty
+    // We suspend the simulator here to stop it from running without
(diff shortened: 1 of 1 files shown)


webapp/src/simulator.ts
```diff
@@ -235,7 +235,7 @@ export function makeDirty() { // running outdated code
     dirty = true;
 
     // No need to continue running if we're dirty
-    driver.suspend();
+    if (driver.state == pxsim.SimulatorState.Running) driver.suspend();
```
pelikhan: This should be handled inside the driver class.

samelhusseini: I think this is the right place for this check, the simdriver should be able to switch to the suspended state from any other state at the demand of the editor. However, in this case, the editor is invalidating the code (change has been made) and we only want to suspend the simulator if it's running. ie: don't want to go from Stopped -> Suspended just because we invalidated code.

PR #4966, opened 2018-11-02
reduce card size in scriptsearch for mobile

pelikhan:
- [ ] merged packagedialog and scriptsearch.less
- [ ] same size as home screen

localhost_3232_index html iphone 6_7_8 1


samelhusseini:
Nice

PR #4971, opened 2018-11-02
Disable import and save in iOS & Mac in browser view

abchatra:
Disable explicit import and save of files in Mac and iOs, enabled now through the inGame=1 for Minecraft


abchatra:
Fixes https://github.com/Microsoft/pxt-minecraft/issues/672

PR #4972, opened 2018-11-02
add option to hide "go back" button

pelikhan:
Hide the "go back" option in scriptsearch when the dialog is waiting on a promise resolve. This is used when a user picks up a board after opening a script or a tutorial.


webapp/src/scriptsearch.tsx
```diff
@@ -388,6 +388,7 @@ export class ScriptSearch extends data.Component<ISettingsProps, ScriptSearchSta
         return (
             <sui.Modal isOpen={this.state.visible} dimmer={true}
                 className="searchdialog" size="fullscreen"
+                hideGoBack={!!this.state.resolve}
```
samelhusseini: Can we not use the closeIcon={false} to indicate that we don't want a go back button?

pelikhan: sure can

PR #4974, opened 2018-11-04
Fix translate note names

sgk:
After upgrading from v0 to v1, the note names do not translate. This change is to fix this and shows translated note names as it is in v0.

v1 (before this change):
image

v0:
image

### DIFF
--- a/pxtblocks/fields/field_note.ts
+++ b/pxtblocks/fields/field_note.ts
@@ -74,57 +74,57 @@ namespace pxtblockly {
     }
 
     const Notes: {[key: number]: NoteData} = {
-        28: {name: "C", prefixedName: "Low C", freq: 131},
-        29: {name: "C#", prefixedName: "Low C#", freq: 139},
-        30: {name: "D", prefixedName: "Low D", freq: 147},
-        31: {name: "D#", prefixedName: "Low D#", freq: 156},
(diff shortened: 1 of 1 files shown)


pxtblocks/fields/field_note.ts
```diff
@@ -74,57 +74,57 @@ namespace pxtblockly {
...
-        67: {name: "D#", prefixedName: "High D#", freq: 1245},
-        68: {name: "E", prefixedName: "High E", freq: 1319},
-        69: {name: "F", prefixedName: "High F", freq: 1397},
-        70: {name: "F#", prefixedName: "High F#", freq: 1478},
-        71: {name: "G", prefixedName: "High G", freq: 1568},
-        72: {name: "G#", prefixedName: "High G#", freq: 1661},
-        73: {name: "A", prefixedName: "High A", freq: 1760},
-        74: {name: "A#", prefixedName: "High A#", freq: 1865},
-        75: {name: "B", prefixedName: "High B", freq: 1976}
+        28: {name: lf("C"), prefixedName: lf("Low C"), freq: 131},
```
pelikhan: This map needs to be cached lazily since we might not have loaded the translations before loading it

sgk:
I will try using rlf() later.

sgk:
@pelikhan Here it is. It's simple. Sorry for the previous commit.

PR #4980, opened 2018-11-07
Support for a group help button and callback

samelhusseini:
This PR adds support for a group help button that calls a callback when a user clicks on it. 

It also adds support for a ``workspaceevent`` which currently only passed up the chain in an iframeworkspace. 
The two events supported: 
- pxt.events.UIEvent for the groupHelpCallback
- pxt.events.CreateEvent for when blocks are created. 

PR also pulls in a Blockly build for support for the help icon with PRs: 
- https://github.com/Microsoft/pxt-blockly/pull/163
- https://github.com/Microsoft/pxt-blockly/pull/166

https://github.com/playi/pxt-wonder/issues/338 and https://github.com/playi/pxt-wonder/issues/337


pxteditor/events.ts
```diff
@@ -0,0 +1,46 @@
+namespace pxt.events {
+
+    export type EditorType = 'blocks' | 'js';
```
pelikhan: to stay consistent, use file extension, ``.ts``

pxteditor/events.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    export interface Event {
+        type: string;
+        editor: EditorType;
+    }
+
+    export abstract class AbstractEvent implements Event {
```
pelikhan: do we need classes, we can typically define these messages as bare interface

pxtlib/service.ts
```diff
@@ -845,6 +846,14 @@ namespace ts.pxtc {
                 res.groupIcons = undefined;
             }
         }
+        if (res.groupHelp) {
+            try {
+                res.groupHelp = JSON.parse(res.groupHelp as any);
+            }
+            catch (e) {
+                res.groupHelp = undefined;
```
pelikhan: report issue

webapp/src/blocks.tsx
```diff
@@ -378,6 +379,7 @@ export class Editor extends toolboxeditor.ToolboxEditor {
                 if (ev.xml.tagName == 'SHADOW')
                     this.cleanUpShadowBlocks();
                 this.parent.setState({ hideEditorFloats: false });
+                workspace.fireEvent(new pxt.events.CreateEvent('blocks', blockId));
```
pelikhan: fireEvent is optional?

pxteditor/events.ts
samelhusseini: Sounds good.

pxteditor/events.ts
samelhusseini: Switching to interfaces

webapp/src/blocks.tsx
samelhusseini: Optional in impl not in workspace

pxteditor/events.ts
```diff
@@ -0,0 +1,20 @@
+namespace pxt.events {
```
pelikhan: can you put this under ``pxt.editor.events`` ? I think all editor stuff is there.

samelhusseini: sure.

pxteditor/events.ts
```diff
@@ -0,0 +1,20 @@
...
+
+    export type EditorType = 'blocks' | 'ts';
+
+    export interface PXTEvent {
```
pelikhan: don't use the product name here. we're already under the pxt namespace. ``HostEvent``?

samelhusseini: Sorted.

PR #4982, opened 2018-11-07
force csv, txt, hex to download

pelikhan:
we need to add a "download" attribute to force .hex file direct downloads. This needs to go in the backend asap.


pelikhan:
Does not seem to work for different domain files.

PR #4986, opened 2018-11-08
monitor storage quota and trigger translation db cleanup as needed

pelikhan:
The translation cache database is used to cache translations. It can grow without bounds and eventually the user will run out of space. With this change we do nuke the translations cache db in these situations:

- [x] if we detect a QuotaExceededError in any transaction
- [x] after any storage operation, we trigger a throttle quota check (10second thorggle). If the storage is > 50% or > 100Mb, we clean up.

Also moved all the indexed DB out of common-utils.ts.

Run ``ts.BrowserUtils.stressTranslationsAsync()`` to store tons of junk.

Tested on Chrome, FF, Edge. storage api not available in Edge, IE.


pxtlib/browserutils.ts
```diff
@@ -527,4 +527,330 @@ namespace pxt.BrowserUtils {
...
+                cursor.onerror = () => this.errorHandler(cursor.error, "getAll", reject);
+            });
+        }
+
+        public setAsync(storeName: string, data: any): Promise<void> {
+            return new Promise((resolve, reject) => {
+                const store = this.getObjectStore(storeName, "readwrite");
+                let request: IDBRequest;
+
+                if (typeof data.id !== "undefined" && data.id !== null) {
```
riknoll: data.id == null

Don't need to use typeof here

pelikhan: Just moving that code.

PR #4988, opened 2018-11-08
handle409 specially

pelikhan:
Ignore 409 errors coming from PouchDb.


riknoll:
Looks fine, but what's the motivation for this?

pelikhan:
When we try to save a project, we might fail because we're out of space. In that case, we try to regain space by clearing the translation cache.
We get 409 when downloading .hex files.

PR #4991, opened 2018-11-09
We don't need to trigger an undo when the editor disables a block

samelhusseini:
PXT is in charge of disabling, we shouldn't trigger record undo when disabling a block. 

PR is into stable, will cherry-pick to master once it's in.


abchatra:
I have created a separate stable for wonder stable4.5. It is possible there might be other fixes for wonder and I would like to keep stable4.4 just for microbit.

PR #4993, opened 2018-11-10
Adding test suite to detect breaking changes to blocks

riknoll:
TODO:

 - [ ] Support non-bundled packages

How the tests work:

1. The PXT CLI checks for `.blocks` files in libs and common-packages and creates a JS file with all of the test cases
2. The test runner loads that file and converts each block XML into a Blockly workspace and back (Blockly removes all fields and values it doesn't recognize)
3. Then it compares the two XML tress and throws on any differences

You should get a nice error output for failed tests that tells you exactly which node is mismatched. Note that this does *not* run any upgrade rules on the XML before testing. Putting DO NOT MERGE on this PR because I don't want everything to break over the weekend.


riknoll:
This PR is ready to go

PR #5002, opened 2018-11-13
Add AudioContextManager.playInstructionsAsync

mmoskal:
This allows for multi-channel sounds in the sim

### DIFF
--- a/cli/buildengine.ts
+++ b/cli/buildengine.ts
@@ -519,7 +519,7 @@ export function buildDalConst(buildEngine: BuildEngine, mainPkg: pxt.MainPackage
         for (let fn of files) {
             let v = extractConstants(fn, fc[fn], true)
             if (v) {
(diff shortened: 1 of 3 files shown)


riknoll:
I think this needs to be cancellable, or maybe you could just check the runtime id in the loop to see if it has changed?

mmoskal:
Well, it's only used to play a single note (the different entries are for attack, decay, etc). I guess there's risk someone will play a really long note.

riknoll:
We've gotten bugs before of people doing that with the old audio context stuff. Would be good to add a `killAllAudio()` function or something that we can call when the simulator is paused or stopped.

mmoskal:
Yeah, can do. IE is not supported (sound playing should be equivalent to pause)

pelikhan:
As long as it does not crash

PR #5006, opened 2018-11-14
Make the serial back button a normal button

samelhusseini:
Today the serial back button is a custom button that is hard to maintain across targets with light and dark themes. There's not much value in that transparent background, so let's just make it a normal button. 
Here's what it currently looks like, and below is the proposal: 
screen shot 2018-11-13 at 7 38 40 pm

This fixes the following issues: 
Fixes https://github.com/Microsoft/pxt-adafruit/issues/843
Fixes https://github.com/Microsoft/pxt-adafruit/issues/724

Here's what it looks like on microbit, adafruit and high contrast: 
screen shot 2018-11-13 at 7 37 12 pm
screen shot 2018-11-13 at 7 35 27 pm
screen shot 2018-11-13 at 7 35 22 pm

### DIFF
--- a/webapp/src/serial.tsx
+++ b/webapp/src/serial.tsx
@@ -422,7 +422,7 @@ export class Editor extends srceditor.Editor {
                 <div id="serialHeader" className="ui serialHeader">
                     <div className="leftHeaderWrapper">
                         <div className="leftHeader">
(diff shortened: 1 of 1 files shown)


samelhusseini:
Nope, I still use the back-button for the extensions dialog.

PR #5008, opened 2018-11-14
Cherry-pick Android note fix and HSV color support for namespace

abchatra:
Fix note picker editing on Android (#4998) 
https://github.com/Microsoft/pxt/commit/04eb86e9f12e561f63f6b9cd17fc2cea01bad018 

And 
Support HSV colors (#4996) 
https://github.com/Microsoft/pxt/commit/3c0069146f604bafca4ab9f7072d5f267b1f7c00


samelhusseini:
Looks good but we should also take this fix: https://github.com/Microsoft/pxt-blockly/pull/170
Which fixes the same issue as the note picker but for the pause block. I can add that to this PR.

PR #5016, opened 2018-11-16
Don't log Blockly events when converting from XML to DOM

samelhusseini:
Blockly logs a BlockCreate event for Top level blocks only (on start, events, forever, etc) when it's converting from XML to DOM. 
I'm not 100% sure why, I'll ask when I get a chance, but I'm sure they have a good reason for it. 

There's no reason for us to have Blockly events turned on as we are converting from xml to dom. ie: when we're loading a project (whether it's a previous script or a shared script), and when we're converting from our decompiled xml to blocks (ie: converting from TS to Blocks).

I've turned off events here when we're doing that conversion.


samelhusseini:
this is better https://github.com/Microsoft/pxt/pull/5015#pullrequestreview-175886538

PR #5026, opened 2018-11-19
Script manager!

samelhusseini:
Adds support for a script manager with two entry points: 
- Clicking on the My projects header
- Clicking the last card of the My projects cards. Cards are limited to 10 on the home screen. 

Features:
- The script manager lets you delete one or a group of projects.
- Duplicate a specific project. We choose a duplicate name for you (eg: Untitled #2) 
- Edit a specific project. 
- Search projects (as you type in the search field, using the worker to do the search as it already has Fuse loaded)
- Sort by name or last modified
- Select multiple projects by using shift and ctrl / cmd (depending on what OS you're on)
- Grid and list view
- Responsive UI, no search in mobile view
- Confirm delete for single and multiple files
- Tooltips for the buttons

The feature is behind the "scriptManager" flag but the hope is to enable that by default and remove the option to disable it in the near future.

Entry point 1: 
screen shot 2018-11-19 at 2 32 31 pm

Entry point 2 (if you have more than 10 projects: 
screen shot 2018-11-19 at 2 34 21 pm

Multi select: using select, shift and select again. 
[screen shot 2018-11-19 at 2 32 45 pm](https://user-images.githubusercontent.com/16690124/48739281-55e89700-ec08-11e8-89fc-63fba7f77e7c.png)

Sort by name or last modified: 
screen shot 2018-11-19 at 2 32 51 pm

List view: 
screen shot 2018-11-19 at 2 32 57 pm
(Sort by selecting the headings)

Specific select, by using ctrl / cmd (os dependent) and click: 
screen shot 2018-11-19 at 2 33 04 pm

Duplicate a project: 
screen shot 2018-11-19 at 2 33 13 pm

Toolbar for actions you can take on the selected project(s): 
screen shot 2018-11-19 at 2 33 18 pm

Project search:
screen shot 2018-11-19 at 2 33 24 pm

No projects: 
screen shot 2018-11-19 at 2 43 55 pm

Delete confirmation: 
screen shot 2018-11-19 at 2 46 35 pm

Strings to verify:
- [x] My Projects (Not New)
- [ ] Name
- [ ] Last Modified
- [ ] Sort by Name {0}
- [ ] Sort by Last Modified {0}
- [ ] ascending
- [ ] descending
- [ ] Select all projects
- [ ] De-select all projects
- [ ] Sort by dropdown
- [ ] Switch sort order to {0}
- [ ] It's empty in here
- [ ] Go back to create a new project
- [ ] List view
- [ ] Grid view
- [x] Delete Project (Not new)
- [ ] Duplicate Project
- [ ] Edit Project
- [x] Search... (Not new)
- [x] {0} result matching '{1}' (Not new)
- [ ] Would you like to delete {0} projects? (For delete confirmation dialog)
- [ ] Choose a new name for your project
- [x] Enter your project name here (Not new)

Note: this introduces the concept of a baseTheme. which can be light or dark, possibly hc (High contrast) in the future. This is an incremental change that will allow us to adapt the look of the editor based on the baseTheme of the target, instead of having all these "invertedMenu / invertedMonaco" all over the place. 

Dark Theme: 
screen shot 2018-11-19 at 2 56 18 pm

Browser Testing: 
- [x] Chrome
- [x] IE
- [x] Edge
- [x] Firefox
- [x] Safari
- [x] IOS
- [x] Android Chrome
- [x] Accessibility Testing

I've also cleaned up search into its own new component in an effort to start moving more of React stuff into their own component.


samelhusseini:
https://github.com/Microsoft/pxt-microbit/issues/1305

samelhusseini:
Custom build to try out these changes: 
- Adafruit: https://makecode.adafruit.com/app/21cc09a083a40740f47cd7abf1a7d10922de627a-b1b1b974d7
- Arcade: https://arcade.makecode.com/app/6138e0ed38e9ff018b548517b7db28bcfaeffbb9-e64628ecd3

pelikhan:
List select dot does not have the "hand" icon.

pelikhan:
IMO, the "My Projects >" should get bolder on hover to highlight that it is a clickeable link.

samelhusseini:
Just FYI, and this is something I noticed in my testing. 
On the home screen we sort by Recent use (ie: opened the project counts as using it). 
But the two options we provide in the script manager which I think are more useful are Last Modified and Name. 

We should keep home screen as recent use as that makes sense. but should we provide a third option in the script manager to sort by Last opened?

pelikhan:
Default sorting should be the same as home screen for consistency.

webapp/src/workspace.ts
```diff
@@ -340,27 +341,33 @@ export function installAsync(h0: InstallHeader, text: ScriptText) {
...
     h.id = U.guidGen()
-    let names = U.toDictionary(allScripts, e => e.header.name)
-    let n = 2
-    while (names.hasOwnProperty(h.name + " #" + n))
-        n++
-    h.name += " #" + n
-    let cfg = JSON.parse(text[pxt.CONFIG_NAME]) as pxt.PackageConfig
-    cfg.name = h.name
-    text[pxt.CONFIG_NAME] = JSON.stringify(cfg, null, 4)
+    if (rename) {
```
abchatra: If rename is false and there is a collision of name?

abchatra:
Looks great

webapp/src/workspace.ts
samelhusseini: id is refreshed, name collisions are allowed.

samelhusseini: hence how you can have so many Untitled projects.

PR #5027, opened 2018-11-20
Fixed inline blocks coloring to reflect updated/overwritten colors

ChaseMor:
Fixes Microsoft/pxt-arcade#355

This removes images color from the toolbox colors. From what I can tell, it shouldn't be there since it isn't a default category, doesn't even have an icon, and is always defined elsewhere if used (micro:bit and arcade). Also, having it in here and defining it elsewhere was creating 2 conflicting css classes for inline blocks

This also changes the default category blocks to check what color they should be instead of just setting their value to their default color.


pelikhan:
Looks great!

PR #5039, opened 2018-11-26
Empty array block

abchatra:
image
They are pretty low in weight and come at the bottom.


pelikhan:
What happens when you assign it to a variable?

PR #5049, opened 2018-11-29
Fix for printing local packages

pelikhan:
Fix for https://github.com/Microsoft/pxt-microbit/issues/1683


abchatra:
stable4.4 is for point fixes. Merge these to master. We are not going to release stable 4.4 unless there is a need for hotfix.

pelikhan:
/beta is still pointing to stable4.4 so my understanding is that the fix is going there.

PR #5051, opened 2018-11-30
Don't use cache when serving locally

jwunderl:
Sets the timeout for the cache to 1 ms when serving locally (e.g. pxt serve). Currently hard to update things as some elements like the front page get cached and you have to change the port every update.

(Looks like it still serves the cached content while it is requesting the new one, so to immediately show updates you need to refresh twice when serving locally)


pxtlib/emitter/cloud.ts
```diff
@@ -90,7 +90,7 @@ namespace pxt.Cloud {
     }
 
     // 1h check on markdown content
-    const MARKDOWN_EXPIRATION = 1 * 60 * 60 * 1000;
+    const MARKDOWN_EXPIRATION = Cloud.isLocalHost() ? 1 : 1 * 60 * 60 * 1000;
```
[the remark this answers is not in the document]
jwunderl: Don't know a way to test electron off hand, but this check should handle that case then I believe?

```typescript
Cloud.isLocalHost() && !(window as any).pxtElectron
```

pelikhan:
In electron, you also run localhost but you want caching.

pxtlib/emitter/cloud.ts
```diff
@@ -89,8 +89,9 @@ namespace pxt.Cloud {
         })
     }
 
+    const isElectron = !!(window as any).pxtElectron || !!(window as any).ipcRenderer;
```
pelikhan: there should be a help on for this

jwunderl: It's in webapp/src/electron currently which is not accessible in here, should I move that around somewhere else? Would need more testing I would assume

pelikhan: should be in browsersutil. let's minize casting window to any (and window might be missing in nodejs)

jwunderl: Sounds good.

pxtlib/emitter/cloud.ts
```diff
@@ -90,7 +90,7 @@ namespace pxt.Cloud {
     }
 
     // 1h check on markdown content
-    const MARKDOWN_EXPIRATION = 1 * 60 * 60 * 1000;
+    const MARKDOWN_EXPIRATION = Cloud.isLocalHost() && !pxt.BrowserUtils.isElectron() ? 1 : 1 * 60 * 60 * 1000;
```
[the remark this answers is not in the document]
jwunderl: Would that live in browserutils too or just in cloud?

pelikhan:
BrowserUtil, near "isLocalHost"

jwunderl:
> BrowserUtil, near "isLocalHost"

that lives in cloud right now, move both then?

PR #5056, opened 2018-12-01
Pin lookup fix

pelikhan:
Use the gpiomap to lookup pins in breadboarding.

### DIFF
--- a/pxtsim/instructions.ts
+++ b/pxtsim/instructions.ts
@@ -123,9 +123,10 @@ namespace pxsim.instructions {
         crocClips?: boolean
     };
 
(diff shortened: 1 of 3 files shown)


samelhusseini:
Did you intend to make boardDef a required field?
That would require target changes namely here: https://github.com/Microsoft/pxt-adafruit/blob/master/sim/dalboard.ts#L176 to pass in the boardDef if it's now required.

pelikhan:
Yes it is mandatory right now but I could make it more resilient if missing.

pelikhan:
It is there, just needs to be passed along.

samelhusseini:
Yeah, its an easy fix in the target, just checking whether it's required now.

Issue #4534, opened 2018-07-06
Tab opens instead of page dropping to defined header


ashaltu:
When creating a .md file and referencing a header later on in the file, a new tab is opened and then it goes down to the header instead of automatically scrolling down on the same tab

![openedtab](https://user-images.githubusercontent.com/10238372/42398403-94175480-811e-11e8-9f06-a89d742e5cba.gif)

DoctaCloak:
Working on this ticket.

DoctaCloak:
Any idea where I should be looking in the codebase for this?  @ashaltu

riknoll:
@DoctaCloak I think you want `pxtlib/docsrender.ts`

DoctaCloak:
Thank you!  @riknoll  😃

PR #5059, opened 2018-12-04
Links that start with '#' appear on the same page instead of a new tab

ChaseMor:
Fixes Microsoft/pxt#4534

This makes it so that if link in the docs starts with a '#', meaning it is an anchor link to the same page, it will open in the same tab instead of a separate tab.

### DIFF
--- a/pxtlib/docsrender.ts
+++ b/pxtlib/docsrender.ts
@@ -475,7 +475,7 @@ namespace pxt.docs {
         setupRenderer(renderer);
         const linkRenderer = renderer.link;
         renderer.link = function (href: string, title: string, text: string) {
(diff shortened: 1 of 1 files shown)


pxtlib/docsrender.ts
```diff
@@ -475,7 +475,7 @@ namespace pxt.docs {
         setupRenderer(renderer);
         const linkRenderer = renderer.link;
         renderer.link = function (href: string, title: string, text: string) {
-            const relative = href.indexOf('/') == 0;
+            const relative = href.charAt(0) === '/' || href.charAt(0) === '#';
```
pelikhan: ``href.test(/^[\/#]/)``

PR #5062, opened 2018-12-05
don't restart sim while compiling

pelikhan:
In arcade, clicking download might trigger a reload and compilation. Simulator should not be running in the background.


samelhusseini:
Possible issue that I'd be concerned about is what happens if: 
- user makes a change (invalidates the sim), ie: it needs to be restarted
- user compiles before throttle timeout, ie: it won't restart
- no action is done to restart the sim, sim stays in a state where it's invalidated but not restarted.

pelikhan:
Let's take this to the UI meeting. Currently the restarts in arcade are quite annoying.

PR #5067, opened 2018-12-06
Name the project before sharing

samelhusseini:
Share dialog allows you to name the project before sharing: 

screen shot 2018-12-06 at 9 47 42 am

and the exit dialog: 

screen shot 2018-12-06 at 9 48 01 am

Note: this only shows if the user has an untitled project, in both cases.


samelhusseini:
Two steps ahead @pelikhan. see: https://github.com/Microsoft/pxt/pull/5067/files#diff-ffda22a895f92240f9d15ba2e95be683R840

PR #5072, opened 2018-12-07
Enum isHash

pelikhan:
Adds a new enum modifier flag, ``EnumIsHash`` that forces MakeCode to generate the enum values as the hash16 of the name. This is particularly useful for radio-broadcast or any other named based broadcast API where users might want to write different programs talking the same protocol.


riknoll:
LGTM. I guess this one doesn't really need tests

pelikhan:
I'm not completely sure how the .blocks files are supposed to be generated.

riknoll:
It gets printed to console with `?dbg=1`

I wasn't hinting at anything, there really isn't anything to test here :smile:

PR #5073, opened 2018-12-07
Update the roll-out section

mmoskal:
@abchatra we have github in all editors now, right?


abchatra:
Probably chibi doesn't have it.

PR #5079, opened 2018-12-10
remove support for gists in cli

pelikhan:
It was a nice experiement but it hasn't materialized as useful.


samelhusseini:
That was quick!

samelhusseini:
Extra reference, Github deprecated anonymous gist sharing earlier this year: https://blog.github.com/2018-02-18-deprecation-notice-removing-anonymous-gist-creation/

PR #5088, opened 2018-12-12
Monaco field editors

riknoll:
Adds support for field editors in the Monaco editor (currently just the sprite editor). Targets can enable this feature by adding a list of field editors they support in `pxtarget.json`.

monaco_field_editor_2

It also works well when starting from typescript:

monaco_field_editor

Field editors get placed based on a regex (I tried doing it with the compiler but it was much too laggy for typing). When you open a TS file, all ranges that are matched by the regex will be automatically folded and get an icon next to them in the margin. Clicking on the icon shows the field editor and clicking it again will close it. Anything that closes the field editor "commits" the change; there is no cancel but undo/redo work great. Folding is only done on file open, but you can also manually fold/unfold.

This PR also refactors the sprite editor into pxtlib so it can be shared between Monaco and blocks. The close button shown in the GIF is also shown in blocks.

Known bugs:
* The palette icon doesn't display properly in IE

Future improvements:
* Should be context-sensitive, right now we can't change the size defaults in the sprite editor like we do for tilemaps
* Maybe targets should be able to add these?

Tested in IE, Edge, Chrome, and Firefox.


localtypings/pxtarget.d.ts
```diff
@@ -298,6 +298,7 @@ declare namespace pxt {
         transparentEditorToolbar?: boolean; // make the editor toolbar float with a transparent background
         hideProjectRename?: boolean; // Temporary flag until we figure out a better way to show the name
         addNewTypeScriptFile?: boolean; // when enabled, the [+] explorer button asks for file name, instead of using "custom.ts"
+        monacoFieldEditors?: string[]; // A list of field editors to show in monaco. Currently only "image-editor" is supported
```
samelhusseini: Move further up, there's a temporary list there.

pxteditor/monaco-fields/field_sprite.ts
```diff
@@ -0,0 +1,89 @@
...
+        private rejecter: (err?: any) => void;
+
+        protected editor: pxtsprite.SpriteEditor;
+        protected editrange: monaco.Range;
+
+        getId() {
+            return fieldEditorId;
+        }
+
+        showEditorAsync(editrange: monaco.Range, host: MonacoFieldEditorHost): Promise<TextEdit> {
```
samelhusseini: It would be good to see us converge the Blockly and Monaco field editors into one interface so we can implement field editors irrespective of where they show up. 

Right now with this current implementation it seems you only allowing field editors to write (set) their value when they close. Here are some examples of field editors that write their value a number of times: 
- note picker
- slider
- turn ratio

pxtlib/browserutils.ts
```diff
@@ -851,4 +851,34 @@ namespace pxt.BrowserUtils {
...
+    export interface IPointerEvents {
+        up: string,
+        down: string[],
+        move: string,
+        enter: string,
+        leave: string
+    }
+
+    export const pointerEvents: IPointerEvents = hasPointerEvents() ? {
+        up: "pointerup",
```
samelhusseini: This is fine for now, but let's look into having a util library that all these utils go to. 
webapp shouldn't need to rely on pxsim just for utils
pxtblocks shouldn't need to rely on pxtblocks just for utils..

you get the idea

webapp/src/monaco.tsx
```diff
@@ -1483,8 +1712,163 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
+        });
+    }
+}
+
+function rangeToSelection(range: monaco.IRange): monaco.Selection {
+    return new monaco.Selection(range.startLineNumber, range.startColumn, range.endLineNumber, range.endColumn);
+}
+
+class ViewZoneEditorHost implements pxt.editor.MonacoFieldEditorHost, monaco.editor.IViewZone {
+    domNode: HTMLDivElement;
```
samelhusseini: Can this go in a separate file. I'd like to avoid having these huge files with a bunch of different components.

pxteditor/monaco-fields/field_sprite.ts
riknoll: The current API was designed to be transactional to prevent code from moving around while the field editor is open. I'll mark this as something to revisit after the holidays (when everyone is back), but we'll merge as-is and tweak it once we have some feedback

riknoll:
Also this fixes https://github.com/Microsoft/pxt-arcade/issues/332

PR #5094, opened 2018-12-13
when dragging create sprite, rev variable number

pelikhan:
When dragging a "create sprite" block variable, automatically rev the variable name so that it generates a new variable in the workspace. This is a simple solution to avoid aliasing of variables in a block program.
slowjs


Jaqster:
Hey, can we talk about this before implementing?  I'm not saying this is a bad idea, but I want to think through this more.  We've now got 3 different ways to create variables - I just don't want users to do stuff unintentionally or be confused.  We got feedback just today from NYC that having default or automatic variables appear was bad.

pelikhan:
When users drag a "create sprite" block into the workspace, they should get a unused variable for it. Reusing existing variables leads to leaking sprites in most cases.

The NYC schools were complaining that the variable flyout was populate with "item" which did not make sense in many programs.

PR #5097, opened 2018-12-13
Introduces simulator "starting" state to prevent button smashing

pelikhan:
Start/stop/reset buttons are disabled while the simulator is starting to prevent button smashing or double start bugs.

This is cirtical for arcade where we have a proeminent reset button and startup can be slow.

Fix for https://github.com/Microsoft/pxt-arcade/issues/410


samelhusseini:
What happens if the simulator is starting and we change tabs? and then go back?

pelikhan:
We are fine there, visilibity change will restart the sim.

PR #5105, opened 2018-12-14
Field position to support sprites

pelikhan:
Needs to be in pxt as sprite apis are in common.
setpos


abchatra:
What happens when simulator is hidden?

abchatra:
Does it work on IE11?

PR #5109, opened 2018-12-14
Fixed centering for all closing icons

ChaseMor:
Does the same fix as Microsoft/pxt#5102, but for all closing icons.

Also makes the three icons, (home page, modal, and sprite editor) more consistent with each other with the same heights/widths and border-shadows.

### DIFF
--- a/theme/home.less
+++ b/theme/home.less
@@ -218,22 +218,21 @@
             z-index: 1;
         }
         >.close .remove.icon {
-            font-size: 34px;
+            font-size: 2.15rem;
+            line-height: 2.15rem;
             cursor: pointer;
-            width: 2.0rem;
+            width: 2.15rem;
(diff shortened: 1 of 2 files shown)


samelhusseini:
Could you add screenshots for the areas you tested please.

ChaseMor:
Here is the normal closing icon. This screenshot was taken from the closing icon from the _Eat The Doughnut_ tutorial. There wasn't really an issue with centering, but the icon was smaller than the others. Previously, the icon font-size was larger than the actual icon which is why the border size increases, but this makes it more consistent with the others.
closing-icon
Here is the closing icon for modals. This one had the same issues as the sprite-editor. Also changed the box-shadow spread to be more consistent.
closing-icon

samelhusseini:
Wonderful, thanks!

PR #5113, opened 2018-12-17
Use the triple-toggle view in all sandbox embeds

riknoll:
triple_embed

I know that GIF shows the full screen button in the sim toolbar, but after I recorded it I removed that button and was too lazy to redo it. It's redundant in the triple-toggle mode.

Tested in Chrome, Edge, IE, and Firefox.

All of the tests were done with arcade, not sure if that matters.

### DIFF
--- a/theme/common.less
+++ b/theme/common.less
@@ -619,6 +619,26 @@ p.ui.font.small {
 
 /* Mobile + Tablet */
 @media only screen and (max-width: @largestTabletScreen) {
+    #root.sandbox .active.item.sim-menuitem ~ .toggle {
+        margin-left: 0 !important;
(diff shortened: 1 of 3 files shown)


theme/common.less
```diff
@@ -944,118 +964,139 @@ p.ui.font.small {
     }
 }
 
-/* Mobile, Tablet AND thin screen */
-@media only screen and (max-width: @largestTabletScreen) and (max-height: @thinEditorBreakpoint) {
-    /* Mobile embedded view */
-    .sandbox #simulators{
+.simView #filelist {
```
pelikhan: should this go into its own .less file?

pelikhan:
We need to check that this change is compatible with the existing script pages and that other editors are not impacted.

PR #5119, opened 2018-12-19
Early semantic init

pelikhan:
* [x] move semantic init before loading PXT compilers. In Arcade, compilation takes a ton of time and the page is in limbo for 10seconds.
* [x] always scroll active header into view
* [x] consistently use /doccdn/ for semantic, /blb/ for other resources

### DIFF
--- a/docfiles/docs.js
+++ b/docfiles/docs.js
@@ -53,6 +53,15 @@ function searchSubmit(form) {
     if (pxt && pxt.tickEvent) pxt.tickEvent("docs.search", { 'source': form.id }, { interactiveConsent: true })
 }
 
+function scrollActiveHeaderIntoView() {
+    var activeHeaders = document.getElementsByClassName("header active");
+    for (var i = 0; i < activeHeaders.length; ++i) {
+        var activeHeader = activeHeaders.item(i);
+        if (activeHeader.scrollIntoView)
+            activeHeader.scrollIntoView()
+    }
+}
+
 function setupSidebar() {
     $('#togglesidebar').on('keydown', handleEnterKey);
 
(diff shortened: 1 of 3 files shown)


pelikhan:
@samelhusseini @riknoll @guillaumejenkins can i get a PR?

webapp/public/embed.js
```diff
@@ -7,14 +7,16 @@
...
     ]
 
     if (typeof jQuery == "undefined")
-        scripts.unshift("/blb/jquery.js")
+        scripts.unshift("/doccdn/jquery.js")
+    if (typeof jQuery == "undefined" || !jQuery.prototype.sidebar)
+        scripts.push("/doccdn/semantic.js")
+    if (!window.pxtTargetBundle)
+        scripts.push("/blb/target.js");
+    scripts.push("/blb/pxtembed.js");
```
guillaumejenkins: Is this intentional? Is it because pxtembed has to be last?

docfiles/docs.js
```diff
@@ -53,15 +53,25 @@ function searchSubmit(form) {
     if (pxt && pxt.tickEvent) pxt.tickEvent("docs.search", { 'source': form.id }, { interactiveConsent: true })
 }
 
+function scrollActiveHeaderIntoView() {
+    var activeHeaders = document.getElementsByClassName("header active");
+    for (var i = 0; i < activeHeaders.length; ++i) {
+        var activeHeader = activeHeaders.item(i);
+        if (activeHeader.scrollIntoView)
+            activeHeader.scrollIntoView()
```
guillaumejenkins: We don't want to scroll multiple times right? After 1st one is scrolled into view, maybe return?

webapp/public/embed.js
pelikhan: Yes it should be last.

docfiles/docs.js
pelikhan: We have 2 TOCa for different views.

pelikhan:
Let's wait on this. Don't have the time to debug if things go south.

PR #5120, opened 2018-12-19
Show download with useUF2

riknoll:


mmoskal:
Oh right, `compile.hasHex` is false before we select compilation variant...

PR #5124, opened 2018-12-21
Changing reporter color back to red

abchatra:
Blue color is impossible to identify as a variable block
image

Changing back to red 
image


abchatra:
Guillaume and I discussed couple of options to improve the experience
1. Add icon to distinguish
2. Slightly different shade of red
3. Variable drop down has a section for local variables
4. If local variable is selected automatically convert to reporter.

pelikhan:
It needs to be a different color and icon. I could easily create a variable “sprite” and drop it in the slot and I would not see any difference. This will lead to confusion.

“If it looks like a variable, it is a variable” — Anonymous

abchatra:
After report changes you cant drop in the event handler. But you need to drop the reporter in the variable slot of destory for example.

Local variables are variables of different kind.

abchatra:
We will take this to next UI meeting. Keeping the status same till then.

PR #5125, opened 2018-12-22
(image editor ) add current image dimensions to preset if non-standard

jwunderl:
If an image has dimensions that were not found by default (e.g. made a different sized sprite in typescript and decompiled), add those dimension to the possible resize options so that the user can cycle back to the same size with the resize button. 

(also serves as a temporary patch for microsoft/pxt-arcade#488, as then they can cycle back to 64x64 or whatever size they had instead of just losing a portion of their sprite)

(thanks @ikeras for pointing out issue)

(and excuse the wonky css in the gif, I hadn't rebuilt css since the monaco field editor update)

### DIFF
--- a/pxtblocks/fields/field_sprite.ts
+++ b/pxtblocks/fields/field_sprite.ts
@@ -99,7 +99,7 @@ namespace pxtblockly {
             });
 
             this.editor.setActiveColor(this.params.initColor, true);
-            if (!this.params.sizes.some(s => s[0] == this.state.width && s[1] == this.state.height)) {
+            if (!this.params.sizes.some(s => s[0] === this.state.width && s[1] === this.state.height)) {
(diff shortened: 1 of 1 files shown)


pxtblocks/fields/field_sprite.ts
```diff
@@ -99,6 +99,9 @@ namespace pxtblockly {
             });
 
             this.editor.setActiveColor(this.params.initColor, true);
+            if (!this.params.sizes.some(s => s[0] == this.state.width && s[1] == this.state.height)) {
```
ikeras: Not sure of your team's style guidelines, but we usually enforce strict equality checks unless we explicitly need them to be loose.

jwunderl: I haven't seen any guidelines for this in particular, but it is always good practice to use strong checks where applicable. Thanks for the note!

PR #5127, opened 2018-12-23
charAt block text change

ganicke:
Seems to need this change for non-LR parms like in the Spanish translation:

EN: ``char from %this=text|at %pos``
vs.
ES: ``devuelve el carácter de la posición %pos|de %this=text``

``%pos`` changed to match the actual parm name ``%index``

RE: https://github.com/Microsoft/pxt-microbit/issues/1311#issuecomment-449068706


pelikhan:
Why are we changing the block id?

ganicke:
No real need to change the ID. At the time I thought that ID should be reserved for the ``string.get()`` as in ``someString[]``.

PR #5137, opened 2019-01-02
String.split

pelikhan:
Our support for string/array is rather spotty. Here is split.

### DIFF
--- a/libs/pxt-common/pxt-helpers.ts
+++ b/libs/pxt-common/pxt-helpers.ts
@@ -207,7 +207,7 @@ namespace helpers {
         const A: string[] = [];
         let lim = 0;
         if (limit === undefined)
-            lim = (1 << 30) - 1; // spec says 1 << 53
+            lim = (1 << 29) - 1; // spec says 1 << 53, leaving it at 29 for constant folding
         else if (limit < 0)
             lim = 0;
         else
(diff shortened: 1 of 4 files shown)


libs/pxt-common/pxt-helpers.ts
```diff
@@ -201,6 +201,63 @@ namespace helpers {
 
         return s.substr(start, end - start);
     }
+
+    export function stringSplit(S: string, separator?: string, limit?: number): string[] {
+        // https://www.ecma-international.org/ecma-262/6.0/#sec-string.prototype.split
+        const A: string[] = [];
+        let lim = 0;
+        if (limit === undefined)
+            lim = (1 << 30) - 1; // spec says 1 << 53
```
mmoskal: better use `29` before we have constant folding in the compiler

libs/pxt-common/pxt-helpers.ts
```diff
@@ -201,6 +201,63 @@ namespace helpers {
...
+        if (s == 0) {
+            let z = splitMatch(S, 0, R);
+            if (z > -1) return A;
+            A[0] = S;
+            return A;
+        }
+        let T: string;
+        let q = p;
+        while (q != s) {
+            let e = splitMatch(S, q, R);
```
mmoskal: using `indexOf` would be faster, though maybe a bit more complicated

pelikhan: I was really just following the spec. We should do a deeper pass on JS support for string/arrays.

PR #5138, opened 2019-01-02
keep menu visible when simulator is up

pelikhan:
When trasitioning between the 3 views, feels weird to have the menu go dark.


pelikhan:
@riknoll is this required by arcade?

riknoll:
I don't think so

PR #5142, opened 2019-01-03
Position picker fixes

samelhusseini:
Only show position picker once:
Fixes https://github.com/Microsoft/pxt-arcade/issues/471
Fixes https://github.com/Microsoft/pxt-arcade/issues/490


pxtblocks/fields/field_position.ts
```diff
@@ -70,6 +71,9 @@ namespace pxtblockly {
             const simFrame = this.getSimFrame();
             if (!simFrame) return;
 
+            if (FieldPosition.eyedropperActivated_) return;
+            FieldPosition.eyedropperActivated_ = true;
```
pelikhan: Could we query the DOM If a particular element is present?

samelhusseini: Yes. this.selectorDiv_

pelikhan:
@samelhusseini if i click on the X editor, then the Y editor, selectorDiv_ leaks in the editor. After that, it is impossible to open the view.

PR #5144, opened 2019-01-03
Support unicode

mmoskal:
This implements emitting different forms of strings, as explained in https://github.com/Microsoft/pxt/blob/utf8/docs/js/values.md#strings

Most of code is in https://github.com/Microsoft/pxt-common-packages/pull/610


mmoskal:
This adds about 2k of code (in flash) - this is why we have flag to disable it, potentially on microbit, since the screen is a bit low-res for many unicode characters.

img_0671

PR #5150, opened 2019-01-04
 Implement wave ~5 - real noise tone

mmoskal:


mmoskal:
Trying to fix safari as well

PR #5155, opened 2019-01-05
WebUSB improvements

mmoskal:
* use `webusb` npm module instead of `node-hid` where WebUSB is enabled
* support for UF2 file family, see https://github.com/Microsoft/uf2#family-list - prevents flashing with the wrong binary
* auto-detect hardware family from WebUSB if available (and skip hardware selection)
* when HF2 flashing, and `HF2_CMD_START_FLASH` (non-reset hand-over to bootloader) fails, just do a full reset into bootloader
* don't assume HF2 payload size to be page size (pages are gigantic on STM32F4)
* fix crash in HF2 version number parsing
* handle webusb device disconnect
* visual fixes to the hardware selection dialog
* some fixes in WebUSB pairing logic, though it still might need some work
* disable WebUSB on Windows 7 and 8 (requires drivers there); enable on Win8.1, Win10, mac, Linux etc


mmoskal:
@abchatra sorry, I wrongly assumed there are no users between 7 and 10, but it seems around 5% sits at 8.1 globally (8.0 is dead it seems). Updated the check.

mmoskal:
Also the descriptors technically require 8.1 (which is 6.3; go figure). My check for Win10 was just because I thought we wouldn't have throughput to test it.

cli/hid.ts
```diff
@@ -2,14 +2,20 @@ import HF2 = pxt.HF2
 import U = pxt.U
 import * as nodeutil from './nodeutil';
 
+function useWebUSB() {
+    return !!pxt.appTarget.compile.webUSB
```
pelikhan: can compile be missing?

mmoskal: no - https://github.com/Microsoft/pxt/blob/master/pxtlib/main.ts#L91

PR #5160, opened 2019-01-08
Fix specific typo

darthwalsh:


pelikhan:
Needs to be cherry picked into master.

PR #5161, opened 2019-01-08
Tutorial mobile / iOS

pelikhan:
- [x] don't show too many dots in tutorials on mobile as it bleeds out the screen

image

- [x] specify xlink namespace for svg:image href so that the image shows in Safari iOS.


pelikhan:
I LUV XML

PR #5163, opened 2019-01-09
PNG Phoenix

pelikhan:
This PR adds a "screenshot" button in the sim toolbox (enabled for arcade), which uses the PNG encoding feature.
phoenix


abchatra:
Does this contain the project? We don't want 3 ways to share the projects. 
Download -> Downloads the .uf2
Save as -> .mkcd

abchatra:
I should say I like the png otherwise. We need png but can it just contain the image and not the project?

pelikhan:
Why not have the project embedded?

pelikhan:
Note that "download" will not generate a file in most cases for Arcade since we'll use the WebUSB route.

abchatra:
>> Note that "download" will not generate a file in most cases for Arcade since we'll use the WebUSB route

This is a misconception. 30% of users are on windows 7 who don't have webusb. 30% others are on non chrome. Only 40% are on chrome with WebUSB.

Also we expect lot of folks to use arcade without hardware, which means they won't have any hardware connected.

PR #5173, opened 2019-01-11
Hardwarevariants

pelikhan:
Additional hw boards loaded from markdown page


mmoskal:
When we have many generic boards, we should move the generic cards on top. But fine for now.

PR #5174, opened 2019-01-11
use choose hw variant dialog when applicable

pelikhan:
Fix for https://github.com/Microsoft/pxt-arcade/issues/539, move variants atop dialog.


webapp/src/container.tsx
```diff
@@ -176,7 +176,11 @@ export class SettingsMenu extends data.Component<SettingsMenuProps, SettingsMenu
 
     showBoardDialog() {
         pxt.tickEvent("menu.changeboard", undefined, { interactiveConsent: true });
-        this.props.parent.showBoardDialogAsync(undefined, true).done();
+        const variants = pxt.getHwVariants();
+        if (variants.length)
```
mmoskal:
```suggestion
        if (variants.length > 1)
```
Only do it when there is more than one.

PR #5175, opened 2019-01-11
added UIntYYYArray.some polyfills

pelikhan:


abchatra:
There is another polyfill in worker, update that too please.

pelikhan:
It is not used in the worker. Fixing duplication of polyfills should be fixed but not here.

abchatra:
Lets add for both as we will hit that bug some other time.

pelikhan:
Polyfills are not consistent at all between worker and sim. It's not just that one.

pelikhan:
Added a few polyfills but it's by no means complete.

PR #5180, opened 2019-01-12
Compiler perf improvements

mmoskal:
WIP

@riknoll any clue if caching type on Node is a good idea? `(node as any).typeOverride = r`

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -3751,6 +3751,9 @@ function buildCoreAsync(buildOpts: BuildCoreOptions): Promise<pxtc.CompileResult
                 }
             }
 
+            if (pxt.appTarget.compile.switches.time)
+                console.log(compileResult.times)
+
             switch (buildOpts.mode) {
                 case BuildOption.GenDocs:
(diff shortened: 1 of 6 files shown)


riknoll:
It's fine if we don't reuse the TS program (which the compiler does not right now). Might cause issues if we ever decide to do that though, because it tries to keep source files (with the AST) around.

cli/cli.ts
```diff
@@ -3751,6 +3751,9 @@ function buildCoreAsync(buildOpts: BuildCoreOptions): Promise<pxtc.CompileResult
                 }
             }
 
+            if (pxt.appTarget.compile.switches.time)
+                console.log(compileResult.times)
```
pelikhan: you can use "pxt.debug"

mmoskal: pxt.debug get lost in all other noise, I guess pxt.log would be good, but need to stringify first

pelikhan: Not worth it.

mmoskal:
should be good to merge @pelikhan

pelikhan:
@riknoll ?

riknoll:
👍

PR #5182, opened 2019-01-14
Embedded simulator with screenshots in share dialog + no auto-run

pelikhan:
2 changes: simulator in share dialog + no-autorun mode with clear icons on simulator.

When an editor supports screenshots for the simulator, host the simulator in the sharing dialog and allow use to snap the right screenshot. It increases the quality of the shared scrips since it has more changes to have a good screenshot. Tested in Chrome, IE, Firefox (Edge untestable right now).

**ps:** we should consider disabling autoRun on slow machines for micro:bit as well.
**pss:** we need a quantitative analysis of compile times in the wild

- [x] Under the hood, the simdriver gives the simulator IFrame on "loan" to the share dialog. As soon as the dialog is done, the IFrame comes back in its original location.

screenshot

- [x] no autorun mode

Using the embed lightbox rendering to clearly display what's going on with the simulator.
autorun

Test: https://arcade.makecode.com/app/6947d73e04dfb5a9c38493e483656230cdca3168-7e1ebf74e5

### DIFF
--- a/webapp/src/sui.tsx
+++ b/webapp/src/sui.tsx
@@ -431,7 +431,7 @@ export class Button extends StatelessUIElement<ButtonProps> {
         const classes = cx([
             color,
             size,
-            disabled || loading ? 'disabled' : '',
+            (disabled || loading) ? 'disabled' : '',
             loading ? 'loading' : '',
             genericClassName("ui button", this.props)
         ])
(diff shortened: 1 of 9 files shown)


webapp/src/sui.tsx
```diff
@@ -426,11 +427,12 @@ export interface ButtonProps extends UiProps, TooltipUIProps {
 
 export class Button extends StatelessUIElement<ButtonProps> {
     renderCore() {
-        const { labelPosition, color, size, disabled } = this.props;
+        const { labelPosition, color, size, disabled, loading } = this.props;
         const classes = cx([
             color,
             size,
-            disabled ? 'disabled' : '',
+            disabled || loading ? 'disabled' : '',
```
riknoll: Parentheses around the ternary?

PR #5186, opened 2019-01-15
Webusbpairdialog

pelikhan:


mmoskal:
I guess the logic should be:
```
try to pair
if (connected)
  if (bootloaderMode)
     flash
     pair again
   else
     switch to bootloaderMode
     flash
     should be paired
```

pelikhan:
Are you missing a pair if the user starts the process in application mode?
```
try to pair
if (connected)
  if (bootloaderMode)
     flash
     pair again
   else
     switch to bootloaderMode
     pair
     flash
     should be paired
```

mmoskal:
right, the flashing will I guess trigger the pairing

pelikhan:
The new flow removes all the various dialog and puts pairing under a single dialog. What about asking the user to enter bootloader mode first?

mmoskal:
I would rather not ask them if we don't have to - depending on what program is flashed they need to hit reset once, twice, or long...

PR #5194, opened 2019-01-17
faster maker build

mmoskal:
This brings down maker build time down to earth (5x probably).

This works by reusing the same C++ image for all boards with the same MCU (by including a bit more C++ than strictly needed). From docs:

> Dependencies under `cppDependencies` are only considered when generating code for the C++ compiler. Usually, one would list all optional packages which contain C++ code in `cppDependencies` of your core package. Then, when user actually adds any of these optional packages, the C++ code doesn't change and re-compilation (and thus cloud round-trip) is not required.

@abchatra this should not affect Arcade, but I guess can wait until after release


mmoskal:
Should be good to merge

PR #5195, opened 2019-01-17
Add missing sim UTF8 methods

mmoskal:
Not yet tested


pelikhan:
Tested rountripping some unicode strings, looks good.

PR #5198, opened 2019-01-17
compliant array slice

pelikhan:
Just annoying...


pelikhan:
@abchatra your call, nice to have.

PR #5199, opened 2019-01-17
Add null+type check on method invocation

mmoskal:


pelikhan:
LGTM. Errors messages are less confusing with this but still very well hidden. @abchatra  your call.

riknoll:
@abchatra I think we should take this change for beta. You get these errors with sprites all the time

abchatra:
lets take it. Merge to stable5.8

pelikhan:
Who does the integration?

abchatra:
I will do it.

PR #5203, opened 2019-01-17
Fixing touch event regression in sprite editor iOS

riknoll:
Fixes https://github.com/Microsoft/pxt-arcade/issues/567
Broke it when I redid the touch events for the Monaco field editor.

### DIFF
--- a/pxtlib/sprite-editor/canvasGrid.ts
+++ b/pxtlib/sprite-editor/canvasGrid.ts
@@ -284,12 +284,12 @@ namespace pxtsprite {
         /**
          * This calls getBoundingClientRect() so don't call it in a loop!
          */
-        protected clientToCell(clientX: number, clientY: number) {
+        protected clientToCell(coord: ClientCoordinates) {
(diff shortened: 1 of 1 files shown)


pxtlib/sprite-editor/canvasGrid.ts
```diff
@@ -493,4 +509,20 @@ namespace pxtsprite {
...
+    export interface ClientCoordinates {
+        clientX: number;
+        clientY: number;
+    }
+
+    function clientCoord(ev: PointerEvent | MouseEvent | TouchEvent): ClientCoordinates {
+        if ((ev as TouchEvent).touches) {
+            const te = ev as TouchEvent;
+            if (te.touches.length) {
+                return te.touches[0];
```
pelikhan: can we handle all the touch events? If the user press both the dpath and A/B,  we get multiple entries in "touches" (future PR)

pelikhan:
Let's get this in and file a bug to track proper fix.

samelhusseini:
Filed a bug here to track proper fix: #5205.
This is good to go.

PR #5216, opened 2019-01-19
Support for GIF generation 

pelikhan:
Uses gif.js to generate gif of the simulator, instead of just screenshots.
- [x] fix leaking workers
- [x] crowss browser test: tested Chrome, Edge, IE, FF. IE is verrrrry slow.
- [x] bring back simple screenshot
- [x] don't mangle target name in shared script name as it is already in "site_name" in metadata
- [x] disable system context menu in simulator
- [x] flag in theme
- [x] support for keyboard shortcut to start/stop recording
- [x] detect when gif is too large, tell user

**Do not merge until cloud supports it**

gifrecording


pelikhan:
@abchatra do you have an ETA on the cloud?

riknoll:
Very cool

pelikhan:
I am looking to getting rid of gif.js and supporting gif encoding directly. There is a layer of useless options in there. Then I will integrate it into our worker.

PR #5242, opened 2019-01-29
Additional file path load issue

pelikhan:
I'm not sure why we did not more issues with this one. Basically, we patch the pxt.json payload before packaging it into target.json.
https://github.com/Microsoft/pxt/pull/5242/files#diff-3395275f9da491d407273b5a3d768ea1R803

However, we also cycle through the files, read pxt.json and override it.

### DIFF
--- a/pxtlib/package.ts
+++ b/pxtlib/package.ts
@@ -800,6 +800,8 @@ namespace pxt {
                     })
                     files[pxt.CONFIG_NAME] = JSON.stringify(cfg, null, 4)
                     for (let f of this.getFiles()) {
+                        // already stored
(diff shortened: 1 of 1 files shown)


riknoll:
I don't understand the issue?

pelikhan:
We patch pxt.json in order to be serialized and store it. Then we iterate all the files, including pxt.json to store them in files. So we override any kind of patching being made in the step.

pelikhan:
We delete additionalFilePath and additionalFilePaths field from the config before serializing. If those stay, the CLI will try to resolve those path and fail to install anything. (so CLI was broken).
Also we patch the github references so that they are not pointing to a local repo.

PR #5244, opened 2019-01-30
fix #5021 by prepending path with "/docs"

### LINKED ISSUE #5021: docs images not showing after setup
**Describe the bug**
Installing this repo and dependencies, then packaging them and hosting them as a static bundle  results in images not loading. The markdown in `codecard` s points to `/static/mb/....` but all of those paths need prefixing with `/docs` to point to the correct images.

**To Reproduce**
1. clone and install
2. run `pxt staticpkg` 
3. host the `built/packaged` directory
4. See images not showing in browser

**Expected behavior**
Visible thumbnails

**Screenshots**

screen shot 2018-11-16 at 7 07 47 am

the 'heart' thumbnail has a path prefixed with /docs, and it loads properly

 - OS: MacOS
 - Browser Chrome
 - Version 70

johnstonmatt:
the thumbnails in docs compiled by pxt are missing the `/docs` prefix, leading to images not being displayed, this fix adds that prefix


johnstonmatt:
@pelikhan thanks for the quick reply! do you know where this functionality lives now?

pelikhan:
probably cli.ts

johnstonmatt:
that does seem to be the case, it is possible the version of pxt my team is using depends on the tdast.js, and should upgrade to fix the issue. I will try that and report back to both here and #5201 with findings.

pelikhan:
Might be accidental. tdast.js is (should) only be used to convert touchdevelop scripts into makecode

pelikhan:
I'm closing this unless we missed something that needed to be done here.

PR #5251, opened 2019-02-01
Serve live sim if local sim not found

guillaumejenkins:


cli/server.ts
```diff
@@ -1036,6 +1036,13 @@ export function serveAsync(options: ServeOptions) {
             }
         }
 
+        if (/simulator\.html/.test(pathname)) {
+            // Special handling for missing simulator: redirect to the live sim
```
pelikhan: how do you know it is missing?

guillaumejenkins: The `for` above would have found it and served it and returned; if we reach this point it means it wasn't found in neither `sim/public` nor `node_modules/pxt-TAREGET_ID-sim/public` (see `dirs` at the top of the file)

PR #5256, opened 2019-02-06
Better variable declarations when going from blocks to JS

riknoll:
This PR overhauls the way we emit variables so that the compiled JS looks closer to something someone would actually write. Changes include:

1. Variables now get declared locally when possible
2. If a variable is never referenced before it's first assignment, we now declare it at the assignment location (e.g. `let item = 0; item = 1` is now `let item = 1`)
3. Type declarations for variables are only emitted when necessary
4. Different variables with the same name (e.g. two for-loops that declare variables named `index`) no longer get renamed unless they actually conflict with one another

To give a feel for it, here is an example program:
image

Here's what the current logic emits:
```typescript
let eventGlobal = 0
let loopLocal = 0
let ifLocal = 0
controller.A.onEvent(ControllerButtonEvent.Pressed, function () {
    if (true) {
        ifLocal = 1
    } else {
        ifLocal = 1
    }
    for (let i = 0; i < 4; i++) {
        loopLocal = 2
    }
    eventGlobal = 27
})
controller.B.onEvent(ControllerButtonEvent.Pressed, function () {
    if (eventGlobal == 0) {
    	
    }
})
```

Here's what the new logic emits:

```typescript
let eventGlobal = 0
controller.A.onEvent(ControllerButtonEvent.Pressed, function () {
    let ifLocal = 0
    if (true) {
        ifLocal = 1
    } else {
        ifLocal = 1
    }
    for (let i = 0; i < 4; i++) {
        let loopLocal = 2
    }
    eventGlobal = 27
})
controller.B.onEvent(ControllerButtonEvent.Pressed, function () {
    if (eventGlobal == 0) {
    	
    }
})
```

This should not be a breaking change.


pelikhan:
Very nice.

pelikhan:
Did you test this with tutorials? We do this bundling of snippets to decompile all code at once... (bundleTutorialCode)

riknoll:
I'll try it out! Shouldn't affect it, since that has more to do with the decompiler than the compiler

pelikhan:
Looks like this is ready?

riknoll:
Yes it's ready

PR #5270, opened 2019-02-08
better auto generation of summaries

pelikhan:
Handles partial urls, youtube videos, learn more, buy, etc...


ganicke:
What about when when both `youTubeId` and `imageUrl` are defined, we get:

image

pelikhan:
Right.... that's a rendering bug. Will fix.

pelikhan:
https://github.com/Microsoft/pxt/pull/5272

PR #5276, opened 2019-02-11
Keep variables global if assigned to

riknoll:
Fixes https://github.com/Microsoft/pxt-arcade/issues/752
Fixes https://github.com/Microsoft/pxt-arcade/issues/751

Unfortunately, this undoes a lot of the logic I added to declare variables locally (oh well). The other improvements still apply (fewer renames, combined declarations, etc.).


pelikhan:
Sad but necessary...

PR #5277, opened 2019-02-12
Fixing function call type inference and name conflicts

riknoll:
Fixes https://github.com/Microsoft/pxt-arcade/issues/748

In addition to the type inference fix, I also fixed another bug I found while testing where functions were getting renamed for no reason (e.g. `doSomething` ->  `doSomething2`). The API info we use to check for name collisions also includes user functions, so functions will collide with themselves after the first compile. My fix for that creates another bug where you can now have collisions with functions defined in `custom.ts`, but I'm not going to fix it since that feature isn't really used that much. I left a comment.

I also "turned on" two test cases for the new functions that I think Guillaume forgot to add to the list of tests.

### DIFF
--- a/pxtblocks/blocklycompiler.ts
+++ b/pxtblocks/blocklycompiler.ts
@@ -409,6 +409,11 @@ namespace pxt.blocks {
                         handleGenericType(b, "LIST");
                         unionParam(e, b, "INDEX", ground(pNumber.type));
                         break;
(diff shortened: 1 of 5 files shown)


jwunderl:
Ah, we talked about that one (user functions renaming too much) before and forgot about it for a bit I think. Let me find where Guillaume and I looked through a bug on it

PR #5279, opened 2019-02-13
Generate the blockly.d.ts from sources

samelhusseini:
In conjunction with https://github.com/Microsoft/pxt-blockly/pull/200. 
Generate blockly.d.ts from sources. Now we have a true holistic view of Blockly in Typescript, and it's maintainable in the future, as it can be simply updated by running a script. 

Changes in PXT revolve around removing any in cases like Blockly.Msg. Since now we have a true idea of what Messages are defined, or in other cases it's using / casting Block to BlockSvg or Workspace to WorkspaceSvg since that's the actual class we use in PXT.

I've also cleaned up a bunch of unused helper methods in blocklyloader.ts that were used by the old toolbox.

I also found a number of issues with our dts interfaces that were previously hidden such as https://github.com/Microsoft/pxt/compare/generatedblocklydts?expand=1#diff-6e6afa1a68b5611fb0fdd0b941924069R20 and https://github.com/Microsoft/pxt/compare/generatedblocklydts?expand=1#diff-e9f64ab81e66f15e0e8de0a44a49425cR21

### DIFF
--- a/localtypings/blockly.d.ts
+++ b/localtypings/blockly.d.ts
@@ -4834,7 +4834,8 @@ declare module Blockly.ContextMenu {
 
     /**
      * Make a context menu option for duplicating the current workspace comment.
-     * @param {!Blockly.WorkspaceCommentSvg} comment The workspace comment where the right-click originated.
+     * @param {!Blockly.WorkspaceCommentSvg} comment The workspace comment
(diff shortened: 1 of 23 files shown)


riknoll:
Awesome

PR #5281, opened 2019-02-15
Add local variables to the variable dropdown

riknoll:
The new function/event blocks in Arcade use draggable function reporters for their arguments. The downside is that those blocks are not "true" Blockly variables so they don't show up in the dropdowns of variable fields. It quickly gets annoying in Arcade, because all of the Sprite related blocks have variable_get blocks as shadows and to use any local variables you need to drag them on top.

This PR adds those local variables to the dropdown, but only in the function scope where they are visible. I only added that for "get" blocks, because function arguments should really be considered readonly. When the user selects a local variable from the dropdown, it gets promoted to a global variable (and thus will also show up in the flyout).

Here's a demo:

local_variable_dropdown

In that GIF, "sprite" and "otherSprite" are local, while the other variables are global.

### DIFF
--- a/pxtblocks/blocklycompiler.ts
+++ b/pxtblocks/blocklycompiler.ts
@@ -17,6 +17,10 @@ namespace pxt.blocks {
         idToComments: Map<Blockly.WorkspaceComment[]>;
     }
 
+    export interface BlockWithScope extends Blockly.Block {
+        _pxtScope?: string[];
+    }
+
     ///////////////////////////////////////////////////////////////////////////////
     // Miscellaneous utility functions
     ///////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


jwunderl:
Minor thing, this will add the ability to rename (and delete?) parameters in the drop down (drag in a getter, switch to param name, rename / delete), correct?

riknoll:
That will delete the "variable", not the parameter. Good point on the rename though; that might not work right...

pelikhan:
What is the status on this?

pelikhan:
@riknoll @abchatra @Jaqster this seems relevant to the new radio blocks.

pelikhan:
We should really get this in. @riknoll @abchatra dragging the blocks from event is not intuititve.

riknoll:
Let me test it a bit

abchatra:
I would like to get this in as well. Lets test it thoroughly before merging.

pelikhan:
@riknoll nudge

pelikhan:
@riknoll @abchatra do we want this?

PR #5291, opened 2019-02-27
check if window is defined

pelikhan:


pelikhan:
Check is better - does not annoy debugger

PR #5298, opened 2019-02-27
Allow type index declarations

riknoll:
Fixes https://github.com/Microsoft/pxt-microbit/issues/1848


riknoll:
Tested in simulator and on hardware (micro:bit)

riknoll:
Yep

PR #5303, opened 2019-03-01
Don't consider "null" and "undefined" different appTargetVariants

darzu:
Fixes https://github.com/Microsoft/pxt-microbit/issues/1875
Partially rolls back behavior in introduced with: #5234 

This PR changes the behavior in setAppTargetVariant so that we don't consider "null" and "undefined" different appTargetVariants

Before this fix, pxt-microbit start up looked like:
call cmds.ts init() which sets the target-agnostic deploy behavior
call initExtensionsAsync() in pxt-microbit which sets the target-specific deploy behavior
call setAppTargetVariant(null) which triggers \/
call cmds.ts init() which **overwrites** the target-specific deploy behavior, which caused the bug

### DIFF
--- a/pxtlib/main.ts
+++ b/pxtlib/main.ts
@@ -162,7 +162,7 @@ namespace pxt {
     // this is set by compileServiceVariant in pxt.json
     export function setAppTargetVariant(variant: string): void {
         pxt.debug(`app variant: ${variant}`);
-        if (appTargetVariant === variant) return;
+        if (!!appTargetVariant === !!variant) return;
         appTargetVariant = variant
         appTarget = U.clone(savedAppTarget)
         if (variant) {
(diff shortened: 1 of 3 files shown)


pxtlib/main.ts
```diff
@@ -162,7 +162,7 @@ namespace pxt {
     // this is set by compileServiceVariant in pxt.json
     export function setAppTargetVariant(variant: string): void {
         pxt.debug(`app variant: ${variant}`);
-        if (appTargetVariant === variant) return;
+        if (!!appTargetVariant === !!variant) return;
```
pelikhan: I don’t think that’s right. For example an app variant might be “foobar”

darzu: Oh of course, duh. Instead:
```suggestion
        if (appTargetVariant === variant || (!appTargetVariant && !variant)) return;
```

darzu:
> You need an additional shortcut in case both target string are equal (and non null / undefined)

I believe the logic works now

PR #5304, opened 2019-03-01
Chunk serial data into lines before consuming in graphs or serial console

darzu:
Fixes https://github.com/Microsoft/pxt-microbit/issues/1823

Previously we were assuming each message from the device was one line. This was not the case. Occasionally lines would be split over multiple messages. The serial console correctly chunked incoming data into lines before displaying them however the graph data did not so we ended up with too many graphs and incorrect data being displayed.

This PR moves the line chunking logic before the serial graph parsing and out of appendConsoleEntry. No other logical changes were made.

after fix:
fix-too-many-consoles


webapp/src/serial.tsx
```diff
@@ -157,17 +157,29 @@ export class Editor extends srceditor.Editor {
...
+        // clean up input
         const data = smsg.data || ""
         const source = smsg.id || "?"
         const receivedTime = smsg.receivedTime || Util.now()
 
         this.appendRawData(data);
         const niceSource = this.mapSource(source);
 
+
+        // chunk into lines
```
pelikhan: Personal preference: use const whenever possible.

PR #5314, opened 2019-03-06
regenerated blockly.d.ts

pelikhan:
Build from https://github.com/Microsoft/pxt-blockly/pull/204. It's not clear if the changes in FieldDropDown are a recent blockly change or just out typechecker who was confused.


pelikhan:
The refactoring changes in blockly have been reverted.

PR #5316, opened 2019-03-07
Speed up toolbox opening by 4X for monaco blocks

darzu:
I've speed up the toolbox opening / switching by rearranging some of the frequently run code.
We only need to compute extended types on blocks if an exact match isn't found.

Before: 1.33s to open toolbox 
speed-up-toolbox-before

After: 0.34s to open toolbox (still too slow IMO)
speed-up-toolbox-after

Perf numbers are from Chrome's profiler with "6x CPU throttle" enabled. Once I get my ChromeBook I'll get some "real world" slow device numbers :)


darzu:
Interesting, the CI stalled out

Issue #5324, opened 2019-03-08
Microbit Curriculum blog post


Jaqster:
Hey Galen,
Can you prep this blog post about the new Educator Materials for the Micro:bit Curriculum?  Feel free to edit.
Files uploaded here: https://1drv.ms/f/s!AmMIW5Hxi0RtgZlWcV8e38x9111pFQ
Thanks!
Jacqueline

ganicke:
Soitenly!

Jaqster:
Hey Galen, can you merge?

PR #5327, opened 2019-03-08
CSIntro educator materials blog post

ganicke:
Make a blog post about the new educator materials available for micro:bit CSIntro.

RE: https://github.com/Microsoft/pxt/issues/5324


pelikhan:
We should use "buttons" macro to clearly indicate where the downloads are.

ganicke:
It would be nice to use a``# ~ button`` section but it only works with links relative to the homeurl. I can use ``[`|Download|`](/path)`` to get a similar result.

image

ganicke:
Ok @Jaqster, GTG for Tuesday.

PR #5328, opened 2019-03-09
5v breadboarding support

pelikhan:
Adding support for routing 5V and 3V power lines on breadboard.
image


pelikhan:
@darzu the allocator is still there waiting for you :)

PR #5329, opened 2019-03-11
Fixes for new functions in decompiler and Blockly compiler

riknoll:
Fixes https://github.com/Microsoft/pxt-microbit/issues/1880
Fixes https://github.com/Microsoft/pxt-microbit/issues/1879
Fixes https://github.com/Microsoft/pxt-microbit/issues/1878

This PR adds a check to ensure only arguments with types that can be declared from blocks are allowed to be decompiled. In other words, just "number", "string", "boolean", and the types listed in pxtarget.json (just "Sprite" in arcade). This only affects user-declared functions.

I also fixed two other new function bugs I found along the way:

1. Array types no longer get mangled by the Blockly compiler's escaping logic
2. Fixed the output checks on custom arg blocks so that they handle our array and type-inheritance logic. I had to do this via a monkey-patch because it doesn't make sense to port that logic into pxt-blockly


riknoll:
Not in this PR

PR #5330, opened 2019-03-11
Resilient loader

pelikhan:
Add some resiliency when loading projects with broken pxt.json or invalid dependencies.

- [x] invalid pxt.json syntax
- [x] invalid built-in library reference


pelikhan:
@riknoll is that what you had in mind?

riknoll:
Yep, and I also wrote some linting for the file editor (but that can be a separate PR)

PR #5333, opened 2019-03-12
Fix offline mode for Safari [investigation]

darzu:
DRAFT PR.

For me to collect notes and invite feedback.

### DIFF
--- a/pxtlib/browserutils.ts
+++ b/pxtlib/browserutils.ts
@@ -878,6 +878,7 @@ namespace pxt.BrowserUtils {
         }
     }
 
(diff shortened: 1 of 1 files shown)


darzu:
# Questions:
[ ] did offline mode ever work on Safari?

# Notes:
## local storage, Safari vs Chrome
image

pelikhan:
I've noticed that it was not working on iPhone as well.

pelikhan:
I noticed lzma is not in the manifest

pelikhan:
Irrelevant with service worker.

PR #5337, opened 2019-03-12
revert flyout opt

pelikhan:
Can't drag blocks on second opening of toolbox
fire


darzu:
Thanks for finding this Peli. Part one of the fix is here: https://github.com/Microsoft/pxt-blockly/pull/212

PR #5338, opened 2019-03-13
Allow a language if base language is available

abchatra:
Simple fix to allow base language if it available. For example: fall back to "nl" if "nl-NL" is not available. Check was in the front end.


abchatra:
Fixes https://github.com/Microsoft/pxt-minecraft/issues/811

PR #5323, opened 2019-03-08
Cache blockly flyouts for faster subsequent loads (~3X speedup)

darzu:
REQUIRES: https://github.com/Microsoft/pxt-blockly/pull/208

Previously, we would recompute and relayout all the blocks each time we opened the "flyout" when clicking on the toolbox.

With this change, we maintain separate flyouts for each toolbox category and toggle their visiblity state when switching between categories. This doesn't speed up the initial load, but subsequent loads are ~3X faster.

I'm also investigating a way to get another 3X (for a total of 9X) speed up by playing nice with React so it doesn't force a style recompute when toggling visibility.

Note this affects all blockly editors.

## Before
[macbookpro, x6 slowdown] 1st load: ~1.3 seconds
[macbookpro, x6 slowdown] nth load: ~600 ms
[chrome book LC330] 1st load: ~1.8 seconds
[chrome book LC330] nth load: ~680ms
flyouts_perf_before

## After
[macbookpro, x6 slowdown] 1st load: ~1.3 seconds (same)
[macbookpro, x6 slowdown] nth load: ~180 ms
flyouts_perf_after

perf numbers are from Chrome profiler using: 
- 6x CPU throttling on MacBook Pro 2018 i9
- Lenovo C330 Chrome book w/ MT8173C

PR #5340, opened 2019-03-13
Fix flyout interaction & re-enable multi-flyout caching

darzu:
Reverts & fixes: https://github.com/Microsoft/pxt/pull/5337

My multi-flyout caching introduced with #5323 broke dragging of blocks from the flyout after it was closed then opened.

This was due to destructive side-effects in the Flyout.hide() code in blockly, which I've proposed a fix for here: https://github.com/Microsoft/pxt-blockly/pull/212

Try the fix here: https://arcade.makecode.com/app/e07373f9b0f4944f2a118372b04b2ae7550353b3-0f3b278df6#editor

Validation:
- [x] Ensure leaving event registrations on hide() doesn't cause further issues
  - confirmed the only events that are registered are: `mousedown, mouseenter, mouseleave, mousedown` which don't trigger on display: none
- [x] complete a tutorial on IE 11


darzu:
Found an issue with zooming in & out. Only the active flyout gets zoomed.
- [ ] Fix zoom in & out issue

webapp/src/blocks.tsx
```diff
@@ -1287,11 +1288,43 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         this.showFlyoutInternal_(this.flyoutXmlList);
     }
 
-    private showFlyoutInternal_(xmlList: Element[]) {
+    private flyouts: pxt.Map<Blockly.VerticalFlyout> = {};
```
pelikhan: When do we clear this cache?

darzu: So far, never. I can't think of a case where we'd want to. This cache doesn't grow unbounded, there's only one one entry per toolbox category. We refresh the contents of the flyouts in these caches when appropriate. Some flyouts that change a lot like variables we don't even cache.

darzu: It's not exactly a cache. We're just maintaining multiple flyouts at once, and they're mutable.

pelikhan:
You need to clear the cache whenever a project gets reload as the fly out might change , e.g. an extension adds new blocks.

pelikhan:
I think there is already a place where we blow out caches

pelikhan:
Try adding an extension, open fly out, remove extension, open fly out.

darzu:
> Try adding an extension, open fly out, remove extension, open fly out. Get Outlook for iOS<https://aka.ms/o0ukef>

I see. I forgot extensions could add blocks to existing flyouts

pelikhan:
Also after creating a new variable, blocks might change to autoincrment the block name.

darzu:
> Also after creating a new variable, blocks might change to autoincrment the block name.

I can't find an example of this outside the Variables flyout (which works). Do you know of any?

darzu:
> > Also after creating a new variable, blocks might change to autoincrment the block name.
> 
> I can't find an example of this outside the Variables flyout (which works). Do you know of any?

Ah found it. Creating the variable "mySprite" causes the sprite blocks to increment to "mySprite2". This is subtle stuff.

darzu:
I've implemented a finer-grain cache invalidation scheme based on the hash of the blocks xml. This seems to work very well and handles all cases I know of.
Test here:
https://arcade.makecode.com/app/a5970054f60bc4ab93403bc282041065a6ad70db-205a1c8059

darzu:
- [ ] pull in the latest pxt-blockly changes

darzu:
Found an issue with the caching only in IE.
- [ ] fix IE issue

darzu:
> I think you should still clean the cached when a header is reloaded.

Why is that? Since this is caching based on the hash of the blocks xml, it should be plenty precise and will save extra work. The first load can be really slow, on the order of several seconds, up to 10 seconds. Avoid re-doing that work as much as possible is very good thing in my mind. I don't see a case where invalidating on header reload would catch anything that invalidating on the xml hash wouldn't.

darzu:
> Found an issue with the caching only in IE. Works in Firefox, Safari, and Chrome..
> 
> * [x]  fix IE issue

Fixed. Here: https://arcade.makecode.com/app/fbaae575ebff2f139efa7b5b80c0a4c040cb371c-4f44116f61#editor

darzu:
Test here: https://arcade.makecode.com/app/6b23f6639d69a49656d0af6dc6495a1b82019220-0bbe4f1578

PR #5350, opened 2019-03-14
Handle double sim for radio in tablet and embed when fullscreen

riknoll:
Fixes https://github.com/Microsoft/pxt-microbit/issues/1700

mbit_radio

Also takes care of a bug on beta where the filelist overlay was showing up when the sim was in sandbox mode, even though the sim is always full screen there.


pelikhan:
Nice!

PR #5353, opened 2019-03-14
Debugger improvements

gushuro:
Rearranging controls toolbar in the toolbox.
Variable explorer:
- Show variables name on hover. Show variables value on hover for
{string, bool, number, array}
- Zebra decoration.
- Background change.

Reduced the amount of green.
Changed color of debug button in the simulator toolbar to be gray/orange for off/on.

Variable Explorer v2 2 1

### DIFF
--- a/pxtsim/libgeneric.ts
+++ b/pxtsim/libgeneric.ts
@@ -20,17 +20,14 @@ namespace pxsim {
         }
 
         toDebugString(): string {
-            return "[" + this.data.join(", ") + "]";
-        }
-        toDebugStringTruncating(): string {
             let s = "[";
             for (let i = 0; i < this.data.length; ++i) {
                 if (i > 0)
                     s += ",";
                 let newElem = RefObject.toDebugString(this.data[i]);
-                if (s.length + newElem.length > 20) {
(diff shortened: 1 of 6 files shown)


pxtsim/libgeneric.ts
```diff
@@ -20,14 +20,17 @@ namespace pxsim {
         }
 
         toDebugString(): string {
+            return "[" + this.data.join(", ") + "]";
```
riknoll: If they have a really big array, this is going to take a lot of memory. I think you should stick with the bail out strategy

gushuro: Yes, makes sense. I rolled this back and just increased the maximum amount of characters we send to 100, which should be enough for most user-created arrays while making sure we don't throttle everything else.

pxtsim/libgeneric.ts
```diff
@@ -20,17 +20,14 @@ namespace pxsim {
         }
 
         toDebugString(): string {
```
pelikhan: this will need rewritting. ``toDebugString`` should also receive the remaining number of characters as an argument -- and shortcut the recursion when the max is reached.
Can be done in future PR.

PR #5357, opened 2019-03-18
Fix browser detection of Firefox on Mac

darzu:
Firefox was being detected as Safari in https://github.com/Microsoft/pxt-microbit/issues/1295


darzu:
Now: 
image

abchatra:
Second condition is redundant if the first condition is true right? 
```javascript
 /Safari/i.test(navigator.userAgent) && /(Macintosh|Safari|iPod|iPhone|iPad)/)/i.test(navigator.userAgent);
```

darzu:
> Second condition is redundant if the first condition is true right?

Yes, good point

PR #5358, opened 2019-03-18
Show error if hex file cannot be read

darzu:
Fixes: https://github.com/Microsoft/pxt-microbit/issues/1560

image

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1097,9 +1097,8 @@ export class ProjectView
         pxt.cpp.unpackSourceFromHexFileAsync(file)
             .then(data => this.importHex(data))
             .catch(e => {
-                pxt.log(e);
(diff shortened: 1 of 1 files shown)


webapp/src/app.tsx
```diff
@@ -1095,7 +1095,11 @@ export class ProjectView
     importHexFile(file: File) {
         if (!file) return;
         pxt.cpp.unpackSourceFromHexFileAsync(file)
-            .done(data => this.importHex(data));
+            .then(data => this.importHex(data))
+            .catch(e => {
+                pxt.log(e);
+                core.warningNotification(lf("Sorry, we could not recognize this file."))
```
darzu: How does localization work? Is this all I need to add or do I need to add an entry somewhere else as well?
@pelikhan

PR #5362, opened 2019-03-18
Handle invalid block XML when creating flyout

darzu:
Before this fix, searching in Arcade would occasionally give you this error due to some invalid blocks xml (separate issue):

image

When there's a problem with xml block xml, we should still gracefully handle the flyout.


darzu:
> invalid

Hmm maybe. Not easily. By the time I'm at this point all I see is null in a list. A better place would be upstream somewhere. Like where this log error is coming from:
image

I'm going to dig into that error now and see if I can't 1) fix the error and 2) report invalid blocks better

PR #5363, opened 2019-03-18
Ensure "builtin" flag is correctly set for search results

darzu:
Before when searching for "string" you could see errors like this in the console:

image

This caused problems downstream occasionally (e.g. https://github.com/Microsoft/pxt/pull/5362)

This was because the builtin blocks (e.g. text_length) are supposed to have the "builtinBlock" flag set to true in the SearchInfo type, but this wasn't being correctly set because of an issue in how the builtin and non-builtin search sets where combined (the non-builtin search set included some builtins that overwrote the builtin search set).

After this PR, these errors no longer appear in the console.


darzu:
@riknoll , here's the search change I made to de-dupe built-in blocks.

PR #5368, opened 2019-03-20
Add TypeScript -> Python decompiler

darzu:
Work in progress decompiler from TypeScript to Python.

So far we can successfully decompile:
https://github.com/Microsoft/pxt/blob/dazuniga/python/tests/pydecompile-test/cases/game.ts
into:
https://github.com/Microsoft/pxt/blob/dazuniga/python/tests/pydecompile-test/baselines/game.py
which in turn successfully compiles into game.ts

Lots of work left to do.
TODO:
- [ ] add py decompile hooks to the triple toggle
- [ ] decompile all .ts from tests/decompile-tests/cases/
- [ ] reuse code between blocks decompiler and pydecompiler
- [ ] handle `myString.length` -> `len(myString)`; might need type info
- [ ] handle more and more of the language

To run the pydecompiler tests:
```
jake testpydecompiler
```

For quick local development,
run this in one terminal:
```
cd pxt/pxtcompiler; tsc -w
```
and this in another window:
```
cd pxt; ./watch-pydecompiletests.sh
```
this will re-run the python decompiler test suite every time you modify the compiler/decompiler. This works much faster than using jake.


darzu:
@pelikhan , @mmoskal  FYI, my work-in-progress decompiler. Works for Michal's sample game

PR #5370, opened 2019-03-20
Revert "Fix browser detection of Firefox on Mac"

abchatra:
Reverts Microsoft/pxt#5357


abchatra:
Change breaks minecraft

PR #5371, opened 2019-03-20
Forcing restart when toggling slow-mo

gushuro:
Fixing https://github.com/Microsoft/pxt-microbit/issues/1869.

- [x] cleaning up startSimulator/stopSimulator/restartSimulator to use the React state to determine if a debug build is necessary
- [x] use options bag in startSimulator to avoid unnoticed breaking changes
- [x] adding a "pending" state in the simulator driver that tells the IDE that an code operation has started (like blocks->js). pre-state before "starting" state.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1871,7 +1871,7 @@ export class ProjectView
             simulator.setTraceInterval(intervalSpeed || simulator.SLOW_TRACE_INTERVAL);
         }
         this.setState({ tracing: !tracing, debugging: debugging && !tracing },
-            () => this.startSimulator())
+            () => this.restartSimulator())
     }
 
     setTrace(enabled: boolean, intervalSpeed?: number) {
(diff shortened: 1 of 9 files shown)


webapp/src/app.tsx
```diff
@@ -1864,15 +1864,16 @@ export class ProjectView
...
+        if (tracing) {
             this.editor.clearHighlightedStatements();
             simulator.setTraceInterval(0);
         }
         else {
             simulator.setTraceInterval(intervalSpeed != undefined ? intervalSpeed : simulator.SLOW_TRACE_INTERVAL);
         }
-        this.setState({ tracing: !this.state.tracing })
-        this.restartSimulator();
+        this.setState({ tracing: !tracing })
```
pelikhan: why is restartSimulator not hanlding the "tracing" mode?

webapp/src/app.tsx
```diff
@@ -2020,6 +2021,15 @@ export class ProjectView
         this.blocksEditor.setBreakpointsFromBlocks();
     }
 
+    restartSimulatorTrace(trace?: boolean) {
+        if (this.state.simState == pxt.editor.SimState.Stopped
```
pelikhan: why is this not part of restartSimulator?

webapp/src/app.tsx
```diff
@@ -1864,15 +1864,16 @@ export class ProjectView
...
 
     toggleTrace(intervalSpeed?: number) {
-        if (this.state.tracing) {
+        let tracing = this.state.tracing;
+        if (tracing) {
             this.editor.clearHighlightedStatements();
             simulator.setTraceInterval(0);
         }
         else {
             simulator.setTraceInterval(intervalSpeed != undefined ? intervalSpeed : simulator.SLOW_TRACE_INTERVAL);
```
pelikhan: ``!== undefined``

pelikhan: ACtualy you could just do ``intervalSpeed || simulator.SLOW_TRACE_INTERNAL``

webapp/src/app.tsx
```diff
@@ -1864,15 +1864,16 @@ export class ProjectView
...
         }
-        this.setState({ tracing: !this.state.tracing })
-        this.restartSimulator();
+        this.setState({ tracing: !tracing })
+        this.startSimulator(true);
```
pelikhan: chain it in the setState -> setState({ tracing: !tracing}, () => this.startSimulator(true))

gushuro: Sure

webapp/src/app.tsx
gushuro: Yep

webapp/src/app.tsx
```diff
@@ -1864,15 +1864,15 @@ export class ProjectView
...
+        } else {
+            simulator.setTraceInterval(intervalSpeed || simulator.SLOW_TRACE_INTERVAL);
         }
-        else {
-            simulator.setTraceInterval(intervalSpeed != undefined ? intervalSpeed : simulator.SLOW_TRACE_INTERVAL);
-        }
-        this.setState({ tracing: !this.state.tracing })
-        this.restartSimulator();
+        this.setState({ tracing: !tracing }, () => this.startSimulator(true))
+        this.startSimulator(true);
```
pelikhan: you are starting twince the simulator + why is the argument true always?

webapp/src/app.tsx
```diff
@@ -2144,16 +2154,15 @@ export class ProjectView
         pxt.log("turning debugging mode to " + state);
         this.setState({ debugging: state, tracing: false }, () => {
             this.renderCore()
+            const blocks = this.blocksEditor.editor.getAllBlocks();
```
pelikhan: it's not guaranteed that the blocks editor is loaded yet. You should not leak this abstraction here.

pelikhan:
In genreal, app.tsx should have have access to the blockly workspace.

PR #5372, opened 2019-03-20
removing ability to use htmlBody

pelikhan:
We are now use React for all our modal dialogs.


riknoll:
👍

PR #5376, opened 2019-03-21
 Adding ability to run getters in the debugger.

gushuro:
We add the ability to run getters in the debugger. 
When sending a variable's children request, now we can specify which ones of the class getters we should call to obtain values.

On the webapp side, we use this api to get getters marked to be displayed in their comment attributes.

### DIFF
--- a/pxtlib/service.ts
+++ b/pxtlib/service.ts
@@ -158,6 +158,7 @@ namespace ts.pxtc {
         advanced?: boolean;
         deprecated?: boolean;
         useEnumVal?: boolean; // for conversion from typescript to blocks with enumVal
+        callInDebugger?: boolean; // for getters, they will be invoked by the debugger.
         // On block
         subcategory?: string;
         group?: string;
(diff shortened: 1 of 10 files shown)


pxtsim/debugger.ts
```diff
@@ -124,29 +124,53 @@ namespace pxsim {
...
             }
             return r
         }
 
-        return frameVars(v);
+        return frameVars(v, fields);
+    }
+
+    function evalGetter(fn: LabelFn, target: RefObject) {
+        let p: any = {
```
riknoll: Let's add some comments (or maybe not use single character var names)

pxtlib/service.ts
```diff
@@ -158,6 +158,7 @@ namespace ts.pxtc {
         advanced?: boolean;
         deprecated?: boolean;
         useEnumVal?: boolean; // for conversion from typescript to blocks with enumVal
+        callInDebugger?: boolean; // for getters, they will be invoked by the debugger.
```
mmoskal: callFromDebugger maybe?

pxtsim/debugger.ts
riknoll: (this was my fault)

gushuro: Sure!

pxtlib/service.ts
gushuro: Already merged it in https://github.com/Microsoft/pxt-common-packages/pull/762, we can change it later if we feel it's not clear enough.

PR #5379, opened 2019-03-21
Numbertostringblock

pelikhan:
Block to convert number to string, should we expose "Object.toString" ?
image

Fix for https://github.com/Microsoft/pxt-microbit/issues/1637

### DIFF
--- a/common-docs/reference/math/to-string.md
+++ b/common-docs/reference/math/to-string.md
@@ -1,0 +1,7 @@
+# to string
+
+Converts a number into its text representation
+
+```sig
+(123).toString()
+```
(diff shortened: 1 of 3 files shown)


jwunderl:
Looks like this is just failing because the ``variables_reserved_names`` blockly compiler test result changed to append a 2 [here](https://github.com/Microsoft/pxt/blob/e398e3bc29a9caa13022f9353269061b946cf9e6/tests/blocklycompiler-test/baselines/variables_reserved_names.ts#L2 ) (``let Number2 = 0``), which is probably good because using ``Number`` as a variable name seems to be broken right now for microbit (https://makecode.microbit.org/_1vcE9pD34H1R) and forcing a renaming fixes that.

Any reason this is marked as a draft besides possibly adding `Object.toString`?

pelikhan:
We should have a block that accepts any reporter and converts it to string. We need compiler/decompiler tests.

jwunderl:
This should be handled by https://github.com/Microsoft/pxt-common-packages/pull/781

PR #5382, opened 2019-03-21
Improvements to the Debug Mode UI.

gushuro:
Removed Blocks/Javascript toggle and added a title instead, centered and
not clickable.
Changed debugMode color to Orange (matching tutorial one).
Disabled Step buttons when simulator not paused on breakpoint.

Disabled Run/Restart button when simulator is loading, and in debug mode.

topBar

### DIFF
--- a/theme/debugger.less
+++ b/theme/debugger.less
@@ -11,6 +11,19 @@
         Debugger
 *******************************/
 
+/* Top bar menu */
+.menubar .debugger-menu-item.centered {
+    position: fixed;
+    place-content: center;
+    right: 0;
+    left: 0;
+    margin-right: auto;
+    margin-left: auto;
+    min-height: inherit;
+    width: 33%;
+}
+
+/* Debugger toolbox */
 .debuggerToolbox.elements {
     height: 100%;
     display: flex;
--- a/theme/debugger.less
+++ b/theme/debugger.less
@@ -73,6 +73,10 @@
     color: rgb(255, 21, 0);
 }
 
+.ui.item.link.dbg-btn.disabled {
+    color: rgba(40, 40, 40, .3);
+}
+
 /* Debugger variables view */
 .ui.segment.debugvariables {
     table-layout: auto;
--- a/theme/themes/pxt/collections/menu.overrides
+++ b/theme/themes/pxt/collections/menu.overrides
@@ -77,3 +77,14 @@
 .debugger .ui.inverted.menu {
     background: @DebuggerBackgroundPrimaryColor !important;
 }
+
+/* Debug mode */
+.menubar .debugger-menu-item.centered {
+    position: fixed;
+    place-content: center;
+    right: 0;
+    left: 0;
+    margin-right: auto;
+    margin-left: auto;
+    min-height: inherit;
+}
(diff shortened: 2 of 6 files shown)


webapp/src/debugger.tsx
```diff
@@ -313,18 +316,18 @@ export class DebuggerToolbar extends data.Component<DebuggerToolbarProps, Debugg
...
-                        <sui.Item key='dbgpauseresume' className={`dbg-btn dbg-pause-resume ${isDebuggerRunning ? "pause" : "play"}`} icon={`${isDebuggerRunning ? "pause blue" : "play green"}`} title={dbgPauseResumeTooltip} onClick={this.dbgPauseResume} />
-                        <sui.Item key='dbgstepover' className={`dbg-btn dbg-step-over`} icon={`xicon stepover ${isDebuggerRunning ? "disabled" : "blue"}`} title={dbgStepOverTooltip} onClick={this.dbgStepOver} />
-                        <sui.Item key='dbgstepinto' className={`dbg-btn dbg-step-into`} icon={`xicon stepinto ${isDebuggerRunning ? "disabled" : ""}`} title={dbgStepIntoTooltip} onClick={this.dbgStepInto} />
-                        <sui.Item key='dbgstepout' className={`dbg-btn dbg-step-out`} icon={`xicon stepout ${isDebuggerRunning ? "disabled" : ""}`} title={dbgStepOutTooltip} onClick={this.dbgStepOut} />
+                        <sui.Item key='dbgpauseresume' className={`dbg-btn dbg-pause-resume ${dbgStepDisabled ? "disabled" : ""} ${isDebuggerRunning ? "pause" : "play"}`} icon={`${isDebuggerRunning ? "pause blue" : "play green"}`} title={dbgPauseResumeTooltip} onClick={this.dbgPauseResume} />
```
riknoll: Might be a little cleaner to do it like this:

```typescript
let debugStepClass = dbgStepDisabled ? "disabled" : "";
```
Instead of replicating the ternary expression

theme/themes/pxt/collections/menu.overrides
```diff
@@ -76,4 +76,15 @@
...
+
+/* Debug mode */
+.menubar .debugger-menu-item.centered {
+    position: fixed;
+    place-content: center;
+    right: 0;
+    left: 0;
+    margin-right: auto;
+    margin-left: auto;
+    min-height: inherit;
```
riknoll: Why is this not in `debugger.less`

gushuro: I thought the top bar css was here. I can move it if you think it should go there.

webapp/src/debugger.tsx
gushuro: Sure, makes sense

webapp/src/debugger.tsx
```diff
@@ -290,6 +290,7 @@ export class DebuggerToolbar extends data.Component<DebuggerToolbarProps, Debugg
 
         const simState = parentState.simState;
         const isRunning = simState == pxt.editor.SimState.Running;
+        const isStarting = simState == pxt.editor.SimState.Starting;
```
pelikhan: There is also a pending state where the editor is transitioning between editors.

gushuro: Yes, but we are not allowing to change editors while in debug mode

theme/debugger.less
```diff
@@ -73,6 +73,10 @@
     color: rgb(255, 21, 0);
 }
 
+.ui.item.link.dbg-btn.disabled {
+    color: rgba(40, 40, 40, .3);
```
pelikhan: theme it.

PR #5384, opened 2019-03-22
respect autorun flag

pelikhan:
When autoRun is false, respect it and don't run the simulator at least once.


mmoskal:
Hold on a second, will push fix in one minute here

PR #5388, opened 2019-03-23
Allow for ![](/docs/static/foo.png) in markdown

mmoskal:
This replaces `/docs/static` with `/static` in image URLs. This lets them be rendered by github as well as our backend. It doesn't touch existing syntax, but going forward we can use `/docs/static/...` to allow rendering both in our websites and on github.

For example see here https://github.com/Microsoft/pxt-arcade/blob/dbgconn/docs/hardware/dbg.md

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -2272,6 +2272,8 @@ function renderDocs(builtPackaged: string, localDir: string) {
                 let str = buf.toString("utf8")
                 if (/\.md$/.test(f)) {
                     str = nodeutil.resolveMd(".", f.substr(5, f.length - 8));
+                    // patch any /static/... url to /docs/static/...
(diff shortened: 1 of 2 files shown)


jwunderl:
Minor benefit, but after everything is updated so the links work, switching the links over to the new syntax would also resolve https://github.com/Microsoft/pxt/issues/5021, correct?

mmoskal:
I suppose, though I guess we should also fix staticpkg...

pelikhan:
We need to patch the code cards too.

pelikhan:
@mmoskal added a patch for static pkg as well.

jwunderl:
Also fixes https://github.com/Microsoft/pxt-arcade/issues/805 besides the simulator portion

PR #5392, opened 2019-03-25
Static Python

pelikhan:
Static Python Experiment build.

* __uses 5.7.* builds__


webapp/src/monaco.tsx
```diff
@@ -63,15 +63,14 @@ class PythonCompletionProvider implements monaco.languages.CompletionItemProvide
...
-        // TODO python
         return compiler.completionsAsync(fileName, offset, source)
             .then(completions => {
                 const items = pxt.Util.values(completions.entries || {}).map(si => {
                     return {
                         label: completions.isMemberCompletion ? si.pyName : si.pyQName,
                         kind: this.tsKindToMonacoKind(si.kind),
                         documentation: si.attributes.jsDoc,
-                        detail: si.parameters ? "(" + si.parameters.map(p => p.name).join(", ") + ")" : "", // TODO improve this
+                        detail: si.pySnippet
```
mmoskal: This is help that shows. There is another field for text to insert.

pelikhan:
There is another service to handle the signature. I need to wire it up.

The problem for events is that we need to insert the function definition above the method call.

darzu:
Alright, I de-duped a bunch of files between the blocks decompiler tests and the python decompiler tests. Now this PR only has 202 files changed instead of 334! :P

PR #5395, opened 2019-03-26
Make breakpoints pause countdown and fibers running in the background.

gushuro:
When hitting a breakpoint, pause all fibers running in the background. After resuming via step or continue, restore all fibers.

### DIFF
--- a/pxtsim/runtime.ts
+++ b/pxtsim/runtime.ts
@@ -440,7 +440,7 @@ namespace pxsim {
     }
 
     export class PausedTimeout {
-        constructor(public fn: any, public timeRemaining: number) { }
+        constructor(public fn: Function, public timeRemaining: number) { }
     }
 
 
@@ -756,7 +756,7 @@ namespace pxsim {
                 Runtime.postMessage(msg)
                 breakAlways = false;
                 breakFrame = null;
-                runtime.pauseScheduled();
+                __this.pauseScheduled();
                 dbgResume = (m: DebuggerMessage) => {
                     dbgResume = null;
                     dbgHeap = null;
@@ -1075,7 +1075,7 @@ namespace pxsim {
         }
 
         // Wrapper for the setTimeout
-        schedule(fn: () => void, timeout: number): number {
+        schedule(fn: Function, timeout: number): number {
             // We call the timeout function and add its id to the timeouts scheduled.
             if (timeout <= 0) return -1;
             let id = setTimeout(fn, timeout);
@@ -1111,8 +1111,9 @@ namespace pxsim {
 
         // Removes from the timeouts scheduled list all the ones that had been fulfilled.
         cleanScheduledExpired() {
+            let now = U.now();
             this.timeoutsScheduled = this.timeoutsScheduled.filter(ts => {
-                let elapsed = U.now() - ts.timestampCall;
+                let elapsed = now - ts.timestampCall;
                 return ts.totalRuntime > elapsed;
             })
         }
--- a/pxtsim/runtime.ts
+++ b/pxtsim/runtime.ts
@@ -436,25 +436,11 @@ namespace pxsim {
     }
 
     export class TimeoutScheduled {
-        constructor(id: number, fn: any, totalRuntime: number, timestampCall: number) {
-            this.id = id;
-            this.fn = fn;
-            this.totalRuntime = totalRuntime;
-            this.timestampCall = timestampCall;
-        }
-        id = -1;
-        fn = () => { };
-        totalRuntime = 0;
-        timestampCall = 0;
+        constructor(public id: number, public fn: Function, public totalRuntime: number, public timestampCall: number) { }
     }
 
     export class PausedTimeout {
-        constructor(fn: any, timeRemaining: number) {
-            this.fn = fn;
-            this.timeRemaining = timeRemaining;
-        }
-        fn = () => { };
-        timeRemaining = 0;
+        constructor(public fn: any, public timeRemaining: number) { }
     }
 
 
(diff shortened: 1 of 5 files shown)


pxtsim/runtime.ts
```diff
@@ -435,6 +435,29 @@ namespace pxsim {
         return pxtcore.mkAction(0, s => _leave(s, f(s)))
     }
 
+    export class TimeoutScheduled {
+        constructor(id: number, fn: any, totalRuntime: number, timestampCall: number) {
```
riknoll: TypeScript has shortcuts for this kind of class; you can rewrite it like this:

```typescript
export class TimeoutScheduled {
    constructor(public id: number, public fn: Function, public totalRuntime: number, public timestampCall: number) { }
}
```

Same for `PausedTimeout`

pxtsim/runtime.ts
```diff
@@ -450,13 +473,18 @@ namespace pxsim {
...
+        pausedTime = 0;
+        lastPauseTimestamp = 0;
         id: string;
         globals: any = {};
         currFrame: StackFrame;
         entry: LabelFn;
         loopLock: Object = null;
         loopLockWaitList: (() => void)[] = [];
 
+        timeoutsScheduled: TimeoutScheduled[] = []
```
riknoll: Why two arrays? Seems like only one will ever be used at a time; you're either paused or running.

gushuro: That's true, only one of them is going to be used at once. But I think having two arrays makes it more clear, since their contents are different. And other issues arise  if we were to use only one, like what would be the ID of a paused timeout? It could be the one from the first call, but it's gonna change when we resume it.

gushuro: I can change it if you disagree with that

pxtsim/runtime.ts
```diff
@@ -751,21 +784,21 @@ namespace pxsim {
 
                     switch (m.subtype) {
                         case "resume":
-                            break
+                            break;
```
pelikhan: adding random ; makes the review harder

pxtsim/runtime.ts
```diff
@@ -1054,6 +1087,47 @@ namespace pxsim {
             c.start = 0;
             c.numstops++;
         }
+
+        // Wrapper for the setTimeout
+        schedule(fn: () => void, timeout: number): number {
+            // We call the timeout function and add its id to the timeouts scheduled.
+            if (timeout <= 0) return -1;
+            let id = setTimeout(fn, timeout);
+            runtime.timeoutsScheduled.push(new TimeoutScheduled(id, fn, timeout, Date.now()));
```
pelikhan: this is an instance method so it can just use "this" ?

pelikhan: we use U.now() in other places.

pxtsim/runtime.ts
```diff
@@ -1054,6 +1087,47 @@ namespace pxsim {
...
+            return id;
+        }
+
+        // On breakpoint, pause all timeouts
+        pauseScheduled() {
+            runtime.timeoutsScheduled.forEach(ts => {
```
pelikhan: instance function should use this

webapp/src/debugger.tsx
```diff
@@ -124,7 +124,7 @@ export class DebuggerVariables extends data.Component<DebuggerVariablesProps, De
                                 });
                         } else {
                             let children: pxt.Map<Variable> = {};
-                            Object.keys(msg.variables).forEach(variableName => {
+                            Object.keys(msg.variables || {}).forEach(variableName => {
```
pelikhan: if(msg.variables) ...

pxtsim/runtime.ts
gushuro: I've seen both around, is the Util one preferred by you?

webapp/src/debugger.tsx
gushuro: Sure

pxtsim/runtime.ts
pelikhan: yes use U.now()

pxtsim/runtime.ts
```diff
@@ -1112,7 +1112,7 @@ namespace pxsim {
         // Removes from the timeouts scheduled list all the ones that had been fulfilled.
         cleanScheduledExpired() {
             this.timeoutsScheduled = this.timeoutsScheduled.filter(ts => {
-                let elapsed = Date.now() - ts.timestampCall;
+                let elapsed = U.now() - ts.timestampCall;
```
pelikhan: move outside of loop

pelikhan:
Please add a description of your changes.

pxtsim/runtime.ts
```diff
@@ -435,6 +435,15 @@ namespace pxsim {
         return pxtcore.mkAction(0, s => _leave(s, f(s)))
     }
 
+    export class TimeoutScheduled {
+        constructor(public id: number, public fn: Function, public totalRuntime: number, public timestampCall: number) { }
+    }
+
+    export class PausedTimeout {
+        constructor(public fn: any, public timeRemaining: number) { }
```
pelikhan: avoid any

pxtsim/runtime.ts
```diff
@@ -739,10 +754,14 @@ namespace pxsim {
                 const { msg, heap } = getBreakpointMsg(s, brkId);
                 dbgHeap = heap;
                 Runtime.postMessage(msg)
+                breakAlways = false;
+                breakFrame = null;
+                runtime.pauseScheduled();
```
pelikhan: runtime == __this ?

pxtsim/runtime.ts
gushuro: Yep

pxtsim/runtime.ts
gushuro: Replaced with Function

pxtsim/runtime.ts
gushuro: it should be `__this`

pelikhan:
Feels like runtime should be in charge of “now” and all other services can rely on it. Avoids propagation of time drift compensation.

PR #5406, opened 2019-03-29
Debug from python

pelikhan:
- [x] a quick cleanup to avoid crashing when starting the debugger from python
- [x] allow debugger to start when simulator is stopped

This goes into the python branch but could be integrated somewhere else.


gushuro:
Yeah, all editor-specific changes should be handled by each editor separately.
LGTM

PR #5407, opened 2019-03-29
Making blocks nonmovable and noneditable in debug mode.

gushuro:
Updating blockly.d.ts and blockly_compressed.
#### Merge only after https://github.com/Microsoft/pxt-blockly/pull/216


gushuro:
> See my PR that fixes crashes when the debugger is started in JavaScript. This should move in the blocks editor class.

Yeah, agreed.

PR #5408, opened 2019-03-29
Allow undefined in arrays

jwunderl:
fixes https://github.com/Microsoft/pxt-arcade/issues/311

Allow undefined / null in arrays rather than using default values to match behavior on hardware / JavaScript

https://github.com/Microsoft/pxt/pull/1055 is where default values for arrays was added to the sim, so worth noting it's a longstanding behavior that has changed for hardware already. For example, the output of the following snippet

```typescript
let myNums: number[] = [];
console.log("a: " + myNums[5]);
myNums[5] = 4;
console.log("b: " + myNums[5]);
myNums[6] += 3;
console.log("c: " + myNums[6]);

let myStrs: string[] = [];
myStrs[5] += "234";
console.log("d: " + myStrs[5]);
```

changes in the simulator from

```
a: 0
b: 4
c: 3
d: 0234
```

to

```
a: undefined
b: 4
c: NaN
d: undefined234
```

### DIFF
--- a/docs/static/playground/pxt-common/pxt-helpers.js
+++ b/docs/static/playground/pxt-common/pxt-helpers.js
@@ -71,7 +71,7 @@ namespace helpers {
         for (let i = 0; i < len; ++i) {
             if (i > 0 && sep)
                 r += sep;
-            r += arr[i] || "";
+            r += (arr[i] === undefined || arr[i] === null) ? "" : arr[i];
         }
         return r;
     }
--- a/docs/static/playground/pxt-common/pxt-helpers.js
+++ b/docs/static/playground/pxt-common/pxt-helpers.js
@@ -65,7 +65,8 @@ namespace helpers {
         return arr.removeAt(0);
     }
 
-    export function arrayJoin<T>(arr: T[], sep: string): string {
+    export function arrayJoin<T>(arr: T[], sep?: string): string {
+        sep = sep || ",";
         let r = "";
         let len = arr.length // caching this seems to match V8
         for (let i = 0; i < len; ++i) {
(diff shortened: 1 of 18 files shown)


docs/static/playground/pxt-common/pxt-helpers.js
```diff
@@ -65,13 +65,14 @@ namespace helpers {
...
-    export function arrayJoin<T>(arr: T[], sep: string): string {
+    export function arrayJoin<T>(arr: T[], sep?: string): string {
+        sep = sep || ",";
         let r = "";
         let len = arr.length // caching this seems to match V8
         for (let i = 0; i < len; ++i) {
             if (i > 0 && sep)
                 r += sep;
-            r += arr[i] || "";
+            r += arr[i] == undefined ? "" : arr[i];
```
pelikhan: use ===

riknoll:
@pelikhan is it breaking if this matches the hardware?

pelikhan:
I mean we need to do testing on this. It impacts all array accesses.

abchatra:
Hmm I didn't know we had undefined in number array at the C++. Are you folks sure? 
Seems like it is returning zero https://github.com/Microsoft/pxt-common-packages/blob/master/libs/base/pxt.cpp#L278

riknoll:
https://github.com/Microsoft/pxt-common-packages/blob/85959d59f561e1cb349e02e1137f6e11a68ff881/libs/base/pxtbase.h#L501

mmoskal:
I think this is unlikely to break much, since it would have been already broken on hardware. I think we should merge it. @abchatra

Issue #5396, opened 2019-03-26
Array join() drops zero values as nulls


ganicke:
Values of `0` are omitted from the text rows on output. Seems that ``array.join()`` treats values of `0` as `null` and drops them out of the result string. See below:

```typescript
let testRows: number[][] = []
testRows = [[23, 9, 56], [87, 0, 45], [75, 0, 0], [0, 32, 45]]
for (let row of testRows) {
	console.log(row.join(','))
}
```

result:

```
23,9,56
87,,45
75,,
,32,45
```

jwunderl:
For completeness;

https://github.com/Microsoft/pxt-arcade/issues/311 is a blocker on fixing this, as it seems like all ``null``s and ``undefined``s in arrays are being stored as just 0s no matter the type (e.g. https://makecode.com/_DTuJb5Ts44F5 and below). https://github.com/jwunderl/pxt/tree/arrjoin is a fix for this issue when that is fixed.

```typescript
let myNumArr = [0, 1, 2, 3, undefined, null];
let myBoolArr = [true, false, undefined, null];
let myStringArr = ["he", null, undefined, " ", "123"];

console.log("nums: " + myNumArr.join(","));
console.log("bools: " + myBoolArr.join(","));
console.log("str: " + myStringArr.join(","));
```

results in

```
nums: 0,1,2,3,0,0
bools: true,false,0,0
str: he,0,0, ,123
```

PR #5412, opened 2019-03-29
Match regular JavaScript for Array.join

jwunderl:
split off from #5408 

fixes https://github.com/Microsoft/pxt-arcade/issues/784
fixes https://github.com/Microsoft/pxt/issues/5396

This matches https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join behavior for ``Array.join`` -- separator defaults to ``,``, and only ``undefined`` and ``null`` are replaced with the empty string, not all falsey values. Note that the second part of this won't behave nicely in the simulator without #5408, as 
all values default to 0. That is, 

```typescript
let mySprites: Sprite[] = [];
mySprites[5] = undefined;
console.log(mySprites.join());
```

results in ``0,0,0,0,0,0`` in the simulator and ``,,,,,`` on hardware

### DIFF
--- a/libs/pxt-common/pxt-helpers.ts
+++ b/libs/pxt-common/pxt-helpers.ts
@@ -65,13 +65,14 @@ namespace helpers {
         return arr.removeAt(0);
     }
 
-    export function arrayJoin<T>(arr: T[], sep: string): string {
+    export function arrayJoin<T>(arr: T[], sep?: string): string {
+        sep = sep || ",";
         let r = "";
         let len = arr.length // caching this seems to match V8
         for (let i = 0; i < len; ++i) {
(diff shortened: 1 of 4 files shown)


libs/pxt-common/pxt-helpers.ts
```diff
@@ -65,13 +65,14 @@ namespace helpers {
         return arr.removeAt(0);
     }
 
-    export function arrayJoin<T>(arr: T[], sep: string): string {
+    export function arrayJoin<T>(arr: T[], sep?: string): string {
+        sep = sep || ",";
```
riknoll: sep could be empty string

PR #5425, opened 2019-04-03
Changed puchdb Version to 7.0.0 

frankbauer:
Fixed install and build errors with node v10.11.0 on MacOS Mojave.


webapp/src/db.ts
```diff
@@ -4,7 +4,9 @@ import * as Promise from "bluebird";
 
 const PouchDB = require("pouchdb");
 /* tslint:disable:no-submodule-imports TODO(tslint) */
-require('pouchdb/extras/memory');
+PouchDB.plugin(require('pouchdb-adapter-cordova-sqlite'));
```
pelikhan: Do we need these?

frankbauer: So far it builds and runs fine without those modules. I included it since it was part of the original pouchdb.

pelikhan:
I think we can (want) drop those if we can :)

frankbauer:
I think we can remove the dependencies as in `package.json` as well. I can build another clean pull request if that is ok?

pelikhan:
You could just update this one.

pelikhan:
Looks like PouchDB is not supported on IE anymore. Needs to revert this.

frankbauer:
Hm, maybe we need the websql plugin for this case:

> PouchDB supports all modern browsers, using IndexedDB under the hood and falling back to WebSQL where IndexedDB isn't supported.
(from [pouchdb.com](https://pouchdb.com/learn.html#browser_support))

Unfortunately I can not test this.

PR #5429, opened 2019-04-04
use icons instead of text

pelikhan:
- [x] avoid using "steps" which takes too much space. Use the "dot" icon instead. 
- [x] display trophy/green when all steps completed
- [x] record finished tutorial data so that trophy stays after finish

(not final design, but small fixes to current story)

image


abchatra:
@pelikhan I don't understand the white filled circle in the progress 3/13 image.

pelikhan:
Dots from the tutorials. Anyhow, seems like we have a new design that will move this label somewhere. Removing the "steps" to avoid ugly overlays.

PR #5435, opened 2019-04-05
Fix svg cursorPoints for safari in iOS

jwunderl:
fixes https://github.com/Microsoft/pxt-microbit/issues/1382 (this behavior also applied to the compass heading and light level at a glance; looks like any svg simulator buttons that required x/y positions)

it looks like the accelerometer is also broken on iOS safari, though; it's probably the same type thing as this, will give it a look tomorrow morning


abchatra:
Awesome @jwunderl

PR #5441, opened 2019-04-07
Add base reference page for 'to-string'

ganicke:
- [x] Put in a base reference page for **toString()**.
- [x] Add help paths for number.toString() and boolean.toString().
- [x] Add 'toString()' to card page.


ganicke:
There's no **convertToText()** in core so I'll close this and add the ref page to _./libs/base_.

PR #5443, opened 2019-04-08
Add builtin 'string compare' block

jwunderl:
fixes https://github.com/Microsoft/pxt-microbit/issues/1050

Adds a new builtin block that is the same as normal comparison, but contains strings by default instead of numbers

Screen Shot 2019-04-08 at 11 35 49 AM


jwunderl:
@jaustin: the block you requested in https://github.com/Microsoft/pxt-microbit/issues/1050 was added here ~

abchatra:
Also please add tests to compiler & decompiler

PR #5445, opened 2019-04-09
Add semantic tests for: ts2py, py2ts, and static typescript compiler based on runtime traces compared with python3 & tsc+node.js as baselines

darzu:
# What
Adds  a test suite that compares the runtime semantics of Static Typescript ["STS"] and Static Python ["SPY"] files with tsc+node.js ["TS"] and python3 ["PY"]. This is useful for testing the correctness of: ts2py, py2ts, and STS.

# Why
This suite was added primarily to test ts2py but also tests py2ts and the sts compiler.

These test supplement but do not replace the pydecompiler-tests. Instead of comparing the syntax of decompiled programs, they test the semantics of decompiled programs by looking at the programs trace. This makes it better for certain kinds of tests like testing variable scope semantics and worse at other types of tests like auto-generated variable names.

This test suite enables TDD of many ts2py features.

# How
The test suite tests all of or any of: STS, ts2py, and py2ts, but will still localize errors to one of {STS, ts2py, py2ts}, as shown below.

The inputs consist of .ts and .py files with console.log/print statements (but without assert statements)

The chain of tests that are run on each .ts file:

1. TS(foo.ts) => baseline trace
2. STS(foo.ts) => static typescript trace, a difference here would indicate a STS problem
3. ts2py(foo.ts) => gives foo.ts.py
4. PY(foo.ts.py) => if different than baseline trace then this indicates a ts2py problem
5. py2ts(foo.ts.py) => gives foo.ts.py.ts, if different than foo.ts it's not necessarily an error but may help localize issues in step 6 & 7
6. TS(foo.ts.py.ts) => if different than baseline trace then this probably indicates a py2ts problem
7. STS(foo.ts.py.ts) => if different than baseline trace then this probably indicates a STS problem

The chain of tests that are run on each .py file:

1. PY(foo.py) => baseline trace
2. py2ts(foo.py) => gives foo.py.ts
3. TS(foo.py.ts) => if different from baseline trace, indicates a py2ts problem
4. STS(foo.py.ts) => if different from baseline trace, indicates a STS problem
5. ts2py(foo.py.ts) => gives foo.py.ts.py, if different than foo.py it's not necessarily an error but may help localize issues in step 6
6. PY(foo.py.ts.py) => if different from baseline trace, probably indicates a ts2py problem
 
Notice how each test step localizes problems. Also, these tests don't depend on checked-in baselines or asserts, don't care about the syntax that ts2py and py2ts output since this may change over time, and can be written in either .py or .ts depending on which is most appropriate but both .ts and .py files will test each of {STS, py2ts, ts2py}.

## Adding a test
Simply add a .ts or .py file to tests/runtime-trace-tests/cases/ that exhibits the behavior you want to test. 

### Example: mutating loop variable semantics
For example, to ensure that ts2py correctly handles for loops where the loop variable is mutated during the loop, you could add a .ts file with:
```
for (let i = 0; i < 5; i++) {
    i++
    console.log(i)
}
```
Then run `jake testtraces`.
At the time of this writing, this test will fail on step #4 mentioned above because ts2py converts the file to:
```
for i in range(5):
  i += 1
  print(i)
```
which produces a different trace when run with python3 than the .ts file does when run with tsc+node.js.
Notice that it only took a simple 4 line program and no baseline files or assert statements to create a test that tests correctness of ts2py, py2ts, and sts.

This example is included as "TODO_scope_for_loop.ts". Note that files with "TODO_" will be ignored by the test runner.

# TODO: py2ts support
Currently the py2ts and later portion of this test suite is disabled due to issues I'm having running py2ts in unit tests. Relevant locations are marked with TODOs in the tracerunner.ts
My plan is to fix this in a follow up PR as the rest of the test suite is still useful for testing ts2py and sts.


tests/runtime-trace-tests/readme.md
```diff
@@ -0,0 +1,64 @@
...
+4. STS(foo.py.ts) => if different from baseline trace, indicates a STS problem
+5. ts2py(foo.py.ts) => gives foo.py.ts.py, if different than foo.py it's not necessarily an error but may help localize issues in step 6
+6. PY(foo.py.ts.py) => if different from baseline trace, probably indicates a ts2py problem
+ 
+Notice how each test step localizes problems. Also, these tests don't depend on checked-in baselines or asserts, don't care about the syntax that ts2py and py2ts output since this may change over time, and can be written in either .py or .ts depending on which is most appropriate but both .ts and .py files will test each of {STS, py2ts, ts2py}.
+
+## Adding a test
+Simply add a .ts or .py file to tests/runtime-trace-tests/cases/ that exhibits the behavior you want to test. 
+
+### Example: mutating loop variable semantics
```
darzu: @riknoll , here's an example of how to add test cases

tests/runtime-trace-tests/cases/for_loop.ts
```diff
@@ -0,0 +1,3 @@
+for (let i = 0; i < 5; i++) {
+    console.log(i)
+}
```
darzu: @riknoll , here's a test case. There's no "actual" and "expected" explicitly written here. Instead we run this .ts program with node.js and that stdout becomes the "expected", then we decompile it python and run the output with python3 and that result is the "actual" and if they disagree, we know there's probably a ts2py issue. See the readme.md for more info. 

We also compare the node.js trace with the sts trace, and we (will soon) compare traces before and after py2ts.

darzu: The point of these tests is to check that our runtime semantics in static typescript/python line up with tsc+node.js/python3

tests/runtime-trace-tests/tracerunner.ts
```diff
@@ -0,0 +1,309 @@
...
+    const isDisabled = (file: string): boolean =>
+        path.basename(file).indexOf("TODO") >= 0
+
+    let tsFiles = util.getFilesByExt(casesDir, ".ts")
+    let pyFiles = util.getFilesByExt(casesDir, ".py")
+    tsAndPyFiles = tsFiles.concat(pyFiles)
+        .filter(f => !isDisabled(f))
+
+    tsAndPyFiles.forEach(file => {
+        it("should preserve the runtime semantics of " + path.basename(file), async function () {
```
darzu: @riknoll here is where the actual mocha cases are generated. See my other comments as to what's going on

tests/runtime-trace-tests/tracerunner.ts
```diff
@@ -0,0 +1,309 @@
...
+        await testSts(tsFile)
+        let pyFile = await testTs2Py(tsFile)
+        // TODO(dazuniga): py2ts doesn't work in unit tests yet, once it does enable it here
+        // let tsfile2 = await testPy2Ts(pyFile)
+        // await testSts(tsfile2)
+    }
+
+    return;
+    // DSL for the tests
+    type Trace = string
```
riknoll: If you want the compiler to show typing errors for Trace and File you can do it like [this](https://www.typescriptlang.org/play/#src=interface%20Trace%20extends%20String%20%7B%0D%0A%20%20%20%20_traceBrand%3F%3A%20any%3B%0D%0A%7D%0D%0A%0D%0Ainterface%20File%20extends%20String%20%7B%0D%0A%20%20%20%20_fileBrand%3F%3A%20any%3B%0D%0A%7D%0D%0A%0D%0Afunction%20foo()%3A%20Trace%20%7B%0D%0A%20%20%20%20return%20%22%22%3B%0D%0A%7D%0D%0A%0D%0Alet%20f%3A%20File%20%3D%20foo()%3B)

tests/runtime-trace-tests/tracerunner.ts
```diff
@@ -0,0 +1,309 @@
...
+        let fnName = isPy ? "py2ts" : "ts2py"
+        let errFile = inFile + `.${fnName}_error`;
+        return convert(inFile)
+            .error(r => {
+                fs.writeFileSync(errFile, JSON.stringify(r))
+                return `${fnName} failed to convert '${inFile}'. Error saved at:\n${errFile}\nError is:\n${r}\n`
+            })
+            .then(async outFile => {
+                let outTrace = await runConverted(outFile)
+                if (!util.compareBaselines(outTrace, baseline)) {
```
riknoll: I think that `compareBaselines` has some decompiler specific stuff it does to the input like stripping some XML attributes. Might need to have two functions.

darzu: Good to know

tests/runtime-trace-tests/tracerunner.ts
darzu: Oh cool! And the "_traceBrand?: any;" is needed to avoid structural typing from seeing File and Trace as interchangeable I'm guessing ?

riknoll: Yep! This is a trick they use a lot in the TypeScript compiler

PR #5447, opened 2019-04-10
Ability to use arrays as shadow blocks

ChaseMor:
This adds the ability to use arrays as shadow blocks so that you can author blocks that take arrays of numbers, strings, or booleans, as parameters.

arrays_as_shadows

Fixes Microsoft/pxt-microbit#1654

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -104,14 +104,46 @@ namespace pxt.blocks {
         const value = document.createElement("value");
         value.setAttribute("name", p.definitionName);
 
-        const shadow = document.createElement(isVariable ? "block" : "shadow");
+
+        const isArray = /(.*)\[\]$/.exec(p.type);
+
+        const shadow = document.createElement(isVariable || isArray ? "block" : "shadow");
+
         value.appendChild(shadow);
 
-        const typeInfo = typeDefaults[p.type];
+        const typeInfo = typeDefaults[isArray && isArray[1] || p.type];
 
-        shadow.setAttribute("type", shadowId || typeInfo && typeInfo.block || p.type);
+        shadow.setAttribute("type", isArray ? 'lists_create_with' : shadowId || typeInfo && typeInfo.block || p.type);
         shadow.setAttribute("colour", (Blockly as any).Colours.textField);
 
+        if (isArray) {
+            const mut = document.createElement('mutation');
+            mut.setAttribute("items", "3");
+            shadow.appendChild(mut);
+            for (let i = 0; i < 3; i++) {
+                const v2 = document.createElement("value");
+                v2.setAttribute("name", "ADD" + i);
+                const shadow2 = document.createElement("shadow");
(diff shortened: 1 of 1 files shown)


pxtblocks/blocklyloader.ts
```diff
@@ -104,14 +104,45 @@ namespace pxt.blocks {
         const value = document.createElement("value");
         value.setAttribute("name", p.definitionName);
 
-        const shadow = document.createElement(isVariable ? "block" : "shadow");
+        const isArray = /(.*)\[\]$/.exec(p.type);
```
riknoll: Also needs to support the alternate syntax for arrays `Array<number>`. See `arrayTypeRegex` at the top of this file

pxtblocks/blocklyloader.ts
```diff
@@ -104,14 +104,45 @@ namespace pxt.blocks {
...
+
+        const shadow = document.createElement(isVariable || isArray ? "block" : "shadow");
+
         value.appendChild(shadow);
 
-        const typeInfo = typeDefaults[p.type];
+        const typeInfo = typeDefaults[isArray && isArray[1] || p.type];
 
-        shadow.setAttribute("type", shadowId || typeInfo && typeInfo.block || p.type);
+        shadow.setAttribute("type", isArray ? 'lists_create_with' : shadowId || typeInfo && typeInfo.block || p.type);
```
riknoll: shadowId should override everything else

pxtblocks/blocklyloader.ts
```diff
@@ -104,14 +104,45 @@ namespace pxt.blocks {
...
 
+        if (isArray) {
+            const mut = document.createElement('mutation');
+            mut.setAttribute("items", "3");
+            shadow.appendChild(mut);
+            for (let i = 0; i < 3; i++) {
+                const innerValue = document.createElement("value");
+                innerValue.setAttribute("name", "ADD" + i);
+                const innerShadow = document.createElement("shadow");
+                innerShadow.setAttribute("type", typeInfo.block);
```
riknoll: typeInfo might be undefined here (if not a number, string, or Boolean). In that case we should probably bail out earlier and just put an empty array for a shadow.

pxtblocks/blocklyloader.ts
```diff
@@ -245,7 +276,7 @@ namespace pxt.blocks {
         }
         if (fn.parameters) {
             comp.parameters.filter(pr => !pr.isOptional &&
-                (/^(string|number|boolean)$/.test(pr.type) || pr.shadowBlockId || pr.defaultValue))
+                (/^(string|number|boolean)(\[\])?$/.test(pr.type) || pr.shadowBlockId || pr.defaultValue))
```
riknoll: Use the arrayTypeRegex here

pxtblocks/blocklyloader.ts
pelikhan: looks like it is guarded in the if condition.

PR #5448, opened 2019-04-10
Handle function parameter's function parameter count mismatch; aka fix "onChat" (part 1/2)

darzu:
We now compile:
```
function onEvent(callback: (a: number) => void) {
    callback(1)
}
onEvent(function () {
    console.log('hi!')
})
```
into
```
def onEvent(callback: Callable[[number], None]):
	callback(1)
def function_0(a: number):
	print('hi!')
onEvent(function_0)
```
Notice the "a" parameter in function_0. This is required to match python semantics.

Note that we still need to handle this case:
```
function onEvent(callback: (a: number) => void) {
    callback(1)
}
function myCallback() {
    console.log('hi!')
}
onEvent(myCallback)
```
this will be in a follow up PR.

Also this PR sneaks in a fix for the python indentation to use '\t' instead of '  '.

Note this PR is waiting on: https://github.com/Microsoft/pxt/pull/5445
and will need to re-target to master once that one is in.

Here's a long discussion we had about this issue captured for posterity:
====================================
I believe the heart of the onChat issue is that this program is valid in typescript:
function fn1() {
 console.log('hi!')
}
function callFn(fn: (a: number, b: number) => void) {
 fn(1, 2)
}
callFn(fn1)
Yet this program isn't valid python:
def fn1():
    print('hi!')
def callFn(fn: Callable[[number, number], None]):
    fn(1, 2)
callFn(fn1)
Python gives this error:
TypeError: fn1() takes 0 positional arguments but 2 were given
which is pretty reasonable. In fact, TS gives a very similar error if u call:
fn1(1, 2) // error TS2554: Expected 0 arguments, but got 2.

It's actually a little bit strange to me that typescript decided to allow that code. It's probably for compatibility with existing javascript libraries. The question is what to do about this..

One solution would be to look for locations where an anonymous function is being passed to another function and the type of the parameter expects more args than the anonymous function is specify and then add extra arguments to the anonymous function. E.g. in TS this:
function onEvent(callback: (a: number, b: number) => void) {
 callback(1, 2)
}
onEvent(function() {
 console.log('hi!')
})
becomes:
def onEvent(callback: Callable[[number, number], None]):
        callback(1, 2)
def function_0(a: number, b: number):
        print('hi!')
onEvent(function_0)
which works for a lot of cases. But a drawback is that it doesn't always work. For example with:
function onEvent1(callback: (a: number, b: number) => void) {
 callback(1, 2)
}
function onEvent2(callback: () => void) {
 callback()
}
function myCallback() {
 console.log('hi!')
}
onEvent1(myCallback)
onEvent2(myCallback)
as far as I know, there's no single definition for myCallback that can be used for both onEvent1 and onEvent2 in python.
EDIT: actually, optional arguments in python might solve this. Investigating..
EDIT 2: optional parameters could solve this case, but I think it'd be harder to implement this correctly than the type conversion lambda I mention below

So another solution might be to add a layer of indirection using lambdas like:
def onEvent1(callback: Callable[[number, number], None]):
    callback(1, 2)
def onEvent2(callback: Callable[[], None]):
    callback()
def myCallback():
    print('hi!')
onEvent1(lambda a, b: myCallback())
onEvent2(lambda: myCallback())
These extra lambdas could be omitted when the arguments match up, so actually it could be:
def onEvent1(callback: Callable[[number, number], None]):
    callback(1, 2)
def onEvent2(callback: Callable[[], None]):
    callback()
def myCallback():
    print('hi!')
onEvent1(lambda a, b: myCallback())
onEvent2(myCallback)
this is essentially injecting an inline type conversion. I'm not sure it'll be easy to make this work for the general case like:
let foo: (a: number) => void;
foo = () => console.log('hi!')
foo(1)
you'd basically have to search out every time this implicit type conversion was happening. I'm not sure how easy that is with the TS language service.

Perhaps a reasonable solution would be a hybrid approach. Have python fill out all the arguments as optional if that's obvious, e.g. here:
onEvent(function() {
  console.log('hi')
}

==>

def myCallback(a: number = None)
    print('hi')
onEvent(myCallback)

this should handle 95% of cases. And then we generate lambdas for conversions in locations where we can't influence the fn definition so easily like here:

onEvent(myCallback)

==>

onEvent1(lambda a: myCallback())


pxtpy/pydecompiler.ts
```diff
@@ -32,7 +32,7 @@ namespace pxt.py {
 ///
 /// UTILS
 ///
-const INDENT = "  "
+const INDENT = "\t"
```
mmoskal: Maybe just four spaces?

darzu: I thought about that, but isn't \t more popular these days? vscode's python extension btw uses tabs by default:
image

Looking at https://stackoverflow.com/questions/119562/tabs-versus-spaces-in-python-programming
you can see there's debate and it does seem clear that the "most correct" way is 4 spaces, but the most popular way answer is tabs.
And personally I'm on the side of tabs for python, even though for most languages I prefer spaces.

PR #5453, opened 2019-04-10
Share to discourse

pelikhan:
Add optional "share to forum" button on share page. Probably need a better icon and text.

image
Used with https://github.com/Microsoft/pxt-arcade/pull/880

### DIFF
--- a/webapp/src/share.tsx
+++ b/webapp/src/share.tsx
@@ -447,7 +447,7 @@ export class ShareEditor extends data.Component<ShareEditorProps, ShareEditorSta
                         {showSocialIcons ? <div className="social-icons">
                             <SocialButton url={url} ariaLabel="Facebook" type='facebook' heading={lf("Share on Facebook")} />
                             <SocialButton url={url} ariaLabel="Twitter" type='twitter' heading={lf("Share on Twitter")} />
-                            {socialOptions.discourse ? <SocialButton url={url} icon={"keyboard"} ariaLabel={lf("Post to Forum")} label={lf("Forum")} type='discourse' heading={lf("Share on Forum")} /> : undefined}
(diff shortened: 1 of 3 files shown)


webapp/src/share.tsx
```diff
@@ -493,36 +497,46 @@ class SocialButton extends data.Component<SocialButtonProps, {}> {
...
+                    twitterText = lf("Check out what I made with @{0}!", socialOptions.orgTwitterHandle);
+                }
+                url = `https://twitter.com/intent/tweet?url=${encodeURIComponent(shareUrl)}` +
+                    `&text=${encodeURIComponent(twitterText)}` +
+                    (socialOptions.hashtags ? `&hashtags=${encodeURIComponent(socialOptions.hashtags)}` : '') +
+                    (socialOptions.related ? `&related=${encodeURIComponent(socialOptions.related)}` : '');
+                break;
+            }
+            case "discourse": {
+                // https://meta.discourse.org/t/compose-a-new-pre-filled-topic-via-url/28074
```
mmoskal: maybe better to use the script title as the "title=" and put the URL in "body=" ?

mmoskal: and even if we don't have the title, I think the URL should be in the body

mmoskal: I think `comments` would be a good icon

Jaqster:
How hard would it be to add a 'send by text' to a mobile phone number?  This is really how kids are communicating with each other - Text, Snapchat and Instagram.

webapp/src/share.tsx
pelikhan: Discourse automatically extracts the title from a URL pasted in the title.

abchatra:
Can the button be same size as facebook twitter?

pelikhan:
You can design it in the UI meeting. If we make it an icon, I don’t think people will discover it since the icon is generic.

PR #5457, opened 2019-04-11
Only starts HID watchers when they aren't activated (Win 10 App)

ChaseMor:
This adds a check to see if the HID watchers are activated before starting them.

Starting them while activated causes an error.

Fixes Microsoft/pxt-microbit#1975


pxtwinrt/hiddeploy.ts
```diff
@@ -196,6 +198,10 @@ namespace pxt.winrt {
             });
             watchers.push(watcher);
         });
-        watchers.forEach((w) => w.start());
+        watchers.forEach((w) => {
+                if (!w.status) {
```
pelikhan: Use filter

PR #5460, opened 2019-04-12
normalize user language + normalize en-US -> en

pelikhan:
Fix for https://github.com/Microsoft/pxt-arcade/issues/883

Normalize "en-US" to "en" which is the neutral language of makecode (en skips all localization work).


pelikhan:
Commented lines are there to debug strings that are loaded before localization data is present.

PR #5464, opened 2019-04-12
shared project: hide "forum", use smaller image

pelikhan:
image


abchatra:
Looks good. Is it possible to do the same for QR code? Like tiny image and when you click pops up the Full QR code? May be bad idea to have a dialog on a dialog.

pelikhan:
The QR code is already “tiny” sized.

abchatra:
That breaking line is hurting my eye

PR #5466, opened 2019-04-12
Fix microbit build error

jwunderl:
Microbit build was failing for me (https://github.com/Microsoft/pxt/commit/d8983d2404342dc7a5ab30e1856531d611bdbf6d#r33160766), looks like the issue is when 'file' is defined / explicitly is set to be undefined.

### DIFF
--- a/pxtcompiler/emitter/plaintsc.ts
+++ b/pxtcompiler/emitter/plaintsc.ts
@@ -14,7 +14,7 @@ namespace ts.pxtc {
 
     export function getDiagnosticString(diagnostic: KsDiagnostic | Diagnostic): string {
         let ksDiagnostic: KsDiagnostic;
-        if ("file" in diagnostic && diagnostic.file != undefined) {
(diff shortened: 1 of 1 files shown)


pxtcompiler/emitter/plaintsc.ts
```diff
@@ -14,7 +14,7 @@ namespace ts.pxtc {
 
     export function getDiagnosticString(diagnostic: KsDiagnostic | Diagnostic): string {
         let ksDiagnostic: KsDiagnostic;
-        if ("file" in diagnostic) {
+        if ("file" in diagnostic && diagnostic.file != undefined) {
```
pelikhan: why not just ``if(diagnostic.file) {`` ?

jwunderl: KsDiagnostic does not have a file property, so it fails

Screen Shot 2019-04-12 at 11 09 01 AM

PR #5467, opened 2019-04-12
move variants after boards

pelikhan:
Reorder download dialog to put physical boards first.
image

### DIFF
--- a/webapp/src/projects.tsx
+++ b/webapp/src/projects.tsx
@@ -944,17 +944,6 @@ export class ChooseHwDialog extends data.Component<ISettingsProps, ChooseHwDialo
             >
                 <div className="group">
                     <div className="ui cards centered" role="listbox">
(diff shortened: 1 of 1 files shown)


abchatra:
Can we move Pi0 to the last?

Jaqster:
Yes, much better.  Let's put the "Build it yourself Adafruit M4" at the end though.  Also, can we remove the Pi0 custom one?  That's only for the Arcade cabinet, right?

pelikhan:
Pi0 is the variant for Adafruit Joybonnet or any Raspberry Pi 0 arcade like the PiCade.

mmoskal:
`Pi0` is the same as `Adafruit Joy Bonnet`, the same way as `F401` is the same as `Brainpad`. I think we should keep all the ones that are "bootloader" screenshots for consistency.

PR #5470, opened 2019-04-15
Tutorials for JavaScript/Python

pelikhan:
Extends the tutorial engine to extract typescript/python code snippets and open the corresponding editor.

image

- [x] desired editor is inferred from the language of the snippets (only 1 kind per tutorial)
- [x] filters are computed using the same logic as blocks -- so the code is decompiled as well
- [x] cleaned up the tutorial parsing a bit.
- [x] added "renderPythonAsync" to editor interface

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1107,7 +1107,7 @@ export class ProjectView
         if (!header || !header.tutorial || !header.tutorial.tutorialMd) return Promise.resolve();
 
         const t = header.tutorial;
-        return tutorial.getUsedBlocksAsync(t.tutorial, t.tutorialMd)
+        return tutorial.getUsedBlocksAsync(t.tutorialCode)
             .then((usedBlocks) => {
                 let editorState: pxt.editor.EditorState = {
                     searchBar: false
@@ -1663,11 +1663,13 @@ export class ProjectView
             cfg.files = cfg.files.filter(f => f != "main.blocks")
             delete files["main.blocks"]
         }
+        if (options.preferredEditor)
+            cfg.preferredEditor = options.preferredEditor;
         files["pxt.json"] = JSON.stringify(cfg, null, 4) + "\n";
         return workspace.installAsync({
             name: cfg.name,
             meta: {},
-            editor: options.prj.id,
+            editor: options.preferredEditor || options.prj.id,
             pubId: "",
             pubCurrent: false,
             target: pxt.appTarget.id,
@@ -2719,11 +2721,13 @@ export class ProjectView
                 tutorialStep: 0,
                 tutorialReady: true,
                 tutorialStepInfo: tutorialInfo.steps,
-                tutorialMd: md
+                tutorialMd: md,
+                tutorialCode: tutorialInfo.code
             };
             return this.createProjectAsync({
                 name: title,
                 tutorial: tutorialOptions,
+                preferredEditor: tutorialInfo.editor,
                 dependencies
             }).then(() => autoChooseBoard ? this.autoChooseBoardAsync(features) : Promise.resolve());
         })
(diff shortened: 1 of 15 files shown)


webapp/src/app.tsx
```diff
@@ -2707,26 +2728,31 @@ export class ProjectView
...
+                tutorialCode: tutorialInfo.code
             };
             return this.createProjectAsync({
                 name: title,
                 tutorial: tutorialOptions,
+                preferredEditor: tutorialInfo.editor,
                 dependencies
-            }).then(() => autoChooseBoard ? this.autoChooseBoardAsync(features) : Promise.resolve());
-        })
-            .then(() => this.loadTutorialFiltersAsync())
```
pelikhan: done in createProject -> loadHeaderAsync

webapp/src/blocks.tsx
```diff
@@ -27,8 +27,6 @@ export class Editor extends toolboxeditor.ToolboxEditor {
     functionsDialog: CreateFunctionDialog = null;
 
     showCategories: boolean = true;
-    filters: pxt.editor.ProjectFilters;
```
pelikhan: using this.parent.state.filters instead

webapp/src/blocks.tsx
```diff
@@ -27,8 +27,6 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
-    showSearch: boolean;
```
pelikhan: use parent state instead

webapp/src/monaco.tsx
```diff
@@ -1501,12 +1501,18 @@ export class Editor extends toolboxeditor.ToolboxEditor {
             const w1 = (f1.attributes.weight || 50) + (f1.attributes.advanced ? 0 : 1000);
             return w2 > w1 ? 1 : -1;
         }).map(fn => {
-            let monacoBlockDisabled = false;
-            const fnState = filters ? (filters.fns && filters.fns[fn.name] != undefined ? filters.fns[fn.name] : (categoryState != undefined ? categoryState : filters.defaultState)) : undefined;
-            monacoBlockDisabled = fnState == pxt.editor.FilterState.Disabled;
+            let fnState = filters.defaultState;
+            if (filters && filters.fns && filters.fns[fn.name] !== undefined) {
+                fnState = filters.fns[fn.name];
+            } else if (filters && filters.blocks && fn.attributes.blockId && filters.blocks[fn.attributes.blockId] !== undefined) {
```
pelikhan: use blocks filter for other toolboxes if needed.

webapp/src/marked.tsx
```diff
@@ -42,7 +42,47 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
...
+            });
+
+        pxt.Util.toArray(content.querySelectorAll(`code.lang-spy`))
+            .forEach((langBlock: HTMLElement) => {
+                const code = langBlock.textContent;
+                const wrapperDiv = document.createElement('div');
+                pxsim.U.clear(langBlock);
+                langBlock.appendChild(wrapperDiv);
+                wrapperDiv.className = 'ui segment raised loading';
+                if (MarkedContent.blockSnippetCache[code]) {
```
jwunderl: At a glance looks this would be the reason I have to restart the local serve - will this grab the previously cached e.g. typescript output and show that instead of the intended python when changing snippet language?

Either way, unlikely to be something that ever matters outside of local serve, and trivially worked around by making a quick change to the content of the snippet that causes the code to need to be rerendered anyways

pelikhan: This cache should have been cleared when closing the turorial

jwunderl: Ah, you're right. I must have had a pending change that forced it to refresh when I tested it with this cache section removed.

PR #5474, opened 2019-04-16
Enable py2ts in unit tests (specifically the runtime-trace tests)

darzu:
Outcomes:
- "print()" compiles in Python editor now
- py2ts is invokable from unit tests

Changes:
- Changes py2ts to filter symbols from "shadowed" .ts files, not from the whole user package. E.g. if a package has file a.ts, b.ts, b.py then b.ts will have its symbols filtered out (so they don't show up as duplicates) during py2ts, but a.ts can still contribute symbols since it doesn't overlap with a .py file.
  - to accomplish this, I track the source file a symbol comes from in SymbolInfo
- Adds some typings for the sake of unit testing without a target (console.log & Buffer)
- Makes other small improvements to the converter test infrastructure

TODO:
- [x] not all tests pass yet
- [ ] mapping of built-ins is defined differently in py2ts and ts2py still; I want to unify these


localtypings/pxtarget.d.ts
```diff
@@ -631,6 +631,7 @@ declare namespace ts.pxtc {
         attributes: CommentAttrs;
         name: string;
         namespace: string;
+        fileName: string;
```
darzu: @mmoskal , I added tracking of the source file into SymbolInfo so I can do more fine-grain filtering of "shadowed" .ts files in py2ts. See below. I'm not sure this is a sane thing to do with these abstractions. Is this okay?

pxtpy/converter.ts
```diff
@@ -204,21 +204,24 @@ namespace pxt.py {
...
         internalApis = {}
         externalApis = {}
 
+        let tsShadowFilesSet = U.toDictionary(tsShadowFiles, t => t)
         for (let sym of U.values(apisInfo.byQName)) {
-            // sym.pkg == null - from main package - skip these
-            if (sym.pkg != null) {
-                let sym2 = sym as SymbolInfo
+            // if the symbol comes from a .ts file that matches one of the .py files we're compiling - skip these
+            if (sym.fileName in tsShadowFilesSet) {
```
darzu: @mmoskal , instead of filtering by package, I filter per file. E.g. filter out symbols from a.ts if a.py exists.

mmoskal: we usually don't use `in`, since `"toString" in {} == true`; use `hasOwnProperty`.

localtypings/pxtarget.d.ts
mmoskal: looks good!

libs/pxt-common/pxt-test-core.d.ts
```diff
@@ -0,0 +1,21 @@
...
+     */
+    //% weight=90
+    //% help=console/log blockGap=8 text.shadowOptions.toString=true
+    //% blockId=console_log block="console|log %text"
+    function log(text: string): void;
+}
+
+//% indexerGet=BufferMethods::getByte indexerSet=BufferMethods::setByte
+declare interface Buffer {
+}
```
mmoskal: maybe you can just add it to the `bare` target directly?

pxtpy/converter.ts
pelikhan: don't we typically do ``!!tsShadowFileSet[sym.fileName]`` ? (assuming non-empty file names)

mmoskal: I guess it's fine for file names, but `!!({})["toString"] == true`

pxtpy/converter.ts
```diff
@@ -444,6 +447,12 @@ namespace pxt.py {
         if (t0 === t1)
             return
 
+        if (t0.primType === "any") {
+            // TODO(dz): consult with Michal about this.
+            t0.union = t1
+            return
+        }
```
darzu: @mmoskal , do you have any thoughts on supporting "any"? Should we?

darzu: Got confirmation this should work.

libs/bare/basic.ts
```diff
@@ -1,16 +1,16 @@
 namespace control {
     //% shim=serial::writeString
-    export function __log(s: string) {
+    export function __log(s: any) {
```
pelikhan: the C++ shim expects a string. You should convert to string before passing to __log.

libs/bare/basic.ts
```diff
@@ -1,16 +1,16 @@
...
     }
 
-    export function dmesg(s: string) {
+    export function dmesg(s: any) {
```
pelikhan: same as __log, C++ expects a string.

pxtpy/converter.ts
```diff
@@ -444,6 +447,12 @@ namespace pxt.py {
         if (t0 === t1)
             return
 
+        if (t0.primType === "any") {
+            // TODO(dz): consult with Michal about this.
```
pelikhan: still relevant?

libs/bare/basic.ts
mmoskal: The thumb code emitter will emit conversion to string regardless of TS type. There is however dmesgVal or something like that which will print more info about non strings.

PR #5482, opened 2019-04-17
Add a file summary to 'translate' page

ganicke:
Add a core file summary and a couple of other notes.

RE: #5477


pelikhan:
A common question is "how long does it take for a translation to make it to the web site?"

pelikhan:
Also maybe link this page from the FAQ

PR #5485, opened 2019-04-18
Fix crashing when GitHub extension doesn't exist

ChaseMor:
Add checks to see if the GitHub extension exists before processing it as an extension.

Fixes Microsoft/pxt-arcade#777

### DIFF
--- a/pxtlib/github.ts
+++ b/pxtlib/github.ts
@@ -556,7 +556,7 @@ namespace pxt.github {
                     tag: rid.tag,
                     status
                 };
-            })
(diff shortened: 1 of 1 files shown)


pxtlib/github.ts
```diff
@@ -556,7 +556,7 @@ namespace pxt.github {
                     tag: rid.tag,
                     status
                 };
-            })
+            }).catch(err => undefined);
```
pelikhan: Add pxt.reportException

PR #5486, opened 2019-04-18
Fix simulator audio on iOS

jwunderl:
Fixes https://github.com/Microsoft/pxt-arcade/issues/626, where the audio wouldn't start until the user actively pressed the start / restart icons on iOS

Mute audio on iOS on page load, and require the user to unmute it themselves. This is directly triggered by the user, and enables the audio for the session.

Also fixes a bug that triggers semi - randomly cross browsers (that I hadn't seen before as it's weird and minor). It looks like the change to allow a quick restart caused the simulator to occasionally use old options due to ``SimulatorDriver._currentRuntime`` not getting updated, which would lead to the simulator being in an incorrect state after restarting the game (i.e. ``simtoolbar`` has it muted, but the sim was just unmuted, or vice-versa). I don't have a consistent way to trigger this, besides randomly pressing the mute, start, and restart buttons until it breaks. But it's fixed.

### DIFF
--- a/pxtsim/simlib.ts
+++ b/pxtsim/simlib.ts
@@ -199,7 +199,7 @@ namespace pxsim {
         // for playing WAV
         let audio: HTMLAudioElement;
 
-        function context(): AudioContext {
+        export function context(): AudioContext {
             if (!_context) _context = freshContext();
             return _context;
         }
@@ -219,6 +219,9 @@ namespace pxsim {
         export function mute(mute: boolean) {
             _mute = mute;
             stopAll();
+            const ctx = context();
+            if (!mute && ctx)
+                ctx.resume();
         }
 
         function stopTone() {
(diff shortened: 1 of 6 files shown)


pxtsim/simlib.ts
```diff
@@ -219,6 +219,9 @@ namespace pxsim {
         export function mute(mute: boolean) {
             _mute = mute;
             stopAll();
+            const ctx = context();
+            if (!mute && ctx && ctx.state === "suspended")
+                ctx.resume();
```
pelikhan: why is mute resuming?

jwunderl: The resume call has to be made explicitly from a user action, and the context state is initially suspended in iOS. I could bump it to context, but I'd still need an arbitrary call to context or something else in mute here

pelikhan: I got confused with what "context" was. LGTM!

PR #5490, opened 2019-04-19
display version info when running in electron

pelikhan:
I find it frustrating that i dont' know if the elecron app is updating or not. Add some info about online version availability in electron.
* not online -> please connect to internet to update
* online, updated -> you're up to date
* online, outdated, an update is avilable
image

### DIFF
--- a/webapp/src/dialogs.tsx
+++ b/webapp/src/dialogs.tsx
@@ -155,7 +155,8 @@ export function showAboutDialogAsync(projectView: pxt.editor.IProjectView) {
 
     pxt.targetConfigAsync()
         .then(config => {
-            const em = pxt.BrowserUtils.isPxtElectron() && config && config.electronManifest;
+            const isElectron = pxt.BrowserUtils.isPxtElectron();
+            const electronManifest = config && config.electronManifest;
             return core.confirmAsync({
                 header: lf("About"),
                 hideCancel: true,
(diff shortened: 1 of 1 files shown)


pelikhan:
@abchatra this is a simple hint like Chrome about the current version.

abchatra:
I don't understand the change. How do you know the update is available? Just based on the live version? We may not have released the latest one yet. It should say update available only if there a latest electron version available.

pelikhan:
My assumption (which may be wrong) is that ``targetConfigAsync()`` pulls the live version of ``targetconfig.json``; then we can compare the version in ``electronManifest.latest`` to the current version. Since the electron layer is already doing that work, it could be added to the ``onStatusUpdate`` event too.

abchatra:
I think you are right and it looks good. Ideally we can link to download page to update too (assuming automatic update is failing for some reason)

pelikhan:
Is there a default url scheme for the download page? I can add a link.

abchatra:
I guess we don't have latest URL probably we should add to the backend api

We have this in the download page. It is annoying to edit it every time. 

```
            window.open("https://makecode.com/api/release/microbit/v1.2.13/win64");
            tickEvent("offlineapp.download", { "target": "microbit", "platform": "win64" });
```

But you could make the url from the electron.manifest version
https://makecode.com/api/release/microbit/[[electronManifest.latest]]/win64

pelikhan:
Just add JavaScript that downloads the target config and generate the url.

Issue #5479, opened 2019-04-17
slice(0) broken.


pelikhan:
The array slice function takes "end" as an optional parameter, we MakeCode passes as 0. This breaks the common use of ``.slice(0)`` to clone the array. Is this a recent regression?

```
let a = [0, 1, 2];
let b = a.slice(0);

console.log(`${JSON.stringify(a)} -> ${JSON.stringify(b)}`);
```
Result:
```
[0,1,2] -> []
```

jwunderl:
Not a recent regression: an optional parameter-ed slice was first available in https://github.com/Microsoft/pxt/releases/tag/v5.4.2 , which is first used in https://arcade.makecode.com/v0.3.33 

alternate test code since that one didn't work right away in the old version:

```typescript
let a = [0, 1, 2];
let b = a.slice(0);

console.log(`${toString(a)} -> ${toString(b)}`);


function toString(a: number[]) {
    return a
        .map(e => e + "")
        .reduce((p, c) => p + ", " + c, "[") + "]";
}
```

jwunderl:
@mmoskal and I figured out what's going on: the ``eg: 0`` in comments are acting as default values instead of just documentation and defaults for the blocks:

```typescript

/**
 * Return a section of an array.
 * @param start The beginning of the specified portion of the array.
 * @param end The end of the specified portion of the array. eg: 0
 */
function slice(start?: number, end?: number) {
    console.log(start + "")
    console.log(end + "")
}

slice();
```

which outputs undefined for start, but 0 for end

He said he's gonna fix it

jwunderl:
fixed with #5492

PR #5492, opened 2019-04-19
Don't treate eg.:... as default parameters

mmoskal:
over to you @jwunderl


jwunderl:
Can confirm it works in arcade and fixes https://github.com/Microsoft/pxt/issues/5479 

``eg`` is used to set default values to populate blocks. Issue was that ``eg:`` was being treated exactly the treated the same as ``variable.defl``, which provides default values for c++, and is basically intended to be the same as setting a default value for a param in typescript. This overrode the defaults parameters for typescript functions with defaults, and passed whatever value is provided instead of undefined for optional parameters.

PR #5493, opened 2019-04-19
Don't decompile globals

ChaseMor:
Checks to see if a variable is declared in another file before recompiling it as a block.

If an identifier is defined in another file, it is flagged as a global and will fail decompilation.

Fixes Microsoft/pxt-arcade#554

### DIFF
--- a/pxtcompiler/emitter/annotate.ts
+++ b/pxtcompiler/emitter/annotate.ts
@@ -42,14 +42,13 @@ namespace ts.pxtc {
                     case SyntaxKind.BinaryExpression:
                         annotateBinaryExpression(child as ts.BinaryExpression);
                         break;
-                    case SK.Identifier:
-                        const decl = getDecl(child);
-                        if (decl.getSourceFile().fileName !== "main.ts") {
+                    case SyntaxKind.Identifier:
+                        const decl: Declaration = getDecl(child);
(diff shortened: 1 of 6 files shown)


ChaseMor:
This works, but it feels a bit hacky so I'm open to suggestions.

riknoll:
I think we should probably move this to `annotate.ts`. The basic approach would be something like this:

1. Add an `IdentifierInfo` interface (see below)
2. In `annotate.ts`, add an `SK.Identifier` case to the switch statement
3. Call `getDecl()` on the identifier node and if it returns something, check if `decl.getSourceFile().name !== "main.ts"`
4. Inject the IdentifierInfo into the Node `(node as any).identifierInfo = // whatever`
5. In the decompiler, consume the injected info in the `checkExpression()` call.


```typescript
interface IdentifierInfo {
    isGlobal: boolean;
}
```

ChaseMor:
@riknoll I've switched the logic to ``annotate.ts``, but I'm noticing that it's flagging all identifiers that are defined elsewhere as global. Not just variables, but namespaces and functions as well. I don't see any side effects though. Is it a problem to flag these or should we limit flagging to only identifiers defined with the ``let`` or ``const`` keywords.

riknoll:
Yeah, let's filter it. You can check the `kind` of the node returned by `getDecl()` and see if it is a variable declaration

PR #5501, opened 2019-04-23
Import extension as file

pelikhan:
To better support offline scenario, this change allows to import project **files** (.hex, .uf2, .png) as extension in a project.

The full source of the project is stored in a file in the project (say ``donut.json`` for the ``donut`` extension file) and the dependecies has a new protocol ``pkg:donut.json``.

This solves the problem of supporting 3rd party extensions in the web and/or offline app. Also shared scripts will correctly load or reload etc...

In the extension dialog, we have a new card:
image


abchatra:
Looks good. Can we also add support for local file system? 

Easiest way is to clone the GitHub repo of extension and load it as a local file system when one is offline.
Instead of https://github.com/microsoft/pxt-neopixel
you load c:/packages/pxt-neopixel.

Offline installer has access to local file system.

pelikhan:
This would be a new feature.

PR #5504, opened 2019-04-24
Monaco debugger part 4 (libs debugging)

riknoll:
Adding an experiment for debugging extensions and libs! Only for TypeScript of course.

@jamesadevine you might find this useful...


jamesadevine:
This sounds totally freakin awesome :smile:

PR #5510, opened 2019-04-24
Add some docs about technical side of Python support

mmoskal:
This goes under https://makecode.com/js/python (which isn't the most fortunate URL, but will do for now)

### DIFF
--- a/docs/js/python.md
+++ b/docs/js/python.md
@@ -25,6 +25,11 @@ This approach has the following advantages:
   smart autocompletion, contextual doc-comment display, etc.
 * features like debugger are easily shared between MPY and STS
 
+The type annotations are technically optional in both STS and MPY -
+TypeScript `any` type is supported in the runtime with dynamic member lookup
+(though it still uses compact C++-like memory layout for classes).
+Some of this have not been fully implemented or exposed to the user yet though.
+
 The main disadvantage are some limitations in compatibility with the full Python language.
 The MPY parser handles the whole of Python, and thus missing features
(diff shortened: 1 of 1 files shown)


docs/js/python.md
```diff
@@ -1,14 +1,56 @@
-# Python Converter
+# Python Support
 
-**This is work in progress**
+**Python support in MakeCode is experimental**
+
+We currently refer to TypeScript version supported by MakeCode as MakeCode TypeScript (MTS)
```
abchatra: MakeCode Typescript(MTS)? It is referred as Static Typescript all over the place

docs/js/python.md
```diff
@@ -1,14 +1,56 @@
...
+* as the MPY->STS converter infers types, the MPY editor supports
+  smart autocompletion, contextual doc-comment display, etc.
+* features like debugger are easily shared between MPY and STS
+
+The type annotations are technically optional in both STS and MPY -
+TypeScript `any` type is supported in the runtime with dynamic member lookup
+(though it still uses compact C++-like memory layout for classes).
+Some of this have not been fully implemented or exposed to the user yet though.
+
+The main disadvantage are some limitations in compatibility with the full Python language.
```
abchatra: I would skip this entire paragraph. Not very useful details here.

docs/js/python.md
mmoskal: well, if it's Static TypeScript, it's also Static Python. Changed.

PR #5514, opened 2019-04-25
Only count blocks, not other pieces of the XML, towards MAX_BLOCKS

jwunderl:
fixes https://github.com/Microsoft/pxt-arcade/issues/915 (there seems to be a blockly related issue in the last one linked there still, but that seems unrelated)

Ties the block count to the number of ``<blocks>`` created, so that it maps directly to the number of blocks listed in the ``Delete all {num} blocks`` dialogue:

Screen Shot 2019-04-24 at 9 21 56 PM

For example, ``Delete all blocks`` on Falling Duck says it has 77 blocks, but currently ends up with ``emittedBlocks`` at 150; from the users perspective there is no real way to tell what they're at until it gives the error.

(Also arbitrarily bumped ``MAX_BLOCKS`` to just more than the largest arcade game I've seen that was built in blocks, but can revert that if preferred)

Making this a draft PR as I'll have to add a test still, but the one in https://github.com/Microsoft/pxt/pull/2925/files is a bit... long... so I'll add it tomorrow

### DIFF
--- a/pxtcompiler/emitter/decompiler.ts
+++ b/pxtcompiler/emitter/decompiler.ts
@@ -20,7 +20,7 @@ namespace ts.pxtc.decompiler {
     /**
      * Max number of blocks before we bail out of decompilation
      */
-    const MAX_BLOCKS = 1000;
+    const MAX_BLOCKS = 1500;
 
     const lowerCaseAlphabetStartCode = 97;
     const lowerCaseAlphabetEndCode = 122;
@@ -466,7 +466,6 @@ ${output}</xml>`;
         }
 
         function mkStmt(type: string): StatementNode {
-            countBlock();
             return {
                 kind: "statement",
                 type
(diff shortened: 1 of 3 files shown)


pelikhan:
How about just removing the count all together?

jwunderl:
Last I heard we wanted to keep the count to avoid breaking on examples where the blocks would be far too large to interact with properly. That's more a decision for @riknoll I think

As a side note, it looks like all the games on the forum that were created in blocks can now decompile properly with this and the other change (a bit weird, as I can't tell what change occurred to fix the variables bug that I was looking at / consistently reproing for ~2 hours yesterday but can't get to repro now at all)

pelikhan:
I see, i taught this was used for the message display only...

riknoll:
We added the count after someone tried to decompile ~1mb of TypeScript and it locked them out of their project. They couldn't recover their code because Blockly would freeze the browser every time they tried to open it.

PR #5515, opened 2019-04-25
Print Dropdown for Docs (Outlines/Color)

toledoal:
Adding the print dropdown for "Outlines" or "Color" options. 

image

image

image


pelikhan:
Could we stay with something more iconic? We are going to have to translate all these strings. 

"Outline" does not ring a bell. Simplify to Color vs BW print.

toledoal:
Sure. Let me ask for an icon option. We didn't considered the translate aspect. I'll post a screenshot here before another commit. :)

toledoal:
Here is another option:

image

pelikhan:
That works.

docfiles/docs.js
```diff
@@ -259,6 +261,35 @@ function renderSnippets() {
...
+        let rules = [].slice.call(stylesSheet.cssRules).filter(function(x,i) { 
+            if (String(x.media).indexOf("only print") !== -1) {
+                return x;
+        }});
+
+        window.printStyles = rules;
+
+        window.printStyles.forEach(function(x) {
+            let index = [].slice.call(stylesSheet.cssRules).indexOf(x);
+            stylesSheet.deleteRule(index);
```
pelikhan: This is really hacky. I think we should add a “print-color” class on the body and have CSS rules get cancelled because of that.

toledoal: Since the default print mode is black & white, we loose the origin colors. We could try removing the current print.less classes and then have a new "black & white" mode on body as you say. I didn't want to move anything in print.less because I don't know how this will behave in lobby and editor.

pelikhan:
I think we can play css tricks within the @media print rules using :not(printcolor)

toledoal:
Good catch. Let me try that.

toledoal:
So, the :not(.print-color) option works as well. But we have several problems. 
1. Since the :not() selector needs a simple selector, each rule in the print.less file should be marked with :not(.print-color) -up to 40- and then we need to iterate the DOM to add/remove those classes to each selector on switching print type. Which adds a bunch of lines of code. :(

```
switch (value) {
                case 'color': 

                    // Add a bunch of print selector
                    [].forEach.call(document.querySelectorAll('.blocklyBlockBackground'), function(el){
                        el.classList.add("print-color");
                    });
                    
                    [].forEach.call(document.querySelectorAll('.blocklyPath'), function(el){
                        el.classList.add("print-color");
                    });  
                    
                       
                    break;
                case 'outlines':
                    [].forEach.call(document.querySelectorAll('.blocklyBlockBackground'), function(el){
                        el.classList.remove("print-color");
                    });
                    
                    [].forEach.call(document.querySelectorAll('.blocklyPath'), function(el){
                        el.classList.remove("print-color");
                    });
                    break;
}
```

2. If we add/modify print rules in print.less we need to go and modify docs.js

I'm going to explore other options if I found a better one :) 
by the way, I'm using [].forEach.call because our dear friend IE11

pelikhan:
Not all the rules need to be disabled. Only the coloring rules which you can group in .LESS probably.

docfiles/docs.js
toledoal: I changed the approach to use add/remove css clases from classList

docfiles/docs.js
```diff
@@ -189,14 +189,58 @@ function setupSemantic() {
...
+
+            switch (value) {
+                case 'color': 
+                    collections.forEach(function(elements) {
+                        if (elements.length != 0){
+                            addRemovePrintColorClass(elements, 'add');
+                        }
+                    });                       
+                    break;
+                case 'outlines':
```
pelikhan: just call it "bw"

docfiles/docs.js
```diff
@@ -189,14 +189,58 @@ function setupSemantic() {
...
+      });
 
     if (/browsers$/i.test(window.location.href))
         $('.ui.footer').append($('<div class="ui center aligned small container"/>').text('user agent: ' + navigator.userAgent))
 }
 
+function addRemovePrintColorClass(elements,operation) {
+    [].forEach.call(elements, function(el){
+        if (operation === 'add'){
+            el.classList.add("print-color");
```
pelikhan: is classList available in IE?

docfiles/docs.js
```diff
@@ -189,14 +189,58 @@ function setupSemantic() {
...
+            var blocklyLedOn = document.querySelectorAll('.blocklyLedOn');
+            var uiimage = document.querySelectorAll('.ui.image');
+            
+            var collections = [ blocklyBlockBackground, blocklyPath, blocklyLedOff, blocklyDropdownText, blocklyText, blocklyLedOn, uiimage]
+
+            switch (value) {
+                case 'color': 
+                    collections.forEach(function(elements) {
+                        if (elements.length != 0){
+                            addRemovePrintColorClass(elements, 'add');
```
pelikhan: use boolean parameter instead of string "add", "remove"

theme/print.less
```diff
@@ -97,20 +101,20 @@
     }
 
     /** blockly **/
-    .blocklyPath {
+    .blocklyPath:not(.print-color) {
```
pelikhan: can we do a document level ``:not(.print-color) .blockPath`` so we don't have to add/remove .print-color to individual items?

toledoal: I tried in different level, but it doesn't work :( The only way I could make the :not() selector work was in the element level itself

docfiles/docs.js
toledoal: It is partially available, but it supports "add" and "remove". https://caniuse.com/#feat=classlist 
I tested it locally :)

theme/print.less
toledoal: Let me try to add new rules to print.less

toledoal: I was testing and it might work :)

toledoal: This one works better. Please check it out

pxtlib/docsrender.ts
```diff
@@ -282,10 +282,18 @@ namespace pxt.docs {
 
         // Add print button
         const printBtnHtml = `
-            <button id="printbtn" class="circular ui icon right floated button hideprint" title="${lf("Print this page")}">
+        <div id="printbtn" class="circular ui icon top right floated pointing dropdown button hideprint" title="${lf("Print this page")}">
```
pelikhan: maybe we can move this HTML into its own .html file under/docfiles; then this change would not require a cloud update.

toledoal: Yes that could work. I'll add a new commit with this

toledoal: Quick question about this @pelikhan, the lf("Color") and lf("Black and White"). If I put this html under docfiles. How should I render this localizable label? I think I could export a function from runner.ts ? Or what would be a better approach? I don't know exactly when the lf() function can be called. Thanks!

pelikhan:
All strings in .html pages are automatically scanned for localizable strings.

pelikhan:
So no need for lf if the string is in an .html page

toledoal:
Yes, the fix for that bug is in the docs styling PR. It basically changes styles.css 
@media only print { #root {
        position: relative;
    }}
But I can add the change here as well, and then merge it.

pelikhan:
Looks good to me. @riknoll @abchatra

PR #5517, opened 2019-04-26
Use Web USB after pairing

riknoll:
The logic was broken when we moved the "Always use web USB" stuff into an experiment. With this change, here's the pairing experience:

1, If you never pair the device, we always download
2. If a paired device is connected, we try to use webUSB
3. If the paired device is disconnected, the user gets an error dialog and we go back to downloading until it gets reconnected

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1838,6 +1838,7 @@ export class ProjectView
             if (res) {
                 return pxt.usb.pairAsync()
                     .then(() => {
+                        cmds.setWebUSBEnabled(true);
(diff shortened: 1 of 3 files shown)


darzu:
There's one critical caveat to the above logic:
1. if the target has overwritten the deploy logic (e.g. microbit does this), then pxt-core shouldn't unset their override. 

This is what was happening when we called `init();` again inside setWebUSBPaired, which broke webusb for micro:bit. It's pretty subtle logic so I'm going to try to make it more explicit in my PR.

FYI @riknoll ^^

PR #5518, opened 2019-04-26
fix parameter hints in python intellisense

darzu:
Before:
image

After:
image


darzu:
Hmm, looks like the "active parameter" isn't updated as you type:
image

TODO: [ ]

pelikhan:
This is a good one.

PR #5528, opened 2019-04-29
Move simulator collapse button out of editor toolbar, CSS updates

shakao:
- Move simulator collapse button (update desktop, tablet, mobile styles)
- Small updates to sidedocs toggle (no "docs" symbol on hover, smaller chevron)
- Update tutorial button styles to match (remove text, move out of tutorial card)

Demo: https://makecode.microbit.org/app/357d62b063de9f7e133ae176f93d541572922dd9-1ffda62afe

### DIFF
--- a/theme/common.less
+++ b/theme/common.less
@@ -414,6 +414,41 @@ div.simframe > iframe {
     padding: 0.3em;
 }
 
+/* Collapse/expand button for simulator, sidedocs */
+#togglesim, #sidedocstoggle {
+    position: absolute;
+    border-top-left-radius: 100px;
+    border-bottom-left-radius: 100px;
+    top: 40%;
+    height: 20%;
+    z-index: @sidedocZIndex;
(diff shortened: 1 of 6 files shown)


pelikhan:
Really not a fan of the chevron for the more/less of the tutorial. I find it much more obvious to have a "More..." test then the down chevron.

shakao:
image

Makes the button a little bit taller so it overlaps the workspace more but it does separate it from the Back/Forward which is nice

pelikhan:
Overlapping is going to be an issue with text-based editors (it will consistently overlap the first line of code). We should avoid IMO.

shakao:
Removed tutorial more/less changes from this PR for later design iteration. New demo: https://makecode.microbit.org/app/fa2739b1b3a588fe63d2b46bd94537a7c2ede8c0-552d1d36ca

PR #5530, opened 2019-04-30
support for "idling"

pelikhan:
- [x] remove generic EventBus and unify support for bus in runtime/bareboard.
- [x] better integration with CODAL "onIdleCallback" infrastructure to avoid spinning fibers for background tasks. in C++, it listens to the IDLE event generated by CODAL. in javascript, starts an interval timer. integrates with eventContext.

REquires https://github.com/Microsoft/pxt-common-packages/pull/806

### DIFF
--- a/pxtsim/runtime.ts
+++ b/pxtsim/runtime.ts
@@ -185,8 +185,15 @@ namespace pxsim {
 
     const SERIAL_BUFFER_LENGTH = 16;
     export class BaseBoard {
-        public runOptions: SimulatorRunMessage;
-        public messageListeners: MessageListener[] = [];
+        id: string;
+        bus: pxsim.EventBus;
+        runOptions: SimulatorRunMessage;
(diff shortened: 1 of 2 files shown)


pxtsim/runtime.ts
```diff
@@ -1097,6 +1102,26 @@ namespace pxsim {
...
 
+        startIdle() {
+            // schedules handlers to run every 20ms
+            if (this.idleTimer === undefined) {
+                this.idleTimer = setInterval(() => {
+                    if (!this.running || this.pausedOnBreakpoint) return;
+                    const bus = this.board.bus;
+                    if (bus)
+                        bus.queueIdle();
+                });
```
mmoskal: why not specify timeout value? the minimum is 10ms.

pelikhan: oopsy

PR #5533, opened 2019-05-01
Fix decompilation of non-boolean conditions

jwunderl:
fixes https://github.com/Microsoft/pxt-arcade/issues/911

(https://github.com/Microsoft/pxt-arcade/issues/946 is related-ish and still causes failures for some expressions, but not fixed with this change)

Emits gray blocks instead of failing to decompile for non-boolean conditions

decompilation test output:

arcade-screenshot (9)

### DIFF
--- a/pxtcompiler/emitter/decompiler.ts
+++ b/pxtcompiler/emitter/decompiler.ts
@@ -1359,7 +1359,7 @@ ${output}</xml>`;
         }
 
         function checkConditionalExpression(n: ts.WhileStatement | ts.IfStatement) {
-            const expr: Expression = n.expression;
+            const expr = unwrapNode(n.expression);
             switch (expr.kind) {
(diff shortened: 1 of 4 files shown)


pxtcompiler/emitter/decompiler.ts
```diff
@@ -1348,6 +1348,66 @@ ${output}</xml>`;
...
+            if (err) {
+                const expr = getTypeScriptExpressionBlock(stmt.expression);
+                return mkValue(name, expr, booleanType)
+            } else {
+                return getValue(name, stmt.expression, booleanType);
+            }
+        }
+
+        function checkConditionalExpression(n: ts.WhileStatement | ts.IfStatement) {
+            const expr: Expression = n.expression;
```
riknoll: Call `unwrapNode()` on the expression to remove any parentheses

PR #5537, opened 2019-05-01
Add 'Open in Forum' display for forum links

jwunderl:
Make forum links say 'Open in Forum' instead of 'Show Instructions'

Screen Shot 2019-05-01 at 1 45 48 PM

### DIFF
--- a/webapp/src/projects.tsx
+++ b/webapp/src/projects.tsx
@@ -579,7 +579,7 @@ export class ProjectsDetail extends data.Component<ProjectsDetailProps, Projects
             className: 'huge positive'
         }]
 
-        const isLink = !cardType && (youTubeId || url) || cardType === "forumLink";
+        const isLink = (!isCodeCardType(cardType) || cardType === "forumLink") && (youTubeId || url);
         const linkHref = (youTubeId && !url) ? `https://youtu.be/${youTubeId}` :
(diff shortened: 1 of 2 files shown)


webapp/src/projects.tsx
```diff
@@ -568,6 +568,7 @@ export class ProjectsDetail extends data.Component<ProjectsDetailProps, Projects
             clickLabel = lf("Start Tutorial");
         }
         else if (cardType == "codeExample" || cardType == "example") clickLabel = lf("Open Example");
+        else if (cardType == "forumLink") clickLabel = lf("Open in Forum");
```
abchatra: See in forum. @ganicke what is the right word here?

jwunderl:
Last commit adds the ability to add code cards in the future without breaking existing docs (eg: in https://github.com/Microsoft/pxt-arcade/pull/948 I can't add the forumLink to the docs until the live site is bumped to include this change, because having a ``cardType`` of any type causes the "Show Instructions" button to not link to anything and be replaced with an undefined onClick behavior)

pelikhan:
Note that discourse has a well defined API that allows to sniff the image, author and title of the forum post -- so we could sniff those on the client side too.

webapp/src/projects.tsx
riknoll: I would say either "Open in forum" or "Go to forum" would work

PR #5543, opened 2019-05-02
Dutch Translation of comment is commentaar

### LINKED ISSUE #5542: Dutch translation of 'comment'
The Dutch translation used for the word 'comment' is strange. It's translated as 'reactie', which is a comment in the sense of a reaction (as in a reply to something).

While this might technically be a correct translation for 'comment', it would be better to use the 'commentaar' translation, which is 'comment' as in a note or remark.

It really bother me, because I'm writing a Dutch course and it's very strange.

I'll open a PR.

editicalu:
The Dutch translation used for the word 'comment' is strange. It's translated as 'reactie', which is a comment in the sense of a reaction (as in a reply to something).

While this might technically be a correct translation for 'comment', it would be better to use the 'commentaar' translation, which is 'comment' as in a note or remark.

It really bother me, because I'm writing a Dutch course and it's very strange.

Closes #5542

### DIFF
--- a/webapp/public/locales/nl/strings.json
+++ b/webapp/public/locales/nl/strings.json
@@ -151,7 +151,7 @@
   "Quit": "Sluit af",
   "Redo": "Opnieuw uitvoeren",
   "Reference, lessons, ...": "Referentie, lessen...",
-  "Remove Comment": "Reactie Verwijderen",
+  "Remove Comment": "Commentaar Verwijderen",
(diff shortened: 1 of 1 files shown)


pelikhan:
Could you fix this in crowdin instead? We have instructions on how to create translations at https://makecode.com/translate. Otherwise, your fix will be overriden the next round of sync crowdin->github.

editicalu:
I see. Thanks

PR #5549, opened 2019-05-07
Image editor shortcuts, eyedropper and shift line drawing

jwunderl:
closes https://github.com/Microsoft/pxt-arcade/issues/951

Matching the shortcuts listed in the issue:

|  Tool | Piskel |
|---|---|
| Paint Bucket | B|
| Pencil | P|
| Draw Rectangle | R|
| Draw Line | L|
| Eraser | E|
| Draw Circle | C|

Colors accessible with 0 - 9, and shift + 1-6 for the rest

holding alt (option on mac) gives eyedropper tool to select colors while drawing

fix keyboard shortcuts for undo / redo (on mac, need to test on a windows machine too)

shift click with line tool to easily draw shapes

eyedropper and a few more shortcuts

tooltips with shortcut hints

(I think this needs a bit of cleanup but checking for initial comments)


pxtlib/sprite-editor/spriteEditor.ts
```diff
@@ -313,62 +314,107 @@ namespace pxtsprite {
...
             }
         }
 
+        setIconsToDefault() {
+            this.switchIconTo(PaintTool.Rectangle);
+            this.switchIconTo(PaintTool.Normal);
+        }
+
         private keyDown = (event: KeyboardEvent) => {
+            if (event.key === "Undo" || (event.ctrlKey && event.key === "z")) {
```
jwunderl: This was already in an event handler, but doesn't actually run -  the blockly undo is run and closes this before the event would be fired. Anyone happen to know of a way to prevent that behavior (preventDefault doesn't.)

jwunderl: Nevermind, fixed by moving the handling of the event to the capture phase and stopping propagation of the event.

pxtlib/sprite-editor/spriteEditor.ts
```diff
@@ -450,15 +516,43 @@ namespace pxtsprite {
...
+                return;
+            switch (this.activeTool) {
+                case PaintTool.Line:
+                    this.setCell(this.paintSurface.mouseCol, this.paintSurface.mouseRow, this.color, false);
+                    break;
+            }
+        }
+
+        private clearShiftAction() {
             switch (this.activeTool) {
```
riknoll: Just make this an if statement

pxtlib/sprite-editor/spriteEditor.ts
```diff
@@ -313,73 +314,125 @@ namespace pxtsprite {
...
             }
 
+            if (event.keyCode == 18 ) { // Alt
+                this.discardEdit();
+                this.paintSurface.setEyedropperMouse(true);
+                this.altDown = true;
+            }
+
+            [
+                PaintTool.Fill,
```
riknoll: Having an array literal as the start of a statement feels weird to me; could we put this in a function or constant?

jwunderl:
And here's the new eyedropper cursor @riknoll sent me:



(and partly me testing out the hot spot for the cursor to make sure it grabs the right spot)

pxtlib/sprite-editor/canvasGrid.ts
```diff
@@ -35,6 +38,23 @@ namespace pxtsprite {
             this.hideOverlay();
         }
 
+        setEyedropperMouse(on: boolean) {
+            const eyedropperClass = "sprite-editor-eyedropper";
+            if (on) {
+                this.paintLayer.classList.add(eyedropperClass);
+                if (!this.lightMode) {
+                    this.backgroundLayer.classList.add(eyedropperClass);
```
riknoll: Can't use classList (not supported by IE). I think we have a helper for this in one of the utils

pxtlib/sprite-editor/spriteEditor.ts
```diff
@@ -331,73 +331,127 @@ namespace pxtsprite {
...
-                btn.setText("\uf096");
-                btn.title(lf("Rectangle"));
-            } else if (tool == PaintTool.Normal) {
-                //Change icon back to pencil
-                btn.setText("\uf040");
-                btn.title(lf("Pencil"));
+        switchIconTo(tool: PaintTool) {
+            if (this.activeTool === tool) return;
+
+            const btn = (this.sidebar.getButtonForTool(tool) as TextButton);
```
riknoll: Nitpick: remove parens

pxtlib/sprite-editor/canvasGrid.ts
jwunderl: It looks like ie10+ supports the simple add/remove cases https://caniuse.com/#feat=classlist (it is also used over in https://github.com/microsoft/pxt/pull/5515#discussion_r279769731 and a handful of other parts of the codebase, so if it is broken we'd probably want to fix those too?)

I see a decent number of different helpers for handling it like you mentioned though, each with slightly different implementations:

* pxtblocks/fields/field_utils.ts
* pxtlib/svgUtil.ts
* pxt/pxtsim/instructions.ts
* pxt/pxtsim/runtime.ts
* pxtsim/svg.ts
* webapp/src/sui.tsx

pelikhan:
Let's do a cleanup of classlist in a sep PR.

pelikhan:
Can you use one of the helpers in svgUtil?

jwunderl:
re: cleaning it up separately, is there a single preferred util file anywhere to put general helpers like that?

pelikhan:
Not really. Some live in sim and lib.

riknoll:
Put it in BrowserUtils maybe?

PR #5550, opened 2019-05-07
Stop sound when user restarts or stops sim

jwunderl:
Fixes https://github.com/microsoft/pxt-arcade/issues/256

Sends a 'stop sound' message whenever the user ends an instance of the sim running, so that the sound doesn't continue for a few moments into the next game / with no game running. (I think for it to work fully with tones it needs the recent common packages changes to mixer, but I'm not certain / didn't see a reason to roll back the common packages bump to check)

For an example of what this fixes, try stopping or restarting [this game](https://makecode.com/_McfaJVTafC8m) after the sound starts - it will continue momentarily after you press the button. A bit worse when playing with a version with a working ringTone as well

### DIFF
--- a/pxtsim/simdriver.ts
+++ b/pxtsim/simdriver.ts
@@ -322,6 +322,10 @@ namespace pxsim {
             this.postMessage({ type: 'mute', mute: mute } as pxsim.SimulatorMuteMessage);
         }
 
+        public stopSound() {
+            this.postMessage({ type: 'stopsound' } as pxsim.SimulatorStopSoundMessage)
+        }
+
         public isLoanedSimulator(el: HTMLElement) {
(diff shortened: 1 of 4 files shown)


pxtsim/simdriver.ts
```diff
@@ -529,7 +533,7 @@ namespace pxsim {
             if (!this.listener) {
                 this.listener = (ev: MessageEvent) => {
                     if (this.hwdbg) return
-                    this.handleMessage(ev.data, ev.source)
+                    this.handleMessage(ev.data, ev.source as Window)
```
jwunderl: This change is unrelated, just casting explicitly as it's missing unnecessary properties of window and showing up as an error in the sidebar without the cast

PR #5552, opened 2019-05-08
Persist image editor undo/redo in blocks editor

jwunderl:
closes https://github.com/microsoft/pxt-arcade/issues/955

ezgif com-optimize (5)

Not added to monaco editor with this, as it doesn't seem like there's necessarily a place to persist that stacks there

### DIFF
--- a/pxtlib/sprite-editor/spriteEditor.ts
+++ b/pxtlib/sprite-editor/spriteEditor.ts
@@ -62,8 +62,8 @@ namespace pxtsprite {
         private cursorCol = 0;
         private cursorRow = 0;
 
-        undoStack: Bitmap[] = [];
-        redoStack: Bitmap[] = [];
+        private undoStack: Bitmap[] = [];
+        private redoStack: Bitmap[] = [];
 
         private columns: number = 16;
         private rows: number = 16;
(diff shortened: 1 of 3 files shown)


pxtlib/sprite-editor/spriteEditor.ts
```diff
@@ -223,6 +223,24 @@ namespace pxtsprite {
...
+                this.undoStack = undoStack;
+            }
+            if (redoStack) {
+                this.redoStack = redoStack;
+            }
+            this.updateUndoRedo();
+        }
+
+        getUndoStack() {
+            return this.undoStack;
```
riknoll: Should probably do a `.slice(0)` here; same with the redo stack. Will this get duplicated if I copy/paste a block?

jwunderl: It doesn't get get duplicated when duplicating the block, though that would be a real nice feature; I haven't interacted with blockly duplication stuff yet though so I'll have to look into how that would work

riknoll: I just checked; doesn't look possible. Copy/paste works with the XML and we don't want to put the stack in the XML

PR #5558, opened 2019-05-11
Send code through data-code attribute instead of src

ChaseMor:
The ```src``` attribute has an obtainable character limit when writing large programs. The ~~```name```~~ ```data-code``` attribute doesn't seem to have as strict limitations. 

This switches to send the code of a program to the docs simulator using the ~~```name```~~ ```data-code``` attribute so that larger programs can run on Edge and IE.

Fixes Microsoft/pxt-arcade#943


pelikhan:
Could we test this in all browser?

ChaseMor:
Tested and works on Chrome, Firefox, Edge, and IE 11.

PR #5563, opened 2019-05-14
Fix decompilation of non-block combined fields

jwunderl:
fixes https://github.com/microsoft/pxt-arcade/issues/966

```typescript
let a = bc.createTest()
a.data = 5
testNamespace.numberArgument(a.data)
```

goes to

arcade-screenshot (5)

(Fixes decompilation of https://forum.makecode.com/t/combo-attack-and-button-combination-extension/169, decompiles successfully to mostly gray blocks)


jwunderl:
Good call, It was just falling through and missing that check

PR #5566, opened 2019-05-14
Don't scroll monaco editor while scrolling flyout

jwunderl:
fixes https://github.com/microsoft/pxt-arcade/issues/205

### DIFF
--- a/webapp/src/monaco.tsx
+++ b/webapp/src/monaco.tsx
@@ -804,6 +804,7 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         if (flyout) {
             pxsim.U.clear(flyout);
             flyout.style.display = 'none';
+            flyout.removeEventListener("wheel", this.handleScroll, true);
         }
 
         // Hide the current toolbox category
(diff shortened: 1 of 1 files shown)


webapp/src/monaco.tsx
```diff
@@ -1606,6 +1606,9 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         monacoFlyout.style.display = 'block';
         monacoFlyout.className = 'monacoFlyout';
         monacoFlyout.style.transform = 'none';
+        monacoFlyout.addEventListener("wheel", e => {
```
riknoll: Don't we reuse this div? We should only register the event once and we should remove it when the flyout is hidden.

webapp/src/monaco.tsx
```diff
@@ -1598,6 +1599,8 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         return document.getElementById('monacoFlyoutWidget');
     }
 
+    handleScroll = (e: WheelEvent) => e.stopPropagation();
```
riknoll: Call this "handleFlyoutScroll" and make it private

PR #5571, opened 2019-05-16
Sprite kind with namespaces

riknoll:
Using a namespace instead of enums so that we can have built-in values without cluttering user code. If the user does define their own kind, they get a declaration like this:

```typescript
namespace SpriteKind {
    export const Rocket = SpriteKind.create();
}
```

The declaration will only decompile if it follows that exact format (the namespace can be omitted on the call to create). Examples of things that will not decompile:

```typescript
// Has to call create function
namespace SpriteKind {
    export const Rocket = 1234
}

// No statements besides exports
namespace SpriteKind {
    let Rocket = SpriteKind.create();

    music.ringTone(1234);
}

namespace SpriteKind {
    // No comments inside the namespace
    export const Rocket = SpriteKind.create();
}

// No nested namespaces
namespace SpriteKind.Whatever {
    export const Rocket = SpriteKind.create();
}
```

And here's the new way of defining SpriteKinds:

```typescript
namespace sprites {
    /**
     * Gets the "kind" of sprite
     */
    //% shim=KIND_GET
    //% blockId=spritekind block="$kind"
    //% kindNamespace=SpriteKind kindMemberName=kind kindPromptHint="e.g. Coin, Fireball, Asteroid..."
    export function _spriteKind(kind: number): number {
        return kind;
    }
}

namespace SpriteKind {
    let nextKind: number;
    export function create() {
        if (nextKind === undefined) nextKind = 1;
        return nextKind++;
    }

    //% isKind
    export const Player = create();

    //% isKind
    export const Projectile = create();

    //% isKind
    export const Food = create();

    //% isKind
    export const Enemy = create();
} 
```


tests/decompile-test/cases/sprite_kind.ts
```diff
@@ -0,0 +1,6 @@
+namespace SpriteKind {
+    export const Rocket = SpriteKind.create();
+    export let Whatever = SpriteKind.create();
```
jwunderl: Should it allow ``let``s? Currently it fails if ``SpriteKind.Whatever`` is reassigned outside the namespace (probably should never happen in real code, but who knows?):

```typescript
namespace SpriteKind {
    export let Whatever = SpriteKind.create()
}
SpriteKind.Whatever = 1
```

riknoll: Good catch

riknoll:
Okay, I have now added the arcade patching hack. @pelikhan  could you please re-review?

pelikhan:
Hack looks good.

pelikhan:
Nice work! This was not trivial!

PR #5580, opened 2019-05-17
Code Box Switch

toledoal:
This PR proposes the switch between code experience for Docs. In this case, the user switches the view of code, between javascript and blocks. This PR asumes we will eventually have Python as another language option. 

image

image

We also consider the scenario where people want to use the previous experience, in this case we added a flag "showToggleCode" to "docs.js" which shows the new behavior. We also added a "showBlocks" flag and "showPython" for the third language which enables the target configuration to show that type of code. We left showBlocks as default.

This changes doesn't affect/change the experience in the Editor.

image


pelikhan:
What problem is this change addressing?

toledoal:
For what I know the idea is for the user to be able to switch between the languages similar to this type of control 
image

pelikhan:
These controls are meant for pro-devs to specify their preferred language and are usually stored in a cookie so that future pages default to that language. In our case, we have an implicit default to "blocks", but we could have buttons to change that default to JS/Py. In that case, all snippets would default to the required rendering.
We are an educational tool and, IMO, there is more value in showing both the blocks and the JS/PY on the page to facilitate transitions from one code representation to the other.

toledoal:
I see the value on showing both block/JS-PY. Let me communicate this to others, just to reach a common agreement on the design side. But I agree 😄.  BTW in the Styling PR we added a style for this. (which might be enough right?)

image

pelikhan:
Style sounds good.

toledoal:
Is it ok for you if I close this PR, I think we will not implement this. The basic styling is already in style docs PR.

pelikhan:
Sounds good to me.

PR #5582, opened 2019-05-17
Prefer touch events to pointer events (fix android image editor)

jwunderl:
Fixes https://github.com/microsoft/pxt-arcade/issues/379

Would be good to test on chromebook and physical android device / windows with touchscreen with a target build of arcade

Issue is that the browsers were using the pointer events when they should have been using touch events.

Broken / current:

Fixed:

testing:

* [x] chrome - mac, no touch screen
* [x] chrome - android emulator
* [x] firefox - mac
* [x] safari - mac
* [x] safari - ipad simulator
* [x] chrome - android hardware
* ~[ ] chrome - touchscreen on chromebook~ don't have a touchscreen chromebook but should match android?
* [x] chrome - touchscreen on windows
* [ ] edge - touchscreen on windows
* [x] firefox - touchscreen on windows

### DIFF
--- a/pxtlib/browserutils.ts
+++ b/pxtlib/browserutils.ts
@@ -916,27 +916,34 @@ namespace pxt.BrowserUtils {
         leave: string
     }
 
-    export const pointerEvents: IPointerEvents = hasPointerEvents() ? {
-        up: "pointerup",
-        down: ["pointerdown"],
-        move: "pointermove",
-        enter: "pointerenter",
-        leave: "pointerleave"
-    } : isTouchEnabled() ?
-            {
+    export const pointerEvents: IPointerEvents = (() => {
+        if (isTouchEnabled()) {
+            return {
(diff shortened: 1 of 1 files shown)


jwunderl:
Works on all the above except for edge; edge seems to be ignoring the ``stopPropogation`` on drag events and treating the left / right touch drags like forward / back browser buttons. (exact same behavior as current live site, except maybe a tiny bit better in that you can draw upwards and downwards)

riknoll:
@jwunderl you probably need to call `e.preventDefault()`

jwunderl:
Oops, yeah said the wrong one; that should be being called in the drag as well though: https://github.com/microsoft/pxt/blob/dca2ad9fcab64f82a9c91aa9e6bd3a81fd75b4bd/pxtlib/sprite-editor/canvasGrid.ts#L355-L366

PR #5583, opened 2019-05-17
Added isEmpty for strings

ChaseMor:
Adds the previously unimplemented String.isEmpty function and block to ts

Fixes microsoft/pxt-microbit#2048


docs/static/playground/pxt-common/pxt-helpers.js
```diff
@@ -296,6 +296,10 @@ namespace helpers {
         }
         return q + r;
     }
+
+    export function stringEmpty(S: string): boolean {
+        return !S || S.length == 0;
```
pelikhan: !S should cover both cases.

PR #5589, opened 2019-05-20
Improves default parameter selection when dragging out snippets

darzu:
For example, dragging "onBlockPlaced" used to give (in TS & PY):
```
blocks.onBlockPlaced(0, function () { })
```
Now it gives:
```
blocks.onBlockPlaced(Block.Grass, function () { })
```

This improves both the typescript and python experience. This is also the foundation for doing the same improvement to our tab-complete. This is a huge usability improvement to our text experience, and it's especially important for our Python deliverable.

Lots of TODOs left in the code before it can be merged.

### DIFF
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -720,8 +720,7 @@ namespace ts.pxtc.service {
         isJsDocTagName: boolean;
     }
 
-    const blocksInfoOp = (apisInfoLocOverride?: pxtc.ApisInfo) => {
-        let bannedCategories = pxt.appTarget.runtime.bannedCategories
+    const blocksInfoOp = (apisInfoLocOverride: pxtc.ApisInfo, bannedCategories: string[]) => {
         if (apisInfoLocOverride) {
             if (!lastLocBlocksInfo) {
                 lastLocBlocksInfo = getBlocksInfo(apisInfoLocOverride, bannedCategories);
@@ -744,10 +743,6 @@ namespace ts.pxtc.service {
     }
 
     const operations: pxt.Map<(v: OpArg) => any> = {
-        initTarget: v => {
-            pxt.appTarget = v.target
-        },
-
         reset: () => {
             service.cleanupSemanticCache();
             lastApiInfo = null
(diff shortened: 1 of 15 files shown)


jwunderl:
Seems like this will fix https://github.com/microsoft/pxt-arcade/issues/321 ?

pxtcompiler/emitter/service.ts
```diff
@@ -1145,20 +1170,98 @@ namespace ts.pxtc.service {
...
+        console.log("attrs")
+        console.dir(attrs)
+        function getShadowSymbol(paramName: string): SymbolInfo | null {
+            let shadowBlock = (attrs._shadowOverrides || {})[paramName]
+            if (!shadowBlock)
+                return null
+            const shadowBlockMap: pxt.Map<string> = {
+                // shadow override -> qname
+                // TODO: determine this programatically somehow
+                "minecraftBlockField": "Block",
```
pelikhan: add section in pxtarget.json

pelikhan:
Is it possible to run some "getSnippet" as part of our test suite?

pxtcompiler/emitter/service.ts
darzu: We should be able to enumerate blocks and find ones with matching IDs

darzu:
> Is it possible to run some "getSnippet" as part of our test suite?

Good question. What sort of test are you thinking? Syntactic matching against well-known "good" results?

darzu:
> microsoft/pxt-arcade#321

Yes it should, thanks for the pointer. I assign it to myself.

pelikhan:
Just a set of know snippets and recorded expected outputs.

darzu:
> Just a set of know snippets and recorded expected outputs.

Currently nothing in the worker thread is tested or testable as far as I know.. I looked and I definitely want to do this but it's going to take some time (for me at least) to grok the structure and refactor things for testing. Pointers welcome. For now I'll not pull on that thread in this PR.

webapp/src/package.ts
```diff
@@ -15,8 +16,8 @@ let extWeight: pxt.Map<number> = {
 }
 
 export function setupAppTarget(trgbundle: pxt.TargetBundle) {
-    //if (!trgbundle.appTheme) trgbundle.appTheme = {};
     pxt.setAppTarget(trgbundle)
+    compiler.initAppTarget(trgbundle)
```
pelikhan: the app target is modified when a different board variant is loaded so you probably need this in more places.

darzu: Thanks, I now also set the app target when the appTargetVariant changes

pelikhan:
target bundle can be very large. best to pass the required data on the service call

darzu:
> target bundle can be very large. best to pass the required data on the service call

Thanks, I changed it to do this. It works

PR #5592, opened 2019-05-22
Fix color parsing in field_colour

jwunderl:
fixes https://github.com/microsoft/pxt-arcade/issues/988 (field editor bug, not decompiler)

Also:

* doesn't fail on invalid indices / super big numbers
* allows shortcut hex codes to decompile as expected when decompiling (e.g. ``0xfff``)

This does have slightly weird behaviors in some cases, though - for example, the ``set background color`` block in arcade will not match the hex inputs, but if you decompile to blocks it will behave as you would normally expect:

https://makecode.com/_UawYdtFWWhsA

The same would apply in adafruit (shortcut syntax doesn't work in typescript but decompiling makes it work). ~Maybe we could add a helper function to parse the input and make the behavior consistent for all the functions that use this field editor?~ I suppose there isn't really too much to do about this, as there isn't anything to disambiguate index vs color with unless you actually have the literal itself - this field editor is just converting typescript that doesn't make sense into blocks that presumably match the intention.

### DIFF
--- a/pxtblocks/fields/field_colour.ts
+++ b/pxtblocks/fields/field_colour.ts
@@ -69,7 +69,7 @@ namespace pxtblockly {
         setValue(colour: string) {
             colour = parseColour(colour);
 
-            if (this.valueMode_ === "index") {
+            if (!/^#/.test(colour) && this.valueMode_ === "index") {
                 const allColors = this.getColours_();
                 if (allColors.indexOf(colour) === -1) {
                     // Might be the index and not the color
@@ -112,7 +112,7 @@ namespace pxtblockly {
     function parseColour(colour: string) {
         if (colour) {
             const enumSplit = /Colors\.([a-zA-Z]+)/.exec(colour);
-            const hexSplit = /0x([0-9a-fA-F]+)/.exec(colour);
+            const hexSplit = /[0x|#]([0-9a-fA-F]+)/.exec(colour);
 
             if (enumSplit) {
                 switch (enumSplit[1].toLocaleLowerCase()) {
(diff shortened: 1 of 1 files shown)


pxtblocks/fields/field_colour.ts
```diff
@@ -67,7 +67,8 @@ namespace pxtblockly {
          * @param {string} colour The new colour in '#rrggbb' format.
          */
         setValue(colour: string) {
-            colour = fixEnumColor(colour);
+            colour = parseColour(colour);
+
             if (colour.indexOf('0x') > -1) {
```
pelikhan: ``/^0x/.test(colour)``

pxtblocks/fields/field_colour.ts
```diff
@@ -90,6 +91,10 @@ namespace pxtblockly {
                 Blockly.Events.fire(new (Blockly as any).Events.BlockChange(
                     this.sourceBlock_, 'field', this.name, this.colour_, colour));
             }
+
+            // if color is not valid, do not apply color
+            if (!/#[0-9a-fA-F]{6}/.exec(colour)) return;
```
pelikhan: use .test instead of .exec

pxtblocks/fields/field_colour.ts
```diff
@@ -67,8 +67,9 @@ namespace pxtblockly {
          * @param {string} colour The new colour in '#rrggbb' format.
          */
         setValue(colour: string) {
-            colour = fixEnumColor(colour);
-            if (colour.indexOf('0x') > -1) {
+            colour = parseColour(colour);
+
+            if (/^0x/.test(colour)) {
```
pelikhan: this should be part of "parseColour"

PR #5593, opened 2019-05-22
Bidirectional builtin function map, override pyQname for special cases

shakao:
Fixes microsoft/pxt-minecraft/issues/858, and max/min/randint conversion between Python and TS.


darzu:
Awesome!!

PR #5594, opened 2019-05-22
Update tutorial hint display

shakao:
tutorial-hint

Latest: https://makecode.adafruit.com/app/b65df62c3b3e2df682137ed0503b7e7bb21edae9-052b95a621

### DIFF
--- a/theme/tutorial.less
+++ b/theme/tutorial.less
@@ -145,7 +145,6 @@ body#docs.tutorial {
     width: 100%;
     font-size: 12pt;
     height: calc(@tutorialCardHeight - 1rem);
-    padding-left: @avatarMargin + 0.5rem;
     margin-bottom: 0.4rem;
     overflow-y: hidden;
     overflow-x: auto;
@@ -164,10 +163,11 @@ body#docs.tutorial {
 }
 
 .tutorial #tutorialcard .tutorialmessage {
-    width: 100%;
+    width: ~'calc(100% - @{avatarMargin} - 0.5rem)';
     padding: 0.5rem;
     height: calc(@tutorialCardHeight - 1.5rem);
     margin-bottom: 1rem;
+    margin-left: @avatarMargin + 0.5rem;
     overflow: hidden;
 }
 
@@ -318,6 +318,7 @@ span.highlight-line {
     z-index: @tutorialLightboxCardZIndex;
     border: 8px solid @primaryColor;
     box-shadow: 0px 0px 8px 1px rgba(0, 0, 0, .3);
+    max-height: 70vh;
 }
 
 .tutorialhint:before {
@@ -332,7 +333,6 @@ span.highlight-line {
     border: 16px solid;
     border-color: transparent @primaryColor transparent transparent;
     transform: rotateZ(-45deg);
-
 }
 
 .tutorialhint.hidden {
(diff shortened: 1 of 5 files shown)


pelikhan:
The hint can not be blown on ios

shakao:
Should the blocks be visible? It's nice to have a separation between the "instruction" and the solution/blocks, so the "See more" only displays the text. Straightforward fix, if we want it though.

abchatra:
Jacqueline doesn't want blocks to be show with more. More is for instructions which didn't fit.

pelikhan:
The reason for "more" is to be able to see both the expected code section and the user code on the same screen without any kind of dialog.
On your first step, the hint dialog shows text. On the second it doesn't. Seems inconsistent.

pelikhan:
On step 1, click two times on the hint and the text goes away.

pelikhan:
The whole editor scrolls up and down when the hint bleeds off.
image

pelikhan:
I think the blocks need to be visible for early tutorials. We are teaching how blocs snap as much as how the editor is being used. Not showing the blocs is a major untested change compared to the current tutorial engine.

shakao:
Updated link with the blocks automatically visible for tutorials with the fullscreen tag: https://makecode.adafruit.com/app/643f557e06a785fd3e789c8e93a909ff9a0cb09d-427b673194

Not sure if this is what you were addressing, Peli?

abchatra:
>> The reason for "more" is to be able to see both the expected code section and the user code on the same screen without any kind of dialog.

@Jaqster strongly belives we shouldn't show answer in more button


 >> On step 1, click two times on the hint and the text goes away.

This intentional to draw attention to the hint on the first UI. First time we show the instruction with the full solution. After that instruction goes to the text area and solution goes to the hint. 


>> I think the blocks need to be visible for early tutorials. We are teaching how blocs snap as much as how the editor is being used. Not showing the blocs is a major untested change compared to the current tutorial engine.

I agree with the above feedback. We should show the solution in each tutorial and shouldn't break existing flow

shakao:
Text is visible in the first step because we automatically flash the dialog, to let users know it's there, but we don't want to display only the solution. Any time the hint is clicked after that, the dialog only shows the solution, since the text is always already visible in the tutorial card.

pelikhan:
The purpose of the tutorial is to take fresh users through the coding experience. Not showing the blocs makes it much harder for them to achieve the steps. This change was made to support Python/TS tutorials where seeing the code is critical for non-bloc based languages where syntax is way more important.

If you want to draw attention on the hint, show the tip for the hint. not the hint itself. Having the hint showing text then no text looks like a bug.

The difference between hint vs no-hint is way too subtle IMO. The little bubble can easily be overlooked.

abchatra:
Shannon latest change shows the full solution when you go to the step. Does that suffice? If not I think we need to discuss on a call. 

This is the same behavior as the live website where the modal shows the full solution at every step if @fullscreen is specified.

pelikhan:
The more section was added to avoid the disconnect between the blocks in dialog and the code. The tooltip is another form of dialog which has the same problem. You can’t compare side by side your solution to the correct solution.

pelikhan:
Hint should show the ### hint content, which may be the code solution. Let the tutorial author decide if the code needs to be in the main section.

And the hint button needs to be much more obvious.

shakao:
Latest: https://makecode.adafruit.com/app/39c1eec283ee78479b2fa6b3728ee3172e89bfb1-2ec7dad50f

theme/tutorial.less
```diff
@@ -216,6 +216,13 @@ i.icon.avatar-image {
     }
 }
 
+.tutorialhint .ui.segment .blocklyPreview {
+    width: 100%;
+    @media only screen and (min-height: 400px) {
```
pelikhan: do we have a breakpoint variable defined in our LESS for this?

shakao: no, i copied this from the previous modal height breakpoint (css above this line), but i can move it into the less? not sure how it was previously determined

PR #5606, opened 2019-05-24
add default array.sort behavior for numbers and strings

jwunderl:
fixes https://github.com/microsoft/pxt-microbit/issues/2072
fixes https://github.com/microsoft/pxt-arcade/issues/782

matches javascript at least? Sorting number arrays isn't generally useful in javascript without adding the callback, but at least it will do something.

### DIFF
--- a/docs/static/playground/pxt-common/pxt-helpers.js
+++ b/docs/static/playground/pxt-common/pxt-helpers.js
@@ -116,11 +116,17 @@ namespace helpers {
     }
 
     export function arraySort<T>(arr: T[], callbackfn?: (value1: T, value2: T) => number): T[] {
-        if (!callbackfn) {
-            //TODO: support native strings and number sorting
-            /* callbackfn = function (value1: string, value2: string) : number {
-                return value1.compare(value2);
-                }*/
+        if (!callbackfn && arr.length > 1) {
+            const firstElement = arr[0];
+            if (typeof(firstElement) == "string") {
+                callbackfn = (a, b) => {
+                    return (a as any as string).compare(b as any as string);
+                }
+            } else if (typeof(firstElement) == "number") {
+                callbackfn = (a, b) => {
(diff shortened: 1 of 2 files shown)


abchatra:
Lets wait on this for microbit release. Lets take the fix after the release.

pelikhan:
Moving to dev branch.

pelikhan:
Any chance to add a few tests?

jwunderl:
Is there a place to add tests for these libs files? I'd be happy to, I think there are a decent number of things missing tests in this file that would be good to add tests for

riknoll:
You could copy the compiler test runner and do it there

riknoll:
It runs the simulator in the command line and you can verify the results with console logs

abchatra:
Why are we changing sorting from selection sort? They are typically great for small array. Are there any cases where user has more than 100 elements in an array?

jwunderl:
I mainly changed it on a whim after richard mentioned it at lunch, the largest typical case would be in arcade with sorting sprites / spritelikes for their ordering on the screen.

Was thinking about reverting that for the same reasoning you mentioned earlier today though

riknoll:
If it gives us a performance gain in arcade it might be worth switching for large arrays. Need to verify that it actually makes a difference though.

riknoll:
Relevant piece of code: https://github.com/microsoft/pxt-common-packages/blob/master/libs/game/scene.ts#L134

jwunderl:
I'll cherry pick the heapsort change (https://github.com/microsoft/pxt/pull/5606/commits/9c3bfa8b06a635f67b3285b35e0d68468513aa2f) to another branch and make that a separate PR / test if there's any perf difference - it makes sense to separate a bug fix from a perf change anyway. Probably also make another separate pr for the libs test runner and some initial tests?

PR #5607, opened 2019-05-24
Tutorial tooltip appears & animates after 30 seconds of inactivity

shakao:
- Timer resets on blockly events (block moved, etc)
- Avatar and tooltip click will open the hint
- Tooltip stays visible for two seconds, then hides

tutorial-tooltip

https://makecode.microbit.org/app/bb1c11d932bf519575143f3bfef42c2c2aa1a844-11baa0cb0c

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -2619,19 +2619,22 @@ export class ProjectView
     }
 
     showImportGithubDialog() {
-        dialogs.showImportGithubDialogAsync().done(url => {
-            if (url === "NEW") {
-                dialogs.showCreateGithubRepoDialogAsync()
-                    .then(url => {
-                        if (url)
-                            importGithubProject(url)
-                    })
-            } else if (!pxt.github.isGithubId(url)) {
-                core.errorNotification(lf("Sorry, the project url looks invalid."));
-            } else {
-                importGithubProject(url);
-            }
-        });
+        dialogs.showImportGithubDialogAsync()
+            .then(url => {
+                if (url === "NEW") {
+                    dialogs.showCreateGithubRepoDialogAsync()
+                        .then(url => {
+                            if (url) {
+                                importGithubProject(url);
+                            }
+                        })
+                } else if (url) {
+                    importGithubProject(url);
+                }
+            }, e => {
+                core.errorNotification(lf("Sorry, that repository looks invalid."));
+            })
+            .done();
     }
 
     showImportFileDialog(options?: pxt.editor.ImportFileOptions) {
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -58,6 +58,7 @@ type ProjectCreationOptions = pxt.editor.ProjectCreationOptions;
 
 import Cloud = pxt.Cloud;
 import Util = pxt.Util;
+import { HintManager } from "./hinttooltip";
 
 pxsim.util.injectPolyphils();
 
@@ -113,6 +114,9 @@ export class ProjectView
 
     private runToken: pxt.Util.CancellationToken;
 
+    // component ID strings
+    private static readonly tutorialCardId = "tutorialcard";
+
     constructor(props: IAppProps) {
         super(props);
         document.title = pxt.appTarget.title || pxt.appTarget.name;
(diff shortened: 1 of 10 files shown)


webapp/src/app.tsx
```diff
@@ -2874,6 +2874,11 @@ export class ProjectView
         tc.toggleHint(showFullText);
     }
 
+    setTutorialHintAnimation() {
+        let tc = this.refs["tutorialcard"] as tutorial.TutorialCard;
+        tc.setHintAnimation();
```
pelikhan: is there a possible race here? (timer triggers while dom is being removed)

webapp/src/blocks.tsx
```diff
@@ -449,6 +449,11 @@ export class Editor extends toolboxeditor.ToolboxEditor {
                     // }
                 }
             }
+
+            // reset tutorial hint animation on any blockly event
+            if (this.parent.state.tutorialOptions != undefined) {
+                this.parent.setTutorialHintAnimation();
```
pelikhan: can we use something a bit more generic like "pokeUserActivity()" ? We should add it to download, menu items and somewhere in monaco where save happens (for python tutorials)

webapp/src/tutorial.tsx
```diff
@@ -343,6 +359,21 @@ export class TutorialCard extends data.Component<ISettingsProps, TutorialCardSta
         this.setState({showSeeMore: show});
     }
 
+    clearHintAnimation() {
+        this.setState({shakeHint: false});
+        clearTimeout(this.timer);
+    }
+
+    setHintAnimation() {
```
pelikhan: rename this to notify that it is only poking.

webapp/src/tutorial.tsx
```diff
@@ -343,6 +359,21 @@ export class TutorialCard extends data.Component<ISettingsProps, TutorialCardSta
...
+        if (this.hasHint()) {
+            this.clearHintAnimation();
+            this.timer = setTimeout(() => {
```
pelikhan: we have a debounce helper for this kind operation

pelikhan: i think something like util.debounce(...

webapp/src/app.tsx
```diff
@@ -2874,6 +2874,16 @@ export class ProjectView
         tc.toggleHint(showFullText);
     }
 
+    ///////////////////////////////////////////////////////////
+    ////////////     User alert/notifications     /////////////
+    ///////////////////////////////////////////////////////////
+
+    pokeUserActivity() {
+        // animate tutorial hint after some time of user inactivity
+        let tc = this.refs["tutorialcard"] as tutorial.TutorialCard;
```
pelikhan: i think we should check the state of the app, e.g. in the editor with tutorial open so that we can start poping other hints in the future.

webapp/src/app.tsx
```diff
@@ -2874,6 +2874,18 @@ export class ProjectView
...
 
+    ///////////////////////////////////////////////////////////
+    ////////////     User alert/notifications     /////////////
+    ///////////////////////////////////////////////////////////
+
+    pokeUserActivity() {
+        if (!!this.state.tutorialOptions && !!this.state.tutorialOptions.tutorial) {
+            // animate tutorial hint after some time of user inactivity
+            let tc = this.refs["tutorialcard"] as tutorial.TutorialCard;
+            if (tc) tc.startPokeUserActivity();
```
pelikhan: IMO, we should have a centralized "tutorial engine" that is in charge of receiving pokes, debouncing and changing the correct state to generate animation -- we don't want timers scattered in the UI. So move the timer from the tutorial card into a "hint engine" (might as well start a new file to keep things tidy) and centralize operations there.

webapp/src/tutorial.tsx
```diff
@@ -153,16 +153,24 @@ interface TutorialCardState {
     popout?: boolean;
     showHintTooltip?: boolean;
     showSeeMore?: boolean;
+    pokeUser?: boolean;
 }
 
 export class TutorialCard extends data.Component<ISettingsProps, TutorialCardState> {
-    public focusInitialized: boolean;
     private prevStep: number;
+    private timer: number;
```
pelikhan: see comment above, refactor in dedicated class -- or this will become messy soon. Eventually the hint engine state machine should be come quite complex.

shakao:
https://makecode.microbit.org/app/bb1c11d932bf519575143f3bfef42c2c2aa1a844-11baa0cb0c

PR #5613, opened 2019-05-29
fixes for large classes and static fields

mmoskal:
* we had a bug where field access in classes over 127 bytes wouldn't be emitted correctly
* static fields used to be included in every instance of a class

@abchatra  After some testing on `dev` you may want this on `master`.


jwunderl:
nice, fixes https://github.com/microsoft/pxt-arcade/issues/1008

abchatra:
Ok

PR #5614, opened 2019-05-29
Fix decompilation for microbit game of life example

jwunderl:
Fixes https://github.com/microsoft/pxt-microbit/issues/2079

Issue was having a user function in a gray block return a boolean; in this case, the CallExpression itself had a ``.callInfo`` property because it was already a gray block, rather than having it on the expression itself

microbit-screenshot (4)


abchatra:
Looks good. Thanks for adding the test.

PR #5617, opened 2019-05-30
[gh-integration] make 'Auto-initialized' contain all necessary files

jwunderl:
fixes https://github.com/microsoft/pxt-arcade/issues/1011

Package all the files into the second 'Auto-initialized' commit instead of separating each one out as 10 separate commits

current behavior: https://github.com/jwunderl/pxt-buncha-commits-test/commits/master
this change: https://github.com/jwunderl/pxt-test-fixes/commits/master

### DIFF
--- a/webapp/src/workspace.ts
+++ b/webapp/src/workspace.ts
@@ -478,7 +478,6 @@ export async function bumpAsync(hd: Header) {
 }
 
 export async function commitAsync(hd: Header, msg: string, tag = "", filenames: string[] = null) {
-    // call this
     let files = await getTextAsync(hd.id)
     let gitjsontext = files[GIT_JSON]
     if (!gitjsontext)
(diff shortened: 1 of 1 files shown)


abchatra:
@jwunderl does this need to be in microbit?

jwunderl:
I would say yes, as it is a regression from the live site for microbit. 

It is mainly just an annoying issue, though (but does create permanent minor annoyance for anyone who creates a new extension), so we can revert and put this in dev if preferred

abchatra:
Makes sense. Lets keep it here. What change regressed this?

jwunderl:
https://github.com/microsoft/pxt/pull/5581 fixed a bug where files were not being generated immediately which would cause annoying conflicts if anyone created the repo and pulled before those files ended up getting generated. It just fixed that issue by putting all the files separately, instead of packaging them into a single commit

abchatra:
Cool thanks!

pelikhan:
Yes we need this in micro:bit. I fixed this issue following up supporting the "let's talk science" team journey through publishing a new package. One of the problem is that the online editor did not generate files like the license, .gitignore and things really got confusing when they had to use VS Code ... which tried to commit tons of build files.

PR #5618, opened 2019-05-30
Sprite editor marquee tool

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/949

sprite_editor_marquee

It works just like the select tool in paint, but with transparency! You can also press backspace to delete a region or use the arrow keys to nudge it around. To clear the selection, you can either click outside the selected area or start using a different tool. Tested in Chrome and Edge but someone should probably test Safari as well.

This PR also fixes a few lifecycle bugs in the sprite editor that caused the undo/redo stack to not always persist.

Still to do:
1. Copy and paste!

Known bugs:
1. The floating layer disappears when you resize the canvas (while the resize overlay animation is showing)


pxtlib/sprite-editor/bitmap.ts
```diff
@@ -41,10 +41,14 @@ namespace pxtsprite {
...
 
-        apply(change: Bitmap) {
+        apply(change: Bitmap, transparent = false) {
+            let current: number;
             for (let c = 0; c < change.width; c++) {
                 for (let r = 0; r < change.height; r++) {
-                    this.set(change.x0 + c, change.y0 + r, change.get(c, r));
+                    current = change.get(c, r);
+
+                    if (!current && transparent) continue;
```
jwunderl: Can just negate the condition here and call next line in conditional here.

pxtlib/sprite-editor/spriteEditor.ts
```diff
@@ -386,6 +395,35 @@ namespace pxtsprite {
                 this.altDown = true;
             }
 
+            if (this.activeTool === PaintTool.Marquee && this.state.floatingLayer) {
+                let didSomething = true;
+                switch (event.keyCode) {
+                    case 8: // Backspace
```
jwunderl: backspace is closing the image editor for me and deleting the block, needs ``event.stopPropgation`` i think

pelikhan:
@riknoll i am rebasing this PR to the dev branch as it is not micro:bit related.

PR #5620, opened 2019-05-31
Sprite markee fixes

pelikhan:
- "m" shortcut
- don't stack overflow on close


pxtlib/sprite-editor/tools.ts
```diff
@@ -26,6 +26,8 @@ namespace pxtsprite {
                 return "l";
             case PaintTool.Erase:
                 return "e";
+            case PaintTool.Marquee:
+                return "m";
```
jwunderl: The rest match piskel, so I think this would be "s"

PR #5623, opened 2019-05-31
Added a tick event to the Make a function button

JacobBovee:
Now we'll be able to track usage of the make a function button.


pxtblocks/blocklyloader.ts
```diff
@@ -2230,6 +2230,8 @@ namespace pxt.blocks {
             workspace.registerButtonCallback('CREATE_FUNCTION', function (button) {
                 let promptAndCheckWithAlert = (defaultName: string) => {
                     Blockly.prompt(newFunctionTitle, defaultName, function (newFunc) {
+                        // Tick event to monitor usage of "Make a function" button
```
darzu: might be redundant since the event it pretty self-explanatory

pxtblocks/blocklyloader.ts
```diff
@@ -2230,6 +2230,8 @@ namespace pxt.blocks {
...
+                        pxt.tickEvent('blocks.make_a_function');
```
pelikhan: there isn't a naming convention for ticks, but we never use _, can you drop them?

darzu:
@JacobBovee , you can merge this PR whenever, maybe after you adjust to the feedback above

pelikhan:
ping

PR #5624, opened 2019-05-31
do not issue error when cancelling github import dialogue

jwunderl:
fixes https://github.com/microsoft/pxt-microbit/issues/2107

The other error I mentioned in the issue, an exception that gets through whenever you create a new extension, is coming from the call in line 668 here:

https://github.com/microsoft/pxt/blob/20595b015c9ba303c24cfe67dedef0061bcec76d/webapp/src/workspace.ts#L667-L679

It gets propagated to the console before the catch handles it, presumably due to the restructuring of async / await functions to make them cross platform. I suppose it could be rewrittten to just use promises instead, but that seems like a bit much just to suppress an error in the console

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -2621,17 +2621,15 @@ export class ProjectView
     showImportGithubDialog() {
         dialogs.showImportGithubDialogAsync()
             .then(url => {
-                if (url) {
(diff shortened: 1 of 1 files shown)


pelikhan:
This one could go in microbit. @abchatra

abchatra:
Sure

PR #5626, opened 2019-06-01
Fix webusb for microbit

darzu:
Fixes WebUSB in microbit. The issue is that we set a "deployCoreAsync" function in pxt-core which can be overridden by the target (microbit may be the only one that does this) and we were reverting the target override after USB pairing. Likely introduced by https://github.com/microsoft/pxt/pull/5517.

The fix was to keep the pxt-core deploy function selection separate from the target's deploy function override and then always prefer the target's override if it exists at the callsite.

related: https://github.com/microsoft/pxt-microbit/pull/2115

fixes: https://github.com/microsoft/pxt-microbit/issues/2089

https://makecode.microbit.org/app/fc71444bb8dcd5f94c4ac4f8ca9c1ba0f0342c34-93265192c2

TODO:
- [x] microbit sibiling PR
- [x] test arcade webusb
- [x] test arcade download
- [x] test adafruit webusb
- [x] test adafruit download
- [x] test microbit download

### DIFF
--- a/pxteditor/editor.ts
+++ b/pxteditor/editor.ts
@@ -330,7 +330,7 @@ namespace pxt.editor {
         resourceImporters?: IResourceImporter[];
         beforeCompile?: () => void;
         patchCompileResultAsync?: (r: pxtc.CompileResult) => Promise<void>;
-        deployCoreAsync?: (r: pxtc.CompileResult) => Promise<void>;
+        deployAsync?: (r: pxtc.CompileResult) => Promise<void>;
         saveOnlyAsync?: (r: ts.pxtc.CompileResult) => Promise<void>;
         saveProjectAsync?: (project: pxt.cpp.HexFile) => Promise<void>;
         showUploadInstructionsAsync?: (fn: string, url: string, confirmAsync: (options: any) => Promise<number>) => Promise<void>;
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -2004,7 +2004,8 @@ export class ProjectView
                         });
                     }
                 }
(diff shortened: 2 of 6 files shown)


webapp/src/app.tsx
```diff
@@ -2004,7 +2004,8 @@ export class ProjectView
                         });
                     }
                 }
-                return pxt.commands.deployCoreAsync(resp, {
+                let doDeploy = pxt.commands.deployAsync.target || pxt.commands.deployAsync.core;
```
riknoll: I feel like this logic should be in `cmds.ts`

pxteditor/editor.ts
```diff
@@ -330,7 +330,7 @@ namespace pxt.editor {
         resourceImporters?: IResourceImporter[];
         beforeCompile?: () => void;
         patchCompileResultAsync?: (r: pxtc.CompileResult) => Promise<void>;
-        deployCoreAsync?: (r: pxtc.CompileResult) => Promise<void>;
+        deployAsync?: (r: pxtc.CompileResult) => Promise<void>;
```
pelikhan: why are we renaming this field?

darzu: Do we have a clean delineation of core <-> target contracts? How do we know what an editor might depend on?

pelikhan:
Every editor ? 😊

darzu:
@pelikhan , I updated it to be backwards compatible. Works in Adafruit, Arcade, and MicroBit.

webapp/src/app.tsx
darzu: done

PR #5627, opened 2019-06-01
use heap sort for sorting large arrays

jwunderl:
Separated from https://github.com/microsoft/pxt/pull/5606

use heap sort for sorting arrays with length > 50. Went with heap sort because:

* javascript sort is not guaranteed to be stable, and none of our use cases require it to be (and typical general purpose fast sorts will require extra space)
* only uses O(1) extra memory (quick sort typically requires O(log(n)) extra space https://en.wikipedia.org/wiki/Quicksort#Space_complexity, although I don't know if that would ever actually matter in our case)
* Θ(n*log(n)) runtime

marking as draft as I'll want to do some perf tests to see if the difference is actually noticeable for typical use cases (e.g. sorting spritelikes in arcade). Also would want to write the start of the libs tests runner before merging either way

### DIFF
--- a/docs/static/playground/pxt-common/pxt-helpers.js
+++ b/docs/static/playground/pxt-common/pxt-helpers.js
@@ -99,11 +99,28 @@ namespace helpers {
         arr[j] = temp;
     }
 
-    function sortHelper<T>(arr: T[], callbackfn?: (value1: T, value2: T) => number): T[] {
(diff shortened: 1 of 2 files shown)


pelikhan:
Should we support both and select the algo based on size?

jwunderl:
Yup, added that in the second commit; I arbitrarily chose 50 to start but will need to test to see where the algos would switch

PR #5629, opened 2019-06-03
add pxt.simpleCompileAsync()

mmoskal:
this is in support of in-app compilation on iOS


mmoskal:
Sure. I think it would be optimal if we could re-work package to have synchronous loading - it would simplify running the scripts from JSC or node.js VM module.

PR #5633, opened 2019-06-03
Recipes

pelikhan:
Essencially, allows to start a tutorial within an existing program. In Aracde, this means that "game concepts" can be followed while making a game.

- [x] slurp list of recipes from the /recipes markdown page
- [x] adding experiment
- [x] don't leave editor when finishing tutorial
- [x] add "tutorials" button in toolbox above advance
- [x] filter tutorials based on editor type

recipe

Use https://github.com/microsoft/pxt-arcade/tree/ututorials to test locally in arcade.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -2822,7 +2822,7 @@ export class ProjectView
                 tutorialRecipe: !!recipe
             };
 
-            // start a tutorial within the context of an exisintg
+            // start a tutorial within the context of an existing program
             if (recipe) {
                 const header = pkg.mainEditorPkg().header;
                 header.tutorial = tutorialOptions;
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -2848,10 +2848,14 @@ export class ProjectView
         // clear tutorial field
         const tutorial = this.state.header.tutorial;
         if (tutorial) {
-            this.state.header.tutorialCompleted = {
-                id: tutorial.tutorial,
-                steps: tutorial.tutorialStepInfo.length
-            }
+            // don't keep track of completion for microtutorials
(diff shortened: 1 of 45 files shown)


webapp/src/app.tsx
```diff
@@ -2719,29 +2725,32 @@ export class ProjectView
...
         let reportId: string = undefined;
-        let tutorialmd: string;
         let title: string;
         let autoChooseBoard: boolean = true;
         let dependencies: pxt.Map<string> = {};
         let features: string[] = undefined;
         let p: Promise<string>;
+
+        // make sure we are in the editor
+        recipe = recipe && !!this.state.header;
```
darzu: Can a regular tutorial run when not in the editor?

webapp/src/app.tsx
```diff
@@ -3299,6 +3328,11 @@ function handleHash(hash: { cmd: string; arg: string }, loading: boolean): boole
             editor.startTutorial(hash.arg);
             pxt.BrowserUtils.changeHash("");
             return true;
+        case "recipe": // shortcut to a tutorial. eg: #tutorial:tutorials/getting-started
+            pxt.tickEvent("hash.tutorial")
```
darzu: new ticket event? hash.recipe

pelikhan: good catch

webapp/src/app.tsx
pelikhan: It can be triggered by a hash change

pelikhan:
@shakao could you check that i merged your changes properly?

pelikhan:
This is ready for review. Updates from UI meeting integrated.

shakao:
looks good to me!

PR #5636, opened 2019-06-04
string.indexOf block

pelikhan:
Add blocks for indexOf, includes, split.
Fix for https://github.com/microsoft/pxt-microbit/issues/1990
image


abchatra:
We don't want to add new blocks at this point of time for microbit. May be after release.

pelikhan:
Moving to dev branch.

PR #5638, opened 2019-06-04
Add Math 'value functions' doc sections

ganicke:
Add sections for ``round``, ``ceiling``, ``floor``, and ``truncate`` to the _blocks/math_ doc.

Fixes https://github.com/microsoft/pxt-microbit/issues/2117

### DIFF
--- a/common-docs/blocks/math.md
+++ b/common-docs/blocks/math.md
@@ -101,6 +101,38 @@ let minval = Math.min(0, 1);
 let maxval = Math.max(8, 2);
 ```
 
+## Round
+
+If a number has a fractional part, you can make the number change to be the closest, next integer value. This is called _rounding_. Rounding the number `6.78` will make it be `7` and rounding `9.3` will give you `9`. If a number has a fractional part greater than or equal to `0.5`, the number will round up to the next whole integer value. Otherwise, it will round down to the next lowest integer value.
+
+```block
+let rounded = Math.round(5.5)
+```
+
+## Ceiling
+
+To make a number change to the next higher whole number (integer), get the number's _ceiling_ value. The ceiling value for `1.234` is `2` since that is the next higher whole number.
+
+```block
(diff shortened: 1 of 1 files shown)


HyperDuino:
Is there any functional difference between truncate and floor? I do professional development for teachers, and write instructional materials for both students and teachers, and every moment spent explaining something like "floor really means round down", or "truncate is the same as floor which really means round down" adds to their cognitive load, frustration, and takes away valuable time.  I very much would like to suggest considering consolidating floor and truncate if there is not really a difference, and renaming floor to round down, ceiling to round up.  Thank you for considering this.

pelikhan:
``trunc`` acts as C trunk, it rounds the argument towards zero.

HyperDuino:
Sorry to be dense on this.  How is that different than floor?

From the proposed [help doc](https://github.com/microsoft/pxt/pull/5638/files): 

> floor:  `8.76` is `8` since that is the next lower whole number.

> truncate: If a number has the value `54.234` its truncated value is `54`.

jwunderl:
``floor`` of negative numbers rounds down (towards negative infinity), ``trunc`` of negative numbers rounds towards 0

HyperDuino:
Thank you. I had been overlooking negative numbers and what to do in those cases.  But now it would seem it's necessary to explain "floor" and "negative infinity" when really all was anted was to get the whole number portion of a value.  This still feels like a case where the simplicity of a programming language for children (and adults who wish to avoid stress, myself included) is being unnecessarily complicated.

What about having the blocks be "round", "whole", "least whole", and "greatest whole", 

This would also be conceptually friendly with the words of the greater-than and less-than logic blocks.

whole (5.84) = 5.00
whole (-5.84) = -5.00

least whole (5.84) = 5.00
least whole (-5.84) = -6.00

greatest whole (5.84) = 6.00
greatest whole (-5.84) = -6.00

PR #5639, opened 2019-06-04
Open doc pages from editor frame

ChaseMor:
Switches docs to open from the editor frame rather than the side doc frame

When the editor frame messages the side docs frame to open the page in a new tab, instead of the side docs frame opening the tab and sending a completion message, this sends back the url to be opened and the editor frame opens the tab instead.

Fixes Microsoft/pxt-microbit#2103

### DIFF
--- a/pxtrunner/runner.ts
+++ b/pxtrunner/runner.ts
@@ -397,11 +397,11 @@ namespace pxt.runner {
                     const docsUrl = pxt.webConfig.docsUrl || '/--docs';
                     let verPrefix = mp[2] || '';
                     let url = mp[3] == "doc" ? (pxt.webConfig.isStatic ? `/docs${mp[4]}.html` : `${mp[4]}`) : `${docsUrl}?md=${mp[4]}`;
-                    window.open(BrowserUtils.urlJoin(verPrefix, url), "_blank");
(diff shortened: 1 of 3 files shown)


pxtrunner/runner.ts
```diff
@@ -397,11 +397,11 @@ namespace pxt.runner {
                     const docsUrl = pxt.webConfig.docsUrl || '/--docs';
                     let verPrefix = mp[2] || '';
                     let url = mp[3] == "doc" ? (pxt.webConfig.isStatic ? `/docs${mp[4]}.html` : `${mp[4]}`) : `${docsUrl}?md=${mp[4]}`;
-                    window.open(BrowserUtils.urlJoin(verPrefix, url), "_blank");
                     // notify parent iframe that we have completed the popout
                     if (window.parent)
-                        window.parent.postMessage(<pxsim.SimulatorDocsReadyMessage>{
-                            type: "popoutcomplete"
+                        window.parent.postMessage(<pxsim.SimulatorOpenDocMessage>{
```
pelikhan: Always check that parent is defined

ChaseMor: Does ```if (window.parent)``` not do that?

PR #5640, opened 2019-06-04
move typings for file headers

jwunderl:
Bump typing for headers to pxtarget so they are accessible across pxt (in particular to pxtlib and electron).

Pretty harmless but not sure it would be worth taking before microbit release just for some missing typing info?


localtypings/pxtarget.d.ts
```diff
@@ -779,4 +779,51 @@ declare namespace pxt.tutorial {
         // number of steps completed
         steps: number;
     }
+}
+
+declare namespace pxt.Cloud {
```
pelikhan: why are we moving this in this file? This is our internal data representation and users should, in general, not take a dependency on this.

pelikhan: place in a separate file.

jwunderl:
They should be in the package as that directory is listed with a glob in package.json, right? I don't have any experience publishing to npm so I can't say for sure

https://github.com/microsoft/pxt/blob/1bfa3fd646c899cfa86ad158dac60bc6062c25b7/package.json#L40

PR #5645, opened 2019-06-06
Upgrade to Typescript 3.5

darzu:
This upgrades us to ts 3.5 which has a lot of performance improvements we want esp. the incremental build support. This will make us faster at every feature we build and bug we fix. We're not sure by how much yet, but it could be 5x or more

The main difficulty of this work is that we use internal interfaces in the typescript compiler that are not supported and they have changed non-trivially since ts 2.6.

TODO:
I've taken a stab at updating our code that uses the ts internals, but we're getting errors in our sts compiler when we try to compile user code still. I'm hoping more eyeballs helps on this problem.

This binary expression:
`r += (arr[i] === undefined || arr[i] === null) ? "" : arr[i]`
from line 78 of pxt-helpers.ts is triggering an exception in our compiler because some type "t" doesn't have the ".symbol" property set.
> INTERNAL ERROR: TypeError: Cannot read property 'flags' of undefined
    at isClassType (/Users/darzu/mc/pxt/built/pxt.js:148924:75)
    at checkType (/Users/darzu/mc/pxt/built/pxt.js:148969:21)
    at typeOf (/Users/darzu/mc/pxt/built/pxt.js:149010:20)
    at emitBinaryExpression (/Users/darzu/mc/pxt/built/pxt.js:151698:26)
    at emitExprCore (/Users/darzu/mc/pxt/built/pxt.js:152347:32)


darzu:
@mmoskal if you feel inclined, I wonder if you could help me navigate the dependencies we have on the typescript internals. The issue is that since we were depending on unsupported APIs and those APIs have changed, we need to rebuild around their new internals. I took a stab in this PR but now it seems some invariants that were true aren't anymore, e.g. the symbol error above.

darzu:
deprecated by: https://github.com/microsoft/pxt/pull/6081

PR #5651, opened 2019-06-07
Only show hint 3 times per tutorial, reduce time to 15s

shakao:
https://makecode.microbit.org/app/156d9cd42589988448c73fafc01a65ef5b3e3076-353a6199b4


shakao:
Redoing without cookie logic

PR #5652, opened 2019-06-07
Only show hint 3 times per tutorial, reduce time to 15s

shakao:
https://makecode.microbit.org/app/68edffa7101138292347819e968e46c0a0d3ce11-c12d57ddaf

### DIFF
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -769,6 +769,7 @@ declare namespace pxt.tutorial {
         tutorialStepInfo?: pxt.tutorial.TutorialStepInfo[];
         tutorialStep?: number; // current tutorial page
         tutorialReady?: boolean; // current tutorial page
+        tutorialHintCounter?: number // count for number of times hint has been shown
         tutorialStepExpanded?: boolean; // display full step in dialog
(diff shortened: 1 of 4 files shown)


localtypings/pxtarget.d.ts
```diff
@@ -769,6 +769,7 @@ declare namespace pxt.tutorial {
         tutorialStepInfo?: pxt.tutorial.TutorialStepInfo[];
         tutorialStep?: number; // current tutorial page
         tutorialReady?: boolean; // current tutorial page
+        tutorialHintCounter?: number // count for number of times hint has been shown
```
pelikhan: this flag gets serialized in the header... you should probably clear it when loading the script. or store it somewhere else. I think the limit should be per "editor session".

webapp/src/hinttooltip.tsx
```diff
@@ -48,9 +49,12 @@ export class HintManager {
     }
 
     // starts a timer, overwriting current timer
-    public pokeUserActivity(id: string) {
-        this.stopPokeUserActivity();
-        this.timer = this.hints[id]();
+    // TODO: if/when we add more hints, should discuss whether this count is across all hints or per-hint
```
pelikhan: one to rule them all

PR #5656, opened 2019-06-10
don't show close icon in tutorial modals

pelikhan:
image


pelikhan:
Fix for https://github.com/microsoft/pxt-microbit/issues/2171

PR #5659, opened 2019-06-10
Add check to see if frameElement exists

ChaseMor:
Adds a check to see if window has access to same origin features like frameElement before trying to use them.

Fixes https://github.com/microsoft/pxt-microbit/issues/2152


pelikhan:
You need support the "old way" otherwise you will break all existing embedded pages. 
https://github.com/microsoft/pxt/commit/209a658b556522c613c8432d713567744a451b56#r33881867

PR #5661, opened 2019-06-10
Sprite wizard

JacobBovee:
Currently the Sprite Wizard takes a static json config and builds a questionnaire with each page having a single question. The answer to the inputs are stored and replace the tokens in the output. When the user clicks Done the output is compiled into blocks and added to the DOM (see attached gif for current functionality). From here I'd like to build in a number of configurable inputs, some could be based on the answers type (Dropdown for enums, slider for numbers) while others could be defined as custom inputs such as the sprite editor for creating an image. The output is generic markdown with the idea that it can be expanded to accept different forms of output such as text. 
I'd like it to to become a "snippet builder" accepting generic config files and generating a modal with inputs and output based on it. This could potentially be used to create new blocks, or even administer quizzes if powerful enough.

8UOm5gsWys


abchatra:
Thanks @JacobBovee  Just a heads up. Merge of this needs to wait till we take a stable branch from master. (Likely tomorrow or day after)

webapp/src/snippetBuilder.tsx
```diff
@@ -0,0 +1,339 @@
...
+        let tokenizedOutput = output;
+        const tokens = Object.keys(defaults);
+
+        // Replaces output tokens with answer if available or default value
+        for (let token of tokens) {
+            if (answers[token]) {
+                tokenizedOutput = tokenizedOutput.split(`$${token}`).join(answers[token]);
+            }
+            else {
+                tokenizedOutput = tokenizedOutput.split(`$${token}`).join(defaults[token]);
```
darzu: I think this if statement could become:
```
const value = answers[token] || defaults[token]
tokenizedOutput = tokenizedOutput.split(`$${token}`).join(value);

```

webapp/src/snippetBuilder.tsx
```diff
@@ -0,0 +1,339 @@
...
+            });
+    }
+
+    confirm() {
+        this.injectBlocksToWorkspace();
+        this.hide();
+    }
+
+    changePage(increment: 1 | -1) {
+        const { currentQuestion } = this.state;
```
darzu: might be worth adding a tick here because we often see >80% of people not getting past first page for things like this

webapp/src/snippetBuilder.tsx
```diff
@@ -0,0 +1,339 @@
...
+
+    nextPage() {
+        const { config } = this.state;
+        const { currentQuestion, output } = this.state;
+        const nextQuestion = config.questions[currentQuestion + 1];
+        // If output exists
+        if (nextQuestion.output) {
+            // If output is not already appended
+            if (output.indexOf(nextQuestion.output) === -1) {
+                this.setState({ output: `${output}\n${nextQuestion.output}`});
```
darzu: I suspect you might actually want to keep each question output separate and then only concatenate them at the very end. You could probably do this by making "output" a string[] instead of a string.
How do you handle revisiting old questions if it's all been combined into one output string?

webapp/src/snippetBuilder.tsx
```diff
@@ -0,0 +1,339 @@
...
+                Blockly.Xml.appendDomToWorkspace(Blockly.Xml.textToDom(resp), mainWorkspace);
+            });
+    }
+
+    confirm() {
+        this.injectBlocksToWorkspace();
+        this.hide();
+    }
+
+    changePage(increment: 1 | -1) {
```
darzu: Good use of union type :)

webapp/src/snippetBuilder.tsx
```diff
@@ -0,0 +1,339 @@
...
+            </sui.Modal>
+        )
+    }
+}
+
+// This will be passed down as a prop
+const staticConfig: any = {
+    name: "Sprite Builder",
+    outputType: 'blocks',
+    output: `enum SpriteKind {
```
darzu: maybe this should be called "initialOutput" ? Because the final output will be a combination of this plus output from each question

pelikhan:
It would be helpful to have a build of the editor uploaded -- or a screen recorded gif of the changes.

webapp/src/snippetBuilder.tsx
```diff
@@ -0,0 +1,339 @@
...
+                                "value": "SpriteKind.Enemy",
+                                "label": "Enemy"
+                            }
+                        ]
+                    }
+                }
+            ],
+            "output": "$spriteName.setKind($spriteKind)",
+            "goTo": {
+                "4": { "spriteKind": 0 },
```
darzu: typically in json schemas like this we try to avoid putting values in the "key" slot out of convention.
Maybe you could do:
```
"goTo": {
  "question": 4,
  "parameters": {
     "spriteKind": 0
  }
}
```

webapp/src/snippetBuilder.tsx
```diff
@@ -0,0 +1,339 @@
...
+    questions: [
+        {
+            "title": "What should your sprite be called?",
+            "inputs": [{
+                    "answerToken": "spriteName",
+                    "defaultAnswer": "mySprite",
+                    "type": "text"
+            }],
+            "output": "",
+            "goTo": 2
```
darzu: nitpick (feel free to ignore nitpick comments): typically goto is lowercase, I'd just use "goto"

webapp/src/blocks.tsx
```diff
@@ -1122,6 +1136,23 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
+        // Inject "Create a Sprite..." button
+        if (ns == "sprites") {
+            // TODO(dz):
+            extraBlocks.push({
+                name: `SPRITE_WIZARD_BUTTON`,
+                type: "button",
+                attributes: {
+                    blockId: `SPRITE_WIZARD_BUTTON`,
+                    label: "Create a Sprite...",
+                    weight: 101,
```
darzu: Can we add a comment about the weight and why it's this value? Otherwise 101 just looks like a magic number :)

darzu:
> Basic functionality of the sprite wizard. Will need to move the config to a more permanent location.

Maybe expand this description a bit. You could describe the goals of the Sprite Wizard & Snippet Builder and the choice of creating a more general SnippetBuilder abstraction and how you plan to eventually separate Sprite Wizard into the target repo (pxt-arcade). Also maybe briefly describe what already works in this PR.

webapp/src/snippetBuilder.tsx
```diff
@@ -0,0 +1,339 @@
...
+    output?: string;
+    projectView?: pxt.editor.IProjectView;
+    answers?: any; // Will be typed once more clearly defined
+    currentQuestion: number;
+    defaults: any; // Will be typed once more clearly defined
+    mainWorkspace?: Blockly.Workspace;
+    config?: any; // Will be a config type
+}
+
+export class CreateSnippetBuilder extends data.Component<ISettingsProps, CreateSnippetBuilderState> {
```
darzu: Should this be called just "SnippetBuilder" ?

Also, would you mind adding a descriptive comment explaining the purpose of this react component? That'll help future developers. You'll notice that our code base doesn't have a lot of descriptive comments like this but we're trying to slowly get the code base into a more readable state so that also means new code should have some top level comments.

darzu:
> It would be helpful to have a build of the editor uploaded -- or a screen recorded gif of the changes.

@JacobBovee do you know how to record gifs on your mac? If not stop by sometime and I can show you :) Or you can probably figure it out: https://giphy.com/apps/giphycapture

Also, next time you stop by, remind me to show you how to upload a build for others to try

ChaseMor:
[Screen To Gif](https://www.screentogif.com/) is also a great tool for recording and editing .gifs

webapp/src/blocks.tsx
```diff
@@ -1108,6 +1110,18 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         return this.filterBlocks(subns, this.nsMap[ns]) || [];
     }
 
+    private openSpriteWizard() {
+        Promise.resolve()
+        .delay(10)
```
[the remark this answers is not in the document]
darzu: For reference, here's the discussion around the delay in the Make a Function code: https://github.com/microsoft/pxt/pull/5164/files#r247268459

webapp/src/snippetBuilder.tsx
JacobBovee: If a question is being revised I have to make sure that the tokens are replaced everywhere the answer correlates to a token. Since they're unique in nature running it over the entire output string is necessary. This could definitely be turned into an array and mapped over if you think it's better. It will have to be combined before being passed to MarkedContent regardless. Otherwise the first item in output[] would be the initial output, the second is the initial output and the new output, and so on.

webapp/src/blocks.tsx
```diff
@@ -1122,6 +1136,23 @@ export class Editor extends toolboxeditor.ToolboxEditor {
                 blockXml: `<block type="pxt-on-start"></block>`
             });
         }
+        // Inject "Create a Sprite..." button
+        if (ns == "sprites") {
```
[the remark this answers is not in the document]
pelikhan: we have some logic to inject buttons loaded from "extension extensions". this needs to be removed and loaded from configurable data.

webapp/src/blocks.tsx
```diff
@@ -1108,6 +1110,14 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         return this.filterBlocks(subns, this.nsMap[ns]) || [];
     }
 
+    private openSpriteWizard() {
```
pelikhan: avoid referring to "SpriteWizard" since you're building a general purpose wizard infrastructure.

webapp/src/snippetBuilder.tsx
```diff
@@ -0,0 +1,398 @@
...
+import * as sui from "./sui";
+import * as md from "./marked";
+import * as compiler from './compiler';
+
+type ISettingsProps = pxt.editor.ISettingsProps;
+
+export interface SnippetBuilderState {
+    visible?: boolean;
+    tsOutput?: string;
+    projectView?: pxt.editor.IProjectView;
```
pelikhan: does not belong to the state

webapp/src/snippetBuilder.tsx
```diff
@@ -0,0 +1,398 @@
...
+    answers?: any; // Will be typed once more clearly defined
+    currentQuestion: number;
+    defaults: any; // Will be typed once more clearly defined
+    mainWorkspace?: Blockly.Workspace;
```
pelikhan: does not belong to the state

webapp/src/snippetBuilder.tsx
```diff
@@ -0,0 +1,398 @@
...
+import * as md from "./marked";
+import * as compiler from './compiler';
+
+type ISettingsProps = pxt.editor.ISettingsProps;
+
+export interface SnippetBuilderState {
+    visible?: boolean;
+    tsOutput?: string;
+    projectView?: pxt.editor.IProjectView;
+    answers?: any; // Will be typed once more clearly defined
```
pelikhan: avoid any even in early phases of the projects. it defeats the benefits of typescript. define interfaces and evolve them as you build the solution.

webapp/src/snippetBuilder.tsx
darzu: this doesn't just capture the final output, but also the intermediate output as you're working through questions. How else should this be stored?

webapp/src/snippetBuilder.tsx
pelikhan: do it in 1 pass: ``tokenizedOutput.replace(/\$(\w+)/g, (m, token) => answers[token] || defaults[token])``

webapp/src/snippetBuilder.tsx
```diff
@@ -0,0 +1,398 @@
...
+        });
+    }
+
+    /**
+     * 
+     * @param projectView - used to access what would traditionally be in the parent prop
+     * @param mainWorkspace  - used to append the final xml to the DOM
+     */
+    show(projectView: pxt.editor.IProjectView, mainWorkspace: Blockly.Workspace) {
+        pxt.tickEvent('snippetBuilder.show', null, { interactiveConsent: false });
```
pelikhan: isn't the sprite builder opened via a button click? in that case, interactiveConsent is true.

webapp/src/snippetBuilder.tsx
```diff
@@ -101,59 +134,86 @@ export class CreateSnippetBuilder extends data.Component<ISettingsProps, CreateS
...
+            if (nextQuestion.output && tsOutput.indexOf(nextQuestion.output) === -1) {
+                this.setState({ tsOutput: `${tsOutput}\n${nextQuestion.output}`});
             }
             // Change page to page + 1
             this.changePage(1);
         }
     }
 
+    /**
+     * Calls changePage with a -1 decrementor
```
darzu: It's great that you added comments but some of them might be too much :) This one for example doesn't describe anything that's not obvious in the code and it's also too specific to the code. Generally the comments shouldn't prescribe an implementation but rather tell the goals and the usage patterns.

webapp/src/snippetBuilder.tsx
```diff
@@ -0,0 +1,398 @@
...
+     * Takes the output from state, runs replace tokens, decompiles the resulting typescript
+     * and outputs the result as a Blockly xmlDOM. This then uses appendDomToWorkspace to attach 
+     * our xmlDOM to the mainWorkspaces passed to the component.
+     */
+    injectBlocksToWorkspace() {
+        const { mainWorkspace, tsOutput } = this.state;
+
+        compiler.getBlocksAsync()
+            .then(blocksInfo => compiler.decompileBlocksSnippetAsync(this.replaceTokens(tsOutput), blocksInfo))
+            .then(resp => {
```
pelikhan: decompile can fail; needs to be handled gracefully.

webapp/src/snippetBuilder.tsx
darzu: not every function needs comments

webapp/src/snippetBuilder.tsx
```diff
@@ -0,0 +1,398 @@
...
+                        {projectView && <md.MarkedContent markdown={this.generateOutputMarkdown(output)} parent={projectView} />}
+                    </div>
+                </div>
+            </sui.Modal>
+        )
+    }
+}
+
+// This will be passed down as a prop but is currently static
+const staticConfig: any = {
```
pelikhan: use a function so that lf() calls a resolve late enough to pick up localization strings. This configuration should come from the target.

webapp/src/snippetBuilder.tsx
```diff
@@ -101,59 +134,86 @@ export class CreateSnippetBuilder extends data.Component<ISettingsProps, CreateS
...
+    /**
+     * Hides modal and injects our blocks to the work space.
+     */
     confirm() {
         this.injectBlocksToWorkspace();
         this.hide();
     }
 
+    /** 
+     * @param increment - this adds either 1 or -1 to the value of currentQuestion
```
darzu: like below, this is too prescriptive. "currentQuestion" is an implementation detail that the reader doesn't need to know about. Omitting implementation specifics helps the reader read faster and it makes it easier for the maintainer since they can change things like how "currentQuestion" works without having to come back and change the comments.
Writing comments is a balancing act between being descriptive enough to be useful while still being vague enough to not be locked to one implementation

webapp/src/snippetBuilder.tsx
pelikhan: It is stored in ISettingsProps as the "parent" field.

darzu:
> Too much "sprite wizard" hard-coding going on. pxt is editor agnostic.

webapp/src/snippetBuilder.tsx
darzu: If you need a good reference on stuff like this, Clean Code is very popular and good:
https://www.amazon.com/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882

webapp/src/snippetBuilder.tsx
JacobBovee: I deleted some of the comments and hopefully am now only describing what could be unclear by reading the code.

webapp/src/snippetBuilder.tsx
JacobBovee: I added a catch that throws an exception and reports the error.

webapp/src/snippetBuilder.tsx
JacobBovee: Updated most of the types to use descriptive interfaces.

webapp/src/snippetBuilder.tsx
JacobBovee: I'm now passing this down as the parent prop.

webapp/src/snippetBuilder.tsx
JacobBovee: I'm now passing this down as the parent prop.

webapp/src/snippetBuilder.tsx
JacobBovee: Yeah I changed the naming to make more sense.

webapp/src/blocks.tsx
JacobBovee: I added a comment to clarify.

webapp/src/snippetBuilder.tsx
JacobBovee: Go to has been changed to lowercase. I've also standardized the goto object to take params.

webapp/src/snippetBuilder.tsx
JacobBovee: Yeah I think that's more clear I updated the format.

webapp/src/blocks.tsx
```diff
@@ -1108,6 +1110,18 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
+        Promise.resolve()
+        .delay(10)
+        .then(() => {
+            if (!this.functionsDialog) {
```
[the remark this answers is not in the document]
JacobBovee: The delay doesn't seem to have any affect. Is it worth updating the functionDialog that is also using this?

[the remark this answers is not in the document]
JacobBovee: Removed the functionsDialog check.

webapp/src/blocks.tsx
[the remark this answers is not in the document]
JacobBovee: The delay doesn't seem to have any affect. Is it worth updating the functionDialog that is also using this?

webapp/src/snippetBuilder.tsx
JacobBovee: Passing this down as a prop now.

JacobBovee:
See here -> https://github.com/microsoft/pxt/pull/5661/commits/17467cd7597ede8b1d9cc5e61aff304f4d4093c8 

in snippetBuilder.tsx for the logic to attach blocks. It's currently hard coded in to attach to on start.

webapp/src/blocks.tsx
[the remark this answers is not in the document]
pelikhan: move as much sprite wizard specific code to the snippetbuilder.tsx

webapp/src/blocks.tsx
```diff
@@ -1108,6 +1110,17 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         return this.filterBlocks(subns, this.nsMap[ns]) || [];
     }
 
+    // TODO(jb): Generalize to snippet builder and pass in specific config.
+    private openSnippetDialog() {
+        if (!this.snippetDialog) {
+            const wrapper = document.body.appendChild(document.createElement('div'));
+            const props = { parent: this.parent, mainWorkspace: this.editor }
+            this.snippetDialog = ReactDOM.render(
```
pelikhan: that's not really how we add new "dialogs" in the react tree.

webapp/src/snippetBuilder.tsx
```diff
@@ -0,0 +1,404 @@
...
+     */
+    injectBlocksToWorkspace() {
+        const { tsOutput } = this.state;
+        const { mainWorkspace } = this.props
+
+        compiler.getBlocksAsync()
+            .then(blocksInfo => compiler.decompileBlocksSnippetAsync(this.replaceTokens(tsOutput), blocksInfo))
+            .then(resp => {
+                // TODO(jb)
+                const xmlDOM = Blockly.Xml.textToDom(resp);
```
pelikhan: this section needs more explanantions

webapp/src/snippetBuilder.tsx
```diff
@@ -0,0 +1,404 @@
...
+                        {parent && <md.MarkedContent markdown={this.generateOutputMarkdown(tsOutput)} parent={parent} />}
+                    </div>
+                </div>
+            </sui.Modal>
+        )
+    }
+}
+// get config
+// This will be passed down as a prop but is currently static
+const staticConfig: ISnippetConfig = {
```
pelikhan: do not preallocate as localization files have not been loaded yet.

webapp/src/snippetBuilder.tsx
JacobBovee: I added a catch so that decompile errors are handled.

webapp/src/blocks.tsx
```diff
@@ -1108,6 +1110,41 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
+                wrapper
+            ) as SnippetBuilder;
+        }
+        this.snippetDialog.show();
+    }
+
+    loadSnippetExtensions(snippetExtensions: pxt.PackageConfig[], ns: string, extraBlocks: (toolbox.BlockDefinition | toolbox.ButtonDefinition)[]) {
+        snippetExtensions.forEach(config => {
+            config.snippetBuilders.forEach(snippet => {
+                if (ns == snippet.namespace) {
```
pelikhan: Use filter

webapp/src/blocks.tsx
```diff
@@ -1108,6 +1110,41 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
+                React.createElement(SnippetBuilder, props),
+                wrapper
+            ) as SnippetBuilder;
+        }
+        this.snippetDialog.show();
+    }
+
+    loadSnippetExtensions(snippetExtensions: pxt.PackageConfig[], ns: string, extraBlocks: (toolbox.BlockDefinition | toolbox.ButtonDefinition)[]) {
+        snippetExtensions.forEach(config => {
+            config.snippetBuilders.forEach(snippet => {
```
pelikhan: snippetBuilders might be undefinex

webapp/src/blocks.tsx
```diff
@@ -1108,6 +1110,41 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
+                if (ns == snippet.namespace) {
+                    extraBlocks.push({
+                        name: `SNIPPET${name}_BUTTON`,
+                        type: "button",
+                        attributes: {
+                            blockId: `SNIPPET${name}_BUTTON`,
+                            label: snippet.label ? Util.rlf(snippet.label) : Util.lf("Editor"),
+                            weight: 101,
+                            group: snippet.group && snippet.group,
```
pelikhan: ?

webapp/src/blocks.tsx
```diff
@@ -1148,6 +1186,17 @@ export class Editor extends toolboxeditor.ToolboxEditor {
                 }
             })
         }
+
+        if (pxt.appTarget.appTheme.spriteWizard) {
```
pelikhan: Move as much logic as possible into snippetBuilder.tsx

webapp/src/blocks.tsx
```diff
@@ -1108,6 +1110,41 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
+            const props = { parent: this.parent, mainWorkspace: this.editor, config };
+            this.snippetDialog = ReactDOM.render(
+                React.createElement(SnippetBuilder, props),
+                wrapper
+            ) as SnippetBuilder;
+        }
+        this.snippetDialog.show();
+    }
+
+    loadSnippetExtensions(snippetExtensions: pxt.PackageConfig[], ns: string, extraBlocks: (toolbox.BlockDefinition | toolbox.ButtonDefinition)[]) {
```
pelikhan: Move to snippeyBuilder.tsx

webapp/src/blocks.tsx
JacobBovee: The only example I could find of a sidebar button opening a dialog was the create function and followed that path. Do you know of a better example somewhere in the codebase?

pelikhan:
We would typically have it in the main react DOM in app.tsx

localtypings/pxtpackage.d.ts
```diff
@@ -161,4 +162,39 @@ declare namespace pxt {
         namespace?: string; // used to construct id
         mimeType: string;
     }
+
+    interface SnippetConfig {
+        name: string;
+        namespace: string;
+        group?: string;
+        label: string;
```
pelikhan: The CLI needs to traverse the config tree and extract the strings that need to be translated into the target-strings.json file. Then you need to use rlf in the UI to localize the strings.

webapp/src/snippetBuilder.tsx
```diff
@@ -0,0 +1,336 @@
...
+}
+
+interface Questions {
+    title: string;
+    output?: string;
+    goto?: GoToOptions;
+    inputs: QuestionInput[];
+}
+
+interface SnippetConfig {
```
pelikhan: Is it the same as in pxtarget.d.ts?

JacobBovee: Yeah I just updated them to use the pxtarget ones.

webapp/src/snippetBuilder.tsx
```diff
@@ -0,0 +1,349 @@
...
+                const xmlOnStartBlock = this.findRootBlock(xmlDOM, 'pxt-on-start');
+                // Finds the on start blocks children
+                const toAttach = this.findRootBlock(xmlOnStartBlock);
+                const rootConnection = Blockly.Xml.domToBlock(toAttach, mainWorkspace);
+                // Connects new blocks to start block
+                this.getOnStartBlock(mainWorkspace)
+                    .getInput("HANDLER").connection.connect(rootConnection.previousConnection);
+            }).catch((e) => {
+                pxt.reportException(e);
+                throw new Error(`Failed to decompile snippet output`);
```
pelikhan: use core.errorNotification(...)

webapp/src/snippetBuilder.tsx
```diff
@@ -0,0 +1,349 @@
...
+    ) as SnippetBuilder;
+    snippetBuilder.show();
+}
+
+export function initializeSnippetExtensions(ns: string, extraBlocks: (toolbox.BlockDefinition | toolbox.ButtonDefinition)[], editor: Blockly.WorkspaceSvg, parent: pxt.editor.IProjectView) {
+    const snippetExtensions = getSnippetExtensions();
+
+    snippetExtensions.forEach(config => {
+        config.snippetBuilders.forEach(snippet => {
+            if (ns == snippet.namespace) {
```
pelikhan: use ``.filter(snippet => snippet.namespace == ns)``

webapp/src/snippetBuilder.tsx
```diff
@@ -0,0 +1,349 @@
...
+                extraBlocks.push({
+                    name: `SNIPPET${name}_BUTTON`,
+                    type: "button",
+                    attributes: {
+                        blockId: `SNIPPET${name}_BUTTON`,
+                        label: snippet.label ? pxt.Util.rlf(snippet.label) : pxt.Util.lf("Editor"),
```
pelikhan: strings need to be injected in target-strings.json when the CLI builds the target

webapp/src/snippetBuilder.tsx
```diff
@@ -328,8 +329,9 @@ export function initializeSnippetExtensions(ns: string, extraBlocks: (toolbox.Bl
     const snippetExtensions = getSnippetExtensions();
 
     snippetExtensions.forEach(config => {
-        config.snippetBuilders.forEach(snippet => {
-            if (ns == snippet.namespace) {
+        config.snippetBuilders
```
pelikhan: can snippetBuilders be undefined?

JacobBovee: getSnippetExtensions filters out anything that doesn't include a snippetBuilders property so it will always be there.

abchatra:
This change should have been merged by squashing. Now history is all messed up

pelikhan:
I forgot to disable that option after merging the dev branch.

PR #5663, opened 2019-06-11
Fix Untranslatable Blockly Strings

sgk:
Make it possible to translate the two newly(?) introduced Blockly strings.

* New variable...
* Rename all '%1' variables to:

image
image


pelikhan:
@abchatra we should take this one

sgk:
Thank you!

PR #5666, opened 2019-06-12
Fix gdb path for docker build with DAL

mmoskal:
this can go on master, either now or later

### DIFF
--- a/cli/gdb.ts
+++ b/cli/gdb.ts
@@ -233,9 +233,8 @@ ${cmds}
 function codalBin() {
     let cs = pxt.appTarget.compileService
 
-    return cs.codalBinary ?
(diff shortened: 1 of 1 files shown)


cli/gdb.ts
```diff
@@ -233,9 +233,8 @@ ${cmds}
 function codalBin() {
     let cs = pxt.appTarget.compileService
 
-    return cs.codalBinary ?
-        buildengine.thisBuild.buildPath + "/build/" + cs.codalBinary :
-        "built/yt/build/" + cs.yottaTarget + "/source/" + cs.yottaBinary.replace(/\.hex$/, "").replace(/-combined$/, "");
+    return buildengine.thisBuild.buildPath + "/build/" + (cs.codalBinary ? cs.codalBinary :
```
pelikhan: you could use "path.join"

PR #5672, opened 2019-06-13
reduce verbosity of breadboard messages

pelikhan:
Removes this benign but annoying message
image


pelikhan:
@abchatra what label do you want for PRs that could make it next week?

PR #5673, opened 2019-06-13
fix single page print

pelikhan:
Print only prints the first page. This fixes it.


pelikhan:
@abchatra we should take this in microbit

PR #5676, opened 2019-06-13
Green screen offline

pelikhan:
Fix for https://github.com/microsoft/pxt-microbit/issues/2198.
-  don't try to use web cams in electron/uwp
- use green screen when no web cams available (including in IE11)


pelikhan:
@abchatra I've added a "microbit v2" label to mark PRs that could be integrated in the microbit v2 stable branch.

PR #5681, opened 2019-06-14
Oops, wrong video for micro:bit release blog

ganicke:
Need the correct video. The previous one was a stub.


ganicke:
@abchatra already got this in https://github.com/microsoft/pxt/commit/365884ebbf4c3ad2fb37532da2afd8dffeca69c2

PR #5684, opened 2019-06-17
Fix ie crash

jwunderl:
Should fix https://github.com/microsoft/pxt-arcade/issues/1061, but I'd love it if anyone could verify that as I don't currently have a windows machine to validate that


pelikhan:
verifying

PR #5686, opened 2019-06-17
Sprite editor fixes

ChaseMor:
Fixes a variety of Sprite Editor bugs

### Preview Not Shown
Fixes https://github.com/microsoft/pxt-arcade/issues/1029
### Marquee Tool Shows Outlines
Selection Outlines
### Marquee Tool can't select bottom or right cells
Fixes https://github.com/microsoft/pxt-arcade/issues/1031
### Can't deselect Marquee Tool if whole screen is selected
If the whole screen is selected, there is currently no way to deselect the selection. This change makes it so a single click, inside or outside of the selection deselects.
### Selection disappears when showing block preview
Selection Disappears
### Top left of the Rectangle Tool starts at click instead of top left of tool when clicked
The top left of the rectangle that is being drawn is set to the center of the preview. This makes the cursor inaccurate. It should be the top left of the rectangle is the top left of the cursor. This way if a user is using a wider tool, all pixels shown colored in cursor are actually painted.
### Circle Tool should show tool width of 1 on preview
The circle tool should show as 1x1 pixel for all tool width until other widths are supported.

### DIFF
--- a/pxtlib/sprite-editor/canvasGrid.ts
+++ b/pxtlib/sprite-editor/canvasGrid.ts
@@ -56,7 +56,7 @@ namespace pxtsprite {
         }
 
         repaint(): void {
-            this.clearContext(this.paintLayer.getContext("2d"));
+            this.clearContext(this.context);
             this.drawImage();
             if (this.state.floatingLayer) this.drawFloatingLayer();
             else this.hideOverlay();
@@ -345,8 +345,7 @@ namespace pxtsprite {
             if (!this.state.floatingLayer) {
                 return;
             }
-            const context = this.paintLayer.getContext("2d");
-            this.drawImage(this.state.floatingLayer, context, this.state.layerOffsetX, this.state.layerOffsetY, true);
+            this.drawImage(this.state.floatingLayer, this.context, this.state.layerOffsetX, this.state.layerOffsetY, true);
 
             this.drawSelectionAnimation();
         }
@@ -378,6 +377,7 @@ namespace pxtsprite {
         }
 
         private clearContext(context: CanvasRenderingContext2D) {
+            // Paint Layer has the same dimensions as all other contexts
             context.clearRect(0, 0, this.paintLayer.width, this.paintLayer.height);
         }
 
--- a/pxtlib/sprite-editor/canvasGrid.ts
+++ b/pxtlib/sprite-editor/canvasGrid.ts
@@ -71,6 +71,11 @@ namespace pxtsprite {
 
             if (cursor) {
                 this.repaint();
+                if (edit.showPreview) {
+                    edit.drawCursor(col, row, (c, r) => {
+                        this.drawColor(c, r, edit.color);
+                    });
+                }
                 this.context.strokeStyle = "#898989";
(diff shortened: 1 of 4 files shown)


pxtlib/sprite-editor/canvasGrid.ts
```diff
@@ -56,6 +56,7 @@ namespace pxtsprite {
         }
 
         repaint(): void {
+            this.clearContext(this.paintLayer.getContext("2d"));
```
riknoll: Use `this.context` instead of calling `getContext()`

pxtlib/sprite-editor/canvasGrid.ts
```diff
@@ -104,10 +110,7 @@ namespace pxtsprite {
                 context.fillStyle = this.palette[color - 1];
                 context.fillRect(x, y, this.cellWidth, this.cellHeight);
             }
-            else if (transparency) {
-                context.clearRect(x, y, this.cellWidth, this.cellHeight);
```
riknoll: Why is this removed?

pxtlib/sprite-editor/canvasGrid.ts
```diff
@@ -333,34 +341,46 @@ namespace pxtsprite {
                 this.mouseRow
             ];
         }
+        protected drawFloatingLayer() {
+            if (!this.state.floatingLayer) {
+                return;
+            }
+            const context = this.paintLayer.getContext("2d");
```
riknoll: `this.context`

pxtlib/sprite-editor/canvasGrid.ts
```diff
@@ -333,34 +341,46 @@ namespace pxtsprite {
...
-                    requestAnimationFrame(() => this.drawFloatingLayer(dashOffset));
+                    requestAnimationFrame(() => this.drawSelectionAnimation(dashOffset));
                 };
 
                 this.selectAnimation = setInterval(drawLayer, 40)
             }
         }
 
+        private clearContext(context: CanvasRenderingContext2D) {
+            context.clearRect(0, 0, this.paintLayer.width, this.paintLayer.height);
```
riknoll: Add a comment that all three canvases are the same dimensions; (i.e. same as paintLayer)

pxtlib/sprite-editor/tools.ts
```diff
@@ -188,35 +217,65 @@ namespace pxtsprite {
      * Tool for drawing empty rectangles
      */
     export class OutlineEdit extends SelectionEdit {
+        showPreview = true;
+
         protected doEditCore(state: CanvasState) {
-            const tl = this.topLeft();
-            const br = this.bottomRight();
+            let tl = this.topLeft();
+            tl.x -= Math.floor(this.toolWidth / 2);
```
riknoll: Nitpick: `this.toolWidth >> 1` gets rid of the call to `Math.floor`

pxtlib/sprite-editor/canvasGrid.ts
ChaseMor: Since the floating layer is drawn onto the same layer as as the paint layer, if it has transparency, this code will remove the color of the painted layer below it. Changed it so that it clears the whole layer before painting anything instead of clearing on a pixel by pixel basis.

PR #5687, opened 2019-06-17
Add list of unsupported browsers to pxttarget json

shakao:
Not styled, just using our current redirect to '/browsers' functionality

### DIFF
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -327,7 +327,7 @@ declare namespace pxt {
         qrCode?: boolean; // generate QR code for shared urls
         importExtensionFiles?: boolean; // import extensions from files
         debugExtensionCode?: boolean; // debug extension and libs code in the Monaco debugger
-        browserBlacklist?: string[] // blacklist of browsers for a specific target (eg IE11 in arcade). check browserutils.js browser() function for strings
+        unsupportedBrowsers?: string[] // list of unsupported browsers for a specific target (eg IE11 in arcade). check browserutils.js browser() function for strings
     }
 
     interface SocialOptions {
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -49,7 +49,11 @@ declare namespace pxt {
         ignoreDocsErrors?: boolean;
         variants?: Map<AppTarget>; // patches on top of the current AppTarget for different chip variants
         queryVariants?: Map<AppTarget>; // patches on top of the current AppTarget using query url regex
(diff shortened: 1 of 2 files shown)


localtypings/pxtarget.d.ts
```diff
@@ -327,6 +327,7 @@ declare namespace pxt {
         qrCode?: boolean; // generate QR code for shared urls
         importExtensionFiles?: boolean; // import extensions from files
         debugExtensionCode?: boolean; // debug extension and libs code in the Monaco debugger
+        unsupportedBrowsers?: string[] // list of unsupported browsers for a specific target (eg IE11 in arcade). check browserutils.js browser() function for strings
```
pelikhan: Move this into the apptarget - it needs to be available in all pages and we delete theme in some cases (see targetlight.js)

pxtlib/browserutils.ts
```diff
@@ -218,6 +218,11 @@ namespace pxt.BrowserUtils {
         if (/bot|crawler|spider|crawling/i.test(navigator.userAgent))
             return true;
 
+        //Check target theme to see if this browser is blacklisted
+        if (pxt.appTarget.appTheme.unsupportedBrowsers && pxt.appTarget.appTheme.unsupportedBrowsers.indexOf(browser()) >= 0) {
```
pelikhan: see comment above.

localtypings/pxtarget.d.ts
```diff
@@ -49,6 +49,7 @@ declare namespace pxt {
         ignoreDocsErrors?: boolean;
         variants?: Map<AppTarget>; // patches on top of the current AppTarget for different chip variants
         queryVariants?: Map<AppTarget>; // patches on top of the current AppTarget using query url regex
+        unsupportedBrowsers?: string[] // list of unsupported browsers for a specific target (eg IE11 in arcade). check browserutils.js browser() function for strings
```
pelikhan: i think we should specify a combo of browser + version (for IE, version == "*"), but we might have to drop older versions of other editors. Maybe change the format to something like ``unsupportedBrowsers?: { id: string;  }`` , this would allow to add a ``minVersion?: string`` field in the future.

pelikhan: or use a map, 
```{
   ie11: "*"
}```

Comment (shakao) [reply localtypings/pxtarget.d.ts]: yeah, that makes sense. i think we could do minVersion as a number and if no minVersion is specified, it's treated as "*"

File: pxtlib/browserutils.ts [hunk 4]
```diff
@@ -218,6 +218,11 @@ namespace pxt.BrowserUtils {
         if (/bot|crawler|spider|crawling/i.test(navigator.userAgent))
             return true;
 
+        //Check target theme to see if this browser is blacklisted
```
pelikhan: blacklisted -> unsupported

PR #5689, opened 2019-06-18
don't show hints for package snippets in tutorials

jwunderl:
fixes https://github.com/microsoft/pxt-arcade/issues/1071

Package snippets are used to load the extensions for the other snippets on the page, and shouldn't be shown as hints

### DIFF
--- a/pxtlib/tutorial.ts
+++ b/pxtlib/tutorial.ts
@@ -53,7 +53,7 @@ namespace pxt.tutorial {
     }
 
     function parseTutorialSteps(tutorialmd: string): TutorialStepInfo[] {
-        const hiddenSnippetRegex = /```(filterblocks|package)\s*\n([\s\S]*?)\n```/gmi;
+        const hiddenSnippetRegex = /```(filterblocks|package|ghost|config)\s*\n([\s\S]*?)\n```/gmi;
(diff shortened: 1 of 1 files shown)


pxtlib/tutorial.ts
```diff
@@ -53,7 +53,7 @@ namespace pxt.tutorial {
     }
 
     function parseTutorialSteps(tutorialmd: string): TutorialStepInfo[] {
-        const filterblocksRegex = /```(filterblocks)\s*\n([\s\S]*?)\n```/gmi;
+        const hiddenSnippetRegex = /```(filterblocks|package)\s*\n([\s\S]*?)\n```/gmi;
```
pelikhan: There is a number of those. See tutorial css

pelikhan:
https://github.com/microsoft/pxt/blob/7e79278244f5ee37a7501c4023ccd616785e5cc6/webapp/public/docs.html#L48

pxtlib/tutorial.ts
jwunderl: so I'd want to add config and ghost here from the link below to this regex, correct? Looks like that's all the ones I see to avoid in https://makecode.com/writing-docs/macros

PR #5691, opened 2019-06-18
Clean for loop decompilation with numeric literal upper bounds

jwunderl:
fixes https://github.com/microsoft/pxt-arcade/issues/1034

Doesn't currently handle prefix unary expressions; can add the handling for that, but it seems really niche to add special handling for - negatives will always basically result in a no-op anyway, and having a unary + in front of a numeric literal is not something that I think I've ever seen :)

current behavior: test decompiles to these blocks:

arcade-screenshot (10)

new: test decompiles to these blocks:

arcade-screenshot (6)

### DIFF
--- a/pxtcompiler/emitter/decompiler.ts
+++ b/pxtcompiler/emitter/decompiler.ts
@@ -1474,15 +1474,16 @@ ${output}</xml>`;
                 r.inputs = [getDraggableVariableBlock("VAR", renamed)];
 
                 if (condition.operatorToken.kind === SK.LessThanToken) {
-                    if (condition.right.kind === SK.NumericLiteral) {
-                        const decrementedValue = parseFloat(condition.right.getFullText()) - 1;
+                    const unwrappedRightSide = unwrapNode(condition.right);
+                    if (unwrappedRightSide.kind === SK.NumericLiteral) {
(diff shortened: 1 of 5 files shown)


pxtcompiler/emitter/decompiler.ts
```diff
@@ -1474,13 +1474,20 @@ ${output}</xml>`;
...
-                    const ex = mkExpr("math_arithmetic");
-                    ex.fields = [getField("OP", "MINUS")];
-                    ex.inputs = [
-                        getValue("A", condition.right, numberType),
-                        getValue("B", 1, numberType)
-                    ];
-                    r.inputs.push(mkValue("TO", ex, wholeNumberType));
+                    const unwrappedRightSide = unwrapNode(condition.right);
+                    if (unwrappedRightSide.kind === SK.NumericLiteral) {
+                        const decrementedValue = parseFloat(unwrappedRightSide.getFullText()) - 1;
```
jwunderl: I've only ever used ``getFullText`` when debugging other than this - is there a helper or better way to handle this sort of thing that I'm forgetting?

riknoll: Just use `unwrappedRightSide.text` (you'll need to cast to a `LiteralExpression). I think `getFullText()` will include comment trivia.

PR #5693, opened 2019-06-19
Fix for IE and Edge Embed

ChaseMor:
IE and Edge can throw an exception when checking if a frameElement exists. This adds a check to see if that exception is thrown.

Fixes https://github.com/microsoft/pxt-microbit/issues/2218

### DIFF
--- a/webapp/public/run.html
+++ b/webapp/public/run.html
@@ -114,8 +114,9 @@
         var deps = /deps(?:[:=])([^&?]+)/i.exec(window.location.href);
 
         var codeFromSrc = /code(?:[:=])([^&?]+)/i.exec(window.location.href);
+        var codeFromData = undefined;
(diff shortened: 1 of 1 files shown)


webapp/public/run.html
```diff
@@ -114,7 +114,14 @@
         var deps = /deps(?:[:=])([^&?]+)/i.exec(window.location.href);
 
         var codeFromSrc = /code(?:[:=])([^&?]+)/i.exec(window.location.href);
-        var codeFromData = window.frameElement ? window.frameElement.getAttribute('data-code') : undefined;
+        try {
+            var codeFromData = window.frameElement ? window.frameElement.getAttribute('data-code') : undefined;
```
pelikhan: Move variable declaration outside try

PR #5695, opened 2019-06-19
fix electron rendering of license

pelikhan:
Use boardname for rendering "micro:bit"


pelikhan:
image

pelikhan:
@abchatra this can be integrated into the micro:bit branch

PR #5696, opened 2019-06-19
Adds tick events for tracking deploy completion time and success/fail stats

darzu:
We're seeing some difficult to repro issues around deployment like this: https://github.com/microsoft/pxt-microbit/issues/2179

This PR adds tick events to help us understand how widespread the issue(s) are and hopefully we can compare failures to determine likely causes.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -2061,23 +2061,27 @@ export class ProjectView
                 }
                 pxt.tickEvent("deploy.start")
                 return pxt.commands.deployAsync(resp, {
-                    reportDeviceNotFoundAsync: (docPath, compileResult) => this.showDeviceNotFoundDialogAsync(docPath, compileResult),
-                    reportError: (e) => core.errorNotification(e),
+                    reportDeviceNotFoundAsync: (docPath, compileResult) => {
+                        pxt.tickEvent("deploy.devicenotfound")
+                        return this.showDeviceNotFoundDialogAsync(docPath, compileResult)
+                    },
+                    reportError: (e) => {
+                        pxt.tickEvent("deploy.reporterror")
+                        return core.errorNotification(e)
+                    },
                     showNotification: (msg) => core.infoNotification(msg)
                 })
                     .then(() => {
                         pxt.tickEvent("deploy.finished")
-                    }, () => {
-                        pxt.tickEvent("deploy.unfinished")
                     })
                     .catch(e => {
-                        pxt.tickEvent("deploy.exception")
                         if (e.notifyUser) {
                             core.warningNotification(e.message);
                         } else {
                             const errorText = pxt.appTarget.appTheme.useUploadMessage ? lf("Upload failed, please try again.") : lf("Download failed, please try again.");
                             core.warningNotification(errorText);
                         }
+                        pxt.tickEvent("deploy.exception", { "notifyUser": e.notifyUser })
                         pxt.reportException(e);
                         if (userContextWindow)
                             try { userContextWindow.close() } catch (e) { }
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -2059,12 +2059,19 @@ export class ProjectView
                         });
                     }
                 }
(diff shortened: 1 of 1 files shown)


webapp/src/app.tsx
```diff
@@ -2059,12 +2059,19 @@ export class ProjectView
...
+                pxt.tickEvent("deploy.start")
                 return pxt.commands.deployAsync(resp, {
                     reportDeviceNotFoundAsync: (docPath, compileResult) => this.showDeviceNotFoundDialogAsync(docPath, compileResult),
                     reportError: (e) => core.errorNotification(e),
                     showNotification: (msg) => core.infoNotification(msg)
                 })
+                    .then(() => {
+                        pxt.tickEvent("deploy.finished")
+                    }, () => {
+                        pxt.tickEvent("deploy.unfinished")
```
riknoll: Won't this prevent the catch clause below from ever executing?

webapp/src/app.tsx
```diff
@@ -2059,12 +2059,19 @@ export class ProjectView
                         });
                     }
                 }
+                pxt.tickEvent("deploy.start")
                 return pxt.commands.deployAsync(resp, {
                     reportDeviceNotFoundAsync: (docPath, compileResult) => this.showDeviceNotFoundDialogAsync(docPath, compileResult),
                     reportError: (e) => core.errorNotification(e),
```
pelikhan: add ticks here too

webapp/src/app.tsx
```diff
@@ -2059,12 +2059,19 @@ export class ProjectView
                         });
                     }
                 }
+                pxt.tickEvent("deploy.start")
                 return pxt.commands.deployAsync(resp, {
                     reportDeviceNotFoundAsync: (docPath, compileResult) => this.showDeviceNotFoundDialogAsync(docPath, compileResult),
```
pelikhan: add ticks here too

webapp/src/app.tsx
```diff
@@ -2059,12 +2059,19 @@ export class ProjectView
...
                     reportError: (e) => core.errorNotification(e),
                     showNotification: (msg) => core.infoNotification(msg)
                 })
+                    .then(() => {
+                        pxt.tickEvent("deploy.finished")
+                    }, () => {
+                        pxt.tickEvent("deploy.unfinished")
+                    })
                     .catch(e => {
+                        pxt.tickEvent("deploy.exception")
```
pelikhan: store whether "notifyUser" is set or not

pelikhan: maybe use pxt.reportError so that we get a stacktrace as well.

webapp/src/app.tsx
darzu: Does it? Shoot. I'll investigate

darzu: Yup, you're right it definitely swallows all rejections. And catch() doesn't just get called for exceptions but any rejection so I can just remove the onreject logic here.

webapp/src/app.tsx
darzu: We use pxt.reportException below. I added the tick so it's easier to see the matching pairs of (deploy.started, deploy.finished) and (deploy.started, deploy.exception). Maybe it's my unfamiliarity with Kusto but I'm not sure how I'd correlate exceptions into (deploy.started, <some exception>) because the exceptions can be anything and don't have a prefix.

I moved the tickEvent to be right above the reportException so it's obvious what's happening.

webapp/src/app.tsx
```diff
@@ -2059,18 +2059,29 @@ export class ProjectView
...
+                        return this.showDeviceNotFoundDialogAsync(docPath, compileResult)
+                    },
+                    reportError: (e) => {
+                        pxt.tickEvent("deploy.reporterror")
+                        return core.errorNotification(e)
+                    },
                     showNotification: (msg) => core.infoNotification(msg)
                 })
+                    .then(() => {
+                        pxt.tickEvent("deploy.finished")
```
pelikhan: you could actually also store the duration of deploy -- just as a nice side data.

PR #5701, opened 2019-06-20
Multi line descriptions for code cards

jwunderl:
closes https://github.com/microsoft/pxt-arcade/issues/371

Allows multi line descriptions using arrays (while still allowing single line versions)

in this gif, most have been 'updated' to use arrays, besides the `which button` tutorial

~Starting as a draft to make sure this seems like a reasonable thing to implement first, I haven't thoroughly tested this, and code cards show up all over the place so I'd really want too. (also would probably want some extra linting as there's a bit of duplication in the code)~ tested all the places I see the modified code used and it seems good

### DIFF
--- a/docs/targets/home-screen.md
+++ b/docs/targets/home-screen.md
@@ -55,7 +55,25 @@ as a JavaScript array of ``CodeCard`` instance.
 ```
 ````
 
-The ``name``, ``description``, ``url``, ``imageUrl`` are mandatory. They are used to display the card in the home screen. A page with a URL that starts with ``https://`` will automatically open in a new tab.
+The ``name``, ``description``, ``url``, ``imageUrl`` are mandatory. They are used to display the card in the home screen. A page with a URL that starts with ``https://`` will automatically open in a new tab. If you want to include a multiple line description, you can
+use an array of strings instead of a single string:
+
+```codecard
+[
+{
+  "name": "Ping Pong",
+  "description": [
+      "A game of bouncing ball",
+      "That you can play"
+  ],
+  "url":"/examples/ping-pong",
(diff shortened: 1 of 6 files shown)


pelikhan:
we not simply use newline characters?

jwunderl:
New line characters don't currently work, but I can switch to support that I suppose. I went with arrays just because it's easier to read / write for me than a bunch of \n's on a long line, and doesn't seem to be any harder to support

docs/targets/home-screen.md
```diff
@@ -55,7 +55,25 @@ as a JavaScript array of ``CodeCard`` instance.
 ```
 ````
 
-The ``name``, ``description``, ``url``, ``imageUrl`` are mandatory. They are used to display the card in the home screen. A page with a URL that starts with ``https://`` will automatically open in a new tab.
+The ``name``, ``description``, ``url``, ``imageUrl`` are mandatory. They are used to display the card in the home screen. A page with a URL that starts with ``https://`` will automatically open in a new tab. If you want to have multiple lines in the description, you can
+use an array of strings instead of a single string. The multiple lines are combined into a single paragraph:
```
ganicke: Single paragraph - T or F?

jwunderl: Each string in the array would be a separate paragraph

ganicke: Multiple sentences per string or are the truncated at the first period?

pelikhan: Please just use \n

ganicke: Looks like you grab just the first element of the split.

jwunderl: Okay, fixed to just be support for `\n` rather than ``string | string[]``

PR #5702, opened 2019-06-20
Update IE11 as not supported for MakeCode Arcade

abchatra:


pelikhan:
@ganicke we should refactor this page to support per editor updates. Good enough for now though.

PR #5703, opened 2019-06-20
base type check off of inferred type rather than explicit typing

jwunderl:
fixes https://github.com/microsoft/pxt-arcade/issues/1091

turns out ``callinfo.decl.type`` only includes explicit types (``: boolean``) and shouldn't be used directly, so this was requiring the type to be explicitly specified to decompile. Using the inferred types instead fixes the issue.

results of test:

arcade-screenshot (12)


jwunderl:
@abchatra this bug does not repro with the blocks in microbit, as all of the boolean blocks there have explicit types set: https://makecode.microbit.org/_5bfdidEJFVoz has all the default ones  and decompiles without gray blocks. Easy to cherry pick there in case it affects any extensions, though.

PR #5715, opened 2019-06-26
Implement friendly snippet auto-complete in Python

darzu:
This PR adds "nice" auto complete.
Before (py):

After (py):

Typescript today:

## Edit Amendments
This PR adds a feature to Monaco I (dz) call "Edit Amendments".
EditAmendments are used by the auto-complete system as a work around for limitations in Monaco.

### Worked example
Consider if a user starts typing this (in Minecraft py):

```python
player.on_cha
```
Our auto-complete suggestion would be:

```python
player.on_chat
```

And the snippet we want to insert is:
```python
def on_chat_handler():
  pass
player.on_chat(on_chat_handler)
```

The problem is that Monaco would paste this snippet as:
```python
player.def on_chat_handler():
  pass
player.on_chat(on_chat_handler)
```
And there's no way I've found to remove the "player." prefix.
Note that we don't have this issue in typescript because of inline callbacks:

```javascript
player.on_chat(() => {

})
```
So the system basically sticks a json payload at the end of the snippet insertion:
```python
player.on_chat#AMENDMENT{behavior: "replaceLine", insertText: "def on_chat_handler(): ..." }
```
which we then sniff for in the onDidChangeModelContent event and apply the changes.

## Known limitations:
- Any time we use an edit amendment, the user's undo stack will have a superfluous entry.
For example, the undo stack for player.on_chat will look like:
After edit:
```python
def on_chat_handler():
  pass
player.on_chat(on_chat_handler)
```
1st undo:
```python
player.
```
2nd undo:
```python
player.on_cha
```
We could remove this extra entry if Monaco gave us a way to pop items from the undo stack,
or better yet if it supported our scenarios so that we didn't need edit amendments.
- The user may see a flicker of "#AMENDMENT:..." after completing a snippet

## Possible future issues:
- The implementation depends on using setTimeout() as a way to wait for Monaco, because
onDidChangeModelContent event notifies us before the change is applied to the buffer
presumably as a way to give us a chance to react somehow, however to apply another edit
we have to wait for the first edit to be applied, thus we use setTimeout().
However in most cases when you use setTimeout(), it's likely that you may be susceptible to
race conditions, especially if anyone else is listening onDidChangeModelContent and using
setTimeout().

### Future work
[ ] Snippet insertions still aren't indent aware.


riknoll:
Documenting offline conversation here: @darzu and I talked it over and I was afraid that this approach might have performance impacts so we will limit it to Python for now

darzu:
> Documenting offline conversation here: @darzu and I talked it over and I was afraid that this approach might have performance impacts so we will limit it to Python for now

Done. Thanks for the suggestion. I also made sure that the changes-listener would use very little CPU in the cases where there is no EditAmendment. I couldn't even get the time scanForEditAmendment spends to register in the chrome profiler with x6 slowdown while mashing keys.

webapp/src/monaco.tsx
```diff
@@ -1038,6 +1076,10 @@ export class Editor extends toolboxeditor.ToolboxEditor {
             // Reload the header if a change was made to the config file: pxt.json
             return this.parent.reloadHeaderAsync();
         }
+        if (this.editAmendmentsListener) {
+            this.editAmendmentsListener.dispose();
```
pelikhan: do you need to ``this.editAmendmentsListener = undefined`` ?

webapp/src/monaco.tsx
```diff
@@ -1936,6 +1978,8 @@ export class Editor extends toolboxeditor.ToolboxEditor {
                 range.startLineNumber + lines.length - 1, lines[lines.length - 1].length)
 
             const disposable = this.editor.onDidChangeModelContent(e => {
+                let changes = e.changes;
```
pelikhan: unused?

webapp/src/monacoEditAmendments.ts
```diff
@@ -0,0 +1,168 @@
...
+//    However in most cases when you use setTimeout(), it's likely that you may be susceptible to
+//    race conditions, especially if anyone else is listening onDidChangeModelContent and using
+//    setTimeout().
+
+export interface EditAmendment {
+    behavior: "replaceLine",
+    insertText: string,
+    selectText?: string,
+}
+type EditAmendmentInstance = {
```
pelikhan: what is this weird syntax?

webapp/src/monaco.tsx
darzu: > EditAmendmentInstance

It's an EditAmendment but with a concrete range in the editor.
An EditAmendment is independent of the current state of the editor.
An EditAmendmentInstance is only valid for a specific state of the editor.

webapp/src/monacoEditAmendments.ts
darzu: It's semantically (nearly) the same as:

interface EditAmendmentInstance extends EditAmendement {
   range: monaco.IRange
}

I can change it. It reads slightly differently, it's saying: "An EditAmendmentInstance has a range and everything that an EditAmendment has".

darzu:
> what is EditAmendmentInstance ?

It's an EditAmendment but with a concrete range in the editor.
An EditAmendment is independent of the current state of the editor.
An EditAmendmentInstance is only valid for a specific state of the editor.

webapp/src/monacoEditAmendments.ts
darzu: I'll change it, interfaces are more idiomatic

darzu: image

darzu: My brain likes to think in terms of algebraic types instead of inheritance :)

webapp/src/monacoEditAmendments.ts
```diff
@@ -0,0 +1,168 @@
...
+            model.pushEditOperations(editor.getSelections(), edits, inverseOp => [rangeToSelection(inverseOp[0].range)])
+        }, 0);
+    }, 0)
+}
+
+export function listenForEditAmendments(editor: monaco.editor.IStandaloneCodeEditor) {
+    return editor.onDidChangeModelContent(e => {
+
+        let amendment = scanForEditAmendment(e)
+        if (!!amendment)
```
riknoll: nitpick: !! is *not* necessary (pun intended)

PR #5724, opened 2019-07-01
Use first 'heading 1' of tutorial as title

shakao:
Looks for single '#' and text, otherwise defaults to file name (as before)


webapp/src/app.tsx
```diff
@@ -2889,7 +2889,7 @@ export class ProjectView
 
             const tutorialOptions: pxt.tutorial.TutorialOptions = {
                 tutorial: tutorialId,
-                tutorialName: title,
+                tutorialName: tutorialInfo.title || title,
```
riknoll: It might be good to rename the title variable to filename or something

PR #5739, opened 2019-07-09
Pipe Python errors into diagnostics

riknoll:
Now if you write some TypeScript that we cannot convert into Python, you get an error span and message. Currently only one error is supported, just like the other decompiler.

python_error


tests/common/testUtils.ts
```diff
@@ -80,13 +80,13 @@ export function ts2pyAsync(f: string): Promise<string> {
             let program = pxtc.getTSProgram(opts);
             // TODO: if needed, we can re-use the CallInfo annotations the blockly decompiler can add
             // annotate(program, tsFile, target || (pxt.appTarget && pxt.appTarget.compile));
-            const decompiled = (pxt as any).py.decompileToPythonHelper(program, "main.ts");
+            const decompiled = (pxt as any).py.decompileToPython(program, "main.ts");
```
jwunderl: Can drop the ``pxt as any`` here, right?

riknoll: I'm not going to change it since it's not a part of this PR

PR #5743, opened 2019-07-10
Allow unicode identifiers python

jwunderl:
fix https://github.com/microsoft/pxt-minecraft/issues/912 (unicode characters being marked as invalid variable names)

As far as I can tell the valid variable names for python3 and javascript, with the exception of python disallowing '$':

### js:

* Names can contain letters, digits, underscores, and dollar signs.
* Names must begin with a letter
* Names can also begin with $ and _
* Names are case sensitive (y and Y are different variables)
* Reserved words (like JavaScript keywords) cannot be used as names *(handled elsewhere)*

### python:

* Identifiers can be a combination of letters in lowercase (a to z) or uppercase (A to Z) or digits (0 to 9) or an underscore _.
* An identifier cannot start with a digit.
* Keywords cannot be used as identifiers. *(handled elsewhere)*
* cannot use special symbols like !, @, #, $, % etc. in identifier.
* Identifier can be of any length.

This change just makes python checks use the ts identifier check; I can explicitly disallow $ if we want to match python, but that seems a bit unnecessary?

I did notice that keywords aren't properly being switched when decompiling (e.g. ``let lambda = 0`` breaks as lambda is a keyword in python but not js), but that's a separate fix


jwunderl:
Side note, if anyone wants to run the trace tests, you'll have to locally add back in the special casing for console.log here: https://github.com/microsoft/pxt/pull/5593/files#diff-1f79ca9bc855a60d993a79f72a2a186dL905

It seems like we'll have to move where the renaming happens soon

PR #5753, opened 2019-07-13
Allow inline snippet insertion, match preceding whitespace

shakao:
Addresses microsoft/pxt-minecraft/issues/856

- If the snippet has a return value, insert the snippet inline (no newlines before or after)
- Match the leading whitespace of the preceding line, if it's a non-blank line (useful for Python)

Demo:
https://minecraft.makecode.com/app/a05e270a13ec83f9eedde8dedf07e69c39b49ad3-34cf3342c0


shakao:
yeah, that's the whitespace change. i didn't see a way to distinguish between snippets that had event handlers and snippets that just invoked an API (and in TS there wouldn't be any difference), so "match preceding whitespace" happens regardless. open to removing this change, though--it's mostly for python, and i could go either way on how much value it adds

pelikhan:
A snippet could have a “noindent” flag to force ignore indentation. Default would be on.

shakao:
added a formatting provider (/pull/5761) so will remove the whitespace change from this one

darzu:
Resurrecting this PR.

pelikhan:
@shakao @darzu is this PR ready?

darzu:
I'm integrating this with: https://github.com/microsoft/pxt/pull/6498

darzu:
Superseded by: https://github.com/microsoft/pxt/pull/6498

PR #5756, opened 2019-07-15
Adding a basic page for testing tutorial authoring

riknoll:
It's a simple page with the monaco editor on one side and the MakeCode editor on the other. Select your target in the dropdown and press the "Run in MakeCode" button to preview the tutorial! The text in the monaco editor also gets persisted in localstorage.

tutorial-tool-demo

Hacks:
1. I had to duplicate the tutorial parsing code from pxtlib. I can fix this later by making some changes in PXT but I wanted this to be standalone for now
2. The pxt.json for each editor is hardcoded in the ts file
3. Obviously it is very ugly

Sorry for the gigantic GIF


docs/static/tutorial-tool/tutorial.ts
```diff
@@ -0,0 +1,462 @@
+/// <reference path="../../../node_modules/monaco-editor/monaco.d.ts" />
```
shakao: maybe add a comment to the pxt tutorial.ts noting that this file should be kept in sync?

pelikhan:
is this a hack or something we want to push later on?

riknoll:
It would be nice to publish a cleaned up version later. It actually isn't too hacky, we could pretty easily move the parsing into PXT and make this more official

PR #5757, opened 2019-07-16
Substr length should not have a default

pelikhan:
Fix for https://github.com/microsoft/pxt-arcade/issues/1125

### DIFF
--- a/docs/static/playground/pxt-common/pxt-core.d.js
+++ b/docs/static/playground/pxt-common/pxt-core.d.js
@@ -244,7 +244,7 @@ declare interface String {
      * @param start first character index; can be negative from counting from the end, eg:0
      * @param length number of characters to extract
      */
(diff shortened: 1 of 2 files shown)


docs/static/playground/pxt-common/pxt-core.d.js
```diff
@@ -244,7 +244,7 @@ declare interface String {
      * @param start first character index; can be negative from counting from the end, eg:0
      * @param length number of characters to extract
      */
-    //% shim=String_::substr length.defl=10
+    //% shim=String_::substr
     //% help=text/substr
     //% blockId="string_substr" block="substring of %this=text|from %start|of length %length" blockNamespace="text"
     substr(start: number, length?: number): string;
```
darzu: don't you want to change this param to be non-optional?

PR #5758, opened 2019-07-16
un-lexically nest function

jwunderl:
Lexically nested functions are not allowed in strict mode in old safari (I'm personally very surprised they're allowed in any strict modes)

should fix https://github.com/microsoft/pxt-microbit/issues/2237, although I haven't yet tested this patch on the current microbit build (not sure which branches to use) - but this fixes it at the point it was first broken, and I don't think this is a common paradigm

### DIFF
--- a/pxtpy/converter.ts
+++ b/pxtpy/converter.ts
@@ -1369,19 +1369,20 @@ namespace pxt.py {
             let tupNames = tup.elts
                 .map(e => e as py.Name)
                 .map(convertName)
-            function convertName(n: py.Name) {
-                // TODO resuse with Name expr
-                markInfoNode(n, "identifierCompletion")
-                typeOf(n)
-                let v = lookupName(n)
-                return possibleDef(n, /*excludeLet*/true)
-            }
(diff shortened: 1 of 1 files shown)


jwunderl:
../node_modules/@types/request/index.d.ts(21,27): error TS7016: Could not find a declaration file for module 'form-data'. '/home/travis/build/microsoft/pxt/node_modules/form-data/lib/form_data.js' implicitly has an 'any' type.
  Try `npm install @types/form-data` if it exists or add a new declaration (.d.ts) file containing `declare module 'form-data';`
jake aborted.

pelikhan:
@jwunderl this fix needs to be integrated into stable6.9

jwunderl:
I thought it was 5.15, got the pr cherry picking to that branch here https://github.com/microsoft/pxt/pull/5759? Either way, I will need to work with @abchatra tomorrow / as soon as we can to get this to the live site

pelikhan:
It is the version referenced in package.json (https://github.com/microsoft/pxt-microbit/blob/master/package.json#L49) so yes 5.15

PR #5763, opened 2019-07-19
compilation speedup and ref-counting removal

mmoskal:
This improves "compileBinary" by ~25%, and overall compilation by ~15%.

TS compiler is now about half of compilation time (on jumpy platformer). I think further gains will have to come from some sort of incremental compilation.

This also removes support for non-GC build (which I'm 100% sure were broken before, but now it's really gone).

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -2822,6 +2822,9 @@ export function timeAsync() {
         mainPkg.getCompileOptionsAsync(mainPkg.getTargetOptions())
             .then(opts => pxtc.compile(opts))
             .then(res => {
+                U.iterMap(res.times, (k, v) => {
+                    res.times[k] = Math.round(v / 1000)
+                })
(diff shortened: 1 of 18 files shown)


mmoskal:
Also, using node.js, first overall compilation time is 3.4s on my box, then 2.0s, 1.7s, and after a few more down to 0.8s. This doesn't seem as terrible as advertised.

If you're getting much worse times in the browser, then it's probably worth investigating.

(BTW, all of these times are for JS compilation; native is much slower)

jwunderl:
re: build failure, looks like this just needs to be dropped?
https://github.com/microsoft/pxt/blob/12539bb7fdd3bf4c4f89338f75fdb73917927e1e/pxtcompiler/emitter/backjs.ts#L68-L69

mmoskal:
build fixed, @pelikhan @riknoll

PR #5765, opened 2019-07-22
Fetch live translations for all minecraft apps

shakao:
ipcRenderer is only used in Minecraft/Code Connection, and we want to fetch the live translations for those (instead of treating as Electron app)


pxtlib/browserutils.ts
```diff
@@ -123,7 +123,7 @@ namespace pxt.BrowserUtils {
         return typeof window != "undefined" && !!(window as any).ipcRenderer;
     }
     export function isElectron() {
-        return isPxtElectron() || isIpcRenderer();
+        return isPxtElectron();
```
pelikhan: maybe we should use "isPxtElectron" when deciding to check for translations?

PR #5766, opened 2019-07-23
Updated Melody Field and Tempo Sync on Block

kimprice:
The melody field on the block now shows the color representation of the melody with a music icon. The tempo on the block and in the editor is synced. The bottom bar is updated to look more similar to the sprite editor.

melodyEditor-V5


pelikhan:
Could we render the editor as a tiny image? In this case, one could clearly see that it is a diagonal songs -- something colors don't really convey.

Also, it's best practice not to rely on color alone and use some kind of shape... hence the curve.

kimprice:
@pelikhan It's an interesting idea but if we make it a miniature rendering of the grid, I think it would be difficult to see. Maybe this is worth discussing in a future design meeting.

pelikhan:
The spatial distribution of notes is meaningful to me, not the colors. Colors are easy to see but there's no easy pattern matching to be done on them since songs don't follow any kind of gradient/rainbow sequence.
image

kimprice:
@pelikhan I'll add an issue to suggest a spatial distribution, but I don't think this is something I'll have time to change before I leave next week.

pxtblocks/fields/field_melodySandbox.ts
```diff
@@ -299,6 +328,47 @@ namespace pxtblockly {
                 if (this.tempoInput) {
                     this.tempoInput.value = this.tempo + "";
                 }
+                // update tempo on block if not a variable
+                const s = this.sourceBlock_;
+                if (s.parentBlock_) {
```
riknoll: Should this just be a call to syncTempoField?

kimprice: One is syncing the field on the block with the field in the dropdown editor and the other is syncing from the editor to the field on the block. There is a lot of overlap in code though so I'll edit the sync function so that it can do both.

PR #5767, opened 2019-07-24
Support for array assignment

mmoskal:
This adds support for things like:

```typescript
let [a, b, c] = pins.unpackBuffer("<HHH", buf);
[x, y] = [y, x];
let {a, b} = someObject
```

The `{a,b} = someObject` (without `let`) is currently unsupported.

CC @jwunderl @darzu @shakao - this applies to Python translation; note that you need semicolon on the line above `[...] = [...]`...


pxtcompiler/emitter/emitter.ts
```diff
@@ -526,10 +542,10 @@ namespace ts.pxtc {
         return isClassType(t)
     }
 
-    function arrayElementType(t: Type): Type {
+    function arrayElementType(t: Type, idx = -1): Type {
```
pelikhan: idx is not used?

mmoskal: yeah, in future this should look if it's a tuple type and return the type for specific index; doesn't seem needed right now though

PR #5771, opened 2019-07-26
Tutorial progress circle component

shakao:
Initial components for circle display for tutorial with activity dropdown. The circle is always displayed if activities are present, and if the tutorial has 8+ steps it is displayed on mobile screen sizes.

Currently uses the H1 (#) tag as the divider for each activity, so it's backward-compatible with the current tutorials and activity segments can easily be added to existing tutorials by inserting H1 tags.

We could also use ## for activities, and ### for steps, using the version metadata to determine what kind of parsing to use, but that will make conversions a little more involved.

tutorial-circle

tutorial-circle-mob

Will also bring this to the design meeting next week for additional CSS/style iteration

### DIFF
--- a/webapp/src/container.tsx
+++ b/webapp/src/container.tsx
@@ -440,7 +440,10 @@ export class MainMenu extends data.Component<ISettingsProps, {}> {
                 {!inTutorial && homeEnabled ? <sui.Item className="icon openproject" role="menuitem" textClass="landscape only" icon="home large" ariaLabel={lf("Home screen")} text={lf("Home")} onClick={this.goHome} /> : null}
                 {showShare ? <sui.Item className="icon shareproject" role="menuitem" textClass="widedesktop only" ariaLabel={lf("Share Project")} text={lf("Share")} icon="share alternate large" onClick={this.showShareDialog} /> : null}
                 {inTutorial && <sui.Item className="tutorialname" tabIndex={-1} textClass="landscape only" text={tutorialOptions.tutorialName} />}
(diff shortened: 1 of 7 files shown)


webapp/src/container.tsx
```diff
@@ -438,7 +439,8 @@ export class MainMenu extends data.Component<ISettingsProps, {}> {
...
                 {!inTutorial && homeEnabled ? <sui.Item className="icon openproject" role="menuitem" textClass="landscape only" icon="home large" ariaLabel={lf("Home screen")} text={lf("Home")} onClick={this.goHome} /> : null}
                 {showShare ? <sui.Item className="icon shareproject" role="menuitem" textClass="widedesktop only" ariaLabel={lf("Share Project")} text={lf("Share")} icon="share alternate large" onClick={this.showShareDialog} /> : null}
-                {inTutorial ? <sui.Item className="tutorialname" tabIndex={-1} textClass="landscape only" text={tutorialOptions.tutorialName} /> : null}
+                {inTutorial && <sui.Item className="tutorialname" tabIndex={-1} textClass="landscape only" text={tutorialOptions.tutorialName} />}
+                {inTutorial && hasActivities && <tutorial.TutorialActivityDropdown parent={this.props.parent} currentActivity={tutorialOptions.tutorialStepInfo[tutorialOptions.tutorialStep].activity} activityInfo={tutorialOptions.tutorialActivityInfo} />}
```
riknoll: Could we break this line up?

PR #5773, opened 2019-07-27
Navigate to error in filelist when clicking on error number

pelikhan:
The error numbers in the filelist can be clicked and jump to the first error. This change also 
fixes the scrolling behavior when restoring the editor state in monaco.
error

- [x] remove file icons which were mostly wrong
- [x] allow all files to be delete minus pxt.json, main.ts
- [x] jump to first error when clicking error number.


webapp/src/filelist.tsx
```diff
@@ -76,11 +76,18 @@ export class FileList extends data.Component<ISettingsProps, FileListState> {
...
+    }
+
     private filesOf(pkg: pkg.EditorPackage): JSX.Element[] {
         const { currentFile } = this.state;
         const deleteFiles = pkg.getPkgId() == "this";
         return pkg.sortedFiles().map(file => {
-            let meta: pkg.FileMeta = this.getData("open-meta:" + file.getName())
+            const meta: pkg.FileMeta = this.getData("open-meta:" + file.getName())
+            const navigateToErrorHandler = () => {
+                this.navigateToError(meta.diagnostics && meta.diagnostics[0]);
```
riknoll: Looks like this could be undefined? navigateToError should guard

PR #5775, opened 2019-07-29
Getting rid of tutorial parsing duplication

riknoll:
In v1 of the tutorial tool I just duplicated the tutorial parsing code because I didn't want to make changes to pxt. Now that it's shipped, I'm removing the duplication and adding an editorcontroller message that imports a tutorial from just the markdown.

This will temporarily break the tutorial tool for all targets until they get bumped. We should do a bump of mc with these changes ASAP once this is merged, so I'm putting the "DO NOT MERGE" label on it

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1469,7 +1469,7 @@ export class ProjectView
 
     importTutorialAsync(md: string) {
         try {
-            const [options, editor] = getTutorialOptions(md, "untitled", "untitled", "", false);
+            const { options, editor } = getTutorialOptions(md, "untitled", "untitled", "", false);
             const dependencies = pxt.gallery.parsePackagesFromMarkdown(md);
 
             return this.createProjectAsync({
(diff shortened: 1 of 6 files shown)


webapp/src/app.tsx
```diff
@@ -3908,3 +3907,30 @@ document.addEventListener("DOMContentLoaded", () => {
         }
     }, false);
 })
+
+
+function getTutorialOptions(md: string, tutorialId: string, filename: string, reportId: string, recipe: boolean): [pxt.tutorial.TutorialOptions, string] {
```
pelikhan: can you return a map that explains what the return values mean? ```{ options: , ???: string }```

webapp/src/app.tsx
```diff
@@ -1462,10 +1462,29 @@ export class ProjectView
...
+
+            return this.createProjectAsync({
+                name: "untitled",
+                tutorial: options,
+                preferredEditor: editor,
+                dependencies
+            });
+        }
+        catch (e) {
+            Util.userError("Could not import tutorial");
```
pelikhan: add a tick?

riknoll: It doesn't swallow it, it rejects the promise. I'm not really sure that there is a good reason to pollute our exceptions with errors generated by the tutorial tool

pelikhan:
Point to /beta?

riknoll:
I have beta and non-beta endpoints for all of the targets in the dropdown. I can comment out the ones that don't work (all but minecraft) if that helps

pelikhan:
Minecraft is the only user of the tool. We haven’t advertised it anywhere.

PR #5776, opened 2019-07-30
Update activity/step and hint syntax for tutorial parsing

shakao:
Tutorial engine side of new hint/activity syntax

### DIFF
--- a/docs/writing-docs/tutorials.md
+++ b/docs/writing-docs/tutorials.md
@@ -126,6 +126,17 @@ a ``"editor": "js"`` entry for JavaScript tutorials and ``"editor": "py"`` entry
 
 The tutorial markdown has a format that the guides the tutorial runner in making a sequence of interactions:
 
+### Metadata
+
+Tutorial metadata can be optionally specified in a table at the top of the document. The table is a key-value pairing of fields. The currrent fields are:
+
+* **v**: version, for the tutorial syntax. When unspecified, default is V1. Used in tutorial parsing.
+
+```markdown
+| v             | 2               |
+| field_name    | field_value     |
+```
+
 ### Title
 
 The title is on the first line and uses a _level 1_ heading, like:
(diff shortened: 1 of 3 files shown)


docs/writing-docs/tutorials.md
```diff
@@ -126,6 +126,17 @@ a ``"editor": "js"`` entry for JavaScript tutorials and ``"editor": "py"`` entry
 
 The tutorial markdown has a format that the guides the tutorial runner in making a sequence of interactions:
 
+### Metadata
+
+Tutorial metadata can be optionally specified in a table at the top of the document. The table is a key-value pairing of fields. The currrent fields are:
+
+* **v**: version, for the tutorial syntax. When unspecified, default is V1. Used in tutorial parsing.
+
+```markdown
```
pelikhan: Does this table get hidden when rendering the tutorials as a docs page?

docs/writing-docs/tutorials.md
```diff
@@ -134,9 +145,40 @@ The title is on the first line and uses a _level 1_ heading, like:
 # Light blaster
 ```
 
+### Activities (V2)
+
+Tutorials with version **2** specified in the metadata table have multiple _activities_, each consisting of several steps. An activity begins with a _level 2_ heading (``##``) followed by the activity name, which will be displayed in a dropdown next to The activity can have any number of steps, but no activity-specific text.
```
pelikhan: I think the “multi level activity” should be a property not a absolute version number. What happens when we come up with tutorials v3?

docs/writing-docs/tutorials.md
```diff
@@ -185,14 +212,15 @@ In each step, any text before the first code snippet or image is automatically d
...
 
 ````markdown
-<!-- Try clicking on the 'Basic' drawer to find the blocks you need! Your code should look like this:
+#### ~ tutorialhint
+Try clicking on the 'Basic' drawer to find the blocks you need! Your code should look like this:
 
 ```blocks
 basic.showString("Micro!")
-``` -->
+```
```
pelikhan: Don you have to close it with #### ~ ?

docs/writing-docs/tutorials.md
shakao: Not right now, I can add that in the docs rendering update as well

docs/writing-docs/tutorials.md
shakao: What do you mean by property? Like, we infer anything with one heading less than the "step" as the "activity" divider? The concept of an "activity" in the code will transfer easily between versions, the only difference would be the regex for parsing

docs/writing-docs/tutorials.md
shakao: No, I was going to update the docsrender to automatically infer the closing--Michal suggested any heading that is equal to or less than the opening, to allow hints with headings

PR #5778, opened 2019-07-31
string.toLowerCase

pelikhan:
Migrate toLowerCase to PXT.

### DIFF
--- a/docs/static/playground/pxt-common/pxt-core.d.js
+++ b/docs/static/playground/pxt-common/pxt-core.d.js
@@ -301,8 +301,6 @@ declare interface String {
      */
     //% helper=stringToLowerCase
     //% help=text/to-lower-case
-    //% blockId="string_to_lower_case" blockNamespace="text"
-    //% block="split %this=text|to lower case"
     toLowerCase(): string;
 
     [index: number]: string;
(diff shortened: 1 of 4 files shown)


libs/pxt-common/pxt-core.d.ts
```diff
@@ -296,6 +296,15 @@ declare interface String {
     //% block="split %this=text|at %separator"
     split(separator?: string, limit?: number): string[];
 
+    /**
+     * Converts the string to lower case characters.
+     */
+    //% helper=stringToLowerCase
+    //% help=text/to-lower-case
+    //% blockId="string_to_lower_case" blockNamespace="text"
+    //% block="split %this=text|to lower case"
```
riknoll: split?

riknoll: Maybe "convert" instead?

docs/static/playground/pxt-common/pxt-core.d.js
```diff
@@ -296,6 +296,15 @@ declare interface String {
     //% block="split %this=text|at %separator"
     split(separator?: string, limit?: number): string[];
 
+    /**
+     * Converts the string to lower case characters.
+     */
+    //% helper=stringToLowerCase
+    //% help=text/to-lower-case
+    //% blockId="string_to_lower_case" blockNamespace="text"
+    //% block="split %this=text|to lower case"
```
mmoskal: block still here

PR #5781, opened 2019-08-01
simulator performance work

mmoskal:
This improves perf of generated JS about 2x.

* the interface between runtime.ts and generated code is cleaner
* call stack frames are now created in a single place per function, with all variables pre-allocated; this has dramatic perf impact (less hidden classes are used in JIT)
* various cleanups in generated code
* more explicit names used in generated code
* simple functions (`return a + b` etc) are now inlined; more functions are inlined in JS than in native
* improvement to perf-counter infrastructure

There are also some new docs on profiling.


riknoll:
Have you tried the debugger with these changes?

mmoskal:
Debugger seems to work. For good measure I disabled inlining in debug mode (probably not strictly necessary, but better safe than sorry)

PR #5790, opened 2019-08-02
Fix crash on m[x] = null

mmoskal:
introduced in https://github.com/microsoft/pxt/pull/5763 ; not sure why we're only seeing it now


abchatra:
Add a test if you can.

PR #5791, opened 2019-08-03
incremental compilation

mmoskal:
Emit simulator code incrementally. Things in `pxt_modules/*` are only emitted once, until Service.reset() (typically `loadHeader` in `app.tsx`). This brings down typical compilation time in Arcade from 1200ms to 100ms.

To measure time, best to use `?compiler=time` and look at the console.

* use language service Program in compile(), decompile() and friends
* TSC info is retained between compilations for `pxt_modules/*`
* our emit info is also retained, provided we're not doing native compilation
* class inheritance info is computed in a separate pass

Changes to allow re-use of generated JS:
* methods are now accessed via name not number in JS (still same semantics as native though)
* map fields are now also accessed by name
* class subtype checks use symbolic class numbers
* hex literals are emitted next to function using them
* context in stored in `Binary` class (framectors mostly)

TODO:
* [ ] support `autoCreate` or just drop it (no one seems to be using this)


mmoskal:
@riknoll @shakao @darzu @abchatra This should be good to give it a spin.

mmoskal:
Oh, and of course @jwunderl!

riknoll:
Test build https://arcade.makecode.com/app/594ebdebf9db6a648e7f227cb81e6d00498c88c2-bdc1b149f4

riknoll:
And here's one with autorun turned on: https://arcade.makecode.com/app/0992c22b25e61ebc4cc4936e3b1ada87b24cdbe7-eb2ec38b22

riknoll:
🎉

PR #5799, opened 2019-08-07
Set tutorial card z index above block fields

shakao:
Fixes microsoft/pxt-minecraft/issues/1104


riknoll:
Nvm I see it elsewhere in the file

PR #5800, opened 2019-08-07
Start of work for python built-in types

riknoll:
I've added two files in the libs folder:
```
libs/pxt-python/pxt-python.d.ts
libs/pxt-python/pxt-python-helpers.ts
```

These are equivalent to the `pxt-common` libs but for Python, completions/types are drawn from the interfaces here. In the `pxt.json` of the blocksprj in each of our targets, we need to add entries for both of these files if they plan on using Python.

For implementing the functions, you can either map to the TypeScript function directly using the `py2tsOverride` comment attribute or you can declare a helper function in the `_py` namespace and set it using the `pyHelper` comment attribute. For example, this python:

```python
x = [1, 2, 3]

x.count(1)
```

converts to this TypeScript:

```typescript
let x = [1, 2, 3]
_py.py_array_count(x, 1)
```

I went ahead and implemented all of the python list methods and I shimmed out all of the string methods. We probably don't actually want ALL of the string methods but I figure we can filter them out later.

My plan for the other direction (blocks => python) is to have another attribute called `ts2pyOverride` that will map all of the functions accessible in the blocks into idiomatic python. That's why the override supports extra features like default values and optional arguments that I'm not actually using yet. The helper functions already map directly in both directions.


mmoskal:
So if we have a block for count() and someone switches to Typescript they will get this _py thingy? I think we should add count() on the TS side.

riknoll:
Everything in blocks will map to non-helper TypeScript. The helper stuff is just for built-ins in python that we want to support but don't have an equivalent in blocks/TypeScript. I agree that if we add a count block then we should add count to the TS as well.

riknoll:
In other words, it's just for python => typescript not python => blocks

mmoskal:
Cool! I think it would be good to limit python helpers to minimum and extend TS in general in places where it makes sense like count(). But I guess we can do it gradually.

PR #5801, opened 2019-08-08
Array default values with shadow blocks

riknoll:
Lets you specify the *inner* shadow block for array shadow blocks. For example, in the animation API that @dannytech and @dharinirajah have been making the default value in the array should be images:

Screen Shot 2019-08-07 at 5 29 21 PM

To use this, all you need to do is set the `defl` for the parameter on the API to the block ID of the shadow. For example,

```typescript
    //% blockId=run_image_animation
    //% block="%sprite=variables_get(mySprite) animate frames %frames=lists_create_with with interval %frameInterval=timePicker ms"
    //% group="Animate"
    //% frames.defl=screen_image_picker
    export function runImageAnimation(sprite: Sprite, frames: Image[], frameInterval?: number): void {
}
```

(note the `frames.defl` line)

### DIFF
--- a/docs/defining-blocks.md
+++ b/docs/defining-blocks.md
@@ -160,6 +160,28 @@ export function showNumber(v: number, interval: number = 150): void
 
 **Playground examples**: [Range](https://makecode.com/playground#field-editors-range), [Default values](https://makecode.com/playground#basic-default-values)
 
+
+## Array default values
+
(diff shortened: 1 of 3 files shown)


riknoll:
Also, I put two items in the array because three took up a lot of space and two is more obvious than one

PR #5824, opened 2019-08-17
remove spritekind enum hack

jwunderl:
all examples are upgraded in arcade now

### DIFF
--- a/pxtrunner/renderer.ts
+++ b/pxtrunner/renderer.ts
@@ -238,9 +238,6 @@ namespace pxt.runner {
         if (!options.layout) options.layout = pxt.blocks.BlockLayout.Align;
         options.splitSvg = true;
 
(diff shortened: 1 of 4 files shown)


pxtrunner/renderer.ts
```diff
@@ -739,8 +736,6 @@ namespace pxt.runner {
 
         function render(e: Node, ignored: boolean) {
             if (typeof hljs !== "undefined") {
-                // FIXME: Remove this patchArcadeSnippets once arcade documentation has been updated from enums to namespace for spritekind
-                $(e).text(pxt.tutorial.patchArcadeSnippets($(e).text().replace(/^\s*\r?\n/, '')))
```
riknoll: I think we want to to keep the replace part of this

PR #5825, opened 2019-08-17
Fix some accessibility issues

jwunderl:
Ended up refactoring a little bit / removing some unused code along the way, will annotate those with some comments at least / can separate them if needed. Most of the commits list the issue they are fixing, though a few were just picking off bugs that didn't have issues associated with them / clean up

### DIFF
--- a/webapp/src/filelist.tsx
+++ b/webapp/src/filelist.tsx
@@ -308,7 +308,8 @@ class FileTreeItem extends sui.StatelessUIElement<FileTreeItemProps> {
     }
 
     renderCore() {
-        const { onClick, onItemClick, onItemRemove, isActive, hasDelete, file, meta, ...rest } = this.props;
+        // onErrorClick is explicitly set so that it is not passed down under the ...rest
+        const { onClick, onItemClick, onItemRemove, isActive, hasDelete, file, meta, onErrorClick, ...rest } = this.props;
 
         return <a
             onClick={this.handleClick}
--- a/webapp/src/filelist.tsx
+++ b/webapp/src/filelist.tsx
@@ -324,8 +324,11 @@ class FileTreeItem extends sui.StatelessUIElement<FileTreeItemProps> {
             {hasDelete ? <sui.Button className="primary label" icon="trash"
                 title={lf("Delete file {0}", file.name)}
                 onClick={this.handleRemove}
-                onKeyDown={this.handleButtonKeydown} /> : ''}
-            {meta &&  meta.numErrors ? <a className='ui label red button' role="button" title={lf("Go to error")} onClick={this.handleErrorClick}>{meta.numErrors}</a> : undefined}
+                onKeyDown={this.handleButtonKeydown} /> : undefined}
+            {meta && meta.numErrors ? <sui.Button className='ui label red button' role="button"
+                title={lf("Go to error")} onClick={this.handleErrorClick}>
+                {meta.numErrors}
+            </sui.Button> : undefined}
         </a>
     }
 }
--- a/webapp/src/projects.tsx
+++ b/webapp/src/projects.tsx
@@ -588,12 +588,12 @@ export class ProjectsDetail extends data.Component<ProjectsDetailProps, Projects
         else if (cardType == "template") clickLabel = lf("New Project");
         else if (youTubeId) clickLabel = lf("Play Video");
 
-        const actions = [{
+        const action = {
             label: clickLabel,
             onClick: this.handleDetailClick,
             icon: '',
             className: 'huge positive'
-        }]
+        }
 
         // featured link: featured_link
         const isForum = cardType == "forumUrl" && url;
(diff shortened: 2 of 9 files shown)


webapp/src/projects.tsx
```diff
@@ -575,66 +599,47 @@ export class ProjectsDetail extends data.Component<ProjectsDetailProps, Projects
...
-                                : <sui.Button
-                                    key={`action_${action.label}`}
-                                    icon={action.icon}
-                                    text={action.label}
-                                    className={`approve ${action.icon ? 'icon right labeled' : ''} ${action.className || ''}`}
-                                    onClick={action.onClick}
-                                    onKeyDown={sui.fireClickOnEnter} />
-                        )}
-                        {isForum && <sui.Button
+                        <sui.Button
```
jwunderl: Needed to switch to just being buttons so that all of them would be focusable in the same way (make it so in keyboard navigation it will immediately focus on the open button you actually want to use). Anchor tags don't support autoFocus, and all of them behaved effectively the same way in the first place.

Also removed the unused array for making more actions here - seems to have been added for extensibility, but it was ignored when the forumLink button was added and would make the focusing issue less consistent / just makes things a bit harder to parse.

webapp/src/projects.tsx
```diff
@@ -645,23 +650,6 @@ export class ProjectsDetail extends data.Component<ProjectsDetailProps, Projects
                 </div>
             </div>
         </div>;
-
-        function isCodeCardType(value: string): value is pxt.CodeCardType {
-            switch (value) {
```
jwunderl: Moved the behavior of buttons up to handleDetailClick

webapp/src/carousel.tsx
```diff
@@ -100,10 +100,10 @@ export class Carousel extends data.Component<ICarouselProps, ICarouselState> {
                     }
                 </div>
             </div>
-            <span role="button" className={"carouselarrow right aligned" + (rightDisabled ? " arrowdisabled" : "")} aria-label={lf("See more")}
-                tabIndex={rightDisabled ? -1 : 0} onClick={this.onRightArrowClick} onKeyDown={sui.fireClickOnEnter} ref={this.handleArrowRefs}>
+            {!rightDisabled && <span role="button" className={"carouselarrow right aligned"} tabIndex={0}
```
riknoll: If both left and right are not disabled (like if you're in the middle of a long carousel) do the tab indices need to be different?

webapp/src/filelist.tsx
```diff
@@ -308,7 +308,8 @@ class FileTreeItem extends sui.StatelessUIElement<FileTreeItemProps> {
     }
 
     renderCore() {
-        const { onClick, onItemClick, onItemRemove, isActive, hasDelete, file, meta, ...rest } = this.props;
+        // onErrorClick is explicitly set so that it is not passed down under the ...rest
+        const { onClick, onItemClick, onItemRemove, isActive, hasDelete, file, meta, onErrorClick, ...rest } = this.props;
```
riknoll: Doesn't look like onErrorClick is being used

webapp/src/projects.tsx
```diff
@@ -401,6 +401,7 @@ export class ProjectsCarousel extends data.Component<ProjectsCarouselProps, Proj
                     </div>
                 </div>
             } else {
+                const selectedElement = cards.filter((scr, index) => index == selectedIndex)[0];
```
riknoll: Isn't this just `cards[selectedIndex]`?

webapp/src/toolbox.tsx
```diff
@@ -716,15 +716,17 @@ export class TreeRow extends data.Component<TreeRowProps, {}> {
             </style>
             iconContent = undefined;
         }
+        const rowTitle = name ? name : Util.capitalize(subns || nameid);
 
         return <div role="button" ref={this.handleTreeRowRef} className={treeRowClass}
             style={treeRowStyle} tabIndex={0}
+            aria-label={lf("Open category {0}", rowTitle)} aria-expanded={selected}
```
riknoll: Toggle category?

webapp/src/projects.tsx
```diff
@@ -554,8 +551,35 @@ export class ProjectsDetail extends data.Component<ProjectsDetailProps, Projects
     }
 
     handleDetailClick() {
-        const { scr, onClick } = this.props;
-        onClick(scr);
+        const  { cardType, url, youTubeId, scr, onClick } = this.props;
```
riknoll: whitespace

webapp/src/carousel.tsx
jwunderl: tab index 0 says to use regular tab order (based off html ordering); only need to be unique if >= 1

webapp/src/projects.tsx
jwunderl: Yup, should have cleaned up more

webapp/src/filelist.tsx
jwunderl: It isn't used here, it's used in the ``handleErrorClick`` method above, this is pulling it out of the ``...rest`` to prevent it from propagating to the anchor element; is there a better way to structure here? I just fixed this one real quick to stop getting errors, nothing came to mind immediately

webapp/src/projects.tsx
```diff
@@ -422,26 +423,22 @@ export class ProjectsCarousel extends data.Component<ProjectsCarouselProps, Proj
...
-                                name={scr.name}
-                                key={'detail' + scr.name}
-                                description={scr.description}
-                                url={scr.url}
-                                imageUrl={scr.imageUrl}
-                                largeImageUrl={scr.largeImageUrl}
-                                youTubeId={scr.youTubeId}
-                                scr={scr}
+                    <div ref="detailView" className={`detailview ${selectedElement ? 'visible' : ''}`}>
+                        {selectedElement && <sui.CloseButton onClick={this.closeDetail}/>}
```
shakao: small: could pull this 'selectedElement' out and wrap the detailView instead of each child element

PR #5832, opened 2019-08-20
Stop simulator when the melody editor is open

riknoll:
Fixes https://github.com/microsoft/pxt-adafruit/issues/1093

### DIFF
--- a/pxtblocks/fields/field_melodySandbox.ts
+++ b/pxtblocks/fields/field_melodySandbox.ts
@@ -78,7 +78,7 @@ namespace pxtblockly {
 
             this.prevString = this.getText();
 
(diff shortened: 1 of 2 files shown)


pxtblocks/fields/field_melodySandbox.ts
```diff
@@ -78,11 +78,17 @@ namespace pxtblockly {
 
             this.prevString = this.getText();
 
+            // The webapp listens fo this event and stops the simulator so that you don't get the melody
```
jwunderl:
```suggestion
            // The webapp listens to this event and stops the simulator so that you don't get the melody
```

PR #5835, opened 2019-08-21
use textColor less var for 'learn more' link

jwunderl:
re: https://github.com/microsoft/pxt-adafruit/issues/1098, will need to set the variable to white in adafruit

also make experiments header display properly like other text boxes, which I'm pretty sure I remember there being an issue for at some point but I don't see it

### DIFF
--- a/theme/common.less
+++ b/theme/common.less
@@ -405,7 +405,7 @@ div.simframe > iframe {
 }
 
 .ui.card .extra.content a.learnmore {
-    color: rgba(0, 0, 0, .6);
+    color: @textColor;
 }
 
 /* Popup message */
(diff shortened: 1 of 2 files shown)


riknoll:
Do we have a secondary text color? I feel like it should be de-emphasized like in micro:bit:

Screen Shot 2019-08-22 at 4 07 08 PM

jwunderl:
Hm, closest seems to be `@secondaryColor`, but that's more of an accent / doesn't have high enough contrast to be used for accessible text (it's a similar color in microbit to the gray text from that screenshot, which had a contrast ratio under 3 and had to be darkened a bit). I'll just add a `@secondaryTextColor` with a gray that works

PR #5839, opened 2019-08-22
Animated melody editor and other fixes

riknoll:
melody-editor-updates

Changes include:
* Melody editor highlights columns as they play
* Clicking on a note while the melody is playing no longer triggers that note
* Changing notes as the melody plays now immediately changes the sound (it used to wait for the next measure)
* Gallery cards now consist of two attached buttons to make the left side look more clickable
* Fixed the tones in the melody editor so that they no longer "clip" when a note ends
* A lot of other tiny bug fixes

I am planning to also fix the sound clipping in the simulator but that will be a separate PR.


pelikhan:
Yes!

PR #5842, opened 2019-08-26
Github actions workflow

pelikhan:
Adding workflow configuration for github actions in generated projects. Tested at https://github.com/pelikhan/pxt-qrcode/actions


pelikhan:
We need to upgrade pouchdb - we don’t build with node 10

PR #5849, opened 2019-08-29
Snippet Builder improvements

darzu:
Helps me prototype the mod-a-game experience and they're generally useful improvements for the SnippetBuilder.

- Allow SnippetBuilder to emit many top-level blocks
- Add support for external .ts files that has snippet content (much easier authoring experience for large snippets)
- Auto-format if <2 existing blocks
- Adds "replace" output mode that replaces all old blocks

The implementation isn't perfect yet since it cannot distinguish on event parameters (on_game_update 500ms vs 1000ms) but it's already more useful for a few scenarios. TODO included in code.

Tested using this custom package: https://github.com/darzu/ModGame

### DIFF
--- a/localtypings/pxtpackage.d.ts
+++ b/localtypings/pxtpackage.d.ts
@@ -164,12 +164,15 @@ declare namespace pxt {
         mimeType: string;
     }
 
+    type SnippetOutputType = 'blocks'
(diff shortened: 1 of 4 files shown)


pelikhan:
is this PR ready?

PR #5852, opened 2019-08-30
Remove reference to IE11 from docs as a supported browsers

abchatra:
We will still support IE11 on microbit for at least this school year. This removes reference to IE11 as supported browsers to discourage users. Also updates some of the chrome versions.


Jaqster:
We should let Adafruit know.

PR #5856, opened 2019-08-30
Add image tags and gallery filter

jwunderl:
Just on a per image editor type basis for now, can add search or user filtering later but probably needs some ui design / feels a bit separate

related prs:

* https://github.com/microsoft/pxt-common-packages/pull/953
* https://github.com/microsoft/pxt-arcade/pull/1271

makes it so you can filter things into or out of the image editor gallery based off tags:

Couple questions;

If we want to use these tags for some sort of search eventually, would need to extract them for localization; it seems like having these as commentAttributes as they are here would allow for that, but just want to confirm

I haven't document the format for the new fields in the docs, but for now its:

* **filter**: a string containing a series of tags, where tags beginning with "!" mean to exclude images with that tag and any other tag means it must be included. In [this](https://github.com/microsoft/pxt-common-packages/pull/953) pr, it's used to remove tiles and dialog images from _spriteImage ("!tile !dialog"), and only include tiles / dialog in the other two blocks
* **tags**: a string with the tags that describe this image; a "?" before a tag means it should not be used to exclude the image. e.g. there are certain tiles that also make sense as sprite types of images, e.g. the [the buttons here](https://github.com/microsoft/pxt-arcade/pull/1271/files#diff-507d0aab0c8cf8238c9c0ac5e6c85a5dR15)

Is that confusing / unmaintainable? I'm happy to go with a different approach, haven't spent too much time thinking of alternatives.

### DIFF
--- a/pxtlib/sprite-editor/gallery.ts
+++ b/pxtlib/sprite-editor/gallery.ts
@@ -94,13 +94,12 @@ namespace pxtsprite {
                 .filter(el => !!el);
 
             return target.filter(el =>
-                includeTags.every(tag => itemContainsTag(el, tag))
-                && excludeTags.every(tag => !itemContainsTag(el, tag))
+                includeTags.every(tag => el.tags.indexOf(tag) !== -1)
+                && excludeTags.every(tag => {
+                    const index = el.tags.indexOf(tag);
+                    return index === -1 || el.tags.charAt(index - 1) === "?";
+                })
             );
-
-            function itemContainsTag(item: GalleryItem, tag: string) {
-                return item.tags.indexOf(tag) !== -1;
-            }
         }
 
         protected buildDom() {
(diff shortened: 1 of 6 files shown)


pxtlib/sprite-editor/gallery.ts
```diff
@@ -93,13 +93,22 @@ namespace pxtsprite {
...
+                    item.tags.some(tag => {
+                        const ind = tag.indexOf(filterTag);
+                        return ind === 0 || ind === 1 && tag.charAt(0) === "?";
+                    })
+                )
+            }
+
+            function checkExclude(item: GalleryItem) {
+                return excludeTags.every(filterTag =>
+                    !item.tags.some(tag => tag.indexOf(filterTag) === -1)
```
jwunderl: as a side note I phrased the commit message for this poorly, these are currently counting as matching the tags as long as there is a tag on the image that **starts with** the given tag; easy enough to change if preferred

jwunderl:
Added some docs, just to the buildsprites doc for now as I don't immediately see anywhere similar things are documented to put it

pxtlib/sprite-editor/gallery.ts
riknoll: I think this should be an explicit equals check (ignoring case, but matching the whole string)

PR #5860, opened 2019-09-03
Pass messages to sim when embedding run page

riknoll:
Needed for the experimental work, just adding a channel to proxy messages through the simulator embed page (the one that compiles the program and creates an iframe to run the sim).


riknoll:
No longer needed

PR #5866, opened 2019-09-05
Add target option to name project when created

mmoskal:
This adds a dialog that pops up when a new project is created using the big button in the home screen. We've discussed that with David a few months back. This is controlled by appTheme option (to be enabled for Arcade).

With the new `settings` namespace, the storage resets when the name changes, and also stays on device when the name is the same on two programs. Because of that it's more important now to name projects.

If you hit "Create!" with no name, you get "Untitled".

Sep-04-2019 17-49-26


abchatra:
I would remove 'cancel' button, they have close (`x`). What is the behavior of 'x'

I love your determination @Mmoskal.

abchatra:
We are moving away from cancel button in all other dialogs

mmoskal:
Cancel is the same as X. Will remove

Jaqster:
Looks great.  I think this feature has been on the backlog for almost 2 years now...

PR #5871, opened 2019-09-05
allow intersection types in blocks

jwunderl:
fix checks for blocks with params that have intersection types

Nice to support in general, but the case I'd be using it for is https://github.com/microsoft/pxt-arcade/issues/1254, as it's rejecting because it doesn't recognize sprite as a structural subtype of ParticleAnchor, and easiest fix without potentially breaking existing scripts / ability to remove from any ParticleAnchor is to make the type for the ``Sprite | ParticleAnchor``.

quick test / example:

```typescript
namespace info {

    //% block="test string int number $myInput"
    export function testBlock(myInput: string | number) { }

    //% block="test sprite int sprite[] $myInput"
    export function testBlock2(myInput: Sprite | Sprite[]) { }
}
```

Can't think of a place to add tests for this immediately, but if possible that seems like it would be good to do

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -931,8 +931,9 @@ namespace pxt.blocks {
                     // to the blocks that accept any array (e.g. length, push, pop, etc)
                     if (isArrayType(subtype)) {
                         if (types.length > 1) {
-                            // type inference will potentially break non-trivial arrays until we have better
-                            // type handling in blocks, so escape and allow any block to be dropped in.
+                            // type inference will potentially break non-trivial arrays in intersections
+                            // until we have better type handling in blocks,
+                            // so escape and allow any block to be dropped in.
                             return null;
                         } else {
                             output.push("Array");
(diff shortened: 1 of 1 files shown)


pxtblocks/blocklyloader.ts
```diff
@@ -768,27 +768,23 @@ namespace pxt.blocks {
...
-                                    inputCheck = "String"
-                                }
-                            } else if (pr.type == "any") {
+                            } else if (pr.type == "number" && pr.shadowBlockId && pr.shadowBlockId == "value") {
+                                inputName = undefined;
+                                fields.push(namedField(new Blockly.FieldTextInput("0", Blockly.FieldTextInput.numberValidator), defName));
+                            } else if (pr.type == "string" && pr.shadowOptions && pr.shadowOptions.toString) {
                                 inputCheck = null;
+                            } else if (pr.type.indexOf("|") !== -1) {
+                                const types = pr.type.split(/\s*\|\s*/);
```
riknoll: Shouldn't this go into `getBlocklyCheckForType`?

PR #5902, opened 2019-09-14
Improved github experience

pelikhan:
Various improvement to the github editing experience, with integrated diff, creation of repo from project, revert support, branch switching, PR or fork creation.

## Testing

* microbit: https://makecode.microbit.org/app/67467aa31cc49d1a804ea3d08630b74ee1273c08-acac7e5eb5
* arcade: https://arcade.makecode.com/app/cfff9ca7fa41c8b8e396d1c05292adee46f82abb-4e51fdfae8 (has the serial bug, fixed in maker)
* maker: https://maker.makecode.com . In this editor, the "alwaysGithub" flag is on so the explorer item will show even if you do not have a token.

## Todos

- [x] guard feature with experiment
- [x] add github item in explorer
- [x] indicate pending changes in github
- [x] create new repo from current project
- [x] use editor instead of dialog for commit (similar to serial)
- [x] display diff of changes
- [x] button to open github repo page
- [x] button to pull only
- [x] detect that repo already exists when creating new repo
- [x] fix "go back" button
- [x] analytics
- [x] skin revert button
- [x] confirm before reverting
- [x] don't start simulator when entering github mode
- [x] show number of changes to pull (if any)
- [x] remove "*" indicator from files, and use that instead of up arrow?
- [x] check for expired/invalid token
- [x] check for token without the right permissions
- [x] inline changes rendering (punted)
- [x] bump when no content changes
- [x] implement pullRequestAsync
- [x] error handling
- [x] handle blocks files in diff
- [x] generate correct githubaction code (trying to install pxt-codal instead of pxt-maker)
- [x] ensure github logo visible on file
- [x] fix gitcurrent logic (can't pull if gitCurrent is false and there are not changes)
- [x] new branch option
- [x] create PR from branch button
- [x] remove old code from app.tsx (https://github.com/microsoft/pxt/pull/5902/commits/e263a422db9723b5a965f2a67bded4515cc65599)
- [x] indicate that pull is needed in filelist item (maybe skip this one?)
- [x] support for github urls in "#pub:"
- [x] new files don't show up in diff
- [x] show removed files in diff?
- [x] add fork option when no permission to commit?
- [x] radio options when creating release (instead of freeform version textbox)
- [x] makes sure to add test file
- [x] **race condition when creating github project generates empty pxt.json**
- [x] indicate that a file was added
- [x] display total number of items in github item
- [x] reuse existing header if available
- [x] clean state when "going home"
- [x] check serial view
- [x] ignore serial.txt


webapp/src/gitjson.tsx
```diff
@@ -164,10 +164,19 @@ export class Editor extends srceditor.Editor {
...
+                    this.parent.setState({ })
+                })
+        }
         cache.diff = (
             <div key={f.name} className="ui segments filediff">
                 <div className="ui segment header">
-                    <p>{f.name}</p>
+                    <span>{f.name}</span>
+                    <sui.Button text={lf("Revert")} className="small" ariaLabel={lf("Revert file")} onClick={revert}>
+                        <sui.Icon icon="undo" />
```
pelikhan: There is a Icon attribute on button

mmoskal: yeah found that ;)

webapp/src/gitjson.tsx
```diff
@@ -538,8 +540,8 @@ export class Editor extends srceditor.Editor {
                         </div>
                     </div>
                     <div className="rightHeader">
-                        <sui.Button className="ui icon button" icon="down arrow"
-                            text={this.needsPull ? lf("Pull changes") : lf("Up to date")} textClass={lf("landscape only")} title={lf("Pull changes")} onClick={this.handlePullClick} onKeyDown={sui.fireClickOnEnter} />
+                        <sui.Button className="ui icon button" icon={`${this.needsPull ? "down arrow" : "check"} ${this.needsPull ? "positive" : ""}`}
+                            text={this.needsPull ? lf("Pull changes") : lf("Up to date")} textClass={lf("landscape only")} title={lf("Pull changes from GitHub to get your code up-to-date.")} onClick={this.handlePullClick} onKeyDown={sui.fireClickOnEnter} />
```
mmoskal: yeah, I think the check mark is enough, no need for "up to date"

webapp/src/gitjson.tsx
```diff
@@ -541,7 +541,7 @@ export class Editor extends srceditor.Editor {
                     </div>
                     <div className="rightHeader">
                         <sui.Button className="ui icon button" icon={`${this.needsPull ? "down arrow" : "check"} ${this.needsPull ? "positive" : ""}`}
-                            text={this.needsPull ? lf("Pull changes") : lf("Up to date")} textClass={lf("landscape only")} title={lf("Pull changes from GitHub to get your code up-to-date.")} onClick={this.handlePullClick} onKeyDown={sui.fireClickOnEnter} />
+                            text={lf("Pull changes")} textClass={lf("landscape only")} title={lf("Pull changes from GitHub to get your code up-to-date.")} onClick={this.handlePullClick} onKeyDown={sui.fireClickOnEnter} />
```
mmoskal: I think this should be `this.needsPull !== false` so that we show down-arrow when re-computing (while it's `null`)

webapp/src/app.tsx
```diff
@@ -774,7 +774,7 @@ export class ProjectView
     }
 
     private pickEditorFor(f: pkg.File): srceditor.Editor {
-        return this.allEditors.filter(e => e.acceptsFile(f))[0]
+        return this.allEditors.filter(e => e.acceptsFile(f), 1)[0]
```
mmoskal: This passes `1` as `this` to the callback. Not very useful?

pelikhan:
Let’s move the file history into its own module and cleanup

mmoskal:
> Let’s move the file history into its own module and cleanup

The current problem isn't history, it's state laying around in ProjectView after going to home.

webapp/src/app.tsx
pelikhan: wrong very wrong

webapp/src/filelist.tsx
```diff
@@ -253,21 +254,102 @@ namespace custom {
...
+            key="github-status"
+            className="item"
+            onClick={this.handleClick}
+            tabIndex={0}
+            role="treeitem"
+            aria-selected="false"
+            aria-label="github status"
+            onKeyDown={sui.fireClickOnEnter}>
+            <i className="github icon" />
+            {ghid ? (ghid.project && ghid.tag ? `${ghid.project}#${ghid.tag}` : ghid.fullName) : lf("create GitHub repository")}
```
jwunderl: nit: maybe ``(ghid.project && ghid.tag ? `${ghid.project}#${ghid.tag}` : ghid.fullName)`` could be pulled out into a const

webapp/src/gitjson.tsx
```diff
@@ -0,0 +1,694 @@
...
+        try {
+            let status = await workspace.pullAsync(this.parent.state.header)
+                .catch(this.handleGithubError)
+            switch (status) {
+                case workspace.PullStatus.NoSourceControl:
+                case workspace.PullStatus.UpToDate:
+                    break
+                case workspace.PullStatus.NeedsCommit:
+                    this.needsCommitMessage = true;
+                    this.parent.setState({});
```
jwunderl: I might be parsing this wrong, but it looks like the finally is force refreshing anyway so this should be redundant, right?

webapp/src/workspace.ts
```diff
@@ -438,45 +438,63 @@ export enum PullStatus {
...
+    cfg.version = newVer || bumpedVersion(cfg)
     files[pxt.CONFIG_NAME] = JSON.stringify(cfg, null, 4)
     await saveAsync(hd, files)
     return await commitAsync(hd, cfg.version, "v" + cfg.version)
 }
 
+export function lookupFile(commit: pxt.github.Commit, path: string) {
+    if (!commit)
+        return null
+    return commit.tree.tree.filter(e => e.path == path)[0]
```
jwunderl:
```suggestion
    return commit.tree.tree.find(e => e.path == path)
```

webapp/src/core.ts
```diff
@@ -154,6 +155,7 @@ export interface DialogOptions {
     header: string;
     body?: string;
     jsx?: JSX.Element;
+    jsxd?: () => JSX.Element; // dynamic-er version of jsx
```
jwunderl: how required is this / how much work would it be to just implement it as jsx? Looks like it's just used in a few dialogues, and it seems like this is a part that's fighting with react a bit and might make it harder to reason about (mainly if it's used more than just one or two places)

webapp/src/app.tsx
```diff
@@ -812,13 +845,16 @@ export class ProjectView
 
                 if (this.state.showBlocks && this.editor == this.textEditor) this.textEditor.openBlocks();
             })
-            .finally(() => this.setStateAsync({ updatingEditorFile: false }))
+            .finally(() => {
+                this.updatingEditorFile = false
+                this.setState({}); // force any kind of state changes that is not tracked by react
```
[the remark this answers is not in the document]
pelikhan: This is a lock variable; it is not used to render the UI in any way. Having out of the React state makes it more deterministics and debuggable (no delayed updates).

webapp/src/gitjson.tsx
pelikhan: Nice catch.

webapp/src/core.ts
pelikhan: The problem with JSX is that it is completely static. So it is not possible to have some local state for the dialog (for example needed to track which radio button is pressed).
Our entire dialog scheme is fighting react.

pelikhan:
``jsxd`` is basically an "display" function for dialogs so that we can still have a local dialog state. ``jsx`` is completely static.

webapp/src/core.ts
jwunderl: Yeah, that makes sense to me. I guess an option would be to make dialogs classes, but that seems like a decent chunk of work that I wouldn't want to see added to a PR that's already this large hah

pelikhan: To be clear, this is already an improvement compared to using the DOM element created by React in some dialogs.

webapp/src/workspace.ts
```diff
@@ -549,29 +567,45 @@ async function githubUpdateToAsync(hd: Header, repoid: string, commitid: string,
...
+                U.userError(lf("Corrupt SHA1 on download of '{0}'.", path))
+        }
+        treeEnt.blobContent = text
+        return treeEnt.blobContent
     }
 
-    await downloadAsync(pxt.CONFIG_NAME)
-    let cfg = JSON.parse(files[pxt.CONFIG_NAME] || "{}") as pxt.PackageConfig
+    const cfgText = await downloadAsync(pxt.CONFIG_NAME)
+    const cfg = JSON.parse(cfgText || "{}") as pxt.PackageConfig
```
riknoll: Error handling?

webapp/src/dialogs.tsx
```diff
@@ -560,40 +561,75 @@ export function showImportUrlDialogAsync() {
...
         if (res) {
             pxt.tickEvent("app.github.create");
-            const name = inputName.value.trim()
-            const desc = inputDesc.value.trim()
 
-            if (/^[\w\-]+$/.test(inputName.value)) {
-                core.showLoading("creategithub", lf("Creating GitHub repo..."))
-                return pxt.github.createRepoAsync(name, desc)
+            if (!repoNameError()) {
+                core.showLoading("creategithub", lf("creating {0} repository...", repoName))
```
riknoll: Capitalization

webapp/src/filelist.tsx
```diff
@@ -99,7 +103,9 @@ export class FileList extends data.Component<ISettingsProps, FileListState> {
                     hasDelete={deleteFiles && !/^(main\.ts|pxt\.json)$/.test(file.name)}
                     className={(currentFile == file ? "active " : "") + (pkg.isTopLevel() ? "" : "nested ") + "item"}
                 >
-                    {file.name} {meta.isSaved ? "" : "*"}
+                    {file.name}
+                    {showStar ? "*" : ""}
+                    {meta.isGitModified ? " ↑" : ""}
```
riknoll: We should probably use an icon and not the unicode character. Semantic has a lot of arrows.

webapp/src/filelist.tsx
```diff
@@ -253,21 +254,102 @@ namespace custom {
...
+    }
+
+    private handleButtonKeydown(e: React.KeyboardEvent<HTMLElement>) {
+        e.stopPropagation();
+    }
+
+    private handleClick(e: React.MouseEvent<HTMLElement>) {
+        const { githubId } = this.props;
+        if (!githubId) {
+            this.createRepositoryAsync().done();
```
riknoll: Why `.done()`?

webapp/src/filelist.tsx
```diff
@@ -253,21 +254,102 @@ namespace custom {
...
+        if (!pxt.github.token) {
+            pxt.tickEvent("github.filelist.create.notoken");
+            return;
+        }
+
+        const repoid = await dialogs.showCreateGithubRepoDialogAsync(this.props.parent.state.projectName);
+        if (!repoid) return;
+
+        pxt.tickEvent("github.filelist.create.export");
+        core.showLoading("creategithub", lf("creating {0} repository...", pxt.github.parseRepoId(repoid).fullName))
```
riknoll: Capitalization

webapp/src/filelist.tsx
riknoll: capitalization

webapp/src/gitjson.tsx
```diff
@@ -0,0 +1,693 @@
...
+        if (!ok)
+            return
+
+        let newv = vpatch;
+        if (bumpType == "major")
+            newv = vmajor;
+        else if (bumpType == "minor")
+            newv = vminor;
+        const newVer = pxt.semver.stringify(newv)
+        core.showLoading("bumpheader", lf("creating release {0}...", newVer));
```
riknoll: Capitalize

webapp/src/gitjson.tsx
```diff
@@ -0,0 +1,693 @@
...
+            this.parent.setState({});
+            return
+        } else {
+            this.previousCfgKey = newKey
+            await this.parent.reloadHeaderAsync();
+        }
+    }
+
+    private async pullAsync() {
+        core.showLoading("loadingheader", lf("pulling changes..."));
```
riknoll: Capitalize

webapp/src/gitjson.tsx
```diff
@@ -0,0 +1,693 @@
...
+            <div key={f.name} className="ui segments filediff">
+                <div className="ui segment diffheader">
+                    <span>{f.name}</span>
+                    <sui.Button className="small" icon="undo" text={lf("Revert")}
+                        ariaLabel={lf("Revert file")} title={lf("Revert file")}
+                        textClass={"landscape only"} onClick={cache.revert} />
+                    {deletedFiles.length == 0 ? undefined :
+                        <p>
+                            {lf("Reverting this file will also restore:")}
+                            {" "}
```
riknoll: Nitpick: combine with previous line `+ " " `

webapp/src/gitjson.tsx
```diff
@@ -0,0 +1,693 @@
...
+            <div id="githubArea">
+                <div id="serialHeader" className="ui serialHeader">
+                    <div className="leftHeaderWrapper">
+                        <div className="leftHeader">
+                            <sui.Button title={lf("Go back")} icon="arrow left" text={lf("Go back")} textClass="landscape only" tabIndex={0} onClick={this.goBack} onKeyDown={sui.fireClickOnEnter} />
+                        </div>
+                    </div>
+                    <div className="rightHeader">
+                        <sui.Button icon={`${this.needsPull !== false ? "down arrow" : "check"} ${this.needsPull !== false ? "positive" : ""}`}
+                            text={lf("Pull changes")} textClass={lf("landscape only")} title={lf("Pull changes from GitHub to get your code up-to-date.")} onClick={this.handlePullClick} onKeyDown={sui.fireClickOnEnter} />
```
riknoll: Looks like you're translating the css class for the text

webapp/src/gitjson.tsx
```diff
@@ -0,0 +1,693 @@
...
+        const gs = this.getGitJson();
+        // don't use gs.prUrl, as it gets cleared often
+        const url = `https://github.com/${githubId.fullName}${master ? "" : `/tree/${githubId.tag}`}`;
+        const needsToken = !pxt.github.token;
+        // this will show existing PR if any
+        const prUrl = !gs.isFork && master ? null :
+            `https://github.com/${githubId.fullName}/compare/${githubId.tag}?expand=1`
+        this.needsLicenseMessage = gs.commit && !gs.commit.tree.tree.some(f =>
+            /^LICENSE/.test(f.path.toUpperCase()) || /^COPYING/.test(f.path.toUpperCase()))
+        return (
```
riknoll: I think this needs to be broken up into several React components

webapp/src/gitjson.tsx
pelikhan: all loading message should be lower cased; just like "loading editor..." or "loading...".

webapp/src/gitjson.tsx
pelikhan: all loading message should be lower cased; just like "loading editor..." or "loading...".

webapp/src/gitjson.tsx
pelikhan: unified into translatable string.

webapp/src/filelist.tsx
pelikhan: all loading message should be lower cased; just like "loading editor..." or "loading...".

webapp/src/filelist.tsx
pelikhan: all loading message should be lower cased; just like "loading editor..." or "loading...".

webapp/src/dialogs.tsx
pelikhan: all loading message should be lower cased; just like "loading editor..." or "loading...".

webapp/src/core.ts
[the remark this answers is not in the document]
pelikhan: Does not seem to exist in our version of React.

webapp/src/filelist.tsx
pelikhan: we tried adding an icon (down arrow) but semantic and the menu CSS insists on docking the icon to the right of the view. Visually, this puts the arrow all the way to the right which is near invisible. Thus keeping it as unicode text right next to the file name.

webapp/src/coretsx.tsx
```diff
@@ -141,6 +148,11 @@ export function dialogIsShowing() {
     return !!currentDialog;
 }
 
+export function forceRender() {
+    if (currentDialog)
+        currentDialog.setState({});
```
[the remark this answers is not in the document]
pelikhan: renamed to forceUpdate

webapp/src/app.tsx
```diff
@@ -812,13 +844,18 @@ export class ProjectView
...
```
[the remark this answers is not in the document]
pelikhan: Added a comment. it's mostly the price to pay to mix in react vs non-react stuff

webapp/src/filelist.tsx
pelikhan: Purely a style gimmic. It clearly state that no one is checking or waiting on this promise; an you cannot accidently return it.

pelikhan:
The problem is that editor are not React components. They all cheat react and manipulate the DOM directly. This editor has a lot of state maintained in the .git.json file so forceRender seems like the most efficient way to get things done.
Typically these force render calls are done after expensive calls and loading dialogs anyway.
Not the best but cleaning up the editor/react story would make this PR touch way more critical code.

webapp/src/gitjson.tsx
pelikhan: this one does not need to be in the state, made it a local var.

pelikhan:
@riknoll moved most of the UI into its own React component. Still going through the weird editor indirection.

PR #5917, opened 2019-09-18
constant folding

mmoskal:
fixes https://github.com/microsoft/pxt-arcade/issues/553


libs/pxt-common/pxt-helpers.ts
```diff
@@ -1,5 +1,29 @@
...
+ * Constant representing Not-A-Number.
+ */
+const NaN = 0 / 0
+
+/**
+ * Constant representing positive infinity.
+ */
+const Infinity = 1 / 0
+
+function isNaN(x: number) {
```
riknoll: `isNaN` only checks if it is `NaN` or not. I think this would return true for `null` and `undefined` right? BTW the semantics for the + operator are hilarious:

```
> +null
0
> +undefined
NaN
> +""
0
> +"1"
1
> +"'"
NaN
> +[]
0
> +[""]
0
> +["", ""]
NaN
```

mmoskal: The `Number.isNaN` is somewhat more sane. I think the `+x` is what JavaScript does, but I'm not sure we do the right thing for all values for `+`...

mmoskal: Maybe we should have `isNaN` be like `Number.isNaN`? I.e., return true only for actual NaN?

riknoll: That's what it does in the browser, so sounds good to me!

mmoskal:
```
> isNaN(undefined)
true
> Number.isNaN(undefined)
false
> Number.isNaN("x")
false
> isNaN("x")
true
```
I believe with this patch we do the same, but I'm wondering if dropping JS compat here wouldn't be a good idea...

PR #5921, opened 2019-09-18
Github explorer PR feedback

pelikhan:
image
image

- [x] remove dots in create github dialog
- [x] add sentence about github in create dialog + link
- [x] don't disable commit/create release and instead pop get token dialog
- [x] move github file item in own tree
- [x] create private repoes
- [x] don't change preferredEditor when going to github view

Not addressed in this PR:
- [ ] revert delete files should bring back the files

### DIFF
--- a/webapp/src/dialogs.tsx
+++ b/webapp/src/dialogs.tsx
@@ -597,8 +597,8 @@ export function showCreateGithubRepoDialogAsync(name?: string) {
         }
     }
 
-    function onPublicChanged(k: string) {
-        const v = k == "checked";
+    function onPublicChanged(v: boolean) {
+        v = !!v;
         if (repoPublic != v) {
             repoPublic = v;
             coretsx.forceUpdate();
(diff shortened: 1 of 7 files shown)


webapp/src/dialogs.tsx
```diff
@@ -596,17 +597,32 @@ export function showCreateGithubRepoDialogAsync(name?: string) {
...
+                </p>
                 <div className="ui field">
-                    <sui.Input type="url" value={repoName} onChange={onNameChanged} label={lf("Repository name.")} placeholder={`pxt-my-gadget...`} class="fluid" error={nameErr} />
+                    <sui.Input type="url" value={repoName} onChange={onNameChanged} label={lf("Repository name")} placeholder={`pxt-my-gadget...`} class="fluid" error={nameErr} />
                 </div>
                 <div className="ui field">
-                    <sui.Input type="text" value={repoDescription} onChange={onDescriptionChanged} label={lf("Repository description.")} placeholder={lf("MakeCode extension for my gadget")} class="fluid" />
+                    <sui.Input type="text" value={repoDescription} onChange={onDescriptionChanged} label={lf("Repository description")} placeholder={lf("MakeCode extension for my gadget")} class="fluid" />
                 </div>
+                <sui.Checkbox checked={repoPublic} inputLabel={repoPublic ? lf("Public repository, anyone can look at your code.") : lf("Private repository, your code is only visible to you.")} onChange={onPublicChanged} />
```
mmoskal: I think this may be confusing that the description changes when you check/uncheck, I've never seen that in a UI.

PR #5924, opened 2019-09-19
github improvements

pelikhan:
- [x] don't show warnings when user is not signed in
- [x] don't expand explorer when coming from blocks
- [x] moving license message under create release
- [x] adding notch on branch to indicate menu
image
- [x] use dropdown instead of toggle for publi/cprivate
image

### DIFF
--- a/webapp/src/dialogs.tsx
+++ b/webapp/src/dialogs.tsx
@@ -597,8 +597,8 @@ export function showCreateGithubRepoDialogAsync(name?: string) {
         }
     }
 
-    function onPublicChanged(v: boolean) {
-        v = !!v;
+    function onPublicChanged(e: React.ChangeEvent<HTMLSelectElement>) {
+        let v = e.currentTarget.selectedIndex == 0;
         if (repoPublic != v) {
             repoPublic = v;
             coretsx.forceUpdate();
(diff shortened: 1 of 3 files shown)


webapp/src/gitjson.tsx
```diff
@@ -693,6 +684,17 @@ class NoChangesComponent extends sui.StatelessUIElement<GitHubViewProps> {
                     </p>
                     <sui.Button className="primary" text={lf("Create release")} onClick={this.handleBumpClick} onKeyDown={sui.fireClickOnEnter} />
                 </div> : undefined}
+            {master && needsLicenseMessage ? <div className="ui message">
+                <div className="content">
+                    {lf("Your project doesn't seem to have a license. This makes it hard for others to use it.")}
+                    {" "}
```
jwunderl: could this be combined with previous line?

webapp/src/dialogs.tsx
```diff
@@ -597,8 +597,8 @@ export function showCreateGithubRepoDialogAsync(name?: string) {
         }
     }
 
-    function onPublicChanged(v: boolean) {
-        v = !!v;
+    function onPublicChanged(e: React.ChangeEvent<HTMLSelectElement>) {
+        let v = e.currentTarget.selectedIndex == 0;
```
jwunderl: const
```suggestion
        const v = e.currentTarget.selectedIndex == 0;
```

webapp/src/gitjson.tsx
pelikhan: no it messes up translations.

PR #5927, opened 2019-09-19
local branch of Print Dropdown for Docs (Outlines/Color)

pelikhan:
See https://github.com/microsoft/pxt/pull/5515#


docfiles/docs.js
```diff
@@ -189,9 +189,24 @@ function setupSemantic() {
...
 
-    $('#printbtn').on("click", function () {
-        window.print();
-    })
+    $('#printbtn').dropdown({
+        action: function(text, value) {
+        
+            switch (value) {
+                case 'color': 
+                document.querySelector('body').classList.add("print-color");
```
jwunderl:
```suggestion
                    document.querySelector('body').classList.add("print-color"); 
```
nit: indentation

docfiles/docs.js
```diff
@@ -189,9 +189,24 @@ function setupSemantic() {
...
-    })
+    $('#printbtn').dropdown({
+        action: function(text, value) {
+        
+            switch (value) {
+                case 'color': 
+                document.querySelector('body').classList.add("print-color");                      
+                    break;
+                case 'bw':
+                document.querySelector('body').classList.remove("print-color");
```
jwunderl:
```suggestion
                    document.querySelector('body').classList.remove("print-color");
```

nit: indentation

docfiles/docs.js
```diff
@@ -189,9 +189,24 @@ function setupSemantic() {
         }).appendTo(outer);
     });
 
-    $('#printbtn').on("click", function () {
-        window.print();
-    })
+    $('#printbtn').dropdown({
+        action: function(text, value) {
```
jwunderl: this doesn't seem to activate on space bar, only enter key

docfiles/printbtn.html
```diff
@@ -0,0 +1,11 @@
+<div id="printbtn" class="circular ui icon top right floated pointing dropdown button hideprint" title="Print this page">
+    <i class="icon print"></i>
+    <div class="menu">
+    <div class="item" data-value="color">
```
jwunderl: nit: indendation here

docfiles/docs.js
darzu: Is this still the case?

jwunderl: Yup, still won't activate on space

darzu: I took a stab at fixing this but couldn't find a working solution. It's even broken on semantic UI's demo page: https://semantic-ui.com/modules/dropdown.html#/examples
I vote we punt this to another work item

darzu:
Things look broken after the new docs update:
image

and selecting "color" didn't print with color. Can't merge this as-is.

PR #5933, opened 2019-09-20
Telemetry fix

abchatra:
This removes all the dynamic properties as  part of our telemetry

What we used to emit
EntityName : activity
PropertyName  :  blocks.create.forever
PropertyValue : 25

New telemetry
EntityName : block.create
PropertyName  :  block
PropertyValue : "blocks.create.forever"

Individual events are sent as part of count. We could add separate count PropertyName to the existing blocks.create as well.

### DIFF
--- a/webapp/src/blocks.tsx
+++ b/webapp/src/blocks.tsx
@@ -385,13 +385,9 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         });
 
         for (const block in deprecatedMap) {
-            if (deprecatedMap[block] === 0) {
-                delete deprecatedMap[block];
+            if (deprecatedMap[block] !== 0) {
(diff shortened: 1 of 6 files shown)


webapp/src/blocks.tsx
```diff
@@ -385,14 +385,10 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         });
 
         for (const block in deprecatedMap) {
-            if (deprecatedMap[block] === 0) {
-                delete deprecatedMap[block];
+            if (deprecatedMap[block] !== 0) {
+                pxt.tickEvent("blocks.usingDeprecated", {block : block, count : deprecatedMap[block] });
```
pelikhan: typically lowercased?

abchatra: I didn't change the tick name. Kept is as before. I can update next round

PR #5937, opened 2019-09-22
Implement inline-diff for single line changes

mmoskal:
This also removes hover highlight. Github doesn't do it, and it seem too much together with inline diff markers.

The inline diff only kicks in for single-line changes (that a single line is updated, with at least one unchanged line above and below).


mmoskal:
Hmm... it seems github does it also when several lines are updated (but none added or removed, as far as I can tell).

PR #5940, opened 2019-09-24
Add 'flyout only' and 'hide iteration' options to tutorial metadata

shakao:
For Minecraft hour of code tutorials.


webapp/src/blocks.tsx
```diff
@@ -930,7 +933,8 @@ export class Editor extends toolboxeditor.ToolboxEditor {
 
         this.clearCaches();
 
-        const hasCategories = this.shouldShowCategories();
+        let forceFlyoutOnly = this.parent.state.editorState && this.parent.state.editorState.hasCategories === false;
```
jwunderl:
```suggestion
        const forceFlyoutOnly = this.parent.state.editorState && this.parent.state.editorState.hasCategories === false;
```

shakao:
CSS changes for minecraft pxt-minecraft/pull/1234

PR #5942, opened 2019-09-25
Adding lock to old issues

abchatra:
Here is the app: https://github.com/apps/lock/


abchatra:
I will move this change to other repositories if this works. I kept old issues to default 365 days.

pelikhan:
Did you install the app?

abchatra:
It said it is enabled for Microsoft repos

PR #5945, opened 2019-09-25
adds meaningless doc change for the purpose of testing our docs versioning

darzu:
@riknoll , should this test the new fix from you and Michal?


darzu:
Shows up on live at:
image
image

But not in staging:
image
image

So LGTM!

PR #5950, opened 2019-09-26
handle field redefinitions

mmoskal:
* compile redefinition of field with a property correctly
* trigger full recompile (instead of incremental) when methods or fields turn virtual

fixes https://github.com/microsoft/pxt-arcade/issues/1195
fixes https://github.com/microsoft/pxt-arcade/issues/1351


pxtcompiler/emitter/service.ts
```diff
@@ -1185,6 +1185,11 @@ namespace ts.pxtc.service {
                     host.opts.skipPxtModulesEmit = true
             }
             res = compile(host.opts, service);
+            if (res.needsFullRecompile) {
+                pxt.log("trigering full recompile")
```
pelikhan: add a tickEvent?

PR #5951, opened 2019-09-26
Crowdin stats upgrade

pelikhan:
- [x] fix branch filtering
- [x] added crowdin file id
- [x] drap all files into single .csv file


pelikhan:
@abchatra you'll need this fix to run the translation stats for editors.

PR #5968, opened 2019-10-01
More accessibility fixes

jwunderl:
Fixes some microbit accessibility bugs

microsoft/pxt-microbit#2330
microsoft/pxt-microbit#2301
microsoft/pxt-microbit#2302 << popout here was from the old lightbox; there's some more unused code relating to that, but leaving it for now / clean up separately
microsoft/pxt-microbit#2329 << thsi is only used in the embed dialog and the icon that goes in the place of the blocks / js toggle when in debug mode
microsoft/pxt-microbit#2334
microsoft/pxt-microbit#2292 ~<< the @red is consistently light against white with the color palettes we've used~ moving this over to microbit PR, no reason to override globally for this


webapp/src/share.tsx
```diff
@@ -582,6 +582,6 @@ class EmbedMenuItem extends sui.StatelessUIElement<EmbedMenuItemProps> {
 
     renderCore() {
         const { label, mode, currentMode } = this.props;
-        return <sui.MenuItem id={`tab${mode}`} active={currentMode == mode} name={label} onClick={this.handleClick} />
+        return <sui.MenuItem id={`tab${mode}`} active={currentMode == mode} tabIndex={0} name={label} onClick={this.handleClick} />
```
pelikhan: MenuItem could set tabIndex={0} by default if not specified?

webapp/src/sui.tsx
```diff
@@ -817,7 +819,17 @@ export class MenuItem extends data.Component<MenuItemProps, {}> {
         }
 
         return (
-            <div id={id} tabIndex={active ? 0 : -1} className={classes} onClick={this.handleClick} role="tab" aria-controls={ariaControls} aria-selected={active} aria-label={content || name}>
+            <div
+                id={id}
+                tabIndex={tabIndex !== null ? tabIndex : -1}
```
pelikhan: what happens with undefined?

jwunderl: ahh good catch

PR #5973, opened 2019-10-02
remove Times New Roman

darzu:
Not sure why we would want Times New Roman. Looks bad and everyone's first instinct is that something broke:
image

after:
image

maybe I'm missing something


pelikhan:
All the variables in docs.less need to go in site.variables.

abchatra:
MakeCode.com has colors inverted for the top bar text and they are not black 
image

PR #5991, opened 2019-10-03
Multiple tutorials per repo

pelikhan:
- [x] support path to load any markdown file in repo
- [x] automatically pick up localized version of file
- [x] add refernce to tutorial repo if it contains code
- [x] add docs


mmoskal:
I think the cloud (gh proxy) might support more than one package per repo, need to check...

pelikhan:
Having multiple files is actually easier to handle when editing.

mmoskal:
just FYI the cloud api allows for subfolders, like here: https://pxt.io/api/gh/microsoft/pxt-microbit/master/libs/hello or https://pxt.io/api/gh/microsoft/pxt-common-packages/master/libs/color-sensor

PR #5994, opened 2019-10-03
Add lifecycle functions before and after tutorial code

shakao:
For the hour of code, the Minecraft team wants to be able to send a command when the code in "onStart" starts, and when it ends. This generates codeStart/codeStop files for a tutorial project & inserts them before and after main.ts. The files call "onCodeStart"/"onCodeStop", which are implemented by the target.

### DIFF
--- a/docs/static/playground/pxt-common/pxt-core.d.js
+++ b/docs/static/playground/pxt-common/pxt-core.d.js
@@ -526,3 +526,11 @@ declare namespace Math {
     //% shim=Math_::idiv
     function idiv(x: number, y: number): number;
 }
+
+declare namespace _lifecycle {
+    //% shim=__lifecycle::onCodeStart
+    export function onCodeStart(arg: any): void;
+
+    //% shim=__lifecycle::onCodeStop
+    export function onCodeStop(arg: any): void;
+}
--- a/docs/static/playground/pxt-common/pxt-core.d.js
+++ b/docs/static/playground/pxt-common/pxt-core.d.js
@@ -527,10 +527,10 @@ declare namespace Math {
     function idiv(x: number, y: number): number;
 }
 
-declare namespace _lifecycle {
-    //% shim=__lifecycle::onCodeStart
-    export function onCodeStart(arg: any): void;
+declare namespace control {
+    //% shim=_control::_onCodeStart
+    export function _onCodeStart(arg: any): void;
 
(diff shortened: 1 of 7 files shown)


docs/static/playground/pxt-common/pxt-core.d.js
```diff
@@ -526,3 +526,11 @@ declare namespace Math {
     //% shim=Math_::idiv
     function idiv(x: number, y: number): number;
 }
+
+declare namespace _lifecycle {
+    //% shim=__lifecycle::onCodeStart
+    export function onCodeStart(arg: any): void;
```
pelikhan: we should move this namespace those under "control" where most of the runtime function exist.

pxtcompiler/emitter/driver.ts
```diff
@@ -333,6 +340,14 @@ namespace ts.pxtc {
             tsFiles.push("main.ts")
             hasMain = true;
         }
+
+        // run post-processing code last, if present
+        const post_idx = tsFiles.indexOf("_onCodeStop.ts")
+        if (post_idx >= 0) {
+            tsFiles.splice(post_idx, 1)
+            tsFiles.push("_onCodeStop.ts")
```
pelikhan: what happens when you have events?

pelikhan:
When happens when the code has events or forever, onCodeStop won't be correct.

pxtcompiler/emitter/driver.ts
shakao: in the pxt-minecraft side, we only send the "onCodeStop" message if no events are registered. basically the minecraft team wants: 1. always send an event before onStart and 2. send "onCodeEnd" after the onStart code, only if no events are registered. i think codeStart might be useful for us in other scenarios but codeStop feels minecraft-HOC-specific to me, since there's no real concept of "code stop" if events are registered

i could rename these to beforeMain/afterMain or something to be clearer? or move codeStop somewhere else maybe...

pxtcompiler/emitter/driver.ts
```diff
@@ -124,6 +124,13 @@ namespace ts.pxtc {
             opts.sourceFiles.push("main.ts")
         }
 
+        // run post-processing code last, if present
+        const post_idx = opts.sourceFiles.indexOf("_onCodeStop.ts")
```
riknoll: Nitpick: lose the underscore in the variable name (also below)

Issue #5987, opened 2019-10-02
Docs 2.0: icons above code snippets


darzu:
Instead of:

![image](https://user-images.githubusercontent.com/6453828/66087872-dddc4480-e52d-11e9-87e9-d613b62b2909.png)

PR #6000, opened 2019-10-04
snippets styling for new docs

jwunderl:
re: https://github.com/microsoft/pxt/issues/5987

move buttons above snippet and subheaders to the right

updated:

also minor patch to headers; a few pages have some weird headers with no content in them just to add some extra targets (e.g. https://github.com/microsoft/pxt/blob/master/common-docs/blocks/on-start.md), which can cause a slightly annoying double line; the h2 changes in docs.less just makes it so those get collapsed when there is more than a single consecutive h2s with no content in between them, and that the h2s otherwise always get a border unless they are the very first thing on the page. Can separate that out if preferred


abchatra:
Test with python enabled as well

jwunderl:
here's with python enabled on https://arcade.makecode.com/reference/game/over

pxtrunner/renderer.ts
```diff
@@ -55,13 +51,17 @@ namespace pxt.runner {
         }
     }
 
+    function appendBlocks($parent: JQuery, $svg: JQuery) {
+        $parent.append($(`<div class="ui content blocks"/>`).append($svg));
+    }
+
     function appendJs($parent: JQuery, $js: JQuery, woptions: WidgetOptions) {
-        $parent.append($('<div class="ui content js"><div><i class="ui icon xicon js"/>JavaScript</div></div>').append($js));
+        $parent.append($(`<div class="ui content js"><div class="subheading"><i class="ui icon xicon js"/>${lf("JavaScript")}</div></div>`).append($js));
```
pelikhan: we are not localizing JavaScript in the app.

pxtrunner/renderer.ts
```diff
@@ -55,13 +51,17 @@ namespace pxt.runner {
...
         highlight($js);
     }
 
     function appendPy($parent: JQuery, $py: JQuery, woptions: WidgetOptions) {
-        $parent.append($('<div class="ui content py"><div><i class="ui icon xicon python"/>Python</div></div>').append($py));
+        $parent.append($(`<div class="ui content py"><div class="subheading"><i class="ui icon xicon python"/>${lf("Python")}</div></div>`).append($py));
```
pelikhan: No localizing Python in the app

jwunderl:
Ahh that makes sense, I thought I saw a comment to localize it in the much reverted PR but I might have read it backwards

PR #6005, opened 2019-10-04
refactor logic to show/hide monaco toolbox

pelikhan:
If you navigate to main.ts, then README.md, you will have a monaco toolbox. The reason is that the logic to hide/show is done in two location and it is different. This PR does a bit of cleanup to ensure we have 1 way of deciding to show/hide.

### DIFF
--- a/webapp/src/monaco.tsx
+++ b/webapp/src/monaco.tsx
@@ -190,6 +190,19 @@ class FormattingProvider implements monaco.languages.DocumentRangeFormattingEdit
     }
 }
 
+const modeMap: pxt.Map<pxt.editor.FileType> = {
+    "cpp": pxt.editor.FileType.CPP,
+    "h": pxt.editor.FileType.CPP,
+    "json": pxt.editor.FileType.JSON,
+    "md": pxt.editor.FileType.Markdown,
+    "py": pxt.editor.FileType.Python,
+    "ts": pxt.editor.FileType.TypeScript,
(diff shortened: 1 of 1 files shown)


webapp/src/monaco.tsx
```diff
@@ -903,6 +917,18 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
 
+    private shouldShowToolbox(): boolean {
+        const readOnly =
+            !this.currFile
+            || this.currFile.isReadonly()
+            || pxt.shell.isReadOnly()
+            || this.isDebugging();
+        return pxt.appTarget.appTheme.monacoToolbox
+            && !readOnly
+            && ((this.fileType == "typescript" && this.currFile.name == "main.ts")
```
jwunderl: Is there a reason to only show for `main.[ts|py]`? That's the current behavior so no reason to hold up this PR on it, but when I'm using the github integration to make a game sometimes I wish I had the toolbox to look up the js equivalent for a block I rarely use / not just have to guess it with intellisense - might be nice to have it show whenever it's editable and js/py

pelikhan: Once you are beyond the default main.ts/.py, you are advanced enough that the toolbox is just annoying and eating up space. Intellisense is more than enough.

pelikhan: Can be negotiated :)

PR #6008, opened 2019-10-04
fix block snippet svg duplicate IDs

jwunderl:
re: https://github.com/microsoft/pxt-microbit/issues/2279; each blocks snippet is duplicating the ids it uses, and on top of that some of the definitions are being duplicated in the ``.injectionDiv`` used for the blockly workspace. This doesn't make a visible difference for the rendering (all the definitions are identical, so the browser implementation's decision of which one gets used doesn't matter at all), but can cause issues for screen readers.

I'm not convinced this is the best solution, but it resolves the issue / I just want to see if anyone immediately knows a more proper way to fix this. The main changes are just disposing of the blockly workspace after rendering the final element (gets rid of ``.injectionDiv``), and patching the svg's as they are rendered to remove the duplicate filters / etc from their definitions.


pelikhan:
ready for review?

jwunderl:
It fixes the problem but feels a hacky / potentially fragile to me. I haven't though of a better place to fix it yet, though, so ready for review

pxtrunner/renderer.ts
```diff
@@ -219,24 +219,41 @@ namespace pxt.runner {
...
-                    errors.forEach(diag => pxt.reportError("docs.decompile", "" + diag.messageText, { "code": diag.code + "" }));
-                render(el, r);
-                el.removeClass("lang-shadow");
-                return consumeRenderQueueAsync();
-            }).catch(e => {
-                pxt.reportException(e);
-                el.append($('<div/>').addClass("ui segment warning").text(e.message));
-                el.removeClass("lang-shadow");
-                return consumeRenderQueueAsync();
+        const existingFilters: Map<Element> = {};
```
pelikhan: do you need to hold a ref to that Element? it is rooting it.

jwunderl: I think I'm missing something here; this exists only in the scope of ``consumeRenderQueueAsync``, and will be gc-ed after all blocks are rendered. No specific reason to ref the element itself though I suppose as I really just need a set, can just do `Map<boolean>`

PR #6014, opened 2019-10-05
Fix python rendering in doc snippets

jwunderl:
pretty colors for python (and javascript that is appended)

pulls in updated hljs from https://highlightjs.org/download/, as the current one didn't have python included. for reference, this one now includes `JavaScript`, `HTML,XML`, `Python`, `Markdown`, and `TypeScript`; if we ever want anything else e.g. json need to pack it again.


pelikhan:
Could you add a quick note in README on how to update highlight.js?

PR #6016, opened 2019-10-05
in context localization for markdown files

pelikhan:
This change adds support for rendering localized markdown files in localhost (downloaded from our cloud) and support for incontext translations using crowdin.
Required:
- [x] move all the crowdin stuff into its own file
- [ ] backend support for "?translate=1" route
- [ ] backend support for @crowdinproject@ macro

Journey: user clicks on "translate this page" link at the bottom of docs page
In Context translations are loaded
image
image


cli/server.ts
```diff
@@ -1102,26 +1111,33 @@ export function serveAsync(options: ServeOptions) {
...
-            const mdopts = <pxt.docs.RenderOptions>{
-                template: expandDocFileTemplate("docs.html"),
-                markdown: md,
-                theme: pxt.appTarget.appTheme,
-                filepath: pathname,
-                TOC: resolveTOC(pathname)
-            };
-            let html = pxt.docs.renderMarkdown(mdopts)
-            sendHtml(html, U.startsWith(md, "# Not found") ? 404 : 200)
+            const lang = opts["translate"] ? "pxt" : opts["lang"] as string;
```
pelikhan: @darzu backend support needed for this route

cli/server.ts
```diff
@@ -1102,26 +1111,33 @@ export function serveAsync(options: ServeOptions) {
...
+                        markdown: md,
+                        theme: pxt.appTarget.appTheme,
+                        filepath: pathname,
+                        TOC: resolveTOC(pathname),
+                        pubinfo: {
+                            locale: lang
+                        }
+                    };
+                    if (opts["translate"])
+                        mdopts.pubinfo["incontexttranslations"] = "1";
```
pelikhan: @darzu backend support to enable this macro

pxtlib/emitter/cloud.ts
```diff
@@ -125,7 +125,7 @@ namespace pxt.Cloud {
             url = `md/${pxt.appTarget.id}/${docid.replace(/^\//, "")}?targetVersion=${encodeURIComponent(targetVersion)}`;
         }
         if (!packaged && locale != "en") {
-            url += `&lang=${encodeURIComponent(Util.userLanguage())}`
+            url += `&lang=${encodeURIComponent(locale)}`
```
pelikhan: this was a silent bug

docfiles/footer.html
```diff
@@ -22,6 +22,9 @@
             <a class="item" href="https://makecode.com/privacy" target="_blank" rel="noopener">Privacy &amp; Cookies</a>
             <a class="item" href="https://makecode.com/termsofuse" target="_blank" rel="noopener"> Terms Of Use</a>
             <a class="item" href="https://makecode.com/trademarks" target="_blank" rel="noopener">Trademarks</a>
+            <!-- @ifdef crowdinproject -->
```
pelikhan: @darzu need support for this macro value

docfiles/docs.html
```diff
@@ -6,6 +6,13 @@
 	<title>@name@</title>
 	<meta name="Description" content="@description@" />
 
+	<!-- @ifdef incontexttranslations -->
+	<script type="text/javascript">
+		var _jipt = [];
+		_jipt.push(['project', '@crowdinproject@']);
```
pelikhan: need support for @crowdinproject@

cli/server.ts
darzu: So if I'm understanding this right, one of the backend changes needed is simply:
If request comes in with translate=1 q.s., then set "incontexttranslations" in pubinfo

It looks like the backend uses renderMarkdown from backend utils which has logic to turn on/off sections via @ifdef

pelikhan: Yes correct. The backends need to set incontexttranslations is translate == 1 and use the "pxt" locale

PR #6022, opened 2019-10-06
Translations in context in editor

pelikhan:
Integration of the Crowdin In Context support in MakeCode. 

* not supported on ie
* live test (gearwheel -> language -> translate the editor): https://makecode.microbit.org/app/eafdb91f42807c21c8fe0214992be13211550d26-f42e68aeb6

image

Add a menu item in blocks context menu to translate.
image

Crowdin has a nifty web widget that hooks into the DOM mutation events to provide live translation. It uses a fake language (code pxt) that contains guid strings; when crowding sees one of the guid added to the DOM tree, it swaps it with the current translation + button to edit.

Things get a bit more tricker for our blocs -- because we parse the translated string. Obviously parsing the GUID does not work. To work around this, MakeCode first add each block string into the DOM tree, let's crowdin modify it and then parses the block. It mostly works

No backend support needed.

### DIFF
--- a/docfiles/docs.html
+++ b/docfiles/docs.html
@@ -12,7 +12,7 @@
 		var _jipt = [];
 		_jipt.push(['project', '@crowdinproject@']);
 		</script>
-		<script type="text/javascript" src="//cdn.crowdin.com/jipt/jipt.js"></script>
+	<script type="text/javascript" src="//cdn.crowdin.com/jipt/jipt.js"></script>
 	<!-- @endif -->
 	<!-- @include meta.html -->
 	<!-- @include head.html -->
--- a/docfiles/docs.html
+++ b/docfiles/docs.html
@@ -11,6 +11,9 @@
 	<script type="text/javascript">
 		var _jipt = [];
 		_jipt.push(['project', '@crowdinproject@']);
+        _jipt.push(['escape', function() {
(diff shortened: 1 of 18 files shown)


docfiles/docs.html
```diff
@@ -11,8 +11,11 @@
 	<script type="text/javascript">
 		var _jipt = [];
 		_jipt.push(['project', '@crowdinproject@']);
-		</script>
-		<script type="text/javascript" src="//cdn.crowdin.com/jipt/jipt.js"></script>
+		_jipt.push(['escape', function () {
+			window.location.href = window.location.href.replace(/[$\?]translate=1/, '');
```
darzu: will this handle "?foo=1&translate=1"  ?

pxtlib/blocks.ts
```diff
@@ -721,5 +726,25 @@ namespace pxt.blocks {
                 message0: Util.lf("continue")
             }
         }
+
+        if (pxt.Util.isTranslationLanguage()) {
```
darzu: Is IE11 blocked from this code path?

docfiles/docs.html
pelikhan: yes this will handle it

pxtlib/blocks.ts
pelikhan: blocking upstream

PR #6029, opened 2019-10-08
Allow one step tutorials, strip metadata before parsing, tests

shakao:
Stripping metadata to ensure parsing of legacy tutorials (### Step) is supported


shakao:
@pelikhan Nope, this should preserve existing behavior

PR #6032, opened 2019-10-08
Docs horizontal spacing

jwunderl:
make horizontal spacing prettier

Eventually plan is to move most of the spacing to use grid (then we can do stuff like make the sidebar not bump the content if there is space), but waiting on doing more until sidebar / top bar are more stable as it would otherwise conflict with other work at the moment.


docfiles/style.css
```diff
@@ -383,17 +356,40 @@ svg {
...
-#docs .ui.container.mainbody {
+.article-inner {
+    -ms-grid-column: 2;
+    grid-column: 2;
+    padding-left: 0;
+}
+
+/* Limit width of content so that it is readable */
+#docs .article .article-inner div {
+    max-width: 1100px !important;
```
shakao: just checking, all of these !importants are still necessary?

jwunderl: for now at least, mostly just from fighting with SUI. Should be able to get rid of more in a future pr that moves more of the content into a grid, but will have to see how that works out

PR #6037, opened 2019-10-08
pxtconfig is null

pelikhan:
pxtConfig can be undefined or null (localhost).


abchatra:
I guess not. Makes sense.

pelikhan:
I kept it just in case it is not defined for some reason.

PR #6046, opened 2019-10-09
upgrade to typescript 3.6; turn on incremental compilation

darzu:
This decouples the version of typescript used by our compiler compared to the version we use to build our editor.

Our editor is now on TS 3.6 while the compiler continues to use TS 2.6.

Speed ups are about 1.9x for the webapp, and 1.2x for the whole build.

We can get much better speed ups if we remove the Jake nonsense as this effectively hamstrings the incremental compiler by chopping up our compilation into little pieces.


pxtcompiler/ext-typescript/readme.md
```diff
@@ -0,0 +1 @@
+This compiled version of typescript was taken from 2.6.1 npm package
```
darzu: @mmoskal , does this seem sane? I included the compiled version of TS 2.6 directly in the pxtcompiler so that it's decoupled from the typescript version used to build our editor, which lets us upgrade to TS 3.6

darzu: Another way to do this would be to check in typescript's source (.ts), but it gets a little weird compiling an old version with itself

mmoskal: looks good!

webapp/src/greenscreen.tsx
```diff
@@ -91,10 +91,6 @@ export class WebCam extends data.Component<WebCamProps, WebCamState> {
     private stop() {
         this.deviceId = undefined;
         if (this.stream) {
-            try {
-                if (this.stream.stop)
-                    this.stream.stop();
```
mmoskal: this wasn't needed?

pxtblocks/blocklylayout.ts
```diff
@@ -338,9 +341,10 @@ namespace pxt.blocks.layout {
 
         if (!BrowserUtils.isEdge()) return Promise.resolve();
 
-        const images = xsg.getElementsByTagName("image") as NodeListOf<Element>;
+        const images = xsg.getElementsByTagName("image")
         const p = pxt.Util.toArray(images)
             .filter(image => /^data:image\/svg\+xml/.test(image.getAttributeNS(XLINK_NAMESPACE, "href")))
+            .map(img => img as unknown as HTMLImageElement)
```
jwunderl: should be able to bump this up to the the first part to avoid a runtime cost just to make the type system happy `(pxt.Util.toArray(images) as HTMLImageElement[])`

pxtlib/browserutils.ts
```diff
@@ -260,10 +260,13 @@ namespace pxt.BrowserUtils {
     export function devicePixelRatio(): number {
         if (typeof window === "undefined" || !window.screen) return 1;
 
-        if (window.screen.systemXDPI !== undefined
-            && window.screen.logicalXDPI !== undefined
-            && window.screen.systemXDPI > window.screen.logicalXDPI) {
-            return window.screen.systemXDPI / window.screen.logicalXDPI;
+        // these are IE specific
+        let sysXDPI = (window.screen as any).systemXDPI
```
jwunderl:
```suggestion
        const sysXDPI = (window.screen as any).systemXDPI
```

pxtlib/browserutils.ts
```diff
@@ -260,10 +260,13 @@ namespace pxt.BrowserUtils {
...
+        let logicalXDPI = (window.screen as any).logicalXDPI
```
jwunderl:
```suggestion
        const logicalXDPI = (window.screen as any).logicalXDPI
```

webapp/src/greenscreen.tsx
darzu: oh strange, this might be a merge gone wrong. I'll investigate

darzu: Ah, this was intentional. MediaStream.stop() was deprecated: https://developers.google.com/web/updates/2015/07/mediastream-deprecations

The right way is to stop the tracks which is what we do below

PR #6048, opened 2019-10-09
fix toc for now

jwunderl:
There was a change to pxtlib/docsrender that relied on docfiles macros being updated, which can't be applied until all targets have the new & improved macros in their docfiles - that is, until all targets are bumped to a version of pxt containing that PR.

this should fix TOC for now in both situations, with some TODOs to clean it up later


pelikhan:
@darzu to double check this

PR #6013, opened 2019-10-05
doc TOC menu improvements

darzu:
- [x] chevron
- [x] better indent
- [x] collapsed by default
- [x] more uniform
- [x] light & dark theme

Before merge:
- [x] Test IE
- [x] Test with microbit docs

Before:
image

Goal:
image

PR #6049, opened 2019-10-09
Fixes table of content rendering

darzu:
This PR: https://github.com/microsoft/pxt/pull/6013
introduced a change that has to be rolled out to all editors simultaneously because of how the docs renderer <-> docfiles contract works.
This is too annoying to do, so this PR rolls back some of those changes and instead keeps the renderer as it was at the cost of some ugly code.

Eventually we should create a better contract between the renderer and the docfiles so changes like this aren't coupled.

### DIFF
--- a/docfiles/macros.html
+++ b/docfiles/macros.html
@@ -110,25 +110,6 @@
     <div class="divider"></div>
 </aside>
 
-<aside id=toc-dropdown class=menu>
-    <nav div class="ui simple dropdown item" title="@TITLE@">
-        @NAME@
-        <i class="dropdown icon"></i>
-        <div class="menu">
-            @ITEMS@
-        </div>
-    </nav>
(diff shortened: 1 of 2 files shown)


jwunderl:
https://github.com/microsoft/pxt/pull/6048?

pelikhan:
> This is too annoying to do,

This is impossible to acheive as we would have to release all editors simultaneously.

jwunderl:
I'd prefer #6048 as the explicit todos seem like it gives it a slightly higher chance of eventually cleaning up the tech debt

(and also just cleaned up a few more places in that branch / fixed a small bug as I was looking at what needed to be done for accessibility there)

darzu:
Redundant with https://github.com/microsoft/pxt/pull/6048#pullrequestreview-300231380

PR #6053, opened 2019-10-09
Support decompiling enums inside of namespaces

riknoll:
Fixes the issue where the "stop all animations" block in arcade doesn't decompile. I couldn't find an issue for it.


pelikhan:
Many extensions in micro:bit have nested enums, this is great!

pelikhan:
This fix should go to microbit branch @abchatra

abchatra:
Only point fixes with clear bugs should go to stable branches. Otherwise chance of regression are high. Why should this go to microbit?

Issue #5979, opened 2019-10-02
Docs 2.0: new footer


darzu:
![image](https://user-images.githubusercontent.com/6453828/66087539-b46ee900-e52c-11e9-91ee-f39c88e111e5.png)

Issue #5984, opened 2019-10-02
Docs 2.0: Use consistent sat & lightness across targets, only change the hue (much easier design and accessibility)


darzu:
![image](https://user-images.githubusercontent.com/6453828/66087701-4ecf2c80-e52d-11e9-96d6-d9fbf1612e23.png)

vs

![image](https://user-images.githubusercontent.com/6453828/66087717-568ed100-e52d-11e9-9f27-e990a2cb06a2.png)

very different saturation & lightness, makes it a pain to test each target and handle accessibility contrast issues

PR #6054, opened 2019-10-10
New footer, consistent styling for header / toc

jwunderl:
add in new footer, and switch up the styling for header and sidebar a bit so that things fit together a bit more nicely (footer looked really out of place on it's own).

**Starting as draft to verify**: is everyone okay with having hue be the main style choice for docs? It seems like the first pass was moving a lot more towards consistency in style for the docs, and just building off a single hue helps with that (e.g. the header background and sidebar accent colors fit together because they're just different shades of the same color; can add more uses of it later). Also makes accessibility a lot easier to handle, as you don't have to worry about color issues as much as they're closer to the same contrast.

Main differences in sidebar:
- light background, colored accents based off hue
- less indentation per level
- a border 'dropping down' from the chevron when expanded, to give an extra indication of indentation level / have some use for the whitespace

the footer Microsoft icon doesn't seem to be showing up correctly in targets, but I think it might just be a local serve issue with images not being shoved into the right place?

heres a few things showing changing hues - we can play with what saturation / luminosity looks best (can't change them all at once as the variable is in less / ie doesn't support css variables :( )
header/footer

chevrons
ezgif com-crop
indent line

highlight color

part of https://github.com/microsoft/pxt/issues/5984 (probably drop use of some of the existing variables?)
closes https://github.com/microsoft/pxt/issues/5979
and also a few drive-by fixes:
* make it so translations button is tab accessible
* drop Instagram as it isn't maintained
* indicate focus on codecards
* one or two potential double sidebars
* open mailto makecode in new tab like all the other buttons


pelikhan:
the social icons were removed in the docs redesign.

pelikhan:
Looks great. Push it in, bump and let's play with it.

jwunderl:
@pelikhan re: the social icons, my understanding from the design meetings was that we decided to drop the `docs global navigation` that was being pulled from the apptheme in the new design, and leave the icons to fill in that space. @abchatra ^?

abchatra:
Yes we wan to keep social icons.

PR #6055, opened 2019-10-10
[stable5.15] integrate translation of docs

pelikhan:
Update docs html to support corwdin in context translation. 
(note: I have not ported the localhost server changes to test this locally to keep the patch small).

This fix is needed to enable in-context translations in micro:bit.


mmoskal:
You can set HEAD_OVERRIDE on staging to stable51docstranslate. You may need the same branch in microbit repo though.

PR #5502, opened 2019-04-23
Langpicker

toledoal:
Adding the language picker for Docs. 
I can successfully switch languages but I'm not able to get any translations for the docs. Maybe we currently don't have any? 
Any tips on where should I look for to verify? Thanks. 

image

PR #6064, opened 2019-10-11
language picker in docs

jwunderl:
building on https://github.com/microsoft/pxt/pull/5502

One main question; is putting the lang cookie into the `/` path alright? it seems like it would be (just a language choice, and won't conflict with other things on the domain as it's pretty specific (PXT_LANG)), and otherwise it will have really weird behavior when you e.g. change lang in `/courses/csintro` and it doesn't change in /docs or the editor.

~draft as I'm still trying to get the modal focus trap working properly, it wasn't correctly trapping focus (keyboard / screenreader accessibility), haven't got a fix quite yet. Might have to switch up the modal implementation (but that is potentially worth doing anyways, as this one used the jquery modal). Current hack solution in https://github.com/microsoft/pxt/pull/6064/commits/ebdaa2ce65d12a7bb06902b5504c2a678328d904, but it's kinda gross.~ Modal behavior effectively matching the main apps modals, needs proper fix later for both.


docfiles/footer.html
```diff
@@ -20,12 +20,12 @@
         </div>
         <div class="ui section divider"></div>
         <div class="ui container horizontal small divided link list">
+            <a id="langpicker" tabindex="0" class="ui item" rel="noopener">
+                <i class="icon xicon globe" aria-hidden="true" role="presentation"></i>Language
+            </a>
             <a class="item" href="https://makecode.com/privacy" target="_blank" rel="noopener">Privacy &amp; Cookies</a>
             <a class="item" href="https://makecode.com/termsofuse" target="_blank" rel="noopener">Terms Of Use</a>
             <a class="item" href="https://makecode.com/trademarks" target="_blank" rel="noopener">Trademarks</a>
-            <!-- @ifdef crowdinproject -->
```
pelikhan: why is this removed?

jwunderl: I bumped it to the bottom of the lang selection page to match editor intro, was going to ask you about that on monday though.  Will put it back here if you prefer

docfiles/docs.js
```diff
@@ -280,6 +281,135 @@ function renderSnippets() {
...
+            var currentOption = langOptions[i];
+
+            currentOption.onclick =  function(e) {
+                var langId = e.currentTarget.dataset.lang;
+                if (!pxt.Util.allLanguages[langId]) {
+                    return;
+                }
+                setCookieLang(langId);
+                if (langId !== initialLang) {
+                    pxt.tickEvent("menu.lang.changelang." + langId);
```
abchatra: Fix this in the new format. dynamic langId shouldn't be done here. 
        pxt.tickEvent(`menu.lang.changelang`, { lang: langId });

docfiles/docs.js
```diff
@@ -280,6 +281,135 @@ function renderSnippets() {
...
+                    // In react app before reload we are using pxt.winrt.releaseAllDevicesAsync()
+                    // In docs we currently don't have access to pxt.winrt
+                    location.reload();
+                } else {
+                    pxt.tickEvent("menu.lang.samelang." + langId);
```
abchatra: pxt.tickEvent(`menu.lang.samelang`, { lang: langId, docs: "true" });

docfiles/docs.js
abchatra: Also probably add a notion of docs too.
pxt.tickEvent("menu.lang.changelang", { lang: langId, docs: "true" });

pxtlib/util.ts
```diff
@@ -970,6 +970,25 @@ namespace ts.pxtc.Util {
...
+        return cookieValue && cookieValue[1] || null;
+    }
+
+    export function setCookieLang(langId: string, docs = "false") {
+        if (!pxt.Util.allLanguages[langId]) {
+            return;
+        }
+
+        if (langId !== getCookieLang()) {
+            pxt.tickEvent(`menu.lang.setcookielang`, { lang: langId, docs});
```
abchatra: needs docs : "true"

docfiles/docs.js
```diff
@@ -280,6 +281,108 @@ function renderSnippets() {
...
+    cardEl.setAttribute('role', 'option');
+    cardEl.setAttribute('aria-label', locale.englishName);
+    cardEl.setAttribute('tabindex', '0');
+    cardEl.appendChild(contentEl);
+
+    return cardEl;
+}
+
+function setupLangPicker() {
+    var appTheme = pxt.appTarget.appTheme;
```
pelikhan: check that appTheme exists

pxtlib/util.ts
```diff
@@ -970,6 +970,25 @@ namespace ts.pxtc.Util {
         return false;
     }
 
+    export function getCookieLang() {
+        const cookiePropRegex = new RegExp(`${pxt.Util.escapeForRegex(pxt.Util.pxtLangCookieId)}=(.*?)(?:;|$)`)
+        const cookieValue = cookiePropRegex.exec(document.cookie);
+        return cookieValue && cookieValue[1] || null;
+    }
+
+    export function setCookieLang(langId: string, docs: "true" | "false" = "false") {
```
pelikhan: "true" | "false" = "false" -> boolean?

pxtlib/util.ts
```diff
@@ -970,6 +970,25 @@ namespace ts.pxtc.Util {
         return false;
     }
 
+    export function getCookieLang() {
```
pelikhan: since this relies on a browser feature, it should be in pxt.BrowserUtils.

jwunderl:
Hmm, both language pickers are breaking in minecraft at the moment due to the `nb` language not being pxt.Util.allLanguages[code]. Should I add checks for that and just drop them from the language selection or should nb info be in or added to pxt.Util.allLanguages[code] ?

pelikhan:
i would recommend fixing nb issues separately from this.

docfiles/docs.js
```diff
@@ -279,6 +280,110 @@ function renderSnippets() {
     });
 }
 
+function languageOption(code) {
+    var locale = pxt.Util.allLanguages[code];
+    if (!locale)
```
abchatra: Lets remove these check. We should crash instead of silently succeed if the apptheme locale is not a subset of all available locales

docfiles/footer.html
```diff
@@ -20,12 +20,12 @@
...
-            <a id="translatebtn" tabindex="0" class="item">Translate this page</a>
```
pelikhan: please keep this. putting this at the bottom of the language picker is burying too much

abchatra: We talked about this in the design meeting.  Consensus language picker is good place to put this in. Consistent with the editor. If you are translating you will click on the language selection to see.

pelikhan:
@jwunderl  is this ready?

PR #6065, opened 2019-10-11
Fix sidebar roles

jwunderl:
Fixes https://github.com/microsoft/pxt-microbit/issues/2339

Fix roles for sidebar and header bar, so screenreaders can parse it properly.

### DIFF
--- a/pxtlib/docsrender.ts
+++ b/pxtlib/docsrender.ts
@@ -319,7 +319,7 @@ namespace pxt.docs {
 
         // Add sidebar toggle
         const sidebarToggleHtml = `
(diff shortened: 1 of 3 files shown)


pxtlib/docsrender.ts
```diff
@@ -319,7 +319,7 @@ namespace pxt.docs {
 
         // Add sidebar toggle
         const sidebarToggleHtml = `
```
pelikhan: This change would require a cloud update.
We should inline all these macros in the HTML, and eventually get rid of those in docsrender.

PR #6067, opened 2019-10-14
Update 'tutorials' page for new block sections

ganicke:
Add some info for the additional block section types.

RE: https://github.com/microsoft/pxt-minecraft/issues/1254

### DIFF
--- a/docs/writing-docs/tutorials.md
+++ b/docs/writing-docs/tutorials.md
@@ -9,6 +9,14 @@ A tutorial is a sequence of steps that the user follows in a limited environment
 Each step of the tutorial has a short description, or explanation, of the activity for the step and
 possibly a block example. If the step includes a block example, the editor will restrict the selection of blocks from the toolbox to only those used in the example.
 
+### ~ hint
+
+#### Tutorial tool
+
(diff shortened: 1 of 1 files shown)


shakao:
Right now the documentation for "ghost" and "template" are with the other macros: https://makecode.com/writing-docs/macros -- should we consolidate the tutorial-specific macros somehow? 

Hint documentation is currently lumped with "activities" and "steps" here: https://github.com/microsoft/pxt/blame/master/docs/writing-docs/tutorials.md#L232 maybe we should remove it from that section entirely, if we want it in the "Using blocks" section?

ganicke:
@shakao thanks for the comments. This is really a matter of how to message use of hints...blocks themselves are implicit hints but they're blocks too and are related to the template and ghost blocks when discussing usage. The 'reference' for the types are in ``macros.md`` but the discussion of use in ``tutorials.md``? What do you think?

shakao:
Oh yeah, that distinction makes a lot of sense--in that case, LGTM!

ganicke:
We'll keep iterating on this page. It will need some shuffling of structure probably.

PR #6069, opened 2019-10-15
faster sidebar animation

jwunderl:
Makes the sidebar transitions a bit quicker, as the defaults feels pretty slow.


docfiles/docs.js
```diff
@@ -127,8 +130,6 @@ function setupSidebar() {
     for (var i = 0; i < searchIcons.length; i++) {
         searchIcons.item(i).onkeydown = handleEnterKey;
     }
-
-    scrollActiveHeaderIntoView();
```
jwunderl: this was dead code as it was trying to scroll the sidebar when it was hidden by default.

theme/theme.config
```diff
@@ -62,7 +62,7 @@
 @sidebar    : 'pxt';
 @sticky     : 'pxt';
 @tab        : 'pxt';
-@transition : 'default';
+@transition : 'pxt';
```
jwunderl: there are a handful more of these in this file that are set to default, and a LOT in the different targets; e.g. in arcade almost all modules just use 'default', including sidebar (which took a little bit to identify).

Not many have rules overridden in pxt, but it feels really confusing so unless there's opposition to it I'll go through the targets and make them all point at pxt for the themes so things don't randomly break

pelikhan: yes we need to systematically fix this.

PR #6073, opened 2019-10-17
Snippet builder tweaks for v1 beta launch

darzu:
https://youtu.be/QmGW2JsBax8

Changes:
- Lot of fixes to spacing and padding/margins
- Remove elevated question shadow box
- Add "graph paper" behind code blocks
- Better responsive window behavior
- Adds "Yes No" input option (used by "Should this sprite use controls")
- Nicer font
- Pairs with: https://github.com/microsoft/pxt-arcade/pull/1403

### DIFF
--- a/theme/snippet-builder.less
+++ b/theme/snippet-builder.less
@@ -41,6 +41,8 @@
                 }
                 padding: 1.5rem;
                 padding-top: 0;
+                padding-left: 0.5rem;
+                overflow-x: hidden;
             }
 
             .ui.segment.raised {
@@ -101,6 +103,7 @@
         .snippet.hint {
             padding: 0;
             padding-left: 1rem;
+            padding-right: 0.5rem;
         }
     }
 
@@ -498,10 +501,10 @@ Inputs
         margin: 0 0 0 -48%;
     }
 }
-@media only screen and (min-width: 1200px)
+@media only screen and (min-width: 1109px)
 {
     .snippet-builder.ui.modal {
-        width: 1150px;
-        margin: 0 0 0 -575px;
+        width: 1060px;
+        margin: 0 0 0 -530px;
     }
 }
--- a/theme/snippet-builder.less
+++ b/theme/snippet-builder.less
@@ -1,47 +1,70 @@
-.snippet-builder {
-    max-height: 95%;
+.snippet-builder.ui.modal > .content {
+    /* Semantic CSS override */
+    padding: 0;
+}
+    
+.snippet-builder.ui.modal {
+    /* Semantic CSS override */
     height: 95%;
-    padding-bottom: 77px;
-    .snippet-question {
-        h6 {
-            font-weight: bolder;
-        }
-    }
 
-    .ui.segment.raised {
-        overflow: visible;
-        width: fit-content;
-        height: fit-content;
-        box-sizing: content-box;
-    } 
+    display: grid !important;
+    grid-template-rows: 
+        70px // header
+        1fr // content
+        77px // footer
+        ;
+}
 
-    .snippet-error {
-        color: red;
-        margin-top: 15px;
-    }
+.snippet-builder {
+    // display: flex;
 
     .content {
+        // display: flex;
         overflow: scroll;
-        height: 100%;
-        max-height: 100%;
-        .ui.equal.width.grid {
-            .snippet.output-section.column {
-                padding-top: 0;
+
+        .snippet-builder-inner {
+            padding: 0;
+            display: grid;
+            grid-template-columns: 1fr 1fr;
+            height: 100%;
+
+            .snippet-question {
+                h6 {
+                    font-weight: bolder;
+                }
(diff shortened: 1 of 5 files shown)


webapp/src/snippetBuilderInputHandler.tsx
```diff
@@ -172,39 +179,74 @@ class RangeInput extends data.Component<IRangeInputProps, {}> {
...
+    }
+
+    onChange(value: string): () => void {
+        const { onChange } = this.props;
+        return () => onChange(value);
+    }
+
+    renderCore() {
+        const { input } = this.props;
+        if (Snippet.isSnippetInputAnswerTypeYesNo(input)) {
```
shakao: is the "input" prop only used in this check? maybe the yesno input component should just render the yes/no buttons and bind the listener, and the parent should control whether it renders?

darzu: Yeah, I agree with you, but for the other input types the input is used more. I think the parent should have controlled the rendering on all of these but I'm not going to change paradigm now. It's wonky but it works :)

webapp/src/snippetBuilder.tsx
```diff
@@ -489,10 +516,14 @@ export class SnippetBuilder extends data.Component<SnippetBuilderProps, SnippetB
                                 {currentQuestion.errorMessage && <p className='snippet-error'>{currentQuestion.errorMessage}</p>}
                             </div>
                             {currentQuestion.hint &&
-                                <div className='snippet hint ui segment'>{pxt.Util.rlf(currentQuestion.hint)}</div>}
+                                <div className='snippet hint ui segment'>{
+                                    pxt.Util.rlf(currentQuestion.hint)
+                                        // replace new lines with <br> elements
+                                        .split("\n")
```
shakao: is there a reason for the split/reduce vs a global string replace?

darzu: I think so, because it's not a string I'm inserting, it's a special JSX <br> tag type.

webapp/src/snippetBuilder.tsx
```diff
@@ -489,10 +516,14 @@ export class SnippetBuilder extends data.Component<SnippetBuilderProps, SnippetB
                                 {currentQuestion.errorMessage && <p className='snippet-error'>{currentQuestion.errorMessage}</p>}
                             </div>
                             {currentQuestion.hint &&
-                                <div className='snippet hint ui segment'>{pxt.Util.rlf(currentQuestion.hint)}</div>}
+                                <div className='snippet hint ui segment'>{
+                                    pxt.Util.rlf(currentQuestion.hint)
```
pelikhan: how is this supposed to work? I don't see where those strings would be injected in our translation system.

theme/snippet-builder.less
```diff
@@ -1,482 +1,510 @@
...
+.snippet-builder.no-back-btn {
+    .actions {
+        .button.left {
+            display: none;
+        }
+    }
+}
+.snippet-builder {
+    .content {
+        font-family: Open Sans,Helvetica Neue,Arial,Helvetica,sans-serif;
```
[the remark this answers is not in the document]
darzu: It's the same font used in the side docs. This consistently came up in the design reviews that the font was too hard to read in the snippet builder. Most of the editor doesn't have as much reading to do.

theme/snippet-builder.less
```diff
@@ -1,482 +1,510 @@
...
+                border: 0;
+            } 
+
+            .snippet-error {
+                color: red;
+                margin-top: 15px;
+            }
+            .snippet.output-section {
+                padding: 1.5rem;
+                background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='16' height='16'%3E%3Cpath d='m0.5,0l0,16M0,8l16,0' style='stroke:black;stroke-width:1;stroke-opacity:0.1;' /%3E%3C/svg%3E");
```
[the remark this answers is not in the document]
darzu: SnippetBuilder works on Firefox, Safari, and Chrome on Mac.
That background-image doesn't work on edge tho. sigh.

webapp/src/snippetBuilder.tsx
darzu: not sure. I didn't touch this in these PRs, so I'd have to check out Jacob implemented this

theme/snippet-builder.less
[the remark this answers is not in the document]
darzu: I spent >1hr trying to get alternatives to work on Edge. Can't figure out how to get doccdn working with less or jsx. But w/e, it's enough time spent, I'm not going to fix this. It looks fine without background grid on Edge:
image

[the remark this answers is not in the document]
darzu: Actually it works in edge. It was the dumb edge developer tools that were giving me errors when I tried to add that background image ad-hoc. Very frusterating

PR #6078, opened 2019-10-17
add doc header field to apptheme

jwunderl:
In theory should fix header name when combined with adding a docsHeader, eg https://github.com/microsoft/pxt-maker/pull/258

Screen Shot 2019-10-17 at 11 36 50 AM

--->

Screen Shot 2019-10-17 at 11 37 13 AM

### DIFF
--- a/pxtlib/docsrender.ts
+++ b/pxtlib/docsrender.ts
@@ -287,7 +287,7 @@ namespace pxt.docs {
             params["homeurl"] = html2Quote(theme.homeUrl);
         params["targetid"] = theme.id || "???";
         params["targetname"] = theme.name || "Microsoft MakeCode";
-        params["docsheader"] = theme.docsHeader || "Microsoft MakeCode";
+        params["docsheader"] = theme.docsHeader || "Documentation";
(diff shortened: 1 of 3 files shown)


pxtlib/docsrender.ts
```diff
@@ -286,7 +286,7 @@ namespace pxt.docs {
         if (theme.homeUrl)
             params["homeurl"] = html2Quote(theme.homeUrl);
         params["targetid"] = theme.id || "???";
-        params["targetname"] = theme.name || "Microsoft MakeCode";
+        params["targetname"] = theme.docsHeader || theme.name || "Microsoft MakeCode";
```
pelikhan: let's use another field for the "docs header". "targetname" is used in many other instance and it refers to pxtarget.name.

pxtlib/docsrender.ts
```diff
@@ -287,6 +287,7 @@ namespace pxt.docs {
             params["homeurl"] = html2Quote(theme.homeUrl);
         params["targetid"] = theme.id || "???";
         params["targetname"] = theme.name || "Microsoft MakeCode";
+        params["docsheader"] = theme.docsHeader || "Microsoft MakeCode";
```
pelikhan: i would do theme.name || "Documentation"

jwunderl:
Hm, should I split the `tocheader.html` change into a separate PR then just to be safe?

PR #6080, opened 2019-10-17
Revert "upgrade to typescript 3.6; turn on incremental compilation"

darzu:
Some unexpected breaks in targets happened.

I'll need to prep and roll out simultaneously with targets more carefully.


darzu:
Holding on merge for a sec, might be able to unblock the targets instead

PR #6083, opened 2019-10-18
Bump pxt-blockly to 2.1.9, begin to remove references to goog.dom.cre…

shakao:
…ateNote. Some small fixes for LED matrix (don't add element if block is insertion marker) and keyboard (set value in doValueUpdate)


pxtblocks/blocklyloader.ts
```diff
@@ -1986,7 +1986,7 @@ namespace pxt.blocks {
                 xmlList.push(headingLabel);
             }
 
-            let button = goog.dom.createDom('button') as HTMLElement;
+            let button = document.createElement('button') as HTMLElement;
```
pelikhan: weird that it does not use the blockly.utils stuff... but it works!

PR #6085, opened 2019-10-18
recenter images in tutorial hints

jwunderl:
Previously images were getting the center class from in docsrender.ts, but with docs2.0 being left aligned that changed.

Screen Shot 2019-10-18 at 11 30 58 AM


abchatra:
Fixed

PR #6102, opened 2019-10-22
Use the new sprite editor in Monaco

riknoll:
Tested in Chrome, Firefox, and Safari. The only thing left that uses the old sprite editor is the snippet builder. I'll take care of that in a separate PR once @darzu has merged his changes.

I have not enabled the animation editor yet.


riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/1392

PR #6105, opened 2019-10-22
moved installedVersion to package

pelikhan:
We keep a cache of the resolved version of a package in pxt.json. This is bad since it modifies program content and potential disastrous in the context of co-authoring and merging.

This change moves installedVersion to the package itself. Everything seems to work.

Note: not systematically cleaning on load as it would modify scripts and have some unintentional results.

### DIFF
--- a/localtypings/pxtpackage.d.ts
+++ b/localtypings/pxtpackage.d.ts
@@ -30,7 +30,7 @@ declare namespace pxt {
     interface PackageConfig {
         name: string;
         version?: string;
-        // installedVersion?: string; ignored going further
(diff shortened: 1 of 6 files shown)


localtypings/pxtpackage.d.ts
```diff
@@ -30,7 +30,7 @@ declare namespace pxt {
     interface PackageConfig {
         name: string;
         version?: string;
-        installedVersion?: string;
+        // installedVersion?: string; ignored going further
```
riknoll: deprecated

PR #6108, opened 2019-10-23
updated logic to init project for github

pelikhan:
A few updates to the logic to populate a github project; to align better with classroom. This PR fixes the following flow:

* teacher creates github classroom assignment with empty repo
* student gets assgiment and generates empty repo
* student imports empty repo in makecode (missing token handled)
* empty project populate landing on blocks

Todos:
- [x] leave .blocks files since we now support editing blocks
- [x] localize readme, clearer instructions
- [x] normalize pxt.json to avoid initial commit
- [x] ask before populating repo
- [x] check that token is available when populating repo

### DIFF
--- a/pxtlib/initprj.ts
+++ b/pxtlib/initprj.ts
@@ -214,7 +214,10 @@ cache:
             "testFiles",
             "testDependencies",
             "public",
-            "targetVersions"
+            "targetVersions",
+            "preferredEditor",
+            "additionalFilePath",
+            "additionalFilePaths"
         ]
 
         config.files = pkgFiles.filter(s => !/test/.test(s));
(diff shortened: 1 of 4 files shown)


pxtlib/initprj.ts
```diff
@@ -172,34 +176,33 @@ cache:
...
+        delete config.additionalFilePaths;
+        // (keep blocks files)
+        config.preferredEditor = config.files.find(f => /\.blocks$/.test(f))
+            ? pxt.BLOCKS_PROJECT_NAME : pxt.JAVASCRIPT_PROJECT_NAME;
         config.name = name;
-        // by default, projects are not public
-        config.public = false;
-
-        if (!config.version) {
+        config.public = true;
```
jwunderl: Was changing this to public by default to avoid students making private repos and not knowing how to update it or something of the sort?

mmoskal: Public false prevents publishing (sharing) i think

PR #6125, opened 2019-10-25
Turns on strict-null-checking in pxtpy/

darzu:
Depends on: https://github.com/microsoft/pxt/pull/5715 completing first

- [x] TODO rebase after above completes 

This adds strict-null-checking to pxtpy (python compiler & decompiler).

This PR includes at least one legitimate bug fix, but possibly more as null/undefined cases are more explicitly handled.

There is a chance of regression since I'm not 100% sure I handled the null/undefined cases right in all cases, but I'd rather get this work in early so that these can be found and fixed.

The most important part of this IMO is so that all the future python work in the next few months can rely on the strict-null-checking and new code written should be designed to pass strict null checking. IMO this will speed up development and increase quality.

The methodology I used was basically:
- If it's obvious what the right way to handle null/undefined is, great, update the code to do that
- If it's not, at least log an error when there is an unexpected null and leave the code untouched (minus type coercion)

As a result, this PR adds quite a few new error logging lines.

This is also a canary experiment for strict-null-checking for PXT. If this goes well, perhaps other areas of PXT can be converted.


pxtpy/ast.ts
```diff
@@ -415,7 +415,7 @@ namespace pxt.py {
     }
     export interface NameConstant extends Expr {
         kind: "NameConstant";
-        value: boolean; // null=None, True, False
+        value: boolean | undefined; // undefined=None, True, False
```
pelikhan: isn't this ``value?: boolean``

mmoskal: I think the current version is clearer since undefined here represents a value

darzu:
@mmoskal , are you okay with turning this on?

PR #6130, opened 2019-10-29
Objectify the web element and format codes

StevenH8:
Updates to format codes so that it has more readable and make some changes to keep it working.


pelikhan:
Could you provide more details about this PR?

PR #6132, opened 2019-10-29
Call textToDomDocument directly for cached SVG elements (#6129)

shakao:


abchatra:
I did the cherry-pick already.

PR #6138, opened 2019-10-30
Removing the old sprite editor (may it rest in peace)

riknoll:
Removing the last remaining usage of the old sprite editor in the snippet builder. I saved the editor/gallery toggle because it was used by the melody editor and refactored the bitmap utils into pxtlib so that they aren't duplicated anymore.

Screen Shot 2019-10-30 at 3 28 32 PM

The new sprite editor is taller than the old one, so it causes this step to scroll on some screens. I removed the gallery to help mitigate that, but if we want it to not scroll we probably need to shrink the top and bottom modal margins. If I add the gallery back it will scroll on every screen.


pxtlib/spriteutils.ts
```diff
@@ -188,10 +221,10 @@ namespace pxtsprite {
         return result;
     }
 
-    export function bitmapToImageLiteral(bitmap: Bitmap, fileType: pxt.editor.FileType): string {
+    export function bitmapToImageLiteral(bitmap: Bitmap, fileType: "typescript" | "python"): string {
```
darzu: I'm curious why we don't use the lang enum here?

pelikhan:
@riknoll ready to merge?

PR #6147, opened 2019-10-31
Github experience

pelikhan:
Round 2 on github signin.

Test build: https://arcade.makecode.com/app/9e6ac0d23e6cbde4d6a341a2c8979cb96a8c15d9-1db5da5931

- [x] add gearwheel menu on home scree: about, reset, language, high contrast, sign in/out of github
image
image
- [x] move github out of bubble head menu
- [x] your github repo always visible in import menu
- [x] move github status button next to save button
image
image
- [x] group extension stuff into a "zone" (hidden in blocks)
image
- [x] minimalist experience in blocks (no extension stuff)
image

### DIFF
--- a/webapp/src/cloudsync.ts
+++ b/webapp/src/cloudsync.ts
@@ -235,7 +235,8 @@ export class ProviderBase {
         pxt.storage.removeLocal(this.name + "token")
         pxt.storage.removeLocal(this.name + "tokenExp")
         pxt.storage.removeLocal(this.name + "AutoLogin")
-        pxt.storage.removeLocal("cloudUser")
+        if (this.hasSync())
+            pxt.storage.removeLocal("cloudUser")
         invalidateData();
     }
 }
@@ -371,7 +372,7 @@ export function resetAsync() {
 function updateNameAsync() {
     const provider = currentProvider;
     let user = pxt.storage.getLocal("cloudUser");
-    if (user || !provider)
+    if (user || !provider || !provider.hasSync())
         return Promise.resolve()
     return provider.getUserInfoAsync()
         .then(info => {
--- a/webapp/src/cloudsync.ts
+++ b/webapp/src/cloudsync.ts
@@ -269,7 +269,8 @@ import * as onedrive from "./onedrive";
 import * as googledrive from "./googledrive";
 import * as githubprovider from "./githubprovider";
 
-export function providers(): IdentityProvider[] {
+// All identity providers, including hithub
+function identityProviders(): IdentityProvider[] {
     if (!allProviders) {
         allProviders = {}
         const cl = pxt.appTarget.cloud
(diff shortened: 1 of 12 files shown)


abchatra:
Signout from github is sufficient in the gear menu, right?  What does the photo do?

abchatra:
Rest looks great

pelikhan:
You need to know who you are logged in as. Otherwise you have no idea "who" you are.

abchatra:
Then may be put the bubble head in place of GitHub icon

webapp/src/cloudsync.ts
```diff
@@ -269,7 +284,8 @@ import * as onedrive from "./onedrive";
 import * as googledrive from "./googledrive";
 import * as githubprovider from "./githubprovider";
 
-export function providers(): IdentityProvider[] {
+// All identity providers, including hithub
```
riknoll: github

pelikhan:
What is the problem with clearly stating who is singed in? Many user do not set their avatar in github.

abchatra:
Everything else does open something when you click on it. This is just a label? Seems odd to me.

pelikhan:
It opens your github profile page.

abchatra:
We all think it doesn't below in the drop down menu. Don't add it now. We can discuss in the next design meeting if you feel strongly about it.

pelikhan:
How do you tell who is the github user? It could be someone else than you since machine are shared.

abchatra:
You could say this one link 
"Your head image" + "Sign out Pelikhan"
Omit GitHub as that is the only way you can sign in today. We can figure out later.

pelikhan:
It's a single menu item "Sign out" as the name bleeds out otherwise. Alea jacta est.

webapp/src/cloudsync.ts
```diff
@@ -656,34 +688,44 @@ function userInitials(username: string): string {
...
-                    return currentProvider.icon;
-                const prs = providers();
-                if (prs.length == 1)
-                    return prs[0].icon;
-                return "user";
-        }
-        return null
-    },
-})
+function syncApiHandler(p: string) {
```
darzu: This seems like the main state machine in the auth flow. Is that correct? Would you mind providing some code comments here or elsewhere in the file giving a high level picture of what's happening?

Some questions:
- How does the data flow with the providers work? e.g. 1st, hit this github API, 2nd pop this dialog to get consent, etc.
- What do we store locally and why?
- When and why do we interact with the backend?
- Any gotcha's or non-standard stuff we do?

darzu:
Nit:
image
the shading on those two buttons is different which is weird since they're side by side.
I'd vote that the save button gets the shading like our other sim buttons, I don't know why it's flat

pelikhan:
@darzu shading is inconsistent
image

darzu:
disabled buttons r flat, clickable are supposed to pop. But not important, since Richard was saying we should probably make them all flat always

PR #6159, opened 2019-11-04
image scripts update

pelikhan:


jwunderl:
Maybe worth adding a check so that the image is only replaced if it's smaller than the old one? looks like a handful of the ones in https://github.com/microsoft/pxt-adafruit/pull/1135/files?short_path=513a469#diff-513a469a8531a5c0a628cd134ae21908 ended up being a tiny bit larger than they were previously (though they all seem to be on the order of < 5kb larger)

pelikhan:
Next update.

PR #6160, opened 2019-11-04
Bump pxt-blockly to 2.1.12

shakao:
Allows custom image URL as Blockly workspace background (Minecraft HOC)


pelikhan:
Does the image get removed from the block screenshot?

shakao:
yep, it replaces the workspace grid pattern so it is also excluded from the screenshot

PR #6163, opened 2019-11-05
Build api info for bundled packages as part of the target build

riknoll:
Instead of building the API info on editor load, this change makes it so that the API info is built at release time and packaged in `target.js`. This speeds up the time for the blocks editor to load by quite a lot (especially in arcade). As long as the project you're trying to load only uses built-in packages, we no longer have to wait for a full compile. If you have added an extension to your project, we fallback to the old slower behavior.

One thing to note: we have a few packages that change the exported values of other packages. For example, many of our targets override namespace colors by including an `ns.ts` file in the core library. With this change, that no longer works for any library *outside* of the one listed in the `corepkg` field of `pxtarget.json`. I don't think we have any extensions that do that today.

In an effort to make local builds stay fast, I cache this API info along with a SHA of the lib files and PXT version. You shouldn't see them rebuilding unless the files change.

I tested this with pxt-arcade and pxt-microbit. And yes, it still works if you remove radio and add bluetooth :smile:


riknoll:
BTW for anyone trying to test this out, arcade has the wrong corepkg. I've opened a PR to fix it: https://github.com/microsoft/pxt-arcade/pull/1441

webapp/src/compiler.ts
```diff
@@ -386,15 +386,76 @@ export function getApisInfoAsync() {
...
+        }}).filter(p => p && p.config);
+
+    const usedPackages = project.pkgAndDeps();
+    const usedInfo: pxt.PackageApiInfo[] = [bundled["libs/" + pxt.appTarget.corepkg]];
+
+
+    for (const dep of usedPackages) {
+        if (dep.id === "built" || dep.isTopLevel()) continue;
+
+        let foundIt = false;
```
jwunderl: nit: should be able to simplify this a bit;

```javascript
const matchedBundle = bundledPackages.find(bundle => bundle.config.name === dep.getKsPkg().config.name)
if (matchedBundled) {
    usedInfo.push(bundled[matchedBundle.dirname]);
} else {
    return undefined;
}
```

webapp/src/compiler.ts
```diff
@@ -386,15 +386,76 @@ export function getApisInfoAsync() {
...
+function getApiInfo(project: pkg.EditorPackage, bundled: pxt.Map<pxt.PackageApiInfo>) {
+    const bundledPackages: BundledPackage[] = pxt.appTarget.bundleddirs.map(dirname => {
+        const pack = pxt.appTarget.bundledpkgs[dirname.substr(dirname.lastIndexOf("/") + 1)];
+
+        if (!pack) return null;
+
+        return {
+            dirname,
+            files: pack,
+            config: JSON.parse(pack[pxt.CONFIG_NAME]) as pxt.PackageConfig
```
pelikhan: pxt.json may be missing or invalid format

pelikhan: it's not the only spot we make assumptions that's it's there but it would be good to know if the service will recover if pxt.json is invalid.

mmoskal: I think this one is coming from pxtarget.json so we should be fine

PR #6169, opened 2019-11-05
YouTube-Embeding as optional feature

Amerlander:
Added an Option to Embed YouTube Videos direct into the CodeCard-Preview:

e.g.:

"youTubeId": "qqBmvHD5bCw",
"embedYouTube": "true"

### DIFF
--- a/docs/extensions/codecard.md
+++ b/docs/extensions/codecard.md
@@ -13,6 +13,7 @@
         typeScript?: string;
         imageUrl?: string;
         youTubeId?: string;
+        embedYouTube?: boolean;
         time?: number;
         url?: string;
         responsive?: boolean;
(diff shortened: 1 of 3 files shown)


Amerlander:
Unfortunately the integration test issues that the iframe uses `sandbox='allow-scripts allow-same-origin'`. But YouTube needs these properties to embed videos. Using no sandbox tag at all would help, but is also not allowed by the integration test.

pelikhan:
We added support for playing .mp4 inline. Is it enough for you?

Amerlander:
Well, using self hosted .mp4 for inline videos could be an option and since it's a more cosmetic change it's okay to have this PR not merged yet.

pelikhan:
YouTube videos are commonly blocked in school networks. With inline hosted .mp4 videos, you don't have this problem (and they are much smaller than gifs). How long are your videos?

pelikhan:
@shakao can we bring back video support?

PR #6174, opened 2019-11-05
Cache API info for external packages in indexeddb

riknoll:
Adds caching for external packages as well. I key the database on the package name, which should be fine unless you have multiple projects with different versions of the package installed. Seems like a small scenario. Most the db code is based off of @pelikhan's translation work

### DIFF
--- a/webapp/src/compiler.ts
+++ b/webapp/src/compiler.ts
@@ -817,9 +817,11 @@ class ApiInfoIndexedDb {
 
         return this.db.getAsync(ApiInfoIndexedDb.TABLE, key)
             .then((value: ApiInfoIndexedDbEntry) => {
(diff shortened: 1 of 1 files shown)


pelikhan:
what happens when you update an extension and the update adds new blocks?

pelikhan:
The multiple version scenario is not so unlikely because the extensions are hardbound. Say 2 projects like to extension v1. Then project A updates extension to v2. You now have 2 version of the extension in flight.

abchatra:
It is handled as the hash will mismatch. Not optimized in that scenario right?

pelikhan:
I see. Looks good. It's kind of weird that we have a persistent cache for the API data and not for the sources. Can we get both?

riknoll:
Yeah, I can cache the sources. Will do in next PR

Issue #6177, opened 2019-11-06
pxt serve --rebundle broken with latest master


pelikhan:
``pxt serve`` works but rebundle option is broken.

```
$ pxt serve --rebundle
Using target arcade with build engine codal
  target: v0.14.29 pxt-arcade
  pxt-core: v5.28.12 pxt
building node_modules/pxt-core...
[run] cd node_modules/pxt-core; npm.cmd run build

> pxt-core@5.28.12 build pxt-arcade\node_modules\pxt-core
> jake

cp -r libs/pxt-common/pxt-core.d.ts docs/static/playground/pxt-common/pxt-core.d.js
cp -r libs/pxt-common/pxt-helpers.ts docs/static/playground/pxt-common/pxt-helpers.js
local pxt-core built.
building pxt-arcade\node_modules\pxt-common-packages\sim...
[run] cd pxt-arcade\node_modules\pxt-common-packages\sim; node ../node_modules/typescript/bin/tsc
log strings: 20 files; 0 strings -> sim-strings.json
copying common-sim...
building target.json in pxt-arcade...
building libs/animation

***
*** Build failed: target build failed: Cannot read property 'options' of null
TypeError: Cannot read property 'options' of null
    at built\pxt.js:149399:30
    at tryCatcher (node_modules\bluebird\js\release\util.js:16:23)
    at Promise._settlePromiseFromHandler (node_modules\bluebird\js\release\promise.js:512:31)
    at Promise._settlePromise (node_modules\bluebird\js\release\promise.js:569:18)
    at Promise._settlePromise0 (node_modules\bluebird\js\release\promise.js:614:10)
    at Promise._settlePromises (node_modules\bluebird\js\release\promise.js:693:18)
    at Promise._fulfill (node_modules\bluebird\js\release\promise.js:638:18)
    at Promise._resolveCallback (node_modules\bluebird\js\release\promise.js:432:57)
    at Promise._settlePromiseFromHandler (node_modules\bluebird\js\release\promise.js:524:17)
    at Promise._settlePromise (node_modules\bluebird\js\release\promise.js:569:18)
    at Promise._settlePromise0 (node_modules\bluebird\js\release\promise.js:614:10)
    at Promise._settlePromises (node_modules\bluebird\js\release\promise.js:693:18)
    at Promise._fulfill (node_modules\bluebird\js\release\promise.js:638:18)
    at Promise._resolveCallback (node_modules\bluebird\js\release\promise.js:432:57)
[...]

PR #6178, opened 2019-11-06
don't crash on --rebundle

jwunderl:
I'm getting a crash on rebundle because it hits the `null` path and tries to destructure null


jwunderl:
Oh this fixes https://github.com/microsoft/pxt/issues/6177, didn't see that

webapp/src/compiler.ts
```diff
@@ -415,6 +415,9 @@ interface BundledPackage {
 }
 
 async function getCachedApiInfoAsync(project: pkg.EditorPackage, bundled: pxt.Map<pxt.PackageApiInfo>): Promise<pxtc.ApisInfo> {
+    if (!bundled || !bundled.length)
```
riknoll: Not an array

PR #6179, opened 2019-11-06
Fixing two API cache bugs

riknoll:
1. In Minecraft we have a package that overrides APIs in the core package, so tweaking the cache logic so that it will override the APIs if they are different. Nothing changes for GitHub packages
1. Fixes the `custom.ts` scenario for exporting blocks from the user project


riknoll:
The tests passed, I'm not sure why GitHub didn't update the checks

PR #6187, opened 2019-11-07
support for setting large image as .mp4

pelikhan:
We serve a large amount of large .gifs that could be 5x smaller as .mp4s.
- [x] support .mp4 files in largeimageurl
- [x] bail out of .gif in electron
diff

### DIFF
--- a/webapp/src/projects.tsx
+++ b/webapp/src/projects.tsx
@@ -710,9 +710,14 @@ export class ProjectsDetail extends data.Component<ProjectsDetailProps, Projects
     renderCore() {
         const { name, description, imageUrl, largeImageUrl, youTubeId, cardType, tags } = this.props;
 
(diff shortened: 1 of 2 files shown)


webapp/src/projects.tsx
```diff
@@ -745,7 +750,8 @@ export class ProjectsDetail extends data.Component<ProjectsDetailProps, Projects
 
         return <div className="ui grid stackable padded">
             {image && <div className="imagewrapper">
-                <div className="image" style={{ backgroundImage: `url("${image}")` }} />
+                {/\.mp4$/.test(image) ? <video className="video" src={image} autoPlay={true} controls={false} loop={true} playsinline={true} />
```
jwunderl:
```suggestion
                {/\.mp4$/.test(image) ? <video className="video" src={image} autoPlay={true} controls={false} loop={true} playsInline={true} />
```

playsinline  is the property of the dom element, but playsInline is the one understood by tsx / react

pelikhan:
For now, don't package .mp4 in the electron app.

PR #6188, opened 2019-11-07
Changing the MemTranslationDb time entry

abchatra:
This used to 24 hours ahead of current time (Date.now). But indexdb stored just the current time. 
This means once cached in the MemTranslationDb the cache never expires for that session. 

It is unclear how this is better, when the time stamp just should match the existing timestamp. Caller of the getAsync decides how to expire the cache.

### DIFF
--- a/pxtlib/browserutils.ts
+++ b/pxtlib/browserutils.ts
@@ -659,16 +659,16 @@ namespace pxt.BrowserUtils {
         getAsync(lang: string, filename: string, branch: string): Promise<ITranslationDbEntry> {
             return Promise.resolve(this.get(lang, filename, branch));
         }
-        set(lang: string, filename: string, branch: string, etag: string, strings?: pxt.Map<string>, md?: string) {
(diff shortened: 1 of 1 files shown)


abchatra:
I don't think there is any actual bug because of this as usually inmemorydb will be destroyed when someone closes the tab, which should not be longer than couple of hours

pxtlib/browserutils.ts
```diff
@@ -659,16 +659,16 @@ namespace pxt.BrowserUtils {
...
                 etag,
-                time: Date.now() + 24 * 60 * 60 * 1000, // in-memory expiration is 24h
+                time,
                 strings,
                 md
             }
         }
         setAsync(lang: string, filename: string, branch: string, etag: string, strings?: pxt.Map<string>, md?: string): Promise<void> {
-            this.set(lang, filename, branch, etag, strings);
+            this.set(lang, filename, branch, etag, Date.now(), strings);
```
pelikhan: i think we have something called Util.now()

PR #6191, opened 2019-11-07
Stay in editor after loading tutorial from hash

shakao:
Kind of an edge case, but if the tutorial is loading and the window refreshes before the #editor is added (pretty hard to hit in the wild, but can repro consistently with a throttled connection + timed refreshing), it will load the home screen. Only comes up in Minecraft, because they are heavily using the #tutorial hash command and often have to work on a slow connection, I guess.


pelikhan:
eventually we should have a proper hash navigation mechanism. all ui navigation should go through a hash change. until then, we'll keep facing this kind of issues.

PR #6194, opened 2019-11-07
Delete templateCode after project is saved, in case of reload during decompile/save

shakao:


jwunderl (dismissed):

webapp/src/app.tsx
```diff
@@ -1318,7 +1314,11 @@ export class ProjectView
                 })
         }
 
-        return decompilePromise.then(() => workspace.saveAsync(header));
+        return decompilePromise.then(() => {
+            // Delete from header after adding template, so that we don't
+            // overwrite the user code if the tutorial is re-opened
+            workspace.saveAsync(header).then(() => { delete header.tutorial.templateCode })
```
pelikhan: missing return

webapp/src/app.tsx
```diff
@@ -1321,7 +1317,11 @@ export class ProjectView
                 })
         }
 
-        return decompilePromise.then(() => workspace.saveAsync(header));
+        return decompilePromise.then(() => {
+            // Delete from header after adding template, so that we don't
+            // overwrite the user code if the tutorial is re-opened
+            return workspace.saveAsync(header).then(() => { delete header.tutorial.templateCode })
```
pelikhan: should you check that header.tutorial is defined?

pelikhan:
Is this not needed? @shakao

pelikhan:
@shakao is this PR still needed or can we close it?

PR #6200, opened 2019-11-09
Add defaults when loading api cache, and remove them when saving

mmoskal:
Could you take a look if this makes sense? It seems to work for me.

It cuts the size of the API info from 31M to 11M and compressed size from 2.6M to 2.0M (1.2M of the compressed size is sources of our packages).

In future we could use the same scheme for cached API info in indexdb.

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -1905,7 +1905,7 @@ function buildSimAsync() {
 function compressApiInfo(inf: Map<pxt.PackageApiInfo>) {
     function leanSymbol(sym: pxtc.SymbolInfo) {
         const isEmpty = (v: any) => !v || Object.keys(v).length == 0
-        const attrs = U.clone(sym.attributes)
+        let attrs = U.clone(sym.attributes)
         if (attrs.callingConvention == 0)
             delete attrs.callingConvention
         if (isEmpty(attrs.paramDefl))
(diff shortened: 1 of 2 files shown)


riknoll:
@mmoskal this broke the monaco toolbox; I think it's removing all of the snippets

mmoskal:
Whoops, didn't realize it was also used for monaco.

mmoskal:
Hmmm... I don't see snippets coming in:

```
[16:00 pxt-32]$ grep snippet temp/api-cache.json |uniq -c
   3339      "snippet": null,
      1       "snippet": "img` `",
   2995      "snippet": null,
[16:00 pxt-32]$
```

I did this change to make sure I don't compress `api-info.json`:

```diff
diff --git a/cli/cli.ts b/cli/cli.ts
index 36eba0c54..6444b7c36 100644
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -2081,8 +2081,8 @@ function buildTargetCoreAsync(options: BuildTargetOptions = {}) {
                 });
             }

+            nodeutil.writeFileSync(apiInfoPath, JSON.stringify(builtInfo, null, 1));
             cfg.apiInfo = compressApiInfo(builtInfo);
-            nodeutil.writeFileSync(apiInfoPath, JSON.stringify(cfg.apiInfo));

             let info = travisInfo()
             cfg.versions = {
```

jwunderl:
Hm, a handful more weird bugs that all seem to stem from this change:

Kinds in arcade not showing up, and events with the fake variable draggable params broken: https://github.com/microsoft/pxt-arcade/issues/1457

fixed instance missing for optional params, and not all fixed instances showing up in dropdowns (at a glance I think it's missing the subclasses e.g. effects like StarField don't seem to be in the list anymore, even though they extend ParticleEffects): https://github.com/microsoft/pxt-arcade/issues/1455 

and https://github.com/microsoft/pxt-arcade/issues/1456 (which is the bug I found yesterday / richard reported here just above)

mmoskal:
I have a PR coming later today

PR #6204, opened 2019-11-12
Fix micro bit static pkg / local serve pathing

jwunderl:
fixes https://github.com/microsoft/pxt-microbit/issues/2514, and a few small bugs I saw along the way (e.g. `.html` files were being converted to ``.h.html`` files).

this was failing because`resolveMD` was failing to locate the files listed as dependencies of bundleddirs (e.g. http://localhost:3232/device/usb/troubleshooting goes to not found even though https://makecode.microbit.org/device/usb/troubleshooting works) -- so I fixed that by including the dependencies in those look ups in the same way https://github.com/microsoft/pxt/pull/6196 did.

Also, building a staticpkg is the only place that we already have the md file readily available when calling it so it seems better to just pass that along anyways.

Also did a quick pass at trying to clean things up a little bit to ideally make it easier to follow along in the future - e.g. using const to clarify when things will mutate, etc.

~I'll take another look at where bundleddirs is referenced real quick too, in case there are some other obvious spots that getting all of the dependencies as well would be needed~ Seems like we should probably grab bundled dirs in the same way in https://github.com/microsoft/pxt/blob/master/cli/crowdin.ts - it doesn't look like it will upload the dependencies docs there currently? @pelikhan

### DIFF
--- a/cli/nodeutil.ts
+++ b/cli/nodeutil.ts
@@ -506,7 +506,7 @@ export function resolveMd(root: string, pathname: string, md?: string): string {
 
     const dirs = [
         path.join(root, "/node_modules/pxt-core/common-docs"),
-        ...getBundledPackages()
+        ...getBundledPackagesDocs()
     ];
 
     for (const d of dirs) {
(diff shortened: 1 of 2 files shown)


cli/nodeutil.ts
```diff
@@ -489,43 +489,76 @@ export function fileExistsSync(p: string): boolean {
...
+            return;
+
+        resolvedDirs[dir] = true;
+
+        const jsonDir = path.join(dir, "pxt.json");
+        const pxtjson = fs.existsSync(jsonDir) && (readJson(jsonDir) as pxt.PackageConfig);
+
+        if (pxtjson) {
+            if (pxtjson.additionalFilePath) {
+                getPackageDocs(path.join(dir, pxtjson.additionalFilePath), folders, resolvedDirs);
```
mmoskal: I think this needs to be recursive - the pxt.json in additionalFilePath can have additionalFilePath, eg mixer---samd in Arcade would point to mixer---samd in common, which would point to mixer in common which has docs

cli/nodeutil.ts
```diff
@@ -489,43 +489,76 @@ export function fileExistsSync(p: string): boolean {
...
+            }
+
+            if (pxtjson.dependencies) {
```
mmoskal: this is probably not needed, since you're going over all bundled packages anyways

cli/nodeutil.ts
mmoskal: though I guess you won't see it now, because you're going over all bundled packages, which will include "mixer" in Arcade pointing to "mixer" in common. maybe that's enough.

jwunderl: Ah, I am recursing here already - for each package, I'm starting a depth first traversal of each other package included as an `additionalFilePath` or in the `dependencies` to make sure all dependencies are included before the files that depend on them; I'll add some comments here as there are some invariants that need to be maintained to make sure everything gets outputted properly

pelikhan:
@jwunderl good catch i don't think the crowdin stuff was updated to support docs in bundled dirs

PR #6205, opened 2019-11-12
Blocks screenshot

pelikhan:
Add diff screenshot when creating a commit in github + cleanup some incorrect typings

Demo see comment on https://github.com/pelikhan/pxt-untitled-21/commit/6572b0de8fd860d32081e75ceba564714f235cab

### DIFF
--- a/webapp/src/workspace.ts
+++ b/webapp/src/workspace.ts
@@ -44,9 +44,11 @@ function lookup(id: string) {
 }
 
 export function gitsha(data: string, encoding: "utf-8" | "base64" = "utf-8") {
-    if (encoding == "base64")
-        data = ts.pxtc.decodeBase64(data);
-    return (sha1("blob " + U.toUTF8(data).length + "\u0000" + data) + "")
+    if (encoding == "base64") {
+        const d = atob(data);
+        return (sha1("blob " + d.length + "\u0000" + d) + "")
+    } else
+        return (sha1("blob " + U.toUTF8(data).length + "\u0000" + data) + "")
 }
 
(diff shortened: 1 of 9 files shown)


webapp/src/workspace.ts
```diff
@@ -43,8 +43,12 @@ function lookup(id: string) {
     return allScripts.filter(x => x.header.id == id || x.header.path == id)[0]
 }
 
-export function gitsha(data: string) {
-    return (sha1("blob " + U.toUTF8(data).length + "\u0000" + data) + "")
+export function gitsha(data: string, encoding: "utf-8" | "base64" = "utf-8") {
+    if (encoding == "base64") {
+        const d = atob(data);
+        return (sha1("blob " + d.length + "\u0000" + d) + "")
```
pelikhan: still no luck with this SHA @mmoskal

mmoskal: It seems crypto-js is doing UTF8 parse. I think we may need something like this:

```
sha1(CryptoJS.enc.Utf8.parse("blob " + len + "\u0000").concat(
CryptoJS.enc.Base64.parse(data)))
```​

https://cryptojs.gitbook.io/docs/#encoders

pelikhan:
That’d be nice... unfortunately I don’t really have the diff on the api info (I only have current). However I think we should be able to generate a rendering of all blocks in the extension (no diff).

jwunderl:
Ahh, thats what I meant actually - if it adds blocks to the workspace / has block definitions. Diff could just be the changed images in the commit itself

pelikhan:
GitHub doesn’t do image diffing unfortunately but we could do an overlay of both. Slurp the old image as a background and blit the new one on top

PR #6206, opened 2019-11-12
Crowdin bundled deps

jwunderl:
re: https://github.com/microsoft/pxt/pull/6204, it looks like these were just pushing / pulling the translations for the bundledDirs themselves, not dependencies of those packages.

I believe this *should* fix that, but I have no clue how to test this / if I'm even able to at the moment :)


pelikhan:
To test this, you need to add a “test mode” to the uploadTrgTranslations command that skips the actual call to crowdin

PR #6213, opened 2019-11-13
Pull merge push

pelikhan:
Before doing a commit, pull changes and merge. If conflicts, bail out. This should take care of most conflict issues and we should never have to resort to creating a PR to handle conflicts.

### DIFF
--- a/pxtlib/github.ts
+++ b/pxtlib/github.ts
@@ -1300,7 +1300,7 @@ namespace pxt.github {
     }
 
     export function hasMergeConflictMarker(content: string) {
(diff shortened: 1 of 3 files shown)


pxtlib/github.ts
```diff
@@ -1299,6 +1299,10 @@ namespace pxt.github {
         }
     }
 
+    export function hasMergeConflictMarker(content: string) {
+        return content && /^(<<<<<<<[^<]|=======|>>>>>>>[^>])/m.test(content);
```
mmoskal: I would check only for <<<>>>> not === because ===== may flag markdown files

PR #6217, opened 2019-11-14
Pair dialog

pelikhan:
- [x] Extend pair logic to display pairing selector when multiple options are available (USB/BLE). 
- [x] "pair device" button shows up in regular download dialog

**Note the dialog is not shown when only 1 connection option is available.** WebBLE is still optional and enabled in this demo.

See https://github.com/microsoft/pxt-microbit/pull/2521

repro
repro


abchatra:
Doesn't webble require switching on experimental features and entering code displayed in the microbit etc? I don't understand this change.

abchatra:
And the instruction in that UI is asking you to connect the USB cable which is not true for BLE. This needs a UI design.

pelikhan:
I enabled WebBLE for the recording, if it is not enabled, the dialog is skipped.

pelikhan:
The USB dialog only shows up for USB pairing

pelikhan:
The gif is misleading because it is dialog then BLE loading screen.

PR #6221, opened 2019-11-15
Fixing language code for packages

abchatra:
We should normalize the languages as they are case sensitive in package _locale folder.
Only fall back to baselang if the full lang is not available

### DIFF
--- a/pxtlib/package.ts
+++ b/pxtlib/package.ts
@@ -676,12 +676,16 @@ namespace pxt {
 
         bundledStringsForFile(lang: string, filename: string): Map<string> {
             let r: Map<string> = {};
+
+            let [initialLang, baseLang]  = pxt.Util.normalizeLanguageCode(lang);
             const files = this.config.files;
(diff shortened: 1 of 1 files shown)


abchatra:
Testing it tonight.

abchatra:
@pelikhan We can't swap between two different locales even if their prefix is same. zh-CN and zh-TW are completely different languages and can't be substituted for one another. Similarly Norwegian nn & nb-NO. Even their prefix is not same.

We can only substitute where prefix language is present. Like "en" for en-GB or en-IN and fr for "fr-CA" etc

These checks are consistent with the backend and other places in the code.

pelikhan:
I understand that. Conceptually if you treat the array returned by the function contains a list of compatible locales, ordered by compatibility preference, then it boils down to
```
const files = this.config.files;
const f = pxt.Util.normalizeLanguageCode(lang).map(code => `_locales/${code}/${filename}-strings.json`).find(fn => files.indexOf(fn) > -1)
const r = fn && JSON.parse(this.readFile(fn)) as Map<string>;
```

PR #6228, opened 2019-11-15
Ask use to prepare repo when importing non-empty repo

pelikhan:
When MakeCode tries to import an existing repository, it scan the existing files for pxt.json. 
If pxt.json is missing, it will ask the user if the project can be initialized. This happens when user creates repo with README.md for example.

- [x] show confirmation dialog to initialize repo
- [x] don't generate .travis.yml anymore
- [x] better pxt.json recostruction logic using target template

### DIFF
--- a/cli/server.ts
+++ b/cli/server.ts
@@ -195,19 +195,19 @@ function writePkgAsync(logicalDirname: string, data: FsPkg) {
                 if (f.name == pxt.CONFIG_NAME) {
                     try {
                         if (!pxt.Package.parseAndValidConfig(f.content)) {
-                            console.log("Trying to save invalid JSON config")
-                            console.log(f.content);
+                            pxt.log("Trying to save invalid JSON config")
+                            pxt.debug(f.content);
                             throwError(410)
                         }
                     } catch (e) {
-                        console.log("Trying to save invalid format JSON config")
-                        console.log(e)
-                        console.log(f.content);
+                        pxt.log("Trying to save invalid format JSON config")
+                        pxt.log(e)
+                        pxt.debug(f.content);
                         throwError(410)
                     }
                 }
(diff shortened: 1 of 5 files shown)


cli/server.ts
```diff
@@ -194,13 +194,15 @@ function writePkgAsync(logicalDirname: string, data: FsPkg) {
             .then(buf => {
                 if (f.name == pxt.CONFIG_NAME) {
                     try {
-                        let cfg: pxt.PackageConfig = JSON.parse(f.content)
-                        if (!cfg.name) {
+                        if (!pxt.Package.parseAndValidConfig(f.content)) {
                             console.log("Trying to save invalid JSON config")
+                            console.log(f.content);
```
riknoll: pxt.debug?

webapp/src/workspace.ts
```diff
@@ -927,29 +928,48 @@ export async function importGithubAsync(id: string): Promise<Header> {
...
         sha = await pxt.github.getRefAsync(parsed.fullName, parsed.tag)
+        // if the repo does not have a pxt.json file, treat as empty 
+        // (must be done before)
+        const commit = await pxt.github.getCommitAsync(parsed.fullName, sha)
+        if (!commit.tree.tree.find(f => f.path == pxt.CONFIG_NAME)) {
+            pxt.log(`github: detected import non-makecode project`)
+            isEmpty = true; // needs initialization
+            forceTemplateFiles = false;
+            // ask user before modifying project
+            const r = await core.confirmAsync({
```
riknoll: Does clicking confirm here also delete non-pxt files?

pelikhan: I've added ``Your existing files will not be modified.``. It leaves all the other files alone.

PR #6230, opened 2019-11-15
missing draggable blocks fix

pelikhan:
And they are back!!!
Fix for https://github.com/microsoft/pxt-arcade/issues/1462


FranklinWhale:
Probably still an issue in `microbit`: microsoft/pxt-microbit#2523

PR #6240, opened 2019-11-19
fix tutorial colors in high contrast mode

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/1470

(also makes the faded next button show up as white instead of black)


pelikhan:
screenshot?

jwunderl:
current:


after change:

PR #6244, opened 2019-11-19
More minification (API Info and JavaScript)

riknoll:
Includes the following changes:

1. Remove the redundant JRES from the cached API info; this shrinks `target.js` in minecraft back to its pre-api-cache size
1. Minify ALL production JavaScript instead of just `main.js` (speeds up load by about ~5 seconds in Chrome with fast 3g network throttling)
1. Stop pretty printing built target files for no reason in production builds

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -455,7 +455,7 @@ function bumpPxtCoreDepAsync(): Promise<void> {
                     U.userError(`Trying to automatically update major version, please edit package.json manually.`)
                 }
                 pkg["dependencies"][knownPackage] = newVer
-                nodeutil.writeFileSync("package.json", JSON.stringify(pkg, null, 2) + "\n")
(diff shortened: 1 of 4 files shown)


FranklinWhale:
`pxt staticpkg` has a `minify` option. Should `process.env["PXT_ENV"]` be set to `production` when `minify` is enabled in `pxt staticpkg`?

cli/cli.ts
```diff
@@ -2897,7 +2901,7 @@ export function initAsync(parsed: commandParser.ParsedCommand) {
 
     return initPromise
         .then(() => {
-            files[pxt.CONFIG_NAME] = JSON.stringify(configMap, null, 4);
+            files[pxt.CONFIG_NAME] = nodeutil.stringify(configMap);
```
pelikhan: use pxt.Package.stringifyConfig(configMap) to keep the endline consistent.

Issue #6241, opened 2019-11-19
CLI uploading unwanted files


pelikhan:
The new logic to pull files from common packages is uploading "junk" files into microbit. As a result, translations are lost https://github.com/microsoft/pxt-microbit/issues/2527

https://travis-ci.org/microsoft/pxt-microbit/jobs/613599901#L2051

Mitigation:
- [x] disabling crowdin upload in master (we should do that for all released editors / flag added https://github.com/microsoft/pxt/pull/6242)
- [x] bumping pxt-microbit in stable2.0 to restore translation. If not enough, manually revert versions in crowdin.
- [x] adding test mode to ``pxt uploadtrgtranslations --test``. It goes through all the files but fakes the call to crowdin.

pelikhan:
This is probably the problem: https://github.com/microsoft/pxt/commit/351f1ea83e2c55c1601455974b61f4c4a7e15b80#diff-87390c8316b0230db07feb98ebca1fc0

pelikhan:
Post mortem: we need to restrict crodwin uploads to stable branches -- or -- use more branches in crowdin. Let's have a meeting to avoid this in the future.

pelikhan:
docs recovery still in progress https://travis-ci.org/microsoft/pxt-microbit/builds/614017441#L1821

jwunderl:
Well presumably the problem would be in https://github.com/microsoft/pxt/pull/6206, so should revert that to start?

pelikhan:
Yes revert for starter

pelikhan:
https://github.com/microsoft/pxt/pull/6243

pelikhan:
I will add a test mode

pelikhan:
@jwunderl you can use ``pxt uploadtrgtranslations --test`` to see which files get uploaded and where in crowdin.

PR #6246, opened 2019-11-20
Bundle docs from dependencies of bundleddirs

jwunderl:
fix https://github.com/microsoft/pxt/issues/6241

Issue was that it was using the `getBundledPackagesDocs ` in `uploadBundledTranslationsAsync`, when it shouldn't have cared about anything but the dirs explicitly listed in bundleddirs.

Also updates `pxt uploadtrgtranslations --test` to dump the files it would have uploaded into `temp/crowdin` so that it's easier to validate changes; e.g. after this change the contents of [microbit.zip](https://github.com/microsoft/pxt/files/3867396/microbit.zip) would be produced in `temp/crowdin/microbit`.

Starting as draft as I don't see a reason to merge this right away (at least until after figuring out what to do with crowdin) - the only target that seems to be significantly negatively affected by this at the moment seems to be arcade, where very few of the ref pages currently were translatable (e.g. anything from [game/docs](https://github.com/microsoft/pxt-common-packages/tree/master/libs/game/docs) wasn't available before, though it looks like the current state of node_modules is going to persist for the time being [on crowdin](https://crowdin.com/project/kindscript/en#/arcade/node_modules))

### DIFF
--- a/cli/crowdin.ts
+++ b/cli/crowdin.ts
@@ -114,11 +114,11 @@ function uploadDocsTranslationsAsync(srcDir: string, crowdinDir: string, branch:
 
 function uploadBundledTranslationsAsync(crowdinDir: string, branch: string, prj: string, key: string): Promise<void> {
     const todo: string[] = [];
-    nodeutil.getBundledPackagesDocs().forEach(docsDir => {
-        const locdir = path.join(docsDir, "..", "_locales");
+    pxt.appTarget.bundleddirs.forEach(dir => {
+        const locdir = path.join(dir, "_locales");
         if (fs.existsSync(locdir))
             fs.readdirSync(locdir)
-                .filter(f => /-strings\.json$/i.test(f) && !/node_modules/.test(f))
+                .filter(f => /-strings\.json$/i.test(f))
                 .forEach(f => todo.push(path.join(locdir, f)))
     });
 
--- a/cli/crowdin.ts
+++ b/cli/crowdin.ts
@@ -118,7 +118,7 @@ function uploadBundledTranslationsAsync(crowdinDir: string, branch: string, prj:
         const locdir = path.join(docsDir, "..", "_locales");
         if (fs.existsSync(locdir))
             fs.readdirSync(locdir)
-                .filter(f => /strings\.json$/i.test(f))
+                .filter(f => /-strings\.json$/i.test(f) && !/node_modules/.test(f))
                 .forEach(f => todo.push(path.join(locdir, f)))
     });
 
--- a/cli/crowdin.ts
+++ b/cli/crowdin.ts
@@ -60,14 +60,14 @@ export function internalUploadTargetTranslationsAsync(uploadDocs: boolean) {
                             pxt.log("uploading docs...");
                             return uploadDocsTranslationsAsync("docs", crowdinDir, cred.branch, cred.prj, cred.key)
                                 // scan for docs in bundled packages
-                                .then(() => Promise.all(pxt.appTarget.bundleddirs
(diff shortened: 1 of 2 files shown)


pelikhan:
Could generate a flat file local file -> crowdin file (including branch). This file would be commuted so that we can see any suspicious move in the diffs

jwunderl:
I added a result output file and a `baseline` option; so e.g. ``pxt uploadtrgtranslations --test --baseline temp/crowdin/fake-baseline.json`` with a `fake-baseline.json` that just has a few random fake edits added on top produces some info like the stuff below; is that along the lines of what you're talking wanting? Or to pull directly from crowdin and compare against current state?

(also I probably need to clean this up a bit before merging / centralize test state better)

```
--- Comparing Result against baseline ---
Missing: fake file
Different hash: microbit/docs/tutorials/hour-of-code/hey-microbit.md - previously: 4100 chars, now 4095 chars
Different hash: microbit/docs/types/buffer.md - previously: 4666 chars, now 4582 chars
New file: microbit/docs/reference/music/ring-tone.md
New file: microbit/docs/reference/music/set-play-tone.md
New file: microbit/docs/reference/music/set-tempo.md
New file: microbit/docs/reference/music/stop-melody.md
New file: microbit/docs/reference/music/tempo.md
New file: microbit/docs/tutorials/getting-started.md
New file: microbit/docs/types/string.md
New file: microbit/docs/types.md
New file: microbit/docs/uploader.md
```

jwunderl:
and here's result of running that on microbit with the baseline from the non-test parts of this pr temporarily reverted

cli/cli.ts
```diff
@@ -5897,7 +5897,12 @@ ${pxt.crowdin.KEY_VARIABLE} - crowdin key
         help: "upload translations for target",
         flags: {
             docs: { description: "upload markdown docs folder as well" },
-            test: { description: "test run, do not upload files to crowdin"}
+            test: { description: "test run, do not upload files to crowdin"},
+            baseline: {
+                description: "baseline to compare test result against",
```
pelikhan: to compare the list of uploaded results

pelikhan:
I think the baseline report should be run on every upload -- regardless of the testmode flag.

cli/crowdin.ts
```diff
@@ -56,26 +92,61 @@ export function internalUploadTargetTranslationsAsync(uploadDocs: boolean) {
                     .then(() => fs.existsSync("built/sim-strings.json") ? execCrowdinAsync("upload", "built/sim-strings.json", crowdinDir) : Promise.resolve())
                     .then(() => uploadBundledTranslationsAsync(crowdinDir, cred.branch, cred.prj, cred.key))
                     .then(() => {
-                        if (uploadDocs) {
+                        if (uploadDocs || pxt.crowdin.testMode) {
```
pelikhan: why are you overloading here?

pelikhan:
It's not clear who will compute the baseline? is it a file updated on every build?

jwunderl:
I just had it for testing changes to upload behavior - that is, run it off master to generate the baseline, save that, do that has an effect on crowdin behavior, and then run this off whatever changes you made. As written it won't work for a CI sort of thing.

If you want it to run as part of CI / on every upload, I'll need to think more on how to best do that, as dropping a file as a build artifact to compare against would likely get way too annoying.

cli/crowdin.ts
jwunderl: Temporary (will add a todo to remove before un-drafting this) - just forcing it so I can see the results of the changes without adding the flags to every other repo

Issue #6136, opened 2019-10-30
local workspaces get published to github


tballmsft:
Scenario:

- I have a github extension E
- I have a script S using E; S also is backed by github.
- Both S and E are loaded into my local cache, so there is a local "workspace" reference from S to E
- If I commit S to github, it will persist the workspace reference to E in the pxt.json

Instead, the workspace reference should be replaced by a github reference.

Related issue: if I modified E then E should be committed and pushed before S. This applies recursively.

pelikhan:
What happens if you have uncommited changes in E, and commit fails? I think we should tell you to commit all dependent libraries before allowing S to be pushed.

pelikhan:
Fixed.

PR #6247, opened 2019-11-20
Rewrite github project references

pelikhan:
Fix for https://github.com/microsoft/pxt/issues/6136

This change handles gracefully the publishing of github project with local references. 

When a project gets published, 
* all github local dependencies (workspace:*) get rewritten to github references (github:*)
* the tag is used otherwise the branch is used
* if the commit is used to create tag, tag requirement is enforced.

Scenario:

repoA and repoB are both used locally. repoB has a dependency on repoA. When commiting, reference to repoA workspace:... should be converted to github:...


webapp/src/data.ts
```diff
@@ -122,7 +122,7 @@ function unsubscribe(component: AnyComponent) {
 
 function expired(ce: CacheEntry) {
     if (!ce.api.expirationTime)
-        return ce.data != null;
+        return !ce.lastRefresh; // needs to be refreshed at least once
```
pelikhan: stumbled on this one: if a api is Async with no expirationTime, it is never called.

webapp/src/package.ts
```diff
@@ -665,9 +674,30 @@ data.mountVirtualApi("open-pkg-meta", {
...
             files.filter(f => !!f.diagnostics).forEach(f => r.diagnostics.concat(f.diagnostics));
         }
-        const hasGit = f.getPkgId() == "this" && !!(f.header && f.header.githubId)
-        if (hasGit)
-            r.numFilesGitModified = files.filter(f => f.baseGitContent != f.content).length;
+        return r;
+    }
+})
+
+export interface PackagetGitStatus {
```
jwunderl:
```suggestion
export interface PackageGitStatus {
```

Think I got all the places it's refd

webapp/src/package.ts
```diff
@@ -665,9 +674,30 @@ data.mountVirtualApi("open-pkg-meta", {
...
+}
+
+/*
+    pkg-git-status:<guid>
+*/
+data.mountVirtualApi("pkg-git-status", {
+    getSync: p => {
+        p = data.stripProtocol(p)
+        const f = allEditorPkgs().find(pkg => pkg.header && pkg.header.id == p);
+        const r: PackagetGitStatus = {};
```
jwunderl:
```suggestion
        const r: PackageGitStatus = {};
```

webapp/src/githubbutton.tsx
```diff
@@ -59,10 +59,8 @@ export class GithubButton extends sui.UIElement<GithubButtonProps, GithubButtonS
                 onClick={this.createRepository} />
 
         // existing repo
-        //const targetTheme = pxt.appTarget.appTheme;
-        const mainPkg = pkg.mainEditorPkg()
-        const meta: pkg.PackageMeta = this.getData("open-pkg-meta:" + mainPkg.getPkgId());
-        const modified = meta && !!meta.numFilesGitModified;
+        const meta: pkg.PackagetGitStatus = this.getData("pkg-git-status:" + header.id);
```
jwunderl:
```suggestion
        const meta: pkg.PackageGitStatus = this.getData("pkg-git-status:" + header.id);
```

pelikhan:
It does restore the workspace reference so you still reference the local projects in your browser but have GitHub refs in GitHub.

PR #6248, opened 2019-11-20
GitHub Oauth confirmation dialog

pelikhan:
Confirm that the user wants to sign in with github before redirecting to the oauth page. Unifies sign in dialog for token/oauth scenarios.
- [x] display dialog before sending user to github (less jaring)
- [x] don't display sign in banner in git view
- [x] ability to still use dev token

repro

### DIFF
--- a/theme/common.less
+++ b/theme/common.less
@@ -1529,3 +1529,14 @@ img.pixelart {
 a.ui.link {
     cursor: pointer;
 }
+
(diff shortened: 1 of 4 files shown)


mmoskal:
Can we have I have developer token link? The oauth sign in doesn’t let you edit in orgs.

pelikhan:
@mmoskal see gif

PR #6250, opened 2019-11-21
Add teachers common-docs page, and a few notes

jwunderl:
close https://github.com/microsoft/pxt-microbit/issues/2524

also add note from https://forum.makecode.com/t/extensions-in-custom-editor/787/4

Adds a start on a /teachers page with some helpful things for teachers, along with a few other updates (e.g. how to include github repos in docs pages, how to link to tutorials). In particular I'm not sure how clear the info on building a link to tutorials / course books.

Also, not sure on the pathing for the gifs in common-docs,  as I haven't added an image there before and it didn't work locally right away. I'll take a look around into how common-docs are handled to see if it should work?


docs/writing-docs/macros.md
```diff
@@ -206,9 +206,15 @@ To avoid changing screenshots, PXT automatically renders code snippets to blocks
 You need declare the packages required to load your snippet, unless they are part of the default empty template.
```
ganicke: "Extensions" formerly called "packages"?

jwunderl: Good catch, maybe we should add ```extensions as an alias for the packages macro too so we can phase that out over time too

pelikhan: file a bug :)

PR #6258, opened 2019-11-25
iOS 13 iPad recognition

Amerlander:
On iOS 13 the user agent for safari got changed and therefore the browser identification stopped working in the WebView on iPads running > iOS 13.

"iPad OS 13 now identifies itself as "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15) AppleWebKit/605.1.15 (KHTML, like Gecko)", which is detected neither as mobile nor desktop."
[https://stackoverflow.com/questions/58019463/how-to-detect-device-name-in-safari-on-ios-13-while-it-doesnt-show-the-correct](https://stackoverflow.com/questions/58019463/how-to-detect-device-name-in-safari-on-ios-13-while-it-doesnt-show-the-correct)

I cant test it on an actual device, so the changes should be tested by someone who can. I just got that Issue report and used google for finding a possible solution.

### DIFF
--- a/pxtlib/browserutils.ts
+++ b/pxtlib/browserutils.ts
@@ -25,7 +25,9 @@ namespace pxt.BrowserUtils {
     }
 
     export function isIOS(): boolean {
-        return hasNavigator() && /iPad|iPhone|iPod/.test(navigator.userAgent);
+        return (hasNavigator() && /iPad|iPhone|iPod/.test(navigator.userAgent)) || 
+        (/iPad|iPhone|iPod/.test(navigator.platform) ||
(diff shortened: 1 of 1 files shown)


Amerlander:
I did, but it fails integration test because `MSStream` does not exist on `window`.
This part should prevent to match against IE.

Amerlander:
Changed this now, but I'm not complete sure wether `!window.hasOwnProperty('MSStream')` has the same result as `!window.MSStream` or not.

riknoll:
You should be able to do `!(window as any).MSStream` to achieve the same effect; I think `hasOwnProperty` has some weird semantics.

pelikhan:
@abchatra does this impact arcade?

abchatra:
IsSafari check should be fine with the new user agent from ipad. As we already check for Macintosh

Only check we need is for isIOS.  We don't special case lot for isIOS. So it works today.  We should take this fix but simplify it for isIOS and do !(window as any).MSStream

PR #6259, opened 2019-11-25
rewrite target.js to use JSON.parse

pelikhan:
Still testing perf. Use JSON.parse('...') instead of pure javascript.


pelikhan:
Needs perf testing.

PR #6266, opened 2019-11-27
Round trip from blocks -> ts -> py -> ts -> blocks gives back identical blocks

darzu:
After much ado, I settled on fixing this bug by centralizing and introducing a cache around the ts->py and py->ts transpile operations. This is done in the compiler code, so in the worker.

Probably a better, longer term fix would be to implement a "MultiViewFile" abstraction in the webapp that encapsulates all language conversions (including blocks, which was the difficult part that I punted from this PR), tracking equivalence and keeps the trio [blocks, ts, py] in sync. Discussed this in person with @riknoll . This would allow us to remove a lot of code in the webapp and elsewhere that does things like "convert from A to B and then back to A to see if things have changed" and it would likely eliminate several race conditions.

But this PR is not that, it just handles the ts <-> py case.

Fixes:

Bug repro:
1. open Space Destroyer in blocks
2. convert to python and back to blocks
3. notice blocks have changed


pxtpy/converter.ts
```diff
@@ -2305,21 +2305,22 @@ namespace pxt.py {
...
     }
     export function py2ts(opts: pxtc.CompileOptions): Py2TsRes {
         let modules: py.Module[] = []
-        const generated: Map<string> = {}
+        const outfiles: Map<string> = {}
         diagnostics = []
 
         if (!opts.sourceFiles)
-            error(null, 9566, lf("missing sourceFiles"));
+            throw Error("missing sourceFiles! Cannot convert py to ts")
```
pelikhan: no more error number?

pelikhan:
Can we revert the "generated" -> "outfiles"? it would remove a bunch of changes.

pxtpy/converter.ts
darzu: In this case it wasn't helpful since this isn't anything to do with the user's code, it's an internal error that should never happen. I was the one that added the error number above but really it should be fatal.

darzu:
> Can we revert the "generated" -> "outfiles"? it would remove a bunch of changes.

Yeah but I think it's worth it. The data structure we use almost everywhere is "CompileResult" which has an "outfiles" property. For whatever reason ts2py had it's own output data structure that had a "generated" property that contained exactly the same information. This was inconvenient in this PR because I was trying to standardize the interface. "TranspileResult" is a strict subset of "CompileResult".

pxtpy/converter.ts
darzu: actually if we don't throw here, the missing property causes a null deref exception before the error is returned. That's why I converted this to a throw because there's no chance to get the error information out anyway, this just catches the error earlier.

pelikhan:
Sounds good!

pelikhan:
Can we get a build for testing?

darzu:
> Can we get a build for testing?

I forget the cmd for that.. what is it again? I can't find it in my notes

pelikhan:
pxt uploadtrg

darzu:
> pxt uploadtrg

ty. my local build isn't working (I always use --rb and there's no --rb option for uploadtrg? Is that intentional), I'll try to get it sorted out..

pelikhan:
What is -rb ?

jwunderl:
rebundle

pxtcompiler/emitter/transpile.ts
```diff
@@ -0,0 +1,115 @@
...
+// so that we can cache results for better performance and user
+// experience (since compile -> decompile round-trips are lossy)
+namespace ts.pxtc.transpile {
+    export interface TranspileResult {
+        diagnostics: pxtc.KsDiagnostic[],
+        success: boolean,
+        outfiles: { [key: string]: string }
+    }
+
+    const mainName = (l: Lang) => `main.${l}`
```
pelikhan: Just a function?

pxtcompiler/emitter/transpile.ts
```diff
@@ -0,0 +1,115 @@
...
+    const MAX_CODE_EQUIVS = 10
+
+    function toComparable(code: string): number {
+        // Ignore whitespace
+        code = code.replace(/\s/g, "")
+        return U.codalHash16(code)
+    }
+    function tryGetLanguageEquivalence(lang: Lang, txt: string): LangEquivSet | undefined {
+        let txtComp = toComparable(txt)
+        for (let eq of codeEquivalences) {
```
pelikhan: Alternatively, use .find(...) on array

darzu:
> Any ticks to be added to measure where those happen?

Like ticks inside the transpile.py2ts and .ts2py functions? It'd would be very noisy. A single ts->py->ts round trip calls those functions ~6 times for unclear reasons.

Coincidentally, I think we'll get a perf win from this PR because of that inefficiency.

I believe the important tick is when the user switched languages which I believe we have already.

pxtcompiler/emitter/transpile.ts
darzu: sure, but this is one line. It's a pattern we use liberally elsewhere.

image

pelikhan:
LGTM

pxtcompiler/emitter/transpile.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    export type Lang = "py" | "blocks" | "ts"
+    export interface LangEquivSet {
+        comparable: { [key in Lang]: number },
+        code: { [key in Lang]: string },
+    }
+    // a circular buffer of size MAX_CODE_EQUIVS that stores
+    // sets of equivalent code files so that when we translate
+    // from ts -> py -> ts we get back the same ts.
+    let codeEquivalences: LangEquivSet[] = []
```
pelikhan: when is the cache cleared? it should be cleared whenever we change header.

pxtcompiler/emitter/transpile.ts
```diff
@@ -0,0 +1,115 @@
...
+    function tryGetLanguageEquivalence(lang: Lang, txt: string): LangEquivSet | undefined {
+        let txtComp = toComparable(txt)
+        for (let eq of codeEquivalences) {
+            if (eq.comparable[lang] === txtComp) {
+                return eq
+            }
+        }
+        return undefined
+    }
+    function setLanguageEquivalence(lang1: Lang, lang1Txt: string, lang2: Lang, lang2Txt: string) {
```
riknoll: I think a clearer name would be cacheLanguageEquivalence or even just cacheTranspile. Took me a bit to realize what was going on in this file.

pxtcompiler/emitter/transpile.ts
riknoll: Actually, should be changed whenever dependencies change as well. Or maybe a hashed version of config.json should be stored with the text.

mmoskal: Don't use hashes unless necessary - storing a string has essentially no cost (just a pointer) and comparison is cheap.

darzu:
> Can we get a build for testing?

https://arcade.makecode.com/app/ae2dc3a32fd5dfc641beadfcad7f6ff70ffb8414-9ab6c5d9b6

darzu:
> project

Do you (also @pelikhan, @mmoskal ) know if anything with:

```
interface CompileOptions {
        fileSystem: pxt.Map<string>;
        target: CompileTarget;
        testMode?: boolean;
        sourceFiles?: string[];
        generatedFiles?: string[];
        jres?: pxt.Map<pxt.JRes>;
        hexinfo: HexInfo;
        extinfo?: ExtensionInfo;
        noEmit?: boolean;
        forceEmit?: boolean;
        ast?: boolean;
        breakpoints?: boolean;
        trace?: boolean;
        justMyCode?: boolean;
        computeUsedSymbols?: boolean;
        name?: string;
        warnDiv?: boolean; // warn when emitting division operator
        apisInfo?: ApisInfo;
        bannedCategories?: string[];
        skipPxtModulesTSC?: boolean; // skip re-checking of pxt_modules/*
        skipPxtModulesEmit?: boolean; // skip re-emit of pxt_modules/*

        syntaxInfo?: SyntaxInfo;

        alwaysDecompileOnStart?: boolean; // decompiler only
        allowedArgumentTypes?: string[]; // decompiler-only; the types allowed for user-defined function arguments in blocks (unlisted types will cause grey blocks)

        embedMeta?: string;
        embedBlob?: string; // base64

        /* @internal */
        ignoreFileResolutionErrors?: boolean; // ignores triple-slash directive errors; debug only
    }
```

changes when the dependencies change? Probably ApisInfo?

My goal is to create a hash representing the current dependencies.

pelikhan:
We always reload the header when dependencies change.

darzu:
> We always reload the header when dependencies change. So resetting your cache is enough to handle this case.

Which header?

darzu:
Using all the qNames as part of the cache key seems to be working well.

(btw I _FAR_ prefer to use a stronger cache key than count on some event to clear the cache. Much less error prone, easier to refactor, no races, and it's truly functional. Invalidating a cache due to some event is a last resort and it's usually an anti-pattern. Caching should simply be memoization whenever possible. IMHO.)

However, now after I've merged master it appears things have regressed. I'm getting an error when going ts->py on Space Destroyer so I'm looking into that.

riknoll:
In `CompileOptions.sourceFiles`, you should be able to look at `pxt.json`. I don't know if that would mess up @mmoskal's two tab extension editing though.

pxtcompiler/emitter/transpile.ts
darzu: Okay, so I'm using the code + all the qNames as the key (no hash). Does that work for you @pelikhan + @riknoll ?

pelikhan:
header is the internal name for a script/program. We reload the entire user package (header) when pxt.json is modified because toolbox and other stuff need to be refreshed.  At that point, you should invalidate this cache.

pxtcompiler/emitter/transpile.ts
riknoll: Extensions can modify existing qNames without adding any new ones, we do that in the Minecraft target for example. Probably better to use all of the extension names and their installed versions.

darzu: Ah okay I'll use pxt.json and/or config.json

pelikhan:
See https://github.com/microsoft/pxt/blob/cf7e498116fef6b52cf22852454b4b0c713550a5/webapp/src/monaco.tsx#L1135

pxtcompiler/emitter/transpile.ts
darzu: pxt.json isn't accessible in many of the places that call the transpilers like service.ts as far as I can tell

Using the filenames seems to capture pretty well which packages are installed but not their versions:
image

@mmoskal , do you know of a way to uniquely identify the versions of packages used from the CompileOptions? that would be the most convenient.

---
Realistically, it should be pretty rare that changing a API could cause different ts<->py transpilation. The transpilation is mostly syntactic. I can't think of any way an API could change that would keep the same .ts but different .py for example.

darzu:
> reloadHeaderAsync

I'd really think it's a bad idea to invalidate caches based on an event. This turned into a huge pain when I did the caching around the blockly flyout and it was brittle an hard to refactor around.

I think its better to just capture the dependencies in the cache key. Then they'll automatically be invalidated and there won't be any race between something trying to read the cache right when the dependencies are invalidated, etc.

pelikhan:
Watching pxt.json is not enough. If you have a local dependency, the content of the dependency may have changed in another tab (this happens with double-tab testing). This is why we reload the program when you enter a tab.

pxtcompiler/emitter/transpile.ts
darzu: Peli and I talked about this offline. I'm going to invalidate the cache based on the event because it's too difficult to do it the other way and because that's how everything else is done.

darzu: Alright, I pushed the change where we invalidate in the service.ts operations.reset handler. Caching still works and if you open a new tab I can see the cache is reset.

I think this PR is ready. @pelikhan ?

darzu:
Wait, I'm hitting an exception, need to investigate:

[ ] Fix null exception, probably https://github.com/microsoft/pxt-minecraft/issues/1472 will need to be fixed.

darzu:
> Cache needs to be cleared when project dependencies change

Fixed. See "resetCache"

PR #6269, opened 2019-11-27
Remove "debugger;" statements as there is no easy way to turn these off; instead devs can opt-in to setting breakpoints where they want

darzu:
Removes "debugger;" statements from our asserts.

I've heard support for this in the office, but want to see if there's support from everyone:
@mmoskal @pelikhan @abchatra @shakao @riknoll 

The problem is that the debugger statement is hard to opt-out of. Manually set breakpoints can be toggled on-off in the Chrome debugger or VSCode. If there's a failed assert, we'll see it in the console and it's trivial to set a breakpoint then.

We have some occasional assert fails (especially when working with multiple tabs) that for better or worse haven't been fixed for many months. I don't see us taking the stance that we must fix every assert failure always, so let's not pretend every dev will jump in the debugger and derail their other work.

### DIFF
--- a/pxtcompiler/emitter/emitter.ts
+++ b/pxtcompiler/emitter/emitter.ts
@@ -1113,7 +1113,7 @@ namespace ts.pxtc {
 
         function unhandled(n: Node, info?: string, code: number = 9202) {
             // if we hit this, and are in debugger, we probably want to look at the node
-            debugger
+            // debugger
 
             // If we have info then we may as well present that instead
             if (info) {
--- a/pxtlib/commonutil.ts
+++ b/pxtlib/commonutil.ts
@@ -10,7 +10,6 @@ import pxtc = ts.pxtc
 namespace ts.pxtc.Util {
     export function assert(cond: boolean, msg = "Assertion failed") {
         if (!cond) {
-            debugger
             throw new Error(msg)
         }
     }
--- a/pxtlib/util.ts
+++ b/pxtlib/util.ts
@@ -205,7 +205,7 @@ namespace ts.pxtc.Util {
     }
 
     export function oops(msg = "OOPS"): Error {
-        debugger
+
         throw new Error(msg)
     }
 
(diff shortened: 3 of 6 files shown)


mmoskal:
I like the debugger statements. The source locations of asserts keep changing so you have to keep moving your breakpoints.

pelikhan:
Maybe guard it with pxt.options.debug ?

jwunderl:
The only one that's been particularly annoying for me has been [`trying to access outdated session`](https://github.com/Microsoft/pxt/blob/master/webapp/src/workspace.ts#L168), which I hit multiple times a day; could we just make that one fail without `debugger`ing first, and see if there are any remaining problematic ones after that?

mmoskal:
+1

If asserts keep failing then they should be removed or fixed. Don’t we report these in AI?

darzu:
Alright, I'll see about fixing or changing that one assert.

So looks like going forward we should fix asserts that fire proactively, otherwise we end up in the situation again where we're constantly thrown into the debugger without actually fixing anything.

darzu:
Looking at the code, it seems the assumption of single tab ownership is pretty ingrained. I filed this issue: https://github.com/microsoft/pxt/issues/6276

but I'm not going to fix it now. I think that will take some serious thought. Instead I'm demoting the assert to a warning since most of the time things seem ok: https://github.com/microsoft/pxt/pull/6277

pelikhan:
I don’t think demoting the assert is the right thing to do. What happens when 2 tabs are opened on the same project?

darzu:
> I don’t think demoting the assert is the right thing to do. What happens when 2 tabs are opened on the same project?

Bad things. I think it's a likely culprit to the bug where they loose all their code. We should absolutely fix it. That's why I filed the issue.

But we aren't. At least not yet. We've explicitly decided not to fix this right now so asserting it every day isn't helpful.

pelikhan:
If you remove the assert, they will loose the oldest code. It will still have weird consequences. If we remove the assert, it will stay broken forever.

riknoll:
Alright well either someone should fix it or we should remove the debugger statement on that particular assert because it steals focus on my dev tools like ten times a day. All it's doing right now is annoying everyone.

PR #6275, opened 2019-12-02
Partial projects

pelikhan:
Annotate packages as "partial" so that we ignore build errors in the CLI. Allows to remove spurious warnings for "base"-like packages.
+ fix a bug in api shrinking.

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -2073,7 +2073,7 @@ function buildTargetCoreAsync(options: BuildTargetOptions = {}) {
     return buildWebStringsAsync()
         .then(() => options.quick ? null : internalGenDocsAsync(false, true))
         .then(() => forEachBundledPkgAsync((pkg, dirname) => {
-            pxt.log(`building ${dirname}`);
+            pxt.log(`building bundled ${dirname}`);
             let isPrj = /prj$/.test(dirname);
             const isHw = /hw---/.test(dirname);
             const config = nodeutil.readPkgConfig(".")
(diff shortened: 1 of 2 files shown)


mmoskal:
How about re-generating shims.d.ts/enums.d.ts on base? Does it have to be done manually (I usually do `pxt run` in base/core/whatever to do that).

pelikhan:
It does the entire build as before but it does not report errors for partial projects. Is that what you meant?

mmoskal:
if the C++ changes, do the shims.d.ts in base get regenerated during a full build?

pelikhan:
it's as broken as before. it runs the build but ignores the errors in "partial" projects.

cli/cli.ts
```diff
@@ -4104,7 +4106,8 @@ function buildCoreAsync(buildOpts: BuildCoreOptions): Promise<pxtc.CompileResult
                 }
             });
 
-            reportDiagnostics(res.diagnostics);
+            if (!config.partial)
```
mmoskal: so no problems with this

cli/cli.ts
```diff
@@ -2091,7 +2091,7 @@ function buildTargetCoreAsync(options: BuildTargetOptions = {}) {
                         pxt.appTarget.simulator.dynamicBoardDefinition)
                         isPrj = true
                 })
-                .then(() => options.quick ? null : testForBuildTargetAsync(isPrj || (!options.skipCore && isCore), builtInfo[dirname] && builtInfo[dirname].sha))
+                .then(() => options.quick || config.partial ? null : testForBuildTargetAsync(isPrj || (!options.skipCore && isCore), builtInfo[dirname] && builtInfo[dirname].sha))
```
mmoskal: but this seems to prevent build of any partial project; it's during the build that we generate enums.d.ts

cli/cli.ts
mmoskal: though, it would be probably good to report errors just before exit(1) regardless of partial

cli/cli.ts
pelikhan: looking

cli/cli.ts
pelikhan: displaying error on exit now

cli/cli.ts
pelikhan: always testForBuild now

Issue #6276, opened 2019-12-02
Possibly inconsistent state when working in multiple tabs: "trying to access outdated session"


darzu:
Repro:
1. open two tabs of arcade.makecode.com
2. in each one, open the debug tools
3. open a new project in each one
4. notice the "trying to access outdated session" assert (or warning) fires

It looks like the local storage code was written with the assumption that only one tab would be accessing it. This is definitely not true for many of our users. We should ensure the property consistency is implemented.

abchatra:
I believe @pelikhan fixed this.

PR #6277, opened 2019-12-02
Demote "trying to access outdated session" to a warning

darzu:
image

We've been living with this bug for probably over a year. We should fix it: https://github.com/microsoft/pxt/issues/6276

but given that we're not fixing it right now, we shouldn't keep throwing devs into the debugger so I demoted this to a warning.

### DIFF
--- a/pxtlib/main.ts
+++ b/pxtlib/main.ts
@@ -305,6 +305,10 @@ namespace pxt {
         ? (msg) => {
             console.log(msg);
         } : () => { };
+    export let warn: (msg: any) => void = typeof console !== "undefined" && !!console.log
+        ? (msg) => {
+            console.warn(msg);
+        } : () => { };
 
     export let reportException: (err: any, data?: Map<string | number>) => void = function (e, d) {
         if (console) {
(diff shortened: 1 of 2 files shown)


pelikhan:
So what happens when you have 2 projects opened in separate tabs and you are switching?

mmoskal:
you could just do `throw new Error()` if you want to keep the current behavior without being thrown into debugger.

darzu:
> So what happens when you have 2 projects opened in separate tabs and you are switching?

Bad things probably. I think it's a likely culprit to the bug where they loose all their code. We should absolutely fix it. I filed: https://github.com/microsoft/pxt/issues/6276

But we aren't. At least not yet. We've explicitly decided not to fix this right now so asserting it every day isn't helpful.

[copy-pasted from another PR]

darzu:
Looking from the git history, it looks like this consistency check has been in the code for 4 years.

darzu:
> > So what happens when you have 2 projects opened in separate tabs and you are switching?
> 
> Bad things probably. I think it's a likely culprit to the bug where they loose all their code. We should absolutely fix it. I filed: #6276
> 
> But we aren't. At least not yet. We've explicitly decided not to fix this right now so asserting it every day isn't helpful.
> 
> [copy-pasted from another PR]

@abchatra , is this an issue we want to tackle now / soon? If not, are you okay with demoting the assert to a warning?

darzu:
Closing this. The right thing is just to fix the issue.

PR #6281, opened 2019-12-03
Video URL in cards

pelikhan:
Introduce separate fields to hold potential .mp4 file location. Since the file location are expanded into blob location, file name extension mangling does not work to switch between .gif or .mp4s.

.mp4 are a much better alternative to .gif (heavy); unfortunately, Safari requires ranges which are not supported by the CDN / Electron does not ship a decoder. Hence the need to support both formats.


cli/cli.ts
```diff
@@ -1706,6 +1706,8 @@ function saveThemeJson(cfg: pxt.TargetBundle, localDir?: boolean, packaged?: boo
                             gcard.imageUrl = card.imageUrl;
                         if (card.largeImageUrl && !gcard.largeImageUrl)
                             gcard.largeImageUrl = card.largeImageUrl;
+                        if (card.videoUrl && !gcard.videoUrl)
+                            gcard.videoUrl = card.videoUrl;
```
```suggestion
                            gcard.videoUrl = card.videoUrl;
```

for build

PR #6286, opened 2019-12-04
Get translations, galleries, and targetconfig from CDN

mmoskal:


pxtlib/emitter/cloud.ts
```diff
@@ -43,17 +73,7 @@ namespace pxt.Cloud {
             options.headers["x-td-access-token"] = accessToken
```
pelikhan: is this token dead?

abchatra:
Isn’t browser caching works by ? being differentiator? Will now change in date ignored by browser cache as you use $?

pxtlib/emitter/cloud.ts
mmoskal: not sure; I guess we might want a token if we ever have a login

mmoskal:
Browser cache will be invalidated every day, but the Etag should still work. $ vs ? won't make a difference, but before were were not appending date. I don't think browser would cache these things for more than a day anyways, as they have not cache expiration headers.

pelikhan:
We cache translations in IndexedDb anyway.

abchatra:
Looks good to me. Good to merge?

mmoskal:
can only merge after cloud deployment

mmoskal:
Houston, we have a problem: 
```
Access to XMLHttpRequest at 'https://pxtstage.azureedge.net/api/md/arcade/tutorials$targetVersion=0.14.46&cdn=20191204' from origin 'https://arcade.staging.pxt.io' has been blocked by CORS policy: Request header field if-none-match is not allowed by Access-Control-Allow-Headers in preflight response.
```

PR #6289, opened 2019-12-05
Note picker reimplementation / fixes

jwunderl:
Cleans up the implementation of the note picker so that it's ideally less fragile / buggy. Lots of minor fixes along the way / dead code removed / etc.

Fixes https://github.com/microsoft/pxt-arcade/issues/1459

One thing in particular to sanity check; I removed the ``// Do not remove this comment.``, and just `lf`-d each name in the prebuilt map; there were a few that weren't used listed in that comment (`deep {0}`) and it feels like it's better to directly map the string instead of comment -> rlf-ing it. (presumably this wouldn't ever matter though, because it's unlikely that we would add more keys to this in the future if we haven't already?)

Other fixes:

* remove closure lib dependencies
* fix decompilation of note enum (`music.playNote(Note.A5)`)
* correctly resize depending on width, as it was only resizing on mobile vs non mobile before
* handle dragging on mobile / don't do the weird drag the whole page thing
* configurable `eps` (this was a public variable that was just never usable properly before, just making it so it's changeable like max / min)
* make the resizing for the text input occur when the text input changes, so that it doesn't grow / shrink randomly as you hit keys (based off the qualified name):

* fix display after typing (when finished editing, convert back to `{0}hz`)
* fix minor bug w/ highlighting on mobile (showed selected note as hover color instead of selected color)
* when opening on a small device, if one of the keys is selected, open to the page containing that key instead of just the first octave
* use standard BrowserUtils click events handling
* minor fix for BrowserUtils.addClass / etc; don't crash on safari when adding empty / non-trimmed class (`classList.add("")` throws, so `.addClass("hello "` throws currently)
* use methods instead of a giant .showEditor that handles all state w/ local-ish variables
* remove a lot of dead / would-be-buggy-if-it-had-worked code 
* really old off by one bug that would cause a crash if anyone ever tried to set maxNote to `76` on a field
* Validator now also prevents you from typing text into the num input, just like the normal number picker (it just ignores the inputs)

If anyone could make a test build it would be appreciated. Still need to test IE, edge, and actual devices for android / ios (emulator / simulator both work for those two though)

~There's also one more known bug; when the field editor is open, it shows up as red (invalid) until you type in something containing a period and reopen the editor (which is absolutely bizarre). Still taking a look at that, but it's not a regression from this change / also happens on arcade/beta.~ hack fix in https://github.com/microsoft/pxt/pull/6289/commits/50dc92e4388ec9fd9ee17a679e0e0e30fb513bb3, which just assigns the property an initial value as it is starting off undefined and only getting assigned with certain changes in input. I think the proper fix is in pxt-blockly, will take a look tomorrow~


pxtblocks/fields/field_note.ts
```diff
@@ -74,260 +74,156 @@ namespace pxtblockly {
...
-        67: {name: "D#", prefixedName: "High D#", freq: 1245},
-        68: {name: "E", prefixedName: "High E", freq: 1319},
-        69: {name: "F", prefixedName: "High F", freq: 1397},
-        70: {name: "F#", prefixedName: "High F#", freq: 1478},
-        71: {name: "G", prefixedName: "High G", freq: 1568},
-        72: {name: "G#", prefixedName: "High G#", freq: 1661},
-        73: {name: "A", prefixedName: "High A", freq: 1760},
-        74: {name: "A#", prefixedName: "High A#", freq: 1865},
-        75: {name: "B", prefixedName: "High B", freq: 1976}
+        28: { name: lf("C"), prefixedName: lf("Low C"), freq: 131 },
```
pelikhan: this is evaluated too early, potentially before the translation files are loaded. Needs to be lazy initialized.

pxtblocks/fields/field_note.ts
```diff
@@ -74,260 +74,156 @@ namespace pxtblockly {
...
+        75: { name: lf("B"), prefixedName: lf("High B"), freq: 1976 }
     }
 
-    let regex: RegExp = /^Note\.(.+)$/;
-
     export interface FieldNoteOptions extends Blockly.FieldCustomOptions {
         editorColour?: string;
         minNote?: string;
         maxNote?: string;
+        eps?: string;
```
pelikhan: what is eps?

pxtblocks/fields/field_note.ts
```diff
@@ -74,260 +74,156 @@ namespace pxtblockly {
...
+            if (!Number.isNaN(eps) && eps >= 0) {
+                this.eps = eps;
             }
 
-            let minNote = parseInt(params.minNote) || this.minNote_;
-            let maxNote = parseInt(params.maxNote) || this.maxNote_;
-            if (minNote >= 28 && maxNote <= 76 && maxNote > minNote) {
+            const minNote = parseInt(params.minNote) || this.minNote_;
+            const maxNote = parseInt(params.maxNote) || this.maxNote_;
+            if (minNote >= 28 && maxNote < 75 && maxNote > minNote) {
```
pelikhan: isn't B == 75?

theme/fieldeditors.less
```diff
@@ -9,59 +9,73 @@
...
+        background-color: white;
+        z-index: 1000;
+    }
 
-.blocklyDropDownContent .blocklyNote:hover {
-    background-color: yellow !important
-}
+    .blocklyNote.black {
+        background-color: black;
+        z-index: 1001;
```
pelikhan: is this supposed to go in the zindex.less?

pelikhan:
build: https://arcade.makecode.com/app/7fa44bb404819029c56da94450cd5442382ccfcb-302b312052

jwunderl:
That build doesn't have the .less changes

pxtblocks/fields/field_note.ts
jwunderl: Ah, makes sense. Moved it to be lazily created in constructor~

pxtblocks/fields/field_note.ts
jwunderl: eps is the 'give' for converting input values into named notes - e.g. Middle D is 294, but by default if you type in / decompile a value between 292.1 and 295.9 (eps of 2) it will switch it to middle d. Can remove from field note options if it doesn't seem useful, it just looked like it was intended to be configurable at some point as it was the only public (non-blockly required) field [here](https://github.com/microsoft/pxt/blob/master/pxtblocks/fields/field_note.ts#L163) from 3 years ago

abchatra:
@pelikhan can you please help review this change and signoff?

pelikhan:
I can test on Safari/Chrome on MacOSX but not on IE. I will produce an editor build.

pelikhan:
Build https://arcade.makecode.com/app/f2778236d0d1c0a18440efde5309088b924007a2-bfcf3d4c5a

pelikhan:
@abchatra looking good on osx/chrome. Someone needs to test IE.

abchatra:
Works well on Edge. We don't support IE anymore.

abchatra:
Good to merge?

jwunderl:
It will end up in microbit/beta next time a feature needs to come out there, which still supports IE, right? Not necessarily a blocker for merging for arcade, but will need to test eventually.

Otherwise should be good, also tested actual android device / osx

abchatra:
Yes for microbit beta, but not blocking merge of this. Lets go ahead and merge this.

PR #6290, opened 2019-12-05
Fix assumption that main.py always exists when switching to python

darzu:
Fixes: https://github.com/microsoft/pxt-minecraft/issues/1472

When switching to Python, we were assuming that main.py always existed. This was causing crashes.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -532,18 +532,23 @@ export class ProjectView
     }
 
     openPythonAsync(): Promise<void> {
-        // convert python to typescript, if same as current source, skip decompilation
-        const tssrc = pkg.mainEditorPkg().files["main.ts"].content;
-        return this.textEditor.convertPythonToTypeScriptAsync()
-            .then(pysrc => {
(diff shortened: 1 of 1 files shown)


webapp/src/app.tsx
```diff
@@ -532,18 +532,23 @@ export class ProjectView
...
+        const pySrcFile = pkg.mainEditorPkg().files["main.py"];
+        if (pySrcFile) { // do we have any python yet?
+            // convert python to typescript, if same as current source, skip decompilation
+            const tsSrc = pkg.mainEditorPkg().files["main.ts"].content;
+            return this.textEditor.convertPythonToTypeScriptAsync()
+                .then(pyAsTsSrc => {
+                    if (pyAsTsSrc == tsSrc) {
+                        // decompiled python is same current typescript, go back to python without ts->py conversion
+                        pxt.debug(`ts -> py shortcut`)
+                        this.setFile(pkg.mainEditorPkg().files["main.py"]);
```
riknoll: Nit: you can reuse the variable you declared here

PR #6298, opened 2019-12-06
Decompiler support for level editor

riknoll:
Adds decompiler support for both the `myTiles` namespace and for the eventual `createTilemap` function. For the namespace, I had to bake it into the decompiler just like I did with SpriteKinds; I might go back and make it more general later but there isn't really any reason to do that now.

For the `createTilemap` block, I didn't want to have the decompiler validating the tilemap so I instead put that duty on the field. If the call to `createTilemap` is invalid in some way, the tilemap field pretends to be a grey block. It ends up looking something like this:

Screen Shot 2019-12-05 at 5 33 10 PM

I think this is the cleanest way to do it without also baking in the tilemap encoding/decoding. We can reuse this trick in the future for other field editors, which should hopefully mean fewer code paths in the decompiler.


pxtblocks/fields/field_tilemap.ts
```diff
@@ -329,4 +372,28 @@ namespace pxtblockly {
...
+                continue;
+            }
+
+            return false;
+        }
+
+        return true;
+    }
+
+    function checkTileset(tileset: pxt.sprite.TileSet, galleryItems: pxt.sprite.GalleryItem[]) {
```
jwunderl: Is this here for future reference or something of the sort? Doesn't seem to be used.

pxtcompiler/emitter/decompiler.ts
```diff
@@ -2821,49 +2851,56 @@ ${output}</xml>`;
...
 
-            const kindInfo = env.blocks.kindsByName[n.name.text]
-            if (!kindInfo) {
-                return Util.lf("Only namespaces with 'kind' blocks can be decompiled");
-            }
+            const tilesetCheck = checkTilesetNamespace(n);
+            if (!tilesetCheck) return undefined;
 
-            const fail = Util.lf("Namespaces may only contain valid 'kind' exports");
+            return kindCheck || tilesetCheck;
```
jwunderl: This just always returns kindCheck at this point, right? if `!kindCheck` it returned undefined earlier, so the `|| tilesetCheck` should always be short circuited? I suppose it doesn't matter as either way it's just returning an error on why this one will be a gray block that isn't commonly seen

pxtblocks/fields/field_tilemap.ts
riknoll: Whoops, this got refactored. I'll delete it

pxtcompiler/emitter/decompiler.ts
riknoll: Yep, I'll remove the `||`. Users never actually see these errors anyways

Issue #6273, opened 2019-11-30
compiler bug (JS only, I think)


tballmsft:
Treatment of any as a field of an object is broken in the JS backend. repro is the following script for arcade (https://makecode.com/_PDsCx5CJseao): 
```

let sprite1 = sprites.create(img`
    . . . . . . . . . . . . . . . .
    . . . . . . . . . . . . . . . .
    . 3 3 . . . . . . . . . . . . .
    . . 3 . . . . . . . . . . . . .
    . . 3 3 . . . 3 3 3 3 . . . . .
    . . . . 3 3 . 3 3 3 . . . . . .
    . . . . . . 3 3 3 . . . . . . .
    . . . . . . 3 . . . . . . . . .
    . . . . . . 3 3 . . . . . . . .
    . . . . . . . 3 3 3 3 3 . . . .
    . . . . . . . . . 3 3 . . . . .
    . . . . . . . . . . . . . . . .
    . . . . . . . . . . . . . . . .
    . . . . . . . . . . . . . . . .
    . . . . . . . . . . . . . . . .
    . . . . . . . . . . . . . . . .
`)

let sprite2 = sprites.create(img`
    . . . . . . . . . . . . . . . .
    . . . . . . . . . . . . . . . .
    . 3 3 . . . . . . . . . . . . .
    . . 3 . . . . . . . . . . . . .
    . . 3 3 . . . 3 3 3 3 . . . . .
    . . . . 3 3 . 3 3 3 . . . . . .
    . . . . . . 3 3 3 . . . . . . .
    . . . . . . 3 . . . . . . . . .
    . . . . . . 3 3 . . . . . . . .
    . . . . . . . 3 3 3 3 3 . . . .
    . . . . . . . . . 3 3 . . . . .
    . . . . . . . . . . . . . . . .
    . . . . . . . . . . . . . . . .
    . . . . . . . . . . . . . . . .
    . . . . . . . . . . . . . . . .
    . . . . . . . . . . . . . . . .
`)

sprite2.x = 100

sprite1.data = sprite2

// this generates a runtime error
sprite1.data.destroy();

// comment out the above line, try this instead:

let s: Sprite = sprite1.data
s.destroy();
``

jwunderl:
Looks like same as https://github.com/microsoft/pxt-arcade/issues/515 - drop the parens following sprite1.data.destroy

mmoskal:
definitely, do not drop the parens, since it will stop working, once I finally fix this!

PR #6299, opened 2019-12-06
rework handling on methods/properties in interfaces

mmoskal:
fixes https://github.com/microsoft/pxt-arcade/issues/1442
fixes https://github.com/microsoft/pxt/issues/6273
fixes https://github.com/microsoft/pxt-arcade/issues/515

maybe fixes some others

tested on JS, SAMD, STM, rpi; VM is already broken it seems

This essentially makes clear distinction in code generation between `x.foo` and `x.foo()`, regardless of type of `x` and if `foo` is declared as method or property.

We now also support the following

```typescript
class A {
  foo() { ... }
}

let a:A
let fn = (a as any).foo
fn()
```

This works as expected (and not as in JS), that is the same as `let fn = () => a.foo()`. I think this departure from JS semantics is warranted because we don't have prototypes.

The `as any` cast is needed only because we otherwise complain. I left the error for now, as it probably helps beginners...


mmoskal:
This now also works as expected (with `log` method in `Logger` and without)

```typescript
interface CanMaybeLog {
    log?: (msg: string) => void;
}

class Logger {
    log(msg: string) {
        console.log(msg ? msg : "why" + msg);
    }
}

let logger: CanMaybeLog = new Logger();

if (logger.log) {
    logger.log("I got here!")
}
```

mmoskal:
build at https://arcade.makecode.com/app/e4dbf0a3c58992f480cb63f0a46ccefe0d0c1e8e-7e84aad5dd

mmoskal:
the behavior is that `x.foo` is equivalent to `x.foo.bind(x)` (at every access to `x.foo` without immediately following arguments)

jwunderl:
Hm, I am getting an error still `cannot read property iface of undefined` after changing https://github.com/jwunderl/pxt-promises/blob/master/main.ts#L235 to `return value && !!(value as any).then;` in that build when passing in a builtin type (num/string/array); is that expected? (``Promise.reject(42)`` fails (as does ``Promise.reject("??")``, but ``Promise.reject({})`` succeeds).

Can always handle literal values explicitly here and everything is working on pygamer and in browser:

```typescript
function isThenable<T>(value: PromiseResult<T>): value is PromiseLike<T> {
    if (typeof value == "number" || typeof value == "string" || Array.isArray(value))
        return false;
    return value && !!(value as any).then;
}
```

mmoskal:
We could probably be smarter about accessing properties on built in types... not this PR though

PR #6303, opened 2019-12-08
clear crowdin translation annotations from blocks

jwunderl:
fixes https://github.com/microsoft/pxt-arcade/issues/1458

Issue was that the block text never directly goes the `rlf`, which typically handles the removal of these annotations; the annotations that are added (like the namespace) are only used to lookup the translations, and are not part of the comment attributes. This just clears any annotations from the blocks when handling localizations. 

(also, I'm out on vacation next week so probably won't be able to update this PR before this needs to be fixed, so anyone can feel free to take this over / fix it in a better way if this is the wrong way to handle it)

re: https://github.com/microsoft/pxt-common-packages/pull/986, https://github.com/microsoft/pxt-common-packages/pull/1020

### DIFF
--- a/pxtlib/service.ts
+++ b/pxtlib/service.ts
@@ -472,13 +472,7 @@ namespace ts.pxtc {
         if (pxtc.Util.userLanguage() == "en") {
             // remove crowdin annotations from blocks
             return Promise.all(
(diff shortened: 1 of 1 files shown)


pelikhan:
Rewritten in #6316

PR #6308, opened 2019-12-10
Tileset field editor

riknoll:
This is the field editor for referencing tiles; it's based on the image dropdown and contains all tiles that are referenced in your current project.

In order to prevent students from breaking their own code, you can't delete any tile from your tilemap that is referenced by one of these field editors. Right now it just silently refuses to do it; I'll be adding an actual warning message in a different PR.

### DIFF
--- a/pxtblocks/fields/field_tilemap.ts
+++ b/pxtblocks/fields/field_tilemap.ts
@@ -132,6 +132,10 @@ namespace pxtblockly {
             return pxt.sprite.encodeTilemap(this.state, "typescript");
         }
 
+        getTileset() {
+            return this.state.tileset;
+        }
+
         doValueUpdate_(newValue: string) {
             if (newValue == null) {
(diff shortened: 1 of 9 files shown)


pxtblocks/fields/field_tilemap.ts
```diff
@@ -276,6 +281,14 @@ namespace pxtblockly {
             }
 
             tilemap.nextId = id + 1;
+
+            const prefix = `${pxt.sprite.TILE_NAMESPACE}.${pxt.sprite.TILE_PREFIX}`;
```
pelikhan: what happens when you change the locale of the editor?

riknoll: This is the qualified name that we generate for the TypeScript. It's the same in all languages. If we want the alt text to be translated, we can do that

PR #6311, opened 2019-12-10
clean up language options settings

pelikhan:
Consistently use userLanguage() and localeLive to pick up the correct locale in the editor.
Fix for https://github.com/microsoft/pxt-arcade/issues/1507

### DIFF
--- a/pxtrunner/runner.ts
+++ b/pxtrunner/runner.ts
@@ -384,20 +384,18 @@ namespace pxt.runner {
     }
 
     export let editorLanguageMode = LanguageMode.Blocks;
-    export let editorLocale = "en";
 
-    export function setEditorContextAsync(mode: LanguageMode, locale: string) {
+    export function setEditorContextAsync(mode: LanguageMode, localeInfo: string) {
         editorLanguageMode = mode;
-        if (locale != editorLocale) {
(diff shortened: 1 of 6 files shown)


pxtrunner/runner.ts
```diff
@@ -384,20 +384,18 @@ namespace pxt.runner {
     }
 
     export let editorLanguageMode = LanguageMode.Blocks;
-    export let editorLocale = "en";
 
-    export function setEditorContextAsync(mode: LanguageMode, locale: string) {
+    export function setEditorContextAsync(mode: LanguageMode, localeInfo: string) {
         editorLanguageMode = mode;
-        if (locale != editorLocale) {
+        if (localeInfo != pxt.Util.localeInfo()) {
```
riknoll: `pxt.Util.localeInfo()` looks different than `editorLocale` (`editorLocale` can't ever have the `live-` prefix). Should this be `pxt.Util.userLanguage()` instead?

pelikhan: We actually pass the localeInfo to the side docs from the editor and this function assumes that live- might be present. I'll take a peek at the calls.

pelikhan: Reviewed the calls and "localeInfo" is passed along.

PR #6314, opened 2019-12-10
Tileset editor decompile

riknoll:
Decompile support for the tileset field editor. Marking as draft because I need to confirm what the  block is going to look like.

It works by using the `//% blockIdentity=""` annotation. Because of that, I had to tweak the emitting for the tileset to add that comment annotation to all the user tiles. I'll also have another PR that does the same for pxt-arcade and all of the gallery tiles (hence the change to `pxt buildsprites`)


riknoll:
Here's the relevant PR in arcade: https://github.com/microsoft/pxt-arcade/pull/1534

tests/decompile-test/cases/testBlocks/spritekind.ts
```diff
@@ -49,4 +49,10 @@ namespace tiles {
     //% group="Animate" duplicateShadowOnDrag
     export function setTilemap(tilemap: TileMapData) {
     }
+
+    //% blockId=til_getter block="tile %tile"
```
shakao:
```suggestion
    //% blockId=tile_getter block="tile %tile"
```

PR #6315, opened 2019-12-11
Fix compilation with disabled blocks

pelikhan:
It turns out that we treat disabled blocks as events and sort them in the compiler. To create a call key for the block we compile each variable which blows up because the context has not been set.

> This change modifies the blocky compiler to sort events **after** disabled blocks have been filtered out. The only impact I can see here is that the decision to disable a bloc is now dependent on their positions rather than their naming -- **which might be a breaking change**.

Fixed. It uses the previous sorting.

@riknoll @abchatra careful review here

Fix for https://github.com/microsoft/pxt-arcade/issues/1485

### DIFF
--- a/pxtblocks/blocklycompiler.ts
+++ b/pxtblocks/blocklycompiler.ts
@@ -1602,6 +1602,7 @@ namespace pxt.blocks {
             // update disable blocks
             updateDisabledBlocks(e, allBlocks, topblocks);
             // drop disabled blocks
+            allBlocks = allBlocks.filter(b => !b.disabled);
             topblocks = topblocks.filter(b => !b.disabled);
             // reorder remaining events by names
             topblocks = topblocks.sort((a, b) => {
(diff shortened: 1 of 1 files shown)


riknoll:
Oh, I see how this could be breaking. Why is it based on positions, though? Can we just sort twice to restore the old behavior?

riknoll:
Hmmmm... I think maybe we should start persisting whether a block is enabled/disabled in the XML. This seems like something we could break again in the future.

pelikhan:
IMO. This is OK. The previous behavior was kind of arbitrary while now it is geometric. Not sure about storing the disabled state in XML since we recompute it on each build.

riknoll:
Right, but we could persist the saved state until the user makes an edit and then recompute it. At least in that case we would prevent shared scripts from breaking.

pelikhan:
Shared scripts don't compile the blocky code; they execute the javascript immediately.

pelikhan:
I'm going to fix this.

pelikhan:
@riknoll ready for review again

PR #6330, opened 2019-12-13
Tilemap resize bugs

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/1557
Fixes https://github.com/microsoft/pxt-arcade/issues/1558
Fixes https://github.com/microsoft/pxt-arcade/issues/1559
Fixes https://github.com/microsoft/pxt-arcade/issues/1563

This also makes it so that the resize controls don't show up within the tile editor


riknoll:
Resize is only disabled when editing tiles (not maps). So, if your tilemap is using 16x16 tiles you can't change the image editor you get when creating/editing tiles to be anything other than 16x16

PR #6332, opened 2019-12-13
Inline wall icon svg to align on mac and PC

shakao:


abchatra:
@shakao this has a conflict with master. Can you resolve?

PR #6334, opened 2019-12-13
Use \r\n new lines on Windows < 10

pelikhan:
Need someone with a Windows 8 machine to test this.
https://github.com/microsoft/pxt-arcade/issues/1241
Test: make sure the exported console text opens in notepad / windows 8.
```
forever(function () {
    console.logValue("x", 0)
})
```


FranklinWhale:
I think `\r\n` should be applied to Windows 10 too, because
1. Only recent Windows 10 releases support `\n` in Notepad (>=1809): https://devblogs.microsoft.com/commandline/extended-eol-in-notepad/
2. It is common to have different versions of Windows being used at the same time within the same school and the files are shared.

PR #6336, opened 2019-12-13
Hardware select dropdown next to download button

shakao:
https://arcade.makecode.com/app/765ff5b298254fce996908797fb97cbad3fe4876-324c7d13f0

### DIFF
--- a/webapp/src/editortoolbar.tsx
+++ b/webapp/src/editortoolbar.tsx
@@ -136,7 +136,11 @@ export class EditorToolbar extends data.Component<ISettingsProps, {}> {
         return saveInput;
     }
 
-    private getCompileButton(view: View, collapsed?: boolean): JSX.Element[] {
+    protected onHwItemClick = () => {
+        this.props.parent.showChooseHwDialog(true);
+    }
+
+    protected getCompileButton(view: View, collapsed?: boolean): JSX.Element[] {
         const targetTheme = pxt.appTarget.appTheme;
         const { compiling, isSaving } = this.props.parent.state;
         const compileLoading = !!compiling;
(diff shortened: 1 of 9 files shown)


webapp/src/editortoolbar.tsx
```diff
@@ -266,17 +323,15 @@ export class EditorToolbar extends data.Component<ISettingsProps, {}> {
                         </div>
                         <div className="five wide column">
                             <div className="ui grid right aligned">
-                                {compileBtn && <div className="row">
-                                    <div className="column">
-                                        <EditorToolbarButton role="menuitem" className={`primary large fluid download-button download-button-full ${downloadButtonClasses}`} icon={downloadIcon} text={downloadText} title={compileTooltip} onButtonClick={this.compile} view='tablet' />
+                                {compileBtn && <div className="ui row items" style={ { paddingBottom: 0, marginBottom: '1rem' } }>
```
riknoll: Why the inline style here?

riknoll: Also on line 332 (though it looks like that one was there before, it should also probably be a class)

webapp/src/editortoolbar.tsx
```diff
@@ -136,6 +136,63 @@ export class EditorToolbar extends data.Component<ISettingsProps, {}> {
...
+            default:
+                downloadButtonClasses += "huge fluid";
+                hwMenuClasses = "large";
+                hwIconClasses = "large";
+        }
+
+        let el = [];
+        el.push(<EditorToolbarButton key="downloadbutton" role="menuitem" icon={downloadIcon} className={`primary download-button ${downloadButtonClasses}`} text={view != View.Mobile ? downloadText : undefined} title={compileTooltip} onButtonClick={this.compile} view='computer' />)
+
+        let deviceName = pxt.hwName || lf("device");
```
riknoll: Capitalize?

shakao: it's used in "Download to {0}" below so "device" is part of the sentence

webapp/src/editortoolbar.tsx
shakao: this was an attempt to avoid having a custom css class/id specifically for the "not collapsed + tablet" view alone, i think in this case inline is a little nicer than a long selector but i'm fine with putting it in the css too

pelikhan:
image

PR #6351, opened 2019-12-17
fix animation editor crash when opening right after tilemap edit

darzu:
Fixes: https://github.com/microsoft/pxt-arcade/issues/1562


webapp/src/components/ImageEditor/ImageEditor.tsx
```diff
@@ -75,20 +75,22 @@ export class ImageEditor extends React.Component<ImageEditorProps, ImageEditorSt
 
         const { editTileValue, editingTile, alert } = this.state;
 
+        const isAnimationEditor = instanceStore.getState().store.present.kind === "Animation"
```
darzu: @riknoll , this is what I came up with, but I'm not sure this is what you had intended with duel mode store. Hope it's not abusing your abstractions

PR #6380, opened 2019-12-19
hide save button if repo is active

pelikhan:
It is confusing for students to have a "save" and "github" button next to each other: they confuse saving to file and github (observed in classroom).

- [x] hide the save button + project name when the project is tied to a repository
- [x] show project name in button in desktop
- [x] show branch name if non-master
- [x] rename local project from github repo name
- [x] don't show project rename dialog when connected to github (on editor exit)
- [x] save for offline button in extension zone (+ help)
- [x] fix for https://github.com/microsoft/pxt-arcade/issues/1676
- [x] move temp files into .github/makecode as this folder is handle specially by GitHub

image
image
image


jwunderl:
Aren't the pngs used for importing extensions offline with the `Import File...` card? Is there still a way to create that? (I suppose any uf2 should work as well, but maybe a bit confusing that you'd have to pick a random device first)

abchatra:
In my opinion users should always be able to save on the desktop. I don't think hiding the button is the right thing to do. We should discuss in the UI meeting before changing this.  

Also why is name of the project still there? Remove that as well.

abchatra:
May be we can find a different location for that GitHub button. 
I think bigger issue here is students don't know what commit means.

pelikhan:
The observed issue was that kids confused “save” as commit. Yes name should go away too.

pelikhan:
Current offline story for extensions (import png) is quite convoluted. Can be made clearer with dedicated button GitHub view.

abchatra:
Can you remove name and display the GitHub repo name there with the save icon? That is lot more clearer, than this change, which just removes the save button.

pelikhan:
Description updated.

pxtlib/initprj.ts
```diff
@@ -54,7 +54,7 @@ ${lf("To edit this repository in MakeCode.")}
 ${lf("This image shows the blocks code from the last commit in master.")}
 ${lf("This image may take a few minutes to refresh.")}
 
-![${lf("A rendered view of the blocks")}](https://github.com/@REPO@/raw/master/.makecode/blocks.png)
+![${lf("A rendered view of the blocks")}](https://github.com/@REPO@/raw/master/.github/makecode/blocks.png)
```
jwunderl: This will break the links for any existing project readmes, right? as the readme will remain the old location but the files will be stored in the new location

PR #6395, opened 2019-12-21
Fix patching of @PXT@ in hex files

mmoskal:
This just does the patching up to 4 times per hex line.

This bug only triggers with particular binary layouts, essentially due to "randomness" in object binary layout


mmoskal:
Hex files have limited line length, and I think we always use 16 bytes, so you cannot have more than 2. The loop might spin 4 times when the first four bytes match, but the next two don't (`patch == null`). I guess let's just leave it at this.

PR #6397, opened 2019-12-23
add `back` translate to ja

NagaoKyota:
The value corresponding to Japanese of `back` was not registered


pelikhan:
All translations are handled via crowdin (https://makecode.com/translate). THis one seems to be available (https://crowdin.com/translate/kindscript/all/en-ja#q=back). Could you attached a screenshot?

PR #6398, opened 2019-12-23
Allow for unary operators on any type, including enums

mmoskal:
Fixes https://github.com/microsoft/pxt-arcade/issues/774

### DIFF
--- a/pxtcompiler/emitter/emitter.ts
+++ b/pxtcompiler/emitter/emitter.ts
@@ -3082,39 +3082,33 @@ ${lbl}: .short 0xffff
             if (folded)
                 return emitLit(folded.val)
 
-            let tp = typeOf(node.operand)
-            if (node.operator == SK.ExclamationToken) {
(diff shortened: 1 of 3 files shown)


pxtcompiler/emitter/emitter.ts
```diff
@@ -3082,39 +3082,33 @@ ${lbl}: .short 0xffff
...
+                case SK.MinusMinusToken:
+                    return emitIncrement(node.operand, "numops::subs", false)
+                case SK.MinusToken: {
+                    let inner = emitExpr(node.operand)
+                    let v = valueToInt(inner)
+                    if (v != null)
+                        return emitLit(-v)
+                    return emitIntOp("numops::subs", emitLit(0), inner)
+                }
+                case SK.PlusToken:
```
jwunderl: This has a comment saying its a no op, but now should coerce to number (`+"1" + 1` should be 2, not 11) - looks like that needs handling here?

mmoskal: Thanks, fixed!

PR #6400, opened 2019-12-29
enable ```python code directive and clarify how highlighting & decompilation works

darzu:
See also: https://github.com/microsoft/pxt-minecraft/pull/1518
Part of: https://github.com/microsoft/pxt-minecraft/issues/1461
This is needed to unblock writing Python content.

TODO:
- [x] Only show the "Python" homepage row if the Python experiment is enabled
- [x] Update checkdocs to check direct python snippets

Enables the ```python tag for writting Python-specific tutorials.

We only show the "Python" carousel if the Python experiment is enabled.
Without experiment:
Screen Shot 2019-12-30 at 3 01 15 PM

With experiment:
Screen Shot 2019-12-30 at 3 01 35 PM

This lets us iterate on Python content work without impacting our users.

### DIFF
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -5,10 +5,14 @@
 
 declare namespace pxt {
     // targetconfig.json
+    interface GalleryProps {
+        url: string,
+        experimentName?: string
+    }
     interface TargetConfig {
         packages?: PackagesConfig;
         // common galleries
-        galleries?: pxt.Map<string>;
+        galleries?: pxt.Map<string | GalleryProps>;
         // localized galleries
         localizedGalleries?: pxt.Map<pxt.Map<string>>;
         windowsStoreLink?: string;
--- a/pxtrunner/renderer.ts
+++ b/pxtrunner/renderer.ts
@@ -43,11 +43,14 @@ namespace pxt.runner {
 
     function highlight($js: JQuery) {
         if (typeof hljs !== "undefined") {
-            if ($js.hasClass("highlight"))
+            if ($js.hasClass("highlight")) {
                 hljs.highlightBlock($js[0]);
-            else $js.find('code.highlight').each(function (i, block) {
-                hljs.highlightBlock(block);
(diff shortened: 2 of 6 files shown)


pxtrunner/renderer.ts
```diff
@@ -831,6 +835,33 @@ namespace pxt.runner {
         })
     }
 
+    function renderDirectPython(options?: ClientRenderOptions) {
+        // Highlight python snippets written with the ```python
+        // language tag (as opposed to the ```spy tag, see renderStaticPythonAsync for that)
```
shakao: does checkdocs need to be updated to handle these snippets?

shakao:
is the main reason for direct python flakiness/verbosity in typescript -> spy conversion?

pxtrunner/renderer.ts
darzu: Good question

darzu:
> is the main reason for direct python flakiness/verbosity in typescript -> spy conversion?

I wouldn't call it flakiness, but since the conversion is automated and conservative it ties your hands a little bit on what you can express. Direct python gives you control to write exactly the python you want and no more which i think is especially important for educational content.

pxtrunner/renderer.ts
darzu: Updating checkdocs is non-obvious, I'm going to hold this PR until I can talk to someone in person that knows more

darzu: Updated check docs. It can now correctly detect if there are errors in the python snippets. Good catch

localtypings/pxtarget.d.ts
```diff
@@ -5,10 +5,14 @@
 
 declare namespace pxt {
     // targetconfig.json
+    interface GalleryProps {
+        url: string,
+        experimentName?: string
```
pelikhan: nice, we could also have locale-spefic galleries too

pelikhan:
Could you also update the tutorial markdown docs?

PR #6411, opened 2020-01-03
Don't rebuild the API info for each snippet in docs

riknoll:
This speeds up the load time of docs pages with a lot of snippets. I used this page for testing: https://arcade.makecode.com/concepts/star-field

On my local machine it took ~6 seconds to fire the onLoad event before my changes and ~3.7 seconds after my changes.

We could improve this even more by using the same API info that the editor stores in the indexed db, but I didn't think it was worth duplicating all of that code in the embed frame. I might try to refactor it out later.

### DIFF
--- a/pxtrunner/runner.ts
+++ b/pxtrunner/runner.ts
@@ -953,8 +953,6 @@ ${linkString}
     function getApiInfo(program: ts.Program, opts: pxtc.CompileOptions) {
         if (!apiCache) apiCache = {};
 
-        // FIXME: If some page theoretically included two different verions of the same package, this might fail.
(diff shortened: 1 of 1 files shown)


riknoll:
Ah, I didn't know it applied to all snippets. I'll remove the comment

jwunderl:
Actually, with that in mind, we probably don't even need the map unless you want to future proof it in case someone adds single-snippet dependencies, right? The snippets should have the same deps always, so it just needs the one ``pxtc.ApisInfo``

riknoll:
Yeah, I'll leave it as is for future-proofing

PR #6418, opened 2020-01-06
Ensure names generated by auto-complete are unique

darzu:
Fixes: https://github.com/microsoft/pxt-minecraft/issues/1479

- In SyntaxInfo, keep a seperate list of global symbols for all types of requests so that we can ensure uniqueness for member completion or identifier completion
- Update py2ts to return these
- Update getCompletions and getSnippet to account for existing symbols
- remove dead code, improve comments, rename a few variables

### DIFF
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -804,7 +804,7 @@ namespace ts.pxtc.service {
         }
     }
 
-    type Py2TsRes = {
+    interface Py2TsRes {
         // TODO: use the real types from converter.ts
         diagnostics: pxtc.KsDiagnostic[],
         success: boolean,
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -669,36 +669,6 @@ namespace ts.pxtc {
             return symbol.name
         return typechecker.getFullyQualifiedName(symbol);
     }
-
-    /*
-    export function fillCompletionEntries(program: Program, symbols: Symbol[], r: CompletionInfo, apiInfo: ApisInfo, opts: CompileOptions) {
-        const typechecker = program.getTypeChecker()
-
-        for (let s of symbols) {
-            let qName = getFullName(typechecker, s)
-            const gsi = Util.lookup(apiInfo.byQName, qName);
-
-            // filter out symbols starting with __
-            if (gsi && /^__/.test(gsi.name))
-                continue;
-
-            if (!r.isMemberCompletion && gsi)
-                continue; // global symbol
-
-            if (Util.lookup(r.entries, qName))
-                continue;
-
-            const decl = s.valueDeclaration || (s.declarations || [])[0]
-            if (!decl) continue;
-
-            const si = createSymbolInfo(typechecker, qName, decl)
-            if (!si) continue;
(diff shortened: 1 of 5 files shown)


pxtpy/converter.ts
```diff
@@ -2433,6 +2434,25 @@ namespace pxt.py {
             if (!syntaxInfo.symbols)
                 syntaxInfo.symbols = []
 
+            // always return global symbols because we might need to check for
+            // name collisions downstream
+            {
```
riknoll: why is this in a block?

pxtcompiler/emitter/service.ts
```diff
@@ -833,10 +804,29 @@ namespace ts.pxtc.service {
...
+        success: boolean,
+        outfiles: { [key: string]: string },
+        syntaxInfo?: pxtc.SyntaxInfo
+    }
+    function updatePySyntaxInfo(opts: CompileOptions) {
+        // TODO: We call py2ts here directly to get the python syntax info.
+        // We likely already have the syntax info from a previous compile,
+        // we should do better about cache the latest results and serving those
+        // here. Also, we shouldn't mutate "opts" (the input).
+        let res = (pxt as any).py.py2ts(opts) as Py2TsRes
```
riknoll: Why the `any` cast here?

pxtcompiler/emitter/service.ts
```diff
@@ -833,10 +804,29 @@ namespace ts.pxtc.service {
         }
     }
 
+    type Py2TsRes = {
```
darzu: swaping:
image

pxtcompiler/emitter/service.ts
```diff
@@ -1254,22 +1248,31 @@ namespace ts.pxtc.service {
 . . . . .
 \``;
 
-    export function getSnippet(apis: ApisInfo, runtimeOps: pxt.RuntimeOptions, fn: SymbolInfo, n: ts.FunctionLikeDeclaration, python?: boolean): string {
+    export function getSnippet(apis: ApisInfo, takenNames: pxt.Map<SymbolInfo>, runtimeOps: pxt.RuntimeOptions, fn: SymbolInfo, decl: ts.FunctionLikeDeclaration, python?: boolean): string {
```
riknoll: We should consider refactoring this at some point; too many arguments

darzu: I definitely agree. I spent a good hour this morning trying to unify lastApisInfo and this syntax info stuff and I had to bail, it was too tangle.

pxtcompiler/emitter/service.ts
darzu: I forget the justification, but Michal tried for a while to get references to the python code to work right with TSC and I believe I gave up. I think I tried at one point too. If you have ideas, I'm all ears.

pxtpy/converter.ts
darzu: Used to have an "if",  i'll remove the extra block

pxtcompiler/emitter/service.ts
darzu: I left a number of new TODO comments detailing some of the things I think are broken / inflexible with our current approach

PR #6421, opened 2020-01-06
support advanced python operators & throw error instead of returning comment

darzu:
Fixes: https://github.com/microsoft/pxt-minecraft/issues/1031


jwunderl:
this fixes https://github.com/microsoft/pxt-minecraft/issues/1464 too, right?

darzu:
Only partly

PR #6422, opened 2020-01-06
Larger sidedocs in Monaco, shrink font & no wrap for python

shakao:
image

Some fixes for https://github.com/microsoft/pxt-minecraft/issues/1521

Sidedocs take up 50% of the screen in text (py/js), unchanged in blocks

### DIFF
--- a/theme/docs.less
+++ b/theme/docs.less
@@ -472,3 +472,14 @@
     }
 }
 .replaceDocsAvatar();
+
+#docs.py {
+    pre {
+        overflow-x: scroll;
+        font-size: 0.875em;
+    }
+
+    pre > code {
+        white-space: pre;
+    }
+}
--- a/theme/docs.less
+++ b/theme/docs.less
@@ -371,8 +371,8 @@
         margin: 0;
     }
 
-    /* python sidedocs tutorials */
-    &.py {
+    /* sidedocs tutorials (text-based) */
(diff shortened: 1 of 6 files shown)


pelikhan:
Looks like the function names are being incremented on each snippet. @darzu

shakao:
oh no this was just a fake/local tutorial where i pasted in the converted python code so i could test

darzu:
Huge improvement, love it!

theme/docs.less
```diff
@@ -472,3 +472,14 @@
     }
 }
 .replaceDocsAvatar();
+
+#docs.py {
```
pelikhan: does this apply to javascript too?

webapp/public/docs.html
```diff
@@ -119,6 +119,9 @@
             var backLabel = document.getElementById('back-label');
             if (backLabel)
                 backLabel.textContent = pxt.Util.lf("Go back")
+            var editorlang = pxt.Util.getEditorLanguagePref();
+            if (editorlang == "py") document.body.classList.add("py");
```
pelikhan: IMO, allways add the editorlang to the class list.

theme/docs.less
shakao: yeah you're right, i think it should. updated to use a single editorlang-text class to style both js/python

Issue #6414, opened 2020-01-04
Make keymap configurable/localizable


pelikhan:
The keymap info is currently hard-coded in pxt.

- [ ] configured via pxtarget.json
- [ ] localizable

https://github.com/microsoft/pxt/commit/c0c841362df7d26d2efab50944b72061e913def5

PR #6428, opened 2020-01-08
Keymap localization + small css tweak

shakao:
Always keep white background, add lf wrapper

Addresses localization of this https://github.com/microsoft/pxt/issues/6414


jwunderl:
I think we might have to lazy load the ``_data`` itself to make sure translations are obtained when it's created? Not 100% sure, @pelikhan will know

PR #6434, opened 2020-01-09
list supported & unsupported python features

darzu:
Lists out supported, partially supported, unstable support, and not supported python language features.

@mmoskal , can you sanity check what I wrote below? I was a little conservative (I think) and listed somethings that might work well as unstable or partially supported. In general I'd like to be conservative to reduce what we have to test thoroughly for this release.

### DIFF
--- a/docs/language.md
+++ b/docs/language.md
@@ -24,6 +24,8 @@ and might not be available in your editor.
 
 ## ~
 
+# Static TypeScript
+
 ## Supported language features
 
 * variable declarations with `let`, `const`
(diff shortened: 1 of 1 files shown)


docs/language.md
```diff
@@ -280,3 +282,61 @@ and dynamic maps.
...
+* generators
+* attributes
+* import, import from
+* sets
+* list comprehensions
+* set comprehensions
+* dictionary comprehensions
+* await
+* yield, yield from
+* "..."
```
mmoskal: what is that?

docs/language.md
```diff
@@ -280,3 +282,61 @@ and dynamic maps.
...
 * how to validate types of C++ classes (Pin mostly)?
+
+
+# Python
+
+## Supported language features
+The following language features should be fully supported and work according to the Python 3 language specification.
+
+* lists
+* dictionaries
```
mmoskal: I guess we still don't do dictionary literals, but should be easy to add (I guess that's assigned to me)

docs/language.md
```diff
@@ -280,3 +282,61 @@ and dynamic maps.
...
+* while loop
+* for-in loop with range(), array or string literal
+* break, continue
+* conditional statements (if, elif, else)
+* pass statement
+* variables*
+* if expression / ternary operator
+* comparison (in, notin)
+* byte literal
+* *args
```
mmoskal: what's that? I don't think we support varargs do we?

docs/language.md
darzu: Fancy slice notation I believe. I removed this bullet point since it is redundant.

docs/language.md
darzu: Think you'll get to this? If not, I'll remove it from the list

docs/language.md
darzu: varargs. Just tried it and it doesn't work. Moving it to unsupported list, thanks

PR #6439, opened 2020-01-11
Always emit type annotations for parameter declaration

riknoll:
In our TS, type annotations are required. This fixes two bugs where they were not being emitted when going from python to TS

1. Type annotations for builtin types (e.g. number) were ignored
1. If we couldn't infer a type for a parameter, we didn't emit any annotation. Now we emit `any`

Because of the second change, I also did a fix to make sure that a parameter declaration with the type `any` is omitted when going from TS to python. Previously it compiled to something like this:

```python
def on_chat5(num: (TODO: Unknown TypeNode kind: 1234)):
    pass
```

### DIFF
--- a/pxtpy/converter.ts
+++ b/pxtpy/converter.ts
@@ -179,7 +179,7 @@ namespace pxt.py {
         if (tp == "T" || tp == "U") // TODO hack!
             return mkType({ primType: "'" + tp })
 
-        // defined by a symbol, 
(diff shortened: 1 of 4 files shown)


jwunderl:
oops, dropped comment when approving by accident; should be able to do a runtime trace test for this, right?

```py
function shouldCompile(a: number):
    pass
```

PR #6442, opened 2020-01-11
Allow list of action cards in homescreen detail view

shakao:
Add 'otherActions' field to codeCard JSON so we can specify additional actions in the homescreen details view. CSS/styling for cards in the details view. 

TODO: add new action for opening arcade forum games in editor, move to markdown.

fix_detailactions

https://minecraft.makecode.com/app/2373e1bc8f68ae7e01ef21027de5ed090430f0f4-82f484cd90


webapp/src/projects.tsx
```diff
@@ -713,11 +724,78 @@ export class ProjectsDetail extends data.Component<ProjectsDetailProps, Projects
...
+                break;
+            case "forumUrl":
+                icon = "comments"
+                break;
+            case "template":
+            default:
+                if (youTubeId) icon = "video camera";
+                break;
+        }
+        return this.isLink() && type != "example" // TODO (shakao)  migrate forumurl to otherAction json in md
```
jwunderl: since the icon and text are both links you have to tab twice per option; probably `tabindex-1` for the icon so it just goes to the button?

shakao:
yeah, removed the background so we could shrink the vertical height and also improve readability of the text/action cards

pelikhan:
I quite liked the gif backgrounds. It’s dull now.

PR #6447, opened 2020-01-14
Python gray blocks

riknoll:
If Python is set as your editor language preference, don't render TypeScript code in gray blocks. Instead, we render `<python code>` for statements and `<python expression>` for expressions. This is more of a placeholder until we have time to do something more comparable to TS.

Here is a GIF that loops perfectly:

### DIFF
--- a/pxtblocks/fields/field_tsexpression.ts
+++ b/pxtblocks/fields/field_tsexpression.ts
@@ -44,7 +44,7 @@ namespace pxtblockly {
         }
 
         protected setPythonText() {
-            this.setText(pxt.Util.lf("<python expression>"))
+            this.setText(pxt.Util.lf("<python code>"))
         }
     }
 }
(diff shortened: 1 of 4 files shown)


pelikhan:
I would just call it “<python code>” in all grey blocks. Expression vs Statement are CS concepts that most students probably grasp yet.

pxtblocks/fields/field_tsexpression.ts
```diff
@@ -3,24 +3,48 @@
...
+            if (enabled) {
+                this.setPythonText();
+            }
+            else {
+                this.setText(this.getValue());
+            }
+        }
+
+        protected setPythonText() {
+            this.setText(pxt.Util.lf("<python expression>"))
```
pelikhan: IMO use ``<python code>`` everywhere.

PR #6452, opened 2020-01-14
Fix docs in menu

jwunderl:
fix https://github.com/microsoft/pxt-minecraft/issues/1430; fix spacing issue when multiple doc items are there, handle outside links correctly, and show all of the snippets

Just noticed you guys said to just drop it instead so if you actively don't want it I can remove the ``renderDocItems(this.props.parent, "ui mobile only inherit", docItems)}`` from settings menu, but it all works now so ¯\\_(ツ)_/¯

### DIFF
--- a/theme/common.less
+++ b/theme/common.less
@@ -814,6 +814,9 @@ p.ui.font.small {
     .ui.mobile.only {
         display:unset !important;
     }
+    .ui.mobile.only.inherit {
+        display:inherit !important;
+    }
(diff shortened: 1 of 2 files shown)


pelikhan:
what happens on a phone form factor?

abchatra:
Is it missing language change option?

jwunderl:
I loaded with the flags for the ingame view

abchatra:
Oh ok makes sense

PR #6456, opened 2020-01-15
don't put member of emit as constant

pelikhan:
Introduces a generic "alias" annotation; emit FORWARD instead of SixDirection.Forward

Before:
image

After:
image

### DIFF
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -1552,7 +1552,8 @@ namespace ts.pxtc.service {
                                 let pxtSym = apis.byQName[fullName]
                                 if (pxtSym) {
                                     if (pxtSym.attributes.alias)
-                                        return pxtSym.attributes.alias; // prefer alias
+                                        // use pyAlias if python; or default to alias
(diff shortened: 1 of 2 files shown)


pxtcompiler/emitter/service.ts
```diff
@@ -927,7 +927,9 @@ namespace ts.pxtc.service {
                     /^__/.test(si.name) || // ignore members starting with __
                     /^__/.test(si.namespace) || // ignore namespaces starting with _-
                     si.attributes.hidden ||
-                    si.attributes.deprecated
+                    si.attributes.deprecated ||
+                    // don't emit members of enum marked as "emitAsConstant"
+                    (si.kind == SymbolKind.EnumMember && lastApiInfo.apis.byQName[si.namespace]?.attributes.emitAsConstant)
```
riknoll: Will this work if the enum is declared inside of a namespace? Wilt the namespace field have the full qualified name?

pelikhan: i'm going to rely on an annotation rather than weird logic

PR #6463, opened 2020-01-16
handle defaults with param.shadow=text

jwunderl:
currently, adding a default value for https://github.com/microsoft/pxt-common-packages/blob/master/libs/base/console.ts#L57 (or in my case making ``player.say`` allow `any` in minecraft) will just get dropped and show up as the empty string, because it doesn't convert the field name from 'text' to 'TEXT' and the value is silently ignored.

This is annoying when using it to give a default for `any` or complex types (e.g. union types involving strings), for example adding an ``e.g.`` default for https://github.com/microsoft/pxt-common-packages/blob/master/libs/base/console.ts#L57 (or in my case making ``player.say`` allow `any` in minecraft) doesn't work currently.

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -141,6 +141,7 @@ namespace pxt.blocks {
         }
 
         const isVariable = shadowId == "variables_get";
+        const isText = shadowId == "text";
 
         const value = document.createElement("value");
         value.setAttribute("name", p.definitionName);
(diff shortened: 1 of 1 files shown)


pelikhan:
possible to get a test?

jwunderl:
I don't know if there is a place to add a test for this one currently; it's not a comment attribute test as the default value was getting parsed accurately but not getting applied, and we don't have tests for the toolbox blocks themselves

PR #6468, opened 2020-01-17
Use alias when avaiable in py decompiler

pelikhan:
Use "alias" info, e.g. short name if possible when decompiling -- otherwise we end up with ugly code if we come from TS/blocks.

Note: the py decompilation looks trashed.


abchatra:
@pelikhan fix the travis issue?

PR #6477, opened 2020-01-19
Hide TD_ID function from toolbox/completion

pelikhan:
Those function are blocks artifacts.

### DIFF
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -933,7 +933,7 @@ namespace ts.pxtc.service {
                     /^__/.test(si.namespace) || // ignore namespaces starting with _-
                     si.attributes.hidden ||
                     si.attributes.deprecated ||
-                    // don't members with an alias
+                    // don't emit members with an alias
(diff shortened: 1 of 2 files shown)


pxtcompiler/emitter/service.ts
```diff
@@ -933,8 +933,10 @@ namespace ts.pxtc.service {
                     /^__/.test(si.namespace) || // ignore namespaces starting with _-
                     si.attributes.hidden ||
                     si.attributes.deprecated ||
-                    // don't members with an alias
-                    si.attributes.alias
+                    // don't emit members with an alias
+                    si.attributes.alias ||
```
darzu: @pelikhan , why do we want to omit members with an alias from completions? This is filtering out entries like "AnimalMob.Chicken" because it has an alias of "CHICKEN". However "CHICKEN" does not appear as its own entry in the apiInfo so CHICKEN is now not appearing at all in completions.

pelikhan: Humm CHICKEN did appear in completion. Do you have a context?

PR #6479, opened 2020-01-21
auto-populate other actions

pelikhan:
Short hand to auto-populate otherActions: empty array means populate otherActions as best as can.

Allows to avoid copy-pasting tons of markdown and URL path errors.

This currently applies to tutorials only but next step will be support for auto multi-editor examples. See https://github.com/microsoft/pxt-maker/pull/276


pelikhan:
This currently applies to tutorials only but next step will be support for auto multi-editor examples.

PR #6489, opened 2020-01-24
add block ids before exporting to github

pelikhan:
To avoid the first empty commit, add block ids as we are exporting to github.


mmoskal:
Not sure hue it works but maybe it’s worth it to keep the ids always except when encoding in uf2?

pelikhan:
I need to move away to use ids and do a structural diff instead

pelikhan:
Maybe do always do a typescript diff them map diff results to the blocky blocks

PR #6493, opened 2020-01-26
Diffing of blocks from decompiled TS

pelikhan:
Uses the TS diff to render a blocks diff. This allows to compute diffs between documentation snippets and tutorial steps; or between blocks that have been rountripped to JS.

* run http://localhost:3232/release-tests/md/diff for testing
* guarded by pxtarget flag to experiment
* automatic step diffing in blocks tutorials

* rendering of diffs in docs

- [x] generate source maps on decompilation
- [x] added 'diffblocks' section
- [x] render js/py diff as well
- [x] handle trailing ; in sources
- [x] auto-diffing in tutorials
- [x] minimize size of serialized tutorial data
- [x] added test + updated test suite (marked as noDiffs)


localtypings/pxtarget.d.ts
```diff
@@ -855,7 +855,6 @@ declare namespace pxt.tutorial {
         // no coding
         unplugged?: boolean;
         tutorialCompleted?: boolean;
-        hasHint?: boolean;
```
pelikhan: inferred from ``!!hintContentMd``

pxtlib/tutorial.ts
```diff
@@ -168,15 +195,19 @@ ${code}
...
-                unplugged: /@unplugged/.test(flags),
-                tutorialCompleted: /@tutorialCompleted/.test(flags),
-                resetDiff: /@resetDiff/.test(flags),
                 contentMd: step,
-                headerContentMd: header,
-                hintContentMd: hint,
-                hasHint: hint && hint.length > 0
+                headerContentMd: header
             }
+            if(/@(fullscreen|unplugged)/.test(flags))
```
pelikhan: avoid useless fields in serialized json

abchatra:
Can we just highlight the block diffed by color and keep the rest the same color? Gray color is hard to read in my opinion.

pelikhan:
The whole point is to focus the attention on the changes on the changes blocks.

abchatra:
Yes highlight it without making everything else look disabled. Like in the 'python' yellow highlighting.

pelikhan:
The highlight does not focus the attention enough , imo.

In python the text is bw by default, which is quite different from the blocs filled with colors.

pelikhan:
This is a UI design question.

abchatra:
> This is a UI design question.

Yes, we can discuss tomorrow in UI meeting.

darzu:
Love it. I like it how it's designed right now. Everything else grey and new thing is colored.

pxtblocks/blocklydiff.ts
```diff
@@ -401,4 +422,75 @@ namespace pxt.blocks {
...
+                    }
+                    // find block ids mapped to the ranges
+                    const oldid = pxt.blocks.findBlockId(oldResp.sourceMap, {
+                        start: oldLineStart,
+                        length: lineLength
+                    });
+                    const newid = pxt.blocks.findBlockId(newResp.sourceMap, {
+                        start: newLineStart,
+                        length: lineLength
+                    });
```
darzu: After my PR, you can provide a position within a single line, in case that helps

pelikhan: right now i don't handle line diff but eventually we could be more precise and flag changes within nested blocs

pelikhan: or after my PR, you can update my decompiler. why do we need this?

abchatra:
Can we add flag to enable\disable this feature in the target on need basis? We have a deadline for Minecraft and hour of code is dependent on the hints. Also hour of code is just one step.

pxtblocks/blocklydiff.ts
darzu: I think the use case is for tutorial steps that just modify a line like "change this number" or "add (pick random) into the second argument". Not sure how many non-trivial examples of this we have.

pelikhan:
You can also disable the feature for a tutorial using
```
### @noDiffs
```

PR #6494, opened 2020-01-27
display beta editor card

pelikhan:
- [x] display "Beta Editor" card in experiments, based on current URL (query and hash carried over)
- [x] deprecated "betaUrl" entry in pxtarget.json (ignored)

### DIFF
--- a/webapp/src/scriptsearch.tsx
+++ b/webapp/src/scriptsearch.tsx
@@ -359,7 +359,7 @@ export class ScriptSearch extends data.Component<ISettingsProps, ScriptSearchSta
             && !searchFor;
         // inject beta at end of / or /#
         // also excludes http://localhost:port/index.html
-        const betaUrl = window.location.href.replace(/\/(#|$)/, "/beta$1")
+        const betaUrl = window.location.href.replace(/\/(#|$|\?)/, "/beta$1")
         const showOpenBeta = mode == ScriptSearchMode.Experiments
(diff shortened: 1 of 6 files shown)


abchatra:
I am not sure this will work on Minecraft. URL vastly differs between code connection, hour of code, in game and nether. Is it possible to just add /beta to the existing URL and #   and reload?

abchatra:
Reload the existing URL with /beta inserted in between the URL.

pelikhan:
Turns out that there were multiple places we were displaying this beta url string. I've deprecated that flag.

pelikhan:
@abchatra

webapp/src/scriptsearch.tsx
```diff
@@ -358,6 +357,11 @@ export class ScriptSearch extends data.Component<ISettingsProps, ScriptSearchSta
             && pxt.appTarget.appTheme.importExtensionFiles
             && !disableFileAccessinMaciOs
             && !searchFor;
+        // inject beta at end of / or /#
+        // also excludes http://localhost:port/index.html
+        const betaUrl = window.location.href.replace(/\/(#|$)/, "/beta$1")
```
abchatra: Does this work with something like below? 
https://minecraft.makecode.com/?ingame=1&ipc=1&noRunOnX=1#tutorial:/hour-of-code/open-door

abchatra: You should add for ? too?

PR #6499, opened 2020-01-27
Add option to choose language restriction on creating project

jwunderl:
add an option to choose to restrict project to be python / javascript only when in the `choose a name for new project` page

keeps you in the selected language (if one is chosen):

(.ts files hidden here because selected language is python)

Screen Shot 2020-01-27 at 2 38 45 PM

in share screen, can toggle between sim and the selected language only:

minor change to share embed view per design meeting -> elements contained in the expandable dropdown should be aligned with the the button's words so there's a clearer hierarchy


webapp/src/app.tsx
```diff
@@ -2053,7 +2064,9 @@ export class ProjectView
                                 })
                         });
                 } else {
-                    opts.tsOnly = !loadBlocks;
```
shakao: can we remove/deprecate "tsOnly" entirely? seems like the property is covered by languageRestriction

jwunderl: It was only ever used / referenced here and in the ProjectCreationOptions interface, and it's removed from both of those places in this pr

shakao: oh oops i didn't see you deleted from the interface as well--and importing old projects with tsOnly is fine?

jwunderl: Yup, that was never persisted in pxt.json so it will just load like any project without blocks; it only ever did this before:

https://github.com/microsoft/pxt/pull/6499/files#diff-325ec83a257a791d33bd4484fd67fb24L1979

pxteditor/editor.ts
```diff
@@ -315,7 +315,7 @@ namespace pxt.editor {
 
         showModalDialogAsync(options: ModalDialogOptions): Promise<void>;
 
-        askForProjectNameAsync(): Promise<string>;
+        askForProjectSettingsAsync(): Promise<ProjectCreationOptions>;
```
pelikhan: rename to ``askForProjectCreationOptionsAsync``()

webapp/src/app.tsx
```diff
@@ -1976,12 +1976,23 @@ export class ProjectView
                 cfg.dependencies[ext.name] = `workspace:${ext.id}`;
             }
         }
-        if (options.tsOnly) {
```
pelikhan: IMO, add code that converts tsOnly to the ``typescript-only`` on the fly, to handle legacy scripts.

pelikhan:
- [ ] add flag in appTheme to enable this feature. This is a UI change and it's not clear that we want this in all editors.

pelikhan:
The ``Project Template`` is misleading. It seems that you are mixing 2 concepts: "template" vs "language restriction". It is not impossible that in the future you will have multiple templates for the same language (``platformer js``, ``schmup js``).

Suggestion to rename to ``I want to code in DROPDOWN``

jwunderl:
Nice, I like `I want to code in {All Languages | ...}`, was going to wait to ask at design meeting for better wording.

in adafruit:

Screen Shot 2020-01-27 at 6 12 03 PM

will take a look at the merge resolution /etc now

(as a side note, I was thinking this might be a good place to add a 'make this a github repository' checkbox eventually)

jwunderl:
I don't think we want anything in 'reconstructConfig' for this - this is only ever a restriction at the moment, and leaving it undefined is fine / equivalent to leaving it as `LanguageRestriction.Standard`

pelikhan:
Language means locale in our UI so I think it’s best to spell in out: “blocks or JavaScript”, “blocks, JavaScript or python”

Yeah you can drop the reconstruct.

webapp/src/projects.tsx
```diff
@@ -1178,7 +1178,7 @@ export class NewProjectDialog extends data.Component<ISettingsProps, NewProjectD
         const langOpts: sui.SelectItem[] = [
             {
                 value: pxt.editor.LanguageRestriction.Standard,
-                display: lf("All Languages")
+                display: python ? lf("Blocks, {0}, and {1}", "JavaScript", "Python") : lf("Blocks and {0}", "JavaScript")
```
pelikhan: what about "blocks and python"?

jwunderl: As in another language restriction option that hides ts? Easy to do if we want

abchatra:
"Give your project name" "I want to code in" use different pronouns. We need to think about the right text here.

webapp/src/projects.tsx
abchatra: Lets keep it to 3 options for now. Too many options will be confusing.

pelikhan:
"Give your project a name" and "Choose how to code"

localtypings/pxtarget.d.ts
```diff
@@ -365,7 +365,8 @@ declare namespace pxt {
         checkForHwVariantWebUSB?: boolean; // check for hardware variant using webusb before compiling
         shareFinishedTutorials?: boolean; // always pop a share dialog once the tutorial is finished
         leanShare?: boolean; // use leanscript.html instead of script.html for sharing pages
-        nameProjectFirst?: boolean;
+        nameProjectFirst?: boolean; // prompt user to name project when creating new one
+        newProjectOptions?: boolean; // include 'options' menu when creating a new project
```
pelikhan: more specific flag maybe: chooseLanguageRestrictionOnNewProject . it's verbose but self-explaijning.

pelikhan:
ready for review?

jwunderl:
Yes, up to date with the decisions in the design meeting

PR #6500, opened 2020-01-28
add flag to control api string upload branch

pelikhan:
Allows to specify a regex to match the branch name to upload api strings.


pelikhan:
Someone has to parse the gitlog to do that.

PR #6503, opened 2020-01-28
React component for Monaco flyout, updated flyout UI

shakao:
- Move Monaco flyout into separate react component
- Add parameters to toolbox block
- Display parameter info on click
- Drag-and-drop behavior enabled for iOS/mobile

toolbox_expand_2

Build: https://minecraft.makecode.com/app/963bb0104ccea6485704df569d9ec7679280c62e-a705a6159a

### DIFF
--- a/webapp/src/monaco.tsx
+++ b/webapp/src/monaco.tsx
@@ -1033,6 +1033,7 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         if (!container || !this.blockInfo) return;
 
         const debugging = this.isDebugging();
+        if (debugging && this.flyout) this.flyout.setState( { groups: undefined } );
         const debuggerToolbox = debugging ? <DebuggerToolbox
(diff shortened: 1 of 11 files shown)


webapp/src/monaco.tsx
```diff
@@ -913,6 +898,36 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
+        if (this.insertionSnippet) {
+            this.dragCurrentPos = {
+                x: ev.clientX,
+                y: ev.clientY
+            }
+            this.onDragBlockThrottled(ev);
+        }
+    };
+
+    protected onPointerUp = (ev: any) => {
```
jwunderl: these `any`s could be `MouseEvent`, right?

shakao: sure, the touch event is different so it'd have to be MouseEvent | TouchEvent but that's probably still better than any

PR #6513, opened 2020-01-30
Update less and semantic

jwunderl:
This would require a ton of testing on every browser / target - leaving up this pr in case it is needed for accessibility fixes

Was doing this for the electron build, but decided to go another way for now; making this as a draft PR for now as it seems to work-ish and might need it later for possible accessibility fixes. Also just need to add a few minor notes for future me

using this ends up being a bit annoying as `npm link` doesn't seem to properly update the transitive dependencies here when you link pxt, so the targets will end up still having the old less / sui, putting things in a weird mismatched state


theme/theme.config
```diff
@@ -87,11 +87,13 @@
          Import Theme
 *******************************/
 
-@import "theme.less";
+@import (multiple) "theme.less";
```
jwunderl: The two changes here would need to be duplicated to every targets theme/theme.config due to changes in less

cli/cli.ts
```diff
@@ -1800,7 +1800,7 @@ function buildSemanticUIAsync(parsed?: commandParser.ParsedCommand) {
             return Promise.resolve();
         return nodeutil.spawnAsync({
             cmd: "node",
-            args: [lessCPath, "theme/blockly.less", "built/web/blockly.css", "--include-path=node_modules/semantic-ui-less:node_modules/pxt-core/theme:theme/foo/bar", "--no-ie-compat"]
+            args: [lessCPath, "theme/blockly.less", "built/web/blockly.css", "--include-path=node_modules/semantic-ui-less:node_modules/pxt-core/theme:theme/foo/bar"]
```
jwunderl: this flag is deprecated / default behavior; there's a bug in the current version of less that this causes an error, but in the future it would just silently ignore the flag so removing it seems good anyways

pelikhan:
This will require extensive testing on all browsers and editors.

Amerlander:
Since semantic ui seems not maintained anymore, have you considered to move on to fomantic ui?

For more details see their repo and the notes to semantic UI on top of the read me:
https://github.com/fomantic/Fomantic-UI

jwunderl:
I did take a look for a little bit, but that was just before I stalled on this branch and moved to another fix for the issue -- if I recall correctly, they don't currently have a fomantic-ui-react equivalent, so it would be using the current suir with fomantic ui / potentially have some weirdness there

PR #6514, opened 2020-01-31
Python general language pages

ganicke:
A modest attempt at a set of analogous basic ref pages for Python to compliment the current [JavaScript](https://makecode.microbit.org/javascript) pages.

RE: https://github.com/microsoft/pxt-minecraft/issues/1463


common-docs/python.md
```diff
@@ -0,0 +1,14 @@
...
+
+The Microsoft MakeCode programming environment uses Python along with the [JavaScript](/javascript) language.
+
+These topics give a brief introduction to Python with MakeCode:
+
+* **[Calling](/python/call)** - How to use a function
+* **[Sequencing](/python/sequence)** - Ordering statements in code
+* **[Variables](/python/variables)** -  Make a way to remember data
+* **[Operators](/python/operators)** - Things to change and compare values
+* **[Statements](/python/statements)** - An elements of to code to take action
```
darzu: Grammar seems off

common-docs/python/classes.md
```diff
@@ -0,0 +1,181 @@
...
+
+Let's take a look at an example:
+
+```python-ignore
+class Animal:
+    name = ""
+    def __init__(self, theName: str):
+        self.name = theName
+    def move(self, distanceInMeters = 0):
+        console.log(self.name + " moved " + str(distanceInMeters) + "m.")
```
darzu: we should use "print()" instead of console.log

common-docs/python/functions.md
```diff
@@ -0,0 +1,159 @@
+# Functions
+
+Functions are the fundamental building block of programs. Here is the simplest
+way to make a function that adds two numbers:
+
+```python
+// Named function
+def add(x: int, y: int):
```
darzu: In Python we try to avoid types whenever possible since that's not commonly used.

So just:
```
def add(x, y):
   return x + y
```
works

common-docs/python/functions.md
```diff
@@ -0,0 +1,159 @@
...
+    return x + y
+
+sum = add(1, 2)
+```
+
+### ~ hint
+
+For the @boardname@, you must specify a [type](/python/types) for each function parameter.
```
darzu: This shouldn't be true

common-docs/python/functions.md
```diff
@@ -0,0 +1,159 @@
...
+```python
+z = 100
+
+def addToZ(x: int, y: int):
+    return x + y + z
+
+sum = addToZ(1, 2)
+```
+
+Let's add a return type to our add function:
```
darzu: I don't think we need this example

common-docs/python/functions.md
```diff
@@ -0,0 +1,159 @@
...
+```python
+def add(x: int, y: int):
+    return x + y
+```
+
+Python can figure the return type out by looking at the return statements, so you can optionally leave this off in many cases.
+
+## Optional and Default Parameters
+
+In TypeScript, the number of arguments given to a function has to match the number of parameters the function expects.
```
darzu: I wouldn't reference TypeScript here. The majority of students are likely learning Python as their first language. You can just cut out the whole TypesScript reference IMO

common-docs/python/types.md
```diff
@@ -0,0 +1,188 @@
+# Types
```
darzu: I would actually leave out this whole page on Types. We try our best to infer types as much as possible and if there are common cases that still need them we're likely to improve our inference in the future. I think this section will just confuse folks, especially those more familiar with Python.

If we find that omitting this is causing too much confusion, we can add some version back in the future.
FYI @abchatra

common-docs/python/variables.md
```diff
@@ -0,0 +1,37 @@
...
+
+Declaring a variable is simply a matter of just assigning a value to a variable identifier:
+
+```python
+a = 10
+```
+
+## Block-scoping
+
+When a variable is declared with and assignment, it uses what some call *lexical-scoping* or *block-scoping*.
```
darzu: *an

common-docs/python.md
ganicke: It was done intentionally for the JavaScript page quite some time ago as a more relaxed style thing.

https://makecode.microbit.org/javascript

darzu:
@ganicke feel free to merge this whenever you feel it's ready.

PR #6515, opened 2020-01-31
Display language type in card icon on home screen

shakao:
Fixes https://github.com/microsoft/pxt-minecraft/issues/1627

Other icon changes (Github, board, etc) take precedence, otherwise display language.

image
image

### DIFF
--- a/theme/home.less
+++ b/theme/home.less
@@ -289,7 +289,7 @@
                 background-blend-mode: color-burn;
                 font-size: 6.5rem!important;
             }
-            .card-action .xicon.python { background-image: @pyIcon; }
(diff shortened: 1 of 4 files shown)


pelikhan:
The icon doesn’t match language selector. Are we changing icon for JS?

shakao:
image

i think it looks better with the square logo but not sure if it will be confusing for existing users--maybe this is a branding kind of decision, i'll check with jacqueline too

pelikhan:
It's a UI and change we need to assess the impact of this icon change. It will also impact all docs and potentially screenshots.

PR #6523, opened 2020-02-03
Gulp, TypeScript, and Monaco

riknoll:
This PR does three things:

1. Gets rid of Jake in favor of Gulp
2. Upgrades our TypeScript version to 3.7.5 
3. Upgrades our Monaco version to 0.19.3

It also removes our dependency on pxt-monaco-typescript and replaces it with the stock TS language service. The plan is to remove that service and switch entirely over to one implemented by us in the near future.

I'm marking this as a draft PR until I've opened PRs on each of the targets to upgrade to the new TS version.


gulpfile.js
```diff
@@ -0,0 +1,557 @@
+const fs = require("fs");
```
darzu: FWIW, gulp seems to support .ts pretty easily: "For TypeScript, rename to gulpfile.ts and install the ts-node module." https://gulpjs.com/docs/en/getting-started/javascript-and-gulpfiles

gulpfile.js
```diff
@@ -0,0 +1,557 @@
...
+    gulp.parallel(pxtjs, pxtdts, pxtapp, pxtworker, pxtembed),
+    targetjs,
+    gulp.parallel(buildcss, buildSVGIcons),
+    webapp,
+    browserifyWebapp,
+    gulp.parallel(semanticjs, copyJquery, copyWebapp, copyPlayground, copySemanticFonts, copyMonaco),
+    runUglify
+);
+
+const travis = gulp.series(buildAll, lint, testAll, targetjs, pxtTravis);
```
darzu: Much much cleaner!

webapp/src/monacoEditAmendments.ts
```diff
@@ -101,8 +101,11 @@ function applyAmendment(amendment: EditAmendmentInstance, editor: monaco.editor.
         // determine where to apply
         let amendRange = Object.assign({}, amendment.range)
```
darzu: u can probably get rid of this object.assign if you're going to do a spread below

darzu:
```suggestion
        let amendRange = amendment.range
```

darzu:
Very excited about gulp. It's a much better foundation.
I suspect we could get even more of an improvement if we could pipe output from tasks directly into each other instead of outputing files to built/ and then consuming from built/, but I would need to look into it more myself.

PR #6534, opened 2020-02-05
Embed sources in .hex using 0x0E record

mmoskal:
Fixes https://github.com/microsoft/pxt-microbit/issues/1631


mmoskal:
needs to be cherry-picked on whatever is the stable channel for micro:bit

PR #6538, opened 2020-02-05
Multi-editor / multi-tab

pelikhan:
Allows to editors to run in multiple tabs. A lock is kept per header and for the header list to trigger a refresh. Additional, a special mode allows to run simulator messages to the parent window rather than spinning multiple sims.
- [x] beefed up the lock situation in worksapce to support simulatneous editors in different windows. I may have gotten it right.
- [x] transfer messages and only spin 1 sim
- [x] figure out which editor url to launch
- [x] use per-header lock

Testing 2 editors together:
http://localhost:3232/multieditor.html


pelikhan:
@darzu please create a stable branch before taking this in. you don't need this for minecraft

webapp/src/workspace.ts
```diff
@@ -121,23 +121,21 @@ export function restoreFromBackupAsync(h: Header) {
...
         .then(() => {
             const backup = getHeader(refId);
             backup.isDeleted = true;
-            return saveAsync(backup);
+            return forceSaveAsync(backup);
         })
         .catch(() => {
             delete h.backupRef;
-            return saveAsync(h);
+            return forceSaveAsync(h);
```
darzu: Do all these force saves have any implications on performance or increase the chance of race conditions from double saves or something similar?

pelikhan: force save means that we override the lock (it has no perf implication) in this case.

webapp/src/workspace.ts
```diff
@@ -1217,10 +1268,12 @@ export function resetAsync() {
             if (Cloud.localToken)
                 pxt.storage.setLocal("local_token", Cloud.localToken);
         })
+        .then(() => syncAsync())
+        .then(() => { });
```
darzu: Do all these header and sync changes make us more robust or give us other advantages?

pelikhan: by doing a sync, we update the headers session marker and it will force other tabs to refresh (eventually).

PR #6541, opened 2020-02-05
Add keyboard button for ipads with ios 13

riknoll:


riknoll:
This check needs to be separate, because monaco handles other versions of ios. We only want this to appear for ipad >= 13

pelikhan:
Microbit is reporting that we don’t detect iOS correctly either.

martinwork:
The option at Settings/Safari/Request Desktop Website/All websites can change the User Agent to

Mozilla/5.0 (iPad; CPU OS 13_0 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.0 Mobile/15E148 Safari/604.1

Would the MacIntel test stop working then?

https://github.com/microsoft/pxt/blob/edfee811f62773c57115ca404a0a89f9d3b4e107/webapp/src/monaco.tsx#L661

PR #6544, opened 2020-02-06
Adding gulp watch task

riknoll:
Adding a "gulp watch" task to the gulpfile and removing that functionality from `pxt serve`.

The new workflow for developing PXT is to run `gulp watch` in one terminal and `pxt serve` in another. Serve watches for changes to the target and gulp watches for changes to core. This has a number of improvements over the old way:

1. Incremental compilation! No more full rebuilds on changes!
2. Better file watching! No more rebuilds triggered by running `git status` in pxt!
3. Faster `pxt serve`! No more waiting for pxt to build when you only made changes in the target!

Note that `pxt serve` no longer builds pxt core *at all*, so make sure you run gulp before starting the server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 * [Try out the editors in your browser...](https://makecode.com)
 
-[![Build Status](https://travis-ci.org/microsoft/pxt.svg?branch=master)](https://travis-ci.org/microsoft/pxt) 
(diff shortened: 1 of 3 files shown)


riknoll:
Also, this does nothing for changing dependencies so you'll have to run a full gulp build if you're working on Blockly, monaco, etc.

pelikhan:
instructions in readme :)

abchatra:
Very cool!

PR #6547, opened 2020-02-06
Always show appropriate language reference in help menu

shakao:
Move language help out of pxtarget and into web app. I'll update to each target to remove these language items from the docsMenu


webapp/src/container.tsx
```diff
@@ -42,13 +42,25 @@ function renderDocItems(parent: pxt.editor.IProjectView, elements: pxt.DocMenuEn
     );
 }
 
-export class DocsMenu extends data.PureComponent<ISettingsProps, {}> {
+// Always append a link to the appropriate language (Blocks, JS, Python) to the help menu
+function getDocsLanguageItem(editor: DocsMenuEditorName, parent: pxt.editor.IProjectView, cls: string = ""): JSX.Element {
+    const path = "/" + editor.toLowerCase();
+    return <DocsMenuItem key={"docsmenu" + path} role="menuitem" ariaLabel={pxt.Util.rlf(editor)} text={pxt.Util.rlf(editor)} className={`ui ${cls}`} parent={parent} path={path} onItemClick={openDocs} />
```
riknoll: I think you should probably use `lf()` instead of `rlf()` since these strings are always the same

PR #6549, opened 2020-02-06
Fix duplicate identifier names when using toolbox snippets

darzu:
Fixes: https://github.com/microsoft/pxt-minecraft/issues/1709


webapp/src/compiler.ts
```diff
@@ -400,6 +400,7 @@ export function formatAsync(input: string, pos: number) {
 }
 
 export function snippetAsync(qName: string, python?: boolean): Promise<string> {
+    console.log(`compiler snippet: ${qName} ${python}`)
```
shakao: small: remove

darzu: ty!!

PR #6565, opened 2020-02-10
don't try and find old copy of same project on load

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/2605; this was changed before in https://github.com/microsoft/pxt/pull/5535 but that basically got reverted in the ~160 commit github integration pr.

The behavior of loading the old project feels very confusing to me -- if I saw that it loaded the project, I'd probably think that meant the project was getting updated on the server as well / be very confused when i send the share link and my friend tells me there's a bug in it that isn't there when I load it.

Also dropped the return of the promise / 'async' suffix because it was just being called with an immediate .done()


webapp/src/app.tsx
```diff
@@ -3958,17 +3958,13 @@ async function importGithubProject(repoid: string, requireSignin?: boolean) {
...
-function loadHeaderBySharedIdAsync(id: string) {
+function loadHeaderBySharedId(id: string) {
     core.showLoading("loadingheader", lf("loading project..."));
 
-    // try to find project with same id
-    const hdid = workspace.getHeaders().find(h => h.pubId == id);
-
-    let p = hdid ? Promise.resolve(hdid) : workspace.installByIdAsync(id);
-    p.then(hd => theEditor.loadHeaderAsync(hd, null))
+    workspace.installByIdAsync(id)
```
pelikhan: return promise

jwunderl:
made it async to match with `importGithubProject`; can make it return the header but it feels a bit weird having that be inconsistent with the function it's always called with (`importGithubProject`) / returning something and always just `.done`ing it.

pelikhan:
let's go back to the non-async version without returning then.

pelikhan:
Needs to be integrated in microbit branch

PR #6568, opened 2020-02-11
handle variables matching top level constants, namespaces

jwunderl:
fix https://github.com/microsoft/pxt-minecraft/issues/1663 (variables conflicting with the new shorthand / alias constants in minecraft, like `LEFT`), as well as other variable conflicts (e.g. namespaces: `console`) that were broken due to the pkg being dropped from the cached api info.

### DIFF
--- a/webapp/src/compiler.ts
+++ b/webapp/src/compiler.ts
@@ -485,10 +485,15 @@ interface BundledPackage {
     files: pxt.Map<string>;
 }
 
+interface UsedPackageInfo {
+    dirname: string,
+    info: pxt.PackageApiInfo
+}
+
 async function getCachedApiInfoAsync(project: pkg.EditorPackage, bundled: pxt.Map<pxt.PackageApiInfo>): Promise<pxtc.ApisInfo> {
     if (!bundled) return null;
-
-    const corePkg = bundled["libs/" + pxt.appTarget.corepkg];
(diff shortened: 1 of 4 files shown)


webapp/src/compiler.ts
```diff
@@ -550,11 +564,15 @@ async function getCachedApiInfoAsync(project: pkg.EditorPackage, bundled: pxt.Ma
...
     };
 
-    for (const used of usedInfo) {
+    for (const used of usedPackageInfo) {
+        const { info, dirname } = used;
         if (!used) continue;
-        pxt.Util.jsonCopyFrom(result.byQName, used.apis.byQName);
+        pxt.Util.jsonCopyFrom(result.byQName, info.apis.byQName);
+        for (const api of Object.keys(info.apis.byQName)) {
+            result.byQName[api].pkg = dirname;
```
riknoll: Isn't the loop overwriting this in each iteration?

riknoll: i.e. for each used package

jwunderl: it's replacing the pkg of each one that has just been copied from info.apis.byQName into result; it's probably confusing to iterate over one map (the package) and update another (the output api info), I'll switch the order and update info.apis.byQName before copying

PR #6570, opened 2020-02-11
Download all projects from a discourse tag

pelikhan:
Download all scripts from discourse posts marked with a specific tag
```
pxt ddt game --md docs/community.md
```
Once drop locally, ``pxt build`` can be run in each folder.
It also optionally patches an existing file of community cards with the latest tags, drops the screenshot on disc.


cli/cli.ts
```diff
@@ -3119,6 +3119,118 @@ export function exportCppAsync(parsed: commandParser.ParsedCommand) {
...
+                .then(id => {
+                    if (!id) {
+                        pxt.log(`  --> unknown project id`)
+                        return Promise.resolve();
+                    }
+                    n++;
+                    return extractAsyncInternal(id, out, false)
+                        .then(() => {
+                            // does the current card have an image?
+                            let card = cards.filter(c => c.url == topic.url)[0];
```
```suggestion
                            let card = cards.find(c => c.url == topic.url);
```

pxtlib/discourse.ts
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+
+    export function extractSharedIdFromPostUrl(url: string): Promise<string> {
+        // https://docs.discourse.org/#tag/Posts%2Fpaths%2F~1posts~1%7Bid%7D.json%2Fget
+        return pxt.Util.httpGetJsonAsync(url + ".json")
+            .then((json: DiscoursePostResponse) => {
+                // extract from post_stream
+                let projectId = json.post_stream
+                    && json.post_stream.posts
```
jwunderl: we probably ought to update tslint (/ move to eslint if easy enough, as tslint is deprecated?), would be great to be able to use optional chaining in places like this / the million `pxt.appTarget.appTheme && ...`s (right now it marks that as an error as it uses an older typescript)

pelikhan: not in this pr

pelikhan:
This is ready to run. now we need to annotate forum posts.

cli/cli.ts
jwunderl: oh my bad here, I hadn't noticed `cli/tsconfig` doesn't define the es6 lib like the others so `find`'s not allowed; e.g. https://github.com/microsoft/pxt/blob/master/pxtcompiler/tsconfig.json#L15

PR #6573, opened 2020-02-11
Improvements for comment decompilation

riknoll:
Fixes https://github.com/microsoft/pxt-minecraft/issues/1581

This overhauls how we do comments in the decompiler (and python decompiler). I now extract all of the comments in the document and group them with the nearest statement. This also means that we shouldn't have disappearing comments anymore; in the worst case they will all become workspace comments.

When going to blocks, we use the following rules:
1. If a comment is followed by an empty line and is top level, it becomes a workspace comment
2. Single line comments are combined with adjacent single line comments but not with adjacent multi line comments
3. Trailing comments (comments at the end of a line that contains a statement) are grouped with the statement they appear on the same line with

There's a more detailed breakdown of the logic in the code. 

To demonstrate the grouping, this example code:
```typescript
player./** A */ onChat("trailing" /** B */, function (/** C */) {
    agent.teleportToPlayer( /** D */)
})

/**
 * E
 */

// F
agent.teleportToPlayer() // G

// H

player.onArrowShot(function() {
    // I
})
```

now converts into this blocks program:

Screen Shot 2020-02-11 at 1 50 13 PM

and this python code: 

```python
"""
A 
"""
"""
B 
"""
"""
C 
"""
"""
D 
"""
def on_chat():
    agent.teleport_to_player()
player.onChat("trailing", on_chat)
"""

E

"""
# F
agent.teleport_to_player()
# G
# H
# I
def arrow_shot():
    pass
player.on_arrow_shot(arrow_shot)
```


riknoll:
Also, I broke most of the test cases (which is why this PR is so huge). I removed all the comments from the decompiler tests that have nothing to do with comments so hopefully this shouldn't happen again.

PR #6582, opened 2020-02-12
tut / docs diffs ignore whitespace, show fileB content on no-diff

jwunderl:
re: https://github.com/microsoft/pxt-minecraft/issues/1744, the diffs renderer wasn't passing in `ignoreWhitespace`

see comment about seemingly minor diff algorithm change \\/

didn't look at the issue about it not rendering diffspy in the docs page, but it looks like that's just https://github.com/Microsoft/pxt/blob/master/pxtrunner/renderer.ts not having handling for  spydiffs  like /marked.tsx does?

### DIFF
--- a/pxtlib/diff.ts
+++ b/pxtlib/diff.ts
@@ -105,7 +105,7 @@ namespace pxt.diff {
             let y = x - k
             const snakeLen = V[MAX + k] - x
             for (let i = snakeLen - 1; i >= 0; --i)
-                diff.push("  " + a[x + i])
+                diff.push("  " + b[y + i])
 
             if (nextK == k - 1) {
                 diff.push("- " + a[x - 1])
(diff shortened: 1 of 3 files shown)


pxtlib/diff.ts
```diff
@@ -105,7 +105,7 @@ namespace pxt.diff {
             let y = x - k
             const snakeLen = V[MAX + k] - x
             for (let i = snakeLen - 1; i >= 0; --i)
-                diff.push("  " + a[x + i])
+                diff.push("  " + b[y + i])
```
jwunderl: does this make sense to take file b's content at the matching line instead of file a's? need that or you get the spacing from the previous step, which will be wrong when the spacing changed / etc - very problematic when using it for step diffs, as it will show up like this:

Screen Shot 2020-02-11 at 5 47 26 PM

instead of

Screen Shot 2020-02-11 at 7 02 47 PM

that change seems to fix it but i haven't looked too deeply at the diff algorithm / it's pretty dense

pelikhan: looks ok. good catch

PR #6588, opened 2020-02-12
Null check for params in block description

shakao:


webapp/src/monacoFlyout.tsx
```diff
@@ -229,7 +229,7 @@ export class MonacoFlyout extends React.Component<MonacoFlyoutProps, MonacoFlyou
         let parts = block.attributes._def && block.attributes._def.parts;
         let name = block.qName || block.name;
         if (parts) {
-            if (parts.filter((p: any) => p.kind == "param").length > params.length) {
+            if (parts.filter((p: any) => p.kind == "param").length > params?.length) {
```
pelikhan: params && .... ? The ?. happens after computation.

PR #6590, opened 2020-02-12
Skip transpile caching on code hover

shakao:
Fixes https://github.com/microsoft/pxt-minecraft/issues/1718

### DIFF
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -866,8 +866,6 @@ namespace ts.pxtc.service {
                 }
                 if (res.globalNames)
                     lastGlobalNames = res.globalNames
-                if (lastPyIdentifierSyntaxInfo)
-                    return lastPyIdentifierSyntaxInfo
             }
             return opts.syntaxInfo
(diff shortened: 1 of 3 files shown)


pxtcompiler/emitter/service.ts
```diff
@@ -866,8 +866,6 @@ namespace ts.pxtc.service {
                 }
                 if (res.globalNames)
                     lastGlobalNames = res.globalNames
-                if (lastPyIdentifierSyntaxInfo)
```
pelikhan: is this symbol used anywhere?

pxtcompiler/emitter/transpile.ts
```diff
@@ -76,9 +76,10 @@ namespace ts.pxtc.transpile {
         }
```
darzu: please add globalNames here if you wouldn't mind
I think that might solve some other naming issues

pxtcompiler/emitter/service.ts
shakao: nope, removed it!

PR #6598, opened 2020-02-14
Fixing autocomplete for TS and improving sorting of completions

riknoll:
Fixes https://github.com/microsoft/pxt-minecraft/issues/1725

Also adds an improvement to the sorting of parameter completions where the parameter value is a property access expression. Now we check the return type of the completion entry and weight it higher if it matches.

### DIFF
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -772,6 +772,11 @@ namespace ts.pxtc.service {
         decls: pxt.Map<Declaration>;
     }
 
+    interface CompletionSymbol {
+        symbol: SymbolInfo;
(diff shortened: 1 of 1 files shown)


darzu:
I noticed we currently have this bug (notice pos and randpos):
image

I wonder if your PR fixes this too?

darzu:
After sleeping on this, I'm leaning towards not including this in the release. It's too much risk, and there will always be another release.

darzu:
Does this PR also fix: https://github.com/microsoft/pxt-minecraft/issues/1769 ?

riknoll:
No, that issue is because of the aliases. We never return positions.random(), only randpos()

PR #6599, opened 2020-02-14
Proper nonlocal & global variable scope analysis for ts->py

darzu:
Fixes: https://github.com/microsoft/pxt-minecraft/issues/1723
as well as many other subtle issues around variables that might have arisen in user programs.

For example, if a user had:
```typescript
let d = 3
player.onChat("set", function on_chat() {
    d = 7
})
player.onChat("run", function () {
    agent.move(FORWARD, d)
})
```
Typing "set" then "run" causes "7" blocks to be placed

If they then switched to Python, they would have gotten:
```python
d = 3
def on_chat():
    d = 7
player.on_chat("set", on_chat)
def on_chat2():
    agent.move(FORWARD, d)
player.on_chat("run", on_chat2)
```
Then, typing "set" then "run" causes "3" blocks to be placed (!!!)

So now after this fix we correctly output:
```python
d = 3
def on_chat():
    global d
    d = 7
player.on_chat("set", on_chat)
def on_chat2():
    agent.move(FORWARD, d)
player.on_chat("run", on_chat2)
```
which places "7" correctly

---

What was needed was a static analysis of all variables to determine which scope they originate from.


pxtpy/pydecompiler.ts
```diff
@@ -769,13 +715,21 @@ namespace pxt.py {
...
             if (stmts.length) {
+                // global or nonlocal declerations
+                let globals = scopeLookup.getExplicitGlobals(s)
+                    .map(g => getName(g))
+                if (globals && globals.length)
+                    stmts.unshift(`global ${globals.join(", ")}`)
+                let nonlocals = scopeLookup.getExplicitNonlocals(s)
+                    .map(n => getName(n))
+                if (nonlocals && nonlocals.length)
+                    stmts.unshift(`nonlocal ${nonlocals.join(", ")}`)
```
darzu: Note that even though this PR is somewhat large, it all comes down to these few lines. These are the only lines that actually change the way python code is generated.

darzu:
After sleeping on this, I'm leaning towards not including this in the release. It's too much risk, and there will always be another release.

pxtpy/varScopes.ts
```diff
@@ -0,0 +1,311 @@
...
+            ts.SyntaxKind.MinusEqualsToken, ts.SyntaxKind.AsteriskEqualsToken,
+            ts.SyntaxKind.AsteriskAsteriskEqualsToken, ts.SyntaxKind.SlashEqualsToken,
+            ts.SyntaxKind.PercentEqualsToken, ts.SyntaxKind.LessThanLessThanEqualsToken,
+            ts.SyntaxKind.GreaterThanGreaterThanEqualsToken,
+            ts.SyntaxKind.GreaterThanGreaterThanGreaterThanEqualsToken,
+            ts.SyntaxKind.AmpersandEqualsToken, ts.SyntaxKind.BarEqualsToken,
+            ts.SyntaxKind.CaretEqualsToken
+        ]
+        return ts.isBinaryExpression(s)
+            && !!AssignmentOperators.find(o => s.operatorToken.kind === o)
```
riknoll: nitpick: `.some()` instead of `.find()`

pelikhan:
looks like you have a good set of tests so we should be good. if you have doubts, keep adding tests.

pxtpy/varScopes.ts
```diff
@@ -0,0 +1,311 @@
...
+    }
+    type VarUse = VarRead | VarAssign
+    type VarRef = VarRead | VarDecl | VarAssign
+    interface VarScope {
+        refs: VarRef[],
+        children: VarScope[],
+        owner: ts.FunctionLikeDeclaration | undefined
+    }
+    function computeVarScopes(node: ts.Node): VarScope {
+        const EMPTY: VarScope = { refs: [], children: [], owner: undefined }
```
riknoll: Is this readonly? If so, could you add that to the interface?

darzu:
> I would prefer if there were non end-to-end tests as well. Looks like the runtime trace test assumes that the pydecompiler is emitting the code in a certain way; would be nice to have code that tests this explicitly and is written in Python

We could do literal syntax checking, however I've found those have been very brittle and require constant tweaking when there are unrelated changes, so I've been prefering to do test that check the runtime observable behavior of programs.

I'm all for new test systems, but it takes work and right now I don't have a clear idea on tests that I could add that would be good ROI.

pxtpy/varScopes.ts
darzu: I forgot about readonly, this is great thanks!

PR #6603, opened 2020-02-14
non-greedy whitespace eating in tutorial parsing

jwunderl:
Right now it's eating up all the whitespace, this makes it so you can put in empty macros to:

````
```template

```
````

Not perfect, as it will still fail on 

````
```template
```
````

but I'm having trouble reasoning through a full fix for that when this part is needed for minecraft  hour of code tuts

### DIFF
--- a/pxtlib/tutorial.ts
+++ b/pxtlib/tutorial.ts
@@ -39,7 +39,7 @@ namespace pxt.tutorial {
     function computeBodyMetadata(body: string) {
         // collect code and infer editor
         let editor: string = undefined;
-        const regex = /```(sim|block|blocks|filterblocks|spy|ghost|typescript|ts|js|javascript|template|python)?\s*\n([\s\S]*?)\n```/gmi;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: pxtlib/tutorial.ts [hunk 1]
```diff
@@ -39,7 +39,7 @@ namespace pxt.tutorial {
     function computeBodyMetadata(body: string) {
         // collect code and infer editor
         let editor: string = undefined;
-        const regex = /```(sim|block|blocks|filterblocks|spy|ghost|typescript|ts|js|javascript|template|python)?\s*\n([\s\S]*?)\n```/gmi;
+        const regex = /```(sim|block|blocks|filterblocks|spy|ghost|typescript|ts|js|javascript|template|python)?\s*?\n([\s\S]*?)\n```/gmi;
```
jwunderl:
```suggestion
        const regex = /```(sim|block|blocks|filterblocks|spy|ghost|typescript|ts|js|javascript|template|python)?\s*?\n([\s\S]*?)\n?```/gmi;
```
seems sufficient to handle the empty case, but allows the last line of code to end in ```, which isn't that big of a deal but ¯\\_(ツ)_/¯

we decided to go with a fix that just doesn't emit empty comments for now, as this would need a LOT of testing / validation

PR #6609, opened 2020-02-18
unbox snippets when rendering in tip

pelikhan:
Hints only contain a single code snippets (they could have more but they shouldn't). We waste space by wrapping the snippet in a segment that adds padding. Don't wrap snippets when rendering tutorial hint.
before:
image
after:
image


pelikhan:
Could probably be done in css...

PR #6611, opened 2020-02-18
Glowing hints in tutorials

jwunderl:
make hints stand out a bit more; this drops the avatar image which we decided in the design meeting didn't really indicate a hint too well, gives them a glow, and moves them to the right side so it is further away from the other enticing content on the left side of the screen

~making this a draft to start as I would imagine we'll want to wait for a design meeting for to verify / confirmation with hour of code version as well~ no releases on the immediate horizon, can discuss at a future design meetinf but not blocking


jwunderl:
changes from design meeting:

* blue instead of orange, to 'blend' with icon
* tiny bit smaller
* tiny offwhite outline
* tie it to tooltip / squiggle
* make tooltip stay up longer (10s), show up faster (5s after first activity)
* clear tooltip on clicking hint
* margin on left so text isn't so smushed

PR #6614, opened 2020-02-18
Upgrade PouchDB version

shakao:
https://arcade.makecode.com/app/10d3c526fe89c85ad523cd2b55393f24c574c813-8cb0113b4c

For compliance with OSS Security review, we need >=6.0.5. Tested in Chrome, Firefox, and (new) Edge, created projects persisted


pelikhan:
What about IE?

abchatra:
>> What about IE?

Dropping IE support everywhere.

pelikhan:
Then we need to enable the flag that forbids IE and update the docs in this PR too.

pelikhan:
This change needs to enable the flag to redirect the user to the browsers page.

shakao:
Fix for IE11 redirect, add to base pxtarget (we will still need to add it to the microbit pxtarget, as that overwrites the base one)

pxtlib/browserutils.ts
```diff
@@ -218,8 +218,10 @@ namespace pxt.BrowserUtils {
         if (/bot|crawler|spider|crawling/i.test(navigator.userAgent))
             return true;
 
-        //Check target theme to see if this browser is supported
-        if (pxt.appTarget.unsupportedBrowsers && pxt.appTarget.unsupportedBrowsers.some(b => b.id == browser())) {
+        // Check target theme to see if this browser is supported
+        let unsupportedBrowsers = pxt.appTarget?.unsupportedBrowsers
```
pelikhan: nit: use const

PR #6620, opened 2020-02-19
add selected language restriction to tick for new project dialog

jwunderl:
add the selected language restriction to the tick for new projects. Currently that will be `"" | "python-only" | "javascript-only" | "no-blocks"`


jwunderl:
not currently; you want one on show?

PR #6625, opened 2020-02-20
use more space of the card to display script title

pelikhan:
Mobile view not modified otherwise, should title over multiple lines. Btw, we don't have any limitation of size on the title...
image


riknoll:
Will this cause bleeding (like https://github.com/microsoft/pxt-minecraft/issues/1631)

pelikhan:
@riknoll the bleeding is the same as the current behavior. this change does not solve this.
image

PR #6629, opened 2020-02-21
github history: restore any commit

pelikhan:
Ability to restore any existing commit.
- [x] view commit list in "history" section
- [x] diff each commits w.r.t. to current files (refactored diff component)
- [x] restore command creates a new commit and copies the tree sha into the new commit

### DIFF
--- a/webapp/src/workspace.ts
+++ b/webapp/src/workspace.ts
@@ -737,7 +737,7 @@ export async function restoreCommitAsync(hd: Header, commit: pxt.github.CommitIn
     }
 
     const commitId = await pxt.github.createObjectAsync(parsed.fullName, "commit", restored)
-    await pxt.github.fastForwardAsync(parsed.fullName, parsed.tag, commitId, true)
+    await pxt.github.fastForwardAsync(parsed.fullName, parsed.tag, commitId)
     await githubUpdateToAsync(hd, {
         repo: gitjson.repo,
         sha: commitId,
--- a/webapp/src/workspace.ts
+++ b/webapp/src/workspace.ts
@@ -735,7 +735,7 @@ export async function restoreCommitAsync(hd: Header, commit: pxt.github.CommitIn
     }
     
     const commitId = await pxt.github.createObjectAsync(parsed.fullName, "commit", restored)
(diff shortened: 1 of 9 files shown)


darzu:
This is awesome!

I think having this for all projects would be awesome. What if we used git in the browser? So before they log into github, they can still get a bunch of these nice features. Then if they do log into github and want to host there, we can literally just push their git repo.

Here's a decent git implementation in JS. I doubt it's very big:
https://github.com/isomorphic-git/isomorphic-git

This might also be a good workaround to users or MakeCode bugs accidently deleting work and not being able to undo.

webapp/src/workspace.ts
```diff
@@ -721,6 +721,29 @@ export async function commitAsync(hd: Header, options: CommitOptions = {}) {
     }
 }
 
+export async function restoreCommitAsync(hd: Header, commit: pxt.github.CommitInfo) {
+    await cloudsync.ensureGitHubTokenAsync();
```
pelikhan: @mmoskal please double check my git skillz here

webapp/src/workspace.ts
```diff
@@ -721,6 +721,29 @@ export async function commitAsync(hd: Header, options: CommitOptions = {}) {
...
+    const parsed = pxt.github.parseRepoId(gitjson.repo)
+    const date = new Date(Date.parse(commit.committer.date));
+    const restored: pxt.github.CreateCommitReq = {
+        message: lf("Restore '{0} {1}'", date.toLocaleString(), commit.message),
+        parents: [gitjson.commit.sha],
+        tree: commit.tree.sha
+    }
+
+    const commitId = await pxt.github.createObjectAsync(parsed.fullName, "commit", restored)
+    await pxt.github.fastForwardAsync(parsed.fullName, parsed.tag, commitId, true)
```
mmoskal: I think there should be no force here.

pelikhan: It does fail saying an update is not a fast forward

mmoskal: If it fails, it means the parent of the commit you're creating is not the current head.

mmoskal: In other words, if it fails and your force it, you *will* lose commits

PR #6634, opened 2020-02-25
Service worker (aka the death of the application cache)

riknoll:
Adding service workers to replace the appcache. There are two service workers:

1. `/ref---serviceworker` covers the webapp
2. `/ref---simserviceworker` covers the simulator

The simulator service worker can be configured per-target by adding a file called `workerConfig.js` inside `sim/public` in the target repo. The script gets imported by the service worker and can add additional URLs like this:

```typescript
self.setSimulatorWorkerOptions({
    urls: ["/sim/sim.css"]
})
```

Just like in the manifest, the `/sim/` and `/cdn/` routes can be used in that file and will be patched by the backend. The `workerConfig.js` file is optional and we just ignore it if it returns a 404.

Because all of our named releases are located in the same 'subdirectory', all refs for a target share the same serviceworker (e.g. `/index.html`, `/beta`, `/stable`, etc.). There's no way to get around this because of how the serviceworker API is set up, but it doesn't really matter because each named release has a separate cache for requests/responses so they shouldn't interfere with each other. However, you might get some weird behavior if you have `/beta` open in one tab and `/index.html` open in another.

The reload logic works the same as before; if there is an update you'll get a refresh after three seconds. You will also get a refresh if you open beta and then index.html because of the shared workers.

This PR is marked as a draft because it relies on a backend change and I need to do some more testing.


riknoll:
Here is a test URL with the application cache disabled:

https://arcade.staging.pxt.io/app/c1cfd4e5a8768eb48acbc7e4304448aea3f29042-43c9bf9eba#editor

Proof that the appcache is disabled is [here](https://arcade.staging.pxt.io/app/c1cfd4e5a8768eb48acbc7e4304448aea3f29042-43c9bf9eba---manifest) and [here](https://trg-arcade.staging.userpxt.io/app/c1cfd4e5a8768eb48acbc7e4304448aea3f29042-43c9bf9eba---simmanifest)

riknoll:
Here is a test URL with the application cache enabled:

https://arcade.staging.pxt.io/app/02579385fa85156b7d5f113e1eb07b74ffba1a94-d3f14ab42e

Proof that the appcache is enabled is [here](https://arcade.staging.pxt.io/app/02579385fa85156b7d5f113e1eb07b74ffba1a94-d3f14ab42e---manifest) and [here](https://trg-arcade.staging.userpxt.io/app/02579385fa85156b7d5f113e1eb07b74ffba1a94-d3f14ab42e---simmanifest)

riknoll:
To test this change:

1. Open one of the test links above
2. Create a new project and wait for the simulator to load
3. Disconnect from the Internet. Also disable the network cache in the devtools (it's usually an option in the `Network` tab)
4. Reload the editor. You should see both the webapp and the simulator

On macOS, I've tested Chrome, Safari, and Firefox

On Windows, I've tested old Edge (v44), new Edge (v80), and Chrome

All of my tests were done with the application cache disabled

riknoll:
Also, the simulator might look a little weird; it's my local build and I made some changes to make sure it was working. As long as the simulator shows up with the network turned off, all is well.

riknoll:
@pelikhan yeah, probably. Do you have any idea how I can test that scenario? I put the relative paths into the webConfig in the cli so I think it should work fine

pelikhan:
you can run "pxt staticpkg --yt" on microsoft/pxt-samples. Or completely disable service worker for staticpkg. Does it interfere with electron?

riknoll:
It shouldn't "interfere". I'm doing feature detection around the feature so at worst it should just be a no-op

pelikhan:
What's blocking here?

pelikhan:
it would be a prefect moment for google to remove service workers since web devs can't go out on the street with pickforks

abchatra:
@riknoll is this ready to be merged?

riknoll:
Let me merge in master and do some testing. Should be ready!

riknoll:
I added one quick change to disable service workers on localhost. @pelikhan @abchatra @jwunderl please review.

Here is a test URL if anyone wants to test it out:
https://makecode.microbit.org/app/e5955391b8e4ff6e21e38b6d7751db2eb2485315-2411bce9c5

That test URL is actually slightly modified to enable service workers on /app urls which are normally disabled. That's the only change.

pelikhan:
Do you have PRs ready for the targets?

riknoll:
@pelikhan no i do not!

riknoll:
Not all the targets need PRs, only ones with extra files in the sim. micro:bit does not need any additional configuration

PR #6636, opened 2020-02-26
Updating the browsers with the edge download link

abchatra:
Fixing some of the existing information and restricting list to what we can test.

### DIFF
--- a/common-docs/browsers.md
+++ b/common-docs/browsers.md
@@ -5,7 +5,7 @@
 You need one of these browsers running on Windows 7, Windows 8, Windows 8.1, or
 Windows 10:
 
-* Microsoft Edge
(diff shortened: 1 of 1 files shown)


common-docs/browsers.md
```diff
@@ -36,12 +38,18 @@ here](/raspberry-pi).
...
 ## Chrome OS
 
-* Version 69 or higher
+* Version 70 or higher
+
+## Internet Explorer
+
+Internet Explorer (IE) is **not** supported. Please upgarde to Microsoft Edge ([Download](https://www.microsoft.com/en-us/edge)). You can check the settings to see if you are running Interent Explorer.
+
+![](/static/configurations/windows-version.png)
```
ganicke: How does this show if you're running Internet Explorer?

```suggestion
![Internet Explorer version](/static/configurations/ie-version.png)
```

Issue #6644, opened 2020-02-28
update python docs


pelikhan:
The first hit on "makecode python" goes to https://makecode.com/js/python which is outdated.

ganicke:
Current search rank for "MakeCode Python":

**Bing**

![image](https://user-images.githubusercontent.com/27789908/75618847-88ca4c00-5b28-11ea-9599-e6f3dcd7a3dd.png)

**Google**

![image](https://user-images.githubusercontent.com/27789908/75618850-9384e100-5b28-11ea-94cd-e08f451e5e47.png)

ganicke:
Python "landing" page is at https://makecode.com/python from #6646.

PR #6646, opened 2020-03-01
Update Python implmentation page

ganicke:
Do we want to generalize this page more, ditch it, or better promote https://makecode.com/language?

RE: #6644


pelikhan:
IMO redirect to /python

pelikhan:
The content of this page should be merged in the "/python". This page should be come a redirect instead (use python-ref.json file).

docs/js/python.md
```diff
@@ -1,21 +1,49 @@
-# Python Support
+# Python in MakeCode
 
-**Python support in MakeCode is experimental**
+Our Python support in MakeCode is called Static Python (SPY). As with TypeScript,
```
abchatra: We decided to call it just MakeCode Python instead of Static Python. Lets omit this line and refer to the python implementation as MakeCode Python.

pelikhan:
ping @ganicke @abchatra what's the status?

ganicke:
Sound good. Also, need to chop the https://minecraft.makecode.com/python page so Minecraft picks up the common-docs/python.md.

Issue #6291, opened 2019-12-05
Python toolbox support


darzu:
Enable the code toolbox in Python.

This issue is for tracking the experimental feature.

PR #6651, opened 2020-03-03
remove the pythonToolbox flag

pelikhan:
The python toolbox is stable and should be always enabled (1 less flag!)


pelikhan:
Closes https://github.com/microsoft/pxt/issues/6291

PR #6654, opened 2020-03-03
Shuffle gallery

pelikhan:
- [x] Ability to specify that a gallery should be shuffled. It provides a stable shuffled based on the day (more options can be added later)
```
        "Community Games": {
            "url": "community",
            "shuffle": "daily"
        },
```
- [x] allow to filter a gallery by locales too

### DIFF
--- a/webapp/src/projects.tsx
+++ b/webapp/src/projects.tsx
@@ -448,7 +448,7 @@ export class ProjectsCarousel extends data.Component<ProjectsCarouselProps, Proj
                     // keep last one
                     const last = this.prevGalleries.pop();
                     // shuffle array
-                    const now = new Date(Date.now());
+                    const now = new Date();
                     const seed = now.toDateString();
                     this.prevGalleries.sort((l, r) =>
                         ts.pxtc.Util.codalHash16(l.name + seed)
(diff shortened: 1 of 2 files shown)


pelikhan:
@ganicke i realize we have not documentation about configuring the home screen...

webapp/src/projects.tsx
```diff
@@ -430,6 +443,21 @@ export class ProjectsCarousel extends data.Component<ProjectsCarouselProps, Proj
                 this.hasFetchErrors = true;
             } else {
                 this.prevGalleries = pxt.Util.concat(res.map(g => g.cards));
+                const shuffle = this.props.shuffle
+                if (shuffle) {
+                    // keep last one
+                    const last = this.prevGalleries.pop();
```
mmoskal: why?

webapp/src/projects.tsx
```diff
@@ -430,6 +443,21 @@ export class ProjectsCarousel extends data.Component<ProjectsCarouselProps, Proj
...
+                    // shuffle array
+                    const now = new Date(Date.now());
+                    const seed = now.toDateString();
```
mmoskal: Could just do `const seed = new Date().toDateString();`

This is now time-zone and locale dependent, but likely the classroom will have matching settings.

mmoskal: I take it back, it's only time-zone dependent, so we're all clear.

webapp/src/projects.tsx
pelikhan: the last card a "add your game"
image

webapp/src/projects.tsx
pelikhan: should i make the change or not? :)

mmoskal: change is just minor opt, I meant I take back the locale - dependent bit. just merge ;)

PR #6669, opened 2020-03-05
setup lang picker in right space so it works on core docs

jwunderl:
render snippets exits if 'probably in pxt docs', and I had this at the end of render snippets for some reason..

Better to initiate with the other set up code anyways

should fix lang picker on makecode.com/blog (well, it will remove it as it langs not selectable in pxtarget right now, needs the flag and a few langs if you want it there: https://github.com/microsoft/pxt/blob/master/pxtarget.json)

### DIFF
--- a/docfiles/docs.js
+++ b/docfiles/docs.js
@@ -247,7 +247,6 @@ function setupBlocklyAsync() {
                 })
         })
     }
-    setupLangPicker();
     return promise;
 }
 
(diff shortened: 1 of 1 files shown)


jwunderl:
actually looks like it needs a bit more work to have a language picker work on makecode.com as well, right now we don't have the pxt.Util / pxt.appTarget available in makecode.com/blog.

Kind of weird note, after everything is set up / if you just use the chrome console without any breakpoints, local serve in pxt docs has pxt.Util available / ksRunnerReady defined, but makecode.com/blog does not; I'm guessing that's known?

PR #6679, opened 2020-03-09
Updating dependencies

riknoll:


pxtlib/docsrender.ts
```diff
@@ -852,13 +862,15 @@ ${opts.repo.name.replace(/^pxt-/, '')}=github:${opts.repo.fullName}#${opts.repo.
...
 
         const markedInstance = pxt.docs.requireMarked();
+        const sanitizer = requireDOMSanitizer();
         const options = {
             renderer: new markedInstance.Renderer(),
             gfm: true,
             tables: false,
             breaks: false,
             pedantic: false,
             sanitize: false,
```
pelikhan: ``sanitize: false`` ?

riknoll:
@pelikhan there are no new files; it's inlined like the lzma worker

pxtlib/docsrender.ts
riknoll: fixed

PR #6680, opened 2020-03-10
delete comments when compiling ts project

qumeta:
delete comments when compiling ts project


riknoll:
I don't see a need for this change. We run uglify in production which takes care of the comments

PR #6688, opened 2020-03-10
unify parsing of "tutorial", "example", "recipe" routes

pelikhan:
- [x] refactor loading code for "tutorial", "example", "recipe" routes
- [x] use common loading code in all instances of loading
- [x] support for loading example from github repo
- [x] support for template route


webapp/src/app.tsx
```diff
@@ -2064,13 +2064,15 @@ export class ProjectView
     importExampleAsync(options: pxt.editor.ExampleImportOptions): Promise<void> {
         const { name, path, loadBlocks, prj, preferredEditor } = options;
         core.showLoading("changingcode", lf("loading..."));
-        let features: string[];
-        return pxt.gallery.loadExampleAsync(name.toLowerCase(), path)
```
shakao: is 'loadExampleAsync' still being used anywhere?

PR #6689, opened 2020-03-10
Image Editor: keybindings for colors, suppress delete key

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/1777 (unless there are other blockly shortcuts to prevent, but I couldn't think of any right away) and fix https://github.com/microsoft/pxt-arcade/issues/1778

had the shortcuts for colors already in the old editor (https://github.com/microsoft/pxt/pull/5549), but never used them as much so I never noticed they were gone in the new one


webapp/src/components/ImageEditor/keyboardShortcuts.ts
```diff
@@ -59,6 +68,12 @@ function handleKeyDown(event: KeyboardEvent) {
             swapForegroundBackground();
             break;
     }
+
+    if (/^Digit\d$/.test(event.code)) {
+        const keyAsNum = +event.code.slice(-1);
+        const color = keyAsNum + (event.shiftKey ? 9 : 0);
```
riknoll: We should double check that we enforce the max index of 16 somewhere

webapp/src/components/ImageEditor/keyboardShortcuts.ts
```diff
@@ -59,6 +68,12 @@ function handleKeyDown(event: KeyboardEvent) {
             swapForegroundBackground();
             break;
     }
+
+    if (/^Digit\d$/.test(event.code)) {
```
riknoll: Should probably disable this for TileMaps

PR #6695, opened 2020-03-11
Adding table of servicing branches

riknoll:


abchatra:
This is go stale soon and no one will update. I will always figure out going to about-> version of pxt and see what is shipped. There should be stable channel corresponding to that.

riknoll:
Yeah, but it took me an hour to figure out all of them yesterday. Sometimes we need to port fixes to all branches

abchatra:
What I mean is, it won't be sufficient to look at this table at all. It will be stale and we will have to do the due diligence every time. We create stable channel every week.

PR #6697, opened 2020-03-11
Updated Blockly with new renderer, field cleanup, fixes

shakao:
Target | URL
-- | --
Microbit | https://makecode.microbit.org/app/37d71e8c2bd3843b169576d48038de235b0761e5-ec23236843
Minecraft | https://minecraft.makecode.com/app/0d322a95b0721c40d71302176752943520e69b2e-781d021a96
Adafruit | https://makecode.adafruit.com/app/49ad4aa3ae8e972ebb06795dc6aa5f66cb3806b3-e6efccdb5f
Arcade | https://arcade.makecode.com/app/1cbbf71dad03e60e84616fb0bba16dbdc7ad25bf-0bd8f72393
 
Known issues (fixing after merge):
- Small text field overlap when piano editor is first opened
- Insertion marker does not consider child blocks (intentional with new renderer)
- (arcade) Image field rendering (white instead of gray transparency)

Non-regression bugs:
- Duplicating function does not correctly rename
- (Arcade) Color field doesn't highlight selected color

Accompanying PRs:
- [x] https://github.com/microsoft/pxt-blockly/pull/269 (Blockly merge)
- [ ] https://github.com/microsoft/pxt-microbit/pull/2658 (Microbit field editor fix)
- [ ] https://github.com/microsoft/pxt-adafruit/pull/1159 (Adafruit field editor fix)


pxtblocks/fields/field_images.ts
```diff
@@ -98,30 +99,28 @@ namespace pxtblockly {
...
             // Update colour to look selected.
-            if (this.sourceBlock_.isShadow()) {
-                this.savedPrimary_ = this.sourceBlock_.getColour();
-                this.sourceBlock_.setColour(this.sourceBlock_.getColourTertiary(),
-                    this.sourceBlock_.getColourSecondary(), this.sourceBlock_.getColourTertiary());
-            } else if (this.box_) {
-                this.box_.setAttribute('fill', this.sourceBlock_.getColourTertiary());
+            let source = this.sourceBlock_ as Blockly.BlockSvg;
+            this.savedPrimary_ = source.getColour();
+            if (source?.isShadow()) {
```
riknoll: Is there a reason this access is guarded but the line above isn't?

pxtblocks/fields/field_toggle.ts
```diff
@@ -161,6 +160,19 @@ namespace pxtblockly {
             return this.sourceBlock_.isShadow() ? this.sourceBlock_.getOutputShape() : Blockly.OUTPUT_SHAPE_SQUARE;
         }
 
+        doClassValidation_(newBool: string) {
+            return typeof this.fromVal(newBool) == "boolean" ? newBool : "false";
+        }
+
+        applyColour = function() {
```
riknoll: should this be an arrow function? Also, why is it a property and not a method?

pxtblocks/fields/field_note.ts
```diff
@@ -228,13 +228,13 @@ namespace pxtblockly {
          * refresh the text to get back into a valid state.
          **/
         protected refreshText() {
-            this.setText(this.getText());
```
jwunderl: the reason the set text was here was that there were a few situations in which the display wouldn't update properly / actually rerender when things have been swapped; I believe the main one was after typing in a note and committing it by pressing enter (or clearing the entry and pressing enter) -- I wouldn't be surprised if that was fixed elsewhere as there wasn't a clear reason why it was broken, but if you could double check that works that would be good~

webapp/public/vs/editor/editor.main.css
```diff
@@ -3,4 +3,4 @@
...
...AABIzU0JisBERQWOwEVIzUzMjY1ESMiBh0BIzUhFyc3FxUHJyEnNycHFRcCsD8SDUASDR+7Hw0SPQ0SPgF3tlwscm8s/dFcWSxvcgJxHw0T/qgNEj8/Eg0BWBMNH339Wy1yLG8tW1gtbyxyAAIAAAAAA6sDrwAiAEoAADciJicuATc2NyY2Nz4BNz4BHwEHFzcXHgEGBwYHDgEnBgcGASIHBgcOAR8BBwYHBgcGFhceATM2NzY/ARcWNjc2Nz4BNTQnByc3Jq8WKQ8eCBp31RECEhA1IzmGPie0T7gRFQEoJhUZNX4742odAcpANBQRLhYbCA5ZX2k2CgQOCBUKEAxi9g4TMWwsFBEeHwSin50UQBIRHFMifdEvYy4nPxUmCxoRuE+0JzFsYyYVDyMPFuZiGwMwIw0QLoA7Ew5XX2k5ECUNCAoBClz4DwkWCB4NEB5MKhQUnZ+iBAAAAAIAAAAAAy0DqgAHABsAABMHERchNxEnBTUhESE1MzUjNTM1IzUzNSM1MzX6Pj4B9D8//gwB9P4MfX36+n19+gOqP/0SPj4C7j99Pv0SPz59P30+fT8AAAAIAAAAAAOqA6oACQANABEAFQAZAB0AIQAlAAATBxEzESERMxEnATUzFTcjFTM3MxUjNyMVMzczFSMhNSMVJzMVI5wfPwKvPyD88z99Pz8+Pz+8Pz8+Pz8BOT99Pz8DqiD9MgKv/VECziD8lT4+Pj4+Pj4+Pj4+Pj4+AAAABwAAAAADqgNrAAcACwAhACsAOwBHAFsAABMHERchNxEnAREhESUzNTQjIgcGBxU2MhUHBhUUFjcyNzM1FRQGIwYmNTQ3FzEVIzUzFTE2MhYVFAYnIicVBhYyNjU0JiIHBhcyNzUOASY0NhYXNSYjIgYVFBYzfT4+Au4/P/0SAu795Cg7DQ8NCxYxJTAXFBsOAQ4NCAsVpSUlDjAaHhgXDAEOGA4NFwgI5x4QDyYYGCYPDBsnLywjA2s+/c0+PgIzPv2PAjP9zcBrTAUDByoTHQYINBkeAR47DBEVAQ0JFQM9E/ZpHSwnKTMBXREOFxsZFBgMDm0KLQ0BHDAdAQ0vCDUtKDIAAAAABgAAAAADqgNrAAcACwATABgAIAAlAAATBxUXITc1JwUhFSEHNzMXFQcjJzcjFTM1ITczFxUHIyc3IxUzNX0+PgLuPz/9EgLu/RI+P
```
riknoll: Was this intentional?

pxtblocks/fields/field_note.ts
shakao: yep, i tested typing for both named and non-named notes when upgrading the field, but the build links above are current if you want to double check!

shakao:
@riknoll the main blockly dropdown now supports rendering an image (https://github.com/microsoft/pxt-blockly/blob/develop/core/field_dropdown.js#L618), settext has been deprecated entirely, partially combined with doValueUpdate and partially moved back to blockly-core, and since the svg arrow is in blockly-core as well we use their built in updateSize_. 

i suspect a bit of this will come back because we need to render selected text for the microbit image dropdown, but for now the base blockly rendering looks reasonable

webapp/public/vs/editor/editor.main.css
shakao: whoops nope. fixed!

pxtblocks/fields/field_note.ts
jwunderl: Ahh, I think the overlap with `for` when opening might have been fixed with resetting the text to get the updated size before and that can be fixed later, everything else seems fine~

riknoll:
Toggles look a little weird:

PR #6698, opened 2020-03-11
Only upload ide strings to crowdin if on tagged release on master

riknoll:
Right now we upload the ide strings for any tagged release and any commit on master. This tweaks that logic so that we now only upload on tagged commits that are on master (and not untagged commits).

We run two builds for every `pxt bump`; one on master and one on the tag that's created. This will only upload strings on the former, the tagged release will skip that step.

I tested my `checkIfTaggedCommitAsync()` locally but can't really test the full scenario. Should be safe to merge and then test out.

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -356,6 +356,27 @@ function semverCmp(a: string, b: string) {
     return parse(a) - parse(b)
 }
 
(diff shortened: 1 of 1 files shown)


riknoll:
@pelikhan I think it varies from platform to platform. This is more consistent

PR #6707, opened 2020-03-12
Fix the upload logic for crowdin

riknoll:
I think I got very confused about the curly braces the first time around. Now it will *actually* only upload on tagged commits on master.


riknoll:
To be clear: this applies to `strings.json`, `webstrings.json`, and the docs in pxt

Issue #6715, opened 2020-03-17
Online Learning blog post


Jaqster:
Can we post a more readable streaming video schedule?
Who knows Markdown well?  And can create some nice looking tables...
Here's the current schedule - https://1drv.ms/x/s!AmMIW5Hxi0RtgZp1V-H3ZIH0ldOCtQ?e=OIaWMv

PR #6717, opened 2020-03-17
table for online learning post

jwunderl:
close https://github.com/microsoft/pxt/issues/6715

~noticed a few minor styling bugs (tables go offscreen in phone size with a different background color)~ fixed in separate pr ~and that hint headers aren't showing up (should be able to just have the header after the hint and have it show up properly next to the icon); will take a look at those real quick (makecode.com/docs files update either immediately with merge or after a bump, so no big release needed)~ fixed in this pr 

Also wrote up short descriptions for each stream, if we could double check those :)

(sorry for huge / 8mb gif, using desktop and haven't nailed gif recorder settings yet)

(see https://github.com/microsoft/pxt/pull/6717#issuecomment-600366559 below for new version with proper headers)

worth noting that view is slightly compressed, here it is at 'full size':

image

### DIFF
--- a/docfiles/style.css
+++ b/docfiles/style.css
@@ -219,17 +219,13 @@ svg {
     align-items: center;
 }
 
-#docs h1:before, #docs h2:before,
-#docs h3:before, #docs h4:before,
-#docs h5:before, #docs h6:before {
-    content: '';
-    display: block;
-    /* keep this aligned with top bar menu height */
-    height: 3.5em;
-    margin-top: -3.5em;
-    visibility: hidden;
+#docs h1, #docs h2,
+#docs h3, #docs h4,
+#docs h5, #docs h6 {
+    /* keep this aligned with top bar menu height;
+    add .5em to keep a small margin*/
+    scroll-margin-top: 4em;
 }
-
 
 /* first children are left & right div wrappers */
 .fixed.menu div {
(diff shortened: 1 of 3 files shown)


jwunderl:
oh one sec, found out how to make a heading for the hint macro but the css has been broken (subheading comes from having a level 4 header in the hint); will add the associated the needed css fix into this pr and then I think everything should be fine

jwunderl:
Turns out I had just broken it a few minutes before in https://github.com/microsoft/pxt/pull/6718 :) looked around for a bit for a better approach and found out there's actually a newer css property that will fix the spacing for you, and it's supported everywhere we support: https://caniuse.com/#feat=mdn-css_properties_scroll-padding-top so that's much better.

also cleaned up a bit of the duplication for the message boxes, as it was easier to do that then duplicate the fix everywhere. 

new look:

andfixed

we could probably find a better icon for streaming / video descriptions (and add another message box along with it) but that doesn't seem immediately necessary :)

pelikhan:
Can we avoid tables? They don’t work on mobile.

jwunderl:
I do generally agree that tables can be annoying, but as a second way to find what you're looking for it seems okay to me

table gives random access to who's streaming when / can be compared to a calendar well; scrolling through a list of times is harder to follow when things are happening as you have to read through all the options to see what fits in your schedule for the day instead of looking at the column for the day

Jaqster:
I love this!  The schedule in a table is so much easier to read!  Thanks so much Joey!  I don't think we should worry overmuch about mobile users Peli - likely people are going to be on a computer to go through this video tutorials.

PR #6718, opened 2020-03-17
make docs tables scrollable iff too wide

jwunderl:
re: table issue in https://github.com/microsoft/pxt/pull/6717

make tables scrollable if too wide to fit on screen; better than scrolling the entire page.

scrollabletables


pelikhan:
Tables are terrible on mobile.

PR #6731, opened 2020-03-20
First draft of online learning landing page

shakao:
online_learning

First pass based on @Jaqster's mockup, still need to hook up some links ("resources" will link to blog posts, updated stream links once we've moved to mixer) and swap out some of the images. We can also talk over design things at the next design meeting.

The schedule and project data is all stored in the typescript so hopefully it should be easy to modify if we have updates/changes.


Jaqster:
Wow!  Love it!  Looks amazing Shannon!!!!

pelikhan:
Woot. Let’s not wait a week on this. Once the fix are ok we push it and iterate.
What links need to be fixed?

pelikhan:
If you would store the program is some markdown format, we could probably localize this too.

shakao:
okay, everything should be linked up! for now all sections of "educator resources" go to the online learning blog post, and i put the aka.ms links for the streams--if we get permanent mixer urls we want to use we can swap those out later. 

for localization, could we use "lf" here? i can also look into moving it to json-y markdown like our gallery definitions

PR #6733, opened 2020-03-20
Online learning page footer placement, video links, timestamp in description

shakao:
image

image


pelikhan:
You don't need to wait for the build to finish on those as the build does not check anything.

PR #6735, opened 2020-03-20
Youtube card on examplers/tutorials

pelikhan:
When a tutorial/example card has a youTubeId entry, display an additional action card to start stream. This allows to surface videos "near" the tutorial.

- [x] detect that youtube is reachable: the card is not displayed if youtube is not reachable.
- [x] display additional action card with video stream

This change is backward compatible: it will be ignored by older makecode instances.

image


abchatra:
Can we not take hard dependency on youtube? Can we generalize for any kind of streaming service?

pelikhan:
We already have a hard dependency on YouTube unfortunately.

PR #6736, opened 2020-03-20
Return values for functions in blocks

riknoll:
Test upload is [here](https://arcade.makecode.com/app/65da5855de9e6c3f90bbd68e1e9a05d1114272ad-fec8cd5320).
Relies on https://github.com/microsoft/pxt-blockly/pull/272

Adds return values for functions!

Looks great, but can it do recursion? Sure can!

Screen Shot 2020-03-20 at 10 26 36 AM

Note that normally this does not emit return annotations for functions, but if you have a recursive function we emit a return type of `any`. This is just to stop TypeScript from complaining about "no implicit any". I chose not to emit any other return types in order to prevent our Blockly type checker from becoming even more complicated.

The build is going to fail until the pxt-blockly changes are merged and bumped.

Some things that can be improved:
1. Error messages are pretty bad when you do something that isn't TypeScript compatible
2. We should support any function argument type. Recursion works right now because all function output blocks have their type check set to "any"


pxtblocks/blocklyloader.ts
```diff
@@ -2378,6 +2378,46 @@ namespace pxt.blocks {
 
         msg.FUNCTIONS_CALL_TITLE = functionCall.block["FUNCTIONS_CALL_TITLE"];
         installBuiltinHelpInfo(functionCallId);
+        installBuiltinHelpInfo("function_call_output");
```
jwunderl: should probably be a const for `"function_call_output"` defined wherever `functionCallId` is?

pxtblocks/blocklycompiler.ts
```diff
@@ -253,6 +257,49 @@ namespace pxt.blocks {
...
+                    const unified = mkPoint(null);
+                    for (const point of returnTypes) {
+                        union(unified, point);
+                    }
+                    res = unified
+                }
+                catch (err) {
+                    e.diagnostics.push({
+                        blockId: definition.id,
+                        message: Util.lf("Function '{0}' has an invalid return type", name)
```
jwunderl: I suppose we don't necessarily have enough info here to give a full description of error ("Function can't return both a string and a number" sort of thing), but maybe something along the lines of `"Function '{0}' has an invalid return type; returning different types of values"` -- something to give a description of why it's invalid, in case it's not obvious to the user that `string|number` is making it invalid

pxtblocks/blocklyloader.ts
riknoll: It's only used here, so I skipped the variable declaration

PR #6743, opened 2020-03-23
generate proper name

pelikhan:
Fix for https://github.com/microsoft/pxt-microbit/issues/2649

Prepend "on_" to generate function that are arguments in toolbox.


pxtcompiler/emitter/service.ts
```diff
@@ -1755,6 +1755,8 @@ namespace ts.pxtc.service {
                     functionArgument = `(${functionSignature.parameters.map(p => p.name).join(', ')})`;
                 let n = fnName || "fn";
                 if (functionCount++ > 0) n += functionCount;
+                if (isArgument) // forever -> on_forever
```
riknoll: Should this check if it already starts with `on`? Otherwise won't `on_chat` in minecraft be `on_on_chat`?

pelikhan: added check

PR #6750, opened 2020-03-23
Foldtutorial. in stable5.30

pelikhan:
Allow to fold the simulator in arcade: blocks tend to be extremely long in arcade and we really need forking.
Test build: https://arcade.makecode.com/app/cfc70ba5eb587dfc0b3e0e5afe46cd6e3e10d6b6-6f860889b2
- [x] mac chrom
- [x] safari
- [x] mac edgium
- [x] ipad simulator in chrome
- [ ] chromebook (no hardware aviable)
- [ ] old edge: no machine available

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -3467,9 +3467,8 @@ export class ProjectView
         const useSerialEditor = pxt.appTarget.serial && !!pxt.appTarget.serial.useEditor;
 
         const showSideDoc = sideDocs && this.state.sideDocsLoadUrl && !this.state.sideDocsCollapsed;
-        const showCollapseButton = showEditorToolbar && !inHome && !inTutorial && !sandbox && !targetTheme.simCollapseInMenu && (!isHeadless || inDebugMode);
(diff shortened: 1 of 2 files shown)


abchatra:
I don't understand this change. Why is this in stable channel?

pelikhan:
This is to release this fix ASAP, while i am recording tutorials.

abchatra:
Why is this fix important? What is the original change? What is the testing matrix? Please test on all platforms before porting.

pelikhan:
This fix is important because i am recording videos and I constantly need to move the blocks in the small coding area.

abchatra:
We can take it up if it is causing issues with live streaming which is important to unblock. I would like to see the this tested in following:

1. iPad
2. Chromebook
3. Edge (old)
4. Mac Safari

Can you test these and let me know? If you point to stable I can test on iPad & Edge (old)

@shakao can you review carefully and see we don't regress anything?

pelikhan:
Added a test build in description

Issue #6749, opened 2020-03-23
Online Learning page


Jaqster:
Love the Online Learning page!!! 🥇 
Here are a few things to fix:

[ ] Include Mixer and YouTube channel link.  "All videos will be live streamed on [Mixer ](https://mixer.com/MakeCode) and available for viewing later in the [MakeCode YouTube Channel](https://youtube.com/microsoftmakecode)".

[ ] Fix broken links:
         https://aka.ms/makecodearcadestreamcommunity doesn't go anywhere - suggest just pointing to https://www.twitch.tv/mr_isaacs/
         https://aka.ms/makecodeadafruitstream doesn't go anywhere either - change to https://aka.ms/makecodecpxstream

[ ] I might also suggest adding links to the video titles in the descriptions 
![image](https://user-images.githubusercontent.com/22108954/77357196-ff600100-6d04-11ea-8728-743955db0c02.png)

[ ] In the Educator Resources section, can we make the hyperlinked text a different color?  Right now it's a bit hidden.

abchatra:
Fixed?

Jaqster:
[ ] Remove Steve Isaacs from the Schedule and Stream Description
![image](https://user-images.githubusercontent.com/22108954/77956286-7efe4a80-7286-11ea-9535-551d96211a87.png)
![image](https://user-images.githubusercontent.com/22108954/77956363-99382880-7286-11ea-8be8-f11ba21490bc.png)

[ ] Add sub-headers under Live Video Streaming section for "Schedule" and "Description"

Jaqster:
Using this same issue for another change request to add the Microsoft Stores "Digital All-Stars" stream.

[ ] Add a 2nd paragraph in the Live Video Streaming section: "Microsoft Stores is proud to host a MakeCode stream with all-star athletes passionate about technology and connecting with their communities.  This week, join Brandon McManus from the Denver Broncos as he helps you build your very own football arcade game with Microsoft MakeCode Arcade!"

[ ] Update the Mixer offline image to include the Digital All-Stars video series
![image](https://user-images.githubusercontent.com/22108954/78291733-c41aba80-74da-11ea-8994-f0d547a80dac.png)

[ ] Add a row for the Digital All-Stars stream to our schedule: use the link https://aka.ms/makecodeDASstream
![image](https://user-images.githubusercontent.com/22108954/78292389-ccbfc080-74db-11ea-8185-00da0c30edc2.png)

[...]

PR #6756, opened 2020-03-24
Alternate links in Online Learning page

ganicke:
Fixes for @Jaqster 's #6749.

- [x] Links for "Streaming" header
- [x] Alternate individual stream links
- [x] Make stream titles into links - someone needs to check that change.

Still need an on hover CSS class for the stream and resource text links.


ganicke:
@shakao fixed this in #6754

PR #6761, opened 2020-03-24
Python debugger

riknoll:
Adding support for Python files in the debugger. I have not tested multi-file python projects.

### DIFF
--- a/webapp/src/package.ts
+++ b/webapp/src/package.ts
@@ -45,7 +45,8 @@ export class File implements pxt.editor.IFile {
     }
 
     getFileNameWithExtension(ext: string) {
-        return this.getTextFileName().replace(this.getExtension(), ext);
+        const base = this.getTextFileName();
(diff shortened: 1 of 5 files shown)


webapp/src/package.ts
```diff
@@ -39,11 +39,15 @@ export class File implements pxt.editor.IFile {
...
     }
 
-    getTypeScriptName() {
+    getTextFileName() {
         if (this.epkg.isTopLevel()) return this.name
         else return "pxt_modules/" + this.epkg.getPkgId() + "/" + this.name
     }
 
+    getFileNameWithExtension(ext: string) {
+        return this.getTextFileName().replace(this.getExtension(), ext);
```
jwunderl:
```suggestion
        return this.getTextFileName().slice(0, -this.getExtension().length) + ext;
```

or something like that; this will fail on converting `grumpy.py` / `knots.ts`

riknoll: yep

PR #6766, opened 2020-03-26
View all button instead of > to get to script manager

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/2418 (well, not remembering home page, but that would be a feature that comes with identity)

image

Went with the primary color to start, as the link blue has issues against the background (it's mostly used against pure white backgrounds like the lang picker, where microbit's home page has a slight blue tint that would make it < 4.5); should be as accessible as the rest of the home page, and it feels like it matches the the projects that way to me.

changed from the sui grid layout for the headers to flexbox as grid wasn't working for small screen sizes (bumping out to new lines unnecessarily, etc) and flexbox handles that properly.

view all button

### DIFF
--- a/webapp/src/projects.tsx
+++ b/webapp/src/projects.tsx
@@ -190,7 +190,7 @@ export class Projects extends data.Component<ISettingsProps, ProjectsState> {
                             onClick={this.showScriptManager} onKeyDown={sui.fireClickOnEnter}>
                             {lf("My Projects")}
                             <span className="view-all-button">
-                                {lf("View all")}
+                                {lf("View All")}
                             </span>
                         </h2> : <h2 className="ui header">{lf("My Projects")}</h2>}
                     </div>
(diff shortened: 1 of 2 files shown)


webapp/src/projects.tsx
```diff
@@ -184,13 +184,13 @@ export class Projects extends data.Component<ISettingsProps, ProjectsState> {
...
-                        {targetTheme.scriptManager ? <h2 role="button" className="ui header myproject-header" title={lf("See all projects")} tabIndex={0}
+                <div className="ui heading">
+                    <div className="column " style={{ zIndex: 1 }}>
+                        {targetTheme.scriptManager ? <h2 role="button" className="ui header myproject-header" title={lf("View all projects")} tabIndex={0}
                             onClick={this.showScriptManager} onKeyDown={sui.fireClickOnEnter}>
                             {lf("My Projects")}
-                            <span className="ui grid-dialog-btn">
-                                <sui.Icon icon="angle right" />
+                            <span className="view-all-button">
+                                {lf("View all")}
```
abchatra: Can we make A in all capital (View All)?  Looks better with that.

PR #6767, opened 2020-03-26
make Code, Editor embeds match version

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/2154

the backend is serving effectively the same string with "---run" at the end as `pxt.webconfig.runUrl` in `serveAppAsync`, which is why that one was working

Potentially better fixed in the backend? Could just make the verprefix also return the tag / versioned release the same as relprefix does, instead of just leaving verprefix solely for 'v0' / 'v1' / etc -- this is the only place that one seems to be used. Figured fixing in frontend when it's easy enough will typically be preferred though

### DIFF
--- a/webapp/src/share.tsx
+++ b/webapp/src/share.tsx
@@ -326,9 +326,7 @@ export class ShareEditor extends data.Component<ShareEditorProps, ShareEditorSta
         if (!/\/$/.test(shareUrl)) shareUrl += '/';
         let rootUrl = pxt.appTarget.appTheme.embedUrl
         if (!/\/$/.test(rootUrl)) rootUrl += '/';
-        const verPrefix = pxt.webConfig.verprefix
-                || pxt.webConfig.relprefix?.replace(/^\//, "").replace(/---$/, "")
-                || '';
+        const verPrefix = pxt.webConfig.verprefix || '';
 
         if (header) {
             if (ready) {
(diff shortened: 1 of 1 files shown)


abchatra:
I don't think we should enable sharing in /beta. We don't users to share and use /beta payload.  So the request is to not append beta at all. 

/v0 & /v1 are fine  as they refer to legacy maintained editor.

riknoll:
@abchatra so there is no way to share beta projects? I disagree. With some of our targets we have a long time between releases

jwunderl:
Ah okay, I had thought it was a bug the other way when making it but easy enough either way. 

Might be worth considering using the semantic release tags (https://arcade.makecode.com/v0.15.17) for these embed views? for normal share links I can see not linking to /stable or /beta etc as it's the common use case, but when it's generating links to be embedded it could be nice to make sure those match the version they were created in (even in the live site) -- otherwise you might get a broken embed at a random date when there's an upgrade rule that can't be applied, or a perf improvement could break some behavior (for example, lots of games on consoles in the past have had trouble on emulators due to physics being based on frame rate, etc)

pelikhan:
Agreed with @abchatra: cloud should be smart and make the decision to about which version to use rendering the shared page. Not the client.

pelikhan:
This is something we should discuss in a meeting - not over a github issue.

jwunderl:
Sounds good, closing this one then~

abchatra:
>> @abchatra so there is no way to share beta projects? I disagree. With some of our targets we have a long time between releases

We shouldn't auto generate it. Users should still be able to go to 
arcade.makecode.com/beta/#xxxx-xxx-xx ... and access the pages.

abchatra:
Otherwise there will be tons of articles in the wild which point to /beta and will be broken time and again. /beta should be choice on the part of users.

pelikhan:
Maybe it is time to remove the embeds. They don’t seem to have much use this people don’t have blogs anymore.

jwunderl:
Could generate pages / links on gh pages for the code embed -- already have a better arcade embed from that anyways

pelikhan:
Right so far I have kept the generate share feature and release feature separate but we could do some kind of takeover. Something to think about.

PR #6774, opened 2020-03-27
Toolbox visibility: animate drawers, hover effect on flyout

shakao:
- Small animation on toolbox drawers when entering monaco
- Display handle on flyout hover to indicate draggable elements

toolbox-gliss

Build: https://makecode.microbit.org/app/dd487eba2ea31ef83f2e2d03e651a5bb200f24d2-1a20e6fcd0

### DIFF
--- a/webapp/src/monacoFlyout.tsx
+++ b/webapp/src/monacoFlyout.tsx
@@ -30,12 +30,14 @@ export interface MonacoFlyoutState {
     icon?: string;
     groups?: toolbox.GroupDefinition[];
     selectedBlock?: string;
+    hoverBlock?: string;
     hide?: boolean;
 }
 
 export class MonacoFlyout extends React.Component<MonacoFlyoutProps, MonacoFlyoutState> {
     protected dragging: boolean = false;
     protected dragInfo: BlockDragInfo;
+    protected lastSelected: string;
 
     constructor(props: MonacoFlyoutProps) {
         super(props);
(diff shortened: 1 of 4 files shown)


webapp/src/monacoFlyout.tsx
```diff
@@ -79,6 +81,15 @@ export class MonacoFlyout extends React.Component<MonacoFlyoutProps, MonacoFlyou
         return () => { this.setState({ selectedBlock: name != this.state.selectedBlock ? name : undefined }) };
     }
 
+    protected getBlockMouseOver = (name: string) => {
+        pxt.tickEvent("monaco.toolbox.itemmouseover", undefined, { interactiveConsent: true });
```
jwunderl: This tick should probably be in the returned fn? right now it sends the tick for every block in the flyout each time the flyout component re-renders - so opening flyout without hovering or hovering over any single snippet would send it for each.

Also, should this one count as interactiveConsent? I suppose it probably doesn't matter at the moment as I don't think we have a monaco `flyoutOnly` mode so they'd have to have clicked on the toolbox to get here, but I wouldn't think of hover as counting as consent generally

riknoll: Yeah, don't pass interactive consent here. It needs to be a click on an interactive element in the page or a link

pelikhan:
- [ ] feels a bit slow, can we try a bit faster?
- [ ] make sure to turn off in "light" mode

How do we test that this feature has any impact?

Jaqster:
I could set up some Teams calls for user testing (with/without the animation)

PR #6781, opened 2020-03-30
Update teachers.md

Isioviel:
Updated URLs for .gif files - they were pointing to an incorrect folder and therefore not showing up in the docs.


pelikhan:
Unfortunately, this is a bug with our cloud rendering engine.

PR #6785, opened 2020-03-30
turn on no shadowed variables lint rule

jwunderl:
starting as a draft as there are a bunch of minor changes / I'm starting to hit the big folders :)

the bug from https://github.com/microsoft/pxt/pull/6784 was from the data variable being clobbered by the value in the given function; when you start to get nested functions it can lead to VERY subtle bugs.

There's one case I've seen so far where this rule is definitely annoying is with implicit elses; i.e.

```typescript
function fn() {
  if (cond) {
    const a = "edge case a"  // error logged here
    // do some trivial work then bail out
    return a;
  }
  const a = "main case a"
  // do the rest of the work in this implicit else branch
  return a;
}
```

the rule gives an error on the first one because the `a` in the conditional does shadow the `a` outside of it, even if it's obvious looking at it that it's a valid case. That one's definitely annoying, but ¯\\_(ツ)_/¯

Also a minor annoying case with constructor parameter field declarations (don't know the proper name off hand, just look at the `pxteditor/localStorage.ts` change; `storageId` is a function in that file so the parameter name conflicts with that in the scope of the constructor)

### DIFF
--- a/docfiles/pxtweb/cookieCompliance.ts
+++ b/docfiles/pxtweb/cookieCompliance.ts
@@ -135,23 +135,23 @@ namespace pxt {
             }
         }
         export function report() {
-            let report = `performance report:\n`
+            let output = `performance report:\n`
             for (let [msg, time] of stats.milestones) {
                 let pretty = prettyStr(time)
-                report += `\t\t${msg} @ ${pretty}\n`
+                output += `\t\t${msg} @ ${pretty}\n`
             }
-            report += `\n`
+            output += `\n`
             for (let [msg, start, duration] of stats.durations) {
                 if (duration > 50) {
                     let pretty = prettyStr(duration)
-                    report += `\t\t${msg} took ~ ${pretty}`
+                    output += `\t\t${msg} took ~ ${pretty}`
                     if (duration > 1000) {
-                        report += ` (${prettyStr(start)} - ${prettyStr(start + duration)})`
+                        output += ` (${prettyStr(start)} - ${prettyStr(start + duration)})`
                     }
-                    report += `\n`
+                    output += `\n`
                 }
             }
-            console.log(report)
+            console.log(output)
             perfReportLogged = true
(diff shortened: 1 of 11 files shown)


jwunderl:
There are a handful of todos that I'll probably want to review tonight (either things that looked suspicious or things I wanted to double check), but for the most part it's a lot of F2 => finding a name to describe the variable and a few refactors / changes to dedup functions. Mainly just annoying because it touches so many files

cli/gitfs.ts
```diff
@@ -49,11 +49,11 @@ export async function uploadRefs(id: string, repoUrl: string): Promise<void> {
     let visited: SMap<boolean> = {};
     let toCheck: string[] = [];
 
-    await processCommit(id);
+    await processCommit(initialCommitId);
 
     await uploadMissingObjects(undefined, true);
 
-    await refreshRefs(id, repoUrl);
+    await refreshRefs(initialCommitId);
```
pelikhan, line 56: parameter gone?

jwunderl: I saw you resolved this conversation, but to clarify for anyone else the dropped parameter was just redundant with the enclosing fns parameter; did that in a few places where the same was true or the parameter was never used

pelikhan:
This is a massive PR. NEeds to be a team PR review.

pelikhan:
Unfortunately this PR is going to be outdated very soon. We should progressively fix this. Is there a way to do this by chunks?

jwunderl:
It'd probably be a pretty easy merge from master as strategy would most always be 'take remote and refix lint', but the size of the pr will always be annoying to deal with. I guess I can split into per folder prs, with tslint.json duplicated into each folder with the flag on, and then when all are fixed dedup tslints and turn the flag on in the main one?

pelikhan:
It should be a team effort to fix this.

PR #6787, opened 2020-03-31
Tutorialparsefix

pelikhan:
It turns out that markdown is merciful and eats white. spaces between back tick and the fence code session name eg. If the user makes this mistake our tutorial parsing completely breaks.

     ```      python
     x = 1
     ```
is valid.

Test both:
```python
while True:
    pass
```

with space
``` python
while True:
    pass
```

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -5346,7 +5346,7 @@ export interface SnippetInfo {
 
 export function getSnippets(source: string): SnippetInfo[] {
     let snippets: SnippetInfo[] = []
-    let re = /^`{3}\s*([\S]+)?\s*\n([\s\S]+?)\n`{3}\s*?$/gm;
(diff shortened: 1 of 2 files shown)


cli/cli.ts
```diff
@@ -5346,7 +5346,7 @@ export interface SnippetInfo {
 
 export function getSnippets(source: string): SnippetInfo[] {
     let snippets: SnippetInfo[] = []
-    let re = /^`{3}([\S]+)?\s*\n([\s\S]+?)\n`{3}\s*?$/gm;
+    let re = /^`{3}\s*([\S]+)?\s*\n([\s\S]+?)\n`{3}\s*?$/gm;
```
```suggestion
    let re = /^`{3}[^\S\r\n]*([\S]+)?\s*\n([\s\S]+?)\n`{3}\s*?$/gm;
```

re: last comment, this should be 'matching any whitespace besides newline / carriage return'

pelikhan: you could just match spaces and keep it simpler.

PR #6805, opened 2020-04-02
Make the function names for python snippets match the pydeompiler

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/2701

I wanted to unite the code paths entirely, but it's really tough to factor out the name hint function from the pydecompiler. Instead I tweaked both of them so that they match. The following changed:

1. Snippets from the toolbox now have enum arguments appended to the end
2. Snippets from the pydecompiler now always begin with `on_`

### DIFF
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -1624,9 +1624,10 @@ namespace ts.pxtc.service {
             return python ? "None" : "null";
         }
 
-        const args = decl.parameters ? decl.parameters
-            .filter(param => !param.initializer && !param.questionToken)
-            .map(getParameterDefault) : [];
+        const includedParameters = decl.parameters ? decl.parameters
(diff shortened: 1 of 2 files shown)


pxtcompiler/emitter/service.ts
```diff
@@ -1624,9 +1624,10 @@ namespace ts.pxtc.service {
             return python ? "None" : "null";
         }
 
-        const args = decl.parameters ? decl.parameters
-            .filter(param => !param.initializer && !param.questionToken)
-            .map(getParameterDefault) : [];
+        const includedParameters = decl.parameters ? decl.parameters
+        .filter(param => !param.initializer && !param.questionToken) : []
```
jwunderl, line 1628:
```suggestion
            .filter(param => !param.initializer && !param.questionToken) : []
```

riknoll: good nitpick!

PR #6807, opened 2020-04-02
Include a News div for new streams on OLP

ganicke:
Put in @Jaqster 's news div to announce the Digital All-Stars stream.

RE: #6804, #6749


pelikhan:
ready?

PR #6809, opened 2020-04-03
Move blockly keyboard accessibility to experiments, hook up to toolbox

shakao:
https://makecode.microbit.org/app/c2c77235b5a93533eb0d5c906d71f11751fae29a-6f11c68cc3


webapp/src/toolbox.tsx
```diff
@@ -518,31 +524,48 @@ export class CategoryItem extends data.Component<CategoryItemProps, CategoryItem
...
 
     handleKeyDown(e: React.KeyboardEvent<HTMLElement>) {
-        const { toolbox, childrenVisible } = this.props;
+        const { toolbox } = this.props;
         const isRtl = Util.isUserLanguageRtl();
 
+        const accessibleBlocksEnabled = (Blockly.getMainWorkspace() as any).keyboardAccessibilityMode;
+        const blocklyNavigationState = (Blockly.navigation as any).currentState_ as number;
+        const keyMap: { [key: string]: number } = {
+            "DOWN": accessibleBlocksEnabled ? 83 : 40, // 'S' || down arrow
```
jwunderl, line 533: So when accessibility is on, navigate toolbox with wasd, otherwise current behavior / arrow keys? I'm pretty sure this has been discussed before but was there a reason to not just handle both for the toolbox at all times? Dont think wasd would conflict with any other mappings while focus is in toolbox

shakao: i think it's confusing to have arrow keys work in the toolbox but stop working on the workspace. the two options i'd propose are: 1. "WASD" exclusively, when the experiment is on, so it's both a test of blockly's keyboard nav and of using "WASD" (vs arrow keys) 2. remap blockly to arrow keys so we're using arrow keys everywhere (sounds like this would break screen readers when blocks accessibility is on)

jwunderl: Sounds good, agree with option 1 then!

PR #6811, opened 2020-04-03
Remove leading whitespace in C++ files

mmoskal:
I'm hitting compile size limits now; this fixes in it for now.

Backend PR pending.


mmoskal:
https://github.com/microsoft/pxt/pull/6811

PR #6812, opened 2020-04-03
collapse all option

pelikhan:
In my arcade streams, the size of the blocks quickly becomes unbearable. A collapse all and format options just shrinks everything and make it easier to scale large programs.

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -1550,7 +1550,7 @@ namespace pxt.blocks {
 
             if (pxt.appTarget.appTheme.blocksCollapsing) {
                 const collapseAllOption = {
-                    text: lf("Collapse All"),
+                    text: lf("Collapse Blocks"),
                     enabled: true,
                     callback: () => {
                         pxt.tickEvent("blocks.context.collapse", undefined, { interactiveConsent: true });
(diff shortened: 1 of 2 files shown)


abchatra:
Do you have uncollapse all too?

pelikhan:
I can easily add it as well. I also want to be smarter about expanding. We should try to expand in a way that minimizes overlaps (eg expand and move up for example if there is space available)

pelikhan:
And expand now

pxtblocks/blocklyloader.ts
```diff
@@ -1548,6 +1548,30 @@ namespace pxt.blocks {
             }
             options.push(formatCodeOption);
 
+            if (pxt.appTarget.appTheme.blocksCollapsing) {
+                const collapseAllOption = {
+                    text: lf("Collapse All"),
```
abchatra, line 1553: Is "Collapse Blocks" & "Expand Blocks" better name for this instead of "All"?

pelikhan:
No preference, technically we collapse events, not all blocks so that might be misleading.

abchatra:
We say "Collapse Block" & "Expand Block" for top level blocks right? That is way I thought about it. I don't have strong preference either. @shakao

pelikhan:
Collapse All Blocks? Plural is too subtle.

pelikhan:
Like delete all blocks

abchatra:
Too big in my opinion. We have "Format Blocks"

abchatra:
Delete is a risky operation. So makes sense to be clear.

abchatra:
Upto you. Ignore my comment if you don't like it.

pelikhan:
Short is better right.

PR #6819, opened 2020-04-06
Rename experiment to Accessible Blocks, default to off, add info dialog

shakao:
- Rename experiment to 'Accessible Blocks' for consistency
- Change behavior so experiment enables settings menu item (Accessible Blocks On) which defaults to off
- Add information dialog when toggling accessible blocks on, to explain keyboard commands

Fixes https://github.com/microsoft/pxt-microbit/issues/2779

### DIFF
--- a/webapp/src/accessibleblocks.tsx
+++ b/webapp/src/accessibleblocks.tsx
@@ -41,7 +41,6 @@ export class AccessibleBlocksInfo extends data.Component<{}, AccessibleBlocksInf
                     {lf("See the official Blockly Documentation for additional detail")}
                 </a></strong>
             </div>
-            
         </sui.Modal>
     }
 }
--- a/webapp/src/accessibleblocks.tsx
+++ b/webapp/src/accessibleblocks.tsx
@@ -1,0 +1,47 @@
+import * as React from "react";
+import * as data from "./data";
+import * as sui from "./sui";
+
+export interface AccessibleBlocksInfoState {
+    shown?: boolean;
+}
+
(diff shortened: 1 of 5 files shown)


webapp/src/accessibleblocks.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+    render() {
+        const shown = this.state.shown;
+        if (shown) return <div />;
+        return <sui.Modal isOpen={!this.state.shown} onClose={this.handleOnClose} closeIcon={true}
+            dimmer={true} header={lf("Move, edit, and insert blocks using your keyboard")}>
+            <div className="ui equal width grid">
+                <div className="column">
+                    <h3>{lf("Navigate")}</h3>
+                    <span className="ui small">{lf("Use the WASD keys to move between blocks. Hold shift to move on the workspace.")}</span>
```
pelikhan, line 28: What is the story for non-US keyboards?

pelikhan: Could we make WASD a localized string and pick keys from there?

shakao: not handled for now--i think the hope is if they integrate screen reader support we can use the arrow keys for everything

i think the most flexible solution is probably to expose some kind of UI to let users remap the controls based on whatever keyboard they have plugged in, but i'm also hoping we can just switch to arrow keys haha

abchatra: This is experimentation feature, lets postpone this problem for now. It unrelated to this checkin

PR #6820, opened 2020-04-06
Debugger variable view improvements

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/2777
Fixes https://github.com/microsoft/pxt-microbit/issues/2738

This PR contains three changes

1. If the code is running, we now show a message instead of the variables from the last stack frame
2. If there are no variables visible at the current breakpoint, we now show a message instead of an empty toolbox
3. Exception messages are now displayed at the top of the variable pane when paused on an exception

And here's a GIF showing all of that off:

Probably needs a UI review

### DIFF
--- a/webapp/src/debuggerVariables.tsx
+++ b/webapp/src/debuggerVariables.tsx
@@ -86,8 +86,17 @@ export class DebuggerVariables extends data.Component<DebuggerVariablesProps, De
             variables = stackFrames[this.props.activeFrame].variables.concat(variables);
         }
 
-        return <DebuggerTable header={variableTableHeader} frozen={frozen}>
-            {this.renderVars(variables)}
+        let placeholderText: string;
+
+        if (frozen) {
+            placeholderText = lf("Code is running...");
+        }
+        else if (!variables.length) {
+            placeholderText = lf("No variables to show");
+        }
+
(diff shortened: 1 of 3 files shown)


webapp/src/debuggerVariables.tsx
```diff
@@ -77,17 +77,38 @@ export class DebuggerVariables extends data.Component<DebuggerVariablesProps, De
...
         const { globalFrame, stackFrames, frozen } = this.state;
+        const { activeFrame, breakpoint } = this.props;
         const variableTableHeader = lf("Variables");
         let variables = globalFrame.variables;
 
         // Add in the local variables.
         // TODO: Handle callstack
-        if (stackFrames && stackFrames.length && this.props.activeFrame !== undefined) {
-            variables = stackFrames[this.props.activeFrame].variables.concat(variables);
+        if (stackFrames && stackFrames.length && activeFrame !== undefined) {
```
jwunderl, line 86:
```suggestion
        if (stackFrames?.length && activeFrame !== undefined) {
```

pelikhan:
Seeing variables in slo-no would be wonderful for teaching purposes.

riknoll:
We could probably do that but it's not related to this PR

PR #6825, opened 2020-04-07
proper js parseint, and add string.trim

jwunderl:
fix for https://github.com/microsoft/pxt-microbit/issues/2664, though will also take a look at fixing ``String_::toNumber`` failing with anything but numbers and spaces but that's in common packages. Can move to cpp if we think that's needed (figured we needed the js anyway for simulator even if I need to move it to cpp)

I did add `string.prototype.trim` too, because I leading trim / ignore whitespace anyways


abchatra:
This needs test. Where did you get the algo for trim from? Can you comment the source?

docs/static/playground/pxt-common/pxt-helpers.js
```diff
@@ -402,6 +454,37 @@ namespace helpers {
         return q + r;
     }
 
+    export function stringTrim(s: string): string {
+        let start = 0;
+        let end = s.length - 1;
```
abchatra, line 459: check for s being undefined?

jwunderl:
can look at adding tests tomorrow, but we don't have a test runner for these builtins yet. Trim is just a quick implementation / not based off any source besides the one linked for listing out whitespace characters (any normal implementation would likely just be a regex match). parseInt is roughly based off https://www.ecma-international.org/ecma-262/5.1/#sec-15.1.2.2

docs/static/playground/pxt-common/pxt-helpers.js
jwunderl: can add that in case someone invokes `helper.stringTrim`, but the normal use case will fail with a null dereference before getting here (``null.trim()``)

mmoskal: yup; no need to check imho

abchatra: Makes sense. We don't have .apply otherwise you can get it as null

jwunderl:
added some tests in compile-test (and made those throw on error instead of just timing out so they show an error message) ~

PR #6826, opened 2020-04-08
Disable keyboard command for accessibility if experiment off, small c…

shakao:
…ollapse all fix

- Only allow ctrl+shift+k shortcut if accessibility experiment is turned on
- Collapse all should only collapse blocks that are enabled (event handlers)

### DIFF
--- a/pxtblocks/blocklylayout.ts
+++ b/pxtblocks/blocklylayout.ts
@@ -138,7 +138,7 @@ namespace pxt.blocks.layout {
 
     export function setCollapsedAll(ws: Blockly.WorkspaceSvg, collapsed: boolean) {
         ws.getTopBlocks(false)
-            .forEach(b => b.setCollapsed(collapsed));
+            .forEach(b => { if (b.isEnabled()) b.setCollapsed(collapsed) });
     }
 
     export function flow(ws: Blockly.WorkspaceSvg, opts?: FlowOptions) {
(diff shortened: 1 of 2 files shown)


webapp/src/blocks.tsx
```diff
@@ -410,6 +410,18 @@ export class Editor extends toolboxeditor.ToolboxEditor {
             document.querySelector("#blocksEditor").addEventListener("keydown", this.handleKeyDown)
         }
 
+        const self = this;
+        const onBlocklyAction = Blockly.navigation.onBlocklyAction;
+        Blockly.navigation.onBlocklyAction = function (action) {
+            if (pxt.appTarget.appTheme?.accessibleBlocks) {
+                if (!(Blockly.getMainWorkspace() as any).keyboardAccessibilityMode &&
```
riknoll, line 417: Instead of getMainWorkspace() you can use this.editor

pxtblocks/blocklylayout.ts
```diff
@@ -138,7 +138,7 @@ namespace pxt.blocks.layout {
 
     export function setCollapsedAll(ws: Blockly.WorkspaceSvg, collapsed: boolean) {
         ws.getTopBlocks(false)
-            .forEach(b => b.setCollapsed(collapsed));
+            .forEach(b => { if (b.isEnabled()) b.setCollapsed(collapsed) });
```
pelikhan, line 141: not use filter

PR #6828, opened 2020-04-08
Fill default values into blockdef for search, add param jsdoc

shakao:
The user reads the default parameter values when they scan the block in the toolbox so it makes sense to add them to the block string for search. Also adds a field for param jsdoc, but weighted much lighter.

Fixes https://github.com/microsoft/pxt-microbit/issues/2467

### DIFF
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -1209,20 +1209,32 @@ namespace ts.pxtc.service {
 
             // Fill default parameters in block string
             const computeBlockString = (symbol: SymbolInfo): string => {
-                let block = symbol.attributes.block;
-                if (symbol.attributes?.paramDefl) {
-                    const paramDefl = symbol.attributes.paramDefl;
-                    for (let param in paramDefl) {
-                        if (block.indexOf(`%${param}`) >= 0) {
-                            block = block.replace(`%${param}`, paramDefl[param]);
-                        } else if (!parseInt(paramDefl[param])) {
-                            // If default param name is not found, and value is not a number
-                            // append to the end of the block string
-                            block = block + " " + paramDefl[param];
+                if (symbol.attributes?.paramDefl && symbol.attributes?._def) {
+                    let block = [];
+                    const blockDef = symbol.attributes._def;
+                    const paramDefl = Object.assign({}, symbol.attributes.paramDefl);
+
+                    // Construct block string from parsed blockdef
+                    for (let part of blockDef.parts) {
+                        switch (part.kind) {
+                            case "label":
+                                block.push(part.text);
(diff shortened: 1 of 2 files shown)


pxtcompiler/emitter/service.ts
```diff
@@ -1207,6 +1207,33 @@ namespace ts.pxtc.service {
                 return Object.keys(tooltipOrBlock).map(k => (<pxt.Map<string>>tooltipOrBlock)[k]).join(" ");
             };
 
+            // Fill default parameters in block string
+            const computeBlockString = (symbol: SymbolInfo): string => {
+                let block = symbol.attributes.block;
+                if (symbol.attributes?.paramDefl) {
+                    const paramDefl = symbol.attributes.paramDefl;
+                    for (let param in paramDefl) {
+                        if (block.indexOf(`%${param}`) >= 0) {
```
riknoll, line 1216: Don't use the block string directly, use the parsed block definition instead. It's stored in _def and _expandedDef on the attributes object.

Also %, is the old way of defining block definitions. All future blocks should use $

riknoll: Any parameters using % are not guaranteed to match the name in paramDefl. That's part of the reason why we switched to the new syntax. In the old way you could put whatever name you wanted into the block string (it was purely based off index)

riknoll: Actually, you might have more luck using https://github.com/microsoft/pxt/blob/master/pxtlib/blocks.ts#L105 which resolves all the param names correctly

shakao: hmm i don't know if the full compile info is worth it, right now if i don't find the matching name i'm just appending the default to the end of the block string and the fuzzy search is generous enough that it mostly works. although i guess it would be good to keep all our "constructing string from blockdef" code in one place

pelikhan:
What about non-default?

shakao:
hmm i don't know, would it be confusing to show non-default members? you'll get search results that don't seem to match the search string at all, because the enum value is hidden in the dropdown

pelikhan:
Say I search for “tilt left” in microbit, what do I find?

shakao:
it makes sense if there's few enum members and the strings are "important" in the block, but it's more confusing when a dropdown has a LOT of options. if you were searching for "square root" and you typed "square" you could get "show icon [square]" instead. or in minecraft, if you typed "spawn" you would get all of the item blocks with the "spawn egg" items

maybe that would be solved if we just weight it pretty lightly though....

pelikhan:
It’s all about weights.

pelikhan:
Not high pri maybe in another fix

PR #6832, opened 2020-04-09
Use language service to fetch help info

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/2676
Fixes https://github.com/microsoft/pxt-microbit/issues/2797

Switching the help command in monaco to use the language service instead of constructing a help URL from whatever happens to be in the editor. Also adding the help URLs for the builtin math functions. @ganicke you might want to do a pass on the APIs in `libs/pxt-common`, I just added the ones I saw pages for.


riknoll:
@jwunderl yep!

PR #6837, opened 2020-04-10
Preserve undo stack when going from serial->blocks

shakao:
Preserves Blockly undo stack for "editors" that use the "Go Back" button (serial, github view, etc). 

Fixes https://github.com/microsoft/pxt-microbit/issues/697

### DIFF
--- a/pxteditor/editor.ts
+++ b/pxteditor/editor.ts
@@ -265,6 +265,7 @@ namespace pxt.editor {
 
         setBannerVisible(b: boolean): void;
         typecheckNow(): void;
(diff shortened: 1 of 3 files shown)


shakao:
yeah, i think ideally we should serialize the undo stack, and also these "overlays" (serial, pxt json, etc) should not unload the editor underneath them. i was looking into doing this with monaco too, but blockly's stack is easily accessible (and we are manually clearing it), and i think for monaco it would be a more manual process

https://github.com/microsoft/monaco-editor/issues/239
https://github.com/Microsoft/monaco-editor/issues/686

PR #6845, opened 2020-04-13
Show keywords in auto-complete

darzu:
before:

after:

keywords can now show up in auto-complete. There's still more work that could be done here, like filtering keywords that don't belong in expressions e.g. "while", but we're getting better.

Fixes: https://github.com/microsoft/pxt-microbit/issues/2806


darzu:
> LGTM. If we can figure out how get tests, we’ll be able to grind through this feature much faster.

Richard has some actually. Let me see if they'll work for these cases

pelikhan:
Too easy to fix 1 and break 3.

pelikhan:
Ready to merge?

pelikhan:
Is this ready to be merged?

darzu:
Shout out to @riknoll , this language service test framework is awesome!!

```
whi // while

; i // if

; func // function

; let foo: nu // number

; function bar(): st // string
```

PR #6852, opened 2020-04-14
Starts the ErrorList experiment and adds a basic div with errors to text editor

darzu:
This adds boiler plate for an error list. This error list is behind an extension flag that is off by default.
See:
https://github.com/microsoft/pxt-arcade/pull/1852
https://github.com/microsoft/pxt-microbit/pull/2817

This works by creating a new react component that renders as part of our monaco editor wrapper and this component subscribes to our monaco editor wrapper for changes to the error list.

Enabling the experiment:

Try it here: https://arcade.makecode.com/app/dde286c7a2fe6f9a8c02a3430bf3f15133e4c3bd-11eb869d7f

FYI: @hristoramirez 

TODO:
- [x] test Safari
- [x] test Edge
- [x] test iPad
- [x] test Firefox
- [x] test mobile


webapp/src/monaco.tsx
```diff
@@ -509,24 +515,41 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
+                <div id="monacoEditorRightArea" className="monacoEditorRightArea">
+                    <div id='monacoEditorInner'>
+                        <MonacoFlyout ref={this.handleFlyoutRef} fileType={this.fileType}
+                            blockIdMap={this.blockIdMap}
+                            moveFocusToParent={this.moveFocusToToolbox}
+                            insertSnippet={this.insertSnippet}
+                            setInsertionSnippet={this.setInsertionSnippet}
+                            parent={this.parent} />
+                    </div>
+                    {showErrorList ? <ErrorList onSizeChange={this.resize.bind(this)} listenToErrorChanges={this.listenToErrorChanges.bind(this)} /> : undefined}
```
darzu, line 534: Note that this "listenToErrorChanges" thing is a work around for the fact that the monaco editor wrapper isn't a react component. Otherwise the current errors could be part of its state and you could naturally trickle that down to the error list as props. Instead we do this event listener stuff.

webapp/src/errorList.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+        this.onCollapseClick = this.onCollapseClick.bind(this)
+        this.onErrorsChanged = this.onErrorsChanged.bind(this)
+
+        props.listenToErrorChanges("errorList", this.onErrorsChanged);
+    }
+
+    render() {
+        const showCollapseButton = true;
+        const collapseTooltip = "Collapse error list"
```
abchatra, line 33: Localizable tooltip?

PR #6853, opened 2020-04-14
Moving transpile cache into webapp and fixing snippet bug

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/2649

As we discussed offline, this PR removes the transpile cache from the language service and puts it into the webapp. Specifically, I'm now storing transpile results on the EditorPackage for the project.

There are two main differences between this and the old cache:

1. Transpile results are only cached when the user hits the toggle to switch languages. All other conversions are not cached
2. The transpile cache is only read from when the user hits the toggle to switch languages. All other conversions for compiling, syntaxInfo, snippets, etc. go through the language service

I left the roundtrip checks intact because they still serve a purpose if you reload the project in between conversions. The transpile cache is not saved with the project but that would be pretty easy to add if we wanted to.

### DIFF
--- a/webapp/src/compiler.ts
+++ b/webapp/src/compiler.ts
@@ -219,7 +219,6 @@ export function py2tsAsync(force = false): Promise<pxtc.transpile.TranspileResul
         .then(() => pkg.mainPkg.getCompileOptionsAsync(trg))
         .then(opts => {
             opts.target.preferredEditor = pxt.PYTHON_PROJECT_NAME
-            if (force) opts.forceTranspile = true;
 
             return workerOpAsync("py2ts", { options: opts })
         })
@@ -431,10 +430,16 @@ export function formatAsync(input: string, pos: number) {
 }
 
 export function snippetAsync(qName: string, python?: boolean): Promise<string> {
-    return workerOpAsync("snippet", {
+    let initStep = Promise.resolve();
+    if (python) {
+        // To make sure that the service is working with the most recent version of the file,
+        // run a typecheck
+        initStep = typecheckAsync().then(() => {})
+    }
+    return initStep.then(() => workerOpAsync("snippet", {
         snippet: { qName, python },
         runtime: pxt.appTarget.runtime
-    }).then(res => res as string)
+    })).then(res => res as string)
 }
 
 export function typecheckAsync() {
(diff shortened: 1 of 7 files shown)


webapp/src/compiler.ts
```diff
@@ -431,10 +430,16 @@ export function formatAsync(input: string, pos: number) {
 }
 
 export function snippetAsync(qName: string, python?: boolean): Promise<string> {
-    return workerOpAsync("snippet", {
+    let initStep = Promise.resolve();
+    if (python) {
+        // To make sure that the service is working with the most recent version of the file,
+        // run a typecheck
+        initStep = typecheckAsync().then(() => {})
```
darzu, line 437: This seems like it should be handled by the language service. It can itself recompile things if need be. Is there a reason it has to be done here?

darzu:
Looks great overall.

One regression concern I have is that the transpile cache was in some cases speeding up repeated compiles of the same code. We tend to be too aggressive triggering compiles even when the code hasn't changed. Can you verify with performance testing or logging that we're not eating up CPU cycles recompiling the same code multiple times?

webapp/src/compiler.ts
riknoll: we don't send the whole project to the language service with each request for snippets. i was seeing a bug where sometimes the version of the file in the language service did not match the version in the editor; calling a typecheck forces the webapp to send over all the files.

riknoll:
Do you have a test in mind? In my experience the decompiler, blockly compiler, and python converter are all pretty speedy.

webapp/src/compiler.ts
darzu: I see. It'd be great if there was an explicit "syncFiles", or even better would be if we had a version notion and when the webapp is calling out to the worker they could check last versions and ensure files are synced. But this is likely out of scope.

darzu:
> Do you have a test in mind? In my experience the decompiler, blockly compiler, and python converter are all pretty speedy.

Maybe some edits to tileworld with chrome CPU throttling enable?

riknoll:
Tileworld is typescript, so it wouldn't be affected by this.

darzu:
> Tileworld is typescript, so it wouldn't be affected by this.

Right. We don't have big python projects yet. I guess one of the larger minecraft examples is the best we have.

PR #6854, opened 2020-04-15
Localize types and variables in the function create dialog

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/2176

@pelikhan can you check and make sure I did this right?


cli/cli.ts
```diff
@@ -1798,7 +1812,7 @@ ${gcards.map(gcard => `[${gcard.name}](${gcard.url})`).join(',\n')}
             .forEach(f => processLf(f, targetStrings))
         );
     let targetStringsSorted: pxt.Map<string> = {};
-    Object.keys(targetStrings).sort().map(k => targetStringsSorted[k] = k);
+    Object.keys(targetStrings).sort().map(k => targetStringsSorted[k] = targetStrings[k]);
```
pelikhan, line 1815: Oh.... this looks like a good fix

PR #6855, opened 2020-04-15
Fix JS intellisense for global variables

darzu:
First batch of JS intellisense fixes

Before:

After:


webapp/public/vs/editor/editor.main.css
```diff
@@ -3,4 +3,4 @@
...
...o-editor .margin-view-overlays .codicon-chevron-down,.monaco-editor .margin-view-overlays .codicon-chevron-right{cursor:pointer;opacity:0;transition:opacity .5s;display:flex;align-items:center;justify-content:center;font-size:140%;margin-left:2px}.monaco-editor .margin-view-overlays .codicon.alwaysShowFoldIcons,.monaco-editor .margin-view-overlays .codicon.codicon-chevron-right,.monaco-editor .margin-view-overlays:hover .codicon{opacity:1}.monaco-editor .inline-folded:after{color:grey;margin:.1em .2em 0;content:"⋯";display:inline;line-height:1em;cursor:pointer}.monaco-editor .peekview-widget .head .peekview-title .severity-icon{display:inline-block;vertical-align:text-top;margin-right:4px}.monaco-editor .marker-widget{text-overflow:ellipsis;white-space:nowrap}.monaco-editor .marker-widget>.stale{opacity:.6;font-style:italic}.monaco-editor .marker-widget .title{display:inline-block;padding-right:5px}.monaco-editor .marker-widget .descriptioncontainer{position:absolute;white-space:pre;user-select:text;-webkit-user-select:text;-ms-user-select:text;padding:8px 12px 0 20px}.monaco-editor .marker-widget .descriptioncontainer .message{display:flex;flex-direction:column}.monaco-editor .mar
```
riknoll, line 6: I'm guessing this is unintentional?

pxtcompiler/emitter/driver.ts
```diff
@@ -271,9 +271,7 @@ namespace ts.pxtc {
         res.times["typescript-sem"] = emitStart - semStart
         res.times["typescript"] = emitStart - startTime
 
-        if (opts.ast) {
```
riknoll, line 274: Why are you removing this option? The AST is pretty big and doesn't need to be sent between the webapp and the webworker

webapp/public/vs/editor/editor.main.css
darzu: yes. Weird

pxtcompiler/emitter/driver.ts
darzu: Thanks, forgot about transfer size.
We couple "opts.ast" to a few things and when talking within the worker than it always makes sense to pass back a reference of the ast but opts.ast also invalidates incremental compile.

I'll decouple this option.

darzu: actually I just put the condition back in.

PR #6862, opened 2020-04-16
Improve auto-complete in JS for local variables

darzu:
We now use the typescript language service to get symbols in scope and surface those in auto-complete.

Before, PY had an advantage b/c the py converter kept track of more local info.

Before:

After:

### DIFF
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -1051,8 +1051,32 @@ namespace ts.pxtc.service {
             }
 
             if (!isPython && !resultSymbols.length) {
-                // TODO: leverage TS language service to get symbols in scope
+                // TODO: share this with the "syntaxinfo" service
+                // start with global api symbols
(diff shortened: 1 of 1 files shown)


pxtcompiler/emitter/service.ts
```diff
@@ -1855,6 +1880,61 @@ namespace ts.pxtc.service {
...
+                paramDefl: {},
+            },
+            fileName: "main.ts",
+            parameters: [],
+            retType: typeName,
+        }
+        return sym;
+    }
+
+    function getPxtSymbolFromTsSymbol(tsSym: ts.Symbol, apiInfo: ApisInfo, tc: TypeChecker): SymbolInfo | undefined {
```
pelikhan, line 1931: any reason why some function do ``| undefined`` and others ``| null``?

darzu: No, they should all use `| undefined`, it's generally better practice and aligns better with the optional type `foo?: number` etc.

PR #6866, opened 2020-04-16
Added line and column numbers for error messages

hristoramirez:
Added error messages with line and column location 
pullrequest-line_column


webapp/src/errorList.tsx
```diff
@@ -44,7 +44,7 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
             <div className="errorListInner" hidden={this.state.isCollapsed}>
                 {
                     (this.state.errors || []).map(e =>
-                        <div key={errorKey(e)}>{e.messageText}</div>)
+                        <div key={errorKey(e)}>{`${e.messageText} - [line ${e.line + 1}: col ${e.column}]`}</div>)
```
darzu, line 47: Looks great! Nice compact change

darzu: This is a good start and we can always tweak how things appear as we go

pelikhan: Unless we plan to translate line and col, you should remove them.

jwunderl: should this display `e.column + 1` like it does for for `e.line`? Just checked vs code and it gives both line and column indexed from 1 ~

darzu:
> It would be good to format the errors in output.txt with the same format (future PR)

Good point, I'll file an item

PR #6869, opened 2020-04-17
Better autocomplete on "this" keyword

darzu:
Fixes: https://github.com/microsoft/pxt-microbit/issues/2822

Before:

After:


cli/cli.ts
```diff
@@ -3031,11 +3031,11 @@ export function serviceAsync(parsed: commandParser.ParsedCommand) {
     return mainPkg.getCompileOptionsAsync()
         .then(opts => {
             pxtc.service.performOperation("reset", {})
-            pxtc.service.performOperation("setOpts", { options: opts })
-            return pxtc.service.performOperation(parsed.args[0], {})
+            pxtc.service.performOperation("setOptions", { options: opts })
```
darzu, line 3034: actual bug caught be the better type checking

webapp/src/compiler.ts
```diff
@@ -360,7 +360,7 @@ function pyDecompileCoreAsync(opts: pxtc.CompileOptions, fileName: string): Prom
     return workerOpAsync("pydecompile", { options: opts, fileName: fileName })
 }
 
-export function workerOpAsync(op: string, arg: pxtc.service.OpArg) {
+export function workerOpAsync<T extends keyof pxtc.service.ServiceOps>(op: T, arg: pxtc.service.OpArg): Promise<any> {
```
riknoll, line 363: should the return type be `Promise< OpResOrError >`?

pxtcompiler/emitter/service.ts
```diff
@@ -796,7 +787,48 @@ namespace ts.pxtc.service {
...
+        };
+        apiInfo: () => ApisInfo;
+        snippet: (v: OpArg) => string;
+        blocksInfo: (v: OpArg) => BlocksInfo;
+        apiSearch: (v: OpArg) => SearchInfo[];
+        projectSearch: (v: OpArg) => ProjectSearchInfo[];
+        projectSearchClear: () => void;
+    };
+
+    export type OpRes = string | void | SyntaxInfo | CompletionInfo | CompileResult | transpile.TranspileResult | {
```
riknoll, line 816: This is crazy, maybe we should just move to something we can pattern match on? Like we do for simulator messages where each result has a type string?

pxtcompiler/emitter/service.ts
```diff
@@ -955,29 +988,47 @@ namespace ts.pxtc.service {
...
-                    let type: Type;
-
                     const symbol = tc.getSymbolAtLocation(propertyAccessTarget);
-                    if (symbol) {
-                        type = tc.getTypeOfSymbolAtLocation(symbol, propertyAccessTarget);
-                    }
-                    else {
-                        type = tc.getTypeAtLocation(propertyAccessTarget);
-                    }
+                    if (symbol?.members?.size > 0) {
```
riknoll, line 992: Instead of separate handling for `this`, could you just call `tc.getPropertiesOfType()` on the type object? You might have to check to see if that works on `this` types but it would unify the two clauses

pxtcompiler/emitter/service.ts
riknoll: At the very least the inline types should get names

webapp/src/compiler.ts
darzu: Yes, but it would require more refactoring and I've already done a fair bit in this PR. This type is more precise than it was and we can improve again.

pxtcompiler/emitter/service.ts
darzu: Yes, I agree. I hesitated to do this as it'd cause more refactoring and I was trying not to pull on too many strings, but as you point out this is confusing and I'll improve this.

pxtcompiler/emitter/service.ts
darzu: Yeah from what I could tell this didn't really work with "this". If you get the symbol for a location with text "this" it returns the type symbol, which is strange, I would have expect a "this" symbol and then checking the type at that location would get the type.

I can take another look, maybe I can unify handling of "this".

darzu: Figured it out. The missing piece was you need to use "getTypeOfSymbolAtLocation" to get the type of symbols for instances of "this", whereas in all other cases you use "getTypeOfSymbolAtLocation".

I still seems a bit odd to me, like they skipped a step. The symbol for "this" should be "this" and getting the type of that symbol at that location should be "Foo". But whatever.

PR #6874, opened 2020-04-19
display state while tracing

pelikhan:
Always send a full "breakpoint" info while tracing. This allows to see the state changing while running the code. I merged the TraceMessage into a breakpoint message (different subtype), there may be cleaner types to do this.


abchatra:
@riknoll

Issue #6873, opened 2020-04-19
compiler bug (regression in pxt)


tballmsft:
The following program gives an error message in makecode.microbit.org:

error: main.ts(8,9): error TS9210: Cannot compute enum value

for the last line (Bar.Delete)

```
enum Foo {
    First,  
    Last,
}

enum Bar{ Last=Foo.Last, Delete };

let x = Bar.Delete;
```

However, in beta, we get a compiler failure on the same line

error: main.ts(8,5): error TS9200: Cannot read property 'pxt' of undefined

tballmsft:
Here is the code that created the error message (master). Looks like failure is probably happening in constantFoldDecl

```
        function emitPropertyAccess(node: PropertyAccessExpression): ir.Expr {
            let decl = getDecl(node);

            const fold = constantFoldDecl(decl)
            if (fold)
                return emitLit(fold.val)

            if (decl.kind == SK.SetAccessor)
                decl = checkGetter(decl)

            if (decl.kind == SK.GetAccessor)
                return emitCallCore(node, node, [], null, decl as GetAccessorDeclaration)

            if (decl.kind == SK.EnumMember) {
                throw userError(9210, lf("Cannot compute enum value"))
```

tballmsft:
More debugging. en.initializer is undefined because of Last=Foo.Last:

```
      function constantFoldDecl(decl: Declaration) {
            if (!decl)
                return null

            const info = pxtInfo(decl)
            if (info.constantFolded)
                return info.constantFolded

            if (isVar(decl) && (decl.parent.flags & NodeFlags.Const)) {
                const vardecl = decl as VariableDeclaration
                if (vardecl.initializer)
                    info.constantFolded = constantFold(vardecl.initializer)
            } else if (decl.kind == SK.EnumMember) {
                const en = decl as EnumMember
                const ev = enumValue(en);
                if (ev == null) {
                    console.log(en.initializer);
                    info.constantFolded = constantFold(en.initializer)
```

tballmsft:
As, a result, we call pxtNode with n==undefined, leading to the dereferencing error:

```
    export function pxtInfo(n: Node) {
        if (!n.pxt) {
```

abchatra:
Fixed

PR #6875, opened 2020-04-19
proposed fix for enum bug

tballmsft:
here is a proposed fix for issue https://github.com/microsoft/pxt/issues/6873. Results in the following error message:

main.ts(8,9): error TS9210: Cannot compute enum value


tballmsft:
added regression test

mmoskal:
Thanks, good investigation!

I tweaked the fix to move the `null` check down, should catch more of these.

PR #6881, opened 2020-04-20
Try to load favicon image from JS if in Edge

shakao:
Fixes https://github.com/microsoft/pxt-microbit/issues/2693

i still suspect the underlying issue has to do with how old edge handles mime types but that's all in the backend and i didn't want to change it for this.


pelikhan:
This is the location where we try to load favicon.ico

https://github.com/microsoft/pxt/blob/ed6cd6533840b31102a9ea6123852b54e49dc859/webapp/src/cloudsync.ts#L767

PR #6886, opened 2020-04-21
setstate should delete element if value is null (same as sim)

tballmsft:
In the simulator code in runtime.ts, we see that setting a key to null value has the effect of deleting this key:

```
        public setStoredState(k: string, value: any) {
            if (value == null)
                delete this.storedState[k]
            else
                this.storedState[k] = value
            Runtime.postMessage({
                type: "simulator",
                command: "setstate",
                stateKey: k,
                stateValue: value
            } as SimulatorCommandMessage)
        }
```

However, the setstate handler in gh-pages-sim.js doesn't contain a delete.


tballmsft:
Peli, how do I test this change?

pelikhan:
You'll need to fix it here too for github: https://github.com/microsoft/pxt-arcade/blob/master/libs/template/assets/js/loader.js#L86

pelikhan:
I don't think this file is used anymore. We can probably get rid of it.

tballmsft:
There are two files that are not referenced by scripts anymore. Can we delete both these?

PR #6888, opened 2020-04-22
added a simple message when there are no errors

hristoramirez:
Displays a message when there are no errors. Basic functionality. 
fixes: microsoft/pxt-microbit#2832
noErrors

### DIFF
--- a/theme/errorList.less
+++ b/theme/errorList.less
@@ -17,7 +17,7 @@
     }
 
     .errorListInner {
-        height: 200px;
+        height: 200rem;
         border-top: 1px #e5e5e5 solid;
         padding: 1em;
     }
--- a/theme/errorList.less
+++ b/theme/errorList.less
@@ -21,4 +21,8 @@
         border-top: 1px #e5e5e5 solid;
         padding: 1em;
     }
+
+    #noErrorsMessage {
+        height: 50px;
+    }
 }
--- a/theme/errorList.less
+++ b/theme/errorList.less
@@ -22,7 +22,7 @@
         padding: 1em;
     }
 
-    #noErrorsMessage {
-        height: 50px;
+    .noErrorsMessage {
+        height: unset;
     }
 }
--- a/theme/errorList.less
+++ b/theme/errorList.less
@@ -17,7 +17,7 @@
     }
 
     .errorListInner {
-        height: 200rem;
+        height: 200px;
         border-top: 1px #e5e5e5 solid;
         padding: 1em;
     }
--- a/webapp/src/errorList.tsx
+++ b/webapp/src/errorList.tsx
@@ -30,7 +30,7 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
 
     render() {
         const {isCollapsed, errors} = this.state;
-        const errorsAvailable = errors.length != 0;
+        const errorsAvailable = !!errors?.length;
         const collapseTooltip = lf("Collapse Error List");
         function errorKey(error: pxtc.KsDiagnostic): string {
             // React likes have a "key" for each element so that it can smartly only
--- a/webapp/src/errorList.tsx
+++ b/webapp/src/errorList.tsx
@@ -56,7 +56,8 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
     }
 
     componentDidUpdate() {
-        // ensures that the shrinking of no errors takes place
+        // notify parent on possible size change so siblings (monaco)
+        // can resize if needed
         this.props.onSizeChange()
     }
 
--- a/webapp/src/errorList.tsx
+++ b/webapp/src/errorList.tsx
@@ -29,42 +29,40 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
     }
 
     render() {
-        const showAvailableErrors = this.state.errors.length != 0;
-        const showCollapseButton = true;
+        const {isCollapsed, errors} = this.state;
+        const errorsAvailable = errors.length != 0;
         const collapseTooltip = lf("Collapse Error List");
         function errorKey(error: pxtc.KsDiagnostic): string {
             // React likes have a "key" for each element so that it can smartly only
             // re-render what changes. Think of it like a hashcode/
             return `${error.messageText}-${error.fileName}-${error.line}-${error.column}`
         }
-            
-        return <div className="errorList" > 
-            {showCollapseButton && showAvailableErrors &&
-                <sui.Button id='toggleErrorList' className={`toggleErrorList collapse-button large`}
-                icon={`inverted chevron ${this.state.isCollapsed ? 'up' : 'down'}`}
-                title={collapseTooltip} onClick={this.onCollapseClick} />}
-            {showAvailableErrors
-                ? <div className="errorListInner" hidden={this.state.isCollapsed}>
-                    {
-                    (this.state.errors || []).map(e =>
-                        <div key={errorKey(e)}>{`${e.messageText} - [line ${e.line + 1}: col ${e.column}]`}</div>)
-                    }
-                  </div>
-                : <div className="errorListInner" id="noErrorsMessage">{lf("You have no errors")}</div>
-            }
+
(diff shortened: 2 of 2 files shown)


webapp/src/errorList.tsx
```diff
@@ -29,27 +29,37 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
     }
 
     render() {
+        const showAvailableErrors = this.state.errors.length != 0;
         const showCollapseButton = true;
```
darzu, line 33: Instead of "showCollapseButton" being always true, you could make it conditional on available errors.

webapp/src/errorList.tsx
```diff
@@ -29,27 +29,37 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
...
+            {showAvailableErrors
+                ? <div className="errorListInner" hidden={this.state.isCollapsed}>
+                    {
                     (this.state.errors || []).map(e =>
                         <div key={errorKey(e)}>{`${e.messageText} - [line ${e.line + 1}: col ${e.column}]`}</div>)
-                }
-            </div>
+                    }
+                  </div>
+                : <div className="errorListInner" id="noErrorsMessage">{lf("You have no errors")}</div>
```
darzu, line 53: typically instead of adding an id to a div dynamically you would just add a class. The "className" can be a list of space seperated names

webapp/src/errorList.tsx
```diff
@@ -29,27 +29,37 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
...
+            }
         </div>
     }
 
+    componentDidUpdate() {
+        // ensures that the shrinking of no errors takes 
+        this.props.onSizeChange()
```
darzu, line 60: since you added this line, you can remove the onSizeChange call in line 67 because it is redundant.

webapp/src/errorList.tsx
```diff
@@ -29,27 +29,37 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
...
-            </div>
+                    }
+                  </div>
+                : <div className="errorListInner" id="noErrorsMessage">{lf("You have no errors")}</div>
+            }
         </div>
     }
 
+    componentDidUpdate() {
+        // ensures that the shrinking of no errors takes
```
darzu, line 59: Incomplete comment?

webapp/src/errorList.tsx
```diff
@@ -29,27 +29,37 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
...
+        return <div className="errorList" > 
+            {showCollapseButton && showAvailableErrors &&
                 <sui.Button id='toggleErrorList' className={`toggleErrorList collapse-button large`}
-                    icon={`inverted chevron ${this.state.isCollapsed ? 'up' : 'down'}`}
-                    title={collapseTooltip} onClick={this.onCollapseClick} />}
-            <div className="errorListInner" hidden={this.state.isCollapsed}>
-                {
+                icon={`inverted chevron ${this.state.isCollapsed ? 'up' : 'down'}`}
+                title={collapseTooltip} onClick={this.onCollapseClick} />}
+            {showAvailableErrors
```
darzu, line 46: You probably don't need to branch here. Instead, you could just add a `|| lf("You have no errors")` to the  end of the map.

Also if things seem to get too messy feel free to break stuff out into functions or methods. Regular functions can return JSX/react syntax in .tsx files.

theme/errorList.less
```diff
@@ -21,4 +21,8 @@
         border-top: 1px #e5e5e5 solid;
         padding: 1em;
     }
-}
+
+    #noErrorsMessage {
+        height: 50px;
```
darzu, line 26: Is this necessary? Instead would it be possible to adjust the line height or padding on the parent such that you get a good height when there's only one line of text?

webapp/src/errorList.tsx
darzu: Usually we think of an id as a constant. A HTML element either has one or it doesn't but it doesn't usually change.

darzu:
Notice that the Travis CI builds failed. If you look at the logs, it looks like the "lint" step failed. You can run "gulp lint" locally to see what errors are produced.

(fyi I recommend: https://stackoverflow.com/a/30884298/814454)

darzu:
I'm thinking about alternate wording a bit..
"Your program appears error free. If MakeCode detects an error, they will be shown here".

This isn't blocking, as I'm sure we'll get to this in a design meeting.

darzu:
I'd recommend adding the text "fixes: https://github.com/microsoft/pxt-microbit/issues/2832" to your description. That will automatically link this to the appropriate issue and once this PR completes that issue will be closed.

pelikhan:
We should discuss the wording of the error list at the design meeting.

webapp/src/errorList.tsx
```diff
@@ -29,27 +29,37 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
     }
 
     render() {
+        const showAvailableErrors = this.state.errors.length != 0;
```
pelikhan, line 32: be defensive and test that errors is defined. In general, we extract the state in const variables:
```
const { errors } = this.state;
```

webapp/src/errorList.tsx
```diff
@@ -29,27 +29,37 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
...
-            {showCollapseButton &&
+            
+        return <div className="errorList" > 
+            {showCollapseButton && showAvailableErrors &&
                 <sui.Button id='toggleErrorList' className={`toggleErrorList collapse-button large`}
-                    icon={`inverted chevron ${this.state.isCollapsed ? 'up' : 'down'}`}
-                    title={collapseTooltip} onClick={this.onCollapseClick} />}
-            <div className="errorListInner" hidden={this.state.isCollapsed}>
-                {
+                icon={`inverted chevron ${this.state.isCollapsed ? 'up' : 'down'}`}
```
pelikhan, line 44: extract ``isCollapsed`` in a const variable at the start of the function, it helps with readability

theme/errorList.less
darzu: Actually, it is necessary since there's a fixed height for when there are errors

shakao: also, preferable to use rem instead of px

webapp/src/errorList.tsx
```diff
@@ -29,32 +29,40 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
     }
 
     render() {
-        const showCollapseButton = true;
+        const {isCollapsed, errors} = this.state;
+        const errorsAvailable = errors.length != 0;
```
pelikhan, line 33: Is error guaranteed to be defined? Assume it’s not (being defensive is good).

    const errorsAvailable = !!errors?.length

webapp/src/errorList.tsx
```diff
@@ -29,32 +29,40 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
...
+                    ? (errors).map(e =>
+                        <div key={errorKey(e)}>{`${e.messageText} - [line ${e.line + 1}: col ${e.column + 1}]`}</div>)
+                    : <div>{lf("You have no errors")}</div>
                 }
             </div>
         </div>
     }
 
+    componentDidUpdate() {
+        // ensures that the shrinking of no errors takes place
```
darzu, line 59: More accurate: notify parent on possible size change so that siblings (e.g. monaco) can resize themselves if needed

Technically the error list itself won't have trouble shrinking or resizing, it's the sibling elements that have trouble

webapp/src/errorList.tsx
hristoramirez: Ok just tweaked that

theme/errorList.less
```diff
@@ -17,8 +17,12 @@
     }
 
     .errorListInner {
-        height: 200px;
+        height: 200rem;
```
jwunderl, line 20: 200rem === 3200px (1 rem == fontsize of root element); probably not quite what you want :) (agreed that it looks great otherwise though!)

PR #6889, opened 2020-04-22
Improve tilemap editor wall fill (microsoft/pxt-arcade#1819).

jacobcarpenter:
For fun (and because it's bugged me too!), I decided to implement what I think https://github.com/microsoft/pxt-arcade/issues/1819 is suggesting:

The fill tool in today's tilemap editor with wall mode selected isn't very useful. It either does nothing if you click on an existing wall tile, or it fills the entire region of connected non-wall locations. This PR updates the wall fill behavior to look at the tiles rather than the walls to determine which walls to set.

I'd love to hear any feedback you have on this PR, regardless of whether or not you decide to merge. Thanks!

### DIFF
--- a/webapp/src/components/ImageEditor/toolDefinitions.ts
+++ b/webapp/src/components/ImageEditor/toolDefinitions.ts
@@ -95,6 +95,8 @@ export function getEdit(tool: ImageEditorTool, state: EditState, color: number, 
             return new OutlineEdit(state.width, state.height, color, width);
         case ImageEditorTool.Fill:
             return new FillEdit(state.width, state.height, color, width);
+        case ImageEditorTool.WallFill:
+            return new WallFillEdit(state.width, state.height, color, width);
         case ImageEditorTool.Line:
             return new LineEdit(state.width, state.height, color, width);
         case ImageEditorTool.Marquee:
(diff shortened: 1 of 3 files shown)


webapp/src/components/ImageEditor/toolDefinitions.ts
```diff
@@ -659,6 +666,20 @@ export class FillEdit extends Edit {
 }
 
 
+export class WallFillEdit extends FillEdit {
```
riknoll, line 669: Instead of creating a new tool, I think you could just replace the existing fill behavior with this. Always fill the active layer but base it off of the tile layer; should work for both cases!

You might need to expose the activeLayer on the tilestate to get around the `if (colorToReplace === this.color)` check for walls

riknoll:
Also, looks like the linting check failed on travis. You can run `gulp lint` to see the report locally. I also recommend turning on the "trimTrailingWhitespace" setting in vscode

jacobcarpenter:
Thanks; I use Prettier generally, but that was reformatting _all_ the code on save, so I turned it off. 😊 I'll get that trailing whitespace cleaned up.

webapp/src/components/ImageEditor/toolDefinitions.ts
jacobcarpenter: Nice. Sounds good.

There is a small optimization in Fill Edit's doEditCore that skips the work if the target tile is the same as the color you're filling. This doesn't make sense when you're operating on different layers, but I'm sure I can figure something reasonable out.

pelikhan:
We use the "stock formatter" in VSCode. We're not cool enough to use prettier.

PR #6901, opened 2020-04-23
Provide block aliases for javascript functions that return values

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/2651
Fixes https://github.com/microsoft/pxt-microbit/issues/2082

This PR adds new blocks for the following APIs:

```ts
Array.pop()
Array.shift()
Array.unshift()
Array.removeAt()
```

The common thread between all of those APIs is that they all have return values that you usually don't care about. The new comment attribute lets you define an alias block for those functions that does not return anything and can be used as a statement.

The old blocks still exist, but I hid them from the toolbox. Existing projects will be unaffected


riknoll:
@abchatra you had comments on this?

abchatra:
Yes give me weekend.

pelikhan:
On a second though we might be swinging the pendulum too much the other way.
I thinl that ``pop`` and ``shift`` should retain the same reporter style but the other ones can be converted. The result ``unshift``, ``push``, ``remove...`` is generall yingore.

docs/static/playground/pxt-common/pxt-core.d.js
```diff
@@ -28,7 +28,7 @@ interface Array<T> {
       * Remove the last element from an array and return it.
       */
     //% help=arrays/pop
-    //% shim=Array_::pop weight=45
+    //% shim=Array_::pop weight=45 blockHidden
```
pelikhan, line 31: i think we should keep pop as a reporter

docs/static/playground/pxt-common/pxt-core.d.js
```diff
@@ -45,7 +45,7 @@ interface Array<T> {
       * Remove the first element from an array and return it. This method changes the length of the array.
       */
     //% help=arrays/shift
-    //% helper=arrayShift weight=30
+    //% helper=arrayShift weight=30 blockHidden
```
pelikhan, line 48: keep shift as reporter too.

riknoll:
Okay, I restored all of the blocks and @shakao helped me categorize them:

Screen Shot 2020-04-28 at 1 09 02 PM


@abchatra can I please remove unshift at least? It returns the *length* of the list.

pelikhan:
I agree with unshift. We end up with blocks with the same wording but reporter/statement.

abchatra:
This looks fantastic

abchatra:
Also may be have a section "Read and Modify" and move some of the "Read" blocks under that. 

Also not sure operations as a sub is required. "Reverse" can go to Modify 
& Unshift can go to "Read and Modify"

riknoll:
I don't think we need a separate heading, the text on the block is pretty explanatory. Also I think that unshift should just go away; it isn't even returning a value from the array like the other blocks (it returns the new length). Just confusing

pelikhan:
Did you remove the unshift reporter?

pelikhan:
image

riknoll:
Nope, is that the consensus? It's a one line change

pelikhan:
Let's hide it. It's confusing to have those two blocks in there.

abchatra:
Lets remove it. Lets keep the old block hidden for backward compat.

PR #6902, opened 2020-04-23
Adding the ability to expand the tilemap categories

riknoll:
We hardcoded the tilemap categories, which was a bad decision! This adds the ability to add extra categories.

To add a category, give a tile a tag of the form `category-<name>` (no angle brackets) and it will be placed in a new category with that name. Those names can be localized if you add a localization file with a key like this: `{id:tilecategory}name`.


riknoll:
@abchatra this is the PR I'd like to do a patch release for

riknoll:
And here's the use case: https://github.com/microsoft/pxt-tilemaps/pull/3

PR #6910, opened 2020-04-24
Scaleledmatrix

pelikhan:
Automatically scale led matrix buttons based on the size of the matrix. Allow allows it to be parameteriszed. Fix for https://github.com/microsoft/pxt-microbit/issues/2891

image


MTKilpatrick:
Thank you for putting this in, that's great. Did you consider that 0.6 is the smallest it should be?

pelikhan:
Trying to avoid tiny buttons that become an accessiliby issue. How many LEDs did you need?

MTKilpatrick:
Maybe not too many!

Is this in the beta release only or will it and the custom size facility appear in the default?

pelikhan:
Beta only for now.

PR #6919, opened 2020-04-27
Update floating workspace elements on resize

shakao:
Fixes https://github.com/microsoft/pxt-microbit/issues/2903

This happens when the workspace resizes without hiding the floating elements (dropdowns, "active" fields, etc). I tested against the collapse/expand simulator case, but it should apply to anywhere we use the editor.resize() call.


pelikhan:
Also happens when a warning sign is removed while editing a text field.

shakao:
hmm let me investigate

PR #6921, opened 2020-04-27
Error list toggles with x button and added header

hristoramirez:
Switched over to using an x to close the error list as mentioned in the design meeting. Also added a simple header.
Fixes: microsoft/pxt-microbit#2834
close_with_xbutton


pelikhan:
Try not to mix pull requests.

pelikhan:
Check semantic ui docs on avaiable classes to style your UI (https://semantic-ui.com/elements/list.html). Then check sui.tsx if it is implemented already.

pelikhan:
Looks good for this iteration.

webapp/src/errorList.tsx
```diff
@@ -38,19 +38,27 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
             return `${error.messageText}-${error.fileName}-${error.line}-${error.column}`
         }
 
-        const toggleButton = <sui.Button id='toggleErrorList' className={`toggleErrorList collapse-button large`}
```
jwunderl, line 41: Should be able to drop the #toggleErrorList rule in errorList.less?

theme/errorList.less
```diff
@@ -20,9 +21,35 @@
         height: 200px;
         border-top: 1px #e5e5e5 solid;
         padding: 1em;
+
+        h4 {
+            margin-bottom: 8px;
```
jwunderl, line 26: To keep consistent with the rest of the spacing, maybe 0.5em here instead of 8px? Up to you ~

webapp/src/errorList.tsx
hristoramirez: Yeah it should just removed it. Thanks

PR #6930, opened 2020-04-28
Add package upgrade rules

mmoskal:
Also improvements to compile switches and variants

### DIFF
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -37,6 +37,10 @@ declare namespace pxt {
         bannedRepos?: string[];
         allowUnapproved?: boolean;
         preferredRepos?: string[]; // list of company/project(#tag) of packages to show by default in search
+        // format:
+        // "acme-corp/pxt-widget": "min:v0.1.2" - auto-upgrade to that version
+        // "acme-corp/pxt-widget": "dv:foo,bar" - add "disablesVariant": ["foo", "bar"] to pxt.json
+        upgrades?: pxt.Map<string>;
     }
 
     interface AppTarget {
--- a/pxtcompiler/emitter/hexfile.ts
+++ b/pxtcompiler/emitter/hexfile.ts
@@ -1088,13 +1088,16 @@ _stored_program: .hex ${res}
     }
 
     function assembleAndPatch(src: string, bin: Binary, opts: CompileOptions, cres: CompileResult) {
-        if (opts.extinfo.disabledDeps) {
+        const dummy = opts.extinfo.disabledDeps
+        if (dummy) {
             src =
-                `; compilation disabled on this variant due to ${opts.extinfo.disabledDeps}` +
+                `${hexfile.hexPrelude()}\n` +
+                `; compilation disabled on this variant due to ${opts.extinfo.disabledDeps}\n` +
                 `.hex 718E3B92C615A841C49866C975EE5197\n` +
                 `.string "${opts.extinfo.disabledDeps}"`
+        } else {
+            src = asmHeader(bin) + src
         }
-        src = asmHeader(bin) + src
         if (opts.embedBlob) {
             bin.packedSource = packSource(opts.embedMeta, ts.pxtc.decodeBase64(opts.embedBlob))
             // TODO more dynamic check for source size
@@ -1106,7 +1109,7 @@ _stored_program: .hex ${res}
         const checksumWords = 8
         const pageSize = hexfile.flashCodeAlign(opts.target)
 
-        if (opts.target.flashChecksumAddr) {
+        if (!dummy && opts.target.flashChecksumAddr) {
             let k = 0
             while (pageSize > (1 << k)) k++;
             let endMarker = parseInt(hexfile.hexTemplateHash().slice(8, 16), 16)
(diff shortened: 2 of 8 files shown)


localtypings/pxtarget.d.ts
```diff
@@ -37,6 +37,10 @@ declare namespace pxt {
         bannedRepos?: string[];
         allowUnapproved?: boolean;
         preferredRepos?: string[]; // list of company/project(#tag) of packages to show by default in search
+        // format:
+        // "acme-corp/pxt-widget": "min:v0.1.2" - auto-upgrade to that version
+        // "acme-corp/pxt-widget": "dv:foo,bar" - add "disablesVariant": ["foo", "bar"] to pxt.json
+        upgrades?: pxt.Map<string>;
```
pelikhan, line 43: do we want to have a mini format like or just have a JSON structure to describe those ranages? what happens if you want to combine min:v0.1.2,dv:foo,bar.

For the target, we have something like this:         
```
patches?: pxt.Map<UpgradePolicy[]>; // semver range -> upgrade policies
```

pxtcompiler/emitter/hexfile.ts
```diff
@@ -1088,13 +1088,16 @@ _stored_program: .hex ${res}
     }
 
     function assembleAndPatch(src: string, bin: Binary, opts: CompileOptions, cres: CompileResult) {
-        if (opts.extinfo.disabledDeps) {
+        const dummy = opts.extinfo.disabledDeps
```
pelikhan, line 1091: extinfo could be undefined if the C++ build failed?

pxtlib/github.ts
```diff
@@ -557,7 +557,16 @@ namespace pxt.github {
             return Promise.resolve<CachedPackage>(undefined);
         }
 
-        return db.loadPackageAsync(p.fullName, p.tag);
+        return db.loadPackageAsync(p.fullName, p.tag)
+            .then(cached => {
+                const dv = upgradedDisablesVariants(config, repoWithTag)
+                if (dv) {
+                    const cfg = JSON.parse(cached.files[pxt.CONFIG_NAME]) as pxt.PackageConfig
```
pelikhan, line 564: use ``Package.parseAndValidConfig``

localtypings/pxtarget.d.ts
mmoskal: I'm not a great fan of using JSON as a DSL... in this case there's also no need to combine - if you enforce a minimum version, you also make sure it has the right properties in its pxt.json

pxtcompiler/emitter/hexfile.ts
mmoskal: I don't believe we get here when the build failed; there was an unguarded reference down below for a while

pxtlib/github.ts
mmoskal: thanks!

localtypings/pxtarget.d.ts
mmoskal: In fact I think the package min version should probably move to UpgradePolicy. Currently, we show the wrong version number in explorer after upgrade.

mmoskal: OK, I'm mistaken. I'll just apply it in the place where UpgradePolicies are applied.

mmoskal: OK done. Package rules are now still in targetconfig.json so they can be updated out of band. The package refs are updated in pxt.json and persisted, and disabledVariants are done on the fly.

pxtlib/github.ts
```diff
@@ -557,7 +557,18 @@ namespace pxt.github {
...
 
-        return db.loadPackageAsync(p.fullName, p.tag);
+        return db.loadPackageAsync(p.fullName, p.tag)
+            .then(cached => {
+                const dv = upgradedDisablesVariants(config, repoWithTag)
+                if (dv) {
+                    const cfg = Package.parseAndValidConfig(cached.files[pxt.CONFIG_NAME])
+                    if (cfg) {
+                        cfg.disablesVariants = dv
+                        cached.files[pxt.CONFIG_NAME] = JSON.stringify(cfg, null, 4)
```
pelikhan, line 567: We also have an api to strongly the pxt.json . I think stringingConfig or something. It makes sure that fields are in same order etc

pxtlib/github.ts
```diff
@@ -557,7 +557,18 @@ namespace pxt.github {
...
         }
 
-        return db.loadPackageAsync(p.fullName, p.tag);
+        return db.loadPackageAsync(p.fullName, p.tag)
+            .then(cached => {
+                const dv = upgradedDisablesVariants(config, repoWithTag)
+                if (dv) {
+                    const cfg = Package.parseAndValidConfig(cached.files[pxt.CONFIG_NAME])
+                    if (cfg) {
+                        cfg.disablesVariants = dv
```
pelikhan, line 566: Add a debug log entry?

pxtlib/package.ts
```diff
@@ -318,6 +318,24 @@ namespace pxt {
             return regex.test(ts);
         }
 
+        private upgradePackagesAsync() {
+            return pxt.packagesConfigAsync()
+                .then(config => {
+                    let numfixes = 0
+                    U.iterMap(this.config.dependencies, (pkg, ver) => {
+                        if (pxt.github.isGithubId(ver)) {
+                            const upgraded = pxt.github.upgradedPackageReference(config, ver)
```
pelikhan, line 327: probably good to first normalize the GitHub id

pelikhan:
```
                            ver = pxt.github.normalizeRepoId(ver);
```

mmoskal: we parse it on the other side, so it's probably OK

mmoskal:
good to go as soon as build passes

PR #6932, opened 2020-04-28
Fixing indent rules for python formatter

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/2869

The indent logic in the python formatter assumes that the pasted text starts at the beginning of the line, which isn't always the case.


pelikhan:
Figuring out how to write tests for those will pay off in the long run.

PR #6954, opened 2020-05-01
Show monaco debugger in non-main files

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/2913


pelikhan:
The monaco toolbox takes a lot of valuable space. Is there anyway to collapse it to icon rendering outside of main.ts?

riknoll:
I'm not opposed to it but I think we should talk about it at the design meeting. I'm a little worried people might be confused as to why it's changing. Agree that it would be nice to minimize it sometimes.

pelikhan:
The only people seeing this change will be people who actually add new files .ts/.py files. Not our regular users.

PR #6958, opened 2020-05-02
Add helpful information about creating an extension when adding a new .ts file

darzu:
Experimenting with the idea of giving a helpful start for extensions when creating a new .ts file.
Otherwise it's pretty hard to find documentation on extension writing.

What do you guys think? @pelikhan , @jwunderl , @shakao , @riknoll ?

### DIFF
--- a/webapp/src/filelist.tsx
+++ b/webapp/src/filelist.tsx
@@ -281,6 +281,21 @@ export class FileList extends data.Component<ISettingsProps, FileListState> {
             return this.props.parent.updateFileAsync(
                 fileName,
                 `${comment} ${pxt.U.lf("Add your code here")}
+                
+                ${comment} ${pxt.U.lf("If this is an extension:")}
+                ${comment} ${pxt.U.lf("1. Uncomment the code below")}
+                ${comment} ${pxt.U.lf("2. Change the namespace name. This name will be shown as a new blocks category.")}
+                ${comment} ${pxt.U.lf("3. Choose a new icon from: https://fontawesome.com/v4.7.0/icons/")}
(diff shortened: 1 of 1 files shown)


jwunderl:
Definitely pro this

One thing is that a lot of the time when I look at user extensions they end up throwing things just in main - which is fine, but they wouldn't ever see this then, or maybe would only see it after working through the initial stuff / finding it themselves. Maybe this could be the initial comment in main.ts if you create a github repo project as well? But that project starts in blocks view, so maybe not.

At first I thought it might get a bit annoying on every file, but then I remembered I always just ctrl+a > delete to clear the initial comment anyway so it's really no different that way. 

Maybe use a block comment for the code section that needs to be uncommented? I'm not certain the difference between the extra comment you need to remove and normal comment attributes will be obvious at first - e.g. might try to remove comments and leave a line starting with %?

pelikhan:
There is already a template for custom.ts to show how do custom blocks. We should have a single template probably.

But it’s also incredibly annoying and people won’t know if they can just delete. I think the dialog that lets you name the file should also have few templates to pick from instead.

shakao:
ooh i really like the idea of choosing a template when you make a new file! maybe to somewhat combat the main.ts issue we could have a link to the extension docs in the github view "release zone" as well...

darzu:
Awesome, yes I agree a template choice during file creation seems ideal.

Then my question is how does anyone know to add a file..
Maybe when the explorer is collapsed, there is some button or hint about adding more files

pelikhan:
It is an advanced scenario so it is ok to keep it hidden. We don’t have great support for multiple files.

PR #6963, opened 2020-05-04
Fix member auto-complete for primative types

darzu:
number, string, and boolean types all weren't working with autocomplete.

Also expands the language service tests to be able to handle negative cases, and provide more info on failure.

Fixes: https://github.com/microsoft/pxt-microbit/issues/2803

Now:
Screen Shot 2020-05-04 at 11 09 33 AM
Screen Shot 2020-05-04 at 11 09 50 AM

### DIFF
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -1008,7 +1008,7 @@ namespace ts.pxtc.service {
                     }
 
                     if (type) {
-                        const qname = type.symbol ? tc.getFullyQualifiedName(type.symbol) : getNameOfWidenedType(type, tc);
+                        const qname = type.symbol ? tc.getFullyQualifiedName(type.symbol) : tsTypeToPxtTypeString(type, tc);
 
                         if (qname) {
                             const props = type.getApparentProperties()
@@ -2146,19 +2146,27 @@ namespace ts.pxtc.service {
         return (n && n.kind === SK.SourceFile) ? null : n;
     }
 
-    function getNameOfWidenedType(t: Type, tc: TypeChecker) {
+    function tsTypeToPxtTypeString(t: Type, tc: TypeChecker) {
         if (t.flags & TypeFlags.NumberLiteral) {
-            return "number";
+            return "Number";
         }
         else if (t.flags & TypeFlags.StringLiteral) {
             return "String";
         }
         else if (t.flags & TypeFlags.BooleanLiteral) {
-            return "boolean";
+            return "Boolean";
         }
 
-        return tc.typeToString(t);
+        const tcString = tc.typeToString(t);
+        const primativeToQname: { [key: string]: string } = {
+            "number": "Number",
+            "string": "String",
+            "boolean": "Boolean"
+        }
+        const pxtString = primativeToQname[tcString] ?? tcString
+        return pxtString
     }
+
 
     function filenameWithExtension(filename: string, extension: string) {
         if (extension.charAt(0) === ".") extension = extension.substr(1);
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -1154,7 +1154,9 @@ namespace ts.pxtc.service {
                     si.attributes.hidden ||
                     si.attributes.deprecated ||
                     // ignore TD_ID helpers
-                    si.attributes.shim == "TD_ID"
+                    si.attributes.shim == "TD_ID" ||
+                    // ignore block aliases like "_popStatement" on arrays
+                    si.attributes.blockAliasFor
                 )
                 return use
             }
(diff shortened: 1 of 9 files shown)


riknoll:
Is this going to cause us to sometimes emit the JS "boxed" types instead of the primitive? We should never produce code with "Number", "String", or "Boolean"

pelikhan:
Let's get in habit of writing tests for those. Is there some docs on how to add tests for completion? It' d be helpful to be able to write those for future bug reports.

darzu:
- [ ] I'll add a doc for writting these tests
- [ ] check that "Number" etc aren't emitted
- [ ] verify this fixes Python too

darzu:
Turns out we're showing "Number" as an option because of another bug: https://github.com/microsoft/pxt-microbit/issues/3001
I'll ignore that for this pr.

pxtcompiler/emitter/service.ts
```diff
@@ -2146,20 +2148,28 @@ namespace ts.pxtc.service {
...
             return "String";
         }
         else if (t.flags & TypeFlags.BooleanLiteral) {
-            return "boolean";
+            return "Boolean";
         }
 
-        return tc.typeToString(t);
+        const tcString = tc.typeToString(t);
+        const primativeToQname: { [key: string]: string } = {
```
pelikhan, line 2163: not we have a Map type

tests/language-service/languageservicerunner.ts
```diff
@@ -46,16 +47,34 @@ describe("language service", () => {
...
+    let filenames: string[] = [];
     for (const file of fs.readdirSync(casesDir)) {
+        // ignore hidden files
         if (file[0] == ".") {
             continue;
         }
 
+        // ignore files that start with TODO_; these represent future work
+        if (file.indexOf("TODO") >= 0) {
+            continue;
```
pelikhan, line 59: Emit a warning?

tests/language-service/languageservicerunner.ts
```diff
@@ -46,16 +47,34 @@ describe("language service", () => {
...
+        }
+
+        // TODO support python as well
+        if (file.substr(-3) !== ".ts") {
+            continue;
```
pelikhan, line 64: Emit a warning? Or we’ll forget

pxtcompiler/emitter/service.ts
darzu: How do we change the policy on this? 
This is standard .ts syntax and I feel it should be prefered to pxt.U.Map. Also, static TS supports { [key: string]: string } syntax but not pxt.U.Map (somewhat obviously)

Not going to hold up this PR on this tho so I'll switch to Map

tests/language-service/languageservicerunner.ts
darzu: Hmm, I don't think a warning is a appropriate here. This is a great way to add a test case that we can reference in an Issue. The priority of this should be decided in the Issue and might not justify a warning every time. We know there will be lots of features we don't get to for a while. This is also a way to document which language service features are missing.

tests/language-service/languageservicerunner.ts
darzu: I filed an issue and will put the issue number here. Again, I think work item priority should be decided outside of the code and therefore the code shouldn't force an importance by throwing a warning.

tests/language-service/languageservicerunner.ts
pelikhan: I disagree this becomes a secret test case that no one cares about. We should emit a message in the console stating we skipping a test. Test do get lost if they are not run.

darzu: I'm okay with the tests becoming "lost". They can be found when they become relevant, e.g. if we decide a feature is important for users or an issue gets raised in priority. Someone exploring this part of the code, e.g. if I go on another intellisense tear in the future, can discover these easily enough.

Warnings that are always present get lost just as easily.
Everytime anyone runs pxt serve we output:
"error TS1084: Invalid 'reference' directive syntax."
no one cares. I believe this has been there for >1 year.

pxtcompiler/emitter/service.ts
pelikhan: (not/nit auto complete...) Discuss it at the team meeting.

tests/language-service/languageservicerunner.ts
pelikhan: a console.log calls does not hurt and reminds everyone about this fact. it'll save someone else an hour in the future to understand why the .ts test does not run. github issues are fragile, they get closed/buried.

tests/language-service/languageservicerunner.ts
pelikhan: I've noticied that my future self forgets these kinds of subtle semantics (TODO means don't run). Adding a simple console.log today might pay off saving a few hours in the future.

> Warnings that are always present get lost just as easily.

They come back on every build.

darzu: Worth noting that I documented the behavior of TODO_ in the .md file for this test suite.

I don't feel strongly about this anymore, so I'm fine adding a warning during my next PR if you still feel strongly.

PR #6964, opened 2020-05-04
Added some of the suggestions made at the design meeting on 4/29

hristoramirez:
- made the header sticky when scrolling and show on collapse, fixes: microsoft/pxt-microbit#2938

- changed to collapse chevron

- border slightly thicker

- start collapsed, fixes: microsoft/pxt-microbit#2953

- changed wording to problems

- switched to just using line numbers

- hide error list when there are no errors

- display number of errors in bubble

- sizing change to accommodate header, fixes: microsoft/pxt-microbit#2895

design-changes-1

### DIFF
--- a/theme/errorList.less
+++ b/theme/errorList.less
@@ -6,24 +6,32 @@
     min-height: 200px;
     border-top: 1px #e5e5e5 solid;
 
-    h4 {
+    .errorListHeader {
         padding: 1em;
         margin-bottom: 0;
-    }
 
-    .collapseButton {
-        float: right;
-        cursor: pointer;
-        position: absolute;
-        top: 1em;
-        right: 1em;
-
-        i {
-            font-size: 2em;
+        h4 {
+            margin: 0;
+            display: inline-block;
+            vertical-align: middle;
         }
 
-        i:not(:hover) {
-            opacity: 0.5;
+        .collapseButton {
+            //float: right;
+            cursor: pointer;
+            position: relative;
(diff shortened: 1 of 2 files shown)


theme/errorList.less
```diff
@@ -1,48 +1,58 @@
...
             }
 
             i:not(:hover) {
                 opacity: 0.5;
             }
         }
     }
+
+    .errorListInner {
+        height: 160px;
```
pelikhan, line 40: use em so you know how many lines of text to display

theme/errorList.less
```diff
@@ -1,48 +1,60 @@
 @summary-hover-color: #00000067;
 
+// Expanded error list
 .errorList {
     position: relative;
-    min-height: 200px;
+    max-height: 16em;
+    min-height: 16em;
+    border-top: 2px #e5e5e5 solid;
```
darzu, line 8: probably worth a less variable
I'd recommend a less variable for all colors by default

theme/errorList.less
```diff
@@ -1,48 +1,60 @@
...
             }
 
             i:not(:hover) {
                 opacity: 0.5;
             }
         }
     }
+
+    .errorListInner {
+        height: 12em;
```
darzu, line 42: Since this height is 16em - 4em from the header, it's probably worth using LESS variables and math here.
https://stackoverflow.com/questions/17433849/less-subtract-from-variable

PR #6966, opened 2020-05-04
Deprecating Math.randomRange in favor of randint

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/2894
Fixes https://github.com/microsoft/pxt-microbit/issues/2955

This deprecates `Math.randomRange(0, 1)` in favor of `randint(0, 1)`.

The reasoning for this change is that we already made up the nonstandard API of `Math.randomRange()` and we need to support `randint()` for python so we might as well unify the two. I'm using our usual strategy of leaving the API accessible in TypeScript but hiding the block from the toolbox.

All existing projects should be unaffected. No upgrade rules required!

### DIFF
--- a/docs/static/playground/pxt-common/pxt-core.d.js
+++ b/docs/static/playground/pxt-common/pxt-core.d.js
@@ -387,6 +387,18 @@ declare interface String {
 //% text.defl="123"
 declare function parseFloat(text: string): number;
 
+/**
+ * Returns a pseudorandom number between min and max included.
+ * If both numbers are integral, the result is integral.
+ * @param min the lower inclusive bound, eg: 0
+ * @param max the upper inclusive bound, eg: 10
+ */
+//% blockId="device_random" block="pick random %min|to %limit"
+//% blockNamespace="Math"
+//% help=math/randint
+//% shim=Math_::randomRange
+declare function randint(min: number, max: number): number;
+
 interface Object { }
 interface Function { }
 interface IArguments { }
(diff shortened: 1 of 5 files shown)


webapp/src/monacoSnippets.ts
```diff
@@ -250,9 +250,9 @@ function cachedBuiltinCategories(): pxt.Map<BuiltinCategoryDefinition> {
                     retType: "number"
                 },
                 {
-                    name: "Math.randomRange",
-                    snippetName: "randomRange",
-                    snippet: `Math.randomRange(0, 10)`,
+                    name: "randint",
```
pelikhan, line 253: do we even need this now?

docs/static/playground/pxt-common/pxt-core.d.js
```diff
@@ -387,6 +387,18 @@ declare interface String {
...
+ * Returns a pseudorandom number between min and max included.
+ * If both numbers are integral, the result is integral.
+ * @param min the lower inclusive bound, eg: 0
+ * @param max the upper inclusive bound, eg: 10
+ */
+//% blockId="device_random" block="pick random %min|to %limit"
+//% blockNamespace="Math"
+//% help=math/randint
+//% shim=Math_::randomRange
+declare function randint(min: number, max: number): number;
```
darzu, line 400: I forget how the playground works... is  this a build artifact or does it need to be kept in sync with pxt-common?

riknoll: Needs to be in sync. I'll update it

PR #6973, opened 2020-05-05
don't turn snippets to darkmode in ts

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/2890

typescript sidedocs were getting [vs2015 dark mode style](https://github.com/microsoft/pxt/blob/master/webapp/public/highlight.js/styles/vs2015.css) instead of [vs light mode style](https://github.com/microsoft/pxt/blob/master/webapp/public/highlight.js/styles/vs.css)

Reason it was only applying to ts and not python is because check [here]( https://github.com/microsoft/pxt/blob/master/pxtrunner/runner.ts#L421) is just against ts (there isn't a LanguageMode.Python atm)

did we actually want dark mode side docs for e.g. adafruit? if not should probably remove the file / not load it in docs in this PR too


pelikhan:
We don’t have the bandwidth to test 2 styles . Keep it light.

PR #6975, opened 2020-05-05
jump to error when the message is clicked

hristoramirez:
Error messages are now clickable and lead to the corresponding error
clickable-errors

### DIFF
--- a/theme/errorList.less
+++ b/theme/errorList.less
@@ -1,19 +1,15 @@
-@total-height: 16em;
-@header-height: 4em;
-@inner-height: (@total-height - @header-height);
-@collapsed-errors-hover-color: rgba(0, 0, 0, 0.05);
-@border-color: #e5e5e5;
+@import 'themes/pxt/globals/site.variables';
 
 // Expanded error list
 .errorList {
     position: relative;
-    max-height: @total-height;
-    min-height: @total-height;
-    border-top: 2px @border-color solid;
+    max-height: @errorListHeight;
+    min-height: @errorListHeight;
+    border-top: 2px @errorListBorderColor solid;
 
     .errorListHeader {
         padding: 1em;
-        height: @header-height;
+        height: @errorListHeaderHeight;
         margin-bottom: 0;
 
         h4 {
@@ -43,7 +39,7 @@
     }
 
     .errorListInner {
-        height: @inner-height;
+        height: @errorListInnerHeight;
         padding: 0 0 1em 0;
         overflow-y: scroll;
 
@@ -52,7 +48,7 @@
 
             &:hover {
                 cursor: pointer;
-                background-color: @collapsed-errors-hover-color;
+                background-color: @errorListMessageHoverColor;
             }
         }
     }
@@ -64,6 +60,6 @@
 
     &:hover {
         cursor: pointer;
-        background-color: @collapsed-errors-hover-color;
+        background-color: @errorListMessageHoverColor;
     }
 }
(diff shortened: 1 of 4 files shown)


webapp/src/errorList.tsx
```diff
@@ -46,8 +48,12 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
             {!isCollapsed && collapseButton}
         </div>
 
+        const createOnErrorMessageClick = (e: pxtc.KsDiagnostic) => () => this.onErrorMessageClick(e.line + 1, e.column + 1)
+
         const errorListContent = (errors).map(e =>
-            <div key={errorKey(e)} className="errorMessage">{`${e.messageText} ${lf("at line {0}", e.line + 1)}`}</div>)
+            <div key={errorKey(e)} className="errorMessage" role="button" onClick={createOnErrorMessageClick(e)}>
```
jwunderl, line 54: This change suppresses the error, but the actual cause is still there -- it'll keep recreating the function each time render is called. It may not be a big deal for a single element, but it's best to avoid that as it adds up.

Here's an example of a workaround for something similar: https://github.com/microsoft/pxt/blob/master/webapp/src/filelist.tsx#L387 - that stores the file, but you could store line / col as props on the message div and reference them from that

theme/errorList.less
```diff
@@ -1,7 +1,7 @@
 @total-height: 16em;
 @header-height: 4em;
 @inner-height: (@total-height - @header-height);
-@collapsed-errors-hover-color: rgba(0, 0, 0, 0.02);
+@collapsed-errors-hover-color: rgba(0, 0, 0, 0.05);
```
pelikhan, line 4: Move all LESS variables to https://github.com/microsoft/pxt/blob/master/theme/themes/pxt/globals/site.variables

webapp/src/monaco.tsx
```diff
@@ -590,6 +591,12 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         this.errorChangesListeners[handlerKey] = handler;
     }
 
+    goToError(line: number, column: number) {
+        this.editor.revealLineInCenterIfOutsideViewport(line, monaco.editor.ScrollType.Smooth);
```
pelikhan, line 595: add a call to pxt.tickEvent(...) to track api usage

webapp/src/errorList.tsx
darzu: Hmm, we could do a separate component per message which basically gives us a way to store some data on the local element.

I feel like in this case it should be fine.. the anonymous functions we're creating are just closing over a couple numbers. It's not obvious to me another approach would use fewer resource

jwunderl: That's true almost any time this shows up though - you typically are only passing a param or two, so it's only ever an anonymous function wrapping a few references - the main cost is in extra rerenders for every item when the fn is considered not equal to the previous one each time the errorlist rerenders.

That won't be a noticeable cost because it's a smaller component, but it's easy enough to break out into an `ErrorListItem` component (and that helps keep `ErrorList.render` nice and tidy)

(This is a rather small thing that can be done separately though)

shakao: i think you could add "data-row=" and "data-col=" attributes onto the div and use e.target to grab them in the event listener

pelikhan: This is a micro-optimization and we can do another PR for this kind of cleanup. Push this fix out into another PR.

PR #6983, opened 2020-05-06
Correct is-module check

mmoskal:
fixes https://github.com/microsoft/pxt-microbit/issues/2653

are there any python tests running on build?


mmoskal:
@darzu do Python tests run on build?

PR #6986, opened 2020-05-07
Adding preview for debugger variable values

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/2979

The preview disappears when it loses focus. It also selects the text contents when it appears so that you can copy with ctrl-c

### DIFF
--- a/theme/debugger.less
+++ b/theme/debugger.less
@@ -203,6 +203,31 @@
     }
 }
 
+.debugger-preview {
+    position: fixed;
+    background-color: white;
+    color: black;
+    border: solid 1px black;
+    padding: 0.25rem;
+
+    -webkit-touch-callout: text; /* iOS Safari */
+    -webkit-user-select: text;   /* Chrome/Safari/Opera */
+    -khtml-user-select: text;    /* Konqueror */
+    -moz-user-select: text;      /* Firefox */
+    -ms-user-select: text;       /* Internet Explorer/Edge */
+    user-select: text;           /* Non-prefixed version, currently
+                                    not supported by any browser */
+    overflow-wrap: break-word;
+    max-width: 500px;
+}
+
+// When the preview appears, the text is selected so that the user can copy it.
+// Hide the selection formatting because it makes it look like the text can be edited
+.debugger-preview::selection {
+    color: black;
+    background-color: white;
+}
+
 .debugging .ui.segment.debugvariables {
     display: block;
     margin: 0;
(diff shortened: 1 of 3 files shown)


theme/debugger.less
```diff
@@ -203,6 +203,31 @@
...
+    color: black;
+    border: solid 1px black;
+    padding: 0.25rem;
+
+    -webkit-touch-callout: text; /* iOS Safari */
+    -webkit-user-select: text;   /* Chrome/Safari/Opera */
+    -khtml-user-select: text;    /* Konqueror */
+    -moz-user-select: text;      /* Firefox */
+    -ms-user-select: text;       /* Internet Explorer/Edge */
+    user-select: text;           /* Non-prefixed version, currently
```
jwunderl, line 218: for what it's worth this seems to be supported across all browsers (with a note on ios): https://caniuse.com/#feat=mdn-css_properties_user-select_text no harm leaving vendor prefixes in though

theme/debugger.less
```diff
@@ -203,6 +203,31 @@
...
+                                    not supported by any browser */
+    overflow-wrap: break-word;
+    max-width: 500px;
+}
+
+// When the preview appears, the text is selected so that the user can copy it.
```
jwunderl, line 224: I don't think I would ever realize it's selected / copyable without the selection highlight - we do have the same issue with the share dialog as it auto selects the url and we leave the selection highlight there, maybe it's fine to leave it?

webapp/src/debuggerVariables.tsx
```diff
@@ -107,17 +117,37 @@ export class DebuggerVariables extends data.Component<DebuggerVariablesProps, De
...
+            <DebuggerTable header={variableTableHeader} placeholderText={placeholderText}>
+                {tableRows}
+            </DebuggerTable>
+            { preview &&
+                <div id="debugger-preview"
+                    role="tooltip"
+                    className="debugger-preview"
+                    ref={this.handlePreviewRef}
+                    tabIndex={0}
+                    style={{
```
jwunderl, line 132: I think an aria-label here saying something like `Current value for {variable}:` would be helpful

theme/debugger.less
riknoll: I think it looks like "edit and continue" but broken because when you click on it the tooltip disappears. I'd rather leave it as is

theme/debugger.less
riknoll: Copy/pasted from elsewhere in the codebase

webapp/src/debuggerVariables.tsx
riknoll: Will do!

PR #6990, opened 2020-05-08
Add support for right-click to copy a marquee selection.

jacobcarpenter:
This PR adds support for right-click to "copy" the current marquee selection.

Currently, as you make a marquee selection, the pixels are "cut" from the image layer and "pasted" to the floating layer. **None of that behavior changes with this PR.**

With this PR, after making a selection, if you right-click within it, the selected pixels are "pasted" back to the image layer and a new copy (to the floating layer) is made **without** cutting.

Here's a gif (black circles are left-clicks; green circles are right-clicks):

Kapture 2020-05-07 at 23 01 46

I tested this within the sprite editor, tilemap editor, and background image editor, and all worked as I'd expected.

I fully recognize right-click is a less discoverable mechanism for exposing this functionality. microsoft/pxt-arcade#1835 mentions Ctrl+C, Ctrl+V. But some of the interactions seemed harder/more awkward as I started to consider the steps required to duplicate pixels with the existing move/cut-by-default behavior of the marquee tool and a clipboard Ctrl+C, Ctrl+V metaphor.

I'd love to hear your feedback on this, and I'm happy to make any changes!

### DIFF
--- a/webapp/src/components/ImageEditor/ImageCanvas.tsx
+++ b/webapp/src/components/ImageEditor/ImageCanvas.tsx
@@ -179,6 +179,11 @@ class ImageCanvasImpl extends React.Component<ImageCanvasProps, {}> implements G
 
         if (!this.inBounds(this.cursorLocation[0], this.cursorLocation[1])) return;
 
+        if (isRightClick && this.isInMarqueeSelection(this.cursorLocation[0], this.cursorLocation[1])) {
+            this.copyMarqueeSelection();
+            return;
+        }
+
         this.startEdit(!!isRightClick);
         this.updateEdit(this.cursorLocation[0], this.cursorLocation[1]);
         this.commitEdit();
@@ -354,16 +359,7 @@ class ImageCanvasImpl extends React.Component<ImageCanvasProps, {}> implements G
             this.edit.doEdit(this.editState);
             this.edit = undefined;
 
-            dispatchImageEdit({
-                bitmap: this.editState.image.data(),
-                layerOffsetX: this.editState.layerOffsetX,
-                layerOffsetY: this.editState.layerOffsetY,
-                floating: this.editState.floating && {
-                    bitmap: this.editState.floating.image ? this.editState.floating.image.data() : undefined,
-                    overlayLayers: this.editState.floating.overlayLayers ? this.editState.floating.overlayLayers.map(el => el.data()) : undefined
-                },
-                overlayLayers: this.editState.overlayLayers ? this.editState.overlayLayers.map(el => el.data()) : undefined
-            });
+            dispatchImageEdit(this.editState.toImageState());
         }
     }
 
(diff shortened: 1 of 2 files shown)


jacobcarpenter:
cc @riknoll

riknoll:
Code looks good! @jwunderl @shakao @darzu what do you think about the interaction?

I honestly don't have any problems with it being right-click, but I don't know if that has semantic meaning in other image editors. Maybe we could reframe it as some sort of "stamp" tool?

riknoll:
Also, @jacobcarpenter thanks for the contribution! You are making all of the changes I wish I had time to do :smile:

darzu:
This is so awesome!
I agree that the user interaction here is a bit tricky.
I think my biggest concern with right-click is that in the future we might want a context menu.

Eventually I do think we should have a UI button of some form for copy-paste for discoverability purposes, but I think it's fine to first at the feature as click or keyboard shortcut only.

For reference, here's what MS paint does on ctrl-c, ctrl-v:


Here's what Mac's Preview does on cmd-c,  cmd-v:


I kind of like these approaches, especially the Mac one: put the copied version slightly offset from the original with the floating copy selected.

riknoll:
@darzu we actually already use right-click in our other paint tools for switching between foreground/background colors (it's pretty common in most image editors).

If right click is an issue, maybe we should use one of the modifier keys? I agree that we need copy/paste, but I think having a "stamp" tool is also pretty neat. Maybe we could tie it to one of the modifier keys?

shakao:
ctrl+drag with the marquee tool is what mspaint uses for "stamp". it's definitely not one of the more commonly-used hotkeys, though, so i think we should feel free to diverge if we want. i also prefer hiding it in the hotkeys (as with the eyedropper) until we add menu items/figure out some way to expose it in the ui--that way we can keep the keyboard shortcut even after we add a button for it

jacobcarpenter:
Yes, "stamp" is definitely a better word than "copy" for the currently implemented interaction. :+1:

It sounds like keyboard shortcut is building support—which is great, I'm happy to update the PR to use any trigger.

I do want to put in a vote for right-click though: it seems like most of the keyboard shortcuts in the editor today are about picking a "mode" (a different tool, or panning controls, or the dropper)... this feature is less a mode and more a canvas modifying action. Those do tend to be clicks on the canvas surface.

... but, again, happy to update to anything that reaches consensus.

jacobcarpenter:
I know this is outside of your normally scheduled/prioritized workload; thanks for continuing to take the time!

Should I close this PR and reimplement the functionality like @darzu suggests? (Ctrl/Cmd+C, Ctrl/Cmd+V creates a new floating layer with a copy of the originally selected pixels; offset slightly from the original selection?)

It's probably the most discoverable/expected mechanism for accessing duplicate functionality.

riknoll:
@jacobcarpenter that sounds great! Sorry for the delayed reply

jacobcarpenter:
No problem! You guys are busy making other awesome things.

Closing with plans to reimplement as described above.

PR #6992, opened 2020-05-08
error list design meeting changes 2

hristoramirez:
Added the suggestions given

- zebra stripe

- lead with line number

- red notification color

- reduced padding around header

- go to end of error with cursor
design2


hristoramirez:
fixes: microsoft/pxt-microbit#2830

PR #6997, opened 2020-05-11
error list now collapses when clicking anywhere on header

hristoramirez:
error list now collapses when clicking anywhere in the header

fixes microsoft/pxt-microbit#3004

### DIFF
--- a/webapp/src/errorList.tsx
+++ b/webapp/src/errorList.tsx
@@ -43,7 +43,7 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
         const createOnErrorMessageClick = (e: pxtc.KsDiagnostic, index: number) => () =>
             this.onErrorMessageClick(e, index)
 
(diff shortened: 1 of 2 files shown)


webapp/src/errorList.tsx
```diff
@@ -43,7 +43,7 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
         const createOnErrorMessageClick = (e: pxtc.KsDiagnostic, index: number) => () =>
             this.onErrorMessageClick(e, index)
 
-        const errorListHeader = <div className="errorListHeader">
+        const errorListHeader = <div className="errorListHeader" role="button" onClick={this.onCollapseClick}>
```
pelikhan, line 46: add this parameter onKeyDown={sui.fire.... you’ll find it in many other UI elements.

hristoramirez: is this something that I should add to any element that is also a button?

pelikhan:
Yep

PR #6998, opened 2020-05-11
Unify types on calls to slice, str

shakao:
Fixes https://github.com/microsoft/pxt-microbit/issues/2931

Additional type unification, fixes the following (previously errored on the call to "len"):
```
c = [[2,3], [4,5,6], [7]]
a = c[1]
b = len(a)
```
```
a = "test"[1:2]
b = len(a)
```
```
a = str(45)
b = len(a)
```

### DIFF
--- a/tests/runtime-trace-tests/cases/slice_str_types.py
+++ b/tests/runtime-trace-tests/cases/slice_str_types.py
@@ -1,0 +1,9 @@
+a = [[2,3], [4,5,6], [7]]
+b = a[1]
+print(len(b))
+
+c = "test"[1]
+print(len(c))
+
+d = str(45)
+print(len(d))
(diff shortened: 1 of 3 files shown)


darzu:
Looks good!

Would you mind adding test cases for these? Should be pretty easy to do by adding a .py file to: tests/runtime-trace-tests/cases/

tests/runtime-trace-tests/cases/slice_str_types.py
```diff
@@ -0,0 +1,9 @@
+a = [[2,3], [4,5,6], [7]]
+b = a[1]
+print(len(b))
+
+c = "test"[1]
+print(len(c))
+
+d = str(45)
+print(len(d))
```
darzu, line 9: awesome!

PR #7002, opened 2020-05-11
Scroll formatted code to top left corner of workspace

shakao:
Scrolls blocks to the top left corner of the workspace after format/flow which I think is reasonable/expected behavior. You can load the test program into /beta and see the current behavior, which changes based on where in the workspace you are (eg try scrolling to the top-right corner and format).

Test program: https://makecode.microbit.org/_Dkw2emRfjKRk
Build: https://makecode.microbit.org/app/e3d66d88919b640778708d582fb76001a12b2085-c5005c49da

Fixes https://github.com/microsoft/pxt-microbit/issues/3007

### DIFF
--- a/pxtblocks/blocklylayout.ts
+++ b/pxtblocks/blocklylayout.ts
@@ -142,6 +142,9 @@ namespace pxt.blocks.layout {
             .forEach(b => b.setCollapsed(collapsed));
     }
 
+    // Workspace margins
+    const marginx = 20;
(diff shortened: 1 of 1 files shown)


shakao:
hmm i think i prefer keeping the zoom the same; it seems jarring to change both the user's position and the zoom level. this just tries to fit as many blocks into the visible workspace as possible

PR #7003, opened 2020-05-11
Check pyRetType when emitting function definitions

shakao:
Fall back to pyRetType if no return expression from parser. Type annotation doesn't default to "any" for now, as this seems to be sufficient for the recursive cases.

Fixes https://github.com/microsoft/pxt-microbit/issues/2719

### DIFF
--- a/pxtpy/converter.ts
+++ b/pxtpy/converter.ts
@@ -1111,9 +1111,10 @@ namespace pxt.py {
                 else
                     nodes.push(B.mkText("export function "), quote(funname))
             }
+            let retType = n.returns ? compileType(n.returns) : sym.pyRetType;
             nodes.push(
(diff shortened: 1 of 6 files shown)


pxtpy/converter.ts
```diff
@@ -1111,9 +1111,10 @@ namespace pxt.py {
                 else
                     nodes.push(B.mkText("export function "), quote(funname))
             }
+            let retType = n.returns ? compileType(n.returns) : sym.pyRetType;
             nodes.push(
                 doArgs(n, isMethod),
-                n.returns ? typeAnnot(compileType(n.returns)) : B.mkText(""))
+                retType ? typeAnnot(retType) : B.mkText(""))
```
riknoll, line 1117: retType !== void?

PR #7005, opened 2020-05-12
can scroll to last few lines of text editor when error list is open

hristoramirez:
fixed the issue where the first or last few lines of code would be unreachable when scrolling

fixes microsoft/pxt-microbit#2824

### DIFF
--- a/webapp/src/errorList.tsx
+++ b/webapp/src/errorList.tsx
@@ -43,12 +43,6 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
         const createOnErrorMessageClick = (e: pxtc.KsDiagnostic, index: number) => () =>
             this.onErrorMessageClick(e, index)
 
-        const errorListHeader = <div className="errorListHeader" role="button" onClick={this.onCollapseClick}>
-            <h4>Problems</h4>
-            <div className="ui red circular label countBubble">{errors?.length}</div>
-            <div className="toggleButton"><sui.Icon icon={`chevron ${isCollapsed ? 'up' : 'down'}`} onClick={this.onCollapseClick} /></div>
-        </div>
-
         const errorListContent = <div className="ui selection list">
             {
                 (errors).map((e, index) =>
@@ -59,14 +53,16 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
             }
         </div>
 
-        return <div className={`errorList ${isCollapsed ? 'errorListSummary' : ''}`} hidden={!errorsAvailable}>
-            {errorListHeader}
-            {!isCollapsed &&
-                <div className="errorListInner">
-                    {errorListContent}
+        return  <div className={`errorList ${isCollapsed ? 'errorListSummary' : ''}`} hidden={!errorsAvailable}>
+                    <div className="errorListHeader" role="button" onClick={this.onCollapseClick}>
+                        <h4>Problems</h4>
+                        <div className="ui red circular label countBubble">{errors?.length}</div>
+                        <div className="toggleButton"><sui.Icon icon={`chevron ${isCollapsed ? 'up' : 'down'}`} onClick={this.onCollapseClick} /></div>
                 </div>
-            }
-        </div>
+                    {!isCollapsed && <div className="errorListInner">
+                        {errorListContent}
+                    </div>}
+                </div>
     }
 
     componentDidUpdate() {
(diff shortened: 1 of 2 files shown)


webapp/src/errorList.tsx
```diff
@@ -43,30 +43,26 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
...
-            {!isCollapsed &&
-                <div className="errorListInner">
-                    {errorListContent}
+                }
+            </div> )
+
+        return (
+            <div className={`errorList ${isCollapsed ? 'errorListSummary' : ''}`} hidden={!errorsAvailable}>
+                <div className="errorListHeader" role="button" onClick={this.onCollapseClick}>
+                    <h4>Problems</h4>
```
pelikhan, line 59: Problems -> {lf(“Problems”)}

webapp/src/errorList.tsx
```diff
@@ -43,30 +43,26 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
...
-        </div>
-
-        const errorListContent = <div className="ui selection list">
-            {
-                (errors).map((e, index) =>
+        const errorListContent = (
+            <div className="ui selection list">
+                {(errors).map((e, index) =>
                     <div className="item" key={errorKey(e)} role="button" onClick={createOnErrorMessageClick(e, index)}>
                         {lf("Line {0}: ", e.line + 1)}
```
pelikhan, line 50: {lf("Line {0}: {1}", e.line + 1, message)}

webapp/src/errorList.tsx
```diff
@@ -43,30 +43,26 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
...
-                <div className="errorListInner">
-                    {errorListContent}
+                }
+            </div> )
+
+        return (
+            <div className={`errorList ${isCollapsed ? 'errorListSummary' : ''}`} hidden={!errorsAvailable}>
+                <div className="errorListHeader" role="button" onClick={this.onCollapseClick}>
+                    <h4>Problems</h4>
+                    <div className="ui red circular label countBubble">{errors?.length}</div>
```
pelikhan, line 60: errors is guaranteed to be defined here or it would have crashed earlier.

pelikhan:
I'm not sure i understand how this issue got fixed. Could you explain a bit more in the comments?

webapp/src/errorList.tsx
```diff
@@ -43,30 +43,26 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
...
-        const errorListHeader = <div className="errorListHeader" role="button" onClick={this.onCollapseClick}>
-            <h4>Problems</h4>
-            <div className="ui red circular label countBubble">{errors?.length}</div>
-            <div className="toggleButton"><sui.Icon icon={`chevron ${isCollapsed ? 'up' : 'down'}`} onClick={this.onCollapseClick} /></div>
-        </div>
-
-        const errorListContent = <div className="ui selection list">
-            {
-                (errors).map((e, index) =>
+        const errorListContent = (
```
pelikhan, line 46: don't compute UI elements eagerly, compute them if not collapsed

webapp/src/monaco.tsx
```diff
@@ -728,6 +728,9 @@ export class Editor extends toolboxeditor.ToolboxEditor {
             });
 
             if (monacoToolboxDiv) monacoToolboxDiv.style.height = `100%`;
+
+            // ensures all lines are displayed
+            this.editor.focus();
```
pelikhan, line 733: Does this mean that the monaco editor will steal focus when the user resizes the window?

darzu: Great question, we don't want to do that. Hristo, we might need to dig into .focus() to see what it's actually doing that fixes the sizing issue and maybe only call a subset of focus()'s behavior.

darzu: Also it's worth testing this with tab navigation. If you're tabbing around the page and this triggers does  the user lose their tab location?

hristoramirez: Ok I'll look into this more to see if there is another way

jwunderl:
```suggestion
            this.editor.layout();
```

@hristoramirez could you check if this fixes the problem? at a glance it looks like the issue is that the dimensions and offset are hardcoded on each of the divs (e.g. `div.monaco-scrollable-element` that's a parent of the individual lines), and that gets re-evaluated on focus: https://github.com/microsoft/monaco-editor/blob/master/monaco.d.ts#L2133

hristoramirez: @jwunderl I tried it, but it gets stuck in a loop while loading

hristoramirez: @jwunderl so far it looks like it fixes the problem

darzu: Might be worth checking that this doesn't slow down the editor resize speed

PR #7012, opened 2020-05-12
fast mode for testghpkgs

pelikhan:
In fast, mode the testghpkgs is not checking the extension tag number and simply looking at the status file directly.


pelikhan:
Adding fast mode (don't rerun if a tracker file is on disk) and repo filter name to rerun a particular set of repoes

PR #7013, opened 2020-05-13
Override codicon css for monaco breakpoints

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/3017

I eventually resorted to doing a manual diff of the css between the live site and master and found the selector that was causing the bug. This restores the old behavior of only showing the breakpoint you're hovering over.


pelikhan:
Yay!

PR #7014, opened 2020-05-13
fixed issue where clicking on error sends to line 1

hristoramirez:
There was an issue with the errors given while using javascript that would send to line 1
fixes: microsoft/pxt-microbit#3005

### DIFF
--- a/webapp/src/errorList.tsx
+++ b/webapp/src/errorList.tsx
@@ -49,7 +49,7 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
                 <div className="ui selection list">
                     {(errors).map((e, index) =>
                     <div className="item" key={errorKey(e)} role="button" onClick={createOnErrorMessageClick(e, index)}>
-                        {lf("Line {0}: {1}", e.line + 1, e.messageText)}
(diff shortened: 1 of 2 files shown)


hristoramirez:
@pelikhan I tried creating several errors and haven't been able to come across a multi line error. One thing though is choosing either to use the error.line or error.endLine number to lead the error message

pelikhan:
You can test ``error.endLine !== undefined && error.endColumn !== undefined`` for example

webapp/src/errorList.tsx
```diff
@@ -49,7 +49,7 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
                 <div className="ui selection list">
                     {(errors).map((e, index) =>
                     <div className="item" key={errorKey(e)} role="button" onClick={createOnErrorMessageClick(e, index)}>
-                        {lf("Line {0}: {1}", e.line + 1, e.messageText)}
+                        {lf("Line {0}: {1}", e.endLine + 1 || e.line + 1, e.messageText)}
```
pelikhan, line 52: what are the semantics of ``undefined + 1``? you probably want to use ``(e.endLine ?? e.endLine + 1)``

PR #7016, opened 2020-05-14
Check that runtime id is still current

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/3012

We were getting interleaved evals that were consuming each other's resumes. Fixed by checking to make sure that the runtime embedded in the binary matches the global one.

### DIFF
--- a/pxtcompiler/emitter/backjs.ts
+++ b/pxtcompiler/emitter/backjs.ts
@@ -221,7 +221,7 @@ s.pc = -1;
         }
         writeRaw(`
 while (true) {
-if (yieldSteps-- < 0 && maybeYield(s, step, r0) || runtime.id !== pxsim.runtime.id) return null;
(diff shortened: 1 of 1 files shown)


pxtcompiler/emitter/backjs.ts
```diff
@@ -221,7 +221,7 @@ s.pc = -1;
         }
         writeRaw(`
 while (true) {
-if (yieldSteps-- < 0 && maybeYield(s, step, r0)) return null;
+if (yieldSteps-- < 0 && maybeYield(s, step, r0) || runtime.id !== pxsim.runtime.id) return null;
```
mmoskal, line 224: can you do this? I think it would be slightly faster.

```suggestion
if (yieldSteps-- < 0 && maybeYield(s, step, r0) || runtime !== pxsim.runtime) return null;
```

PR #7023, opened 2020-05-14
Fix typo in flash checksum replacement

mmoskal:
This should fix maker build


mmoskal:
(bug introduced in https://github.com/microsoft/pxt/pull/6891)

PR #7026, opened 2020-05-14
Markdown rendering

pelikhan:
Fix preview rendering.


webapp/src/container.tsx
```diff
@@ -622,27 +622,30 @@ export class SideDocs extends data.Component<SideDocsProps, SideDocsState> {
         this.popOut = this.popOut.bind(this);
     }
 
+    private rootDocsUrl(): string {
+        return pxt.webConfig.docsUrl || '/--docs#';
```
jwunderl, line 626: Isn't pxt.webConfig.docsUrl sent from backend without the #? (As just `/--docs`)

pelikhan: great catch

PR #7032, opened 2020-05-14
More intellisense fixes

darzu:
Fixes: https://github.com/microsoft/pxt-microbit/issues/2893
Fixes: https://github.com/microsoft/pxt-microbit/issues/2707
Fixes: https://github.com/microsoft/pxt-microbit/issues/2736
Fixes: https://github.com/microsoft/pxt-microbit/issues/2977
Fixes: https://github.com/microsoft/pxt-microbit/issues/3056

Lots of miscellaneous auto-complete improvements.
We better handle member completion and don't show bogus symbols when we can tell the type  we're completing on.
We make sure we at minimum can complete global symbols in Python.
We fix the ordering on some completions, like putting local variables higher.
We more correctly identify where a completion snippet should start (if it's qualified).
We correctly handle showing local variables after a parenthesis.
We recompute auto-complete after a ":".
Probably more little things..

Language service test enhancements:
- Support for .py files
- Support for more expressions like `some.fn(foo // foobar`
- Support for order sensativity

TODO:
https://github.com/microsoft/pxt-microbit/issues/3056

### DIFF
--- a/tests/language-service/languageservicerunner.ts
+++ b/tests/language-service/languageservicerunner.ts
@@ -107,21 +107,18 @@ function getTestCases() {
 
                 const lineWithoutCommment = line.substring(0, commentIndex);
 
-                const dotPosition = lineWithoutCommment.lastIndexOf(".");
-                const isCompletionAtDot = dotPosition !== -1
-
-                let relativeCompletionPosition;
-                if (!isCompletionAtDot) {
-                    // find last non-whitespace character
-                    for (let i = lineWithoutCommment.length - 1; i >= 0; i--) {
-                        relativeCompletionPosition = i
-                        if (lineWithoutCommment[i] !== " ") {
-                            break
-                        }
+                // find last non-whitespace character
+                let lastNonWhitespace: number;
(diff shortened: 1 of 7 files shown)


darzu:
fixes rest of: microsoft/pxt-microbit#2707

sorting is still a little off

darzu:
Fixed: https://github.com/microsoft/pxt-microbit/issues/2736

tests/language-service/languageservicerunner.ts
```diff
@@ -123,13 +140,15 @@ function getTestCases() {
     return testCases;
 }
 
+const fileName = (isPython: boolean) => isPython ? "main.py" : "main.ts"
```
pelikhan, line 143: This is not new but what about other files, does it matter?

darzu: This is just for shimming the compile options. We assume each test case can be in a single file, which I don't see any need to support more. We have a "test package" already that lets us add arbitrary amounts of support files.

PR #7033, opened 2020-05-14
Handling invalid github token

pelikhan:
It seems that GitHub is expeiring OAuth tokens (maybe there is a limit of clients) or the user might just invalidate it too. SInce any github request may fail, we handle this at the request level and simply erase the token from the browser.
This change also guarantees that loading a tutorial does **not**rely on a token since that token might be bogus.
This fix is not bullet proof, i should really be wrapping every github call with a retry without token that goes through the proxy.
Added tick for errors to track them.

### DIFF
--- a/pxtlib/github.ts
+++ b/pxtlib/github.ts
@@ -109,29 +109,37 @@ namespace pxt.github {
         loadPackageAsync(repopath: string, tag: string): Promise<CachedPackage>;
     }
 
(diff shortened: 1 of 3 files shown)


riknoll:
LGTM but where are you deleting the token?

pelikhan:
The handleGithubNetworkError clears the token on the webapp side.

PR #7038, opened 2020-05-15
fix width of tutorial message on mobile

jwunderl:
fix mobile width, the change to the selector made the normal one overpower the mobile only one

curr:

image

fixed:

image


shakao:
would double check that you don't need `width: 100%` for very short messages without a hint

PR #7044, opened 2020-05-18
Add pxtarget.compile.shimRenames

mmoskal:


pxtcompiler/emitter/emitter.ts
```diff
@@ -2134,6 +2134,9 @@ ${lbl}: .short 0xffff
                 return emitExpr(args[0])
             }
 
+            if (opts.target.shimRenames && U.lookup(opts.target.shimRenames, nm))
+                nm = opts.target.shimRenames[nm]
```
pelikhan, line 2138: nit, all in 1 shot?

    nm = opts.target?.shimRenames[nm] || nm

mmoskal: I think that would protect against `opts.target` being `null`, not `opts.target.shimRenames`

mmoskal: I guess there's `opts.target.shimRenames.?[nm]`...

mmoskal: or something

PR #7045, opened 2020-05-19
Mobile and tablet editor toolbar fixes

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/1919 (this ones fixed by adding it to fullscreen view)
fix https://github.com/microsoft/pxt-microbit/issues/2260
fix https://github.com/microsoft/pxt-microbit/issues/2487
and fixes a bunch of other minor resizing bugs

Here's some of the resizing (pretty big so hidden behind details):

<details>

editor-toolbar

ezgif com-resize
</details>

side notes;

* this reverts a change to the breakpoints for mobile vs tablet; that one was to make landscape iphone x count as a mobile device still, but that ends up causing most all tablets to be treated either as mobile or desktop
* There's a bit more code I can remove in `editortoolbar.tsx` and a few more redundant rules in `common.less` - that cleanup was a large part of the casualties when my I needed to wipe and reinstall my computer after it crashed. Figured it's probably better just to finish this up and do a follow up PR for that.
* currently there's a very slight diff in tablet / mobile view on minecraft (left new, right is current) -- can special case back if we're worried about it:
tablet adds back redo button

<details>

image

</details>

mobile no zoom but adds back redo

<details>

image

</details>

* arcade has some target specific button recoloring that's overriding the stop / pause button so can fix that later if we want; it's obvious enough against the blocks that I'm not sure it matters though. here's arcade on iphone5:

<details>

image

</details>


webapp/src/blocks.tsx
```diff
@@ -561,8 +561,6 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         const blocklyDiv = this.getBlocksEditorDiv();
         // Position blocklyDiv over blocklyArea.
         if (blocklyDiv && this.editor) {
-            blocklyDiv.style.width = blocklyArea.offsetWidth + 'px';
```
jwunderl, line 564: @shakao does this look alright / reasonable (along with changes in blockly-core.less: https://github.com/microsoft/pxt/pull/7045/files#diff-f2432c11e75a6cd6a1d5f64674a06542R23)? It was the quick workaround for some resizing bugs (so I didn't have to duplicate the size logic all over the place), but it feels like it must have been done this way for a reason..

Easy enough to fix and maintain this if needed (e.g. if svgResize peeks at the width maybe? haven't pulled blockly on this machine to check quickly)

shakao: this looks good to me! accessing the .width property will still return the correct measurement with this moved to css

webapp/src/editortoolbar.tsx
```diff
@@ -221,20 +226,14 @@ export class EditorToolbar extends data.Component<ISettingsProps, {}> {
     }
 
     renderCore() {
-        const { home, tutorialOptions, hideEditorFloats, collapseEditorTools, projectName, compiling, isSaving, simState, debugging, header } = this.props.parent.state;
-
-        if (home) return <div />; // Don't render if we're in the home screen
```
shakao, line 226: why remove this?

jwunderl: The component is not shown in app.tsx when in the home screen (it's toggled by `showEditorToolbar` which is false if `inHome`)

PR #7048, opened 2020-05-19
display runtime errors in the error list

hristoramirez:
Display simple runtime error in the error list
fixes: microsoft/pxt-microbit#3089
runtime-to-errorlist

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -782,7 +782,8 @@ export class ProjectView
                 // TODO: start debugging session
                 // TODO: user friendly error message
                 //core.warningNotification(lf("Program Error: {0}", brk.exceptionMessage));
-                this.textEditor.onExceptionDetected(brk)
+                //this.currentEditor.onExceptionDetected(brk)
+                this.editor.onExceptionDetected(brk)
             },
             highlightStatement: (stmt, brk) => {
                 if (this.state.debugging && !simulator.driver.areBreakpointsSet() && brk && !brk.exceptionMessage) {
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -781,7 +781,8 @@ export class ProjectView
             orphanException: brk => {
                 // TODO: start debugging session
                 // TODO: user friendly error message
-                core.warningNotification(lf("Program Error: {0}", brk.exceptionMessage));
+                //core.warningNotification(lf("Program Error: {0}", brk.exceptionMessage));
+                this.textEditor.onExceptionDetected(brk)
             },
             highlightStatement: (stmt, brk) => {
                 if (this.state.debugging && !simulator.driver.areBreakpointsSet() && brk && !brk.exceptionMessage) {
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -783,7 +783,7 @@ export class ProjectView
                 // TODO: user friendly error message
                 //core.warningNotification(lf("Program Error: {0}", brk.exceptionMessage));
                 //this.currentEditor.onExceptionDetected(brk)
-                this.editor.onExceptionDetected(brk)
+                this.editor?.onExceptionDetected(brk)
             },
             highlightStatement: (stmt, brk) => {
                 if (this.state.debugging && !simulator.driver.areBreakpointsSet() && brk && !brk.exceptionMessage) {
--- a/webapp/src/errorList.tsx
+++ b/webapp/src/errorList.tsx
@@ -6,11 +6,13 @@ import * as sui from "./sui";
 export interface ErrorListProps {
     onSizeChange: () => void,
     listenToErrorChanges: (key: string, onErrorChanges: (errors: pxtc.KsDiagnostic[]) => void) => void,
+    listenToExceptionChanges: (handlerKey: string, handler: (exception: pxsim.DebuggerBreakpointMessage) => void) => void,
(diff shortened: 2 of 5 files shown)


webapp/src/app.tsx
```diff
@@ -781,7 +781,8 @@ export class ProjectView
             orphanException: brk => {
                 // TODO: start debugging session
                 // TODO: user friendly error message
-                core.warningNotification(lf("Program Error: {0}", brk.exceptionMessage));
+                //core.warningNotification(lf("Program Error: {0}", brk.exceptionMessage));
+                this.textEditor.onExceptionDetected(brk)
```
pelikhan, line 785: don't assume the user is in the textEditor, use currentEditor.

webapp/src/errorList.tsx
```diff
@@ -6,11 +6,13 @@ import * as sui from "./sui";
...
     onSizeChange: () => void,
     listenToErrorChanges: (key: string, onErrorChanges: (errors: pxtc.KsDiagnostic[]) => void) => void,
+    listenToExceptionChanges: (handlerKey: string, handler: (exception: pxsim.DebuggerBreakpointMessage) => void) => void,
     goToError: (error: pxtc.KsDiagnostic) => void
 }
 export interface ErrorListState {
-    isCollapsed: boolean
+    isCollapsed: boolean,
     errors: pxtc.KsDiagnostic[],
+    exception: pxsim.DebuggerBreakpointMessage
```
pelikhan, line 15: optional field ``exception?``?

webapp/src/errorList.tsx
```diff
@@ -92,4 +98,10 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
             isCollapsed: errors?.length == 0 || this.state.isCollapsed
         })
     }
+
+    onExceptionDetected(exception: pxsim.DebuggerBreakpointMessage) {
+        this.setState({
+            exception: exception
```
pelikhan, line 104: shorten to ``exception`` -- typescript does the magic

webapp/src/monaco.tsx
```diff
@@ -603,18 +605,29 @@ export class Editor extends toolboxeditor.ToolboxEditor {
                             setInsertionSnippet={this.setInsertionSnippet}
                             parent={this.parent} />
                     </div>
-                    {showErrorList ? <ErrorList onSizeChange={this.resize} listenToErrorChanges={this.listenToErrorChanges} goToError={this.goToError} /> : undefined}
+                    {showErrorList ? <ErrorList onSizeChange={this.resize} listenToErrorChanges={this.listenToErrorChanges}
```
pelikhan, line 608: ``showErrorList &&`` ...

webapp/src/errorList.tsx
hristoramirez: yeah thats probably more appropriate

webapp/src/app.tsx
hristoramirez: since the errorlist is currently only available in the text editor would it be safe to only call this if its the current active editor and use the previous call for the other editors for the time being?

pelikhan: This should be set on all editors and have a base function that does nothing

webapp/src/app.tsx
```diff
@@ -781,7 +781,9 @@ export class ProjectView
             orphanException: brk => {
                 // TODO: start debugging session
                 // TODO: user friendly error message
-                core.warningNotification(lf("Program Error: {0}", brk.exceptionMessage));
+                //core.warningNotification(lf("Program Error: {0}", brk.exceptionMessage));
+                //this.currentEditor.onExceptionDetected(brk)
+                this.editor.onExceptionDetected(brk)
```
pelikhan, line 786: check that editor is set 
```
this.editor?.onException...
```

webapp/src/errorList.tsx
```diff
@@ -6,11 +6,13 @@ import * as sui from "./sui";
 export interface ErrorListProps {
     onSizeChange: () => void,
     listenToErrorChanges: (key: string, onErrorChanges: (errors: pxtc.KsDiagnostic[]) => void) => void,
+    listenToExceptionChanges: (handlerKey: string, handler: (exception: pxsim.DebuggerBreakpointMessage) => void) => void,
     goToError: (error: pxtc.KsDiagnostic) => void
 }
 export interface ErrorListState {
-    isCollapsed: boolean
+    isCollapsed: boolean,
     errors: pxtc.KsDiagnostic[],
```
pelikhan, line 14: this could be optional too

webapp/src/errorList.tsx
```diff
@@ -45,22 +50,26 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
...
-                <div className="ui selection list">
-                    {(errors).map((e, index) =>
-                    <div className="item" key={errorKey(e)} role="button" onClick={createOnErrorMessageClick(e, index)}>
-                        {lf("Line {0}: {1}", (e.endLine) ? e.endLine + 1 : e.line + 1, e.messageText)}
-                    </div>)
-                    }
-                </div>
-            )
+            if (exception) {
+                errorListContent = <div className="exceptionMessage">{lf(exception.exceptionMessage)}</div>
```
pelikhan, line 54: Use ``rlf(exception.exceptionMessage)`` for runtime generated error messages

PR #7050, opened 2020-05-19
Ensure name uniqueness for auto-complete and toolbox snippets

darzu:
when generating snippets for auto-complete  or toolbox, track when we introduce new definitions so that we can make sure we regenerate those snippets when names might have changed.

I believe this regression was introduced as a side effect of better caching of symbol info between intellisense calls. Really snippets are very stateful and so we need extra info to know when to invalidate them.

Fixes: https://github.com/microsoft/pxt-microbit/issues/3068
Fixes: https://github.com/microsoft/pxt-microbit/issues/3074

### DIFF
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -767,6 +767,7 @@ declare namespace ts.pxtc {
         combinedProperties?: string[];
         pyName?: string;
         pyQName?: string;
+        snippetAddsDefinitions?: boolean;
     }
 
     interface ApisInfo {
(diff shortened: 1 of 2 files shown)


darzu:
> Is there any way to add tests for those?

I started making one, but then I realize it wouldn't have caught the last regression. It's possible to create some tests but they'd only guard against certain regressions and wouldn't help new feature development much because in this case it'd be very specific to name-uniqueness. Didn't seem like it was worth it. Still keeping it in mind though and if I think of a better or more general set of tests I'd like to add them

PR #7053, opened 2020-05-19
Embed icon font in svgs for screenshots

shakao:
- Embed semantic CSS in SVG for icons
- Add pxt-renderer class to generated SVG

We could also strip out the font and include just that in the CSS, but it felt like including the whole thing was better than doing a lot of parsing through the CSS file.

Fixes https://github.com/microsoft/pxt-microbit/issues/3084


pelikhan:
Semantic css is hugesque.... maybe our build can produce a font.css  or screenshot.css to be included in those.

PR #7056, opened 2020-05-20
Adding detection for ios 13 ipads

riknoll:
Should fix https://github.com/microsoft/pxt-microbit/issues/2601

I can't really test it locally but this is the same check we use in monaco for determining if we're on an ipad. I tested that it does return true in the ios 13.3 ipad simulator.


pelikhan:
who has the ipad? :)

abchatra:
I have one and I also have browserstack.

PR #7057, opened 2020-05-20
use editor tools colors instead of red / green, padding in tablet view

jwunderl:
quick changes we discussed on teams:

image

move fullscreen button to top in tablet mode, make the collapse / expand show up in same color as other editor tools, and a little space around download button in tablet view

### DIFF
--- a/theme/common.less
+++ b/theme/common.less
@@ -1092,6 +1092,12 @@ p.ui.font.small {
         transform: translateX(0%)!important;
         left: -2rem;
     }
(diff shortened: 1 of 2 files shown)


jwunderl:
Oh hm, quick tangent - I just realized this button overlays on top of the error list now: 

image

image

Will think about this, probably a simple solution we can come up with in design meeting

abchatra:
File an issue to track

PR #7059, opened 2020-05-20
Replace 'devdocs' JS links

### LINKED ISSUE #7058: Remove links to devdocs.io from our docs
**Describe the bug**
We link to https://devdocs.io/ in some of our JS docs, but the site does not look like it's being maintained.

Eg:
https://makecode.microbit.org/javascript/statements
https://makecode.microbit.org/javascript/operators

ganicke:
The links to `devdocs.io` pages have equivalent pages at `https://developer.mozilla.org`. Those page links are replaced with the MDN page links. If linking to MDN is not appropriate, then we'll have to figure something else out.

Closes #7058


shakao:
looks great to me! i think MDN is pretty reliable, i feel like it should be all right to link to?

PR #7066, opened 2020-05-21
Special case range in for-loops for symbol info

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/3046

We don't treat `range()` as a symbol when it's in a for-loop because we convert it to the TS style of for-loops (not a function call). Adding a special case to the language service to make sure that you get the same hover text in a for-loop as you do in standalone calls to `range()`

### DIFF
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -880,6 +880,12 @@ namespace ts.pxtc.service {
             if (isSymbol && !opts.syntaxInfo.symbols?.length) {
                 const possibleKeyword = getWordAtPosition(v.fileContent, v.position);
                 if (possibleKeyword) {
+                    // In python if range() is used in a for-loop, we don't convert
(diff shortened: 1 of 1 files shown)


pxtcompiler/emitter/service.ts
```diff
@@ -877,27 +877,37 @@ namespace ts.pxtc.service {
...
-                    const help = getHelpForKeyword(possibleKeyword.text, isPython);
-
-                    if (help) {
-                        opts.syntaxInfo.auxResult = {
-                            documentation: help,
-                            displayString: displayStringForKeyword(possibleKeyword.text, isPython),
-                        };
-                        opts.syntaxInfo.beginPos = possibleKeyword.start;
-                        opts.syntaxInfo.endPos = possibleKeyword.end;
+                    if (isPython && possibleKeyword.text === "range") {
```
darzu, line 883: Maybe add a comment about why we're special casing 'range' here

PR #7068, opened 2020-05-21
Show deprecated blocks if they're being used in a tutorial

livcheerful:
fix https://github.com/microsoft/pxt-microbit/issues/3096

I tested this with the random-range deprecated block, and the radio.receiveString() because they seem to be deprecated in different ways?

### DIFF
--- a/webapp/src/blocks.tsx
+++ b/webapp/src/blocks.tsx
@@ -1229,7 +1229,8 @@ export class Editor extends toolboxeditor.ToolboxEditor {
     }
 
     private filterBlocks(subns: string, blocks: toolbox.BlockDefinition[]) {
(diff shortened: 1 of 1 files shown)


riknoll:
Yes, that's correct. Those blocks should only show up as shadows

livcheerful:
@riknoll what is a shadow block?

riknoll:
Shadow blocks are the blocks you can't remove from inputs. In the toolbox, we use them for default values. You can place other blocks on top of them and they stay if you pull those blocks out again.

We generally try to make it so that you can't have "holes" in the program so that it always converts cleanly to text. Shadow blocks can't be removed so there will always be *something*. The exception is variable and array blocks, because those can take any type and we can't populate default values.

PR #7085, opened 2020-05-26
error list now displays the stacktrace for runtime errors

hristoramirez:
image

### DIFF
--- a/webapp/src/errorList.tsx
+++ b/webapp/src/errorList.tsx
@@ -118,7 +118,7 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
         return <div>
             {(exception.stackframes || []).map(sf =>
                 <div className="stackframe">
-                    {pxt.Util.rlf("at {0} ({1}:{2}:{3})", sf.funcInfo.functionName, sf.funcInfo.fileName, sf.funcInfo.line + 1, sf.funcInfo.column + 1)}
+                    {lf("at {0} ({1}:{2}:{3})", sf.funcInfo.functionName, sf.funcInfo.fileName, sf.funcInfo.line + 1, sf.funcInfo.column + 1)}
                 </div>)
             }
         </div>;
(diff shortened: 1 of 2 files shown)


webapp/src/errorList.tsx
```diff
@@ -108,4 +113,16 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
             exception
         })
     }
+
+    private generateStackTraces(exception: pxsim.DebuggerBreakpointMessage) {
+        return (
```
pelikhan, line 118: parenthesis not needed?

webapp/src/errorList.tsx
```diff
@@ -108,4 +113,16 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
...
+            <div>
+                {(exception.stackframes).map(sf =>
```
pelikhan, line 120: make sure "stackFrames" is defined

webapp/src/errorList.tsx
```diff
@@ -108,4 +113,16 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
...
+                    <div className="stackframe">
```
pelikhan: Clickeable?

webapp/src/errorList.tsx
```diff
@@ -108,4 +113,14 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
             exception
         })
     }
+
+    private generateStackTraces(exception: pxsim.DebuggerBreakpointMessage) {
+        return <div>
+            {(exception.stackframes || []).map(sf =>
+                <div className="stackframe">
+                    {pxt.Util.rlf("at {0} ({1}:{2}:{3})", sf.funcInfo.functionName, sf.funcInfo.fileName, sf.funcInfo.line + 1, sf.funcInfo.column + 1)}
```
pelikhan, line 121: Use lf

pelikhan: funcinfo guaranteed?

webapp/src/errorList.tsx
hristoramirez: Yeah I'm going to add that in the next pr

webapp/src/errorList.tsx
hristoramirez: It looks like it's a required property of the StackFrameInfo interface. Also, since funcInfo is of any type is it safe that access the properties like I'm doing right now?

PR #7089, opened 2020-05-27
Invite button

pelikhan:
- [x] use regular button fix for https://github.com/microsoft/pxt-microbit/issues/3168
- [x] only show button (1) if not blocks, (2) if repo is owned, (3) if desktop fix for https://github.com/microsoft/pxt-arcade/issues/1906
- [x] always "pull changes" text, we have space

Blocks is not quite ready for collaboration so hiding that button for now.


webapp/src/gitjson.tsx
```diff
@@ -656,7 +656,8 @@ class GithubComponent extends data.Component<GithubProps, GithubState> {
         const pr: pxt.github.PullRequest = this.getData("pkg-git-pr:" + header.id)
         const showPr = pr !== null && (gs.isFork || !master);
         const showPrResolved = showPr && pr && pr.number > 0;
-        const showPrCreate = showPr && pr && pr.number <= 0
+        const showPrCreate = showPr && pr && pr.number <= 0;
+        const isOwner = user && user.id === githubId.owner;
```
pelikhan, line 660: this does not cover all cases but good enough approximation for now.

PR #7091, opened 2020-05-27
Disable performance reporting if APIs not found

shakao:
Haven't tested this on iOS (no BrowserStack access yet), but this only runs the performance measurements if performance.mark is defined. mark/measure/getEntriesByName are all enabled from iOS 11+

Fixes https://github.com/microsoft/pxt-microbit/issues/3038

### DIFF
--- a/docfiles/pxtweb/cookieCompliance.ts
+++ b/docfiles/pxtweb/cookieCompliance.ts
@@ -111,7 +111,7 @@ namespace pxt {
             stats.milestones.push([msg, time])
         }
         export function init() {
-            enabled = performance && !!performance.mark;
+            enabled = performance && !!performance.mark && !!performance.measure;
             if (enabled) {
                 performance.measure("measure from the start of navigation to now")
                 let navStartMeasure = performance.getEntriesByType("measure")[0]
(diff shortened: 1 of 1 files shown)


docfiles/pxtweb/cookieCompliance.ts
```diff
@@ -109,16 +111,18 @@ namespace pxt {
             stats.milestones.push([msg, time])
         }
         export function init() {
-            performance.measure("measure from the start of navigation to now")
-            let navStartMeasure = performance.getEntriesByType("measure")[0]
-            startTimeMs = navStartMeasure.startTime
-            // startTimeMs = performance.timing.navigationStart
+            enabled = performance && !!performance.mark;
+            if (enabled) {
+                performance.measure("measure from the start of navigation to now")
```
pelikhan, line 116: test for performance.measure as well?

docfiles/pxtweb/cookieCompliance.ts
```diff
@@ -109,16 +111,18 @@ namespace pxt {
             stats.milestones.push([msg, time])
         }
         export function init() {
-            performance.measure("measure from the start of navigation to now")
-            let navStartMeasure = performance.getEntriesByType("measure")[0]
-            startTimeMs = navStartMeasure.startTime
-            // startTimeMs = performance.timing.navigationStart
+            enabled = performance && !!performance.mark;
```
shakao, line 114:
```suggestion
            enabled = performance && !!performance.mark && !!performance.measure;
```

PR #7095, opened 2020-05-28
added a debugger suggestion in the error list for runtime errors

hristoramirez:
Added the debugger suggestion from the design meeting on 5/27
fixes: microsoft/pxt-microbit#3095
pr_debugSuggest

### DIFF
--- a/theme/errorList.less
+++ b/theme/errorList.less
@@ -77,9 +77,10 @@
             padding-top: 0.5em;
             padding-bottom: 0.5em;
 
-            .debuggerSuggestion {
+            .debuggerSuggestion.blue {
                 float: right;
                 margin-right: 1em;
+                color: @errorListDebugSuggestionColor;
                 cursor: pointer;
 
                 &:hover {
--- a/theme/errorList.less
+++ b/theme/errorList.less
@@ -76,6 +76,21 @@
             padding-left: 1em;
             padding-top: 0.5em;
             padding-bottom: 0.5em;
+
+            .debuggerSuggestion {
+                float: right;
+                margin-right: 1em;
+                cursor: pointer;
+
+                &:hover {
+                    opacity: 0.75;
+                }
+
+                .debug-icon.blue {
+                    margin-left: 0.5em;
+                    vertical-align: top;
+                }
+            }
         }
 
         .ui.selection.list.exception {
--- a/theme/errorList.less
+++ b/theme/errorList.less
@@ -26,32 +26,18 @@
             margin-left: 1em;
         }
 
-        .cornerIcons {
+        .toggleButton {
             float: right;
+            cursor: pointer;
+            position: relative;
+            vertical-align: middle;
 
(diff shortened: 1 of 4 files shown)


pelikhan:
Nice!

Can we simplify the wording? like “debug this problem”. “Try starting a debug session” is somewhat complicated.

abchatra:
_"Debug this project"_ is better.  

It doesn't look clickable and looks like part of the text. 
Does clicking on the whole line launches the debugger? 
I would prefer something like link in _"Debug this project"_ which is clearly actionable.

pelikhan:
Might be worth investigating rendering it as a button so that it is clearly an action.

abchatra:
Discussed the button in the UI meeting and decided against it. Call of the action was better option here.

pelikhan:
It’s just more text on top of a lot of text.

webapp/src/errorList.tsx
```diff
@@ -57,7 +59,7 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
                         <div className="exceptionMessage">
                             {pxt.Util.rlf(exception.exceptionMessage)}
                             <span className="debuggerSuggestion" role="button" onClick={this.props.startDebugger}>
-                                {lf('Try starting a debug session')}
+                                <a>{lf("Debug this project")}</a>
```
pelikhan, line 62: this use of <a> look wrong

hristoramirez:
@abchatra I made the text and icon clickable because the icon by itself was too small of an area to click

abchatra:
LGTM

theme/errorList.less
```diff
@@ -75,11 +75,31 @@
         .exceptionMessage {
             padding-left: 1em;
             padding-top: 0.5em;
+            padding-bottom: 0.5em;
+
+            .debuggerSuggestion.blue {
```
pelikhan, line 80: Why is the class called blue?

theme/errorList.less
```diff
@@ -75,11 +75,31 @@
...
+                }
+            }
         }
 
-        .stackframe {
-            padding-left: 2em;
-            padding-top: 0.25em;
+        .ui.selection.list.exception {
+                margin-top: 0;
+            .stackframe.item {
```
pelikhan, line 99: already nestled under .stackframe

theme/errorList.less
hristoramirez: It was residual from when I was trying to turn the icon and text blue, but I removed it since its unnecessary

darzu:
Would it make sense to have a (?) info/help button? I'm not sure "Debug this project" will make sense to beginners.
@abchatra , @pelikhan

darzu:
Also can you show what it looks like when the error text is really long and wraps to multiple lines?

hristoramirez:
@darzu I placed a sample text that would overflow and this it looks like
runtime-over

PR #7106, opened 2020-05-29
account for error list in portrait simulator positioning

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/3106

woo-resizing

Few notes on errors I was thinking about while doing this (@hristoramirez maybe interesting?)

* we'll need to think about this for minecraft (transparent editor tools) as well; could just apply the same extra padding to the editor tools when they're transparent easily enough, but that'll need a design meeting so I left it out of this PR, maybe outside scope of internship as it wouldn't ship for a while?
* the 2 second debounce feels alright to me when adding the error (longer than I would personally like but I agree it's probably better for kids), but it's pretty painful when you fix an error as it feels like you're doing wrong / 2 seconds is long enough to second guess yourself. Maybe if we could keep the squiggle after .5 seconds, and delay showing the error list for the rest of the time or something? Not sure

### DIFF
--- a/pxteditor/editor.ts
+++ b/pxteditor/editor.ts
@@ -37,6 +37,11 @@ namespace pxt.editor {
         fileHistory: FileHistoryEntry[];
     }
 
+    export enum ErrorListState {
+        HeaderOnly = "errorListHeader",
+        Expanded = "errorListExpanded"
+    }
+
     export interface IAppProps { }
     export interface IAppState {
         active?: boolean; // is this tab visible at all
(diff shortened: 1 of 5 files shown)


hristoramirez:
I'd like to take a look at the minecraft transparency tools if I get the time. But yeah I like the idea of keeping the squiggles for immediate feedback because it's difficult coming up with a one size fits all solution. I can try looking into it

pelikhan:
We can fiddle around as much as we want we won't get the perfect timing because our users have vastly different keyboarding skills.

PR #7111, opened 2020-05-29
Custom tracking file for initializing telemetry without cookie

shakao:
The idea is we can include this file in static docs pages (instead of tracking.html). The page will then manually call "loadAppInsights(true/false)" depending on whether it wants the cookie or not. Also bundles a tick event helper--not sure if saving that on the window is the right place?

### DIFF
--- a/docfiles/customtracking.html
+++ b/docfiles/customtracking.html
@@ -1,0 +1,69 @@
+<script type="text/javascript">
+    // Similar to tracking.html and apptracking.html except we load appinsights but
+    // do not initialize. The individual pages will call loadAppInsights with custom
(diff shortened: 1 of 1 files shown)


abchatra:
Do we need to duplicate the entire file? Can we pass false as a parameter in-place of includeCookie?

abchatra:
As long as we don't drop the cookie we are good right?

shakao:
the problem is both existing files call pxt.initAnalyticsAsync which automatically initializes appinsights with a cookie unless it sees sandbox=1 / #sandbox / etc in the url. i didn't want to change the existing files because we use them in a bunch of places already

PR #7118, opened 2020-05-31
hide fullscreen button in mobile view instead of debug btn

jwunderl:
close https://github.com/microsoft/pxt-microbit/issues/3200

Starting as a draft as adding back fullscreen button in mobile was a request when it was fullscreen & restart; I think showing debug button is more important, just want to confirm.

### DIFF
--- a/theme/common.less
+++ b/theme/common.less
@@ -1103,12 +1103,6 @@ p.ui.font.small {
         left: -2rem;
     }
 
-    #filelist .simtoolbar > .buttons {
-        > .debug-button {
-            display: block !important;
-        }
-    }
-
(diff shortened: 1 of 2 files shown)


jwunderl:
Ah, yes, no difference on >= tablet size, and here's mobile:

mobile-w-debug

Maybe worth looking at the sui mobile breakpoint in the future and consider making mobile more as 'portrait mobile' instead of trying to include landscape phones -- e.g. there's room for the name field on mobile landscape if we want it. But that's a change with wide impact / would be worth sitting in beta first.

pelikhan:
Let’s just do a minimal change here

PR #7122, opened 2020-06-01
Show deprecated blocks in Python/JS tutorial

livcheerful:
fixes https://github.com/microsoft/pxt-microbit/issues/3198#event-3389041242

related to https://github.com/microsoft/pxt/commit/2e9fde5a3a9f6900d652442d4c031b1995e0344f

Show the deprecated blocks if we're in a tutorial in Python or JS

### DIFF
--- a/webapp/src/toolboxeditor.tsx
+++ b/webapp/src/toolboxeditor.tsx
@@ -119,7 +119,7 @@ export abstract class ToolboxEditor extends srceditor.Editor {
         let that = this;
 
         function filterNamespaces(namespaces: [string, pxtc.CommentAttrs][]) {
-            return namespaces.filter(([, md]) => !(md.deprecated && that.parent.state.tutorialOptions == undefined)&& (isAdvanced ? md.advanced : !md.advanced));
+            return namespaces.filter(([, md]) => !(md.deprecated && that.parent.state.tutorialOptions == undefined) && (isAdvanced ? md.advanced : !md.advanced));
         }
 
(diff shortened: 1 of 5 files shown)


webapp/src/toolboxeditor.tsx
```diff
@@ -118,7 +119,7 @@ export abstract class ToolboxEditor extends srceditor.Editor {
         let that = this;
 
         function filterNamespaces(namespaces: [string, pxtc.CommentAttrs][]) {
-            return namespaces.filter(([, md]) => !md.deprecated && (isAdvanced ? md.advanced : !md.advanced));
+            return namespaces.filter(([, md]) => !(md.deprecated && that.parent.state.tutorialOptions == undefined)&& (isAdvanced ? md.advanced : !md.advanced));
```
riknoll, line 122: I think this should maybe just get pulled into a method, i.e. `this.parent.inTutorial()`

jwunderl, line 122:
```suggestion
            return namespaces.filter(([, md]) => !(md.deprecated && that.parent.state.tutorialOptions == undefined) && (isAdvanced ? md.advanced : !md.advanced));
```

to fix lint ~

livcheerful: totally, I'll do that!

livcheerful:
> LGTM, just a refactor. Also I think you might have committed using the wrong email address

PR #7125, opened 2020-06-02
Update partner approval message

ganicke:
The current message just says:

`"The content below is provided by partner."`

image

Here, I've **only** added the **"a"** as a minimum to update the message.

`"The content below is provided by a partner."`

However, the ask from https://github.com/microsoft/pxt-microbit/issues/3207 is to have something more like:

image

This could be done through use of a macro like `@approver@` or something like that.

```
 <div class="content">
   <h3 class="header">
       Approved content
   </h3>
   <p>
       The content below is provided by a partner and approved by the @approver@.
   </p>
</div>
```

### DIFF
--- a/docfiles/package.html
+++ b/docfiles/package.html
@@ -35,7 +35,7 @@
                     Approved content
                 </h3>
                 <p>
-                    The content below is provided by partner.
+                    The content below is provided by a partner.
                 </p>
             </div>
         </div>


pelikhan:
@approver@ macro would require cloud backend update.

Jaqster:
I don't think this is needed - "provided by a partner" is fine.

ganicke:
OK, approve the **"a"** and we'll be done with it.

PR #7128, opened 2020-06-03
Fix WindowsRT download

pelikhan:
This PR fixes download in Windows10 app. Tested also that webusb still works. The heart of the issue was that we running cmds.init before configurating DAPLink.

- [x] disable webusb in electron
- [x] rename mk*PacketIOAsync to relevant names to help with stacktraces
- [x] removed drive deployment in WinRT - this is dead code.
- [x] ignore .simstate.json conflict state
- [x] initialize commands **after** loading extensions -- since micro:bit loads DAP extension.
- [x] always disconnect before running cmds.init
- [x] avoid ping2s animation, does not work in Edge

### DIFF
--- a/cli/hid.ts
+++ b/cli/hid.ts
@@ -127,7 +127,7 @@ export function hf2ConnectAsync(path: string, raw = false) {
         }
 
         return pxt.usb.pairAsync()
-            .then(() => pxt.usb.mkPacketIOAsync())
+            .then(() => pxt.usb.mkUSBPacketIOAsync())
             .then(io => new HF2.Wrapper(io))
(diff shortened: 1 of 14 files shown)


abchatra:
Will review. Probably you need to test Adafruit app with these changes.

abchatra:
Also big thank you!

pelikhan:
Adafruit mostly works. There is some weird error after flashing in winrt but mostly it works. This will need to be revisted when Adafruit gets upgraded.

abchatra:
I think this broke serial data from the webusb. It doesn't work always now and very spotty.

abchatra:
Above may be a false alarm. I can't find a repro.

PR #7130, opened 2020-06-04
disable oauth in pxt electron

jwunderl:
disable oauth in pxt electron for now (I'll fix it later, but have a few other issues that are more pressing / it's not the typical scenario for the offline build to need oauth)

I guess question if we should disable these in minecraft too? Haven't tested if they work but I don't think they will?


jwunderl:
@abchatra If you could port this one to stable I'll kick off the builds for testing tomorrow ~

abchatra:
Please disable all github in offline app. We will address all issues later. We don’t want to keep difference in experience. Main scenario for offline is no internet and github doesn’t work there.

abchatra:
Make this experience consistent with Windows app

abchatra:
Also testing cost we can’t pay now. Like we have all right error messages when we have no internet

PR #7134, opened 2020-06-04
[stable6.0] disable oauth in pxt electron

jwunderl:
port #7130 to stable

### DIFF
--- a/webapp/src/githubprovider.tsx
+++ b/webapp/src/githubprovider.tsx
@@ -70,6 +70,7 @@ export class GithubProvider extends cloudsync.ProviderBase {
         pxt.tickEvent("github.login.dialog");
         // auth flow if github provider is prsent
         const oAuthSupported = pxt.appTarget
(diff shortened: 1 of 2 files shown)


jwunderl:
@abchatra Also need this one to hide crowdin translate editor button

abchatra:
What is the issue with translate? It opens in browser right?

jwunderl:
The redirect link will likely be wrong / not load you in correctly, and the electron app will freeze on the open authentication page. (refreshing won't help, have to close and reopen)

abchatra:
There is a close button at the crowdin login page (kind of hidden). Did you try that?

jwunderl:
Yeah, that's just getting to refreshing the page currently in the app: 

stuck-in-translation

Hm actually don't know if I've checked that in the browser to confirm it's electron only bugginess

abchatra:
change looks good to me. If it is not working, we don't want to fix it for offline.

PR #7137, opened 2020-06-04
added tab navigation for the error list

hristoramirez:
fixes: microsoft/pxt-microbit#3139
tab_nav_2
tab_nav_hc2

### DIFF
--- a/theme/errorList.less
+++ b/theme/errorList.less
@@ -94,7 +94,7 @@
                 opacity: 0.75;
             }
 
-            .debug-icon.blue {
+            .debug-icon {
                 margin-left: 0.5em;
                 vertical-align: text-top;
             }
--- a/theme/errorList.less
+++ b/theme/errorList.less
@@ -75,7 +75,7 @@
                     background-color: @errorListOdd;
                 }
 
-                &.stackFrame {
+                &.stackframe {
                     padding-left: 2em;
                 }
             }
--- a/theme/errorList.less
+++ b/theme/errorList.less
@@ -34,6 +34,10 @@
             margin-left: 1em;
         }
 
+        &:hover > .toggleButton > i , &:focus > .toggleButton > i{
+            font-size: 1.5em;
+        }
+
         .toggleButton {
             float: right;
             position: relative;
--- a/theme/errorList.less
+++ b/theme/errorList.less
@@ -35,7 +35,7 @@
         }
 
         &:hover > .toggleButton > i , &:focus > .toggleButton > i{
-            font-size: 1.5em;
+            font-size: 1.4em;
         }
 
         .toggleButton {
--- a/theme/errorList.less
+++ b/theme/errorList.less
@@ -20,10 +20,6 @@
             cursor: pointer;
         }
 
-        &:focus {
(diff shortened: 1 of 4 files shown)


theme/errorList.less
```diff
@@ -7,6 +7,10 @@
     min-height: @errorListHeight;
     border-top: 2px @errorListBorderColor solid;
 
+    *:focus {
+        outline: none;
```
jwunderl, line 11: could we see a gif of how it looks w/o this line? The focus is pretty subtle in non high contrast mode, and blended in with the zebra striping a bit to my eyes

hristoramirez: Here's what it looks without it The one thing was that the outline was always persistent
tab_nav_outline

jwunderl: Hm, I suppose that's a bit aggressive when you're clicking on it (though I think I might like it for the errors themselves).

Maybe just a tiny bit darker for focus to differentiate it a bit more? Assuming you can do that without the contrast with the text becoming an issue

webapp/src/errorList.tsx
```diff
@@ -77,15 +75,15 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
 
         return (
             <div className={`errorList ${isCollapsed ? 'errorListSummary' : ''}`} hidden={!errorsAvailable}>
-                <div className="errorListHeader" role="button" onClick={this.onCollapseClick}>
+                <div className="errorListHeader" role="button" onClick={this.onCollapseClick} onKeyDown={sui.fireClickOnEnter} tabIndex={0}>
```
jwunderl, line 78: Oh woops almost forgot, these need aria labels to indicate what they do for screen readers; each div just needs a short description of what it'll do

```suggestion
                <div className="errorListHeader" role="button" aria-label={lf("{0} error list", isCollapsed ? lf("Expand") : lf("Collapse"))} onClick={this.onCollapseClick} onKeyDown={sui.fireClickOnEnter} tabIndex={0}>
```

(note that I just typed that out in the github comment window, sorry if there's a typo ~)

hristoramirez: Thanks I'll add those on and leave the outline on for the errors

hristoramirez:
@jwunderl for elements that contain text, do I add an aria-label with the same text as contained in the element?

jwunderl:
You don't need one if the text already fully describes what will happen

webapp/src/errorList.tsx
```diff
@@ -66,7 +64,7 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
                 errorListContent = (
                     <div className="ui selection list">
                         {(errors).map((e, index) =>
-                        <div className="item" key={errorKey(e)} role="button" onClick={createOnErrorMessageClick(e, index)}>
+                        <div className="item" key={errorKey(e)} role="button" onClick={createOnErrorMessageClick(e, index)} onKeyDown={sui.fireClickOnEnter} tabIndex={0}>
```
jwunderl, line 67: Maybe have a label that specifies that this will "go to error ", but thats a bit of a nitpick

hristoramirez:
@abchatra

PR #7140, opened 2020-06-04
Set enabled/disabled on child blocks, prevent unnecessary warnings

shakao:
Partial fix for https://github.com/microsoft/pxt-arcade/issues/1692

This prevents the warnings that pop up on disabled blocks. Real cause of the issue is when a big stack of blocks with many warnings overlaps another one, it triggers a chain of calls to bumpNeighbours that bumps the stack way down into the right corner. 

The other part of the fix will be on the blockly end to either tweak the bump functionality or just disable it in some cases--in general I feel it's better to have overlapping blocks than to move blocks that the user did not drag.

\+ small update to remove unnecessary getTopLevelParent function

### DIFF
--- a/pxtblocks/blocklycompiler.ts
+++ b/pxtblocks/blocklycompiler.ts
@@ -1887,6 +1887,15 @@ namespace pxt.blocks {
         };
     }
 
+    function setChildrenEnabled(block: Blockly.Block, enabled: boolean) {
+        block.setEnabled(enabled);
+        // propagate changes
+        var children = block.getDescendants(false);
+        for (var i = 0, child; (child = children[i]); i++) {
+            child.setEnabled(enabled);
+        }
+    }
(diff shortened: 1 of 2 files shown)


pxtblocks/blocklycompiler.ts
```diff
@@ -1887,6 +1887,15 @@ namespace pxt.blocks {
         };
     }
 
+    function setChildrenEnabled(block: Blockly.Block, enabled: boolean) {
+        block.setEnabled(enabled);
+        // propagate changes
+        var children = block.getDescendants(false);
+        for (var i = 0, child; (child = children[i]); i++) {
```
jwunderl, line 1894:
```suggestion
        const children = block.getDescendants(false);
        for (const child of children) {
```

to pass lint

PR #7148, opened 2020-06-08
fix extension search in electron app

jwunderl:
Fix extension search in electron app (and normal staticpkg, I would assume?)

it was hitting `/gh/owner/repo` instead of `/api/gh/owner/repo` in the electron app as it was using `relprefix`

### DIFF
--- a/pxtlib/emitter/cloud.ts
+++ b/pxtlib/emitter/cloud.ts
@@ -3,7 +3,7 @@ namespace pxt.Cloud {
     import Util = pxtc.Util;
 
     // hit /api/ to stay on same domain and avoid CORS
-    export let apiRoot = pxt.BrowserUtils.isLocalHost() || Util.isNodeJS ? "https://www.makecode.com/api/" : "/api/";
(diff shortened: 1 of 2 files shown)


abchatra:
Can you explain the fix more?

I only see  `forceLiveEndpoint: true` being  the relevant  code changes. Rest looks like style change.

abchatra:
Oh I see the braces.

jwunderl:
without `forceLiveEndpoint`, the github search will get set to use relprefix in the line I changed to use a safe navigation operator (?.) - that is, privateRequestAsync would make it `localhost:3232/gh/microsoft/pxt-neopixel`; this fails because the github search is an api call -- https://makecode.microbit.org/gh/microsoft/pxt-neopixel is a 404, where https://makecode.microbit.org/api/gh/microsoft/pxt-neopixel is correct.

You are right on the other two lines though, they were just the style change as I was about to change the line && a sanity on the order of operations for || and ternary operators

PR #7151, opened 2020-06-08
Always take screenshot when share button is clicked

shakao:
Fixes https://github.com/microsoft/pxt-arcade/issues/1916

### DIFF
--- a/webapp/src/share.tsx
+++ b/webapp/src/share.tsx
@@ -227,7 +227,6 @@ export class ShareEditor extends data.Component<ShareEditorProps, ShareEditorSta
     screenshotAsync = () => {
         return this.props.parent.requestScreenshotAsync()
             .then(img => {
-                console.log("reqd screenshot")
                 const st: ShareEditorState = { recordingState: ShareRecordingState.None, recordError: undefined };
                 if (img) st.screenshotUri = img;
                 else st.recordError = lf("Oops, screenshot failed. Please try again.")
--- a/webapp/src/share.tsx
+++ b/webapp/src/share.tsx
@@ -221,15 +221,17 @@ export class ShareEditor extends data.Component<ShareEditorProps, ShareEditorSta
         if (this.state.recordingState != ShareRecordingState.None) return;
 
         this.setState({ recordingState: ShareRecordingState.ScreenshotSnap, recordError: undefined },
-            () => {
-                this.props.parent.requestScreenshotAsync()
-                    .then(img => {
-                        const st: ShareEditorState = { recordingState: ShareRecordingState.None, recordError: undefined };
-                        if (img) st.screenshotUri = img;
-                        else st.recordError = lf("Oops, screenshot failed. Please try again.")
-                        this.setState(st);
-                    });
+            () => this.screenshotAsync());
+    }
 
+    screenshotAsync = () => {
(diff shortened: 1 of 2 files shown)


webapp/src/share.tsx
```diff
@@ -221,15 +221,17 @@ export class ShareEditor extends data.Component<ShareEditorProps, ShareEditorSta
...
         this.setState({ recordingState: ShareRecordingState.ScreenshotSnap, recordError: undefined },
-            () => {
-                this.props.parent.requestScreenshotAsync()
-                    .then(img => {
-                        const st: ShareEditorState = { recordingState: ShareRecordingState.None, recordError: undefined };
-                        if (img) st.screenshotUri = img;
-                        else st.recordError = lf("Oops, screenshot failed. Please try again.")
-                        this.setState(st);
-                    });
+            () => this.screenshotAsync());
```
riknoll, line 224:
```suggestion
            this.screenshotAsync);
```

webapp/src/share.tsx
```diff
@@ -221,15 +221,17 @@ export class ShareEditor extends data.Component<ShareEditorProps, ShareEditorSta
...
+    }
 
+    screenshotAsync = () => {
+        return this.props.parent.requestScreenshotAsync()
+            .then(img => {
+                console.log("reqd screenshot")
```
pelikhan, line 230: pxt.debug(...

riknoll, line 230: Was this log left in intentionally?

webapp/src/share.tsx
```diff
@@ -382,7 +384,11 @@ export class ShareEditor extends data.Component<ShareEditorProps, ShareEditorSta
                 // save project name if we've made a change change
                 p = this.props.parent.updateHeaderNameAsync(newProjectName);
             }
-            p.then(() => this.props.parent.anonymousPublishAsync(screenshotUri))
+            // if screenshots are enabled, always take one
+            if (targetTheme.simScreenshot && !screenshotUri) {
+                p = p.then(() => this.screenshotAsync());
```
riknoll, line 389:
```suggestion
                p = p.then(this.screenshotAsync);
```

webapp/src/share.tsx
```diff
@@ -382,7 +384,11 @@ export class ShareEditor extends data.Component<ShareEditorProps, ShareEditorSta
                 // save project name if we've made a change change
                 p = this.props.parent.updateHeaderNameAsync(newProjectName);
             }
-            p.then(() => this.props.parent.anonymousPublishAsync(screenshotUri))
+            // if screenshots are enabled, always take one
+            if (targetTheme.simScreenshot && !screenshotUri) {
```
pelikhan, line 388: we should not attach a screenshot "silently", the user should always be able to review it. This should be done before publishing.

webapp/src/share.tsx
shakao: oops nope, missed it while cleaning up

shakao:
hmm i don't want to add an extra click to the publishing flow though. this doesn't overwrite the user screenshot (or gif) if they have already taken one, this just ensures that all shared urls have SOME kind of image

pelikhan:
Ok. Make sure things still work when sharing a project with errors (simulator not running).

PR #7152, opened 2020-06-08
added clickable stackframes to error list

hristoramirez:
Made each individual stackframe clickable
Also removed workaround with creating the error click function in render
fixes: microsoft/pxt-microbit#2996
clickable_stacktrace

### DIFF
--- a/webapp/src/errorList.tsx
+++ b/webapp/src/errorList.tsx
@@ -157,7 +157,7 @@ class ErrorListItem extends React.Component<ErrorListItemProps, ErrorListItemSta
     render() {
         const {error, stackFrame, location} = this.props
 
-        const message = stackFrame ?  lf("line {0} at {1}", location.line + 1, stackFrame.funcInfo.functionName)
+        const message = stackFrame ? lf("at {0} line {1}", stackFrame.funcInfo.functionName, location.line + 1)
             : lf("Line {0}: {1}", error.endLine ? error.endLine + 1 : error.line + 1, error.messageText)
 
         return <div className={`item ${stackFrame ? 'stackframe' : ''}`} role="button"
--- a/webapp/src/errorList.tsx
+++ b/webapp/src/errorList.tsx
@@ -128,7 +128,7 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
 
                     if (!location) return null;
 
-                    return <ErrorListItem key={index} index={index} stackFrame={sf} location={location} revealError={this.onErrorMessageClick}/>
+                    return <ErrorListItem key={index} index={index} stackframe={sf} location={location} revealError={this.onErrorMessageClick}/>
                 })}
             </div>
         </div>;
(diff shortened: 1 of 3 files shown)


webapp/src/errorList.tsx
```diff
@@ -133,17 +107,70 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
...
+    constructor(props: ErrorListItemProps) {
+        super(props)
+
+        this.onErrorListItemClick = this.onErrorListItemClick.bind(this)
+    }
+
+    render() {
+        const {error, stackFrame, location} = this.props
+
+        const message = stackFrame ?  lf("line {0} at {1}", location.line + 1, stackFrame.funcInfo.functionName)
```
jwunderl, line 160:
```suggestion
        const message = stackFrame ? lf("at {0} line {1}", stackFrame.funcInfo.functionName, location.line + 1)
```

maybe to match how stack traces are usually shown? The current wording feels a bit off to me, as the "at" feels like it needs to come before the line number in the message

webapp/src/errorList.tsx
```diff
@@ -133,17 +107,70 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
...
+            : lf("Line {0}: {1}", error.endLine ? error.endLine + 1 : error.line + 1, error.messageText)
+
+        return <div className={`item ${stackFrame ? 'stackframe' : ''}`} role="button"
+                    onClick={this.onErrorListItemClick}
+                    onKeyDown={sui.fireClickOnEnter}
+                    aria-label={lf("Go to {0}", stackFrame ? 'stackframe' : 'error')}
```
jwunderl, line 166: Maybe include the message in the aria label - so same as current w/ message at end (aria label usually replaces message https://www.w3.org/TR/WCAG20-TECHS/ARIA8.html#ARIA8-ex1)

also nitpick: saying stackframe feels off as you're going to the line the stackframe originated from / not viewing a stack frame that currently exists like you would in the debugger. Maybe worth splitting into two messages / translations?

```suggestion
                    aria-label={lf("Go to {0}: {1}", stackFrame ? '' : 'error', message)}
```

webapp/src/errorList.tsx
```diff
@@ -133,17 +107,70 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
...
         </div>;
     }
 }
+
+interface ErrorListItemProps {
+    index: number;
+    error?: pxtc.KsDiagnostic;
+    stackFrame?: pxsim.StackFrameInfo;
+    location?: pxtc.LocationInfo;
+    revealError: (location: pxtc.LocationInfo, index: number) => void;
```
jwunderl, line 143: nitpick: I prefer to keep the required attributes grouped together so it's obvious when you look at the interface what you know exists; maybe bump this one up to be adjacent to index

hristoramirez:
@abchatra

PR #7157, opened 2020-06-09
Adjust tileset field to show correct tile in preview

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/1909

The issue was in the constructor, we were checking `text` to see if we should set the icon to the transparent tile, but `text` was always null. I dunno if the better fix is to figure out why it's null if we're expecting it to not be. Lemme know if you think I should change it :D


riknoll:
I'm guessing this probably also fixes the issue I filed yesterday: https://github.com/microsoft/pxt-arcade/issues/1939

riknoll:
Hmm... I'm guessing that the text is null because something changed internally in Blockly with the new upgrade. @shakao that's probably fine right?

livcheerful:
> I'm guessing this probably also fixes the issue I filed yesterday: [microsoft/pxt-arcade#1939](https://github.com/microsoft/pxt-arcade/issues/1939)

Yup, I just tested it and it's fixed

livcheerful:
I just realized this adds a bug where a small hamburger is always added to the tileset :\ gonna figure that out before I push these changes

livcheerful:
I added a change to filter out the default sprite :\ I don't like how it's hard coded but couldn't find a better way to do it, since we just fetch the value (which will just be the string sprite.food.smallBurger). If there's a better solution pleeeeease let me know

riknoll:
I don't think we should merge this with the hard coded fix.

Here is the line in the super-super class that is causing this bug:

https://github.com/microsoft/pxt-blockly/blob/develop/core/field_dropdown.js#L91

Easiest fix would be to override that `getOptions()` method in FieldTileset and make sure that if menuGenerator isn't set, it returns an array with just the transparent tile: 

https://github.com/microsoft/pxt/blob/master/pxtblocks/fields/field_tileset.ts#L166

livcheerful:
> I don't think we should merge this with the hard coded fix.
> 
> Here is the line in the super-super class that is causing this bug:
> 
> https://github.com/microsoft/pxt-blockly/blob/develop/core/field_dropdown.js#L91
> 
> Easiest fix would be to override that `getOptions()` method in FieldTileset and make sure that if menuGenerator isn't set, it returns an array with just the transparent tile:
> 
> https://github.com/microsoft/pxt/blob/master/pxtblocks/fields/field_tileset.ts#L166

Oooo okay I'll do that. Thanks!!!

jwunderl:
@livcheerful did you mean to merge this? It got closed ~

livcheerful:
> @livcheerful did you mean to merge this? It got closed ~

omggg thank you joey I did a dumb lol

pxtblocks/fields/field_tileset.ts
```diff
@@ -169,7 +169,7 @@ namespace pxtblockly {
 
             return this.menuGenerator_.call(this);
         }
-      
+
```
jwunderl, line 172: Maybe helpful, my favorite small setting:

image

livcheerful: suuuuuuuuuuuuuuuuuuuuuuper helpful

PR #7168, opened 2020-06-10
Embed blocks xml in screenshot, load block snippet on drag

shakao:
Build: https://arcade.makecode.com/app/da6fe3fd22e4f79bb98659b267d92aaffd357e11-229de52ccf

demo-snippet2

- Right-click -> "snapshot" to download magic png
- Drag png into editor to load blocks

Notes: 
- Should we mark in the image some way that this is a blocks screenshot that contains a snippet? Otherwise some snapshots mysteriously work and some don't
- Imported functions rename, imported variables create if they are not created but will use the existing one if it's already in the project
- Fails if the snippet contains blocks from an extension that has not been added. Could do some nicer UI for this in v2 I think
- Right now it only embeds the blocks for the right-click screenshot and not other screenshots (blocks diff, github preview)


pxtlib/util.ts
```diff
@@ -1194,6 +1194,155 @@ namespace ts.pxtc.Util {
...
+            ptr = encode(imgdat.data, ptr, bpp, blob)
+        else {
+            let firstChunk = imgCapacity - header.length
+            ptr = encode(imgdat.data, ptr, bpp, blob.slice(0, firstChunk))
+            ptr = encode(imgdat.data, addedOffset, 8, blob.slice(firstChunk))
+        }
+        // set remaining alpha
+        ptr |= 3
+        while (ptr < imgdat.data.length) {
+            if (!transparent) imgdat.data[ptr] = 0xff
```
shakao, line 1287: i extracted the encode/decode functions from screenshot.ts and did not change them except for here--added a flag to leave the remaining alpha pixels as they are to preserve the blocks screenshot transparency

shakao:
Actually let me use the uidesign channel for this

pxtlib/util.ts
mmoskal: I guess the `if (!transparent)` could be around the whole loop?

PR #7177, opened 2020-06-12
2020 blog add new videos

pelikhan:
Removing beta references, adding new videos.


ganicke:
Ah, we have colliding posts: #7179.

pelikhan:
Could you add the latest video and playlist in yours?

PR #7178, opened 2020-06-12
Embed blocks xml in screenshot, load snippet on drag

shakao:
Build: https://arcade.makecode.com/app/45ec34344bacb933977c28ac820e5dd53a686b58-235a5c84dc

Right-click -> screenshot generates a screenshot with a save icon:

arcade-screenshot (29)

Dragging this screenshot into the editor will load that block snippet. Currently does not import extensions, will save that for a future iteration. This will not affect the blocks screenshot outside of right-click (github preview, etc).


pxtlib/util.ts
```diff
@@ -1188,6 +1188,155 @@ namespace ts.pxtc.Util {
         if (/xml|svg/.test(mimetype)) return `data:${mimetype},${encodeURIComponent(data)}`
         else return `data:${mimetype || "image/png"};base64,${encodeBase64(toUTF8(data))}`;
     }
+
+    export const imageMagic = 0x59347a7d // randomly selected
+    export const imageHeaderSize = 36 // has to be divisible by 9
+    export function encodeBlobAsync(canvas: HTMLCanvasElement, blob: Uint8Array) {
```
shakao, line 1194: enocdeBlobAsync/decodeBlobAsync are unchanged, just extracted from screenshot.ts

pxtblocks/blocklylayout.ts
```diff
@@ -205,7 +217,15 @@ namespace pxt.blocks.layout {
                     ctx.drawImage(img, 0, 0, width, height, 0, 0, cvs.width, cvs.height);
                     canvasdata = cvs.toDataURL("image/png");
                 }
-                resolve(canvasdata);
+                if (text) {
+                    let p = pxt.lzmaCompressAsync(text).then(blob => {
+                        const datacvs = pxt.Util.encodeBlobAsync(cvs, blob);
+                        resolve(datacvs.toDataURL("image/png"));
+                    });
+                    p.done();
```
riknoll, line 225: Why call done here?

pxtblocks/blocklyimporter.ts
```diff
@@ -2,6 +2,10 @@
 /// <reference path="../built/pxtlib.d.ts" />
 
 namespace pxt.blocks {
+    export interface BlockSnippet {
+        xml: string[];
+        extensions?: string[];
```
riknoll, line 7: Add comment saying currently unused?

pxtblocks/blocklyimporter.ts
```diff
@@ -2,6 +2,10 @@
 /// <reference path="../built/pxtlib.d.ts" />
 
 namespace pxt.blocks {
+    export interface BlockSnippet {
+        xml: string[];
```
pelikhan, line 6: why not joining the line then encoding?

pelikhan:
Could we also add the save icon in the "saved project" image? (future PR maybe)

pxtblocks/blocklyimporter.ts
```diff
@@ -2,6 +2,10 @@
 /// <reference path="../built/pxtlib.d.ts" />
 
 namespace pxt.blocks {
+    export interface BlockSnippet {
```
pelikhan, line 5: add the "versions" field of targets so that down the road we can do clever app compat to import snippets.

pxtblocks/blocklylayout.ts
riknoll: Ah, rethrowing exceptions

pxtblocks/blocklyimporter.ts
shakao: blockly's xml -> block only works on one block stack, so i have to call it separately for each top level block

shakao:
really easy to enable for the github preview image! i'll turn it on and we can turn it off if it turns out to be not useful

PR #7180, opened 2020-06-12
Adding all lowercase locale as part of normalization

abchatra:
Correcting for the extension author writing all lower case locales.
Fixes https://github.com/microsoft/pxt-microbit/issues/3265

### DIFF
--- a/pxtlib/commonutil.ts
+++ b/pxtlib/commonutil.ts
@@ -71,10 +71,15 @@ namespace ts.pxtc.Util {
         return _localizeLang;
     }
 
+    // This function returns normalized language code
+    // For example: zh-CN this returns ["zh-CN", "zh", "zh-cn"]
+    // First two are valid crowdin\makecode locale code,
+    // Last all lowercase one is just for the backup when reading user defined extensions & tutorials.
     export function normalizeLanguageCode(code: string): string[] {
         const langParts = /^(\w{2})-(\w{2}$)/i.exec(code);
         if (langParts && langParts[1] && langParts[2]) {
-            return [`${langParts[1].toLowerCase()}-${langParts[2].toUpperCase()}`, langParts[1].toLowerCase()];
+            return [`${langParts[1].toLowerCase()}-${langParts[2].toUpperCase()}`, langParts[1].toLowerCase(),
+             `${langParts[1].toLowerCase()}-${langParts[2].toLowerCase()}`];
         } else {
             return [(code || "en").toLowerCase()];
         }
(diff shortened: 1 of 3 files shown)


abchatra:
I need to release this to live to fix extensions not loading the zh-cn locale.

shakao:
will there also be issues if some user defines the language code as "ZH-CN"? maybe we should also have them update their extension repo?

pxtlib/commonutil.ts
```diff
@@ -71,10 +71,15 @@ namespace ts.pxtc.Util {
...
+    // This function returns normalized language code
+    // For example: zh-CN this returns ["zh-CN", "zh", "zh-cn"]
+    // First two are valid crowdin\makecode locale code,
+    // Last all lowercase one is just for the backup when reading user defined extensions & tutorials.
     export function normalizeLanguageCode(code: string): string[] {
         const langParts = /^(\w{2})-(\w{2}$)/i.exec(code);
         if (langParts && langParts[1] && langParts[2]) {
-            return [`${langParts[1].toLowerCase()}-${langParts[2].toUpperCase()}`, langParts[1].toLowerCase()];
+            return [`${langParts[1].toLowerCase()}-${langParts[2].toUpperCase()}`, langParts[1].toLowerCase(),
+             `${langParts[1].toLowerCase()}-${langParts[2].toLowerCase()}`];
```
pelikhan, line 82: feels like the order should be ``zn-cn, zn-CN, zn``

pelikhan: we shoudl always try regional codes first

abchatra: No the right locale is case sensitive. See https://support.crowdin.com/api/language-codes/

abchatra: zn-CN is the right locale. first part of the locale is lower case and second is capital case.

abchatra:
@shakao This fix is just for backward compat. We had a bug in the past where we did lower case for everything and it worked while comparing the locale. I don't want to do all the combination of lower case & upper case.

shakao:
ah okay, makes sense

pxtlib/commonutil.ts
pelikhan: then we should reorder to ``zh-CN, zh-cn, zh``

abchatra: Ok I changed the order for 3rd party extensions & tutorial. Doesn't make sense other places.

PR #7185, opened 2020-06-12
Fix nonlocal and global variable tracking and type unification

darzu:
Fixes: https://github.com/microsoft/pxt-microbit/issues/2637
Fixes: https://github.com/microsoft/pxt-microbit/issues/3271
and adds tests for above.

There were some issues with how we were tracking nonlocal and global variables in py->ts.
Before we were creating new symbol info for each variable in each scope when encountering "global" and "nonlocal" directives. However this was causing issues with type unification and proper variable lookup via traversing parent scopes since. The essential issue is we had duplicate sources of truth on where a symbol was declared.

Instead, we now keep a seperate data structure, the "ScopeSymbolInfo" for each variable declare in a scope (either global/nonlocal or other). This struct passes through to the original symbol info at the outermost declaration, so we properly unify types across scopes and we can traverse parent scopes properly.

Tagging new folks @livcheerful, @eanders-MS  in case you'd like some exposure to Python->TypeScript compiler. Happy to explain over chat :)

(Big thanks to @mmoskal for helping brainstorm the solutions here.)


tests/runtime-trace-tests/cases/global_array_len.py
```diff
@@ -0,0 +1,9 @@
+choices = ["rock", "paper", "scissors"]
+
+x = len(choices)
+
+def next_choice(forward):
```
riknoll, line 5: Why is this in the trace test if it never gets called?

pxtpy/converter.ts
```diff
@@ -355,24 +355,35 @@ namespace pxt.py {
...
+                    || modifier === VarModifier.NonLocal)) {
+                varSym = addSymbol(SK.Variable, name)
+            }
+            else if (isLocalScope(scope))
+                varSym = mkSymbol(SK.Variable, name)
             else
-                v = addSymbol(SK.Variable, qn)
-            scope.vars[n] = v
+                varSym = addSymbol(SK.Variable, qualifiedName)
+            varScopeSym = scope.vars[name] = {
```
riknoll, line 378: Nitpick: make this two lines?

pxtpy/converter.ts
```diff
@@ -1106,7 +1130,8 @@ namespace pxt.py {
                     unifyClass(n, sym.pyRetType!, ctx.currClass!.symInfo)
                 } else {
                     if (funname == "__get__" || funname == "__set__") {
-                        let vv = n.vars["value"]
+                        let scopeVv = n.vars["value"]
```
riknoll, line 1133: Nitpick: as long as we're renaming things, vv?

tests/runtime-trace-tests/cases/global_array_len.py
darzu: Trace tests also test that compilation doesn't error py->ts and back to py.
But yeah, I might as well make a trace here

PR #7186, opened 2020-06-13
Avoid single letter note name to conflict with other meaning like button names.

sgk:
Single letter note names like "A" and "B" may conflict with other meanings especially in MakeCode variants other than micro:bit. It is better to add something like "{id:note}" before the note names used as a translation key strings.


abchatra:
I don't see a repro for microbit today. We should take it in master for arcade and others and will be integrated in microbit in future.

pelikhan:
The repro is that both button A and note A are translated the same — and that won’t work in some languages.

PR #7194, opened 2020-06-15
Embedfix

pelikhan:
Fix for https://github.com/microsoft/pxt-microbit/issues/3276

This test should have fixed this: https://makecode.microbit.org/release-tests/sharing#embed


pxtlib/package.ts
```diff
@@ -567,12 +570,32 @@ namespace pxt {
...
                 initPromise = initPromise.then(() => this.downloadAsync())
 
+            // we are installing the script, and we've download the original version and we haven't upgraded it yet
+            // do upgrade and reload as needed
+            if (this.level == 0 && isInstall) {
+                initPromise = initPromise.then(() => this.upgradePackagesAsync())
+                    .then(fixes => {
+                        if (fixes) {
+                            // worst case scenario with double load
+                            pxt.tickEvent("package.doubleload", fixes);
```
abchatra, line 589: This tickEvent needs careful review. Does it add tons of indices? Like versions. Then we need to remove the fixes array. Can you provide me a sample

abchatra:
The scenario this is broken is only when we auto upgrade packages?

pxtlib/package.ts
abchatra: You can split into multiple events in a loop. Something like "extension": "xyz", "version": 1.0.1 , "upgraded version": 2.05

abchatra:
How did we regress this scenario? What original change caused this?

pelikhan:
All embeded simulators are broken.

pelikhan:
We added patching support for outdated github extensions. It blows up in the embed load case.

abchatra:
Looks good, can you fix the tickEvent. Rest looks good.

pxtlib/package.ts
pelikhan: adding one tick per extension

PR #7199, opened 2020-06-15
Fix array indexing in Python

darzu:
Fixes: https://github.com/microsoft/pxt-microbit/issues/3277
Fixes: https://github.com/microsoft/pxt-microbit/issues/3254

Previously we were unifying this expression "foo[index]" with the type of "index".
Now we unify this expression "foo[index]" with the element type of "foo" if it's an array, or `any` if it's an object.

The changes are just a few lines. The rest of this PR is test infrastructure upgrades that made finding and fixing this easier.


pxtpy/converter.ts
```diff
@@ -2181,7 +2181,16 @@ namespace pxt.py {
         },
         Subscript: (n: py.Subscript) => {
             if (n.slice.kind == "Index") {
-                unifyTypeOf(n, typeOf(n.value));
```
mmoskal, line 2184: I guess this never worked in the slightest!

darzu: Actually I see the intent now. It works for strings. I added a case back for strings.

PR #7218, opened 2020-06-17
Relax table of contents parsing regex for docs TOC localization

shakao:
Crowdin returns list headings with extra whitespace at the end (eg `* [따라해보기](/tutorials)`); the regex doesn't match this and so all localized TOC headings are empty. I'm modifying the regex so it doesn't match start/end of line but making the .* groups nongreedy, which I think should be pretty safe. It should still always match the first link of `[text](path)` format. 

I've tested this locally by calling buildTOC manually on an input string I copied from Crowdin, but haven't been able to deploy to staging to test there.

Fixes https://github.com/microsoft/pxt-microbit/issues/3263


shakao:
ok i changed my mind, the trimming seems like overkill so now it's just the regex. >> @riknoll we run it through marked's lexer so that strips the "* " but i guess not whitespace changes haha

PR #7229, opened 2020-06-18
Sticky the debug toolbar

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/1972

Here's what it looks like!

sticky debug


MINTGenie:
Thanks for the fix @livcheerful 👍

PR #7231, opened 2020-06-19
Move to default branch

pelikhan:
We shouldn't assume that master is the default branch. Cleaning up to remove this assumption.

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -400,18 +400,18 @@ function ciAsync() {
         fs.writeFileSync(npmrc, cfg)
     }
 
-    const latest = branch == "master" ? "latest" : "git-" + branch
-    // upload locs on build on master
-    const masterOrReleaseBranchRx = /^(master|v\d+\.\d+\.\d+)$/;
-    const apiStringBranchRx = pxt.appTarget.uploadApiStringsBranchRx
-        ? new RegExp(pxt.appTarget.uploadApiStringsBranchRx)
-        : masterOrReleaseBranchRx;
+    const defaultBranch = pxt.github.isDefaultBranch(branch)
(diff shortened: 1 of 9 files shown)


pelikhan:
This is not ready at all.

abchatra:
@pelikhan what is pending in this PR?
@aznhassan

pelikhan:
Backend need to be able to pull the default branch when not specified.

Switch template to use "main" be default?

Otherwise lots of testing.

pelikhan (commented):
When we load a configuration from the cloud, we don't know the default branch. The could should support a tag called "default" and use "master" or "main" as needed.

cli/cli.ts
```diff
@@ -450,18 +407,18 @@ function ciAsync() {
...
-    const latest = branch == "master" ? "latest" : "git-" + branch
-    // upload locs on build on master
-    const masterOrReleaseBranchRx = /^(master|v\d+\.\d+\.\d+)$/;
-    const apiStringBranchRx = pxt.appTarget.uploadApiStringsBranchRx
-        ? new RegExp(pxt.appTarget.uploadApiStringsBranchRx)
-        : masterOrReleaseBranchRx;
+    const defaultBranch = pxt.github.isDefaultBranch(branch)
+    const latest = defaultBranch ? "latest" : "git-" + branch
+    // upload locs on build on default or releases
+    const apiStringBranchRx: (t: string) => boolean = ((t:string) => new RegExp(pxt.appTarget.uploadApiStringsBranchRx).test(t))
```
jwunderl, line 413:
```suggestion
    const apiStringBranchRx: (t: string) => boolean = ((t: string) => new RegExp(pxt.appTarget.uploadApiStringsBranchRx).test(t))
```

to make the linter happen

PR #7239, opened 2020-06-23
Add editor controller command for importing custom.ts file

shakao:
New "importcustomblocks" editor controller command, to be used by updated playground.


riknoll:
Should we just use the existing importproject? Otherwise, code looks good

shakao:
i can modify importproject! but i don't think there's a way to do this without changes to pxt since i have to pull the default project template from pxt.appTarget, right?

pelikhan:
once the editor is opened, you should receive the project when it tries to save it. You can use that to add/modify custom.ts and reinject it back.

riknoll:
@shakao check out the workspacesync event

PR #7241, opened 2020-06-23
Update playground to load blocks in iframe instead of custom sandbox

shakao:
- Core logic is in playground.ts
- Moved the common CSS from /tutorial-tool into a /shared folder. I thought about doing this for the shared JS too but it's kind of scattered and didn't seem entirely worth it
- Did a pass over the samples for formatting. Might go back and do more proofreading later
- Removed all unused files

I checked that all samples load correctly in Arcade (beta), except the text dropdown, which I think is probably due to blockly upgrade issues--I'll investigate separately.

For https://github.com/microsoft/pxt-arcade/issues/1942


docs/static/playground/playground.ts
```diff
@@ -0,0 +1,311 @@
...
+                name: "beta",
+                url: "https://makecode.adafruit.com/beta?controller=1"
+            },
+            {
+                name: "released",
+                url: "https://makecode.adafruit.com?controller=1"
+            }
+        ],
+    }, {
+        name: "Micro:bit",
```
jwunderl, line 79:
```suggestion
        name: "micro:bit",
```

should probably follow standard formatting for their name?

docs/playground.html
```diff
@@ -1,70 +1,65 @@
...
-    <nav class="ui menu inverted fixed borderless">
-        <div class="menu left">
-            <div class="ui item">
-                MakeCode Playground
-            </div>
-        </div>
+		<div class="menu right">
+			<a href="/defining-blocks" target="_blank" class="ui item">
+				Defining Blocks
+			</a>
```
riknoll, line 25: this file has tabs instead of spaces

docs/playground.html
```diff
@@ -1,70 +1,65 @@
...
-    <section id="playground">
-    </section>
 
-    <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/jquery/1.9.1/jquery.min.js" integrity="sha256-wS9gmOZBqsqWxgIVgA8Y9WcQOa7PgSIX+rPA0VL2rbQ="
-        crossorigin="anonymous"></script>
-
-    <!-- @include playground-apptrackingweb.html -->
-    <!-- @include apptracking.html -->
+	<script type="text/javascript" src="https://pxt.azureedge.net/blob/795d8506c80a04f5ca26f577a8d6152e2fa3e7a6/doccdn/pxtweb.js"></script>
+    <!-- @include tracking.html -->
```
riknoll, line 52: The old page was using a different tracking template, why the change? Also do you know why it was including both apptracking and playground-apptracingweb?

shakao: tracking.html is appinsights + pxtweb and apptracking is appinsights alone, and playground-apptrackingweb was pxtweb alone SO...... i don't know why it had a separate playground-apptracking file but this should add up to the same thing (appinsights + pxtweb). i think i can actually delete the pxtweb script tag here so i'll do that along with addressing the other comments!

docs/static/playground/playground.js
```diff
@@ -1,404 +1,266 @@
...
-        var sampleSwitcher = document.createElement('select');
-        var sampleChapter;
-        PLAY_SAMPLES.forEach(function (sample) {
-            if (!sampleChapter || sampleChapter.label !== sample.chapter) {
-                sampleChapter = document.createElement('optgroup');
-                sampleChapter.label = sample.chapter;
-                sampleSwitcher.appendChild(sampleChapter);
+        ]
+    }, {
+        name: "LEGO EV3",
```
pelikhan, line 66: You could load the editors.json for list of editors (since you are cleaning)

pelikhan: Maybe next iteration

docs/static/playground/playground.js
```diff
@@ -1,404 +1,266 @@
...
+var STORAGE_KEY = "SAVED_SAMPLE";
+var ENDPOINT_KEY = "SAVED_ENDPOINT";
+var existing = localStorage.getItem(STORAGE_KEY);
+// @ts-ignore
+var editor = monaco.editor.create(document.getElementById("container"), {
+    value: existing || "",
+    language: "typescript",
+    minimap: { enabled: false }
+});
+var targets = [
```
pelikhan, line 12: I've also refactored our list of editors in the 'editors.json" file, which you can fetch. WOuld be good to eventually unify all these tools to use the same list of editors.

docs/static/playground/playground.js
```diff
@@ -1,404 +1,266 @@
...
+                url: "https://makecode.mindstorms.com?controller=1"
             }
-            var sampleOption = document.createElement('option');
-            sampleOption.value = sample.id;
-            sampleOption.appendChild(document.createTextNode(sample.name));
-            sampleChapter.appendChild(sampleOption);
+        ]
+    }
+];
+var CUSTOM_FILE = "_custom.ts";
```
pelikhan, line 80: why not just "custom.ts" ? (not that it matters at all)

docs/static/playground/playground.js
```diff
@@ -1,404 +1,266 @@
...
+}
+function loadIframe(selected) {
+    if (selected === selectedEndpoint)
+        return;
+    for (var _i = 0, targets_2 = targets; _i < targets_2.length; _i++) {
+        var target = targets_2[_i];
+        for (var _a = 0, _b = target.endpoints; _a < _b.length; _a++) {
+            var endpoint = _b[_a];
+            if (!selected || selected === target.name + "-" + endpoint.name) {
+                iframe.setAttribute("src", endpoint.url);
```
pelikhan, line 195: add ``ws=mem`` to avoid polluting the script list of the user.

docs/static/playground/playground.js
shakao: oh i use the underscore for files that users shouldn't modify, i feel like it makes the file look less... friendly? just habit in this case haha

docs/static/playground/playground.js
shakao: updated! i'll make a separate pr after this one to use editors.json, it's not a lot of code but i want to add some targets to the json so i might tweak streamer to add a list of disabled targets

pelikhan: you can add a flag in editors.json to specify that it is supported in  particular tool...
```
"microbit": {
    "tutorial-tool": true,
    ...
}
```
to avoid interference

docs/static/playground/playground.js
pelikhan: the editor creates a "custom.ts" file so let's drop the ``_``

docs/static/playground/playground.ts
```diff
@@ -0,0 +1,311 @@
...
+            },
+            {
+                name: "released",
+                url: "https://makecode.mindstorms.com?controller=1&ws=mem"
+            }
+        ],
+    }
+];
+
+const CUSTOM_FILE = "_custom.ts";
```
pelikhan, line 107: custom.ts

shakao: updated! i ended up removing ws=mem because it isn't compatible with controller=1, and controller=1 also does not pollute the user's real projects

PR #7243, opened 2020-06-24
Serial console: Stop scrolling when simulator stops

eanders-MS:
This PR is larger than strictly necessary in order to preserve the serial console very-smooth chart scrolling.  I mention a simpler alternative below, with the trade off that chart scrolling would happen in discrete steps and look chunky.

### Details
The serial console chart (from the "Smoothie Chart" package) can operate in either "realtime" or "non-realtime" mode. In realtime mode, the scroll animation is continuous and looks very smooth, but there's no way to pause it. In non-realtime, scrolling only happens when new data arrives. Because of this, scrolling is choppy.

I wanted both continuous scrolling and the ability to pause. This can be accomplished by toggling the realtime mode of the chart at the right times. To do this, I added an interface on `Editor` called `simStateChanged`, which allows the serial console to change the chart's mode in response to the simulator stopping and starting.

All that so the chart will scroll continuously instead of in discrete steps.

### Simpler alternative
If chunky chart scrolling is preferred over modifying Editor's interface, the solution is simpler: Have the chart always run in non-realtime mode.

### Dependency changes
Non-realtime chart support was not available in the version of Smoothie Charts that the code was referencing, but a newer version of the package was already included in the project as a pxt dependency, so I switched to using that one. Let me know if there is more I need to do here (packaging/webpacking considerations, etc.). The older package is part of Blockly.

---
Fixes https://github.com/microsoft/pxt-arcade/issues/1955


webapp/src/serial.tsx
```diff
@@ -84,6 +83,15 @@ export class Editor extends srceditor.Editor {
         }
     }
 
+    simStateChanged() {
+        // We're leveraging the chart's `nonRealtmeData` flag to get both smooth scrolling
+        // and the ability to pause the animation.
+        // When the chart is in realtime mode, scrolling is very smooth but cannot be paused.
+        // In non-realtime mode, scrolling only occurs when new data arrives, but scrolling
+        // is choppy (being tied to new data arriving).
+        this.charts.forEach((chart) => chart.setRealtimeData(this.parent.isSimulatorRunning()));
```
riknoll, line 92: We should only pause charts if they are coming from the simulator.

You can also have charts that come from the hardware (using serial over WebUSB)

webapp/src/serial.tsx
```diff
@@ -224,30 +224,28 @@ export class Editor extends srceditor.Editor {
...
+        for (let i = 0; i < this.charts.length; ++i) {
+            let chart = this.charts[i]
+            if (chart.shouldContain(source, variable)) {
+                homeChart = chart
+                break
+            }
+        }
+        if (!homeChart) {
+            homeChart = new Chart(source, variable, this.chartIdx, this.currentLineColors)
+            homeChart.setRealtimeData(this.parent.isSimulatorRunning())
```
riknoll, line 239: Same comment as above

webapp/src/serial.tsx
```diff
@@ -84,6 +83,15 @@ export class Editor extends srceditor.Editor {
         }
     }
 
+    simStateChanged() {
+        // We're leveraging the chart's `nonRealtmeData` flag to get both smooth scrolling
```
jwunderl, line 87:
```suggestion
        // We're leveraging the chart's `nonRealtimeData` flag to get both smooth scrolling
```

webapp/src/serial.tsx
```diff
@@ -224,30 +224,28 @@ export class Editor extends srceditor.Editor {
...
+        //See if there is a "home chart" that this point belongs to -
+        //if not, create a new chart
+        let homeChart: Chart = undefined
+        for (let i = 0; i < this.charts.length; ++i) {
+            let chart = this.charts[i]
+            if (chart.shouldContain(source, variable)) {
+                homeChart = chart
+                break
+            }
+        }
```
jwunderl, line 236:
```suggestion
        const homeChart = this.charts.find(chart => chart.shouldContain(source, variable));
```

This portions not new, but easy enough to cleanup; just personal preference though, up to you

webapp/src/serial.tsx
eanders-MS: Oh nice catch. How can I test this? Will web search it.

pelikhan: Create a Microbit script that uses plotBarGraph, flash it to the device and you’ve a second console button for device data.

eanders-MS: It isn't clear to me how to connect the running microbit program to the serial console. Do I need to pair the device? It isn't being detected when I try to pair it, but I can flash it by copying the hex file.

eanders-MS: Ok after updating the microbit firmware I was able to pair it and get the serial stream.

abchatra: Instructions here if you need it: https://makecode.microbit.org/device/usb/webusb

pelikhan:
For this kind of big changes, it's very helpful to upload a local build using

```
pxt uploadpkg
```

riknoll:
@pelikhan I think you mean `pxt uploadtrg`

eanders-MS:
I'm unable to `pxt uploadtrg` due to my GitHub username having uppercase letters in it. I think I'll take a look at that bug today.

darzu:
Any before  & after gifs?  Recommend gifox on Mac and screenToGif on Windows

eanders-MS:
Sure I can supply some before/after gifs. My fav tool is Snagit from TechSmith. Not free, but worth every cent.

eanders-MS:
#### Before
BEFORE-2020-06-26_13-08-15

#### After
AFTER-2020-06-26_13-10-01

darzu:
Awesome! So what's the behavior after u resume? Does it pick up exactly where it left off? Is there any marker to indicate "this is a new run" ? I imagine a vertical red line or something could be useful

eanders-MS:
That behavior is unchanged. It clears the console and starts a fresh chart.

darzu:
I forget, do we have a history of past runs? If not, that'd be neat. Especially if someone accidently stops a run or drags a block which causes a new run.

Out of scope likely

eanders-MS:
No support for history of runs at the moment. That would be a new feature.

Some initial thoughts about it:
* It could produce a lot of junk in the history as the app is developed (multiple starts/restarts). We might want it to be something the user has to enable.
* Alternatively: the console could support streaming out to a rest endpoint for more persistent data gathering.

pelikhan:
KISS: no history.

eanders-MS:
I've uploaded a local build Arcade build: https://arcade.staging.pxt.io/app/9053693207593e6e569b75ed5cdefe8e01e5dbe7-048e7f5a4b

Working on getting permission to upload a microbit build (allows for testing both serial console and simulator console behavior)

eanders-MS:
microbit was user error (forgot to set access token), but is unnecessary anyway. Arcade hardware can write to serial console.

I think this change is ready for merge. Any additional feedback or gates?

PR #7246, opened 2020-06-25
Fix type checking of function args by allowing narrower types instead of always unifying with formal param

darzu:
Fixes: https://github.com/microsoft/pxt-arcade/issues/2000
Part of fix for: https://github.com/microsoft/pxt-arcade/issues/2001

We were always unifying the types of an argument expression and the formal parameter type of the function being called. This is sound but too strict, so long as the argument is assignable it should be allowed. This PR carves out an assignability exception for interface/class inheritance. Future work would do more complete assignability check and also track the "narrowest" type a free type must satisfy and ensure it's enforced as unification happens in later iterations.


pxtpy/converter.ts
```diff
@@ -609,20 +619,39 @@ namespace pxt.py {
...
-        if (currIteration > 1) {
+        if (isFunType(t0) || isFunType(t1)
+            || isPrimativeType(t0) || isPrimativeType(t1)) {
             unify(e, typeOf(e), t1)
+        } else {
+            // if we're not sure about assinability or unification, we do nothing as future
+            // iterations may unify or make assinability clear.
+            // TODO: Ideally what we should do is track a "constraining type" similiar to how we track .union
+            // per type, and ensure the constraints are met as we unify or narrow types. The difficulty is that
+            // this depends on a more accurate assignability check which will take some work to get right.
```
jwunderl, line 649: this is where `unifyOrConstrain` would be called, right? maybe leave it as a no-op call to keep track for future?

pxtpy/converter.ts
```diff
@@ -491,7 +497,7 @@ namespace pxt.py {
         error(a, 9500, U.lf("types not compatible: {0} and {1}", t2s(t0), t2s(t1)))
     }
 
-    function typeCtor(t: Type): any {
+    function typeCtor(t: Type): string | SymbolInfo | {} | null {
```
jwunderl, line 500: Does this type do anything? Wouldn't it evaluate to just == `{} | null`?

pxtpy/converter.ts
```diff
@@ -587,6 +596,64 @@ namespace pxt.py {
...
+
+    function narrow(e: Expr, constrainingType: Type): void {
+        const t0 = canonicalize(typeOf(e))
+        const t1 = canonicalize(constrainingType)
+
+        if (isAssignable(t0, t1)) {
+            return;
+        }
+
+        // if we don't know if two types are assignable, we can try to unify them ssin common cases
```
jwunderl, line 639:
```suggestion
        // if we don't know if two types are assignable, we can try to unify them in common cases
```

pxtpy/converter.ts
```diff
@@ -587,6 +596,64 @@ namespace pxt.py {
...
+            // if we're not sure about assinability or unification, we do nothing as future
+            // iterations may unify or make assinability clear.
+            // TODO: Ideally what we should do is track a "constraining type" similiar to how we track .union
+            // per type, and ensure the constraints are met as we unify or narrow types. The difficulty is that
+            // this depends on a more accurate assignability check which will take some work to get right.
+        }
+    }
+
+    function isSymbol(c: string | SymbolInfo | {} | null): c is SymbolInfo {
+        return !!c && !!(c as SymbolInfo).name
```
jwunderl, line 654:
```suggestion
        return !!(c as SymbolInfo)?.name
```

pxtpy/converter.ts
darzu: Woops, unifyOrConstrain is dead code I removed

pxtpy/converter.ts
darzu: I'm not sure what you mean. All four types can be returned.

pxtpy/converter.ts
darzu: ty!

pxtpy/converter.ts
jwunderl: String and SymbolInfo are both subtypes of {}, but doesn't really make a difference - I suppose this will probably show up with the other two possible types explicitly in intellisense / as hover text, which is better.

Issue #7248, opened 2020-06-25
One off page issues on safari


jwunderl:
**Describe the bug**

A few bugs running makecode tools on safari:

makecode.com/multi:

```
[Error] SecurityError: IDBFactory.open() called in an invalid security context
	e (main.js:1:967098)
	(anonymous function) (main.js:1:967155)
[Error] TypeError: undefined is not an object (evaluating 'H.reloadEditor')
	(anonymous function) (main.js:1:97845)
	(anonymous function) (main.js:1:105550)
```

first happens each reload, second happened only first time after bump. Stuck spinning for me, works on all but microbit.

makecode.com/tutorial-tool:

```
[Error] Failed to load resource: the server responded with a status of 404 () (dapjs.js.map, line 0)
https://pxt.azureedge.net/commit/23c3d8c7f8a0f44f7681d1e3de648cb6b67220d2/dapjs.js.map
```

looks like this:

![image](https://user-images.githubusercontent.com/5615930/85783649-5c8c9980-b6dc-11ea-97ad-00f714077d60.png)

makecode.com/playground:

<img width="1665" alt="Screen Shot 2020-06-25 at 12 12 40 PM" src="https://user-images.githubusercontent.com/5615930/85784650-577c1a00-b6dd-11ea-9562-1b6f043d270c.png">

streamer:

```
[Error] Error while parsing the 'sandbox' attribute: 'allow-downloads' is an invalid sandbox flag. (streamer, line 1073)
[Error] Error while parsing the 'sandbox' attribute: 'allow-downloads' is an invalid sandbox flag. (streamer, line 1074)
[Error] Failed to load resource: the server responded with a status of 404 (Not Found) (undefined, line 0) //// NOTE; this one's trying to access https://makecode.com/undefined, which seems odd?
[Error] SecurityError: IDBFactory.open() called in an invalid security context
	e (main.js:1:967098)
	(anonymous function) (main.js:1:967155)
[Error] NotSupportedError: The operation is not supported.
(anonymous function) — script.js:1475
forEach
(anonymous function) — script.js:1472
asyncFunctionResume
[native code]
promiseReactionJob

	(anonymous function) (script.js:80)
	asyncFunctionResume
	(anonymous function)
	promiseReactionJob
```

[...]

PR #7249, opened 2020-06-25
fix playground / tutorial-tool on safari

jwunderl:
fix view on current safari re: https://github.com/microsoft/pxt/issues/7248 (was there another issue for this?)

It's possible there's an issue on much older versions of safari (as there are a few 'still used but old' versions that don't support some es5 functions), but if there is we'd probably need browserstack to get them; oldest safari I can get access to on my device is ios 10.3.1 and with this change it works there ~


abchatra:
Lets not worry about older safari for this site.

jwunderl:
Sounds good, this is all that should be needed for newish safari versions then ~

PR #7252, opened 2020-06-26
Use editors.json in playground, additional editors + "unsupported" flag

shakao:
- Remove target list from playground, use editors.json
- Add "unsupported" array with endpoint names + read this in playground/streamer
- Added new endpoints for playground

I looked into updating the tutorial tool as well but the share code would require some changes to the editor.json structure and it grew to be a slightly larger change so I'm leaving it as-is for now.

### DIFF
--- a/docs/editors.json
+++ b/docs/editors.json
@@ -1,6 +1,5 @@
 {
     "arcade": {
-        "id": "arcade-beta",
         "name": "Arcade BETA",
         "title": "MakeCode Arcade",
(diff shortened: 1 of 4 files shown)


docs/editors.json
```diff
@@ -1,5 +1,6 @@
 {
     "arcade": {
+        "id": "arcade-beta",
```
pelikhan, line 3: id is the key in the map....

shakao: oh yeah good point, updated!

PR #7259, opened 2020-06-29
Remove playground tasks from gulpfile

shakao:


livcheerful:
Why don't we want playground in the gulpfile now? 😅

jwunderl:
These files were copying over the builtin livs for e.g. array methods and string methods - now that its using targets directly we dont need those anymore. 

Can we delete the duplicates of the builtin libs this copies over now too?

shakao:
no, i believe the pxt-common files are still used by the test host/runners

jwunderl:
Ahh oops, I meant the copies in the docs not the sources, but I just noticed that these were already deleted with your previous change. So good to go!

PR #7260, opened 2020-06-30
Asset editor setup

kimprice:
This is the basic setup of the asset editor html page which will be used to embed the sprite/animation/tilemap editor into the vs code extension.

image


gulpfile.js
```diff
@@ -131,7 +131,8 @@ function initWatch() {
         targetjs,
         webapp,
         browserifyWebapp,
-        gulp.parallel(semanticjs, copyJquery, copyWebapp, copySemanticFonts, copyMonaco)
+        browserifyAssetEditor,
+        gulp.parallel(semanticjs, copyJquery, copyWebapp, copyPlayground, copySemanticFonts, copyMonaco)
```
shakao, line 135: remove "copyPlayground" task

webapp/public/release.manifest
```diff
@@ -16,6 +16,7 @@ CACHE:
 /blb/pxteditor.js
 /blb/pxtsim.js
 /blb/pxtembed.js
+/blb/mainAsset.js
```
pelikhan, line 19: mainAsset is too generic, it should be assetEditor.js or pxtasseteditor.js (all lowercase)

pelikhan: The manifest is for the pxt editor, I don’t think you need this here since the asset editor is already built into Pxt

webapp/public/assetEditor.html
```diff
@@ -0,0 +1,22 @@
...
+<head> 
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
+    <link rel="stylesheet" data-rtl="/blb/rtlsemantic.css" href="/blb/semantic.css" type="text/css">
+    <!-- <link rel="stylesheet" href="https://pxt.azureedge.net/blob/c5005664a6804b0cd60045f2c25abb69ece0576d/semantic.css" type="text/css"> -->
+    <title>Asset Editor</title>
+</head>
+
+<body>
+    <h1>Sprite Editor</h1>
```
pelikhan, line 13: MakeCode Asset Editor

webapp/src/serviceworker.ts
```diff
@@ -56,6 +56,7 @@ function initWebappServiceWorker() {
         `/blb/pxteditor.js`,
         `/blb/pxtsim.js`,
         `/blb/pxtembed.js`,
+        '/blb/mainAsset.js',
```
pelikhan, line 59: Same as manifest: unless needed by editor, should not be here.

pelikhan:
You will need to add a route in the backend to serve this html page. See my previous PR to add the —-multi route

kimprice:
> You will need to add a route in the backend to serve this html page. See my previous PR to add the —-multi route

Awesome thanks for the reference!

PR #7262, opened 2020-06-30
Rearranging editors in editors.json

abchatra:
Rearranged to 
- Live editors first,
- Beta editors next,
- Other editors at the end.

### DIFF
--- a/docs/editors.json
+++ b/docs/editors.json
@@ -1,5 +1,5 @@
 {
-    "arcade": {
+    "arcadestable": {
         "name": "Arcade",
         "title": "MakeCode Arcade",
         "url": "https://arcade.makecode.com/",
@@ -34,7 +34,7 @@
         "url": "https://makecode.mindstorms.com/",
         "unsupported": ["streamer"]
     },
-    "arcadebeta": {
+    "arcade": {
         "name": "Arcade BETA",
         "title": "MakeCode Arcade",
         "url": "https://arcade.makecode.com/beta",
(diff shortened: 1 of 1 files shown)


docs/editors.json
```diff
@@ -1,18 +1,43 @@
...
             "menu": "#F76820",
             "background": "linear-gradient(90deg, #B1E3DA 0%, #E91E63 100%)"
         }
     },
-    "arcadestable": {
-        "name": "Arcade",
+    "microbit": {
+        "name": "micro:bit",
+        "title": "MakeCode for micro:bit",
+        "url": "https://makecode.microbit.org/?github=1",
```
pelikhan, line 15: We can drop the GitHub flag now

docs/editors.json
```diff
@@ -1,18 +1,43 @@
 {
     "arcade": {
-        "name": "Arcade BETA",
+        "name": "Arcade",
```
pelikhan, line 3: This will break the GitHub explorer which requires beta.

docs/editors.json
abchatra: Removed for microbit.

docs/editors.json
abchatra: I don't understand. There is BETA editor too in this list.

pelikhan: The key is used also used as the target id when resolving an editor for a github repo. Once we ship the update, we can have "arcade" point to the released version.

abchatra: Ok fixed, I need to remember to fix this once we relase.

shakao:
looks fine for playground!

PR #7263, opened 2020-07-01
Fix keyboard shortcuts in nested image editor

shakao:
Two bugs:
- We were always dispatching keyboard shortcut events to "mainStore", even when the nested image editor had a separate store
- We were unbinding events on component unmount, which removed the event binding for the tilemap below as well


webapp/src/components/ImageEditor/keyboardShortcuts.ts
```diff
@@ -1,6 +1,9 @@
-import { ImageEditorTool } from './store/imageReducer';
+
+import { Store } from 'react-redux';
+import { ImageEditorTool, ImageEditorStore } from './store/imageReducer';
 import { dispatchChangeZoom, dispatchUndoImageEdit, dispatchRedoImageEdit, dispatchChangeImageTool, dispatchSwapBackgroundForeground, dispatchChangeSelectedColor} from './actions/dispatch';
-import store from './store/imageStore';
+import mainStore from './store/imageStore';
```
jwunderl, line 5:
```suggestion
import { mainStore } from './store/imageStore';
```

I think should be that ^ re: lint ~

PR #7269, opened 2020-07-02
Experiment: open in streamer

pelikhan:
Adds  an experiment to display button in the share dialog as an entry point to the streamer. The button opens the streamer on the same project.

image


webapp/src/sui.tsx
```diff
@@ -1350,6 +1350,7 @@ class ModalButtonElement extends data.PureComponent<ModalButton, {}> {
         return <Button
             icon={action.icon}
             text={action.label}
+            title={action.title}
```
pelikhan, line 1353: found this one bug!

PR #7270, opened 2020-07-03
allow pasting image literals from text into image editor

jwunderl:
Enable pasting into image editor from elsewhere if the plain text looks roughly like an image literal. Also see https://forum.makecode.com/t/copying-makecode-image-data-from-another-application/2656

re: @jacobcarpenter tried to add you as a reviewer too but it didn't let me :(

### DIFF
--- a/webapp/src/components/ImageEditor/ImageCanvas.tsx
+++ b/webapp/src/components/ImageEditor/ImageCanvas.tsx
@@ -292,7 +292,15 @@ class ImageCanvasImpl extends React.Component<ImageCanvasProps, {}> implements G
             return;
         }
 
-        const imageData = ev.clipboardData.getData('application/makecode-image');
+        let imageData = ev.clipboardData.getData('application/makecode-image');
+
+        if (!imageData) {
+            const textData = ev.clipboardData.getData("text/plain");
+            if (/^\s*img`(.|\s)+`\s*$/im.test(textData)) {
+                imageData = textData;
+            }
+        }
+
         const image = imageData && pxt.sprite.imageLiteralToBitmap(imageData);
 
         if (image && image.width && image.height) {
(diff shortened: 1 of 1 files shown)


webapp/src/components/ImageEditor/ImageCanvas.tsx
```diff
@@ -292,7 +292,15 @@ class ImageCanvasImpl extends React.Component<ImageCanvasProps, {}> implements G
             return;
         }
 
-        const imageData = ev.clipboardData.getData('application/makecode-image');
+        let imageData = ev.clipboardData.getData('application/makecode-image');
+
+        if (!imageData) {
+            const textData = ev.clipboardData.getData("text/plain");
+            if (/^\s*img`[\s\S]+`\s*$/im.test(textData)) {
```
jacobcarpenter, line 299: `[\s\S]+` seems equivalent to `.+`. Is there any benefit to the former?

Would there be any benefit to more strictly matching the expected input?

```
^\s*img`[.1-9a-f\s]+`\s*$
```

jacobcarpenter: Apart from regex alternatives, looks good to me!

jwunderl: `.` doesn't match new line characters, so it fails for matching multiline strings.

Matching to the valid input is probably the right thing to do, I'll be a bit more strict than just allowing any input; the image parser allows a few more alternate characters though (like `R` instead of `A`) but I've never used any besides `.` and `#`: https://github.com/microsoft/pxt/blob/master/pxtlib/spriteutils.ts#L532

jacobcarpenter: > `.` doesn't match new line characters

Ha. Wild. Looks like there's a [new `s` flag](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#Advanced_searching_with_flags) to change that behavior, but it isn't supported in all browsers yet.

jwunderl: Ahh, nice! Have to wait a while to use that though, as using new regex flags is always a little bit dangerous. For a little bit we had a crash when loading in IE and older versions of safari due to a regex having the `sticky` flag, as adding an invalid (/ not valid yet) flag to a regex literal causes all of the javascript parsing to fail

PR #7276, opened 2020-07-06
Shift docs header when we have cookies

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/1655

push everything down 😡

### DIFF
--- a/docfiles/docs.html
+++ b/docfiles/docs.html
@@ -23,14 +23,15 @@
 </head>
 
 <body id='root' class='root'>
-
+	<div id="cookiebanner"></div>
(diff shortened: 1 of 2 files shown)


shakao:
oh wait, can you try the hamburger menu and make sure that's still ok?

livcheerful:
> oh wait, can you try the hamburger menu and make sure that's still ok? i think maybe the pusher was used for aligning that, not 100% sure

ooooo yes I will do that

livcheerful:
okay so this is what it looks like - it's different from before. I'll find a different fix 😅 thanks for looking out shannon

with changes:
image

without:
image

PR #7281, opened 2020-07-08
Update indent to handle multiline strings

shakao:
- Indents the inner part of template literals
- Updates the getCommaSep to handle multiline elements (skip indent for first line, not entire first element)
- Updates indent function to indent every line, instead of just the first

Fixes https://github.com/microsoft/pxt-arcade/issues/2036

### DIFF
--- a/pxtpy/pydecompiler.ts
+++ b/pxtpy/pydecompiler.ts
@@ -45,7 +45,7 @@ namespace pxt.py {
     ///
     export const INDENT = "    "
     export function indent(lvl: number): (s: string) => string {
-        return s => `${INDENT.repeat(lvl)}${s}`
+        return s => (s || "").split('\n').map(line => `${INDENT.repeat(lvl)}${line}`).join('\n')
     }
     export const indent1 = indent(1)
 
@@ -1034,6 +1034,12 @@ namespace pxt.py {
                 return exps.map((el, i) => {
                     let sep = el.charAt(el.length - 1) == "," ? "" : ",";
                     if (i == 0) {
+                        let lines = el.split('\n');
+                        // for multiline element, indent all lines after first
+                        if (lines.length > 1) {
+                            let first = lines.shift();
+                            el = first + '\n' + indent1(lines.join('\n'));
+                        }
                         return el + sep;
                     } else if (i == exps.length - 1) {
                         return indent1(el);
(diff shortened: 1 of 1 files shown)


pxtpy/pydecompiler.ts
```diff
@@ -45,7 +45,7 @@ namespace pxt.py {
     ///
     export const INDENT = "    "
     export function indent(lvl: number): (s: string) => string {
-        return s => `${INDENT.repeat(lvl)}${s}`
+        return s => (s || "").split('\n').map(line => `${INDENT.repeat(lvl)}${line}`).join('\n')
```
shakao, line 48: we use indent everywhere, so this could break things if we were dependent on the previous behavior for multiline strings--i don't think we were, and i tested jumpy platformer, cat jumper, raptor run, and castle crawler (just converting to python and playing the game) and they all seemed to work, but would love a double-check from any of the python people

riknoll: I don't *think* it matters

darzu: Seems like a good solution to me!

pxtpy/pydecompiler.ts
```diff
@@ -1034,6 +1034,12 @@ namespace pxt.py {
                 return exps.map((el, i) => {
                     let sep = el.charAt(el.length - 1) == "," ? "" : ",";
                     if (i == 0) {
+                        let lines = el.split('\n');
+                        // for multiline element, indent all lines after first
+                        if (lines.length > 1) {
+                            let first = lines.shift();
+                            el = first + '\n' + indent1(lines.join('\n'));
+                        }
```
darzu, line 1042: Do you still need this since you made the change to "indent" above?

pxtpy/pydecompiler.ts
```diff
@@ -1213,8 +1219,10 @@ namespace pxt.py {
             return asExpRes(`(${expToStr(inner)})`, innerSup)
         }
         function emitMultiLnStrLitExp(s: ts.NoSubstitutionTemplateLiteral | ts.TaggedTemplateExpression): ExpRes {
-            if (ts.isNoSubstitutionTemplateLiteral(s))
-                return asExpRes(`"""${s.text}"""`)
+            if (ts.isNoSubstitutionTemplateLiteral(s)) {
+                let lines = s.text.split('\n').map((l, i) => !!l.length ? indent1(l) : l);
+                return asExpRes(`"""${lines.join('\n')}"""`)
```
darzu, line 1224: likewise, can't indent1 now handle multiline without splitting before hand?

pxtpy/pydecompiler.ts
shakao: this skips the first line for cases like 

```
let arr = [img("""
    . . .  """)]
```

where we don't want to indent the `img("""` part

pxtpy/pydecompiler.ts
shakao: this is a little bit of a hack but the split->indent->join actually skips the first and last lines, since the symbol text we generate has empty newlines before/after the literal, so we output something like

```
"""
    . . . 
    . 1 .
"""
```

where only the inside is indented..... now that i'm writing this out i think i should probably strip empty lines and just wrap the indented block in non-indented triple quotes. will update!

pxtpy/pydecompiler.ts
darzu: Ohh gotcha

PR #7283, opened 2020-07-08
Optimize use of regex in hotpath

eanders-MS:
This improves compilation speed for large projects. Testing with the game mentioned in https://github.com/microsoft/pxt-arcade/issues/1928, cumulative time spent in the `getCurrentLine` method:
* Before: ~30s
* After: ~2s

### DIFF
--- a/pxtlib/jsoutput.ts
+++ b/pxtlib/jsoutput.ts
@@ -32,12 +32,16 @@ namespace pxt.blocks {
 
     let placeholders: Map<Map<any>> = {};
 
+    const Regex_Newline = /\n/;
+    const Regex_Newline_g = /\n/g;
+    const Regex_LastIndent = /\n *$/;
+
     export function backtickLit(s: string) {
         return "`" + s.replace(/[\\`${}]/g, f => "\\" + f) + "`"
     }
 
     export function stringLit(s: string) {
-        if (s.length > 20 && /\n/.test(s))
+        if (s.length > 20 && Regex_Newline.test(s))
             return backtickLit(s)
         else return JSON.stringify(s)
     }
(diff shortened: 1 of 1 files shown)


mmoskal:
Do the other changes also help? Or is it only the match().length in getCurrentLine? I though JS engines would cache RegExp objects.

mmoskal:
I also wonder if counting \n manually with .charCodeAt() would be faster - there are good chances this would get heavily optimized by v8

eanders-MS:
getCurrentLine was the main hot path by far, but removeLastIndent was present too, though to a much lesser degree. The other replacements came along as a consequence of a search/replace. I'm happy to undo them as they're not contributing any actual improvements.

I will try .charCodeAt and report back.

eanders-MS:
@mmoskal How can I test this? What should I look out for?

eanders-MS:
To clarify: Test against regressing the compiler.

mmoskal:
I was only wondering if pulling out the regexps helps in any way. I assume you can compare profile info. If it doesn't help then we shouldn't do it, especially so that people don't think they have to do it elsewhere. AFAIK V8 optimizes these, and some people say the optimizations may change in future blah blah, but I don't think so.

I don't know much about the blockly tests, I think they run automatically during cloud build, so you should be fine.

eanders-MS:
Interesting, precompiled regex wasn't helping. Here are the times spent in `compileAsync` with different implementations of `getCurrentLine`:
| charAt | precompiled regex | inline regex |
|---|---|--|
| 22s | 6s | 6s |

charAt implementation looked like this:
```ts
for (let i = 0; i < output.length; ++i)
    count += +(output.charAt(i) === '\n');
```

eanders-MS:
Just for curiosity I tested with charCodeAt, it was exactly the same as charAt.

PR #7284, opened 2020-07-09
Fix snippets for shim=TD_ID parameters

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/2068

Is there a way to add snippet tests?


darzu:
LGTM. No way to add snippet tests yet, but I'm working on adding a way. I'll see about including a test case for this

PR #7285, opened 2020-07-09
Use rightmost part for python callback naming

shakao:
We were previously concatenating the entire qname which resulted in controller.combos.attachCombo -> combos_controller.combos.attach_combo during the recursion.

Tested: Warehouese, Jumping Mateo, HackMan, Raptor Run

Fixes Warehouse and Jumping Mateo from https://github.com/microsoft/pxt-arcade/issues/2083

### DIFF
--- a/pxtpy/pydecompiler.ts
+++ b/pxtpy/pydecompiler.ts
@@ -934,11 +934,11 @@ namespace pxt.py {
             // TODO(dz): Impl skip namespaces properly. Right now we just skip the left-most part of a property access
             if (ts.isPropertyAccessExpression(s)) {
                 let nmPart = getName(s.name)
+                let nmRight = nmPart.substr(nmPart.lastIndexOf(".") + 1)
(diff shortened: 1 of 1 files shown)


pxtpy/pydecompiler.ts
```diff
@@ -934,11 +934,11 @@ namespace pxt.py {
             // TODO(dz): Impl skip namespaces properly. Right now we just skip the left-most part of a property access
```
darzu, line 934: I think this addresses this comment now.

darzu:
If it's easy, want to add a test case?

shakao:
i couldn't quite figure out how to get a test going since we only hit this case when we have the qname from the typechecker, otherwise we output generic my_function, my_function2 names; including the namespaces as part of the test file didn't work, let me know if i'm missing something obvious!

abchatra:
Merging now for tomorrow testing. Feel free to add test case later.

PR #7286, opened 2020-07-09
Fix snippets for fixedInstances outside block namespace

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/2074

this uses the the namespace from the fixedInstance symbol we're emitting instead of from the blocknamespace

I don't think we have a test suite for these yet?


riknoll:
Wait, should we ever use blockNamespace? Isn't it only there to override the actual namespace for the categories in the toolbox?

jwunderl:
Right, I left that in as it would only be hit for the path of 'no matching fixedInstance available', but just using the namespace from the method itself seems like the right behavior in that case (and I don't think it's really a supported scenario? I feel like I've ended up with broken toolboxes from not having defined fixedInstances in the past)

abchatra:
Merging now for tomorrow testing

PR #7287, opened 2020-07-09
Fix the monaco sprite editor for python

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/2079

Three changes:

1. Emit the python syntax and not typescript when we are editing python
2. Made the regex a little more robust
3. Register the folding range provider for python (this is what actually makes the icon show up)


abchatra:
Merging for tomorrow testing.

PR #7288, opened 2020-07-09
Compiler perf: Count line numbers progressively

eanders-MS:
This is another attempt to squeeze perf out of `compileAsync`. Previous attempt here: https://github.com/microsoft/pxt/pull/7283, which brought it from 20s to 6s. That was a nice improvement, but still noticeably slow. Getting better results this time:

Time spent in `compileAsync`:
| Before | After |
|---|---|
| 6s | 500ms |

Testing with this project: https://arcade.makecode.com/74548-67333-76127-33508

Related: https://github.com/microsoft/pxt-arcade/issues/1928

### DIFF
--- a/pxtlib/jsoutput.ts
+++ b/pxtlib/jsoutput.ts
@@ -288,7 +288,7 @@ namespace pxt.blocks {
 
         function append(s: string) {
             // At runtime sometimes `s` is a number.
-            if (s?.match) {
+            if (typeof(s) === 'string') {
                 currentLine += (s.match(/\n/g) || []).length;
             }
             output += s;
--- a/pxtlib/jsoutput.ts
+++ b/pxtlib/jsoutput.ts
@@ -284,6 +284,15 @@ namespace pxt.blocks {
         let output = ""
         let indent = ""
         let variables: Map<string>[] = [{}];
+        let currentLine = 0;
+
+        function append(s: string) {
+            // At runtime sometimes `s` is a number.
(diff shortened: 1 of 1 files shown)


eanders-MS:
Hm, the PR title isn't isn't accurate (will update). Still using a regexp, but no longer scanning the whole program each time.

pxtlib/jsoutput.ts
```diff
@@ -284,6 +284,15 @@ namespace pxt.blocks {
         let output = ""
         let indent = ""
         let variables: Map<string>[] = [{}];
+        let currentLine = 0;
+
+        function append(s: string) {
+            // At runtime sometimes `s` is a number.
+            if (s && s.match) {
```
riknoll, line 291:
```suggestion
            if (s?.match) {
```

pxtlib/jsoutput.ts
```diff
@@ -284,6 +284,15 @@ namespace pxt.blocks {
         let output = ""
         let indent = ""
         let variables: Map<string>[] = [{}];
+        let currentLine = 0;
+
+        function append(s: string) {
+            // At runtime sometimes `s` is a number.
+            if (s?.match) {
```
mmoskal, line 291: maybe `typeof(s) === "string"` ?

eanders-MS: I think I will go with that. It is more accurate in this case, and reinforces the comment above it.

PR #7290, opened 2020-07-10
Use annotation in python type unification

shakao:
We're not actually using the annotation during assignment type unification, which causes problems for this case:

```
arr: List[Sprite] = []
for s in arr:
    s.destroy()  # <- can't find property "destroy"
```

Fixes "Chomper" for https://github.com/microsoft/pxt-arcade/issues/2083


abchatra:
Merging for tomorrow testing.

PR #7291, opened 2020-07-10
Fix auto-complete snippet cursor placement

darzu:
Fixes: https://github.com/microsoft/pxt-arcade/issues/2076

Notice how awesome the last behavior is!! It both auto completes with a working example AND we get the cursor over the first arg.

Basically we have to track when generating snippets where the identifiers that we might want to replace are. So we defer stringify'ing the snippet.

Before (last release):
before

Regression (beta):
regression

After:

Some TODOs:
- [x] multiple args "(false, false)" issue
- [x] function selection
- [x] python testing


pxtcompiler/emitter/service.ts
```diff
@@ -1254,11 +1256,18 @@ namespace ts.pxtc.service {
...
+                        const addsDefinitions = snippetAddsDefinitions(snippetNode)
+                        if (isPython) {
                             si.pySnippet = snippet
-                        else
+                            si.snippetWithMarkers = snippetWithMarkers
+                        }
+                        else {
                             si.snippet = snippet
+                            si.snippetWithMarkers = snippetWithMarkers
+                        }
```
jwunderl, line 1270:
```suggestion
                        }
                        else {
                            si.snippet = snippet
                        }
                        si.snippetWithMarkers = snippetWithMarkers
```

maybe pull out ` si.snippetWithMarkers = snippetWithMarkers` to be outside of the conditional as it's the same in both cases?

pxtcompiler/emitter/service.ts
```diff
@@ -1723,22 +1734,61 @@ namespace ts.pxtc.service {
...
-            addsDefinitions: false
+    function isSnippetReplacePoint(n: SnippetNode): n is SnippetReplacePoint {
+        return !!(n as SnippetReplacePoint).default
+    }
+    function isSnippetNodeList(n: SnippetNode): n is SnippetNode[] {
+        return typeof (n) === "object" && typeof ((n as SnippetNode[]).length) === "number"
+    }
+    export function snippetStringify(snippet: SnippetNode, emitMonacoReplacementPoints = false): string {
+        // The format for monaco snippets is: 
+        //      foo(${1:bar}, ${2:baz},  ${1:bar})
```
jwunderl, line 1750: Hm, so is the multicursor behavior for same type parameters intended? I always found that behavior a bit confusing personally

pxtcompiler/emitter/service.ts
```diff
@@ -1723,22 +1734,61 @@ namespace ts.pxtc.service {
...
+        // so both instances of "bar" will start highlighted, then tab will cycle to "baz", etc.
+        const namesToReplacementNumbers: { [key: string]: number } = {}
```
jwunderl, line 1752:
```suggestion
        const namesToReplacementNumbers: pxt.Map<Number> = {}
```

pxtcompiler/emitter/service.ts
```diff
@@ -1723,22 +1734,61 @@ namespace ts.pxtc.service {
...
+                    nextNum++
+                    namesToReplacementNumbers[name] = num
+                }
+                return "${" + num + ":" + name + "}"
+            } else {
+                return snippetStringify(snippet.default, emitMonacoReplacementPoints)
+            }
+        } else if (isSnippetNodeList(snippet)) {
+            const res = snippet.map(s => snippetStringify(s, emitMonacoReplacementPoints)).join("")
+            return res
```
jwunderl, line 1769:
```suggestion
            return snippet.map(s => snippetStringify(s, emitMonacoReplacementPoints)).join("")
```

pxtcompiler/emitter/service.ts
darzu: Multi cursor should only be for the same identifiers. I have a bug right now that hit happens for the same types like "false" but I'll fix that.

pxtcompiler/emitter/service.ts
darzu: Oops it's supposed to be different in each case, thanks!

PR #7293, opened 2020-07-10
Add select all, delete, and cancel selection to sprite editor

jacobcarpenter:
closes microsoft/pxt-arcade#2093

This PR adds shortcuts for the following actions in the sprite/tile/background image editor:

* `Cmd-A`/`Ctrl-A`: Select all
* `Esc`: Cancel selection
* `Backspace`/`Delete`: Delete selection

Kapture 2020-07-09 at 20 58 38

### DIFF
--- a/webapp/src/components/ImageEditor/ImageCanvas.tsx
+++ b/webapp/src/components/ImageEditor/ImageCanvas.tsx
@@ -247,16 +247,14 @@ class ImageCanvasImpl extends React.Component<ImageCanvasProps, {}> implements G
                 ev.stopPropagation();
             }
 
-            if ((ev.ctrlKey || ev.metaKey) && ev.key === 'a') {
-                // only handle select all if the canvas has the focus or if nothing has the focus
-                const shouldHandleSelectAll = document.activeElement === this.refs["canvas-bounds"] ||
-                    document.activeElement === document.body ||
-                    !document.activeElement;
+            if ((ev.ctrlKey || ev.metaKey) && ev.key === 'a' && this.shouldHandleCanvasShortcut()) {
+                this.selectAll();
+                ev.preventDefault();
+            }
 
-                if (shouldHandleSelectAll) {
-                    this.selectAll();
-                    ev.preventDefault();
-                }
+            if ((ev.key === "Backspace" || ev.key === "Delete") && this.editState?.floating?.image && this.shouldHandleCanvasShortcut()) {
+                this.deleteSelection();
+                ev.preventDefault();
             }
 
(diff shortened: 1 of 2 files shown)


webapp/src/components/ImageEditor/ImageCanvas.tsx
```diff
@@ -95,6 +95,9 @@ class ImageCanvasImpl extends React.Component<ImageCanvasProps, {}> implements G
     }
 
     componentDidMount() {
+        // move initial focus off of the blockly surface
+        (document.activeElement as HTMLElement)?.blur();
```
riknoll, line 99: @jwunderl is this okay for accessibility?

riknoll:
@jacobcarpenter did focusing the canvas not work? I notice it was in the history of commits

jacobcarpenter:
@riknoll yes, an earlier version did make canvas focusable. It had a couple of drawbacks though:

1. After the `canvas-bounds` received focus, the browser outlined it, which I disabled—and I'm aware that carries accessibility implications...
2. Setting `tabindex` on the canvas-bounds wasn't sufficient to get the browser to move the focus to that element when the user clicked on it. I believe other click interactions on the canvas surface are already `preventDefault`-ed, but I didn't thoroughly investigate. [Instead, I temporarily added a click handler that explicitly moved the focus to the element...]
3. Interactions with other parts of the editor also end up blurring the focus. The tool buttons, for instance, are `div`s rather than buttons, and they don't accept the focus when the user clicks. There were enough normal interactions that ended up clearing the focus that the code needed to account for that state too, so I ended up removing the focusable `canvas-bounds`.

Let me know if this isn't an acceptable compromise; I'd be happy to make changes to get this functionality in.

webapp/src/components/ImageEditor/ImageCanvas.tsx
jwunderl: I think it's better than the current state (currently focus is being left on the svg of the preview image they clicked on)

That said definitely some accessibility work in general for image editor - right now it doesn't capture focus like a modal typically would (which is presumably why this is here in the first place, to get focus off hidden elements), and only the size inputs have tabstops - there's actually no obvious way to escape once you've entered the image editor keyboard, as the done button doesn't have a tabstop.

jacobcarpenter:
@riknoll @jwunderl any further thoughts on this one? I completely understand you guys are busy prepping for a new release.

But I'd love to do what I can to help get this functionality in. 😅

jwunderl:
As always, thank you so much for your contribution @jacobcarpenter! :)

PR #7295, opened 2020-07-10
Fixes sprite kinds completion / toolbox snippet

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/1898

took the kindNamespace resolution (defined kindNamespace >  blockNamespace > namespace) from https://github.com/microsoft/pxt/blob/master/pxtlib/service.ts#L512

I wasn't completely sure where in `getBlocksInfo` to include this check, but this seemed like a reasonable spot to start with.

fix-kinds

Side note, looks like block kinds still don't allow default values; that is, adding

```typescript
//% kind.defl=SpriteKind.Player
//% otherKind.defl=SpriteKind.Enemy
```

here https://github.com/microsoft/pxt-common-packages/blob/master/libs/game/spriteevents.ts#L51 doesn't cause `otherKind` to start off as `Enemy` in blocks. (It does in the js/py completions / snippets though)

Maybe something to try and fix, I can give it a look? I guess depends on when we end up releasing / if there's time.

### DIFF
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -1958,7 +1958,7 @@ namespace ts.pxtc.service {
                     const kindNamespace = shadowAttrs.kindNamespace || shadowAttrs.blockNamespace || fn.namespace;
                     const defaultValueForKind = pxtc.Util.values(apis.byQName).find(api => api.namespace === kindNamespace && api.attributes.isKind);
                     if (defaultValueForKind) {
-                        return defaultValueForKind.qName;
+                        return python ? defaultValueForKind.pyQName : defaultValueForKind.qName;
                     }
                 }
 
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -1955,7 +1955,7 @@ namespace ts.pxtc.service {
 
                 const shadowAttrs = shadowSymbol.attributes;
                 if (shadowAttrs.shim === "KIND_GET" && shadowAttrs.blockId) {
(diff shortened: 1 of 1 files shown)


pelikhan:
> Maybe something to try and fix, I can give it a look? I guess depends on when we end up releasing / if there's time.

Separate fix.

pxtcompiler/emitter/service.ts
```diff
@@ -1953,6 +1953,15 @@ namespace ts.pxtc.service {
                     }
                 }
 
+                const shadowAttrs = shadowSymbol.attributes;
```
darzu, line 1956: Maybe add a quick comment saying "kindNamespace > blockNamespace > namespace" ?

riknoll: block namespace is just for blocks, should not be used for snippets

riknoll: it's for overriding the "real" namespace but in TS that doesn't matter

jwunderl: Oh, I thought it would matter in this case as it's being used to look up the kind in `apis.byQName`, so we would want to look it up with the same namespace as a 'key'.

riknoll: How so? The only information on the namespace is toolbox info, right? like color, icon, group names, etc?

jwunderl: Just so that it matches and the same default value would be picked for blocks and typescript (as the initial members get picked here: https://github.com/microsoft/pxt/blob/master/pxtlib/service.ts#L522 and the first member found will be the one it defaults to).

I don't think this is an actual case that ever gets hit though, as kinds should have a defined kindNamespace that would take priority; will drop the blockNamespace here

PR #7296, opened 2020-07-10
Handle the other format of the tile tag in sprite meta json

riknoll:
Part of the fix for https://github.com/microsoft/pxt-arcade/issues/2103
Used to build https://github.com/microsoft/pxt-arcade/pull/2109

@jwunderl any other forms of the tags I should know about?


jwunderl:
The other form is "!{tag}"; I suppose technically we wanna bail out if there's a !tile before these, in case a metajson says 'these are all tiles' and the png specific json says 'just kidding, these ones aren't actually tiles'.

I find it very unlikely we'll ever end up doing / running into that though

pelikhan:
Documented?

jwunderl:
under https://makecode.com/cli/buildsprites

PR #7299, opened 2020-07-11
Group errors by line number and show count

eanders-MS:
Duplicate compiler errors are grouped, and a count bubble is shown (if count > 1).

Before:
image

After:
image

(screenshots not taken from the same project, so total error count is different)

Fixes: https://github.com/microsoft/pxt-arcade/issues/2105

### DIFF
--- a/webapp/src/errorList.tsx
+++ b/webapp/src/errorList.tsx
@@ -209,7 +209,5 @@ function groupErrors(errors: pxtc.KsDiagnostic[]) {
             grouped.get(key).count++;
         }
     }
-    let sorted: GroupedError[] = [];
-    grouped.forEach(value => sorted.push(value));
-    return sorted.sort((a, b) => a.index - b.index);
+    return [...grouped.values()].sort((a, b) => a.index - b.index);
 }
--- a/webapp/src/errorList.tsx
+++ b/webapp/src/errorList.tsx
@@ -49,15 +49,15 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
     render() {
         const {isCollapsed, errors, exception} = this.state;
 
-        const errorCount = exception ? 1 : errors.length;
-        const errorsAvailable = errorCount > 0 || !!exception;
         const errorListContent = !isCollapsed ? (exception ? this.generateStackTraces(exception) : this.getCompilerErrors(errors)) : undefined;
+        const errorsAvailable = !!errors?.length || !!exception;	
+        const collapseTooltip = lf("Collapse Error List");
 
         return (
             <div className={`errorList ${isCollapsed ? 'errorListSummary' : ''}`} hidden={!errorsAvailable}>
                 <div className="errorListHeader" role="button" aria-label={lf("{0} error list", isCollapsed ? lf("Expand") : lf("Collapse"))} onClick={this.onCollapseClick} onKeyDown={sui.fireClickOnEnter} tabIndex={0}>
                     <h4>{lf("Problems")}</h4>
-                    <div className="ui red circular label countBubble">{exception ? 1 : errorCount}</div>
+                    <div className="ui red circular label countBubble">{exception ? 1 : errors.length}</div>
                     <div className="toggleButton"><sui.Icon icon={`chevron ${isCollapsed ? 'up' : 'down'}`}/></div>
                 </div>
                 {!isCollapsed && <div className="errorListInner">
--- a/webapp/src/errorList.tsx
+++ b/webapp/src/errorList.tsx
@@ -48,8 +48,9 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
 
     render() {
         const {isCollapsed, errors, exception} = this.state;
-        const errorsAvailable = !!errors?.length || !!exception;
         const collapseTooltip = lf("Collapse Error List");
+        const errorsAvailable = !!errors?.length || !!exception;
+    
         const errorListContent = !isCollapsed ? (exception ? this.generateStackTraces(exception) : this.getCompilerErrors(errors)) : undefined;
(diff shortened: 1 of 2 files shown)


webapp/src/errorList.tsx
```diff
@@ -42,16 +48,17 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
 
     render() {
         const {isCollapsed, errors, exception} = this.state;
-        const errorsAvailable = !!errors?.length || !!exception;
-        const collapseTooltip = lf("Collapse Error List");
 
+        let errorCount = 0;
+        (errors || []).forEach(e => errorCount += e.count);
```
pelikhan, line 53: nit: ``(errors || []).reduce((e, count) => e.count + count)``

webapp/src/errorList.tsx
```diff
@@ -158,20 +165,53 @@ class ErrorListItem extends React.Component<ErrorListItemProps, ErrorListItemSta
...
+function errorMessageStringWithLineNumber(error: pxtc.KsDiagnostic) {
+    return lf("Line {0}: {1}", error.endLine ? error.endLine + 1 : error.line + 1, error.messageText);
+}
+
+function stackFrameMessageStringWithLineNumber(stackframe: pxsim.StackFrameInfo, location: pxtc.LocationInfo) {
+    return lf("at {0} (line {1})", stackframe.funcInfo.functionName, location.line + 1);
+}
+
+function groupErrors(errors: pxtc.KsDiagnostic[]) {
+    const grouped: { [key: string]: GroupedError} = {};
```
pelikhan, line 197: nit - we have a type def for maps: ``Map<GroupedError>``

webapp/src/errorList.tsx
```diff
@@ -94,7 +101,7 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
 
     onErrorsChanged(errors: pxtc.KsDiagnostic[]) {
         this.setState({
-            errors,
+            errors: groupErrors(errors),
```
pelikhan, line 104: IMO, we should lazily group errors **if** the error list is visible. In most cases, it is collapsed and this computation is not required.

webapp/src/errorList.tsx
eanders-MS: > nit - we have a type def for maps: `Map<GroupedError>`

Oh yes, the Map type!

webapp/src/errorList.tsx
eanders-MS: > IMO, we should lazily group errors **if** the error list is visible. In most cases, it is collapsed and this computation is not required.

Agreed that's better. Updated.

webapp/src/errorList.tsx
eanders-MS: > nit: `(errors || []).reduce((e, count) => e.count + count)`

Delayed grouping made this unnecessary. Reverted my changes in this method.

webapp/src/errorList.tsx
darzu: `{ [key: string]: GroupedError}` is standard TS syntax and we should prefer that IMO. I'd like to bring this up at the next code review day

webapp/src/errorList.tsx
```diff
@@ -158,20 +165,51 @@ class ErrorListItem extends React.Component<ErrorListItemProps, ErrorListItemSta
...
+                count: 1,
+                error
+            });
+        }
+        else {
+            grouped.get(key).count++;
+        }
+    }
+    let sorted: GroupedError[] = [];
+    grouped.forEach(value => sorted.push(value));
```
darzu, line 213: sorted = [...grouped].sort((a, b) => a.index - b.index)

webapp/src/errorList.tsx
```diff
@@ -158,20 +165,51 @@ class ErrorListItem extends React.Component<ErrorListItemProps, ErrorListItemSta
...
+function errorMessageStringWithLineNumber(error: pxtc.KsDiagnostic) {
+    return lf("Line {0}: {1}", error.endLine ? error.endLine + 1 : error.line + 1, error.messageText);
+}
+
+function stackFrameMessageStringWithLineNumber(stackframe: pxsim.StackFrameInfo, location: pxtc.LocationInfo) {
+    return lf("at {0} (line {1})", stackframe.funcInfo.functionName, location.line + 1);
+}
+
+function groupErrors(errors: pxtc.KsDiagnostic[]) {
+    const grouped = new Map<string, GroupedError>();
```
darzu, line 197: I'm curious, what are the advantages of using a map type over vanilla objects?

pelikhan: we've been using consistently the Map type accross the entire PXT codebase; so to stay consistent, keep using it. Also, don't have to lookup the TS index syntax every time.

webapp/tsconfig.json
```diff
@@ -16,6 +16,7 @@
         "emitDecoratorMetadata": true,
         "declaration": false,
         "preserveConstEnums": true,
+        "downlevelIteration": true,
```
eanders-MS, line 19: @pelikhan use of spread operator on map values required the addition of this compiler option (due to targeting es5). Any issue?

pelikhan: You'd have to ask this to the language gurus, @riknoll @mmoskal . we should decide where to draw the line w.r.t. to typescript features, specially because we might start to confuse STS vs TS versions.
imo: not worth it.

eanders-MS: I'll revert the change to avoid risk. Curious what the language gurus think though.

mmoskal: I think it should be fine. There's lots of things we do in TS that we cannot do in STS (like async).

PR #7302, opened 2020-07-13
A couple tilemap bugs

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/2098

Also fixes part of the issue https://github.com/microsoft/pxt-arcade/issues/2125

And makes it so that the generated tilemap and tileset files can't be deleted

### DIFF
--- a/pxteditor/monaco-fields/field_tilemap.ts
+++ b/pxteditor/monaco-fields/field_tilemap.ts
@@ -170,14 +170,6 @@ namespace pxt.editor {
         }
     }
 
-    const defaultTileset = `{
-        "*": {
-            "namespace": "myTiles",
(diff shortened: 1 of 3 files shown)


pxteditor/monaco-fields/field_tilemap.ts
```diff
@@ -8,15 +8,6 @@ namespace pxt.editor {
         protected tilemapName: string;
         protected isTilemapLiteral: boolean;
 
-        protected initAsync(): Promise<void> {
```
shakao, line 11: was this just unused? can 'defaultTileset' below be removed as well?

riknoll: yep will fix

PR #7304, opened 2020-07-13
github: revert all + support for generated files

pelikhan:
* revert all button
image
* revert sibling generated files
image

fix for https://github.com/microsoft/pxt-arcade/issues/1964


pelikhan:
Double check on the revert logic please. There might be a much easier way that I completely overlooked.

PR #7306, opened 2020-07-14
Auto-complete: Rank symbols matching current parameter type higher; rank keywords lower

darzu:
Fixes: https://github.com/microsoft/pxt-arcade/issues/2070

Fixes the way we were sorting completion symbols that match the type of the current parameter. Also deprioritized keywords by default since we only have them in the completion list so that we don't accidently auto complete a symbol when someone types "if".

After:

### DIFF
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -1161,8 +1161,7 @@ namespace ts.pxtc.service {
                             paramIdx = callSym.parameters.length - 1
                         const paramType = getParameterTsType(callSym, paramIdx, blocksInfo)
                         if (paramType) {
-                            // if this is a property access, then weight the results higher if they return the
-                            // correct type for the parameter
+                            // weight the results higher if they return the correct type for the parameter
                             const MATCHING_TYPE_WEIGHT = 10
                             const matchingApis = getApisForTsType(paramType, call, tc, resultSymbols);
                             matchingApis.forEach(match => match.weight = MATCHING_TYPE_WEIGHT);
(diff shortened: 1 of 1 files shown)


pxtcompiler/emitter/service.ts
```diff
@@ -1158,26 +1163,18 @@ namespace ts.pxtc.service {
                         if (paramType) {
                             // if this is a property access, then weight the results higher if they return the
```
jwunderl, line 1164: should these two lines of comments be dropped? Does not appear to only be applying to property accesses anymore.

pxtcompiler/emitter/service.ts
```diff
@@ -1203,7 +1200,7 @@ namespace ts.pxtc.service {
                     .map(s => completionSymbol(s))
 
                 // in scope locals should be weighter higher
-                inScopePxtSyms.forEach(s => s.weight += 100)
+                inScopePxtSyms.forEach(s => s.weight += 5)
```
jwunderl, line 1203: Not in this PR, but maybe worth going through and setting some constants to make 'visualizing' relative weights easier.

pxtcompiler/emitter/service.ts
```diff
@@ -1107,6 +1107,11 @@ namespace ts.pxtc.service {
                 }
             }
 
+            if (resultSymbols.length === 0) {
```
pelikhan, line 1110: what does this mean, if the user enters garbage it receives a ton of suggestions?

darzu: It means that: "if by this point we don't yet have a specialized set of results (like those for member completion or a specific type for a call expression), use all symbols as the start"
So yes, this results in a flood of symbols but monaco filters and searches these down based on what you're typing. This is how vs code works. Type "fo" and it'll show you all symbols with "f" and "o" in them sorted appropriately.

darzu: I updated the comment to clarify

PR #7311, opened 2020-07-14
Option to disable auto-opening of readme in makecode

pelikhan:
image


pelikhan:
@shakao it would be nice to pickup the docsrender.ts changes in the cloud too.

PR #7313, opened 2020-07-15
collect client size

pelikhan:
We don't collect that actual size of the document in our analytics; so we don't know how much space is lost with toolbars, etc...


webapp/src/app.tsx
```diff
@@ -3962,6 +3962,9 @@ function enableAnalytics() {
         stats["screen.innerWidth"] = window.innerWidth;
         stats["screen.innerHeight"] = window.innerHeight;
         stats["screen.devicepixelratio"] = pxt.BrowserUtils.devicePixelRatio();
+        const body = document.firstElementChild; // body
```
abchatra, line 3965: Can this be null?

PR #7314, opened 2020-07-15
Asset editor message handling and loading image from jres

kimprice:
Set up message handling from vscode extension and loading image in the editor from the JRes URL.


webapp/src/assetEditor.tsx
```diff
@@ -16,11 +16,26 @@ function init() {
 }
 
 export class AssetEditor extends React.Component {
+    private editor: FieldEditorComponent<any>;
+
+    handleMessage = (msg: any)  => {
+        if (msg.data._fromVscode && msg.data.type === "update") {
```
pelikhan, line 22: you should define interface that specify the message format.

pelikhan: can data be null/undefined? use ?.

webapp/src/assetEditor.tsx
```diff
@@ -16,11 +16,26 @@ function init() {
...
+            this.editor.loadJres(msg.data.message);
+        }
+    }
+
+    refHandler = (e: FieldEditorComponent<any>) => {
+        this.editor = e;
+    }
+
+    componentDidMount() {
+        window.onmessage = this.handleMessage;
```
pelikhan, line 32: best to use ``window.addEventHandler("message", this.handleMessage, null)`` and remove handler in unmount

webapp/src/assetEditor.tsx
```diff
@@ -16,11 +16,26 @@ function init() {
...
+        window.parent.postMessage({type: "ready" }, "*")
```
pelikhan, line 33: specify protocol of message using interfaces

webapp/src/assetEditor.tsx
pelikhan: why is there a underscore in ``_fromVscode``? it does not really follow any convention.

webapp/src/assetEditor.tsx
```diff
@@ -16,11 +16,26 @@ function init() {
...
+    }
+
     render() {
-        return <ImageFieldEditor singleFrame={true} />
+        return <div><ImageFieldEditor ref={this.refHandler} singleFrame={true} />
```
pelikhan, line 37: <div> not needed?

webapp/src/components/ImageFieldEditor.tsx
```diff
@@ -167,6 +167,10 @@ export class ImageFieldEditor<U extends ImageType> extends React.Component<Image
         });
     }
 
+    loadJres(jres: string) {
+        this.ref.setCurrentFrame(pxt.sprite.getBitmapFromJResURL(jres));
```
pelikhan, line 171: make sure to handle invalid jres here.

webapp/src/assetEditor.tsx
kimprice: image
For the '_fromVscode', I just tried to follow what was written for simulator messages in the vscode extension. I'm not sure the reason for the underscore, maybe because it should always be set to true for messages from the extension?

kimprice: Yea the '_fromVscode' is the way it's specified in the simulator messages in the vscode extension so I just thought I should be consistent, but I can change it.

pelikhan: we should specify the message format here in makecode for consumers. I would drop the ``_fromVscode`` and instead use something like ``"source": "vscode"``

kimprice: The message will always have a data attribute.

PR #7317, opened 2020-07-15
Objects should not be assignable to Function type

mmoskal:
Before this change the following code would pass the type checker:

```typescript
const foobar = 1
foobar()
```

This was problematic because people were using `sprite.x()` or something, which would kind of work, except the return type would be `any`.

Note that this may break existing scripts, but I would rather take it and have people fix the scripts. The current behavior is non-standard, confusing, and bug-prone.

Fixes https://github.com/microsoft/pxt-arcade/issues/2019

### DIFF
--- a/libs/pxt-common/pxt-core.d.ts
+++ b/libs/pxt-common/pxt-core.d.ts
@@ -400,9 +400,15 @@ declare function parseFloat(text: string): number;
 declare function randint(min: number, max: number): number;
 
 interface Object { }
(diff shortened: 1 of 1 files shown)


mmoskal:
Also note that there are two errors like that in arcade library

pelikhan:
All games using pxt-sprite-data will break.

riknoll:
We should force a package update for that one...

PR #7318, opened 2020-07-15
Set canIndentInside to true for multiline arguments

shakao:
Fixes TS image literal indentation https://github.com/microsoft/pxt-arcade/issues/2143


pelikhan:
is it possible to get a general test for indentation?

PR #7320, opened 2020-07-15
alpha-sort deps for c++ compilation to avoid missing valid hex files

jwunderl:
should fix (important part of) https://github.com/microsoft/pxt-arcade/issues/2127 -- the sha was changing because the package's dependencies would change the topo sort order of the files, even though the set of c++ files wasn't changing. Michal mentioned the order doesn't really matter for these c++ / that making alphabetical is fine. (I'll see about fixing the message in a separate PR)

Tested with a trivial arcade program and it worked (&& worked when reversing the sort order as well, so should be good)

### DIFF
--- a/pxtlib/cpp.ts
+++ b/pxtlib/cpp.ts
@@ -170,7 +170,7 @@ namespace pxt.cpp {
             .sort((a, b) => {
                 if (a.id == "this") return 1;
                 else if (b.id == "this") return -1;
-                else return a.id.localeCompare(b.id);
+                else return U.strcmp(a.id, b.id);
             });
 
         for (let pkg of mainPkgDeps) {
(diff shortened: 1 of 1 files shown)


pxtlib/cpp.ts
```diff
@@ -163,7 +163,17 @@ namespace pxt.cpp {
...
         let mainDeps: Package[] = []
-        for (let pkg of mainPkg.sortedDeps(true)) {
+
+        // order shouldn't matter for c++ compilation,
+        // so use a stable order to prevent order changes from fetching a new hex file
+        const mainPkgDeps = mainPkg.sortedDeps(true)
+            .sort((a, b) => {
+                if (a.id == "this") return 1;
+                else if (b.id == "this") return -1;
+                else return a.id.localeCompare(b.id);
```
mmoskal, line 173: Use U.strcmp or at least compare. Locale may make it change based on language settings

PR #7322, opened 2020-07-15
Try to parse 'any' values when emitting snippets

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/2078

Two issues:
1. I'm a Javascript beginner, so lemme know if there are better ways to check for all these cases. 
2. Also I only tested for the sprite.say string changes. Are there other blocks I should test with?

### DIFF
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -1919,6 +1919,9 @@ namespace ts.pxtc.service {
                     return snippet
                 }
                 return attrs.paramDefl[name];
+            } else {
+                // default to empty string
+                return "\"\"";
             }
 
             function getDefaultValueOfType(type: ts.Type): SnippetNode | undefined {
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -1919,9 +1919,6 @@ namespace ts.pxtc.service {
                     return snippet
                 }
                 return attrs.paramDefl[name];
-            } else {
-                // default to empty string
-                return "\"\"";
             }
 
             function getDefaultValueOfType(type: ts.Type): SnippetNode | undefined {
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -1897,21 +1897,22 @@ namespace ts.pxtc.service {
 
             // check for explicit default in the attributes
             if (attrs && attrs.paramDefl && attrs.paramDefl[name]) {
-                // if (typeNode.kind == SK.AnyKeyword) {
-                //     if (typeNode.) {
-                //         // try to parse as a number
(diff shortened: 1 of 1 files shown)


pxtcompiler/emitter/service.ts
```diff
@@ -1897,6 +1897,22 @@ namespace ts.pxtc.service {
 
             // check for explicit default in the attributes
             if (attrs && attrs.paramDefl && attrs.paramDefl[name]) {
+                if (typeNode.kind == SK.AnyKeyword) {
+                    const defaultName = attrs.paramDefl[name].toUpperCase();
+                    if (!Number.isNaN(+defaultName)) {
+                        // try to parse as a number
+                        typeNode.kind = SK.NumberKeyword;
+                    } else if (defaultName == "FALSE" || defaultName == "TRUE") {
```
riknoll, line 1905: should probably lowercase it before checking, in TS it's lowercase and in python it's uppercase

riknoll: never mind! I can't read

livcheerful: ahaha i can see into the future

riknoll:
As for other blocks, check out console log and logValue

livcheerful:
> As for other blocks, check out console log and logValue

cool cool i'll do that now :D

pxtcompiler/emitter/service.ts
```diff
@@ -1897,6 +1897,22 @@ namespace ts.pxtc.service {
 
             // check for explicit default in the attributes
             if (attrs && attrs.paramDefl && attrs.paramDefl[name]) {
+                if (typeNode.kind == SK.AnyKeyword) {
+                    const defaultName = attrs.paramDefl[name].toUpperCase();
+                    if (!Number.isNaN(+defaultName)) {
```
jwunderl, line 1902:
```suggestion
                    if (!isNaN(defaultName)) {
```

isNaN coerces then checks, Number.isNaN does not coerce / behaves wonkily on non num input; for that reason I much prefer isNaN personally so that you don't get the "well it's not technically NaN" result (that is, `Number.isNaN("abc")` returns false, even though "abc" is not a number, which feels wrong / top level isNaN behaves as I would expect).

(if you / anyone else has a particular opinion against that I'd yield though, hah)

livcheerful: oo okay so I just tried to do this and it's mad because defaultName isn't a number. is this not supposed to happen?

jwunderl: Oh hmm, I didn't realize typescript had the typing for isNaN stricter than browsers behavior. I guess it doesn't make a difference in that case at all if you're forced to coerce, I'm used to using it from javascript code but no difference here then!

PR #7323, opened 2020-07-15
Show 'hex file is not available' error when offline

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/2127 (the other error re: not getting the right hex file was fixed in https://github.com/microsoft/pxt/pull/7320)

Propagate the error message from https://github.com/microsoft/pxt/blob/master/pxtcompiler/emitter/emitter.ts#L967 in the toast

image


jwunderl:
It's the only diagnostic being returned from that path right now, but switched it up as it's clearer that way / future proof in case some day we decide to attach a reason the hex build failed or something like that

PR #7326, opened 2020-07-16
fix importing share url for stable versions (v0, v1, ...)

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/2155

was waiting on arcade build so I figured I'd fix this one, didn't realize till after that it was assigned, oops


abchatra:
Looks good. Thanks!

PR #7327, opened 2020-07-16
bring image blocks size back down to normal (reduce vert padding)

jwunderl:
Reduce vertical padding around img blocks back down to what it is on live site:

live site:

image

/beta:

image

after this change:

image

Not certain why this behavior changed; this code hasn't changed for ~2 years, so it seems like it must have been being compressed vertically in blockly and that got changed with the new renderer? I could look around if we want, but maybe after release

### DIFF
--- a/pxtblocks/fields/field_animation.ts
+++ b/pxtblocks/fields/field_animation.ts
@@ -20,12 +20,13 @@ namespace pxtblockly {
     // 32 is specifically chosen so that we can scale the images for the default
     // sprite sizes without getting browser anti-aliasing
     const PREVIEW_WIDTH = 32;
-    const PADDING = 5;
+    const X_PADDING = 5;
+    const Y_PADDING = 1;
     const BG_PADDING = 4;
     const BG_WIDTH = BG_PADDING * 2 + PREVIEW_WIDTH;
     const ICON_WIDTH = 30;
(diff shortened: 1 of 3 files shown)


jwunderl:
Yeah, that's what I was thinking.

Just noticed I should apply same fix for the animation and tilemap field editors, I had thought they were extending from this one; 'do not merging' this one for a second till I fix those too

jwunderl:
Okay here's a new comparison with animation / tmap fixed

image

the reporter blocks (e.g. sprite of kind and animation), but the ones that are on the blocks directly (set background image, set tilemap) get a bit of extra vertical padding; @shakao is that something we've seen before with new blockly bump?

shakao:
this blockly update overhauled their rendering system; it unified a lot of the code for blocks layout (padding, spacing before/after text and fields, etc) so i'm guessing some of this stuff was handled in individual blocks/field editors before. we're trying to defer to the blockly rendering whenever possible to reduce issues with future merges, so the extra padding on set tilemap/set background seems ok to me

jwunderl:
Sounds good, overall that group of code in the last comment ended up getting a bit shorter in total (410px w/ this change vs 417px in live vs 446px in beta), I think it's a nice change overall ~

shakao:
oh yeah the image blocks fix is good, that's a custom field so we own all the rendering for it anyway

jwunderl:
@abchatra for release ~

PR #7329, opened 2020-07-16
Only add parenthesis to method and function snippets

shakao:
Fixes https://github.com/microsoft/pxt-arcade/issues/2177

Fixes completions for properties, also adds back in the "(method)", "(property)" preview text in the autocomplete list--easy to remove if it seems cluttered but it felt helpful to me.

fix-completions


darzu:
The icons are supposed to be informative about whether they are properties or methods but I think we're maybe not setting the icon right

PR #7332, opened 2020-07-17
Fix emit order

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/2190

I'm not sure how this got out of sync, but for some reason the emit order sometimes changes on the program. Make sure main.ts is always last.


mmoskal:
Also fixes https://github.com/microsoft/pxt-arcade/issues/2186

PR #7339, opened 2020-07-20
Fix blocks upgrades with tilemaps

jwunderl:
re: https://github.com/microsoft/pxt-arcade/pull/2197, this fixes blocks upgrades when there's a tilemap.

The issue was that the blocks compile was generating `main.ts` as if `tilemap.g.ts` exists, so it didn't have the namespace at the top. This adds on to the blocks patch logic to copy over the generated files to the patch when we're validating that it creates an okay program. 

This feels like potentially a bit too much tilemap specific logic in the upgrade logic, but not sure where a better fix would be.

### DIFF
--- a/webapp/src/compiler.ts
+++ b/webapp/src/compiler.ts
@@ -718,11 +718,20 @@ function upgradeFromBlocksAsync(): Promise<UpgradeResult> {
 
             const xml = Blockly.Xml.textToDom(text);
             pxt.blocks.domToWorkspaceNoEvents(xml, ws);
-            patchedFiles["main.blocks"] = text;
+            pxtblockly.upgradeTilemapsInWorkspace(ws, pxt.react.getTilemapProject());
+            const upgradedXml = Blockly.Xml.workspaceToDom(ws);
+            patchedFiles["main.blocks"] = Blockly.Xml.domToText(upgradedXml);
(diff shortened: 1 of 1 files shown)


webapp/src/compiler.ts
```diff
@@ -718,11 +718,24 @@ function upgradeFromBlocksAsync(): Promise<UpgradeResult> {
...
+
             return pxt.blocks.compileAsync(ws, info)
         })
         .then(res => {
             patchedFiles["main.ts"] = res.source;
+            return project.buildTilemapsAsync();
+        })
+        .then(() => {
+            const compiledFiles = project.getAllFiles();
+            const generatedFileNames = Object.keys(compiledFiles).filter(el => /\.g\.ts/.test(el));
```
riknoll, line 733: why not just copy over all non-main.ts files?

riknoll:
I think the way to make this less tilemappy would be to refactor so that there is only one codepath that compiles projects, but it's outside the scope of this change.

webapp/src/compiler.ts
jwunderl: No particular reason, these were just the ones that were breaking things.

I'll switch it so it hopefully avoids the issue in the future~

PR #7341, opened 2020-07-20
Asset editor sends message when user clicks done

kimprice:
The asset editor sends a message with the updated jresURL when the user clicks done. This is used by the vs code extension to update the underlying jres document of the asset.


webapp/src/assetEditor.tsx
```diff
@@ -25,10 +25,18 @@ export interface MessageData {
...
-        if (msg.data._fromVscode && msg.data.type === "update") {
-            this.editor.loadJres(msg.data.message);
+        if (msg.data._fromVscode) {
+            if (msg.data.type === "initialize") {
+                this.editor.loadJres(msg.data.message);
+                this.currentJres = msg.data.message;
+            } else if (msg.data.type === "update") {
+                this.updateAndSendJres();
+            } else if (msg.data.type === 'updateWebview') {
+                this.editor.loadJres(this.currentJres);
```
shakao, line 38: when do you send the updateWebview message? i'm not sure i see why you need to keep a separate "currentJres" variable

webapp/src/components/ImageFieldEditor.tsx
```diff
@@ -6,6 +6,7 @@ import { setTelemetryFunction } from './ImageEditor/store/imageReducer';
 
 export interface ImageFieldEditorProps {
     singleFrame: boolean;
+    parentCallback?: () => void;
```
shakao, line 9: use a more descriptive name--"doneButtonCallback" or "onDoneButtonClicked", for example

webapp/src/assetEditor.tsx
```diff
@@ -49,8 +57,22 @@ export class AssetEditor extends React.Component {
...
+    callbackOnDoneClick = () => {
+        this.updateAndSendJres();
+    }
+
+    updateAndSendJres() {
+        this.currentJres = this.editor.getJres();
+        const updateMsg: MessageData = {
+            type: "update",
+            message: this.currentJres,
+            _fromVscode: false,
```
shakao, line 69: you can delete this line, as it's an optional parameter

webapp/src/assetEditor.tsx
kimprice: I added this for the case when the user has the file open in two or more places in vscode like when they use the split editor. When they edit one and click done, the other open versions need to have the webview updated, but I'll remove this for now because I didn't finish implementing it on the vscode side and now that I'll thinking about it again, I think I should be able to take care of the issue using just the vscode side.

PR #7343, opened 2020-07-20
Run url replacements on asseteditor.html

shakao:


shakao:
@abchatra can we also take this one? it's for the /--asseteditor page

abchatra:
Sure. This is CLI only change though.

PR #7344, opened 2020-07-21
setBackgroundImage snippet improvements

eanders-MS:
Resolves https://github.com/microsoft/pxt-arcade/issues/2169

With this change, when the user types or drags `scene.setBackgroundImage(...`, it creates an image literal the same dimensions as the screen.

**Rough edge**: The resulting snippet is 120-ish lines long, _and isn't automatically collapsed after being inserted_. Which is kind of bad. One fix would be to make snippets optionally foldable, perhaps parameterized from a block comment. That might be a lot of additional code, and seems separate from this change.

Tested with JS and PY.

### DIFF
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -1857,6 +1857,7 @@ namespace ts.pxtc.service {
         // TODO: a lot of this is duplicate logic with blocklyloader.ts:buildBlockFromDef; we should
         //  unify these approaches
         const PY_INDENT: string = (pxt as any).py.INDENT;
+        const fileType = python ? "python" : "typescript";
 
         let preStmt: SnippetNode[] = [];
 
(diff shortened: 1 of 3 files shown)


pxtcompiler/emitter/service.ts
```diff
@@ -1970,6 +1971,83 @@ namespace ts.pxtc.service {
...
+                    const res = parseInt(raw);
+                    if (isNaN(res)) {
+                        return def;
+                    }
+                    return res;
+                }
+            }
+
+            function getDefaultValueFromFieldEditor(paramName: string): SnippetNode | null {
+                const blockParam = attrs._def?.parameters?.find((p) => p.name === name);
```
riknoll, line 2028: Don't reference _def directly, in pxtlib we have a function called `pxt.blocks.compileInfo` that parses it

eanders-MS:
Here's a visual:

pxtcompiler/emitter/service.ts
```diff
@@ -1970,6 +1971,83 @@ namespace ts.pxtc.service {
...
+                if (!blockParam?.shadowBlockId)
+                    return null
+                let sym = blocksById[blockParam.shadowBlockId]
+                if (!sym)
+                    return null
+                if (sym.attributes?.blockId !== blockParam.shadowBlockId)
```
riknoll, line 2034: I don't think this check is necessary, should always be true

eanders-MS: I was wondering about that!

eanders-MS:
@darzu, @abchatra curious what you think of the user experience here. Not sure this is better than existing until the code can be folded.

PR #7345, opened 2020-07-21
github integration: include version info in pxt.json and use it on import

jwunderl:
re: https://github.com/microsoft/pxt-arcade/pull/2197 not upgrading scripts imported from github

The issue is that we don't store any editor version info in the github repo, and instead just make it apply the current version, so none of the upgrade rules were applying when importing a github script. (Previously loaded ones are fine)

Unfortunately, I don't think there's actually a way to fix this one for the current release, as the information just isn't there; we can't safely pretend like the script is from `0.0.0` and run all the upgrade rules because they don't necessarily work when reapplied (e.g. SpriteKind upgrade rule would break new scripts, etc). This change should make it work for the future as it makes the version number available in pxt.json

if we want to hack in a fix for any particular repo, easy enough to patch the pxt.json and add in the targetVersions field, and this change would then pick that up / apply upgrade rules

as a side fix, also changes the upgrade logic to apply upgrade rules when any part of the version is different; it was bailing out early on patch changes, which could result in patches being missed / we don't consistently update minor version along with minor breaking changes that need patches.

### DIFF
--- a/pxtlib/diff.ts
+++ b/pxtlib/diff.ts
@@ -533,6 +533,7 @@ namespace pxt.diff {
                         break;
                     case "languageRestriction":
                     case "preferredEditor":
+                    case "targetVersion":
                         r[key] = vA; // keep current one
                         break;
                     case "public":
--- a/webapp/src/compiler.ts
+++ b/webapp/src/compiler.ts
@@ -657,7 +657,7 @@ export function applyUpgradesAsync(): Promise<UpgradeResult> {
     const pkgVersion = pxt.semver.parse(epkg.header.targetVersion || "0.0.0");
     const trgVersion = pxt.semver.parse(pxt.appTarget.versions.target);
 
-    if (pxt.semver.cmp(pkgVersion, trgVersion) !== 0) {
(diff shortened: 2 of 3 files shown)


jwunderl:
Hm, is that a thing that is done? Either way I made it track the target ID too and only upgrade if from same target. Will find the 'merge pxt.json' logic again and make sure it handles changing IDs like it does changing versions

And the field is already in the schema [here](https://github.com/microsoft/pxt/blob/master/localtypings/pxtpackage.d.ts#L44) so should be good.

pelikhan:
A library can be used in multiple targets - this change does not seem to handle that.

jwunderl:
This change applies when importing the github repo as a project (through import dialog -> your github repo), not as a dependency. Right now it doesn't keep any version info from the target, so it never applies upgrade rules for those projects.

I can make it keep track of the last version opened in each target separately, but then we could get duplicate updates if something in pxt or common packages changes that applies to each target; the intention of this change is that if you have a working github project in one version of the editor, you can at least come to the newer version of that same editor and get a working version automatically from our upgrade rules.

PR #7346, opened 2020-07-21
fix runtime error toast

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/2211

Add back in toast when we get a runtime error. This got dropped when adding support for runtime errors to error list ([here](https://github.com/microsoft/pxt/pull/7048/files#diff-325ec83a257a791d33bd4484fd67fb24R784)); this adds it in as the default behavior for the method that gets overwritten to propagate to the error list, so that it shows the error as a toast when the editor doesn't have error list support.


abchatra:
I thought this behavior was intentional. Error are propagated to error panel.

PR #7350, opened 2020-07-23
Remove 'Blocks' text for forumUrl card, use pencil icon

shakao:
image

Match previous behavior of "open in editor" text without specifying language. Since forum games are just a share URL, we don't know what language they're in, and this saves us the effort of having to manually check every game.

(Still eventually want to move the extra "open in editor" card to the otherActions field but that change touches more places and requires updating the discourse download script to output the updated format)

Fixes https://github.com/microsoft/pxt-arcade/issues/2229

Also a build: https://arcade.makecode.com/app/52c89c868dd520e607c5abb60364871a26703ea0-6e76a989ff

### DIFF
--- a/webapp/src/projects.tsx
+++ b/webapp/src/projects.tsx
@@ -821,12 +821,15 @@ export class ProjectsDetail extends data.Component<ProjectsDetailProps, Projects
             case "forumUrl":
                 icon = "comments"
                 break;
+            case "forumExample":
+                icon = "pencil"
+                break;
             case "template":
             default:
                 if (youTubeId || youTubePlaylistId) icon = "youtube";
                 break;
         }
-        return this.isLink() && type != "example" // TODO (shakao)  migrate forumurl to otherAction json in md
+        return this.isLink() && type != "forumExample" // TODO (shakao)  migrate forumurl to otherAction json in md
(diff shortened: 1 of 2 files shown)


pelikhan:
These icons are barely visible to my eyes.

webapp/src/projects.tsx
```diff
@@ -821,12 +821,15 @@ export class ProjectsDetail extends data.Component<ProjectsDetailProps, Projects
...
+            case "forumExample":
+                icon = "pencil"
+                break;
             case "template":
             default:
                 if (youTubeId || youTubePlaylistId) icon = "youtube";
                 break;
         }
-        return this.isLink() && type != "example" // TODO (shakao)  migrate forumurl to otherAction json in md
+        return this.isLink() && type != "forumExample" // TODO (shakao)  migrate forumurl to otherAction json in md
```
pelikhan, line 832: what about existing "example" cards?

shakao: no change in those--they still have the blocks/text/js. ultimately we can just merge forumExample and sharedExample (https://github.com/microsoft/pxt/pull/7352) but it's messy to maintain compatibility between beta/live so i'm leaving it for now

PR #7351, opened 2020-07-23
performance improvements for intellisense

darzu:
Fixes: https://github.com/microsoft/pxt-arcade/issues/2239

Perf improvements:
1. Global symbols pre-filter
When we don't have a specific  set of symbols to return like for member completion, we would return all PXT APIs and allow Monaco to fuzzy filter and sort based on the users input text. However this means we're generating snippets and doing other processing for potentially hundreds or thousands of APIs. So instead we now filter these global symbols to only those who partially match what the user has already entered.
2. Add snippets late
When returning completion results, wait to generate and add completion snippets until all other filtering is done
3. Enforce a total max number of symbols

TODO:
- [x] validate perf improvements are sufficient
- [x] test to make sure quality of intellisense hasn't regressed
- [x] fix language service tests

### DIFF
--- a/docfiles/pxtweb/cookieCompliance.ts
+++ b/docfiles/pxtweb/cookieCompliance.ts
@@ -138,16 +138,19 @@ namespace pxt {
                 performance.clearMeasures(`${name} elapsed`)
             }
         }
-        export function report() {
+        export function report(filter: string = null) {
             if (enabled) {
                 let report = `performance report:\n`
                 for (let [msg, time] of stats.milestones) {
-                    let pretty = prettyStr(time)
-                    report += `\t\t${msg} @ ${pretty}\n`
+                    if (!filter || msg.indexOf(filter) >= 0) {
+                        let pretty = prettyStr(time)
+                        report += `\t\t${msg} @ ${pretty}\n`
+                    }
                 }
                 report += `\n`
                 for (let [msg, start, duration] of stats.durations) {
-                    if (duration > 50) {
+                    let filterIncl = filter && msg.indexOf(filter) >= 0
+                    if ((duration > 50 && !filter) || filterIncl) {
                         let pretty = prettyStr(duration)
                         report += `\t\t${msg} took ~ ${pretty}`
                         if (duration > 1000) {
@@ -156,7 +159,7 @@ namespace pxt {
                         report += `\n`
                     }
                 }
-                console.log(report)
+                console.log(report || "nothing to report")
             }
             perfReportLogged = true
         }
--- a/docfiles/pxtweb/cookieCompliance.ts
+++ b/docfiles/pxtweb/cookieCompliance.ts
@@ -159,7 +159,7 @@ namespace pxt {
                         report += `\n`
                     }
                 }
-                console.log(report || "nothing to report")
+                console.log(report)
             }
             perfReportLogged = true
         }
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -1178,7 +1178,7 @@ namespace ts.pxtc.service {
                             paramIdx = callSym.parameters.length - 1
                         const paramType = getParameterTsType(callSym, paramIdx, blocksInfo)
                         if (paramType) {
-                            /// weight the results higher if they return the correct type for the parameter
+                            // weight the results higher if they return the correct type for the parameter
                             const matchingApis = getApisForTsType(paramType, call, tc, resultSymbols);
                             matchingApis.forEach(match => match.weight = COMPLETION_MATCHING_PARAM_TYPE_WEIGHT);
                         }
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -1032,6 +1032,11 @@ namespace ts.pxtc.service {
 
             const isMemberCompletion = dotIdx !== -1
 
+            const partialWord = isMemberCompletion ? src.slice(dotIdx + 1, wordEndPos) : src.slice(wordStartPos, wordEndPos)
+
+            const MAX_SYMBOLS_BEFORE_FILTER = 50
+            const MAX_SYMBOLS = 200
+
             if (isMemberCompletion)
                 complPosition = dotIdx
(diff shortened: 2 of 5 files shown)


darzu:
Using MacBook Pro 15" 2018 i9 w/ 6x CPU throttling in Chrome

Before (JS)

Before (PY)


After (JS)


After (PY)

pelikhan:
There should be a tick to measure these perf metric and a dashboard entry to visualize it.

docfiles/pxtweb/cookieCompliance.ts
```diff
@@ -156,7 +159,7 @@ namespace pxt {
                         report += `\n`
                     }
                 }
-                console.log(report)
+                console.log(report || "nothing to report")
```
eanders-MS, line 162: >report [](start = 28, length = 6)

minor: `report` will never be null.

webapp/src/monaco.tsx
```diff
@@ -69,8 +69,10 @@ class CompletionProvider implements monaco.languages.CompletionItemProvider {
         const wordEndOffset = model.getOffsetAt({ lineNumber: position.lineNumber, column: word.endColumn })
 
 
+        pxt.perf.measureStart("getCompletions")
```
eanders-MS, line 72: >pxt.perf.measureStart("getCompletions") [](start = 8, length = 39)

I think you can already get this measurement by setting `pxt.appTarget.compile.switches.time`

pxtcompiler/emitter/service.ts
```diff
@@ -1088,6 +1093,13 @@ namespace ts.pxtc.service {
                         tsPos = smallest.ts.startPos
                     }
                 }
+
+                // filter based on word match if we get too many (ideally we'd leave this filtering for monaco as it's
+                // better at fuzzy matching and fluidly changing but for performance reasons we want to do it here)
+                if (!isMemberCompletion && resultSymbols.length > MAX_SYMBOLS_BEFORE_FILTER) {
```
eanders-MS, line 1099: I think this will help. It looks like the issue is primarily in Monaco. Profiling it, all the time is going to rendering the suggestion elements.

darzu:
> There should be a tick to measure these perf metric and a dashboard entry to visualize it.

Any example? Not sure how to add or track a number like this

webapp/src/monaco.tsx
darzu: Hmm, that's not giving me anything?
image

darzu: Ah I get it now:
image

Thanks! didn't know this

darzu:
You can test new perf here: https://arcade.makecode.com/app/6d4a9c5b0e7fa39bd448d0a14c4f15b91bc5e972-d7e6fc7eda

pxtcompiler/emitter/service.ts
```diff
@@ -1088,6 +1093,13 @@ namespace ts.pxtc.service {
...
+                    resultSymbols = resultSymbols
+                        .filter(s => (isPython ? s.symbol.pyQName : s.symbol.qName).indexOf(partialWord) >= 0)
```
jwunderl, line 1101: question; my understanding was that the result list always had to filter symbols down / it typically doesn't add back in symbols (I think for example this was why we don't show completions for things starting with `_`, as it was non trivial to 'add back in' those symbols). If that's the case, it seems like this could lose fidelity of completions / not accurately gather completions later on, if we're semi-arbitrarily filtering early. Is my understanding of this incorrect / am I missing context here?

(as a sidenote for anyone who doesn't have experience with that, we filter out symbols that start with `-` as they're seen as 'internal data' the user won't typically want to see, like sprites `._x` field / etc)

eanders-MS:
Profiling in Chrome, I see a perf hot spot in Monaco's rendering of suggestion elements. Monaco is setting `innerHTML` on each visible suggestion, each taking about 20ms. That isn't much by itself, but it adds up when the list is long.

Does not repro for me in Firefox.

webapp/src/monaco.tsx
eanders-MS: I found it while tracing through the code. I suspect it might be a config setting read from a file.

---
In reply to: [459766521](https://github.com/microsoft/pxt/pull/7351#discussion_r459766521) [](ancestors = 459766521)

pxtcompiler/emitter/service.ts
darzu: Yes, for the most part we filter symbols down over time, but we also do sometimes add new symbols (like for member completion, local vars, or keywords) back in, so it should be that it's only global APIs we're filtering this way.

We definitely lose some fidelity compared to before like "gm" won't show "game.*" members any more, but I think the trade off for performance is worth it.

darzu:
> Profiling in Chrome, I see a perf hot spot in Monaco's rendering of suggestion elements. Monaco is setting `innerHTML` on each visible suggestion, each taking about 20ms. That isn't much by itself, but it adds up when the list is long.
> 
> Does not repro for me in Firefox.

Doesn't repro meaning firefox is faster at rendering intellisense?

darzu:
I lowered the max symbol limit again to 100. I didn't notice a drop in quality and based on what you said of the rendering this should help significantly.

darzu:
@eanders-MS , I'm happy to merge whenever you feel it's ready

eanders-MS:
LGTM!

darzu:
Another test target: https://arcade.makecode.com/app/603a2215f53787304aaf63c35bf919607dbd986a-2cf7c30eb4

Just waiting on the builds.

eanders-MS:
Some perf comparisons:

### Chrome
| Beta | Test Target |
|---|---|
| 781ms | 587ms |

### Firefox
| Beta | Test Target |
|---|---|
| 1921ms | 1244ms |

Test case: Starting from an empty main.ts, type 's' and wait. Timing `getCompletions` call, using `pxt.appTarget.compile.switches.time=true`.

pelikhan:
We should systematically log these in ticks to analyze them in the wild. We can also do that for 5% of sessions to minimize impact.

PR #7359, opened 2020-07-31
handle changes in github apis

pelikhan:
The API to enable pages in github changed slightly breaking the github integration. This fixes it.

https://developer.github.com/v3/repos/pages/#enable-a-pages-site

* catch and report any failure of the api, ignore result


abchatra:
Does this needs to be ported to stable channel for microbit & arcade?

pelikhan:
Yes

pelikhan:
We should have a script to set those up, it’s a pain.

PR #7361, opened 2020-07-31
github pages fix [stable6.0]

pelikhan:
Port of https://github.com/microsoft/pxt/pull/7359
Tested on localhost.


abchatra:
Needs to port to stable6.1 as well.

PR #7063, opened 2020-05-21
invalidation of release cache

pelikhan:
Invalidate the cloud release cachen when the client creates a new release. This is critical to improve the authoring flow for extensions/tutorials.
- [x] add cache bustom query argument
- [x] add nocache for pxt cloud
- [x] rename a few function "withCdn" to make it clearer that those are 24hour caches.

PR #7363, opened 2020-07-31
[stable5.33] Port versioned tutorials fix from beta to live

jwunderl:
this commit is the change: https://github.com/microsoft/pxt/commits?author=jwunderl, but I had to port over the gulp watch change for pxt-minecraft to build on my machine: https://github.com/microsoft/pxt/pull/6544 (it kept getting random ts ref errors when running gulp as part of pxt serve). I can drop that if we want to minimize, but I don't think that change has any effect outside of our local builds

partial port of https://github.com/microsoft/pxt/pull/7063, https://github.com/microsoft/pxt/pull/7029 and some other PR that added a `pxt.BrowserUtils.cacheBustingUrl`, and cherry picks finally https://github.com/microsoft/pxt/pull/7359

Also, note the comment starting with [DIFF](https://github.com/microsoft/pxt/pull/7363/files#diff-325ec83a257a791d33bd4484fd67fb24R3180); the behavior is slightly different from master, but I wanted to minimize the behavioral difference from live site, which would mean not breaking repos that do not have a release. If we're not worried about that can change back to normal.

Started as a draft because I wanted to test for a few more minutes after the live stream / run through making a test tutorial and doing a few updates.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -3174,14 +3174,14 @@ export class ProjectView
                     return (ghid.tag ? Promise.resolve(ghid.tag) : pxt.github.latestVersionAsync(ghid.fullName, config, true))
                         .then(tag => {
                             if (!tag) {
-                                // pxt.log(`tutorial github tag not found at ${ghid.fullName}`);
-                                // return undefined;
                                 /**
                                  * DIFF: in /beta and above, this path returns undefined,
                                  * but I'm just returning based off main branch to minimize
                                  * the chance of breaking existing tutorials with the patch /
                                  * give time to check existing tutorials for release.
                                  */
+                                pxt.log(`tutorial github tag not found at ${ghid.fullName}`);
+                                // return undefined;
                                 return pxt.github.downloadPackageAsync(ghid.fullName, config);
                             }
                             ghid.tag = tag;
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -3171,7 +3171,23 @@ export class ProjectView
                             reportId = "https://github.com/" + ghid.fullName;
                             break;
                     }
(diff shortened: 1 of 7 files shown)


webapp/src/app.tsx
```diff
@@ -3171,7 +3171,23 @@ export class ProjectView
...
+                            if (!tag) {
+                                // pxt.log(`tutorial github tag not found at ${ghid.fullName}`);
+                                // return undefined;
+                                /**
+                                 * DIFF: in /beta and above, this path returns undefined,
+                                 * but I'm just returning based off main branch to minimize
+                                 * the chance of breaking existing tutorials with the patch /
+                                 * give time to check existing tutorials for release.
+                                 */
+                                return pxt.github.downloadPackageAsync(ghid.fullName, config);
```
pelikhan, line 3185: emit warning in the console

webapp/src/workspace.ts
```diff
@@ -640,7 +640,10 @@ export async function commitAsync(hd: Header, options: CommitOptions = {}) {
         })
         if (options.createRelease) {
             await pxt.github.createReleaseAsync(parsed.fullName, options.createRelease, newCommit)
-            await pxt.github.enablePagesAsync(parsed.fullName); // ensure pages are on
+             // ensure pages are on
+            await pxt.github.enablePagesAsync(parsed.fullName);
```
pelikhan, line 644: you're going to want my fix about github pages as well

abchatra:
Thanks @jwunderl for trying this out. I do feel this is bit risky change and requires end to end testing. Can we hold off for some time and give beta as workaround for content developer?

jwunderl:
Ahh sounds good if they're good with using beta to test, if it seems risky then I'll close ~

PR #7366, opened 2020-08-06
Fix crash when opening packages with no readme

mmoskal:
repro:
* create new project
* switch to TS
* remove README.md
* go back to home
* re-open project
* hang!

Similarly, will hang when opening published projects without readme.

@abchatra this seem to warrant point-fix


abchatra:
Ported to stable6.1 for arcade.
stable6.0 for microbit doesn't have this issue.

PR #7371, opened 2020-08-07
Initial Tilemap setup in Asset Editor

kimprice:
Asset Editor uses query parameters to determine which editor (sprite or tilemap) to invoke. Both editors have the galleries hidden. This page is embeded in the MakeCode VS Code extension.

### DIFF
--- a/pxtlib/package.ts
+++ b/pxtlib/package.ts
@@ -1237,7 +1237,7 @@ namespace pxt {
         return {
             id,
             data: jres.data,
-            dataEncoding: jres.dataEncoding || base.dataEncoding || "base64",
+            dataEncoding,
             icon,
             namespace: ns,
             mimeType,
--- a/pxtlib/package.ts
+++ b/pxtlib/package.ts
@@ -1218,4 +1218,31 @@ namespace pxt {
     export function isPkgBeta(cfg: { description?: string; }): boolean {
         return cfg && /\bbeta\b/.test(cfg.description);
     }
+
+    export function parseJResFile(key: string, jres: JRes, base: JRes): JRes {
+        if (key == "*") return null;
+        if (typeof jres == "string") {
+            // short form
+            jres = { data: jres } as any
+        }
+        let ns = jres.namespace || base.namespace || ""
+        if (ns) ns += "."
+        let id = jres.id || ns + key
+        let icon = jres.icon
+        let mimeType = jres.mimeType || base.mimeType
+        let dataEncoding = jres.dataEncoding || base.dataEncoding || "base64"
+        if (!icon && dataEncoding == "base64" && (mimeType == "image/png" || mimeType == "image/jpeg")) {
+            icon = "data:" + mimeType + ";base64," + jres.data
+        }
+        return {
+            id,
+            data: jres.data,
+            dataEncoding: jres.dataEncoding || base.dataEncoding || "base64",
+            icon,
+            namespace: ns,
+            mimeType,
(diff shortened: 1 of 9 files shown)


pxtlib/package.ts
```diff
@@ -1213,4 +1190,31 @@ namespace pxt {
...
+        let icon = jres.icon
+        let mimeType = jres.mimeType || base.mimeType
+        let dataEncoding = jres.dataEncoding || base.dataEncoding || "base64"
+        if (!icon && dataEncoding == "base64" && (mimeType == "image/png" || mimeType == "image/jpeg")) {
+            icon = "data:" + mimeType + ";base64," + jres.data
+        }
+        return {
+            id,
+            data: jres.data,
+            dataEncoding: jres.dataEncoding || base.dataEncoding || "base64",
```
riknoll, line 1212: nitpick: reuse the dataEncoding variable here

webapp/src/assetEditor.tsx
```diff
@@ -14,39 +17,126 @@ function init() {
...
-    componentDidMount() {
-        window.addEventListener("message", this.handleMessage, null);
-        this.postMessage({type: "ready"});
+    initTilemap(s?: string) {
+        this.tilemapProject = new pxt.TilemapProject();
+        this.tilemapProject.loadJRes(s ? this.parseJRes(s) : {});
+        let project = this.tilemapProject.getTilemap(this.tilemapName);
+
+        if (!project) {
+            const [ name, map ] = this.tilemapProject.createNewTilemap(this.tilemapName, 16, 16, 16);
```
riknoll, line 92: we should think about ways to pass in a tile width

webapp/src/assetEditor.tsx
```diff
@@ -14,39 +17,126 @@ function init() {
...
+            project = map;
+            this.tilemapName = name;
+        }
+
+        this.editor.init(this.tilemapProject.getTilemap(this.tilemapName), this.callbackOnDoneClick);
```
riknoll, line 97: use project variable here

webapp/src/assetEditor.tsx
```diff
@@ -14,39 +17,126 @@ function init() {
...
+                this.tilemapProject.deleteTile(deleted);
+            }
+        }
+
+        if (data.editedTiles) {
+            for (const edit of data.editedTiles) {
+                const editedIndex = data.tileset.tiles.findIndex(t => t.id === edit);
+                const edited = data.tileset.tiles[editedIndex];
+
+                if (edited.id.startsWith("*")) continue;
```
riknoll, line 113: edited might be undefined

webapp/src/assetEditor.tsx
shakao: yeah probably some UI on the vscode side for tile width + optional initial tilemap width/height, for the extension. i moved the tile width out so it can be set with that first 'initialize' message to the iframe

webapp/src/assetEditor.tsx
```diff
@@ -58,15 +154,41 @@ export class AssetEditor extends React.Component {
...
-            type: "update",
-            message: this.editor.getJres(),
+        let message: string;
+        switch (this.state.viewType) {
+            case "sprite":
+                message = this.editor.getJres();
+                break;
+            case "tilemap":
+                this.updateTilemap();
+                message = JSON.stringify(this.tilemapProject.getProjectTilesetJRes());
```
riknoll, line 164: Probably also want to send the corresponding ts file

PR #7374, opened 2020-08-07
Suppress tilemap shortcut handling when editing a tile.

jacobcarpenter:
@riknoll @livcheerful 

Here's a fix for microsoft/pxt-arcade#2317. I hadn't realized that when editing an individual tile, there were _two_ `ImageCanvas`es being rendered! 😱 😅

I'm also **totally** happy for you to independently solve this one, if you'd prefer.

I haven't actually done much with Redux... perhaps it's unusual to directly pass props to components like I do in ImageEditor line 87.

### DIFF
--- a/webapp/src/components/ImageEditor/ImageCanvas.tsx
+++ b/webapp/src/components/ImageEditor/ImageCanvas.tsx
@@ -32,6 +32,8 @@ export interface ImageCanvasProps {
     tilemapState?: TilemapState;
     imageState?: pxt.sprite.ImageState;
     prevFrame?: pxt.sprite.ImageState;
+
+    suppressShortcuts: boolean;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


webapp/src/components/ImageEditor/ImageCanvas.tsx
```diff
@@ -32,6 +32,8 @@ export interface ImageCanvasProps {
     tilemapState?: TilemapState;
     imageState?: pxt.sprite.ImageState;
     prevFrame?: pxt.sprite.ImageState;
+
+    observeShortcuts: boolean;
```
jacobcarpenter, line 36: Whoops! Obviously I should invert this prop so that the default behavior is to observer. Fix incoming...

riknoll:
This is a good place to pass props from the parent, I don't think you could achieve this behavior with mapStateToProps()

PR #7378, opened 2020-08-07
Dont show intellisense in comments or string literals; intellisense refactor (pt 1)

darzu:
Fixes: https://github.com/microsoft/pxt-arcade/issues/2282

Also adds capability to test auto-complete snippet results in the languageservicetests

I'll point out in comments which code is new or changed. The rest has just been rearranged into new files for easier editing and reasoning.


pxtcompiler/emitter/languageservice.ts
```diff
@@ -0,0 +1,650 @@
...
+        lastNl = Math.max(0, lastNl)
+        const lineText = src.substring(lastNl + 1, position)
+
+        // are we on a line comment, if so don't show completions
+        // NOTE: multi-line comments and string literals are handled
+        //  later as they require parsing
+        const lineCommentStr = isPython ? "#" : "//"
+        if (lineText.trim().startsWith(lineCommentStr)) {
+            return r;
+        }
```
darzu, line 328: This is new. This aborts if we're on a comment line

pxtcompiler/emitter/languageservice.ts
```diff
@@ -0,0 +1,650 @@
...
+        if (inComment) {
+            return r;
+        }
+
+        // abort if we're in a string literal
+        const stringLiteralKinds = [SK.StringLiteral, SK.FirstTemplateToken, SK.NoSubstitutionTemplateLiteral];
+        const inLiteral = stringLiteralKinds.some(k => tsNode.kind === k)
+        if (inLiteral) {
+            return r;
+        }
```
darzu, line 439: `const commentMap [...]` until here is new. We abort if we're in a multi-line comment or string literal

pxtcompiler/emitter/languageservice.ts
```diff
@@ -0,0 +1,650 @@
...
+            }
+        }
+
+        return [...retApis, ...completionSymbols(enumVals, COMPLETION_DEFAULT_WEIGHT)]
+    }
+
+
+    export function getBasicKindDefault(kind: SyntaxKind, isPython: boolean): string | undefined {
+        switch (kind) {
+            case SK.StringKeyword: return "\"\"";
```
darzu, line 83: This used to be:
```
case SK.StringKeyword: return "";
```
which caused no default value to be emitted for string parameters. This was found by the new test suite.

tests/language-service/languageservicerunner.ts
```diff
@@ -1,6 +1,5 @@
 /// <reference path="../../built/pxtcompiler.d.ts"/>
```
darzu, line 2: All changes in this file are new. I added support for testing snippets (toolbox or auto-complete).

pxtcompiler/emitter/languageservice.ts
jwunderl: Checking my understanding here; this won't bail out on the following,

```typescript
console.log("got here") // I got here!!
```

but that would get handled in the commentMap section below, it just doesn't bail out early.

pxtcompiler/emitter/languageservice.ts
jwunderl: Is there a perf hit here? I suppose the type checking / etc that will be running will take significantly longer than a linear pass over the file so it should be fine.

pxtcompiler/emitter/languageservice.ts
darzu: Correct, the above is just a quick abort for perf reasons. It could actually be made more general to capture end-of-line comments as well.

pxtcompiler/emitter/languageservice.ts
darzu: Yeah, it really should be incremental. As should most of the language service.

I'll measure the perf hit.
- [ ] TODO

darzu: <10ms for PlanetPuttPutt, so barely measurable

PR #7379, opened 2020-08-07
fix error list colors in inverted-editor

jwunderl:
fix https://github.com/microsoft/pxt-minecraft/issues/1918, but there's a minor diff in styling for other editors I'll add a note on below; can maintain the same behavior but it'd require a bit of duplication / an extra less var defined for each editor

image

w/ highlight on error:

image

### DIFF
--- a/theme/errorList.less
+++ b/theme/errorList.less
@@ -55,21 +55,21 @@
         overflow-y: auto;
 
         .exceptionMessage {
-            color: black;
+            color: @textColor;
             padding: 0.5em 1em;
             width: 70%;
         }
 
         // errors
         .ui.selection.list {
-            color: black;
+            color: @textColor;
             cursor: pointer;
             margin: 0;
 
             .item {
                 border-radius: unset;
                 padding: 0.5em 1em;
-                color: black;
+                color: @textColor;
 
                 &:hover, &:focus {
                     background-color: @errorListMessageHoverColor !important;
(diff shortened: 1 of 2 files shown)


theme/themes/pxt/globals/site.variables
```diff
@@ -358,6 +358,8 @@
 @errorListInnerHeight: (@errorListHeight - @errorListHeaderHeight);
 @errorListMessageHoverColor: rgba(0, 0, 0, 0.1);
 @errorListBorderColor: #e5e5e5;
-@errorListOdd: @homeScreenBackground;
```
jwunderl, line 361: This was assuming homeScreenBackground === monaco editor background, but that's only true if the editor specifically sets them the same; monaco colors are set like this https://github.com/microsoft/pxt-microbit/blob/master/pxtarget.json#L509 and they've been set to match for arcade & microbit but not minecraft (which has a light homescreen / dark editor).

As written this changes to just lighten / darken but it goes towards gray more than the current behavior with lesses darken; current arcade:

image

this change: 

image

I personally prefer the grayer one as it matches with the highlight color, but can add a variable to maintain hue like it currently does if preferred ~

theme/errorList.less
```diff
@@ -117,4 +113,35 @@
...
+        color: @errorListInvertedTextColor;
+    }
+
+    .errorListInner {
+        .exceptionMessage {
+            color: @errorListInvertedTextColor;
+        }
+        .ui.selection.list {
+            color: @errorListInvertedTextColor;
+            cursor: pointer;
```
shakao, line 129: do we need this in the non-inverted theme as well?

shakao: same question for the border-radius/padding styles below

jwunderl: Oops! I forgot to commit removing these, copy / paste error when grabbing the selectors.

PR #7382, opened 2020-08-11
Pass tutorial snippets as array, isolate python snippets in modules

shakao:
Main changes: instead of concatenating the code when we parse the tutorial, save an array of snippets. When decompiling to get the used blocks, TS behavior is unchanged (concat snippets, decompile). For Python, each snippet is put in a separate module (file) to convert to TS, then we strip out the "namespace" declarations generated from the modules, concatenate the snippets and decompile.

Long-term, it could be useful to precompute or cache the used blocks; this data only really changes when the tutorial content changes so we probably don't need to generate it every time a tutorial is loaded.

Fixes https://github.com/microsoft/pxt-minecraft/issues/1926

### DIFF
--- a/webapp/src/compiler.ts
+++ b/webapp/src/compiler.ts
@@ -314,6 +314,39 @@ export function pySnippetToBlocksAsync(code: string, blockInfo?: ts.pxtc.BlocksI
         });
 }
 
+// Py[] -> blocks
+export function pySnippetArrayToBlocksAsync(code: string[], blockInfo?: ts.pxtc.BlocksInfo): Promise<pxtc.CompileResult> {
+    const namespaceRegex = /^\s*namespace\s+[^\s]+\s*{([\S\s]*)}\s*$/im;
+    const snippetBlocks = "main.blocks";
+    let trg = pkg.mainPkg.getTargetOptions()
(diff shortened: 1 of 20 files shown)


webapp/src/compiler.ts
```diff
@@ -314,6 +314,39 @@ export function pySnippetToBlocksAsync(code: string, blockInfo?: ts.pxtc.BlocksI
...
+            if (opts.sourceFiles.indexOf(snippetBlocks) === -1) {
+                opts.sourceFiles.push(snippetBlocks);
+            }
+            return workerOpAsync("py2ts", { options: opts });
+        })
+        .then(res => {
+            // strip the namespace declaration out of the converted snippets and concat to convert to blocks
+            let ts = "";
+            for (let file of files) {
+                let match = res.outfiles[file + ".ts"].match(namespaceRegex);
```
jwunderl, line 343: To confirm, this will work for arcade games with custom spritekinds (e.g. free throw) as it would just unwrap the outermost namespace, right? Looks like it should be fine

shakao: yep! it matches the start/end of file and the outermost namespace is generated as part of the py2ts conversion

PR #7383, opened 2020-08-11
Allow for expressions in pxt.json:fileDependencies

mmoskal:
still needs testing


pelikhan:
is this pr dead? @mmoskal

PR #7384, opened 2020-08-11
Allow duplicate comment to be translated

jwunderl:
fix https://github.com/microsoft/pxt-minecraft/issues/1911

I guess minor question, is there value in it specifying duplicate *comment*? Could just use the same duplicate as we use for DUPLICATE_BLOCK if we want, but maybe specifying it is a comment is valuable for students.


riknoll:
it's part of Blockly, so no reason to diverge I think

PR #7387, opened 2020-08-12
Adding bobbing animation to gutter icons on selected line

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/2265

Implementing the quick-fix proposed by @soniakandah in https://github.com/microsoft/pxt-arcade/issues/2265

The gutter icon bobs on the selected line:

### DIFF
--- a/webapp/src/monacoFieldEditorHost.ts
+++ b/webapp/src/monacoFieldEditorHost.ts
@@ -299,7 +299,7 @@ export class FieldEditorManager implements monaco.languages.FoldingRangeProvider
     }
 
     protected reapplyDecorations(owner: string, highlightLine = -1) {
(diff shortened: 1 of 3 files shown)


webapp/src/monacoFieldEditorHost.ts
```diff
@@ -279,6 +298,33 @@ export class FieldEditorManager implements monaco.languages.FoldingRangeProvider
         return editorRanges.concat(indentRanges);
     }
 
+    protected reapplyDecorations(owner: string, highlightLine = -1) {
+        const fe = this.fieldEditors.filter(f => f.id === owner)[0];
```
jwunderl, line 302:
```suggestion
        const fe = this.fieldEditors.find(f => f.id === owner);
```

PR #7394, opened 2020-08-14
Add tick event for successful github commit

livcheerful:
Do we want to add events so we can complete the quiz? We have a tick for when users click the commit button, but nothing to see if the commit was successful.

### DIFF
--- a/webapp/src/gitjson.tsx
+++ b/webapp/src/gitjson.tsx
@@ -559,16 +559,20 @@ class GithubComponent extends data.Component<GithubProps, GithubState> {
     }
 
     async commitAsync() {
+        let success = true;
         this.setState({ needsCommitMessage: false });
         this.showLoading("github.commit", true, lf("commit and push changes to GitHub..."));
         try {
             await this.commitCoreAsync()
             await this.maybeReloadAsync()
         } catch (e) {
+            success = false;
             pxt.tickEvent("github.commit.fail");
             this.handleGithubError(e);
         } finally {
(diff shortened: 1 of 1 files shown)


webapp/src/gitjson.tsx
```diff
@@ -568,6 +568,7 @@ class GithubComponent extends data.Component<GithubProps, GithubState> {
             this.handleGithubError(e);
         } finally {
             this.hideLoading()
+            pxt.tickEvent("github.commit.success");
```
abchatra, line 571: Can you add a failure path too? Like github.commit.failure?finally will execute even though there is an exception.

livcheerful: oh yes totally! I added a failure message instead of this success one, since we already have a tick event for when the commit button is pressed. I can add this back though if you think it'd be better to have the two events closer

abchatra: Good thumb of rule is, you don't want to join the two independent events to figure out the information. So just add the success event as well here. So you have to just query one thing.

livcheerful: I made the change, but I don't fully understand - won't that be two events anyways?

abchatra: Sorry, I was not clear. I pushed the commit to check the success and only emit then.

livcheerful: oh yeah that totally makes sense! thanks :D

PR #7395, opened 2020-08-14
Cache tutorial usedBlocks in IndexedDb

shakao:
- Database class based off the api info and translations DBs
- Hashes the code snippets and invalidates the blocks based on those--I think this should functionally sync up with invalidating when the cached markdown invalidates/changes
- Next steps are to get the list of used blocks at build time (like the api info) to also speed up the first load

Build: https://minecraft.makecode.com/app/29d4f5abb6b4e1d67b538a9ed500175b4b1eaa18-87b91c9b85

You can compare with beta or just compare first load vs subsequent loads

### DIFF
--- a/pxtlib/browserutils.ts
+++ b/pxtlib/browserutils.ts
@@ -946,12 +946,12 @@ namespace pxt.BrowserUtils {
     interface TutorialInfoIndexedDbEntry {
         id: string;
         hash: string;
-        blocks: { [ key: string ]: number };
+        blocks: Map<number>;
     }
 
     export interface ITutorialInfoDb {
         getAsync(filename: string, code: string[], branch?: string): Promise<TutorialInfoIndexedDbEntry>;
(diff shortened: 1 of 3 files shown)


pxtlib/browserutils.ts
```diff
@@ -932,6 +932,99 @@ namespace pxt.BrowserUtils {
...
+    }
+
+    function getTutorialInfoKey(filename: string, branch?: string) {
+        return `${filename}|${branch || "master"}`;
+    }
+
+    interface TutorialInfoIndexedDbEntry {
+        id: string;
+        hash: string;
+        blocks: { [ key: string ]: number };
```
riknoll, line 949: we usually use `Map<number>` (or `pxt.Map`)

riknoll:
maybe we should start this fetching/caching when they click on the card in the homescreen (to expand the info/language buttons). would improve the perceived load time at least.

abchatra:
@shakao Should we clear the cache when the hash changes? Otherwise it might just grow. Corner case may be worth the thought?

PR #7396, opened 2020-08-17
support for jres snippets in tutorials

riknoll:
Part of a fix for https://github.com/microsoft/pxt-arcade/issues/2250

You can add tilemaps and custom tiles to a tutorial by using a `jres` snippet. Just copy/paste the contents of `tilemap.g.jres` (no need to copy `tilemap.g.ts`). The tiles will also show up in the user project if they are custom.

### DIFF
--- a/pxtblocks/fields/field_tilemap.ts
+++ b/pxtblocks/fields/field_tilemap.ts
@@ -262,6 +262,11 @@ namespace pxtblockly {
         private parseBitmap(newText: string) {
             if (!this.blocksInfo) return;
 
+            if (newText) {
+                // backticks are escaped inside markdown content
+                newText = newText.replace(/&#96;/g, "`");
+            }
+
             const match = /^\s*tilemap\s*`([^`]*)`\s*$/.exec(newText);
 
             if (match) {
--- a/pxtlib/package.ts
+++ b/pxtlib/package.ts
@@ -205,35 +205,7 @@ namespace pxt {
         parseJRes(allres: Map<JRes> = {}) {
             for (const f of this.getFiles()) {
                 if (U.endsWith(f, ".jres")) {
-                    let js: Map<JRes> = JSON.parse(this.readFile(f))
-                    let base: JRes = js["*"] || {} as any
-                    for (let k of Object.keys(js)) {
-                        if (k == "*") continue
-                        let v = js[k]
-                        if (typeof v == "string") {
-                            // short form
-                            v = { data: v } as any
-                        }
-                        let ns = v.namespace || base.namespace || ""
-                        if (ns) ns += "."
-                        let id = v.id || ns + k
-                        let icon = v.icon
-                        let mimeType = v.mimeType || base.mimeType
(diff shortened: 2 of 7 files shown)


pxtlib/package.ts
```diff
@@ -1206,6 +1178,39 @@ namespace pxt {
...
+            let icon = v.icon
+            let mimeType = v.mimeType || base.mimeType
+            let dataEncoding = v.dataEncoding || base.dataEncoding || "base64"
+            if (!icon && dataEncoding == "base64" && (mimeType == "image/png" || mimeType == "image/jpeg")) {
+                icon = "data:" + mimeType + ";base64," + v.data
+            }
+            allres[id] = {
+                id,
+                data: v.data,
+                dataEncoding: v.dataEncoding || base.dataEncoding || "base64",
```
shakao, line 1202: small: reuse dataEncoding variable

pxtblocks/fields/field_tilemap.ts
```diff
@@ -262,6 +262,11 @@ namespace pxtblockly {
         private parseBitmap(newText: string) {
             if (!this.blocksInfo) return;
 
+            if (newText) {
+                // backticks are escaped inside markdown content
+                newText = newText.replace(/&#96;/g, "`");
```
shakao, line 267: can we do this in tutorial.js when the jres is parsed?

pxtlib/tilemap.ts
```diff
@@ -356,6 +376,54 @@ namespace pxt {
             for (const id of Object.keys(allJRes)) this.state.takenNames[id] = true;
         }
 
+        loadJres(jres: Map<JRes>, skipDuplicates = false) {
+            jres = inflateJRes(jres)
```
shakao, line 380: need to set `this.extensionTilesets = []` as in loadPackage ?

pxtlib/tilemap.ts
```diff
@@ -356,6 +376,54 @@ namespace pxt {
...
+                            id = tileMapping[id];
+                        }
+
+                        return this.resolveTile(id)
+                    })
+                }));
+
+            this.state.projectTilemaps.push(...maps);
+
+            for (const id of Object.keys(jres)) this.state.takenNames[id] = true;
```
shakao, line 423: is it possible for these tile ids to have changed from createNewTile?

pxtlib/tilemap.ts
riknoll: no, this adds to the existing project instead of overwriting it. The gallery won't get added twice

pxtlib/tilemap.ts
riknoll: good catch!

pxtblocks/fields/field_tilemap.ts
riknoll: actually, i'm not sure where the escape is happening. I tried moving this to tutorial.ts and marked.tsx. I think it might be the decompiler somehow? anyways, I think this is the best place for it.

riknoll:
@jwunderl forgot to do that, I'll do it in a separate PR

PR #7398, opened 2020-08-18
Offline app page updates

jwunderl:
align offline app page with docs styling, and clean up a bit of unused imports / rules / etc.

current state is https://arcade.makecode.com/offline-app ; if you want to see what it looks like when expanded need to open console and run:

```javascript
showAgree(); document.querySelector("#no-download-segment").remove();
```

~worth a note I left the generated css / html for the legal stuff  the same for now.~ last two commits update to current legal text; probably easiest to compare https://github.com/microsoft/pxt/pull/7398/files/a7f1279255aa091f83718926bd47975e7d63347e for the main primarily non-generated diffs

Doesn't currently change microbit as that has it's own page set / doesn't use these rules yet https://github.com/microsoft/pxt-microbit/blob/master/docs/offline-app.html (same with lego)

image

gif \\/

<details>
  <summary>Click to expand (7mb gif)</summary>
  
offline-app-page

</details>

### DIFF
--- a/docfiles/offline-app-body.html
+++ b/docfiles/offline-app-body.html
@@ -1,5 +1,5 @@
 
-    <div class="ui inverted vertical center aligned segment content">
+    <div id="offline-app-body" class="ui inverted vertical center aligned segment content">
         <div class="topbar @tocclass@">
             <div class="left flex-wrapper">
                 <img class="headerlogo" src="@cardLogo@" />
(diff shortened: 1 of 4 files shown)


jwunderl:
Okay, updated the legal stuff too @abchatra . There were a few things that needed manual fixing for the formatting to look nice so I put up another change in our docs to keep track of them. Also added an `offline-app-trademarks` docfile so we can include those on a target specific basis when needed (e.g. micro:bit, etc)

legal-mumbo-jumbo

PR #7406, opened 2020-08-19
Add ability to disable package help and specify custom blocks help

shakao:
Right now the right-click context help for blocks from an extension always opens the extension README in a new window. This adds two additional path formats:

- `//% help=none` to explicitly hide the help
- `//% help=github:extension-name/file-name` for custom help that will be loaded into the sidedocs

It also auto-expands the sidedocs when this help option is selected.

@abchatra I think this would be useful for Arcade as well; WeCodeMakeCode was asking how to it for her extension

Fixes https://github.com/microsoft/pxt-minecraft/issues/1881


webapp/src/blocks.tsx
```diff
@@ -801,9 +801,19 @@ export class Editor extends toolboxeditor.ToolboxEditor {
                                 return; // TODO support serving package docs in docs frame.
                             }
                         };
-                        if (/^\//.test(url))
+                        if (/^github:/.test(url)) {
+                            url = url.replace(/^github:\/?/, '') + ".md";
```
riknoll, line 805: what if they have .md in the path?

abchatra:
Awesome. I would document in the extension authoring.

PR #7413, opened 2020-08-22
Check block translations have equivalent parameters

jwunderl:
re: https://github.com/microsoft/pxt-microbit/issues/3351 and a ton of equivalent errors throughout all targets; when a user translates a block, it's really easy to accidentally break the block. This can be done by change parameter order (with % params), changing a shadow value, dropping a parameter, etc -- in the linked issue it was translating the shadow block from `device_arrow` to `devise_arrow`. This changes makes `localizeApisAsync` ignore translations when the blocks parameters get broken

image

The code was there to try and check it, but it was basically a noop; `ps` and `locps` were always equal because `pxt.blocks.compileInfo` uses the `._def` that is built and attached to the function's attributes by `updateBlockInfo`, which wasn't run until after the comparison.

Worth noting I left out two things from the comparison that can be changed in the block definition; `param.defaultValue` and `param.definitionName`. `defaultValue` seems fine to be able to change, e.g. to change a string's value to a localized equivalent, and `definitionName` seems fine as it already checks the types match / it is associated properly with a parameter as defined in typescript. Can add those to the checks and fail them if we want, but it would probably invalidate extra translations that are mostly valid / not currently breaking blocks.

### DIFF
--- a/pxtlib/service.ts
+++ b/pxtlib/service.ts
@@ -658,8 +658,7 @@ namespace ts.pxtc {
                             fn.attributes.block = nsDoc;
                         }
                         updateBlockDef(fn.attributes);
-                    }
-                    else if (fn.attributes.block && locBlock) {
+                    } else if (fn.attributes.block && locBlock) {
                         const ps = pxt.blocks.compileInfo(fn);
                         const oldBlock = fn.attributes.block;
                         fn.attributes.block = pxt.blocks.normalizeBlock(locBlock, err => errors[`${fn.attributes.blockId}.${lang}`] = 1);
@@ -674,6 +673,8 @@ namespace ts.pxtc {
                                 updateBlockDef(fn.attributes);
                             }
                         }
+                    } else {
+                        updateBlockDef(fn.attributes);
                     }
                 })
             })))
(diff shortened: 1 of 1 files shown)


pxtlib/service.ts
```diff
@@ -688,6 +692,21 @@ namespace ts.pxtc {
...
+    function hasEquivalentParameters(a: pxt.blocks.BlockCompileInfo, b: pxt.blocks.BlockCompileInfo) {
+        if (a.parameters.length != b.parameters.length)
+            return false;
+
+        for (const aParam of a.parameters) {
+            const bParam = b.actualNameToParam[aParam.actualName];
+            if (!bParam
+                || aParam.type != bParam.type
+                || aParam.shadowBlockId != bParam.shadowBlockId) {
+                return false;
```
pelikhan, line 704: add ``pxt.debug`` logging, it's very helpful to pin point exctaly what happened here.

pelikhan:
Is there a way to run these tests from the CLI? We could run tests daily use a GitHub Action cron job.

jwunderl:
I think it'd be simple enough to check as a cron job- it's already running this function whenever we load blocks anyways in `getBlocksAsync` - so it would mainly be a matter of pulling all translations and loading packages in each language.

I guess the question would be what we do with failures; we could remove approval with crowdins api, but with this we'll be ignoring / not breaking for users anyway, and removing the approval would make it harder to see what's wrong with the translation. Maybe post a comment or an issue on the translation?

PR #7416, opened 2020-08-25
Get usedblocks when checking snippets, save to target.json

shakao:
Build https://minecraft.makecode.com/app/4d2ba8743c44fbea28e32007b8aed731d205caeb-0268f72fcd

Marks tutorial snippets with the source file URL and an "extractIds" flag. When decompiling each snippet during checkdocs, we extract the blockIds and save them to a file (used when pxt serve called after checkdocs, eg locally) and patch them into the target.json (when uploading, the target is built before checkdocs is called).

Also includes a change for deleting stale usedblocks entries from IndexedDb

### DIFF
--- a/webapp/src/tutorial.tsx
+++ b/webapp/src/tutorial.tsx
@@ -22,6 +22,8 @@ type ISettingsProps = pxt.editor.ISettingsProps;
 export function getUsedBlocksAsync(code: string[], id: string, language?: string): Promise<pxt.Map<number>> {
     if (!code) return Promise.resolve({});
 
+    if (pxt.appTarget?.tutorialInfo && pxt.appTarget.tutorialInfo[id]) return Promise.resolve(pxt.appTarget.tutorialInfo[id]);
+
     return pxt.BrowserUtils.tutorialInfoDbAsync()
         .then(db => db.getAsync(id, code)
             .then(entry => {
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -1984,6 +1984,13 @@ function updateTOC(cfg: pxt.TargetBundle) {
     }
 }
 
+function updateTutorialInfo(cfg: pxt.TargetBundle) {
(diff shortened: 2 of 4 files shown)


webapp/src/tutorial.tsx
```diff
@@ -22,6 +22,8 @@ type ISettingsProps = pxt.editor.ISettingsProps;
 export function getUsedBlocksAsync(code: string[], id: string, language?: string): Promise<pxt.Map<number>> {
     if (!code) return Promise.resolve({});
 
+    if (pxt.appTarget?.tutorialInfo && pxt.appTarget.tutorialInfo[id]) return Promise.resolve(pxt.appTarget.tutorialInfo[id]);
```
jwunderl, line 25:
```suggestion
    if (pxt.appTarget?.tutorialInfo?.[id]) return Promise.resolve(pxt.appTarget.tutorialInfo[id]);
```

pelikhan (changes requested):
You should key the cached data on the hash of the tutorial content, not the path so that any live update To the tutorial invalidate they cache. Otherwise this will lead to very confusing bugs.
What is the story for localized tutorials?

shakao:
@jwunderl hmm well, we could would only need an update when the code snippets in a tutorial change and not for general typos/instruction updates, but that's still a good point. i'll brainstorm some options/alternatives

@pelikhan it's saving the block ids so it will also work for localized tutorials, but yeah, i'll think about a way to live update this json or maybe limit it to the HoC tutorials or something...

pelikhan:
If you use the hash, you will just loose caching when you make an update. This is safer; otherwise we'll surely forget about this caching and eventually wonder why the tutorials are not loading blocks although we've made markdown changes.
A better solution would be to compute the block ids in the cloud on the fly and store it in a blob at the hash location of the markdown. This is a common caching paradigm we have.

PR #7420, opened 2020-08-25
Several small arcade bug fixes

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/2367
Fixes https://github.com/microsoft/pxt-arcade/issues/2368
Fixes https://github.com/microsoft/pxt-arcade/issues/2366
Fixes https://github.com/microsoft/pxt-arcade/issues/2363
Fixes https://github.com/microsoft/pxt-arcade/issues/2357

Also fixes a longstanding bug with the folding behavior where monaco folds image literals in such a way that it becomes impossible to enter a newline after them. 

For example, here is sprites.create in the current editor:
Screen Shot 2020-08-25 at 3 13 25 PM

And here it is with these changes:
Screen Shot 2020-08-25 at 3 14 04 PM

In the first picture you would need to unfold the code before you could add another line the document because pressing enter would just edit the image literal and cause it to fold again.


riknoll:
Also, each commit is 1 bug so that might be the easiest way to review

PR #7422, opened 2020-08-26
Confirm github dependencies available before loading packages

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/2371; this tries to fetch the github dependencies of the file before installing the header. Happy to look for an alternative approach if this feels wrong, but it seems okay to me as we'll need to fetch those dependencies for the project to actually load any way, and it shouldn't take any longer as the extensions get cached locally during `downloadPackageAsync`

And a few related fixes; quit the loading screen and return home with an error message if you try and load an old project and something prevents the header from being loaded (the second part of that issue), as well as a toast for when it fails to load an example (we didn't run into this in microbit as there are only a few non tutorial/project examples, and none of them had extensions)

Also, finally fixes https://github.com/microsoft/pxt-electron/issues/196 (tutorials occasionally failing to load until you refresh), as https://github.com/microsoft/pxt-arcade/issues/2371 actually just ended up being the repro case for it after I fixed the first portions. The fix for that is https://github.com/microsoft/pxt/commit/f5746ea5ef656f1ce120bfe7bd2e3c4a28c2ceb7 - when we sent out a request to something other than the local server while offline, it could potentially get an offline error, which would flip`_isOnline` and cause all future requests to fail, even when they're just to the server. This just says to always attempt the request if we're in pxt electron

Also, we were sending requests for `/docs//tutorial/...`, which we technically handle by ignoring the double slash but no reason to make things hard on ourselves.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -2193,24 +2193,7 @@ export class ProjectView
         }
         files[pxt.CONFIG_NAME] = pxt.Package.stringifyConfig(cfg);
 
-        const ghExtensions = Object.keys(cfg.dependencies)
-                ?.filter(dep => pxt.github.isGithubId(cfg.dependencies[dep]));
-
-        if (ghExtensions.length) {
-            const pkgConfig = await pxt.packagesConfigAsync();
-            // Make sure external packages load before installing header.
-            await Promise.all(
-                ghExtensions.map(
-                    async ext => {
-                        const extSrc = cfg.dependencies[ext];
-                        const ghPkg = await pxt.github.downloadPackageAsync(extSrc, pkgConfig);
-                        if (!ghPkg) {
-                            throw new Error(lf("Cannot load extension {0} from {1}", ext, extSrc));
-                        }
-                    }
-                )
-            );
-        }
+        await pxt.github.cacheProjectDependenciesAsync(cfg);
 
         const hd = await workspace.installAsync({
             name: cfg.name,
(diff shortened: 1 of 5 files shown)


webapp/src/app.tsx
```diff
@@ -2192,7 +2192,27 @@ export class ProjectView
...
+        if (ghExtensions.length) {
+            const pkgConfig = await pxt.packagesConfigAsync();
+            // Make sure external packages load before installing header.
+            await Promise.all(
+                ghExtensions.map(
+                    async ext => {
+                        const extSrc = cfg.dependencies[ext];
+                        const ghPkg = await pxt.github.downloadPackageAsync(extSrc, pkgConfig);
+                        if (!ghPkg) {
+                            throw new Error(lf("Cannot load dependency {0} from {1}", ext, extSrc));
```
pelikhan, line 2208: dependency is an internal jargon, should be "extension"

webapp/src/app.tsx
```diff
@@ -2269,7 +2290,10 @@ export class ProjectView
                         })
                 }
             })
-
+            .catch(e => {
+                core.warningNotification(lf("Please check your internet connection and check the example is valid."));
```
pelikhan, line 2294: add reporting pxt.reportException

jwunderl:
Added in one last fix with the last commit; right now if you try and load a share link with `pub:` while offline or the link is invalid it just sticks you in the editor in a broken state like this: https://arcade.makecode.com/#pub:32akjsdlfajksdlfjasdklfasd -- this was happening with the games in the game jam row in the offline app.

https://github.com/microsoft/pxt/pull/7422/commits/8465a3885cf20d1388080e83135ddb1f8bd5287c makes it bounce you back out to the homescreen when the load fails due to missing project / dependencies

pelikhan (approved):

PR #7423, opened 2020-08-26
Rembember me

pelikhan:
Adds a ``remember me`` checkbox to github login dialogs. Remember me in formation is used to decide to store the token in localstorage or not.

When a user checks ``remember me``, we store the OAuth ``state`` in a new localstorage key, ``oauthRememberState`` when starting the login flow. After redirecting from the OAuth, we check if the ``state`` matches the ``oauthRememberState`` key to determine if ``remember me`` was selected.

- [x] force rememberme parameter in all paths
- [x] use state to check remember me
- [x] updated github-explorer (tested both cases)
- [x] updated github-classroom-redirect
- [x] use in-memory token instead of always reading from localstorage

Notes:

It's unclear if the onedrive/google auth path are still working with this since they are mostly dead code/untestable.

### DIFF
--- a/docs/github-classroom-redirect.html
+++ b/docs/github-classroom-redirect.html
@@ -37,7 +37,8 @@
                 }
 
                 const assignment = localStorage["core/ghclassroomassigment"];
-                localStorage["core/githubtoken"] = keys["access_token"];
+                // do not store access token as the user has not explicitely allow storage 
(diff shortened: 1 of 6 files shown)


abchatra:
Looks good to me. This does remove the existing token if the user doesn't select remember me right? I couldn't follow where this w as done.

abchatra:

pelikhan:
it's ready.

PR #7424, opened 2020-08-26
Consider activeLayer data in addition to image data when filling

jacobcarpenter:
Adjusts fill to be based on a computed value that includes data from the `activeLayer` in addition to the pixel value from the `image` layer.

fixes microsoft/pxt-arcade#2374

Kapture 2020-08-26 at 12 27 48

### DIFF
--- a/webapp/src/components/ImageEditor/toolDefinitions.ts
+++ b/webapp/src/components/ImageEditor/toolDefinitions.ts
@@ -650,9 +650,11 @@ export class FillEdit extends Edit {
     }
 
     protected doEditCore(state: EditState) {
-        // Read image layer, but write to the active layer; allows wall fill to fill based on tiles
-        const colorToReplace = state.image.get(this.col, this.row);
-        if (state.activeLayer === state.image && colorToReplace === this.color) {
+        const includeActiveLayerData = state.activeLayer !== state.image;
+        const getData = (col: number, row: number) => (includeActiveLayerData ? state.activeLayer.get(col, row) << 8 : 0) + state.image.get(col, row);
+
+        const colorToReplace = getData(this.col, this.row);
+        if (colorToReplace === this.color) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


webapp/src/components/ImageEditor/toolDefinitions.ts
```diff
@@ -650,9 +650,11 @@ export class FillEdit extends Edit {
     }
 
     protected doEditCore(state: EditState) {
-        // Read image layer, but write to the active layer; allows wall fill to fill based on tiles
-        const colorToReplace = state.image.get(this.col, this.row);
-        if (state.activeLayer === state.image && colorToReplace === this.color) {
+        const includeActiveLayerData = state.activeLayer !== state.image;
+        const getData = (col: number, row: number) => (includeActiveLayerData ? state.activeLayer.get(col, row) << 4 : 0) + state.image.get(col, row);
```
jacobcarpenter, line 654: Oh! While looking at @riknoll's tilemap copy+paste PR, I realized: maybe the data in `state.image` isn't guaranteed to fall within the range of `0..15`... I don't actually know how tilemaps with >16 unique tiles are stored.

Is my assumption wrong about the range of data in `state.image` wrong? If so, the shift + addition to combine the activeLayer with the image layer could result in collisions...

jwunderl: Ah, right. I read it backwards and was thinking the wall color was getting stored in the spot for 0-15.

the color for the tile is just an id for the tile as well, and should be from `0..255` I believe

jacobcarpenter: @jwunderl Yes, it looks like `EditState` references [this `Bitmap`](https://github.com/microsoft/pxt/blob/ed25e1fbbdcf8dee691274bde4a4c227842f520a/pxtlib/spriteutils.ts#L38-L42) which is backed by a `Uint8ClampedArray` (0..255).

I kind of like keeping the image data in the low order bits. Any objection to just moving the activeLayer data to `<< 8`?

jwunderl: No objection from me at least!

PR #7426, opened 2020-08-26
Add new FieldVariableDropdown subclass 

livcheerful:
fixes https://github.com/microsoft/pxt-minecraft/issues/657

Subclass of FieldVariable so we can filter out the "Delete" option

### DIFF
--- a/pxtblocks/fields/field_argumentvariable.ts
+++ b/pxtblocks/fields/field_argumentvariable.ts
@@ -4,7 +4,7 @@ namespace pxtblockly {
      * variables are part of a function argument (or else the whole function
      * gets deleted).
     */
-    export class FieldArgumentVariableDropdown extends Blockly.FieldVariable {
+    export class FieldArgumentVariable extends Blockly.FieldVariable {
         constructor(varName: string) {
             super(varName);
             this.menuGenerator_ = this.dropdownCreate;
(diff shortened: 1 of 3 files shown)


pxtblocks/fields/field_variabledropdown.ts
```diff
@@ -0,0 +1,13 @@
+namespace pxtblockly {
+    export class FieldVariableDropdown extends Blockly.FieldVariable {
```
shakao, line 2: LGTM, maybe give this a name that's a little more descriptive of the fact that it's a "variable dropdown minus delete option" eg "FieldVariableSelectDropdown", "FunctionArgumentVariableDropdown" or something.... like that but better? or you could just leave a comment explaining what it is

pxtblocks/fields/field_argumentvariable.ts
```diff
@@ -0,0 +1,18 @@
...
+    export class FieldArgumentVariable extends Blockly.FieldVariable {
+        constructor(varName: string) {
+            super(varName);
+            this.menuGenerator_ = this.dropdownCreate;
+        }
+        dropdownCreate(): any {
+            const options = Blockly.FieldVariable.dropdownCreate.call(this);
+            return options.filter((opt: any) => opt[1] != Blockly.DELETE_VARIABLE_ID);
+        }
+
```
jwunderl, line 16:
```suggestion
```

fix lint

pxtblocks/fields/field_variabledropdown.ts
livcheerful: changed the name to be FieldArgumentVariable, and added a comment!

PR #7428, opened 2020-08-27
CLI command for collecting block ids used in tutorial markdown

shakao:
- Separate CLI command for collecting blockIds, run on an allowlist of directories specified in pxtarget.json (eg ["hour-of-code", "tutorials"])
- Key on a hash of the tutorial code, pxt version, and target version--only use prebuilt blocks if the hash matches
- Used blocks saved in target.json

### DIFF
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -425,7 +425,11 @@ declare namespace pxt {
         bundleddirs: string[];
         versions: TargetVersions;        // @derived
         apiInfo?: Map<PackageApiInfo>;
-        tutorialInfo?: Map<Map<number>>;
+        tutorialInfo?: Map<BuiltTutorialInfo>; // hash of tutorial code mapped to prebuilt info for each tutorial
+    }
+
+    interface BuiltTutorialInfo {
+        usedBlocks: Map<number>;
     }
 
     interface PackageApiInfo {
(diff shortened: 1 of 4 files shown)


webapp/src/tutorial.tsx
```diff
@@ -22,6 +22,12 @@ type ISettingsProps = pxt.editor.ISettingsProps;
 export function getUsedBlocksAsync(code: string[], id: string, language?: string): Promise<pxt.Map<number>> {
     if (!code) return Promise.resolve({});
 
+    // check to see if usedblocks has been prebuilt. this is hashed on the tutorial code + pxt version + target version
+    if (pxt.appTarget?.tutorialInfo) {
+        const hash = pxt.BrowserUtils.getTutorialInfoHash(code);
+        if (pxt.appTarget.tutorialInfo[hash]) return Promise.resolve(pxt.appTarget.tutorialInfo[hash]);
```
pelikhan, line 28: add tick event for cache hit and non-hit

localtypings/pxtarget.d.ts
```diff
@@ -424,6 +425,7 @@ declare namespace pxt {
         bundleddirs: string[];
         versions: TargetVersions;        // @derived
         apiInfo?: Map<PackageApiInfo>;
+        tutorialInfo?: Map<Map<number>>;
```
pelikhan, line 428: add comment to describe the meaning of this data structure?

PR #7430, opened 2020-08-27
if cached markdown > a week old, wait for new markdown

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/2361

The issue is that we currently always return the markdown from markdownAsync, even if it's weeks / months / years old. In the linked issue, it appears a prior test version of the electron app had been installed a long time ago, so it had a very old / broken homepage until a refresh.

This gives a second tier - if the content is over a week old, ignore the cached content completely and wait for the new content to come in. This helps minimize the chance old content just 404s, or is completely invalid due to e.g. changes in external dependencies. I think this along with the change @livcheerful is working on to invalidate the caches when we get new editor versions should handle most cases of broken links / examples on home page?

I have another branch with a bit more refactoring to make it use async / await instead of the nested promise; putting that up as a separate PR https://github.com/microsoft/pxt/pull/7431 to separate the behavioral change from a refactoring.

(also as a side effect, this change will make sure we never use cached content in pxt serve, which is very helpful for iterating on docs content)

### DIFF
--- a/pxtlib/emitter/cloud.ts
+++ b/pxtlib/emitter/cloud.ts
@@ -114,7 +114,9 @@ namespace pxt.Cloud {
     }
 
     // 1h check on markdown content if not on development server
-    const MARKDOWN_EXPIRATION = pxt.BrowserUtils.isLocalHostDev() ? 1 : 1 * 60 * 60 * 1000;
+    const MARKDOWN_EXPIRATION = pxt.BrowserUtils.isLocalHostDev() ? 0 : 1 * 60 * 60 * 1000;
+    // 1w check don't use cached version and wait for new content
+    const FORCE_MARKDOWN_UPDATE = MARKDOWN_EXPIRATION * 24 * 7;
     export function markdownAsync(docid: string, locale?: string): Promise<string> {
         locale = locale || pxt.Util.userLanguage();
         const live = pxt.Util.localizeLive;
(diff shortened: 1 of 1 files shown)


pxtlib/emitter/cloud.ts
```diff
@@ -114,25 +114,34 @@ namespace pxt.Cloud {
     }
 
     // 1h check on markdown content if not on development server
-    const MARKDOWN_EXPIRATION = pxt.BrowserUtils.isLocalHostDev() ? 1 : 1 * 60 * 60 * 1000;
+    const MARKDOWN_EXPIRATION = pxt.BrowserUtils.isLocalHostDev() ? 0 : 1 * 60 * 60 * 1000;
```
livcheerful, line 117: I know this isn't a change but I'm curious - what's the 1 here for?

jwunderl: the 1 in `1 * 60`? Just for expressing the time fully - 1 hour * 60 minutes * 60 seconds * 1000 ms. Not necessary at all, but I've seen it written both ways / possibly prevents silly typos if someone wanted to change it to 2 hours later.

pelikhan (approved):

pxtlib/emitter/cloud.ts
livcheerful: oo cool

PR #7433, opened 2020-08-28
Default to focus on monaco loading

livcheerful:
fixes https://github.com/microsoft/pxt-minecraft/issues/1761

Not sure if this will affect accessibility or anything?

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1149,6 +1149,9 @@ export class ProjectView
             if (isCompleted && pxt.commands.onTutorialCompleted) pxt.commands.onTutorialCompleted();
             // Hide flyouts and popouts
             this.editor.closeFlyout();
+            if (this.textEditor.giveFocusOnLoading) {
+                this.textEditor.editor.focus();
+            }
         }
     }
 
@@ -3512,7 +3515,7 @@ export class ProjectView
             .finally(() => core.hideLoading("tutorial"));
     }
 
-    startActivity(activity: pxt.editor.Activity, path: string, title?: string, editorProjectName?: string) {
+    startActivity(activity: pxt.editor.Activity, path: string, title?: string, editorProjectName?: string, focus?: boolean) {
         pxt.tickEvent(activity + ".start", { editor: editorProjectName });
         switch (activity) {
             case "tutorial":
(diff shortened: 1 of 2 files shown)


jwunderl:
I think it's fine accessibility wise when you navigate to the page from within the app, but if you were given a direct link (e.g. if someone shared a link to `https://minecraft.makecode.com/beta?ipc=1&inGame=1#tutorial:ts:/tutorials/chicken-rain`) it might not be ideal. When we load in for the first time we need the accessibility toolbar to be focusable easily / as the first few stops when outside modals, but if we focus to the monaco editor to start it's hard to get there - the menu isn't available when you tab, and you also need to know to use `shift+m` or `ctrl+shift+m` to switch tab from inserting a tab to navigating the page.

livcheerful:
ooo good point. I'll make the changes

livcheerful:
I just tested it for accessibility, and there's currently no way to get to the tutorial steps anyways, even with tab navigation. I'll file a bug in arcade, since Minecraft isn't accessible

webapp/src/app.tsx
```diff
@@ -3512,7 +3515,7 @@ export class ProjectView
             .finally(() => core.hideLoading("tutorial"));
     }
 
-    startActivity(activity: pxt.editor.Activity, path: string, title?: string, editorProjectName?: string) {
+    startActivity(activity: pxt.editor.Activity, path: string, title?: string, editorProjectName?: string, focus?: boolean) {
```
jwunderl, line 3518:
```suggestion
    startActivity(activity: pxt.editor.Activity, path: string, title?: string, editorProjectName?: string, focus = true) {
```

and you can change portion below to

```typescript
this.textEditor.giveFocusOnLoading = focus;
```

livcheerful: don't the optional parameters need to be last though? I can move everything XD I guess I was just scared of changing all the calls to startActivity

livcheerful: oh lol I just tried it and it works. Dunno why I thought the optional parameters needed to be last!

jwunderl: Optional parameters have to be set after required parameters, but optional parameters and default parameters can be mixed / order doesn't matter -- default parameters just emit to javascript with an extra if at the front of the function e.g.

image

livcheerful: oooooo cool thanks Joey!

PR #7436, opened 2020-08-31
clean up config.name vs pkg id in package api caching

jwunderl:
re: last portion of message in https://github.com/microsoft/pxt/pull/7427 and discussion last friday

**[edit]** + a quick fix along the same code path for extension caching when we add in extra built files, like tilemap of code start.

After looking through stuff a bit more I think this is the right way to handle things. I don't think this necessarily needs to go into this release, as it hasn't been run into in arcade yet / is a pretty subtle case

quick context on all the relevant fields;
* `SymbolInfo.pkg` is the package that the api came from. this gets picked from the symbol info we get out of the ts compiler base off the name of the folder the symbol is found in https://github.com/microsoft/pxt/blob/7589ccebd32df24e1cb7ceb8e620eb4e617ab55c/pxtcompiler/emitter/service.ts#L218-L225
* epkg.getPkgId gets the id for the editor package (this is equal to pxt.package / kspkg .id, if kspkg exists). This id typically matches either pkg.config.name or an alias defined by the user if it's not a builtin pkg (as described in the linked PR), and is used when passing things to the compiler, eg https://github.com/microsoft/pxt/blob/7589ccebd32df24e1cb7ceb8e620eb4e617ab55c/pxtlib/package.ts#L1071-L1081
* pkg.config.name is the name of the package as listed in pxt.json
* Package._verspec describes the packages version; for github dependencies this will be `github:{repo slug}(#{tag})?`, for builtin packages this will usually be '*', and for the main package / user local deps / etc this will be something like `workspace:60fc5898-522b-40a5-7c4f-77754c92fce5`

So, at runtime, the information we have about a block <-> api correlation is the package id, which will link to a specific package. This value is meaningless when persisting the apis though / is specific to a project, so we would want to key it off some other value -- verspec if there's a meaningful one is useful, otherwise just config.name is likely the best. So, we want to accumulate the apis using the package id, and store them in the db based off the relevant key.

** side note, we had discussed adding a field containing the verspec to the block; after looking into it this seems like it would be either a hacky implementation of building it wherever we call `getApiInfo`, or require a decent amount reworking / passing things around that I'm not sure is warranted. This is because the point at which we create the api info only really has context for the ts program, and not our view of the program; which is why we e.g. get the package of the block by pulling it out of the file path. If we want that information, I think a method on `pkg.EditorPackage` would be better (`getApiSource(info: ApiInfo): pkg.EditorPackage`), that takes an api's .pkg, uses that to find the package with the same id, and returns that (or it's `.config.name` / `._verspec` like I have it doing in `getPackageKey`)

### DIFF
--- a/pxtlib/main.ts
+++ b/pxtlib/main.ts
@@ -502,6 +502,8 @@ namespace pxt {
     export const DEFAULT_GROUP_NAME = "other"; // used in flyout, for snippet groups
     export const TILEMAP_CODE = "tilemap.g.ts";
     export const TILEMAP_JRES = "tilemap.g.jres";
+    export const TUTORIAL_CODE_START = "_onCodeStart.ts"
+    export const TUTORIAL_CODE_STOP = "_onCodeStop.ts"
 
     export function outputName(trg: pxtc.CompileTarget = null) {
         if (!trg) trg = appTarget.compile
--- a/pxtlib/main.ts
+++ b/pxtlib/main.ts
@@ -503,8 +503,8 @@ namespace pxt {
     export const DEFAULT_GROUP_NAME = "other"; // used in flyout, for snippet groups
     export const TILEMAP_CODE = "tilemap.g.ts";
     export const TILEMAP_JRES = "tilemap.g.jres";
-    export const TUTORIAL_CODE_START = "_onCodeStart.ts"
-    export const TUTORIAL_CODE_STOP = "_onCodeStop.ts"
+    export const TUTORIAL_CODE_START = "_onCodeStart.ts";
+    export const TUTORIAL_CODE_STOP = "_onCodeStop.ts";
     export const TUTORIAL_INFO_FILE = "tutorial-info-cache.json";
 
     export function outputName(trg: pxtc.CompileTarget = null) {
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -2181,14 +2181,12 @@ export class ProjectView
         }
         if (options.tutorial && options.tutorial.metadata) {
             if (options.tutorial.metadata.codeStart) {
-                let codeStart = "_onCodeStart.ts";
-                files[codeStart] = "control._onCodeStart('" + pxt.U.htmlEscape(options.tutorial.metadata.codeStart) + "')";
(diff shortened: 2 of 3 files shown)


riknoll:
Wasn't there also a problem of the APIs getting misattributed to the wrong package when the ID doesn't match? Or else how were we cacheing an empty API info before?

jwunderl:
Yeah, that was the comparison between `.pkg` and `.getKsPkg().config.name` here: https://github.com/microsoft/pxt/pull/7436/files#diff-ee7e9c6bd3fb366a240fa278c79b8ebfR664-R676

The issue was that `.getPkgId` and that were usually the same, but not when the id was explicitly specified - so the .pkg was equal to the ID in that case but not the name, and we were only comparing against the name. The id should always match the `.pkg` so that would remain fixed here

jwunderl:
@shakao the discussion this morning about reminded me I had to complete this pr, and while looking it over I think I found the bug we were hitting for hoc tutorials:

https://github.com/microsoft/pxt/blob/49e3a67fa2235899a31560f096c540a4613ffe90/webapp/src/compiler.ts#L549-L553

the hoc tutorials have `_onCodeStart.ts` / `_onCodeStop.ts` (and arcade now has `.g.ts` asset files) so this check probably just needs to be updated?

shakao:
awesome! are you seeing it skip the first compile for HoC tutorials now? (just to verify that there aren't any additional issues with the HoC extension) i'll also try it out ingame later today after testing

jwunderl:
I need to merge in master real quick to do a full test but it's hitting the cached api info with this change~

abchatra:
Awesome. Does this bug exists in todays live editor (1.4.10) though?

jwunderl:
yup: https://github.com/microsoft/pxt/blob/v5.33.16/webapp/src/compiler.ts#L499-L503

here's a build that I think should be up to date with master in both repos: https://minecraft.makecode.com/app/6de0b281e406e62d589fb8c14bb4492816eefa40-e838f285ae

abchatra:
Awesome @jwunderl & @shakao

abchatra:

PR #7446, opened 2020-09-03
minor fixes for api cards

jwunderl:
A few minor fixes api cards I noticed while testing https://github.com/microsoft/pxt-arcade/pull/2325

* Filter out deprecated, ~`shim=TD_ID`~, and blockAliasFor (e.g. the second version of array.pop as a statement)
* create code card for `.blockBuiltin` (only hits array.length and string.length I believe?)
* use `.blockNamespace` if it exists (make arrays, images, loops show up)

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -409,12 +409,15 @@ namespace pxt.blocks {
         // inject Blockly with all block definitions
         return blockInfo.blocks
             .map(fn => {
+                const comp = compileInfo(fn);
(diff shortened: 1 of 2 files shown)


riknoll:
Shouldn't we leave the shim=TD_ID ones? some of those are in the toolbox as standalone blocks

jwunderl:
Right; I didn't realize there were some shown as standalone blocks / thought they were all just used as shadows. removed that check, thanks

riknoll (approved):

PR #7449, opened 2020-09-03
Clear indexed db on reset

livcheerful:
Lemme know if we don't want this - I'm adding this for https://github.com/microsoft/pxt-minecraft/issues/1936 , since Sheila expected the Reset button to help her recover, but it didn't

### DIFF
--- a/webapp/src/workspace.ts
+++ b/webapp/src/workspace.ts
@@ -1363,6 +1363,7 @@ export function resetAsync() {
         .then(() => {
             pxt.storage.clearLocal();
             data.clearCache();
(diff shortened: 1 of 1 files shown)


riknoll:
Seems like the right behavior to me. @pelikhan @abchatra FYI

jwunderl (approved):

pelikhan:
Not waiting on the promise?

abchatra:
Cleanup is the right thing to do. But should await on the async call?

eanders-MS:
You can explicitly not await by calling `.done()`. It's essentially an empty `.then` handler.

PR #7450, opened 2020-09-03
Bumping Blockly

riknoll:
Localization for https://github.com/microsoft/pxt-blockly/pull/325 and https://github.com/microsoft/pxt-blockly/pull/322

and bumping blockly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
   },
   "devDependencies": {
     "monaco-editor": "0.19.3",
-    "pxt-blockly": "3.0.21",
+    "pxt-blockly": "3.0.23",
     "react": "16.8.3",
     "react-dom": "16.11.0",
     "react-modal": "3.3.2",
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
   },
   "devDependencies": {
     "monaco-editor": "0.19.3",
-    "pxt-blockly": "3.0.23",
+    "pxt-blockly": "3.0.24",
     "react": "16.8.3",
     "react-dom": "16.11.0",
     "react-modal": "3.3.2",
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -2402,6 +2402,7 @@ namespace pxt.blocks {
         const functionCall = pxt.blocks.getBlockDefinition(functionCallId);
 
         msg.FUNCTIONS_CALL_TITLE = functionCall.block["FUNCTIONS_CALL_TITLE"];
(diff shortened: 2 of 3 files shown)


pxtblocks/blocklyloader.ts
```diff
@@ -2635,6 +2636,29 @@ namespace pxt.blocks {
                 setOutputCheck(this, typeName, cachedBlockInfo);
             };
         }
+
+        /**
```
jacobcarpenter, line 2640: Is this chunk generated? This seems like kind of a weird place to preserve the jsdoc comment.

pxtblocks/blocklyloader.ts
```diff
@@ -2635,6 +2636,29 @@ namespace pxt.blocks {
...
+            let functionName = block.getField("function_name").getText();
+
+            let mutation = goog.dom.createDom('mutation');
+            mutation.setAttribute('name', functionName);
+            let callBlock = goog.dom.createDom('block', null, mutation);
+            callBlock.setAttribute('type', 'function_call');
+
+            let option = {
+                enabled: block.workspace.remainingCapacity() > 0,
+                text: Util.lf("Create 'call {0}'", functionName),
```
jacobcarpenter, line 2657: Is the only functionality that needs to be patched the `text` property? Would it be better to avoid the code duplication by calling the original method and then overwrite the `text` property of the return value?

```ts
const makeCreateCallOptionOriginal = (Blockly as any).Functions.makeCreateCallOption;
(Blockly as any).Functions.makeCreateCallOption = function (block: Blockly.Block) {
   let option = makeCreateCallOptionOriginal(block);
   let functionName = block.getField("function_name").getText();
   option.text = Util.lf("Create 'call {0}'", functionName);
   return option;
}
```

pxtblocks/blocklyloader.ts
riknoll: No, we just usually preserve those in this file

pxtblocks/blocklyloader.ts
riknoll: true, I'll make the update!

jwunderl (approved):

abchatra:

shakao (approved):

package.json
```diff
@@ -84,7 +84,7 @@
   },
   "devDependencies": {
     "monaco-editor": "0.19.3",
-    "pxt-blockly": "3.0.21",
+    "pxt-blockly": "3.0.23",
```
jwunderl, line 87:
```suggestion
    "pxt-blockly": "3.0.24",
```

to pick up https://github.com/microsoft/pxt-blockly/pull/326 as well?

riknoll:
will merge today

PR #7451, opened 2020-09-04
Custom completion filtering for enums and enum constants

shakao:
- Filter completion symbols by parameter type (for call expressions) after adding local variables
- When filtering, enum members and enum constants have return type 'AnimalMob.Chicken', and we are checking against 'AnimalMob', so do some additional parsing

Fixes https://github.com/microsoft/pxt-minecraft/issues/1949


riknoll (approved):

jwunderl (approved):

darzu (approved):
Awesome!

PR #7458, opened 2020-09-11
Avoid github throttling in CI when caching used blocks

shakao:
Same as testSnippetsAsync


jwunderl (approved):
Hmm, looks like we do that in several places; maybe worth setting the environment variable in CI instead of flipping it on in different portions? https://github.com/microsoft/pxt/blob/4012f9014982904b73b7a873b6a6c16b4db44234/cli/cli.ts#L6680

but that would be a separate change / LGTM as is

shakao:
yeah, i'm not really sure what the implications of turning it on/off globally would be!

PR #7461, opened 2020-09-14
Add high contrast rules for Minecraft/inverted toolbox

livcheerful:
Here are some screenshots, but also, a target! https://minecraft.makecode.com/app/f16ee16d161060d22fb9d9005df5bce0b7f6fda8-783dcb27ba

I haven't tested the other targets to make sure they haven't changed from these edits yet, but I'll do that next. I'll do it for arcade, microbit, and LEGO, but lemme know if this list should be changed

It looks like that it's not possible to have a HC change for the field editors yet/it would need more work than just CSS changes? Tell me if this is wrong :D or if there's a field editor where it's been done before so I can take a look

the corresponding pxt-minecraft PR: https://github.com/microsoft/pxt-minecraft/pull/1992

Blocks:
image
image

Homepage:
image

Monaco:
image

### DIFF
--- a/theme/highcontrast.less
+++ b/theme/highcontrast.less
@@ -235,9 +235,9 @@
         }
     }
     #mainmenu.inverted, #homemenu.inverted {
-        background-color: @HCmainMenuTextColor !important;
         color: @HCmainMenuBackground !important;
         border-bottom: 4px solid @HCtextColor;
+        background: @HCmainMenuTextColor !important;
 
         .ui.item {
             color: @HCmainMenuBackground;
--- a/theme/highcontrast.less
+++ b/theme/highcontrast.less
@@ -115,6 +115,12 @@
         }
     }
 
+    .pxtToolbox {
+        .blocklyTreeRow:not(.blocklyTreeSelected) {
+            background-color: @HCblocklyToolboxColor !important;
+        }
+    }
+
     #monacoEditor .monacoDraggableBlock {
         background: none !important;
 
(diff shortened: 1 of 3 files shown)


theme/highcontrast.less
```diff
@@ -459,6 +473,14 @@
             }
         }
     }
+    .ui.card.link.newprojectcard{
+        background-color: #1B8800 !important;
```
jwunderl, line 477: This will definitely change the color across the editors; this button is typically the same color as the download button.

https://github.com/microsoft/pxt/blob/095250200b6d106b1b5b3c0f81bd23c2693f1e9f/theme/themes/pxt/views/card.overrides#L151-L158

Maybe fine to darken it like the script manager card in hc mode instead of pegging to a different hue?

theme/highcontrast.less
```diff
@@ -168,8 +179,10 @@
         }
     }
 
+    rect.blocklyFlyoutButtonBackground,
```
jwunderl, line 182: Is there a case of there being a `blocklyFlyoutButtonBackground` that isn't a rect? I think should be fine to just add rect without duplicating selector here.

theme/highcontrast.less
```diff
@@ -186,6 +199,7 @@
     }
     svg.blocklySvg {
         background-color: @HCblocklySvgColor !important;
+        background: @HCblocklySvgColor !important;
```
jwunderl, line 202:
```suggestion
        background: @HCblocklySvgColor !important;
```
setting background will set background-color as well / is shorthand for setting all the background- props

jwunderl (approved):

theme/highcontrast.less
livcheerful: oooo cool ty joey!

theme/highcontrast.less
livcheerful: good point good point I'll try that out!

shakao (approved):

PR #7464, opened 2020-09-16
Don't strip out python namespaces when getting used blocks

livcheerful:
fixes https://github.com/microsoft/pxt-minecraft/issues/1979

the issue was with these two snippets from the tutorial:
image

since it gets converted to "export let p = ....", it was hitting an assert in "getExportInfo" in typescript.

This is one way to fix it, Shannon also suggested just stripping out the "export" for export let and export function, but I'm still working on those regexes XD figured I'd put this up so I can get feedback on which way would be better

### DIFF
--- a/webapp/src/compiler.ts
+++ b/webapp/src/compiler.ts
@@ -316,6 +316,9 @@ export function pySnippetToBlocksAsync(code: string, blockInfo?: ts.pxtc.BlocksI
 
 // Py[] -> blocks
 export function pySnippetArrayToBlocksAsync(code: string[], blockInfo?: ts.pxtc.BlocksInfo): Promise<pxtc.CompileResult> {
+    const namespaceRegex = /^\s*namespace\s+[^\s]+\s*{([\S\s]*)}\s*$/im;
+    const exportLetRegex = /^\s*export\s(let\s*[\S\s]*)\s*$/im;
+    const exportFunctionRegex = /^\s*export\s(function\s*[\S\s]*)\s*$/im;
     const snippetBlocks = "main.blocks";
     let trg = pkg.mainPkg.getTargetOptions()
     let files: string[] = [];
@@ -339,7 +342,23 @@ export function pySnippetArrayToBlocksAsync(code: string[], blockInfo?: ts.pxtc.
             // strip the namespace declaration out of the converted snippets and concat to convert to blocks
             let ts = "";
             for (let file of files) {
-                ts += `{\n${res.outfiles[file + ".ts"]}\n}\n`
+                let match = res.outfiles[file + ".ts"].match(namespaceRegex);
+                if (match && match[1]) {
+                    let noNamespace = match[1];
+                    let lines = noNamespace.split("\n");
+                    let cleanLines = lines.map((element: string)=>{
+                        match = element.match(exportLetRegex);
+                        let strippedExportLet;
+                        match? strippedExportLet = match[1] : strippedExportLet = element;
+                        match = strippedExportLet.match(exportFunctionRegex);
+                        if (match) {
+                            return match[1];
+                        } else {
+                            return strippedExportLet;
+                        }
+                    })
+                    ts += `{\n${cleanLines.join("\n")}\n}\n`;
+                }
             }
             return decompileBlocksSnippetAsync(ts, blockInfo)
         });
--- a/webapp/src/compiler.ts
+++ b/webapp/src/compiler.ts
@@ -316,7 +316,6 @@ export function pySnippetToBlocksAsync(code: string, blockInfo?: ts.pxtc.BlocksI
 
 // Py[] -> blocks
 export function pySnippetArrayToBlocksAsync(code: string[], blockInfo?: ts.pxtc.BlocksInfo): Promise<pxtc.CompileResult> {
-    const namespaceRegex = /^\s*namespace\s+[^\s]+\s*{([\S\s]*)}\s*$/im;
     const snippetBlocks = "main.blocks";
     let trg = pkg.mainPkg.getTargetOptions()
     let files: string[] = [];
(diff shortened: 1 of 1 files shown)


shakao:
hmm, are you sure this still works for existing tutorials? i'm getting an empty toolbox when i open up python tutorials:

image

jwunderl (commented):
Hm I'm not sure I follow this change by itself; previously it was unwrapping from the namespace and passing the contents of the namespace inside it's own closure (basically just slicing off the namespace keyword and it's name), but wouldn't this just append the namespace wrapped in a closure?

e.g. if the file in res.outfiles is `namespace myStuff { console.log("hi") }` wouldn't this make it 

```
{
namespace myStuff { console.log("hi") }
}
```

which wouldn't convert to blocks?

livcheerful:
ahhh shoot shannon you're right. I thought I cleared the indexeddb but I didn't :\ will make the better fix

webapp/src/compiler.ts
```diff
@@ -341,7 +343,22 @@ export function pySnippetArrayToBlocksAsync(code: string[], blockInfo?: ts.pxtc.
...
                 let match = res.outfiles[file + ".ts"].match(namespaceRegex);
-                if (match && match[1]) ts += `{\n${match[1]}\n}\n`
+                if (match && match[1]) {
+                    let noNamespace = match[1];
+                    let lines = noNamespace.split("\n");
+                    let cleanLines = lines.map((element: string)=>{
+                        match = element.match(exportLetRegex);
+                        let strippedExportLet;
+                        match? strippedExportLet = match[1] : strippedExportLet = element;
+                        match = strippedExportLet.match(exportFunctionRegex);
```
jwunderl, line 353: maybe worth creating a new const for each math? It ends up feeling a bit hard to track for me when we reuse match variables but that's just me~

webapp/src/compiler.ts
```diff
@@ -317,6 +317,8 @@ export function pySnippetToBlocksAsync(code: string, blockInfo?: ts.pxtc.BlocksI
 // Py[] -> blocks
 export function pySnippetArrayToBlocksAsync(code: string[], blockInfo?: ts.pxtc.BlocksInfo): Promise<pxtc.CompileResult> {
     const namespaceRegex = /^\s*namespace\s+[^\s]+\s*{([\S\s]*)}\s*$/im;
+    const exportLetRegex = /^\s*export\s(let\s*[\S\s]*)\s*$/im;
```
jwunderl, line 320: Hm, it seems like this would break tutorials with custom spritekinds e.g. https://github.com/microsoft/pxt-arcade/blob/master/docs/tutorials/paddle.md? I might be missing something though

jwunderl (commented):
question \\/

webapp/src/compiler.ts
livcheerful: yeah I'll do that! I did it both ways, but switched it to this since I wasn't sure what would be preferred :)

webapp/src/compiler.ts
livcheerful: ooh I'll try it. I only tested it with minecraft tutorials

livcheerful: hmmm so it seems to work fine for the paddles ( I made a new python version of it to test ), but why do you think it would break tutorials? maybe I'm not testing rigorously enough

jwunderl: Okay I pulled for a sec and checked and we don't hit it because it uses `const` instead of `let` -- the issue I was thinking of was that if we had something like:

```typescript
namespace SpriteKind {
    export const LeftPaddles = SpriteKind.create()
}
let left_paddle = sprites.create(img`
    a 8 1
    8 8 1
`, SpriteKind.LeftPaddles)
```

it would swap to

```typescript
namespace SpriteKind {
    const LeftPaddles = SpriteKind.create()
}
let left_paddle = sprites.create(img`
    a 8 1
    8 8 1
`, SpriteKind.LeftPaddles)
```

(but again those are `const`s not `let`s so it doesn't get hit here)

livcheerful: ahhh but we don't add blocks in the toolbox for creating the custom spritekinds, right?

jwunderl: Not blocks in toolbox, but the references to `SpriteKind.LeftPaddles` would be invalid when compiling

livcheerful: hmm so I think this is only used to filter the toolbox? but maybe it's not clear and dangerous to put here?

jwunderl: The issue would be if it fails to compile and get the blocks info for the toolbox. But since we always use const for our generated values inside namespaces, we won't hit that here -- maybe worth a note in a comment that we don't want to remove the export for consts but I don't think it's something to block this fix on :)

livcheerful: sounds good I'll add that!

abchatra:

jwunderl (approved):
Change looks good to me, but do you have context on why it was converting to `export let p = ....` from python / could you explain why it's occurring a little bit for future reference? is it just the conversion to python when wrapped in a namespace forcing things to get exported?

livcheerful:
> Change looks good to me, but do you have context on why it was converting to `export let p = ....` from python / could you explain why it's occurring a little bit for future reference? is it just the conversion to python when wrapped in a namespace forcing things to get exported?



> Change looks good to me, but do you have context on why it was converting to `export let p = ....` from python / could you explain why it's occurring a little bit for future reference? is it just the conversion to python when wrapped in a namespace forcing things to get exported?

Yeah things were getting exported when we took the python snippets and converted them to TS. I think in converter.ts py2ts(), we "always return global symbols because we might need to check for name collisions downstream" (and I THINK that's what's adding the export, though I haven't nailed it down yet).

Should we just not add the export to the beginning then? I think this pattern of writing a tutorial is reasonable, where you repeat the same lines of code you had before and highlight the new ones.

PR #7465, opened 2020-09-17
allow using local translations when building staticpkg

jwunderl:
When building a staticpkg, allow user to specify a directory to act as the source of translations, instead of fetching translations from crowdin. Directory needs to have the same set up as our crowdin, i.e. the files should be found like

```
{locs-src}/{lang id}/target-strings.json
{locs-src}/{lang id}/sim-strings.json
{locs-src}/{lang id}/{package id}-strings.json
{locs-src}/{lang id}/{package id}-jsdoc-strings.json
```

I had to split up `downloadTargetTranslations` quite a bit to add the right hook in, so it ended up being easier to asyncify it and separate it out a bit. Ran test builds of arcade electron app with both downloaded translations from `locs` and prebuilt translations from `locs-src` and both worked ~

### DIFF
--- a/cli/crowdin.ts
+++ b/cli/crowdin.ts
@@ -119,7 +119,7 @@ function uploadBundledTranslationsAsync(crowdinDir: string, branch: string, prj:
         if (fs.existsSync(locdir))
             fs.readdirSync(locdir)
                 .filter(f => /strings\.json$/i.test(f))
-                .forEach(f => todo.push(path.join(locdir, f)))
+                .forEach(f => todo.push(path.join(locdir, f)));
     });
 
     pxt.log(`uploading bundled translations to Crowdin (${todo.length} files)`);
@@ -202,6 +202,7 @@ export async function buildAllTranslationsAsync(langToStringsHandlerAsync: (file
                     .filter(k => !!dataLang[k] && !strings[k])
                     .forEach(k => strings[k] = dataLang[k]);
             }
+
             const errorIds = Object.keys(errors);
             if (errorIds.length) {
                 pxt.log(`${errorIds.length} errors`);
--- a/cli/crowdin.ts
+++ b/cli/crowdin.ts
@@ -135,95 +135,90 @@ function uploadBundledTranslationsAsync(crowdinDir: string, branch: string, prj:
     return nextFileAsync();
 }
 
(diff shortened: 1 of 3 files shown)


pelikhan:
Update docs?

abchatra (approved):
Looks good to me. I am assuming there is no change in the crowdin download path, other than refactoring.

cli/crowdin.ts
```diff
@@ -135,95 +135,89 @@ function uploadBundledTranslationsAsync(crowdinDir: string, branch: string, prj:
...
             const crowdf = path.join(crowdinDir, fn);
-            const locdir = path.dirname(f);
+            const locdir = path.dirname(filePath);
             const projectdir = path.dirname(locdir);
-            pxt.log(`downloading ${crowdf}`);
-            pxt.debug(`projectdir: ${projectdir}`)
-            return pxt.crowdin.downloadTranslationsAsync(cred.branch, cred.prj, cred.key, crowdf, { translatedOnly: true, validatedOnly: true })
-                .then(data => {
-                    Object.keys(data)
-                        .filter(lang => Object.keys(data[lang]).some(k => !!data[lang][k]))
```
jwunderl, line 180: dropped this filter as it's handled better by the `if (!dataLang || !stringifyTranslations(dataLang))` inside the foreach, which will return undefined if there aren't any non empty values

cli/crowdin.ts
```diff
@@ -135,95 +135,89 @@ function uploadBundledTranslationsAsync(crowdinDir: string, branch: string, prj:
...
 }
 
-export function downloadTargetTranslationsAsync(parsed?: commandParser.ParsedCommand) {
-    const name = (parsed && parsed.args[0]) || "";
-    const crowdinDir = pxt.appTarget.id;
-    return crowdinCredentialsAsync()
-        .then(cred => {
-            if (!cred) return Promise.resolve();
+export async function downloadTargetTranslationsAsync(parsed?: commandParser.ParsedCommand) {
+    const name = parsed?.args[0];
```
jwunderl, line 139: keeping this as `undefined` instead of `""`; we did a check against falsey values when using this anyway, no reason to coerce to string

jwunderl (commented):
Yup, just refactoring along that path besides a few falsey checks -- I added some annotations to point them out for clarity's sake

PR #7466, opened 2020-09-18
Compute TS AST for Python completions, remove completion list truncation

shakao:
Fixes https://github.com/microsoft/pxt-minecraft/issues/1990

Issue here is the intellisense for call expressions requires our annotated TS AST, which wasn't being computed for Python. I added an additional compile (after the py2ts) to fill in this info.

Fixes https://github.com/microsoft/pxt-minecraft/issues/1989

This is because getCompletions is only called when intellisense initially pops up, and subsequent keystrokes just filter the existing list (eg if you type "PRI" quickly and wait, you'll see "PRIMED_TNT", but if you type "P", wait, then add "RI" you will not). PRIMED_TNT isn't in the initial list because it's a "ProjectileMob" and mobs.spawn is preferencing AnimalMobs. Adding `incomplete: true` will recompute the completions each time.

I also updated our pre-monaco completion result filtering to be case-insensitive (so that "pri" will also bring up "PRIMED_TNT").

These are all kind of perf hits--here's a build, I do think it makes the intellisense noticeably slower so it might be worth discussing whether we should ditch some of these fixes in favor of better perf. Or if folks have ideas for better fixes! (cc @darzu)

Build: https://minecraft.makecode.com/app/e1999a6e25f69f5e6ddbfe5ae1b26d967894fe03-36536a76cc

### DIFF
--- a/pxtcompiler/emitter/languageservice.ts
+++ b/pxtcompiler/emitter/languageservice.ts
@@ -591,7 +591,7 @@ namespace ts.pxtc.service {
         resultSymbols.sort(compareCompletionSymbols);
 
         // limit the number of entries
-        if (resultSymbols.length > MAX_SYMBOLS) {
+        if (v.light && resultSymbols.length > MAX_SYMBOLS) {
             resultSymbols = resultSymbols.splice(0, MAX_SYMBOLS)
         }
 
--- a/pxtcompiler/emitter/languageservice.ts
+++ b/pxtcompiler/emitter/languageservice.ts
@@ -409,8 +409,11 @@ namespace ts.pxtc.service {
             // better at fuzzy matching and fluidly changing but for performance reasons we want to do it here)
             if (!isMemberCompletion && resultSymbols.length > MAX_SYMBOLS_BEFORE_FILTER) {
                 resultSymbols = resultSymbols
-                    .filter(s => (isPython ? s.symbol.pyQName : s.symbol.qName).indexOf(partialWord) >= 0)
+                    .filter(s => (isPython ? s.symbol.pyQName : s.symbol.qName).toLowerCase().indexOf(partialWord.toLowerCase()) >= 0)
             }
+
+            opts.ast = true;
+            const ts2asm = compile(opts, service);
         } else {
             tsPos = position
             opts.ast = true;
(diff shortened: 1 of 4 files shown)


jwunderl (approved):
LGTM, looks like the case insensitive match change is causing the test to fail though - but really, that issue is https://github.com/microsoft/pxt-arcade/issues/2430, as it's not narrowing to just types and is providing a bunch of other invalid translations and bumping out string. It's currently suggesting the wrong string (`String` instead of `string`) and not high enough to be valuable anyways:

live:

image

and beta's worse

livcheerful (approved):

darzu:
This is tricky territory. The "PRIMED_TNT" is interesting. It used to be that this would show up because the initial computation of the dropdown returned all possible matches and the subsequent filtering would always find the item no matter what you typed.

However we felt perf was too slow so I deliberately introduced limits on how many items were returned by intellisense in order to speed the initial call up. The downside of this capping is that the situation above happens.

You've found another way to do this which is to keep getting items. But I worry this will be pretty bad for perf. It used to be that filtering was very instantaneous because that was all handled by monaco.

The perfect solution I think would be if we could return some hits immediately and slowly fill in more and more intellisense items with more time, but leave all the filtering to the monaco client side.

My feeling now is that removing the capping I added is probably the right solution. The capping improved perf a bit, but I think the delta of that gain is less than the net loss from the compute it every time strategy.

darzu:
Thinking a little more, I believe the best solution is if we divide up and remember the work done by the compiler and language service better.

It should be the case that a change by the user causes one compile which will then store a bunch of analysis state that was necessary for the compilation process anyway, and then subsequent calls to the language service functions (completions, snippets, hover help, etc) will be snappy because the bulk of the work was done for them already during that first compile.

Or maybe refining this a bit more, the one compile after a change can return the errors to the webapp asap and then once that's been returned the compile service immediately starts computing all remaining auxiliary data structures needed by the language service so that once a language service call is made the answer should be a quick lookup.

darzu:
For this PR, my recommendation is we remove the hard limit on number of results returned.

shakao:
@darzu Cool, that makes sense! Abhijith had a good suggestion, I've removed the capping in general to fix this bug, but I'm leaving it in/turning it on if the user is in light mode--my feeling is the tradeoff for slightly less accurate intellisense vs better perf goes the other way in that scenario

PR #7469, opened 2020-09-22
guard localStorage + webusb access

pelikhan:
We should not access localStorage directly as it throws in the wrong security context. Same for webusb.

Impacts micro:bit, will require a hot fix.

### DIFF
--- a/pxteditor/shell.ts
+++ b/pxteditor/shell.ts
@@ -60,4 +60,26 @@ namespace pxt.shell {
         init();
         return layoutType == EditorLayoutType.Controller;
     }
+
+    export function isPyLangPref(): boolean {
+        return pxt.storage.getLocal("editorlangpref") == "py";
+    }
+
+    export function getEditorLanguagePref(): string {
+        return pxt.storage.getLocal("editorlangpref");
+    }
+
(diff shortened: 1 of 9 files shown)


pxteditor/shell.ts
```diff
@@ -60,4 +60,26 @@ namespace pxt.shell {
...
+    export function getEditorLanguagePref(): string {
+        return pxt.storage.getLocal("editorlangpref");
+    }
+
+    export function setEditorLanguagePref(lang: string): void {
+        if (lang.match(/prj$/)) lang = lang.replace(/prj$/, "")
+        pxt.storage.setLocal("editorlangpref", lang);
+    }
+
+    export function getToolboxAnimation(): string {
```
pelikhan, line 77: @shakao can we avoid using localStorage for this state? can it be in memory only?

mmoskal (approved):

pxteditor/shell.ts
shakao: we could move it! it would make the animation show up more frequently though. i guess the right thing to do would be to check the data and see if this is even helping people find the toolbox, and remove it if not... what's the downside of having it be in localStorage? are we mostly trying to reduce places where we access it?

PR #7473, opened 2020-09-23
Avoid Cross Site Scripting Vectors To/From the Simulator Iframe

aznhassan:
### Problem
The simulator is loaded in an iframe and a separate domain from the editor/page that hosts it. We communicate with the simulator via the _postMessage()_ command, and receive messages via a message event listener. When we post messages to the simulator iframe, we currently pass the _*_ wild card as the _targetOrigin_. When we receive messages from the simulator iframe, we don't check that the origin matches the domain of the simulator iframe. 

### Solution
The SimulatorDriver is run in the parent window of the simulator iframe. It facilitates communication between the simulator iframe, the page that loaded the SimulatorDriver, and even it's parent window. The latter case comes into play when you visit a page like https://makecode.com/multi, which can load multiple editor pages with simulators that will be able to communicate with one another. The solution comes in a few parts:
- Include the origin of the parent window as a URL search parameter when loading in editor in either makecode.com/multi or makecode.com/streamer
- Check that the origin of messages sent to the editor/SimulatorDriver match either the _simUrl_ of the simulator iframe, the editor's origin, or the _parentOrigin_ if that is specified.
- Use the URL of the simulator, _simUrl_, when _postMessage_ is called on the simulator iframe
- Use the origin of the editor, or the _parentOrigin_, when broadcasting message up to the parent window
- Use the origin of the editor when broadcasting messages to dependent editors

### Tests
I did lots of manual testing to make sure that messages were still going back and forth as expected. 
- Check that broadcasting still works by using multi (I had to manually edit the multi.html to use my changes to include the parent origin in the editor's URI) to load two Arcade pages; one page includes a radio broadcaster and the other a radio listener.
- Did the same on the test as above on the streamer page with no issues.
- Check that broadcasting between an editor and a dependent editor works by opening a dependent editor in a new tab that received messages from the current tab (the inverse does not work even before my changes).
- Check that simulator to simulator communication still works by creating a single project with a radio broadcaster and radio listener. This will create two simulators, both changing their color based on the broadcasted values.
- Check that editor to simulator communication still works via taking a screenshot/recording the simulator in the share modal.
- I couldn't test my changes with the share page, but that doesn't listen to messages like the streamer or multi pages do. So, the editor running in the iframe should run normally. Once this change is in, I can update one of the targets, say maker.makecode.com, to use the new pxt version to confirm the share project page works as expected.

### Note
This is only half of the problem of cross scripting vectors. The next step would be to put the origin of the editor page as a search parameter in the URI of the simulator iframe. The simulator iframe would then check against that editor origin when it receives or sends messages to the editor. Should I create a separate bug to track this, or should I consider the bug unresolved?

Here is the link to uploaded MakeCode Arcade that I used to test my changes: https://arcade.makecode.com/app/39fe5a831cd7b837be2e61d6f2f65e9ce59a347c-f5305f6e21

Fixes https://github.com/microsoft/pxt-backend/issues/109

### DIFF
--- a/docs/multi.html
+++ b/docs/multi.html
@@ -78,7 +78,7 @@
             var selectRight = document.getElementById("selectright");
             var divider = document.getElementById("divider");
             var localhost = /localhost=1/.test(window.location.href)
-            var flags = "?nestededitorsim=1&editorlayout=ide&nosandbox=1";
+            var flags = "?nestededitorsim=1&editorlayout=ide&nosandbox=1&parentOrigin=" + window.location.origin;
             var ratio = .5;
             var dividerWidth = 14;
 
@@ -167,7 +167,7 @@
 
             setWidths();
             handleHash();
-        })();        
+        })();
     </script>
     <!-- @include tracking.html -->
 </body>
--- a/docs/multi.html
+++ b/docs/multi.html
@@ -78,7 +78,7 @@
             var selectRight = document.getElementById("selectright");
             var divider = document.getElementById("divider");
             var localhost = /localhost=1/.test(window.location.href)
-            var flags = "?nestededitorsim=1&editorlayout=ide&nosandbox=1&parentOrigin=" + window.location.origin;
+            var flags = "?nestededitorsim=1&editorlayout=ide&nosandbox=1&parentOrigin=" + encodeURIComponent(window.location.origin);
             var ratio = .5;
             var dividerWidth = 14;
 
--- a/pxtsim/embed.ts
+++ b/pxtsim/embed.ts
@@ -249,6 +249,7 @@ namespace pxsim {
 
         function receiveMessage(event: MessageEvent) {
             let origin = event.origin; // || (<any>event).originalEvent.origin;
+            console.log(`[Embed] Origin: ${event.origin}, Data: ${event.data}, location: ${location}`)
             // TODO: test origins
             let data: SimulatorMessage = event.data || {};
             let type = data.type;
--- a/pxtsim/embed.ts
+++ b/pxtsim/embed.ts
@@ -249,7 +249,6 @@ namespace pxsim {
 
         function receiveMessage(event: MessageEvent) {
             let origin = event.origin; // || (<any>event).originalEvent.origin;
-            console.log(`[Embed] Origin: ${event.origin}, Data: ${event.data}, location: ${location}`)
             // TODO: test origins
             let data: SimulatorMessage = event.data || {};
             let type = data.type;
--- a/pxtsim/embed.ts
+++ b/pxtsim/embed.ts
@@ -250,6 +250,7 @@ namespace pxsim {
         function receiveMessage(event: MessageEvent) {
             let origin = event.origin; // || (<any>event).originalEvent.origin;
             // TODO: test origins
+
             let data: SimulatorMessage = event.data || {};
             let type = data.type;
             if (!type) return;
--- a/pxtsim/embed.ts
+++ b/pxtsim/embed.ts
@@ -250,7 +250,11 @@ namespace pxsim {
         function receiveMessage(event: MessageEvent) {
             let origin = event.origin; // || (<any>event).originalEvent.origin;
             // TODO: test origins
-
+            if (U.isLocalHost()) {
+                // no-op
+            } else {
+                if (!U.messageOriginExpected(origin, "https://userpxt.io")) return
+            }
             let data: SimulatorMessage = event.data || {};
             let type = data.type;
             if (!type) return;
--- a/pxtsim/embed.ts
+++ b/pxtsim/embed.ts
@@ -250,11 +250,6 @@ namespace pxsim {
         function receiveMessage(event: MessageEvent) {
             let origin = event.origin; // || (<any>event).originalEvent.origin;
             // TODO: test origins
-            if (U.isLocalHost()) {
-                // no-op
-            } else {
-                if (!U.messageOriginExpected(origin, "https://userpxt.io")) return
-            }
             let data: SimulatorMessage = event.data || {};
             let type = data.type;
             if (!type) return;
--- a/pxtsim/runtime.ts
+++ b/pxtsim/runtime.ts
@@ -193,15 +193,10 @@ namespace pxsim {
          */
         export function messageOriginExpected(origin: string, expectedOrigin: string, ignoreSubDomains: boolean = true): boolean {
             try {
-                console.log(`[RunTime] origin: ${origin}, expectedOrigin: ${expectedOrigin}`)
                 const originUrl = new URL(origin)
                 const expectedOriginUrl = new URL(expectedOrigin)
 
-                console.log(`[RunTime] Origin Protocol: ${originUrl.protocol}`)
-                console.log(`[RunTime] Expected Origin Protocol: ${expectedOriginUrl.protocol}`)
                 if (originUrl.protocol != expectedOriginUrl.protocol) return false
-                console.log(`[RunTime] Origin port: ${originUrl.port}`)
-                console.log(`[RunTime] Expected Origin port: ${expectedOriginUrl.port}`)
                 if (originUrl.port != expectedOriginUrl.port) return false
 
                 // Ignore the subdomains
@@ -211,8 +206,6 @@ namespace pxsim {
                     return originUrl.hostname === expectedOriginUrl.hostname
                 }
             } catch (error) {
-                // TODO: Consider logging an error here
-                console.log(`[RunTime] Exception while checking expected origin: ${error}`)
                 return false
             }
         }
@@ -909,7 +902,6 @@ namespace pxsim {
         // communication
         static messagePosted: (data: SimulatorMessage) => void;
         static postMessage(data: SimulatorMessage) {
-            console.log(`[Runtime] Post Message. Location ${window?.location.href}, Data: ${data}`)
             if (!data) return;
             // TODO: origins
             if (typeof window !== 'undefined' && window.parent && window.parent.postMessage) {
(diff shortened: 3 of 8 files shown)


docs/multi.html
```diff
@@ -78,7 +78,7 @@
             var selectRight = document.getElementById("selectright");
             var divider = document.getElementById("divider");
             var localhost = /localhost=1/.test(window.location.href)
-            var flags = "?nestededitorsim=1&editorlayout=ide&nosandbox=1";
+            var flags = "?nestededitorsim=1&editorlayout=ide&nosandbox=1&parentOrigin=" + window.location.origin;
```
pelikhan, line 81: encodeURIComponent?

pxtsim/embed.ts
```diff
@@ -250,7 +250,6 @@ namespace pxsim {
         function receiveMessage(event: MessageEvent) {
             let origin = event.origin; // || (<any>event).originalEvent.origin;
             // TODO: test origins
-
```
pelikhan, line 253: since this change does not help, best to remove it

pxtsim/runtime.ts
```diff
@@ -174,6 +174,41 @@ namespace pxsim {
...
+         * @param origin The origin of the received message
+         * @param expectedOrigin The expected origin of the received message
+         * @param ignoreSubDomains Whether the subdomains of the origins should be ignored
+         */
+        export function messageOriginExpected(origin: string, expectedOrigin: string, ignoreSubDomains: boolean = true): boolean {
+            try {
+                const originUrl = new URL(origin)
+                const expectedOriginUrl = new URL(expectedOrigin)
+
+                if (originUrl.protocol != expectedOriginUrl.protocol) return false
```
pelikhan, line 199: !==

pxtsim/simdriver.ts
```diff
@@ -601,6 +608,24 @@ namespace pxsim {
             if (!this.listener) {
                 this.listener = (ev: MessageEvent) => {
                     if (this.hwdbg) return
+
+                    if (U.isLocalHost()) {
+                        // no-op
+                    } else {
+                        let expectedOrigins: string[] =  [this.getSimUrl(), window.location.origin]
```
pelikhan, line 615: we are processing lots of messages when doing radio or serial messages, can we streamline this check? could this whole thing be condensed to a single hashmap lookup?

webapp/src/simulator.ts
```diff
@@ -45,6 +45,11 @@ export function init(root: HTMLElement, cfg: SimulatorConfig) {
     root.appendChild(debuggerDiv);
 
     const nestedEditorSim = /nestededitorsim=1/i.test(window.location.href);
+    let parentOrigin: string | null = null;
```
pelikhan, line 48: i don't think we use | null checking in this part of pxt yet.

aznhassan: Is this due to some limitation that I am currently unaware of, a style decision, or something else? Should I use `| undefined` instead? We currently use let variable: type | null in the `DebuggerVariablesState` interface, and in `monacoEditAmendments` in pxt/webapp. There are more instances of type | undefined used in pxt/webapp.

pelikhan:
Only some parts of the codebase are annotated; not that one I believe. Daryl has led this initiative.

webapp/src/simulator.ts
aznhassan: I poked Daryl and he was able to clarify. Appending `| null` to a type is only necessary when strict null checking is enabled. Otherwise, all types are implicitly unioned with `null` and `undefined`. Only the pxt/pxtpy enabled strict null checking. So I'll remove `| null` from the type.

webapp/public/multi.html
```diff
@@ -53,7 +53,7 @@
             var divider = document.getElementById("divider");
             var localhost = window.location.hostname == "localhost";
             var editor = (pxtConfig ? pxtConfig.relprefix : '/').replace(/-*$/, '');
-            var flags = "?nestededitorsim=1&editorlayout=ide&nosandbox=1";
+            var flags = "?nestededitorsim=1&editorlayout=ide&nosandbox=1&parentOrigin=" + window.location.origin;
```
darzu, line 56: I noticed that sometimes parentOrigin is a parameter and sometimes you just grab the location.origin

Is there a quick description for why these can differ?

darzu (approved):
Woo hoo!

webapp/public/multi.html
aznhassan: It's the `parentOrigin` from the perspective of the iframe, which is the location of the current window spawning said iframe. 

There are three origins that I care about in the editor. There is the origin of my parent window, my current origin, and the origins of any simulators the editor has spawned. 
#### Posting Messages
Since the `SimulatorDriver` can send broadcast messages to its parent window, it needs to know its parent origin. Sending messages to simulator iframes will always use the `simUrl`, and broadcasts to dependent editors will use the same origin as the current editor (working on this now). 
#### Receiving Messages
On the receiving end, messages can come from any of those three origins I mentioned before. In that case, the message just has to come from one of them to not be discarded.

Please let me know if this has answered your question.

pxtsim/simdriver.ts
aznhassan: Totally. I created `messageOriginExpected` back when I didn't know there was the `simUrl` variable. I can actually just create a set of the origins I expect in the SimulatorDriver constructor and check against that set when I receive messages. 

#### Note
It seems TypeScript doesn't support HashSets, then again this set will only create 3 items so the O(n) lookup cost is negligible when using a normal Set. This will be in my next commit.

docs/multi.html
aznhassan: I now encode the URI component.

webapp/src/simulator.ts
aznhassan: I've removed the `| null` from the variable declaration.

pxtsim/simdriver.ts
aznhassan: I now just check if the event origin is in the set of allowed origins.

pxtsim/runtime.ts
aznhassan: These changes no longer exist.

pxtsim/embed.ts
aznhassan: I've removed this change.

pxtsim/simdriver.ts
```diff
@@ -77,12 +77,22 @@ namespace pxsim {
         public state = SimulatorState.Unloaded;
         public hwdbg: HwDebugger;
         private _dependentEditors: Window[];
+        private _allowedOrigins: Set<string> = new Set();
```
pelikhan, line 80: is Set supported in all our browser matrix? There are not many origins, you can use a string[] and do linear lookups.

pxtsim/simdriver.ts
```diff
@@ -77,12 +77,22 @@ namespace pxsim {
...
 
         // we might "loan" a simulator when the user is recording
         // screenshots for sharing
         private loanedSimulator: HTMLDivElement;
 
         constructor(public container: HTMLElement, public options: SimulatorDriverOptions = {}) {
+            this._allowedOrigins.add(window.location.origin);
+            if (options.parentOrigin)
+            this._allowedOrigins.add(options.parentOrigin)
```
pelikhan, line 89: formatting

pxtsim/simdriver.ts
```diff
@@ -77,12 +77,22 @@ namespace pxsim {
...
+            try {
+                const simUrl = new URL(this.getSimUrl())
+                this._allowedOrigins.add(simUrl.origin)
+            } catch (e) {
+                console.error(`Invalid sim url ${this.getSimUrl()}`)
```
pelikhan, line 94: use ``pxt.reportException``

pxtsim/simdriver.ts
```diff
@@ -255,19 +270,23 @@ namespace pxsim {
...
                 // the editor is hosted in a multi-editor setting
                 // don't start extra frames
                 const parentWindow = window.parent && window.parent !== window.window
                     ? window.parent : window.opener;
                 if (this.options.nestedEditorSim && parentWindow) {
                     // if message comes from parent already, don't echo
-                    if (source !== parentWindow)
-                        parentWindow.postMessage(msg, "*");
+                    if (source !== parentWindow) {
+                        const parentOrigin = this.options.parentOrigin ? this.options.parentOrigin : window.location.origin
```
pelikhan, line 282: nit: ``this.options.parentOrigin || window.location.origin``

pxtsim/simdriver.ts
aznhassan: Where can I find our browser matrix? Looking at [the Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) I see that the Set constructor is supported by browsers. `add()` is supported by all the browsers listed there, Internet Explorer will not return the Set object after add is called. This is fine, since I'm not chaining calls to together.

pxtsim/simdriver.ts
aznhassan: I've made the change, but I had to include `/// <reference path="../built/pxtlib.d.ts"/>` if that is something to consider.

aznhassan: Okay, I was told that I should not include `pxtlib` in `pxtsim` due to how much that will increase the file size of `pxtsim`. Without referencing the lib, the `reportException` function is not available to me. I'll reverse my change to just use the console for the time being.

It seems there is work to introduce analytics to the sim in a more lightweight way than including all of pxtlib. Thoughts @pelikhan?

pxtsim/simdriver.ts
aznhassan: I changed it to use a string[] as suggested.

pelikhan:
Right we don’t need typescript and such in the simulator so pxsim is standalone.
I remember Guillaume did work on analytics there but that can be tackled in the future.

aznhassan:
I'll be completing this pull request now. I've created a new bug to track the work needed to make sure the simulator communicates with the editor appropriately. microsoft/pxt-backend#293.

PR #7477, opened 2020-09-24
Add gh action for extensions to check files exists

jwunderl:
And make a PR removing them from the list if they are invalid

close https://github.com/microsoft/pxt-minecraft/issues/1951

Very similar to the one I added for the minecraft content repo, e.g. of the generated pr for that: https://github.com/Mojang/EducationContent/pull/14

made a quick test repo: https://github.com/jwunderl/test-pxt-gh-action-repo/blob/master/.github/workflows/cfg-check.yml but I have that using a separate file because the checkpkgcfg isn't available without this pr

Could limit to just main/master branch, but it works for others too / doesn't seem like there's a reason to wait till the failure is fully merged?

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -3171,7 +3171,6 @@ export async function validateAndFixPkgConfig(parsed: commandParser.ParsedComman
         U.userError("Could not parse pxt.json");
     }
 
-
     const trimmedFiles = validateFileList("files", cfg.files);
     if (trimmedFiles) {
         cfg.files = trimmedFiles;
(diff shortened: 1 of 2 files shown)


cli/cli.ts
```diff
@@ -3163,6 +3163,64 @@ export function downloadPlaylistsAsync(parsed: commandParser.ParsedCommand): Pro
...
+    if (validFilesInFileDependencies) {
+        for (const key of Object.keys(cfg.fileDependencies)) {
+            if (validFilesInFileDependencies.indexOf(cfg.fileDependencies[key]) === -1) {
+                delete cfg.fileDependencies[key];
+            }
+        }
+    }
+
+    if (trimmedFiles || trimmedTestFiles || validFilesInFileDependencies) {
+        console.log("Updating pxt.json");
```
pelikhan, line 3195: use pxt.log, add logging before after

pelikhan (approved):

PR #7478, opened 2020-09-24
Don't request all files at once when building course book

jwunderl:
Right now, when building a course using `--docs#book`, we fetch all pages at once; when this is done in a language other than english, this means we need to fetch translations for all those files (which is ~~ 40-50+ md files per course). This will result in some of the requests for updating translations getting rejected, and the user getting out of date / possibly no translations for random entries

testing this is a bit tedious as localhost 403s on `/api/md/arcade/courses/...` when trying to build the book locally, so here's a link to a course in a build: https://arcade.makecode.com/app/5be2b998acd9bb6b02c82e20bef342be8d2acc7a-841de5f31f--docs#book:/courses/csintro1/SUMMARY It doesn't appear to be significantly slower than it currently is, as a significant portion of the time is spent rendering the page after fetching everything

worth a note, `pxtc.Util.promisePoolAsync` is a slight generalization of the implementation I added in pxt-translations to build translations in parallel.

### DIFF
--- a/pxtlib/util.ts
+++ b/pxtlib/util.ts
@@ -847,6 +847,28 @@ namespace ts.pxtc.Util {
         }
     }
 
+    export async function promisePoolAsync<T, V>(maxConcurrent: number, inputValues: T[], handler: (input: T) => Promise<V>) {
+        let curr = 0;
+        const promises = [];
+        const output: V[] = [];
+
+        for (let i = 0; i < maxConcurrent; i++) {
+            const thread = (async () => {
+                while (curr < inputValues.length) {
+                    const id = curr++;
+                    const input = inputValues[id];
+                    output[id] = await handler(input);
+                }
+            })();
+
+            promises.push(thread);
+        }
+
+        await Promise.all(promises);
+
+        return output;
+    }
+
     export function now(): number {
         return Date.now();
     }
--- a/pxtlib/util.ts
+++ b/pxtlib/util.ts
@@ -847,7 +847,7 @@ namespace ts.pxtc.Util {
         }
     }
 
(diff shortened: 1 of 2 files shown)


abchatra:
I am not sure I get the context. What bug is this fixing?

jwunderl:
All 50 files hit backend at once for translation, this is hitting the limit for concurrent requests whenever someone requests a book in a non english language.

pxtlib/util.ts
```diff
@@ -847,6 +847,28 @@ namespace ts.pxtc.Util {
...
+
+            promises.push(thread);
+        }
+
+        await Promise.all(promises);
+
+        return output;
+    }
+
     export function now(): number {
```
pelikhan, line 850: Don't we have this already?

pelikhan (approved):
Looks good. I think we already have U.now()

pxtlib/util.ts
jwunderl: Yeah `U.now` is existing code, not part of this change ~

abchatra:
Is this a safe fix? Can we port to the stable channels for microbit which has the course?

jwunderl:
Should be, here's a microbit build of each stable branch + this change: https://makecode.microbit.org/app/a17f67f1d8f68ee4157385d8f1e26987abebbcb5-c3a52af5a4

and the first csintro course: https://makecode.microbit.org/app/a17f67f1d8f68ee4157385d8f1e26987abebbcb5-c3a52af5a4--docs#book:/courses/csintro/SUMMARY

I'll put up a PR porting it ~

abchatra:
Thanks a ton @jwunderl

PR #7482, opened 2020-09-25
translate headers for monaco flyout

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/3394

image


shakao (approved):

abchatra:
Thanks! Looks good. Probably should port to microbit stable channel.

PR #7488, opened 2020-09-28
Added custom.ts

kharar:
Because I accidentally deleted mine and couldn't retrieve it so I spent half an hour to search the web until I found a Taiwanese translation that made it possible for me to fit the pieces back together.


pelikhan:
This looks good to me. We need to add a note in the sources to keep the docs in sync. @abchatra

ganicke (approved):

abchatra (approved):
Looks good to me as well.

PR #7490, opened 2020-09-29
Rename 'unplugged' -> 'showdialog' and 'fullscreen' -> 'showhint'

shakao:
Fixes https://github.com/microsoft/pxt-minecraft/issues/2017

- `@unplugged` -> `@showdialog`
- `@fullscreen` -> `@showhint`

Also renaming it in the code to avoid confusion. Old flags will still work as before, just improving the naming for future content creators.

### DIFF
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -912,9 +912,11 @@ declare namespace pxt.tutorial {
     }
 
     interface TutorialStepInfo {
-        fullscreen?: boolean;
-        // no coding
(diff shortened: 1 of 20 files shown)


riknoll (approved):
much better! Is the focus change intentional?

shakao:
oops yeah forgot to add in my commit message. additional tiny fix, noticed while testing this that we were calling focus on the text editor in blocks which was throwing a console error in each tutorial step

ganicke (approved):

jwunderl (approved):

PR #7494, opened 2020-10-01
Removing old cookie banner

riknoll:
All of our cookies are "essential" so we no longer need to show the cookie banner. Here's a test build:

https://arcade.makecode.com/app/046fc70ff689f31065a793dca1af3521006fafbc-b1aacb9a6c


riknoll:
@abchatra I've made the discussed changes and the link above is updated. Here's a summary of the behavior:

1. cookie is dropped immediately on production sites and electron
1. cookie is not dropped and cookie=false if staticpkg, localhost, sandbox iframe, or if pxtarget skipAnalyticsCookie is set
1. cookie=false on production sites and electron until the user does something that constitutes interactive consent (but cookie is already dropped) and is true afterwards

abchatra (approved):

PR #7495, opened 2020-10-02
add example of using handlerStatement to playground

jwunderl:
Add example to the playground using handlerStatement to make an event you can put inline:

image

I also made it so each of the examples calls the handlers; something similar to that will need to show up in their block anyways, and someone who's new might not understand how to get the code to actually run / treating a function as a parameter might be non-obvious

This issue did stand out a bit https://github.com/microsoft/pxt-microbit/issues/3303 as it causes the blocks in the toolbox to be reversed from their definitions, which feels odd. I can take a look at what's going on unless we know where / why that happened?

### DIFF
--- a/docs/static/playground/samples/language/events/sample.js
+++ b/docs/static/playground/samples/language/events/sample.js
@@ -4,11 +4,11 @@
 //% color="#AA278D"
 namespace language {
     /**
-     * A simple event taking an function handler
+     * A simple event taking a function handler
      */
     //% block="on event"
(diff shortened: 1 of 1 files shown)


shakao (approved):

shakao:
hmm yeah, first thought is something to do with the default ordering changed during the blockly upgrade, but that's mostly an educated guess! i don't think it's super urgent since the bug isn't marked next-release and weighting the blocks should still work to override default ordering

PR #7496, opened 2020-10-06
Beginnings of new asset editor work

riknoll:
PR Includes:

* Generalizes the tilemap project to include all asset types
* Uniting the tilemap/sprite field editors (and animation in future)
* Adding a "my assets" tab to the image editor and a name field for naming them

This is definitely a work in progress, but it includes most of the "all at once" refactoring so we should hopefully be able to work incrementally on top of it. Ideally i would have split out the UI changes but I needed them to test the other stuff.


webapp/src/components/ImageEditor/store/imageReducer.ts
```diff
@@ -500,6 +512,12 @@ const editorReducer = (state: EditorState, action: any, store: EditorStore): Edi
                 ...state,
                 resizeDisabled: true
             }
+        case actions.CHANGE_ASSET_NAME:
+            tickEvent("change-asset-name");
```
jwunderl, line 516: sidenote: maybe worth including the type of asset as another field in the `tickEvent`'s data, as it could be interesting to see which types of assets get named most often. Looks like `setTelemetryFunction` is already set up to allow for that anyways

webapp/src/components/ImageFieldEditor.tsx
```diff
@@ -11,12 +11,12 @@ export interface ImageFieldEditorProps {
 }
 
 export interface ImageFieldEditorState {
-    galleryVisible: boolean;
+    galleryState: "editor" | "gallery" | "my-assets";
```
jwunderl, line 14: minor nit: `currentView` maybe? `galleryState` feels like it would be giving info about the current gallery to me, not the current state of the editor

webapp/src/package.ts
```diff
@@ -484,11 +484,16 @@ export class EditorPackage {
         return this.filterFiles(f => f.getName() == name)[0]
     }
```
jwunderl, line 486: will want to count generated image files as auto generated above https://github.com/microsoft/pxt/blob/3611c3b7f06cf878561dc61fc38ec86344fb6851/webapp/src/package.ts#L144

jwunderl (approved):
LGTM to merge / iterate on; just one api caching thing and the rest are nits / fine to ignore / handle later

Did notice some bugs when renaming tiles (but that seems fine to fix in a separate pr) -- it's allowing duplicate names for tiles, adding extra tiles after closing / reopening, and sometimes losing some tiles altogether

renaming tile causes duplicate

webapp/src/components/ImageEditor/BottomBar.tsx
```diff
@@ -205,6 +235,34 @@ export class BottomBarImpl extends React.Component<BottomBarProps, BottomBarStat
         }
     }
 
+    protected handleAssetNameChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+        let errorMessage = null;
+
+        if (!validateAssetName(event.target.value)) {
+            errorMessage = lf("Names must begin with a letter and can only contain letters, numbers, and _");
```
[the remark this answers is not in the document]
riknoll: yeah, I'm gonna leave it in for now but should be much more lenient in the future

webapp/src/components/ImageEditor/store/imageReducer.ts
riknoll: I think I'm actually going to remove this tick.... I'll add a different tick elsewhere and think about how to add that info to the general function.

PR #7497, opened 2020-10-06
Tick event for 'exiting' a recipe

shakao:
Current approach is that "finishing" a recipe is the same as loading a new recipe into a project that already has a recipe open. When users are done with the entire chain of recipes, they would hit the "Finish" button which would be a different tick that represents being done with all the recipes in that progression path.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -3487,7 +3487,7 @@ export class ProjectView
         // custom tick for recipe "completion". recipes use links in the markdown to
         // progress, so we track when a user "exits" a recipe by loading a new one
         if (this.state.header?.tutorial?.tutorialRecipe) {
-            pxt.tickEvent("recipe.exit", { id: this.state.header?.tutorial?.tutorial, goto: tutorialId });
+            pxt.tickEvent("recipe.exit", { tutorial: this.state.header?.tutorial?.tutorial, goto: tutorialId });
         }
 
         core.hideDialog();
@@ -3545,7 +3545,7 @@ export class ProjectView
     }
 
     completeTutorialAsync(): Promise<void> {
-        pxt.tickEvent("tutorial.complete");
+        pxt.tickEvent("tutorial.complete", { tutorial: this.state.header?.tutorial?.tutorial });
         core.showLoading("leavingtutorial", lf("leaving tutorial..."));
 
         // clear tutorial field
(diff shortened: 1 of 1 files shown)


Jaqster:
Awesome, thanks so much for looking into this Shannon!  What happens if they just click Ok?  So they finish that chapter, but don't move on to the next recipe?  Will they get an exit event?
image

Jaqster:
Also, probably a separate issue, but when they do click Ok, I feel like they should automatically "Finish" the tutorial and be taken to the editor.  Instead of staying in the last step with a blank instruction pane.  Eventually, we should think about how to allow a user to take their project and jump back into the recipe flow later... maybe part of the Skills Map brainstorm...?
image

shakao:
no, if they click "Ok" they are still in the tutorial (could theoretically be going back to look at previous steps, or messing with their existing code, so they won't get an exit even. if they click "Ok" then afterwards click "Finish", they'll get a tutorial.complete event.

and yeah, definitely agreed! i remember david came up with a nicer design for the final recipe modal ("Ok" was "Finish", and there were buttons under each GIF, i think?), it would be great to have custom UI for it instead of just reusing our markdown rendering, which is a little limiting.

Jaqster:
@soniakandah - another little project for you to think about in your spare time...
@shakao - if this is true, I should see a tutorial.complete event for /recipes/shark-splash/01-character, but that's not the case

abchatra (approved):

shakao:
@Jaqster ah, looks like there was a bug with the tutorial.complete event--i'm not seeing any completion events for chase the pizza or galga either. i added the tutorial name to those ticks as well!

Jaqster:
Oh dear.  How long has this bug been messing things up?  Was it not firing at all?  Or just not associated with a tutorial?  Wondering if I need to throw out our Arcade completion rates entirely...?  Does it affect other targets?

shakao:
the event was firing at the right time, so the existing data is still correct! it just wasn't associated with a specific tutorial. there might be some query with session id and the tutorial.start event ("this user started galga and sometime later in the session a tutorial.complete event was fired") we could use to estimate that information as well

webapp/src/app.tsx
```diff
@@ -3539,7 +3545,7 @@ export class ProjectView
     }
 
     completeTutorialAsync(): Promise<void> {
-        pxt.tickEvent("tutorial.complete");
+        pxt.tickEvent("tutorial.complete", { tutorial: this.state.header?.tutorial?.tutorial });
```
abchatra, line 3548: @shakao can we add this tutorial custom dimension to all the following ticks? 

tutorial.step
tutorial.start
tutorial.showhint
tutorial.shared
tutorial.previous
tutorial.next
tutorial.hint.next
tutorial.faileddecompile
tutorial.exit
tutorial.complete
menu.exittutorial
hash.tutorial

abchatra (approved):

PR #7498, opened 2020-10-06
don't log spurious sim url error on localhost

jwunderl:
This currently logs an error message on localhost as `new URL("/sim/simulator.html")` is invalid. No functional difference, just avoids an unnecessary error 

I suppose it could also just return before adding any of the _allowedOrigins on localhost since they are ignored here anyways https://github.com/microsoft/pxt/blob/981db6465023458f99506bf73be4cb57b8013c40/pxtsim/simdriver.ts#L626-L630 but that feels like a bug waiting to happen if someone wants to add something to the simdriver constructor in the future

edit: I suppose this also closes https://github.com/microsoft/pxt-microbit/issues/3410 now ~

### DIFF
--- a/pxtsim/simdriver.ts
+++ b/pxtsim/simdriver.ts
@@ -92,7 +92,9 @@ namespace pxsim {
                 const simUrl = new URL(this.getSimUrl())
                 this._allowedOrigins.push(simUrl.origin)
             } catch (e) {
-                console.error(`Invalid sim url ${this.getSimUrl()}`)
+                if (!U.isLocalHost()) {
+                    console.error(`Invalid sim url ${this.getSimUrl()}`)
+                }
             }
         }
 
--- a/pxtsim/simdriver.ts
+++ b/pxtsim/simdriver.ts
@@ -92,9 +92,7 @@ namespace pxsim {
                 const simUrl = new URL(this.getSimUrl())
                 this._allowedOrigins.push(simUrl.origin)
             } catch (e) {
-                if (!U.isLocalHost()) {
-                    console.error(`Invalid sim url ${this.getSimUrl()}`)
-                }
+                console.error(`Invalid sim url ${this.getSimUrl()}`)
             }
         }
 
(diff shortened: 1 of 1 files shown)


pxtsim/simdriver.ts
```diff
@@ -92,7 +92,9 @@ namespace pxsim {
                 const simUrl = new URL(this.getSimUrl())
                 this._allowedOrigins.push(simUrl.origin)
             } catch (e) {
-                console.error(`Invalid sim url ${this.getSimUrl()}`)
+                if (!U.isLocalHost()) {
```
aznhassan, line 95: This looks fine to me. We really only want to log an error when the sim URL is invalid when we're not running on localhost. If the sim URL were invalid in that case, then the simulator iframe would not load correctly and this error message would make it clear that the malformed URL is the culprit. 

Something we can consider is changing
```ts
private getSimUrl(): string {
    return this.options.simUrl || ((window as any).pxtConfig || {}).simUrl || "/sim/simulator.html"
}
```
to
```ts
private getSimUrl(): string {
    return this.options.simUrl || ((window as any).pxtConfig || {}).simUrl || location.origin + "/sim/simulator.html"
}
```
This way we'd have a valid URL even when we're on localhost. When we're creating the simulator iframe:
```ts
frame.src = this.getSimUrl() + '#' + frame.id;
```
in the 
```ts
private createFrame(light?: boolean): HTMLDivElement {
```
function, setting the frame src to "/sim/simulator.html" + '#' + frame.id will actually include the origin of the window creating the frame (localhost in this case). So including the origin ourselves in the localhost case means we don't have to check for local host before logging an error about an invalid URL. It's probably what I should have done when I first created this PR. 

Either way, looks good to me.

aznhassan (approved):
LGTM. I've left some thoughts about an alternative change in a comment.

pxtsim/simdriver.ts
jwunderl: Good call, that sounds better to me too~

PR #7500, opened 2020-10-07
Don't infinite load when on error in loading path

jwunderl:
If we hit an error while loading in the DOMContentLoaded event, right now we just infinite load in the home screen. These aren't necessarily recoverable -- e.g. the case this last came up in was a crash in the targets editor/extension.ts's `initExtensionsAsync` (https://github.com/microsoft/pxt-minecraft/pull/2031), and a failure there will mean editor specific behavior didn't get loaded / something is very wrong.

It uses the same one as when the main app fails to load, which pops this message and tries to refresh once in case the error was just random.

image

A few small fixes along the way, I'll add a few comments pointing at them

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -865,6 +865,10 @@ export class ProjectView
 
     // Add an error guard for the entire application
     componentDidCatch(error: any, info: any) {
+        this.handleCriticalError(error, info);
+    }
+
+    handleCriticalError(error: any, info?: any) {
         try {
             core.killLoadingQueue();
             pxsim.U.remove(document.getElementById('loading'));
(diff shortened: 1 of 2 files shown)


webapp/src/app.tsx
```diff
@@ -865,17 +865,21 @@ export class ProjectView
...
         try {
             core.killLoadingQueue();
             pxsim.U.remove(document.getElementById('loading'));
             this.setState({ hasError: true });
             // Log critical error
             pxt.tickEvent('pxt.criticalerror', { error: error || '', info: info || '' });
             // Reload the page in 2 seconds
-            const lastCriticalError = pxt.storage.getLocal("lastcriticalerror") ?
-                Date.parse(pxt.storage.getLocal("lastcriticalerror")) : Date.now();
+            const lastCriticalError = pxt.storage.getLocal("lastcriticalerror") &&
```
jwunderl, line 879: I don't think this code ever worked / it never actually reloaded the editor. If this wasn't set, it defaults to date.now, which will trivially be within the last 60 seconds

webapp/src/app.tsx
```diff
@@ -865,17 +865,21 @@ export class ProjectView
...
+                Date.parse(pxt.storage.getLocal("lastcriticalerror"));
             // don't refresh if we refreshed in the last minute
-            if (!lastCriticalError || (!isNaN(lastCriticalError) && Date.now() - lastCriticalError > 60 * 1000)) {
+            if (!lastCriticalError || isNaN(lastCriticalError) || Date.now() - lastCriticalError > 60 * 1000) {
```
jwunderl, line 882: If it's nan then it's invalid / should continue and try refreshing

webapp/src/blocks.tsx
```diff
@@ -481,6 +481,8 @@ export class Editor extends toolboxeditor.ToolboxEditor {
     private prepareBlockly(forceHasCategories?: boolean) {
         pxt.perf.measureStart("prepareBlockly")
         let blocklyDiv = document.getElementById('blocksEditor');
+        if (!blocklyDiv)
```
jwunderl, line 484: This was adding an extra crash because blocksEditor doesn't exist if we crash, and this happens in a separate promise thread. Can do something different if we want here?

jwunderl (commented):
minor bug notes ~

livcheerful (approved):

PR #7502, opened 2020-10-08
fixes for error reporting on block localizations

jwunderl:
Right now we're getting a ton of errors from these that aren't easily investigated, as it's sending each block id as a separate property in the object, and sending the error in a lot of cases where we actually fix the block. 

two fixes for the errors:

* Send a tick event instead of an error when `normalizeBlock` has modified a block -- this isn't really an error case as we are theoretically fixing these with the replaces, so if it parses properly through the `hasEquivalentParameters` check below then in theory the block is fine / just trivia differences

* changing the shape of the error reporting - I believe these should be easier to investigate, as we can get all the info we need to find the errors from the target, version, & language id as that's enough to load in and view the errors in the console (I also included the broken block id as a pipe delimited list)

Also drops the logic from cli/crowdin; this wasn't actually changing the blocks, and these will get the ticks sent whenever users load them. Can add them back in with the tickevent behavior in pxtlib/service.ts if we still want to send those tickevents when packaging the app instead of on use.

also a few minor fixes to block normalization logic, will add notes explaining inline

### DIFF
--- a/pxtlib/blocks.ts
+++ b/pxtlib/blocks.ts
@@ -82,15 +82,15 @@ namespace pxt.blocks {
         if (!b) return b;
         // normalize and validate common errors
         // made while translating
-        let nb = b.replace(/[^\\]%\s+/g, '%');
+        let nb = b.replace(/(?:^|[^\\])([%$])\s+/g, '$1');
         if (nb != b) {
             err(`block has extra spaces: ${b}`);
-            return b;
+            b = nb;
         }
 
         // remove spaces around %foo = ==> %foo=
         b = nb;
-        nb = b.replace(/(%\w+)\s*=\s*(\w+)/, '$1=$2');
+        nb = b.replace(/([%$]\w+)\s*=\s*(\w+)/, '$1=$2');
         if (nb != b) {
             err(`block has space between %name and = : ${b}`)
             b = nb;
--- a/pxtlib/service.ts
+++ b/pxtlib/service.ts
@@ -610,7 +610,7 @@ namespace ts.pxtc {
         const lang = pxtc.Util.userLanguage();
         if (pxtc.Util.userLanguage() == "en") return Promise.resolve(cleanLocalizations(apis));
 
-        const errors: pxt.Map<number> = {};
+        const errors: string[] = [];
         const langLower = lang.toLowerCase();
         const attrJsLocsKey = langLower + "|jsdoc";
         const attrBlockLocsKey = langLower + "|block";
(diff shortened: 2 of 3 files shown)


pxtlib/blocks.ts
```diff
@@ -82,15 +82,15 @@ namespace pxt.blocks {
         if (!b) return b;
         // normalize and validate common errors
         // made while translating
-        let nb = b.replace(/[^\\]%\s+/g, '%');
+        let nb = b.replace(/(?:^|[^\\])([%$])\s+/g, '$1');
```
jwunderl, line 85: two fixes; make this also check blocks using the $ format, and add a check to also fix blocks that start with a parameter

pxtlib/blocks.ts
```diff
@@ -82,15 +82,15 @@ namespace pxt.blocks {
...
         if (nb != b) {
             err(`block has extra spaces: ${b}`);
-            return b;
+            b = nb;
```
jwunderl, line 88: also want to handle improper spacing between default values, don't want to bail out early

pxtlib/blocks.ts
```diff
@@ -82,15 +82,15 @@ namespace pxt.blocks {
...
         }
 
         // remove spaces around %foo = ==> %foo=
         b = nb;
-        nb = b.replace(/(%\w+)\s*=\s*(\w+)/, '$1=$2');
+        nb = b.replace(/([%$]\w+)\s*=\s*(\w+)/, '$1=$2');
         if (nb != b) {
             err(`block has space between %name and = : ${b}`)
```
jwunderl, line 95: I'm attaching the error in the tickevent we send, so it makes sense to send both to me if it hits both spacing errors

pxtlib/service.ts
```diff
@@ -680,8 +686,11 @@ namespace ts.pxtc {
             })))
             .then(() => cleanLocalizations(apis))
             .finally(() => {
-                if (Object.keys(errors).length)
-                    pxt.reportError(`loc.errors`, `invalid translation`, errors);
+                if (errors.length)
+                    pxt.reportError(`loc.errors`, `invalid translations`, {
+                        blockIds: errors.join("|"),
```
abchatra, line 691: Instead of this, send multiple reportError with each reportErro to have a single blockId. You can do this in a loop. That will be way easier to index in _by_ query.

abchatra (approved):

abchatra:
Awesome :-)

PR #7503, opened 2020-10-08
Base identity platform - initial add

darzu:
Adds a identity experiment which enables a new "login" button on the homepage which pops a dialog with a few login options.


eanders-MS (approved):
LGTM, but does this need to be merged to master? Seems early.

pxtlib/emitter/cloud.ts
```diff
@@ -2,8 +2,11 @@
 namespace pxt.Cloud {
     import Util = pxtc.Util;
 
+    const baseEndpoint = "https://www.makecode.com";
+    //const baseEndpoint = "http://localhost:5500";
```
darzu, line 6: Actually, this might not work with staging.pxt.io
I'll need to rethink this

darzu:
> LGTM, but does this need to be merged to master? Seems early.

Merging into master is easiest since arcade staging will point to that naturally. Also that way I'm not slowing down others that want to test things in staging.

The thought is that since it'll be behind an experiment it should be out of the way, and it gives us a hook to try other identity stuff.

pxtlib/emitter/cloud.ts
eanders-MS: Does the client app have a per-environment config file?

darzu: @eanders-MS , can you think of any reason why we wouldn't use "window.location.origin" here ?

eanders-MS: That should work, but might need to special case it for localhost to fixup the port number.

webapp/src/identity.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import Cloud = pxt.Cloud;
+import Util = pxt.Util;
+
+export function showLoginDialogAsync(projectView: pxt.editor.IProjectView) {
+    const targetTheme = pxt.appTarget.appTheme;
+
+    const searchParams = new URLSearchParams(window.location.search);
+    const experimentalIdentityFlagEnabled = !!searchParams.get("experimentalIdentity")
+    if (!experimentalIdentityFlagEnabled) {
```
darzu, line 16: extra query param flag makes it harder to stumble upon

pxtlib/emitter/cloud.ts
darzu: I decided against this change since "baseEndpoint" is only used when running in node or localhost, otherwise we use site relative "/api" as the api root which is effectively like using window.location.origin

darzu:
Most of this was reviewed here: https://github.com/microsoft/pxt/pull/7506#event-3868305713

darzu:
@eanders-MS , think this is ready for merge?

eanders-MS:
@darzu, we should hold off merging until we have privacy & security approval.

eanders-MS:
@darzu, there's also the CORS thing. I think I have a solution for this now.

eanders-MS:
@darzu, I also want to get session tokens in place before this gets merged to master.

abchatra (approved):

eanders-MS (approved):
:shipit:

PR #7505, opened 2020-10-09
fixes for idb workspaces

jwunderl:
re: https://github.com/microsoft/pxt-minecraft/issues/1936
closes https://github.com/microsoft/pxt-minecraft/issues/1893

Handful of fixes for idb workspaces (I awaitified it in the first commit as I was reading through all the code to make sure I had context on everything).

* If we fail to write in idb workspace, try again after clearing translation / tutorial info caches to clear up some room. This behavior is largely the same as https://github.com/microsoft/pxt/blob/76381d21d3f3574e7bd1bee1459e80af2dea9690/webapp/src/db.ts#L84
* If writing still fails, we're out of space / something is wrong and we can't continue to store stuff. bubble up the error to workspace `saveAsync`, and transition to using the in memory db so it at least works for the session / e.g. you can at least load tutorials still. (this is the fix for 1936 above).
* Fix race condition in getDbAsync -- This was immediately setting `_db`, and then going on to open it in a promise; if something else tried called `getDbAsync` it would get immediate access, potentially before `openAsync` finished. Worth a note it looks like we have a similar problem in db.ts (though I don't think this one breaks anything really): https://github.com/microsoft/pxt/blob/76381d21d3f3574e7bd1bee1459e80af2dea9690/webapp/src/db.ts#L43
This one defers setting it at until after the test get is complete, so it's possible multiple pouchdbs will get instantiated simultaneously / could probably be better of storing memoryDb if there's an error here. Can look at that more, but didn't want to make this PR any bigger right away.

I think it's worth discussing the in memory db behavior in a design meeting; feels like there needs to be some way we can convey that they need to make other plans to persist their projects -- at least something like an icon maybe?

### DIFF
--- a/webapp/src/idbworkspace.ts
+++ b/webapp/src/idbworkspace.ts
@@ -22,39 +22,32 @@ let _db: pxt.BrowserUtils.IDBWrapper;
 
 // This function migrates existing projectes in pouchDb to indexDb
 // From browserworkspace to idbworkspace
-function migrateBrowserWorkspaceAsync(): Promise<void> {
-    return getDbAsync()
-        .then((db) => {
-            return db.getAllAsync<pxt.workspace.Header>(HEADERS_TABLE);
-        })
-        .then((allDbHeaders) => {
-                if (allDbHeaders.length) {
-                    // There are already scripts using the idbworkspace, so a migration has already happened
-                    return Promise.resolve();
-                }
+async function migrateBrowserWorkspaceAsync(): Promise<void> {
+    const db = await getDbAsync();
+    const allDbHeaders = await db.getAllAsync<pxt.workspace.Header>(HEADERS_TABLE);
+    if (allDbHeaders.length) {
+        // There are already scripts using the idbworkspace, so a migration has already happened
+        return;
+    }
 
-                const copyProject = (h: pxt.workspace.Header): Promise<string> => {
-                    return browserworkspace.provider.getAsync(h)
-                        .then((resp) => {
-                            // Ignore metadata of the previous script so they get re-generated for the new copy
-                            delete (<any>h)._id;
-                            delete (<any>h)._rev;
-                            return setAsync(h, undefined, resp.text);
-                        })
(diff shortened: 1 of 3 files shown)


webapp/src/idbworkspace.ts
```diff
@@ -18,116 +18,134 @@ const TEXTS_TABLE = "texts";
...
-                id: h.id,
-                files: text,
-                _rev: prevVer
-            };
-            return (text ? db.setAsync(TEXTS_TABLE, dataToStore) : Promise.resolve())
-                .then(() => {
-                    return db.setAsync(HEADERS_TABLE, h);
-                });
-        });
+async function setAsync(h: Header, prevVer: any, text?: ScriptText): Promise<void> {
```
jwunderl, line 93: Worth a note on the return type change; the return type was void / nothing was actually getting returned by `db.setAsync` before so no actual difference, it was just passing as written because `pxt.workspace.Version` is `any` and that allows void returns

abchatra (approved):

pelikhan:
We should not fail silently. We need a banner that warns the user that all changes will be lost.

livcheerful (approved):

pelikhan (approved):

pelikhan:
We could use the notification bar (the one we use for the windows app) to tell the user that his changes are not stored in the browser. Could be done as a future PR>

jwunderl:
Oh, yes, @soniakandah @livcheerful @abchatra and I discussed this in a meeting; what we decided on was popping a confirmation dialogue when we switched to in mem db explaining that we were unable to save / suggesting to save their project by other means (probably not when in an iframe though? Wouldn't want that popping up in all embeds that don't have the right sandbox). That said I don't remember the notification bar at all so that might be good too, will take a look in a separate PR

abchatra:
Don't use notification bar. It has tons of bugs and messes up the CSS a lot. For now just do a quick pop up in non embed scenarios.

PR #7506, opened 2020-10-09
Client API for auth

eanders-MS:
This PR merges my and @darzu's work. It adds an api for authentication, and a nascent "virtual api" for user profile. This virtual api will serve the data source for React components that show user profile state.

### DIFF
--- a/pxtlib/util.ts
+++ b/pxtlib/util.ts
@@ -650,6 +650,7 @@ namespace ts.pxtc.Util {
         responseArrayBuffer?: boolean;
         forceLiveEndpoint?: boolean;
         successCodes?: number[];
+        withCredentials?: boolean;
     }
 
     export interface HttpResponse {
@@ -1434,6 +1435,8 @@ namespace ts.pxtc.BrowserImpl {
             client = new XMLHttpRequest();
             if (options.responseArrayBuffer)
                 client.responseType = "arraybuffer";
+            if (options.withCredentials)
+                client.withCredentials = true;
             client.onreadystatechange = () => {
                 if (resolved) return // Safari/iOS likes to call this thing more than once
 
--- a/pxtlib/util.ts
+++ b/pxtlib/util.ts
@@ -1429,6 +1429,8 @@ namespace ts.pxtc.BrowserImpl {
 
             let headers = Util.clone(options.headers) || {}
 
+            options.url = /^http/.test(options.url) ? options.url : pxt.Cloud.appRoot + options.url;
+
             client = new XMLHttpRequest();
             if (options.responseArrayBuffer)
                 client.responseType = "arraybuffer";
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -112,6 +112,7 @@ export class ProjectView
     scriptManagerDialog: scriptmanager.ScriptManagerDialog;
     importDialog: projects.ImportDialog;
     signInDialog: cloud.SignInDialog;
+    identityLoginDialog: identity.LoginDialog;
     exitAndSaveDialog: projects.ExitAndSaveDialog;
     newProjectDialog: projects.NewProjectDialog;
     chooseHwDialog: projects.ChooseHwDialog;
@@ -163,6 +164,7 @@ export class ProjectView
         this.toggleGreenScreen = this.toggleGreenScreen.bind(this);
         this.toggleSimulatorFullscreen = this.toggleSimulatorFullscreen.bind(this);
         this.cloudSignInComplete = this.cloudSignInComplete.bind(this);
+        this.identityLoginComplete = this.identityLoginComplete.bind(this);
         this.toggleSimulatorCollapse = this.toggleSimulatorCollapse.bind(this);
         this.showKeymap = this.showKeymap.bind(this);
         this.toggleKeymap = this.toggleKeymap.bind(this);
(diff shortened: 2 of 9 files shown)


webapp/src/app.tsx
```diff
@@ -4451,7 +4457,7 @@ document.addEventListener("DOMContentLoaded", () => {
     else if (isSandbox) workspace.setupWorkspace("mem");
     else if (pxt.winrt.isWinRT()) workspace.setupWorkspace("uwp");
     else if (pxt.BrowserUtils.isIpcRenderer()) workspace.setupWorkspace("idb");
-    else if (pxt.BrowserUtils.isLocalHost() || pxt.BrowserUtils.isPxtElectron()) workspace.setupWorkspace("fs");
+    else if (/*pxt.BrowserUtils.isLocalHost() ||*/ pxt.BrowserUtils.isPxtElectron()) workspace.setupWorkspace("fs");
```
eanders-MS, line 4460: I'm curious why workspace is set to "fs" for localhost. Is this safe to remove? I'm seeing a network error locally when this is the case. It tries to GET /api/list, which doesn't exist in the backend.

jwunderl: Projects are saved under the `projects` folder in the directory you run `pxt serve` in from with fs, instead of browser storage; very helpful for debugging shared scripts / etc

eanders-MS: Oh! I will revert this change.

pxtlib/util.ts
```diff
@@ -1429,6 +1429,8 @@ namespace ts.pxtc.BrowserImpl {
 
             let headers = Util.clone(options.headers) || {}
 
+            options.url = /^http/.test(options.url) ? options.url : pxt.Cloud.appRoot + options.url;
```
darzu, line 1432: When is this needed?

webapp/src/auth.ts
```diff
@@ -0,0 +1,218 @@
+import * as core from "./core";
+import * as data from "./data";
+
+import U = pxt.Util;
+
+/**
+ * Virtual API keys
```
darzu, line 7: What do you mean by "virtual api" in this file?

pxtlib/util.ts
eanders-MS: For routing to the configured backend when the url is relative (see my changes in cloud.ts -- the `BACKEND` field). I needed this so that calls to endpoints like `/auth/login` are able to hit the locally running backend.

webapp/src/auth.ts
eanders-MS: The "virtrual api" is the mechanism for injecting state into React components. It is defined in data.ts, here: https://github.com/microsoft/pxt/blob/541d9c867beaed9029aefb2cf65a3c4bcb1efaba/webapp/src/data.ts#L32

webapp/src/app.tsx
```diff
@@ -2018,6 +2020,9 @@ export class ProjectView
             }).done();
     }
 
+    identityLoginComplete() {
+    }
```
darzu, line 2024: TODO comment?

webapp/src/auth.ts
```diff
@@ -159,46 +155,92 @@ export function identityProviders(): pxt.AppCloudProvider[] {
...
-    const user: UserProfile = resp.json;
-
-    setUser(user);
-}
-
-function mkLoginStateKey(id: string): string {
-    return `auth.login.state:${id}`;
-}
+    // We already have a user, no need to get it again.
+    if (state.user) return;
```
darzu, line 208: What's the expiration on this?

webapp/src/auth.ts
```diff
@@ -159,46 +155,92 @@ export function identityProviders(): pxt.AppCloudProvider[] {
...
 function idpEnabled(idp: pxt.IdentityProviderId): boolean {
     return identityProviders().filter(prov => prov.id === idp).length > 0;
 }
 
-function setUser(user: UserProfile) {
+function setUser(user: UserProfile, invalidate: boolean = true) {
     state_.user = user;
-    data.invalidate(AUTH_USER);
+    if (invalidate)
+        data.invalidate(USER);
```
darzu: in what case wouldn't we invalidate?

webapp/src/auth.ts
```diff
@@ -159,46 +155,92 @@ export function identityProviders(): pxt.AppCloudProvider[] {
...
     return U.requestAsync({
-        url: "/api/" + path,
-        headers: { "Cache-Control": "no-store" },
+        url: path,
+        headers: { "Cache-Control": "no-store" },   // don't cache responses
         method: data ? "POST" : "GET",
-        data: data || undefined
-    }).then(r => r.json).catch(core.handleNetworkError);
+        data: data || undefined,
+        withCredentials: true,  // include access token in request
```
darzu, line 245: Don't we have APIs that don't require credentials?

eanders-MS: This just instructs the call to include cookies in cross-site requests (with matching origin, hence my need to change the CORS policy on the backend -- where I made that "CONFLICT" comment). Without this, the x-pxt-access-token cookie doesn't get sent.

webapp/src/auth.ts
eanders-MS: On `state.user`? It lives in memory, never stored locally.

webapp/src/app.tsx
eanders-MS: This is unused. Will remove.

eanders-MS: I take that back. We will need it later. Will add a TODO comment.

darzu (approved):
LGTM!

PR #7508, opened 2020-10-09
Additional telemetry data for various tutorial ticks

shakao:
- `tutorial.start` will have `{ activity: ["tutorial"|"recipe"|"example"], path: "/tutorials/chase-the-pizza", editor: ["blocksprj"|"tsprj"|"pyprj"] }`
- `tutorial.shared` has the script id under "tutorial"
- `tutorial.faileddecompile` (not modified in this PR) has the tutorial name/path under "tutorialId"

@abchatra do we want the "tutorial id" dimension to have the same key in all of our ticks? not sure if it matters

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1439,7 +1439,7 @@ export class ProjectView
             })
             .catch(e => {
                 // Failed to decompile
-                pxt.tickEvent('tutorial.faileddecompile', { tutorialId: t.tutorial });
+                pxt.tickEvent('tutorial.faileddecompile', { tutorial: t.tutorial });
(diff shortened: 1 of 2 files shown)


abchatra:
It is easier to query something like following if they are under the same cusotmdimension such as tutorialID.  You can do it otherwise too. But will have to add bunch or _or clauses_
```
where  customdimensions.tutorialid == "/tutorials/chase-the-pizza"
summarize count() by name
```

abchatra (approved):

shakao:
if the tick has already existed for a while as "tutorialid", is it better change it to "tutorial" now to unify, or leave it?

abchatra:
Change it. We are the only one who query for this and we can change our queries.

abchatra (approved):

PR #7509, opened 2020-10-12
don't broadcast random messages to iframes + dev token fix

pelikhan:
- [x] when receiving messages, make sure to only broadcast message that have .broadcast = true. Not every messages like react debugging stuff...
- [x] dev token lookup is broken due to "remember me " feature. We used to lookup the first input in the dialog.


pelikhan:
@abchatra  i'll need to port the devtoken fix in the microbit/arcade branch

PR #7513, opened 2020-10-14
micro:bit play/ring tone fixes

jwunderl:
closes https://github.com/microsoft/pxt-microbit/issues/3371
re: https://github.com/microsoft/pxt-microbit/issues/3400

* sim: 0's a valid frequency, corresponding to no sound (commonly used to clear out a current sound); do not ignore it
* hw/sim: Volume should apply to active playTone / ringTone; add a method to apply the volume (hardware side needs microbit pr). 

Worth a note, the mixer package (e.g. arcade) already handles this https://makecode.com/_1adXPC4ivao2 (though it does have an annoying recurring pop each beat), so this fix is just for microbit


mmoskal (approved):

pxtsim/simlib.ts
```diff
@@ -249,12 +249,27 @@ namespace pxsim {
...
         }
 
         export function stop() {
             stopTone();
+            clearVca();
+        }
+
+        function clearVco() {
+            if (_vco) {
+                _vco.stop();
```
pelikhan, line 263: i would try catch around those.

pxtsim/simlib.ts
```diff
@@ -249,12 +249,27 @@ namespace pxsim {
...
+                _vco.disconnect();
+                _vco = undefined;
+            }
+        }
+        function clearVca() {
+            if (_vca) {
+                _vca.disconnect();
```
pelikhan, line 270: defensive try catch here too

pxtsim/simlib.ts
```diff
@@ -587,6 +598,12 @@ namespace pxsim {
             _vca.gain.value = gain;
```
pelikhan, line 598: should this use setCurrentToneGain ?

pelikhan (approved):

PR #7515, opened 2020-10-14
don't ignore sim msgs

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/2459
fix https://github.com/microsoft/pxt-arcade/issues/2460
fix https://github.com/microsoft/pxt-microbit/issues/3413
fix mute button

mostly reverts the change in https://github.com/microsoft/pxt/pull/7509; that prevented sending any messages (besides broadcasts) to the simulator. I'm not sure I understand the case the PR said it was fixing, but it'll definitely need a narrower fix for those / if you want to give me a few examples I can see where to cull them @pelikhan

### DIFF
--- a/pxtsim/simdriver.ts
+++ b/pxtsim/simdriver.ts
@@ -268,39 +268,38 @@ namespace pxsim {
                 return
             }
 
-            const broadcastmsg = msg as pxsim.SimulatorBroadcastMessage;
-            if (!source || !broadcastmsg?.broadcast)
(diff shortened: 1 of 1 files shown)


pelikhan (approved):

pelikhan:
The issue currently is that we are sending unrelated messages like debugger iframe messages that the react debugger sends. It's mostly ignored but not generally useful.

PR #7520, opened 2020-10-15
Error message on workspace db failure

jwunderl:
Pop an error message when we transition into an in memory workspace; no issue but some discussion here: https://github.com/microsoft/pxt/pull/7505 and offline

pouch db had a separate form of the in memory db (using pouch dbs apis) -- no reason to split up, so I made it just use the same memoryworkspace as others. also incidentally fixes the minor race condition mentioned in the pr above where we'd potentially make more than one db object

the `disableMemoryWorkspaceWarning` flag is so we can disable this warning in situations like hour of code where we don't really care to notify user we couldn't save.

can't save anything modal

(I'm not too attached to the wording, and we can always throw in a sad emoji to the header if we want!)

### DIFF
--- a/common-docs/browsers/no-auto-save.md
+++ b/common-docs/browsers/no-auto-save.md
@@ -13,3 +13,4 @@ There are several ways to do this depending on the editor you are using:
 
 * Private browsing: If you open makecode in your browsers 'private browsing' mode, it will often fail to be able to save projects - depending on the browser, they might be saved only until the browser window is closed, or only until the window is refreshed.
 * Storage full: The browser allocates a certain amount of space for each website, so it's possible that you have run out of memory. If this is the case, you might need to try to clear space on your hard drive, or delete old projects.
+* Storage disabled: Many browsers have options to disable local storage or cookies entirely. If these settings are enabled in your browser, you may need to disable them for auto save to work again.
--- a/common-docs/browsers/no-auto-save.md
+++ b/common-docs/browsers/no-auto-save.md
@@ -1,6 +1,6 @@
 # No auto save
 
-In some cases, MakeCode is unable to save your project on it's own.
+In some cases, MakeCode is unable to save your project on its own.
 When this happens, you should be careful to back up your projects before leaving the window to make sure you do not lose any progress.
 There are several ways to do this depending on the editor you are using:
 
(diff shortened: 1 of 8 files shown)


pelikhan:
We should have a help link with detailed instructions; button should be red really. Danger zone.

Docs could give tips like make sure your HARD DRIVE IS NOT FULL!

soniakandah:
A couple of suggested wording edits: 

---

Title: Warning! Project Auto-Save Disabled 

Body: We are unable to save your projects at this time. You can still manually save your project via direct download, exporting to GitHub, or publishing/sharing your project. [Learn more about this issue](<link to some documentation>). 

Button: 
Continue

---

I would also put a caution icon before the title text, and make the title segment red. I don't know if the button needs to be red, I think typically in Arcade the button is red when you're changing some data stored. Here, the button isn't really changing anything, it's just letting you move forward.

pelikhan:
Don’t recommend GitHub, it will require a page reload to login.

jwunderl:
@soniakandah did you have a particular red in mind for the title segment / suggestion to make it look better?

image

I tried lightening it a bit but it feels off next to the blue background / etc.

image

soniakandah:
@jwunderl How about this? 

image

Background color of the header is #F5BC5F. Everything else would be the same.

common-docs/browsers/no-auto-save.md
```diff
@@ -0,0 +1,16 @@
+# No auto save
+
+In some cases, MakeCode is unable to save your project on it's own.
```
livcheerful, line 3: small: should be "its" instead of "it's"

common-docs/browsers/no-auto-save.md
```diff
@@ -0,0 +1,16 @@
...
+There are several ways to do this depending on the editor you are using:
+
+* Create a share link for project, which will store it in the cloud.
+* Save your project using the "Save Project" option in the project settings menu.
+* Save your project as a PNG
+* For editors that include hardware, a downloaded hex file can be loaded back into the editor
+
+## Common causes of auto save being disabled
+
+* Private browsing: If you open makecode in your browsers 'private browsing' mode, it will often fail to be able to save projects - depending on the browser, they might be saved only until the browser window is closed, or only until the window is refreshed.
```
livcheerful, line 14: MakeCode

livcheerful (approved):
looks great! just some small grammar stuff

PR #7521, opened 2020-10-15
Upgrade to es2017 emit, drop bluebird

jwunderl:
I wanted to just change the emit to a newer one so that we wouldn't have to have all the wonky polyfill stuff for async/await / etc, but then I remembered that the non-polyfilled async/await will return a native promise, so we'd have to drop bluebird at the same time. Decided to see how much work that would be, now it doesn't seem nearly as bad as I thought it would be.

From the few paths I've clicked through (make a project, load tutorial, sim works, download creates valid hex, etc) it seems to be mostly working. I'm sure there are lots of bugs that I'd need to squash before getting this in though

🦀🦀 open the (browser) debugger, search for `async function` & set a breakpoint, and behold the beauty of not having weird wrappers all over the place 🦀🦀

caveats
* node took a while to add promise finally; added a polyfill in nodeutil https://github.com/es-shims/Promise.prototype.finally
* no more .done, I replaced them all with then to get things working for now. My understanding is that these were used to catch unhandled rejections
* Not 100% sure on `U.nextTick`; it's slightly different than just setTimeout, as I believe it was enqueuing the function to run before promise handlers would run next? Not 100% sure, looking into that a bit more
* added a few helpers https://github.com/microsoft/pxt/compare/upgradeToES2017?expand=1#diff-c66ca22b14f9512edcc64b334c968354a32f478e2c2d7fd5eccef126c3e4f3d6R451 for the bluebird stuff we were using
* the typing for setTimeout is bad. Node returns a `Timer` object, browsers return a number. Since we include node typings for parts of the app, just changing the timer holder to any to support both :(
* Probably will also need to upgrade backend so the promise implementations match (due to utils), but it might be fine to leave that for now? will take a look.
* ~might want to polyfill promise.finally across whole app~ should be done
* upgraded minify-js -> terser; this is a fork of minify-js that supports es6+, and is already a dependency of uglifyify. (also made the build fail if minification fails, as right now we're just ignoring it https://github.com/microsoft/pxt/runs/2082287467?check_suite_focus=true#step:5:168)

(do not merge tag as there will be some timing to consider / need to be ready to finish up and coordinate the prs in all the linked tags. Also need oss review for new dep, and make sure everyone knows to `npm i` everything)


tests/blocklycompiler-test/test.spec.ts
```diff
@@ -163,8 +163,10 @@ class BlocklyCompilerTestHost implements pxt.Host {
     }
 }
 
-function fail(msg: string) {
+function fail(msg: string): never {
     chai.assert(false, msg);
+    // convince compiler we won't return
+    throw msg;
```
riknoll, line 169: you can also do `// @ts-ignore`

riknoll (approved):
Gotta say, I was expecting this to be a much scarier change. LGTM!

pxtlib/util.ts
```diff
@@ -445,14 +445,127 @@ namespace ts.pxtc.Util {
...
+            promise: new Promise<T>(function (r, j) {
+                if (result) {
+                    r(result);
+                } else {
+                    resolve = r;
+                    reject = j;
+                }
+            }),
+
+            isResolved
```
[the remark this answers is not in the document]
jwunderl: Oops, good catch. This one's not required / just had it there so it would be available while I was testing it, so I'm just going to remove it

pxtblocks/composablemutations.ts
```diff
@@ -313,12 +313,13 @@ namespace pxt.blocks {
         Blockly.Extensions.apply('inline-svgs', b, false);
 
         let returnValueVisible = true;
-        updateShape();
 
         // When the value input is removed, we disconnect the block that was connected to it. This
         // is the id of whatever block was last connected
         let lastConnectedId: string;
 
+        updateShape();
```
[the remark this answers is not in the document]
jwunderl: Yes; I'm not sure certain why it wasn't complaining about this before, but it was hitting a compiler error here as `updateShape` references `lastConnectedId`. I'm guessing it was part of the downleveling for arrow functions moving code around?

abchatra:
ES2018 means dropping support for old edge. I am fine with it. But we need to update the docs.
What about safari? I would look at the our telemetry on Safari versions.

webapp/src/cloud.ts
```diff
@@ -175,15 +175,18 @@ function resetAsync(): Promise<void> {
     return Promise.resolve();
 }
 
-let inProgressSyncPromise: Promise<any> = Promise.resolve()
+let inProgressSyncPromise: Promise<pxt.workspace.Header[]>;
```
darzu, line 178: If you await a null or undefined promise, is that okay?

webapp/src/app.tsx
```diff
@@ -1389,16 +1389,16 @@ export class ProjectView
...
+            return pxt.U.promiseTimeout(
+                1500,
                 // start a partial cloud sync
                 cloud.syncAsync([h])
-                    .then((changes) => {
+                    .then(changes => {
                         if (changes.length) {
                             const elapsed = Util.nowSeconds() - timeoutStart;
-                            if (timeout.isResolved()) {
+                            if (timedOut) {
```
darzu, line 1401: So this still runs even though it timed out? Or does it cancel the callback?

If it still runs, then this looks good (if uglier haha)!

darzu: Actually it looks like promiseTimeout is implement in terms of "Promise.race", could you just use that instead of Promise.any ?

webapp/src/cloud.ts
jwunderl: awaiting null / a non promise is fine, implementation wise it either is a no-op or just wraps in Promise.resolve(..)

webapp/src/app.tsx
jwunderl: promises will continue to run when created / timeout doesn't stop the callback from running on any promise inside of it, it just rejects after the timeout (standard promises aren't cancellable - the bluebird implementation for cancellable promises is just a promise wrapper around other promises that rejects when cancelled, so any callbacks contained up till that point still run).

promise.isResolved is the api that doesn't exist / that timedOut is 'replacing' here, if we want can just do `elapsed < 1.5` but that feels like an impossible to debug race condition waiting to happen.

I'll see if I can make it prettier by just directly using race instead of the timeout util though ~

darzu: Ah got it. Thanks for looking into this! Your solution also works

webapp/src/cloud.ts
darzu: Awesome thanks!

webapp/src/app.tsx
jwunderl: You were right, promise.race came out looking better ~

libs/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
     "compilerOptions": {
-        "target": "es5",
+        "target": "esnext",
```
eanders-MS, line 3: >esnext [](start = 19, length = 6)

Wondering: why not es2018 here?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "pxt-core",
-  "version": "6.11.1",
+  "version": "6.10.2",
```
eanders-MS, line 3: >6.10.2 [](start = 14, length = 6)

Intentional version downgrade?

pxtcompiler/emitter/backjs.ts
```diff
@@ -554,7 +554,7 @@ function ${id}(s) {
                     write(`if ((${args[0]}) && (${args[0]}).vtable) {`)
                 }
                 if (topExpr.callingConvention == ir.CallingConvention.Promise) {
-                    write(`(function(cb) { ${text}.done(cb) })(buildResume(s, ${loc}));`)
+                    write(`(function(cb) { ${text}.then(cb) })(buildResume(s, ${loc}));`)
```
eanders-MS, line 557: >then [](start = 51, length = 4)

The code being emitted here.. was the environment where it executes also upgraded from bluebird?

pxtlib/browserutils.ts
```diff
@@ -1020,7 +1019,7 @@ namespace pxt.BrowserUtils {
             return deleteDbAsync();
         return _translationDbPromise
             .then(db => db.clearAsync())
-            .catch(e => deleteDbAsync().done());
+            .catch(e => deleteDbAsync().then());
```
eanders-MS, line 1022: >.then() [](start = 39, length = 7)

extraneous ".then()"?

pxtlib/initprj.ts
```diff
@@ -1,7 +1,7 @@
 namespace pxt.template {
     export const TS_CONFIG = `{
     "compilerOptions": {
-        "target": "es5",
+        "target": "es2017",
```
eanders-MS, line 4: >es2017 [](start = 19, length = 6)

Wondering: why not es2018 here?

package.json
jwunderl: I'll be removing this before merging, just leaving it now in case I need to make any more changes / handle targets -- when we have changes that we need to test in targets / make pr builds for but aren't ready to merge into master, we'll occasionally bump the minor version to make a 'test branch' we can reference in builds -- in this case, I bumped master's minor branch from 6.9 to 6.11, and used 6.10 to create builds on npm for this PR, to make sure it's merge-able without breaking targets / so I could create builds like this one: https://github.com/microsoft/pxt-arcade/pull/3276

pxtcompiler/emitter/backjs.ts
jwunderl: Right, all code in the frontend currently uses bluebird (`.done` isn't actually part of the ecmascript promise spec in the first place)

pxtlib/browserutils.ts
jwunderl: I removed the trivially extraneous `thens` in one of the last commits, but I left the one's where it's not necessarily obvious -- I'm going to file a few issues for 'further cleanup we should do' on monday. The reason I left this one (and ones like it) is that it technically changes the behavior slightly -- when it's at the end of a promise chain that is stored in a variable / returned, it clears the return value. In this case, that means if deleteDbAsync returns something for some reason (or is changed to return something) the value will not be propagated to the caller.

This is probably just being overly careful, just wanted to minimize the mental load on this pr / it's already a large one that is a lot to review.

libs/tsconfig.json
jwunderl: We have two versions of typescript that we use -- one used locally to build most parts of the webapp (e.g. the app itself, the cli), which is currently 3.7.something, and one that we used in the client itself (or in portions of the code that are trying to match the client), which is 2.7.something (it's here: https://github.com/microsoft/pxt/tree/master/pxtcompiler/ext-typescript) -- this is the one we use for building user programs. The tsconfigs that use esnext are the ones that use the 2.7 compiler; in that one es2018 is 'latest' / 'esnext' because it wasn't completely finalized yet, but the behavior matches properly for our purposes.

I'll file an issue on monday for upgrading the shipped compiler as well -- there are some nice features we're missing out in our editor due to the old version we have, but upgrading is non trivial -- a while back we used the same one we build the app with, but we currently rely on some internal api's in typescript for our static typescript compiler so when we upgraded the webapp to typescript 3 we split off the shipped compiler to keep the old version. Bringing it back up to date + ideally communicating what we need to see if we can use supported api's seems like a worthwhile investment to me (at some point at least..)

(We split it in https://github.com/microsoft/pxt/pull/6046 for reference)

eanders-MS: That makes sense. I really want the spread operator in editor :)

pxtlib/initprj.ts
jwunderl: Linking https://github.com/microsoft/pxt/pull/7521#discussion_r593599459 for reference, this is the default tsconfig when we create a new github repo in the editor. (this should be esnext though, I missed it in my grep; thanks for pointing it out~)

eanders-MS (approved):
Awesome work.
:shipit:

PR #7522, opened 2020-10-15
Remove delete button when asset is referenced anywhere in project

shakao:
Trying to prevent any situation where a user could irreparably break their project from the asset editor. Assets cannot be deleted if they are used in any way (referenced in blocks, used in a tilemap, referenced in TS/Python).

### DIFF
--- a/pxtlib/tilemap.ts
+++ b/pxtlib/tilemap.ts
@@ -758,7 +758,7 @@ namespace pxt {
                 for (let filename of Object.keys(files)) {
                     const f = files[filename];
                     // Match .ts files that are not generated (.g.ts)
-                    if (filename.match(/(?<!\.g)\.ts$/i)) {
+                    if (filename.match(/((?!\.g).{2}|^.{0,1})\.ts$/i)) {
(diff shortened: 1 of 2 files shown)


jwunderl (approved):

pxtlib/tilemap.ts
```diff
@@ -677,6 +677,98 @@ namespace pxt {
...
+                        assetPyRefs = `assets.image\("""${shortId}"""\)`;
+                        if (displayName) assetPyRefs += `|assets.image\("""${displayName}"""\)`;
+                        break;
+                }
+                const assetPyRegex = new RegExp(assetPyRefs, "gm");
+
+                for (let filename of Object.keys(files)) {
+                    const f = files[filename];
+                    // Match .ts files that are not generated (.g.ts)
+                    if (filename.match(/(?<!\.g)\.ts$/i)) {
```
jwunderl, line 761: Ahh wait a second, forgot about this; safari doesn't support lookbehind :( https://caniuse.com/mdn-javascript_builtins_regexp_lookbehind_assertion

PR #7530, opened 2020-10-17
Resize toolbar when window resizes

livcheerful:
Add a state for the workspace header so we can rerender when the window size changes. The toolbar won't overlap the toolbox anymore!

resize!


jwunderl (approved):
🦀🦀🦀🦀

PR #7535, opened 2020-10-20
[pxt] Latest changes for identity

eanders-MS:
Latest collection of identity changes to support sign in.
* Better protection against CSRF/XSS attack vectors.
* Improved sign in UI (still WIP and incomplete).
* Added sign in UI to editor view (was only on home screen).
* Stubs for integration with existing cloud sync codepath.


webapp/src/app.tsx
```diff
@@ -3989,7 +3983,6 @@ export class ProjectView
                 {sandbox ? undefined : <scriptsearch.ScriptSearch parent={this} ref={this.handleScriptSearchRef} />}
                 {sandbox ? undefined : <extensions.Extensions parent={this} ref={this.handleExtensionRef} />}
                 {inHome ? <projects.ImportDialog parent={this} ref={this.handleImportDialogRef} /> : undefined}
-                {hasCloud ? <cloud.SignInDialog parent={this} ref={this.handleSignInDialogRef} onComplete={this.cloudSignInComplete} /> : undefined}
```
darzu, line 3992: Was this the old OneDrive / g-drive sync?

webapp/src/app.tsx
```diff
@@ -4466,6 +4454,13 @@ document.addEventListener("DOMContentLoaded", () => {
         return;
     }
 
+    const query = core.parseQueryString(window.location.href);
+
+    // Handle auth callback redirect.
+    if (query["authcallback"]) {
+        auth.loginCallback(query);
```
darzu, line 4461: What's the difference between here and line 4295?

webapp/src/auth.ts
```diff
@@ -102,15 +103,27 @@ export async function loginAsync(idp: pxt.IdentityProviderId, persistent: boolea
...
+
+    // Update state and UI to reflect logged out state.
     clearState();
-    if (wasLoggedIn) {
-        core.infoNotification(lf("Signed out"));
+
+    // Redirect to provider's logout endpoint. If all goes well there, we'll
+    // be redirected back here.
+    if (result.resp?.logout_uri) {
+        window.location.href = result.resp.logout_uri;
```
darzu, line 126: I know other sites like Tynker do this, but it still seems wrong to me to log out of the provider when the user just said "logout" on our site. E.g. logout from MakeCode shouldn't mean logout from MSFT and GitHub too.
But I'm not going to die on that hill

webapp/src/auth.ts
```diff
@@ -76,23 +79,21 @@ export async function loginAsync(idp: pxt.IdentityProviderId, persistent: boolea
 
     pxt.tickEvent('auth.login.start', { 'provider': idp });
 
-    const stateKey = ts.pxtc.Util.guidGen();
     const stateObj: AuthState = {
-        continuationState,
+        continuationHash,
```
darzu, line 83: I'm curious, why remove the generality of arbitrary state and restrict it to just a hash?

darzu (approved):
LGTM!

webapp/src/app.tsx
eanders-MS: Yep. Keeping the implementation around for reference, but removing the UI.

webapp/src/app.tsx
eanders-MS: Line 4295 was handling the callback hash `#authcallback`, but with the hash, all url params were getting stripped off the url. I couldn't figure out where. I was driving me crazy so I switched to using a url parameter to indicate its an auth callback.

webapp/src/auth.ts
eanders-MS: Oh I meant to switch this back to `continuationState`. The hash thing was an experiment in making the callback work. Not using hashes anymore. Good catch.

webapp/src/auth.ts
eanders-MS: I think you're right that we shouldn't sign out of the upstream provider. This will sign the user out from our app on all devices. If we want to support that, it should be a choice: "Sign out from: [Just here] [Everywhere]"

webapp/src/auth.ts
eanders-MS: I updated this to `callbackLocation`. It can be any url fragment, e.g.: `#editor`. It can contain url parameters too, if needed. This should be adequate for the app to pick up where it left off after sign in is complete. This isn't state sent to the server. It is stored locally and read back when the auth flow completes.

webapp/public/images/google-logo.svg
```diff
@@ -0,0 +1 @@
+<svg viewBox="0 0 533.5 544.3" xmlns="http://www.w3.org/2000/svg"><path d="M533.5 278.4c0-18.5-1.5-37.1-4.7-55.3H272.1v104.8h147c-6.1 33.8-25.7 63.7-54.4 82.7v68h87.7c51.5-47.4 81.1-117.4 81.1-200.2z" fill="#4285f4"/><path d="M272.1 544.3c73.4 0 135.3-24.1 180.4-65.7l-87.7-68c-24.4 16.6-55.9 26-92.6 26-71 0-131.2-47.9-152.8-112.3H28.9v70.1c46.2 91.9 140.3 149.9 243.2 149.9z" fill="#34a853"/><path d="M119.3 324.3c-11.4-33.8-11.4-70.4 0-104.2V150H28.9c-38.6 76.9-38.6 167.5 0 244.4l90.4-70.1z" fill="#fbbc04"/><path d="M272.1 107.7c38.8-.6 76.3 14 104.4 40.8l77.7-77.7C405 24.6 339.7-.8 272.1 0 169.2 0 75.1 58 28.9 150l90.4 70.1c21.5-64.5 81.8-112.4 152.8-112.4z" fill="#ea4335"/></svg>
```
eanders-MS, line 1: @darzu What is the right way to add static content so it will go to cdn?

PR #7537, opened 2020-10-21
Fixing some bugs in temporary assets

riknoll:
We were losing references to temporary assets when switching between the asset editor and blocks editor. This fixes that and also makes it so that the source of truth for assets is the block data and not the field value (though that still gets used if the data doesn't exist)


shakao:
also fixes https://github.com/microsoft/pxt-arcade/issues/2493 !

shakao (approved):

PR #7538, opened 2020-10-21
Update simulator from asset editor view, fix for tileset block

shakao:
https://github.com/microsoft/pxt/commit/026a824a30ff8d5c4041cf0026fa730b593c60b5 - mark simulator as dirty, update main.ts for unnamed/temporary assets since they're emitted as literals in the case of images and animations (second half of fix for https://github.com/microsoft/pxt-arcade/issues/2488)
https://github.com/microsoft/pxt/commit/4a6d4be6d8f2df822b5d63a4f606c073f6d09da1 - null check, emit id directly for gallery tiles (fixes https://github.com/microsoft/pxt-arcade/issues/2512)

### DIFF
--- a/webapp/src/components/assetEditor/editor.tsx
+++ b/webapp/src/components/assetEditor/editor.tsx
@@ -159,7 +159,12 @@ export class AssetEditor extends Editor {
         fieldView.onHide(() => {
             const result = fieldView.getResult();
             if (asset.type == pxt.AssetType.Tilemap) result.data = this.updateTilemapTiles(result.data);
-            cb(result);
+
+            Promise.resolve(cb(result)).then(() => {
+                this.parent.saveBlocksToTypeScriptAsync().then((src) => {
+                    pkg.mainEditorPkg().setContentAsync("main.ts", src)
+                })
+            });
         });
         fieldView.show();
     }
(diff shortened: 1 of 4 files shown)


jwunderl (approved):

webapp/src/components/assetEditor/editor.tsx
```diff
@@ -159,7 +159,12 @@ export class AssetEditor extends Editor {
         fieldView.onHide(() => {
             const result = fieldView.getResult();
             if (asset.type == pxt.AssetType.Tilemap) result.data = this.updateTilemapTiles(result.data);
-            cb(result);
+
+            Promise.resolve(cb(result)).then(() => {
+                this.parent.saveBlocksToTypeScriptAsync().then((src) => {
+                    pkg.mainEditorPkg().setContentAsync("main.ts", src)
```
riknoll, line 165: Could we unify this with the normal blocks saving codepath somehow?

shakao: it's a little weird because the default save path is very tied to the current editorFile, but in this case what we really want is to pretend the current file is main.blocks instead of assets.json. i could pass a file override to saveFileAsync/saveTypeScriptAsync? or if you have a better suggestion, totally welcome!

shakao: also i'm gonna update so that we only do this save if we're editing a temporary asset

riknoll: ehhh, okay

riknoll (approved):

PR #7545, opened 2020-10-22
Set toggle position based on number of toggle elements

shakao:
Build: https://arcade.makecode.com/app/497f26dff3463a0cb641fec965a062b57bc34850-5ffe727f79

Cleans up the various cases we had and positions the toggle strictly based on the number of children in the toggle and which child is active. Elements after the dropdown are slightly special-cased since the dropdown changes the offset amount. container.tsx changes are mostly just shifting those conditionals around, but I also added base-editormenuitem CSS class.

### DIFF
--- a/theme/common.less
+++ b/theme/common.less
@@ -329,6 +329,8 @@ div.simframe > iframe {
     background: none;
     color: @mainMenuInvertedBackground !important;
 }
+
+/* toggle style */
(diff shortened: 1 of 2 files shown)


jwunderl (approved):
Awesome

riknoll (approved):
1000% better

PR #7547, opened 2020-10-23
Resize headers ALL the time

livcheerful:
Resize the headers when we scroll with mouse! and touch

The slider will update for the wheel scroll, but not with touch. Do you guys think that's okay?

resize


pelikhan:
Unrelated but the step description should zoom too.

jwunderl (approved):

PR #7551, opened 2020-10-23
Add user profile page

eanders-MS:
* Added user profile page.
* Reworked the login API to allow the backend to know the origin of the request in order to relax samesite cookie restriction for localhost.
* Ensure all identity functionality is behind the experiment.
* Preserve the url path when signing in. Allows auth to work from an `uploadtrg` deployment.


darzu (approved):
LGTM!

PR #7556, opened 2020-10-26
User preferences syncing for high-contrast and language

darzu:
**Semantics**
When logged in, the user will now have their high-contrast and language preferences synced with the cloud.
When loading MakeCode, we'll do a background fetch of the user preferences.
If the language in the cloud is different from the local one, we'll update the local one and reload the page.
If the contrast in the cloud is different from the local one, we'll update the local setting mode without needing to reload the page.
If the contrast or language changes, we push those changes to the cloud.

**For high contrast:**
A lot of the work in this PR is about creating  a single source of truth for high-contrast mode and using the virtual APIs (from data.ts) to inform all React and non-react renderables to update their contrast mode if contrast changes. This saves us lots of state duplication and "this.props.parent.setHighContrast(this.state.highContrast)" type calls (bubbling state changes up and down react hierarchies). 

**For language:**
Previously language state was stored in a local cookie. For now we're keeping the cookie around and we keep its value in sync with the cloud state. Theoretically we could skip the cookie for logged in users.
One tricky thing with language state changes is we don't have the infrastructure in place to dynamically swap out or re-render strings. So we just always reload the page if the language changes.

**Virtual APIs change**
To support all of this, I added a way to subscribe to change virtual API notifications outside of react components, and generalized the notion of a virtual api subscriber.

**Auth changes**
I've made the "loggedIn()" function async and the user profile is now awaitable so that cloud operations that need to assume we're logged in or have the profile can await these.


eanders-MS:
>     await fetchUserAsync();

I was lazily fetching the user to test if we're already logged in, but I don't think this will scale. We should make a new endpoint to fetch all user state in a single call (profile, preferences, more as we add things).

---
Refers to: webapp/src/auth.ts:147 in 82a0a70. [](commit_id = 82a0a7092e274cb7902c71b224d462eb25a169ad, deletion_comment = False)

webapp/src/auth.ts
```diff
@@ -256,6 +263,43 @@ async function fetchUserAsync() {
     }
 }
 
+
+export async function updateUserPreferencesAsync(newPref: Partial<UserPreferences>) {
```
eanders-MS, line 267: >export [](start = 0, length = 6)

Code organization: Put exported functions above the "Private functions" section.

webapp/src/auth.ts
```diff
@@ -256,6 +263,43 @@ async function fetchUserAsync() {
...
+
+    if (!loggedIn()) { return; }
+
+    // If the user is logged in, save to cloud
+    const result = await apiAsync<UserPreferences>('/api/user/preferences', newPref);
+    if (result.success) {
+        console.log("updating userpref from POST")
+        // Set user profile from returned value
+        state_.preferences = { ...state_.preferences, ...result.resp }
+        data.invalidate("user-pref");
```
eanders-MS, line 279: >"user-pref" [](start = 24, length = 11)

Duplicating "user-pref" everywhere can lead to problems later. I think it would be better to define a const for it at the top of this file.

webapp/src/auth.ts
```diff
@@ -327,4 +371,19 @@ function clearState() {
     data.invalidate(NEEDS_SETUP);
 }
 
+function userPreferencesHandler(path: string): UserPreferences {
```
eanders-MS, line 374: >userPreferencesHandler [](start = 9, length = 22)

Can this be rolled into the auth virtual api? By the way, I want to rename this  to "user" api.

webapp/src/data.ts
```diff
@@ -2,7 +2,10 @@ import * as React from "react";
 import * as core from "./core";
 
 export type Action = () => void;
-export type AnyComponent = Component<any, any>;
+export type DataSubscriber = {
```
eanders-MS, line 5: >DataSubscriber [](start = 12, length = 14)

Nice!

webapp/src/sui.tsx
```diff
@@ -5,6 +5,7 @@ import * as ReactTooltip from 'react-tooltip';
 
 import * as data from "./data";
 import * as core from "./core";
+import * as auth from "./auth";
```
eanders-MS, line 8: >auth [](start = 25, length = 4)

I want to rename ./auth.ts to ./user.ts

webapp/src/auth.ts
```diff
@@ -327,4 +371,19 @@ function clearState() {
...
+        }
+        fetchUserPreferencesAsync();
+    }
+    return state_?.preferences 
+}
+
 data.mountVirtualApi("auth", { getSync: authApiHandler });
+
+console.log("mounting user-pref") // TODO @darzu: 
+data.mountVirtualApi("user-pref", { getSync: userPreferencesHandler });
```
jwunderl, line 389: Question - if it's getting all the preferences at once here, will it end up making any change in a single preference rerender all components that refer to any preference (e.g. changing language also rerenders all high contrast elements)? Would it make sense to subscribe to each pref separately ('user-pref:high-contrast', 'user-pref:language') so that each component only listens for the preferences it depends upon?

eanders-MS:
>     await fetchUserAsync();

We can do it in a separate PR, but the login endpoint needs to get more sophisticated and be able to serve as a login test, returning all the user data if they're already authenticated.

---
In reply to: [716886184](https://github.com/microsoft/pxt/pull/7556#issuecomment-716886184) [](ancestors = 716886184)

---
Refers to: webapp/src/auth.ts:147 in 82a0a70. [](commit_id = 82a0a7092e274cb7902c71b224d462eb25a169ad, deletion_comment = False)

webapp/src/auth.ts
darzu: You've got it exactly right. For simplicity I was using user-pref everywhere but we should use :high-contrast etc so we can have more fine-grain subscriptions.

darzu:
> > ```
> > await fetchUserAsync();
> > ```
> 
> I was lazily fetching the user to test if we're already logged in, but I don't think this will scale. We should make a new endpoint to fetch all user state in a single call (profile, preferences, more as we add things).
> 
> Refers to: webapp/src/auth.ts:147 in [82a0a70](https://github.com/microsoft/pxt/commit/82a0a7092e274cb7902c71b224d462eb25a169ad). [](commit_id = [82a0a70](https://github.com/microsoft/pxt/commit/82a0a7092e274cb7902c71b224d462eb25a169ad), deletion_comment = False)

Yes, a bundled call makes sense once we know what we'll need. We'll probably want to send projects and skills map metadata too. I figure for now it was better to keep things conceptually separated and take steps once there's a performance need. Otherwise we run the risk of premature optimization.

webapp/src/app.tsx
```diff
@@ -133,6 +133,13 @@ export class ProjectView
     private openingTypeScript: boolean;
     private preserveUndoStack: boolean;
 
+    private highContrastSubscriber: data.DataSubscriber = {
```
eanders-MS, line 136: >DataSubscriber [](start = 41, length = 14)

This component already extends data.Component. Do you need this here too?

webapp/src/auth.ts
darzu: Hmmm, we want to handle subfields like high-contrast and language. Can we nest multiple levels? Like "auth:user:preferences:high-contrast" ?

webapp/src/auth.ts
darzu: Okay, fine-grain subscriptions should be working in theory; gotta do a little testing to make sure

webapp/src/app.tsx
darzu: Yes, because we need to get updates outside of the render lifecycle. The data.Component only updates the component by calling render

webapp/src/auth.ts
```diff
@@ -151,20 +171,32 @@ export async function logout() {
...
 }
 
+let initialAuthCheck_: Promise<UserProfile | undefined> = undefined;
 /**
  * Checks to see if we're already logged in by trying to fetch user info from
  * the backend. If we have a valid auth token cookie, it will succeed.
  */
-export async function authCheck() {
-    if (!hasIdentity()) { return; }
+export async function authCheck(): Promise<UserProfile | undefined> {
```
darzu, line 179: @eanders-MS , I've reworked authCheck to only run once using the initialAuthCheck_ promise so that it can be called many times and later calls will just return the first result.

I do a similar thing with fetching user preferences.

webapp/src/auth.ts
```diff
@@ -254,17 +286,16 @@ export function hasIdentity(): boolean {
     return experimentEnabled && identityProviders().length > 0;
 }
 
-export function loggedIn(): boolean {
-    if (!hasIdentity()) { return false; }
-    const state = getState();
-    return !!state.user?.id;
+export async function loggedIn(): Promise<boolean> {
+    await authCheck();
+    return loggedInSync();
```
darzu, line 291: @eanders-MS , I've made loggedIn async and it blocks waiting on the initial user profile fetch. There is a sync version too that checks the current state but it seemed like most locations really wanted the async version to avoid race conditions.

webapp/src/auth.ts
```diff
@@ -299,15 +330,70 @@ export class Component<TProps, TState> extends data.Component<TProps, TState> {
...
+    if (oldPref?.highContrast !== state_.preferences?.highContrast) {
+        data.invalidate(HIGHCONTRAST)
+    }
+    if (oldPref?.language !== state_.preferences?.language) {
+        data.invalidate(LANGUAGE)
+    }
+}
+
+let initialUserPreferences_: Promise<UserPreferences | undefined> = undefined;
+export async function initialUserPreferences(): Promise<UserPreferences | undefined> {
```
darzu, line 369: @eanders-MS , this is the pattern I'm using for getting preferences. We first wait for a loggin check and then we do a single fetch of the preferences and subsequent calls use this. I think this makes sense.

Eventually we should return more data from the serve in the initial payload but I think this works well for now.

webapp/src/auth.ts
eanders-MS: Seems like a good change.

---
In reply to: [521609168](https://github.com/microsoft/pxt/pull/7556#discussion_r521609168) [](ancestors = 521609168)

eanders-MS (approved):
Nice cleanup on the high contrast front!
:shipit:

PR #7557, opened 2020-10-27
Fix large hex downloads on chrome <=79

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/3417 - the dataurl was just too long for older versions of chrome to handle it. `createObjectURL` is listed as a working draft but browser support is good https://caniuse.com/bloburls

~I added a TODO for it, but will want to look at this a bit more and `revokeObjectUrl` -- as written the blob will persist in memory until refresh most likely~ added handling to revoke url when it goes out of use. Mainly wanted initial thoughts as this part of the code hasn't changed much / in case there is something known that I'm overlooking about `createObjectURL`

here's a build since I'm using it for browserstack anyways https://makecode.microbit.org/app/662dfecbb1be03dc15f0348e51683672c8e02097-64245805e1

### DIFF
--- a/pxtlib/browserutils.ts
+++ b/pxtlib/browserutils.ts
@@ -369,18 +369,27 @@ namespace pxt.BrowserUtils {
     export function browserDownloadBase64(b64: string, name: string, contentType: string = "application/octet-stream", userContextWindow?: Window, onError?: (err: any) => void): string {
         pxt.debug('trigger download')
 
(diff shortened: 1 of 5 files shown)


pelikhan (approved):

pelikhan:
@jwunderl  ready?

jwunderl:
Yup, pushed a last cleanup change. I did add the blob url revoke logic to prevent memory leaks since you reviewed if you wanted to look over that, otherwise I'll merge in a few~

abchatra:
ported to stable6.0

PR #7558, opened 2020-10-28
move responsability of reconnect to reflashAsync

pelikhan:
For the micro:bit, don't disconnect from DAPlink after flashing -- since it stays alive anyway. This reduces the number of disconnect/reconnect.

The responsibility of reconnect is now moved to the reflashAsync function.
Tested v1/v2.


pelikhan:
@abchatra needs integration

mmoskal (approved):

abchatra:
ported to stable6.0

PR #7561, opened 2020-10-28
Parse skills map

riknoll:
Very simple syntax:

Single # header defines the skills map
Double ## header defines an activity
Key value pairs are just `key: value` and are case insensitive for the keys

For map prerequisites, they can be either skills maps you must have finished or tags

### DIFF
--- a/skillsmap/src/lib/skillMap.d.ts
+++ b/skillsmap/src/lib/skillMap.d.ts
@@ -1,0 +1,55 @@
+interface SkillsMap {
+    mapId: string;
+    displayName: string;
+    description?: string;
+    prerequisites: MapPrerequisite[];
+
+    actvities: {[index: string]: MapActivity};
+    root: MapActivity;
+}
+
+type MapPrerequisite = TagPrerequisite | ActivityPrerequisite;
+
+interface TagPrerequisite {
+    type: "tag"
+    tag: string;
+    numberCompleted: number;
+}
+
+interface ActivityPrerequisite {
+    type: "activity";
+    mapId: string;
+}
+
+interface MapActivity {
+    activityId: string;
+
+    displayName: string;
+    description: string;
+    tags: string[];
+
+    type: "tutorial";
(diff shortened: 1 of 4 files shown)


skillsmap/src/lib/skillMap.d.ts
```diff
@@ -0,0 +1,55 @@
...
+interface ActivityPrerequisite {
+    type: "activity";
+    mapId: string;
+}
+
+interface MapActivity {
+    activityId: string;
+
+    displayName: string;
+    description: string;
```
shakao, line 28: required?

skillsmap/src/lib/skillMap.d.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+interface MapState {
+    mapId: string;
+    activityState: {[index: string]: ActivityState};
+}
+
+interface ActivityState {
+    activityId: string;
+    headerId: string;
```
shakao, line 53: just checking: this is so we can load them back into a partially completed activity?

skillsmap/src/lib/skillMapParser.ts
```diff
@@ -0,0 +1,262 @@
...
+}
+
+export function test() {
+    return parseSkillsMap(testMap);
+}
+
+export function parseSkillsMap(text: string) {
+    const sections = getSectionsFromText(text);
+
+    const result = inflateSkillMap(sections[0]);
```
shakao, line 58: later we would expand this to theoretically return an array of skill maps?

skillsmap/src/lib/skillMap.d.ts
riknoll: yep!

skillsmap/src/lib/skillMap.d.ts
riknoll: Changed

shakao (approved):

PR #7562, opened 2020-10-28
Cache namespaces that are defined in multiple packages in each of those packages

aznhassan:
# Problem
When you load multiple extensions that are in the same namespace, only one extension will have it's namespace API info cached. This means that if you load two such extensions in a project, back-out and create a new project with only one of the extensions (the one you loaded first), you will not find that extension's blocks in the toolbox. No category shows up in the toolbox for that extensions namespace since the namespace API info was not present in the cache for that extension. You can still search for the blocks, just not drag-and-drop them from the toll-box. 

# Solution
### Creating the API info list
There is already code to resolve when multiple packages define the same API info. If it's not two interface types that collided we usually make sure to either combine their sources or keep one if it contains the other. After that work is done, we'll check if we're dealing with two module (namespace) types. If so, if they come from different packages we'll add the existing namespace's package ID to a list of packages IDs on the new API info. The new API info is then stored in place of the existing.

### Caching it
We end up with a single namespace API info object that contains a list of the packages it was defined in. When caching this, we'll now check that an extension's package ID matches any of those defined in the current API's package list as well as its package ID. We make sure to remove the list of packages from the API before storing it in the cache. Now each extension will keep a copy of the namespace API info, so when they're loaded separately the namespace will show up in the toolbox.

### Reading it back from the cache
I actually don't do anything here since the namespace we save when there are duplicates are the same in all the packages/extensions we save them in. Meaning when you read back the namespaces from the cache only one of them will end up in the list of API info objects.

# Testing
Using Microbit, try loading the makerbit-lcd1602 extension. You'll see the `Makerbit` category in the toolbox along with subcategories for LCD1602 and LCD2004 blocks. Load the makerbit extension next. You'll still see the `Makerbit` category, curiously the LCD2004 subcategory no longer appears, that is a separate issue. Go back to the home screen and load a new project. If you add the LCD1602 extension again you'll see that there is no `Makerbit` category in the toolbox, though you can still search for the blocks.

After my changes, you can do the following and the `Makerbit` category will now show up. Though, much like when we loaded both the makerbit-lcd1602 and makerbit extensions together, there will be no subcategory for LCD2004. You can reverse the order that you load the extensions and then load a project with just the LCD extension and the `Makerbit` category will be present in the toolbox.

### After Loading Just the makerbit-lcd1602 extension
JustLCD

### After loading the makerbit extension as well
MakerbitAndLcd

### Creating a new project with just the makerbit-lcd1602 extension after the previous step
JustLcdAfterMakerbit

fixes microsoft/pxt-microbit#3332

### DIFF
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -216,7 +216,7 @@ namespace ts.pxtc {
             let hasParams = kind == SymbolKind.Function || kind == SymbolKind.Method
 
             let pkg: string = null
-            let pkgs: string[] = []
+            let pkgs: string[] = null
 
             let src = getSourceFileOfNode(stmt)
             if (src) {
@@ -531,10 +531,13 @@ namespace ts.pxtc {
                             }
                             si.attributes = parseCommentString(source);
 
+                            // Check if the colliding symbols are namespace definitions. The same namespace can be
+                            // defined in different packages/extensions, so we want to keep track of that information.
+                            // That way, we can make sure each cached extension has a copy of the namespace
                             if (existing.kind === SymbolKind.Module) {
-                                // Copy the existing package array to the new symbol info
+                                // Reference the existing array of packages where this namespace has been defined 
                                 si.pkgs = existing.pkgs || []
-                                if (existing?.pkg !== si?.pkg) {
+                                if (existing.pkg !== si.pkg) {
                                     if (!si.pkgs.find(element => element === existing.pkg)) {
                                         si.pkgs.push(existing.pkg)
                                     }
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -216,6 +216,7 @@ namespace ts.pxtc {
             let hasParams = kind == SymbolKind.Function || kind == SymbolKind.Method
 
             let pkg: string = null
+            let pkgs: string[] = []
 
             let src = getSourceFileOfNode(stmt)
             if (src) {
@@ -251,6 +252,7 @@ namespace ts.pxtc {
                 fileName: stmt.getSourceFile().fileName,
                 attributes,
                 pkg,
+                pkgs,
                 extendsTypes,
                 retType:
                     stmt.kind == SyntaxKind.Constructor ? "void" :
@@ -528,6 +530,17 @@ namespace ts.pxtc {
                                 source = foundSrc;
                             }
                             si.attributes = parseCommentString(source);
+
+                            if (existing.kind === SymbolKind.Module) {
+                                // Copy the existing package array to the new symbol info
+                                si.pkgs = existing.pkgs || []
+                                if (existing?.pkg !== si?.pkg) {
+                                    if (!si.pkgs.find(element => element === existing.pkg)) {
+                                        si.pkgs.push(existing.pkg)
+                                        console.log(`${si.pkg}/${si.name} adding package ${existing.pkg} to the list of packages ${existing.pkgs}`)
+                                    }
+                                }
+                            }
                             if (existing.extendsTypes) {
                                 si.extendsTypes = si.extendsTypes || []
                                 existing.extendsTypes.forEach(t => {
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -537,7 +537,6 @@ namespace ts.pxtc {
                                 if (existing?.pkg !== si?.pkg) {
                                     if (!si.pkgs.find(element => element === existing.pkg)) {
                                         si.pkgs.push(existing.pkg)
-                                        console.log(`${si.pkg}/${si.name} adding package ${existing.pkg} to the list of packages ${existing.pkgs}`)
                                     }
                                 }
                             }
--- a/webapp/src/compiler.ts
+++ b/webapp/src/compiler.ts
@@ -701,10 +701,13 @@ async function cacheApiInfoAsync(project: pkg.EditorPackage, info: pxtc.ApisInfo
             for (const api of apiList) {
                 const apiInfo = info.byQName[api];
 
-                if (apiInfo.pkg === pkgId || apiInfo?.pkgs.find(element => element === pkgId)) {
+                // We include an API in the list of APIs for an external package in two cases:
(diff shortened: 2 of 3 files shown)


jwunderl (approved):

pxtcompiler/emitter/service.ts
```diff
@@ -216,6 +216,7 @@ namespace ts.pxtc {
             let hasParams = kind == SymbolKind.Function || kind == SymbolKind.Method
 
             let pkg: string = null
+            let pkgs: string[] = []
```
pelikhan, line 219: don't eagerly allocate this array as it is most of the time not used

pxtcompiler/emitter/service.ts
```diff
@@ -528,6 +530,16 @@ namespace ts.pxtc {
                                 source = foundSrc;
                             }
                             si.attributes = parseCommentString(source);
+
+                            if (existing.kind === SymbolKind.Module) {
+                                // Copy the existing package array to the new symbol info
+                                si.pkgs = existing.pkgs || []
+                                if (existing?.pkg !== si?.pkg) {
```
pelikhan, line 537: you are passing a reference to the array. did you want to ``.slice(0)`` it?

pxtcompiler/emitter/service.ts
```diff
@@ -528,6 +530,16 @@ namespace ts.pxtc {
                                 source = foundSrc;
                             }
                             si.attributes = parseCommentString(source);
+
+                            if (existing.kind === SymbolKind.Module) {
+                                // Copy the existing package array to the new symbol info
+                                si.pkgs = existing.pkgs || []
```
pelikhan, line 536: don't pre-allocate pkgs; only do it when necessary since most of the time it's not used

webapp/src/compiler.ts
```diff
@@ -700,8 +700,13 @@ async function cacheApiInfoAsync(project: pkg.EditorPackage, info: pxtc.ApisInfo
 
             for (const api of apiList) {
                 const apiInfo = info.byQName[api];
-                if (apiInfo.pkg === pkgId) {
-                    entry.apis.byQName[api] = apiInfo;
+
+                if (apiInfo.pkg === pkgId || apiInfo?.pkgs.find(element => element === pkgId)) {
+                    // Create a copy of the API info since we want to remove the package list.
+                    // Most info objects won't have a package list
+                    entry.apis.byQName[api] = Object.assign({}, apiInfo);
```
pelikhan, line 707: you can ``U.cone(...)``

pelikhan:
- what happens when you unload an extension?
- what happens when you update an extension?

pelikhan:
This area is incredibly confusing. Defnitely appreciate the comments in the code; maybe migrate more of the PR comments in the code.

pxtcompiler/emitter/service.ts
aznhassan: Good point, I've changed this to be null by default.

pxtcompiler/emitter/service.ts
aznhassan: I actually do want it to be a reference vs copying it. We're replacing the `existing` symbol info in the `byQname` object with current symbol info `si` object, so the current symbol info will be the only one referencing the package array. Since the `existing` object reference will no won't come up again, there is no harm in copying it's reference to the package array vs copying the contents of said array.

I updated the comment to make my intention here clearer.

webapp/src/compiler.ts
aznhassan: I've updated the code to do just that.

pxtcompiler/emitter/service.ts
aznhassan: At this point we already have two colliding namespace symbols, we've also already either combined or picked one of their sources to keep. So when we set `pkgs` at this point, it's because we need to.

aznhassan:
> This area is incredibly confusing. Defnitely appreciate the comments in the code; maybe migrate more of the PR comments in the code.

Take a look, I've updated service.ts and compiler.ts with comments that hopefully give context around why I'm adding these changes.

pelikhan:
> Though, much like when we loaded both the makerbit-lcd1602 and makerbit extensions together, there will be no subcategory for LCD2004

Isn't that mean that the solution is partially fixed?

aznhassan:
> > Though, much like when we loaded both the makerbit-lcd1602 and makerbit extensions together, there will be no subcategory for LCD2004
> 
> Isn't that mean that the solution is partially fixed?

This happens when you combine both extensions in the same project. It's a separate issue from not including the namespace API info in each extension/package that defines it leading to missing categories. The latter is what I fixed, the former I'll need create an issue to investigate this more.

pelikhan:
Can we disable the caching for microbit?

aznhassan:
> * what happens when you unload an extension?
> * what happens when you update an extension?

If by unload you mean remove an extension from a project, then that extension will still stay in the cache. I believe we won't update the cached extension unless it's removed first but I'm sure. @riknoll or @jwunderl would probably be able to shine more light on this.

jwunderl:
When you remove an extension it'll go through the same path and just not be included when building the set of extensions (i.e. it won't be in the list of files that get pulled from the cache in compiler.ts anymore); when you update one, the hash won't match (it's a hash over all the files + the target versions so if nothing else pxt.json will have changed version numbers) so it will rebuild and cache the new content through this path in compiler.ts

There are ways to construct packages that will break the api caching logic right now; e.g. if you make an extension that references another github extension, and make it just declare the functions from the dependency with different comment attributes (`//% stuff`), it's very likely the dependency will get cached with the wrong block attributes. I guess we could cache based off the set of extensions instead of a per extension basis? That would also help with the fact we don't currently handle `fileDependencies`~

aznhassan:
> Can we disable the caching for microbit?
> […](#)

I'm not sure, though this would fix the issue. I'm not sure if the trade-off would be worth it since we're caching for a reason. If you feel strongly about this, feel free to open up an issue so we can further discuss if this is the move to make.

pelikhan:
I'm slightly worried that the caching strategy is too convoluted; and will lead to future (acidental) breaks. But we don't have the time to revisit it on this round so go ahead.

PR #7564, opened 2020-10-29
Fix 'pxt heap' crash on micro:bit

mmoskal:
Add example pointers

@abchatra no need to merge; I don't think anyone besides me is using this


pelikhan (approved):

abchatra:
Sounds good. Thanks!

PR #7565, opened 2020-10-29
Markdown format for codecard

pelikhan:
Backward compatible support for a markdown codecard section.

Our current code card format is cumbersome: it is really easy to create invalid json, it does not integrate well at all in crowdin. I was trying to translate the micro:bit cards... and it's impossible.

    ```codecard
    [{
        "name": "A card",
        "url": "...."
    }, {
        "name": "Another card",
        "url": "...."
    }]
    ```
Crowdin decides to chunk it randomly.

image

The new format is based on markdown and guarantees a smooth integration in crowdin (screenshot below).

```
    ### ~ codecard
    * name: A card
    * url: ...
    ---
    * name: Another card
    * url: ...
    ---
    * name: Yet another card
    * url: ...
    ### ~
```
image

To support this format, we will also need to update the backend renderer.

### DIFF
--- a/pxtlib/gallery.ts
+++ b/pxtlib/gallery.ts
@@ -98,7 +98,7 @@ namespace pxt.gallery {
             .filter(cmd => !!cmd)
             .map(cmd => {
                 let cc: any = {};
-                cmd.replace(/^\s*\*\s+(\w+)\s*:\s*(.*)$/gm, (m, n, v) => {
+                cmd.replace(/^\s*(-|\*)\s+(\w+)\s*:\s*(.*)$/gm, (m, n, v) => {
                     if (n == "flags")
                         cc[n] = v.split(',')
                     else
@@ -119,7 +119,7 @@ namespace pxt.gallery {
         let card: any;
         Array.from(el.children)
             .forEach(child => {
-                if (child.tagName === "UL") {
+                if (child.tagName === "UL" || child.tagName === "OL") {
                     if (!card)
                         card = {};
                     // read fields into card
--- a/pxtlib/gallery.ts
+++ b/pxtlib/gallery.ts
@@ -85,6 +85,53 @@ namespace pxt.gallery {
         return prj;
     }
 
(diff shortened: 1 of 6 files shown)


pxtrunner/runner.ts
```diff
@@ -875,6 +875,10 @@ ${linkString}
     </div>
 </aside>
 
+<aside id=codecard class=box>
+    <pre><code class="lang-codecard">@BODY@</code></pre>
+</aside>
```
pelikhan, line 880: this change needs to be ported to the server side.

pxtlib/gallery.ts
```diff
@@ -85,6 +85,68 @@ namespace pxt.gallery {
...
+            cards = [cards];
+        if (cards?.length)
+            return cards;
+
+        // not json, try parsing as sequence of key,value pairs, with line splits
+        cards = md.split(/^---$/gm)
+            .filter(cmd => !!cmd)
+            .map(cmd => {
+                let cc: any = {};
+                cmd.replace(/^\s*\*\s+(\w+)\s*:\s*(.*)$/gm, (m, n, v) => {
```
shakao, line 101: we could support both "*" and "-" bullet points like markdown

pxtlib/gallery.ts
```diff
@@ -85,6 +85,68 @@ namespace pxt.gallery {
...
+                    if (n == "flags")
+                        cc[n] = v.split(',')
+                    else
```
shakao, line 104: do we also plan to support arrays of objects like "CodeCardAction" in this format?

shakao (approved):

pxtlib/gallery.ts
pelikhan: added

jwunderl (approved):

pxtlib/gallery.ts
pelikhan: good added.

PR #7567, opened 2020-10-29
Filter out hide iteration tutorials

jwunderl:
fix https://github.com/microsoft/pxt-minecraft/issues/2051

the hideIteration tutorial flag hides the home button / makes tutorials inescapable. Don't show those under my projects so users can't get stuck.

As an alternative I took a look at doing it with the [temporary header flag](https://github.com/microsoft/pxt/blob/c73b189198b392ab371beab46a57f030c12ca51f/localtypings/projectheader.d.ts#L15) so the project wouldn't just sit around in local storage in perpetuity, but that will lose the project on refresh, and drop you into the normal editor.

I think one better thing we could do would be to change the behavior of the `temporary` flag to make the project have a lifespan of one day instead of not storing it at all, hide them the carousels, and then clear old ones in [`migratePrefixesAsync`](https://github.com/microsoft/pxt/blob/c73b189198b392ab371beab46a57f030c12ca51f/webapp/src/browserworkspace.ts#L10) and on oom errors -- where I've seen it used I actually would want the project to persist past a refresh, just not indefinitely. If that sounds alright with everyone, I'll do that after I finish the other stuff needed for release~

@livcheerful quick fix while I was looking through in tutorials.tsx, the flyout header is a part of the ui / not something the user will intend to copy so I disabled selection on it ~

### DIFF
--- a/webapp/src/projects.tsx
+++ b/webapp/src/projects.tsx
@@ -623,6 +623,10 @@ export class ProjectsCarousel extends data.Component<ProjectsCarouselProps, Proj
             const headers = this.fetchLocalData()
             const showNewProject = pxt.appTarget.appTheme && !pxt.appTarget.appTheme.hideNewProjectButton;
             const showScriptManagerCard = targetTheme.scriptManager && headers.length > ProjectsCarousel.NUM_PROJECTS_HOMESCREEN;
+
+            const headersToShow = headers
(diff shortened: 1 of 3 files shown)


livcheerful (approved):
(the tutorial.tsx change is 👨‍🍳)

abchatra (approved):

abchatra (approved):

PR #7569, opened 2020-10-29
don't show import from file if on android

livcheerful:
workaround for https://github.com/microsoft/pxt-minecraft/issues/1996 

If we're running in minecraft and we're on a chromebook / android device, don't show the option to import from files. it doesn't work!

new target here: https://minecraft.makecode.com/app/3a840092483734d565d7fe3f74da9f2a37939181-da851b6b60?ipc=1&ingame=1


jwunderl (approved):

abchatra (changes requested):
There are more places disableFileAccessinMaciOs is used. Make it consistent with this.

abchatra (approved):

abchatra:
Fixed

PR #7576, opened 2020-10-30
if in ipcRenderer, send that state in sim msg

jwunderl:
The host is implemented differently for ios / android / desktop, so loadIpc isn't guaranteed to exist in non-desktop. ipcRenderer will eventually get loaded, so include that in the run msg so we can guarantee the simulator knows it is in game~


livcheerful (approved):

abchatra:
Why is loadIpc  not guaranteed? What is the expected flow? I am not sure I understand why  pxt.BrowserUtils.isIpcRenderer() would not be deterministic.

abchatra (approved):

PR #7584, opened 2020-11-03
Open skills map activities in the makecode editor and save in indexeddb

riknoll:
Adds a few things:

1. A workspace for save projects and user state in indexeddb
2. An iframe that hosts the makecode editor
3. New message in pxt for starting an activity. Basically it's the same as doing a `#tutorial:/path` url but in message form. Right now this means you have to also serve localhost for testing but I'll work on doing an uploadtrg to make it easier

Known bugs: 
1. sometimes the editor doesn't save immediately, you might need to wait until the simulator to refresh before your project is saved


skillsmap/src/components/CarouselItem.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from "react";
 import { connect } from 'react-redux';
 
 import { SkillsMapState } from '../store/reducer';
-import { dispatchChangeSelectedItem } from '../actions/dispatch';
+import { dispatchChangeSelectedItem, dispatchOpenActivity } from '../actions/dispatch';
```
shakao, line 5: dispatchOpenActivity not used?

shakao (approved):

PR #7585, opened 2020-11-03
Profile page updates

eanders-MS:
* Remove "Request data" button
* Combine Privacy and Account panels
* Simplify CSS

### DIFF
--- a/webapp/src/user.tsx
+++ b/webapp/src/user.tsx
@@ -58,7 +58,6 @@ export class ProfileDialog extends auth.Component<ProfileDialogProps, ProfileDia
             >
                 <AccountPanel parent={this} />
                 <GitHubPanel parent={this} />
-                <PrivacyPanel parent={this} />
             </sui.Modal>
         );
     }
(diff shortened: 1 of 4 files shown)


webapp/src/user.tsx
```diff
@@ -77,8 +76,25 @@ class AccountPanel extends sui.UIElement<AccountPanelProps, {}> {
...
 
+    handleDeleteAccountClick = async () => {
+        const result = await core.confirmAsync({
+            header: lf("Delete Account"),
+            body: lf("You are about to delete your account. You can't undo this. Are you sure?"),
+            agreeClass: "red",
+            agreeIcon: "delete",
+            agreeLbl: lf("Delete my account"),
+        });
+        if (result) {
```
pelikhan, line 87: for this kind of really deadly action, we use to ask it a second time in TouchDevelop. And github asks to actually write a piece of text as well so you can't "butt-delete" your account.

eanders-MS: For sure. Plan is for the user to have to type the name of their account to enable the delete button.

darzu (approved):
LGTM

PR #7589, opened 2020-11-04
Attach root classes to minecraft modals

livcheerful:
fixes https://github.com/microsoft/pxt-minecraft/issues/2076

This fix is two parts:
1. Updating the HC less so that for inverted themed editors, the modal's outline is the text color (white)
2. Make the root's classes accessible so that we can attach them to the modals when we make them

I initially updated all the modals with new classes, but changed it so that only the modals we'll see in Minecraft have the classes. I can add them back tho if you guys think that'd be better

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -118,7 +118,6 @@ export class ProjectView
     chooseHwDialog: projects.ChooseHwDialog;
     prevEditorId: string;
     screenshotHandlers: ((msg: pxt.editor.ScreenshotData) => void)[] = [];
-    rootClasses: string[];
 
     private lastChangeTime: number;
     private reload: boolean;
@@ -130,6 +129,7 @@ export class ProjectView
     private loadingExample: boolean;
     private openingTypeScript: boolean;
     private preserveUndoStack: boolean;
+    private rootClasses: string[];
 
     // component ID strings
     static readonly tutorialCardId = "tutorialcard";
@@ -3219,6 +3219,16 @@ export class ProjectView
     ///////////////////////////////////////////////////////////
     ////////////             Dialogs              /////////////
     ///////////////////////////////////////////////////////////
+
+    // Classes to add to modals
+    createModalClasses(classes?: string): string {
+        const rootClassList = [
+            classes,
+            this.rootClasses.indexOf("flyoutOnly") != -1 ? "flyoutOnly": "",
+            this.rootClasses.indexOf("inverted-theme") != -1 ? "inverted-theme" : "",
+        ]
+        return sui.cx(rootClassList);
+    }
 
     showReportAbuse() {
         const pubId = (this.state.tutorialOptions && this.state.tutorialOptions.tutorialReportId)
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -118,6 +118,7 @@ export class ProjectView
     chooseHwDialog: projects.ChooseHwDialog;
     prevEditorId: string;
     screenshotHandlers: ((msg: pxt.editor.ScreenshotData) => void)[] = [];
+    rootClasses: string[];
 
     private lastChangeTime: number;
     private reload: boolean;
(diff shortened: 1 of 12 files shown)


webapp/src/app.tsx
```diff
@@ -3906,6 +3907,7 @@ export class ProjectView
             this.editor == this.textEditor && this.state.errorListState,
             'full-abs',
         ];
+        this.rootClasses = rootClassList;
```
jwunderl, line 3910: I'm a little hesitant on this because there are just so many classes this adds / wide variety of scenarios that make it a little hard to know we've tested everything. Could we pull out just the ones that we definitely need here into a list for now, and maybe see about expanding / cleaning it up later? I think that's invertedTheme, hc

webapp/src/createFunction.tsx
```diff
@@ -151,9 +151,9 @@ export class CreateFunctionDialog extends data.Component<ISettingsProps, CreateF
             }
         ];
         const types = this.getArgumentTypes().slice();
-
+        const classes = ["createfunction", this.props.parent.rootClasses.filter((e: string) => e != "full-abs").join(" ")].join(" ");
```
jwunderl, line 154:
```suggestion
        const classes = sui.cx(["createfunction", ...this.props.parent.rootClasses.filter((e: string) => e != "full-abs")]);
```

we have a helper for this https://github.com/microsoft/pxt/blob/70c8a663c90fc5a14fc5cb3a05343fee785644c7/webapp/src/sui.tsx#L32-L34

(along the lines of the comment above, though, I'd suggest making this a getter instead of just a property; `this.props.parent.rootClasses()` or something like that, keep the logic for getting all the classes in one place / minimize chance of someone mutating it by accident and breaking things)

jwunderl (commented):
Might want to check with daryl on persisting high contrast like this, seems like there might be some changes you'd want to make it have it hook into the theme he's adding / not maintain the hc state on the editor (though I suppose that might be made harder if the high contrast changes are going into the identity branch...). Maybe this classlist itself should get mounted as a virtual api / the modals should be data.Components and subscribe to the changes

webapp/src/app.tsx
livcheerful: yeah that sounds good! I think hc actually is handled in a different way, so I won't touch it here

webapp/src/extensions.tsx
```diff
@@ -289,7 +289,7 @@ export class Extensions extends data.Component<ISettingsProps, ExtensionsState>
         const actions: sui.ModalButton[] = action ? [{ label: action, onclick: actionClick }] : undefined;
         if (!needsConsent && visible) this.initializeFrame();
         return (
-            <sui.Modal isOpen={visible} className={`${needsConsent ? 'extensionconsentdialog' : 'extensiondialog'}`}
+            <sui.Modal isOpen={visible} className={(needsConsent ? "extensionconsentdialog" : "extensiondialog")}
```
jwunderl, line 292: did we need createModalClasses here?

jwunderl (approved):

webapp/src/extensions.tsx
livcheerful: oh yes I thought it'd be good to take it away from extensions since minecraft doesn't have extensions (right?)

jwunderl: Minecraft has extensions / you can still get to the dialog, it just doesn't have any approved extensions currently (which I think we want to change, we just need to find some time to think of some good ones to make and do it~)

livcheerful: oo cool cool tyty i'll add it back

livcheerful: okay so I added it to scriptsearch.tsx, which seems to be the modal that shows when we do the + extensions. I only dug a little into the modal where we ask for permissions - would you happen to know a case off the top of your head where that one shows up 😅 i will keep looking if not

livcheerful: (for future reference) to test this modal, you can use this extension in micro:bit https://github.com/samelhusseini/pxt-midi-tone

(for today) i'm gonna not update this modal since I don't think there's a minecraft use for it, but we can always add in the future

MicahM-MSFT:
Thanks! I'll have our QA team take another look once it hits live. Is it available in beta yet?

PR #7590, opened 2020-11-05
Cleaning up tutorial chrome in skills map frame

riknoll:
Note that this change requires the `skills-map` branch in pxt-arcade, which has the necessary changes to pxtarget.json.


shakao (approved):

shakao:
looks SO much slicker with the loader

PR #7594, opened 2020-11-05
Add PXT_TARGET #define in codal

mmoskal:
This is generic version of https://github.com/microsoft/pxt-microbit/pull/3597 for future use

### DIFF
--- a/pxtlib/cpp.ts
+++ b/pxtlib/cpp.ts
@@ -989,6 +989,8 @@ namespace pxt.cpp {
                 .forEach(k => optSettings["YOTTA_CFG_" + k] = optSettings[k]);
         }
 
+        optSettings["PXT_TARGET"] = JSON.stringify(appTarget.id)
+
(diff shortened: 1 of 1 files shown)


pxtlib/cpp.ts
```diff
@@ -989,6 +989,8 @@ namespace pxt.cpp {
                 .forEach(k => optSettings["YOTTA_CFG_" + k] = optSettings[k]);
         }
 
+        optSettings["PXT_TARGET"] = JSON.stringify(appTarget.id)
```
pelikhan, line 992: maybe have a optional field to override this? we have platformid already.

pelikhan (approved):

pxtlib/cpp.ts
mmoskal: I think platformid is usually "codal", so maybe not so much use here

abchatra:
Do I need to port this? Or is this just for future?

mmoskal:
Just for future i think

PR #7595, opened 2020-11-06
fix code stop / minecraft hour of code

jwunderl:
hot fix for hour of code tutorials

~I'm fairly confident this will work but waiting for my local build to test real quick, putting up now to get build started~ tested and appears to work, though that's just comparing the generated code / haven't got an easy way to test in game at the moment.

For reference, generated code with current release (important bit is relative position of `pxsim._control._onCodeStop` vs on start / main.ts code):

MicrosoftTeams-image (1)

vs with previous release

MicrosoftTeams-image (2)

and with this change:

MicrosoftTeams-image (3)

### DIFF
--- a/pxtcompiler/emitter/driver.ts
+++ b/pxtcompiler/emitter/driver.ts
@@ -133,10 +133,10 @@ namespace ts.pxtc {
         }
 
         // run post-processing code last, if present
-        const postIdx = opts.sourceFiles.indexOf("_onCodeStop.ts")
+        const postIdx = opts.sourceFiles.indexOf(pxt.TUTORIAL_CODE_STOP)
         if (postIdx >= 0) {
             opts.sourceFiles.splice(postIdx, 1)
-            opts.sourceFiles.push("_onCodeStop.ts")
+            opts.sourceFiles.push(pxt.TUTORIAL_CODE_STOP)
         }
 
         res.times["conversions"] = U.cpuUs() - startTime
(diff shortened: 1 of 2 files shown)


shakao:
change looks fine to me, but which part of the new release caused the behavior to change?

shakao (approved):

jwunderl:
the sorting pushing main.ts to the end was added in https://github.com/microsoft/pxt/pull/7332, previously it was just done on the `emitter/driver.ts` side

shakao:
shouldn't that have caused it to break for all hoc2019 tutorials and not just the fifth one?

jwunderl:
It was breaking for me on all of them (showing the 'failure' message) but inconsistent on whether that reset things / terminated the code and tried to restart - I couldn't get past the maze one personally.

abchatra:
I believe this particular tutorial has while <condition> code in the on start. That condition won't be true if the TUTORIAL_CODE_STOP is executed?

abchatra:
image

jwunderl:
Ahh you're right I think, hazards remain must return false after the 'code finished' setting ran

abchatra:
You can modify the world to link to /beta URL and see if the bug is fixed.

shakao:
code stop just executes a slash command specified in the markdown--i think for these tutorials it was just setting a scoreboard value to 1 and the world would listen for the scoreboard change then check if the task was successfully completed. i'm a little surprised it even worked for "open door" since it was running before the "open door" block would run, but maybe there's some time delay when they check for success/failure...

jwunderl:
I realized I had a debug build loaded up that I could use to test, can confirm this seems to fix it.

PR #7598, opened 2020-11-06
Keep track of hints we've seen

livcheerful:
fixes https://github.com/microsoft/pxt-minecraft/issues/2080

Keep track of all the hints we've seen so we don't jiggle the hint button if it's already been used.
the path is basically:

**hintManager** is in charge of the wiggling and also keeps track of who has/hasn't been seen
**app** prompts **hintManager** to start the timer to jiggle
**tutorial** tells **app** when we've seen a hint who then passes it along to **hintManager**

This plays nicely with having multiple tutorials


jwunderl (approved):
Looks good!

MicahM-MSFT:
Thanks!

PR #7604, opened 2020-11-10
fix for #7602, and address feedback

### LINKED ISSUE #7602: All galleries are loaded from the cloud on localhost.
Try loading the editor in localhost and see the URL of galleries. It's downloading from makecode.com.

image

Due to https://github.com/microsoft/pxt/commit/ed6180765c67f822b312eb3ea27f6e4b8faf4a3b#commitcomment-44040161

eanders-MS:
Reverted changes in:
* cloud.ts
* util.ts
* fileworkspace.ts

Moved `BACKEND` selection to auth.ts. This gives backend devs flexibility to run against localhost or not.

fixes #7602

### DIFF
--- a/webapp/src/auth.ts
+++ b/webapp/src/auth.ts
@@ -349,14 +349,14 @@ type ApiResult<T> = {
     errmsg: string;
 };
 
-const BACKEND_DEFAULT = "";
-const BACKEND_PROD = "https://www.makecode.com";
-const BACKEND_STAGING = "https://staging.pxt.io";
+const DEV_BACKEND_DEFAULT = "";
+const DEV_BACKEND_PROD = "https://www.makecode.com";
+const DEV_BACKEND_STAGING = "https://staging.pxt.io";
 // tslint:disable-next-line:no-http-string
-const BACKEND_LOCALHOST = "http://localhost:5500";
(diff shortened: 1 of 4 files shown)


webapp/src/auth.ts
```diff
@@ -343,14 +349,23 @@ type ApiResult<T> = {
     errmsg: string;
 };
 
+const BACKEND_DEFAULT = "";
+const BACKEND_PROD = "https://www.makecode.com";
```
pelikhan, line 353: to be friendly to OSS, a 3rd party should be able to override this. Our API used ``/api/`` to allow this.

pelikhan: (something to ponder about)

eanders-MS: In the prod environment the url is relative and will hit whatever backend the site was served from. This is a dev-only override.

eanders-MS: I scoped it to only apply when running on localhost/dev.
https://github.com/microsoft/pxt/pull/7604/files/ca9311a75f5f0aa58f911043a42544b2708ae7c3..e5bea850a769873bda22234bc26f4c967a27fc34

pelikhan (approved):

PR #7606, opened 2020-11-10
Tracking progress in the skill map

riknoll:
Fixes a few bugs around saving projects and also actually tracks the activity progress/completion. Still need to hook up the completion modal and add the debug hooks for marking activities as completed


skillmap/src/components/SkillCard.tsx
```diff
@@ -62,6 +70,9 @@ export class SkillCardImpl extends React.Component<SkillCardProps> {
                 <div className="skill-card-label">
                     <div className="skill-card-title">
                         {(status === "locked" || status === "completed") && <i className={`icon ${status === "locked" ? "lock" : "check circle"}`} />}
+                        {(status === "inprogress" || status === "notstarted") && maxSteps &&
```
shakao, line 73: so "notstarted" shows 0/maxSteps?

skillmap/src/styles/skillcard.css
```diff
@@ -182,6 +182,16 @@
     vertical-align: bottom;
 }
 
+.circular-label {
+	background: #656565;
```
shakao, line 186: small: tabs instead of spaces?

shakao (approved):

skillmap/src/styles/skillcard.css
riknoll: HOW darn it.

skillmap/src/components/SkillCard.tsx
riknoll: oh no it doesn't but this is definitely confusing so I'll change it!

shakao: ohh notstarted doesn't have maxsteps, i see

Issue #7258, opened 2020-06-29
groups attribute is not properly processed


FranklinWhale:
**Describe the bug**

According to https://makecode.com/defining-blocks#category,

> To define your groups, add the groups attribute to your namespace. The groups attribute is an array of group names.

> The order in which you define your groups is the order in which the groups will appear in the toolbox flyout

, and it has a sample syntax of the `groups` attribute:

```ts
//% groups=['LED matrix', 'Control flow', 'others']
```

Unfortunately, that does work.

On https://makecode.com/playground#basic-groups, another syntax is shown:

```ts
//% groups="['Fruits', 'Veggies']"
```

Sadly, that does not work either.

The only syntax that appears to work is as follows:

```ts
//% groups='["Fruits", "Veggies"]'
```

I think either the above samples should amended, or the parser should support all of them.

**To Reproduce**
1. Go to https://makecode.com/playground#basic-groups
2. Click "Run"
3. Replace `//% groups="['Fruits', 'Veggies']"` with `//% groups=['Fruits', 'Veggies']`
4. Click "Run"
5. Replace `//% groups=['Fruits', 'Veggies']` with `//% groups=["Fruits", "Veggies"]`
6. Click "Run"
7. Replace `//% groups=["Fruits", "Veggies"]` with `//% groups='["Fruits", "Veggies"]'`
8. Click "Run"

**Actual behavior**
After Step 2, 4 and 6:
![image](https://user-images.githubusercontent.com/8416067/85979774-b14d4e80-ba13-11ea-97ca-bc4d19ab48e6.png)

After Step 8:
![image](https://user-images.githubusercontent.com/8416067/85980073-45b7b100-ba14-11ea-9ab3-fdd3de6ee960.png)

pelikhan:
@ganicke looks like a docs issue.

ganicke:
@pelikhan - It seems that the first **member** of the namespace with its `group` attribute set determines the display order in the Toolbox and not the order set for `groups`. The playground example below will cause the `Fruits` group to display first despite the order of `groups`.

```typescript
/**
 * Organize your blocks in groups
 */
//% color="#AA278D"
//% groups="['Veggies', 'Fruits']"
namespace food {
    //% block
    //% group="Fruits"
    export function apple() {

    }
    //% block
    //% group="Veggies"
    export function potato() {

    }
    //% block
    //% group="Fruits"
    export function banana() {

    }
    //% block
    //% group="Veggies"
    export function bean() {

    }
[...]

PR #7609, opened 2020-11-10
Clarify group ordering better in 'Defining blocks'

ganicke:


pelikhan:
@aznhassan did you look at the group logic lately?

FranklinWhale:
I think this PR may need to be amended: https://github.com/microsoft/pxt/issues/7258#issuecomment-728755161

PR #7612, opened 2020-11-11
Save user state

riknoll:
Saves the user state locally and also adds a couple debug flags:

1. `debugNewUser=1` will get you a new user with no progress (does not overwrite local user)
2. `debugCompleted=1` will mark all tutorials in the map as completed (also does not overwrite local user)


skillmap/src/lib/browserUtils.ts
```diff
@@ -11,6 +11,24 @@ export function parseHash() {
     return hash;
 }
 
+export function parseQuery() {
+    const out: {[index: string]: string} = {};
+    const query = window.location.search;
+
+    if (query) {
+        const params = new URLSearchParams(query);
```
shakao, line 19: whoa didn't know about this, very cool

shakao (approved):

PR #7614, opened 2020-11-12
win10 flash: make sure to reconnect dap wrapper

pelikhan:
We need to run reconnectAsync after detecting a new device so that we can detect which hardware is connected.
+ increase timeout for winrt deployment.


pelikhan:
Needs to be ported in stable6.0

PR #7615, opened 2020-11-12
show all links for multivariant editors

pelikhan:
Display links to each compile service in multi-variant editors.

image

### DIFF
--- a/webapp/src/dialogs.tsx
+++ b/webapp/src/dialogs.tsx
@@ -15,8 +15,6 @@ export function showAboutDialogAsync(projectView: pxt.editor.IProjectView) {
     const githubUrl = pxt.appTarget.appTheme.githubUrl;
     const targetTheme = pxt.appTarget.appTheme;
     const versions: pxt.TargetVersions = pxt.appTarget.versions;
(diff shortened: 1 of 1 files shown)


pelikhan:
Worth adding to stable6.0 so that we can easily find the CODAL repoes.

webapp/src/dialogs.tsx
```diff
@@ -68,7 +74,7 @@ export function showAboutDialogAsync(projectView: pxt.editor.IProjectView) {
 }
 
 
-function renderCompileLink(cs: pxt.TargetCompileService) {
+function renderCompileLink(variantName: string, cs: pxt.TargetCompileService) {
```
jwunderl, line 77: Is `variantName` intended to be used anywhere in here? Seems fine as it's taking the info from the cs

jwunderl (approved):

webapp/src/dialogs.tsx
pelikhan: right, i taught i would never but eventually i didn't. The repo description is actually enough.

mmoskal (approved):

livcheerful (approved):

abchatra:
I will not port this. At this point only absolutely necessary fixes will be ported. I will take this post release as a patch.

PR #7616, opened 2020-11-12
add apptheme field to use data uri instead of blob object

jwunderl:
re: https://github.com/microsoft/pxt-microbit/issues/3674, will toggle this on with https://github.com/microsoft/pxt-microbit/blob/master/pxtarget.json#L563 (and if needed for iosapp)

### DIFF
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -405,6 +405,7 @@ declare namespace pxt {
         identity?: boolean; // login with identity providers
         assetEditor?: boolean; // enable asset editor view (in blocks/text toggle)
         disableMemoryWorkspaceWarning?: boolean; // do not warn the user when switching to in memory workspace
+        disableBlobObjectDownload?: boolean; // use data uri downloads instead of object urls
     }
 
     interface SocialOptions {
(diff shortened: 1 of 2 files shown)


localtypings/pxtarget.d.ts
```diff
@@ -405,6 +405,7 @@ declare namespace pxt {
         identity?: boolean; // login with identity providers
         assetEditor?: boolean; // enable asset editor view (in blocks/text toggle)
         disableMemoryWorkspaceWarning?: boolean; // do not warn the user when switching to in memory workspace
+        disableBlobObjectDownload?: boolean; // use data uri downloads instead of object urls
```
pelikhan, line 408: why is this a flag for an entire editor, isn't this browser/host specific?

jwunderl: It is host specific; I was going to use the query variants to toggle it, but if preferred can just change this to check for the query parameters in the browserdownload function?

pelikhan: would be best to detect this at runtime one way or the other i guess. This is for electron/ios?

jwunderl: for the android / probably ios mobile apps; for android users going through the browser we still would want the blob object download, it's just not as accessible to the app to intercept / store that way

abchatra (approved):

PR #7619, opened 2020-11-12
Fix IndexedDB reset/clear, update global hash when loading new activity

shakao:
- The "clearDB" promises for the TutorialInfo and ApiInfo were never resolving, so resetAsync would never reload the editor. I'm guessing something about the updated CEF changed the behavior of this; I've updated the clear functions to delete the table instead of the entire DB whenever possible.
- The global "hash" variable is not updated when we navigate between activities (page isn't reloaded), so in the Minecraft HoC case reloadEditor will always reload into the first tutorial; I don't know if any parts of our code depend on this behavior so I'm only updating it in the case of a new activity which should be fairly safe.

Fixes https://github.com/microsoft/pxt-minecraft/issues/2093

### DIFF
--- a/pxtlib/browserutils.ts
+++ b/pxtlib/browserutils.ts
@@ -975,6 +975,7 @@ namespace pxt.BrowserUtils {
     export interface ITutorialInfoDb {
         getAsync(filename: string, code: string[], branch?: string): Promise<TutorialInfoIndexedDbEntry>;
         setAsync(filename: string, blocks: Map<number>, code: string[], branch?: string ): Promise<void>;
+        clearAsync(): Promise<void>;
     }
 
     class TutorialInfoIndexedDb implements ITutorialInfoDb {
@@ -1048,6 +1049,14 @@ namespace pxt.BrowserUtils {
                     pxt.perf.measureEnd("tutorial info db setAsync")
                 })
         }
+
+        clearAsync(): Promise<void> {
+            return this.db.deleteAllAsync(TutorialInfoIndexedDb.TABLE)
+                .then(() => console.debug(`db: all clean`))
+                .catch(e => {
+                    console.error('db: failed to delete all');
+                })
+        }
     }
 
     let _tutorialInfoDbPromise: Promise<TutorialInfoIndexedDb>;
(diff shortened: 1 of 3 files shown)


riknoll (approved):

jwunderl (approved):

livcheerful (approved):

abchatra (approved):

MicahM-MSFT:
Thanks @shakao! Can we test this in /beta now?

abchatra:
Yes available in /beta

MicahM-MSFT:
@shakao @abchatra maybe I'm doing something wrong, but I'm in the Hour of Code 2020 world that I attached to the other bug (with the reset button added), I added a command block pointed to the beta URL: /codebuilder navigate @p true https://minecraft.makecode.com/beta?ingame=1&ipc=1&norunonx=1#tutorial:/hour-of-code/2020/fishing and the reset button still doesn't do anything.

shakao:
hmm, taking a look now!

shakao:
@MicahM-MSFT aha we were a couple updates behind on /beta, but if you try again now it should work!

MicahM-MSFT:
@shakao The reset button is working now and is clearing the whole cache, but in the HOC world I am missing the warning dialog telling me I'm about to reset my whole cache and lose my projects:
image

shakao:
@MicahM-MSFT yes, that's intentional--the thought was that hour of code projects are temporary anyway so we directly reset all the state. do you want the warning dialog? it's a straightforward change i think

abchatra:
Question is does it just delete hour of code projects or all other projects as well?

MicahM-MSFT:
@shakao this option will appear in both the library and the demo and in the future will likely appear in more places as well. If the reset button is only clearing content for this one world, we could leave it as is for now, but I believe this is clearing everything, so we should warn the users.

shakao:
sounds good! it does clear all the content on the device (hard reset) so that makes sense

MicahM-MSFT:
Cool, let me know when the update is on beta and I'll send it over to QA team for some more in-depth testing.

shakao:
@MicahM-MSFT updated in /beta!

MicahM-MSFT:
Thanks! Confirmed working.

PR #7620, opened 2020-11-12
shuffle in-context

pelikhan:
Updated on the in-context translation page

### DIFF
--- a/docs/translate.md
+++ b/docs/translate.md
@@ -14,25 +14,6 @@ Just need a quick answer to a translation question? You can check the [FAQ](#faq
 
 ### ~
 
-## Crowdin project
-
-The Crowdin translation project below contains both the resources and menu items from @homeurl@.
-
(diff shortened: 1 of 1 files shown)


ganicke (approved):

docs/translate.md
```diff
@@ -58,8 +58,8 @@ You will be prompted with a dialog that contains the block translatable string.
 
 ### Documentation page translation
 
-In any documention page, click on the **Language** button 
-in the footer; then select translation this page in the dialog.
+When on any documention page, click on the **Language** button
```
pelikhan, line 61: there seems to be a problem with this feature. it works on localhost but not in the cloud. The current way to do this is to add ?translate=1 on the page

ganicke: Add a note for that?

pelikhan: let's see if it's a quick fix first

PR #7627, opened 2020-11-16
[stable6.0] support for Markdown codecards

pelikhan:
- The markdown-based format is optimized for crowdin and make its much easier to localize. 
- The changes are easily testable: if the gallery render in the home screen and on the docs pages; they still work
- We have a CLI too to batch convert all codecards into markdown once this is live.

### DIFF
--- a/docfiles/macros.html
+++ b/docfiles/macros.html
@@ -49,6 +49,10 @@
     </div>
 </aside>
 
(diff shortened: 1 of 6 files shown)


pelikhan:
@abchatra

abchatra (approved):
Looks good to me. Will test home screen before release.

PR #7629, opened 2020-11-16
Locked skill maps

riknoll:
Add support for maps with prerequisites!

Screen Shot 2020-11-16 at 2 54 01 PM


riknoll:
This also changes the user state so that completed tags are indexed on the fetched page url

shakao (approved):

PR #7633, opened 2020-11-17
Settings dropdown with report abuse option

shakao:
Going to have to do a merge dance with the telemetry PR but I'll cross that bridge when I get to it.

Adds a settings gearwheel which has a dropdown with one option: 
image

Which pops up the report abuse modal:
image

Which sends a POST request to `/api/[URI encoded pageSourceURL/abusereports` with `{ text: [user description] }` in the body. The localhost request looks correctly formatted to me, but I'm not quite sure how to test that it's being received on the other end. (I also haven't sent any real abuse reports because I didn't want to trigger a false Avert warning?)


abchatra:
Can we reuse the existing dialog for consistency? 

image

riknoll (approved):

shakao:
@abchatra that dialog is pretty tied into the pxt codebase (the skill map doesn't have access to the modal component or the pxt cloud util functions right now). i could use something like the share page report abuse:

image

but either case would require changes to the modal, since it expects to be reporting a shared project id and the skills map would be reporting the markdown url

abchatra:
@shakao sounds good. Nit pick: May be move "Cancel" to "x" at the top?

PR #7634, opened 2020-11-17
Dont mutate type node during completions of parameters with type 'any'

jwunderl:
fix https://github.com/microsoft/pxt-minecraft/issues/2094

When a symbol with 'any' type had a default value, this was mutating parameter node / changing it from `any` to the type of the default value. So in minecraft, currently `player.say`'s block can take numbers / any, but if you switch to typescript and happen to get a completion generated for `player.say` the apiInfo will get changed / the block will no longer accept numbers or other values until you refresh.

### DIFF
--- a/pxtcompiler/emitter/snippets.ts
+++ b/pxtcompiler/emitter/snippets.ts
@@ -281,25 +281,26 @@ namespace ts.pxtc.service {
 
             // check for explicit default in the attributes
             if (attrs && attrs.paramDefl && attrs.paramDefl[name]) {
+                let deflKind: SyntaxKind;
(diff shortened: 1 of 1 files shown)


abchatra:
Add a test? Also what change regressed this?

jwunderl:
https://github.com/microsoft/pxt/pull/7322, I'll see about adding a test - a bit tricky / might require adding more to the language-service tests to be able to check something like this, as currently they just check the completions that get outputted.

abchatra (approved):

livcheerful (approved):

PR #7637, opened 2020-11-18
show vars with non-ts-identifier chars in debug view

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/2573

Worth noting this would show the variables with invalid chars escaped:

image

as that's how they're stored in the simulator; if we want to, we could defer escaping the name till the filter is used https://github.com/microsoft/pxt/blob/d1d8cd45952216d88c79130b4c04bc926586b281/webapp/src/debuggerVariables.tsx#L189 and then use the name in `filters` instead of `updatedGlobals.variables` when displaying the variable? Not sure it's necessary so I left it as the simple fix for now

### DIFF
--- a/webapp/src/blocks.tsx
+++ b/webapp/src/blocks.tsx
@@ -986,7 +986,8 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         if (!this.parent.state.debugging) return;
 
         if (this.debuggerToolbox) {
-            const visibleVars = Blockly.Variables.allUsedVarModels(this.editor).map((variable: any) => variable.name as string);
+            const visibleVars = Blockly.Variables.allUsedVarModels(this.editor)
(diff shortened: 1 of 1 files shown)


riknoll:
Why are we making this change? Shouldn't we show them as they appear in the Blockly program?

jwunderl:
Right now they're not showing up at all -- the  `visibleVars` filter has them unescaped, and we're checking for matches against names that were escaped when used in the sim here

https://github.com/microsoft/pxt/blob/d1d8cd45952216d88c79130b4c04bc926586b281/webapp/src/debuggerVariables.tsx#L187-L189

We could see about using the filter list to swap back to the original names when displaying the variables -- I'm not sure we have the context currently to accurately map to renames (though I did just check and any variables that get a number appended during the blockly compile go missing for a similar reason to this change; I think that would mean we will need to get the scopes / variable bindings from the blocklycompiler to fully fix it anyways)

riknoll (approved):

PR #7639, opened 2020-11-18
Clear editor filters when opening a new empty project

shakao:
We carryover filters from the previous editor state in internalLoadHeaderAsync which means if you call newEmptyProject from inside a tutorial, the toolbox filters are not cleared. 

@abchatra can we take this as a patch fix for Minecraft? i think it will help the Prodigy Learning folks with the tutorial thing they're doing


jwunderl (approved):

abchatra (approved):

abchatra:
Sounds good. Port to stable6.4 and bump

PR #7643, opened 2020-11-18
Merging skills-map into master

riknoll:
Merging the skills-map branch into master. Should be pretty safe.

Most of the changes here are contained within the skillmap/ directory, but there are a few changes to pxt. All of these have been review by @shakao or I.


cli/cli.ts
```diff
@@ -439,6 +439,8 @@ function ciAsync() {
                             .then(() => crowdin.execCrowdinAsync("upload", "built/webstrings.json"));
                     if (uploadApiStrings)
                         p = p.then(() => crowdin.execCrowdinAsync("upload", "built/strings.json"))
+                            // TODO: uncomment once verified that this is correct
```
jwunderl, line 442: RE: todo here, this looks like it'd be adding to pxt-core's build, but these would typically be specific to targets, right? I believe that'd be here ish: 

https://github.com/microsoft/pxt/blob/f5a9eab92148aae6e045ae31b03aaaba708d2503/cli/crowdin.ts#L57

skillmap/src/App.tsx
```diff
@@ -0,0 +1,294 @@
...
+        e.preventDefault();
+    }
+
+    protected async initLocalizationAsync() {
+        const theme = pxt.appTarget.appTheme;
+
+        const href = window.location.href;
+        let force = false;
+        let useLang: string | undefined = undefined;
+        if (/[&?]translate=1/.test(href) && !pxt.BrowserUtils.isIE()) {
```
jwunderl, line 75: sidenote: I see this was just taken from another spot, but we're probably safe removing all the `pxt.BrowserUtils.isIE` references in the codebase, right? Seems fine to leave in here for now for consistency sake though~

skillmap/src/App.tsx
```diff
@@ -0,0 +1,294 @@
...
+
+function mapStateToProps(state: SkillMapState, ownProps: any) {
+    if (!state) return {};
+    return {
+        skillMaps: state.maps,
+        activityOpen: !!state.editorView
+    };
+}
+
+async function updateLocalizationAsync(targetId: string, baseUrl: string, code: string, pxtBranch: string, targetBranch: string, live?: boolean, force?: boolean) {
```
jwunderl, line 255: nit: I'm assuming we're duping this portion to avoid refactoring it into in pxtlib/util's version? Maybe worth adding a note here and in the original definition pointing to eachother / that if we change anything in the structure of localizations (beginning / end of fn) we should propagate it to both sections

skillmap/src/lib/indexedDBWrapper.ts
```diff
@@ -0,0 +1,125 @@
+// Duplicated from pxtlib/browserutils.ts
+// TODO: Move to common location
+
+
+// IndexedDB wrapper class
+export type IDBUpgradeHandler = (ev: IDBVersionChangeEvent, request: IDBRequest) => void;
+
+export class IDBWrapper {
```
jwunderl, line 8: nit: since browserutils is being copied over now, this file can go away, right?

jwunderl (approved):
LGTM to merge ~

(None of my comments are blocking / need to be addressed right away)

cli/cli.ts
riknoll: This is for the skill map UI, which isn't specific to targets (just like the core pxt strings)

skillmap/src/lib/indexedDBWrapper.ts
shakao: i think we're doing a separate pass to dedup everything (there are a couple functions in the skillmap/lib/browserutils that also need to be removed)

livcheerful (approved):

PR #7645, opened 2020-11-19
try adding UTF-8 BOM on sha miss

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/1999

This seems to behave fine to me? Causes the project to load properly (both the one from the issue and the other one I ran into today), and doesn't show the files in the diff. I suppose we'd need to check for utf16 big / little endian BOM as well if we wanted to be fully thorough, but notepad produces UTF-8 + BOM which I believe was the issue here / before as well?

### DIFF
--- a/webapp/src/workspace.ts
+++ b/webapp/src/workspace.ts
@@ -919,8 +919,11 @@ async function githubUpdateToAsync(hd: Header, options: UpdateOptions) {
         }
         text = await pxt.github.downloadTextAsync(parsed.fullName, sha, path)
         treeEnt.blobContent = text
-        if (gitsha(text) != treeEnt.sha)
+        if (gitsha(text) != treeEnt.sha
+                // Try adding UTF-8 BOM; this can be stripped over web requests
+                && gitsha(`\uFEFF${text}`) !== treeEnt.sha) {
             U.userError(lf("Corrupt SHA1 on download of '{0}'.", path))
+        }
         if (options.tryDiff3 && hasChanges) {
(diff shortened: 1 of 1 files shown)


pelikhan (approved):
LGTM

mmoskal (approved):
Maybe you want to add BOM to `text` when the sha matches that? I guess if diffing works without this, maybe we're fine. Also BOM may confused TypeScript or Blockly.

Also there's only one UTF-8 BOM. We do not support UTF16 or UTF32 so no need to worry about other BOMs.

jwunderl:
Right, I was thinking about adding the BOM but wasn't certain how the rest of the app would handle it; it doesn't show in the diffs on our end as we just compare against the text, though when the user makes the first commit after pulling will clear all the BOMs / list them as changed in the log if that's a concern (Potentially for merge conflicts?):

https://github.com/jwunderl/test-bom-clone/commit/e78f50db29e26d2c27e1d36ae0c1057d5d8f74c4

pelikhan:
As long as our merge algo is smart about those BOM changes we're fine. Did you try this any this? This is a cool fix.

jwunderl:
Okay, tested out a few merges and looks like we handle everything nicely that I saw: https://github.com/jwunderl/test-bom-clone

As far as I can tell, any place that we merge things we will already be comparing with the text on the tree, which wouldn't have the bom anyways; so as is, we'll be handling any merge with a bom by just stripping it. This seems fine / like the right behavior to make sure everything works in the editor to me~

PR #7649, opened 2020-11-20
Embed Usabilla widget on skill map page

shakao:
image

### DIFF
--- a/skillmap/public/index.html
+++ b/skillmap/public/index.html
@@ -17,5 +17,11 @@
     <script type="text/javascript" src="/pxtlib.js"></script>
 
     <div id="root"></div>
+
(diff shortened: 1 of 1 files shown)


skillmap/public/index.html
```diff
@@ -17,5 +17,11 @@
...
+    window.usabilla_live = lightningjs.require("usabilla_live", "//w.usabilla.com/1bba04d66c15.js");
```
riknoll, line 23: Not sure this is compliant, we might need to host this js ourselves.

riknoll: Never mind, disregard!

riknoll (approved):

PR #7654, opened 2020-11-23
 porting over skills map changes to stable 6.2

riknoll:
Note that I am only porting the pxt changes and not the skillmap itself to avoid any confusion. The skillmap should be built using master


riknoll:
Hm.... on second thought, maybe i should port over the skillmap stuff too?

riknoll:
Okay, updated to move over all skillmap changes. Talked to @shakao and decided we'd probably need it for the eventual build changes.

abchatra (approved):

PR #7655, opened 2020-11-23
Dont lose tick events, attach target

jwunderl:
setting up window.appInsights.trackEvent is async, so it's possible for tick events to be fired before that gets loaded; in particular this happens for https://github.com/microsoft/pxt-arcade/blob/master/docfiles/script.html#L174 on my machine.

This queues the events in that case, and checks every second until trackEvent is loaded / fires off all the ticks at that point.

Also updated the tick posting portion so https://github.com/microsoft/pxt-arcade/pull/2587 would be unnecessary (though not really sure if we want cookie = true for sure? Not sure where it landed on whether we should be setting that or not now that when the cookie banner is gone).

### DIFF
--- a/docfiles/tickevent.html
+++ b/docfiles/tickevent.html
@@ -18,7 +18,7 @@
                     }, 1000);
                 }
 
-                if (queuedTicks.length >= 10) {
+                if (queuedTicks.length >= 100) {
                     queuedTicks.shift();
                 }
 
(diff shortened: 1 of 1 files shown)


abchatra (approved):
I would put a max limit on the queuedTicks in case application insights is blocked (as some extensions do that) and we don't exhaust the memory.

abchatra (approved):
I would keep 100 or even 1000. 100 is still small :-).

PR #7658, opened 2020-11-23
Adjust tutorial 'show see more' when window resizes

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/2462

Decide if we wanna show the "More/Less" button every time the window resizes. I also changed the check so that if the tutorial text scroll size and its container have the same size, we still show the button (on my window the bottom part of the text was being cut off)


webapp/src/tutorial.tsx
```diff
@@ -433,12 +434,19 @@ export class TutorialCard extends data.Component<TutorialCardProps, TutorialCard
         }
     }
 
+    private handleResize() {
+        const options = this.props.parent.state.tutorialOptions;
+        this.setShowSeeMore(options.autoexpandStep);
```
riknoll, line 439: should probably throttle or at least use requestAnimationFrame()

livcheerful: good point good point I added a 500ms debounce time - do you think that's enough?

webapp/src/tutorial.tsx
```diff
@@ -433,12 +435,22 @@ export class TutorialCard extends data.Component<TutorialCardProps, TutorialCard
         }
     }
 
+    private handleResize() {
+        const options = this.props.parent.state.tutorialOptions;
+        if (this.resizeDebounceTimer) {
+            clearTimeout(this.resizeDebounceTimer);
+        }
+        this.resizeDebounceTimer = setTimeout(()=>(this.setShowSeeMore(options.autoexpandStep)), 500);
```
pelikhan, line 443: we have a helper to do debouncing https://github.com/microsoft/pxt/blob/d57cb0b70b95b552c26e8e8cf8d8685d2b7b1c82/pxtlib/util.ts#L470

livcheerful: ooooooooooo great I'll use that! tytytyty

webapp/src/tutorial.tsx
```diff
@@ -433,12 +434,19 @@ export class TutorialCard extends data.Component<TutorialCardProps, TutorialCard
...
+    }
+
     componentDidMount() {
         this.setShowSeeMore(this.props.parent.state.tutorialOptions.autoexpandStep);
+        window.addEventListener('resize', pxt.Util.debounce(this.handleResize, 500));
     }
 
     onMarkdownDidRender() {
         this.setShowSeeMore(this.props.parent.state.tutorialOptions.autoexpandStep);
+        window.removeEventListener('resize', pxt.Util.debounce(this.handleResize, 500));
```
pelikhan, line 449: debounce creates a new function instance so this won't remove the one added in componentDidMount. Store the debounce function in componentDidMount and remove that one.

pelikhan (changes requested):

webapp/src/tutorial.tsx
livcheerful: ahhh yeah that makes sense. Thanks, Peli!

pelikhan:

webapp/src/tutorial.tsx
```diff
@@ -315,6 +315,7 @@ interface TutorialCardProps extends ISettingsProps {
 export class TutorialCard extends data.Component<TutorialCardProps, TutorialCardState> {
     private prevStep: number;
     private cardHeight: number;
+    private resizeDebouncer: any;
```
pelikhan, line 318: don't use any

pelikhan (dismissed):

livcheerful:
> What is the status of this PR?

it'll be good to go after review!

webapp/src/tutorial.tsx
```diff
@@ -433,12 +435,20 @@ export class TutorialCard extends data.Component<TutorialCardProps, TutorialCard
...
+
     componentDidMount() {
         this.setShowSeeMore(this.props.parent.state.tutorialOptions.autoexpandStep);
+        this.resizeDebouncer = pxt.Util.debounce(this.handleResize, 500);
+        window.addEventListener('resize', this.resizeDebouncer);
     }
 
     onMarkdownDidRender() {
         this.setShowSeeMore(this.props.parent.state.tutorialOptions.autoexpandStep);
+        window.removeEventListener('resize', this.resizeDebouncer);
```
jwunderl, line 451: Should this one be in `componentWillUnmount`? Removing the event listener in `onMarkdownDidRender` seems like it wouldn't fix the issue as it would stop resizing when the page has loaded

livcheerful: oh yeah you're right thanks joey! It was rudely working, fooling me!

jwunderl (approved):

PR #7659, opened 2020-11-24
Add new telemetry properties for isProduction and interactiveConsent (renaming "cookie")

shakao:
Previously we were setting "cookie" to "includeCookie" which corresponds to whether or not we dropped a cookie, and not to the "interactiveConsent" flag (https://github.com/microsoft/pxt/pull/7494/files#diff-8957ba59dbfacd963c66897334cdb7040d8a10b172f100090eb857c2e7e28245L22)

We could revert to this behavior (set cookie directly equal to "isProduction" here: https://github.com/microsoft/pxt/blob/master/docfiles/pxtweb/cookieCompliance.ts#L220), but I think it's more correct to automatically set "interactiveConsent" to true for native apps (since we automatically trust them). 

@abchatra thoughts?

### DIFF
--- a/docfiles/pxtweb/cookieCompliance.ts
+++ b/docfiles/pxtweb/cookieCompliance.ts
@@ -216,9 +216,12 @@ namespace pxt {
 
         // "cookie" does not actually correspond to whether or not we drop the cookie because we recently
         // switched to immediately dropping it rather than waiting. Instead, we maintain the legacy behavior
-        // of only setting it to true for production sites where interactive consent has been obtained
-        // so that we don't break legacy queries
-        telemetryItem.properties["cookie"] = interactiveConsent && isProduction;
+        // of only setting it to true for production sites so that we don't break legacy queries
+        telemetryItem.properties["cookie"] = isProduction;
+
+        // "interacted" corresponds to whether the user has interacted meaningfully with the site (ie
+        // the "interactiveConsent" flag is set to true). This is automatically set for native apps.
+        telemetryItem.properties["interacted"] = interactiveConsent && isProduction;
     }
 
     export function setInteractiveConsent(enabled: boolean) {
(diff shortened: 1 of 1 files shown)


abchatra:
Probably we need to discuss with @Jaqster I like new behavior but unfortunately it causes issue with the consistency with legacy metrics.

shakao:
@abchatra actually, better idea--we can keep the cookie behavior as it was before, since i think we do still need some kind of flag for discarding forked editor ticks, and add a new custom metric that is "interactiveConsent" = true (or some more descriptive name), for if we want to filter on this new metric?

abchatra:
@shakao that sounds good. I though @riknoll had a interactiveconset as  a new tick just for this.

riknoll:
Yes I do remember adding that

shakao:
okay so here is the updated behavior:
1. cookie = true if "includeCookie" is true (anywhere we request that the cookie be dropped, regardless of the "interactiveConsent" flag)
2. interacted = true (<- can rename this) if cookie is dropped (production sites & electron) and the user meaningfully interacts with the page
3. interactiveConsent set to true automatically for native apps (minecraft, electron, etc)

@abchatra

Jaqster:
A few questions:
- What's the difference between interacted and interactiveConsent?  Don't they mean the same thing?  Just that for apps we're setting the flag automatically?
- What's the new cookie behavior again?  We drop it automatically on page load?
- How is this going to affect our MAU queries?

abchatra:
This needs a quick sync for us to discuss.

abchatra:
@shakao your proposal is good for me. Lets discuss in a 15 min sync tomorrow?

shakao:
sounds good, i can do a quick chat whenever!

abchatra (approved):

PR #7669, opened 2020-11-30
Don't save non-tutorial headers and wait until load is complete

riknoll:
Hopefully fixes https://github.com/microsoft/pxt-arcade/issues/2595 and https://github.com/microsoft/pxt-arcade/issues/2597

I was finally able to repro this by turning on the maximum cpu throttling in Chrome.

The other part of the change is to make it so that we wait until the activity/project is started instead of using the editor load event because when it's slow the editor loads an empty project and *then* loads the actual thing.

### DIFF
--- a/skillmap/src/components/makecodeFrame.tsx
+++ b/skillmap/src/components/makecodeFrame.tsx
@@ -144,6 +144,10 @@ class MakeCodeFrameImpl extends React.Component<MakeCodeFrameProps, MakeCodeFram
                 });
             }
         }
+
+        if (original.action === "importproject" || original.action === "startactivity") {
+            this.onEditorLoaded();
+        }
     }
 
     protected sendMessage(message: any, response = false) {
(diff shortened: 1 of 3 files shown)


skillmap/src/components/makecodeFrame.tsx
```diff
@@ -188,7 +192,9 @@ class MakeCodeFrameImpl extends React.Component<MakeCodeFrameProps, MakeCodeFram
             }
         };
 
-        await saveProjectAsync(project);
+        if (project.header.tutorial || project.header.tutorialCompleted) {
```
shakao, line 195: is there a way to also check if the activity we're currently in is of "tutorial" type? otherwise maybe add a comment that this will have to be updated if/when we generalize the types of activities

riknoll: Added to the props

shakao (approved):

PR #7672, opened 2020-12-01
Handle null tile nicely

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/2568

The issue was trying to render any hints with a tilemap block in it, and we were trying to access things from a null value :( I'm assuming the null tile is just the transparent tile? 

I tested all the tutorials mentioned in the issue and the hints show fine now! I'm seeing this error though:
`OmittedExpression not supported`
which might just be a whole other issue? It doesn't seem to affect hint rendering though! (but i haven't looked into it at all)

### DIFF
--- a/pxtlib/spriteutils.ts
+++ b/pxtlib/spriteutils.ts
@@ -196,10 +196,14 @@ namespace pxt.sprite {
             const tm = this.tilemap.copy();
             const tileset: TileSet = {
                 tileWidth: this.tileset.tileWidth,
-                tiles: this.tileset.tiles.map(t => ({
-                    ...t,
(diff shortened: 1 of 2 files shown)


shakao:
hmm, i don't think the tileset should have a null tile in it, the transparency is just a starter tile that's added to every tilemap: https://github.com/microsoft/pxt/blob/master/pxtlib/tilemap.ts#L568

something might be going wrong where we are generating the tileset? ("decodeTilemap", i think)

livcheerful:
OOO okay I'll look more into that then >> thanks shannon!

PR #7677, opened 2020-12-02
c++ compilation analytics

pelikhan:
- [x] Add analytics around C++ build attempts and failure.
- [x] give up after some time since it's not uncommon our build is down

It roughly takes a minute to fail.

Add ``cppcompile.poll``, ``cppcompile.cancel``, ``cppcompile.done`` ticks.


pxtlib/cpp.ts
```diff
@@ -1349,12 +1350,25 @@ namespace pxt.hexloader {
...
                             .then(ret => new Promise<string>((resolve, reject) => {
-                                let tryGet = () => {
+                                let retry = 0;
+                                const maxRetry = 7;
+                                let delay = 1000; // ms
+
+                                const tryGet = () => {
+                                    retry++;
+                                    if (retry > maxRetry) {
+                                        pxt.log(`abandonning C++ build`)
```
jwunderl, line 1360:
```suggestion
                                        pxt.log(`abandoning C++ build`)
```

nit

pxtlib/cpp.ts
```diff
@@ -1367,8 +1381,10 @@ namespace pxt.hexloader {
                                             }
                                         },
                                             e => {
-                                                setTimeout(tryGet, 1000)
-                                                return null
+                                                delay = Math.min(30000, delay * 1.5); // exponential back off, max 30sec
```
jwunderl, line 1384:
```suggestion
                                                delay = Math.min(30000, delay * 2); // exponential back off, max 30sec
```

possibly, or maybe just a few more retries? at 1.5* and 7 max retries it would time out after ~48seconds with a max delay of 17s, I think 48s is still in the realm of "possibly going to succeed"

jwunderl (approved):
LGTM though we might need to adjust retry count / backoff amount

mmoskal:
I'm not sure we need exponential back-off. If the build is on-going, the cloud will block the requested for JSON file for up to 30s. Hitting that JSON also doesn't start new compilation (only a POST request can do that), so there seems no harm in the current behavior. We should timeout though as you do now after a few tries (maybe 2-3 minutes).

pelikhan:
Nobody is going to stay in front of the editor for 3 minutes thinking it's doing work. They'll refresh the page and go somewhere else.

pxtlib/cpp.ts
pelikhan: Removed backoff

pelikhan:
> I'm not sure we need exponential back-off. If the build is on-going, the cloud will block the request for JSON file for up to 30s (long-polling). Hitting that JSON also doesn't start new compilation (only a POST request can do that), so there seems no harm in the current behavior. We should timeout though as you do now after a few tries (maybe 2-3 minutes).

Retrying for up to 2min then give up.

mmoskal:
Does this actually wait for 120s? I think it will wait for 12*(30+10), because the http get will hang for 30s.

pelikhan:
@mmoskal  using time now

mmoskal (approved):

pxtlib/cpp.ts
```diff
@@ -1334,13 +1334,14 @@ namespace pxt.hexloader {
         let hexurl = ""
 
         showLoading(pxt.U.lf("Compiling (this may take a minute)..."));
+        pxt.tickEvent("cppcompile.start")
         return downloadHexInfoLocalAsync(extInfo)
             .then((hex) => {
                 if (hex) {
                     // Found the hex image in the local server cache, use that
+                    pxt.tickEvent("cppcompile.cachehit")
```
pelikhan, line 1342: @abchatra is this how you want ticks?

PR #7681, opened 2020-12-03
Allow prebuilt js in run.html links

jwunderl:
Add option to use prebuilt js on share page in arcade to speed up loading; this is mainly just an intermediary step towards prebuilding sim js in general (hopefully!)

(also added a quick fix for docs sims, which is crashing off master right now: https://github.com/microsoft/pxt/pull/7681/commits/b534fe7d608177a368f830aab6fa009cd0ddf1a4)

loading-loading-loading

asphodel diff:
time from page load -> first asphodel paint on my work computer:
image

with this (worth noting this is local serve / loading time may vary):
image

other notes:
* I'm appending version suffixes so any bump would need to rebuild the source -- this isn't usually necessary except for major releases, but 
* plays a bit weirdly with local host, as we replace `@versionsuff@` with "" there; for testing locally I just renamed the file without the suffix
* I only made the cli command to build one share link's js at a time (instead of all approved sharelinks), but it'll be easy enough to add one that builds all approved links + a github action to handle checking out the right release / build all / make a pr; just a few other improvements I wanted to look at before the release / it's not a big deal to manually update when there's only one approved link at the moment anyways
* There's not a clean point between pxt.runner and the cli where we can get the compilation portion the same, so I left off the upgrade rules + python support from the cli for this PR so it's easier to port; will see about getting that cleaned up in follow afterwards~

and example built off current arcade (not stable branch): https://github.com/microsoft/pxt-arcade/pull/2622

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -5014,21 +5014,14 @@ export async function buildShareSimJsAsync(parsed: commandParser.ParsedCommand) 
         throw new Error(`Failed to compile share id: ${id}`);
     }
 
-    // todo: this is duped from runner.ts, maybe move to pxtc so can just keep in one loc?
-    const builtJs = {
-        js: compileResult.outfiles[pxtc.BINARY_JS],
-        targetVersion: targetVersion,
-        fnArgs: compileResult.usedArguments,
-        parts: pxtc.computeUsedParts(compileResult, "ignorebuiltin"),
-        usedBuiltinParts: pxtc.computeUsedParts(compileResult, "onlybuiltin"),
-    };
+    const builtJsInfo = pxtc.buildSimJsInfo(compileResult);
 
     const outdir = parsed.flags["output"] as string || path.join(cwd, "docs", "static", "builtjs");
     nodeutil.mkdirP(outdir);
     const outputLocation = path.join(outdir, `${id}v${targetVersion}.json`);
     fs.writeFileSync(
         outputLocation,
-        JSON.stringify(builtJs)
+        JSON.stringify(builtJsInfo)
     );
 
     process.chdir(cwd);
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -4991,6 +4991,50 @@ export function buildAsync(parsed: commandParser.ParsedCommand) {
         }).then((compileOpts) => { });
 }
 
+export async function buildShareSimJsAsync(parsed: commandParser.ParsedCommand) {
+    const id = parsed.args[0];
+    console.log(`Building sim js for ${id}`);
+    const targetVersion = pxt.appTarget.versions.target;
+    const cwd = process.cwd();
+    const builtFolder = path.join("temp", id);
+    nodeutil.mkdirP(builtFolder);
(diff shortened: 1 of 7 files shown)


pelikhan (dismissed):
Does this allow injecting any javascript file as prebuildSimJs?

pelikhan:
This is slightly better but it still creates nested iframes instead running the simulator in the main window.

pelikhan:
- [ ] our "buildtarget" command should regenerate all these files.

jwunderl:
I can take a look at cutting the second iframe, it shouldn't be that hard as it would just be taking portions of run.html and inserting them into arcades docfiles/script.html; that's all over in arcade though.

For buildtarget do you mean dropping them in the built arcade repo instead of docs or just as part of the build? Currently these are a bit closer to the electron app than other docs, as you would want the ones in docs to always match either the current index-ref or the next one you are planning on releasing. I was planning on doing a github action for these post release that would handle cleaning and building these (take a tag you're planning on releasing as a param, delete any old builtjs that don't match target or index-ref, builds all approved share links, makes a PR updating everything) but I can do that tonight if preferred.

pelikhan:
Look at the GitHub script page for skipping the idéale.

abchatra:
>> I can take a look at cutting the second iframe, it shouldn't be that hard as it would just be taking portions of run.html and inserting them into arcades docfiles/script.html; that's all over in arcade though.**

Lets do it in a different PR.

pelikhan:
For the build, yes i think it should be part of "pxt buildtarget". It also acts as a integration test too.

jwunderl:
If they're built during buildtarget we'd have to create a new release any time we want to add a new prebuilt sample game, which is doable but maybe a bit tedious. Also, ideally we wanted to get the share page faster for this patch release so it's better for hour of code week, would it be alright to keep it as is for this pr and do a follow up pr changing where it's stored?

If we are baking them into releases though, my understanding is that during the build I'd be adding each game to `/built/prebuiltsim/@id@.json`, and then they'd be available at `/blb/prebuiltsim/@id@.json` in the html (hopefully? I'll look at the order things get resolved, this would have to assume that `@id@` gets replaced before the `@blobCdnUrl@` is portion is handled); I haven't added anything new there yet so need to look that over.

pelikhan:
Don't we want this for hour of code?

pelikhan:
> If they're built during buildtarget we'd have to create a new release any time we want to add a new prebuilt sample game, which is doable but maybe a bit tedious.

I don't understand that reasoning. Probably a mis understanding there. I think we should be them on every build as a test suite; probably not to commit them.

jwunderl:
Yes we wanted to get it out in the skillmap patch, @pelikhan @mmoskal do you have time to review? Trying to keep changes here minimal for porting

> I don't understand that reasoning. Probably a mis understanding there. I think we should be them on every build as a test suite; probably not to commit them.

Ah, yes, I thought you meant to save them in buildtarget, happy to add as part of ci tests but would prefer to do that in another PR for porting

PR #7682, opened 2020-12-03
Mono repo support

pelikhan:
Support for PXT package in sub-folder of repositories, not just top level. This would allow to break down massive projects and still enjoy a sane versioning system. The only weird requirement is to have at least a top level dummy ``pxt.json`` file

- [x] version is still handled at the repo level, e.g. applies to all sub-repoes
- [x] import from home screen
- [x] loaded as dependency in editor
- [x] loaded as dependency in sim
- [x] added as extension in editor
- [x] load multiple sub-projects in same project
- [x] github flow: commit / diff / revert / push
- [x] show proper names in home screen
- [x] search
- [x] update github populate dialog to account for nested package
- [x] docs
- [x] Test all the above with a classic repo.

You can test out a mono repo at https://github.com/pelikhan/mono-demo/demo1 and https://github.com/pelikhan/mono-demo/demo2

A few bugs fixed along the way:
- [x] when setDiagnostics triggers while in the home screen, ignore. This happens a lot with big programs as setDiagnostics always finishes after leaving the editor
- [x] check that targetVersions is in the correct target for checks (this should really have been an array)
- [x] query latest tags before doing a release, instead of relying on pxt.json


mmoskal:
can you rewrite `https://github.com/pelikhan/mono-demo/tree/master/demo1` to `https://github.com/pelikhan/mono-demo/demo1` (`/tree/anything/path` -> `/path`) ?

pelikhan:
> can you rewrite `https://github.com/pelikhan/mono-demo/tree/master/demo1` to `https://github.com/pelikhan/mono-demo/demo1` (`/tree/anything/path` -> `/path`) ?

It already does :)

mmoskal:
> > can you rewrite `https://github.com/pelikhan/mono-demo/tree/master/demo1` to `https://github.com/pelikhan/mono-demo/demo1` (`/tree/anything/path` -> `/path`) ?
> 
> It already does :)

jwunderl (approved):
LGTM

PR #7697, opened 2020-12-08
Patch header name to match card

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/2661

Changes it to be the name of the activity

### DIFF
--- a/skillmap/src/components/makecodeFrame.tsx
+++ b/skillmap/src/components/makecodeFrame.tsx
@@ -185,7 +185,7 @@ class MakeCodeFrameImpl extends React.Component<MakeCodeFrameProps, MakeCodeFram
     }
 
     protected async handleWorkspaceSaveRequestAsync(request: pxt.editor.EditorWorkspaceSaveRequest) {
-        const { dispatchSetHeaderIdForActivity, activityHeaderId, activityType } = this.props;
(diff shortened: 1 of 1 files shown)


skillmap/src/components/makecodeFrame.tsx
```diff
@@ -195,6 +195,17 @@ class MakeCodeFrameImpl extends React.Component<MakeCodeFrameProps, MakeCodeFram
             }
         };
 
+        // Patch the name, otherwise it will be the name of the GitHub repo
+        if (project.header?.name !== title) {
+            project.header!.name = title;
+            const pxtJSON = project.text!["pxt.json"];
+            if (pxtJSON) {
+                const config = JSON.parse(pxtJSON);
+                config.name = title;
```
shakao, line 204: do we need a null check for config if the JSON.parse fails, or is this pretty safe?

shakao (approved):

skillmap/src/components/makecodeFrame.tsx
riknoll: I think it's safe? if pxt.json isn't there then I think we have other problems

shakao: oh yeah good point

PR #7709, opened 2020-12-10
Add tutorial.finish tick for combining recipe and tutorial exit events

shakao:


jwunderl (approved):
To confirm, looks like this

PR #7711, opened 2020-12-10
Fixing default properties in the analytics

abchatra:
We are not doing a clone but accumulating the default properties as we go.  Yikes


pelikhan:
change made 2 years ago :(

pelikhan (approved):

PR #7713, opened 2020-12-10
External message simulators

pelikhan:
Adds support for registering external simulator URLs to be loaded when specific channel are used in control.simmessages.

A message simulator is just any other MakeCode simulator iframe; it receives the same iframe messages and it's lifecycle is maintained by the simdriver. Once a message from control.simmessages is detected, we lazily spin up the message simulator if registered.

* this change is a no-op if the messageSimulators settings are not in pxtarget.json
* only URLs specified in messageSimulators will load


pxtlib/cpp.ts
```diff
@@ -1465,7 +1465,7 @@ namespace pxt.hexloader {
                 keys.unshift(newkey)
                 let todel = keys.slice(maxLen)
                 keys = keys.slice(0, maxLen)
-                return Promise.map(todel, e => host.cacheStoreAsync(e, null))
+                return Promise.map(todel, e => host.cacheStoreAsync(e, "[]"))
```
pelikhan, line 1468: bug found by running v14+

pxtsim/embed.ts
```diff
@@ -78,6 +78,13 @@ namespace pxsim {
     export interface SimulatorBroadcastMessage extends SimulatorMessage {
         broadcast: boolean;
     }
+
+    export interface SimulatorControlMessage extends SimulatorBroadcastMessage {
```
pelikhan, line 82: this used to be in common-packages

mmoskal (approved):

PR #7717, opened 2020-12-11
add css classes to sim messages iframes for skinning

pelikhan:
Add a couple css classes to enable editor/specific skinning.


pxtsim/simdriver.ts
```diff
@@ -333,6 +333,8 @@ namespace pxsim {
                             this.container.appendChild(wrapper);
                             messageFrame = wrapper.firstElementChild as HTMLIFrameElement;
                             messageFrame.dataset[FRAME_DATA_MESSAGE_CHANNEL] = messageChannel;
+                            pxsim.U.addClass(wrapper, "simmsg")
+                            pxsim.U.addClass(wrapper, "simmsg" + messageChannel)
```
mmoskal, line 337: wouldn't it be better to have "simmsg " + messageChannel so you get two classes? or you just don't want the sim styling to apply to jd iframe?

mmoskal (approved):

pxtsim/simdriver.ts
pelikhan: in less, in have ``.simmsg > iframe { ... }`` to skin the any message simulator iframe and ``.simmsgjacdac > iframe { ... }`` for jacdac specifics.

PR #7718, opened 2020-12-11
fix pathing for local serve in docs, issue with lang picker on /pkg

jwunderl:
Hassan pointed out last night that the `/pkg/{repo slug}` page on local serve was behaving poorly -- since most of our icon urls are defined as relative paths (https://github.com/microsoft/pxt-arcade/blob/1517905b5d614a77a433dd7328b6be1ae92586df/pxtarget.json#L438) they get messed up (in this case, it ended up making requests to github as it was relative under the `/pkg/` endpoint, which causes the local serve to crash.

Fix that and a few other minor issues I noticed along the way
* fix refs to images defined in `pxt/docs/static`
* patch relative links to ./static/ (these get automatically swapped out for the cdn links in the backend anyways)
* Don't error when pulling in docs.js on a page without langpicker
* add langpicker to /pkg/ so the button works (the translations don't work locally for me, but they appear to if you go to pkg/{slug}?lang=zh-CN on a live site?

### DIFF
--- a/cli/server.ts
+++ b/cli/server.ts
@@ -35,6 +35,7 @@ function setupRootDir() {
     console.log(`With pxt core at ${nodeutil.pxtCoreDir}`)
     dirs = [
         "built/web",
+        path.join(nodeutil.targetDir, "docs"),
         path.join(nodeutil.targetDir, "built"),
         path.join(nodeutil.targetDir, "sim/public"),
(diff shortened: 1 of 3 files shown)


riknoll (approved):
Not super familiar with this part of the code, but LGTM

PR #7722, opened 2020-12-14
Upgrade tutorial tiles from old tilemaps into user's project

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/2568

corresponding pxt-arcade PR: https://github.com/microsoft/pxt-arcade/pull/2727

The issue was when we initiate the tilemap block, we try to decode all the tiles, but the tiles don't exist in the project. This will add all the custom tiles to the user's tileset -- it looks like this for Side Scroller:
image

Questions I have:
- Should I be worried about adding another decompile in? Should I try to store what we had before and use that?
- Is it best to separate out the XML upgrade logic into pxt-arcade? Or should I move it to here? 

Try out the upload hereee: https://arcade.makecode.com/app/5f84aab6ae563167ac13981b277805cda8157511-9d50993b91

The tutorials we have with the old tilemaps are: 

- Barrel Dodger
- Simple Extensions
- Maze
- Side Scroller
- Setting the Scene
- Breadcrumb Trail

### DIFF
--- a/pxteditor/editor.ts
+++ b/pxteditor/editor.ts
@@ -545,6 +545,9 @@ namespace pxt.editor {
         blockId?: string;
     }
 
+
+    export let upgradeXml: (code: string) => void;
+
     export let initExtensionsAsync: (opts: ExtensionOptions) => Promise<ExtensionResult>
         = opts => Promise.resolve<ExtensionResult>({});
 
--- a/pxteditor/editor.ts
+++ b/pxteditor/editor.ts
@@ -545,9 +545,6 @@ namespace pxt.editor {
         blockId?: string;
     }
 
-
-    export let upgradeXml: (code: string) => void;
-
     export let initExtensionsAsync: (opts: ExtensionOptions) => Promise<ExtensionResult>
         = opts => Promise.resolve<ExtensionResult>({});
 
--- a/pxtlib/tilemap.ts
+++ b/pxtlib/tilemap.ts
@@ -816,6 +816,9 @@ namespace pxt {
                 case AssetType.Image:
                     return this.state.images.getByID(name) || this.gallery.images.getByID(name);
                 case AssetType.Tile:
+                    if (name == "myTiles.tile0") {
+                        return this.getTransparency(16);
+                    }
                     return this.state.tiles.getByID(name) || this.gallery.tiles.getByID(name);
                 case AssetType.Tilemap:
                     return this.state.tilemaps.getByID(name) || this.gallery.tilemaps.getByID(name);
--- a/pxtlib/tilemap.ts
+++ b/pxtlib/tilemap.ts
@@ -552,13 +552,6 @@ namespace pxt {
             return new pxt.sprite.TilemapData(tilemap, tileset, layers.data());
         }
 
-        public getTileWidth(): number {
-            const tilemaps = this.state.tilemaps.getSnapshot();
-            if (tilemaps.length == 0)
-                return 16; // default tile size
-            return tilemaps[0].data.tileset.tileWidth; // Assuming that all tilemaps have the same size tiles
-        }
-
         public resolveTile(id: string): Tile {
             return this.lookupAsset(AssetType.Tile, id) as Tile;
         }
@@ -823,9 +816,6 @@ namespace pxt {
                 case AssetType.Image:
                     return this.state.images.getByID(name) || this.gallery.images.getByID(name);
                 case AssetType.Tile:
-                    if (name == "myTiles.tile0") {
-                        return this.getTransparency(this.getTileWidth());
-                    }
                     return this.state.tiles.getByID(name) || this.gallery.tiles.getByID(name);
                 case AssetType.Tilemap:
                     return this.state.tilemaps.getByID(name) || this.gallery.tilemaps.getByID(name);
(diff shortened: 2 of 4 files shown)


webapp/src/app.tsx
```diff
@@ -1470,6 +1470,7 @@ export class ProjectView
         return this.loadBlocklyAsync()
             .then(() => tutorial.getUsedBlocksAsync(t.tutorialCode, t.tutorial, t.language))
             .then((usedBlocks) => {
+                tutorial.xmlUpgrades(t.tutorialCode, t.language);
```
jwunderl, line 1473: do we not need to wait on this? It's kicking off async work right?

webapp/src/tutorial.tsx
```diff
@@ -16,6 +16,21 @@ import * as editortoolbar from "./editortoolbar";
 
 type ISettingsProps = pxt.editor.ISettingsProps;
 
+export function xmlUpgrades(code: string[], language?: string,) {
+    compiler.getBlocksAsync()
```
jwunderl, line 20: maybe wrap the entire thing in `if (pxt.editor.upgradeXml)` instead of just the bit in the last `then` to avoid unnecessary work?

jwunderl (commented):
Question; would it be possible to just update the tutorial content in pxt-arcade instead of doing the upgrade at runtime? my understanding is that we own all of the tutorials using the old tilemaps at this point, so if we were to upgrade them ourselves we could avoid the runtime cost of decompiling / upgrading every single time a tutorial is loaded in the future.

We did something similar in the past with tilemaps where we had a change like this one temporarily, upgraded all the content manually, and then removed that upgrade path. If we made this a no-op except when upgradeXml is defined then it would be a nice way to maintain a temporary upgrade path for breaking docs changes.

livcheerful:
ooo good point @jwunderl. Do we expect that there are third party tutorials that we don't want to break?

pxtlib/tilemap.ts
```diff
@@ -816,6 +816,9 @@ namespace pxt {
                 case AssetType.Image:
                     return this.state.images.getByID(name) || this.gallery.images.getByID(name);
                 case AssetType.Tile:
+                    if (name == "myTiles.tile0") {
```
shakao, line 819: i thought transparency was named "myTiles.transparency16" or whatever the tile size is? or did it get renamed during the refactor process?

pxtlib/tilemap.ts
```diff
@@ -816,6 +816,9 @@ namespace pxt {
...
+                        return this.getTransparency(16);
```
shakao, line 820: this should probably be the tilewidth instead of 16

pxtlib/tilemap.ts
livcheerful: oh yes the old way was myTiles.tile0

webapp/src/tutorial.tsx
livcheerful: :O yes I will do that tytyty

pxtlib/tilemap.ts
livcheerful: ahhh yes agreed. I was having issues with tilewidths - are all the tilemaps in a tilemapProject supposed to have the same tilewidth?

riknoll: no, they can have different widths

livcheerful: oh okay hmmmm i need to fix some things

pxteditor/editor.ts
```diff
@@ -545,6 +545,9 @@ namespace pxt.editor {
         blockId?: string;
     }
 
+
+    export let upgradeXml: (code: string) => void;
```
riknoll, line 549: is there something preventing this upgrade from being inside the other xml upgrade code we already have? The ones returned by initExtensionsAsync?

riknoll (changes requested):

pxteditor/editor.ts
riknoll: blocklyPatch

livcheerful: oh yes initially it was because the dom we were getting there was for the workspace and not the tutorial.  Should I change it to also take in tutorial XML?

riknoll: the tutorial xml is also a workspace, it just isn't loaded in the editor (it gets rendered to an svg). i think it's fair to call the blocklypatch function on the decompiled tutorial xml as well.

riknoll: you shouldn't have to change the function though, just also call it in the tutorial loading code

livcheerful: 🎉 i will do that!

livcheerful:
this issue has been fixed with a different change >< Closing this PR!

riknoll:
@livcheerful what change?

livcheerful:
It does it in https://github.com/microsoft/pxt/blob/106dd01ff00d3605c4ae43e3007a27c0c99eb8f1/webapp/src/app.tsx#L1539

But I'm still not sure what the specific change is yet... I was worried that it was always fixed and I just didn't try to repro on beta? But then I wouldn't have seen it in my local build right?? I'm gonna hunt for it

PR #7723, opened 2020-12-14
Update skillmap build to support versioned skillmap

shakao:
new flow!
- `gulp skillmap` in pxt to copy all files to the correct locations. this:
   - builds the skillmap
   - copies the index.html into `webapp/public` as `skillmap.html`
   - copies the built js and css into `built/web/skillmap`
- build locally by running `pxt skillmap --serve` in the target directory
- the previous build command (`pxt skillmap` in the target) is still around in case we need to deploy before we swap over to the new url

changes:
- all skillmap assets are now permanently kept in `docs/static/skillmap`
- added npm prebuild and postbuild commands to generate an index.html formatted to our specifications (don't remove comments, don't collapse whitespace), so the backend replacements should all work
- added the skillmap url to the webconfig

for https://github.com/microsoft/pxt-arcade/issues/2584


pelikhan (changes requested):
Please add build instructions to the read me.

pelikhan:
Does the pxt ci run gulp skillmap as part of the build?

shakao:
yep! although it doesn't have to, i can take it out for now since the backend route changes aren't in either

riknoll (approved):

jwunderl (approved):

riknoll:
@pelikhan can you mark your changes as resolved?

pelikhan (approved):

riknoll:
thanks!

PR #7734, opened 2020-12-21
option to run single simulators

pelikhan:
With single=1, the sim runner (run.html) never spins up an additional iframe (neighter duplicate or using message packet). Use in JACDAC docs to have better control if frame being spinned up

### DIFF
--- a/pxtsim/simdriver.ts
+++ b/pxtsim/simdriver.ts
@@ -441,7 +441,6 @@ namespace pxsim {
             this.setState(starting ? SimulatorState.Starting : SimulatorState.Stopped);
             if (unload)
                 this.unload();
-            this.removeEventListeners();
         }
 
         public suspend() {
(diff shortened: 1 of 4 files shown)


pxtsim/simdriver.ts
```diff
@@ -438,6 +441,7 @@ namespace pxsim {
             this.setState(starting ? SimulatorState.Starting : SimulatorState.Stopped);
             if (unload)
                 this.unload();
+            this.removeEventListeners();
```
mmoskal, line 444: was this missing before?

mmoskal (approved):

pxtsim/simdriver.ts
pelikhan: Yes. It seemed to be dead code. Will do further testing.

jwunderl (approved):

pxtsim/simdriver.ts
pelikhan: Add was idempotent, but we never registered those events, so we were still listening from the home screen.

pelikhan:
Removed removeEventListener -> for separate PR

PR #7736, opened 2020-12-23
Make sure currentRuntime exists before accessing stuff

livcheerful:


jwunderl (approved):
LGTM, if we're getting a broadcast message its coming from an existing runtime ~

pelikhan:
Message might come from a host

PR #7742, opened 2021-01-04
Skillmap css, educator info url, map reset

shakao:
each commit is a separate fix:

- add 'editor' class to root element when iframe is open, only specify height:100% in iframe case (fixes https://github.com/microsoft/pxt-arcade/issues/2697)
(tested that the iframe looks all right locally, and the added css seems to work on live when adding via chrome inspector)

- add 'Reset' to gearwheel, hooks up to reset warning modal. "reset" erases the current user's map progress/completed tags (fixes https://github.com/microsoft/pxt-arcade/issues/2699)
image

- adds an info icon by the page title if the markdown has an infoUrl (fixes https://github.com/microsoft/pxt-arcade/issues/2698)
image


skillmap/src/components/AppModal.tsx
```diff
@@ -90,6 +93,25 @@ export class AppModalImpl extends React.Component<AppModalProps> {
         </Modal>
     }
 
+    renderResetWarning() {
+        const  { dispatchResetUser, dispatchHideModal } = this.props;
+        const resetModalTitle = lf("Reset All Activities?");
+        const resetModalText = lf("Are you sure you want to reset? This will permanently erase all progress and delete the current projects. This action cannot be undone.");
```
riknoll, line 99: "delete the projects you created in tutorials"? struggling with how to convey that it's really just deleting the projects in the skillmap and not the main editor

riknoll (approved):

jwunderl (approved):

PR #7744, opened 2021-01-05
Support projects that use assets from the asset editor in tutorials

riknoll:
I've added documentation on how to use this to the docs section on tutorials.

This is definitely not ideal, but it will serve as a stopgap until we have time to make a better solution (right now you can't write tutorials with the asset editor at all 😦). @abchatra and I brainstormed a few options earlier today that we'll discuss with the team.

### DIFF
--- a/pxtlib/gallery.ts
+++ b/pxtlib/gallery.ts
@@ -54,6 +54,16 @@ namespace pxt.gallery {
         return undefined;
     }
 
+    export function parseTemplateProjectJSON(md: string): pxt.Map<string> {
+        const pm = /```assetjson\s+((.|\s)+?)\s*```/i.exec(md);
+
+        if (pm) {
+            pxt.tutorial.parseAssetJson(pm[1]);
+        }
+
+        return {};
+    }
+
     export function parseExampleMarkdown(name: string, md: string): GalleryProject {
         if (!md) return undefined;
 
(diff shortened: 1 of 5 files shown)


pxtlib/gallery.ts
```diff
@@ -54,6 +54,16 @@ namespace pxt.gallery {
         return undefined;
     }
 
+    export function parseTemplateProjectJSON(md: string): pxt.Map<string> {
+        const pm = /```assetjson\s+((.|\s)+?)\s*```/i.exec(md);
+
+        if (pm) {
+            pxt.tutorial.parseAssetJson(pm[1]);
```
jwunderl, line 61:
```suggestion
            return pxt.tutorial.parseAssetJson(pm[1]);
```

I believe?

webapp/src/app.tsx
```diff
@@ -3730,6 +3743,14 @@ export class ProjectView
         return null;
     }
 
+    private saveTutorialTemplateAsync() {
+        return pkg.mainEditorPkg().buildAssetsAsync()
+            .then(() => pkg.mainPkg.saveToJsonAsync())
+            .then(project => {
+                pxt.BrowserUtils.browserDownloadText(project.source, pkg.genFileName(".txt.mkcd"))
```
jwunderl, line 3750:
```suggestion
                pxt.BrowserUtils.browserDownloadText(project.source, pkg.genFileName(".assets.mkcd"))
```

Fine either way, but maybe a bit more descriptive than .txt? I also personally get weirded out  a bit when I see `filename.{common file extension}.{something else}` as people occasionally try and use that for viruses (`famous-book.pdf.exe`).

jwunderl (approved):
LGTM

shakao (approved):

PR #7745, opened 2021-01-05
Store skillmap progress for page source url rather than carousel name

shakao:
previously mapProgress was a map of carousel name to progress (per activity)--this would cause issues if different pages had the same name for a carousel. adding an extra layer where mapProgress uses the page source URL (same as completedTags, should be unique) to map to individual carousels

also makes the page source URL required (defaults to "default" if one is not found)


riknoll (commented):
LGTM, but do you need to update the saved user state at all? or did that turn out not to be an issue?

shakao:
@riknoll updated with the upgrade logic! we need the pagesourceurl + current maps present, so i'm doing it in the reducer when the user is first loaded. also added a version for the user state

jwunderl (approved):

riknoll (approved):

PR #7748, opened 2021-01-06
high contrast fixes for github view

pelikhan:
There's probably a million discussion about dark theme high contrast text diff; this is one way to do it.

image

image


jwunderl (approved):
LGTM

PR #7749, opened 2021-01-06
Scale screenshots / gifs to display on social media better

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/2600

The issue was that Facebook won't show images if they're too small (less than 200x200 px), so I added an optional scaling factor to converting the screenshot image -> PNG and also when we make the gifs. Are we worried about this taking too much storage?

[edit] I had a screenshot that I forgot to share!
image
it's a little blurry still, but I think that's on FB's end not liking pixel art. better than having part of the microsoft logo?

### DIFF
--- a/pxtlib/browserutils.ts
+++ b/pxtlib/browserutils.ts
@@ -429,14 +429,17 @@ namespace pxt.BrowserUtils {
             })
     }
 
-    export function imageDataToPNG(img: ImageData): string {
+    export function imageDataToPNG(img: ImageData, scale = 1): string {
         if (!img) return undefined;
 
         const canvas = document.createElement("canvas")
-        canvas.width = img.width
-        canvas.height = img.height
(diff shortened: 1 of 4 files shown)


pelikhan:
Storage may or may not be an issue. What kind of size increase are you seeing? Looks good otherwise.

pelikhan (approved):

livcheerful:
> Storage may or may not be an issue. What kind of size increase are you seeing? Looks good otherwise.

It'll increase by a factor of 3 for all screenshots/gifs

pelikhan:
There is definitely a maximum size. Did you hit with this new change?

livcheerful:
oh you're totally right it hit the limit! I changed it so that gifs only scale by 2, which will look like this on FB:

image

PR #7751, opened 2021-01-06
[stable6.0] alpha-sort deps for c++ compilation to avoid missing valid hex files …

jwunderl:
Should be fix for https://github.com/microsoft/pxt-microbit/issues/3826 (need to do a patch release for the random crash fix as well), but still in process of getting everything checked out / making a local build to test

* https://github.com/microsoft/pxt/pull/7320
* https://github.com/microsoft/pxt/pull/7429
* https://github.com/microsoft/pxt/pull/7465
* https://github.com/microsoft/pxt/pull/7763
* https://github.com/microsoft/pxt/pull/7764

** edit ** also fixing the bug for https://github.com/microsoft/pxt-microbit/issues/3397, testing that now as well


abchatra (approved):

jwunderl:
neither of the linked issues are fully fixed by the two cherry picked fixes, I'm fixing them both on this branch & will port those fixes to master / get reviews there before undrafting this (easier to fix that way as they're both microbit specific).

For reference, I believe the hex file download issue one is due to multivariant but still figuring it out, the lang one took me by surprise a bit -- it's actually been using these ones instead of downloading the strings.json the whole time: https://github.com/microsoft/pxt/tree/master/webapp/public/locales. Fix for that is https://github.com/microsoft/pxt/pull/7751/commits/5fa4d9e0df9509bd3617a422dbee3bcdc7985dc6, currently all translations from strings.json are out of date by 2 years

(also cherry picked my change to use pxt-translations so I can test / fix things, also makes it easier to port things back to master)

jwunderl:
Okay, all fixed up, also includes https://github.com/microsoft/pxt/pull/7763 and https://github.com/microsoft/pxt/pull/7764 to fix the issues

PR #7760, opened 2021-01-08
Run expandHtml on skillmap before uploading file

shakao:


cli/cli.ts
```diff
@@ -1095,6 +1095,13 @@ function uploadCoreAsync(opts: UploadOptions) {
         "skillmap.html"
     ]
 
+    // expandHtml is manually called on these files before upload
+    // runs <!-- @include --> substitutions, fills in locale, etc
+    let expandFiles = [
+        "index.html",
```
jwunderl, line 1101: Should these be qualified names instead of just any index.html? I feel like we might end up having multiple `index.html`'s   in different dirs which could have this with weird behavior / not matching other htings

jwunderl (approved):
LGTM but one question / possible change \\/

cli/cli.ts
shakao: hm, i wanted to exactly match the current cli behavior, i'm a little leery of changing it because i don't know if anything is dependant on this

jwunderl: Oh I read that index of backwards, you're right / already doing what I was meaning to suggest!

PR #7774, opened 2021-01-13
fix skillmap step race condition and change is completed logic

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/2789

We now display the last step as 7/7 unless you actually click finish.
Also fixes a race condition when unloading the iframe where we weren't updating the activity progress before the iframe had been unloaded.


jwunderl (approved):

shakao:
i think it should be fine, but auto-displaying the modal after completing a carousel still works as expected with this?

riknoll:
yep, still works!

PR #7778, opened 2021-01-13
Fix infinite reload if language wasnt set; fix setting High Contrast

darzu:
It was checking for "undefined" !== "null" and returning true.

Fixes: https://github.com/microsoft/pxt-arcade/issues/2864
Fixes: https://github.com/microsoft/pxt-arcade/issues/2722

We really should remove the location.reloal() all together.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -4592,7 +4592,7 @@ document.addEventListener("DOMContentLoaded", () => {
             const cloudLang = auth.getState()?.preferences?.language;
             // kick of a user preferences check; if the language is different we'll reload
             auth.initialUserPreferences().then((pref) => {
-                if (pref && pref.language !== pxt.BrowserUtils.getCookieLang()) {
+                if (pref && pref.language != pxt.BrowserUtils.getCookieLang()) {
                     pxt.BrowserUtils.setCookieLang(pref.language);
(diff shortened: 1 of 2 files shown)


eanders-MS (approved):

webapp/src/app.tsx
```diff
@@ -4592,7 +4592,7 @@ document.addEventListener("DOMContentLoaded", () => {
             const cloudLang = auth.getState()?.preferences?.language;
             // kick of a user preferences check; if the language is different we'll reload
             auth.initialUserPreferences().then((pref) => {
-                if (pref && pref.language !== pxt.BrowserUtils.getCookieLang()) {
+                if (pref && pref.language != pxt.BrowserUtils.getCookieLang()) {
                     pxt.BrowserUtils.setCookieLang(pref.language);
```
pelikhan, line 4596: if the user does not allow you to set cookies for any kind of reason, you reload forever.

darzu: Good point.

We need to purge this reload and be non-cookie safe.

darzu: I removed the reload entirely and put a info message. I also added a check to see if language cookie exists at all. This branch should only be hit if you change preferences on a different computer.

PR #7779, opened 2021-01-13
Fix hang in safari when dragging a png from the gallery

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/2802

For some reason, Safari is just reading the gallery png as a buffer full of 0's. Handle that case


jwunderl (approved):
Maybe worth a comment on why since it's such a weird bug but fine either way~

livcheerful (approved):

PR #7782, opened 2021-01-14
Try loading the latest cloud version when opening a project; skip it if it takes too long

darzu:
Try loading the latest cloud version when opening a project; skip it if it takes too long
Adds support for partial synchronizations.

The new behavior is:
When opening a cloud project,
Start fetching that project's latest state from the cloud,
If the fetch completes with 1.5 seconds, great, load that,
Else, load the local project and if the sync comes back with changes, refresh the editor

Fixes: https://github.com/microsoft/pxt-arcade/issues/2866

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -44,6 +44,7 @@ import * as socketbridge from "./socketbridge";
 import * as webusb from "./webusb";
 import * as keymap from "./keymap";
 import * as auth from "./auth";
+import * as cloud from "./cloud";
 import * as user from "./user";
 
 import * as monaco from "./monaco"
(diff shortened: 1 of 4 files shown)


pelikhan:
I'd recommend adding ticks to track these timeout "performance" in the wild.

darzu:
> I'd recommend adding ticks to track these timeout "performance" in the wild.

Great idea. I've got another PR I'll put up that has all my ticks. Tracking the time it takes to sync will be one of them.

webapp/src/app.tsx
```diff
@@ -1304,27 +1305,38 @@ export class ProjectView
...
+            return Promise.any([
+                timeout,
+                // start a partial cloud sync
+                cloud.syncAsync([h])
+                    .then((changes) => {
+                        if (changes.length) {
+                            if (timeout.isResolved()) {
+                                // We are too late; the editor has already been loaded.
+                                // Call the onChanges handler to update the editor.
+                                cloud.onChangesSynced(changes)
```
eanders-MS, line 1329: Thinking about when we have a conflict resolution dialog available... can a conflict be detected here?

eanders-MS (approved):

webapp/src/app.tsx
pelikhan: Let's assume we did a minor update to the editor (so we automatically modify the pxt.json file with the new target version information) and the connection is slow (1 minute to sync), as soon as the editor will load the project will be "modified" and the local copy will win.
Do you merge changes with the cloud or do you blanket replace the version?

pelikhan, line 1329: Let's call this project A. Let's assume the connection is slow and the user had the time to leave the editor again and load another project B, are we overriding A sources with B?

pelikhan:
I would recommend fault injection (inject large delay semi-randomly) in any code path that talks to a cloud API. This will surface a lot of races naturally.

pelikhan:
> Else, load the local project and if the sync comes back with changes, refresh the editor

Do you merge local and cloud version before refresh? What about local changes made between opening the project and receiving the cloud data?

webapp/src/app.tsx
eanders-MS: > I would recommend fault injection (inject large delay semi-randomly) in any code path that talks to a cloud API. This will surface a lot of races naturally.

Yes this is important.

eanders-MS:
1.5 seconds seems optimistic. My kid's school-issued laptop takes more than that to do anything. Its borderline unusable, but it's what they have.

webapp/src/app.tsx
darzu: No, the closure references the "h" that belongs to the particular invocation of loadHeaderAsync that created it. Otherwise there be all sorts of race conditions in async-await code.

darzu: Blanket replace. Yes, if you edit locally before the cloud has synced, you'll overwrite whatever is in the cloud. This is bad, and a dialog would help, but really we need history. We need git.

darzu:
> I would recommend fault injection (inject large delay semi-randomly) in any code path that talks to a cloud API. This will surface a lot of races naturally.

Good idea. Using the "Network throttle" setting in Chrome dev tools is also particularly helpful.

darzu:
> 1.5 seconds seems optimistic. My kid's school-issued laptop takes more than that to do anything. Its borderline unusable, but it's what they have.

Yes, that's why it's short. A longer timeout would make for a worse slow-connection experience. This is the time that we block their project load to wait on the cloud. We want this to be as short as possible.

pelikhan:
>  We want this to be as short as possible especially for slow connections.

I don't understand this logic. Say a slow connection takes 1 minute to complete, the user will loose 1 minute of changes potentially.

pelikhan:
I think this kind of logic will lead to error loss and frustration.

eanders-MS:
I agree with Peli. This kind of hopeful check will lead to unpredictability. We should block on the check, or not do the check.

PR #7784, opened 2021-01-14
Fix docs sims w jres, support assetjson in docs

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/2566 (include the jres when compiling simjs), and also support the \`\`\`assetjson snippets in docs as I noticed that was missing.

(also fix a minor bug where there was still a box where \`\`\`jres snippets used to be, you can see it at the bottom of https://arcade.makecode.com/blocks-games/jumpy-platformer)

~in the test md file I made it looks like images / animations from the jres aren't getting loaded properly into the blocks, will take a look at that (but wanted to put up PR first as I think this will be the majority of the changes / probably just need an extra line in runner.ts somewhere to patch in the images properly)~ ~Fixed images but animations are still showing up as grey blocks for some reason~ fixed all, but could use some double checking on these two commits: https://github.com/microsoft/pxt/pull/7784/files/04aac748dd41d5255fe3713ad72f6dc999c3ed84..3c3f8a281b53b7014795b8ad6ff3c54bd36ffed7 as well as https://github.com/microsoft/pxt-common-packages/pull/1214

===> test file for assetjson snippets I made
[test-asset-json.zip](https://github.com/microsoft/pxt/files/5812320/test-asset-json.zip)

### DIFF
--- a/pxtlib/tutorial.ts
+++ b/pxtlib/tutorial.ts
@@ -416,7 +416,7 @@ ${code}
             [pxt.TILEMAP_JRES]: files[pxt.TILEMAP_JRES],
             [pxt.TILEMAP_CODE]: files[pxt.TILEMAP_CODE],
             [pxt.IMAGES_JRES]: files[pxt.IMAGES_JRES],
-            [pxt.IMAGES_CODE]: files[pxt.IMAGES_CODE]
+            [pxt.IMAGES_CODE]: files[pxt.IMAGES_CODE],
         }
     }
 }
--- a/pxtrunner/renderer.ts
+++ b/pxtrunner/renderer.ts
@@ -33,8 +33,8 @@ namespace pxt.runner {
         packageClass?: string;
         package?: string;
         jresClass?: string;
-        jres?: string;
-        assetJSON?: string;
+        assetJSONClass?: string;
+        assetJSON?: Map<string>;
         showEdit?: boolean;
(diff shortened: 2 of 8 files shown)


pxtrunner/renderer.ts
```diff
@@ -217,11 +219,13 @@ namespace pxt.runner {
                 pkg.setPreferredEditor(pxt.JAVASCRIPT_PROJECT_NAME);
             }
 
-            if (options.jres) {
```
riknoll, line 220: are you removing support for jres? is this going to break existing docs?

riknoll (approved):

pxtlib/tutorial.ts
```diff
@@ -416,7 +416,7 @@ ${code}
             [pxt.TILEMAP_JRES]: files[pxt.TILEMAP_JRES],
             [pxt.TILEMAP_CODE]: files[pxt.TILEMAP_CODE],
             [pxt.IMAGES_JRES]: files[pxt.IMAGES_JRES],
-            [pxt.IMAGES_CODE]: files[pxt.IMAGES_CODE]
+            [pxt.IMAGES_CODE]: files[pxt.IMAGES_CODE],
```
shakao, line 419:
```suggestion
            [pxt.IMAGES_CODE]: files[pxt.IMAGES_CODE]
```

PR #7789, opened 2021-01-14
Add identity & cloud sync tick events

darzu:
I believe these tick events can help us answer most anonymous questions about auth and cloud sync. See Identity chat for more discussion.

Fixes: https://github.com/microsoft/pxt-arcade/issues/2884

### DIFF
--- a/webapp/src/workspace.ts
+++ b/webapp/src/workspace.ts
@@ -624,7 +624,7 @@ export function duplicateAsync(h: Header, text: ScriptText, newName?: string): P
     delete h.githubTag;
 
     if (h.cloudVersion) {
-        pxt.tickEvent(`cloud.duplicatingCloudProject`);
+        pxt.tickEvent(`identity.duplicatingCloudProject`);
     }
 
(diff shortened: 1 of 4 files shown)


abchatra:
Instead of _cloud._ I would do _identity._  as cloud is very generic term.

webapp/src/workspace.ts
```diff
@@ -1504,11 +1508,16 @@ export function installByIdAsync(id: string) {
 export async function saveToCloudAsync(h: Header) {
     pxt.debug(`cloud save to ${h.name} (${h.id})`)
     checkHeaderSession(h);
+    const saveStart = U.nowSeconds()
```
eanders-MS, line 1511: No StopWatch utility class?

eanders-MS (approved):

webapp/src/workspace.ts
darzu: There probably is, but it doesn't seem simpler than this?

PR #7791, opened 2021-01-14
auth: logout if we receive 401/Unauthorized

eanders-MS:
If any authenticated api call returns 401/Unauthorized, logout the user. This indicates the user did something to invalidate their account such as delete their account from another browser or tab.


eanders-MS:
@darzu can you review?

darzu (approved):
LGTM

PR #7793, opened 2021-01-15
Don't add forumExample card if forumUrl has otherActions

shakao:
right now we manually add a "forumExample" action card ("Open in Editor") to every "forumUrl" card type. this PR changes it so we skip this step if the forumUrl has otherActions present. will do a separate PR with CLI changes to output the featured games markdown in the new format, but that requires some additional tweaks to support the images, and is all run locally to generate the markdown. this should be all that's required for the release.

also makes no assumptions about what editor (blocks/js/py) the sharedExample should be opened in--if nothing is specified, it simply says "Open in Editor"

### DIFF
--- a/pxtlib/discourse.ts
+++ b/pxtlib/discourse.ts
@@ -53,7 +53,7 @@ namespace pxt.discourse {
 
     export function topicsByTag(apiUrl: string, tag: string): Promise<pxt.CodeCard[]> {
         apiUrl = apiUrl.replace(/\/$/, '');
-        const q = `${apiUrl}/tags/${tag}.json`;
+        const q = `${apiUrl}/tag/${tag}.json`;
         return pxt.Util.httpGetJsonAsync(q)
             .then((json: TagsResponse) => {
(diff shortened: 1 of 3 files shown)


pxtlib/discourse.ts
```diff
@@ -53,7 +53,7 @@ namespace pxt.discourse {
 
     export function topicsByTag(apiUrl: string, tag: string): Promise<pxt.CodeCard[]> {
         apiUrl = apiUrl.replace(/\/$/, '');
-        const q = `${apiUrl}/tags/${tag}.json`;
```
jwunderl, line 56: To confirm, nothing changed on discourse side / this was just the wrong api it looks like, right? https://docs.discourse.org/#tag/Tags/paths/~1tags/get

jwunderl (approved):

livcheerful (approved):

pxtlib/discourse.ts
shakao: i've run this command before in the past so i think it used to work or they were more forgiving?

PR #7804, opened 2021-01-20
Identity UI updates

eanders-MS:
### Summary
* Update experiment description and add link to feedback.
* Add "experimental feature ahead!" warning to sign in dialog.
* When a user deletes their account, convert cloud projects to local.
* Update delete account to mention the above behavior.
* Add feedback link on My Profile screen.

### Feedback panel
image

### Updated delete account verbiage
image

### Warning on sign in
image


webapp/src/auth.ts
```diff
@@ -394,13 +395,21 @@ export async function updateUserProfile(opts: {
...
     pxt.storage.removeLocal(CSRF_TOKEN);
 
+    // Convert cloud-saved projects to local projects.
+    await cloud.convertCloudToLocal(userId);
+
     // Update state and UI to reflect logged out state.
     clearState();
+
+    // Reload page
+    window.location.reload();
```
eanders-MS, line 412: This reload can be removed once the project list refresh issue is fixed.

theme/common.less
```diff
@@ -1713,6 +1713,11 @@ div.simframe.ui.embed {
     div.description {
         line-height: 1.1em;
         margin-bottom: 20px;
+
+        .warning {
```
darzu, line 1717: don't we have css for standard warnings? Is this special somehow?

webapp/src/cloud.ts
```diff
@@ -342,6 +342,23 @@ export function onChangesSynced(changes: Header[]) {
...
+        const localCloudHeaders = workspace.getHeaders(true)
+            .filter(h => h.cloudUserId && h.cloudUserId === userId);
+        const tasks: Promise<void>[] = [];
+        localCloudHeaders.forEach((h) => {
+            // Clear cloud header and re-save the header.
+            delete h.cloudCurrent;
+            delete h.cloudLastSyncTime;
+            delete h.cloudUserId;
+            delete h.cloudVersion;
+            tasks.push(workspace.saveAsync(h));
```
darzu, line 356: saveAsync should be called like: `saveAsync(h, null, true)` in this case. The last parameter, fromCloudSync, indicates this isn't a user change and we shouldn't touch the timestamps on the header.

theme/common.less
eanders-MS: Yes, probably, but this warning message is temporary and will be removed once the feature is out of experiment. If it were going to be long-lived, I would for sure want to use a standard css class.

theme/common.less
```diff
@@ -1805,6 +1815,15 @@ div.simframe.ui.embed {
                 justify-self: start;
                 max-width: 100%;
             }
+
+            div.row2 {
```
darzu, line 1819: I don't suppose there is a more descriptive name we could use?

darzu (approved):
LGTM

theme/common.less
eanders-MS: That's fair :)

PR #7805, opened 2021-01-20
Filter assets by tag in gallery

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/2901

- Added fetching tags for assets
- Added filtering by tags 

I copied the tag filtering from https://github.com/microsoft/pxt/blob/9e24403a17e2df8aa2261572b653662dbc9cbf5d/pxtlib/spriteutils.ts#L414

Also the dialogs gallery looks like this now, I can't remember if that's how it was before or if this is what we want?
image

### DIFF
--- a/webapp/src/components/ImageFieldEditor.tsx
+++ b/webapp/src/components/ImageFieldEditor.tsx
@@ -189,9 +189,10 @@ export class ImageFieldEditor<U extends pxt.Asset> extends React.Component<Image
         if (filterString) {
             assets.forEach(a => {
                 if (!a.meta.tags && this.options) {
-                    a.meta.tags = this.blocksInfo.apis.byQName[a.id]?.attributes.tags?.split(" ") || [""];
+                    a.meta.tags = this.blocksInfo.apis.byQName[a.id]?.attributes.tags?.split(" ") || [];
                 }})
 
+        // Keep tag filtering unified with pxtlib/spriteutils:filterItems
             const tags = filterString.split(" ")
                 .filter(el => !!el)
                 .map(el => el.toLowerCase());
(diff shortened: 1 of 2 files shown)


webapp/src/components/ImageFieldEditor.tsx
```diff
@@ -186,6 +186,24 @@ export class ImageFieldEditor<U extends pxt.Asset> extends React.Component<Image
             assets = assets.map(t => (t.type !== this.asset.type || t.id !== this.asset.id) ? t : assetToGalleryItem(this.getValue()))
         }
 
+        if (filterString) {
+            assets.forEach(a => {
+                if (!a.meta.tags && this.options) {
+                    a.meta.tags = this.blocksInfo.apis.byQName[a.id]?.attributes.tags?.split(" ") || [""];
```
jwunderl, line 192:
```suggestion
                    a.meta.tags = this.blocksInfo.apis.byQName[a.id]?.attributes.tags?.split(" ") || [];
```
nit; should be fine to just do []?

webapp/src/components/ImageFieldEditor.tsx
```diff
@@ -210,6 +228,23 @@ export class ImageFieldEditor<U extends pxt.Asset> extends React.Component<Image
                     return assets.filter(t => t.type === pxt.AssetType.Tilemap);
             }
         }
+
+        function checkInclude(item: pxt.Asset, includeTags: string[]) {
+            const tags = item.meta.tags ? item.meta.tags: [""];
```
jwunderl, line 233:
```suggestion
            const tags = item.meta.tags ? item.meta.tags: [];
```

webapp/src/components/ImageFieldEditor.tsx
```diff
@@ -210,6 +228,23 @@ export class ImageFieldEditor<U extends pxt.Asset> extends React.Component<Image
...
+            return includeTags.every(filterTag => {
+                const optFilterTag = `?${filterTag}`;
+                return tags.some(tag =>
+                    tag === filterTag || tag === optFilterTag
+                )
+            });
+        }
+
+        function checkExclude(item: pxt.Asset, excludeTags: string[]) {
+            const tags = item.meta.tags? item.meta.tags: [""];
```
jwunderl, line 243:
```suggestion
            const tags = item.meta.tags? item.meta.tags: [];
```

webapp/src/components/ImageFieldEditor.tsx
```diff
@@ -186,6 +186,24 @@ export class ImageFieldEditor<U extends pxt.Asset> extends React.Component<Image
             assets = assets.map(t => (t.type !== this.asset.type || t.id !== this.asset.id) ? t : assetToGalleryItem(this.getValue()))
         }
 
+        if (filterString) {
+            assets.forEach(a => {
+                if (!a.meta.tags && this.options) {
+                    a.meta.tags = this.blocksInfo.apis.byQName[a.id]?.attributes.tags?.split(" ") || [""];
+                }})
+
+            const tags = filterString.split(" ")
```
jwunderl, line 195: It looks like the pxt.sprite.filter version is used in a few spots and would be a bit tedious to dedup due to the different interfaces so probably fine, but if we're duplicating this would be good to add a comment in both spots pointing at eachother to avoid them getting out of sync

jwunderl (approved):

jwunderl (approved):

PR #7810, opened 2021-01-21
Add flipping and rotating to the image/tilemap/animation editor

riknoll:
What....? Could it be....?? IT IS!!! Flip and rotate are finally in the image editor!!!!!

They both work basically like you would expect. If you select something with the marquee tool, then only that part of the image is flipped/rotated. Other random things:

1. If you are in the tilemap editor, walls are also rotated (just like how we handle them in the marquee tool)
2. In the animation editor this will only rotate the current frame and will not resize the image. If anything would be cut off (e.g. width != height), then the image gets converted into the floating layer so nothing is lost
3. You can also now click outside the canvas to clear the marquee selection (this always bothered me)

And the new keyboard shortcuts:
* **SHIFT + V** is flip vertical
* **SHIFT + H** is flip horizontal
* **[** is rotate counterclockwise
* **]** is rotate clockwise

All of this is pretty self contained (except the buttons), so should be safe to merge.
Thanks to @soniakandah for the icons and @shakao for putting the buttons in the editor.


abchatra:
Alt Text

webapp/src/components/ImageEditor/toolDefinitions.ts
```diff
@@ -727,3 +727,102 @@ export class MarqueeEdit extends SelectionEdit {
...
+        }
+    }
+}
+
+export function flipEdit(image: EditState, vertical: boolean, isTilemap: boolean) {
+    const source = image.floating || image;
+
+    const newImage = isTilemap ? new pxt.sprite.Tilemap(source.image.width, source.image.height) :
+        new pxt.sprite.Bitmap(source.image.width, source.image.height);
+    const newOverlayLayers = source.overlayLayers ? source.overlayLayers.map(layer => new pxt.sprite.Bitmap(layer.width, layer.height)) : undefined;
```
jwunderl, line 794:
```suggestion
    const newOverlayLayers = source.overlayLayers?.map(layer => new pxt.sprite.Bitmap(layer.width, layer.height));
```

jwunderl (approved):
nice! LGTM

PR #7812, opened 2021-01-21
Show tile assets if editing a tile

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/2922

also don't always filter for tags because user assets don't have tags

### DIFF
--- a/webapp/src/components/ImageFieldEditor.tsx
+++ b/webapp/src/components/ImageFieldEditor.tsx
@@ -81,7 +81,8 @@ export class ImageFieldEditor<U extends pxt.Asset> extends React.Component<Image
             <div className="image-editor-gallery-content">
                 <ImageEditor ref="image-editor" singleFrame={this.props.singleFrame} onDoneClicked={this.onDoneClick} onTileEditorOpenClose={this.onTileEditorOpenClose} />
                 <ImageEditorGallery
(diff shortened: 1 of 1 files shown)


jwunderl (approved):

webapp/src/components/ImageFieldEditor.tsx
```diff
@@ -224,6 +224,7 @@ export class ImageFieldEditor<U extends pxt.Asset> extends React.Component<Image
             );
         }
 
+        type = this.state.editingTile ? pxt.AssetType.Tile : type;
```
shakao, line 227: is the wrong asset type being passed into this function?

livcheerful: oh it thinks it's still in the Tilemap editor so it's looking at the wrong asset

shakao: ah! hmm maybe this could be passed in in the this.filterAssets(this.userAssets) call like it is for gallery assets? or you could remove it from the gallery assets call and just use this

livcheerful: oh yeah that makes sense! I'll do that. thanks shannon :D

shakao (approved):

PR #7822, opened 2021-01-22
Hide my assets option when editing a tile or in asset editor

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/2963

The "My Assets" tab only really makes sense when you are in a block, because it replaces the current thing you're editing with a reference to the asset you selected. We don't really have a concept of references in the asset editor, so remove the tab altogether. This also removes it when editing a tile in the tilemap editor which has the same issue.


riknoll:
Also, this means the header is completely removed if you edit a tilemap from the asset editor (we already hid the gallery, so there would be a toggle with a single "editor" option which doesn't make much sense)

jwunderl (approved):

livcheerful (approved):

PR #7829, opened 2021-01-25
Improve intellisense completions with complex qualifiers

darzu:
Before:

After:

It's not a perfect solution, but it's better than what was.

### DIFF
--- a/webapp/src/monaco.tsx
+++ b/webapp/src/monaco.tsx
@@ -124,10 +124,12 @@ class CompletionProvider implements monaco.languages.CompletionItemProvider {
                         // remove what precedes the "." in the full snippet.
                         // E.g. if the user is typing "mobs.", we want to complete with "spawn" (name) not "mobs.spawn" (qName)
                         if (completions.isMemberCompletion && snippet) {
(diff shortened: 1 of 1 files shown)


pelikhan:
Outdated?

webapp/src/monaco.tsx
```diff
@@ -124,10 +124,12 @@ class CompletionProvider implements monaco.languages.CompletionItemProvider {
                         // remove what precedes the "." in the full snippet.
                         // E.g. if the user is typing "mobs.", we want to complete with "spawn" (name) not "mobs.spawn" (qName)
                         if (completions.isMemberCompletion && snippet) {
-                            const nameStart = snippet.lastIndexOf(name);
-                            if (nameStart !== -1) {
-                                snippet = snippet.substr(nameStart)
+                            let snippetQualifiers = getQualifiers(snippet);
+                            if (qName.indexOf("anyButton") >= 0) {
+                                console.log("snippet");
+                                console.dir({ snippet, snippetQualifiers })
```
darzu, line 130: TODO: cleanup

darzu:
Might still be needed for .py work

PR #7832, opened 2021-01-25
fix simulator iframe origin in electron app

jwunderl:
Right now builds of the electron app are crashing immediately as simulator gets packaged as `/simulator.html` (specifically, it's crashing on the new URL(furl) here  https://github.com/microsoft/pxt/blob/569b5e6a6ccc604c29cf9c4447861a894b0ffc67/pxtsim/simdriver.ts#L393-L397 .

Crashing here is no fun, and we have to assume the path is valid for the simulator to work in the first place, so if it fails to parse then just return it as relative to the local origin

(currently electron app is failing to create any file).

it looks like there's a separate error on production builds that is occurring as well which I'll look at, but for now this should let us do local tests / make sure things like asset editor are working as intended.


riknoll (approved):
Is the relative error case just for local host?

jwunderl:
Should be localhost, staticpkg (electron), maybe the old windows app as well (I don't know much about that one but I'm assuming it's also effectively just wrapping a staticpkg). The crash was introduced with https://github.com/microsoft/pxt/pull/7720/files so it isn't released anywhere else yet

PR #7836, opened 2021-01-26
Improve UX for cloud save

darzu:
- in the editor, shows only temporary "saving..." and "saved!" messages while/after a cloud save happened.
- Remove the double page reload after signing in
- Projects list in the Home Screen and "All Projects" view sync correctly and without a reload
- Fix several bugs around the virtual APIs
- Fix bugs around project delete not working right with the cloud

Fixes: https://github.com/microsoft/pxt-arcade/issues/2988
Fixes: https://github.com/microsoft/pxt-arcade/issues/3017

In editor "saving..." and "saved!" message:

In project view dynamic syncing and deleting without a reload!


darzu:
Pausing this as we re-evaluate priorities.

darzu:
Fixes: https://github.com/microsoft/pxt-arcade/issues/2988

darzu:
progress:

darzu:
Trying a dimmer color:

webapp/src/app.tsx
```diff
@@ -1332,7 +1332,8 @@ export class ProjectView
                                 // We are too late; the editor has already been loaded.
                                 // Call the onChanges handler to update the editor.
                                 pxt.tickEvent(`identity.syncOnProjectOpen.timedout`, { 'elapsedSec': elapsed})
-                                cloud.onChangesSynced(changes)
+                                if (changes.length)
+                                    cloud.forceReloadForCloudSync()
```
eanders-MS, line 1336: Can this call trigger a reload while in the editor? I've seen the toast "Cloud synchronization complete. Reloading" happen in the editor, which was jarring. Logged it here: https://github.com/microsoft/pxt-arcade/issues/3005

eanders-MS (approved):
I know this work is still in progress, but this looks like a good merge point if it is stable enough.

darzu:
> I know this work is still in progress, but this looks like a good merge point if it is stable enough.

Yes, this is meant to be a complete PR. Future work will be in a future PR.

webapp/src/app.tsx
darzu: Yes, it can. Unfortunately I don't know a good way to avoid this.

The problem is that if the user opens up a project at version A but after that we sync with the cloud and determine they should be fast-forwarded to version B, then we need to reload the editor from underneath them. Currently, there isn't a way to reload the editor's content without doing a page reload. Likely we could fix this using virtual APIs but it's another chunk of work..

webapp/src/cloud.ts
```diff
@@ -216,31 +252,55 @@ async function syncAsyncInternal(hdrs?: Header[]): Promise<Header[]> {
...
+            const remote = remoteHeadersToProcess[local.id];
+            delete remoteHeadersToProcess[local.id];
             if (remote) {
                 local.cloudLastSyncTime = remote.cloudLastSyncTime
                 // Note that we use modification time to detect differences. If we had full (or partial) history, we could
                 //  use version numbers. However we cannot currently use etags since the Cosmos list operations
                 //  don't return etags per-version. And because of how etags work, the record itself can never
                 //  have the latest etag version.
-                if (local.modificationTime !== remote.modificationTime) {
+                if (local.modificationTime !== remote.modificationTime || local.isDeleted !== remote.isDeleted) {
```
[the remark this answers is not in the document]
darzu: Yes, I'm pro switching to etags. It'll take a cloud deploy.

PR #7841, opened 2021-01-28
fix diagnostic ticks

jwunderl:
I think this should fix the diagnostics ticks so they're queryable? Figured should do this before we forget~


abchatra (approved):

abchatra:
Thanks @jwunderl

PR #7842, opened 2021-01-28
Disable identity in offline app

jwunderl:
For offline app, work needs to be done to enable oauth / etc (similar to github integration); disabling for now and we can reassess later on as needed. (I'm also filtering out the option for identity & a few github experiments from the experiments list to avoid potential confusion / people trying to enable when they can't)


eanders-MS (approved):
:shipit:

pxteditor/experiments.ts
```diff
@@ -3,7 +3,8 @@ namespace pxt.editor.experiments {
         id: string; // == field in apptheme also assumes image at /static/experiments/ID.png
         name: string;
         description: string;
-        feedbackUrl: string; // allows user to put feedback
+        feedbackUrl?: string; // allows user to put feedback
+        disableOffline?: boolean;
```
darzu, line 7: The double negative is a little confusing... can this just be called enableOnline?

darzu (approved):

pxteditor/experiments.ts
jwunderl: Ah good point, thanks! I did one more minor change as well to remove a cast, as casting actually removes the type checking for the experiments there (instead of telling compiler to confirm they are Experiments, it tells the compiler that you confirm they are Experiments)

PR #7845, opened 2021-01-29
Put plus/minus buttons at the end of the input list for mutations

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/2979

The issue was the inputs for the block and the secret shadow block in https://github.com/microsoft/pxt-blockly/blob/8496a44d4ffeddc24affebd91d509773fbb29209/core/insertion_marker_manager.js#L273 weren't ordered the same, so sometimes it would try to set values on things that didn't exist :\ it's weird that it worked for so many blocks???

Anyways this change makes the plus/minus buttons just be at the end of the inputList always. I don't 100% understand mutations still so let me know if this is bad -- I tested this with all the arcade blocks w/ the expanding (I also was going to try microbit/minecraft but it looks like there aren't that many for them?)

Try it here!: https://arcade.makecode.com/app/c844ab3378a131c5ee565b297df9c92ad13d49f5-0e1c9eee29

Also I changed it from showAdd to showPlus because addPlusButton was confusing me ><


jwunderl (approved):
LGTM

Should be able to remove https://github.com/microsoft/pxt/blob/ebb4c04857603419851adfb538f3b509b8b6d199/pxtblocks/composablemutations.ts#L129-L130

and the two assignments to those vars in `addPlusButton` / `addMinusButton` as well

shakao:
looks good! should we also add a check to make sure the input field exists before setting it in the insertion marker manager? it will make it harder to notice errors like this with the insertion marker blocks but also i think it's not a HUGE deal if the insertion block is slightly the wrong size... up to you!

shakao (approved):

livcheerful:
yeah I'll do that!

PR #7848, opened 2021-02-01
Clean up temporary asset when exiting field editor

shakao:
When going from temporary asset -> named asset in the image/animation editor, clean up the previous temporary asset. This will delete it from the project but you can get it back using undo/redo if it was a mistake.

Fixes https://github.com/microsoft/pxt-arcade/issues/2987

### DIFF
--- a/pxtblocks/fields/field_asset.ts
+++ b/pxtblocks/fields/field_asset.ts
@@ -113,6 +113,8 @@ namespace pxtblockly {
                     if (pxt.assetEquals(this.asset, result)) return;
 
                     this.pendingEdit = true;
+
+                    if (result.meta?.displayName) this.disposeOfTemporaryAsset();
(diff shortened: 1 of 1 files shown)


livcheerful (approved):
Is it undo/redo within the image editor or in the blocks editor?

jwunderl (approved):

shakao:
@livcheerful both kind of--if you draw a temporary asset and click on a named one you can undo (while still in the image editor). but once you go out, the workspace undo won't affect it. if you've run the project with a temporary asset and you go in and replace it with an named one and exit the image editor, the workspace undo WILL change it back--in this case it sees that the block has a changed value and is undoing that.

livcheerful:
oooo okay cool ty

riknoll (approved):

PR #7851, opened 2021-02-02
flag to supress bundled pkg from default extensions view

jwunderl:
For https://github.com/microsoft/pxt-microbit/issues/3854 we don't want the extension to show up by default, but it should be available for people who have content based around it / that it actually works for. Adds a flag to suppress bundled packages from the default extensions view, while still leaving them available if you search for them.


abchatra:
I don't think there is any content dependent on it. I would just say "mark this extension hidden". We don't want user to use it.

jwunderl:
Ah, okay, will just mark that as hidden then~

PR #7854, opened 2021-02-02
parse new format first

pelikhan:
Fix for https://github.com/microsoft/pxt-arcade/issues/3020#issuecomment-771888706

Try to parse cards HTML as UL/OL tree; and bail out to JSON if failing.


jwunderl (approved):

pelikhan:
Would be good to cherry pick this one into arcade.

PR #7855, opened 2021-02-02
Use tile id if no display name present

shakao:
fixes https://github.com/microsoft/pxt-arcade/issues/3018

the tilemap extension doesn't have display names for tiles and it was filling in eg assets.tile\`tiles.util.door2\` which is not a real tile

### DIFF
--- a/pxtblocks/fields/field_tileset.ts
+++ b/pxtblocks/fields/field_tileset.ts
@@ -94,7 +94,7 @@ namespace pxtblockly {
         getValue() {
             if (this.selectedOption_) {
                 const tile = this.selectedOption_[2];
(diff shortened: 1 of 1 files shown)


jwunderl (approved):
LGTM but I prefer parentheses if possible :)

livcheerful (approved):

riknoll (changes requested):
This should check the prefix of the qname instead of always defaulting to inserting the ID. If it starts with the "myTiles" namespace, it should use the short id

riknoll:
maybe should fix this function to return undefined or something: https://github.com/microsoft/pxt/blob/master/pxtlib/tilemap.ts#L1498

riknoll:
Wait, I'll unify the code for this in my PR

PR #7856, opened 2021-02-03
When syncing projects, prefer to use the project version returned from cloud list

darzu:
Due to some confusion, we weren't returning version tags from the project list call. So we used the modification time and delete status to check for differences between the local project and the cloud project.

Now thanks to: https://github.com/microsoft/pxt-backend/pull/364
project versions are returned by list and can be used instead.

Because of how this is written, this is backwards compatible so that this PR doesn't need the backend to be updated first. So this is safe to merge.


eanders-MS (approved):
:shipit:

PR #7858, opened 2021-02-03
check shortId as well when creating new id

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/3026, easy repro in https://github.com/microsoft/pxt-arcade/issues/3026#issuecomment-772104528

Check the shortId as well when generating checking whether a name is taken, to avoid overlapping between shortId & display names

image

Also changes default display name from `tile` to ~`customTile`~ see comment \\/, to minimize overlapping between displayName & shortId when possible

image

### DIFF
--- a/pxtlib/tilemap.ts
+++ b/pxtlib/tilemap.ts
@@ -714,7 +714,10 @@ namespace pxt {
                 }
             }
 
-            return isTaken(name) || isTaken(getShortIDCore(assetType, name));
+            const shortId = getShortIDCore(assetType, name);
+            const checkShortId = shortId && shortId !== name;
+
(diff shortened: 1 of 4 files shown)


riknoll (approved):

jwunderl:
Made two more minor changes; handle getShortIdCore returning undefined or same as name without rechecking for those (for future safety / comment in https://github.com/microsoft/pxt/pull/7855#issuecomment-772691948 seems reasonable), and also avoid having same displayname schema as shortId for animations, images for same reason as tiles (when made from asset editor) - so `myAnim`, `myImage`, `myTile`

shakao (approved):

PR #7860, opened 2021-02-03
Unite asset emitting code paths and fix tile parsing

riknoll:
Fixes  microsoft/pxt-arcade#2986
Fixes microsoft/pxt-arcade#3018

This unites all of the asset emitting and all of the parsing except in the monaco fields (which have some special behavior around creating new assets if the name does not exist)

@shakao you should be able to close https://github.com/microsoft/pxt/pull/7855

@jwunderl this might conflict with https://github.com/microsoft/pxt/pull/7858 because getShortIDCore can now return undefined.

I think I tested this pretty thoroughly, but please review carefully because it touches some key stuff

### DIFF
--- a/pxtblocks/fields/field_sprite.ts
+++ b/pxtblocks/fields/field_sprite.ts
@@ -34,18 +34,9 @@ namespace pxtblockly {
         protected createNewAsset(text?: string): pxt.Asset {
             const project = pxt.react.getTilemapProject();
             if (text) {
-                const match = /^\s*assets\s*\.\s*(image|tile)\s*`([^`]+)`\s*$/.exec(text);
-                if (match) {
-                    const asset = project.lookupAssetByName(
-                        match[1] === "image" ? pxt.AssetType.Image : pxt.AssetType.Tile, match[2].trim()
-                    );
-                    if (asset) return asset;
-                    else if (!this.getBlockData()) {
-                        this.isGreyBlock = true;
-                        this.valueText = text;
-                        return undefined;
-                    }
-                }
+                const asset = pxt.lookupProjectAssetByTSReference(text, project);
+
+                if (asset) return asset;
             }
 
             if (this.getBlockData()) {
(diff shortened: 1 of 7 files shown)


pxtblocks/fields/field_sprite.ts
```diff
@@ -34,16 +34,9 @@ namespace pxtblockly {
         protected createNewAsset(text?: string): pxt.Asset {
             const project = pxt.react.getTilemapProject();
             if (text) {
-                const match = /^\s*assets\s*\.\s*image\s*`([^`]+)`\s*$/.exec(text);
-                if (match) {
-                    const asset = project.lookupAssetByName(pxt.AssetType.Image, match[1].trim());
-                    if (asset) return asset;
-                    else if (!this.getBlockData()) {
-                        this.isGreyBlock = true;
```
jwunderl, line 42: Does it not need to handle grey block here anymore?

jwunderl (commented):
one comment / question but otherwise LGTM

shouldn't conflict with my pr, already made a change to handle undefined when I saw your comment on the other one: https://github.com/microsoft/pxt/pull/7858/commits/2f77692a2cf5cd80bf125cd510c8d10611a12951

(and in case anyone else always checks longs regexes by hand just to make sure, here's a quick link with a couple test cases I made https://regex101.com/r/XBc5vB/1)

pxtblocks/fields/field_sprite.ts
riknoll: I think this should get caught by the one below

jwunderl (approved):

riknoll:
Pushed a quick change to fix python field editors and decompiling from python

pxtblocks/fields/field_tileset.ts
```diff
@@ -94,7 +94,7 @@ namespace pxtblockly {
         getValue() {
             if (this.selectedOption_) {
                 const tile = this.selectedOption_[2];
-                return isGalleryTile(tile) ? tile.id : `assets.tile\`${displayName(tile)}\``
```
shakao, line 97: can `isGalleryTile` be deleted?

shakao (approved):

PR #7861, opened 2021-02-04
compile / decompile emojis

jwunderl:
I like using emojis for examples sometimes but it gets bad when you compile (just some combination of `_` and numbers), so this would let allow emojis to persist between compile -> decompile:

shiny shiny

Obviously the generated names aren't great (they're just "Ex" + emoji's codepoint in base 16) but I don't know it's any worse than just `_` / `_2` / etc; starting as a draft though as I don't know if we'd want to take it necessarily for this reason

test build with a sample project: https://arcade.makecode.com/app/eed3a5d454d55f0f6587cf8534d40ceb3b7021bd-6b430144ba#pub:_PrbKdTAmkYdk 

I messed up switching between branches after my local version of the branch broke and accidentally pulled when I should have checkout'd, so I accidentally merged this into master / reverted it already D:. The commits from when I wrote it are in https://github.com/microsoft/pxt/compare/8e3181e...bb4281a, and I reverted that in https://github.com/microsoft/pxt/commit/596e90c79336f81c1a086ec05149ec8bdc544ff5, so this pr would... revert that.


tests/decompile-test/baselines/tagged_templates.blocks
```diff
@@ -15,12 +15,18 @@
 <value name="VALUE">
 <shadow type="math_number"><field name="NUM">0</field></shadow>
 <block type="spritescreate">
-<field name="img">img&#38;&#35;96&#59;&#38;&#35;10&#59;0&#38;&#35;10&#59;&#38;&#35;10&#59;&#38;&#35;10&#59;1&#38;&#35;10&#59;2&#38;&#35;10&#59;3&#38;&#35;96&#59;</field>
+<field name="img">img`
```
riknoll, line 18: this doesn't look right... shouldn't newlines get escaped?

riknoll:
i wonder if we could do a mapping to the unicode names instead? Might look a bit nicer (but would definitely be more complicated).

tests/decompile-test/baselines/tagged_templates.blocks
jwunderl: My understanding is that they shouldn't need to be escaped in xml like that re: https://www.w3.org/TR/xml-c14n/#Example-WhitespaceInContent 

I believe this example would break on decompiling the image if it weren't being read correctly, right? As the https://arcade.makecode.com/app/0610dda95754bed7e6df6ce7485f7d128dd0f4ea-1375337f35#pub:_Rd0UJmiP0RWq

can always add a few more values to be escaped if we're worried about it, would probably be better to just use https://www.npmjs.com/package/he if we want the full solution though as the current escaping is broken -- e.g. it's causing crashes due to malformed encodings here https://github.com/microsoft/pxt-arcade/issues/2774

riknoll: iiirc we had bugs around that before. can't remember if it was us or blockly

jwunderl:
For using the unicode names, I believe the values should be the same in general -- that is, it's currently encoding `U+1f4da` as `E1f4daX` in the name just to make it easier to escape / not run into a case where it tries to swap a users actual intended variable name with an emoji that happens to match. I can make it `U1f4da$` or something instead (just want to keep a marker on each side so it's definite which things are part of emoji and which are just letters that are valid hexchars) -- only needs changing like this: https://github.com/microsoft/pxt/pull/7861/commits/db107f0cc42c068a613aee5c8d3b0b8688586ea4

riknoll:
@jwunderl I meant the name of the emoji. e.g. "grinningFace" and "personInSuitLevitating"

jwunderl:
Okay finally brought this up to date, I'm gonna try and get this in for next arcade release :) here's a test build, everything appears to still be working but I'll poke around a bit more 

https://arcade.makecode.com/app/ef36c7fc259238e12d680c626d4af6c6bbf1b771-418c08602e#pub:_Rd0UJmiP0RWq

abchatra:
Is this PR still valid?

PR #7862, opened 2021-02-04
Conflict resolution for loosely versioned dependencies.

pelikhan:
When add a new github dependency without a tag, default to existing tag if already reference. Eg if adding "foo/bar", but "foo/bar#0.0.0" is already reference, use "foo/bar#0.0.0".

### DIFF
--- a/pxtlib/package.ts
+++ b/pxtlib/package.ts
@@ -405,11 +405,29 @@ namespace pxt {
                                     }
                                 }
                             }
-                            if (!foundYottaConflict && pkgCfg.name === depPkg.id && depPkg._verspec != version && !/^file:/.test(depPkg._verspec) && !/^file:/.test(version)) {
-                                const conflict = new cpp.PkgConflictError(lf("version mismatch for extension {0} (installed: {1}, installing: {2})", depPkg, depPkg._verspec, version));
-                                conflict.pkg0 = depPkg;
-                                conflict.isVersionConflict = true;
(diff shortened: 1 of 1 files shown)


pxtlib/package.ts
```diff
@@ -405,11 +405,30 @@ namespace pxt {
...
+                                // check if versions are semver compatible for github refs
+                                const ghCurrent = /^github:/.test(depPkg._verspec)
+                                    && pxt.github.parseRepoId(depPkg._verspec);
+                                const ghNew = /^github:/.test(version)
+                                    && pxt.github.parseRepoId(version);
+                                if (!ghCurrent || !ghNew // only for github refs
+                                    || ghCurrent.fullName !== ghNew.fullName // must be same extension
+                                    // if newversion does not have tag, it's ok
+                                    // note: we are upgrade major versions as well
+                                    || (ghNew.tag && pxt.semver.strcmp(ghCurrent.tag, ghNew.tag) < 0)) {
```
mmoskal, line 423: aren't `ghNew` and `ghCurrent` symmetric?

pelikhan: you mean you want to automitcally upgrade the reference and take the highest?

mmoskal: ghNew/ghCurrent are just different references in the project. I think they can be swapped, essentially randomly. It's new/current only in the sense that we got to it first. Or am I wrong?

PR #7866, opened 2021-02-04
proper check of org for github

pelikhan:
Using the wrong API to query if a repo is from an org or a user.


jwunderl (approved):
LGTM, maybe should port this to arcade too?

pelikhan:
You hit it when you don't have rights... so yes.

PR #7867, opened 2021-02-05
Add "learn more" link on sign in dialog

eanders-MS:
Link takes the user to aka.ms/cloudsave, which will redirect to the forum post once we have the url.

image


eanders-MS:
I will cherry-pick this to stable once approved.

abchatra (approved):

PR #7872, opened 2021-02-08
add styling for tutorial hovertext

jwunderl:
Looks like this as written, but the selector is the more interesting part / easy to swap any styling in:

image

Always underlined + black + italic without any hover effect reads as "hover but don't click" to me :)


livcheerful:
Should we do talk about this in a design office hours? I just tried to click a hovertext and it was sad, so I remembered why we wanted this :D

Jaqster:
Yes, please! 👍

PR #7873, opened 2021-02-08
Add autoexpandOff metadata option for tutorial testing

shakao:
tutorial steps will default to "collapsed" ("Show more...") when the tutorial is first opened. as with current behavior, if a user clicks "Show more/less" we remember that state and will persist it with that project.

to use, add this to the start of a tutorial:

```
### @autoexpandOff true
```

for https://github.com/microsoft/pxt-arcade/issues/3047


jwunderl (approved):
If we're just doing for internal testing, maybe query parameter would be better for a/b testing? Either way seems fine to me~

shakao:
hmm good point, let me sync with kiki--i don't think a query parameter will work inside the skillmap but if these are standalone tutorials that would probably be better

shakao:
chatted with kiki, sticking with the markdown flag for now since they're not doing full a/b testing and this keeps the change pretty small/contained

PR #7876, opened 2021-02-09
Minecraft UI fixes

livcheerful:
fixes https://github.com/microsoft/pxt-minecraft/issues/2109

- Change modal confirmation button from 'Start' -> 'OK'
- Fix workspace creeping on the top toolbar
- Add classes to the modal to only use HOC styling on tutorials with hideIteration used

### DIFF
--- a/webapp/src/tutorial.tsx
+++ b/webapp/src/tutorial.tsx
@@ -272,6 +272,8 @@ export class TutorialHint extends data.Component<ISettingsProps, TutorialHintSta
     renderCore() {
         const { visible } = this.state;
         const options = this.props.parent.state.tutorialOptions;
+        const flyoutOnly = this.props.parent.state.editorState && this.props.parent.state.editorState.hasCategories === false;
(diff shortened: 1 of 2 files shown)


webapp/src/tutorial.tsx
```diff
@@ -288,7 +288,7 @@ export class TutorialHint extends data.Component<ISettingsProps, TutorialHintSta
         } else {
             let onClick = tutorialStep < tutorialStepInfo.length - 1 ? this.next : this.closeHint;
             const actions: sui.ModalButton[] = [{
-                label: lf("Start"),
+                label: lf("Ok"),
```
jwunderl, line 291: Did we want this to be start for hoc still? / should this be `Start` when in hideiteration + flyoutonly?

jwunderl (approved):

shakao (approved):

webapp/src/tutorial.tsx
livcheerful: oh sure sure good point! I'll fix

PR #7877, opened 2021-02-09
set default color for note picker so it's not white notes on offwhite bkgd

jwunderl:
Just before stream we tried to use the block that is the notepicker on it's own and noticed it was messed up; the default color we get for the notepicker when it isn't slotted into another block is white / light gray, which makes it hard to see / very low contrast:

image

(sometimes as this: 
image)

The music category color is different depending on the color (orange in micro:bit, pink in arcade) -- I can make it try and grab the default but it's a bit tedious, and I kind of like the dark gray electric piano look anyways / the color of the piano changes whenever it's used anyways:

image

### DIFF
--- a/pxtblocks/fields/field_note.ts
+++ b/pxtblocks/fields/field_note.ts
@@ -369,9 +369,8 @@ namespace pxtblockly {
                 this.borderColour = b.getColourTertiary();
             }
             else {
-                let b = this.sourceBlock_ as Blockly.BlockSvg;
-                this.primaryColour = b.getColourTertiary();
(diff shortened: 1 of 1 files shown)


Jaqster:
How come this doesn't show the little callout connector?
image

jwunderl:
Oh, I think that's https://github.com/microsoft/pxt-arcade/issues/2895 (@shakao correct me if I'm wrong ~)

Jaqster:
Got it.  Looks good!

shakao (approved):

PR #7880, opened 2021-02-11
Adding empty card shadows in the asset editor gallery

riknoll:
Screen Shot 2021-02-11 at 11 11 56 AM

Don't love doing this in JavaScript but @shakao and I couldn't think of a clever way to do it in css. Originally I tried putting the number of empty cards in the react state but it was pretty buggy.

### DIFF
--- a/webapp/src/components/assetEditor/assetCardList.tsx
+++ b/webapp/src/components/assetEditor/assetCardList.tsx
@@ -76,7 +76,7 @@ export class AssetCardList extends React.Component<AssetCardListProps> {
         this.hasUpdatedCards = true;
 
         const columns = Math.floor(parentBounds.width / elementWidth);
-        const rows = Math.floor(this.container.parentElement.getBoundingClientRect().height/ elementWidth)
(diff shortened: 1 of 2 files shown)


webapp/src/components/assetEditor/assetCardList.tsx
```diff
@@ -8,11 +8,117 @@ interface AssetCardListProps {
...
+            const siblingBounds = child.nextElementSibling.getBoundingClientRect();
+
+            // Guard against a single column
+            elementWidth = Math.max(elementWidth, siblingBounds.left - childBounds.left);
+        }
+
+        this.hasUpdatedCards = true;
+
+        const columns = Math.floor(parentBounds.width / elementWidth);
+        const rows = Math.floor(this.container.parentElement.getBoundingClientRect().height/ elementWidth)
```
jwunderl, line 79:
```suggestion
        const rows = Math.floor(this.container.parentElement.getBoundingClientRect().height / elementWidth)
```

jwunderl (approved):

PR #7886, opened 2021-02-12
resolve conflicts by creating a copy; fix auto-save

darzu:
# Cloud Conflict Resolution
Resolves project conflicts (when the same user edits the same project on two browsers/computers without syncing first) by create a duplicate copy.

Here's a scenario:
1. User opens cloud project Foo v0 on computer A
2. User opens cloud project Foo v0 on computer B
3. User edits cloud project Foo on computer A which upload new version v1a to the cloud
4. User edits cloud project Foo on computer B which tries and fails to upload version v1b to the cloud

(Importantly, a cloud sync on computer B didn't happen between steps 3 and 4.)

step 4 is where we've detected a conflict. At this point we do the following:
1. Create a copy of Foo on computer B (called "Foo - Copy")
2. Switch the user's editor from "Foo" to "Foo - Copy" on computer B
3. Replace "Foo" on computer B with version v1a from the cloud
4. Save "Foo - Copy" to the cloud

Now both versions are uploaded to the cloud and there is no more conflict.

Before this PR, we just overwrote both the remote and local version with the last edited version (so v1b).

The technical crux is here: https://github.com/microsoft/pxt/pull/7886/files#diff-16f87a7e55e24fbb6bf8f4b42bd5491b13ac71ae41ff1618a08116e62133c275R204

Example dialog to user:
Screen Shot 2021-02-25 at 2 09 14 PM

Fixes: https://github.com/microsoft/pxt-arcade/issues/2990
Fixes: https://github.com/microsoft/pxt-arcade/issues/3108
Fixes: https://github.com/microsoft/pxt-arcade/issues/3129
Fixes: https://github.com/microsoft/pxt-arcade/issues/3173

## Interactions with GitHub

Assume you have a github project open in two computers, both linked to the same github repo and both linked to the same cloud project. If you encounter a cloud project conflict, we have the same behavior of creating a copy with your local conflict changes. Both of these projects will link to the same github repo and both can push and pull to the github repo just fine.

### auto-save fixes
Before, many places in the code were explicitly calling "saveToCloud" or "cloudSaveAsync" to trigger a push to the cloud. This was problem because this approach requires that callsites call these functions in all the right places and also not too much. Turns out both were wrong: these function were being called way too often in many cases and also not at all in certain important cases. Futher, not all of these callsites were debouncing, only those that went through package.ts did and even those often used a "immediate" flag to force a save. This resulted in unnecessary work but also cases were changes weren't being synced to the cloud. 

So I revamped how cloud auto-saving works. Instead we use the virtual APIs notification system to subscribe to all change to the "header:*" path. Then we maintain our own list if headers with pending saves and invoke a save after 3 seconds of inactivity or 15 seconds maximum.

This has the benefit of centralizing our cloud sync logic, reducing surface area, and giving more consistent and complete guarantees about save.

### virtual APIs support "path" param
To make the above changes possible, I enhanced the virtual API system so that subscribers are notified of _which_ path has been invalidated in case the subscriber is listening to many paths or a glob of paths (e.g. "header:*").
 
## PR final TODOs:
- [x] hunting bug that has to do with inconsistent local state for the header's cloud local params
- [x] fix project rename being reverted
- [x] show popup or toast to user when conflict happens
- [x] verify conflict resolution happens correctly on home screen
  - [x] seems to be a bug where one project overwrites the other.. (https://github.com/microsoft/pxt-arcade/issues/3173) [fixed by: https://github.com/microsoft/pxt/pull/7886/commits/1d238132e01f81e35d52cd793f94780a257020e3]
- [x] use popup instead of toast
- [x] final verify 2990
- [x] final verify 3108
- [ ] final verify 3107 (Edit: nvm, 3107 is a deeper issue that this PR doesn't fully solve)
- [x] final verify 3129 (not solved, but mitigated enough to punt this until we do a workspace & package layer refactor)
- [x] test github repo clone

Also, investigate these to see if they're now fixed or need more work:
- [ ] https://github.com/microsoft/pxt-arcade/issues/3150
- [ ] https://github.com/microsoft/pxt-arcade/issues/3138


darzu:
- [ ] hunting bug that has to do with inconsistent local state for the header's cloud local params

darzu:
- [ ] fix project rename being reverted

webapp/src/cloud.ts
```diff
@@ -211,19 +193,88 @@ async function transferFromCloud(local: Header | null, remoteFile: File): Promis
     } else {
         await workspace.importAsync(newHeader, remoteFile.text, true)
     }
-    return newHeader
+    return workspace.getHeader(newHeader.id)
+}
+
+function getConflictCopyName(hdr: Header): string {
+    // TODO: do we want a better or more descriptive name?
+    return hdr.name + " - Copy";
```
darzu, line 201: Any thoughts on what we want to name the copy?

webapp/src/cloud.ts
```diff
@@ -418,6 +476,80 @@ export async function convertCloudToLocal(userId: string) {
     }
 }
 
+const CLOUDSAVE_DEBOUNCE_MS = 3000;
+const CLOUDSAVE_MAX_MS = 15000;
```
darzu, line 480: @eanders-MS , I re-worked the auto-save system entirely. Now we listen for "header:*" virtual API changes and then debounce them between 3-15 seconds.

I yanked out all the many places in the cloud that were making explicit cloud saves. Now instead the only ways to trigger a cloud save is via the virtual APIs or the syncAsync function.

One of the problems we had before is that we had to trust that callsites were calling "saveToCloud" at appropriate times and for good reason. And of course this wasn't true, it was being both overcalled in many places and also not called in important scenarios.

webapp/src/app.tsx
```diff
@@ -4601,7 +4602,7 @@ document.addEventListener("DOMContentLoaded", () => {
     else if (isSandbox) workspace.setupWorkspace("mem");
     else if (pxt.winrt.isWinRT()) workspace.setupWorkspace("uwp");
     else if (pxt.BrowserUtils.isIpcRenderer()) workspace.setupWorkspace("idb");
-    else if (pxt.BrowserUtils.isLocalHost() || pxt.BrowserUtils.isPxtElectron()) workspace.setupWorkspace("fs");
+    else if (pxt.BrowserUtils.isPxtElectron()) workspace.setupWorkspace("fs");
```
darzu, line 4605: TODO: allow optional workspace override var via local storage

webapp/src/app.tsx
```diff
@@ -76,7 +76,11 @@ let theEditor: ProjectView;
 let hash: { cmd: string, arg: string };
 let pendingEditorRequests: ((p: ProjectView) => void)[];
 
-function getEditorAsync() {
+export function hasEditor() {
+    return !!theEditor;
+}
+
+export function getEditorAsync() {
```
darzu, line 83: I'm really unsure about this approach. The goal is to detect if we have an editor open and if so allow getting a handle so we can change which project is being edited.

However, these weren't public before and it doesn't seem great to allow arbitrary parts of the code to reach and mutate the editor like this.

Yet I don't know of a better way with our current abstractions.

(This is another case where something like redux would be great: I'd like to change the currently edited project state and that should cascade automatically into updating the relevant editors if needed)

eanders-MS: Agreed this isn't great. Virtual api is our redux analog. Can it be used here to load the new header?

darzu: The problem is it doesn't support a quarter of the things proper state management in React or Redux would do. For example, each header load causes at least one save to happen as a biproduct which invalidates the header. If the editor reacted to header invalidations, we'd be in an endless loop. The virtual API invalidation is so noisy around headers that it really can't be relied upon for many tasks.

webapp/src/cloud.ts
eanders-MS: Nice cleanup.

eanders-MS (approved):
LGTM! Tested locally and couldn't break it :)

webapp/src/cloud.ts
```diff
@@ -248,15 +248,6 @@ async function resolveConflict(local: Header, remoteFile: File) {
         pxt.tickEvent(`identity.sync.conflict.reloadEditorFailed`, {exception: e});
     }
 
-    // 2a. tell the user a conflict occured
-    try {
-        core.infoNotification(lf(`Project '${local.name}' was editted in two places and the changes conflict. Creating backup copy...`));
```
eanders-MS: editted -> edited

eanders-MS (approved):
LGTM. Needs team bug bashing, so let's get it out to beta :)

PR #7890, opened 2021-02-12
Use github slug for tutorial project dependency

shakao:
i think the monorepo stuff gets a little weird when loading non-README tutorials (many of these are in folders, etc); this change fixes tutorial loading but i think it also means if you're loading tutorials from a monorepo it'll load the entire repository?

wanted to do a small change to fix this for current arcade since i don't think anyone is using the monorepo functionality for tutorial content yet

### DIFF
--- a/pxtlib/github.ts
+++ b/pxtlib/github.ts
@@ -983,9 +983,9 @@ namespace pxt.github {
         }
     }
 
-    export function toGithubDependencyPath(id: ParsedRepo): string {
-        let r = "github:" + id.fullName;
-        if (id.tag) r += "#" + id.tag;
+    export function toGithubDependencyPath(path: string, tag?: string): string {
+        let r = "github:" + path;
+        if (tag) r += "#" + tag;
         return r;
     }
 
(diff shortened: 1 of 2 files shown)


pelikhan:
If you use the slug, you will not load the sub package so this is potentially breaking since it only loads files listed in pxt.json.

What is the bug?

shakao:
it's this line: https://github.com/microsoft/pxt/pull/7890/files#diff-7047859ba8a434609145f1795eb0ee90ad1991dcdb4e22fc986bc181f2a55862L3553

if your repo has ts files and you load a non-README tutorial (eg `#tutorial:https://github.com/shakao-test/zookeeper-2/tutorials/activity1`) the dependency is: `zookeeper-2: github:shakao-test/zookeeper-2/tutorials/activity1`, which doesn't have a pxt.json. not sure if there's a way to distinguish "tutorial in a subfolder" from "tutorial in a subpackage"... i guess we could first check the full path and if there's no pxt.json, fall back to the slug?

pelikhan:
On a second thought, it's probably ok.

pelikhan (approved):

jwunderl (approved):

PR #7892, opened 2021-02-12
Check slug instead of full name for repo status

shakao:
fixes https://github.com/microsoft/pxt-arcade/issues/3077, will address https://github.com/microsoft/pxt-arcade/issues/3078 but leaving that one open as we don't actually check trusted/untrusted for individual tutorials and we should consider that


pelikhan:
I think the proper fix is to add each individual subfolder to the targetconfig.json.

PR #7896, opened 2021-02-16
Add filter panel to gallery

livcheerful:
This adds a button to the gallery view in the image editor to list all the relevant tags and to be able to filter by them. The categories are anything we tag with; if we want to add more tags than we want to show in categories (like for search), I can make a change later to do https://github.com/microsoft/pxt/blob/501f8f644fca2cb16130683af9c9252ee97f5377/webapp/src/components/ImageEditor/tilemap/TilePalette.tsx#L56

Try it out here
https://arcade.makecode.com/app/34b2c66fd592763c2b832448b548f90dfe57cea8-495463c3e6

gallery filter


webapp/src/components/ImageFieldEditor.tsx
```diff
@@ -42,16 +45,22 @@ export class ImageFieldEditor<U extends pxt.Asset> extends React.Component<Image
 
     constructor(props: ImageFieldEditorProps) {
         super(props);
+        this.toggleFilter = this.toggleFilter.bind(this);
```
livcheerful, line 48: I was confused why I needed to do the binding here but the other functions didn't? I feel like I'm missing somethin obvious >>

jwunderl: You need to do it for ones that are being passed as parameters to react components (e.g. in this case it's acting as a handler for an event, so you need to do a bind or define it as a field (e.g. instead of `toggleFilter() {` defining it as `toggleFilter = () => {`) Just a perf thing, if I recall correctly it's to make it compare as equal when react checks props / to prevent unnecessary renders

webapp/src/components/FilterPanel.tsx
```diff
@@ -0,0 +1,68 @@
...
+export class FilterTag extends React.Component<FilterTagProps> {
+    constructor(props: FilterTagProps) {
+        super(props)
+        this.clickHandler = this.clickHandler.bind(this);
+    }
+
+    render() {
+        return <div className="filter-tag">
+            <div className="filter-tag-box" role="checkbox" onClick={this.clickHandler} aria-checked={this.props.selected}>
+                <i className={this.props.selected ? `icon check square outline` : `icon square outline`}></i>
```
jwunderl, line 18:
```suggestion
                <i className={`icon square outline ${this.props.selected ? "check" : ""}`}></i>
```

nit: helps to minimize the difference so it's clear what the prop is changing

webapp/src/components/FilterPanel.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export class FilterTag extends React.Component<FilterTagProps> {
+    constructor(props: FilterTagProps) {
+        super(props)
+        this.clickHandler = this.clickHandler.bind(this);
+    }
+
+    render() {
+        return <div className="filter-tag">
+            <div className="filter-tag-box" role="checkbox" onClick={this.clickHandler} aria-checked={this.props.selected}>
```
jwunderl, line 17:
```suggestion
            <div className="filter-tag-box" role="checkbox" onClick={this.clickHandler} onKeyDown={sui.fireClickOnEnter} aria-checked={this.props.selected}>
```

don't apply this one as it requires sui import / it'll autoimport it if you type it out, the image editor isn't currently keyboard accessible but still good practice to handle key events to make things easier for future us :) Applies pretty much anywhere you add an `onClick`, it binds enter/space with focus to trigger `onClick`

webapp/src/components/FilterPanel.tsx
```diff
@@ -0,0 +1,68 @@
...
+interface FilterPanelProps {
+    enabledTags: string[];
+    tagClickHandler: (tag: string) => void;
+    clearTags: () => void;
+    tagOptions: string[];
+}
+
+export class FilterPanel extends React.Component<FilterPanelProps> {
+    protected isTagSelected(tag: string) {
+        return this.props.enabledTags.indexOf(tag.toLowerCase()) < 0 ? false : true;
```
jwunderl, line 55:
```suggestion
        return this.props.enabledTags.indexOf(tag.toLowerCase()) >= 0;
```
nit

webapp/src/components/ImageFieldEditor.tsx
livcheerful: ooooo gotcha thanks!

livcheerful: do we have a preference for binding vs field?

webapp/src/components/FilterPanel.tsx
```diff
@@ -0,0 +1,68 @@
...
+    buttonText?: string;
+    buttonAction?: () => void;
+}
+
+export class FilterPanelSubheading extends React.Component<FilterPanelSubheadingProps> {
+
+    render() {
+        return <div className="filter-subheading-row">
+            <div className="filter-subheading-title">{`${lf(this.props.subheading)}:`}</div>
+            {this.props.buttonText && <div className="filter-subheading-button" role="button" onClick={this.props.buttonAction}>{lf(this.props.buttonText)}</div>}
```
jwunderl, line 41: can't use lf like this, parameter to lf has to be a string literal. I'm surprised nothing in the build is yelling here.

(solution is either `lf` when passing as a prop, which is preferable, or `rlf` here and making sure any possible text is either extracted from definitions during build or in a hacky commented out `lf` like https://github.com/microsoft/pxt/blob/master/webapp/src/container.tsx#L25)

```suggestion
            {this.props.buttonText && <div className="filter-subheading-button" role="button" onClick={this.props.buttonAction}>{rlf(this.props.buttonText)}</div>}
```

webapp/src/components/ImageFieldEditor.tsx
```diff
@@ -197,7 +226,82 @@ export class ImageFieldEditor<U extends pxt.Asset> extends React.Component<Image
...
+                    });
+                }
+            })
+
+            return collectedTags;
+        }
+        return [];
+
+        function sanitize(tag: string) {
+            let sanitizedTag = (tag.indexOf("!") === 0 || tag.indexOf("?") === 0) && tag.length > 1 ? tag.substring(1) : tag;
```
jwunderl, line 252: probably don't need to trim `!` - `!` is only defined for gallery filters and not image tags so it wouldn't really need to be trimmed here / the tag shouldn't hit here? re: https://makecode.com/cli/buildsprites#tags (I guess no harm leaving though? just an extra that won't be hit in validly defined tags)

webapp/src/components/ImageFieldEditor.tsx
```diff
@@ -197,7 +226,82 @@ export class ImageFieldEditor<U extends pxt.Asset> extends React.Component<Image
...
+                if (!asset.meta.tags) {
+                    return false;
+                }
+                for (let i = 0; i < asset.meta.tags.length; i++) {
+                    const t = asset.meta.tags[i];
+                    if (this.state.gallerySelectedTags.indexOf(t) >= 0) {
+                        return true;
+                    }
+                }
+                return false;
```
jwunderl, line 298:
```suggestion
                return !!assets.meta.tags?.find(t => this.state.gallerySelectedTags.indexOf(t) >= 0);
```

nit; that should work? I wouldn't trust it without checking real quick as I did type it out in the github comment box and not vs code or anything :)

webapp/src/components/ImageFieldEditor.tsx
```diff
@@ -197,7 +226,82 @@ export class ImageFieldEditor<U extends pxt.Asset> extends React.Component<Image
...
+                if (asset.meta.tags) {
+                    asset.meta.tags.forEach(t => {
+                        let sanitizedTag = sanitize(t);
+                        if (specialTags.indexOf(sanitizedTag) == -1) {
+                            if (collectedTags.indexOf(sanitizedTag) < 0) {
+                                collectedTags.push(sanitizedTag);
+                            }
+                        }
+                    });
+                }
```
jwunderl, line 244:
```suggestion
                asset.meta.tags?.forEach(t => {
                    const sanitizedTag = sanitize(t);
                    if (specialTags.indexOf(sanitizedTag) == -1 && collectedTags.indexOf(sanitizedTag) < 0) {
                        collectedTags.push(sanitizedTag);
                    }
                });
```

nit, a lot of indentation here so maybe worth

jwunderl (changes requested):
mostly just nits that you can take / leave as you want, but the lf vs rlf comment does need fixing or it will mess with the translations~

webapp/src/components/ImageFieldEditor.tsx
jwunderl: my preference is towards arrow function defined as a property over `bind` / I think that's common to other people on the team, but it doesn't make a big difference at all

webapp/src/components/FilterPanel.tsx
livcheerful: ooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo cool good to know!

webapp/src/components/FilterPanel.tsx
livcheerful: ahhhhhhhhhhhhhhhhh okay so for tags, should I rlf, and add comments in the top the hacky way, and then we just gotta remember to add it here if we add more tags?

jwunderl: For now that would work, we should probably be pulling out the tags in the build and uploading them to crowdin if we're displaying them to the user but that's more work / not something that feels necessary for this change.

livcheerful: ooo yeah makes sense makes sense

webapp/src/components/ImageFieldEditor.tsx
```diff
@@ -197,7 +227,77 @@ export class ImageFieldEditor<U extends pxt.Asset> extends React.Component<Image
...
+        // lf("People")
+        // lf("Animals")
+        // lf("Food")
+        // lf("Dungeon")
+        // lf("Forest")
+        // lf("Space")
+        // lf("Aquatic")
+
+        if (this.galleryAssets) {
+            const specialTags = ["tile", "dialog", "background"];
```
shakao, line 242: maybe this should be passed in to "getAvailableTags" as a "tags to omit" param or something? i could picture a case where having these in the filter panel might be useful

abchatra:
@livcheerful is there a newer link with the updated CSS?

livcheerful:
@abchatra oh no I was going to do the CSS today

livcheerful:
Mobile views!

image

And tablet!
image

livcheerful:
and new upload: https://arcade.makecode.com/app/34b2c66fd592763c2b832448b548f90dfe57cea8-495463c3e6

There's a bug with viewing with iPad Pro -- I'm guessing our largest tablet size doesn't cover it. Not sure if that should be bumped up??

abchatra:
Looks great.

jwunderl (approved):
LGTM, one question; from the design check it sounded like Sonia was thinking of moving the editor/gallery/my assets toggle to the left in mobile / maybe small tablet view (e.g. like this:
image
), is that still something we're thinking about? On some devices (e.g. default moto g4 device under chrome dev tools) the filter button is hard to click as it overlays on top of the my assets button 
image
and I think it looks a bit better on phone sizes when it hangs to the left personally.

livcheerful:
ooo yeah let me do that

theme/image-editor/imageEditor.less
```diff
@@ -91,13 +95,67 @@
...
+
+
+// Mobile Only
+@media only screen and (max-width: @largestMobileScreen) {
+    .image-editor-gallery-content {
+        flex: 3!important;
+    }
+
+    .gallery-editor-toggle {
+        margin-left: .2rem!important;
```
jwunderl, line 129:
```suggestion
        margin-left: .25rem!important;
```

nitpickiest of possible nitpicks but the top-margin here is .25 so matching looks ever so slightly better to my eyes, feel free to take or leave :)

jwunderl (approved):

Jaqster:
Can we add a Filter by Size too?

livcheerful:
Oh hmm as a separate heading or as part of categories?

Jaqster:
Separate heading - you can filter by Category (what the images are) or Size of image.
Also, I'm wondering if this is going to be a universal Filter (across all Galleries) then should we also have a Filter by Type (image, animation, tile, etc.)?

livcheerful:
Oh yeah I'll look into adding the size filter too -- for type, I'm not sure if it makes sense since if it's visible in the gallery, it'll just import it as whatever type of editor you have opened

Jaqster:
Sounds good.  Looks amazing Vivian!!! So excited to have this in now!  :-)

abchatra:
Great job @livcheerful

PR #7898, opened 2021-02-17
name the default colors in the image editor for accessibility

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/3135

colors colors colors

mouse is only huge due to gif maker / it doesn't overlap normally

Took the names from https://arcade.makecode.com/reference/scene/background-color

(I think there's still value in showing the number as it corresponds to shortcut / etc so I just appended the name of the color to the end if present)

There are some node packages that will translate hexcode -> nearest named color, but that seems like a lot for now (the ones that pop up right away include like 28k colors), especially when changing the palette isn't used too much.

### DIFF
--- a/webapp/src/components/ImageEditor/sprite/Palette.tsx
+++ b/webapp/src/components/ImageEditor/sprite/Palette.tsx
@@ -53,14 +53,18 @@ class PaletteImpl extends React.Component<PaletteProps,{}> {
                 </g>
             </svg>
             <div className="image-editor-color-buttons" onContextMenu={this.preventContextMenu}>
-                {this.props.colors.map((color, index) =>
-                    <div key={index}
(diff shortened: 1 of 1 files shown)


livcheerful (approved):

webapp/src/components/ImageEditor/sprite/Palette.tsx
```diff
@@ -83,6 +87,43 @@ class PaletteImpl extends React.Component<PaletteProps,{}> {
     protected preventContextMenu = (ev: React.MouseEvent<any>) => ev.preventDefault();
 }
 
+function hexToNamedColor(color: string) {
```
riknoll, line 90: Add a comment explaining where these came from?

riknoll (approved):

PR #7904, opened 2021-02-22
add arcade skillmap to list of editors

pelikhan:
Streamer:
- [x] added arcade skillmap
- [x] fixed url.startWith function
- [x] compile to es6.


abchatra:
Thanks for building this super fast.

PR #7909, opened 2021-02-25
Code persistence for skillmap activities

riknoll:
Allow the option of carrying code from the previous code card to the next one in skillmaps. Code can only be carried over if the previous code card is completed. When the carryover happens, we merge the assets of the two projects so that whichever choice the user makes, their images, tilemaps, animations, and tiles will all still be present in the next step as well as the assets that are used in the tutorial.

If the user makes the wrong choice, the restart button is now always present for code cards that are in progress (used to only be there for completed cards). Restarting will give them the same choice.

Right now, we do not show the user's code in the modal when making the choice. I have some code for that, but I want to get this checked-in now because it has a ways to go. Eventually we will have a scrolling snippet like the ones you see in docs pages.

This PR also fixes an unreported tilemap bug when selecting tilemaps in the "my assets" tab where new tiles that aren't used in the tilemap don't show up in the tile gallery. It was also affecting carried over tilemaps, so I fixed it in both places.


skillmap/src/lib/codeCarryover.ts
```diff
@@ -0,0 +1,185 @@
+import { lookupPreviousActivityStates } from "./skillMapUtils";
+import { getProjectAsync, saveProjectAsync } from "./workspaceProvider";
+
+export async function carryoverProjectCode(user: UserState, pageSource: string, map: SkillMap, activityId: string, carryoverCode: boolean) {
+    const progress = user.mapProgress[pageSource][map.mapId];
+
+    const headerId = progress.activityState[activityId]?.headerId;
+
+    const previous = lookupPreviousActivityStates(user, pageSource, map, activityId);
+    const previousHeaderId = previous.find(state => state?.isCompleted &&
```
jwunderl, line 10: To confirm my reading of this; `lookupPreviousActivityStates` returns an array of elements that could potentially unlock the current activity (so parent nodes in the graph, not grandparents+), and this will grab the first one from that list that is completed.

This definitely falls into future work as the rendering is not supported yet, but out of curiosity -- in the future when shannon's stuff comes in and we can have skillmaps that have 2 parents to an activity, if both are completed then I guess the expectation would be that this would pull from first one defined in the markdown, and maybe allow an override as a property in the markdown (by specifying the id of the activity to pull from just in general I suppose) -- sound about right? Something to discuss in future I suppose.

jwunderl (approved):
LGTM, just one clarification / possible question for future plans~

skillmap/src/lib/codeCarryover.ts
shakao: in my ideal world we'd have UI to let the user choose, or we would take the more recently opened activity (along with displaying the activity name--"You are importing code from [Creating Enemies]" eg). but yeah definitely worth stepping through the flow as a team once we have all the pieces together

pxtlib/spriteutils.ts
```diff
@@ -647,6 +647,51 @@ namespace pxt.sprite {
         return result;
     }
 
+    export function addMissingTilemapTilesAndReferences(project: TilemapProject, asset: ProjectTilemap) {
```
shakao, line 650: i think there's one more instance of both of these functions in the asset editor's `editor.tsx`

skillmap/src/components/AppModal.tsx
```diff
@@ -129,6 +135,39 @@ export class AppModalImpl extends React.Component<AppModalProps> {
...
+
+        const actions = [
+            { label: lf("START FRESH"), onClick: async () => {
+                tickEvent("skillmap.startfresh");
+
+                dispatchSetReloadHeaderState("reloading")
+
+                await carryoverProjectCode(userState!, pageSourceUrl!, skillMap!, activity!.activityId, false)
+
+                dispatchSetReloadHeaderState("reload");
```
shakao, line 152: should it be "reload" in both start fresh/carryover cases? we don't need to reload if they select the default option, right?

skillmap/src/lib/codeCarryover.ts
riknoll: yeah this is absolutely just a placeholder for now. my plan was to do it based on whichever was most recent completed, but ui would be cool also.

skillmap/src/components/AppModal.tsx
riknoll: we do actually, because i merge the user's project into the currently loaded one.

shakao: hmm, feasible/worth it to kick off the "start fresh" merge as soon as the project is loaded? i don't know how long it's really taking--maybe it's fast?

livcheerful (approved):

skillmap/src/components/AppModal.tsx
riknoll: i think it's pretty fast currently. The first editor load is what takes the most time

shakao (approved):

PR #7911, opened 2021-02-26
Render branching layout for skill maps

shakao:
image

Trying to keep things cleanly split between the graph layout algorithm itself and the rendering. The React components only require a "depth" and "offset" per node (depth is distance along a primary axis--theoretically horizontal for desktop, vertical for tablet, offset is distance along the perpendicular axis) and then converts those to SVG units to draw. 

The graph layout is not bulletproof & works best for lower-degree graphs; we also don't handle edge crossings. We render the nodes in the order that the content creator has specified. If it turns out we do need a more robust algorithm to support content, it should be straightforward to swap out the layout function since it's separate from the rest of the implementation.

Basic layout:
- DFS traversal. Place node one unit horizontal (in the "depth" direction) from the parent if possible, otherwise place at the bottom of the current graph.
- If we encounter a child for the second time, increase the offset by 1 so that it is between both parents. (Previous parent will be above, given placement rules.)

Shrink long edges:
- If we have a very long leaf node with one parent, shift it back so that it is one unit below the nearest sibling (or parent, if it has no siblings). (This is for cases like the top-right graph, where otherwise the leaf nodes don't look great.)

Edges:
- Edges can have at most one bend. By default all edges extend horizontally from the child then move vertically to the parent. However, if there is a parent node on the horizontal axis or more than two parents, this will overlap, so we try to flip the edges to extend vertically from the child.

This is roughly based on: https://core.ac.uk/download/pdf/82771879.pdf (modified a bit to fit our use case)

If we need a more accurate algorithm, I thought this one also looked okay: https://www.sciencedirect.com/science/article/pii/S0925772197000266

### DIFF
--- a/skillmap/src/components/SkillGraph.tsx
+++ b/skillmap/src/components/SkillGraph.tsx
@@ -7,16 +7,17 @@ import { SkillGraphNode } from './SkillGraphNode';
 import { SkillGraphPath } from "./SkillGraphPath";
 
 import { isActivityUnlocked } from '../lib/skillMapUtils';
-import { GraphCoord, simpleGraph } from '../lib/skillGraphUtils';
+import { GraphCoord, GraphEdgeDirection, orthogonalGraph } from '../lib/skillGraphUtils';
 
 interface GraphItem {
     activity: MapActivity;
-    offset: GraphCoord;
+    position: GraphCoord;
 }
 
 interface GraphPath {
     start: GraphCoord;
     end: GraphCoord;
+    direction?: GraphEdgeDirection;
 }
 
 interface SkillGraphProps {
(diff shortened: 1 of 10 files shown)


skillmap/src/components/SkillGraph.tsx
```diff
@@ -0,0 +1,130 @@
...
+                if (current.edges) {
+                    current.edges.forEach(edge => {
+                        const points: SvgCoord[] = [];
+                        edge.forEach(n => points.push(this.getPosition(n.depth, n.offset)));
+                        paths.push({ points });
+                    });
+                }
+
+                this.size.height = Math.max(this.size.height, current.offset);
+                this.size.width = Math.max(this.size.width, current.depth);
```
jwunderl, line 67:
```suggestion
            items.push({
                activity: current,
                position: this.getPosition(depth, offset)
            });

            if (current.edges) {
                current.edges.forEach(edge => {
                    const points: SvgCoord[] = [];
                    edge.forEach(n => points.push(this.getPosition(n.depth, n.offset)));
                    paths.push({ points });
                });
            }

            this.size.height = Math.max(this.size.height, current.offset);
            this.size.width = Math.max(this.size.width, current.depth);
```

nit: indentation

skillmap/src/components/SkillGraphPath.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+interface SkillGraphPathProps {
+    points: SvgCoord[];
+    strokeWidth: number;
+    color: string;
+}
+
+export class SkillGraphPath extends React.Component<SkillGraphPathProps> {
+    render() {
+        const  { points, strokeWidth, color } = this.props;
```
jwunderl, line 17:
```suggestion
        const { points, strokeWidth, color } = this.props;
```

nit

jwunderl (approved):
LGTM, only a few minor spacing nits~

livcheerful (approved):

riknoll (approved):

abchatra:
Nicely done.

PR #7912, opened 2021-02-26
Disable auth in skillmap

eanders-MS:
Disable auth in skillmap until properly supported.

Also in this PR: force editor reloaded only if current project is in the list of synced changes. @darzu tell me if this looks ok.

Fixes: https://github.com/microsoft/pxt-arcade/issues/3138
Fixes: https://github.com/microsoft/pxt-arcade/issues/3150

cc @kiki-lee

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1335,7 +1335,7 @@ export class ProjectView
                                 // We are too late; the editor has already been loaded.
                                 // Call the onChanges handler to update the editor.
                                 pxt.tickEvent(`identity.syncOnProjectOpen.timedout`, { 'elapsedSec': elapsed})
-                                if (changes.length)
+                                if (changes.some(header => header.id === h.id))
                                     cloud.forceReloadForCloudSync()
                             } else {
                                 // We're not too late, update the local var so that the
@@ -4565,7 +4565,11 @@ document.addEventListener("DOMContentLoaded", () => {
         auth.loginCallback(query);
     }
 
-    auth.init();
+    // Disable auth in skillmap until it is properly supported.
+    // See https://github.com/microsoft/pxt-arcade/issues/3138
+    const disableAuth = query["skillsMap"] == "1";
+
+    auth.init(disableAuth);
     cloud.init(); // depends on auth.init() and workspace.ts's top level
     cloudsync.loginCheck()
     parseLocalToken();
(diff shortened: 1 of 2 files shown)


webapp/src/app.tsx
```diff
@@ -1335,7 +1335,7 @@ export class ProjectView
                                 // We are too late; the editor has already been loaded.
                                 // Call the onChanges handler to update the editor.
                                 pxt.tickEvent(`identity.syncOnProjectOpen.timedout`, { 'elapsedSec': elapsed})
-                                if (changes.length)
+                                if (changes.some(header => header.id === h.id))
```
eanders-MS, line 1338: @darzu does this make sense? I was getting an infinite reload and this fixed it. The current project wasn't in `changes`, yet it was still reloading the editor.

darzu: Ahh I see, good catch.
A few lines up, you can see cloud.syncAsync only passes a single header, h, the one that we're just loading. So I thought it'd be safe to just check length here.
But it's not because the way cloud.syncAsync works is that if a sync is already in progress it just returns a promise to that one which might include more headers than the one you asked for.

 (Really the whole syncAsync thing needs to be redone.. which I've done in my refactor branch but that's not coming in yet.)

webapp/src/auth.ts
```diff
@@ -366,7 +369,7 @@ export function hasIdentity(): boolean {
     // Must read storage for this rather than app theme because this method
     // gets called before experiments are synced to the theme.
     const experimentEnabled = pxt.editor.experiments.isEnabled("identity");
-    return !pxt.BrowserUtils.isPxtElectron() && experimentEnabled && identityProviders().length > 0;
+    return !authDisabled && !pxt.BrowserUtils.isPxtElectron() && experimentEnabled && identityProviders().length > 0;
```
darzu, line 372: Can you query the query string right here?
Might save some plumbing.

otherwise, this seems okay, but I'm a little surprised this solves the issues.
Cloud sync should only be touching projects that make it to the main webapp workspace.
And skillsmap projects live in their own DB so they shouldn't appear there unless the user explicitly said "save to my projects".
And even if auth is disabled while in the skillsmap, it should kick in again as soon as you're in the webapp normally which should sync any pending projects.

So I guess my question is: what behavior are you seeing that this fixes?

eanders-MS: What I think is happening is:
* skillmap loads the arcade editor in an iframe.
* iframe'd webapp has access to its normal cookies and db, so identity works (it can see the auth cookie and token).
* the editor, with a skillmap project loaded, auto saves, triggering a cloud save.
* cloud-saved projects sync to another client and are written to the regular workspace db.

eanders-MS: > Can you query the query string right here?

I'd considered that, but this seemed like a better separation of concerns. Let app logic happen at the app level. Auth shouldn't know about skill map.

darzu (approved):

PR #7913, opened 2021-02-26
Adding interactive consent to blocks.create

abchatra:
- Removing debouncing as appinsights already does debouncing.
- Adding interactive consent to activity.edit & blocks.create. (PLEASE Review this)


webapp/src/app.tsx
```diff
@@ -822,7 +822,7 @@ export class ProjectView
         let changeHandler = () => {
             if (this.editorFile) {
                 if (this.editorFile.inSyncWithEditor)
-                    pxt.tickActivity("edit", "edit." + this.editor.getId().replace(/Editor$/, ''))
+                    pxt.tickEvent("activity.edit", { editor: this.editor.getId().replace(/Editor$/, '')}, { interactiveConsent: true } );
```
jwunderl, line 825: This one gets hit in the arcade share page after pressing the `show code` (doesn't get hit in micro:bit immediately); that seems valid as it comes after the user has clicked the button to interact, but just want to double check

jwunderl (approved):
LGTM, one question on a tick firing in the share page but that's just double checking

webapp/src/app.tsx
abchatra: Good point. Seems ok to me.

abchatra:
@riknoll ?

riknoll:
blocks.create is not necessarily triggered by the user's interaction, i think there might be instances where it is triggered by things like loading a tutorial (but i could be wrong)

abchatra:
I have validated in arcade & minecraft in loading tutorials scenario (example: https://minecraft.makecode.com/#tutorial:/hour-of-code/glass) that blocks.create is only invoked when the block is dragged. Merging this.

PR #7914, opened 2021-02-26
[stable6.8] Cherry-pick fix checkdocs with autoexpandOff flag (#7884)

shakao:


jwunderl:
Is this one necessary on the stable branch? We should only need it if we merging new content into stable6.8, but I don't think we need that as it's usually just going into master

PR #7916, opened 2021-03-01
no more simulator flickering

pelikhan:
In certain sizes, our simulators are just at the limit of displaying the scrollbar. At this point, a bunch of resize flickering may happen.

The solution here is to always have a vertical scrollbar, but unhide only when it is needed. So, no more resizing and flickering.

scroll

ps: the scrollbar customizations for the image editor impact the whole of PXT.

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,6 +128,7 @@
     "@types/react-dom": "16.0.3",
     "@types/react-redux": "5.0.0",
     "@types/request": "2.48.2",
+    "@types/resize-observer-browser": "^0.1.5",
     "@types/web-bluetooth": "0.0.4"
   },
   "scripts": {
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -128,6 +128,7 @@
     "@types/react-dom": "16.0.3",
     "@types/react-redux": "5.0.0",
     "@types/request": "2.48.2",
+    "@types/resize-observer-browser": "^0.1.5",
```
jwunderl, line 131: Is this one ok to use? It seems pretty recent, I think we typically try to keep to supporting safari ~10+? https://caniuse.com/?search=resize%20observer Maybe worth the polyfill though, seems nice

pelikhan: I'm adding a guard to just ignore ResizeObserver if not supported.

jwunderl (approved):

PR #7920, opened 2021-03-02
Fill in skill map info panel with selected activity details

shakao:
skill-map-info

Filling in the info panel: displays activity info if an activity is selected, otherwise defaults to skill map (page-level) info. Right now we only support the title and description.

### DIFF
--- a/skillmap/src/components/ActionButtons.tsx
+++ b/skillmap/src/components/ActionButtons.tsx
@@ -62,7 +62,7 @@ export class ActionButtonsImpl extends React.Component<ActionButtonsProps> {
 
     render() {
         const { status } = this.props;
-        const completed = this.isCompleted(status || "notstarted");
+        const showRestart = (status && status !== "notstarted" && status !== "locked");
 
         if (status === "locked") return <div />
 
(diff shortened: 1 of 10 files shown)


skillmap/src/components/ActionButtons.tsx
```diff
@@ -0,0 +1,85 @@
...
+        const { status } = this.props;
+        const completed = this.isCompleted(status || "notstarted");
+
+        if (status === "locked") return <div />
+
+        return <div className="activity-actions">
+            <div className="action-button" role="button" onClick={this.handleActionButtonClick}>
+                {this.getSkillCardActionText()}
+            </div>
+            {completed && <div className="action-button" role="button" onClick={this.handleRestartButtonClick}>
```
jwunderl, line 73: Should this one be behind completed? I thought we were showing reset button when started but not completed as well, to allow for resetting when someone imports previous code and it doesn't work / they decide they need to go back to default project

jwunderl (approved):
LGTM besides the one comment on when to show restart button

skillmap/src/components/ActionButtons.tsx
shakao: oh yeah good catch, thanks!

riknoll (approved):

PR #7921, opened 2021-03-02
Fix setContentAsync to save to disk on new file creation; fix cloud sync race conditions

darzu:
Fixes: https://github.com/microsoft/pxt-arcade/issues/3194
Fixes: https://github.com/microsoft/pxt-arcade/issues/3206

- Changes Package.ts's setContentAsync to correctly save to disk when creating a new file
- Remove obsolete "inSyncWithDisk" flag in File interface
- Ensure cloud sync and saving queues properly so that we don't have conflicting cloud saves/syncs


webapp/src/cloud.ts
```diff
@@ -161,13 +161,15 @@ function resetAsync(): Promise<void> {
     return Promise.resolve();
 }
 
-let _inProgressSyncAsync: Promise<any> = Promise.resolve()
+let inProgressSyncPromise: Promise<any> = Promise.resolve()
 export async function syncAsync(hdrs?: Header[]): Promise<Header[]> {
```
eanders-MS, line 165: >export [](start = 0, length = 6)

Does this function need to be exported? Seems dangerous.

eanders-MS (approved):
:shipit:

webapp/src/cloud.ts
darzu: It should be safe to use liberally, it'll just hit the cloud a lot. And today it is called rather aggressively...

Ideally this should be called sparingly and probably use a similar debounce logic as our file saving. TODO :)

PR #7924, opened 2021-03-03
Adding reward/completion nodes

riknoll:
Mostly just a refactoring of types/parsing to support different node types in the skillmap graph. I added circular nodes but they currently don't do anything. My test url is

http://localhost:3000/?nolocalhost=1#github:https://github.com/riknoll/skillmap-branch/blob/main/test.md

Screen Shot 2021-03-02 at 5 11 35 PM

### DIFF
--- a/skillmap/src/lib/skillMapParser.ts
+++ b/skillmap/src/lib/skillMapParser.ts
@@ -108,7 +108,8 @@ function getSectionsFromText(text: string) {
 
 function buildMapFromSections(header: MarkdownSection, sections: MarkdownSection[]) {
     const result = inflateSkillMap(header);
-    const activities = sections.map(inflateActivity);
+    const nodes = sections.map(inflateMapNode);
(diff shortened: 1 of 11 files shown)


skillmap/src/lib/skillMapParser.ts
```diff
@@ -117,18 +118,17 @@ function buildMapFromSections(header: MarkdownSection, sections: MarkdownSection
         }
     });
 
-    result.root = activities[0];
-    result.root.allowCodeCarryover = false;
```
shakao, line 121: don't need this any more?

shakao:
not for this PR, but we will eventually need to update the getPreviousActivity function to skip these nodes for code carryover, correct?

shakao (approved):

skillmap/src/lib/skillMapParser.ts
riknoll: yeah i don't think it was ever needed

PR #7927, opened 2021-03-03
Fix race condition between multi-tab syncAsync and partialSaveAsync

darzu:
Fixes: https://github.com/microsoft/pxt-arcade/issues/3207

What I didn't realize is that our in memory cache in workspace.ts gets wiped rather aggressively by the multi-tab sync code even when there aren't any multi-tab changes.. we should fix that but for now if we just work with the assumption that u might need to rehydrate the cache from disk at any moment, things seem to work again.


eanders-MS (approved):
:shipit:

PR #7932, opened 2021-03-05
Fix false conflicts when doing multi-tab editting

darzu:
Fixes: https://github.com/microsoft/pxt-arcade/issues/3118
Fixes: https://github.com/microsoft/pxt-arcade/issues/3167

We use the "cloudCurrent" flag on Headers to determine if work needs to be pushed to the cloud. The problem is that each tab was assuming it should push changes for all headers that have cloudCurrent == false.
The fix was to not bother saving to cloud projects that aren't owned by this tab during the normal save loop.

Also, we were too aggressively reloading the whole editor in multi-tab scenarios. Instead we only need to reload the editor if the specific header we're editing is being edited by another tab.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -337,12 +337,13 @@ export class ProjectView
             this.saveFileAsync().done();
         } else if (active) {
             data.invalidate("header:*")
-            if (workspace.isHeadersSessionOutdated()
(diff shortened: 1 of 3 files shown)


eanders-MS (approved):
:shipit:

PR #7933, opened 2021-03-05
Fix rename not working w/ cloud save

darzu:
Fixes: https://github.com/microsoft/pxt-arcade/issues/3233

The problem was that the package layer was responsible for keeping the pxt.json changes in sync with the header fields, and they became desync'ed when using the single file save path.

Also there was an issue with diffing previous and new project state for the purpose of determining which changes are worth syncing to the cloud. We were updating the in-memory cache before doing the diff so we had no way to actually compare the header.


eanders-MS (approved):
:shipit:

PR #7935, opened 2021-03-05
fix project export on profile delete

eanders-MS:
Also: added related ticks.

### DIFF
--- a/webapp/src/auth.ts
+++ b/webapp/src/auth.ts
@@ -398,19 +398,28 @@ export async function deleteAccount() {
 
     const userId = getState()?.profile?.id;
 
-    await apiAsync('/api/user', null, 'DELETE');
+    const res = await apiAsync('/api/user', null, 'DELETE');
+    if (res.err) {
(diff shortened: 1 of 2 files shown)


webapp/src/auth.ts
```diff
@@ -398,19 +398,28 @@ export async function deleteAccount() {
...
-    pxt.storage.removeLocal(CSRF_TOKEN);
-
-    // Convert cloud-saved projects to local projects.
-    await cloud.convertCloudToLocal(userId);
-
-    // Update state and UI to reflect logged out state.
-    clearState();
-
-    // Reload page
-    window.location.reload();
```
darzu, line 413: no more reload necessary?

darzu (approved):
LGTM!

webapp/src/auth.ts
eanders-MS: I removed it to see what would happen. All seems to work. Do you have any concerns?

PR #7938, opened 2021-03-08
Render API request render to png

pelikhan:
Support for generating png in "renderblocks" javascript API.

It turns out that the franken-SVG we generated using "renderblocks" does not work with popular SVG rendering engine like Sharp. This SVG currently bundles the entire semantic.css so it's always at least 1.2Mb thick. This PR also generates PNGs from the iframe process so that all fonts resolved, etc...

To test: run http://localhost:3232/rendertest.html from the microbit target.

- [x] render URI field as png instead of encoded SVG (as it is so hugesque)
- [x] fix some missing guards on accessing options
- [x] refactoring logic to generate png with encoded data into browser utils

### DIFF
--- a/pxtblocks/blocklylayout.ts
+++ b/pxtblocks/blocklylayout.ts
@@ -363,7 +363,7 @@ namespace pxt.blocks.layout {
                 let pngUri = imageIconCache[svgUri];
 
                 return (pngUri ? Promise.resolve(pngUri)
-                    : pxt.BrowserUtils.encodeToPngAsync(svgUri, { width, height, pixelDensity: 4 }))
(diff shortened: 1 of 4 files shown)


pxtblocks/blocklylayout.ts
```diff
@@ -186,57 +186,19 @@ namespace pxt.blocks.layout {
...
+                        pixelDensity: (pixelDensity | 0) || 4,
+                        text: encodeBlocks ? JSON.stringify(blockSnippet, null, 2) : null
+                    });
             }).catch(e => {
                 pxt.reportException(e);
                 return undefined;
             })
     }
 
-    const MAX_SCREENSHOT_SIZE = 1e6; // max 1Mb
```
pelikhan, line 198: moved to browserutils

PR #7947, opened 2021-03-10
Check for "cloudCurrent" field when comparing local vs cloud

darzu:
Our sync algorithm wasn't pushing local changes in all cases.
Also improves our ticks during conflict resolution.

Fixes: https://github.com/microsoft/pxt-arcade/issues/3254

_might_ fix this too: https://github.com/microsoft/pxt-arcade/issues/3228
but since I don't have a consistent repro it's hard to say. Certainly the window for conflicts is larger without this PR since changes are synced less often.

### DIFF
--- a/webapp/src/cloud.ts
+++ b/webapp/src/cloud.ts
@@ -246,6 +246,7 @@ async function resolveConflict(local: Header, remoteFile: File) {
     const newName = getConflictCopyName(local);
     let newCopyHdr = await workspace.duplicateAsync(local, newName);
     pxt.tickEvent(`identity.sync.conflict.createdDuplicate`);
+    let anyError = false;
 
     // 2. swap current project to the new copy
     try {
@@ -259,6 +260,7 @@ async function resolveConflict(local: Header, remoteFile: File) {
         // we want to swallow this and keep going since step 3. is the essentail one to resolve the conflcit.
         pxt.reportException(e);
         pxt.tickEvent(`identity.sync.conflict.reloadEditorFailed`, { exception: e });
+        anyError = true;
     }
 
     // 3. overwrite local changes in the original project with cloud changes
@@ -269,11 +271,17 @@ async function resolveConflict(local: Header, remoteFile: File) {
         //  since this is a bad case (may lead to repeat duplication).
         pxt.reportException(e);
         pxt.tickEvent(`identity.sync.conflict.overwriteLocalFailed`, { exception: e });
+        anyError = true;
         throw e;
     }
 
     // 4. upload new project to the cloud (network op)
     const copyUploadHdr = await transferToCloud(newCopyHdr, null);
+    if (!copyUploadHdr) {
+        // nothing to do but report this
+        pxt.tickEvent(`identity.sync.conflict.uploadCopyToCloudFailed`);
+        anyError = true;
+    }
 
     // 5. tell the user a conflict occured
     try {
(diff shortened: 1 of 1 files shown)


webapp/src/cloud.ts
```diff
@@ -354,12 +370,8 @@ async function syncAsyncInternal(hdrs?: Header[]): Promise<Header[]> {
             if (remote) {
                 local.cloudLastSyncTime = remote.cloudLastSyncTime
                 // Resolve local and cloud differences.
-                const areDifferent = remote.cloudVersion
-                    ? local.cloudVersion !== remote.cloudVersion
-                    // TODO: once we deploy the backend to correctly return the cloudVersion from listAsync, then
```
darzu, line 359: Cloud always sends back the cloudVersion so this can be removed.

webapp/src/cloud.ts
```diff
@@ -269,11 +270,17 @@ async function resolveConflict(local: Header, remoteFile: File) {
         //  since this is a bad case (may lead to repeat duplication).
         pxt.reportException(e);
         pxt.tickEvent(`identity.sync.conflict.overwriteLocalFailed`, { exception: e });
+        anyError = true;
         throw e;
```
eanders-MS, line 274: Throwing here will skip processing of anyError below.

webapp/src/cloud.ts
```diff
@@ -382,28 +393,27 @@ async function syncAsyncInternal(hdrs?: Header[]): Promise<Header[]> {
                         // No local changes, download latest.
                         const newHdr = await fromCloud(local, remoteFile);
                         pxt.tickEvent(`identity.sync.noConflict.localProjectUpdatedFromCloud`)
+                    } else if (local.cloudVersion === remoteFile.version) {
+                        // Local has unpushed changes, push them now
```
eanders-MS, line 397: If local and remote versions are the same, wouldn't that mean there are no local changes to push?

webapp/src/cloud.ts
```diff
@@ -382,28 +393,27 @@ async function syncAsyncInternal(hdrs?: Header[]): Promise<Header[]> {
...
+                        pxt.debug(`local project '${local.name}' has changes that will be pushed to the cloud.`)
+                        const newHdr = await toCloud(local, remoteFile.version);
+                        U.assert(!!newHdr, `Failed to save ${local.id} to the cloud.`);
+                        pxt.tickEvent(`identity.sync.noConflict.localProjectUpdatedToCloud`)
                     } else {
-                        // Possible conflict.
-                        const conflictStr = `conflict found for ${projShorthand}. Last cloud change was ${agoStr(remoteFile.header.modificationTime)} and last local change was ${agoStr(local.modificationTime)}.`
+                        // Conflict. Local changes exist and we aren't on the latest cloud version.
                         if (local.modificationTime > remoteFile.header.modificationTime) {
```
eanders-MS, line 404: >modificationTime [](start = 34, length = 16)

Is this a client-generated timestamp? If so, it can't be relied upon.

webapp/src/cloud.ts
darzu: This is really confusing, but `local.cloudVersion` is the last e-tag we received from the cloud. `local.cloudCurrent` says that there have been changes to the project since it was last saved to the cloud.

eanders-MS: Oh I see. We get here if cloudCurrent is false.

---
In reply to: [592894008](https://github.com/microsoft/pxt/pull/7947#discussion_r592894008) [](ancestors = 592894008)

webapp/src/cloud.ts
darzu: it is client generated but in this case all it's informing is which tick to send since we always do the same copy solution.

darzu:
@eanders-MS , mind approving or offering critique?

eanders-MS (approved):
:shipit:

PR #7949, opened 2021-03-11
add cloud save icons; add cloud save button instead of text

darzu:
As we agreed on in the design meeting (quite a while ago now), here's the new cloud status indicator reworked as a button instead of icon + text.

Currently clicking the button does nothing except send a tick event. Since we're always auto-saving, it's not clear what the button could do..

Before:

After:


eanders-MS (approved):
:shipit:

darzu:
@eanders-MS , how do you feel about this change? Now that it's here, I'm not sold as it's less informative and very subtle. Also the "uploading" icon is a bit hard to parse.

eanders-MS:
Well, I was never onboard with this change, but willing to try it out. Agreed the saving icon could be improved. An upward pointing arrow typically means "has local changes". Saving to cloud should be an animated rotating arrow.

darzu:
I'll merge this for now since it was the roughly agreed upon goal, but I want to bring this up again at the next design office hours and hopefully we can at least improve the icon. It's relatively easy to iterate on this.

PR #7951, opened 2021-03-11
Sort headers consistently when cloud saved

darzu:
Fixes: https://github.com/microsoft/pxt-arcade/issues/3232

We used to always sort by "recent use". Instead, we now sort by either the last cloud save time or the modification time.

For non-cloud projects, this is a slight change since we're showing by mod time instead of view time (but I think we agreed on this.)

For cloud projects, the reason we chose the min(lastCloudSave, mod time) is because that is the safest time to show: it'll never show a timestamp beyond when it was last saved so that the user can trust that the timestamp means "it was saved no older than this time".


eanders-MS (approved):
This will work most of the time .. when modificationTime isn't a random value from another device. Only affects sorting so not really a concern.
:shipit:

PR #7956, opened 2021-03-15
Add banner image for info panel, tablet simulator CSS

shakao:
some small css, markdown changes for skillmap:
- move tablet simulator view to bottom of page (fixes https://github.com/microsoft/pxt-arcade/issues/3146)
- add a 'bannerurl' property for the info panel image for the top level page
- use the target's logo url if present


jwunderl (approved):

livcheerful (approved):

skillmap/src/components/InfoPanel.tsx
```diff
@@ -125,7 +125,7 @@ function mapStateToProps(state: SkillMapState, ownProps: any) {
         title: node?.displayName || state.title,
         subtitle,
         description: isActivity ? (node as MapActivity).description : state.description,
-        imageUrl: node?.imageUrl,
+        imageUrl: node ? node?.imageUrl : state.bannerImageUrl,
```
riknoll, line 128:
```suggestion
        imageUrl: node?.imageUrl || state.bannerImageUrl,
```

riknoll (approved):

skillmap/src/components/InfoPanel.tsx
shakao: ah wait sorry i committed then uncommited, i think we actually want it to be null if a node is present but doesn't have an imageUrl (eg some certificate nodes, maybe?), it'll just be the default "no image" icon in that case

PR #7960, opened 2021-03-16
Small CSS fixes for skillmap

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/3266
Fixes https://github.com/microsoft/pxt-arcade/issues/3267
Requires https://github.com/microsoft/pxt-arcade/pull/3282

Also fixes the skillmap background stretching in Safari (Safari doesn't like it if you put images in flexboxes).

Also includes a bunch of mobile fixes for the header. We still have lots of problems on mobile, but at least the header looks nice now:

Screen Shot 2021-03-15 at 6 12 30 PM

### DIFF
--- a/skillmap/src/styles/graphnode.css
+++ b/skillmap/src/styles/graphnode.css
@@ -19,3 +19,12 @@
     font-family: Icons;
     font-size: 1rem;
 }
+
+.graph-activity {
+    cursor: pointer;
+    user-select: none;
+}
+
+.graph-activity:hover .graph-icon, .graph-activity:hover .graph-icon-x {
+    opacity: 0.6;
+}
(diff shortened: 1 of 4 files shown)


jwunderl (approved):

livcheerful (approved):

skillmap/src/styles/graphnode.css
```diff
@@ -18,4 +18,13 @@
 .graph-status-icon {
     font-family: Icons;
     font-size: 1rem;
+}
+
+.graph-activity {
+    cursor: pointer;
```
jwunderl, line 24: Oh, just noticed, the status icon && graph icon need this too as they're also selectable

shakao (approved):

skillmap/src/styles/graphnode.css
riknoll: I put it on the outer <g>, so it should apply to all of them

PR #7964, opened 2021-03-16
Code carries over from most recently completed activity

shakao:
there's some merge weirdness going on, but here's the carryover pulling from the latest completed activity

fixes https://github.com/microsoft/pxt-arcade/issues/3280

### DIFF
--- a/skillmap/src/components/AppModal.tsx
+++ b/skillmap/src/components/AppModal.tsx
@@ -171,9 +171,7 @@ export class AppModalImpl extends React.Component<AppModalProps, AppModalState> 
     renderCodeCarryoverModal() {
         const  { skillMap, activity, pageSourceUrl, userState, dispatchHideModal, dispatchSetReloadHeaderState } = this.props;
         const carryoverModalTitle = lf("Keep code from previous activity?");
-        const carryoverModalText = lf("Do you want to start with your code from {0} \
-            or start fresh with starter code? Your images, tilemaps, \
-            tiles, and animations will stick around either way.");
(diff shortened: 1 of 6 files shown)


skillmap/src/components/AppModal.tsx
```diff
@@ -169,9 +169,17 @@ export class AppModalImpl extends React.Component<AppModalProps, AppModalState>
     }
 
     renderCodeCarryoverModal() {
-        const  { dispatchHideModal, skillMap, activity, pageSourceUrl, userState, dispatchSetReloadHeaderState } = this.props;
+        const  { skillMap, activity, pageSourceUrl, userState, dispatchHideModal, dispatchSetReloadHeaderState } = this.props;
         const carryoverModalTitle = lf("Keep code from previous activity?");
-        const carryoverModalText = lf("Do you want to start with your code from the previous activity or start fresh with starter code? Your images, tilemaps, tiles, and animations will stick around either way.");
+        const carryoverModalText = lf("Do you want to start with your code from {0} \
+            or start fresh with starter code? Your images, tilemaps, \
```
jwunderl, line 175: Huh, I hadn't realized escaping new lines in strings was supported in js; does the spacing come out okay? / do we extract it & upload properly to crowdin? I'd expect it to have an extra three tabs of spaces between each line

jwunderl (approved):
I think it might be preferable to leave carryoverModalText as one long line instead of splitting, but other than that lgtm

skillmap/src/components/AppModal.tsx
shakao: ah, i hadn't thought of crowdin, good call--will update!

PR #7968, opened 2021-03-16
Removing edit options from workspace in debug mode

riknoll:
Requires https://github.com/microsoft/pxt-blockly/pull/341
Fixes https://github.com/microsoft/pxt-arcade/issues/3220

Removes the remaining context menu edit options and also the "create get 'var'" option on for-loops which does not work and hasn't ever since we switched it to be draggable. I guess we just never noticed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
   },
   "devDependencies": {
     "monaco-editor": "0.19.3",
-    "pxt-blockly": "3.0.29",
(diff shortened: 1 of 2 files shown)


jwunderl (approved):

riknoll:
@jwunderl I'm bumping blockly as part of this change which includes your slider fixes. Is that okay to merge?

jwunderl:
Should be good yeah~

PR #7970, opened 2021-03-17
Add Immersive Reader

livcheerful:
The Immersive Reader gets a token from the cloud, then uses that token to make calls to the SDK

I did some initial cleaning of the tutorial text, but I think there's a lot of formatting that I missed. It should be easy to add regexes to pick off each case

Once I get the cloud going I'll drop an upload target!

I'll make a separate PR to make it work nicely on skillmap -- should only be adjusting the button


webapp/src/immersivereader.tsx
```diff
@@ -0,0 +1,126 @@
...
+    })
+
+    return contentWIP;
+
+    // Change ``|| around blocks to {}
+    function cleanBlockAnnotation(content: string): string {
+        const blockLiteral = /``\|\|([\w|\s]+:[\w|\s]+)\|\|``/
+        const blockRegex = new RegExp(blockLiteral, "g");
+        const blocksRemoved = content.replace(blockRegex, (blockDef) => {
+            const match = blockDef.match(blockLiteral);
```
jwunderl, line 37: Shouldn't need to rematch in here -- you should get the matched groups as parameters to this fn https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_function_as_a_parameter

webapp/src/immersivereader.tsx
```diff
@@ -0,0 +1,126 @@
...
+        const imagesRemoved = content.replace(imageRegEx, "");
+        return imagesRemoved;
+    }
+
+    // Remove ** around bold text, replace with "
+    function cleanBold(content: string): string {
+        const boldLiteral = /\*\*([\w|\s]+)\*\*/
+        const boldRegex = new RegExp(boldLiteral, "g");
+        const boldRemoved = content.replace(boldRegex, (boldWord) => {
+            const match = boldWord.match(boldLiteral);
```
jwunderl, line 55: same as above ^^

webapp/src/immersivereader.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+export class ImmersiveReaderButton extends data.Component<ImmersiveReaderProps, {}> {
+    private buttonClickHandler = () => {
+        launchImmersiveReader(this.props.content);
+    }
+
+    render() {
+        return <div className='immersive-reader-button ui item' onClick={this.buttonClickHandler}
+        aria-label={lf("Launch Immersive Reader")} role="button" onKeyDown={sui.fireClickOnEnter} tabIndex={0}>
+        </div>
```
jwunderl, line 124:
```suggestion
            aria-label={lf("Launch Immersive Reader")} role="button" onKeyDown={sui.fireClickOnEnter} tabIndex={0} />
```

nit spacing, also can make divs self closing in tsx if no children~

webapp/src/tutorial.tsx
```diff
@@ -99,6 +100,11 @@ export class TutorialMenu extends data.Component<ISettingsProps, {}> {
 
     renderCore() {
         let tutorialOptions = this.props.parent.state.tutorialOptions;
+        const tutorialCardContent = tutorialOptions.tutorialStepInfo ?
+            tutorialOptions.tutorialStepInfo[tutorialOptions.tutorialStep].headerContentMd :
+            null;
```
jwunderl, line 105:
```suggestion
        const tutorialCardContent = tutorialOptions.tutorialStepInfo?.[tutorialOptions.tutorialStep].headerContentMd
```

jwunderl (approved):
LGTM, pending backend side ~

package.json
```diff
@@ -128,7 +131,8 @@
     "@types/react-redux": "5.0.19",
     "@types/request": "2.48.5",
     "@types/resize-observer-browser": "^0.1.5",
-    "@types/web-bluetooth": "0.0.4"
+    "@types/web-bluetooth": "0.0.4",
+    "@microsoft/immersive-reader-sdk": "1.1.0"
```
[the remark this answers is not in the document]
livcheerful: ooo

pxtlib/emitter/cloud.ts
```diff
@@ -214,6 +214,8 @@ namespace pxt.Cloud {
     export function isOnline() {
         if (typeof navigator !== "undefined" && isNavigatorOnline()) {
             _isOnline = true;
+        } else {
```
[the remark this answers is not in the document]
livcheerful: oh hm maybe just a local test thing -- it wasn't being toggled to false when i was using chrome devtools to turn off network, but I can remove this

webapp/src/immersivereader.tsx
livcheerful: :O

pxtlib/emitter/cloud.ts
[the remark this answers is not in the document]
jwunderl: Right, it will only toggle to false after the first network call failed with 'disconnected' 0 response / doesn't trust navigator.isOnline response enough to aggressively fail things

Probably fine as nowadays that api is widely supported (though we could just simplify it a bit and `return navigator === "undefined" ? _isOnline : isNavigatorOnline()`) but if it was just from local testing maybe better to leave as is? Will still end up with the same "you're offline" error in either case, and if it actually is offline it won't take any time at all to error out~

theme/tutorial.less
```diff
@@ -471,6 +471,50 @@ code.lang-filterblocks {
     margin: 0.5em;
 }
 
+
+/******************************
+    Immersive Reader button
+******************************/
+#mainmenu .immersive-reader-button.ui.item,
+#mainmenu .immersive-reader-button.ui.item:focus,
+.hc #mainmenu .immersive-reader-button.ui.item:focus {
```
shakao, line 480: are the high contrast/focus selectors here necessary? will the first one cover those cases?

theme/tutorial.less
```diff
@@ -471,6 +471,50 @@ code.lang-filterblocks {
...
+    background-repeat: no-repeat;
+    background-size: contain;
+    background-color: transparent;
+    box-shadow: none;
+    transition: none;
+    float: left;
+    margin-top: 10px;
+}
+
+.hc .modals .ui.button.immersive-reader-button {
```
shakao, line 502: small: imo it's preferable to have all the .hc styles in highcontrast.less

webapp/src/immersivereader.tsx
```diff
@@ -0,0 +1,126 @@
...
+        return boldRemoved;
+    }
+}
+
+function getTokenAsync(): Promise<ImmersiveReaderToken> {
+    if (Cloud.isOnline()) {
+        const storedToken = pxt.storage.getLocal('immReader');
+        const cachedToken: ImmersiveReaderToken = storedToken ? JSON.parse(storedToken) : undefined;
+
+        if (!storedToken || (Date.now() / 1000 > cachedToken.expiration)) {
```
shakao, line 67: `!cachedToken` ?

webapp/src/immersivereader.tsx
```diff
@@ -0,0 +1,126 @@
...
+            return "\"" + match[1] + "\""
+        })
+        return boldRemoved;
+    }
+}
+
+function getTokenAsync(): Promise<ImmersiveReaderToken> {
+    if (Cloud.isOnline()) {
+        const storedToken = pxt.storage.getLocal('immReader');
+        const cachedToken: ImmersiveReaderToken = storedToken ? JSON.parse(storedToken) : undefined;
```
shakao, line 65: there's a pxt.Util.jsonTryParse that will catch errors and return undefined (https://github.com/microsoft/pxt/blob/master/pxtlib/util.ts#L276)

webapp/src/immersivereader.tsx
```diff
@@ -0,0 +1,126 @@
...
+            pxt.tickEvent("immersiveReader.cachedToken");
+            return Promise.resolve(cachedToken);
+        }
+    } else {
+        return Promise.reject(new Error("offline"));
+    }
+}
+
+export function launchImmersiveReader(content: string) {
+    pxt.tickEvent("immersiveReader.launch");
```
shakao, line 88: hmm i wonder if it would be useful to know what specific content the immersive reader is being launched for? maybe not the entire content but the tutorial id? not sure... and i guess this could also be done with an "immersiveReader.button.click" handler on the tutorial.tsx end

theme/tutorial.less
livcheerful: oh yeah in high contrast the background gets overridden by the high contrast rules -- I can try to add !important here instead if that's better

theme/tutorial.less
livcheerful: oh makes sense I'll move

webapp/src/immersivereader.tsx
livcheerful: oooo nice I will use instead!

webapp/src/immersivereader.tsx
livcheerful: oh yeah I'll add the id!

theme/tutorial.less
shakao: does the high contrast case below fix the background issue? or does it get overriden by the focus rules?

livcheerful: the case below does fix until it gets focused -- then it turns black

livcheerful:
Just realized I never dropped gifs >>

immersive reader

immersive reader hc

theme/tutorial.less
shakao: oh yeah maybe just add the :focus to the one below, then? just so all the overriding is happening in one place

pxtlib/emitter/cloud.ts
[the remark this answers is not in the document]
livcheerful: nicenice i will revert

webapp/src/immersivereader.tsx
livcheerful: (is tutorial id the tutorial name, or the tutorial? or something else >>)
image

shakao: oh yeah, i think most of our ticks use tutorialOptions.tutorial... we're not entirely consistent >>

shakao (approved):

jwunderl (approved):

webapp/src/immersivereader.tsx
livcheerful: excellentttt

webapp/src/immersivereader.tsx
```diff
@@ -0,0 +1,124 @@
...
+        }]
+    }
+
+    const options = {
+        onExit: () => {pxt.tickEvent("immersiveReader.close")}
+    }
+
+    getTokenAsync().then(res => ImmersiveReader.launchAsync(res.token, res.subdomain, data, options)).catch(e => {
+        switch (e.message) {
+            case "offline": {
```
abchatra, line 96: Is it worthwhile to have ticks for these error scenarios?

abchatra (approved):

webapp/src/immersivereader.tsx
livcheerful: oh totally. ticks are easier to watch i think. I'll add them!

livcheerful:
🚨🚨🚨 Uploadtrg here!!!!!!
https://arcade.staging.pxt.io/app/764b6aa1736bcd8c2d8708b50723a0fa460b7511-56cb6c0231

running with staging cloud, has all changes (including new error ticks that i'm gonna upload)

lemme know what you think!

abchatra:
Looks great. Ship it.

PR #7982, opened 2021-03-19
Add Immersive Reader for skillmap and recipes, add hover text

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/3317 tooltip says "Launch Immersive Reader"

uploadtrg! https://arcade.makecode.com/app/2bf8ddd106e873c066702f4ea6e3ee00aca4b2e5-72e4c2c947

skillmap:
skillmap

recipe:
recipe

ALSO fixes this bug, sorry guys >>
image

### DIFF
--- a/theme/highcontrast.less
+++ b/theme/highcontrast.less
@@ -826,6 +826,14 @@
         background-position: center!important;
     }
 
+    // skillmap embedded tutorials
+    .tutorial.tutorial-embed {
+        .immersive-reader-button.ui.item,
+        .immersive-reader-button.ui.item:focus {
+            background-size: 1.7rem;
+        }
+    }
+
     /* github */
     #githubEditor {
         background-color: @HCbackground !important;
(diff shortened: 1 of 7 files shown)


jwunderl (approved):
Yay my favorite test shark attack works :D 🎆 🔥

abchatra (approved):

pelikhan:
We should consider cleaning the markdown so that it makes more sense in the reader. Our "macro" encoding to generate inline blocks gets pasted verbatim and does not make much sense in "reader" mode.
image

pelikhan:
Nit: the active style is rather subtle, it's not obvious that the immersive reader is selected. This is a problem with the other buttons too IMO.

livcheerful:
> We should consider cleaning the markdown so that it makes more sense in the reader. Our "macro" encoding to generate inline blocks gets pasted verbatim and does not make much sense in "reader" mode.
> image

haha that's actually my cleaned up version -- the blocks before are the ||`` notation >> but yeah they definitely need more cleaning, I didn't hit all the cases

livcheerful:
I filed a bug for the button style here https://github.com/microsoft/pxt-arcade/issues/3318 I'll fix it!

PR #7984, opened 2021-03-19
cut sim volume a bit

jwunderl:
Make it less startling: can compare https://arcade.makecode.com/79623-16989-79723-53384 and https://arcade.makecode.com/app/ec5c0258e39de822efe8382124500bd128aff490-58aedfd1f9#pub:79623-16989-79723-53384

### DIFF
--- a/pxtsim/simlib.ts
+++ b/pxtsim/simlib.ts
@@ -506,7 +506,7 @@ namespace pxsim {
             channels.push(ch)
 
             /** Square waves are perceved as much louder than other sounds, so scale it down a bit to make it less jarring **/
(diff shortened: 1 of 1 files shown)


Jaqster:
The wawawa is especially obnoxious...  other sounds seem fine at this level, but I almost want even lower volume...

jwunderl:
I already started special casing the wawawa with the square noise, could lower it from .7 to .5?

Jaqster:
Yeah, let's try that

jwunderl:
Okay here it is with that: https://arcade.makecode.com/app/f3a549e77b25e3a011e4af1079c3ff3d1aa11d5e-9378dbdaee#pub:79623-16989-79723-53384

PR #7989, opened 2021-03-22
Separate live updates from unapproved translations for md

jwunderl:
Right now we're sending out `live=1` VERY aggressively in our translated markdown requests, and we should not be doing that. It looks like over time two distinct uses of live translations got combined -- updating the translations that are packaged with the editor, and fetching translations that are "live" / not yet approved

The change in `downloadLiveTranslationsAsync->downloadFromCloudAsync` isn't strictly necessary / I was mainly just checking if that needed to be fixed as well, but small enough it seemed like it should be included so those can get fetched appropriately as well?

### DIFF
--- a/pxtlib/util.ts
+++ b/pxtlib/util.ts
@@ -1044,10 +1044,9 @@ namespace ts.pxtc.Util {
         // hitting the cloud
         function downloadFromCloudAsync(strings?: pxt.Map<string>) {
             pxt.debug(`downloading translations for ${lang} ${filename} ${branch || ""}`);
-            const approvedOnly = !pxt.Util.unapprovedTranslationsEnabled();
             let host = pxt.BrowserUtils.isLocalHost() || pxt.webConfig.isStatic ? "https://makecode.com/api/" : ""
             // https://pxt.io/api/translations?filename=strings.json&lang=pl&approved=true&branch=v0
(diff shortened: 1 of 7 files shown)


pxtlib/util.ts
```diff
@@ -1187,22 +1205,21 @@ namespace ts.pxtc.Util {
...
                     setUserLanguage(code);
                     setLocalizedStrings(translations);
-                    if (live) {
-                        localizeLive = true;
-                    }
                 }
 
                 // Download api translations
-                return !live ? ts.pxtc.Util.downloadTranslationsAsync(
+                return liveUpdateStrings ? ts.pxtc.Util.downloadTranslationsAsync(
```
jwunderl, line 1220: This is reversing the behavior but it definitely *looks* like it was incorrect before to me / not triggering correctly -- before it was saying 'if we're not doing live translation updates, try and download translations for apis'.

livcheerful (approved):

riknoll (approved):
I'm not super familiar with this code path, but LGTM. @pelikhan FYI

pelikhan:
I would recommend to stop supporting unapproved translations all together. Translators can use the "live translation feature" to try out unapproved translations.

jwunderl:
Sounds good / agree with that, I dropped client side requesting of unapproved translations in the last 2 commits~

PR #7994, opened 2021-03-23
Add customts snippet type for loading hidden ts into tutorials

shakao:
proposing a new tutorial snippet type! the `customts` snippet allows you to write some typescript that's added to a hidden file in the tutorial project--essentially this is like adding an extension, but the code lives with the tutorial and works better with the existing `assetjson` snippet

build + example content, the hidden code creates a "camera" sprite that allows the user to pan the game around with the dpad--we want to hide this to make the game code easier to debug:
https://arcade.makecode.com/app/026265abbadd99d99736bf3e271d1a5b621c64d3-f5083f4725#tutorial:49840-40808-29707-30223

### DIFF
--- a/docs/writing-docs/tutorials/control-options.md
+++ b/docs/writing-docs/tutorials/control-options.md
@@ -65,22 +65,22 @@ This template example gives some initial code as a starting point for making a g
 ````
 ```template
 controller.A.onEvent(ControllerButtonEvent.Pressed, function () {
-	
+
 })
 scene.setBackgroundColor(9)
 let mySprite = sprites.create(img`
-    . . . . . . . . 
-    . . . . . . . . 
-    . . . . . . . . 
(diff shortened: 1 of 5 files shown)


riknoll (approved):

jwunderl (approved):
I like this idea a lot, only concern is that we won't handle name collisions nicely (e.g. if user tries to make Camera kind I think we'd just collide with them right now). Maybe we can stick the code in a namespace (or just have that be a suggestion in the docs for this)

shakao:
ooh good thought.... i think for now i'll add it to the documentation, since wrapping the SpriteKind namespace in another namespace causes an error and i want the content creators to be able to directly copy/paste our emitted code

riknoll:
Wrapping in a namespace would also break any blocks you are exporting

PR #7995, opened 2021-03-24
Only match recognized snippets when parsing tutorial metadata

shakao:
there was a bug where having an unrecognized snippet (eg `package`) would break all subsequent snippets since the regex matches the trailing \`\`\` . this restricts it to match only on snippets that we know; i don't think we were relying on replacing the untagged snippets anywhere?


jwunderl (approved):
LGTM

riknoll (approved):

PR #7996, opened 2021-03-24
make useLoc work for jsdoc, also take localizations with blockAliasFor

jwunderl:
Fix https://github.com/microsoft/pxt-arcade/issues/3110

Make it so blockAliasFor also takes localizations if the other block is translated and this one is not (iff the block has the same definition).

Also make it so blocklAliasFor and useLoc take the localizations for the parameters / jsdoc if available, and support the alternate translation in jsdoc format.

Minor change that this will use the parameters localization even if the regular jsdoc isn't translated; easy to change back but not sure why we would ignore those translations~

### DIFF
--- a/pxtlib/service.ts
+++ b/pxtlib/service.ts
@@ -638,9 +638,12 @@ namespace ts.pxtc {
                 const altLocSrc = fn.attributes.useLoc || fn.attributes.blockAliasFor;
                 const altLocSrcFn = altLocSrc && apis.byQName[altLocSrc];
 
-                const attrLocs = fn.attributes.locs || {};
-                const locJsDoc = loc[fn.qName] || attrLocs[attrJsLocsKey]
-                    || (altLocSrcFn && (loc[altLocSrcFn.qName] || altLocSrcFn.attributes.locs?.[attrJsLocsKey]));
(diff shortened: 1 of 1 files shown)


pxtlib/service.ts
```diff
@@ -634,27 +634,39 @@ namespace ts.pxtc {
...
                     fn.attributes._untranslatedJsDoc = fn.attributes.jsDoc;
                     fn.attributes.jsDoc = locJsDoc;
-                    if (fn.parameters)
-                        fn.parameters.forEach(pi => pi.description = loc[`${fn.qName}|param|${pi.name}`] || attrLocs[`${langLower}|param|${pi.name}`] || pi.description);
                 }
+
+                if (fn.parameters) {
+                    fn.parameters.forEach(pi => {
+                        const paramSuff = `|param|${pi.name}`;
+                        const paramLocs = loc[`${fn.qName}${paramSuff}`] || attrLocs[`${langLower}${paramSuff}`]
```
riknoll, line 652: nitpick: could pull this lookup into a local function

riknoll (approved):

livcheerful (approved):

PR #7997, opened 2021-03-24
More robust cloud sync error reporting

darzu:
We're getting some cloud sync errors and we don't have enough information to isolate where they are coming from yet, so we add a bunch of tick events.

Might help with: https://github.com/microsoft/pxt-arcade/issues/3285, https://github.com/microsoft/pxt-arcade/issues/3258, https://github.com/microsoft/pxt-arcade/issues/3251 and possibly others.


eanders-MS (approved):
:shipit:

PR #7998, opened 2021-03-24
Stop storing temporary assets in the tilemap project

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/3296

This PR changes how we store temporary assets a bit. We used to store them in the tilemap project, which was very error prone because it was difficult to keep the tilemap project and blocks file in sync. Now it just stores them in the blocks themselves, and fetches them from the workspace when opening the asset editor and "my assets" tab.

This fixes a lot of undo/redo bugs and probably quite a few other "disappearing asset" cases that we haven't reported. It also should fix the bug we saw on stream the other day where we lost all of our assets when skipping javascript decompile.

From my testing, this seems much more stable than it was before. Still we should make sure we test this before release.

### DIFF
--- a/pxtblocks/fields/field_asset.ts
+++ b/pxtblocks/fields/field_asset.ts
@@ -264,12 +264,7 @@ namespace pxtblockly {
                             this.asset.meta.blockIDs = this.asset.meta.blockIDs.filter(id => id !== this.sourceBlock_.id);
 
                             if (!this.isTemporaryAsset()) {
-                                if (this.asset.meta.blockIDs.length === 0 && !this.asset.meta.displayName) {
-                                    project.removeAsset(this.asset);
-                                }
-                                else {
-                                    project.updateAsset(this.asset);
-                                }
+                                project.updateAsset(this.asset);
                             }
                         }
                     }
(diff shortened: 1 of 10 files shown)


jwunderl (approved):

pxtblocks/fields/field_asset.ts
```diff
@@ -245,11 +263,13 @@ namespace pxtblockly {
                         if (this.sourceBlock_ && this.asset.meta.blockIDs) {
                             this.asset.meta.blockIDs = this.asset.meta.blockIDs.filter(id => id !== this.sourceBlock_.id);
 
-                            if (this.asset.meta.blockIDs.length === 0 && !this.asset.meta.displayName) {
-                                project.removeAsset(this.asset);
-                            }
-                            else {
-                                project.updateAsset(this.asset);
+                            if (!this.isTemporaryAsset()) {
+                                if (this.asset.meta.blockIDs.length === 0 && !this.asset.meta.displayName) {
```
shakao, line 267: not sure i understand this check--won't the display name always be present if it is not a temporary asset?

shakao:
one question, lgtm otherwise!

shakao (approved):

pxtblocks/fields/field_asset.ts
riknoll: whoops, i think this is just left over from refactoring

PR #8000, opened 2021-03-25
Fix blockCombine blocks in incontext translation mode

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/1793
fix https://github.com/microsoft/pxt-arcade/issues/2824
fix https://github.com/microsoft/pxt-arcade/issues/3120

Have to kick off the work async in a sync method here -- in practice it looks like this runs before the page is fully loaded, but I gave it a default english localization while the real one is loading just in case

### DIFF
--- a/pxtlib/service.ts
+++ b/pxtlib/service.ts
@@ -478,11 +478,17 @@ namespace ts.pxtc {
                 updateBlockDef(ex.attributes)
                 if (pxt.Util.isTranslationMode()) {
                     ex.attributes.translationId = ex.attributes.block;
-                    pxt.crowdin.inContextLoadAsync(ex.attributes.block)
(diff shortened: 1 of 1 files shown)


pelikhan:
On a slow enough connection it might create issues?

jwunderl:
Worst case the block will show up in english but 'translate this block' menu should still show the correct thing; similar to these built in ones that are also kicked off async in a synchronous portion of the code I believe https://github.com/microsoft/pxt/blob/d61dbc45b04867826176b1a14c9a66ea034958ed/pxtlib/blocks.ts#L764

Currently the toolbox just crashes due to these not getting replaced ~

pelikhan (approved):

PR #8002, opened 2021-03-25
More tutorial cleaning for Immersive Reader

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/3325
fixes https://github.com/microsoft/pxt-arcade/issues/3323

A couple of new rules for tutorials -- it looks really pretty now! mostly 😅 if there are other cases you've seen that are egregious lemme know

uploadtrg:
https://arcade.makecode.com/app/79562dd4db313b61b72dd4f3d901983d057ba09b-77eee626c1


webapp/src/immersivereader.tsx
```diff
@@ -15,17 +15,25 @@ export type ImmersiveReaderToken = {
...
+        cleanItalics,
         cleanNewLines,
+        cleanHorizontalRule,
+        cleanBlockquotes,
+        cleanEmojis,
+        convertBoolAngles
     ];
     let contentWIP = content;
-
+    console.log("cleaning: \n" + content );
```
jwunderl: remove logs?

webapp/src/immersivereader.tsx
```diff
@@ -46,29 +62,98 @@ function beautifyText(content: string): string {
...
+            }
         );
     }
 
     // replace horizontal rules with new lines
     function cleanHorizontalRule(content: string): string {
         return content.replace(
-            /^\* \* \*$/gum,
-            "- - -"
+            /^\* \* \*$|^- - -$|^<hr\/>$/gum,
```
jwunderl, line 93: I might be misreading this one but did this not get included in the build? looks like it still has the behavior I had it as with swapping horizontal rule for `- - -`? 
image from galga step 2

webapp/src/immersivereader.tsx
```diff
@@ -46,29 +62,98 @@ function beautifyText(content: string): string {
...
+            /^```block$[^`]*^```$/gum,
+            ""
+        );
+    }
+
+    // Remove ** and __ around bold text, replace with " if it's only one word
+    function cleanBold(content: string): string {
+
+        let singleWordBold = content.replace(
+            /\*\*([^*^\s]+)\*\*|__([^\s^_]+)__/gu,
```
jwunderl, line 76: nit: feel free to ignore -- if you want to avoid the ternary / matching two groups you could use this regex instead here:

``/(?:\*\*|__)([^*^\s]+)(?:\*\*|__)/``

?: at the start of a group denotes it as a 'non capture group' / gets skipped for the groupings. Applies to both this fns regexes and the clean italics I think?

webapp/src/immersivereader.tsx
shakao, line 33: were these logs intentionally left in?

livcheerful: D: no, thanks shannon!

webapp/src/immersivereader.tsx
```diff
@@ -46,29 +62,98 @@ function beautifyText(content: string): string {
...
 
-    // Remove ** around bold text, replace with "
-    function cleanBold(content: string): string {
+    // If the ```block ``` annotation is used to display a block in
+    // a modal, remove it
+    function cleanHintText(content: string): string {
         return content.replace(
-            /\*\*([^*]+)\*\*/gu,
-            (matched, word, offset, s) => lf("\"{0}\"", word)
+            /^```block$[^`]*^```$/gum,
```
jwunderl, line 67: There's a laundry list of these over in https://github.com/microsoft/pxt/blob/ec20d5991e3c23707a6bd5c85cd3a2a913d49a29/pxtlib/tutorial.ts#L53 -- might be able to unify that a bit? but probably good for now!

jwunderl (approved):
LGTM, though I'd drop the logs at least :)

webapp/src/immersivereader.tsx
livcheerful: oh man that's great

webapp/src/immersivereader.tsx
livcheerful: OH thanks for finding that i will investigate 🔍

webapp/src/immersivereader.tsx
```diff
@@ -46,29 +60,98 @@ function beautifyText(content: string): string {
...
+    function cleanBlockquotes(content: string): string {
+        return content.replace(
+            /^>>/gu,
+            ""
+        )
+    }
+
+    // Remove all emojis because they get read out. Characters that aren't
+    // emojis don't get hit by this RegEx, but they're silent
+    function cleanEmojis(content: string): string {
```
shakao, line 116: question: are we leaving the unicode characters (Ⓐ, Ⓑ, etc) for now?

webapp/src/immersivereader.tsx
```diff
@@ -46,29 +60,98 @@ function beautifyText(content: string): string {
...
+
+        let singleWordBold = content.replace(
+            /\*\*([^*^\s]+)\*\*|__([^\s^_]+)__/gu,
+            (matched, boldStar, boldUnder, offset, s) => {
+                return "\"" + lf("{0}", (boldStar ? boldStar : boldUnder)) + "\""
+            }
+        )
+
+        return singleWordBold.replace(
+            /\*\*([^*]+)\*\*|__([^_]+)__/gu,
```
shakao, line 83: not blocking for this PR, but this doesn't catch something like https://github.com/microsoft/pxt-arcade/blame/master/docs/skillmap/platformer/activity2.md#L209 which has italicized text inside a bold. we could also just modify the tutorial to work, though!

webapp/src/immersivereader.tsx
livcheerful: yesss because I don't think they get read out -- at least for the ones I've seen. Do you think it'd be better if it was replaced by the letters instead?

webapp/src/immersivereader.tsx
livcheerful: Ah yeah :\ I couldn't think of a good way to catch this case, and hadn't seen many places where it was used >> I can put up a PR to modify the tutorial

webapp/src/immersivereader.tsx
```diff
@@ -46,29 +60,98 @@ function beautifyText(content: string): string {
...
+        return content.replace(
+            /\*([^*]+)\*|_([^_]+)_/gu,
+            (matched, starItalics, underlineItalics, offset, s) =>
+                lf("{0}", starItalics ? starItalics : underlineItalics)
+        );
+    }
+
+    function convertBoolAngles(content: string): string {
+        return content.replace(
+            /<(true|false)>/gu,
```
shakao, line 149: small: do we also want this for the other toggles (win/lose, on/off etc)? there's a pretty good list of them here https://github.com/microsoft/pxt/blob/master/libs/pxt-common/pxt-helpers.ts#L588, but i tried it on some platformer tutorials with the win/lose and it seemed fine with the angle brackets also

webapp/src/immersivereader.tsx
livcheerful: oooh yeah I didn't do this because then it could grab a **(wordswordswords)__ case, I think?

livcheerful: but also I didn't konw about ?: lol

webapp/src/immersivereader.tsx
shakao: yeah i think eventually the button ones should be replaced with "A" or "B" (since which button to select is important info for the game) but the other characters we probably don't need! could also add this later though, up to you

webapp/src/immersivereader.tsx
livcheerful: !!! yes we definitely want for other toggles tytyty Immersive Reader ignores everything inside a <> so it sounds silly when it takes it away :(

webapp/src/immersivereader.tsx
shakao: not urgent, i think!

shakao (approved):

webapp/src/immersivereader.tsx
jwunderl: Ohhh good point, then maybe use a backref `/(\*\*|__)([^*^\s]+)\1/gu`? have to skip the first in the replace function but don't have to do ternaries in each replace later on:

https://regex101.com/r/iwUPLA/1

livcheerful: :O that's amazingggg

webapp/src/immersivereader.tsx
```diff
@@ -60,11 +61,11 @@ function beautifyText(content: string): string {
         );
     }
 
-    // If the ```block ``` annotation is used to display a block in
-    // a modal, remove it
-    function cleanHintText(content: string): string {
+    // Remove any triple tick annotations ```
+    // This RegEx was copied from pxtlib/tutorial.ts
```
jwunderl, line 65: Should probably drop a note there or we'll never remember to update this -- might be easy to put a function in `pxtlib/tutorial.ts` that returns this regex so we don't have to dup it though

abchatra:
This PR reverted change by following PR
_Only match recognized snippets when parsing tutorial metadata_ (#7995)

@livcheerful & @shakao is that intentional?

livcheerful:
! not intentional

abchatra:
Can you fix in master? I have fixed in stable .

livcheerful:
yupyup. thanks!!

PR #8011, opened 2021-03-26
Make it more obvious when Immersive Reader is focused/hovered on

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/3318

for non-HC mode, it was really hard to see when the button was focused/hovered on. it's better now!

image
image
image
image
image
image

### DIFF
--- a/theme/highcontrast.less
+++ b/theme/highcontrast.less
@@ -818,6 +818,16 @@
 
     /* immersive reader */
     .modals .ui.button.immersive-reader-button,
+    #mainmenu .immersive-reader-button.ui.item {
+        border: 1px solid #fff!important;
+        background-image: @immersiveReaderLightIcon !important;
+        background-repeat: no-repeat !important;
+        background-size: 80% !important;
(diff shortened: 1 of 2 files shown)


jwunderl (approved):

theme/highcontrast.less
```diff
@@ -818,19 +818,24 @@
...
     }
 
     // skillmap embedded tutorials
     .tutorial.tutorial-embed {
         .immersive-reader-button.ui.item,
-        .immersive-reader-button.ui.item:focus {
-            background-size: 1.7rem;
+        .immersive-reader-button.ui.item:focus,
+        .immersive-reader-button.ui.item:hover {
+            background-size: 80%!important;
```
jwunderl, line 836: For high contrast to match the other buttons it might be better to do white outline / black background / white version of the icon I suppose (but still LGTM)

image

livcheerful:
non hover HC looks like this now!
image
image

hover still looks the same

PR #8014, opened 2021-03-26
More robust cloud sync error reporting (#7997)

darzu:
* Adds more identity tick events
* more robust cloud sync error handling
* wrap in Error()

### DIFF
--- a/webapp/src/cloud.ts
+++ b/webapp/src/cloud.ts
@@ -80,8 +80,10 @@ async function listAsync(): Promise<Header[]> {
                 header.cloudVersion = proj.version;
                 return header;
             });
+            pxt.tickEvent(`identity.cloudApi.list.success`, { count: headers.length });
             resolve(headers);
         } else {
+            pxt.tickEvent(`identity.cloudApi.list.failed`);
             reject(result.err);
         }
     });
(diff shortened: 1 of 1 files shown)


darzu:
Force merging so hopefully it makes it in the release testing. @abchatra

abchatra:
Give me some time. I am going to take a look

abchatra:
Typically, ticks are for user action and it is ok to use them for debug purpose if the ticks are limited.

These are debug ticks. What is the number of these ticks per session?

darzu:
For normal sessions, we currently expect about ~2 syncs and ~15 saves, each of which will generate on average 1-2 of these new ticks. So about 15-30ish ticks per session.

Do we have a system for debug ticks? They are a useful pattern for debugging occasional and hard to repro issues.

abchatra:
30ticks is ok per session. If it is like 500 I would be concerned. We use log to console (using debug flag) and ask folks to send the logs

abchatra:
I am going to merge this. But keep an eye on the number of these ticks per session. If it is too much lets dial down.

abchatra (approved):

darzu:
Log to console with debug flag and manual sending error reports is great for some scenarios but really not good if you're trying to figure out how many people are hitting certain error cases or investigating a rare issue after it has already happened.

We have a system for reporting and tracking exception telemetry already, right? Does this use the normal app insights channel or something else? Can we extend that for debug ticks?

abchatra:
You can log failure scenario. But you are also logging success scenario. 

We already send all exceptions to application insights.

darzu:
Okay, I'll remove some/all of the success scenario logging as we expand identity to many new users. The error cases are the most important. The exceptions are only occasionally useful since they don't capture all the context needed.

PR #8017, opened 2021-03-29
Fix String.fromCharCode py error

darzu:
Fixes: https://github.com/microsoft/pxt-microbit/issues/3152

When emitting Name expressions, we shouldn't emit our internal `@type` specifiers.

### DIFF
--- a/pxtpy/converter.ts
+++ b/pxtpy/converter.ts
@@ -2387,6 +2387,8 @@ namespace pxt.py {
 
             let scopeV = lookupName(n)
             let v = scopeV?.symbol
+
+            // handle import
             if (v && v.isImport) {
                 return quote(v.name) // it's import X = Y.Z.X, use X not Y.Z.X
             }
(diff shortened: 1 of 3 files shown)


riknoll (approved):

pxtpy/converter.ts
```diff
@@ -2387,18 +2387,30 @@ namespace pxt.py {
...
 
             if (n.ctx.indexOf("Load") >= 0) {
-                if (v && !v.qName)
+                if (!v)
+                    return quote(getName(n))
+                if (!v?.qName) {
                     error(n, 9561, lf("missing qName"));
-                return quote(v ? v.qName! : getName(n))
-            } else
+                    return quote("unknown")
```
riknoll, line 2403: is this unreachable? or is there a chance of getting "unknown" in your code?

shakao (approved):

pxtpy/converter.ts
darzu: Should be unreachable. The previous behavior (if reachable) was emitting "", "undefined", or "null".

Ideally I'd convince the type system it's unreachable too, but it didn't appear trivial to do so.

PR #8026, opened 2021-03-30
Don't let the header overlap with each other

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/3322

This is the best way I could think of to fix this :\ -> After we render the menu bar, check all the UI elements under the parent and add up the widths. If it's bigger than the screen width, add a class to the menu bar, "shrink", which will hide all elements with the landscape class (like tutorial name, text on the exit button)

One weird thing that's happening that I don't know if it's a result of me not understanding React or this change, is when the MainMenu element rerenders, it keeps the "shrink" class, which makes going from not shrunk -> shrunk a one way street until you refresh the page.

This is a window at (1200x649), which is the dimension of the Pixelbook I have:
image

And with these changes:
image

And this shows the one-way street ness of it :\

width changes

### DIFF
--- a/theme/tutorial.less
+++ b/theme/tutorial.less
@@ -696,7 +696,7 @@ code.lang-filterblocks {
 
 /* > Small Monitor */
 @media only screen and (min-width: @computerBreakpoint) {
-    .shrink {
+    .thin {
         .landscape.only {
             display: none;
         }
--- a/theme/tutorial.less
+++ b/theme/tutorial.less
@@ -696,6 +696,11 @@ code.lang-filterblocks {
 
 /* > Small Monitor */
 @media only screen and (min-width: @computerBreakpoint) {
+    .shrink {
+        .landscape.only {
+            display: none;
+        }
+    }
 }
 
 /* Large Monitor */
--- a/theme/tutorial.less
+++ b/theme/tutorial.less
@@ -700,6 +700,15 @@ code.lang-filterblocks {
         .landscape.only {
             display: none;
         }
+
+        .portrait.hide {
+            display: none;
+        }
+
+        .portrait.only {
+            display: block!important;
+            height: auto!important;
+        }
     }
 }
 
--- a/webapp/src/container.tsx
+++ b/webapp/src/container.tsx
@@ -528,6 +528,44 @@ export class MainMenu extends data.Component<ISettingsProps, {}> {
         this.props.parent.toggleDebugging();
     }
 
+    componentDidUpdate() {
+        const tutorialOptions = this.props.parent.state.tutorialOptions;
+        const inTutorial = !!tutorialOptions && !!tutorialOptions.tutorial;
+        const header = document.getElementById("mainmenu")
+        if (inTutorial && header && !header.classList.contains("shrink")) {
+            const shrinkHeader = this.shouldShrinkItems();
+            if (shrinkHeader) {
+                header.classList.add("shrink");
+            }
+        }
+
+    }
+
+    shouldShrinkItems(): boolean {
+        const header = document.getElementById("mainmenu")
+        if (header) {
+            let headerWidth = 0;
+            for (let i = 0; i < header.children.length; i++) {
+                headerWidth += addUiSize(header.children[i]);
+            }
+            return headerWidth > window.innerWidth;
+        }
+        return false;
+
+        function addUiSize(elem: Element): number {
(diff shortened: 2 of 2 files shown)


theme/tutorial.less
```diff
@@ -696,6 +696,11 @@ code.lang-filterblocks {
 
 /* > Small Monitor */
 @media only screen and (min-width: @computerBreakpoint) {
+    .shrink {
+        .landscape.only {
```
jwunderl, line 700: I think I'd personally prefer the button to always go small in those dimension rather than only when strictly necessary -- it's very cramped even when everything shows up on screens like that and shrinking the button actually leaves a nice amount of room for everything to settle in. Maybe a quick topic for a design office hour?

webapp/src/container.tsx
```diff
@@ -528,6 +528,44 @@ export class MainMenu extends data.Component<ISettingsProps, {}> {
...
+    componentDidUpdate() {
+        const tutorialOptions = this.props.parent.state.tutorialOptions;
+        const inTutorial = !!tutorialOptions && !!tutorialOptions.tutorial;
+        const header = document.getElementById("mainmenu")
+        if (inTutorial && header && !header.classList.contains("shrink")) {
+            const shrinkHeader = this.shouldShrinkItems();
+            if (shrinkHeader) {
+                header.classList.add("shrink");
+            }
+        }
```
jwunderl, line 540:
```suggestion
        const headerContainsShrink = header.classList.contains("shrink");
        if (inTutorial && header && !headerContainsShrink && this.shouldShrinkItems()) {
            header.classList.add("shrink");
        } else if (headerContainsShrink) {
            header.classList.remove("shrink");
        }
```

it looks like the reason for the weird behavior is because it's not cleared in the update here -- I think this would fix it?

theme/tutorial.less
livcheerful: oh yeah sure! perfect timing

webapp/src/container.tsx
livcheerful: hahaha I thought that when the render() returned the div, it would only have classes given by className :\ okay I'll come up with a way to fix it -- i think doing it this way makes it flicker but i will try it again

shakao:
one possible alternative is you could add a CSS class if there are many steps in a tutorial (eg > 10 steps), then in the CSS you could set a media breakpoint to shrink the button if the "many tutorial steps" CSS class is present--if that makes sense?

livcheerful:
> one possible alternative is you could add a CSS class if there are many steps in a tutorial (eg > 10 steps), then in the CSS you could set a media breakpoint to shrink the button if the "many tutorial steps" CSS class is present--if that makes sense?

oh yeah makes sense! I'll bring to office hours today and we can decide which way we prefer?

livcheerful:
@shakao okay I think I *actually* understood what you meant this time -- instead of calculating the rendered width of all the elements, I made an estimate by the number of tutorial steps, and added a class in the mainmenu div.

(I wasn't 100% sure about the media breakpoint though and took a stab 😓 let me know if that was off base lol)

shakao:
awesome!! so just to double-check, the expected behavior is:
- thin width + few tutorial steps: no collapse
- thin width + many tutorial steps: collapse

is that right? is there a plan to "always collapse" for certain widths?

livcheerful:
yup that sounds right! It'll always collapse when it hits the tablet width (~990 I think)

(i forgot I made this uploadtrg! https://arcade.makecode.com/app/7c2dc53f82bfabdbde839f8d3070e0b89ca3cef0-e76ed3e8b4 )

shakao (approved):

jwunderl (approved):

webapp/src/container.tsx
```diff
@@ -540,6 +540,9 @@ export class MainMenu extends data.Component<ISettingsProps, {}> {
         const sandbox = pxt.shell.isSandboxMode();
         const inTutorial = !!tutorialOptions && !!tutorialOptions.tutorial;
 
+        // Approximate each tutorial step to be 22 px
+        const manyTutorialSteps = inTutorial && (tutorialOptions.tutorialStepInfo.length * 22 > window.innerWidth / 3);
```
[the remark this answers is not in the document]
livcheerful: oh yeah :\ I think it will come back if you switch to the next step too so hopefully that won't be too confusing...

PR #8029, opened 2021-03-30
Add bug button to skillmap header to toggle Usabilla

shakao:
skillmap-usabilla


livcheerful (approved):

abchatra:
I am not sure I understand the change. Don't we have a feedback button on the side?  We are adding another bug button at the top too? 

image

PR #8032, opened 2021-03-31
Add temporary assets to JRes when doing skillmap carryover

shakao:
now that temporary blocks assets are stored in the workspace, we need to separately pull them out in order to carry them over in the skillmap. the updated flow:

if we are not carrying over the code (if we are carrying over the code the temporary assets are included, we keep the present behavior):
- regex to strip temporary image and animation literals from the blocks xml
- convert to jres and append to the project's jres
- continue with merge

plus an additional fix to add the tilemap/image jres file to the pxt.json if it is not present

### DIFF
--- a/skillmap/src/lib/codeCarryover.ts
+++ b/skillmap/src/lib/codeCarryover.ts
@@ -1,3 +1,4 @@
+import { guidGen } from "./browserUtils";
 import { lookupPreviousCompletedActivityState } from "./skillMapUtils";
 import { getProjectAsync, saveProjectAsync } from "./workspaceProvider";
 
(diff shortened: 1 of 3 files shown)


skillmap/src/lib/codeCarryover.ts
```diff
@@ -181,4 +193,94 @@ function mergeJRES(previous: string, next: string) {
...
+    return JSON.stringify(jres)
+}
+
+/**
+ *  <field name="img">img`
+ *      . .
+ *      . .
+ *  `</field>
+ */
+function getImages(text: string): pxt.sprite.BitmapData[] {
```
riknoll, line 251: hmmm.. I wonder if we should hardcode the field name as that is part of the block definition and might not always be "img". For example, if someone made an extension that uses images

skillmap/src/lib/codeCarryover.ts
```diff
@@ -181,4 +193,94 @@ function mergeJRES(previous: string, next: string) {
...
+ *  <field name="frames">[img`
+ *      . .
+ *      . .
+ *  `,img`
+ *      . .
+ *      . .
+ *  `]</field>
+ */
+function getAnimations(text: string): { frames: pxt.sprite.BitmapData[], interval: number}[] {
+    const animRegex = /<field name=\"frames\">\s*\[((?:img`[\s\da-f.#tngrpoyw]+`,?)+)\]\s*<\/field>.*<shadow type=\"timePicker\"><field name=\"ms\">(\d+)<\/field><\/shadow>/gim;
```
riknoll, line 273: same comment here

jwunderl (approved):

livcheerful (approved):

PR #8034, opened 2021-03-31
Rewrite checkdocs - snippet checks to use incremental compile

jwunderl:
And a handful of other changes along the way (test cycle was long so I made a few incremental fixes between changes), I'll add a few annotations inline.

~One other thing I was thinking about was changing the order in ci so that checkdocs runs last (i.e. after uploading target, release, etc) -- that should be a very easy change and makes it so checkdocs still runs but is less irritating (e.g. if we want to do a hotfix or last build before a testing session). Will add that to this pr if no one is opposed to that~ Sounds like opposition to this so won't do for now, we can reconsider later on if we want.

on my machine, timing speed:
arcade: ~30-35minutes currently, ~9minutes after this change
microbit: ~12-15minutes currently, ~5 minutes after this change


cli/cli.ts
```diff
@@ -4084,212 +4115,263 @@ function testSnippetsAsync(snippets: CodeSnippet[], re?: string, pyStrictSyntaxC
...
-                    })
-                    pkg.filesToBePublishedAsync()
-                        .then(files => {
-                            Object.keys(files).forEach(f => {
-                                const fn = path.join(dir, f);
-                                pxt.debug(`writing ${fn}`);
-                                nodeutil.writeFileSync(fn, files[f], 'utf8');
-                            })
+            if (resp.outfiles && snippet.file && isLocal) {
+                const dir = snippet.file
```
jwunderl, line 4159: I disable the file logs in ci as it adds an extra few minutes / uses a bunch of extra as we have to keep around pkgs as filesToBePublished / those writes run async

cli/cli.ts
```diff
@@ -4084,212 +4115,263 @@ function testSnippetsAsync(snippets: CodeSnippet[], re?: string, pyStrictSyntaxC
...
 
-        let isPy = snippet.ext === "py"
-        let inFiles;
-        let extra = snippet.extraFiles || {};
-        if (isPy)
-            inFiles = { "main.ts": "", "main.py": snippet.code, "main.blocks": "", ...extra }
-        else
-            inFiles = { "main.ts": snippet.code, "main.py": "", "main.blocks": "", ...extra }
+        if (++codeSnippetCount % 40 == 0) {
+            // clean up cache every once in a while to speed up / clear up mem
```
jwunderl, line 4127: I fixed this here for now, but we should investigate this further -- it grinds to a complete halt after ~400-500 snippets without this & uses ~2 gb of ram at that point (vs ~400-700mb with this cleanup). Also note the change in service.ts related to reset, seems like that is likely a cause of us crashing at some points during long sessions.

cli/cli.ts
```diff
@@ -4084,212 +4115,263 @@ function testSnippetsAsync(snippets: CodeSnippet[], re?: string, pyStrictSyntaxC
...
+                    .replace(/\.ts$/, '')
+                    .replace(/\.py$/, '');
+                nodeutil.mkdirP(dir);
+                nodeutil.mkdirP(path.join(dir, "built"));
+                Object.keys(resp.outfiles).forEach(outfile => {
+                    const ofn = path.join(dir, "built", outfile);
+                    pxt.debug(`writing ${ofn}`);
+                    nodeutil.writeFileSync(ofn, resp.outfiles[outfile], 'utf8');
+                })
+                pkg.filesToBePublishedAsync()
```
jwunderl, line 4169: Maybe should be an await here, would take longer but it does slow significantly when a lot of these get backed up.

cli/cli.ts
```diff
@@ -4084,212 +4115,263 @@ function testSnippetsAsync(snippets: CodeSnippet[], re?: string, pyStrictSyntaxC
...
+            // ensure decompile to blocks works
+            const bresp = pxtc.service.performOperation("decompile", {
+                fileName: pxt.MAIN_TS,
+                options: opts,
+            }) as pxtc.CompileResult;
 
-                        // NOTE: neither of these decompile steps checks that the resulting code is correct or that
-                        // when the code is compiled back to ts it'll behave the same. This could be validated in
-                        // the future.
+            if (!bresp.success) {
```
jwunderl, line 4193: This will catch a few more errors we were ignoring-- we were passing an option to error on snippets that resulted in greyblocks but never reading it / just looking if it produced blocks. Looks like 1 snippet in arcade and ~3 in microbit, but all easy to fix (e.g. in microbit  they're in the blocks->javascript course, so should probably in either switched to a js snippet or just blocks-ignore'd)

cli/cli.ts
```diff
@@ -4084,212 +4115,263 @@ function testSnippetsAsync(snippets: CodeSnippet[], re?: string, pyStrictSyntaxC
...
+
+    pxt.log(`${successes.length}/${successes.length + failures.length} snippets compiled to blocks and python (and back), ${failures.length} failed`);
+    if (ignoreCount > 0) {
+        pxt.log(`Skipped ${ignoreCount} snippets`);
+    }
+    if (failures.length > 0) {
+        const msg = `${failures.length} snippets not compiling in the docs`
+        pxt.log(`${msg}:`);
+
+        failures.forEach(failure => {
```
jwunderl, line 4267: I found it annoying looking inline for the errors / having them trimmed / not listing the origin file, so I made it read them out in detail at the end

cli/cli.ts
```diff
@@ -4084,212 +4115,263 @@ function testSnippetsAsync(snippets: CodeSnippet[], re?: string, pyStrictSyntaxC
...
+                pxt.log(`    ${pxtc.getDiagnosticString(diag)}`);
+            });
+        });
+
+        if (!pxt.appTarget.ignoreDocsErrors) U.userError(msg);
+    }
+
+    function tsServiceCompile(options: pxtc.CompileOptions, retryOnError?: boolean): pxtc.CompileResult {
+        let resp = tryCompile(options);
+        if (retryOnError && (!resp.success || resp.diagnostics.length)) {
```
jwunderl, line 4279: This isn't currently getting hit except in real error cases, but was useful while fixing things / seems like little harm in leaving it in to make sure the error is in the snippet and not the incremental compile. takes ~3 seconds extra per ts snippet error

cli/cli.ts
```diff
@@ -4084,212 +4115,263 @@ function testSnippetsAsync(snippets: CodeSnippet[], re?: string, pyStrictSyntaxC
...
+    function getCompileOptsKey(pkgs: pxt.Map<string>) {
+        const keys = Object.keys(pkgs).sort();
+        return keys.reduce(
+            (prv, curr) => `${prv},${curr}:${pkgs[curr]}`,
+            ""
+        );
+    }
+
+    function pyComparisonString(s: string): string {
+        return s.split("\n")
```
jwunderl, line 4365: I rewrite this a little bit from the original replaces during one of the initial long tests, I think it behaves the same as before but it's not in a path that is used often.

cli/cli.ts
```diff
@@ -5434,13 +5518,20 @@ function internalCheckDocsAsync(compileSnippets?: boolean, re?: string, fix?: bo
             }
         });
 
-    function addSnippet(snippet: CodeSnippet, entryPath: string, snipIndex: number) {
+    function addSnippet(snippet: CodeSnippet, entryPath: string, snipIndex: number, src: string) {
+        const key = `${src}${snipIndex}`;
+        if (existingSnippets[key])
```
jwunderl, line 5523: If a file is linked both in the summary and the home page galleries, it was getting checked twice. This cuts out about ~125 snippets from arcade that were getting run twice.

cli/cli.ts
```diff
@@ -5616,14 +5707,16 @@ function internalCheckDocsAsync(compileSnippets?: boolean, re?: string, fix?: bo
...
+                                        .forEach((step, stepIndex) => getCodeSnippets(`${card.name}-step${stepIndex}`, step.contentMd)
                                             .forEach((snippet, snippetIndex) => {
                                                 snippet.packages = pkgs;
                                                 snippet.extraFiles = extraFiles;
+                                                const indexInStep = snippetIndex ? `-${snippetIndex}` : "";
                                                 addSnippet(
                                                     snippet,
-                                                    `tutorial${card.name}-${stepIndex}-${snippetIndex}`,
-                                                    cardIndex
+                                                    `tutorial/${card.name}-${indexInStep}step`,
```
jwunderl, line 5717: the numbering was a bit hard to parse which snippet was erroring (as it was something like `tutorial-example-4-0-5`) so should be a bit more clear now

pxtcompiler/emitter/service.ts
```diff
@@ -701,8 +701,14 @@ namespace ts.pxtc.service {
                 if (this.opts.fileSystem[fn] != v) {
                     this.fileVersions[fn] = (this.fileVersions[fn] || 0) + 1
                 }
-            })
-            this.opts = o
+            });
+            // shallow copy, but deep copy the file system
```
jwunderl, line 705: Was hitting a bug with this where it was manipulating the file system service host's file systems without updating file versions (we don't hit this in the webapp as the opts are always stringified / destringified when being passed to the worker). Can move this to be cloned in the cli but little cost to doing a shallow clone here

pxtcompiler/emitter/service.ts
```diff
@@ -851,7 +857,7 @@ namespace ts.pxtc.service {
 
     const operations: ServiceOps = {
         reset: () => {
-            service.cleanupSemanticCache();
+            service = ts.createLanguageService(host)
```
jwunderl, line 860: Can hide this behind a param to reset if it's concerning, but cleanupSemanticCache wasn't doing enough to clear / still resulting growing significantly and slowing things down / using tons of memory.

pxtcompiler/emitter/service.ts
```diff
@@ -1306,13 +1312,19 @@ namespace ts.pxtc.service {
...
             }
             let ts2asm = compile(host.opts, service)
-            res = { sourceMap: res.sourceMap, ...ts2asm }
-            if (res.needsFullRecompile) {
-                pxt.log("trigering full recompile")
+            res = {
+                sourceMap: res.sourceMap,
+                ...ts2asm,
+            }
+            if (res.needsFullRecompile || ((!res.success || res.diagnostics.length) && host.opts.clearIncrBuildAndRetryOnError)) {
```
jwunderl, line 1319: was running into some occasional (~5 times in arcade checkdocs) errors where having it retry here fixed things. This seems to imply a bug in the incremental compiles to me, but having the param to do a full recompile on error seemed fine to me.

cli/cli.ts
```diff
@@ -4084,212 +4115,263 @@ function testSnippetsAsync(snippets: CodeSnippet[], re?: string, pyStrictSyntaxC
...
+            const py = decompiled.outfiles[pxt.MAIN_PY];
+            const pySuccess = !!py && decompiled.success;
+            if (!pySuccess) {
+                console.log("ts2py error");
+                return addFailure(snippet.src, decompiled.diagnostics);
+            }
+            opts.fileSystem[pxt.MAIN_PY] = py;
+
+            // py to ts
+            const ts1 = opts.fileSystem["main.ts"];
```
riknoll, line 4212: pxt.MAIN_TS? also below

riknoll (approved):

pxtcompiler/emitter/service.ts
riknoll: as long as there isn't any noticeable perf drop, i think it's fine

pxtcompiler/emitter/service.ts
riknoll: @mmoskal fyi

livcheerful (approved):

darzu (approved):
This is awesome!

PR #8040, opened 2021-04-01
[stable6.12] Cherry-pick debug button and asset carryover changes

shakao:
- https://github.com/microsoft/pxt/pull/8029 - request from Kiki, she said that in user testing people don't think of the "Feedback" button as a way to report bugs or issues, so she wanted a separate "Bug" button that will toggle the Usabilla form
- https://github.com/microsoft/pxt/pull/8032 - slightly bigger change, but is related to https://github.com/microsoft/pxt/pull/7998; this fixes the asset carryover when you click "Start Fresh" on a skillmap activity. we can leave this out and not have the asset carryover for now but should probably sync with kiki to make sure that's okay

### DIFF
--- a/pxtlib/spriteutils.ts
+++ b/pxtlib/spriteutils.ts
@@ -649,6 +649,26 @@ namespace pxt.sprite {
         return result;
     }
 
+    export function encodeAnimationString(frames: BitmapData[], interval: number) {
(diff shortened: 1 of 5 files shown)


abchatra:
I would like to discuss #8029

PR #8042, opened 2021-04-01
Add filter options to monaco image editor

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/3384

- don't show the filter button if the filter option isn't there
- add filter to the field editor option for monaco 

The issue is tags weren't being populated in monaco, adding the option fixes it

### DIFF
--- a/pxteditor/monaco-fields/field_sprite.ts
+++ b/pxteditor/monaco-fields/field_sprite.ts
@@ -57,7 +57,7 @@ namespace pxt.editor {
                 initWidth: 16,
                 initHeight: 16,
                 blocksInfo: this.host.blocksInfo(),
-                filter: ""
+                filter: " "
             };
         }
     }
--- a/pxteditor/monaco-fields/field_sprite.ts
+++ b/pxteditor/monaco-fields/field_sprite.ts
@@ -57,7 +57,7 @@ namespace pxt.editor {
                 initWidth: 16,
                 initHeight: 16,
                 blocksInfo: this.host.blocksInfo(),
-                filter: "!tile !dialog !background"
+                filter: ""
             };
         }
     }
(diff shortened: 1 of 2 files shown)


pxteditor/monaco-fields/field_sprite.ts
```diff
@@ -56,7 +56,8 @@ namespace pxt.editor {
             return {
                 initWidth: 16,
                 initHeight: 16,
-                blocksInfo: this.host.blocksInfo()
+                blocksInfo: this.host.blocksInfo(),
+                filter: "!tile !dialog !background"
```
jwunderl, line 60: does this remove these from the gallery entirely? If it removes them completely it would make these inaccessible in text which isn't good (particularly for tile, the others are a bit rarely used)

livcheerful: oh yeah you're right. I'll just remove the filter and give it a whirl

webapp/src/components/ImageFieldEditor.tsx
```diff
@@ -51,7 +51,8 @@ export class ImageFieldEditor<U extends pxt.Asset> extends React.Component<Image
             currentView: "editor",
             headerVisible: true,
             filterOpen: false,
-            gallerySelectedTags: []
+            gallerySelectedTags: [],
+            galleryFilter: " "
```
jwunderl, line 55:
```suggestion
            galleryFilter: ""
```

Just for cleanliness, no need to have it be truthy here I think.

Also probably should update https://github.com/microsoft/pxt/pull/8042/commits/8ea2dd866d21ccf8437814d76d94f0767edf85ad#diff-d27a987847063b48bbd409fa5ea5df1e2d7655def806979aba7492322faa0e36R24 to remove the ?, as it's no longer optional? or leave as optional  and handle in the filter part below like

```
const tags = this.state.galleryFilter?.split(" ")
                .filter(el => !!el)
                .map(el => el.toLowerCase()) ?? [];
```

jwunderl (approved):

PR #8046, opened 2021-04-02
Green screen safari fix

pelikhan:
Safari changed it's policy and will not provide label, deviceId on enumerated user media devices until getUserMedia is called, and user consent is given.
Fix for https://github.com/microsoft/pxt-microbit/issues/3633

Tested on Safari MaxOSX, Edge.

image

### DIFF
--- a/webapp/src/greenscreen.tsx
+++ b/webapp/src/greenscreen.tsx
@@ -40,6 +40,7 @@ export class WebCam extends data.Component<WebCamProps, WebCamState> {
 
     handleDeviceClick(deviceId: any) {
         this.setState({ hasPrompt: false });
+        pxt.debug(`greenscreen: device ${deviceId}`)
         this.deviceId = deviceId;
         // deviceId is "" if green screen selected
         if (this.deviceId) {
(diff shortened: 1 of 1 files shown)


webapp/src/greenscreen.tsx
```diff
@@ -77,11 +82,18 @@ export class WebCam extends data.Component<WebCamProps, WebCamState> {
...
+            // first ask for permission from ther user so that
+            // labels are populated in enumerateDevices
+            navigator.mediaDevices.getUserMedia({ audio: false, video: true })
+            .then(() => navigator.mediaDevices.enumerateDevices())
+            .then(devices => {
+                this.setState({ devices: devices.filter(device => device.kind == "videoinput") });
+            }, e => {
+                pxt.debug(`greenscreen: enumerate devices failed`)
+                console.error(e);
+            });
```
jwunderl, line 95:
```suggestion
                .then(() => navigator.mediaDevices.enumerateDevices())
                .then(devices => {
                    this.setState({ devices: devices.filter(device => device.kind == "videoinput") });
                }, e => {
                    pxt.debug(`greenscreen: enumerate devices failed`)
                    console.error(e);
                });
```

nit indentation

jwunderl (approved):

webapp/src/greenscreen.tsx
pelikhan: the default text formatter does not enforce those, might be time to move to prettier who is much more aggressive about indentation,

abchatra (approved):

PR #8039, opened 2021-04-01
Migration from TSLint to ESLint

darzu:
# Migration from TSLint to ESLint

I primarily followed: https://code.visualstudio.com/api/advanced-topics/tslint-eslint-migration.

I used "tslint-to-eslint-config" CLI tool to generate a .eslintrc file using our tslint.json, however there were lots of rules that tslint-to-eslint-config didn't know about.

Some of the gaps from "tslint-microsoft-contrib" where filled in by "@microsoft/eslint-plugin-sdl" which is sort of its successor.

Some of the gaps from "react-a11y" tslint plugin where filled by "jsx-a11y".

Bizarrely, a couple of files (blocks.tsx and snippetBuilder.tsx; see .eslintignore) cause the ESLint to crash deep inside the TS compiler ☹️. So I had to disable these files from linting for now, but I suspect we just need to update our TS version (future PR).

## Steps for Devs
After this PR is merged, do the following in pxt (no other repo should be affected):
0. `git pull`
1. `rm -rf node_modules/`
2. `npm install`
3. in vscode, uninstall or disable tslint extension (if you want)
4. in vscode, install "dbaeumer.vscode-eslint" extension
5. run `gulp && gulp lint` to make sure everything works
6. in vscode, inside a file (say webapp/src/app.tsx), try adding some trailing whitespace. You should see this squigly error: `Trailing spaces not allowed. eslintno-trailing-spaces`
7. you're done!

## Removed rules
These rules where removed and tslint-to-eslint-config couldn't find replacements:
<details>
    <summary>Expand</summary>

 - "jquery-deferred-must-complete"
 - "mocha-avoid-only"
 - "mocha-no-side-effect-code"
 - "no-delete-expression"
 - "no-disable-auto-sanitization"
 - "no-document-domain"
 - "no-document-write"
 - "no-exec-script"
 - "no-function-constructor-with-string-args"
 - "no-http-string"
 - "no-jquery-raw-elements"
 - "no-string-based-set-immediate"
 - "no-string-based-set-interval"
 - "no-string-based-set-timeout"
 - "no-unnecessary-bind"
 - "no-unnecessary-override"
 - "no-unsupported-browser-code"
 - "no-with-statement"
 - "non-literal-require"
 - "possible-timing-attack"
 - "react-anchor-blank-noopener"
 - "react-no-dangerous-html"
 - "react-this-binding-issue"
 - "react-unused-props-and-state"
</details>
<br>

[...]

PR #8047, opened 2021-04-02
Upgrade Typescript from v3.7 to v4.2

darzu:
Now that we've moved to ESLint (https://github.com/microsoft/pxt/pull/8039), we can now upgrade our Typescript version! Hurray!

Also, this lets us remove the lint exceptions I had to introduce for blocks.tsx and snippetBuilder.tsx because those were crashing eslint due to a typescript bug.

The upgrade went surprisingly smooth... just a few new complaints about void types and removing the "name: void" global by default.

### DIFF
--- a/pxtsim/runtime.ts
+++ b/pxtsim/runtime.ts
@@ -156,7 +156,7 @@ namespace pxsim {
 
         export async function promiseTimeout<T>(ms: number, promise: T | Promise<T>, msg?: string): Promise<T> {
             let timeoutId: any;
-            let res: () => void;
+            let res: (_?: any) => void;
 
             const timeoutPromise: Promise<T> = new Promise((resolve, reject) => {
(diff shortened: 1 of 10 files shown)


pxtsim/runtime.ts
```diff
@@ -156,7 +156,7 @@ namespace pxsim {
 
         export async function promiseTimeout<T>(ms: number, promise: T | Promise<T>, msg?: string): Promise<T> {
             let timeoutId: any;
-            let res: () => void;
+            let res: (_?: any) => void;
```
jwunderl, line 159:
```suggestion
            let res: (v?: T | PromiseLike<T>) => void;
```

to keep in sync with util.ts:promiseTimeout / not use any

jwunderl (approved):
LGTM

PR #8048, opened 2021-04-02
Update monaco 0.19 -> 0.23

darzu:
Prerequisite for: https://github.com/microsoft/pxt-microbit/issues/3940

## Steps for updating Monaco (in this PR and hopefully useful for the future):

0. bump monaco-editor version in package.json & npm install
1. Patch monaco.tsx and other places where API changes might have occured
1. Download latest codicon (https://github.com/microsoft/vscode-codicons/blob/main/dist/codicon.ttf) and replace version at (theme/external-font/codicon.ttf)
2. Verify all the paths match up (in this PR, `ui/codiconLabel/codicon/codicon.ttf` had to be changed to `ui/codicons/codicon/codicon.ttf` based on new monaco-editor paths)
0. run gulp (this runs the `copyMonaco` task), check in all the changes
0. update localtypings/monaco.d.ts from node_modules/monaco-editor/

## Test plan:

0. test specifically on iOS and android to ensure we don't regress
1. Check that all the icons are present
2. Test large JS and PY projects, try features like find and replace, text selection, go to line, command palette

Try it here: https://arcade.makecode.com/app/e1d966184d6fd748667a9a38283c3c02f00df890-68169a79a7

PR #8049, opened 2021-04-02
Re-enable default Monaco .js/.ts language services (auto-complete / intellisense)

darzu:
~~DRAFT:~~
- [x] waiting for: https://github.com/microsoft/pxt/pull/8048 
  - [x] rebase after
- [x] Ensure python doesn't regress

# Re-enable default Monaco TS language service

Fixes: https://github.com/microsoft/pxt-microbit/issues/3940

Basically undo'ing these two PRs:
https://github.com/microsoft/pxt/pull/6587
https://github.com/microsoft/pxt/pull/6531

Test here:
https://arcade.makecode.com/app/4ff7fbf8b794a7593b0d7ec08f38a9c45ee4ef02-ff2060082f
compare to:
https://arcade.makecode.com

# MakeCode TS vs. Monaco TS lang-service
Here's some comparison between our Typescript vs Monaco Typescript:


darzu:
Should we resurrect https://github.com/microsoft/pxt-monaco-typescript ?
To figure out: what advantage is there to having our own fork instead of using the vanilla service?

pelikhan:
My vote for no fork.

We started doing some custom filtering, sorting there but I am thinking that we should really make sure that the out-of-box experience works great (and it'll also work great in vscode too).

darzu:
> My vote for no fork.
> 
> We started doing some custom filtering, sorting there but I am thinking that we should really make sure that the out-of-box experience works great (and it'll also work great in vscode too).

I'm leaning that way as well. I think there's value in matching VSCode's experience (this makes sense for our transition story), and it's also a nice workload reduction for ourselves. Turns out, maintain a language service, even if you're just doing a few tweaks, is cumbersome.

riknoll (approved):
LGTM. We should make sure that we enumerate the differences for the whole team though. I think they mainly are:

Completions no longer include nice snippets for function signatures (e.g. images, functions)
No sorting based on block weight
No filtering of _ names

We might be able to "extend" the language service instead of forking by intercepting the results of the language service calls. Might let us bring back the *nice* things we had (sorting based on weight, filtering _). But it could be a rabbit hole so maybe not. Also it would be tough to do our old signature completions without forking.

jwunderl (approved):

darzu:
> LGTM. We should make sure that we enumerate the differences for the whole team though. I think they mainly are:
> 
> * Completions no longer include nice snippets for function signatures (e.g. images, functions)
> * No sorting based on block weight
> * No filtering of _ names
> 
> We might be able to "extend" the language service instead of forking by intercepting the results of the language service calls. Might let us bring back the _nice_ things we had (sorting based on weight, filtering _). But it could be a rabbit hole so maybe not. Also it would be tough to do our old signature completions without forking.

Yes, I agree those are the main trade offs. I'll take a stab at "extending" the language service.

----
@abchatra , @Jaqster , are you okay with these trade-offs? Or should we have a meeting to discuss?
To summarize: We're bringing back the VSCode/Monaco language service which generally preforms much better and more reliably than the one we built ourselves (which we built to support Python and presumed (wrongly) could preform just as well for TS). See the gifs above for many super nice features that are present in VSCode. The downsides compared to what we have now are:

- Completions no longer include nice snippets for function signatures (e.g. images, functions)
- No sorting based on block weight
- No filtering of _ names
- in general, less control to tweak things

We might be able to overcome these by extending or forking the language service, but this is a road that got us into trouble last time. It's unsupported, difficult to maintain esp. when upgrading Monaco & typescript, and the difficulty of it was a major factor in why we thought rolling our own language service was the better path.

Jaqster:
Let's discuss when Abhijith gets back.  In theory I like sticking with main Monaco language service and getting all their latest updates, but I also think that we need a kid-friendly version of it.  I like the idea of having some sort of "MakeCode extension" for the language service that adds filters/sorting and 'nice snippets'.

darzu:
i'm going to merge this for now because it is much better than the status quo and helps my work on other bugs and we'll continue the discussion after Abhijith gets back.
Abhijith and I talked about this before he left and we do think going with Monaco is the right approach, and we'll hash out the other details in future PRs.

pelikhan:
Maybe a sustainable long term strategy is to slowly add the required extension points upstream in Monaco to avoid forking and allow refinements on completion results.

abchatra:
I am not quite happy about the following 2:

abchatra:
@dbaeumer might just see this thread and help us here.

darzu:
Filter at least should be fairly easily achievable with a thin wrapper. 

Weight maybe too, but we'll need to interface with our symbol lookups which might be tough; might be the easiest way is to fork again.. which we can do b/c we know the pattern but I honestly don't think it's worth the maintenance cost.

abchatra:
Let's discuss with the vscode first on a long term solution.

dbaeumer:
@abchatra actually I am not involved in the Monaco editor language services. Best would be to open an issue here: https://github.com/microsoft/monaco-editor

darzu:
These issues are relavent:

https://github.com/microsoft/monaco-editor/issues/17
- This is people asking to filter "_" just like us, it ends with a link to:

https://github.com/microsoft/monaco-typescript/pull/65
- This makes it easy to subclass our own TypescriptWorker which can selectively override the default TS worker behavior.

https://github.com/microsoft/monaco-editor/issues/2368
- Someone asks about sorting and concludes using "sortText" via the completion API (like we used to) is the way

https://github.com/microsoft/monaco-editor/issues/1077
- Again discussing filtering and sorting, the solution seems to be shimming the worker

https://github.com/microsoft/monaco-editor/issues/1850
- Some discussion of extending completions but the attempts seem centered around shimming the completion provider again

To summarize: lots of people have asked / are asking for functionality like this. The official solution is to subclass the TypescriptWorker in our own custom worker and selectively extend or shim. This is perfectly adequate and better than what we had before. All the complexity will revolve around interfacing with our own compiler output which is entirely our own doing.

PR #8050, opened 2021-04-02
Fix immersive reader in electron

jwunderl:
Api requests in electron expect the local token for authentication, where auth.apiAsync attaches the user's token for authentication - so typically we wanna use Cloud.apiRequestWithCdnAsync. In this case we always want to hit the live endpoint too (makecode.com/api/immreader instead of localhost/api/immreader), so this fixes in electron. My understanding is that we want these to hit through the cdn anyways? e.g. https://pxt.azureedge.net/api/immreader?cdn=20210402 (I believethe cdn= changes every 12 hours so that should be good with the refreshes, though I'll double check -- if it's 24 hours will have to fix that or do a different fix.)

test https://arcade.makecode.com/app/d32c7f199fbdc6a758b45eb7255cda3b1caacada-7050ed1730

image

Also I think small side fix in https://github.com/microsoft/pxt/commit/61eba0bed2cda340551e41292c19664b337f35b4 @livcheerful, let me know if it was intentional but the removeLocal wasn't removing the token as the key didn't match

### DIFF
--- a/pxtlib/emitter/cloud.ts
+++ b/pxtlib/emitter/cloud.ts
@@ -34,18 +34,18 @@ namespace pxt.Cloud {
             && !BrowserUtils.isLocalHost() && !!pxt.webConfig.cdnUrl
     }
 
-    export function cdnApiUrl(url: string) {
+    export function cdnApiUrl(url: string, hourly?: boolean) {
         url = url.replace(/^\//, '');
         if (!useCdnApi())
             return apiRoot + url;
 
         const d = new Date()
-        const timestamp = d.getUTCFullYear() + ("0" + (d.getUTCMonth() + 1)).slice(-2) + ("0" + d.getUTCDate()).slice(-2)
+        const timestamp = d.getUTCFullYear() + ("0" + (d.getUTCMonth() + 1)).slice(-2) + ("0" + d.getUTCDate()).slice(-2);
         if (url.indexOf("?") < 0)
             url += "?"
         else
             url += "&"
-        url += "cdn=" + timestamp
+        url += `cdn=${timestamp}${hourly ? d.getUTCHours() : ""}`;
         // url = url.replace("?", "$")
         return pxt.webConfig.cdnUrl + "/api/" + url
     }
@@ -53,7 +53,7 @@ namespace pxt.Cloud {
     export function apiRequestWithCdnAsync(options: Util.HttpRequestOptions) {
         if (!useCdnApi())
             return privateRequestAsync(options)
-        options.url = cdnApiUrl(options.url)
+        options.url = cdnApiUrl(options.url, options.hourlyCache)
         return Util.requestAsync(options)
             .catch(e => handleNetworkError(options, e))
     }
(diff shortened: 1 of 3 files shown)


webapp/src/immersivereader.tsx
```diff
@@ -171,21 +171,23 @@ function beautifyText(content: string): string {
...
-                    pxt.storage.removeLocal("/api/immreader");
-                    console.log("immersive reader fetch token error: " + JSON.stringify(res.err));
-                    pxt.tickEvent("immersiveReader.error", {error: JSON.stringify(res.err)})
+            return Cloud.apiRequestWithCdnAsync({ url: "immreader", forceLiveEndpoint: true }).then(
+                res => {
+                    pxt.storage.setLocal(IMMERSIVE_READER_ID, JSON.stringify(res.json));
+                    return res.json;
+                },
+                e => {
+                    pxt.storage.removeLocal(IMMERSIVE_READER_ID);
```
livcheerful, line 185: wow good call omg

livcheerful (approved):

jwunderl:
Ahh looks like the cdn= is [only changed every 24 hours](https://github.com/microsoft/pxt/blob/61eba0bed2cda340551e41292c19664b337f35b4/pxtlib/emitter/cloud.ts#L43) and not 12 hours like I though, will see how else to fix. (I guess could just add an optional param to cdnApiUrl that would make it refresh hourly? Seems reasonable to me at least)

jwunderl:
Okay updated to make it hit the cdn with an hourly cache instead of a 24hour cache - e.g. https://pxt.azureedge.net/api/immreader?cdn=2021040223. All other requests still hit daily, it's toggled by an option to Cloud.apiRequestWithCdnAsync Here's an updated build:

https://arcade.makecode.com/app/b3170605f9310b1258308471a6e9c2caa016b2bd-6aec3e2b8f

eanders-MS:
> Okay updated to make it hit the cdn with an hourly

Why does it need to hit cdn at all? Could this make the immersive reader fail for up to an hour if the token expired in the meantime?

livcheerful:
The backend actually provides a new token every 12 hours, and each token is valid for 24 hours so.....I think it will be fine?

eanders-MS:
> The backend actually provides a new token every 12 hours, and each token is valid for 24 hours so.....I think it will be fine?

Oh that makes sense!

pxtlib/emitter/cloud.ts
```diff
@@ -34,26 +34,26 @@ namespace pxt.Cloud {
...
 
         const d = new Date()
-        const timestamp = d.getUTCFullYear() + ("0" + (d.getUTCMonth() + 1)).slice(-2) + ("0" + d.getUTCDate()).slice(-2)
+        const timestamp = d.getUTCFullYear() + ("0" + (d.getUTCMonth() + 1)).slice(-2) + ("0" + d.getUTCDate()).slice(-2);
         if (url.indexOf("?") < 0)
             url += "?"
         else
             url += "&"
-        url += "cdn=" + timestamp
+        url += `cdn=${timestamp}${hourly ? d.getUTCHours() : ""}`;
```
abchatra, line 48: Hmm. I am not sure this is how it works. Backend looks for timestamp https://github.com/microsoft/pxt-backend/blob/e952a311066087dcbb42c872a26bd8504bff1d94/node/src/main.ts#L104

abchatra (changes requested):

pxtlib/emitter/cloud.ts
jwunderl: Ahh you're right, I misunderstood the comparison on the timestamp; closing for now and will get a better fix then

abchatra:
For now you could just disable immersive reader for offline app. We can take our time and figure out.

eanders-MS:
Skip cdn?

PR #8051, opened 2021-04-03
Fix immersive reader in electron (v2)

jwunderl:
Similar to previous one but not using cdn, just going through requestAsync rather than cdnapiurl. Added a redirect in the local serve rather than an explicit check like in auth.ts so `pxt serve` can still fetch tokens.

### DIFF
--- a/cli/server.ts
+++ b/cli/server.ts
@@ -973,6 +973,13 @@ export function serveAsync(options: ServeOptions) {
                 return
             }
 
+            if (elts[1] == "immreader") {
+                let trg = Cloud.apiRoot + elts[1];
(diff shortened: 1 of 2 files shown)


pelikhan:
Électron being a full offline scenario, you probably want to better handle slow/no internet scenarios.

livcheerful (approved):

livcheerful:
Hmmm I think to be able to handle poor-internet cases, we'd have to do some hacking around the Immersive Reader -- most of the functionality is hidden away in the SDK. Definitely doable if we want to, [since it's open](https://github.com/microsoft/immersive-reader-sdk), but will probably be a big change!

PR #8053, opened 2021-04-05
need to allow A-F in github token check

tballmsft:


pelikhan (approved):
Or do /i

jwunderl:
Hm looks like we also need to allow `_` and e-z now https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats/

tballmsft:
yep.

abchatra:
@tballmsft  @jwun Does this need to ported to other target like microbit?

jwunderl:
It's an easy port and we still need to release the live=1 fix for micro:bit anyway so might as well (along with https://github.com/microsoft/pxt/pull/8055 which fixes the rest of the problem).

This only hits dev tokens right now though, and not oauth which I would expect to be much more commonly used, so it's not *that* urgent (and can work around by logging in in beta and then switching to live)

abchatra:
Can you please port it to stable channel in microbit? I am doing a point fix release this week.

PR #8054, opened 2021-04-05
Fix crash in getSnippet for PY auto-complete

darzu:
Fixes: https://github.com/microsoft/pxt-arcade/issues/3379


jwunderl (approved):
To confirm, the crash was from getParameterDefault eventually calling createDefaultFunction which refs functionCount before it was defined?

darzu:
> To confirm, the crash was from getParameterDefault eventually calling createDefaultFunction which refs functionCount before it was defined?

Yup!

PR #8055, opened 2021-04-05
allow new github token format

jwunderl:
github changed their format today so our 'looks kinda like a token' check is failing. New format info: https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats/

could trim down the regex a bit more based off the valid token types but I don't see much of a reason to do so, if the user puts in something invalid it will fail a bit further down. (Conversely, could be a bit more general with just `/^[a-z0-9_]+$/i` to allow old tokens, but this is only run through the 'login to github dialog' so they'd have to be using a token created before today / really they *should* create a new token anyway in that case.)

here's a build: https://arcade.makecode.com/app/0f247e4a7a42ba773e2a1810f4acec7bc9c59e1f-20861107f6

### DIFF
--- a/webapp/src/githubprovider.tsx
+++ b/webapp/src/githubprovider.tsx
@@ -189,36 +189,30 @@ export class GithubProvider extends cloudsync.ProviderBase {
         core.showLoading(LOAD_ID, lf("validating GitHub token..."));
         return Promise.resolve()
             .then(() => {
-                if (hextoken.length != 40 || !/^gh[pousr]_[a-z0-9_]+$/i.test(hextoken)) {
-                    pxt.tickEvent("github.token.invalid");
-                    core.errorNotification(lf("Invalid token format"))
-                    return Promise.resolve();
-                } else {
-                    pxt.github.token = hextoken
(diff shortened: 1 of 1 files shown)


webapp/src/githubprovider.tsx
```diff
@@ -189,7 +189,7 @@ export class GithubProvider extends cloudsync.ProviderBase {
         core.showLoading(LOAD_ID, lf("validating GitHub token..."));
         return Promise.resolve()
             .then(() => {
-                if (hextoken.length != 40 || !/^[a-fA-F0-9]+$/.test(hextoken)) {
+                if (hextoken.length != 40 || !/^gh[pousr]_[a-z0-9_]+$/i.test(hextoken)) {
```
eanders-MS, line 192: I suggest removing the validation check entirely. Let GitHub validate it and we handle the result code accordingly.

jwunderl: Seems good to me!

eanders-MS (approved):
:shipit:

PR #8063, opened 2021-04-06
Apply high contrast styles when OS high contrast is enabled

shakao:
this fix adds our "hc" class when it detects that the operating system's high contrast is on. this seems like the right behavior because it ensures all future fixes to our high contrast will also propagate to windows/os high contrast users

one gotcha is if a user is swapping repeatedly between high-contrast on/off (i'm not sure if this is a real scenario?)--if you turn windows high contrast on, you'll need to refresh makecode to apply our high contrast styles. we also save high contrast preferences, so if you turn off the OS high contrast, makecode will still be high contrast for that session.

\+ one additional fix to invert the logo color if the user is in "light-mode" high contrast

fixes https://github.com/microsoft/pxt-microbit/issues/3551
fixes https://github.com/microsoft/pxt-microbit/issues/3549
fixes https://github.com/microsoft/pxt-microbit/issues/3562
fixes https://github.com/microsoft/pxt-microbit/issues/3565

### DIFF
--- a/theme/highcontrast.less
+++ b/theme/highcontrast.less
@@ -874,6 +874,14 @@
     }
 }
 
+@media (forced-colors: active) and (prefers-color-scheme: light) {
+    #homemenu .logo.organization img,
+    #homemenu .logo.brand img {
+        filter: invert(1);
+    }
+}
+
+
 /* > Small Monitor */
 @media only screen and (min-width: @computerBreakpoint) {
     /* Hide the editor toolbor in tutorial mode */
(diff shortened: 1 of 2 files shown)


jwunderl (approved):
Nice! I think there are a few more this probably fixes too but we can find those when we get to them

(out of curiousity, does this mean if user is logged in while in high contrast mode, we'll persist high contrast mode with user settings in cloud sync? Not sure that's the wrong behavior / just curious)

shakao:
yep! initially i was like "that's a feature!" but i'm willing to be convinced otherwise if people think it's weird haha. the other thing we could do is move this check to the render function so it applies the "hc" class without setting high contrast to true internally

jwunderl:
Sounds good to me, I'm leaning towards "that's a feature" as well as I can't imagine anyone using os high contrast unintentionally.

livcheerful (approved):

PR #8072, opened 2021-04-08
Always use new project pxt.json for skillmap carryover

shakao:
bug: picking "Keep Code" from space explorer activity3 -> activity4 was loading without the Status Bars extension, because we were overwriting the config of the new project and erasing the dependencies.

we never want to take the pxt.json of the old project when doing code carryover, we only want to push the tilemap/image.g.jres files into the new config if the filenames are not there (eg activity A1 has images, activity A2 does not, so when carrying over assets from A1 -> A2 we need to add `images.g.jres` to the list of files in pxt.json)


livcheerful (approved):

riknoll (approved):

riknoll:
Might want to make sure all the appropriate files are in the files array

jwunderl (approved):

PR #8073, opened 2021-04-08
Fix py string.replace

darzu:
The JS and PY String.replace behaviors are similar enough we can do a straight translation. Python technically has an option for a third param, but this is rare and if it comes up we can use a py override.

Fixes: https://github.com/microsoft/pxt-microbit/issues/3831

### DIFF
--- a/libs/pxt-python/pxt-python-helpers.ts
+++ b/libs/pxt-python/pxt-python-helpers.ts
@@ -119,11 +119,6 @@ namespace _py {
         return str;
     }
 
-    export function py_string_replace(str: string, oldString: string, newString: string, count?: number): string {
-        nullCheck(str);
-        return str;
-    }
-
     export function py_string_rfind(str: string, sub: string, start?: number, end?: number): number {
         nullCheck(str);
         return 0;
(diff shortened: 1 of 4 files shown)


libs/pxt-python/pxt-python-helpers.ts
```diff
@@ -119,11 +119,6 @@ namespace _py {
         return str;
     }
 
-    export function py_string_replace(str: string, oldString: string, newString: string, count?: number): string {
```
jwunderl, line 122: Just to confirm, my recollection here is that this is only ever used as a shim / is not available to be called directly from the editor, right? (reason for asking being whether a user project could have a direct dependency on the function being available)

jwunderl (approved):
LGTM (though one question \\/  in case I'm misremembering and we do need an upgrade rule)

libs/pxt-python/pxt-python-helpers.ts
darzu: Good point. It could be called directly from the editor (in JS). However in this case it isn't implemented, it's just the identity function, so it'd be a very odd dependency for someone to have.

darzu: But you're point stands: in the future, we can't get rid of these functions once people start using them. (When they have implementations.)

PR #8074, opened 2021-04-08
Properly exclude python keywords & reserved words from ALL declarations

darzu:
Fixes: https://github.com/microsoft/pxt-microbit/issues/3145

- buildRenameMap considers ALL declarations (functions, classes, etc) instead of just variable declarations
- Add PY keywords to the reserved words list (should always have been there.)

### DIFF
--- a/pxtcompiler/emitter/decompiler.ts
+++ b/pxtcompiler/emitter/decompiler.ts
@@ -273,13 +273,18 @@ namespace ts.pxtc.decompiler {
         }
     }
 
+    export type RenameMapOpts = {
+        declarations: "variables" | "all"
+        takenNames: NamesSet,
+    }
     export type NamesSet = pxt.Map<boolean | {}>
     /**
      * Uses the language service to ensure that there are no duplicate variable
      * names in the given file. All variables in Blockly are global, so this is
      * necessary to prevent local variables from colliding.
      */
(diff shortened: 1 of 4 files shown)


pxtcompiler/emitter/decompiler.ts
```diff
@@ -293,12 +293,12 @@ namespace ts.pxtc.decompiler {
 
             function checkChildren(n: Node): void {
                 ts.forEachChild(n, (child) => {
-                    if (child.kind === SK.VariableDeclaration && (child as ts.VariableDeclaration).name.kind === SK.Identifier) {
-                        const name = (child as ts.VariableDeclaration).name.getText();
+                    if (ts.isDeclarationName(child)) {
```
darzu, line 296: Note that this change expands the rename map to consider all declarations not just variable declarations. This is used in both the PY and Blocks decompiler. @riknoll , can you think of any reason why we wouldn't want all declarations?

darzu: Nvm, I'm rolling this back so blockly decompiler only looks for variable declarations and py decompiler looks at all declarations. There were specific unit tests for blockly decomp looking for only variable renames (and not functions) so I'll leave this one alone.

jwunderl (approved):

PR #8076, opened 2021-04-08
Handle snippets with default values translating to py

darzu:
Fixes: https://github.com/microsoft/pxt-microbit/issues/3281

We weren't doing any sort of TS->PY translation for specified defaults on blocks.

### DIFF
--- a/pxtcompiler/emitter/languageservice.ts
+++ b/pxtcompiler/emitter/languageservice.ts
@@ -77,7 +77,7 @@ namespace ts.pxtc.service {
     }
 
     export function tsKeywordToPyKeyword(keyword: string): string | undefined {
-        const map: { [tsType: string]: string } = {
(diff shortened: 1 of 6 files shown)


pxtcompiler/emitter/languageservice.ts
```diff
@@ -76,6 +76,14 @@ namespace ts.pxtc.service {
         return [...retApis, ...completionSymbols(enumVals, COMPLETION_DEFAULT_WEIGHT)]
     }
 
+    export function tsKeywordToPyKeyword(keyword: string): string | undefined {
+        const map: { [tsType: string]: string } = {
```
jwunderl, line 80:
```suggestion
        const map: pxt.Map<string> = {
```

nit, we have our map type we use all over~

jwunderl (approved):
LGTM

PR #8077, opened 2021-04-08
Surface more Immersive Reader errors

livcheerful:
Add an immersiveReader.error tick for when the immersive reader returns an error -- right now we can't see when the users are getting the "Immersive Reader is already launching" issue.

Also @jwunderl I think you pointed this out in the initial PR and I forgot, but I removed the double package dependency -- it was required for prod and dev before, and this removes it for dev.

### DIFF
--- a/webapp/src/immersivereader.tsx
+++ b/webapp/src/immersivereader.tsx
@@ -216,8 +216,12 @@ export function launchImmersiveReader(content: string, tutorialOptions: pxt.tuto
                 core.warningNotification(lf("Immersive Reader cannot be used offline"));
                 break;
             }
+            case "token": {
+                break;
+            }
             default: {
                 core.warningNotification(lf("Immersive Reader could not be launched"));
+                pxt.tickEvent("immersiveReader.error", {error: JSON.stringify(e)});
             }
         }
         console.log("Immersive Reader Error: " + e);
--- a/webapp/src/immersivereader.tsx
+++ b/webapp/src/immersivereader.tsx
@@ -182,7 +182,8 @@ function getTokenAsync(): Promise<ImmersiveReaderToken> {
                 },
                 e => {
                     pxt.storage.removeLocal(IMMERSIVE_READER_ID);
-                    pxt.tickEvent("immersiveReader.error", e);
+                    pxt.reportException(e)
+                    pxt.tickEvent("immersiveReader.error", {error: e.statusCode, message: e.message});
                     return Promise.reject(new Error("token"));
                 }
             );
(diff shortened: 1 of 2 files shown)


webapp/src/immersivereader.tsx
```diff
@@ -216,8 +216,12 @@ export function launchImmersiveReader(content: string, tutorialOptions: pxt.tuto
                 core.warningNotification(lf("Immersive Reader cannot be used offline"));
                 break;
             }
+            case "token": {
+                break;
+            }
             default: {
                 core.warningNotification(lf("Immersive Reader could not be launched"));
+                pxt.tickEvent("immersiveReader.error", {error: JSON.stringify(e)});
```
jwunderl, line 224:
```suggestion
                pxt.reportError("immersiveReader.error", JSON.stringify(e));
```

I think? ping @abchatra to make sure as I'm not sure I remember the normal way to post error ticks

jwunderl (approved):
Ah just noticed the devdep warning for that as well, good timing!

webapp/src/immersivereader.tsx
abchatra: Both are fine. reportError goes to exceptions table which is harder to use. I am not sure why we are stringify the result code. 

I see ticks like this:

{"target":"arcade","cookie":"false","stage":"","ElectronVersion":"8.3.1","Electron":"1","PxtElectronVersion":"0.7.5","PxtElectronIsProd":"1","PxtTargetVersion":"1.4.33","ChromiumVersion":"80.0.3987.165","PxtCoreVersion":"6.12.14","PxtElectron":"1","NodeVersion":"12.13.0","error":"{\"statusCode\":403}"}
--

{"target":"arcade","cookie":"true","stage":"","error":"{\"statusCode\":0}"}
--

Why not just write the { error : e.statusCode} ?

abchatra: I actually prefer tickEvent

livcheerful: some of the errors don't have status code, or a message. I thought it would be good to just have a catch all, but I can change it

pelikhan: There should be reportException to log catch all errors.

webapp/src/immersivereader.tsx
```diff
@@ -182,8 +182,7 @@ function getTokenAsync(): Promise<ImmersiveReaderToken> {
                 },
                 e => {
                     pxt.storage.removeLocal(IMMERSIVE_READER_ID);
-                    console.log("immersive reader fetch token error: " + JSON.stringify(e));
-                    pxt.tickEvent("immersiveReader.error", { error: JSON.stringify(e) });
+                    pxt.tickEvent("immersiveReader.error", e);
```
pelikhan, line 185: I don't think we ever log the entire exception in tickEvent, it will grab the stack trace which might be large.

pelikhan: you can call reportException to get the exception details and also call tickEvent to see in aggregate analytics

abchatra (approved):

PR #8080, opened 2021-04-09
Support sort() and sort() w/ lambda in py

darzu:
Fixes: https://github.com/microsoft/pxt-microbit/issues/3374

We supported lambdas in ts2py but not py2ts so I created a clearer error message and disabled them in ts2py for now.

Also, I allow the optional sort function in ts<->py even though py doesn't technically support that.

### DIFF
--- a/pxtpy/pydecompiler.ts
+++ b/pxtpy/pydecompiler.ts
@@ -41,6 +41,11 @@ namespace pxt.py {
     }
 
     ///
+    /// FLAGS
+    ///
+    const SUPPORT_LAMBDAS = false;
+
+    ///
     /// UTILS
     ///
     export const INDENT = "    "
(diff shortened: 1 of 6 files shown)


pxtpy/pydecompiler.ts
```diff
@@ -40,6 +40,11 @@ namespace pxt.py {
         throw err;
     }
 
+    ///
+    /// FLAGS
+    ///
+    const SUPPORT_LAMBDAS = false;
```
jwunderl, line 46: I'm guessing his is non-permanent / just to make it easier to find in the future when we do support lambdas, right?

jwunderl (approved):

pxtpy/pydecompiler.ts
darzu: Yes exactly.

PR #8081, opened 2021-04-09
Always use the pxt.json name in cards, even for github projects

pelikhan:
Don't try to do too much magic and simply display the project name from pxt.json.

This solves rendering issues with mon-repo github projects.


pelikhan:
Fix for https://github.com/microsoft/pxt-microbit/issues/3914

jwunderl (approved):

PR #8082, opened 2021-04-09
always update header name on script load

pelikhan:
On load, always make sure the header name matches what is in pxt.json (or "untitled")


pelikhan:
Fix for https://github.com/microsoft/pxt-microbit/issues/3914

jwunderl (approved):

PR #8088, opened 2021-04-12
Set sidedocs height instead of padding, add thin editor style

shakao:
- sets sidedoc height instead of padding-bottom, so the editor toolbar buttons are still clickable
- additional height for thin editor toolbar with desktop simulator (sidedocs can be taller here)

fixes https://github.com/microsoft/pxt-microbit/issues/3707
fixes https://github.com/microsoft/pxt-microbit/issues/3014


livcheerful (approved):
🧮✨

PR #8089, opened 2021-04-13
Fix crash on project delete

darzu:
Fixes: https://github.com/microsoft/pxt-microbit/issues/3959
Fixes: https://github.com/microsoft/pxt-arcade/issues/3413

This was a regression introduced by the cloud work. It's not cloud related, just that I changed the toolboxeditor from fetching the header from the props to instead fetch from the virtual APIs but there's a case (during delete) where the header from the virtual API is null.


jwunderl (approved):

abchatra:
Thanks for the quick fix @darzu

PR #8097, opened 2021-04-14
Enabling scrollbar in tutorial card by removing y-overflow and adjust…

Jxwoon:
…ing height

PR #8099, opened 2021-04-15
serve local file in pxt serve again

jwunderl:
When I gave https://github.com/microsoft/pxt/pull/8097 a quick check locally, my local markdown updates weren't going through and I noticed it was fetching everything from the server.

Turns out this weird behavior https://github.com/microsoft/pxt/pull/7989/files#r598994321 of updating as a side effect was the thing preventing it from serving docs from the server instead of the local versions when doing a `pxt serve`, as setting that to true makes us skip the local request path here:

https://github.com/microsoft/pxt/blob/108e3785e2ae20fdeb5b59e4f0956fd83092b4e1/pxtlib/emitter/cloud.ts#L184-L195

Fix that so we don't go to makecode.com/api/* for markdown on pxt serve. (the changes are just guarding with `pxt.BrowserUtils.isLocalHostDev()` in a few places where we don't want to default to fetching from server -- e.g. we still want to get updates when if we're doing incontext translations or if we pass `livelang=de`)

### DIFF
--- a/skillmap/src/App.tsx
+++ b/skillmap/src/App.tsx
@@ -124,7 +124,7 @@ class AppImpl extends React.Component<AppProps, AppState> {
         const targetId = pxt.appTarget.id;
         const pxtBranch = pxt.appTarget.versions.pxtCrowdinBranch;
         const targetBranch = pxt.appTarget.versions.targetCrowdinBranch;
-        if (!pxt.BrowserUtils.isLocalHostDev()) {
+        if (!pxt.BrowserUtils.isLocalHostDev() && !pxt.BrowserUtils.isPxtElectron()) {
             pxt.Util.enableLiveLocalizationUpdates();
         }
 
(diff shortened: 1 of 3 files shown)


skillmap/src/App.tsx
```diff
@@ -124,7 +124,9 @@ class AppImpl extends React.Component<AppProps, AppState> {
         const targetId = pxt.appTarget.id;
         const pxtBranch = pxt.appTarget.versions.pxtCrowdinBranch;
         const targetBranch = pxt.appTarget.versions.targetCrowdinBranch;
-        pxt.Util.enableLiveLocalizationUpdates();
+        if (!pxt.BrowserUtils.isLocalHostDev()) {
```
shakao, line 127: don't need the electron check here?

shakao (approved):

skillmap/src/App.tsx
jwunderl: Not really as we just bounce out to the livesite for skillmap in offline app anyways, but probably should just include it anyways to save having to remember it whenever we do support skillmaps there.

```suggestion
        if (!pxt.BrowserUtils.isLocalHostDev() && !pxt.BrowserUtils.isPxtElectron()) {
```

PR #8093, opened 2021-04-13
Create codeql-analysis.yml

abchatra:
https://docs.github.com/en/code-security/secure-coding/configuring-code-scanning

PR #8100, opened 2021-04-15
Update codeql-analysis.yml

jwunderl:
re: https://github.com/microsoft/pxt/pull/8093#issuecomment-819790820, looks like it's hitting a parser error due to the [ ] in a list being a bit ambiguous https://github.com/microsoft/pxt/actions/runs/750203599/workflow

I believe this should hit the right ones, if it still doesn't work can also do
```
    branches:
      - master
      - stable*
      - v[0-9]
```

Looks like it's parsing it as valid with this change: https://github.com/microsoft/pxt/runs/2348513966

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -13,7 +13,7 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ master, stable*, v[0-9] ]
+    branches: [ 'master', 'stable*', 'v[0-9]' ]
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [ master ]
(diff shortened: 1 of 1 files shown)


eanders-MS:
Should the version match be `v[0-9]*`?

jwunderl:
I suppose that would work -- these aren't regex patterns so it would match `v9a` or `v3jjjjjj`, but probably fine / save some confusion in the future when we eventually hit v10

.github/workflows/codeql-analysis.yml
```diff
@@ -13,7 +13,7 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ master, stable*, v[0-9] ]
+    branches: [ 'master', 'stable*', 'v[0-9]' ]
```
jwunderl, line 16:
```suggestion
    branches: [ 'master', 'stable*', 'v[0-9]*' ]
```

abchatra (approved):

PR #8101, opened 2021-04-15
Add skip the shift of nodes we've already seen if they are adjacent

shakao:
checks if nodes are adjacent siblings (can be directly connected by a vertical line) and does not shift the node in that case. this makes the handling of connections within one depth layer a little better (used in Kiki's collector map)


jwunderl (approved):
LGTM, could we get a before & after screenshot to help visualize though / for future reference?

shakao:
sure! before (this had issues when you tried to add a third node connecting to "B", and also "A" is the required activity and "B" is optional, but this makes it look like B is the suggested first activity):
image

now:
image

riknoll (approved):

PR #8102, opened 2021-04-15
Production website check which inserting the telemetry cookie

abchatra:
Checks if our cloud has patched @id@.
I have also made embed view set the cookie. This should be ok as we look for interactive consent while sending telemetry and that should be low enough. I will remove the check in future if there are too many anamolies.

I have not tested this on share page. Let me know how to test this if you have ideas.
I have also not taken care of offline app which I will do separately. This needs to go tomorrow.


docfiles/apptracking.html
```diff
@@ -1,14 +1,16 @@
 <script type="text/javascript">
     window.loadAppInsights = function (includeCookie, telemetryInitializer) {
```
riknoll, line 2: remove includecookie?

docfiles/tracking.html
```diff
@@ -1,15 +1,17 @@
 <script type="text/javascript" src="/doccdn/pxtweb.js"></script>
 <script type="text/javascript">
     window.loadAppInsights = function (includeCookie) {
```
riknoll, line 3: remove includecookie?

riknoll (approved):
might need to check if any targets override tracking.html or apptracking.html but i don't think they do

docfiles/tracking.html
abchatra: I still want to includeCookie for the case of pxtSkipAnalyticsCookie

docfiles/apptracking.html
```diff
@@ -1,14 +1,16 @@
...
+        //Backend will patch @id@ and the check would fail for string concat
+        var isProduction = includeCookie && ("@id@" !== "@"+"id"+"@");
```
jwunderl, line 4: I'm not certain this works in the backend currently; my understanding is that the backend only runs replacements once / on the docfile, and this one is `@included` so it might not get the replacement on `@id@`. That's one that we've run into with docs / share page changes before. I can point you at how to test it on staging though so probably fine to merge and test at least?

riknoll: you should be able to test it by simply doing an uploadtrg

jwunderl: Oh right, this is the one that is used in `webapp/public` so this one is testable -- I guess this comment applies more to `docfiles/tracking.html` which is used in other docfiles. Not certain if it goes through the same path though

abchatra: I modified the I tried uploadtrg but the cli kept changing this with local values. Probably uploadtrg handles the html modificiations. I am going to debug later. For now I am going merge and bump and test to see it works. 
https://arcade.makecode.com/app/20f5f48d56153ea3ba8ae89a66761254161d37ed-22a00706f2

Debuging this is hard.

PR #8103, opened 2021-04-15
Handle WebUSB sharing between windows/tabs in the service worker

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/3706

Only one MakeCode tab in the browser can use WebUSB at a time without running into trouble. Previously, we used to handle the multiple tab problem by disconnecting whenever you switch tabs (to another MakeCode tab or otherwise). That worked, but it had the downside of cutting off long operations like deploying or logging to serial when it really didn't need to.

This PR moves the WebUSB sharing logic into the ServiceWorker. The ServiceWorker maintains a lock on the WebUSB that the clients must request before they can connect.

Test build: 

https://makecode.microbit.org/app/04420c730ef24e5a7e5ed610f47fab9e36c27f0c-bb44734a6d

Note: you might have to go into the "Application" tab of Chrome and click the "skipWaiting" button to force the service worker to update immediately. For users, this should automatically happen when they refresh.


webapp/src/cmds.ts
```diff
@@ -450,7 +464,128 @@ export function showDisconnectAsync(): Promise<void> {
...
+    const p =  new Promise<void>(resolve => {
+        serviceWorkerSupportedResolver = resolve;
+        sendServiceWorkerMessage({
+            type: "serviceworkerclient",
+            action: "packet-io-supported"
+        });
+    });
+
+    try {
+        await pxt.U.promiseTimeout(1000, p);
```
jwunderl, line 580: Making sure I follow here as it took me a sec -- we're effectively making a deferred above with a timeout here, where `serviceWorkerSupportedResolver` is being kept for `handleServiceWorkerMessage` to resolve w/in 1 second?

webapp/src/serviceworker.ts
```diff
@@ -217,3 +224,184 @@ function initWebappServiceWorker() {
...
+        return Promise.race([ promise, timeoutPromise ])
+            .then(resp => {
+                clearTimeout(ref);
+                statusResolver = undefined
+                return resp;
+            });
+    }
+
+
+    function delay(millis: number): Promise<void> {
```
jwunderl, line 402: Just to confirm, not using the helpers in `pxtlib/util.ts` for things like delay / timeout here because serviceworker file is run separate from the rest / don't want any deps on other files?

jwunderl (approved):
LGTM, few questions / clarifications & the rest are just nits

webapp/src/app.tsx
```diff
@@ -320,18 +320,8 @@ export class ProjectView
         data.invalidate('pkg-git-pr');
         data.invalidate('pkg-git-pages')
 
-        // there is a race with another tab that has a lock on the device
```
pelikhan, line 323: This hack was in place in the case of using webusb in MakeCode.com/multi . Probably worth testing this change in that page too

webapp/src/cmds.ts
```diff
@@ -450,7 +464,128 @@ export function showDisconnectAsync(): Promise<void> {
...
+let serviceWorkerSupported: boolean | undefined = undefined;
+
+async function requestPacketIOLockAsync() {
+    if (hasLock) return;
+    if (pendingPacketIOLockResolver) return Promise.reject("Already waiting for packet lock");
+    const supported = await checkIfServiceWorkerSupportedAsync();
+    if (!supported) return;
+
+    if (navigator?.serviceWorker?.controller) {
+        return new Promise<void>((resolve, reject) => {
```
pelikhan, line 501: Is it good to mix async function and "raw" promises?

pelikhan (approved):
LGTM. Would be nice to have a special logging function that can be turned on with url flag to be able to trace the state of the state machine.
+ ticks at strategic location maybe as well.

pelikhan:
We have webusbdbg=1 in microbit

webapp/src/serviceworker.ts
riknoll: correct!

webapp/src/cmds.ts
riknoll: yep! i'll add a comment to clarify. this is just to handle the case where the new service worker has been downloaded, but hasn't activated yet

webapp/src/cmds.ts
riknoll: I think you have to... promises have to come from somewhere, right?

webapp/src/app.tsx
riknoll: Is there a way to test that locally?

riknoll: @pelikhan what is the expected behavior here? As far as I can tell it works the same as the current site, only one of the frames is able to claim the webusb and you can't switch between them.

pelikhan: i can repro what was the issue. i think you overall change might just make this hack irrevelant anymore. 
Do you handle the case where 2 different microbits are connect to different tabs? This scenari works currently in multi.

riknoll: I don't think that would work, right now its a lock on webUSB as a whole

pelikhan: I guess it's not a core scenario but it's a bit unfortunate.

riknoll: lemme take a look at how hard it would be to support that

PR #8107, opened 2021-04-16
Moving to cdn & doccdn checks

abchatra:
I moved the check to /doccdn/ & /blb/ which for sure gets replaced by backend.

Works in the https://arcade.makecode.com/app/20f5f48d56153ea3ba8ae89a66761254161d37ed-22a00706f2


jwunderl (approved):
Seems fine to me, we can test the docs in staging by adding `HEAD_OVERRIDE: "stable1.4"` to staging's env

PR #8109, opened 2021-04-16
Break on first breakpoint in debugger

riknoll:
Potential fix for https://github.com/microsoft/pxt-microbit/issues/3280

This makes it so that we automatically pause on the first breakpoint when running a program in the debugger. It does *not* pause if you have slow-mo turned on, but I can change that if desired. Also, should it pause if you hit restart?

### DIFF
--- a/pxteditor/editor.ts
+++ b/pxteditor/editor.ts
@@ -79,6 +79,7 @@ namespace pxt.editor {
         }; // ensure that this line is visible when loading the editor
         tracing?: boolean;
         debugging?: boolean;
+        debugFirstRun?: boolean;
         bannerVisible?: boolean;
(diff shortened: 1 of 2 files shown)


jwunderl (approved):

abchatra:
Yes good with slow mo doesn't stop & I am also good with restart not hitting the first pause

abchatra (approved):

riknoll:
Okay, I'll update this with no pause on restart

riknoll:
Okay it now only pauses on the first run when you enable the debugger (even if slow-mo was previously enabled). All subsequent runs in the same debugger session (e.g. triggered by clicking restart/toggling slow-mo) will not pause.

PR #8117, opened 2021-04-23
experiment: simple reshuffling of buttons

pelikhan:
Made "pair device" primary color and move "download again" as secondary.

See diff without whitespace https://github.com/microsoft/pxt/pull/8117/files?w=1 . What code formatter settings is the team using?

image

### DIFF
--- a/webapp/src/cmds.ts
+++ b/webapp/src/cmds.ts
@@ -105,23 +105,24 @@ function showUploadInstructionsAsync(fn: string, url: string, confirmAsync: (opt
         helpUrl,
         className: 'downloaddialog',
         buttons: [
+            downloadAgain && {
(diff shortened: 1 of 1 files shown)


Jaqster:
How hard would it be to put like a big Red arrow overlay to draw attention to where the Download file is?  I feel like that's one of the biggest problems with this dialog... there's a disconnect between what just happened and what it's telling you to do.
Chrome
Edge
Firefox

pelikhan:
It's doable with a bit of CSS trickery. We'll probably want to support this for a subset of browser/versions where we **known** where the files are displayed in the browser shell.

pelikhan:
Unrelated to this change though.

Jaqster (approved):
Let's try this

PR #8122, opened 2021-04-27
Immersive Reader -- More Parsing, better button

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/3371
fixes https://github.com/microsoft/pxt-microbit/issues/4008
fixes https://github.com/microsoft/pxt-microbit/issues/4002

- Change @boardname@ to the actual board name
- Scrub \`\`| |`` annotation (inline buttons?)
- Make the hover of the button dark instead of green lol

### DIFF
--- a/theme/tutorial.less
+++ b/theme/tutorial.less
@@ -493,7 +493,7 @@ code.lang-filterblocks {
 
 #mainmenu .immersive-reader-button:focus,
 #mainmenu .immersive-reader-button:hover {
-    background-color: #096975!important;
+    background-color: #00000059!important;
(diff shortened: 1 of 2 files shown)


jwunderl (approved):
re: board name, there are a bunch more like this that get added all over the place so might be good to give a quick look for other `@blah@` in tutorials? I believe they all end up going through here: https://github.com/microsoft/pxt/blob/master/pxtlib/docsrender.ts#L293 (but some get added as params / etc so easiest to set a breakpoint in that function and check what all exists)

PR #8123, opened 2021-04-28
Add description text to serial placeholder graph

shakao:
changes from design office hours

image

### DIFF
--- a/webapp/src/serial.tsx
+++ b/webapp/src/serial.tsx
@@ -475,7 +475,7 @@ export class Editor extends srceditor.Editor {
                     </div>
                 </div>
                 {this.charts?.length == 0 && <div id="serialPlaceholder" className="ui segment">
-                    <div className="ui bottom left attached no-select label seriallabel">{lf("Values will be logged when the device or simulator receives data")}</div>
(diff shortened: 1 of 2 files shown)


webapp/src/serial.tsx
```diff
@@ -475,7 +475,7 @@ export class Editor extends srceditor.Editor {
                     </div>
                 </div>
                 {this.charts?.length == 0 && <div id="serialPlaceholder" className="ui segment">
-                    <div className="ui orange bottom left attached no-select label seriallabel">0</div>
+                    <div className="ui bottom left attached no-select label seriallabel">{lf("Values will be logged when the device or simulator receives data")}</div>
```
jwunderl, line 478:
```suggestion
                    <div className="ui bottom left attached no-select label seriallabel">{lf("Values will be logged when the {0} sends data", this.isSim ? lf("simulator") : lf("device"))}</div>
```

^^ would that make sense? I think only showing simulator / device makes sense, sends vs receives I'm not certain on but I think of it as the simulator sending / creating the data personally?

jwunderl (approved):

webapp/src/serial.tsx
shakao: sounds good to me!

pelikhan (approved):

PR #8125, opened 2021-05-03
Decompile tutorial snippets separately, hash usedBlocks to individual code snippet

shakao:
- adds a new op to the service to decompile an array of ts snippets (strings) & return an array of xml snippets
- where previously we were caching "usedBlocks" (map of blockId to the number 1) we now also cache "snippetBlocks" (map of code snippet hash, to <map of blockIds to the number 1>). usedBlocks is kept as-is to prevent weirdness with existing caches (in git repos, indexeddb, etc)
- also checked with python

perf-wise, it did make things slightly slower (increased a ~2s load time by about ~0.1s on average, on my machine), but that should only be for the initial load, before the blocks are cached

build: https://arcade.makecode.com/app/779fd95829c1d84fb875251dbb9e2f8dc047d2ca-ea561d6c56

### DIFF
--- a/pxtcompiler/emitter/driver.ts
+++ b/pxtcompiler/emitter/driver.ts
@@ -299,11 +299,40 @@ namespace ts.pxtc {
             includeGreyBlockMessages,
             generateSourceMap: !!opts.ast,
             allowedArgumentTypes: opts.allowedArgumentTypes || ["number", "boolean", "string"],
-            errorOnGreyBlocks: !!opts.errorOnGreyBlocks,
+            errorOnGreyBlocks: !!opts.errorOnGreyBlocks
         };
         const [renameMap, _] = pxtc.decompiler.buildRenameMap(program, file, { declarations: "variables", takenNames: {} })
         const bresp = pxtc.decompiler.decompileToBlocks(blocksInfo, file, decompileOpts, renameMap);
         return bresp;
+    }
+
+    // Decompile an array of code snippets (sourceTexts) to XML strings (blocks)
+    export function decompileSnippets(program: Program, opts: CompileOptions, includeGreyBlockMessages = false) {
+        const apis = getApiInfo(program, opts.jres);
+        const blocksInfo = pxtc.getBlocksInfo(apis, opts.bannedCategories);
+        const renameMap = new pxtc.decompiler.RenameMap([]); // Don't rename for snippets
+
+        const decompileOpts: decompiler.DecompileBlocksOptions = {
+            snippetMode: opts.snippetMode || false,
+            alwaysEmitOnStart: opts.alwaysDecompileOnStart,
+            includeGreyBlockMessages,
+            generateSourceMap: !!opts.ast,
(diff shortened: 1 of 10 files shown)


pxtcompiler/emitter/driver.ts
```diff
@@ -299,13 +299,42 @@ namespace ts.pxtc {
...
+            errorOnGreyBlocks: !!opts.errorOnGreyBlocks
         };
         const [renameMap, _] = pxtc.decompiler.buildRenameMap(program, file, { declarations: "variables", takenNames: {} })
         const bresp = pxtc.decompiler.decompileToBlocks(blocksInfo, file, decompileOpts, renameMap);
         return bresp;
     }
 
+    // Decompile an array of code snippets (sourceTexts) to XML strings (blocks)
+    export function decompileSnippets(program: Program, opts: CompileOptions, includeGreyBlockMessages = false) {
+        const apis = getApiInfo(program, opts.jres);
```
riknoll, line 311: Hmmm... I wonder if we can use lastApiInfo here (and above). Doesn't need to happen in this pr

pxtcompiler/emitter/driver.ts
```diff
@@ -299,13 +299,42 @@ namespace ts.pxtc {
...
+            allowedArgumentTypes: opts.allowedArgumentTypes || ["number", "boolean", "string"],
+            errorOnGreyBlocks: !!opts.errorOnGreyBlocks
+        };
+
+        const xml: string[] = [];
+        opts.sourceTexts?.forEach((text: string) => {
+            opts.fileSystem[pxt.MAIN_TS] = text;
+            opts.fileSystem[pxt.MAIN_BLOCKS] = "";
+
+            program = getTSProgram(opts);
```
riknoll, line 329: You should pass in the old program here (optional second argument). It should speed up the TS language service generating the new program.

webapp/src/tutorial.tsx
```diff
@@ -38,7 +43,9 @@ export function getUsedBlocksAsync(code: string[], id: string, language?: string
             .then(entry => {
                 if (entry?.blocks && Object.keys(entry.blocks).length > 0 && !skipCache) {
                     pxt.tickEvent(`tutorial.usedblocks.indexeddb`, { tutorial: id });
-                    return Promise.resolve(entry.blocks);
+                    // populate snippets if usedBlocks are present, but snippets are not
+                    if (!entry?.snippets) getUsedBlocksInternalAsync(code, id, language, db, skipCache);
```
riknoll, line 47: should this be returned?

riknoll (changes requested):

webapp/src/tutorial.tsx
shakao: nope, it only needs the usedblocks for the tutorial filter. this just populates the snippets so they're cached next time. (snippetblocks aren't used anywhere yet, they'll be pulled into jasmine's stuff later)

livcheerful (approved):

riknoll (approved):

jwunderl (approved):

PR #8131, opened 2021-05-05
actually restart simulator when hc setting changes

jwunderl:
Fix https://github.com/microsoft/pxt-microbit/issues/3126. We were trying to restart here already but it was getting ignored in multiple places -- the restart always hits the 'fast restart' path which doesn't update the iframe options, and even with that fixed the start would ignore it as the simulator was already running.

We also don't want to use restart here because it steals focus to the simulator, which messes with tab accessibility a bit.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -3048,7 +3048,10 @@ export class ProjectView
 
     onHighContrastChanged() {
         this.clearSerial();
-        this.restartSimulator();
+        // Not this.restartSimulator; need full restart to consistently update visuals,
+        // and don't want to steal focus.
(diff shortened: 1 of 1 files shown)


shakao:
hmm weird, i think this bug was filed before that HC function was added (for identity)? but maybe we were restarting somewhere else, haha

shakao (approved):

livcheerful (approved):

jwunderl:
Ah yeah, this is a longstanding bug, not caused by identity change (at least goes back to microbit/v2, worked in microbit/v1) -- the identity change just moved the attempt at restarting here

PR #8133, opened 2021-05-07
Persist Immersive Reader preferences

livcheerful:
fixes https://github.com/microsoft/pxt-microbit/issues/4005

Use the nice auth APIs to store the reader preferences, which will use local storage (not idb) if not logged in, and will sync w/ cloud if you are :D it works!

Also snuck in the user language preferences, so the UI in Immersive Reader is the one you selected in Arcade! wow

image

oh also i found some logs i left before so i took those out sorry

### DIFF
--- a/webapp/src/auth.ts
+++ b/webapp/src/auth.ts
@@ -56,6 +56,7 @@ export type UserPreferences = {
 const DEFAULT_USER_PREFERENCES: () => UserPreferences = () => ({
     highContrast: false,
     language: pxt.appTarget.appTheme.defaultLocale,
+    reader: ""
 })
 
 /**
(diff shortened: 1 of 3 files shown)


webapp/src/auth.ts
```diff
@@ -48,6 +50,7 @@ export type UserProfile = {
 export type UserPreferences = {
     language?: string;
     highContrast?: boolean;
+    reader?: string;
 }
 
 const DEFAULT_USER_PREFERENCES: () => UserPreferences = () => ({
```
eanders-MS, line 56: Is it worth adding a default value here for the reader? Maybe in empty string?

eanders-MS (approved):
LGTM!

webapp/src/auth.ts
livcheerful: hahaha yeah i'll do that!

PR #8140, opened 2021-05-11
Adding audio context function for streaming buffers for sound emoji

riknoll:
Adding simulator support for streaming audio buffers for the micro:bit sound emoji.

Tested in Firefox, Safari, and Chrome on macOS

### DIFF
--- a/pxtsim/simlib.ts
+++ b/pxtsim/simlib.ts
@@ -645,6 +645,74 @@ namespace pxsim {
             })
         }
 
+        const MAX_SCHEDULED_BUFFER_NODES = 3;
+
+        export function playPCMBufferStreamAsync(pull: () => Float32Array, sampleRate: number, volume = 0.3) {
+            return new Promise<void>(resolve => {
+                let nodes: AudioBufferSourceNode[] = [];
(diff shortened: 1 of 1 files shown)


pelikhan (approved):

abchatra:
@jaustin fyi

pxtsim/simlib.ts
```diff
@@ -645,6 +645,74 @@ namespace pxsim {
             })
         }
 
+        const MAX_SCHEDULED_BUFFER_NODES = 3;
+
+        export function playPCMBufferStreamAsync(pull: () => Float32Array, sampleRate: number, volume = 0.3) {
```
abchatra, line 650: Can you please add the reference to CODAL code this is trying to mirro?

abchatra (approved):

pelikhan:
Should we also repipe this function to use your PCM api?

https://github.com/microsoft/pxt/blob/d0ec5d92f120492142330e264cff57ba5d5cb364/pxtsim/simlib.ts#L629

riknoll:
@pelikhan no, I don't think it's necessary. Works well enough for the .wav files

PR #8146, opened 2021-05-12
Fix local host serve when pointing to production sim

riknoll:
This fixes local serve in arcade when using the production simulator. It was a regression accidentally caused by #7720

If the simulator is missing, our CLI [redirects requests to the production simulator](https://github.com/microsoft/pxt/blob/master/cli/server.ts#L1136) which ends up changing the domain of the iframe. Because the origin changes with the redirect, the messages we send with `postMessage` never reach the content window.

This fixes that by sending to both domains when on localhost. Should be harmless since one of the messages will just always be ignored. I don't love how I got access to the target id (apptarget isn't usually visible in pxsim), but piping everything from the webapp for local serve seemed messy.

This really only effects arcade

### DIFF
--- a/pxtsim/simdriver.ts
+++ b/pxtsim/simdriver.ts
@@ -393,7 +393,7 @@ namespace pxsim {
         private postMessageCore(frame: HTMLIFrameElement, msg: SimulatorMessage) {
             frame.contentWindow.postMessage(msg, frame.dataset['origin']);
 
-            if (U.isLocalHost() && (pxt as any)?.appTarget?.id) {
+            if (U.isLocalHostDev() && (pxt as any)?.appTarget?.id) {
                 // If using the production simulator on local serve, the domain might have been
                 // redirected by the CLI server. Also send to the production domain just in case
                 try {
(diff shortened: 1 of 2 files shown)


pxtsim/simdriver.ts
```diff
@@ -390,6 +390,21 @@ namespace pxsim {
             }
         }
 
+        private postMessageCore(frame: HTMLIFrameElement, msg: SimulatorMessage) {
+            frame.contentWindow.postMessage(msg, frame.dataset['origin']);
+
+            if (U.isLocalHost() && (pxt as any)?.appTarget?.id) {
```
jwunderl, line 396:
```suggestion
            if (U.isLocalHostDev() && (pxt as any)?.appTarget?.id) {
```

to filter out electron

jwunderl (approved):

pelikhan:
Oops. Do we have support to force using the production simulator?

pelikhan (approved):

pxtsim/simdriver.ts
riknoll: good call, but i need to port over that function into pxtsim

riknoll:
@pelikhan not that I'm aware, but we totally should! This was a pain to test

livcheerful (approved):

PR #8159, opened 2021-05-17
Added Markdown Flag for Code Validation

Jxwoon:
In a markdown file, this flag can be added:
`### @tutorialCodeValidation true/false`

If this flag is set to true, then the tutorial will go through the process of tutorial code validation. Otherwise features for tutorial code validation will not be included when a user is going through a tutorial.


shakao (changes requested):
please add it to here as well! https://github.com/microsoft/pxt/blob/master/pxtlib/docsrender.ts#L20 all tutorials can also be rendered from our docs (eg https://arcade.makecode.com/tutorials/barrel-dodger) and this replaces the `### @flag` with a markdown comment so that it's hidden

shakao (approved):

pelikhan:
Can you add a description of your pull request? (For future review)

PR #8162, opened 2021-05-17
Add ?compiler=size switch

mmoskal:
This generates `built/sizes.csv` that can be visualized like so (with Excel):

image


pelikhan (approved):

eanders-MS (approved):
Nice! eventContext is bigger than I expected.

mmoskal:
note that this is picture for `game---light`, so the whole code is like 10k not 90k as it is with normal `game`. So eventContext is not that big ;)

Here's a picture for a normal game.

image

pelikhan:
eventually, it would be nice to run this tool in the build and track the size of the runtime. (eg. fail the build if we increase more than xx)

PR #8166, opened 2021-05-18
Fix spacing in skillmap files

shakao:
missed some lint things! not sure why the build passed!


livcheerful (approved):

jwunderl (approved):
I think lint was disabled for skillmap when we upgraded from tslint to eslint, there should be a todo somewhere hah

PR #8170, opened 2021-05-19
Translate Python snippets

livcheerful:
fixes https://github.com/microsoft/pxt-microbit/issues/4007

Similar to https://github.com/microsoft/pxt/pull/8076 except there wasn't a translation path for enums. Added a case to manually update enums, and also added the same function to webapp so the snippets will display correctly in the toolbox (for example, true **should** be capitalized here 
image )

I couldn't find a good way to share code between pxtcompiler/webapp so I copied it, but if there's a better way please let me know 😭

### DIFF
--- a/pxtcompiler/emitter/languageservice.ts
+++ b/pxtcompiler/emitter/languageservice.ts
@@ -92,7 +92,7 @@ namespace ts.pxtc.service {
             const dotIdx = param.lastIndexOf(".");
             const left = param.substr(0, dotIdx)
             let right = param.substr(dotIdx + 1)
-            right = snakify(right).toUpperCase();
+            right = U.snakify(right).toUpperCase();
             return `${left}.${right}`
         }
         return undefined;
@@ -285,8 +285,6 @@ namespace ts.pxtc.service {
         }
         return null
     }
-
-
 
     export function getDefaultEnumValue(t: Type, python: boolean): string {
         // Note: AFAIK this is NOT guranteed to get the same default as you get in
--- a/pxtcompiler/emitter/languageservice.ts
+++ b/pxtcompiler/emitter/languageservice.ts
@@ -76,13 +76,22 @@ namespace ts.pxtc.service {
         return [...retApis, ...completionSymbols(enumVals, COMPLETION_DEFAULT_WEIGHT)]
     }
 
-    export function tsKeywordToPyKeyword(keyword: string): string | undefined {
-        const map: pxt.Map<string> = {
+    export function tsSnippetToPySnippet(param: string): string | undefined {
+        // Keep this unified with webapp/monacoSnippets.ts
+        const keywords: pxt.Map<string> = {
             "true": "True",
             "false": "False",
             "null": "None"
         }
-        return map[keyword]
+        const key = keywords[param];
+        if (key) {
+            return key
+        }
+        if (param.includes(".")) {
+            const match = /(.+)\.(.+)/.exec(param);
+            return `${match[1]}.${match[2].toUpperCase()}`
+        }
+        return undefined;
     }
 
     export function getBasicKindDefault(kind: SyntaxKind, isPython: boolean): string | undefined {
--- a/pxtcompiler/emitter/languageservice.ts
+++ b/pxtcompiler/emitter/languageservice.ts
@@ -88,6 +88,7 @@ namespace ts.pxtc.service {
             return key
         }
         if (param.includes(".")) {
+            // Python enums are all caps
             const match = /(.+)\.(.+)/.exec(param);
             return `${match[1]}.${match[2].toUpperCase()}`
         }
--- a/pxtcompiler/emitter/languageservice.ts
+++ b/pxtcompiler/emitter/languageservice.ts
@@ -76,28 +76,6 @@ namespace ts.pxtc.service {
         return [...retApis, ...completionSymbols(enumVals, COMPLETION_DEFAULT_WEIGHT)]
     }
 
-    export function tsSnippetToPySnippet(param: string): string | undefined {
-        // Keep this unified with webapp/monacoSnippets.ts
-        const keywords: pxt.Map<string> = {
-            "true": "True",
-            "false": "False",
-            "null": "None"
-        }
-        const key = keywords[param];
-        if (key) {
-            return key
-        }
-        if (param.includes(".")) {
-            // Python enums are all caps
-            const dotIdx = param.lastIndexOf(".");
-            const left = param.substr(0, dotIdx)
-            let right = param.substr(dotIdx + 1)
-            right = U.snakify(right).toUpperCase();
-            return `${left}.${right}`
-        }
-        return undefined;
-    }
-
     export function getBasicKindDefault(kind: SyntaxKind, isPython: boolean): string | undefined {
         switch (kind) {
             case SK.StringKeyword: return "\"\"";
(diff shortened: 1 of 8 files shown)


webapp/src/monacoSnippets.ts
```diff
@@ -617,6 +617,25 @@ export function blockIdMap() {
...
+        "false": "False",
+        "null": "None"
+    }
+    const key = keywords[param];
+    if (key) {
+        return key
+    }
+    if (param.includes(".")) {
+        // Python enums are all caps
+        const match = /(.+)\.(.+)/.exec(param);
```
jwunderl, line 633:
```suggestion
        const match = /(.+)\.([^.]+)$/.exec(param);
```

maybe to only grab the final? we have some enums (e.g. `sprites.Flag.Ghost`) in namespaces but I feel like I remember them having trouble with blocks so maybe it's not actually an issue (in this case we have a trimmed version of the enum at the top level anyways)

jwunderl (commented):
Will this hit fixedinstance params like `mySprite.startEffect(effects.spray)`?

image


also one suggestion \\/

webapp/src/monacoSnippets.ts
livcheerful: ooooh yeah I'll do that and test

livcheerful:
This isn't playing nicely with the sprite flags in Arcade (since underscores are getting added there)-- closing for now

livcheerful:
> Will this hit fixedinstance params like `mySprite.startEffect(effects.spray)`?
> 
> image
> 
> also one suggestion \/

Right now no blocks have a default  fixed instance ever explicitly defined I think so this code won't affect it, but if we add/change a block, it will display as all caps with this change in the toolbox, but when dragged out it will have the right code

pxtcompiler/emitter/snippets.ts
```diff
@@ -311,9 +311,13 @@ namespace ts.pxtc.service {
                 if (typeNode.kind === SK.StringKeyword || deflKind === SK.StringKeyword) {
                     return paramDefl.indexOf(`"`) != 0 ? `"${paramDefl}"` : paramDefl;
                 }
-
+                let type = checker && checker.getTypeAtLocation(param);
```
jwunderl, line 314:
```suggestion
                const type = checker?.getTypeAtLocation(param);
```

nit

pxtcompiler/emitter/languageservice.ts
```diff
@@ -76,13 +76,26 @@ namespace ts.pxtc.service {
         return [...retApis, ...completionSymbols(enumVals, COMPLETION_DEFAULT_WEIGHT)]
     }
 
-    export function tsKeywordToPyKeyword(keyword: string): string | undefined {
-        const map: pxt.Map<string> = {
+    export function tsSnippetToPySnippet(param: string): string | undefined {
```
jwunderl, line 79: I guess this could move to pxtlib/util.ts too so it doesn't need to be duped?

pxtcompiler/emitter/snippets.ts
```diff
@@ -311,9 +311,13 @@ namespace ts.pxtc.service {
...
+                const typeSymbol = getPxtSymbolFromTsSymbol(type.symbol, apis, checker);
+                if (typeSymbol && typeSymbol.attributes.fixedInstances) {
```
jwunderl, line 316:
```suggestion
                if (typeSymbol?.attributes.fixedInstances) {
```

pxtcompiler/emitter/snippets.ts
```diff
@@ -311,9 +311,13 @@ namespace ts.pxtc.service {
                 if (typeNode.kind === SK.StringKeyword || deflKind === SK.StringKeyword) {
                     return paramDefl.indexOf(`"`) != 0 ? `"${paramDefl}"` : paramDefl;
                 }
-
+                let type = checker && checker.getTypeAtLocation(param);
+                const typeSymbol = getPxtSymbolFromTsSymbol(type.symbol, apis, checker);
```
jwunderl, line 315: The line above would indicate that checker could be undefined, which would cause it to get a null deref here

```suggestion
                const typeSymbol = getPxtSymbolFromTsSymbol(type?.symbol, apis, checker);
```

webapp/src/monacoSnippets.ts
```diff
@@ -617,6 +617,28 @@ export function blockIdMap() {
...
+    }
+    if (param.includes(".")) {
+        // Python enums are all caps
+        const dotIdx = param.lastIndexOf(".");
+        const left = param.substr(0, dotIdx)
+        let right = param.substr(dotIdx + 1)
+        right = pxt.Util.snakify(right).toUpperCase();
+        return `${left}.${right}`
+    }
+    return undefined;
```
jwunderl, line 639: Might be nice to just return param here as the default case, since it seems like everywhere it is called it ends up getting either `| inputted` in some way or another anyways. Up to you though~

jwunderl (approved):
LGTM

pxtcompiler/emitter/snippets.ts
```diff
@@ -311,11 +311,13 @@ namespace ts.pxtc.service {
                 if (typeNode.kind === SK.StringKeyword || deflKind === SK.StringKeyword) {
                     return paramDefl.indexOf(`"`) != 0 ? `"${paramDefl}"` : paramDefl;
                 }
-
+                const type = checker?.getTypeAtLocation(param);
+                const typeSymbol = getPxtSymbolFromTsSymbol(type?.symbol, apis, checker);
+                if (typeSymbol?.attributes.fixedInstances) {
+                    return paramDefl;
```
riknoll, line 317: Do you need to check if we're in python here? Does paramDefl work for both languages?

livcheerful: oh for some reason I thought fixedInstances meant they had the same name in py/ts. I'll just remove this part

riknoll: fixedInstances are when we export a bunch of instances and use them to populate a dropdown. For example:

https://github.com/microsoft/pxt-common-packages/blob/master/libs/game/controlleroverrides.ts

riknoll: It's sorta like an enum, but it compiles differently. Those are for the controller blocks in arcade

livcheerful: okay hmm looks like it's just a snakify? I will add

pxtlib/service.ts
```diff
@@ -634,6 +634,29 @@ namespace ts.pxtc {
...
+        const keywords: pxt.Map<string> = {
+            "true": "True",
+            "false": "False",
+            "null": "None"
+        }
+        const key = keywords[param];
+        if (key) {
+            return key
+        }
+        if (param.includes(".")) {
```
riknoll, line 648: iirc the symbol info should have an array for all of the parameters and the entry should have an "isEnum" propeerty that you could check. This is probably fine, but it could get a little weird in some cases (e.g. if the default is a string with a period in it or something like ``assets.tilemap`whatever` ``)

riknoll (approved):

pxtlib/service.ts
livcheerful: ! okay I made a change to use the symbol info if it's available to check for enum. it'll be more accurate when dragging out the snippet than in the toolbox, but I think that's okay?

PR #8171, opened 2021-05-19
Updates for hero banner gallery

jwunderl:
Updates to hero banner per design meetings

image

here's a build: https://makecode.microbit.org/app/5815523cda0fd9ce28690b493d7346eae12648ea-7b4356e8ad (updated to https://github.com/microsoft/pxt/pull/8171/commits/68bd0c128c2831a1c4b617a5850de8bff5e3e888)

to make the default hero banner show up with actions also need to add https://github.com/microsoft/pxt-microbit/pull/4077

Two other things I'll probably look at today -- supporting swipe to skip / go back, and possibly changing the transition between images from default browser crossfade to an opacity fade (but that can be surprisingly tedious to deal with / might not be worth the time right now.)

One implementation note; the reason for leaving the homeScreenHero in and not just moving complete to a gallery is because the homeScreenHero is defined in pxtarget, so it will be present immediately on load and show up in first sweep -- the gallery is loaded asynchronously so it will populate *after* the first load. If we were to define only a gallery and no homescreenhero banner, it would take a moment to load and then pop into view / move the contents of the screen around.


webapp/src/projects.tsx
```diff
@@ -410,26 +410,28 @@ export class ProjectsMenu extends data.Component<ISettingsProps, {}> {
 }
 
 interface HeroBannerState {
-    cardIndex?: number;
+    cardIndex: number;
     paused?: boolean;
 }
 
-const HERO_BANNER_DELAY = 10000; // 10 seconds per card
+// todo make it not so speedy before putting up pr
```
shakao, line 417: leftover comment?

jwunderl: Oops yes thanks

webapp/src/projects.tsx
```diff
@@ -489,43 +515,57 @@ class HeroBanner extends data.Component<ISettingsProps, HeroBannerState> {
...
+        const card = cards[cardIndex];
+        if (!card)
+            return null; // nothing to see here
+
         const handleSetCard = (i: number) => () => this.handleSetCardIndex(i)
         let url = card.url;
         // open tutorials in browser
-        if (card.cardType === "tutorial" && /^\//.test(url))
-            url = `#tutorial:${url}`
+        if (card.cardType === "tutorial" && /^\//.test(url)) {
```
shakao, line 539: are we only handling tutorials for now? will this break if someone puts an example card in the gallery?

theme/home.less
```diff
@@ -54,26 +54,53 @@
...
             }
         }
+
+        .gradient-overlay {
+            position: relative;
+            height: 100%;
+            width: 40rem;
+            top: 0;
+            left: 0;
+            background: linear-gradient(to right, rgba(0, 0, 0, 1) 0%, rgba(0, 0, 0, 0) 100%);
```
shakao, line 100: small: should the color we're fading to be a less variable/configurable per theme?

theme/home.less
```diff
@@ -54,26 +54,53 @@
...
+            font-weight: 600;
+            color: @white;
+        }
+
+        .action {
+            margin: 1rem 0;
+        }
+
+        .dots button {
+            border: 2px solid #6D6D6D;
```
shakao, line 85: small: #000 at partial opacity instead of grey?

shakao:
if we can ensure the images are all the same size then the crossfade thing is a nonissue also!

webapp/src/projects.tsx
jwunderl: It'll point it at the url in there -- so we can point at anything from the md (this will point at a 'hero' gallery or something of the sort that won't be a normal gallery).

You're right though that should probably just take the same behavior from the detail view, I'll see how much work it is to pull that out

jwunderl: Okay, pulled out the logic for the detail buttons in https://github.com/microsoft/pxt/pull/8171/commits/192698ad9ed1732216c31e493b0f0cc8d942b151 so should work with all cardTypes -- not the prettiest but that path is a bit messy already / probably put off cleaning it up for a later PR?

livcheerful:
small thing that doesn't have to be this PR, it looks like when the description is 2 lines and the screen is short, it looks a little cramped

image

jwunderl:
Ooh good catch thanks, I missed that there was a 'very short wide' setting for banner too. Fixed by swapping it to mobile view:

image

livcheerful (approved):

PR #8177, opened 2021-05-21
Tutorial Validation Extracts Rules from Markdown

Jxwoon:
Tutorial Code validation now extracts rules from a markdown file in the format:

```tutorialValidationRules
{"rule1":true, "rule2":false}
```
"rule1" corresponds to the name of the rule, and the boolean after the colon indicates if the rule is turned on/off. Associated with each rule is a boolean status flag that indicates if the user passes this rule and a rule message.

### DIFF
--- a/pxtlib/tutorial.ts
+++ b/pxtlib/tutorial.ts
@@ -10,7 +10,10 @@ namespace pxt.tutorial {
             return undefined; // error parsing steps
 
         // collect code and infer editor
-        const { code, templateCode, editor, language, jres, assetJson, customTs } = computeBodyMetadata(body);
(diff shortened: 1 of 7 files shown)


livcheerful (approved):

pxtlib/tutorial.ts
```diff
@@ -282,6 +290,23 @@ ${code}
         return { header, hint };
     }
 
+    /* Parse Tutorial Validation Rules */
+    function parseTutorialValidationRules(body: string, tutorialCodeValidation?: boolean): pxt.Map<boolean> {
```
shakao, line 294: (late comment/this can be in a separate PR) i think this should use JSON.parse instead of manually parsing (see https://github.com/microsoft/pxt/blob/master/pxtlib/util.ts#L276)

PR #8181, opened 2021-05-22
New download dialog/webusb flow

riknoll:
Implementing the new download dialog/web usb pairing flow. Here is a test URL:

https://makecode.microbit.org/app/981a204b8eab269ab2f58bc2993534add19e6162-440b8f8359

Note that the target specific icon is checked into pxt-core (the micro:bit logo) because we don't really have a way for targets to load their own icons right now. I figure it's not that big a deal. All of the images, icons, and help urls are controlled by a theme set in the target so all of this should also work for pxt-arcade. Targets can also still implement their own dialogs, this just improves the default.


jwunderl (approved):

martinwork:
@riknoll Please see comment https://github.com/microsoft/pxt-microbit/issues/3299#issuecomment-846398316

webapp/src/dialogs.tsx
```diff
@@ -714,3 +717,74 @@ export function promptTranslateBlock(blockid: string, blockTranslationIds: strin
...
+                    inputLabel={lf("Don't show this again")}
+                    onChange={onCheckboxClicked}
+                />
+            </div>
+        </div>
+    </div>;
+}
+
+export function isDontShowDownloadDialogCookieSet() {
+    return pxt.BrowserUtils.getCookie(DONT_SHOW_COOKIE_ID) === "true";
```
abchatra, line 789: Can we use localStorage instead of cookie. This doesn't need to be cookie in my opionion.

abchatra (approved):

abchatra:
> @riknoll Please see comment [microsoft/pxt-microbit#3299 (comment)](https://github.com/microsoft/pxt-microbit/issues/3299#issuecomment-846398316)

Yes probably we can fix this in a separate issue.

webapp/src/dialogs.tsx
abchatra: I mearging this PR now. But we should do this seperately.

pelikhan: So what happens on a shared computer when a previous user has dismissed the dialogue?

It feels like this setting should be in memory or a user-settings is signed in.

PR #8185, opened 2021-05-24
Custom Pop-Up Messages and Toggle Pop-Up Message

Jxwoon:
Few things in this PR:

Added checks for when promised variables in the first tutorial step are undefined. 

When the user has the pop-up message opened and clicks the hint, the pop-up message will close. 

disappearPopUpWhenHintPressed

For each validation rule, there is a custom message. For this rule, validate the number of blocks, the custom message lists the blocks. 

customMissingBlockMessage


webapp/src/tutorial.tsx
```diff
@@ -697,7 +700,7 @@ export class TutorialCard extends data.Component<TutorialCardProps, TutorialCard
                 </div>
                 {hasNext ? <sui.Button icon={`${isRtl ? 'left' : 'right'} chevron large`} className={`nextbutton right attached ${!hasNext ? 'disabled' : ''}  ${tutorialCodeValidated ? 'isValidated' : ''}`} text={lf("Next")} textClass="widedesktop only" ariaLabel={lf("Go to the next step of the tutorial.")}
                     onClick={nextOnClick} onKeyDown={sui.fireClickOnEnter} /> : undefined}
-                {showMissingBlockPopupMessage && <TutorialCodeValidation.MoveOn onYesButtonClick={this.nextTutorialStep} onNoButtonClick={this.showUnusedBlocksMessage} initialVisible={this.state.showUnusedBlockMessage} isTutorialCodeInvalid={!tutorialCodeValidated} parent={this.props.parent} />}
+                {showMissingBlockPopupMessage && <TutorialCodeValidation.MoveOn onYesButtonClick={this.nextTutorialStep} onNoButtonClick={this.showUnusedBlocksMessage} initialVisible={this.state.showUnusedBlockMessage} isTutorialCodeInvalid={!tutorialCodeValidated} ruleComponents={stepInfo.codeValidated} parent={this.props.parent} />}
```
shakao, line 703: (can be in a later pr) maybe "stepInfo.codeValidated" should be renamed; right now it looks like a boolean instead of a list which is a bit confusing for anyone reading through the code

webapp/src/tutorialCodeValidation.tsx
```diff
@@ -38,15 +39,25 @@ export class MoveOn extends data.Component<TutorialCodeValidationProps, tutorial
...
         const vis = this.props.isTutorialCodeInvalid;
+        const message = this.compileMessage(this.props.ruleComponents);
         return <div>
             <div className={`tutorialCodeValidation no-select ${!vis ? 'hidden' : ''}`}>
                 <div className="codeValidationPopUpText">
-                    {lf("These are the blocks you seem to be missing.")}
-                </div>
-                <div className="codeValidationPopUpText">
-                    { /* Will add a list a blocks here in future PR */}
+                    {lf(message)}
```
shakao, line 60: the "lf" should be where you are generating the string (in tutorialValidator.ts), this will treat each combination of rule strings as unique

webapp/src/tutorialCodeValidation.tsx
```diff
@@ -38,15 +39,25 @@ export class MoveOn extends data.Component<TutorialCodeValidationProps, tutorial
...
+                    s = s.concat(s, rules[i].RuleMessage);
+                }
+            }
+        }
+        return s;
+    }
+
     renderCore() {
         const vis = this.props.isTutorialCodeInvalid;
+        const message = this.compileMessage(this.props.ruleComponents);
```
shakao, line 56: you can iterate over the messages inside the react component using `.map` (see https://github.com/microsoft/pxt/blob/master/webapp/src/errorList.tsx#L150); concatenating the strings means there won't be line breaks between messages, and you won't be able to adjust the styles/spacing

pxtlib/tutorialValidator.ts
```diff
@@ -123,23 +125,37 @@ namespace pxt.tutorial {
...
         }
         for (let i: number = 0; i < tutorialBlockKeys.length; i++) {
             let tutorialBlockKey = tutorialBlockKeys[i];
             if (!usersBlockUsed[tutorialBlockKey]) { // user did not use a specific block
-                return false;
-            }
-            if (usersBlockUsed[tutorialBlockKey] < tutorialBlockUsed[tutorialBlockKey]) { // user did not use enough of a certain block
-                return false;
+                sArr[arrayIndex] = "- " + tutorialBlockKey;
+                arrayIndex++;
```
shakao, line 148: is there a reason to manually increment the index instead of using `sArr.push(item)` ?

pxtlib/tutorialValidator.ts
```diff
@@ -123,23 +125,37 @@ namespace pxt.tutorial {
...
+                isValid = false;
+            } else if (usersBlockUsed[tutorialBlockKey] < tutorialBlockUsed[tutorialBlockKey]) { // user did not use enough of a certain block
```
shakao, line 150: could this check be combined into the previous `if`?

shakao (approved):

jwunderl (approved):

PR #8191, opened 2021-05-25
Update monaco to 0.24.0

livcheerful:
fixes https://github.com/microsoft/pxt-microbit/issues/3987

I wasn't sure if I was supposed to add all the public/vs* files, but it looked like they were recently updated so I did. Only change is updating monaco-editor from 0.23.0 -> 0.24.0 in package.json

And an uploadtrg to try here!
https://makecode.microbit.org/app/b163c36f033854ba751c5d0ae4da16119ab322fa-447c9e7176


jwunderl (approved):
LGTM -- I'm also not certain I have context on the checked in monaco files but it looks expected that they'd change (e.g. workerMain showing version+hash) -- maybe for /playground

riknoll:
This fixes ios 13.7!  I can't test windows app...

shakao (approved):

livcheerful:
> This fixes ios 13.7! I can't test windows app...

I'll try the windows app!

jwunderl:
Ah so it looks like this breaks all monaco click behavior on safari (besides one that richard caught, which is 'click just above line number to select whole line'. :(

jwunderl:
^^ looks like that was https://github.com/microsoft/pxt/pull/8189 slipping into the uploaded build I tested so everything looks good to merge now~

livcheerful:
Hm it looks like it might be hard to test windows app with local build, so since this seems to work okay for everything else I'll just merge it and test with beta. I'll close https://github.com/microsoft/pxt-microbit/issues/4091 if this fixes it

PR #8201, opened 2021-05-27
Extend monaco-typescript worker

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/3978

Extend the TypeScriptWorker from monaco-typescript to bring back the feature of filtering out symbols that start with an underscore. This requires a minor change to the backend because of the new entry in the WebConfig, so making it a draft for now. That being said merging it should be safe (it just won't do anything).


jwunderl (approved):

livcheerful (approved):

abchatra (approved):

abchatra:
Is this ready to be merged?

PR #8208, opened 2021-05-28
Fixed Next Button Color When Code Is Validated

Jxwoon:
See title - made changes from 5/25 Design Office Hours
Note: These colors are for Arcade 

Screenshot (33)

### DIFF
--- a/theme/themes/pxt/globals/site.variables
+++ b/theme/themes/pxt/globals/site.variables
@@ -394,5 +394,3 @@
 ----------------------------*/
 @tutorialCodeValidationCorrectBackground: @green;
 @tutorialCodeValidationCorrectTextColor: @white;
-@tutorialCodeValidationCorrectBackgroundMicrobit: #cd0365;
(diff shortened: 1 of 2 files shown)


theme/themes/pxt/globals/site.variables
```diff
@@ -392,4 +392,7 @@
 /*---------------------------
    Tutorial Code Validation
 ----------------------------*/
-@tutorialCodeValidationMoveOn: @green;
+@tutorialCodeValidationCorrectBackground: @green;
+@tutorialCodeValidationCorrectTextColor: @white;
+@tutorialCodeValidationCorrectBackgroundMicrobit: #cd0365;
+@tutorialCodeValidationCorrectTextColorMicrobit: @white;
```
jwunderl, line 398: we shouldn't need two variables for this; should just override the two above in target

jwunderl (changes requested):

PR #8209, opened 2021-05-28
Added a Loose Rule and Fixed UI for Validation

Jxwoon:
### New Loose Rule: 
Checks for at least one block type for each tutorial step. In Chase The Pizza, two sprite blocks are in the hint, but this rule allows the user to move on despite having one sprite block.

Screenshot (35)

------

### UI Behavior when user clicks next button and code is incorrect

Only Strict Rules Turned on: 
A pop-up message will appear to inform the user how to correct their code.

strictRuleEnable

Only Loose Rules Turned On:
No pop-up message will appear when the code is incorrect, and users are allowed to move on to the next step.

LooseRuleBehavior

Strict and Loose Rules Turned On:
This behavior will follow "Only Strict Rules Turned On"  behavior.

StrictAndLooseRuleBehavior

### DIFF
--- a/pxtlib/tutorialValidator.ts
+++ b/pxtlib/tutorialValidator.ts
@@ -1,11 +1,11 @@
 namespace pxt.tutorial {
 
     export interface TutorialRuleStatus {
-        RuleName: string;
-        RuleTurnOn: boolean;
-        RuleStatus: boolean;
-        RuleMessage: string;
-        isStrict: boolean;
+        ruleName: string;
+        ruleTurnOn: boolean;
+        ruleStatus?: boolean;
+        ruleMessage?: string;
+        isStrict?: boolean;
     }
 
     /**
@@ -32,7 +32,7 @@ namespace pxt.tutorial {
             const tutorialBlockUsed = extractBlockSnippet(tutorial, indexdb);
             for (let i = 0; i < TutorialRuleStatuses.length; i++) {
                 let currRuleToValidate = TutorialRuleStatuses[i];
-                const ruleName = TutorialRuleStatuses[i].RuleName;
+                const ruleName = TutorialRuleStatuses[i].ruleName;
                 switch (ruleName) {
                     case "validateExactNumberOfBlocks":
                         currRuleToValidate = validateExactNumberOfBlocks(usersBlockUsed, tutorialBlockUsed, currRuleToValidate);
@@ -61,7 +61,7 @@ namespace pxt.tutorial {
         for (let i = 0; i < ruleNames.length; i++) {
             const currRule: string = ruleNames[i];
             const ruleVal: boolean = listOfRules[currRule];
-            const currRuleStatus: TutorialRuleStatus = { RuleName: currRule, RuleTurnOn: ruleVal, RuleStatus: false, RuleMessage: "", isStrict: false };
+            const currRuleStatus: TutorialRuleStatus = { ruleName: currRule, ruleTurnOn: ruleVal, ruleStatus: false };
             listOfRuleStatuses.push(currRuleStatus);
         }
         return listOfRuleStatuses;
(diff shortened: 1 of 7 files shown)


pxtlib/tutorialValidator.ts
```diff
@@ -5,6 +5,7 @@ namespace pxt.tutorial {
         RuleTurnOn: boolean;
         RuleStatus: boolean;
         RuleMessage: string;
+        isStrict: boolean;
```
jwunderl, line 8: for later; it might be worth considering leaving some of these as optional (e.g. `isStrict?: boolean`); on line 64 down below for example, RuleStatus, RuleMessage, and isStrict don't appear to necessarily have full values set yet / it looks like they're only being filled in to satisfy the compiler, whereas leaving them undefined might make it easier to identify in the future (having false or "" explicitly set can sometimes feel like it's intentional, where undefined is pretty clear it's missing)

pxtlib/tutorialValidator.ts
```diff
@@ -5,6 +5,7 @@ namespace pxt.tutorial {
         RuleTurnOn: boolean;
         RuleStatus: boolean;
```
jwunderl, line 6: nit: usually fields are camelCased like isStrict; feel free to leave for later as it's not part of this pr anyways~

pxtlib/tutorialValidator.ts
```diff
@@ -107,7 +111,9 @@ namespace pxt.tutorial {
     function extractBlockSnippet(tutorial: TutorialOptions, indexdb: pxt.Map<pxt.Map<number>>) {
         const { tutorialStepInfo, tutorialStep } = tutorial;
         const body = tutorial.tutorialStepInfo[tutorialStep].hintContentMd;
+
         let hintCode = "";
+        let requiredBlocks = "";
```
jwunderl, line 116: This isn't used? Fine to leave in if you were planning on adding something else in the future and it just snuck in now by accident~

pxtlib/tutorialValidator.ts
```diff
@@ -156,4 +162,62 @@ namespace pxt.tutorial {
...
+    function validateAtleastOneBlocks(usersBlockUsed: pxt.Map<number>, tutorialBlockUsed: pxt.Map<number>, currRule: TutorialRuleStatus): TutorialRuleStatus {
+        const userBlockKeys = Object.keys(usersBlockUsed);
+        let tutorialBlockKeys: string[] = []
+        if (tutorialBlockUsed != undefined) {
+            tutorialBlockKeys = Object.keys(tutorialBlockUsed);
+        }
+        let isValid: boolean = true;
+        if (userBlockKeys.length < tutorialBlockKeys.length) { // user doesn't have enough blocks
+            isValid = false;
+        }
```
jwunderl, line 182:
```suggestion
        let isValid = userBlockKeys.length >= tutorialBlockKeys.length; // user has enough blocks
```

webapp/src/tutorial.tsx
```diff
@@ -644,6 +644,17 @@ export class TutorialCard extends data.Component<TutorialCardProps, TutorialCard
...
 
+    areStrictRulesPresent(rules: pxt.tutorial.TutorialRuleStatus[]) {
+        if (rules != undefined) {
+            for (let i = 0; i < rules.length; i++) {
+                if (rules[i].RuleTurnOn && rules[i].isStrict) {
+                    return true;
+                }
+            }
+        }
+        return false;
```
jwunderl, line 655:
```suggestion
    areStrictRulesPresent(rules: pxt.tutorial.TutorialRuleStatus[]) {
        return rules?.some(rule => rule.RuleTurnOn && rule.isStrict) ?? false;
```

nit / option

pxtlib/tutorialValidator.ts
```diff
@@ -156,4 +162,62 @@ namespace pxt.tutorial {
...
+    * @param tutorialBlockUsed the next available index
+    * @param currRule the current rule with its TutorialRuleStatus
+    * @return a tutorial rule status for currRule
+    */
+    function validateAtleastOneBlocks(usersBlockUsed: pxt.Map<number>, tutorialBlockUsed: pxt.Map<number>, currRule: TutorialRuleStatus): TutorialRuleStatus {
+        const userBlockKeys = Object.keys(usersBlockUsed);
+        let tutorialBlockKeys: string[] = []
+        if (tutorialBlockUsed != undefined) {
+            tutorialBlockKeys = Object.keys(tutorialBlockUsed);
+        }
```
jwunderl, line 178:
```suggestion
        const tutorialBlockKeys = Object.keys(tutorialBlockUsed ?? {});
```

jwunderl (approved):
LGTM, some nits you can take / leave as you prefer ~

PR #8211, opened 2021-06-01

Jxwoon:
Added telemetry when the user clicks 'Continue Anyway' and 'Keep Editing' button. The telemetry will monitor which rules were turned on for this tutorial and state which rules were incorrect. Below is a sample telemetry output with console.log.

Screenshot (36)


livcheerful (approved):

Jxwoon:
Closing this PR because it has dynamically changing keys

PR #8212, opened 2021-06-02
Fix on-demand sync from cloud

eanders-MS:
When the user clicks a project to open it, we try to pull the latest version from the cloud before loading it into the editor. This process involves a variable number of asynchronous steps. The prior implementation chained these steps together as a set of `then` clauses appended to a resolved promise. It isn't entirely clear why, but this wasn't working correctly; the chained `then` clauses weren't executing in the expected order. This resulted in the project being loaded into the editor before the cloud sync completed. Note: This is a common pattern in the codebase, which is worrisome. I wonder if we were previously relying on a Bluebird-specific behavior.

I also increased the cloud sync timeout from 1.5 to 15 secs. I'd rather be generous here, and looking at the telemetry from different geographical regions, 1.5 secs is too short.

Fixes https://github.com/microsoft/pxt-arcade/issues/3549

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1385,7 +1385,7 @@ export class ProjectView
         return undefined;
     }
 
-    loadHeaderAsync(h: pxt.workspace.Header, editorState?: pxt.editor.EditorState): Promise<void> {
+    async loadHeaderAsync(h: pxt.workspace.Header, editorState?: pxt.editor.EditorState): Promise<void> {
         if (!h)
             return Promise.resolve()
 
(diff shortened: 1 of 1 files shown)


riknoll (approved):
hmmmm.... that really shakes up my understanding of promises if that is the case. it would kind of defeat the entire purposes of promises if they weren't guaranteed to execute in order, wouldn't it?

eanders-MS:
The pattern is so common in the codebase I think we'd see a lot of bugs if it were the case. Still, I'm not sure why this code was behaving the way it was!

livcheerful (approved):

webapp/src/app.tsx
```diff
@@ -1385,56 +1385,53 @@ export class ProjectView
...
-                    })
-            ]);
-        });
+                    }
+                })
+        ]);
 
-        return p.then(() => {
+        if (h) {
             workspace.acquireHeaderSession(h);
```
eanders-MS, line 1432: @riknoll This code was executing before the `then` containing `Promise.race` above. There must be a reasonable explanation.

riknoll: hmmm yeah the code looks fine to me. no idea what it could be

riknoll: unless it's getting called more than once and interleaving

PR #8216, opened 2021-06-04
Updated branding on header for documentation pages

shakao:
image

### DIFF
--- a/pxtlib/docsrender.ts
+++ b/pxtlib/docsrender.ts
@@ -297,13 +297,19 @@ namespace pxt.docs {
             params["drivename"] = html2Quote(theme.driveDisplayName);
         if (theme.homeUrl)
             params["homeurl"] = html2Quote(theme.homeUrl);
+
+
         params["targetid"] = theme.id || "???";
         params["targetname"] = theme.name || "Microsoft MakeCode";
         params["docsheader"] = theme.docsHeader || "Documentation";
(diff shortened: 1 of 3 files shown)


livcheerful (approved):

jwunderl (approved):
Hm, with this there's no way to get from docs back to the main site from docs since it's dropping the link; I guess it's not much better now as it isn't clear clicking the logo will take you back home though.

shakao:
yeah i was debating this, i thought it was a little weird to have the "MakeCode" text take you to the specific editor, but i'm willing to be convinced otherwise!! i think some of the docs (eg courses, etc) have links in the markdown itself but i definitely agree it's good to have something in the header

pxtlib/docsrender.ts
```diff
@@ -301,6 +301,10 @@ namespace pxt.docs {
         params["targetname"] = theme.name || "Microsoft MakeCode";
         params["docsheader"] = theme.docsHeader || "Documentation";
         params["targetlogo"] = theme.docsLogo ? `<img aria-hidden="true" role="presentation" class="ui ${theme.logoWide ? "small" : "mini"} image" src="${theme.docsLogo}" />` : ""
+        params["orglogo"] = (theme.organizationLogo || theme.organizationWideLogo) ? `<img aria-hidden="true" role="presentation" class="ui image" src="${theme.organizationLogo || theme.organizationWideLogo}" />` : ""
```
riknoll, line 304: should we be html escaping these? should be fine to do on images

riknoll (approved):

PR #8218, opened 2021-06-04
Added Strict Rule: Required Blocks

Jxwoon:
### Markdown
Under each tutorial step, the tutorial maker will have the ability to inform the user that a block or blocks are required for a tutorial step. To enable required blocks in markdown, the tutorial maker must use...

```
```requiredTutorialBlock
[Insert block here]
```'
```

### Behavior

 A custom pop-up message will appear informing the user what blocks they are required to use. In the screenshot below, this rule is combined with other rules.
Screenshot (40)

This rule can be used as a standalone too, but if a step does not have required block specified, the 'next' button will turn green. 
required blocks behavior stand alone

### DIFF
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -1072,6 +1072,7 @@ declare namespace pxt.tutorial {
         contentMd?: string;
         headerContentMd?: string;
         hintContentMd?: string;
+        requiredBlockMd?: string;
         activity?: number;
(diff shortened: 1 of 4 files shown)


jwunderl (approved):
Nice!

pelikhan:
Are you planning to render those block identifier?

Jxwoon:
> Are you planning to render those block identifiers?

Ideally, I would like to render blocks if it's not too complicated

PR #8219, opened 2021-06-04

Jxwoon:
Added telemetry to indicate

- a map of rules that are enabled/ not enabled 
- a map of valid and invalid code for ...
    - when the user clicks pop-message button "continue anyway"
    - when the user clicks pop-message button "keep editing"
    - when the user clicks hint button with pop up still in view
    - when the user clicks the previous button with pop-up still in view

Below is a sample console.log of the telemetry: 
Screenshot (43)

### DIFF
--- a/pxtlib/tutorial.ts
+++ b/pxtlib/tutorial.ts
@@ -16,6 +16,7 @@ namespace pxt.tutorial {
         let tutorialValidationRules: pxt.Map<boolean>;
         if (metadata.tutorialCodeValidation) {
             tutorialValidationRules = pxt.Util.jsonTryParse(tutorialValidationRulesStr);
+            categorizingValidationRules(tutorialValidationRules);
         }
 
         // noDiffs legacy
@@ -303,6 +304,27 @@ ${code}
             }
         }
         return { header, hint, requiredBlocks };
+    }
+
+    function categorizingValidationRules(listOfRules: pxt.Map<boolean>) {
+        const ruleNames: string[] = Object.keys(listOfRules);
+        let enabledRulesMessageArr: string[] = [];
+        let turnedOffRulesMessageArr: string[] = [];
+        for (let i = 0; i < ruleNames.length; i++) {
+            const currRule: string = ruleNames[i];
+            const ruleVal: boolean = listOfRules[currRule];
+            if (ruleVal) {
+                enabledRulesMessageArr.push(currRule);
+            } else {
+                turnedOffRulesMessageArr.push(currRule);
+            }
+        }
+        const enabledRulesMessage = enabledRulesMessageArr.join(', ');
+        const turnedOffRulesMessage = turnedOffRulesMessageArr.join(', ');
+        let enabledRulesMap: pxt.Map<string> = {};
+        enabledRulesMap['enableRules'] = enabledRulesMessage;
+        enabledRulesMap['turnedOffRules'] = turnedOffRulesMessage;
+        pxt.tickEvent('tutorial.validation.listOfRules ', enabledRulesMap);
     }
 
     /* Remove hidden snippets from text */
--- a/pxtlib/tutorial.ts
+++ b/pxtlib/tutorial.ts
@@ -323,7 +323,7 @@ ${code}
         const turnedOffRulesMessage = turnedOffRulesMessageArr.join(', ');
         let enabledRulesMap: pxt.Map<string> = {};
         enabledRulesMap['enableRules'] = enabledRulesMessage;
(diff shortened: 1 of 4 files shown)


pxtlib/tutorial.ts
```diff
@@ -305,6 +306,27 @@ ${code}
         return { header, hint, requiredBlocks };
     }
 
+    function categorizingValidationRules(listOfRules: pxt.Map<boolean>) {
+        const ruleNames: string[] = Object.keys(listOfRules);
```
jwunderl, line 310: nit: for the most part it's fine / preferable to omit explicit typings like these, as you get the same from type inference / it's often just extra noise.

pxtlib/tutorial.ts
```diff
@@ -305,6 +306,27 @@ ${code}
...
+        let enabledRulesMessageArr: string[] = [];
+        let turnedOffRulesMessageArr: string[] = [];
```
jwunderl, line 312:
```suggestion
        const enabledRulesMessageArr = [];
        const turnedOffRulesMessageArr = [];
```

nit: preferable to have const for these & enabledRulesMap.

(maybe worth a sidenote, having these as let makes it feel like you're thinking of const as frozen / immutable -- only the original reference / pointer is immutable, so it always points to the same array but the contents of that array can be modified after setting a const)

pxtlib/tutorial.ts
```diff
@@ -305,6 +306,27 @@ ${code}
...
+            } else {
+                turnedOffRulesMessageArr.push(currRule);
+            }
+        }
+        const enabledRulesMessage = enabledRulesMessageArr.join(', ');
+        const turnedOffRulesMessage = turnedOffRulesMessageArr.join(', ');
+        let enabledRulesMap: pxt.Map<string> = {};
+        enabledRulesMap['enabledRules'] = enabledRulesMessage;
+        enabledRulesMap['notEnabledRules'] = turnedOffRulesMessage;
+        pxt.tickEvent('tutorial.validation.listOfRules ', enabledRulesMap);
```
jwunderl, line 327:
```suggestion
        pxt.tickEvent('tutorial.validation.listOfRules', enabledRulesMap);
```

pxtlib/tutorial.ts
```diff
@@ -305,6 +306,27 @@ ${code}
...
+                enabledRulesMessageArr.push(currRule);
+            } else {
+                turnedOffRulesMessageArr.push(currRule);
+            }
+        }
+        const enabledRulesMessage = enabledRulesMessageArr.join(', ');
+        const turnedOffRulesMessage = turnedOffRulesMessageArr.join(', ');
+        let enabledRulesMap: pxt.Map<string> = {};
+        enabledRulesMap['enabledRules'] = enabledRulesMessage;
+        enabledRulesMap['notEnabledRules'] = turnedOffRulesMessage;
```
jwunderl, line 326: It's probably sufficient to just have enabledRules right, since we can infer the unenabled ones.

Another think to think of; it's a bit harder to query accurately when they're joined like this -- might be better to just send 1 tick per enabled rule. (I'll defer to @abchatra on that though)

webapp/src/tutorialCodeValidation.tsx
```diff
@@ -31,12 +32,18 @@ export class MoveOn extends data.Component<TutorialCodeValidationProps, tutorial
         this.setState({ visible: vis });
     }
 
+
+
     moveOnToNextTutorialStep() {
+        const sortedValidAndInvalidRules = this.props.codeValidAndInvalidRuleMap;
+        pxt.tickEvent('tutorial.validation.continueAnyway ', sortedValidAndInvalidRules);
```
jwunderl, line 39:
```suggestion
        pxt.tickEvent('tutorial.validation.continueAnyway', sortedValidAndInvalidRules);
```

webapp/src/tutorialCodeValidation.tsx
```diff
@@ -31,12 +32,18 @@ export class MoveOn extends data.Component<TutorialCodeValidationProps, tutorial
...
         this.props.onYesButtonClick();
         this.showUnusedBlocksMessage(false);
     }
 
     stayOnThisTutorialStep() {
+        const sortedValidAndInvalidRules = this.props.codeValidAndInvalidRuleMap;
+        pxt.tickEvent('tutorial.validation.keepEditing ', sortedValidAndInvalidRules);
```
jwunderl, line 46:
```suggestion
        pxt.tickEvent('tutorial.validation.keepEditing', sortedValidAndInvalidRules);
```

PR #8221, opened 2021-06-07
Handle generics from array methods in pyconverter

riknoll:
We still don't support generics except for arrays

Fixes https://github.com/microsoft/pxt-microbit/issues/3579

### DIFF
--- a/pxtpy/converter.ts
+++ b/pxtpy/converter.ts
@@ -2223,8 +2223,8 @@ namespace pxt.py {
                 if (!fun.pyRetType)
                     error(n, 9549, lf("function missing pyRetType"));
 
-                if (isGenericType(fun.pyRetType!) && recv && isArrayType(recv) && recvTp) {
-                    unifyTypeOf(n, recvTp.typeArgs![0])
+                if (recv && isArrayType(recv) && recvTp) {
+                    unifyArrayType(n, fun, recvTp);
                 }
                 else {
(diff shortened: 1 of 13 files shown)


jwunderl (approved):

livcheerful (approved):

riknoll:
@jwunderl @livcheerful Please take another look, I just updated this to handle all the generic methods for Arrays. I cheated on map though.

pxtpy/converter.ts
```diff
@@ -2937,4 +2937,37 @@ namespace pxt.py {
...
+            case "Array.shift":
+            case "Array.find":
+            case "Array.get":
+            case "Array._pickRandom":
+                unifyTypeOf(e, arrayType.typeArgs![0]);
+                break;
+            case "Array.concat":
+            case "Array.slice":
+            case "Array.filter":
+            case "Array.fill":
```
jwunderl, line 2955: Should `Array.sort` be in here? It returns the input array https://github.com/microsoft/pxt/blob/master/libs/pxt-common/pxt-core.d.ts#L109

```suggestion
            case "Array.fill":
            case "Array.sort":
```

jwunderl (approved):

pxtpy/converter.ts
riknoll: It doesn't in Python

PR #8222, opened 2021-06-07

Jxwoon:
Telemetry cases:

Once the tutorial loads, all validation rules are add to the telemetry. 

All user cases require the validation pop-up to be visible
- User clicks the next button
- User clicks 'Continue Anyways' 
- User clicks 'Keep Editing' 

Below is a sample console.log of telemetry with two rules enabled
Screenshot (48)

### DIFF
--- a/pxtlib/tutorial.ts
+++ b/pxtlib/tutorial.ts
@@ -16,6 +16,7 @@ namespace pxt.tutorial {
         let tutorialValidationRules: pxt.Map<boolean>;
         if (metadata.tutorialCodeValidation) {
             tutorialValidationRules = pxt.Util.jsonTryParse(tutorialValidationRulesStr);
+            categorizingValidationRules(tutorialValidationRules);
         }
 
         // noDiffs legacy
@@ -303,6 +304,27 @@ ${code}
             }
         }
         return { header, hint, requiredBlocks };
+    }
+
+    function categorizingValidationRules(listOfRules: pxt.Map<boolean>) {
+        const ruleNames: string[] = Object.keys(listOfRules);
+        let enabledRulesMessageArr: string[] = [];
+        let turnedOffRulesMessageArr: string[] = [];
+        for (let i = 0; i < ruleNames.length; i++) {
+            const currRule: string = ruleNames[i];
+            const ruleVal: boolean = listOfRules[currRule];
+            if (ruleVal) {
+                enabledRulesMessageArr.push(currRule);
+            } else {
+                turnedOffRulesMessageArr.push(currRule);
+            }
+        }
+        const enabledRulesMessage = enabledRulesMessageArr.join(', ');
+        const turnedOffRulesMessage = turnedOffRulesMessageArr.join(', ');
+        let enabledRulesMap: pxt.Map<string> = {};
+        enabledRulesMap['enableRules'] = enabledRulesMessage;
+        enabledRulesMap['turnedOffRules'] = turnedOffRulesMessage;
+        pxt.tickEvent('tutorial.validation.listOfRules ', enabledRulesMap);
     }
 
     /* Remove hidden snippets from text */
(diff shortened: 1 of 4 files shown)


pxtlib/tutorial.ts
```diff
@@ -305,6 +306,17 @@ ${code}
         return { header, hint, requiredBlocks };
     }
 
+    function categorizingValidationRules(listOfRules: pxt.Map<boolean>, title: string) {
+        const ruleNames = Object.keys(listOfRules);
+        for (let i = 0; i < ruleNames.length; i++) {
+            const setValidationRule: pxt.Map<string | number> = {};
+            setValidationRule["ruleName"] = ruleNames[i];
+            setValidationRule["enabled"] = listOfRules[ruleNames[i]] ? 'true' : 'false';
+            setValidationRule["tutorial"] = title;
```
jwunderl, line 315:
```suggestion
            const setValidationRule: pxt.Map<string> = {
                ruleName: ruleNames[i],
                enabled: listOfRules[ruleNames[i]] ? 'true' : 'false',
                tutorial: title,
            };
```

nit: usually a bit easier to use an an object literal vs explicitly indexing into it and setting properties, as it cuts out a bunch repetition / focuses attention more on the key itself / the values associated with them

webapp/src/tutorialCodeValidation.tsx
```diff
@@ -32,15 +33,36 @@ export class MoveOn extends data.Component<TutorialCodeValidationProps, tutorial
...
+        const rules = stepInfo.listOfValidationRules;
+
+        if (rules != undefined) {
+            const validationRuleStepStatus: pxt.Map<string | number> = {};
+            validationRuleStepStatus["tutorial"] = tutorialName;
+            validationRuleStepStatus["step"] = tutorialStep;
+            for (let i = 0; i < rules.length; i++) {
+                if (rules[i].ruleTurnOn) {
+                    validationRuleStepStatus["ruleName"] = rules[i].ruleName;
+                    let str = 'tutorial.validation' + (isEditing ? '.edit' : '.continue') + (rules[i].ruleStatus ? '.pass' : '.fail');
```
jwunderl, line 59: should state / cause of tick (edit/continue/next) and result (pass/fail) be part of `validationRuleStepStatus`?

webapp/src/tutorialCodeValidation.tsx
```diff
@@ -32,15 +33,36 @@ export class MoveOn extends data.Component<TutorialCodeValidationProps, tutorial
...
+                    pxt.tickEvent(str, validationRuleStepStatus);
+                }
+            }
+        }
```
jwunderl, line 63: Might be worth extracting this to a helper function as this is duplicated from `doubleClickedNextStep` and we tend to have issues consistently updating tick events when they're duplicated like this (e.g. we'll add a property in one place and forget the other)

abchatra (approved):

webapp/src/tutorial.tsx
```diff
@@ -432,22 +433,7 @@ export class TutorialCard extends data.Component<TutorialCardProps, TutorialCard
...
-            validationRuleStepStatus["step"] = tutorialStep;
-            for (let i = 0; i < rules.length; i++) {
-                if (rules[i].ruleTurnOn) {
-                    validationRuleStepStatus["ruleName"] = rules[i].ruleName;
-                    let str = 'tutorial.validation.next' + (rules[i].ruleStatus ? '.pass' : '.fail');
-                    pxt.tickEvent(str, validationRuleStepStatus);
-                }
-            }
-        }
+        this.validationTelemetry('.next');
```
shakao, line 436: small/can be done later: move the "." to be part of the helper function (`"tutorial.validation." + command`) instead of passed in

shakao (approved):

jwunderl (approved):

webapp/src/tutorialCodeValidation.tsx
Jxwoon: Yes, by combining the cause of the tick and result to get a faster query to see which rules have passed/failed.

PR #8224, opened 2021-06-08
Get typescript snippet from block id for rendering missing blocks in tutorial validation

shakao:
i think the better way to do this would be with react hooks, but sticking with the old lifecycle for now, for simplicity's sake

### DIFF
--- a/webapp/src/tutorialCodeValidation.tsx
+++ b/webapp/src/tutorialCodeValidation.tsx
@@ -50,7 +50,7 @@ export class ShowValidationMessage extends data.Component<TutorialCodeValidation
     componentDidMount() {
         this.props.ruleComponents.forEach(rule => {
             if (rule.blockIds && !this.state.ruleSnippets?.[rule.ruleName]?.length) {
-                this.getRuleSnippets(rule);
+                this.getRuleSnippetsAsync(rule);
             }
         })
     }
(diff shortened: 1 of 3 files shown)


jwunderl (approved):

webapp/src/tutorialCodeValidation.tsx
```diff
@@ -40,6 +44,60 @@ export class MoveOn extends data.Component<TutorialCodeValidationProps, tutorial
...
+            if ((rule.blockIds && rule.blockIds.sort().toString() != prev?.blockIds?.sort()?.toString())
+                || rule.blockIds.length != nextState.ruleSnippets?.[rule.ruleName]?.length) {
+                this.getRuleSnippets(rule);
+            }
+        })
+    }
+
+    renderRule(rule: pxt.tutorial.TutorialRuleStatus, index?: number) {
+        const snippets = this.state.ruleSnippets?.[rule.ruleName];
+        if (!snippets) {
```
riknoll, line 68: Aren't both of these cases the same?

webapp/src/tutorialCodeValidation.tsx
```diff
@@ -40,6 +44,60 @@ export class MoveOn extends data.Component<TutorialCodeValidationProps, tutorial
...
+            return <p key={index + rule.ruleName}>{(rule.ruleTurnOn && !rule.ruleStatus) ? rule.ruleMessage : ''}</p>
+        } else {
+            // TODO (jxwoon): render snippets
+            return <p key={index + rule.ruleName}>{(rule.ruleTurnOn && !rule.ruleStatus) ? rule.ruleMessage : ''}</p>
+        }
+    }
+
+    getRuleSnippets(rule: pxt.tutorial.TutorialRuleStatus) {
```
riknoll, line 76: getRuleSnippetsAsync?

riknoll (approved):

webapp/src/tutorialCodeValidation.tsx
shakao: yep, one's for jasmine to fill out!

livcheerful (approved):

Jxwoon (approved):

PR #8229, opened 2021-06-10
Use blocks toolbox instead of monaco snippet to get blocks image

shakao:
this still doesn't have quite everything from https://github.com/microsoft/pxt/blob/master/webapp/src/blocks.tsx#L1635 but it should be closer than the monaco snippet is!

### DIFF
--- a/webapp/src/tutorialCodeValidation.tsx
+++ b/webapp/src/tutorialCodeValidation.tsx
@@ -60,7 +60,7 @@ export class ShowValidationMessage extends data.Component<TutorialCodeValidation
             const prev = this.props.ruleComponents.find(r => r.ruleName == rule.ruleName);
             // Update if the list of blockids is different
             if ((rule.blockIds && rule.blockIds.sort().toString() != prev?.blockIds?.sort()?.toString())
-                || rule.blockIds.length != nextState.ruleBlocks?.[rule.ruleName]?.length) {
+                || rule.blockIds?.length != nextState.ruleBlocks?.[rule.ruleName]?.length) {
                 this.getRuleBlocksAsync(rule);
             }
         })
@@ -133,4 +133,3 @@ export class ShowValidationMessage extends data.Component<TutorialCodeValidation
         </div>;
     }
 }
-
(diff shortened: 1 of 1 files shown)


webapp/src/tutorialCodeValidation.tsx
```diff
@@ -60,42 +60,52 @@ export class ShowValidationMessage extends data.Component<TutorialCodeValidation
...
             return <p key={index + rule.ruleName}>{(rule.ruleTurnOn && !rule.ruleStatus) ? rule.ruleMessage : ''}</p>
         } else {
-            // TODO (jxwoon): render snippets
+            // TODO (jxwoon): render block dataUris
             return <p key={index + rule.ruleName}>{(rule.ruleTurnOn && !rule.ruleStatus) ? rule.ruleMessage : ''}</p>
         }
     }
 
-    getRuleSnippetsAsync(rule: pxt.tutorial.TutorialRuleStatus) {
+    getRuleBlocksAsync(rule: pxt.tutorial.TutorialRuleStatus) {
```
jwunderl, line 79: not in this pr; this seems like it might be a good function to async-ify to clean up eventually

jwunderl (approved):

webapp/src/tutorialCodeValidation.tsx
```diff
@@ -60,42 +60,52 @@ export class ShowValidationMessage extends data.Component<TutorialCodeValidation
             const prev = this.props.ruleComponents.find(r => r.ruleName == rule.ruleName);
             // Update if the list of blockids is different
             if ((rule.blockIds && rule.blockIds.sort().toString() != prev?.blockIds?.sort()?.toString())
-                || rule.blockIds.length != nextState.ruleSnippets?.[rule.ruleName]?.length) {
-                this.getRuleSnippetsAsync(rule);
+                || rule.blockIds.length != nextState.ruleBlocks?.[rule.ruleName]?.length) {
```
Jxwoon, line 63: small nit and you don't have to fix this. I can address this change when I create a PR.
```suggestion
                || rule.blockIds?.length != nextState.ruleBlocks?.[rule.ruleName]?.length) {
```

Jxwoon (approved):

riknoll (approved):

PR #8231, opened 2021-06-10
Disable Monaco in Windows App

livcheerful:
fixes https://github.com/microsoft/pxt-microbit/issues/4091 kinda

The help page definitely needs more fleshing out. I wasn't sure if I should include more info about sharing projects there or not.

I haven't figured out how to test it in the actual app without it being /beta? But let me know if there's a way. Also wasn't sure if that was the right way to link to an md in documentation, lemme know

trytry (no app needed) https://makecode.microbit.org/app/d4cbc2291e86f05bb5c85ba20f183aee6e732de6-c2db3709f5

the banner:
image

modal after trying to switch to text:
image

opening a text tutorial:
image

If you already have a project open in text, you'll get this. It's not possible to switch to blocks, so the only way to get to your code is to share it, and import into the editor. I could probably add switching to blocks too if you think that's important
image


webapp/src/dialogs.tsx
```diff
@@ -689,6 +689,25 @@ export function showReportAbuseAsync(pubId?: string) {
     })
 }
 
+export function showWinAppDeprecateAsync() {
+    pxt.tickEvent("winApp.dialog", undefined)
+    return core.confirmAsync({
+        header: lf("You can't get there from here!"),
+        hideAgree: true,
+        hasCloseIcon: true,
+        helpUrl: "/windows-app",
```
livcheerful, line 698: is the right way to do the help like this? .md-less?

shakao:
you could try entering "/@app/d4cbc2291e86f05bb5c85ba20f183aee6e732de6-c2db3709f5" into the extensions bar in the windows app?

livcheerful:
oooooooooooooooooooooooooooooooooo

webapp/public/vs/editor/editor.main.css
```diff
@@ -3,4 +3,4 @@
...
...action-label,.monaco-editor.vs .monaco-menu .monaco-action-bar.vertical .action-menu-item:hover .action-label{-ms-high-contrast-adjust:none;background:transparent!important;border:1px solid highlight;box-sizing:border-box}.monaco-diff-editor.vs-dark .diffOverviewRuler,.monaco-diff-editor.vs .diffOverviewRuler{display:none}.monaco-editor.vs-dark .line-delete,.monaco-editor.vs-dark .line-insert,.monaco-editor.vs .line-delete,.monaco-editor.vs .line-insert{background:transparent!important;border:1px solid highlight!important;box-sizing:border-box}.monaco-editor.vs-dark .char-delete,.monaco-editor.vs-dark .char-insert,.monaco-editor.vs .char-delete,.monaco-editor.vs .char-insert{background:transparent!important}}.monaco-action-bar .action-item.menu-entry .action-label.icon{width:16px;height:16px;background-repeat:no-repeat;background-position:50%;background-size:16px}.monaco-action-bar .action-item.menu-entry .action-label{background-image:var(--menu-entry-icon-light)}.hc-black .monaco-action-bar .action-item.menu-entry .action-label,.vs-dark .monaco-action-bar .action-item.menu-entry .action-label{background-image:var(--menu-entry-icon-dark)}.context-view .monaco-menu{min-width:130px}
```
shakao, line 6: this file/change included intentionally?

webapp/src/monaco.tsx
```diff
@@ -1311,6 +1333,13 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         let mode = pxt.editor.FileType.Text;
         this.currSource = file.content;
 
+
+        const isWinApp = pxt.BrowserUtils.isWinRT();
+        if (isWinApp) {
+            this.currFile = file;
```
livcheerful, line 1339: just setting currFile here so it doesn't send errors later

webapp/src/monaco.tsx
```diff
@@ -624,26 +629,43 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
-                        <MonacoFlyout ref={this.handleFlyoutRef} fileType={this.fileType}
-                            blockIdMap={this.blockIdMap}
-                            moveFocusToParent={this.moveFocusToToolbox}
-                            insertSnippet={this.insertSnippet}
-                            setInsertionSnippet={this.setInsertionSnippet}
-                            parent={this.parent} />
+
+                { isWinApp ?
+                    <div id="winAppError">
+                        <img className="ui medium centered image" alt={lf("An image of a shrugging board")} src={pxt.appTarget.appTheme.winAppDeprImage}/>
```
livcheerful, line 642: The full text is "Oops! Text editing is only available on the MakeCode website. Go here for more information

livcheerful: wasn't sure if there was an easier way to embed links in a sentence

webapp/public/vs/editor/editor.main.css
livcheerful: !!! no ty for the catch

riknoll:
Maybe a transparent background for the image?

webapp/src/monaco.tsx
shakao: yeah hmm i don't know if this is better, but you could do something like 

```
const msg = lf("Oops! etc {0}MakeCode website{0}. Go {0}here{0} for more..")
const parts = msg.split("{0}")
```

and then use `parts[0]`, `parts[1]` or whatever in your dialog... maybe a better unique separator than "{0}". this WOULD be a little easier to localize, i think

shakao: right the less dumb way is 

```
const msg = lf("Oops! ... available on the {0}. Go {1} for more..")
```

then split on `{\d}` and replace the {0} with `lf("MakeCode website") in a link, etc

webapp/src/dialogs.tsx
```diff
@@ -689,6 +689,25 @@ export function showReportAbuseAsync(pubId?: string) {
     })
 }
 
+export function showWinAppDeprecateAsync() {
+    pxt.tickEvent("winApp.dialog", undefined)
+    return core.confirmAsync({
+        header: lf("You can't get there from here!"),
```
livcheerful, line 695: oh also wasn't sure about this header text. stole this from the dialog i see anytime i try to log into teams on an illegal device 🙂

webapp/src/monaco.tsx
livcheerful: ohoooo

jwunderl: Yeah, should be what shannon says here ^^ -- as written the words won't be ordered correctly in other languages.

webapp/src/notification.tsx
```diff
@@ -121,9 +121,21 @@ export class NotificationBanner extends data.Component<ISettingsProps, {}> {
...
+        const isWinApp = pxt.BrowserUtils.isWinRT();
+
+        if (isWinApp) {
+            return <GenericBanner id="winAppBanner" parent={this.props.parent} bannerType={"negative"}>
+                <sui.Icon icon="warning circle" />
+                <div className="header">{lf("This app is being deprecated. Please use ")}
+                <a href={`https://${pxt.appTarget.name}`} target="_blank" rel="noopener noreferrer" onClick={()=>{pxt.tickEvent("winApp.openSite", undefined)}}>
+                    {lf("the website")}
+                </a>
+                {lf(" instead.")}
```
jwunderl, line 133: this one needs lf formatting attention too ~

jwunderl (approved):

PR #8233, opened 2021-06-10
Align share page branding with editor and home screen

shakao:
image

leaving the arcade share page alone for now, since long-term we plan to revisit the share page (and the tutorial) header


jwunderl (approved):

livcheerful (approved):

Jaqster (approved):
Sounds good.

PR #8235, opened 2021-06-10
Renamed Tutorial Validation Variables, Toggle Validation Message, and Render Blocks in Validation Message

Jxwoon:
Few things in this PR

- Renamed variables

- Validation Blocks are rendered into the message
Screenshot (53)

- Validation Messages disappears when anywhere in the workspace is clicked 
click workspace validation message dissapears

### DIFF
--- a/webapp/src/tutorial.tsx
+++ b/webapp/src/tutorial.tsx
@@ -715,7 +715,6 @@ export class TutorialCard extends data.Component<TutorialCardProps, TutorialCard
         const isRtl = pxt.Util.isUserLanguageRtl();
         return <div id="tutorialcard" className={`ui ${tutorialStepExpanded ? 'tutorialExpanded' : ''} ${tutorialReady ? 'tutorialReady' : ''} ${this.state.showSeeMore ? 'seemore' : ''}  ${!this.state.showHint ? 'showTooltip' : ''} ${hasHint ? 'hasHint' : ''}`} style={tutorialStepExpanded ? this.getExpandedCardStyle('height') : null} >
             {hasHint && this.state.showHint && !showDialog && <div className="mask" role="region" onClick={this.closeHint}></div>}
-            {validationEnabled && showTutorialValidationMessage && <div className="mask" role="region" onClick={this.closeTutorialValidationMessage}></div>}
             <div className='ui buttons'>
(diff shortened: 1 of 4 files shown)


webapp/src/tutorialCodeValidation.tsx
```diff
@@ -118,6 +124,7 @@ export class ShowValidationMessage extends data.Component<TutorialCodeValidation
         const rulesDefined = (rules != undefined);
         const strictRulePresent = this.props.areStrictRulesPresent;
         return <div>
+            {this.state.visible && <div className="mask" role="region" onClick={this.props.onNoButtonClick}></div>}
```
shakao, line 127: do you need both this mask and the one in tutorial.tsx?

webapp/src/tutorial.tsx
```diff
@@ -715,7 +715,7 @@ export class TutorialCard extends data.Component<TutorialCardProps, TutorialCard
         const isRtl = pxt.Util.isUserLanguageRtl();
         return <div id="tutorialcard" className={`ui ${tutorialStepExpanded ? 'tutorialExpanded' : ''} ${tutorialReady ? 'tutorialReady' : ''} ${this.state.showSeeMore ? 'seemore' : ''}  ${!this.state.showHint ? 'showTooltip' : ''} ${hasHint ? 'hasHint' : ''}`} style={tutorialStepExpanded ? this.getExpandedCardStyle('height') : null} >
             {hasHint && this.state.showHint && !showDialog && <div className="mask" role="region" onClick={this.closeHint}></div>}
-
+            {validationEnabled && showTutorialValidationMessage && <div className="mask" role="region" onClick={this.closeTutorialValidationMessage}></div>}
```
jwunderl, line 718:
```suggestion
            {validationEnabled && showTutorialValidationMessage && <div className="mask" role="region" onClick={this.closeTutorialValidationMessage} />}
```

just personal preference, completely fine to ignore; any tag can be self closing in [jt]sx so it's sometimes easier to follow when that is used~

jwunderl (approved):

webapp/src/tutorialCodeValidation.tsx
Jxwoon: Good point, I don't think it's necessary to have it in two places.

PR #8248, opened 2021-06-17
ignore memory ws error in windows app

jwunderl:
Do not switch to memory workspace in windows app, get back the old behavior to close https://github.com/microsoft/pxt-microbit/issues/4095


abchatra (approved):
Lets try this. Did you try reset  with this change as well? Just to be sure.

PR #8250, opened 2021-06-18
Update classes.md to fix a few issues.

jaydeetay:
I'm no Python expert, so let me know if I have this wrong, but...
* I've removed the member variables defined in the classes (rather than in the constructor). These are static members and thus aren't the ones being used in the methods.
* I've corrected the error that's generated when you try to access a private (ie name mangled) member
* I've removed the section on structural typing - that doesn't look right.  The code as presented works just fine for me.
* I've tried to clarify that the single underscore meaning 'protected' is just a convention. Python won't stop you accessing the member anyway.

[Or have I misunderstood and is Microbit Python a little 'non-standard' due to it needing to convert to/from JS?  I've tried to test out the code in MakeCode but have been running into issues like https://github.com/microsoft/pxt-microbit/issues/4201

### DIFF
--- a/common-docs/python/classes.md
+++ b/common-docs/python/classes.md
@@ -6,7 +6,6 @@ Let's take a look at a simple class-based example:
 
 ```python-ignore
 class Greeter:
-    greeting = ""
     def __init__(self, message):
         self.greeting = message
     def greet(self):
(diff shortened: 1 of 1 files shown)


jaydeetay:
Alas, I can't sign the CLA due to restrictions from my employer.  I'll withdraw this PR and file a bug instead.

jaydeetay:
Filed: https://github.com/microsoft/pxt/issues/8252

PR #8256, opened 2021-06-21
Add cancellation function to playPCMBufferAsync

riknoll:
Part of a fix for this issue: https://github.com/microsoft/pxt-microbit/issues/4212

Sometimes Safari delays sounds before the user interacts with the page instead of just cancelling them. That means when the user does finally do something to unmute, whatever sound they tried to play five minutes ago will play. Adding a cancellation token like function to be used in pxt-microbit to prevent that from happening.

### DIFF
--- a/pxtsim/simlib.ts
+++ b/pxtsim/simlib.ts
@@ -649,7 +649,7 @@ namespace pxsim {
 
         const MAX_SCHEDULED_BUFFER_NODES = 3;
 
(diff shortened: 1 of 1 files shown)


jwunderl (approved):
will the microbit / pxt-common-packages side just be a settimeout -> cancel and promise.race then?

riknoll:
yup!

livcheerful (approved):

PR #8259, opened 2021-06-22
Show offline message for Immersive Reader

livcheerful:
fixes https://github.com/microsoft/pxt-microbit/issues/4213

-> in the scenario where you don't already have the token and you're offline:
- this change properly handles the error and shows the offline message

-> in the scenario where you do have the token and you're offline
- I added a ping to check the connection before starting the launch to the immersive reader. the "Launch Immersive Reader" call doesn't fail quickly enough if there's no internet, so this shows the offline error instead of white-screening it for 15 seconds first

### DIFF
--- a/webapp/src/immersivereader.tsx
+++ b/webapp/src/immersivereader.tsx
@@ -231,11 +231,12 @@ export function launchImmersiveReader(content: string, tutorialOptions: pxt.tuto
         onPreferencesChanged: (pref: string) => {
             auth.updateUserPreferencesAsync({reader: pref})
         },
-        preferences: userReaderPref,
-        timeout: 5000
+        preferences: userReaderPref
     }
 
-    getTokenAsync().then(res => {
+    getTokenAsync().then(res =>{
+        return testConnectionAsync(res);
+    }).then(res => {
         if (Cloud.isOnline()) {
             const launchStart = pxt.Util.now();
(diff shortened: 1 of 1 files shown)


jwunderl (approved):

abchatra:
Can you elaborate what was the issue? I am not able to understand the fix.

livcheerful:
to my knowledge, we don't have a super reliable way to tell if a user is offline. this change uses a different get request api that also interacts with the offline code to help trigger that, and also shortens the timeout of the immersive reader. default is 15 seconds. also adds a tick to see how long it typically takes to launch so we can adjust that

abchatra:
5 sec may be too short for some geography like China. Is the bug because Cloud.isOnline() returned true but still there is no connectivity?

abchatra:
Or another way to ask is, do we show the error message after 15 sec?

abchatra:
Seems like share page does the right thing in offline app where it shows error message if there is no connectivity almost immediately. How much timeout does it have?

livcheerful:
share page tries to make a post request. difference here is that instead of making a call to an API, we have to call to the Immersive Reader SDK. I can change it so it tries to ping one of our APIs first, and then launch the SDK instead

abchatra (approved):

PR #8262, opened 2021-06-23
Send long press event on header bar back button press (for iOS app)

shakao:
tested this on my surface with touchscreen/mouse click, but have not tested on an iOS device. apparently iOS default long press duration is 500ms but that felt really fast to me

### DIFF
--- a/webapp/src/cmds.ts
+++ b/webapp/src/cmds.ts
@@ -185,6 +185,17 @@ export function nativeHostBackAsync(): Promise<void> {
     return Promise.resolve();
 }
 
+export function nativeHostLongpressAsync(): Promise<void> {
+    log(`native longpress`)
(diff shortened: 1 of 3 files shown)


jwunderl (approved):

livcheerful (approved):
do you want me to try it out on my ios device?

shakao:
@livcheerful oh, it's just posting a message so it won't really have an effect or do anything until the microbit iOS app is updated! my thought is i can quickly put it in beta and then the microbit foundation can test/let us know if there are any issues haha

martinwork:
@shakao When will it get to beta?

shakao:
@martinwork i'm going to make one more pull request with the updated message structure, but definitely by end of day today!

PR #8265, opened 2021-06-23
Always hit the live endpoint for compile requests

riknoll:
Fixes the main problem with https://github.com/microsoft/pxt-microbit/issues/4207

I will file an issue for fixing the hexcache so this also works better online

### DIFF
--- a/pxtlib/cpp.ts
+++ b/pxtlib/cpp.ts
@@ -1371,7 +1371,7 @@ namespace pxt.hexloader {
                         return U.httpGetTextAsync(hexurl + ".hex")
                     })
                     .then(r => r, e =>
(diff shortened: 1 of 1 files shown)


jwunderl (approved):

livcheerful (approved):

abchatra:
Is this a regression? What is the impact of forceLiveEndpoint?

abchatra:
I guess it is only used in isStatic scenario. I am good with this change.

abchatra (approved):

riknoll:
yeah, and the isStatic will fail because it doesn't implement the compile endpoint. Only the live service will ever respond to that request.

PR #8268, opened 2021-06-24
Compare text content instead of etags when fetching markdown

shakao:
we're requesting markdown content from `makecode.microbit.org` and our markdown is hosted on `makecode.com` so i believe we're hitting CORS issues where the etag is not accessible from javascript: https://stackoverflow.com/questions/5822985/cross-domain-resource-sharing-get-refused-to-get-unsafe-header-etag-from-re

i think we can fix this by adding an `Access-Controls-Expose-Headers` header to the backend response but in the meantime, comparing the markdown should also work


riknoll (approved):

shakao:
oh my local backend repo was a few (hundred) commits behind, it looks like we ARE setting the header to expose the etags... hmm

livcheerful (approved):

PR #8276, opened 2021-06-30
Also cache common project variations in staticpkg

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/4232

When building hexcache entries for the default project (e.g. libs/blocksprj), also cache hexfiles for those projects plus the other bundled packages.

This is a pretty naive implementation because I didn't want to go through the trouble of actually verifying which projects are compatible (that code is impossible to factor out cleanly). Instead I just try all of them and cache the ones that don't error.

I also took this opportunity to switch to async await, I recommend ignoring whitespace when reviewing the changes


eanders-MS (approved):
Seems good to me!

PR #8277, opened 2021-06-30
Patch top level shadow blocks

riknoll:
If a shadow block somehow becomes a top-level block (e.g. via https://github.com/microsoft/pxt-arcade/issues/3623), blockly refuses to load the workspace and throws an error instead. We don't really care about that restriction, so just convert any top-level shadow blocks in the XML to regular blocks before loading the workspace.

This will make it so that if any projects were affected by https://github.com/microsoft/pxt-arcade/issues/3623, they can now be opened. The actual fix for that issue is here:

https://github.com/microsoft/pxt-blockly/pull/354

### DIFF
--- a/pxtblocks/blocklyimporter.ts
+++ b/pxtblocks/blocklyimporter.ts
@@ -246,6 +246,18 @@ namespace pxt.blocks {
                     }));
             }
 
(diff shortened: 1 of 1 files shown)


eanders-MS (approved):
LGTM

jwunderl (approved):

PR #8281, opened 2021-07-07
Cloud status with spinner

eanders-MS:
This version more closely resembles the design, and uses a spinner to indicate save in progress.

cloud2

As you can see, it is vastly superior.

Notice the spinner has two states:
* Spinner rotating counterclockwise: Debouncer is active, waiting for a span of 1.5 seconds with no new edits.
* Spinner rotating clockwise (and more quickly): Cloud save is in progress.

If this feels too busy we can just use the clockwise spinner for both. To me that makes cloud sync feel slower than it actually is.

### DIFF
--- a/webapp/src/identity.tsx
+++ b/webapp/src/identity.tsx
@@ -205,13 +205,13 @@ export class CloudSaveStatus extends data.Component<CloudSaveStatusProps, {}> {
         const syncing = preparing || cloudStatus.value === "syncing";
 
         return (<div className="cloudstatusarea">
-            {!syncing && <i className={"ui tiny cloudicon xicon " + cloudStatus.icon}></i>}
-            {syncing && <div className={"ui tiny inline loader active cloudprogress " + (preparing ? "indeterminate" : "")}></div>}
-            {cloudStatus.value === "localEdits" && <span className="ui mobile hide cloudtext">{lf("saving...")}</span>}
-            {cloudStatus.value === "syncing" && <span className="ui mobile hide cloudtext">{lf("saving...")}</span>}
-            {cloudStatus.value === "justSynced" && <span className="ui mobile hide cloudtext">{lf("saved!")}</span>}
-            {cloudStatus.value === "offline" && <span className="ui mobile hide cloudtext">{lf("offline")}</span>}
-            {cloudStatus.value === "conflict" && <span className="ui mobile hide cloudtext">{lf("conflict!")}</span>}
+            {!syncing && <sui.Item className={"ui tiny cloudicon xicon " + cloudStatus.icon} title={cloudStatus.tooltip}></sui.Item>}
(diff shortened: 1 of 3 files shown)


eanders-MS:
Resolves https://github.com/microsoft/pxt-arcade/issues/3608

webapp/src/identity.tsx
```diff
@@ -181,11 +181,37 @@ export class UserMenu extends auth.Component<UserMenuProps, UserMenuState> {
...
+        const cloudMd = this.getData<cloud.CloudTempMetadata>(`${cloud.HEADER_CLOUDSTATE}:${this.props.headerId}`);
+        const cloudStatus = cloudMd.cloudStatus();
+        const showCloudButton = !!cloudStatus && cloudStatus.value !== "none" && auth.hasIdentity();
+        if (!showCloudButton) { return null; }
+        const preparing = cloudStatus.value === "localEdits";
+        const syncing = preparing || cloudStatus.value === "syncing";
+
+        return (<div className="cloudstatusarea">
+            {!syncing && <i className={"ui tiny cloudicon xicon " + cloudStatus.icon}></i>}
+            {syncing && <div className={"ui tiny inline loader active cloudprogress " + (preparing ? "indeterminate" : "")}></div>}
```
riknoll, line 209: huh, is this a semantic ui class? didn't know we had this.

riknoll (approved):
LGTM! I'm undecided on the reversing spinner, but no need to hold up this PR!

Jaqster:
Can we keep the "saved" text there?  Or if that's too busy maybe have a hover tooltip that says "saved" or something?  I do want to provide strong reassurance to people somehow.  The little cloud checkmark seems too subtle.  No opinion on the spinner - I don't get the difference between clockwise vs. counter-clockwise, and I don't think users will either.

eanders-MS (commented):
> Can we keep the "saved" text there? Or if that's too busy maybe have a hover tooltip that says "saved" or something? I do want to provide strong reassurance to people somehow. The little cloud checkmark seems too subtle. No opinion on the spinner - I don't get the difference between clockwise vs. counter-clockwise, and I don't think users will either.

Tooltip is a great idea, I will add that! As for keeping the text around, this could go a few different ways. I'd like to get this in as-is and give folks a chance to experience it. Maybe there's a middle ground. I agree the icon alone might be too subtle.

Spinner direction follows the language of Chrome's tab loading state, where it goes from "connecting", to "loading":

cloud3

One kid in the study mentioned that cloud save felt slow, and requested we speed it up :). Save itself is fast, but rolling the debouncer's wait time into the indicator makes it look otherwise. Kids in our user study immediately connected the cloud icon with "iCloud" and were able to infer a lot about how it worked from that association. My thought was to similarly leverage the ubiquity of Chrome to indicate the states of our cloud save process. It's not exactly the same, but there are parallels.

jwunderl (approved):
I'm impressed by how simple the flipping spinner is -- I hadn't ever noticed it explicitly in other places but here I definitely understood the 'something is queued to be saved' -> 'saving is happening' right away.

PR #8282, opened 2021-07-07
string.toUpperCase

Jacob-Tate:
Added string.toUpperCase as there is a project I'm working on that would benefit from it!

### DIFF
--- a/libs/pxt-common/pxt-helpers.ts
+++ b/libs/pxt-common/pxt-helpers.ts
@@ -402,7 +402,6 @@ namespace helpers {
         return stringSubstrHelper(s, start, end - start);
     }
 
-    // TODO move to PXT
     // also note this doesn't handle unicode, but neither does JS (there's toLocaleUpperCase())
     export function stringToUpperCase(s: string): string {
         let r = ""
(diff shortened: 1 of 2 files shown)


libs/pxt-common/pxt-helpers.ts
```diff
@@ -402,6 +402,22 @@ namespace helpers {
...
+            if (97 <= c && c <= 122) {
+                r += s.slice(prev, i) + String.fromCharCode(c - 32)
+                prev = i + 1
+            }
+        }
+        r += s.slice(prev)
+        return r
+    }
+
     // TODO move to PXT
```
jwunderl, line 421:
```suggestion
```

libs/pxt-common/pxt-helpers.ts
```diff
@@ -402,6 +402,22 @@ namespace helpers {
         return stringSubstrHelper(s, start, end - start);
     }
 
+    // TODO move to PXT
```
jwunderl, line 405:
```suggestion
```

I see this is duplicated from tolowercase / was accidentally left in when that was originally moved into pxt, but might as well drop the errant comments now ~

jwunderl (approved):
Looks good to me, thanks for the contribution!

riknoll (approved):
LGTM!

jwunderl:
Thanks again for fixing this!

PR #8286, opened 2021-07-08
Fix lf checks in gulpfile to grab missing translations

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/3627

one of the strings was formatted incorrectly (\` instead of \"), and when I was finding it I noticed a few others were broken as well (but not getting reported). This is because we were not properly extracting from a number of files (subfolders that weren't being checked, as well as a number of folders that had lfs but weren't included in the list: cli, pxtcompiler, pxteditor, pxtpy, pxtsim)

with this change updatestrings builds `334 files; 1811 strings`, where before it had `196 files; 1349 strings`

Also add an escape for skipping lfs on a given line to avoid a hack we had before for not erroring out on the lf function definitions.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -214,7 +214,20 @@ function pxtcommon() {
 
 // TODO: Copied from Jakefile; should be async
 function updatestrings() {
-    return buildStrings("built/strings.json", ["pxtlib", "pxtblocks", "pxtblocks/fields", "webapp/src"]);
+    return buildStrings("built/strings.json", [
+        "cli",
+        "pxtblocks",
+        "pxtblocks/fields",
+        "pxtcompiler",
+        "pxteditor",
+        "pxteditor/monaco-fields",
+        "pxtlib",
+        "pxtlib/emitter",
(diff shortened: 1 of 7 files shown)


gulpfile.js
```diff
@@ -249,8 +261,7 @@ function buildStrings(out, rootPaths, recursive) {
                             err("cannot JSON-parse " + m[1])
                         }
                     } else {
-                        if (!/util\.ts$/.test(filename))
-                            err("invalid format of lf() argument: " + args)
+                        err("invalid format of lf() argument: " + args)
```
eanders-MS, line 264: // @ignorelf@ needed here?

eanders-MS (approved):

gulpfile.js
jwunderl: Ah, this one's just part of the gulpfile so we don't pull translations from here / no lf's run in this context -- that said, I'll add it here for consistency with the other spots this function exists.

riknoll (approved):

PR #8293, opened 2021-07-13
Move virtual api and storage into pxtlib

eanders-MS:
In preparation to move the auth system to pxtlib, I'm moving "virtual api" and "storage" in a separate step. auth has dependencies on these systems. This operation was mostly a lift-and-shift, whereas moving auth will require some mild refactoring.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -147,7 +147,7 @@ export class ProjectView
     private preserveUndoStack: boolean;
     private rootClasses: string[];
 
-    private highContrastSubscriber: data.DataSubscriber = {
+    private highContrastSubscriber: pxt.data.DataSubscriber = {
         subscriptions: [],
         onDataChanged: () => {
             this.onHighContrastChanged();
@@ -317,9 +317,9 @@ export class ProjectView
         let active = document.visibilityState == 'visible';
         pxt.debug(`page visibility: ${active}`)
         this.setState({ active: active });
-        data.invalidate('pkg-git-pull-status');
-        data.invalidate('pkg-git-pr');
-        data.invalidate('pkg-git-pages')
+        pxt.data.invalidate('pkg-git-pull-status');
+        pxt.data.invalidate('pkg-git-pr');
+        pxt.data.invalidate('pkg-git-pages')
 
         // disconnect devices to avoid locking between tabs
         if (!active && !navigator?.serviceWorker?.controller)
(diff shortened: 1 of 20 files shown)


jwunderl (approved):

webapp/src/app.tsx
```diff
@@ -4726,6 +4726,12 @@ pxt.winrt.captureInitialActivation();
 document.addEventListener("DOMContentLoaded", () => {
     pxt.perf.recordMilestone(`DOM loaded`)
 
+    // Initialization order is important here.
```
eanders-MS, line 4729: Curious what people think of this change.

jwunderl: It seems reasonable to me; this whole event handler is already pretty order dependent anyways, with all the initializations & the promise chain at the end, so keeping these important inits in one place seems good

jwunderl (approved):

webapp/src/app.tsx
riknoll: LGTM. Not sure it should impact much of anything to initialize the shell early. Could maybe move loadcache outside of the event, since it would have been loaded on module load before.

riknoll: oh wait, i guess you can't move it out unless you also move out the init. in any case, shouldn't matter much!

riknoll (approved):

webapp/src/app.tsx
eanders-MS: Or I could move into pxtlib. It's pretty small. Let me try that.

eanders-MS: shell. move shell :)

webapp/src/cloudsync.ts
```diff
@@ -120,8 +120,8 @@ export class ProviderBase {
             pxt.storage.setLocal(this.name + CLOUD_USER, JSON.stringify(user))
         else
             pxt.storage.removeLocal(this.name + CLOUD_USER);
-        data.invalidate("sync:user")
-        data.invalidate("github:user")
+            pxt.data.invalidate("sync:user")
```
shakao, line 123:
```suggestion
        pxt.data.invalidate("sync:user")
```

small: extra indent, these aren't meant to be part of the else above, right?

webapp/src/cloudsync.ts
```diff
@@ -120,8 +120,8 @@ export class ProviderBase {
...
+            pxt.data.invalidate("github:user")
```
shakao, line 124:
```suggestion
        pxt.data.invalidate("github:user")
```

shakao (approved):

PR #8299, opened 2021-07-16
Move virtual api back to webapp

eanders-MS:
As I started integrating auth into skillmap, it became clear that it was unnecessary to move the virtual api into pxtlib. Probably better to keep it out of there anyway, so I moved it back to webapp.

Sorry for the churn!


jwunderl (approved):
I don't have a strong opinion one way or the other on whether this should live in pxtlib or webapp, but I guess keeping it in webapp does decrease skillmap bundle size a bit so probably better there if there's no need for pxtlib

riknoll (approved):

PR #8300, opened 2021-07-16
Cloud sync UI Changes

livcheerful:
- Add the Cloud Card to projects list if you're not signed in
- Add custom icon / sign in text to the header

based on Design office hour changes

looks like this!

image


theme/common.less
```diff
@@ -1938,6 +1938,21 @@ div.simframe.ui.embed {
...
+div.signin-button {
+    display: flex;
+    align-items: center;
+
+    .text {
+        margin-right: .5rem;
+    }
+}
+
+@media only screen and (max-width: @largestTabletScreen) {
```
livcheerful, line 1950: This is to keep the icon floating right. The default rules are lost bc I added an extra div around the whole sign in button 

image
 vs 
image

livcheerful: (in not desktop view)

theme/themes/pxt/views/card.overrides
```diff
@@ -152,6 +152,17 @@ a.ui.card:focus,
...
     background: fade(@primaryColor, 90%) !important;
 }
+
+.ui.card.link.cloudprojectscard {
+    background: @cloudProjectCardBackground!important;
+}
+
+@media only screen and (min-height: tallEditorBreakpoint), (min-width: @computerBreakpoint) {
+    .ui.card.link.cloudprojectscard .icon {
+        margin-left: -1rem
```
livcheerful, line 162: for some reason the new cloud svg floats right and looks silly when large. This bumps it to the left until it hits the smaller icon size, where it looks properly centered

without the extra rule:
image

livcheerful: with:
image

webapp/src/projects.tsx
```diff
@@ -747,6 +748,13 @@ export class ProjectsCarousel extends data.Component<ProjectsCarouselProps, Proj
                         <span className="header">{lf("New Project")}</span>
                     </div>
                 </div> : undefined}
+                {showCloudProjectsCard ? <div role="button" className="ui card link buttoncard cloudprojectscard" title={lf("Sign in to see cloud projects")}
```
eanders-MS, line 751: super minor suggestion: using `{showCloudProjectCard && ...}` would make this statement slightly more compact.

theme/themes/pxt/views/card.overrides
eanders-MS: How does the card look on mobile? Wondering if "Cloud Projects" wraps, or doesn't but should...

theme/themes/pxt/globals/site.variables
```diff
@@ -176,6 +176,7 @@
 @sidebarTutorialColor: @lightGrey;
 
 @homeScreenBackground: #ecf0f1;
+@cloudProjectCardBackground: #58AFBA;
```
eanders-MS, line 179: This css variable name is very specific. Might be better to make it categorical, like `@simulatorBackground` (or whatever makes sense).

eanders-MS (commented):
Looks good! Left a few minor comments/questions.

theme/themes/pxt/views/card.overrides
livcheerful: it doesn't, I'll add it

image

theme/themes/pxt/globals/site.variables
livcheerful: hmmm yeah I was thinking about that, but I was hesitant to give it a grand name when the color is just used once for now.

eanders-MS (approved):
LGTM!

PR #8302, opened 2021-07-19
Don't clear script selection when changing sort

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/1084


jwunderl (approved):

theme/scriptsearch.less
```diff
@@ -93,4 +93,8 @@
     .ui.card.file .content {
         margin-right: 3rem;
     }
+}
+
+.dropdown-menu {
```
shakao, line 98: theoretically this https://github.com/microsoft/pxt/blob/master/theme/themes/pxt/modules/dropdown.overrides#L53 should be setting the z-index, not sure if something changed or there was a missed case

riknoll: Yep, my css must have been super out of date. removed from pr

webapp/src/scriptmanager.tsx
```diff
@@ -300,6 +308,10 @@ export class ScriptManagerDialog extends data.Component<ScriptManagerDialogProps
         return sortingFunction;
     }
 
+    private getId(scr: pxt.workspace.Header) {
```
eanders-MS, line 311: I think this might have broken the Delete button. See https://github.com/microsoft/pxt-arcade/issues/3676

PR #8306, opened 2021-07-20
Fix exception when client is undefined

riknoll:
The short-circuit in these functions is making them return undefined instead of a promise that resolves to undefined when `client()` is falsy. That causes an exception downstream when whatever is calling them tries to `.then()` the promise. This was causing my local serve to fail.

Converting them to async await which handles undefined.


eanders-MS (approved):
Good catch, thanks.

PR #8307, opened 2021-07-20
Fix pwa scope and starturl

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/4273

With this change we can't test the beta or /app/ urls, but that isn't a big deal. Testing the normal site should suffice in those scenarios


jwunderl (approved):

abchatra:
LGTM. Please port to stable after merging.

livcheerful (approved):

PR #8308, opened 2021-07-20
Move editor sidebar into a tabbed panel component

shakao:
this PR adds a TabPane component; the tabs are pulled from the children of the component, if there is only one child then no tabs are displayed (current scenario). this should make it easy to add the tutorial and sidedocs view later. 

it also separates the "filelist" div element (previously the sidebar, but also the floating simulator in mobile/tablet views) into an "editorSidebar" parent and a "simPanel" child element, so that we can later show the tutorial view and the small simulator. 

builds (everything should look basically identical to before):
https://arcade.makecode.com/app/17c2c55576a2d6c83d762d2a0e0e6a4342685280-77a8e65f69
https://minecraft.makecode.com/app/794632fa97b320d7ad5ebb0c65497a9e7a720d60-cc4756b75c

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -25,10 +25,8 @@ import * as accessibility from "./accessibility";
 import * as tutorial from "./tutorial";
 import * as sidebarTutorial from "./sidebarTutorial";
 import * as editortoolbar from "./editortoolbar";
-import * as simtoolbar from "./simtoolbar";
 import * as dialogs from "./dialogs";
 import * as identity from "./identity";
-import * as filelist from "./filelist";
 import * as container from "./container";
 import * as scriptsearch from "./scriptsearch";
 import * as projects from "./projects";
@@ -47,6 +45,7 @@ import * as auth from "./auth";
 import * as cloud from "./cloud";
 import * as user from "./user";
 import * as headerbar from "./headerbar";
+import * as sidepanel from "./sidepanel";
 
 import * as monaco from "./monaco"
 import * as pxtjson from "./pxtjson"
@@ -134,7 +133,6 @@ export class ProjectView
     chooseHwDialog: projects.ChooseHwDialog;
     prevEditorId: string;
     screenshotHandlers: ((msg: pxt.editor.ScreenshotData) => void)[] = [];
-    fileListRef: HTMLDivElement;
 
     private lastChangeTime: number;
     private reload: boolean;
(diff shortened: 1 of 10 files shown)


webapp/src/sidepanel.tsx
```diff
@@ -0,0 +1,74 @@
...
+    renderCore() {
+        const { parent, inHome, showKeymap, showSerialButtons, showFileList, showFullscreenButton,
+            collapseEditorTools, simSerialActive, deviceSerialActive,
+            handleHardwareDebugClick, handleFullscreenButtonClick } = this.props;
+
+        return <div id="simulator" className="simulator">
+            <TabPane id="editorSidebar">
+                <TabContent name="tab-simulator">
+                    <div className="ui items simPanel">
+                        <div id="boardview" className={`ui vertical editorFloat`} role="region" aria-label={lf("Simulator")} tabIndex={inHome ? -1 : 0} />
```
riknoll, line 57:
```suggestion
                        <div id="boardview" className="ui vertical editorFloat" role="region" aria-label={lf("Simulator")} tabIndex={inHome ? -1 : 0} />
```

webapp/src/sidepanel.tsx
```diff
@@ -0,0 +1,74 @@
...
+                        <simtoolbar.SimulatorToolbar parent={parent} collapsed={collapseEditorTools} simSerialActive={simSerialActive} devSerialActive={deviceSerialActive} />
+                        {showKeymap && <keymap.Keymap parent={parent} />}
+                        <div className="ui item portrait hide hidefullscreen">
+                            {pxt.options.debug && <sui.Button key='hwdebugbtn' className='teal' icon="xicon chip" text={"Dev Debug"} onClick={handleHardwareDebugClick} />}
```
riknoll, line 61:
```suggestion
                            {pxt.options.debug && <sui.Button key="hwdebugbtn" className="teal" icon="xicon chip" text={"Dev Debug"} onClick={handleHardwareDebugClick} />}
```

riknoll (approved):

webapp/src/app.tsx
```diff
@@ -4295,30 +4280,19 @@ export class ProjectView
...
-                            <div id="serialPreview" className="ui editorFloat portrait hide hidefullscreen">
-                                <serialindicator.SerialIndicator ref="simIndicator" isSim={true} onClick={this.openSimSerial} parent={this} />
-                                <serialindicator.SerialIndicator ref="devIndicator" isSim={false} onClick={this.openDeviceSerial} parent={this} />
-                            </div> : undefined}
-                        {showFileList ? <filelist.FileList parent={this} /> : undefined}
-                        {!isHeadless && <div id="filelistOverlay" role="button" title={lf("Open in fullscreen")} onClick={this.toggleSimulatorFullscreen}></div>}
-                    </div>
-                </div>
+                <sidepanel.Sidepanel parent={this} inHome={inHome}
+                    showKeymap={this.state.keymap && simOpts.keymap}
```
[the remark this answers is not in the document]
shakao: no, it's better to pass them in so that react can handle the component update--recalculating them can cause problems with the render loop. there's some duplicate buttons and code in the different sim views (fullscreen button, the serial stuff) but that's probably going to be dealt with separate from the tutorial stuff

jwunderl (approved):

dmonroym (approved):

livcheerful (approved):

Issue #5390, opened 2019-03-25
Static Python Feature


pelikhan:
## Description

Convert Blocks and JavaScript to Static Python and back.

## How to use it?

Click on the **Python** tab in the language selection toggle above the editor.

![py3](https://user-images.githubusercontent.com/4175913/54951490-080ad500-4f01-11e9-86ed-0f2f31fb4774.gif)

If you cannot see that button and you editor supports this experiment, go to **About**, then click **Experiments** and **Static Python**. If you cannot see this card, your editor does not support yet this feature.

## Feedback

Do you use this feature? Is there something you'd like to change? Did you encounter issues? Please drop a comment on this issue to send us your feedback.

pelikhan:
- [ ] use 4 spaces instead of 2
- [ ] auto-indent

codinggirl:
Where is my editor?

codinggirl:
Is there any live samples?

pelikhan:
In arcade.makecode.com, go to about -> experiments -> Static Python.
In maker.makecode.com, enabled by default. Looking forward your feedback.

pokon549:
Hi, all.
I've met some strange problems reported by the Static Python editor, but I actually write no codes inside the textarea. Is that a bug?
The Arcade version is 0.13.20.
![深度截图_20190805174056](https://user-images.githubusercontent.com/51010793/62455106-3ef2d080-b7a8-11e9-9b6d-312edd1ad250.png)

darzu:
Hi, yes these are known issues will be fixed in: https://github.com/microsoft/pxt/pull/6167

I'm going to close this issue as it is too broad. We're tracking individual Python issues with the "python" label in the pxt, pxt-arcade, and pxt-minecraft repo.

Thanks for your feedback. Major work is being done on the Python support now (we have lots of issues we know about) and over the next few months we'll likely have something more stable that we'd love to get more community feedback on.

danielbloy:
Just some feedback guys and a big thumbs up. I've been using MakeCode arcade in my code clubs for around 6 months now and the children (around 10 years old) and have been confined to Blocks as the children really struggle with JavaScript. They do respond better to Python as it looks less scary and I'm currently writing new material for them in Python. So far it's been fairly easy as I can navigate between Blocks and Python and use the generated results to guide my materials but it would be really useful to have a more comprehensive guide to the API in Python.

PR #8309, opened 2021-07-20
update language about static python experimental

mavaddat:
Re enabling experiments, I borrowed language from @pelikhan on https://github.com/microsoft/pxt/issues/5390

### DIFF
--- a/docs/language.md
+++ b/docs/language.md
@@ -24,7 +24,7 @@ so may be unavailable in your MakeCode editor.
 To enable an experiment, click the gear icon ⚙️,
 go to `About` , then click `Experiments`.
 Experiments are released selectively, so 
-if an experiment you want does not show for you, 
+if an experiment you want does not show yet, 
 please be patient &emdash; it will be rolled to 
 your editor soon!
 
--- a/docs/language.md
+++ b/docs/language.md
@@ -19,8 +19,14 @@ functions, lambdas, classes).
 
 ## ~ hint
 
-Static Python is not still in an Experimental phase 
-and might not be available in your editor.
+Static Python is still in an experimental phase, 
+so may be unavailable in your MakeCode editor.
+To enable an experiment, click the gear icon ⚙️,
+go to `About` , then click `Experiments`.
+Experiments are released selectively, so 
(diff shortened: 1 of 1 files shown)


mavaddat (commented):
I added the word "yet" instead of "for you" to avoid redundancy and build positive anticipation.

dmonroym (approved):

jwunderl:
Hm, were there any editors with python enabled only as an experiment? I thought it was fully enabled in any of the editors where it is possible to enable it at the moment (micro:bit, minecraft, arcade, maker) -- change looks good if it is hidden behind an experiment anywhere though!

jwunderl (approved):

mavaddat:
> Hm, were there any editors with python enabled only as an experiment? 

I was wondering that too: thinking maybe the hint author was saying that static Python was *no longer* experimental. If that's the case, I can modify the first sentence so it says something like, "Static Python is now available for all MakeCode releases!" or, "Static Python has graduated from experimental release, so all MakeCode editors can now use it."

N.B., I try to avoid language like *no longer experimental*, since it can appear to the eye (skim reading) as suggesting unavailability.

jwunderl:
Yes, that's a good point -- I'm not sure I would say 'All MakeCode releases' as we have several editors that are supported but have not been updated since the implementation of python support (e.g. adafruit). Tagging in @ganicke here as I'm sure he'll have a good idea on how to phrase it :)

ganicke:
With Python as a standard language option for micro:bit, Arcade, and Minecraft, I think note should say more that "Support for Static Python might still be experimental for some MakeCode editors. If `Python` isn't shown as a language choice in the editor you're using, you can enable it by..."

jwunderl:
Thank you for your contribution! :)

mavaddat:
Ah shoot &mdash; the `&mdash;` html entity does not render on [the webpage](https://makecode.com/language) as an em-dash. Can someone just change the html entity to [the equivalent unicode character](https://www.compart.com/en/unicode/U+2014) “—”?

jwunderl:
Good catch, put up change replacing it~

PR #8313, opened 2021-07-20
Escape characters so regex creation doesn't fail

### LINKED ISSUE #3642: The scrollbar is displayed differently in Firefox
**Repro steps**:
• Navigating to https://makecode.microbit.org/beta
• Click "LED" and choose any one block
• Change data and observe the scroll bar that appears 
**Expected Result**: 
safari
**Actual Result**:
The scrollbar is displayed differently in Firefox than safari and chrome
firefox

dmonroym:
Fixes https://github.com/microsoft/pxt-arcade/issues/3642

When we attempted to create a RegExp, if the sprite had been generated with an ID that contained reserved regex symbols, this would fail. This fix aims at escaping all these special characters.

It was tested using https://github.com/eanders-MS/palette-sprite project.

### DIFF
--- a/pxtlib/tilemap.ts
+++ b/pxtlib/tilemap.ts
@@ -766,8 +766,8 @@ namespace pxt {
             }
 
             if (files) {
-                const shortId = getShortIDForAsset(asset);
-                const displayName = asset.meta?.displayName || "";
+                const shortId = Util.escapeForRegex(getShortIDForAsset(asset));
+                const displayName = Util.escapeForRegex(asset.meta?.displayName) || "";
 
                 let assetTsRefs: string;
                 switch (asset.type) {


jwunderl (approved):
Nice! I didn't know we already had that helper, good to know for the future~

riknoll (approved):

jwunderl:
(for reference / making sure there's a backref to this pr the issue is microsoft/pxt-arcade#3642 vs the one in this repo)

dmonroym:
> Nice! I didn't know we already had that helper, good to know for the future~

Neither did I but when I went to implement it the util.ts place looked to be the most sensible place to do it. Turns out it's been there for over 5 yrs

> (for reference / making sure there's a backref to this pr the issue is [microsoft/pxt-arcade#3642](https://github.com/microsoft/pxt-arcade/issues/3642) vs the one in this repo)

Ah yes, still need to get used with working on multiple repos. Fixed it in the description

livcheerful (approved):

PR #8317, opened 2021-07-21
Add storage api to pxt server

eanders-MS:
This PR adds a new storage API to pxtlib: `pxt.storage.localhost`. When calls to this API are made while running on localhost, it will route them to the pxt server running in the background on port 3232. In production, it routes calls to local browser storage. This API enables multiple localhost document frames to share local storage, the way they would if they were running in production (assuming the frames were served from the same domain). 

#### Why do we need this?
Browsers consider port number when determining same-origin, and skillmap and editor are served on different ports when running locally. This means their same-origin status is different when developing locally vs. running in production. In production, the two will share browser local storage, whereas they won't when running locally in development. This matters now because the two need to share some auth state via browser storage. Integrating auth and cloud sync into skillmap would be very difficult without this workaround.

#### Are there side effects?
There are a few:
1. This change has a non-obvious side effect which we might want to curb at some point: *All* localhost frames using this API will map to the same local storage regardless of browser or in-private status.
2. Stored data is persisted in files in a subfolder off the current working directory, at `.pxt/store/...`. I will open a PR in pxt-arcade to .gitignore this path.

#### Anything else?
I needed to update @types/node to get a new mkdir signature. **You will need to npm install after syncing these changes.**

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -7137,7 +7137,7 @@ function errorHandler(reason: any) {
         process.exit(1)
     }
 
-    let msg = reason.stack || reason.message || (reason + "")
+    let msg = reason.stack || reason.message || reason.statusMessage || (reason + "")
     console.error("INTERNAL ERROR:", msg)
     process.exit(20)
 }
--- a/cli/server.ts
+++ b/cli/server.ts
@@ -7,6 +7,7 @@ import * as nodeutil from './nodeutil';
 import * as hid from './hid';
 import * as net from 'net';
 import * as crowdin from './crowdin';
+import * as storage from './storage';
 
 import { promisify } from "util";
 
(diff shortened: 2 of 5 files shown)


jwunderl (approved):
Seems fine to me, but might be good to just use the same storage location as fileworkspace if possible? https://github.com/microsoft/pxt/blob/master/webapp/src/fileworkspace.ts (/projects/*)

eanders-MS:
Putting storage in the projects folder feels like conflating two very different things. Kind of related: for a long time I didn't realize that the projects folder was generated. The name seems like a reasonable top-level folder name in a repo. Moving it below .pxt/ might help clarify this.

eanders-MS:
Before I merge this PR could someone kindly approve https://github.com/microsoft/pxt-arcade/pull/3673?

PR #8321, opened 2021-07-22
Auth system to use shared local storage

eanders-MS:
Update auth system to use the pxt server-backed local storage.

### DIFF
--- a/pxtlib/auth.ts
+++ b/pxtlib/auth.ts
@@ -1,9 +1,10 @@
 namespace pxt.auth {
 
-    const CSRF_TOKEN = "csrf-token";
-    const AUTH_LOGIN_STATE = "auth:login-state";
-    const AUTH_USER_STATE = "auth:user-state";
-    const X_PXT_TARGET = "x-pxt-target";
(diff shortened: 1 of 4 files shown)


riknoll (approved):
LGTM but imo it is very confusing that all of these functions are in a namespace with "localhost" in it.

eanders-MS:
> LGTM but imo it is very confusing that all of these functions are in a namespace with "localhost" in it.

Agreed. I really struggled to find a good name. I don't love "localhost". If you have a suggestion I'm all ears.

eanders-MS:
Alternative names I considered:
* `pxt.storage.routed` -> because the calls are routed one direction or the other.
* `pxt.storage.pxt` -> because the calls go to the pxt server, sometimes. Having "pxt" appear twice in the namespace felt odd.
* `pxt.storage.local` -> because the calls go to local storage one way or another. This one would have been nice except that "local" is already used in another context in the same file.
* `pxt.storage.dev` -> because the calls have special behavior in development scenarios. Seemed odd for code shipping to production exist in a "dev" namespace.

PR #4209, opened 2018-04-25
Tweaks for electron

guillaumejenkins:
Required for full end-to-end solution test

To test:
- [x] pxt serve
- [x] pxt serve --pkg (my changes fix a bunch of issues with pxt serve --pkg)
- [x] pxt staticpkg
- [x] pxt staticpkg —gh (do this on your own sample target as it uploads to github)
- [x] pxt uploadtrg (do this on one of our own targets)

PR #8331, opened 2021-07-27
Use whole Immersive Reader API URL

livcheerful:
fixes https://github.com/microsoft/pxt-microbit/issues/4281

The issue was in the Electron app it was resolving to makecode.microbit.org/immreader (when it should be makecode.microbit.org**/api/**immreader ) so I'm just hardcoding the whole URL and also reverted to the old API for safety

https://github.com/microsoft/pxt/pull/8259/files?diff=split&w=1

I had changed it in this PR bc I thought it might be better (for offline use) but I don't think it really matters.

### DIFF
--- a/webapp/src/immersivereader.tsx
+++ b/webapp/src/immersivereader.tsx
@@ -190,10 +190,10 @@ function getTokenAsync(): Promise<ImmersiveReaderToken> {
     const cachedToken: ImmersiveReaderToken = pxt.Util.jsonTryParse(storedTokenString);
 
     if (!cachedToken || (Date.now() / 1000 > cachedToken.expiration)) {
-        return pxt.Cloud.privateGetAsync("immreader").then(
+        return pxt.Util.requestAsync({ url: "https://makecode.com/api/immreader", method: "GET" }).then(
             res => {
-                pxt.storage.setLocal(IMMERSIVE_READER_ID, JSON.stringify(res));
-                return res;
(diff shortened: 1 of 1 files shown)


jwunderl (approved):

webapp/src/immersivereader.tsx
```diff
@@ -190,10 +190,10 @@ function getTokenAsync(): Promise<ImmersiveReaderToken> {
     const cachedToken: ImmersiveReaderToken = pxt.Util.jsonTryParse(storedTokenString);
 
     if (!cachedToken || (Date.now() / 1000 > cachedToken.expiration)) {
-        return pxt.Cloud.privateGetAsync("immreader").then(
+        return pxt.Util.requestAsync({ url: "https://makecode.com/api/immreader", method: "GET" }).then(
```
abchatra, line 193: This needs to be relative path otherwise arcade.staging.pxt.io is going to point to production.

eanders-MS: Agreed it needs to be relative. For staging and also localhost development. Maybe condition this behind an isElectron check.

livcheerful: 👍 I changed it to use the forceliveendpoint parameter (found from https://github.com/microsoft/pxt/pull/4209/commits/8d6ce291ef2641107fdfbd3d9c74436087e4d953 )

PR #8333, opened 2021-07-27
sync skillmap progression

eanders-MS:
Sync skillmap progress to cloud. It does this by hooking into the location where the skillmap loads and saves user state and injecting some state from the cloud. I *think* this is a good approach. Would appreciate feedback!

### DIFF
--- a/pxtlib/auth.ts
+++ b/pxtlib/auth.ts
@@ -20,7 +20,7 @@ namespace pxt.auth {
     const DEV_BACKEND_LOCALHOST_5500 = "http://localhost:5500"; // if running in Docker container
     const DEV_BACKEND_LOCALHOST_8080 = "http://localhost:8080"; // if not running in Docker
 
-    const DEV_BACKEND = DEV_BACKEND_PROD;
+    const DEV_BACKEND = DEV_BACKEND_LOCALHOST_8080;
 
     let authDisabled = false;
 
@@ -40,6 +40,11 @@ namespace pxt.auth {
         };
     };
 
+    export type UserSkillmapState = {
+        mapProgress: any;
+        completedTags: any;
+    };
+
     /**
      * User preference state that should be synced with the cloud.
      */
@@ -47,8 +52,8 @@ namespace pxt.auth {
         language?: string;
         highContrast?: boolean;
         reader?: string;
+        skillmap?: UserSkillmapState;
     };
-    export type UserPreferencePart = "language" | "high-contrast" | "immersive-reader";
 
     export const DEFAULT_USER_PREFERENCES: () => UserPreferences = () => ({
         highContrast: false,
@@ -57,7 +62,7 @@ namespace pxt.auth {
     });
 
     /**
-     * In-memory auth state. Changes to this state trigger virtual API subscription callbacks.
+     * Cloud-synced user state.
      */
     export type State = {
         profile?: UserProfile;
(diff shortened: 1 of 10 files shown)


pxtlib/auth.ts
```diff
@@ -20,7 +20,7 @@ namespace pxt.auth {
     const DEV_BACKEND_LOCALHOST_5500 = "http://localhost:5500"; // if running in Docker container
     const DEV_BACKEND_LOCALHOST_8080 = "http://localhost:8080"; // if not running in Docker
 
-    const DEV_BACKEND = DEV_BACKEND_PROD;
```
livcheerful, line 23: is this a change we want to commit, or just for testing?

pxtlib/util.ts
```diff
@@ -333,6 +333,115 @@ namespace ts.pxtc.Util {
...
+                } else if (!Object.is(atObjA, atObjB)) {
+                    // Value exists in both, and they're not the same instance.
+                    if (typeof atObjA !== typeof atObjB) {
+                        // Type changed. Replace the whole subtree.
+                        diff.replace.push({
+                            op: 'replace',
+                            path: basePath.concat(key),
+                            value: atObjB
+                        });
+                    } else {
```
riknoll, line 390: Nitpick:  should the clause here just be combined with the above case?

pxtlib/util.ts
```diff
@@ -333,6 +333,115 @@ namespace ts.pxtc.Util {
...
+            const lastKey: any = path.pop();
+            if (lastKey == null) {
+                throw new Error("jsonPath: missing last key");
+            }
+            let subObj = obj;
+            // Add sub-objects for each path element except the last one.
+            let currKey: any;
+            while (((currKey = path.shift())) != null) {
+                if (!(currKey in subObj)) {
+                    subObj[currKey] = {};
```
riknoll, line 432: will this work with arrays? should recreate all of the properties, but will they be proper arrays? i.e. can I do subObj[currKey].map()?

skillmap/src/lib/workspaceProvider.ts
```diff
@@ -20,10 +23,26 @@ export async function saveProjectAsync(project: pxt.workspace.Project): Promise<
...
+        userState = {
+            id: guidGen(),
+            completedTags: {},
+            mapProgress: {},
+            version: pxt.skillmap.USER_VERSION
+        };
+    }
+    // Read synchronized skillmap state from cloud profile.
+    const skillmapState = await authClient.getSkillmapStateAsync();
+    userState = {
```
riknoll, line 40: this should use the "version" from the authClient

riknoll (changes requested):

pxtlib/util.ts
eanders-MS: Zooming out, this feels like something we shouldn't invent. There are edge cases (like creating arrays vs objects, as you point out), and we need unit tests. I generally try to avoid taking additional dependencies, especially client side, but in this case I'd guess that a well-tested version of this algorithm already exists. I'll poke around npm for a good package.

skillmap/src/lib/workspaceProvider.ts
eanders-MS: Can you expand? The meaning of the field hasn't changed, just moved where it is initialized.

pxtlib/util.ts
eanders-MS: Ended up reinventing the wheel anyway. Works with arrays now.

pxtlib/auth.ts
eanders-MS: Reverted this change to point to the staging backend (what it should have been -- prod was incorrect). Identity is working again in staging, and has been updated to support skillmap sync.

skillmap/src/lib/workspaceProvider.ts
eanders-MS: @riknoll My sense is that skillmap should own this version number. Does that square with your thinking?

PR #8338, opened 2021-07-29
include lang query parameter when fetching md

jwunderl:
Include query parameter for language when fetching markdown, even when it is packaged (part of: https://github.com/microsoft/pxt-microbit/issues/4286)

Also minor fix for md expiration which didn't matter before this -- in pxt electron the window.pxtElectron doesn't consistently exist immediately on load, so we need to calculate expiration inside the function instead of at the top level for it to expire local markdown properly (and properly use the cache for the translated ones)

### DIFF
--- a/pxtlib/emitter/cloud.ts
+++ b/pxtlib/emitter/cloud.ts
@@ -113,11 +113,12 @@ namespace pxt.Cloud {
         })
     }
 
-    // 1h check on markdown content if not on development server
-    const MARKDOWN_EXPIRATION = pxt.BrowserUtils.isLocalHostDev() ? 0 : 1 * 60 * 60 * 1000;
-    // 1w check don't use cached version and wait for new content
-    const FORCE_MARKDOWN_UPDATE = MARKDOWN_EXPIRATION * 24 * 7;
     export async function markdownAsync(docid: string, locale?: string): Promise<string> {
+        // 1h check on markdown content if not on development server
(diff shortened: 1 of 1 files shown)


livcheerful (approved):

dmonroym (approved):

pxtlib/emitter/cloud.ts
```diff
@@ -182,8 +183,8 @@ namespace pxt.Cloud {
         } else {
             url = `md/${pxt.appTarget.id}/${docid.replace(/^\//, "")}?targetVersion=${encodeURIComponent(targetVersion)}`;
         }
-        if (!packaged && locale != "en") {
-            url += `&lang=${encodeURIComponent(locale)}`
+        if (locale != "en") {
+            url += `${packaged ? "?" : "&"}lang=${encodeURIComponent(locale)}`
```
shakao, line 187: just to double-check, in the fully offline case this parameter is just ignored if the file isn't found?

shakao (approved):

pxtlib/emitter/cloud.ts
jwunderl: Yes, currently it is completely ignored, and with the change in pxt-electron to make use of the parameter the logic is basically 'if lang specified, try to fetch online and fall back to local copy'

abchatra (approved):

PR #8352, opened 2021-08-04
Fix delete project button

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/3676

Check projects by id, not index


eanders-MS (approved):
LGTM! There is a similar bug I think shares the same solution. Let me find it and pass it to you. If not same solution feel free to unassign it and we'll retriage.

PR #8357, opened 2021-08-05
Fix temporary asset duplication

shakao:
- don't set the display name on temporary assets at all
- some null checks
- move the default display name code to tilemap so it can be shared
- skips dispatching the change gallery view event if you're already on the correct (user assets) view, as it makes this very slow

fixes https://github.com/microsoft/pxt-arcade/issues/3068

### DIFF
--- a/webapp/src/components/assetEditor/assetSidebar.tsx
+++ b/webapp/src/components/assetEditor/assetSidebar.tsx
@@ -107,7 +107,7 @@ class AssetSidebarImpl extends React.Component<AssetSidebarProps, AssetSidebarSt
         pxt.tickEvent("assets.duplicate", { type: this.props.asset.type.toString(), gallery: isGalleryAsset.toString() });
 
         const asset = this.props.asset;
-        const displayName = asset.meta.displayName
+        const displayName = asset.meta?.displayName
             || getDisplayNameForAsset(asset, isGalleryAsset)
             || pxt.getDefaultAssetDisplayName(asset.type);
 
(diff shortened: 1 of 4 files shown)


webapp/src/components/assetEditor/assetSidebar.tsx
```diff
@@ -103,16 +103,19 @@ class AssetSidebarImpl extends React.Component<AssetSidebarProps, AssetSidebarSt
     }
 
     protected duplicateAssetHandler = () => {
-        pxt.tickEvent("assets.duplicate", { type: this.props.asset.type.toString(), gallery: this.props.isGalleryAsset.toString() });
+        const { isGalleryAsset } = this.props;
+        pxt.tickEvent("assets.duplicate", { type: this.props.asset.type.toString(), gallery: isGalleryAsset.toString() });
 
         const asset = this.props.asset;
-        if (!asset.meta?.displayName) asset.meta = { ...asset.meta, displayName: getDisplayNameForAsset(asset, this.props.isGalleryAsset) }
+        const displayName = asset.meta.displayName
```
jwunderl, line 110: Just to confirm, asset.meta can't be null here, right? it was doing a null propagation there before but if it was null it'd have crashed on the `...asset.meta` anyway

jwunderl (approved):

webapp/src/components/assetEditor/assetSidebar.tsx
shakao: ah there was a null check in the if statement before! i'll add it back here, thanks!

PR #8362, opened 2021-08-06
NavBar sizing fixes

livcheerful:
- Fixes login button sizing
- Collapses the docs button into settings earlier (bc the sign in button is fatter now)

fixes https://github.com/microsoft/pxt-arcade/issues/3731

image
image
image
image

### DIFF
--- a/theme/common.less
+++ b/theme/common.less
@@ -1699,6 +1699,20 @@ p.ui.font.small {
     }
 }
 
+/* Thin view, for navbar size management */
+@media only screen and (max-width: @mediumScreenWidth) {
+    .ui.thin.hide {
+        display: none!important;
+    }
+}
+
+@media only screen and (min-width: @mediumScreenWidth) {
+    .ui.thin.only {
+        display: none!important;
+    }
+}
+
+
 /* Custom width treshold to force vertical stacking on the start page */
 @media only screen and (max-width: 1095px) {
     .ui.modal.projectsdialog h2.editorname {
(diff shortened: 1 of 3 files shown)


theme/themes/pxt/globals/site.variables
```diff
@@ -55,6 +55,7 @@
 
 @tallEditorBreakpoint: 44rem;
 @thinEditorBreakpoint: 50rem;
+@mediumScreenWidth: 1313px;
```
eanders-MS, line 58: use rems?

eanders-MS (approved):

theme/common.less
```diff
@@ -1682,6 +1699,20 @@ p.ui.font.small {
     }
 }
 
+/* Thin view, for navbar size management */
+@media only screen and (max-width: @mediumScreenWidth) {
```
shakao, line 1703: hmm i would slightly prefer not to add another breakpoint (just because it's another view we'll have to remember to test), is it possible to reduce the padding on the button or the spacing between the sign in button and the settings gear at all?

theme/themes/pxt/globals/site.variables
```diff
@@ -55,6 +55,7 @@
 
 @tallEditorBreakpoint: 44rem;
 @thinEditorBreakpoint: 50rem;
+@mediumScreenWidth: 82rem;
```
shakao, line 58: small: can also be removed? otherwise lgtm!

shakao (approved):

PR #8368, opened 2021-08-06
Fix false-positive sync failures

eanders-MS:
The syncing of deleted projects was falsely detecting errors.
This is a quick fix. I'd like to clean up this code later to return a richer sync result.


shakao (approved):
maybe leave a TODO in saveTasks just so we know to come back to it?

PR #8372, opened 2021-08-07
Just use logoURL for skillmap logo

livcheerful:
This should fix the broken logo in the skillmap page, but I'm not sure how to test it :\ I can't uploadtrg bc my build is super mad about it, but the issue was it was trying to fetch the logo at:

/static/skillmap/https://pxt.azureedge.net/blob/d6139e3d0af51f02aa5f8765ecb0985acbd98551/static/Micorsoft_logo_rgb_W-white_D.png

bc I was using resolvePath and it was prepending /static/skillmap.

This doesn't fix the localhost serve logo. I could only ever get it to work if I dropped something in pxt/skillmap/public, which *feels* like a place I shouldn't be touching???? Any advice is appreciated

### DIFF
--- a/skillmap/src/components/HeaderBar.tsx
+++ b/skillmap/src/components/HeaderBar.tsx
@@ -52,9 +52,10 @@ export class HeaderBarImpl extends React.Component<HeaderBarProps> {
     protected getOrganizationLogo(targetTheme: pxt.AppTheme) {
         const logoUrl = targetTheme.organizationWideLogo;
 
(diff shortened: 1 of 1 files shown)


shakao:
for localserve you can update the cli to pull the image file you need into skillmap public/assets: https://github.com/microsoft/pxt/blob/master/cli/cli.ts#L2021 but not a blocker for this pr

shakao (approved):

livcheerful:
I'm going to commit this change now! But I'll put up another PR later today for local serve 👀

PR #8373, opened 2021-08-08
sign in dialog refresh

eanders-MS:
Old:
image

New:
image

Later today I'll draft a blog post and point the "Learn More" link to it. I'm thinking we can detail how to create a Microsoft Account from a gmail address or phone number there.

### DIFF
--- a/common-docs/identity/sign-in.md
+++ b/common-docs/identity/sign-in.md
@@ -1,0 +1,35 @@
+# Microsoft Accounts and MakeCode
+
+> Note: This feature only available in [**MakeCode Arcade**](https://arcade.makecode.com).
+
+## What is a Microsoft Account?
+
+A Microsoft Account, or MSA, is the personal user account you create when you want to sign into Microsoft services like outlook.com, Xbox, Skype, etc. Read more [here](https://aka.ms/AAdd6f8).
+
+## How does it work in MakeCode?
+
+A number of new features are coming that will only be available to signed-in users. These features need to know your identity so they can do things for you like store data in the cloud, and connect you to other users. The first of these features is one we're calling [Cloud Sync](/identity/cloud-sync.md). With Cloud Sync, when you're signed into MakeCode, we'll store your projects in the cloud where they're accessible from anywhere.
+
+## Is it required?
+
+No. Sign in only if you want to take advantage of features like Cloud Sync.
+
+## Does it cost money?
+
+Nope! Creating an MSA account is free. Using MakeCode is also free. As a reminder, MSA accounts work across a range of Microsoft products and services, so you can use it outside MakeCode. Some of these products and services may not be free (Minecraft Marketplace content, for example), but this is outside the MakeCode ecosystem.
+
+## What will other users see?
+
+Your signed in status is visible only to you. In the future we may add features that allow others to see your presence. When that time comes, we would ask you to create a MakeCode Profile so that your name and picture are decoupled from your MSA profile.
+
+## Do I already have a Microsoft Account?
+
+It is possible! Do you play Minecraft, game on an Xbox, subscribe to Office 365, or use any other Microsoft products or services? If so, you probably already have a Microsoft Account.
+
+## Does this mean I have to create a new email address?
+
+No. You can create an MSA using your current email address if it is not hosted by Microsoft (e.g.: GMail, Yahoo Mail, and others). When creating your MSA account, just enter your existing email address.
+
+## Can I use my school or work account?
+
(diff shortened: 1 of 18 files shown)


Jaqster:
Looks great!  I wonder if we should point the Learn More to a documentation page though instead of blog post?

Jaqster:
Maybe add a "Sign In" page under Miscellaneous?  @ganicke can you help?
image

eanders-MS:
Great idea! Should probably have one for Sign In and another for Cloud Sync. Will add them.

eanders-MS:
@ganicke Is it possible to filter docs by target? We would want to only show these docs pages in Arcade, but would be great to also link to them from Arcade's table of contents.

ganicke:
I guess the question is that you'd like them to go in as `common-docs` for use later with other targets but want to have it exposed now only for Arcade? It can go into the SUMMARY for Arcade but it could also be seen when someone inadvertently accesses `https://makecode.microbit.org/sign-in`. There is no per target doc filter. I think SUMMARY kind of is supposed to do that but it doesn't restrict the doc set.

We can just put it in Arcade docs and then move it down to pxt as an extended page when other targets will need it.

eanders-MS:
Ok, sounds good! Where do you suggest I put the files? pxt-arcade/docs/static?

theme/common.less
```diff
@@ -1915,33 +1915,8 @@ div.simframe.ui.embed {
...
-            justify-self: center;
-        }
-
-        div.providers {
-            display: grid;
-            row-gap: 20px;
-            grid-template-rows: auto;
-            justify-content: center;
-        }
+    div.remember-me {
```
livcheerful, line 1918: should this be generalize CSS for the actions added to the modal?

ganicke:
For any related doc pages, let's make a path under `./common-docs` such as:

https://github.com/microsoft/pxt/tree/master/common-docs/identity

-- or --

https://github.com/microsoft/pxt/tree/master/common-docs/accounts

This will allow a target to override the page if needs to include some additional info. We do that with extensible pages like https://github.com/microsoft/pxt/blob/master/common-docs/translate.md and some others.

eanders-MS:
@ganicke I've added docs with some seed content. Would love a review, and please make changes as you see fit!

ganicke (approved):

theme/common.less
eanders-MS: > should this be generalized CSS for the actions added to the modal?

Good call. Done.

common-docs/identity/sign-in.md
```diff
@@ -24,7 +24,7 @@ Your signed in status is visible only to you. In the future we may add features
 
 ## Do I already have a Microsoft Account?
 
-It is possible! Do you play Minecraft, game on an Xbox, subscribe to Office 365, or use any other Microsoft products or services? If so, you probably already have a Microsoft Account.
+It is possible! Do you play Minecraft, games on an Xbox, subscribe to Office 365, or use any other Microsoft products or services? If so, you probably already have a Microsoft Account.
```
eanders-MS, line 27: @ganicke I was using "game" as a verb here, making a loose affinity to gamer culture. "games" doesn't sound right to me.

common-docs/identity/sign-in.md
```diff
@@ -0,0 +1,35 @@
...
+
+Your signed in status is visible only to you. In the future we may add features that allow others to see your presence. When that time comes, we would ask you to create a MakeCode Profile so that your name and picture are decoupled from your MSA profile.
+
+## Do I already have a Microsoft Account?
+
+It is possible! Do you play Minecraft, games on an Xbox, subscribe to Office 365, or use any other Microsoft products or services? If so, you probably already have a Microsoft Account.
+
+## Does this mean I have to create a new email address?
+
+No. You can create an MSA using your current email address if it is not hosted by Microsoft (e.g.: GMail, Yahoo Mail, and others). When creating your MSA account, just enter your existing email address.
```
livcheerful, line 31: I think it should be styled as "Gmail" (at least that's what it looks like on their docs https://developers.google.com/gmail/design/css)

livcheerful (approved):

common-docs/identity/sign-in.md
eanders-MS: I went ahead and reverted this to "game". If you like we can reword it more generally, but in this context "game" felt right to me :).

ganicke: I see. "game" or maybe "do gaming on Xbox".

PR #8377, opened 2021-08-09
Show Sign In modal directly if no Github

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/3760


eanders-MS (approved):
Might be able to generalize this by inspecting dropdown's children, but this is fine for now!

dmonroym (approved):

PR #8384, opened 2021-08-10
Always keep the skillmap iframe in the DOM

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/3747

This changes the skillmap to always keep the iframe around in the background. When no tutorial is open, the editor unloads the current project and has its visibility hidden until the next activity is opened.

In the meantime, you can send whatever messages you want to the editor. For sharing, i'm doing it through the redux state (see makecodeframe's componentDidUpdate) but @eanders-MS I expect you might want to use the frame directly. Easiest way to do that is to probably pass a message channel of some kind from the makecodeframe in App.tsx

Not sure if this will improve perf at all, but it might! I think @shakao tried out something similar in the past but the gains weren't huge.

### DIFF
--- a/pxteditor/editor.ts
+++ b/pxteditor/editor.ts
@@ -253,6 +253,7 @@ namespace pxt.editor {
         updateFileAsync(name: string, content: string, open?: boolean): Promise<void>;
 
         openHome(): void;
+        unloadProjectAsync(home?: boolean): Promise<void>;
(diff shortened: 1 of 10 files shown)


dmonroym (approved):

eanders-MS (approved):
nice work

shakao (approved):

PR #8397, opened 2021-08-12
Fix the editor toolbar project name to use local state

riknoll:
The project name input in the editor toolbar uses the project name that comes from the parent editor state, which means that every time you do a keystroke the effect of that keystroke needs to propagate from the top down. That makes editing laggy and sometimes causes characters to be missed or the cursor to jump around. Fixing it by using local state.

This bug is easier to see if you open a big project and trigger a compile (i.e. drag a block) right before you start typing in the input.

I don't think this is filed anywhere but it has been annoying me a ton lately. Not a regression but should probably get cherry picked for either this release or next.

### DIFF
--- a/webapp/src/editortoolbar.tsx
+++ b/webapp/src/editortoolbar.tsx
@@ -543,29 +543,60 @@ interface EditorToolbarSaveInputProps extends React.DetailedHTMLProps<React.Inpu
     onChangeValue: (value: string, view: string) => void;
 }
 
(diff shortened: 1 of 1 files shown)


eanders-MS (approved):
Could this be generalized to all text input controls?

shakao (approved):

riknoll:
I might take a look at unifying this in the future, but not for this release

PR #8398, opened 2021-08-12
Handle iframe reload in skillmap

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/3782

When the iframe is loading again, kick us back into the loading state for the makecodeFrame. I tested this with an automatic reload (suggested by richard!!) and not the scenario in the bug, but I *think* it should work. 

I will see if an uploadtrg might trigger the bug, and if it does, I'll try testing it there


shakao (approved):

eanders-MS (approved):
Seems ok!

abchatra (approved):
Looks good

PR #8405, opened 2021-08-12
Don't run sim in background when no project is loaded in the skillmap iframe

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/3794

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -3979,7 +3979,6 @@ export class ProjectView
                 return this.exitTutorialAsync()
                     .finally(() => {
                         core.hideLoading("leavingtutorial")
-                        this.postTutorialProgress();
                     })
             }
             else {
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -3975,22 +3975,31 @@ export class ProjectView
             return Promise.resolve(); // TODO cleanup
         }
         else {
-            return this.exitTutorialAsync()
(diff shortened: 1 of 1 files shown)


webapp/src/app.tsx
```diff
@@ -3971,22 +3975,31 @@ export class ProjectView
...
-                        pxt.appTarget.appTheme.shareFinishedTutorials) {
-                        pxt.tickEvent("tutorial.share", undefined, { interactiveConsent: false });
-                        this.showShareDialog(lf("Well done! Would you like to share your project?"));
-                    }
-                })
+            if (pxt.BrowserUtils.isSkillmapEditor()) {
+                return this.exitTutorialAsync()
+                    .finally(() => {
+                        core.hideLoading("leavingtutorial")
+                        this.postTutorialProgress();
```
shakao, line 3982: do we need this? we're posting the tutorial completed event above. could we just use the same IE check above?

abchatra:
@riknoll I am still waiting for this. Can you address review comment from Shannon?

webapp/src/app.tsx
riknoll: Can't use the same check because the IE one opens the home screen. I'll fix the progress event

shakao (approved):

dmonroym (approved):

eanders-MS (approved):

PR #8411, opened 2021-08-13
Add user profile page

livcheerful:
Adds the User Profile page to skillmap. It looks basically the same as the regular webapp one!

Also I fixed the local serve of the MS logo in skillmap! Will put up pxt-arcade changes for that


skillmap/src/components/AppModal.tsx
```diff
@@ -347,17 +350,54 @@ export class AppModalImpl extends React.Component<AppModalProps, AppModalState>
...
+        const { checkboxSelected } = this.state;
+
+        const buttons = [];
+        buttons.push({
+            label: lf("Confirm"),
+            className: checkboxSelected ? "confirm" : "confirm disabled",
+
+            onClick: async () => {
+                if (checkboxSelected) {
+                    tickEvent("skillmap.profile.delete", { link: "https://aka.ms/cloudsave" });
```
shakao, line 375: why do we need this link in the tick event?

skillmap/src/components/UserProfile.tsx
```diff
@@ -0,0 +1,130 @@
...
+        );
+
+        return <div className="account panel ui card">
+            <div className="header-text">
+                <label>{lf("Profile")}</label>
+            </div>
+            {profile?.idp?.picture?.dataUrl ? avatarElem : initialsElem}
+            <div className="row-span-two">
+                <label className="title">{lf("Name")}</label>
+                <p className="value">{profile?.idp?.displayName}</p>
```
shakao, line 62: is it possible that the displayName/username/provider.name values will be undefined? (if not then disregard!) otherwise should we have some kind of check?

skillmap/src/styles/modal.css
```diff
@@ -60,6 +128,22 @@
     text-decoration: none;
 }
 
+.modal-button.disabled {
+    background-color: var(--inactive-color);
+    color: var(--inactive-hover-color);
+    border: 1px solid var(--inactive-hover-color);
+}
+
+.modal-button.disabled:hover {
```
shakao, line 137: should we have hover changes when the button is disabled?

shakao (approved):
some small comments, feel free to address/disregard as you see fit. otherwise lgtm!

eanders-MS (approved):
Looking good!

skillmap/src/components/AppModal.tsx
eanders-MS: I think it was just copied from the tick in webapp. Probably don't need it in either place. Not useful data.

skillmap/src/components/UserProfile.tsx
eanders-MS: It is possible for displayName to be undefined. Username and provider should always be populated. I think a lot of this was just lifted from webapp.

skillmap/src/components/UserProfile.tsx
```diff
@@ -0,0 +1,130 @@
...
+interface UserProfileProps {
+    signedIn: boolean;
+    profile: pxt.auth.UserProfile
+    showProfile: boolean;
+    dispatchCloseUserProfile: () => void;
+    dispatchShowDeleteAccountModal: () => void;
+}
+
+
+export class UserProfilImpl extends React.Component<UserProfileProps, {}> {
```
eanders-MS, line 18: spelling: UserProfileImpl?

skillmap/src/components/UserProfile.tsx
```diff
@@ -0,0 +1,130 @@
...
+                <p className="value">{profile?.idp?.username}</p>
+            </div>
+            <div className="row-span-two">
+                <label className="title">{lf("Provider")}</label>
+                <p className="value">{provider?.name}</p>
+            </div>
+            <div className="row-span-two" onClick={this.handleSignout}>
+                <div className="sign-out ui icon button" >
+                    <span className={`xicon ${profile?.idp?.provider}`} />
+                    <span> {lf("Sign out")} </span>
```
eanders-MS, line 75: I ran this locally and this sign out button looked like the buttons from the webapp. Felt inconsistent with the rest of skillmap's design. (we should bring them together)
image

skillmap/src/components/AppModal.tsx
livcheerful: ! i will remove in both

skillmap/src/components/UserProfile.tsx
livcheerful: oh yeah I just copied. I'll give this and webapp a pass for undefined ty for noticing!!

skillmap/src/components/UserProfile.tsx
livcheerful: yeah, but I thought since the header between skillmap/webapp are the same, user wouldn't really know the difference between skillmap and regular app? So it would be confusing if the profile looked different from before

livcheerful: I could also change the regular app to match skillmap buttons?

eanders-MS: My 2c: I think it would be best for each to be consistent with its own style. Ideally they would share one common control lib and style sheet. But I don't know if that will ever happen due to the amount of reengineering.

shakao: yeah most buttons in webapp already use the same component, we should at some point do the same in the skillmap--just add a button component and ensure that any buttons in the skillmap are using that component (shouldn't be TOO bad since the skillmap is small). at least this'll make it easier longer-term, if we eventually want to unify the two components or something

shakao: maybe for this PR we could just make the user profile buttons look like the other skillmap buttons?

PR #8413, opened 2021-08-13
blog link for tweeted url

eanders-MS:
Tweeted link was truncated to https://makecode.com/blog/arcade/in, so let's put the blog post there.


ElizabethRiffle (commented):
Thanks for doing this!

PR #8426, opened 2021-08-19
Document existing docs and tutorial snippets

shakao:
i went through the tutorial markdown parser and the docs renderer code to get a list of the snippets we handle and tested them to double-check their behavior! this is HOPEFULLY a comprehensive list of everything our markdown can do

@ganicke there's some overlap with the writing-tutorials docs here, i tried to link back to avoid duplicate documentation but i'm happy to rearrange or split this up as you see fit!!

@kiki-lee @anzhou5 i know this is super long so no need to read all of it but it should be mostly summarized in the compatibility tables at the top of each section--if i'm missing anything or there are any gaps in "general markdown documentation" that you'd have found useful when onboarding just let me know!


shakao:
thanks so much @ganicke!!

PR #8456, opened 2021-09-01
Make copies of local projects when importing skillmap progress to cloud

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/3861
Fixes https://github.com/microsoft/pxt-arcade/issues/3858

Makes a few changes to preserve the local user state when logging in on the skillmap page:

1. When syncing local skillmap headers to the cloud, syncs duplicates instead of the original
2. Local user state is no longer overwritten when logged in
3. Always copies state from the local user and not whatever is in the store

The copy only happens for maps that the signed in user has no progress in.

### DIFF
--- a/skillmap/src/App.tsx
+++ b/skillmap/src/App.tsx
@@ -5,7 +5,7 @@ import { connect } from 'react-redux';
 
 import store from "./store/store";
 import * as authClient from "./lib/authClient";
-import { getFlattenedHeaderIds as getHeaderIdsForUnstartedMaps } from "./lib/skillMapUtils";
+import { getHeaderIdsForUnstartedMaps } from "./lib/skillMapUtils";
 
 import {
     dispatchAddSkillMap,
--- a/skillmap/src/App.tsx
+++ b/skillmap/src/App.tsx
@@ -5,7 +5,7 @@ import { connect } from 'react-redux';
 
 import store from "./store/store";
 import * as authClient from "./lib/authClient";
-import { getHeaderIdsForUnstartedMaps } from "./lib/skillMapUtils";
+import { getFlattenedHeaderIds } from "./lib/skillMapUtils";
 
 import {
     dispatchAddSkillMap,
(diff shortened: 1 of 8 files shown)


jwunderl (approved):

skillmap/src/lib/skillMapUtils.ts
```diff
@@ -160,9 +160,10 @@ export function isRewardNode(node: MapNode) {
     return node.kind === "reward" || node.kind === "completion";
 }
 
-export function getFlattenedHeaderIds(user: UserState, pageSource: string): string[] {
+export function getFlattenedHeaderIds(user: UserState, pageSource: string, existing: UserState): string[] {
```
livcheerful, line 163: Does this only get the ids for unstarted projects in the cloud? should this be called getHeaderIdsForUnstartedMaps ?

riknoll: yeah, i guess "rename symbol" was trying to be clever and renamed the import but not the definition. Fixed!

skillmap/src/App.tsx
```diff
@@ -211,13 +211,79 @@ class AppImpl extends React.Component<AppProps, AppState> {
...
+                        }
+                    }
+                }
+
+                this.props.dispatchSetUser(newUser)
+                await saveUserStateAsync(newUser);
+            }
+
             // Tell the editor to send us the cloud status of our projects.
             await this.sendMessageAsync({
```
livcheerful, line 288: I think with the headerId changes, we won't fetch the status of our....already synced projects ever? I think this should request the status of all headerIds, not just the new ones?

riknoll: true! fixed

skillmap/src/lib/workspaceProvider.ts
```diff
@@ -57,14 +61,7 @@ export async function saveUserStateAsync(user: UserState): Promise<void> {
...
     const ws = await getWorkspaceAsync();
-    if (await pxt.auth.client()?.loggedInAsync()) {
-        // When signed in, clear locally saved progress.
-        await ws.saveUserStateAsync({
-            ...user,
-            mapProgress: { },
-            completedTags: { }
-        });
-    } else {
+    if (!(await pxt.auth.client()?.loggedInAsync())) {
```
eanders-MS, line 64: minor: this could be simplified to `if (!(await authClient.loggedInAsync()))`

eanders-MS (approved):
Looks great.

PR #8459, opened 2021-09-01
Display tutorial step title, formatted bullets for markdown

shakao:
tutorial-md

tutorial markdown styling! and an unrelated bug fix (https://github.com/microsoft/pxt/commit/ade0d76dd96e2f47f91503d93c7b46006a9d9fa5) sandwiched in there.

this displays every step title by default. the syntax for hiding a title is `{STEP TITLE HERE}`. the plan is to run a script to hide the titles on all our existing content, or do a manual pass to ensure the titles are useful--will depend on time/bandwidth

general remaining work:
- hint CSS changes from design office hours
- tutorial in headless targets (minecraft)
- special cases: flyout only, locked editor, etc


theme/tutorial-sidebar.less
```diff
@@ -30,6 +32,8 @@
     margin: 1rem 0;
     overflow-x: hidden;
     overflow-y: auto;
+    font-family: "Segoe UI",Tahoma,Geneva,Verdana;
```
riknoll, line 35: should define a variable for this

webapp/src/marked.tsx
```diff
@@ -303,6 +303,40 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
...
+                    li.innerHTML = "";
+                    li.appendChild(p);
+                }
+            })
+
+        pxt.Util.toArray(content.querySelectorAll("li > p"))
+            .forEach((p: HTMLElement) => {
+                const match = p.innerText.match(bulletRegex);
+                if (match?.[1]) {
+                    p.innerHTML = p.innerHTML.replace(bulletRegex, "");
```
riknoll, line 323: you can probably just run this on the first text node child to get around this warning

webapp/src/marked.tsx
```diff
@@ -303,6 +303,40 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
...
+
+                    const li = p.parentElement;
+                    li.className += " formatted-bullet-li";
+                    const ul = li.parentElement;
+                    if (!ul.classList.contains("formatted-bullet-ul")) ul.className += " formatted-bullet-ul";
+
+                    const bullet = document.createElement("div");
+                    bullet.className = "formatted-bullet";
+                    const icon = document.createElement("i");
+                    icon.className = `ui icon ${match[1]}`;
```
riknoll, line 333: do we need to sanitize this? I think we should allowlist the semantic icons (must be a list of them somewhere). Otherwise some clever folks might abuse this to thwart our formatting

riknoll (approved):

webapp/src/marked.tsx
riknoll: e.g. they could put in "red cubes" and change the color

PR #8464, opened 2021-09-02
Speed up the clone operation for api info and apptarget

riknoll:
I was doing some profiling on performance and noticed we spend a lot of time in our clone function in util.ts. That clone function just looks like this:

```ts
    export function clone<T>(v: T): T {
        if (v == null) return null
        return JSON.parse(JSON.stringify(v))
    }
```

Turns out, the vast majority of the time spent cloning was in two places: cloning the apptarget when initially setting it up and cloning the cached api info in webapp/src/compiler.ts. We were basically cloning the entirety of target.js (a 10.1 MB file in pxt-arcade) multiple times!

This PR speeds things up considerably by implementing functions for specifically cloning the SymbolInfo and TargetBundle types. In my testing, it cut it down to about 1/3 of what it was before when setting up the AppTarget (I didn't time the cached API info changes but there should be similar gains)

### DIFF
--- a/pxtlib/main.ts
+++ b/pxtlib/main.ts
@@ -48,7 +48,7 @@ namespace pxt {
     export function setAppTarget(trg: TargetBundle) {
         appTarget = trg || <TargetBundle>{};
         patchAppTarget();
-        savedAppTarget = U.clone(appTarget)
+        savedAppTarget = U.cloneTargetBundle(appTarget)
     }
 
     let apiInfo: Map<PackageApiInfo>;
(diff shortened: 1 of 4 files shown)


riknoll:
We could speed this up even further if we treated the TargetBundle and SymbolInfo types as readonly (and avoid cloning altogether), but that would be a much bigger change. The localization and hardware variant code relies pretty heavily on cloning

jwunderl (approved):

dmonroym (approved):

abchatra:
This is an awesome find!

PR #8470, opened 2021-09-03
cleanup cloudSyncCheckAsync trigger

eanders-MS:
The way I was determining whether cloudSyncCheckAsync should execute was gross. This cleans it up (hopefully!) into something more clear and maintainable (hopefully!).


eanders-MS:
Will reopen against stable7.1

PR #8473, opened 2021-09-07
Align icons in skillmap firefox

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/3859

The units in firefox are just different. Both are using pixels, but firefox needs more? I tried some other things but nothing worked for me other than special casing it. If there's a better way lemme know 😓

### DIFF
--- a/skillmap/src/lib/browserUtils.ts
+++ b/skillmap/src/lib/browserUtils.ts
@@ -283,38 +283,3 @@ export function tickEvent(id: string, data?: { [key: string]: string | number })
 
     (window as any).pxtTickEvent?.(id, data);
 }
-
-export function hasNavigator(): boolean {
-    return typeof navigator !== "undefined";
-}
-
-// The browser matching code is the same as pxtlib/browserutils.ts
-//Microsoft Edge lies about its user agent and claims to be Chrome, but Microsoft Edge/Version
-//is always at the end
-export function isEdge(): boolean {
-    return hasNavigator() && /Edge/i.test(navigator.userAgent);
-}
-
-//IE11 also lies about its user agent, but has Trident appear somewhere in
-//the user agent. Detecting the different between IE11 and Microsoft Edge isn't
-//super-important because the UI is similar enough
-export function isIE(): boolean {
-    return hasNavigator() && /Trident/i.test(navigator.userAgent);
-}
-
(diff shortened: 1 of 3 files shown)


skillmap/src/lib/browserUtils.ts
```diff
@@ -282,4 +282,39 @@ export function tickEvent(id: string, data?: { [key: string]: string | number })
     data["pageTitle"] = pageTitle;
 
     (window as any).pxtTickEvent?.(id, data);
-}
\ No newline at end of file
+}
+
+export function hasNavigator(): boolean {
```
eanders-MS, line 287: why copy these methods? pxtlib's browserutils is accessible from skillmap.

livcheerful: lol I forgot. i'll fix it

eanders-MS (approved):

riknoll:
can i get a test build? this also often breaks on mac vs pc

livcheerful:
yeah i'll spin one up! i tried to do one just now and saw something weird, so it might be a bit...but i'm working on it!

livcheerful:
here it is!! https://arcade.makecode.com/app/87b96eebfb815fa52e9bd07975590a9647845449-d68d14a157---skillmap 

lemme know if you see bad things on mac :O

riknoll:
Yeah, they're pretty off on mac:

Screen Shot 2021-09-08 at 9 15 48 AM

livcheerful:
ahhhhhh how do they look normally?

riknoll:
exactly the same! so i'm guessing the check is just failing on mac. In any case, not a regression so feel free to merge

shakao:
does adding `dominant-baseline="middle"` to the text nodes do anything? (based on https://stackoverflow.com/questions/44545006/svg-shows-different-on-firefox-and-edge) 

also definitely a lingering bug from when this was first written, but the dy/yOffset should be calculated from the width/selectedUnit probably

livcheerful:
ooo yeah it does i'll push it. Richard i'll also make another upload for mac??

livcheerful:
https://arcade.makecode.com/app/497e4b372fbf2a00029b6003585e4f05603e2a05-d864d31504---skillmap

riknoll:
no change

riknoll:
The "isFirefox" check returns false on mac (I just debugged it)

shakao:
(i think maybe that build doesn't have the latest changes? it isn't using the isFirefox check any more)

livcheerful:
whatttt that upload shouldn't use isfirefox though. maybe i uploaded wrong hmmmmm

PR #8474, opened 2021-09-07
[stable7.1] Expand inline blocks regex to allow arrays (#8471)

shakao:
hoping to get this in with the upcoming skillmap fixes, angel is writing some documentation that uses arrays and this has been causing issues


shakao:
also cherry-picking 
- https://github.com/microsoft/pxt/pull/8424
- https://github.com/microsoft/pxt/pull/8478

for kiki

abchatra (approved):

PR #8476, opened 2021-09-07
cleanup cloudSyncCheckAsync trigger

eanders-MS:
The way I was determining whether cloudSyncCheckAsync should execute was gross. This cleans it up (hopefully!) into something more clear and maintainable (hopefully!).

Targeting stable7.1 branch directly. We can merge stable7.1 back to master after we ship it.

### DIFF
--- a/skillmap/src/App.tsx
+++ b/skillmap/src/App.tsx
@@ -66,18 +66,62 @@ interface AppState {
     error?: string;
 }
 
+interface ReadyResources {
+    sendMessageAsync?: (message: any) => Promise<any>;
+}
+
+class ReadyPromise {
+    private promise_: Promise<ReadyResources>
+    private resources: ReadyResources;
+    private mounted?: boolean;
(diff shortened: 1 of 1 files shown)


eanders-MS:
Bumping. Can someone please review this PR?

skillmap/src/App.tsx
```diff
@@ -66,19 +66,63 @@ interface AppState {
     error?: string;
 }
 
+interface ReadyResources {
```
shakao, line 69: (for later/when making this change on master) could this class be moved out of app.tsx?

shakao (approved):

skillmap/src/App.tsx
eanders-MS: That might make sense. It really is just an api to the editor that isn't available right away.

abchatra:
@eanders-MS this needs to be in the master as well.

PR #8479, opened 2021-09-08
Cleanup console output

eanders-MS:
Fixes https://github.com/microsoft/pxt-arcade/issues/3912


jwunderl (approved):

abchatra:
@eanders-MS can we port this to master? I prefer first check-in to master and then backport to stable. It is easy to forget the other way.

eanders-MS:
Master is diverging a lot from stable. It's simpler and faster to just work in stable for now, then do a single merge back to master after we ship it. Cherry-picking every change and retesting in stable is more overhead.

abchatra:
Ok sounds good. As long as we don't lose the changes it is ok.

PR #8481, opened 2021-09-08
Align sign in button icon

livcheerful:
looks like this now 
image

before it was like... 
image

but with the right icons lol


eanders-MS (approved):
LGTM

PR #8482, opened 2021-09-09
skillmap: export cloud projects to local on profile delete

eanders-MS:
When the user deletes their profile from within skillmap, we need to export all their cloud projects to local. This PR does that, but does not export their skillmap state to local. I'll handle that in a future PR. Exporting regular projects is the high-order bit (from a compliance perspective).

Fixes https://github.com/microsoft/pxt-arcade/issues/3878


eanders-MS:
Test url: https://arcade.makecode.com/app/bc663a757be03a43980b587f26922c326282ebab-40e16a4b41

webapp/src/cloud.ts
```diff
@@ -353,7 +353,7 @@ function getLocalCloudHeaders(allHdrs?: Header[]) {
     if (!auth.hasIdentity()) { return []; }
     if (!auth.loggedIn()) { return []; }
     const userId = auth.userProfile()?.id;
-    return (allHdrs || workspace.getHeaders(true))
+    return (allHdrs || workspace.getHeaders(true/*withDeleted*/, false/*filterByEditor*/))
```
eanders-MS, line 356: This code was filtering out non-skillmap projects when the editor was loaded in the skillmap iframe. But the cloud was sending down all headers, not just skillmap activities. This resulted in an insidious issue where the client would think all non-skillmap remote projects were new projects and try to import them. This would fail with couchdb conflicts.

skillmap/src/lib/readyResources.ts
```diff
@@ -0,0 +1,50 @@
+import { dispatchSetReadyResources } from "../actions/dispatch";
+import store from "../store/store";
+
+export class ReadyResources {
```
eanders-MS, line 4: @shakao I factored this out to its own file :). I don't love the name though.

eanders-MS:
Bumping. This needs to go to stable prior to testing at 10am today.

shakao:
hmm should it also be converting skillmap progress to local?

eanders-MS:
> hmm should it also be converting skillmap progress to local?

It should. I will do that in a separate PR. This was the min bar for compliance.

shakao:
ah! sorry you totally said that in the PR description and i completely missed it, my bad

shakao (approved):

PR #8483, opened 2021-09-09
Copy skillmap headers over for cloud users

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/3898

The issue was multilayered 🗻 --

🌸 First, this was only happening for completely new users who have had not visited skillmap before, which is why I could repro it by deleting the account, but not Eric by resetting progress in the skillmap.

🌸 Second, there was an issue where in cloudSyncCheckAsync, when we were figuring out what project headers we wanted to upload, it was always an empty array. 
https://github.com/microsoft/pxt/blob/290c02b2143b3dc9cd3f27762d37e61a6654be09/skillmap/src/App.tsx#L224-L231
This was an empty array because there was a race(?) condition where we were writing the local user to the cloud user in onStoreChange
https://github.com/microsoft/pxt/blob/290c02b2143b3dc9cd3f27762d37e61a6654be09/skillmap/src/App.tsx#L391-L399
So we weren't uploading a duplicated copy of the header because there was no duplicate.

🌸 Third, when the user hadn't visited skillmaps before, their user preferences just...didn't have an entry for their skillmap progress. But we were patching their progress with local progress when fetching the user state. We don't want to do the patching here, we'll do the patching in cloudSyncCheckAsync
https://github.com/microsoft/pxt/blob/290c02b2143b3dc9cd3f27762d37e61a6654be09/skillmap/src/lib/workspaceProvider.ts#L41-L57

🌸 Fourth, the way we were keying into what skillmaps to upload were by what the cloud user had visited

...I'll add inline comments to put what fixes what. I don't know if this is readable sorry 😓


skillmap/src/App.tsx
```diff
@@ -222,72 +226,62 @@ class AppImpl extends React.Component<AppProps, AppState> {
...
                 headerIds
             } as pxt.editor.EditorMessageSaveLocalProjectsToCloud)).resp.headerIdMap;
-
             if (headerMap) {
-                headerIds = headerIds.map(h => headerMap[h] || h);
-
-                // Patch all of the header ids in the user state and copy
-                // over the local progress that doesn't exist in the signed in
-                // user
-                const urls = Object.keys(currentUser.mapProgress);
```
livcheerful, line 239: I think we should be checking things based on which urls we've visited as local users. Otherwise we might miss something

skillmap/src/App.tsx
```diff
@@ -222,72 +226,62 @@ class AppImpl extends React.Component<AppProps, AppState> {
...
-                                const base = signedInActivity || localActivity;
-                                activityState[activity] = {
-                                    ...base,
-                                    headerId: localActivity.headerId ? (headerMap[localActivity.headerId] || localActivity.headerId) : localActivity.headerId
-                                };
-                            }
+                const visitedUrls = Object.keys(currentUser.mapProgress)
+                // Copy progress from cloud user for all visited URLs.
+                for (const url of visitedUrls) {
+                    if (hasUrlBeenStarted(currentUser, url)) {
```
livcheerful, line 277: Let the cloud progress override the local progress for the user.

skillmap/src/App.tsx
```diff
@@ -222,72 +226,62 @@ class AppImpl extends React.Component<AppProps, AppState> {
...
+                            const activityState: {[index: string]: ActivityState} = {};
+                            newUser.mapProgress[url][map].activityState = activityState;
+
+                            const localProgress = localUser.mapProgress[url][map].activityState
+                            for (const activity of Object.keys(localProgress)) {
+                                const localActivity = localProgress[activity];
+                                if (localActivity.headerId) {
+                                    activityState[activity] = {
+                                        ...localActivity,
+                                        headerId: headerMap[localActivity.headerId] || localActivity.headerId
```
livcheerful, line 266: I think...the headerId should theoretically ALWAYS be in the headerMap. But I don't want to bet on it, so I copied a lot of the old logic here for safety...

skillmap/src/App.tsx
```diff
@@ -300,6 +289,7 @@ class AppImpl extends React.Component<AppProps, AppState> {
                 headerIds: getFlattenedHeaderIds(currentUser, state.pageSourceUrl)
             } as pxt.editor.EditorMessageRequestProjectCloudStatus);
         }
+        this.readyPromise.cloudSyncCheckHasFinished = true;
```
livcheerful, line 292: This is the unblock for onStoreChange.

skillmap/src/lib/skillMapUtils.ts
```diff
@@ -111,6 +111,20 @@ export function lookupActivityProgress(user: UserState, pageSource: string, mapI
     return lookupMapProgress(user, pageSource, mapId)?.activityState[activityId]
 }
 
+export function hasUrlBeenStarted(user: UserState, pageSource: string): boolean {
```
livcheerful, line 114: The logic for this was kinda complicated, so I threw it into a helper here.

skillmap/src/lib/workspaceProvider.ts
```diff
@@ -45,13 +45,15 @@ export async function getUserStateAsync(): Promise<UserState> {
...
-        ...userState,
-        mapProgress: skillmapState?.mapProgress ?? userState.mapProgress,
-        completedTags: skillmapState?.completedTags ?? userState.completedTags
-    };
+    if (await authClient.loggedInAsync()) {
+        // Read synchronized skillmap state from cloud profile. Fallback to workspace-saved state.
+        const skillmapState = await authClient.getSkillmapStateAsync();
+        userState = {
+            ...userState,
+            mapProgress: skillmapState?.mapProgress || {},
```
livcheerful, line 53: Since we're now defaulting the user preferences to always have an empty mapProgress, this shouldn't be necessary. But for safety...

skillmap/src/lib/workspaceProvider.ts
```diff
@@ -45,13 +45,15 @@ export async function getUserStateAsync(): Promise<UserState> {
     // when auth is enabled.
     let userState = await getLocalUserStateAsync();
 
-    // Read synchronized skillmap state from cloud profile. Fallback to workspace-saved state.
-    const skillmapState = await authClient.getSkillmapStateAsync();
-    userState = {
-        ...userState,
-        mapProgress: skillmapState?.mapProgress ?? userState.mapProgress,
```
livcheerful, line 52: This was what was copying the local user to the cloud user when we hadn't visited skillmaps before

webapp/src/cloud.ts
```diff
@@ -598,6 +598,7 @@ export async function saveLocalProjectsToCloudAsync(headerIds: string[]) {
         for (const h of headers) {
             const newHeader = await workspace.duplicateAsync(h, h.name);
             guidMap[h.id] = newHeader.id;
+            newHeaders.push(newHeader);
```
livcheerful, line 601: also we want to actually sync the headers

skillmap/src/lib/readyResources.ts
```diff
@@ -16,10 +16,12 @@ export class ReadyPromise {
     private promise_: Promise<ReadyResources>
     private resources: ReadyResources;
     private appMounted?: boolean;
+    public cloudSyncCheckHasFinished: boolean
```
eanders-MS, line 19: Odd to have `cloudSyncCheckHasFinished` in here. It isn't part of the ready check. It's only used by App.tsx. I think it could live on AppImpl.

webapp/src/cloud.ts
eanders-MS, line 601: Oh this must have gotten dropped. Good catch.

eanders-MS:
I pulled this branch to try it out, and I think it has a regression: my local progress isn't being imported when I sign in. Repro steps:
- Delete profile
- Reset progress
- Start an activity
- Sign in
- Notice progress wasn't imported

Working as expected in /stable.

livcheerful:
Here is an uploadtrg to try out https://arcade.makecode.com/app/02497ed2bc68eb3139615162d0ba255fb5746b3d-fefeaf11fe

eanders-MS:
> Here is an uploadtrg to try out https://arcade.makecode.com/app/02497ed2bc68eb3139615162d0ba255fb5746b3d-fefeaf11fe

Import is working reliably there! I can't repro the delete profile bug either. Nice.

livcheerful:
cool cool cool that's the dreeeeaaaam

eanders-MS (approved):
Seems fine in local testing (on the target you linked). It's a big enough change that I think we should do a quick 2nd round of testing @abchatra.

skillmap/src/App.tsx
```diff
@@ -222,68 +222,57 @@ class AppImpl extends React.Component<AppProps, AppState> {
...
+                // over the local progress that doesn't exist in the signed in
+                // user
+                const localUrls = Object.keys(localUser.mapProgress);
+                for (const url of localUrls) {
                     newUser.mapProgress[url] = {
-                        ...currentUser.mapProgress[url],
-                    };
-
-                    if (!localUser.mapProgress[url]) continue;
+                        ...localUser.mapProgress[url]
```
riknoll, line 243: shouldn't this be currentuser?

skillmap/src/App.tsx
riknoll: there is an (admittedly very brief) time where this can happen. for example, if they refresh when loading a tutorial

skillmap/src/App.tsx
livcheerful: my logic is that we copy the local progress, and overwrite with the cloud one at the end if it exists?

riknoll: oh, can you add a comment explaining that?

skillmap/src/lib/skillMapUtils.ts
```diff
@@ -111,6 +111,20 @@ export function lookupActivityProgress(user: UserState, pageSource: string, mapI
...
+    // The user has no progress on the given page if:
+    // 1. They've never visited any skillmaps before
+    // 2. They've never visited this skillmap page before
+    // 3. They've visited this page, but have reset the skillmap state.
+    if (!user.mapProgress
+        || !user.mapProgress[pageSource]
+        || Object.keys(user.mapProgress[pageSource]).length == 0
+        || Object.keys(user.mapProgress[pageSource]).filter(mapId => Object.keys(user.mapProgress[pageSource][mapId].activityState).length != 0).length == 0) {
```
riknoll, line 122: nitpick: you can use `.some()` instead of `.filter().length === 0`

riknoll (approved):

jwunderl (approved):

abchatra:
Great job @livcheerful

eanders-MS:
Agreed, nice sleuthing.

PR #8485, opened 2021-09-10
Use arcade-sim icons.css :)

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/3925

the icons.css in the sim also has....semantic icons at the end. I can't figure out how to append that in our build, so I'm just gonna update icons.css in arcade-sim.

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -717,7 +717,6 @@ function targetFileList() {
     if (simDir() != "sim")
         lst = lst.concat(nodeutil.allFiles(path.join("sim", "public"), 5, true))
     pxt.debug(`target files (on disk): ${lst.join('\r\n    ')}`)
-    lst = lst.filter(f => !f.includes("icons.css")) // We want to pull icons.css from pxt-core, not the target
     return lst;
 }
 


cli/cli.ts
```diff
@@ -717,7 +717,6 @@ function targetFileList() {
     if (simDir() != "sim")
         lst = lst.concat(nodeutil.allFiles(path.join("sim", "public"), 5, true))
     pxt.debug(`target files (on disk): ${lst.join('\r\n    ')}`)
-    lst = lst.filter(f => !f.includes("icons.css")) // We want to pull icons.css from pxt-core, not the target
```
eanders-MS, line 720: Cloud icon still there?

livcheerful: yup with https://github.com/microsoft/pxt-arcade-sim/pull/50

eanders-MS: Oh! Do we need to do anything special to bump pxt-arcade-sim?

livcheerful: hmmmmmmm i think @riknoll knows? not sure what being an unpublished npm package means...

riknoll: You need to manually do a bump (change the version in package.json and push a tag).

Then you update the version in pxt-arcade in three places: The package.json and in tow of the github actions builds. Here's an example PR: https://github.com/microsoft/pxt-arcade/pull/3034/files

eanders-MS (approved):

PR #8491, opened 2021-09-15
Running npm audit fix --force

abchatra:


riknoll (approved):

riknoll:
hmmmm... looks like this broke our testing code

package.json
```diff
@@ -62,24 +62,24 @@
     "browserify": "16.2.0",
     "chai": "^3.5.0",
     "cssnano": "4.1.10",
-    "dompurify": "2.0.8",
+    "dompurify": "^2.3.1",
```
eanders-MS, line 65: We don't have a package-lock.json, so version numbers should not be loose matches; no `^`, `~`, etc. We need to ensure that the build machine uses the same package versions we developed with and tested locally. Why are we not using package-lock.json here?

eanders-MS (changes requested):
Lock version numbers or introduce package-lock.json. Otherwise we don't know what we're shipping.

abchatra:
Some of these are test only like karma & mocha. We should be ok without exact versions right?

eanders-MS:
> Some of these are test only like karma & mocha. We should be ok without exact versions right?

Sure, but would it not be simpler to lock them all? My preference would be to add a package-lock file actually.

PR #8493, opened 2021-09-15
A couple small perf improvements

riknoll:
A couple small changes to improve tutorial loading perf. All of these measurements were done in chrome with the 6x CPU slowdown:

1. When rendering the markdown snippets, do it outside the dom to prevent a style recalculation (went from ~90ms to ~45ms)
2. Replaced the UTF8 encoding/decoding we do in a few places to use the TextEncoder/TextDecoder APIs which are now supported in all the browsers we support and node. I mainly did this because I was noticing a lot of time spent getting a SHA256 hash on extensions (went from ~150ms to ~60ms but this is called a few times with different timings)

AFAIK the result of the TextEncoder/TextDecoder should be the same as our implementations. I confirmed it at least returned the same results in the calls to the SHA256 function.

Doesn't make a *ton* of difference on my machine but I can see the improvements in the performance tab of devtools. Hopefully I will have time to go into the office and pick up a chromebook this week (my work laptop is a little too speedy for accurate numbers).

### DIFF
--- a/cli/hid.ts
+++ b/cli/hid.ts
@@ -64,7 +64,7 @@ export function dmesgAsync() {
     return initAsync()
         .then(d => d.talkAsync(pxt.HF2.HF2_CMD_DMESG)
             .then(resp => {
-                console.log(U.fromUTF8(U.uint8ArrayToString(resp)))
+                console.log(U.fromUTF8Array(resp))
                 return d.disconnectAsync()
             }))
 }
--- a/pxtlib/hf2.ts
+++ b/pxtlib/hf2.ts
@@ -538,7 +538,7 @@ namespace pxt.HF2 {
                     return this.talkAsync(HF2_CMD_INFO)
                 })
                 .then(buf => {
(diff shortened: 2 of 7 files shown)


livcheerful (approved):

jwunderl (approved):

abchatra (approved):

dmonroym (approved):

mmoskal (approved):
LGTM. You could also use sha256 from crypto.subtle if available.

pelikhan:
saved by node 16.0 https://nodejs.org/api/util.html#util_class_util_textencoder

mmoskal:
Just to followup - @pelikhan found that the text encoder was in node since 11.0.0, but I found mkc wasn't exposing it. This is fixed in npm makecode package v0.13.1 https://github.com/microsoft/pxt-mkc/commit/fbfa3a707af5a0c59b919428fe0472b2739a6f32

PR #8496, opened 2021-09-16
Add editor message for restricting editor languages for loaded project

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/3976

Adds an editor message to control the allowed languages of a project inside an iframe. The options are:

* "python-only",
* "javascript-only",
* "blocks-only",
* "no-blocks",
* "no-python",
* "no-javascript",
* "" (all languages)

The message should look like this:

```
{
    type: "pxteditor",
    action: "setlanguagerestriction",
    restriction: "blocks-only"
}
```

### DIFF
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -597,7 +597,10 @@ declare namespace pxt.editor {
         Standard = "",
         PythonOnly = "python-only",
         JavaScriptOnly = "javascript-only",
-        NoBlocks = "no-blocks"
+        BlocksOnly = "blocks-only",
+        NoBlocks = "no-blocks",
+        NoPython = "no-python",
+        NoJavaScript = "no-javascript"
     }
 }
 
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -605,6 +605,42 @@ export class ProjectView
         this.shouldTryDecompile = false;
     }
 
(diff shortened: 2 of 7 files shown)


jwunderl (approved):

localtypings/pxtarget.d.ts
```diff
@@ -597,7 +597,10 @@ declare namespace pxt.editor {
         Standard = "",
         PythonOnly = "python-only",
         JavaScriptOnly = "javascript-only",
-        NoBlocks = "no-blocks"
+        BlocksOnly = "blocks-only",
+        NoBlocks = "no-blocks",
```
abchatra, line 601: I don't think we should expose "no-blocks". Lets remove this option.

riknoll: This was not added as part of this PR, I don't know what parts of the codebase are using this...

webapp/src/app.tsx
```diff
@@ -605,6 +605,42 @@ export class ProjectView
         this.shouldTryDecompile = false;
     }
 
+    async openBlocksAsync() {
```
livcheerful, line 608: wait what's the difference between this openBlocks and the one above? should there be a comment?

livcheerful (approved):

dmonroym (approved):

abchatra (approved):

PR #8498, opened 2021-09-17
Display meta data for skillmap cards

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/3835

looks like 
image
with tags, but these are just test tags for now :)

### DIFF
--- a/webapp/src/codecard.tsx
+++ b/webapp/src/codecard.tsx
@@ -7,7 +7,11 @@ const repeat = pxt.Util.repeatMap;
 
 export interface CodeCardState { }
 
-export class CodeCardView extends data.Component<pxt.CodeCard, CodeCardState> {
+interface CodeCardProps extends pxt.CodeCard {
+    tallCard?: boolean;
+}
+
+export class CodeCardView extends data.Component<CodeCardProps, CodeCardState> {
 
     public element: HTMLDivElement;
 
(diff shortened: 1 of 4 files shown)


webapp/src/codecard.tsx
```diff
@@ -122,8 +126,10 @@ export class CodeCardView extends data.Component<pxt.CodeCard, CodeCardState> {
                     </div>
                 </div> : undefined}
             {(card.shortName || name || descriptions) ?
-                <div className="content">
-                    {card.shortName || name ? <div className="header">{card.shortName || name}</div> : null}
+                <div className={`content ${this.props.tallCard? "tall" : ""}`}>
+                    {card.shortName || name ? <div className="header">{card.shortName || name}
+                            <div className="tags">{card.tags}</div>
```
jwunderl, line 131: I'm not certain, but do we want an rlf here? I can't recall if these can / should be translated so deferring to richard / shannon

Also, should the tags have a borderbox similar to the display on the skillmap
image
 or maybe just separate with commas? Might be just me but the spacing looks close enough that I'd read it as one phrase rather than separate tags.

jwunderl (approved):
two questions / possible nitpicks but lgtm

webapp/src/codecard.tsx
livcheerful: oooh yeah I wasn't thinking about translations

and for the tags I copied the design mentioned in office hours -- i'm not sure if there'll be more than one tag (these were just for test). I think we can just add commas in authoring? but we can take to office hours again!

jwunderl: Ah sounds good then, I think I missed that office hours so my bad~

PR #8504, opened 2021-09-21
Sound editor

microbit-josephine:
Here a new `play sound` block is introduced with a field editor drop down, as requested for in https://github.com/microbit-foundation/v2/issues/864. It was based on the original melody field editor, and hence there may be still some legacy code that can be got rid of or re-worded. For instance, the classes referenced are often the same as the ones used in the melody editor (ie `melody-editor-dropdown`) which I didn't change for consistency, and time, but in future, more work should be done into the css and styling, as for the time being I had primarily pushed surface changes to rendering in pxt-microbit to override some styling features (see blockly.less in pxt-microbit). The gallery section of the field editor also fulfils the criteria of https://github.com/microsoft/pxt-microbit/issues/4049 which asks for a way of selecting present parameters.

Some work should be done to reduce the file size, such as getting rid of large sections of repeat. A good example would be the sync field functions such as `syncVolumeField` or `syncRepeatField` which are largely all the same (except the elements they pass) except for `syncWaveField` (as it is a drop down). 

Future work needed to be done:

- Add more interpolations (perhaps some arpeggios or more novel envelopes)
- Add a vibrato tick box with a choice of range of vibrato 
- Add randomise parameter field
- Repeat functionality need to be completed. I have implemented it in the field editor, but it is not yet done for the simulator. Depending how the sounds are ulimately made, the changes will need to be made there (see note at bottom). Also, the method may need to be changed, as it is not that robust (atm it is just a for loop).
- Make a small 'sound representation image' which is like the frequency graph of each sound sample, which will represent this sound. This will then sit in the block where you click to open the editor (at the moment there are just two musical notes). This sound representation will also sit next to each pre-set sound in the gallery section, taking place of the turquoise square I put there for the time being).

Possible additions:

- Add an `add effect` field (ie some kind of EQ or other cool fields which may change the overall character of the sound)
- Make frequency graph interactive so frequency envelope can be changed this way also.

Important note about sound production:

The simulator currently uses the same method Josh used to create his prototype sound blocks to make a noise. (I think this is mainly in CODAL?). However, it was not possible to refer to this from pxt and so I created my own `sound()` function in the audioUtil.ts file in pxt. Another reason I did this is because it was also where the `tone()` function was, which was used in the other field editors... and I wanted to follow suit to pre existing implementations. Whether we use the CODAL or the play tone method, i dont know, but eventually the simulator and the editor will need to be the same.


package.json
```diff
@@ -72,10 +74,12 @@
     "lzma": "2.3.2",
     "marked": "0.3.19",
     "mocha": "5.1.0",
+    "plotly.js": "^2.5.0",
```
jwunderl, line 77: The dependencies this change adds are pretty large -- this one alone is 81.9mb unzipped https://www.npmjs.com/package/plotly.js -- this is notably larger than all the other dependencies we have in pxt combined. I'm not certain the backend would handle things properly / it could cause issues if we tried to create a build with this in it

jaustin:
```suggestion
```

@jwunderl are you able to apply this and try again please?

jwunderl: Ah, I missed the previous message, sorry! I have a few things on my plate right now that will be a bit hard to shuffle off of at the moment, would it be possible to coordinate with @abchatra on getting help with this?

PR #8505, opened 2021-09-21
Tutorial telemetry pass

shakao:
check to make sure all tick events from the old `tutorial.tsx` file are represented in the new components as well

the one i am not carrying over is `tutorial.hint.next` (from clicking "OK" on a modal in a tutorial), instead i'm adding an `isModal` measurement to the previous and next tick events, @abchatra does that work? i don't think we used the hint.next tick in many queries


abchatra (approved):

abchatra:
Looks good to me.

livcheerful (approved):

PR #8506, opened 2021-09-21
Upgrade npm dependencies

riknoll:
Upgrading less, semantic-ui-less, karma, and karma-mocha. Once this is merged, I'll go around and do PRs on all the targets with the necessary less changes.

I *think* I fixed all of the problems with updating semantic-ui-less but we should definitely do some thorough UI testing on this.

Thanks to @jwunderl for reaching into his memory palace and remembering how to fix the less build.

### DIFF
--- a/theme/theme.config
+++ b/theme/theme.config
@@ -17,6 +17,8 @@
    specify theme name below
 */
 
+@placeholder: 'default';
+
 /* Global */
 @site       : 'pxt';
 @reset      : 'pxt';
@@ -87,7 +89,7 @@
          Import Theme
 *******************************/
 
-@import "theme.less";
+@import (multiple) "theme.less";
 
 @fontPath  : 'fonts';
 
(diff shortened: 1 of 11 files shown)


jwunderl (approved):
LGTM

(for anyone else's ref, here is the link to my old branch https://github.com/microsoft/pxt/commits/updateLess which was just the less changes, I think we landed at pretty much the same point though with a few extra fixes in this change)

theme/theme.config
```diff
@@ -17,6 +17,8 @@
    specify theme name below
 */
 
+@placeholder: 'default';
```
shakao, line 20: what is this overriding?

shakao (approved):

theme/theme.config
riknoll: this fixes a build issue with less. tbh, i have no idea what it *does* but here's where I got it: https://github.com/Semantic-Org/Semantic-UI-LESS/issues/53

PR #8511, opened 2021-09-22
Force update DOM when markdown blocks have been rendered

shakao:
fixes https://github.com/microsoft/pxt-arcade/issues/4000
fixes https://github.com/microsoft/pxt-arcade/issues/4005


jwunderl (approved):
I'm a bit surprised this fixes it / the dom doesn't update on the innerhtml being mutated; guess I need to reread the dom lifecycle a bit.

shakao:
using innerhtml destroys references to child nodes, which is what the renderSnippets function is using to asynchronously replace the loading animation with the rendered blocks

shakao:
i think this is slightly less performant, but should still keep some gains from not repeatedly looping over all the elements in the generated DOM!

PR #8516, opened 2021-09-23
Minecraft -- Fix tutorial load

livcheerful:
Fixes the issue that Andrew was reporting! Gotta make sure things exist before using them :)

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -3673,7 +3673,7 @@ export class ProjectView
         let headerHeight = 0;
         if (flyoutOnly) {
             let headers = document.getElementById("headers");
-            headerHeight += headers.offsetHeight;
+            headerHeight += headers?.offsetHeight;
         }
         if (tc) {
             // maxium offset of 18rem


jwunderl (approved):

abchatra:
Is this a regression? Has this ever worked?

livcheerful:
hmmm it works for you, right?

abchatra:
I am asking from the code side. Is this a new changes we shipped couple of weeks caused this or was this code was always there?

livcheerful:
OH yeah it's from the fix of https://github.com/microsoft/pxt/pull/7876 "fix workspace creeping on the top toolbar" I'm assuming

abchatra:
Sounds good. If you port to /stable we can ship it today.

livcheerful:
oh this is straight into stable I think? The code in master is completely different so we don't need a fix there

abchatra:
Awesome sounds good

abchatra:
@livcheerful  I am forgetting typescript, but **_headers?.offsetHeight_** will resolve to undefined and NaN right instead of 0?

livcheerful:
oh yeah i can put up a fix

PR #8536, opened 2021-09-29
Don't auto sync. Add sync button.

eanders-ms:
Our cosmos backend started getting a lot more traffic after the last release. I believe it's due to a change I made to do a cloud sync whenever an Arcade tab gains focus. This backs that out and adds a "Sync" button to the home screen, allowing the user to choose when to sync cloud projects.

This change should cut down the number of cross-tab conflicts as well.

sync-button

### DIFF
--- a/webapp/src/projects.tsx
+++ b/webapp/src/projects.tsx
@@ -7,7 +7,7 @@ import * as sui from "./sui";
 import * as core from "./core";
 import * as cloudsync from "./cloudsync";
 import * as auth from "./auth";
-import * as identity from "./identity";
+import * as cloud from "./cloud";
 import * as codecard from "./codecard"
 import * as carousel from "./carousel";
 import { showAboutDialogAsync } from "./dialogs";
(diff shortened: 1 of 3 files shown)


webapp/src/projects.tsx
```diff
@@ -104,6 +104,14 @@ export class Projects extends auth.Component<ISettingsProps, ProjectsState> {
         applyCodeCardAction(this.props.parent, "projects", scr, action);
     }
 
+    async syncCloudProjects() {
+        await cloud.syncAsync();
```
shakao, line 108: does this need `{ direction: "down" }` ?

shakao (approved):

abchatra:
What if the user doesn't click the button ever? Is that ok?  Do we atleast sync once we load the page? 

Instead of the button, can this be done based on the time heuristics? Like if there was sync in last 10 minutes we don't trigger another one, otherwise we do?

webapp/src/projects.tsx
eanders-ms: Default is to sync both directions, which I think is the more appropriate behavior when the sync is initiated by the user.

eanders-ms:
> What if the user doesn't click the button ever? Is that ok? Do we atleast sync once we load the page?
> 
> Instead of the button, can this be done based on the time heuristics? Like if there was sync in last 10 minutes we don't trigger another one, otherwise we do?

We sync when the page loads, and we do a sync on a project when it's opened. I want to stay away from any kind of auto sync until we have a more efficient API for it. It needs to be a super lightweight check that doesn't hit cosmos.

PR #8539, opened 2021-09-29
fix scroll in safari microsoft/pxt-arcade#4097

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/4097

### DIFF
--- a/theme/themes/pxt/modules/modal.overrides
+++ b/theme/themes/pxt/modules/modal.overrides
@@ -50,7 +50,7 @@
 }
 
 .ui.modal {
-  position: fixed;
+  position: absolute;
   display: none;
   z-index: @zIndex;
   text-align: left;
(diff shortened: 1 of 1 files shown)


shakao:
it looks like in arcade v1.5 this selector https://github.com/microsoft/pxt/blob/master/theme/themes/pxt/modules/modal.overrides#L48 had `position: relative !important;` on it, could we restore that instead of changing it on the base modal?

jwunderl:
Hmm I can move it to position: absolute there yeah? It looks like some deep changes in sui cause position: relative to break position even more

shakao:
sure, that sounds good to me!

shakao:
does it need the !important? or is it fine with the new styles

jwunderl:
It works fine in the safari devtools, it's overriding the .ui.modals selector just below so this selector is more specific~

shakao (approved):

PR #8541, opened 2021-09-30
Debounce user preferences save

eanders-ms:
Our Cosmos backend is getting hammered by calls to the save preferences api right now. This change adds a debounce. Looking at telemetry, this should result in a 3-4x reduction.

### DIFF
--- a/pxtlib/auth.ts
+++ b/pxtlib/auth.ts
@@ -75,6 +75,11 @@ namespace pxt.auth {
     let _client: AuthClient;
     export function client(): AuthClient { return _client; }
 
+    const PREFERENCES_DEBOUNCE_MS = 5 * 1000;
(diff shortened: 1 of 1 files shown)


shakao:
we do have an existing debounce util but i don't think it has support for this max_ms concept, not sure if it could be used here? https://github.com/microsoft/pxt/blob/master/pxtlib/util.ts#L645 otherwise lgtm!

jwunderl (approved):

shakao (approved):

dmonroym (approved):

eanders-ms:
Thanks @shakao for pointing that out. I'd like to ship this as-is, but then look at switching to an existing debouncer or adding this variant to the lib. This logic is also used over in cloud.ts, so some duplication happening.

PR #8544, opened 2021-10-05
Add home and share tooltips

dmonroym:
Fixes https://github.com/microsoft/pxt-arcade/issues/4101

The buttons were missing a title which is used to determine the text in the tooltip. We now show it for the Home and the Share buttons.
Since the share button is also used to publish your project I've modified the text to be more descriptive.

image

image

### DIFF
--- a/webapp/src/container.tsx
+++ b/webapp/src/container.tsx
@@ -282,8 +282,8 @@ export class SettingsMenu extends data.Component<SettingsMenuProps, SettingsMenu
         const showCenterDivider = targetTheme.selectLanguage || targetTheme.highContrast || showGreenScreen || githubUser;
 
         return <sui.DropdownMenu role="menuitem" icon={'setting large'} title={lf("More...")} className="item icon more-dropdown-menuitem">
-            {showHome && <sui.Item className="mobile only inherit" role="menuitem" icon="home" text={lf("Home")} ariaLabel={lf("Home screen")} onClick={this.showExitAndSaveDialog} />}
-            {showShare && <sui.Item className="mobile only inherit" role="menuitem" icon="share alternate" text={lf("Share")} ariaLabel={lf("Share Project")} onClick={this.showShareDialog} />}
+            {showHome && <sui.Item className="mobile only inherit" role="menuitem" icon="home" title={lf("Home")} text={lf("Home")} ariaLabel={lf("Home screen")} onClick={this.showExitAndSaveDialog} />}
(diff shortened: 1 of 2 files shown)


dmonroym:
I wasn't able to figure it out but what's the difference between these two files?

@shakao, you modified these last, do you know?

shakao:
headerbar.tsx is the header that appears at the top of the page, container.tsx has a collection of small components--the one you're editing is the is the settings dropdown that appears when you click the gear wheel icon. in mobile views, the share functionality is moved from the header bar to the settings gearwheel:

image

dmonroym:
Ah makes sense. Thanks for clarifying!
In that case I've also changed the text. @team feel free to give feedback if you think I should leave the text simply as Share and the tooltip with share/publish

image

shakao:
maybe post a screenshot in the design channel on teams? i think PMs etc don't routinely check github pull requests unless they're tagged in them!

riknoll:
How does the slash localize?

dmonroym:
> How does the slash localize?

Shouldn't be any different than the ellipsis we have in `about...` or `more...`

webapp/src/container.tsx
```diff
@@ -282,8 +282,8 @@ export class SettingsMenu extends data.Component<SettingsMenuProps, SettingsMenu
         const showCenterDivider = targetTheme.selectLanguage || targetTheme.highContrast || showGreenScreen || githubUser;
 
         return <sui.DropdownMenu role="menuitem" icon={'setting large'} title={lf("More...")} className="item icon more-dropdown-menuitem">
-            {showHome && <sui.Item className="mobile only inherit" role="menuitem" icon="home" text={lf("Home")} ariaLabel={lf("Home screen")} onClick={this.showExitAndSaveDialog} />}
-            {showShare && <sui.Item className="mobile only inherit" role="menuitem" icon="share alternate" text={lf("Share")} ariaLabel={lf("Share Project")} onClick={this.showShareDialog} />}
+            {showHome && <sui.Item className="mobile only inherit" role="menuitem" icon="home" title={lf("Home")} text={lf("Home")} ariaLabel={lf("Home screen")} onClick={this.showExitAndSaveDialog} />}
+            {showShare && <sui.Item className="mobile only inherit" role="menuitem" icon="share alternate" title={lf("Share/Publish")} text={lf("Share/Publish")} ariaLabel={lf("Share Project")} onClick={this.showShareDialog} />}
```
jwunderl, line 286: With the text showing up in the dropdown menu having `Share/Publish` here feels bad to me; having a `/ synonym` on a button would make me think it leads to two options with a subtle difference between them and look a bit unfinished. Maybe we can have the title as something even more descriptive like 'Publish your game to create a shareable link' or similar and just leave the button as 'share'?

abchatra:
I think tooltip on the hover is great. However tooltip on the dropdown seems not great. It hides the other options. Can we just do tooltip on the hover?

PR #8546, opened 2021-10-07
[stable 7.2] Cherry pick #8545

riknoll:
#8545


abchatra:
This can be closed right? We shipped the fix with the latest release.

PR #8550, opened 2021-10-12
Get rid of extraneous dependencies to resurrect the ---asseteditor page

riknoll:
This looks way, way, scarier than it actually is! All this PR does is move things around a bit to make our dependency graph a little less complicated. Basically, the /---asseteditor page broke a while back when the cloudsync stuff went in. Obviously the cloudsync stuff shouldn't even be in the  /---asseteditor page at all, so I had to untangle our imports a bit.

The big culprit was `fireClickOnEnter` which we use all over the place and is exported by `sui.tsx` (which itself requires `app.tsx`). I factored that out into a new file along with some of the asset editor utils that were also drawing in `package.ts` for no good reason.

To fix this, I wrote a script to trace the dependency graph and figure out the culprit. Here's how the dependency graph looked for the `package.ts` issues:

```
cloudsync-> /dialogs.js-> /container.js-> /headerbar.js-> /app.js-> /cloud.js-> /auth.js-> /core.js->
/compiler.js-> /workspace.js-> /package.js-> /components/assetEditor/store/assetEditorReducer.js->
/components/assetEditor/assetCard.js-> /components/ImageFieldEditor.js-> /assetEditor.js
```

### DIFF
--- a/docs/develop/accessibility/accessibility-keyboard.md
+++ b/docs/develop/accessibility/accessibility-keyboard.md
@@ -24,15 +24,15 @@ The tab index is defined in HTML or JavaScript depending the page initialization
 
 Many elements respond to a mouse click. This is usually defined with the ``onClick`` attribute. If no additonal interaction is needed with a component, it is required that pressing the ``Enter`` or ``Space`` key has the same behavior to the ``onClick``.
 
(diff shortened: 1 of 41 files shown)


dmonroym (approved):

abchatra:
@riknoll what is the status of this?

riknoll:
I need to resolve the conflicts, but i've been busy.

I'll close this and resurrect it later

PR #8555, opened 2021-10-15
[stable7.0] restore github signout

eanders-ms:
Implementation of https://github.com/microsoft/pxt/pull/8553 for stable7.0. A small difference was necessary to avoid having to pull in the auth refactor we did for skillmap.


eanders-ms:
Fixes https://github.com/microsoft/pxt-microbit/issues/4396

PR #8576, opened 2021-11-04
Updates to sync with Blockly changes through July 2021

shakao:
requires https://github.com/microsoft/pxt-blockly/pull/359

mostly small fixes to sync with blockly-side changes. i also removed all the code for handling the accessible blocks experiment, since it's moved out of blockly-core into an extension and we need to re-integrate with it. 

(this will fail the build until i can update the npm package, since that's where the typings come from)

builds!
https://arcade.makecode.com/app/894739978b4be11010c202ea303e01047d102420-e3d1f85bce
https://makecode.microbit.org/app/9e9f521b48ac103e33139fe979e611db76c9eab1-d62781acca

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -819,7 +819,7 @@ namespace pxt.blocks {
                                 inputCheck = pr.type;
                             } else if (pr.type == "number" && pr.shadowBlockId && pr.shadowBlockId == "value") {
                                 inputName = undefined;
-                                fields.push(namedField(new Blockly.FieldTextInput("0", Blockly.FieldTextInput.numberValidator), defName));
+                                fields.push(namedField(new Blockly.FieldNumber("0"), defName));
                             } else if (pr.type == "string" && pr.shadowOptions && pr.shadowOptions.toString) {
                                 inputCheck = null;
                             } else {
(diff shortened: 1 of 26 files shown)


pxtblocks/blocklyloader.ts
```diff
@@ -2124,11 +2124,6 @@ namespace pxt.blocks {
                     let blockText = '<xml>' +
                         '<block type="variables_change" gap="' + gap + '">' +
                         Blockly.Variables.generateVariableFieldXmlString(mostRecentVariable) +
-                        '<value name="DELTA">' +
```
jwunderl, line 2127: is this not necessary anymore / coming from generateVariableFieldXmlString? or was it just always an ignored field?

pxtblocks/fields/field_turnratio.ts
```diff
@@ -36,23 +36,23 @@ namespace pxtblockly {
...
+            let defs = Blockly.utils.dom.createSvgElement<SVGElement>('defs', {}, svg);
+            let marker = Blockly.utils.dom.createSvgElement<SVGElement>('marker', {
                 'id': 'head',
                 'orient': "auto",
                 'markerWidth': '2',
                 'markerHeight': '4',
                 'refX': '0.1', 'refY': '1.5'
             }, defs);
-            let markerPath = Blockly.utils.dom.createSvgElement('path', {
+            let markerPath = Blockly.utils.dom.createSvgElement<SVGElement>('path', {
```
jwunderl, line 55: nit for future (looks like this is more blockly typing side) -- I believe we should be able to use [template literal types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) to make it so we don't need to respecify the types each time like this. I suppose maybe overdoing it though depending on how often it is used.

jwunderl (approved):

livcheerful (approved):

pxtblocks/blocklyloader.ts
shakao: always ignored! we had the warning for it suppressed--i actually also re-suppressed it but before i did that, i was trying to fix the things causing the warning so this change is from that pass

dmonroym (approved):

shakao:
updated builds! 
https://arcade.makecode.com/app/894739978b4be11010c202ea303e01047d102420-e3d1f85bce
https://makecode.microbit.org/app/9e9f521b48ac103e33139fe979e611db76c9eab1-d62781acca

PR #8578, opened 2021-11-05
Don't let blocks overflow, add scrollbar

livcheerful:
fixes https://github.com/microsoft/pxt-minecraft/issues/1991

but needs this change! https://github.com/microsoft/pxt-blockly/pull/360

looks like this:
image

ideally would have more CSS changes to restrict the size of the flyout to always be a certain width. Right now the sizes get kinda wild sometimes. Another EZ option would be removing the double scroll, and just cutting off the blocks. You can see the whole block when you drag one out (shannon's idea!) 

a good way to test is to make your screen small, and then press the + button 

(I still need to test this in game)

Try here:
https://minecraft.makecode.com/app/ab924030ba350913bc71b432bfb45f4b33feb968-9ded06f80b#tutorial:/hour-of-code/reforestation

also fllyoutonly no hideiteration
https://minecraft.makecode.com/app/ab924030ba350913bc71b432bfb45f4b33feb968-9ded06f80b#tutorial:36607-16422-18962-51568

and the basic upload 
https://minecraft.makecode.com/app/ab924030ba350913bc71b432bfb45f4b33feb968-9ded06f80b

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
   },
   "devDependencies": {
     "monaco-editor": "0.19.3",
-    "pxt-blockly": "3.1.2",
+    "pxt-blockly": "3.1.3",
     "react": "16.8.3",
     "react-dom": "16.11.0",
     "react-modal": "3.3.2",
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
   },
   "devDependencies": {
     "monaco-editor": "0.19.3",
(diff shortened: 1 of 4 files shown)


shakao (approved):

shakao:
looks good to me! definitely a huge improvement over the existing state already

dmonroym (approved):

PR #8586, opened 2021-11-09
Badge fixes

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/4286
Fixes https://github.com/microsoft/pxt-arcade/issues/4290
Fixes https://github.com/microsoft/pxt-arcade/issues/4295

Also does a ton of css tweaks based on feedback @soniakandah sent me!

### DIFF
--- a/react-common/components/profile/BadgeInfo.tsx
+++ b/react-common/components/profile/BadgeInfo.tsx
@@ -15,24 +15,24 @@ export const BadgeInfo = (props: BadgeInfoProps) => {
         <div className="profile-badge-info-image">
             <Badge badge={badge} disabled={!badge.timestamp} />
         </div>
-        <div className="profile-badge-info-header">
-            {lf("Awarded For:")}
-        </div>
-        <div className="profile-badge-info-text">
-            {badgeDescription(badge)}
-        </div>
-        { badge.timestamp ?
+        <div className="profile-badge-info-item">
             <div className="profile-badge-info-header">
-                {lf("Awarded On:")}
+                {lf("Awarded For:")}
             </div>
-            : undefined
-        }
-        { badge.timestamp ?
             <div className="profile-badge-info-text">
-                {date.toLocaleDateString(pxt.U.userLanguage())}
+                {badgeDescription(badge)}
             </div>
-            : undefined
-        }
+        </div>
+        { badge.timestamp ?
+            <div className="profile-badge-info-item">
+                <div className="profile-badge-info-header">
+                    {lf("Awarded On:")}
+                </div>
+                <div className="profile-badge-info-text">
+                    {date.toLocaleDateString(pxt.U.userLanguage())}
+                </div>
+            </div>
+        : undefined }
     </div>
 }
 
--- a/react-common/styles/profile/profile.css
+++ b/react-common/styles/profile/profile.css
@@ -3,7 +3,17 @@
     flex-direction: row;
     height: 100%;
 
+    --header-text-color: #000000;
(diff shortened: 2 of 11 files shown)


jwunderl (approved):

livcheerful (approved):

react-common/components/profile/BadgeInfo.tsx
```diff
@@ -15,24 +15,24 @@ export const BadgeInfo = (props: BadgeInfoProps) => {
...
-        }
-        { badge.timestamp ?
             <div className="profile-badge-info-text">
-                {date.toLocaleDateString(pxt.U.userLanguage())}
+                {badgeDescription(badge)}
             </div>
-            : undefined
-        }
+        </div>
+        { badge.timestamp ?
```
shakao, line 26: small: `badge.timestamp &&` ?

riknoll: timestamp is a number, so if it's zero it results in a 0 being printed. I think I was initializing it to 0 at some point?

shakao: could do !== undefined or something but it's also not a huge deal

react-common/styles/profile/profile.css
```diff
@@ -104,6 +115,11 @@
     box-shadow: none;
 }
 
+.profile-badge-subtitle {
+    font-style: italic;
+    color: #aba9a5;
```
shakao, line 120: should this be in a less variable?

skillmap/src/lib/skillMapParser.ts
```diff
@@ -336,6 +336,9 @@ function inflateMapReward(section: MarkdownSection, base: Partial<MapRewardNode>
             }
         }
 
+        const priority = ["completion-badge", "certificate"];
```
shakao, line 339: we're not going to let skillmap authors determine their own reward order?

skillmap/src/components/AppModal.tsx
```diff
@@ -419,77 +421,126 @@ export class AppModalImpl extends React.Component<AppModalProps, AppModalState>
...
+                    <span>{lf("Next Reward")}</span>
+                    <i className="icon right circle arrow" />
+                </button>
+            )
         }
 
-        return <Modal title={title} actions={buttons} onClose={this.handleOnClose}>
+        return <Modal title={title} onClose={this.handleOnClose}>
             {lf("Use the button below to get your completion certificate!")}
+            <div className="reward-buttons">
```
shakao, line 469: are we doing some special/different design for the reward modal buttons? can we make these changes across all the skillmap modal buttons? don't love moving them out like this, "open reward link in new tab" and "move to the next modal" both still feel like modal actions to me

riknoll: We were already doing that for the completion modal, I'm just reusing the css from the button in that modal

react-common/styles/profile/profile.css
riknoll: I'll move all colors into variables

skillmap/src/components/AppModal.tsx
shakao: the completion modal also has separate follow-up actions though, these are the follow-up actions for this modal. this is more a design question than an implementation one, i guess. could we pass in the modal button class as a property to the actions?

eanders-ms (approved):

shakao (approved):

shakao (approved):

PR #8587, opened 2021-11-09
Sync language change immediately, don't wait for debouncer to fire

eanders-ms:
Because skillmap updates user preferences so frequently, I added a debouncer to ease network traffic. Changes to user preferences are accumulated and sent in batch after a timeout. For signed in users, language setting is also saved in this blob. The issue was that after changing language, the page is reloaded. Due to the debounce, the change hadn't been synced to the cloud yet. When the page reloads, it pulls latest preferences from the cloud, which still has the old language selection.

I also added a queue to accumulate pending changes. I think we had a bug where only the most recent patch would be synced up.

Fixes: https://github.com/microsoft/pxt-arcade/issues/4310

### DIFF
--- a/pxtlib/auth.ts
+++ b/pxtlib/auth.ts
@@ -305,7 +305,9 @@ namespace pxt.auth {
             return result.success;
         }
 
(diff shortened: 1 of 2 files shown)


eanders-ms:
I need to write up an identity test script.

livcheerful (approved):

PR #8592, opened 2021-11-11
Adding certificate preview to the rewards modal and skillmap tests

riknoll:
More design changes from @soniakandah


jwunderl (approved):

skillmap/src/lib/skillMapParser.ts
```diff
@@ -317,22 +366,43 @@ function inflateMapReward(section: MarkdownSection, base: Partial<MapRewardNode>
...
-
-            switch (kind) {
-                case "certificate":
-                    parsedRewards.push({
+        for (const reward of rewards.items) {
+            if (typeof reward === "string") {
+                let [kind, ...value] = reward.split(":");
+
+                switch (kind) {
+                    case "certificate":
```
shakao, line 374: do we need to support this case? the only map that could've been using it (before this current pr changed the syntax) was our own hour of code, right?

shakao (approved):

skillmap/src/lib/skillMapParser.ts
riknoll: there are a few right now, i can remove this once release shenanigans are settled but best to keep it for now

shakao: sounds good! lgtm otherwise

livcheerful (approved):

riknoll:
I've updated this PR to also include the necessary change for https://github.com/microsoft/pxt-arcade/issues/4320

@shakao @livcheerful please take a look at the most recent commit

dmonroym (approved):

PR #8597, opened 2021-11-13
Delay the readyPromise until the workspace is loaded, add skillmap loader to main page

shakao:
fixes https://github.com/microsoft/pxt-arcade/issues/4331

we're still not really sure what's causing this bug.... but delaying the progress syncing until the workspace has loaded seems to fix it (for now). this means there's a noticable pause between logging in and your local progress importing, so we also added a loader during that state. 

build: https://arcade.makecode.com/app/de3eb561356a7951f9b6614556c3b56c968ee044-5f756ce822--skillmap

### DIFF
--- a/skillmap/src/App.tsx
+++ b/skillmap/src/App.tsx
@@ -358,7 +358,7 @@ class AppImpl extends React.Component<AppProps, AppState> {
         const maps = Object.keys(skillMaps).map((id: string) => skillMaps[id]);
         return (<div className={`app-container ${pxt.appTarget.id}`}>
                 <HeaderBar />
-                {syncingLocalState &&<div className={"makecode-frame-loader"}>
+                {syncingLocalState && <div className={"makecode-frame-loader"}>
                     <img src={resolvePath("assets/logo.svg")} alt={lf("MakeCode Logo")} />
                 <div className="makecode-frame-loader-text">{lf("Saving to cloud...")}</div>
                 </div>}
--- a/skillmap/src/App.tsx
+++ b/skillmap/src/App.tsx
@@ -30,7 +30,7 @@ import { InfoPanel } from './components/InfoPanel';
 
 import { parseSkillMap } from './lib/skillMapParser';
 import { parseHash, getMarkdownAsync, MarkdownSource, parseQuery,
-    setPageTitle, setPageSourceUrl, ParsedHash } from './lib/browserUtils';
+    setPageTitle, setPageSourceUrl, ParsedHash, resolvePath } from './lib/browserUtils';
 
 import { MakeCodeFrame } from './components/makecodeFrame';
 import { getLocalUserStateAsync, getUserStateAsync, saveUserStateAsync } from './lib/workspaceProvider';
@@ -70,6 +70,7 @@ interface AppState {
     error?: string;
     cloudSyncCheckHasFinished: boolean;
     badgeSyncLock: boolean;
+    syncingLocalState?: boolean;
 }
 
 class AppImpl extends React.Component<AppProps, AppState> {
@@ -236,6 +237,7 @@ class AppImpl extends React.Component<AppProps, AppState> {
         if (!await authClient.loggedInAsync()) {
             this.setState({cloudSyncCheckHasFinished: true});
         } else {
+            this.setState({syncingLocalState: true});
             const doCloudSyncCheckAsync = async () => {
                 const state = store.getState();
                 const localUser = await getLocalUserStateAsync();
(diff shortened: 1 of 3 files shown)


skillmap/src/App.tsx
```diff
@@ -352,18 +354,22 @@ class AppImpl extends React.Component<AppProps, AppState> {
...
                 <HeaderBar />
-                    <div className={`skill-map-container ${activityOpen ? "hidden" : ""}`} style={{ backgroundColor: theme.backgroundColor }}>
-                        { error
-                            ? <div className="skill-map-error">{error}</div>
-                            : <SkillGraphContainer maps={maps} backgroundImageUrl={backgroundImageUrl} />
-                        }
-                        { !error && <InfoPanel />}
-                    </div>
-                    <MakeCodeFrame onFrameLoaded={this.onMakeCodeFrameLoaded}/>
+                {syncingLocalState &&<div className={"makecode-frame-loader"}>
```
jwunderl, line 361:
```suggestion
                {syncingLocalState && <div className={"makecode-frame-loader"}>
```

nit spacing

riknoll (approved):

jwunderl (approved):

skillmap/src/App.tsx
```diff
@@ -427,6 +433,10 @@ class AppImpl extends React.Component<AppProps, AppState> {
             }
         }
 
+        if (!this.props.signedIn || (this.props.signedIn && this.state.cloudSyncCheckHasFinished)) {
+            this.setState({ syncingLocalState: false });
```
eanders-ms, line 437: Curious why this needs to happen on every state change.

skillmap/src/components/makecodeFrame.tsx
```diff
@@ -165,6 +163,10 @@ class MakeCodeFrameImpl extends React.Component<MakeCodeFrameProps, MakeCodeFram
 
         switch (data.action) {
             case "newproject":
+                if (!this.state.workspaceReady) {
+                    this.setState({ workspaceReady: true });
```
eanders-ms, line 167: >this.setState({ workspaceReady: true }); [](http://example.com/codeflow?start=20&length=40)

Will queued messages be sent at this point?

skillmap/src/App.tsx
shakao: oh i just added it in the spirit of overkill because i was afraid of having the new loader permanently hide the skillmap in weird edge cases; probably setting it anywhere we set cloudSyncCheckStateHasFinished to true would be enough though

eanders-ms: "spirit of overkill" :) :). I think it will trigger a redraw every time (i could be wrong!). Maybe add a condition to check for `this.state.syncingLocalState == true`? Or trust that it's always getting set correctly in the cloud sync check.

skillmap/src/components/makecodeFrame.tsx
shakao: oops, clearing the queue got dropped while i was cleaning stuff up! added it back

skillmap/src/App.tsx
shakao: ah you're right! i thought setstate was smart about setting the same value but it definitely redraws in this case. will update!

riknoll: oh, huh... i also thought setstate was smart

shakao: only for functional components it seems

livcheerful (approved):

skillmap/src/App.tsx
eanders-ms: I'm loving the trend in github profile icons happening here :)

riknoll: well you're a trendsetter!

shakao:
updated build: https://arcade.makecode.com/app/f74e28638f6c8fd2eccbd1d8249a8caa9645e275-4bd1e9bbd4--skillmap

eanders-ms (approved):
It does have one slightly concerning edge... on a slow machine it takes a long time for the ready promise to resolve. Enough time that it is possible to start an activity before the "Syncing to cloud" blocker appears. Nice thing is the cloud sync will happen in the background while the user is in the activity. If the current activity is stomped by the cloud, local progress seems to win without a conflict. 🤞 I think it's worth having in as a safety measure!

shakao:
oof yeah, i'd love to revisit this entire loader/maybe figure out a way to fix the underlying issue after this release... i didn't want to put the loader before the readypromise because it seemed too intrusive?

PR #8600, opened 2021-11-16
Make loading start earlier and add locked badges

riknoll:
This PR does two things:

1. Moves the "saving to cloud" loader earlier so that it shows up when you navigate to the page if you're signed in. This prevents the bait and switch of appearing to load before showing a loader
2. Adds support for locked badge images:

Screen Shot 2021-11-15 at 4 13 51 PM

### DIFF
--- a/skillmap/src/App.tsx
+++ b/skillmap/src/App.tsx
@@ -70,7 +70,6 @@ interface AppState {
     error?: string;
     cloudSyncCheckHasFinished: boolean;
     badgeSyncLock: boolean;
-    syncingLocalState?: boolean;
     showingSyncLoader?: boolean;
 }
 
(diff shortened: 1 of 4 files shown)


dmonroym (approved):

skillmap/src/App.tsx
```diff
@@ -354,11 +359,11 @@ class AppImpl extends React.Component<AppProps, AppState> {
 
     render() {
         const { skillMaps, activityOpen, backgroundImageUrl, theme } = this.props;
-        const { error, syncingLocalState } = this.state;
+        const { error, showingSyncLoader } = this.state;
         const maps = Object.keys(skillMaps).map((id: string) => skillMaps[id]);
         return (<div className={`app-container ${pxt.appTarget.id}`}>
                 <HeaderBar />
-                {syncingLocalState && <div className={"makecode-frame-loader"}>
+                {showingSyncLoader && <div className={"makecode-frame-loader"}>
```
shakao, line 366: do we even need `syncingLocalState` any more? could we remove it entirely?

livcheerful (approved):

jwunderl (approved):

eanders-ms (approved):

skillmap/src/App.tsx
riknoll: nope, removed!

PR #8605, opened 2021-11-18
Add email opt-in checkbox to User Profile ✅👍

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/4357

Adds a checkbox to the User Profile Pane to opt in to emails. I moved the Notifications to react-common, and added a watch in the gulpfile to react-common too

a build to try! https://arcade.makecode.com/app/d941d5d3a379015428bcdf82f60830cf52589d88-0440bbf0eb

image
when you click the check there's a lil spinner that goes until the actual cloud save has happened
image
if you're offline, you'll get a red toast and the setting will revert to original


webapp/src/coretsx.tsx
```diff
@@ -256,84 +256,3 @@ export class LoadingDimmer extends React.Component<LoadingDimmerProps, LoadingDi
         </sui.Dimmer>;
     }
 }
-
-
-export interface NotificationOptions {
```
livcheerful, line 261: move this to react-common

webapp/src/user.tsx
```diff
@@ -17,14 +18,17 @@ type ProfileDialogState = {
...
 };
 
 export class ProfileDialog extends auth.Component<ProfileDialogProps, ProfileDialogState> {
     constructor(props: ProfileDialogProps) {
         super(props);
+        const pref = auth.userPreferences();
         this.state = {
             visible: props.visible,
             location: 'my-stuff',
+            emailSelected: pref?.email ? CheckboxStatus.Selected : CheckboxStatus.Unselected
```
livcheerful, line 31: this is this bc we also have a third 'Waiting' status

react-common/components/Notification.tsx
```diff
@@ -0,0 +1,82 @@
+import * as React from "react";
```
livcheerful, line 1: this is all pulled from coretsx.tsx!

gulpfile.js
```diff
@@ -171,6 +173,8 @@ function initWatch() {
     gulp.watch("./pxtwinrt/**/*", gulp.series(pxtwinrt, ...tasks.slice(5)));
     gulp.watch("./cli/**/*", gulp.series(cli, ...tasks.slice(5)));
 
+    gulp.watch("./react-common/styles/**/*.css", gulp.series(reactCommonCss, ...tasks.slice(9)))
```
livcheerful, line 176: you still gotta do a pxt buildcss --force here

webapp/src/user.tsx
```diff
@@ -60,6 +64,21 @@ export class ProfileDialog extends auth.Component<ProfileDialogProps, ProfileDia
...
+        } else {
+            core.errorNotification(lf("Oops, something went wrong"))
+        }
+
+        this.setState({emailSelected: pref.email ? CheckboxStatus.Selected : CheckboxStatus.Unselected})
+    }
+
+    onEmailCheckClicked = (isSelected: boolean ) => {
+        auth.setEmailPrefAsync(isSelected, this.onEmailPrefUpdateComplete);
+        this.setState({emailSelected: CheckboxStatus.Waiting})
```
riknoll, line 79:
```suggestion
        this.setState({ emailSelected: CheckboxStatus.Waiting })
```

webapp/src/user.tsx
```diff
@@ -60,6 +64,21 @@ export class ProfileDialog extends auth.Component<ProfileDialogProps, ProfileDia
...
     }
 
+    onEmailPrefUpdateComplete = (success: boolean, pref: pxt.auth.UserPreferences) => {
+        if (success) {
+            core.infoNotification(lf("Settings saved"))
+        } else {
+            core.errorNotification(lf("Oops, something went wrong"))
+        }
+
+        this.setState({emailSelected: pref.email ? CheckboxStatus.Selected : CheckboxStatus.Unselected})
```
riknoll, line 74:
```suggestion
        this.setState({ emailSelected: pref.email ? CheckboxStatus.Selected : CheckboxStatus.Unselected })
```

skillmap/src/components/UserProfile.tsx
```diff
@@ -140,6 +147,22 @@ export class UserProfileImpl extends React.Component<UserProfileProps, UserProfi
...
     }
 
+    onEmailPrefUpdateComplete = (success: boolean, pref: pxt.auth.UserPreferences) => {
+        if (success) {
+            infoNotification(lf("Settings saved"))
+            this.setState({emailSelected: pref.email ? CheckboxStatus.Selected : CheckboxStatus.Unselected})
+        } else {
+            errorNotification(lf("Oops, something went wrong"))
+            const localPref = this.props.preferences;
+            this.setState({emailSelected: localPref?.email ? CheckboxStatus.Selected: CheckboxStatus.Unselected})
```
riknoll, line 157:
```suggestion
            this.setState({ emailSelected: localPref?.email ? CheckboxStatus.Selected: CheckboxStatus.Unselected })
```

skillmap/src/components/UserProfile.tsx
```diff
@@ -140,6 +147,22 @@ export class UserProfileImpl extends React.Component<UserProfileProps, UserProfi
         this.props.dispatchShowDeleteAccountModal();
     }
 
+    onEmailPrefUpdateComplete = (success: boolean, pref: pxt.auth.UserPreferences) => {
+        if (success) {
+            infoNotification(lf("Settings saved"))
+            this.setState({emailSelected: pref.email ? CheckboxStatus.Selected : CheckboxStatus.Unselected})
```
riknoll, line 153:
```suggestion
            this.setState({ emailSelected: pref.email ? CheckboxStatus.Selected : CheckboxStatus.Unselected })
```

riknoll (changes requested):

pxtlib/auth.ts
```diff
@@ -347,6 +349,9 @@ namespace pxt.auth {
                 } else {
                     pxt.reportError("identity", "update preferences failed", result as any);
                 }
+                if (callback) {
+                    callback(result.success, result.resp)
```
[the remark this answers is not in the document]
livcheerful: hmmmm my thinking was I don't know how to return a promise in the case where we're debouncing the set preferences call? but if there's a way I will switch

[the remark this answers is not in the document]
eanders-ms: Agreed, it would be better to use the returned promise. I wonder if it will work correctly for debounced saves, but could sort that out later. This is an immediate save so should be fine. Update the promise to return a success/fail boolean.

skillmap/src/lib/core.ts
```diff
@@ -0,0 +1,25 @@
+import { pushNotificationMessage } from "react-common/Notification"
+
+///////////////////////////////////////////////////////////
```
eanders-ms, line 3: The name of this file could be more descriptive. Maybe `notificationService.ts`? or `notifications.ts`? `toast.ts`?

livcheerful: oh yeah I was just copying the structure over in the webapp in case we....ever wanted to match more things between the two? But I'm also happy  to make this a notifications only file

pxtlib/auth.ts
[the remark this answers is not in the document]
eanders-ms: For debouncing, keep a promise around and just return it if it already exists, otherwise create it. We do this in a few places. cloud.ts syncAsync is one example. I think it's ok to not add that today, since we don't have a test case. Just leave a comment that it needs implementing.

skillmap/src/lib/core.ts
riknoll: yeah let's do notifications only. we've mostly been doing 1 component per file in skillmap

livcheerful: 👍

pxtlib/auth.ts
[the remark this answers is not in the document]
livcheerful: I'll try it out! I do want to mention -- there is no promise returned here currently, but I'll add it.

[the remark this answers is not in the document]
riknoll: it's an async function so a promise is implicitly returned. You can just put a `return` here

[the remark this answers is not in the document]
livcheerful: ooooooooooooooooo i see!

pxtlib/auth.ts
```diff
@@ -305,17 +312,19 @@ namespace pxt.auth {
             return result.success;
         }
 
+        private debounceResult: Promise<boolean>;
```
eanders-ms, line 315: Is debounceResult used?

pxtlib/auth.ts
```diff
@@ -305,17 +312,19 @@ namespace pxt.auth {
...
             ops = Array.isArray(ops) ? ops : [ops];
             ops = ops.filter(op => !!op);
-            if (!ops.length) { return; }
             const curPref = await this.userPreferencesAsync();
+            if (!ops.length) { return {success: true, res: curPref} }
             ts.pxtc.jsonPatch.patchInPlace(curPref, ops);
             await this.setUserPreferencesAsync(curPref);
             // If we're not logged in, non-persistent local state is all we'll use
-            if (!await this.loggedInAsync()) { return; }
+            if (!await this.loggedInAsync()) { return {success: false, res: undefined}; }
```
eanders-ms, line 327: I feel like this should return `{ success: true, res: currPref }`. It isn't an error condition to be signed out.

eanders-ms: Nit/Suggestion: add whitespace after `{` and before `}`. We could use a lint rule for this.

pxtlib/auth.ts
```diff
@@ -347,17 +356,19 @@ namespace pxt.auth {
                 } else {
                     pxt.reportError("identity", "update preferences failed", result as any);
                 }
+                return { success: result.success, res: result.resp};
```
eanders-ms, line 359: nit: inconsistent whitespace.

eanders-ms (changes requested):

pxtlib/auth.ts
livcheerful: oh shoot no that was a test good catch

pxtlib/auth.ts
livcheerful: hmm but if you're signed out then the choice won't be saved? I guess at least in email case you can't even change it

eanders-ms: If you're signed out, right now the profile page will be inaccessible. Maybe that will change one day.

eanders-ms (approved):
:shipit:

jwunderl (approved):

react-common/components/profile/UserPane.tsx
```diff
@@ -36,6 +41,13 @@ export const UserPane = (props: UserPaneProps) => {
         </div>
         { notification && <UserNotification notification={notification}/> }
         <div className="profile-spacer"></div>
+        <div className="profile-email">
+            <Checkbox isChecked={emailChecked} onClick={onEmailCheckClick} label={checkboxLabel}/>
+            <div className={checkboxLabel}>
```
riknoll, line 46: should be id, not className

riknoll: to work with labelledby

react-common/components/Checkbox.tsx
```diff
@@ -16,7 +16,7 @@ export const Checkbox = (props: CheckboxProps) => {
     const { isChecked, label } = props;
     return (
         <div className="checkbox">
-            {isChecked == CheckboxStatus.Waiting ? <div className={`ui inline mini loader active`}/> :
+            {isChecked == CheckboxStatus.Waiting ? <div className={`ui inline mini loader active`} aria-checked="mixed"/> :
```
riknoll, line 19: this doesn't need an aria-checked, because it doesn't have role="checkbox"

riknoll (approved):

PR #8607, opened 2021-11-19
More careful patching of user preferences

eanders-ms:
Fixes https://github.com/microsoft/pxt-arcade/issues/4367, and should go into the hotfix.

Turns out there were multiple issues at play here that could result in loss of skillmap progress. First, the code to update skillmap state was just overwriting the entire state object, and thereby erasing state made concurrently in other tabs/browsers. I put a mitigation in place to not simply stomp over the old state, but do extra work to calculate the actual diff.

These changes helped, but another issue emerged where the server and client versions of the state can diverge when working in multiple tabs to such a degree that the patch generated by a client might not be applicable to the server's version of the state. I mitigated this issue by having the client fetch the latest state from the backend and carefully apply the local changes. From the result of that, produce a final patch by diffing it and the original from the server. This should ensure we send a diff to the server that can be processed.

Sigh. These are all just mitigations and not a long-term solution. Long term, this whole diff/patch system needs a different approach entirely. I've captured some of that work in this item: https://github.com/microsoft/pxt-backend/issues/538. Keeping all the state in one json blob isn't a robust solution. Fields need to be broken out into their own table entries. Separate the concerns.


pxtlib/auth.ts
```diff
@@ -305,59 +305,92 @@ namespace pxt.auth {
...
-                    this.prefPatchOps[iExisting] = incoming;
-                    // Clear from incoming so we don't add it below
-                    (ops as ts.pxtc.jsonPatch.PatchOperation[])[iIncoming] = null;
-                    return true;
-                });
-            });
-            // Add remaining ops to the queue
-            ops.filter(op => !!op).forEach(op => this.prefPatchOps.push(op));
+            // Process incoming patch operations to produce a more fine-grained set of diffs. Incoming patches may be overly destructive
+            {
```
jwunderl, line 341: this is just a bit nitpicky but is there a reason for the closure here / concern of variables conflicting or is it just for visually separating it out within the function? We don't tend to use non-behavioral closures too often so it stood out a bit (I don't know if there's an official term for this, I just mean 'a closure that only separates scope and does not have a keyword affecting it').

eanders-ms:
Test target! https://arcade.staging.pxt.io/app/604a3a3546957fb45a68032a17532b5d2140357e-45606a806e

If this doesn't make it into the build in time for testing, we can use this target.

pxtlib/auth.ts
```diff
@@ -312,61 +312,96 @@ namespace pxt.auth {
...
+                const pDst = U.deepCopy(pSrc);
+                ts.pxtc.jsonPatch.patchInPlace(pDst, ops);
+                let diff = ts.pxtc.jsonPatch.diff(pSrc, pDst);
+                // Run caller-provided filter
+                if (diff.length && filter) { diff = diff.filter(filter); }
+                return diff;
+            }
+
+            // Process incoming patch operations to produce a more fine-grained set of diffs. Incoming patches may be overly destructive
+            {
```
livcheerful, line 341: I might be super confused, but I can't figure out what these braces are for????

eanders-ms: Just scoping the variables. I use `diff` and `curPrefs` a lot in this function.

jwunderl (approved):

pxtlib/auth.ts
eanders-ms: Just scoping the variables. I use `diff` and `curPrefs` a lot in this function. But @livcheerful also commented on it being confusing. I'll clean it up :).

PR #8610, opened 2021-11-20
Fix high contrast for non-identity platforms

eanders-ms:
The high contrast setting on non-identity targets has been broken since the auth refactor to support identity in skillmap. This fixes it.

Fixes https://github.com/microsoft/pxt-arcade/issues/4388

PR #8613, opened 2021-11-22
[stable7.3] More careful patching of user preferences (#8607)

shakao:
* more careful patching of user preferences

* remove variable scoping

### DIFF
--- a/pxtlib/auth.ts
+++ b/pxtlib/auth.ts
@@ -73,7 +73,7 @@ namespace pxt.auth {
         highContrast: false,
         language: pxt.appTarget.appTheme.defaultLocale,
         reader: "",
-        skillmap: {mapProgress: {}, completedTags: {}},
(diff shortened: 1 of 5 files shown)


jwunderl (approved):
LGTM, probably should port https://github.com/microsoft/pxt/pull/8610 too

shakao:
hmm, isn't that fix for non-identity targets?

jwunderl:
Hm doesn't it also apply when the user is just not signed in when sign in is available? 

(either way it would be needed for offline app most likely, but if it doesn't repro in the not signed in case I suppose we could port it later for an offline app only portion)

dmonroym (approved):

shakao:
yeah i think it's specifically in the case of auth being disabled for the target, not just when the user is logged out. pulled it over for the offline app though! we should remember to thoroughly test language stuff in arcade this afternoon

PR #8618, opened 2021-11-30
Converting react-common css into less

riknoll:
Since I was changing the code, I also cleaned up our `pxt buildcss` command a bit. The `--force` flag is no longer required and I converted everything to `async/await`. It's much easier to read now.

The actual css in react-common is completely unchanged, since .less is a superset of .css


cli/cli.ts
```diff
@@ -1929,98 +1929,123 @@ ${gcards.map(gcard => `[${gcard.name}](${gcard.url})`).join(',\n')}
...
     nodeutil.mkdirP(path.join("built", "web"));
     const lessPath = require.resolve('less');
     const lessCPath = path.join(path.dirname(lessPath), '/bin/lessc');
-    return nodeutil.spawnAsync({
+
+    const lessIncludePaths = [
+        "node_modules/semantic-ui-less",
+        "node_modules/pxt-core/theme",
+        "theme/foo/bar",
+        "theme",
```
shakao, line 1948: this wasn't included before, is there a reason we're adding it now?

shakao (approved):
had one question, LGTM otherwise

cli/cli.ts
riknoll: yeah, I added it for the imports in `react-common-skillmap.less`. Specifically trying to include this file: https://github.com/microsoft/pxt-arcade/blob/master/theme/site/globals/site.variables

I did it here because using a relative path that reaches out of `node_modules/pxt-core` and into the target felt bad.

PR #8619, opened 2021-11-30
Check skillmap user locale against target's available locales before setting

shakao:
we were always passing the user's locale to crowdin, even if it wasn't in the list of available locales for that target. this checks if there's a `force` flag or if the locale has been enabled, and if not it resets to the target's default locale. 

i tested `forcelang=ru`, as well as setting the browser locale to an enabled (chinese) and non-enabled (russian) locale and it seems to be working as expected (localizes for the first two cases, doesn't for the third), but let me know if there's any scenarios i missed!

### DIFF
--- a/skillmap/src/App.tsx
+++ b/skillmap/src/App.tsx
@@ -168,7 +168,10 @@ class AppImpl extends React.Component<AppProps, AppState> {
             force: force,
         });
 
-        pxt.BrowserUtils.setCookieLang(useLang!);
+        if (pxt.Util.isLocaleEnabled(useLang!)) {
+            pxt.BrowserUtils.setCookieLang(useLang!);
+            pxt.Util.setUserLanguage(useLang!);
+        }
     }
 
     protected async fetchAndParseSkillMaps(source: MarkdownSource, url: string) {
@@ -515,7 +518,6 @@ async function updateLocalizationAsync(opts: LocalizationUpdateOptions): Promise
         ts.pxtc.Util.TranslationsKind.SkillMap
     );
 
-    pxt.Util.setUserLanguage(code);
     if (translations) {
         pxt.Util.setLocalizedStrings(translations);
     }
--- a/skillmap/src/App.tsx
+++ b/skillmap/src/App.tsx
@@ -138,13 +138,17 @@ class AppImpl extends React.Component<AppProps, AppState> {
             force = !!mlang && !!mlang[2];
         }
 
+        const defLocale = pxt.appTarget.appTheme.defaultLocale;
+        if (!force && !pxt.Util.isLocaleEnabled(useLang!)) {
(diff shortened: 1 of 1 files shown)


dmonroym (approved):

skillmap/src/App.tsx
```diff
@@ -164,9 +168,7 @@ class AppImpl extends React.Component<AppProps, AppState> {
             force: force,
         });
 
-        if (pxt.Util.isLocaleEnabled(useLang!)) {
-            pxt.BrowserUtils.setCookieLang(useLang!);
-        }
+        pxt.BrowserUtils.setCookieLang(useLang!);
```
abchatra, line 171: We set he cookie to userLang only if isLocaleEnabled right? Why this change?

skillmap/src/App.tsx
```diff
@@ -138,13 +138,17 @@ class AppImpl extends React.Component<AppProps, AppState> {
             force = !!mlang && !!mlang[2];
         }
 
+        const defLocale = pxt.appTarget.appTheme.defaultLocale;
+        if (!force && !pxt.Util.isLocaleEnabled(useLang!)) {
+            useLang = defLocale;
```
abchatra, line 143: I don't get this one as well. You are setting useLang if it is not isLocaleEnabled?

abchatra: Sorry ! tripped me off. This looks good.

skillmap/src/App.tsx
shakao: oh it's being covered by the above case (invalid languages are overwritten by the default), by the time it gets here the useLang should only be an enabled locale! i can add it back though, it doesn't hurt either way

abchatra: But when forcelang is used its not overridden right?

shakao: OH yes you're totally right, i'll replace the check! hmm @jwunderl do we want to set the user language when the forcelang flag is used? or should we also be skipping that?

jwunderl: I would lean towards maintaining the current behavior of not persisting unapproved languages / only keeping them for the session when forcelang is used; forcelang is typically just a test scenario anyway

jwunderl (approved):
Besides maybe not persisting forcelang lgtm~

skillmap/src/App.tsx
shakao: sounds good! i'm moving the setUserLanguage out of `updateLocalizationAsync` and next to the cookie setting, then wrapping them both in this check. this is different from the previous skillmap behavior but i think matches the webapp?

abchatra (approved):

abchatra:
port to stable

PR #8622, opened 2021-12-02
Use default preferences for users if none exist

livcheerful:
fixes https://github.com/microsoft/pxt-arcade/issues/4429

The issue is when we fetch the user preferences from the cloud, for new users they won't have any, and we don't pass in default preferences here. Adding default preferences


eanders-ms (approved):
Nice fix!

PR #8623, opened 2021-12-02
Display Back and Next buttons on simulator tab in tutorial

shakao:
image
image

build: https://arcade.makecode.com/app/417ebce3e68c882a7e60fc089eddf4028974077b-ca22d467a8

### DIFF
--- a/webapp/src/sidepanel.tsx
+++ b/webapp/src/sidepanel.tsx
@@ -2,14 +2,13 @@
 
 import * as React from "react";
 import * as data from "./data";
-import * as sui from "./sui";
-
 import * as filelist from "./filelist";
 import * as keymap from "./keymap";
 import * as serialindicator from "./serialindicator"
 import * as simtoolbar from "./simtoolbar";
 import * as simulator from "./simulator";
 
+import { fireClickOnEnter, Button } from "./sui";
 import { TabContent } from "./components/core/TabContent";
 import { TabPane } from "./components/core/TabPane";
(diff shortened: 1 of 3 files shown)


webapp/src/sidepanel.tsx
```diff
@@ -153,6 +155,11 @@ export class Sidepanel extends data.Component<SidepanelProps, SidepanelState> {
                         {showFileList && <filelist.FileList parent={parent} />}
                         {showFullscreenButton && <div id="miniSimOverlay" role="button" title={lf("Open in fullscreen")} onClick={this.handleSimOverlayClick} />}
                     </div>
+                    {isTabTutorial && <div className="tutorial-controls">
+                        { backButton }
+                        <Button icon="lightbulb" disabled={true} className="tutorial-hint" />
```
dmonroym, line 160: is it worth showing this button if it's always disabled?

dmonroym (approved):

webapp/src/sidepanel.tsx
shakao: yeah, technically all three of these are fake buttons (the back and next directly navigate you to the tutorial tab and don't change the step), the thought was that users are familiar with the three buttons at the bottom, so we shouldn't change the positioning/text/etc because a new kind of button might still be confusing

jwunderl (approved):

shakao:
fixes https://github.com/microsoft/pxt-arcade/issues/4395

PR #8631, opened 2021-12-14
New react-common button implementation and using fontawesome directly

riknoll:
This does a few things:

1. Implements a new button component in react-common
2. Replaces the buttons in the skillmap with the new button component
3. Rips out semantic icon classes from the skillmap and replaces them with fontawesome's less classes 
4. Adds fontawesome as an npm dependency

Eventually I will also replace the buttons/icons in the main webapp, but that is a PR for another day. This also does a little bit of tweaking across the board for how buttons look in the skillmap. They're now much closer to the webapp, but this will require a design pass later.

Also, there is a list component that I started implementing but isn't actually used anywhere yet.


skillmap/src/components/Modal.tsx
```diff
@@ -38,30 +40,29 @@ export class Modal extends React.Component<ModalProps> {
...
-                    {!fullscreen && <div className="modal-close-icon" onClick={this.handleCloseClick} role="button"><i className="icon close remove circle" /></div>}
+                    {!fullscreen && <div className="modal-close-icon" onClick={this.handleCloseClick} role="button"><i className="fas fa-times-circle" /></div>}
                 </div>
                 <div className="modal-body">
                     {this.props.children}
                 </div>
                 {actions && actions.length > 0 && <div className="modal-actions">
-                    {actions.map((el, i) => {
-                        return el.url
-                            ? <a key={i} className={`modal-button ${el.className || ""}`} href={el.url} onClick={el.onClick} target="_blank" rel="noopener noreferrer" role="button">{el.label}</a>
```
shakao, line 54: i think these were the only two places using `modal-button` so we could possibly delete all CSS related to it

react-common/components/controls/Button.tsx
```diff
@@ -0,0 +1,62 @@
...
+    let clickHandler = () => {
+        if (onClick) onClick();
+        if (href) window.open(href, target || "_blank", "noopener,noreferrer")
+    }
+
+    return (
+        <button
+            id={id}
+            className={classes}
+            title={title}
```
shakao, line 47: could have the title default to the label value unless otherwise specified? i think this is the behavior we generally want but it might be a little too much behind the scenes, idk

jwunderl: ^^ I think this would be good myself, we've historically missed a LOT of titles & had a lot of cases where it's just set to aria-label anyways

jwunderl (approved):

react-common/components/controls/Button.tsx
riknoll: So title is a required property and label is not. I did it that way so that icon buttons would always have a title, even when they don't have a label (e.g. they're just an icon). I can split this into two components if we want (Button and IconButton) and make label required in one and title required in the other.

shakao: hmm yeah i was trying to think of a way to avoid the duplication of the same string in the title/label all over the place but i do think i prefer to have a single button component.... not very strongly either way though

riknoll: Alright I'm going to stick with it as-is and we'll see if it's to annoying

PR #8632, opened 2021-12-15
Image Editor Resize

livcheerful:
https://arcade.makecode.com/app/4e9fa04d8b2d41666c2bc308572b02a4e737d047-600825771d#editor

Adds resizing handles to the marquee tool. I put them at the corners similarly to how Aesprite does it! Let me know how it feels for you -- I can change the size of the squares if they need to be bigger

image


webapp/src/components/ImageEditor/ImageCanvas.tsx
```diff
@@ -207,14 +214,19 @@ export class ImageCanvasImpl extends React.Component<ImageCanvasProps, {}> imple
             }
             return;
         }
-
+        if (this.touchesResize(coord.clientX, coord.clientY)) {
+            this.isResizing =  true;
```
jwunderl, line 218:
```suggestion
            this.isResizing = true;
```

spacing nit

webapp/src/components/ImageEditor/ImageCanvas.tsx
```diff
@@ -207,14 +214,19 @@ export class ImageCanvasImpl extends React.Component<ImageCanvasProps, {}> imple
...
+        }
         this.startEdit(!!isRightClick);
         this.updateEdit(this.cursorLocation[0], this.cursorLocation[1]);
         this.commitEdit();
     }
 
     onDragStart(coord: ClientCoordinates, isRightClick?: boolean): void {
         this.hasInteracted = true
+        if (this.touchesResize(coord.clientX, coord.clientY)) {
+            this.isResizing =  true;
```
jwunderl, line 228:
```suggestion
            this.isResizing = true;
```

jwunderl (approved):
Hm, separate from this we should work on the behavior when mouse leaves the canvas -- it's a little wonky for different tools but in particular the canvas, and makes scaling up / etc here a bit of a pain. That's separate from this change though! LGTM

livcheerful:
> Hm, separate from this we should work on the behavior when mouse leaves the canvas -- it's a little wonky for different tools but in particular the canvas, and makes scaling up / etc here a bit of a pain. That's separate from this change though! LGTM

LOL yes we do D:<

dmonroym:
This doesn't repro 100% of the time but if I've got a selection and try to do another selection, then resize, the selection box simply moves
selection

livcheerful:
> This doesn't repro 100% of the time but if I've got a selection and try to do another selection, then resize, the selection box simply moves selection

Yup that's a result of the resize handles being kinda hard to grab...I think? I will mess around with it a bit more. Easy solution might be to just make them bigger? but it also looks like at the last second before grabbing your mouse moves to the top right above the handle

webapp/src/components/ImageEditor/ImageCanvas.tsx
```diff
@@ -977,6 +1046,16 @@ export class ImageCanvasImpl extends React.Component<ImageCanvasProps, {}> imple
         return this.props.tool === ImageEditorTool.Pan;
     }
 
+    protected touchesResize(cursorX: number, cursorY: number) {
+        const hovered = document.querySelectorAll( ":hover" );
+        for ( let i = 0; i < hovered.length; i ++){
+            if (hovered[i].className ==  "image-editor-floating-layer-corner") {
```
riknoll, line 1052:
```suggestion
            if (hovered[i].className == "image-editor-floating-layer-corner") {
```

riknoll (approved):

dmonroym (approved):

PR #8648, opened 2022-01-11
New checkbox and loading spinner for react-common

riknoll:
Creating a new checkbox component that uses native checkboxes (better for accessibility). It's now tab navigable and has styling for when it's focused.

I also stole the spinner component from fluentui to replace the one we use in the async checkbox of the user profile since that one is from semantic. The fluentui is also much simpler implementation wise, so that's a bonus.


jwunderl (approved):

riknoll:
oh, whoops, i broke the webapp build. will reopen

riknoll:
Fixed! Ready for review

shakao:
also swap out loader here? https://github.com/microsoft/pxt/blob/master/skillmap/src/components/AppModal.tsx#L351

livcheerful (approved):

shakao (approved):

PR #8656, opened 2022-01-18
Adding a couple useful keyboard shortcuts to the image/tilemap editor

riknoll:
On Friday we did an art livestream and it reminded me of two keyboard shortcuts I've been meaning to add to the image editor forever. These are both very self contained and should be safe to include in the patch release

1. Outline image (image editor only): **ctrl + shift + [1-9a-f]**
This one outlines the current sprite in the color indicated by the number. Outlining sprites is a super tedious process to do by hand

2. Replace color/tile: **ctrl + shift + r**
Replaces the currently selected background color/tile with the currently selected foreground color/tile. Another super tedious thing to do by hand, especially for larger images with a lot of colors
replace color

### DIFF
--- a/webapp/src/components/ImageEditor/keyboardShortcuts.ts
+++ b/webapp/src/components/ImageEditor/keyboardShortcuts.ts
@@ -3,7 +3,7 @@ import { Store } from 'react-redux';
 import { ImageEditorTool, ImageEditorStore, TilemapState, AnimationState } from './store/imageReducer';
 import { dispatchChangeZoom, dispatchUndoImageEdit, dispatchRedoImageEdit, dispatchChangeImageTool, dispatchSwapBackgroundForeground, dispatchChangeSelectedColor, dispatchImageEdit} from './actions/dispatch';
 import { mainStore } from './store/imageStore';
-import { EditState, flipEdit, getEditState, rotateEdit } from './toolDefinitions';
+import { EditState, flipEdit, getEditState, outlineEdit, rotateEdit } from './toolDefinitions';
 let store = mainStore;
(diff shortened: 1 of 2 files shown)


aznhassan:
`ctrl + shift + r` will refresh the tab in Windows.

riknoll:
oh... right.... @shakao does photoshop have a shortcut for this? Aseprite uses just shift + R which I'm also okay with

shakao:
hmm i don't know of any photoshop shortcut for it. won't shift+r possibly get weird if people are typing asset names or something?

riknoll:
true! tempted to leave it as is... the preventdefault() should stop the refresh I think

riknoll:
maybe ctrl + shift + s (for swap)?

eanders-ms:
my 2c, I'd prefer ctrl + shift + s over + r. I don't think I would ever be comfortable pressing ctrl + shift +r in a browser for anything other than doing a hard page refresh.

riknoll:
changed! It's now ctrl+shift+s

aznhassan:
> changed! It's now ctrl+shift+s

That shortcut activates the snip tool in Edge...

eanders-ms:
Or, maybe just `s`? Sans modifiers.

riknoll:
just s conflicts with typing a name into the name field... though i guess we already handle that for tool switching

riknoll:
just confirmed that we do in fact suppress keyboard shortcuts when typing in the name field!

so i'm going to copy aseprite and do shift + r

shakao (approved):

PR #8659, opened 2022-01-20
React common menubar and menudropdown

riknoll:
New MenuBar, MenuDropdown, and FocusTrap components. I'll be documenting all of the accessibility stuff at some point, but here are the basics of how the menu works:

1. The entire menu is a single tab stop in the page navigation
2. When the menu is focused, the arrow keys can be used to navigate within the menu
3. When a menu dropdown is expanded, the submenu traps focus just like a modal. It can be closed using escape or by focusing something else on the page

This is all based on the [w3 example](https://www.w3.org/TR/wai-aria-practices-1.2/examples/menubar/menubar-navigation.html) and [fluent ui implementation](https://developer.microsoft.com/en-us/fluentui#/controls/web/commandbar)


jwunderl (approved):

react-common/styles/controls/Button.less
```diff
@@ -166,6 +166,10 @@
     background: rgba(0,0,0,.1);
 }
 
+.common-button.menu-button:active {
+    background: rgba(0,0,0,.15);
```
shakao, line 170: small/can do later: should this be in a less variable?

shakao: same for expanded dropdown bg below

react-common/components/controls/MenuBar.tsx
```diff
@@ -0,0 +1,94 @@
...
+        for (const element of focusable.values()) {
+            focusableElements.push(element as HTMLElement);
+
+            // Remove them from the tab order, menu items are navigable using the arrow keys
+            element.setAttribute("tabindex", "-1");
+            element.setAttribute("data-isfocusable", "true");
+        }
+    }
+
+    const onKeyDown = (e: React.KeyboardEvent<HTMLElement>) => {
```
shakao, line 37: also for later: we could consider moving this key nav stuff into a common util or react hook or something--not sure if there will be other components that use the arrow key navigation?

shakao (approved):

dmonroym (approved):

livcheerful (approved):

PR #8661, opened 2022-01-20
print label in build output

pelikhan:
Print label when refused for versioned build to diagnose build issues.


abchatra (approved):

pelikhan:
this pr is not technically needed since it turns out we print it at the start of the flow, does not hurt though. @abchatra  your call, you could just close it too.

abchatra:
Lets merge it. Good to see the label in the error message as well.

PR #8664, opened 2022-01-21
Enable high contrast styles for skillmap

shakao:
adds high contrast styles, adds settings item to toggle, hooks it up to the main editor high contrast (which is already all nicely piped through the auth service thanks to eric & vivian's work haha)

build: https://arcade.makecode.com/app/4883fe9e2d5a3716f460f3557bf4e7b7eb6fd973-0d61bad887--skillmap

### DIFF
--- a/pxteditor/editor.ts
+++ b/pxteditor/editor.ts
@@ -320,6 +320,7 @@ namespace pxt.editor {
         renderPythonAsync(req: EditorMessageRenderPythonRequest): Promise<EditorMessageRenderPythonResponse>;
 
         toggleHighContrast(): void;
+        setHighContrast(on: boolean): void;
         toggleGreenScreen(): void;
         toggleAccessibleBlocks(): void;
         setAccessibleBlocks(enabled: boolean): void;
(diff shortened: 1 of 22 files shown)


riknoll (approved):

riknoll:
hmmm i think the uploadtrg is broken for skillmap

abchatra:
How do I test the skillmap? Following link doesn't open for me. 
https://arcade.makecode.com/app/6642671ff704d47006b60c72bb9c6b276350b0a0-48f7bed654--skillmap

shakao:
ahh hm i wonder if that webpack alias fix for the local build broke the uploadtrg scenario.... let me investigate

shakao:
@riknoll @abchatra fixed build https://arcade.makecode.com/app/4883fe9e2d5a3716f460f3557bf4e7b7eb6fd973-0d61bad887--skillmap

abchatra:
Thanks this link works. 

I think setting high contrast in the skillmap should set high contrast in the iframe as well. Otherwise students have to go to editor and set it there and it is not intuitive.

jwunderl (approved):
It looks like the 'saving to cloud' got some style leak as it loads as black in non-high contrast mode now: 
image

(vs 
image
)

LGTM besides that & abhijiths point (which should just be loading with `?hc=1` iff high contrast?)

shakao:
@jwunderl nice catch, thanks! i don't want to fully reload the iframe when high contrast is toggled just cause that's kind of disruptive and error-prone, so i'm using messages to pipe that state through

PR #8671, opened 2022-01-25
Suggested fix for tutorial docs headings

### LINKED ISSUE #8670: Tutorials have curly braces
Hey Galen, can you look at this?
Seeing curly braces in the headings for the Tutorials
https://arcade.makecode.com/tutorials/happy-flower
image

ganicke:
The braces on the tutorial doc headings aren't removed when served from azure. I suspect there is a difference in the behavior of `string.trim()`. Here is a suggested fix for `render.heading` to address this. An additional change is in there to better form the header id's as well.

Fixes #8670

### DIFF
--- a/pxtlib/docsrender.ts
+++ b/pxtlib/docsrender.ts
@@ -441,7 +441,6 @@ namespace pxt.docs {
             // remove tutorial macros
             if (text)
                 text = text.replace(/@(fullscreen|unplugged|showdialog|showhint)/gi, '');
-            text = text.trim();
             // remove brackets for hiding step title
             if (text.match(/\{([\s\S]+)\}/))
                 text = text.substr(1, text.length - 2).trim()
--- a/pxtlib/docsrender.ts
+++ b/pxtlib/docsrender.ts
@@ -437,16 +437,17 @@ namespace pxt.docs {
             if (m) {
                 text = m[1]
                 id = m[2]
-            } else {
-                id = raw.toLowerCase().replace(/[^\w]+/g, '-')
             }
             // remove tutorial macros
             if (text)
                 text = text.replace(/@(fullscreen|unplugged|showdialog|showhint)/gi, '');
             text = text.trim();
             // remove brackets for hiding step title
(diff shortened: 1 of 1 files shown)


pxtlib/docsrender.ts
```diff
@@ -437,16 +437,16 @@ namespace pxt.docs {
...
-                id = raw.toLowerCase().replace(/[^\w]+/g, '-')
             }
             // remove tutorial macros
             if (text)
                 text = text.replace(/@(fullscreen|unplugged|showdialog|showhint)/gi, '');
-            text = text.trim();
             // remove brackets for hiding step title
-            if (text.match(/^\{([\s\S]+)\}$/))
-                text = text.substr(1, text.length - 2);
+            if (text.match(/\{([\s\S]+)\}/))
```
shakao, line 445: hmm i think maybe something like 
```
if (text.match(/\{([\s\S]+)\}/)?.[1])
     text = text.match(/\{([\s\S]+)\}/)[1].trim()
```
would ensure that we're just getting the text between the brackets.... if someone has a title that's like `{TITLE} other text` it would only translate to `TITLE` but maybe that's okay? would result in a cleaner ID anyway

ganicke: Right, to make it simple, maybe just have the "rules" be that the heading is either all-visible or all-hidden. Heading text that breaks the rules will just need to get fixed when discovered. Go ahead and make the changes as you see fit. Thanks.

shakao (approved):

pxtlib/docsrender.ts
shakao: cool, updated! (also great branch name hahaha)

ganicke: Thanks. BTW, this is the best regex I can come up with that discriminates something out of the first occurrence of "{blahblah123}".

`\{([^\{][\s\S]+?)\}`

yields: "Introduction " when the heading is crazy like:

`jjkkjk }}} {{Introduction }}{{ {more} or {less}`

PR #8674, opened 2022-01-31
Show ctrl+f 'Find...' option in context menu

shakao:


jwunderl (approved):

pxtblocks/blocklyloader.ts
```diff
@@ -1583,6 +1583,17 @@ namespace pxt.blocks {
                 options.push(screenshotOption);
             }
 
+            if (pxt.appTarget.appTheme.workspaceSearch) {
+                options.push({
+                    text: lf("Find..."),
+                    enabled: topBlocks.length > 0,
+                    callback: () => {
+                        pxt.tickEvent("blocks.context.workspacesearch", undefined, { interactiveConsent: true });
+                        this.getComponentManager()?.getComponent("workspaceSearch")?.open();
```
riknoll, line 1592: what is this here? does the arrow function not thwart this?

shakao: it's the workspace--configureContextMenu is called from the workspace svg which is what we want to access anyway (rather than the menu option itself)

riknoll: oh i see, it's within a normal function

riknoll (approved):

livcheerful (approved):

PR #8675, opened 2022-01-31
Fix asset editor modal high contrast, z-index

shakao:
fixes https://github.com/microsoft/pxt-arcade/issues/4544
fixes https://github.com/microsoft/pxt-arcade/issues/4549 

general CSS fixes for high contrast, updates the modal component to use a react portal so that we can render it at the root of the DOM

there's another issue with fontawesome icons not showing up in the webapp (looks like the font isn't included somehow?) which i'm still investigating

### DIFF
--- a/react-common/components/controls/Modal.tsx
+++ b/react-common/components/controls/Modal.tsx
@@ -24,7 +24,7 @@ export interface ModalProps extends ContainerProps {
     actions?: ModalAction[];
     onClose?: () => void;
     fullscreen?: boolean;
-    parentElement?: HTMLElement;
+    parentElement?: Element;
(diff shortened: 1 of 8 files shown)


jwunderl (approved):

riknoll:
did you try the fullscreen modals with this? e.g. the user profile in the skillmap. also the badge info modals within the fullscreen modal

shakao:
@riknoll oh yeah good call, moved the fullscreen one under the app container element so it has the correct CSS!

livcheerful (approved):

riknoll (approved):

PR #8676, opened 2022-01-31
Tab navigation for skill graph (the actual map part)

riknoll:
Adds tab navigation to the skill graph. Since SVGs don't play super nice with aria roles, I made it so that each graphnode now has a secret HTML button element behind it. All of the focus/tab navigation occurs on those buttons instead of the SVG elements.

The tab navigation works exactly like the menu bar: the entire graph is one tab stop and the buttons inside are navigable with the arrow keys. Pressing enter/space moves you into the info panel which traps focus until you escape. Right now, I'm not doing anything fancy here for the navigation order; it just goes left->right then top->bottom for any branches.

I plan to do more for the aria labels, descriptions, etc. but I might try to wrangle a Windows box to do it because voiceover on mac is not the best and not the primary use case.


jwunderl (approved):

skillmap/src/App.tsx
```diff
@@ -432,6 +435,14 @@ class AppImpl extends React.Component<AppProps, AppState> {
         }
     }
 
+    protected focusCurrentActivity = () => {
+        const node = document.querySelector("[data-activity=" + this.props.activityId + "] button");
+        (node as HTMLElement | SVGElement).focus();
```
shakao, line 440: can node be null? or is this pretty safe

shakao (approved):

skillmap/src/App.tsx
riknoll: nope, this should never be null

dmonroym (approved):

PR #8680, opened 2022-02-02
Skillmap activities 'Keep Code' by default, add 'Replace my code' to tutorial pane

shakao:
fixes https://github.com/microsoft/pxt-arcade/issues/4203

image

changes:
- add a TutorialCallout component that has the hint styling, so that we can reuse the bubble styles 
- adds the 'Replace my code' link and callout UI
- no longer deletes the template code from the header tutorialOptions; adds a 'templateLoaded' flag so that we can reload if necessary
- removes the keep code modal from the skillmap; always automatically carryover code

build: https://arcade.makecode.com/app/7cb86b6806a1f8147e41f0a59d907ca24aee0b15-694e1027e7--skillmap

### DIFF
--- a/theme/tutorial-sidebar.less
+++ b/theme/tutorial-sidebar.less
@@ -541,8 +541,6 @@
             flex-direction: column;
             align-items: center;
             justify-content: center;
-            width: 4rem;
-            height: unset;
             margin: 0;
             font-size: 1rem;
 
@@ -550,7 +548,6 @@
                 display: flex;
                 align-items: center;
                 justify-content: center;
-                margin-bottom: 0.5rem!important;
                 font-size: 2rem;
             }
 
@@ -575,6 +572,10 @@
         }
     }
 
+    .tutorial-container > .ui.button i.icon {
+        margin-bottom: 0.5rem!important;
+    }
+
     /*******************************
             Tutorial Hint
     *******************************/
@@ -586,7 +587,6 @@
     }
 
     .tutorial-hint .tutorial-callout-button.ui.button {
-        border-radius: 0.2rem;
         color: @tutorialPrimaryColor;
         background: @tutorialSecondaryColor;
         margin: 1rem;
@@ -621,8 +621,9 @@
 
     .tutorial-replace-code {
         position: absolute;
-        right: 5rem;
+        right: 4.5rem;
         bottom: -0.5rem;
+        width: 4rem;
 
         .tutorial-callout {
             top: 22.5rem;
@@ -632,6 +633,10 @@
         .tutorial-replace-code-actions .ui.button {
             width: unset;
         }
+    }
+
+    .tutorial-replace-code + .tutorial-controls > .tutorial-hint {
+        width: 5rem;
     }
 
 
--- a/theme/tutorial-sidebar.less
+++ b/theme/tutorial-sidebar.less
@@ -62,8 +62,12 @@
     position: absolute;
     height: 3rem;
     width: calc(~'100% - 1rem');
-    bottom: 9rem;
+    bottom: 6rem;
     background-image: linear-gradient(to bottom, rgba(255, 255, 255, 0), @white);
+}
+
+.tutorial-replace-code + .tutorial-scroll-gradient {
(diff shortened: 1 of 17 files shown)


theme/tutorial-sidebar.less
```diff
@@ -274,6 +295,41 @@
     border-bottom: 2px solid @tutorialPrimaryColor;
 }
 
+/*******************************
+    Tutorial Replace Code
+*******************************/
+.tutorial-replace-code {
+    display: flex;
+    justify-content: center;
+    font-family: 'Segoe UI', Roboto, 'Open Sans', 'Helvetica Neue', sans-serif;
```
riknoll, line 304: should this be in a variable?

theme/tutorial-sidebar.less
```diff
@@ -274,6 +295,41 @@
...
+
+    .tutorial-callout-button.ui.button {
+        padding: 0;
+        margin: 0 0 2rem;
+        color: #0064BF;
```
riknoll, line 309: also the color and font here

webapp/src/app.tsx
```diff
@@ -1881,6 +1882,24 @@ export class ProjectView
...
+        if (keepAssets) {
+            // Convert all temporary assets to named assets before we load in the template
+            let currentText = await workspace.getTextAsync(header.id);
+            const imageJres = appendTemporaryAssets(currentText[pxt.MAIN_BLOCKS], currentText[pxt.IMAGES_JRES]);
+            pkg.mainEditorPkg().setFile(pxt.IMAGES_JRES, imageJres);
+            await mainPkg.saveFilesAsync();
+        }
+
+        header.tutorial.templateLoaded = false;
+        delete header.tutorial.mergeHeaderId;
```
riknoll, line 1899: what happens if they restart the tutorial a second time? is that possible?

webapp/src/components/tutorial/TutorialCallout.tsx
```diff
@@ -0,0 +1,56 @@
...
+    const { children, className, buttonIcon, buttonLabel } = props;
+    const [ visible, setVisible ] = React.useState(false);
+
+    const captureEvent = (e: any) => {
+        e.preventDefault();
+        e.stopPropagation();
+        e.nativeEvent?.stopImmediatePropagation();
+    }
+
+    const closeCallout = (e: any) => {
```
riknoll, line 23: not for this pr, but this "click anywhere and dismiss" behavior would be a good candidate for react-common

jwunderl (approved):

webapp/src/app.tsx
shakao: yeah, when they restart it re-imports the code so the merge id is reset and there shouldn't be any issue (the mergeProjects call happens after the template code check, so previously it only ran once per project anyway). they can also reset the template code more than once and that shouldn't have any issues

webapp/src/components/tutorial/TutorialCallout.tsx
shakao: oh yeah definitely, i was a little torn on whether to start integrating the react-common components (buttons, etc) here, but stuck with SUI for now just for visual consistency within the webapp

dmonroym (approved):

webapp/src/app.tsx
```diff
@@ -1757,14 +1758,14 @@ export class ProjectView
...
-        if (!header || !header.tutorial || !header.tutorial.templateCode)
+        if (!header || !header.tutorial || !header.tutorial.templateCode || header.tutorial.templateLoaded)
             return;
 
         const template = header.tutorial.templateCode;
 
-        // Delete from the header so that we don't overwrite the user code if the tutorial is
-        // re-opened
-        delete header.tutorial.templateCode;
+        // Mark that the templated has been loaded so that we don't overwrite the
```
livcheerful, line 1766:
```suggestion
        // Mark that the template has been loaded so that we don't overwrite the
```
so small feel free to ignore

livcheerful (approved):

PR #8681, opened 2022-02-02
Marquee resize fixes :)

livcheerful:
GROUP A
fixes https://github.com/microsoft/pxt-arcade/issues/4540
fixes https://github.com/microsoft/pxt-arcade/issues/4543
fixes https://github.com/microsoft/pxt-arcade/issues/4548

GROUP B
fixes https://github.com/microsoft/pxt-arcade/issues/4541

GROUP C
fixes https://github.com/microsoft/pxt-arcade/issues/4563

This PR fixes 3 things, and the first thing fixes all the bugs in GROUP A.

Group A's fix: The cursor wasn't registering anything when it moved off the canvas, so I allowed the cursor to be updated even if it was beyond the canvas.

Group B's fix: There's a bunch of control points and canvas surfaces being made when doing the tile editor within the tilemap editor. So I added some rules to only add control points if it's NOT the tilemap editor, and to pick the LAST main canvas surface so it can properly calculate the resize. Before it was null-ing out. I added a fix for Group C too where we don't show the handles when it's not the marquee tool too

### DIFF
--- a/webapp/src/components/ImageEditor/ImageCanvas.tsx
+++ b/webapp/src/components/ImageEditor/ImageCanvas.tsx
@@ -96,15 +96,15 @@ export class ImageCanvasImpl extends React.Component<ImageCanvasProps, {}> imple
         return <div ref="canvas-bounds" className={`image-editor-canvas ${isPortrait ? "portrait" : "landscape"}`} onContextMenu={this.preventContextMenu}>
             <div className="paint-container">
                 {!this.props.lightMode && <canvas ref="paint-surface-bg" className="paint-surface" />}
-                <canvas ref="paint-surface" className="paint-surface" />
+                <canvas ref="paint-surface" className="paint-surface main" />
                 {overlayLayers.map((layer, index) => {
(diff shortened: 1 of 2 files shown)


webapp/src/components/ImageEditor/ImageCanvas.tsx
```diff
@@ -506,7 +505,7 @@ export class ImageCanvasImpl extends React.Component<ImageCanvasProps, {}> imple
     }
 
     protected updateEdit(x: number, y: number) {
-        if (this.edit && this.edit.inBounds(x, y)) {
+        if ((this.edit && this.edit.inBounds(x, y)) || (this.edit && this.isResizing)) {
```
jwunderl, line 508:
```suggestion
        if (this.edit && (this.edit.inBounds(x, y)) || this.isResizing)) {
```

nit; also we might want to consider this for ones like the line / circle / rect tool, it's annoying when you go off canvas slightly and it just... stops

jwunderl (approved):

webapp/src/components/ImageEditor/ImageCanvas.tsx
livcheerful: yes hahahahhahahaha

riknoll (approved):

shakao (approved):

PR #8697, opened 2022-02-07
Various skillmap fixes for February release

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/4584
Fixes https://github.com/microsoft/pxt-arcade/issues/4591
Fixes https://github.com/microsoft/pxt-arcade/issues/4578

Also fixes two un-filed bugs:

1. Modals were not respecting high-contrast. I moved the high-contrast class up to the document body
2. The sign in button was no longer present when no activity was selected in the skillmap

The commits break down to one fix per commit, so might be easier to review. I also refactored the focus outline less variables a bit because I was getting very confused as to what inverted/not-inverted meant.

### DIFF
--- a/skillmap/src/App.tsx
+++ b/skillmap/src/App.tsx
@@ -382,10 +382,10 @@ class AppImpl extends React.Component<AppProps, AppState> {
     }
 
     render() {
-        const { skillMaps, activityOpen, backgroundImageUrl, theme, highContrast } = this.props;
+        const { skillMaps, activityOpen, backgroundImageUrl, theme } = this.props;
         const { error, showingSyncLoader } = this.state;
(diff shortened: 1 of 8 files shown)


skillmap/src/App.tsx
```diff
@@ -361,6 +361,19 @@ class AppImpl extends React.Component<AppProps, AppState> {
         }
     }
 
+    componentDidUpdate() {
+        const { highContrast } = this.props;
+
+        const bodyIsHighContrast = document.body.classList.contains("high-contrast");
```
shakao, line 367: can we remove the high-contrast class from the app-container element with this change?

riknoll: done

shakao (approved):

PR #8705, opened 2022-02-09
Add media query breakpoints

riknoll:
This PR duplicates the semantic ui screen size breakpoints in react common and adds new macros to make them easier to read. I only added helpers for the max-width media queries for each screen size; @shakao and I talked it over and we felt the best way to do media queries was to always cascade from large screen to small

I think going forward we want to standardize on this style (cascading from large to small) rather than our current jumble of media queries which are hard to read/interpret. In particular we want to move away from queries that specify both a min and max width like this:

```less
@media only screen and (min-width: @tabletBreakpoint) and (max-width: @largestTabletScreen) {
}
```

As we component-ize things, we'll slowly start removing the old ones (unless folks object).

@livcheerful @jwunderl FYI

### DIFF
--- a/react-common/styles/profile/profile.less
+++ b/react-common/styles/profile/profile.less
@@ -335,7 +335,7 @@
     }
 }
 
-@media only screen and (max-width: 1200px) and (min-width: 992px) {
+@media @computerAndBelow {
     .profile-badges, .profile-badges-background {
         background-size: 25%;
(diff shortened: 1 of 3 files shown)


jwunderl (approved):
Sounds good to me~

riknoll:
Of course this is all open to debate if people like the old way!

livcheerful (approved):
A+ love

PR #8712, opened 2022-02-11
Chaning the name from kindscript to makecode

abchatra:
Documentation and other changes for prepping name change.

### DIFF
--- a/common-docs/faq.md
+++ b/common-docs/faq.md
@@ -58,7 +58,7 @@ When you download the program for your project, the name you gave the project be
 
 ## I just added / approved a new translation in Crowdin, when will it show up?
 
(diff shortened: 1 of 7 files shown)


pelikhan:
Can you ask crowdin for a redirect?

abchatra:
Yes we should ask that too.

abchatra:
@anzhou5 can you ask if they can do redirect if anyone tries to reach kindscript to new URL?

jwunderl (approved):
I believe there's an env variable in the backend (CROWDIN_PROJECT) set to kindscript as well that gets used when only when initializing a page for inContextTranslationMode (in serveFileAsync 
 in routing.ts)

abchatra:
@jwunderl is that part of env.json?

jwunderl:
Yes, it's in env.json

abchatra:
Ideally it should be in pxt-deployment-config. It is not a secret.

PR #8719, opened 2022-02-23
[stable7.4] port kindscript->makecode change

jwunderl:
for arcade

Porting non-docs versions of https://github.com/microsoft/pxt/commit/32723c6912405c067eedb7b6b4dc7a89dfd4a6ea to microbit, arcade, minecraft


livcheerful (approved):

jwunderl:
@riknoll looks like I have to port https://github.com/microsoft/pxt/pull/8717 as well? Is that expected (looks like it's popping up after a eslint patch update so I guess it's fine)

PR #8722, opened 2022-02-24
Allow literals to be compared in blocks by casting to any

riknoll:
Screen Shot 2022-02-23 at 4 32 00 PM


jwunderl (commented):
Do we need to do anything for python? e.g. if you take a program with if (5==7){} and convert to python would it work?

riknoll:
@jwunderl it now supports python! will add tests...

jwunderl (approved):

PR #8732, opened 2022-03-07
Save generated ID to the project header when the user shares

shakao:
when we add the persistent share UI we can hook it up to this as well. open to other thoughts on the type of version timestamp, i'm using "temporary" and "persistent" because that's what we've been calling them but not sure if "onetime"/"persistent" or something would be better


livcheerful (approved):
I don't have opinions on temporary/onetime! o(><；)oo looks good to me either wayyyyyy

riknoll (approved):
I vote we change temporary -> snapshot

shakao:
hmm what about "snapshot" / "permalink" ?

PR #8733, opened 2022-03-08
Add a secret accessibility menu to the editor toggle

riknoll:
Add a secret accessibility menu to the editor toggle

This is sort of similar to the secret menu you get when you tab in the main editor. This is to work around our UI pattern of a toggle with a dropdown which doesn't really have a corresponding aria role. The secret menu takes the form of a tab list where all navigable items are present (including the ones buried in the dropdown). The menu is hidden until it receives focus; then the current tab takes over the toggle.

The tab list is navigable using the arrow keys, just like a menu bar. Unlike a menu bar, the first item to get focus when you tab in is always the active tab.

### DIFF
--- a/react-common/components/controls/FocusList.tsx
+++ b/react-common/components/controls/FocusList.tsx
@@ -1,0 +1,120 @@
+import * as React from "react";
+import { ContainerProps } from "../util";
+
+export interface FocusListProps extends ContainerProps {
+    role: string;
+    childTabStopId?: string;
+}
+
+/**
+ * A list of focusable items that represents a single tab stop in the tab order. The
+ * children of the list can be navigated between using the arrow keys. Any child with
+ * a tabindex other than -1 will be included in the list.
+ *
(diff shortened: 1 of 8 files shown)


react-common/components/controls/FocusList.tsx
```diff
@@ -0,0 +1,120 @@
...
+            }
+            e.preventDefault();
+            e.stopPropagation();
+        }
+        else if (e.key === "Home") {
+            focusableElements[0].focus();
+            e.preventDefault();
+            e.stopPropagation();
+        }
+        else if (e.key === "End") {
```
jwunderl, line 101: Should there be an esc one here to skip 'out' / past this tab menu? Not sure since it's basically a fake menu so I can't really decide what I would expect the behavior

jwunderl (approved):

react-common/components/controls/FocusList.tsx
riknoll: you escape using tab

PR #8748, opened 2022-03-25
New share modal components in react-common

shakao:
react-common modal, new share components following Alex's fullscreen design for when screenshot/simulator is enabled and just the old modal design for when screenshots are disabled (microbit)

builds:
https://arcade.makecode.com/app/4fd34183d6d629192d9a075628999ec00a366090-b5a5921b34
https://makecode.microbit.org/app/548b0b0e67ceeaa611ad69936640cdd2ae58727f-93ebe2f317

i still need to clean up/reconnect the gif hotkeys but it's a little finicky because the gif encoder lives in the webapp (to avoid react-common dependencies) and all the gif state is in the new component, will do in a separate PR

### DIFF
--- a/react-common/components/share/GifInfo.tsx
+++ b/react-common/components/share/GifInfo.tsx
@@ -1,0 +1,55 @@
+import * as React from "react";
+import { Button } from "../controls/Button";
+import { GifRecorder } from "./GifRecorder";
+
+export interface GifInfoProps {
+    initialUri?: string;
+
+    onApply: (uri: string) => void;
+    onCancel: () => void;
+
+    screenshotAsync?: () => Promise<string>;
+    gifRecordAsync?: () => Promise<void>;
+    gifRenderAsync?: () => Promise<string | void>;
+}
+
+export const GifInfo = (props: GifInfoProps) => {
+    const { initialUri, onApply, onCancel, screenshotAsync, gifRecordAsync, gifRenderAsync } = props;
+    const [ uri, setUri ] =  React.useState(initialUri);
+
+    const handleApplyClick = (evt?: any) => {
+        onApply(uri);
+    }
+
+    const handleScreenshotClick = async () => {
+        const screenshotUri = await screenshotAsync();
+        setUri(screenshotUri);
+    }
+
+    const handleRecordStopClick = async () => {
+        const gifUri = await gifRenderAsync();
+        if (gifUri) setUri(gifUri);
+    }
+
+    return <>
+        <span className="thumbnail-label">{lf("Current Thumbnail")}</span>
+        <div className="thumbnail-image">
+            {uri
+                ? <img src={uri} />
+                : <div className="thumbnail-placeholder" />
+            }
+        </div>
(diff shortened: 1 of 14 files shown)


react-common/components/share/GifInfo.tsx
```diff
@@ -0,0 +1,55 @@
...
+        const screenshotUri = await screenshotAsync();
+        setUri(screenshotUri);
+    }
+
+    const handleRecordStopClick = async () => {
+        const gifUri = await gifRenderAsync();
+        if (gifUri) setUri(gifUri);
+    }
+
+    return <>
```
riknoll, line 34: is there a reason this isn't wrapped in a div?

react-common/components/share/Share.tsx
```diff
@@ -0,0 +1,44 @@
...
+    gifRenderAsync: () => Promise<string | void>;
+    publishAsync: (name: string, screenshotUri?: string) => Promise<ShareData>;
+    showModalAsync(options: DialogOptions): Promise<void>;
+}
+
+export const Share = (props: ShareProps) => {
+    const { projectName, screenshotUri, screenshotAsync, gifRecordAsync,
+        gifRenderAsync, publishAsync, showModalAsync } = props;
+
+return <div className="project-share">
```
riknoll, line 33: nit: whitespace

react-common/components/share/Share.tsx
```diff
@@ -0,0 +1,44 @@
...
+    }
+    qr?: string;
+    error?: any;
+}
+
+export interface ShareProps {
+    projectName: string;
+    screenshotUri?: string;
+
+    screenshotAsync: () => Promise<string>;
```
riknoll, line 22: is there a reason these are optional in shareinfo and not here?

riknoll (approved):

riknoll:
In the arcade build, I can't seem to get gif recording to work

shakao:
@riknoll hmm so the R hotkey doesn't work yet, but you should be able to click "record" -> do some stuff in the sim -> click "stop recording" and see the changes in the preview on the side? what's happening when you try to record?

riknoll:
the thumbnail never changes. it does work with the screenshot, just not the gif.

riknoll:
okay figured out a repro:

1. Press record
2. Move around in the game until it automatically stops recording (i.e. don't hit stop recording, wait until the GIF is too long and the simulator stops)
3. See that the ui breaks

shakao:
ah! forgot about the auto-stop for the recording, i'll update

shakao:
@riknoll fixed! turns out the work to fix the auto-stop was the same work as fixing the keyboard shortcuts and removing the leftover gif state from share.tsx, so keyboard shortcuts should also now work.

updated builds:
https://arcade.makecode.com/app/16fe11cf642aa4a5c252b520978a75d192b38ae7-dfbee4a739
https://makecode.microbit.org/app/c7adef8fd6892d4e436869ecedb15059ddd9855d-8631ebca3a

riknoll (approved):

PR #8755, opened 2022-03-31
Hook logged in share up to persistent API, add dropdown for temporary share

shakao:
plus some cleanup in `share.tsx` ! basically hooks the "Publish" button up to the new persistent share API when the user is logged in and adds a ellipsis menu with a "Create snapshot" option for generating the temporary link. the not-logged-in behavior should be exactly the same as before

requires https://github.com/microsoft/pxt-backend/pull/626 to be deployed


jwunderl (approved):

shakao:
updated build with deployed backend changes! https://arcade.makecode.com/app/1958c63dc9796642d263abc7f6d0b6eee1ccc131-ed65fdbb17

PR #8761, opened 2022-04-05
Add slug to packages.github message to make queries easier

riknoll:
The name parameter just includes whatever the user typed, which can have a number of inconsistencies:

1. trailing slash
2. random uppercase/lowercase letters
3. may not include the repo owner if the extension is approved (e.g. pxt-color instead of jwunderl/pxt-color)

The slug is much more consistently formatted and should always be in the format owner/repo


jwunderl (approved):

pelikhan (approved):

abchatra (approved):

abchatra:
This is exactly what we should do. Improve telemetry when it is not good enough.

PR #8763, opened 2022-04-05
Always try to resolve the latest release for github dependencies

pelikhan:
Currently unbound github dependencies (no version tag specified) are bound to "master". The desired behavior should be to grab the latest release. There is also some caching involved which leads to some confusing situation where things might not work based on the state of the cache.

## Fixes

- [x] This fix auto-binding of no-tag -> latest tag + bailout to "master" if needed; in the in-memory github database.
- [x] various functions rewritten to async for readability

## Test

This project won't load in the current beta https://github.com/pelikhan/pxt-kitronik-air-quality/blob/master/jacdac/pxt.json if you happen to have an older version of jacdac cached.

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -232,6 +232,10 @@ class FileGithubDb implements pxt.github.IGithubDb {
             })
     }
 
+    latestVersionAsync(repopath: string, config: pxt.PackagesConfig): Promise<string> {
+        return this.db.latestVersionAsync(repopath, config)
+    }
+
     loadConfigAsync(repopath: string, tag: string): Promise<pxt.PackageConfig> {
         return this.loadAsync(repopath, tag, "pxt", (r, t) => this.db.loadConfigAsync(r, t));
     }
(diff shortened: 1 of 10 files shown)


cli/cli.ts
```diff
@@ -251,49 +255,6 @@ function searchAsync(...query: string[]) {
         })
 }
 
-function pkginfoAsync(repopath: string) {
```
pelikhan, line 254: looks like dead code

webapp/src/data.ts
```diff
@@ -284,13 +284,6 @@ mountVirtualApi("gh-search", {
     isOffline: () => !Cloud.isOnline(),
 })
 
-mountVirtualApi("gh-pkgcfg", {
```
pelikhan, line 287: looks like deadcode

webapp/src/data.ts
```diff
@@ -315,7 +308,6 @@ mountVirtualApi("target-config", {
                         pxt.storage.setLocal("targetconfig", JSON.stringify(js))
                         invalidate("target-config");
                         invalidate("gh-search");
-                        invalidate("gh-pkgcfg");
```
pelikhan, line 318: route is unused

jwunderl (approved):

PR #8779, opened 2022-04-14
Extensions Update

livcheerful:


webapp/src/extensionsBrowser.tsx
```diff
@@ -0,0 +1,453 @@
...
+        </MenuBar>
+        <div className="extensionSearchHeader">
+            <div className="header">{(lf("Do more with your micro:bit"))}</div>
+            <SearchInput searchHandler={setSearchFor} />
+            <div className="extensionTags">
+                {categoryNames.map(c =>
+                    <div className={"extensionTag " + (selectedTag == c ? "selected" : "")} onClick={() => handleCategoryClick(c)}>{c}</div>
+                )}
+            </div>
+            {/* <div className="importButton">
```
riknoll, line 416: delete?

webapp/src/extensionsBrowser.tsx
```diff
@@ -0,0 +1,453 @@
...
+                </div> :
+                currentTab == TabState.Recommended ? preferredExts.map(e => <ExtensionCard scr={e} name={e.name} onCardClick={installExtension} imageUrl={e.imageUrl} description={e.description}
+                        learnMoreUrl={e.fullName ? `/pkg/${e.fullName}`: undefined} loading={e.loading} />) :
+                    currentTab == TabState.Installed ? installedExtensions.map(e =>
+                        <ExtensionCard scr={e} name={e.name} onCardClick={() => handleInstalledCardClick(e)} imageUrl={e.imageUrl} description={e.description} learnMoreUrl={e.fullName ? `/pkg/${e.fullName}` : undefined}/>
+                    ) :
+                        local.forEach(p => {
+                            <ExtensionCard scr={p} name={p.name} description={lf("Local copy of {0} hosted on github.com", p.githubId)} onCardClick={addLocal} />
+                        })
+            }
```
riknoll, line 450: these nested ternaries are a little hard to parse. is there any way to break them up?

webapp/src/toolbox.tsx
```diff
@@ -362,6 +376,29 @@ export class Toolbox extends data.Component<ToolboxProps, ToolboxState> {
...
     }
 
+    private async deleteExtension(ns: string) {
+        this.setState({
+            tryToDeleteNamespace: undefined
+        })
+        const extensionsToDelete = [];
+        const extension = pkg.findExtForNs(ns);
+        extensionsToDelete.push(extension.id)
+        while ( extensionsToDelete.length > 0) {
```
riknoll, line 386:
```suggestion
        while (extensionsToDelete.length > 0) {
```

riknoll (changes requested):

react-common/components/extensions/ExtensionCard.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from "react";
+
+interface ExtensionCardProps extends pxt.CodeCard {
+    scr: any;
```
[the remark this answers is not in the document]
livcheerful: ohhhh jajajajajajaj fixing

webapp/src/toolboxeditor.tsx
```diff
@@ -151,6 +154,25 @@ export abstract class ToolboxEditor extends srceditor.Editor {
...
+                const namespaces = pkg.getNsForPkg(ext.id)
+                if (!namespaces) return false
+                return namespaces.indexOf(ns) >= 0
+            })
+            const extensionPkg = foundExtension?.getKsPkg();
+
+            let trgConfigFetch = data.getDataWithStatus("target-config:");
+            let trgConfig = trgConfigFetch.data as pxt.TargetConfig;
+
+            const hasDel = (nsAttr._def && nsAttr._def.parts.length > 0) || (foundExtension && extensionPkg.id != "core" && !trgConfig.packages.extensionsToIgnore.includes(extensionPkg.id))
```
jwunderl, line 172: do we need to account for ---{variant}s here? e.g. when you select hardware to download and it swaps over to hardware variant pkgs https://github.com/microsoft/pxt-common-packages/blob/master/libs/core---stm32/pxt.json#L2

localtypings/pxtarget.d.ts
```diff
@@ -40,13 +40,27 @@ declare namespace pxt {
...
+        preferredRepoLib?: RepoData[];
+        approvedRepoLib?: RepoData[];
         // format:
         // "acme-corp/pxt-widget": "min:v0.1.2" - auto-upgrade to that version
         // "acme-corp/pxt-widget": "dv:foo,bar" - add "disablesVariant": ["foo", "bar"] to pxt.json
         upgrades?: pxt.Map<string>;
         // list of trusted custom editor extension urls
         // that can bypass consent and send/receive messages
         approvedEditorExtensionUrls?: string[];
+        extensionsToIgnore?: string[]; // List of extensions to ignore when allowing for deletion
```
jwunderl, line 52: nit: my reading of this / the code using looks like this field is used to mark 'extensions that we don't want to mark as isExtension', but my first thought was that it was 'extensions to hide from search / tag results' -- maybe 'unmarkedExtensions' or something similar to that? Not sure

react-common/components/extensions/DeleteConfirmationModal.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from "react";
+import { Modal } from "../controls/Modal"
+
+export interface ImportModalProps {
+    onCancelClick: () => void,
+    onDeleteClick: (url: string) => void,
+    ns: string
```
jwunderl, line 7:
```suggestion
    onCancelClick: () => void;
    onDeleteClick: (url: string) => void;
    ns: string;
```

nit

react-common/components/extensions/ExtensionCard.tsx
```diff
@@ -0,0 +1,31 @@
...
+    function handleClick() {
+        props.onCardClick(props.scr);
+    }
+
+    function handleLearnMoreClick(e: React.MouseEvent) {
+        e.stopPropagation();
+        window.open(props.learnMoreUrl, "_blank")
+    }
+
+    return <div className={(props.loading ? "loading ": "") +"ui card extensionCard"} onClick={handleClick}>
```
jwunderl, line 20:
```suggestion
    return <div className={(props.loading ? "loading ": "") + "ui card extensionCard"} onClick={handleClick}>
```

nit

react-common/components/extensions/ExtensionCard.tsx
```diff
@@ -0,0 +1,31 @@
...
+        <img src={props.imageUrl}/>
+        <div className="content">
+            <div className="name">{props.name}</div>
+            <div className="description">{props.description}</div>
+        </div>
+
+        {props.learnMoreUrl ?
+            <div className="extra content">
+                <a className="learnmore" onClick={handleLearnMoreClick} rel="noopener noreferrer">{lf("Learn More")}</a>
+            </div>: undefined}
```
jwunderl, line 30:
```suggestion
            </div> : undefined}
```

nit

webapp/src/blocks.tsx
```diff
@@ -1392,6 +1392,7 @@ export class Editor extends toolboxeditor.ToolboxEditor {
     }
 
     private filterBlocks(subns: string, blocks: toolbox.BlockDefinition[]) {
+        if (!blocks) return [];
```
jwunderl, line 1395: question - is there a known case where blocks is not set that we're already hitting? This change seems fine but just want to make sure it's not suppressing a real subtle issue (or does this just end up falling to a path of "filterBlocks returns empty so just return all blocks")

webapp/src/extensionsBrowser.tsx
```diff
@@ -0,0 +1,453 @@
...
+        }
+    }
+    function handleHomeButtonClick() {
+        setSelectedTag("")
+        setSearchFor("")
+    }
+
+    function fetchBundled(): Map<string, ExtensionMeta> {
+        const bundled = pxt.appTarget.bundledpkgs;
+        const extensionsMap = new Map<string, ExtensionMeta>();
```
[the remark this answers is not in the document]
jwunderl: it's another real obj, but we don't really use it https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

maybe better to use `= {}: pxt.Map<string>`

webapp/src/blocks.tsx
livcheerful: no I think was only hit with the new changes

localtypings/pxtarget.d.ts
livcheerful: yeah it's being used for hiding for toolbox, but I think maybe it'd be good to use to hide from installed extensions too? Wanted to see if people had feedback about it during testing

livcheerful:
Here's a target, but you need to change the targetconfig in localstorage manually to the one here: https://github.com/microsoft/pxt-microbit/pull/4511

https://makecode.microbit.org/app/87ec16f5347fed149ce24af3c61429b780e4a4c6-3e52918070

webapp/src/extensionsBrowser.tsx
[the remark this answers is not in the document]
livcheerful: ohh is there....a reason we don't use it?

webapp/src/extensionsBrowser.tsx
livcheerful: I want to bring in the import button.....eventually. But the modal was kinda tricky to do in time. Should I still delete and just add back later?

webapp/src/extensionsBrowser.tsx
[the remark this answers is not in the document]
jwunderl: no important ones, originally wasn't used due to inconsistent browser support / being a feature that was relatively new / unused but it's available everywhere now. just pretty very minor things

* since we use the Map type used all over the place it can be a little misleading at a glance -- doesn't really matter and easy to find when you're aware but if you e.g. pass it as a parameter or try to store it incorrectly as a pxt.Map the message is not ideal for debugging 
image
* whenever our existing functions use object literals they don't play well / have to convert to pass them in which is a bit magicky -- have to know to use `new Map<string, any>(Object.entries(obj))` for plain object -> map and `Object.fromEntries(myMap)` for map -> obj. 

which is not a big deal at all, just have to make sure everyone remembers that the map builtin type exists if we want to use it to avoid confusion there

pelikhan:
can you provide a built editor link?

localtypings/pxtpackage.d.ts
```diff
@@ -73,6 +73,7 @@ declare namespace pxt {
         weight?: number;
         gistId?: string;
         extension?: PackageExtension; // describe the associated extension if any
+        isExtension?: boolean; // is this package an extension
```
pelikhan, line 76: who is supposed to set this flag? is there a UI for it?
what about the existing extension, do they have to be updated?

react-common/components/extensions/ExtensionCard.tsx
pelikhan: maybe time to start running prettier systematically

webapp/src/package.ts
```diff
@@ -764,6 +781,48 @@ export function getEditorPkg(p: pxt.Package): EditorPackage {
...
+
+export function findExtForNs(ns: string): pxt.Package {
+    const pkgId = Object.keys(mainPkg.deps).find( dep => {
+        const associatedNamespaces = getNsForPkg(dep)
+        return associatedNamespaces?.includes(ns)
+    })
+    return mainPkg.deps[pkgId]
+}
+
+export function getNsForPkg(pkgId: string) {
```
pelikhan, line 808: how do you handle radio-broadcast who adds blocks to "radio"?

pelikhan:
IMO, "Do more with your micro:bit" is a nice slogan but we need to have a descriptive name so that people can reference to this screen (and verify that they are in the correct screen). There is nothing in the UI that says "extensions" although that's the keyword.

pelikhan:
On the 'tagged extensions' screenshot, some extensions are duplicated.

livcheerful:
> On the 'tagged extensions' screenshot, some extensions are duplicated.

Yessss I saw that after and push a fix for that! Good eye :D

livcheerful:
> can you provide a built editor link?

oh yes so here's a link, but you have to manually update the targetconfig from the pxt-microbit pr :(

https://github.com/microsoft/pxt-microbit/pull/4511

[makecode.microbit.org/app/87ec16f5347fed149ce24af3c61429b780e4a4c6-3e52918070](https://makecode.microbit.org/app/87ec16f5347fed149ce24af3c61429b780e4a4c6-3e52918070)

localtypings/pxtpackage.d.ts
livcheerful: This flag gets set automatically in the toolbox code!

webapp/src/extensionsBrowser.tsx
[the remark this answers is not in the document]
riknoll: huh TIL

webapp/src/extensionsBrowser.tsx
riknoll: you can leave it, just add a comment!

react-common/components/extensions/DeleteConfirmationModal.tsx
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export const DeleteConfirmationModal = (props: ImportModalProps) => {
+    const actions = [
+        {label: lf("Cancel"), onClick: props.onCancelClick},
+        {label: lf("Remove"), className: "red", onClick: () => {props.onDeleteClick(props.ns)}}
+    ]
+    return <Modal title={lf("Removing extension")} actions={actions} onClose={props.onCancelClick}>
+        {lf("Are you sure you want to remove this extension? Doing so will remove any of its blocks that you used in your code and other extensions that use it.")}
```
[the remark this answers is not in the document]
livcheerful: yeah 1000% agree

webapp/src/extensionsBrowser.tsx
```diff
@@ -0,0 +1,453 @@
...
+import { DeleteConfirmationModal } from "../../react-common/components/extensions/DeleteConfirmationModal";
+import { ExtensionCard } from "../../react-common/components/extensions/ExtensionCard";
+
+type ExtensionMeta = pxtc.service.ExtensionMeta;
+const emptyCard = { name: "", loading: true }
+
+interface ExtensionsProps {
+    isVisible: boolean;
+    hideExtensions: () => void;
+    parent: pxt.editor.IProjectView;
```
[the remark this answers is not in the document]
livcheerful: :( okay i tried to just pass in reloadHeadersAsync but it got mad about not finding the headers

webapp/src/extensionsBrowser.tsx
```diff
@@ -0,0 +1,453 @@
...
+        })
+        setAllExtensions(addedExtensions);
+    }
+
+    function getExtensionFromFetched(extensionUrl: string) {
+        const fullName = allExtensions.get(extensionUrl.toLowerCase())
+        if (fullName) {
+            return fullName
+        }
+        const githubRepoName = /\S*\/(\S*)#[\S]*/g.exec(extensionUrl)
```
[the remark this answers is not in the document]
livcheerful: !!! fixed great tip haha

webapp/src/extensionsBrowser.tsx
```diff
@@ -0,0 +1,453 @@
...
+            const repoName = allExtensions.get(githubRepoNameWithoutHash[1])
+            return repoName;
+        }
+        return undefined
+    }
+
+    async function removeDepAsync(dep: ExtensionMeta) {
+        setDeletionCandidate(undefined)
+        props.hideExtensions()
+        pkg.mainEditorPkg().removeDepAsync(dep.name)
```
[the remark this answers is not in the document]
livcheerful: (yes o.o fixed)

webapp/src/extensionsBrowser.tsx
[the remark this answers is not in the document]
riknoll: Did you pass in the function directly? try wrapping it in an arrow function, e.g.:

```tsx
 <Extensions reloadHeaderAsync={() => this.reloadHeadersAsync()} />
```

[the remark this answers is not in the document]
livcheerful: ohhh bet let me try

webapp/src/package.ts
livcheerful: There'll just be one x for the whole category, and it'll delete radio and radio-broadcast. I think the eventual plan is to change all the extensions that add blocks to an existing category to make their own (since I think we're the only ones who do that?)

localtypings/pxtarget.d.ts
```diff
@@ -40,13 +40,27 @@ declare namespace pxt {
...
+        categories?: ExtensionCategory[];
+    }
+
+    interface RepoData {
+        slug: string;
+        tags: string[]
```
jwunderl, line 58: This is pretty minor, but looking at the other pr there are a lot with empty tags arrays, it might be nice to set this as optional field / make sure it's guarded against it being undefined wherever it's referenced -- targetconfig is loaded live and updated regularly so it can help minimize future mistakes by minimizing the amount of things that are required to be set.

livcheerful: ah yeah i'll change it!

pelikhan:
getting this crash when importing a jacdac repo
image

image

I did not update target-config, but we should be defensive vs this as it may have a delay updating or be caching.

webapp/src/package.ts
pelikhan: what if an extension add a block to "basic"? :)

pelikhan:
There should be a way to filter out v2 only extensions, like microphone or data blocks.

webapp/src/package.ts
livcheerful: Do you know if any extensions do that? I think since basic isn't considered an extension to begin with, it won't be deleted (whereas radio is a package!)

abchatra: I don't think this is a concern that extensions add to basic. We can decide not approve such extensions.

webapp/src/extensionsBrowser.tsx
```diff
@@ -0,0 +1,478 @@
...
+            }
+        }
+    }, [searchFor])
+
+    /**
+     * Github search
+     */
+    async function searchForGithubAsync() {
+        setExtensionsToShow([emptyCard, emptyCard, emptyCard, emptyCard])
+        const exts = await fetchGithubDataAsync([searchFor])
```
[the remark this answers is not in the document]
eanders-ms: May be true for other methods here as well

webapp/src/extensionsBrowser.tsx
```diff
@@ -0,0 +1,478 @@
...
+                    extensions: {
+                        srcs: Array.from(allExtensions.values())
+                    }
+                }).then(e => {
+                    setExtensionsToShow(e)
+                    setSelectedTag("")
+                })
+            }
+        }
+    }, [searchFor])
```
eanders-ms, line 81: Does `allExtensions` need to be listed in dependencies?

eanders-ms (commented):
A few comments...

webapp/src/extensionsBrowser.tsx
livcheerful: I think no, adding it causes an infinite loop

webapp/src/extensionsBrowser.tsx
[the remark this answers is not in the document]
eanders-ms: Actually no, it isn't a future bug. useCallback would be an optimization but not a fix for anything. Please disregard!

webapp/src/extensionsBrowser.tsx
```diff
@@ -0,0 +1,478 @@
...
+        const parsedGithubRepo = pxt.github.parseRepoId(extensionUrl)
+        if (!parsedGithubRepo) return undefined;
+        return allExtensions.get(parsedGithubRepo.slug.toLowerCase())
+    }
+
+    async function removeDepAsync(dep: ExtensionMeta) {
+        setDeletionCandidate(undefined)
+        props.hideExtensions()
+        await pkg.mainEditorPkg().removeDepAsync(dep.name)
+        await pxt.Util.delay(1000) // TODO VVN: Without a delay the reload still tries to load the extension
```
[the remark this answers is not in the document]
eanders-ms: Wondering if we understand the underlying reason for needing a delay here. It would be worth a comment, for posterity.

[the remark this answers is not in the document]
livcheerful: +1 agree we should understand why we need a delay. My TODO here is more to figure this out than....needing to fix it? I think it's shippable as it is -- it doesn't feel like anything at all?

Is my current comment okay? Would you like to see something else haha

eanders-ms:
In a new blank microbit project, it looks like I can delete the radio category. Is that expected?

image

livcheerful:
> In a new blank microbit project, it looks like I can delete the radio category. Is that expected?
> 
> image

Yup that's expected. We can add it to the list of extensions that we don't want to be able to delete by default though in the pxt-microbit targetconfig!

eanders-ms:
Running your branch locally, I quickly started getting throttled when just opening/closing the extension view twice. It looks like it's making a bunch of calls all at once. Maybe that's triggering the throttle.
image

eanders-ms (approved):
:shipit:!
I wasn't able to repro the throttling issue again. If we get into beta, more folks can try it. Maybe we'll get a repro there.

pelikhan:
Do you have the stack trace to those requests?

eanders-ms:
> Do you have the stack trace to those requests?

```
Error: Bad HTTP status code: 429 at https://www.makecode.com/api/gh/SGBotic/pxt-SGBotic-Ultimate-SR04-RGB; message: {"message":""}
    at pxtapp.js:1921:37
    at async repoAsync (pxtapp.js:10502:28)
    at async Promise.all (index 31)
pxt.reportException @ pxtapp.js:3633
pxt.reportException @ pxtapp.js:56
(anonymous) @ pxtapp.js:10532
Promise.catch (async)
proxyRepoAsync @ pxtapp.js:10531
repoAsync @ pxtapp.js:10502
(anonymous) @ pxtapp.js:10541
searchAsync @ pxtapp.js:10541
(anonymous) @ data.js:232
Promise.then (async)
getAsync @ data.js:232
queueNotify @ data.js:115
(anonymous) @ data.js:178
getAsync @ data.js:174
fetchGithubDataAsync @ extensionsBrowser.js:172
fetchGithubDataAndAddAsync @ extensionsBrowser.js:175
handleCategoryClick @ extensionsBrowser.js:250
onClick @ extensionsBrowser.js:388
callCallback @ react-dom.development.js:336
invokeGuardedCallbackDev @ react-dom.development.js:385
invokeGuardedCallback @ react-dom.development.js:440
invokeGuardedCallbackAndCatchFirstError @ react-dom.development.js:454
executeDispatch @ react-dom.development.js:584
executeDispatchesInOrder @ react-dom.development.js:609
executeDispatchesAndRelease @ react-dom.development.js:713
executeDispatchesAndReleaseTopLevel @ react-dom.development.js:722
forEachAccumulated @ react-dom.development.js:694
runEventsInBatch @ react-dom.development.js:739
runExtractedPluginEventsInBatch @ react-dom.development.js:880
handleTopLevel @ react-dom.development.js:5807
batchedEventUpdates$1 @ react-dom.development.js:24368
batchedEventUpdates @ react-dom.development.js:1419
dispatchEventForPluginEventSystem @ react-dom.development.js:5898
attemptToDispatchEvent @ react-dom.development.js:6014
dispatchEvent @ react-dom.development.js:5918
unstable_runWithPriority @ scheduler.development.js:818
runWithPriority$2 @ react-dom.development.js:12130
discreteUpdates$1 @ react-dom.development.js:24384
discreteUpdates @ react-dom.development.js:1442
dispatchDiscreteEvent @ react-dom.development.js:5885
```

eanders-ms:
Oh that one happened in a search.

pelikhan:
Yes this is expected.

PR #8780, opened 2022-04-14
Page to detect reachable domains for MakeCode

abchatra:
This site will work even if the css doesn't load. Only thing this has hard dependency is the main @homeurl@ itself. 

loading


abchatra:
Fixes https://github.com/microsoft/pxt-microbit/issues/4115

common-docs/domains.html
```diff
@@ -0,0 +1,151 @@
...
+        content="Reachable MakeCode domains"
+    />
+
+    <meta name="msapplication-config" content="none" />
+    <meta name="theme-color" content="#3DC1AA" />
+
+    <title>Reachable MakeCode domains</title>
+
+    <link rel="stylesheet" href="https://pxt.azureedge.net/blob/2163189fd5e35c0981ed55318415582a7c9aeb12/doccdn/semantic.css" type="text/css">
+    <link rel="stylesheet" href="domains/style.css" type="text/css">
```
jwunderl, line 36: typically we put css files under static e.g. `/static/domains/style.css`; in this case since it's such a small page already it might be better just to directly insert into a style tag here vs. dealing with backend fetching the right file / etc

jwunderl (approved):
lgtm

common-docs/domains.html
abchatra: Good point. I think we need this page to be up irrespective of other pages. I will in line the style tag.

PR #8785, opened 2022-04-17
Ensure github token is present before creating subproject

pelikhan:
Before starting the process of populating a sub-folder in a repo, make sure we have a github token.


pelikhan:
@riknoll @jwunderl @eanders-ms  low risk change, mostly for my sanity

jwunderl (approved):

PR #8786, opened 2022-04-18
Better mute behavior and fix sound queuing

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/4541
Fixes https://github.com/microsoft/pxt-microbit/issues/4537

This restructures the audio context manager so that nodes that produce sound now all pass through a single gain node that can be used to mute/unmute sounds. Previously we just disconnected nodes whenever we received a mute request which made it tricky to restart sounds when the mute ends.

This also fixes a bug where long running soundexpressions weren't being cancelled properly when the sim restarted.

### DIFF
--- a/pxtsim/simlib.ts
+++ b/pxtsim/simlib.ts
@@ -261,10 +261,10 @@ namespace pxsim {
 
             const ctx = context();
             if (mute) {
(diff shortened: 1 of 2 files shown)


eanders-ms (approved):
Seems good to me

jwunderl (approved):

PR #8794, opened 2022-04-21
Add flag to make a block the child of another in the toolbox

riknoll:
Let blocks specify that they should be a the child of another block. This can sort of be thought of as a more generalized version of the flag that makes it so a block is within a "variable set" block. This is in support of the requested changes for 

https://github.com/microsoft/pxt-microbit/pull/4545


jwunderl (approved):
Ah great, this is a nice option to have. I've wanted this before but always forgot to file an issue or implement it haha

PR #8798, opened 2022-04-25
Tweaks to the new share dialog to make it more like the old one

riknoll:
https://makecode.microbit.org/app/2d53ab608c8370a19300c47ed485d6c25b1bc99d-f7e828fc73

Fixes https://github.com/microsoft/pxt-microbit/issues/4524
Fixes https://github.com/microsoft/pxt-microbit/issues/4504
Fixes https://github.com/microsoft/pxt-microbit/issues/4529
Fixes https://github.com/microsoft/pxt-microbit/issues/4505

### DIFF
--- a/react-common/components/share/ShareInfo.tsx
+++ b/react-common/components/share/ShareInfo.tsx
@@ -35,6 +35,7 @@ export const ShareInfo = (props: ShareInfoProps) => {
     const [ showQRCode, setShowQRCode ] = React.useState(false);
 
     const showSimulator = !!screenshotAsync || !!gifRecordAsync;
+    const showDescription = shareState !== "publish";
 
     React.useEffect(() => {
         setThumbnailUri(screenshotUri)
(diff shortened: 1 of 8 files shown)


react-common/components/share/ShareInfo.tsx
```diff
@@ -195,6 +239,14 @@ export const ShareInfo = (props: ShareInfoProps) => {
                 gifAddFrame={gifAddFrame}
                 registerSimulatorMsgHandler={registerSimulatorMsgHandler}
                 unregisterSimulatorMsgHandler={unregisterSimulatorMsgHandler} />}
+
+            {showQRCode &&
```
jwunderl, line 243: nit: could we get back the onclick behavior of the qr code display here (the 'toggle back to normal share screen' / exit big qr screen)?  it feels more natural to me vs. clicking x which I would almost expect to toggle out of the whole share dialog.

jwunderl (approved):
nit on one missing qr code behavior but otherwise lgtm

PR #8803, opened 2022-04-26
Add app theme flag for using legacy tutorial layout

riknoll:
Forces the editor to use the old tutorial layout. While the layout is the legacy one, the other new tutorial features like better markdown rendering still work fine. Here's a view of a tutorial in arcade that uses the fancy icon bullets:

Screen Shot 2022-04-26 at 4 28 27 PM


Jaqster (approved):
Awesome, thanks for enabling this Richard!

abchatra (approved):

PR #8806, opened 2022-04-28
[core] Support for online multiplayer

eanders-ms:
Adds support for online multiplayer messaging. Changes in this PR:
* Added a new optional URL parameter to sim frame: `mp=client|server`. When passed to the simulator, the new `MultiplayerState` board component is enabled.

Associated PRs:
https://github.com/microsoft/pxt-common-packages/pull/1339
https://github.com/microsoft/pxt-arcade-sim/pull/55


pxtsim/simdriver.ts
```diff
@@ -314,7 +315,7 @@ namespace pxsim {
                 const single = !!this._currentRuntime?.single;
                 const parentWindow = window.parent && window.parent !== window.window
                     ? window.parent : window.opener;
-                if (this.options.nestedEditorSim && parentWindow) {
+                if (parentWindow) {
```
jwunderl, line 318: After taking another look I think we might want to separate this out from the two connected 'else if's that follow now that it's not limited to `nestedEditorSim` -- otherwise those won't trigger in any embed scenario. So just leaving this if as is & having line 325 cut out the 'else' ~

jwunderl (approved):

pxtsim/simdriver.ts
eanders-ms: When remove the else on 325, I get multiple simulators instanced. I think I need a better understanding of this function.

jwunderl: Ahh gotcha, I see. We were kinda bypassing an issue with how it was set up by having it in the same branch.

This issue is that broadcast messages we use to send multiplayer state are often used to represent e.g. radio, so it falls into the the path below where it says "oh you want to simulate something that communicates with other boards? Okay, make another sim".

So, to get that to work nicely, we should probably do two things -- add an optional param to SimulatorBroadCastMessage (https://github.com/microsoft/pxt/blob/viewBtnControls/pxtsim/embed.ts#L81) for 'toParentIFrameOnly' which we'll set to true in the call to `Runtime.postMessage` (in `pxt-common-packages/libs/game/sim/multiplayer.ts`), and wrap the second two branches that start on 325 of this in another `if (!this.options.nestedEditorSim && !broadcastmsg?.toParentIFrameOnly)` to avoid double sim when we only really want to post to the outer frame

webapp/public/run.html
```diff
@@ -166,6 +167,10 @@
                     // propagate message packets to parent frames
                     window.parent.postMessage(msg.data, "*");
                 }
+                else if (mpRole && data.type === "multiplayer") {
+                    // propagate message packets to parent frames
+                    window.parent.postMessage(msg.data, "*");
```
eanders-ms, line 172: @jwunderl I'm seeing incorrect behavior here in production (beta). But I think it makes sense. This issue is that the client isn't rendering the screen. What I believe is happening: on the client, this method receives a "broadcast-screen" message, passed to this frame from the parent. And what we do here it pass it right back to the parent! I can confirm it is getting received by the client parent. I wasn't seeing this behavior when hosting locally, for reasons unknown.

My understanding is this method can receive messages from the child iframe or the parent. I think the solution is that we need to know which direction to route the message. Does that sound right to you? Do we do this elsewhere?

eanders-ms: Made a fix here: https://github.com/microsoft/pxt/pull/8814

PR #8813, opened 2022-05-03
Fix py for-loop conversion when using three range arguments

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/4403


abchatra:
Looks good. Add some description to the PR for future reference.

abchatra (approved):

dmonroym (approved):

PR #8814, opened 2022-05-04
Simulator: fix for cross-domain bidirectional multiplayer messages

eanders-ms:
Hosting the simulator within another app is a multiple nested iframe situation. The middle frame - run.html - has the responsibility of marshaling multiplayer message traffic to/from the simulator component, which is hosted in a child iframe. Because multiplayer messaging is bidirectional, traffic passing through the middle frame's singular event handler must be explicitly routed inbound or outbound.

In development, things were working fine on localhost without this routing, but I'm not certain why. There might be a special case somewhere, or the same-origin nature of localhost resulted in different behavior. In any case, this change makes it work in both scenarios.

### DIFF
--- a/pxtrunner/runner.ts
+++ b/pxtrunner/runner.ts
@@ -363,7 +363,7 @@ namespace pxt.runner {
             })
     }
 
-    export async function simulateAsync(container: HTMLElement, simOptions: SimulateOptions): Promise<pxtc.BuiltSimJsInfo> {
+    export async function simulateAsync(container: HTMLElement, simOptions: SimulateOptions): Promise<pxsim.SimulatorDriver> {
         const builtSimJS = simOptions.builtJsInfo || await buildSimJsInfo(simOptions);
         const {
             js,
@@ -411,7 +411,7 @@ namespace pxt.runner {
                 ? pxt.appTarget.simulator.partsAspectRatio
                 : pxt.appTarget.simulator.aspectRatio;
         driver.run(js, runOptions);
-        return builtSimJS;
+        return driver;
     }
 
     export async function buildSimJsInfo(simOptions: SimulateOptions): Promise<pxtc.BuiltSimJsInfo> {
--- a/webapp/public/run.html
+++ b/webapp/public/run.html
@@ -116,6 +116,7 @@
         var single = !!/single(?:[:=])1/i.test(window.location.href);
         var server = !!/server(?:[:=])1/i.test(window.location.href);
         var mpRole = /[\&\?]mp=(server|client)/i.exec(window.location.href)?.[1]?.toLowerCase();
+        var simdriver = null; /*pxsim.SimulatorDriver;*/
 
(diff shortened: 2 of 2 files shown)


webapp/public/run.html
```diff
@@ -168,8 +169,14 @@
...
                 else if (mpRole && data.type === "multiplayer") {
-                    // propagate message packets to parent frames
-                    window.parent.postMessage(msg.data, "*");
+                    // Multiplayer messages are bi-directional
+                    if (mpRole === msg.data.origin) {
+                        // propagate to parent frame
+                        window.parent.postMessage(msg.data, "*");
+                    } else {
+                        // propagate to child frames
+                        if (simdriver) simdriver.postMessage(msg.data);
```
jwunderl, line 178: I'm not certain I understand the need for posting to simdriver here -- as those aren't getting lost / get handled elsewhere right as it had previously gotten sent to the sim iframe, so would this double-send the message?

webapp/public/run.html
```diff
@@ -168,8 +169,14 @@
                     window.parent.postMessage(msg.data, "*");
                 }
                 else if (mpRole && data.type === "multiplayer") {
-                    // propagate message packets to parent frames
-                    window.parent.postMessage(msg.data, "*");
+                    // Multiplayer messages are bi-directional
+                    if (mpRole === msg.data.origin) {
```
jwunderl, line 173: would a

```
if (event.source !== window.parent)
```

work to prevent passing back / forth? I believe we do that elsewhere so it feels like it must have just gotten missed / dropped here

pxtrunner/runner.ts
```diff
@@ -363,7 +363,7 @@ namespace pxt.runner {
             })
     }
 
-    export async function simulateAsync(container: HTMLElement, simOptions: SimulateOptions): Promise<pxtc.BuiltSimJsInfo> {
+    export async function simulateAsync(container: HTMLElement, simOptions: SimulateOptions): Promise<pxsim.SimulatorDriver> {
```
jwunderl, line 366: we actually use the value returned here in arcades share page for perf when toggling between editor view & full screen game -- so ideally we should either not modify or return an object with both values available

webapp/public/run.html
eanders-ms: The intention here is to simply forward multiplayer messages along. Messages from the parent are forwarded to the simulator (via simdriver). Messages from the simulator are forwarded to the parent. It isn't clear how a double-send could happen. Is there something I'm not considering? There are so many message handlers scattered about. I might have missed where these messages are forwarded.

jwunderl: ah, what I meant was that if it was working on localhost, then the messages were getting passed to child frames somewhere else -- which would mean adding this would add a second path that intended to pass the same message from run.html -> simframes, right (with the first path not working right currently on live which might be related to missing parentOrigin)?

PR #8815, opened 2022-05-04
allow extension to hide parts

pelikhan:
Adding a new block attribute, `hiddenParts` that allows to filter out specific parts in the parts computation phase.

```
//% hideParts=servo <-- don't collect 'servo' parts
function someHighLevelfunctionUsingServo() {
    ...
    // this triggers part 'servo'
    pins.setServoValue(...)
}
```

In Jacdac, we use API that enable specific part but eventually the rendering is disabled. Downside of this flag is that it is pretty blunt and does not discriminate where the original part was used.


riknoll (approved):
i think this is probably fine... having both the jacdac ui and the servo ui in any case would be a bad scenario. probably not worth coming up with a more elegant solution

PR #8819, opened 2022-05-04
Various CSS Fixes

aznhassan:
### Problem
There are a number of UI issues related to the new extension page UI/UX. 

### Solution
* Add help button to the react common modal when a help URL is specified
* Change extensions view to use react-common modal to fix issues related to the editor page underneath peeking through
* Category buttons can be tabbed
* Category buttons can no longer have their text selectable
* Category buttons can by selected via the _enter_ key
* Category buttons now have a hover state
* Category buttons now have a focus state
* Extension cards can be navigated with tab
* Learn more links can be navigated to with tab
* Tab focus stays on the modal page
* Fix regression where extension cards didn't get smaller on mobile displays
* Extensions card icons now load when they intersect the page
* Category extension results are now cached for 60 minutes

### Validation
You can view the changes yourself at: https://microbit.staging.pxt.io/app/7d4f2cbcdc290eee97e1adf36a18cdb11eab985d-09292bbbde

### Notes
Fixes microsoft/pxt-microbit#4591
Fixes microsoft/pxt-microbit#4580
Fixes microsoft/pxt-microbit#4579 (@riknoll You also mentioned a border-right property being set on the _In Development_ tab which does not appear to be the case)
Fixes microsoft/pxt-microbit#4572
Fixes microsoft/pxt-microbit#4571
Fixes (and caused): microsoft/pxt-microbit/issues/4617


react-common/components/controls/Modal.tsx
```diff
@@ -65,6 +67,7 @@ export const Modal = (props: ModalProps) => {
                 {fullscreen &&
                     <div className="common-modal-back">
                         <Button
+                            key={"back-button"}
```
riknoll, line 70: key is not required unless the element is in a list

react-common/components/controls/Modal.tsx
```diff
@@ -76,9 +79,21 @@ export const Modal = (props: ModalProps) => {
                 <div id="modal-title" className="common-modal-title">
                     {title}
                 </div>
+                {fullscreen && helpUrl &&
+                    <div className="common-modal-help">
+                        <Button
+                            key={"help-button"}
```
riknoll, line 85:
```suggestion
```

react-common/components/controls/Modal.tsx
```diff
@@ -76,9 +79,21 @@ export const Modal = (props: ModalProps) => {
...
+                            title={lf("Help on {0} dialog", title)}
+                            onClick={() => { window.open(props.helpUrl, "_blank") }}
+                            rightIcon={"icon help"}
+                        />
+                    </div>
+                }
                 {!fullscreen &&
                     <div className="common-modal-close">
                         <Button
+                            key={"close-button"}
```
riknoll, line 96:
```suggestion
```

theme/common.less
```diff
@@ -882,23 +882,36 @@ Field editors
...
+.extension-card.loading {
+    background: @loadingAnimation;
+    background-size: 400% 400%;
+    animation: cardGradient 3s infinite alternate;
+}
 
-    .extensionsHeader {
+
+.fullscreen.extensions-browser> {
+    .common-modal> .common-modal-header {
```
riknoll, line 909:
```suggestion
    .common-modal > .common-modal-header {
```

theme/common.less
```diff
@@ -882,23 +882,36 @@ Field editors
...
+
+.extension-card.loading {
+    background: @loadingAnimation;
+    background-size: 400% 400%;
+    animation: cardGradient 3s infinite alternate;
+}
 
-    .extensionsHeader {
+
+.fullscreen.extensions-browser> {
```
riknoll, line 908:
```suggestion
.fullscreen.extensions-browser > {
```

theme/common.less
```diff
@@ -917,8 +930,26 @@ Field editors
             justify-content: center;
         }
     }
+}
+@media @mobileAndBelow {
+    .extensions-browser {
+        .common-modal> .common-modal-header> .common-modal-back {
```
riknoll, line 936:
```suggestion
        .common-modal > .common-modal-header > .common-modal-back {
```

theme/common.less
```diff
@@ -1025,13 +1062,80 @@ Field editors
         }
     }
 
+    .common-modal> .common-modal-body {
```
riknoll, line 1065:
```suggestion
    .common-modal > .common-modal-body {
```

theme/common.less
```diff
@@ -1025,13 +1062,80 @@ Field editors
...
     .common-modal-body {
         display: flex;
         flex-direction: column;
     }
 }
+
+// reduce all card sizes
+@media @mobileAndBelow {
+    .extensions-browser {
+        .ui.card, .ui.cards>.card {
```
riknoll, line 1126:
```suggestion
        .ui.card, .ui.cards > .card {
```

webapp/src/extensionsBrowser.tsx
```diff
@@ -33,31 +33,19 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
 
     const [searchFor, setSearchFor] = useState("");
     const [allExtensions, setAllExtensions] = useState(fetchBundled());
-    const [extensionsToShow, setExtensionsToShow] = useState([]);
+    const [extensionsToShow, setExtensionsToShow] = useState(new Array<ExtensionMeta & EmptyCard>());
```
riknoll, line 36: in general, no one really uses array constructors in js. if you wan to hint the type, you can do it like this
```suggestion
    const [extensionsToShow, setExtensionsToShow] = useState<(ExtensionMeta & EmptyCard)[]>([]);
```

webapp/src/extensionsBrowser.tsx
```diff
@@ -33,31 +33,19 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
...
     const [selectedTag, setSelectedTag] = useState("");
     const [currentTab, setCurrentTab] = useState(TabState.Recommended);
     const [showImportExtensionDialog, setShowImportExtensionDialog] = useState(false);
-    const [installedExtensions, setInstalledExtensions] = useState([])
-    const [lastVisibleState, setLastVisibleState] = useState(props.isVisible)
+    const [installedExtensions, setInstalledExtensions] = useState(new Array<ExtensionMeta & EmptyCard>())
```
riknoll, line 40:
```suggestion
    const [installedExtensions, setInstalledExtensions] = useState<(ExtensionMeta & EmptyCard)[]>([])
```

webapp/src/extensionsBrowser.tsx
```diff
@@ -33,31 +33,19 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
...
     const [deletionCandidate, setDeletionCandidate] = useState(undefined)
-    const [preferredExts, setPreferredExts] = useState([])
+    const [preferredExts, setPreferredExts] = useState(new Array<ExtensionMeta & EmptyCard>())
```
riknoll, line 42:
```suggestion
    const [preferredExts, setPreferredExts] = useState<(ExtensionMeta & EmptyCard)[]>([])
```

webapp/src/extensionsBrowser.tsx
```diff
@@ -408,71 +405,148 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
...
+    return (
+        <Modal
+            title={lf("Extensions")}
+            fullscreen={true}
+            className={"extensions-browser"}
+            onClose={props.hideExtensions}
+            helpUrl={"/extensions"}
+        >
+            <div className="ui">
+                {showImportExtensionDialog ? <ImportModal onCancelClick={() => setShowImportExtensionDialog(false)} onImportClick={handleImportUrl} /> : undefined}
```
riknoll, line 418: use && instead of a ternary (also the line below). For tags with multiple props, it's also nice to break them out into multiple lines

webapp/src/extensionsBrowser.tsx
```diff
@@ -408,71 +405,148 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
...
+                {deletionCandidate ? <DeleteConfirmationModal ns={deletionCandidate.name} onCancelClick={() => { setDeletionCandidate(undefined) }} onDeleteClick={() => { removeDepAsync(deletionCandidate) }} /> : undefined}
+                <div className="extension-search-header">
+                    <div className="header">{(lf("Do more with your micro:bit"))}</div>
```
riknoll, line 421: instead of hardcoding micro:bit, use the boardname. I think it's stored in the apptarget

webapp/src/extensionsBrowser.tsx
```diff
@@ -408,71 +405,148 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
...
+                    <SearchInput searchHandler={setSearchFor} key={"search-input"}/>
```
riknoll, line 422:
```suggestion
                    <SearchInput searchHandler={setSearchFor}/>
```

webapp/src/extensionsBrowser.tsx
```diff
@@ -408,71 +405,148 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
...
+                    <div className="extension-tags">
+                        {categoryNames.map(c =>
+                            <Button title={lf(c)}
```
riknoll, line 425: instead of lf, use rlf. lf only works with static strings, rlf will do a lookup against the list of known strings.

webapp/src/extensionsBrowser.tsx
```diff
@@ -408,71 +405,148 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
...
-                            name={scr.name} imageUrl={scr.imageUrl} description={scr.description} loading={scr.loading} />)}
-                </div>
-            </div>}
-        {displayMode == ExtensionView.Tags &&
-            <div className="extension-display">
-                <div className="breadcrumbs">
-                    <span className="link" onClick={handleHomeButtonClick}>{lf("Home")}</span>
-                    <span>/</span>
-                    <span>{selectedTag}</span>
+    let loadingCardId = 0
```
riknoll, line 408: i don't think you need this variable. in the callbacks to map below, an index is also passed in after the element

webapp/src/extensionsBrowser.tsx
```diff
@@ -408,71 +405,148 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
...
+                    <div className="extension-display">
+                        <div className="breadcrumbs">
+                            <span className="link" onClick={handleHomeButtonClick}>{lf("Home")}</span>
+                            <span>/</span>
+                            <span>{selectedTag}</span>
+                        </div>
+                        <div className="ui cards left">
+                            {extensionsToShow?.map(scr =>
+                                <ExtensionCard
+                                    key={scr.loading ? `loading ${loadingCardId++}`: 'tagged:' + scr.name}
```
riknoll, line 469: you can probably just pass the index here. i'm not sure you need the loading/tagged bit either

webapp/src/extensionsBrowser.tsx
```diff
@@ -408,71 +405,148 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
...
+                                    learnMoreUrl={scr.fullName ? `/pkg/${scr.fullName}` : undefined}
+                                    loading={scr.loading}
+                                    role="button"
+                                />)}
+                        </div>
+                    </div>}
+                {displayMode == ExtensionView.Tabbed &&
+                    <div className="extension-display">
+                        <div className="tab-header">
+                            <Button key={"Recommended"} title={lf("Recommended")} label={lf("Recommended")} onClick={() => { setCurrentTab(TabState.Recommended) }} className={currentTab == TabState.Recommended ? "selected" : ""} />
```
riknoll, line 484: break these into multiple lines

riknoll (changes requested):

webapp/src/data.ts
```diff
@@ -269,6 +269,14 @@ mountVirtualApi("cloud-search", {
     isOffline: () => !Cloud.isOnline(),
 })
 
+mountVirtualApi("extension-search", {
+    getAsync: query => pxt.targetConfigAsync()
+        .then(config => pxt.github.searchAsync(stripProtocol(query), config ? config.packages : undefined))
```
pelikhan, line 274: `config ? config.packages : undefined` -> `config?.packages`

webapp/src/extensionsBrowser.tsx
```diff
@@ -339,14 +332,15 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
         }
         setPreferredExts([...repos, ...loadingCards])
 
-        fetchGithubDataAndAddAsync(toBeFetched, (exts) => setPreferredExts([...repos, ...exts]));
+        const exts = await fetchGithubDataAndAddAsync(toBeFetched);
+        setPreferredExts([...repos, ...exts])
```
pelikhan, line 336: the page might be unmounted here already

pelikhan:
in general, having "useEffect + leaking promises" is a recipe for some fun race conditions between the promise and the page lifecycle (eg promise returns when the page is already unmounted)

webapp/src/extensionsBrowser.tsx
aznhassan: Ah, sweet! I definitely was aiming for just knowing the type so that when referencing it later on I could see its members.

webapp/src/extensionsBrowser.tsx
aznhassan: Interestingly, the name for the micro:bit target is actually _microbit_.

riknoll: @aznhassan you want pxt.appTarget.appTheme.boardName. that should have the colon!

aznhassan: Sweet. I changed this to `Do more with <boardName>` since `Do more with Arcade` would sound bad for example.

webapp/src/extensionsBrowser.tsx
riknoll: I don't think this is actually a problem in this case because this is going through the virtualapi. this will just be a no-op

riknoll (approved):

webapp/src/extensionsBrowser.tsx
aznhassan: Calling setState on an unmounted component is a no-op (though it will print a warning to the console). The long-running async operations all use the virtual API (retrieving GitHub repos) which has no concept of cancellation tokens to stop operations when it no longer makes sense to continue. All we could do is check the cancellation token after the long-running virtual API has returned before calling setState, which has the same effect as just calling setState. We could add cancellation tokens to virtual API to have them return sooner, but that is outside the scope of this PR. I think adding cancellation tokens and validation functions (validating that the results are what you expect before caching them) would be useful extensions of the virtual API. Thoughts?

@pelikhan

aznhassan: Leaving this be for now. We discussed creating a new hook that uses the subscribe and unsubscribe features of the virtual API to do away with the promises (thus not calling the setState functions if unsubscribed) but that work is outside the scope of this PR. Please feel free to create an issue, I'll see if I can work it in to a future PR.

PR #8826, opened 2022-05-09
Add Search Placeholder Text

aznhassan:
* Add search placeholder text
* Add an aria message and paragraph to let the user know there were no search results
* Remove 'Do more with your micro:bit' to match Alex's design

Latest: https://microbit.staging.pxt.io/app/c2a02986df7a1bc570d667e648740413eede93e8-8c6cbca310
Previous:
* https://microbit.staging.pxt.io/app/4ad87889fbd28313f3c8f0ab8dcd02b867c2fbbc-75b1423254

### DIFF
--- a/webapp/src/extensionsBrowser.tsx
+++ b/webapp/src/extensionsBrowser.tsx
@@ -429,7 +429,6 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
                     />
                 }
                 <div className="extension-search-header">
-                    <div className="header">{(lf(`Do more with ${pxt.appTarget.appTheme.boardName}`))}</div>
                     <SearchInput
                         ariaMessage={searchComplete && pxt.U.rlf(`${extensionsToShow.length} results matching '${searchFor}'`)}
                         placeholder={lf("Search or enter project URL...")}
--- a/webapp/src/extensionsBrowser.tsx
+++ b/webapp/src/extensionsBrowser.tsx
@@ -430,7 +430,11 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
                 }
                 <div className="extension-search-header">
                     <div className="header">{(lf(`Do more with ${pxt.appTarget.appTheme.boardName}`))}</div>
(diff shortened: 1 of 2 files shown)


aznhassan:
Related to microsoft/pxt-microbit/issues/4577 but this build still uses the breadcrumbs, so it doesn't fix the issue entirely.

webapp/src/extensionsBrowser.tsx
```diff
@@ -463,6 +469,11 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
                                     role="button"
                                 />)}
                         </div>
+                        {searchComplete && extensionsToShow.length == 0 &&
+                            <div aria-label="Extension search results">
+                                <p>{lf("We couldn't find any extensions matching ")}{pxt.Util.rlf(`'${searchFor}'`)}</p>
```
riknoll, line 474: Use the format string (e.g. `{0}`)to append the string instead of putting it at the end. Different language grammars might require it to be placed elsewhere in the sentence.

Also, does it make more sense to put the rlf on the inner string? I mean, to not include the quotes when doing the lookup and just adding them to the result

riknoll (changes requested):

webapp/src/extensionsBrowser.tsx
jwunderl: do we want to translation for the inner rlf at all? If I'm reading right that's a value from user input in the first place so it'd make or sense to leave as whatever they typed in (that is, if they type in 'controller' to the searchbox while in spanish it'd be weird to show it as 'Controles')

that said we could do an lf with `lf("'${0}'", searchFor)` to translate only the surrounding quotation marks to the given language's format

aznhassan: Won't the inner string, `searchFor` already be in the appropriate language since it's user input?

aznhassan: I've updated it to use format strings!

jwunderl (approved):

riknoll (approved):

PR #8833, opened 2022-05-11
Remove deletion of extension from the toolbox

abchatra:
This removes the code logic to remove the extension from the toolbox. Basically, isTopLevelExtension returns false, so user doesn't see any option to delete the extension. 

I will do a separate check-in adding the logic to enable this feature later. Idea is to allow only top-level extension to be deleted and if there is any block is used, we won't allow for deletion.

### DIFF
--- a/webapp/src/toolbox.tsx
+++ b/webapp/src/toolbox.tsx
@@ -380,15 +380,9 @@ export class Toolbox extends data.Component<ToolboxProps, ToolboxState> {
         this.setState({
             tryToDeleteNamespace: undefined
         })
-        const extensionsToDelete = [];
-        const extension = pkg.findExtForNs(ns);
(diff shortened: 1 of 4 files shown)


webapp/src/toolbox.tsx
```diff
@@ -685,7 +679,7 @@ export interface ToolboxCategory {
 
     customClick?: (theEditor: editor.ToolboxEditor) => boolean;
     advanced?: boolean; /*@internal*/
-    isExtension?: boolean;
+    delete?: boolean;
```
eanders-ms, line 682: `delete` is a reserved word. What about `deletable` or `allowDelete`.

abchatra: Changed to allowDelete

eanders-ms (approved):
:shipit:

PR #8835, opened 2022-05-13
Fix new tutorial in headless mode

riknoll:
The simulator was broken in headless mode even though the UI worked fine.

The fix is just to make sure the div exists. This change puts it in the DOM, but it isn't visible anywhere.


jwunderl (approved):

abchatra (approved):

riknoll:
added another quick css fix and an option to remove the "replace my code" button for templated tutorials

PR #8846, opened 2022-05-19
Fix crash when runoptions is not defined

riknoll:
micro:bit beta currently can't open any projects because of https://github.com/microsoft/pxt/pull/8844/files

Turns out, runOptions can be undefined, which is news to me!


jwunderl (approved):
Yeah that's pretty surprising to me too hah

PR #8853, opened 2022-05-19
Add Card and LazyImage to react-common and use in extension dialog

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/4635
Fixes https://github.com/microsoft/pxt-microbit/issues/4626

Adds two new generic components to react-common:

1. Card - A container for other components that acts like a button. Also includes support for those ribbon labels (most the css I just copied from semantic for those)
2. LazyImage - An image that is lazy loaded when it appears on screen. Also shows a spinner while it is loading.

This also uses those components in the extension browser to make things look much nicer. Changes include:

1. Text is no longer huge. Now follows the conventions for the rest of the webapp
2. Descriptions now get ellipsized after three lines
3. Extension cards and images now show loading spinners when they are loading and fade in when they finish

Here's an upload target:

https://makecode.microbit.org/app/2592776d32dddb4e0d913b66caafcf8f15b8c992-1ca81c68e3

### DIFF
--- a/react-common/components/controls/Card.tsx
+++ b/react-common/components/controls/Card.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { classList, ContainerProps } from "../util";
+import { classList, ContainerProps, fireClickOnEnter } from "../util";
 
 export interface CardProps extends ContainerProps {
     onClick?: () => void;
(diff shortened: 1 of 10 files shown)


jwunderl:
I see this hitting /null for some images: 
image
when opening led / robotics tags & also an infinite spin on 'individual' tag with bluetooth-temperature-sensor after that 404s -- 
image

image

riknoll:
@jwunderl that extension actually doesn't have an image. i think there's a separate issue for handling that case... in the current beta it just shows a broken link

jwunderl (approved):

riknoll:
@jwunderl I just fixed the null request error also. I'll update the uploadtrg link

PR #8854, opened 2022-05-20
support upgrade rules for nested repoes

pelikhan:
Use the slug instead of fullname to resolve upgrade rules, so that nested repoes are also always upgraded.


pelikhan:
Fix for https://github.com/microsoft/pxt-microbit/pull/4668

jwunderl (approved):

PR #8855, opened 2022-05-20
Add Import File Button in Extension Browser

aznhassan:
#### Problem
Can no longer import an extension from a file when one could before in the old extension modal.

#### Solution
Create an import button similar to what is on the home screen for importing extensions from the filesystem.

#### Validation
* Created a test extension locally and downloaded it to a file
* Created a new project and imported the extension using the import button without issue
* Made the browser windows smaller than the largest tablet screen and confirmed that the button text is removed

https://microbit.staging.pxt.io/app/7cd9e113ad0beb9c93eb4ba879b979b0b105c54d-dd7b117945

#### Notes
Fixes microsoft/pxt-microbit/issues/4576


webapp/src/extensionsBrowser.tsx
```diff
@@ -411,108 +418,132 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
...
-                                className={currentTab == TabState.InDevelopment ? "selected" : ""}
+                                ariaLabel={"Open files from your computer"}
+                                title={lf("Import File")}
+                                key="import-file"
+                                label={lf("Import File")}
+                                leftIcon="fas fa-upload"
+                                className="common-button gray"
+                                onClick={importExtension}
+                                onKeydown={importExtension}
+                            // className={"extension-tag " + ("selected")}
```
riknoll, line 463: delete?

webapp/src/extensionsBrowser.tsx
```diff
@@ -411,108 +418,132 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
...
+                            <div className="breadcrumbs">
+                                <span className="link" onClick={handleHomeButtonClick}>{lf("Home")}</span>
+                                <span>/</span>
+                                <span>{selectedTag}</span>
+                            </div>
+                        }
+                        {displayMode == ExtensionView.Tabbed &&
+                            <div className="tab-header">
+                                <Button
+                                    key={"Recommended"}
```
riknoll, line 438: delete key

webapp/src/extensionsBrowser.tsx
```diff
@@ -411,108 +418,132 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
...
+                                    title={lf("Recommended")}
+                                    label={lf("Recommended")}
+                                    onClick={() => { setCurrentTab(TabState.Recommended) }}
+                                    className={currentTab == TabState.Recommended ? "selected" : ""}
+                                />
+                                <Button
+                                    key={"In Development"}
```
riknoll, line 445: delete key

webapp/src/extensionsBrowser.tsx
```diff
@@ -17,6 +17,7 @@ const emptyCard: EmptyCard = { name: "", loading: true }
 
 interface ExtensionsProps {
     hideExtensions: () => void;
+    parent: pxt.editor.IProjectView;
```
riknoll, line 20: Don't pass in the IProjectView. It makes the coupling too tight so it's hard to refactor this in the future. Instead, pass in a callback that calls the specific method you need.

theme/common.less
```diff
@@ -1033,6 +1041,22 @@ Field editors
...
+            align-items: flex-end;
+            justify-content: space-between;
+        }
+
+        .import-button>.common-button {
+            padding: .8rem 1rem .95rem
+        }
+
+        .import-button>.common-button:hover {
+            filter: grayscale(.85) brightness(.75) contrast(1.3)
```
riknoll, line 1058: common-buttons already have hover behavior. No need to overwrite it here. Same with the padding above; just use the default react-common styles

riknoll (changes requested):

theme/common.less
aznhassan: When I use the default style the button looks like:
image
vs with padding:
image
which looks cleaner and close to the import button on the home page:
image

aznhassan: The filter was to make it even darker than the react-common button on hover to better match the behavior it has on the target home pages.

webapp/src/extensionsBrowser.tsx
```diff
@@ -411,108 +418,132 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
...
+                        <div className="import-button">
                             <Button
-                                key={"In Development"}
-                                title={lf("In Development")}
-                                label={lf("In Development")}
-                                onClick={() => { setCurrentTab(TabState.InDevelopment) }}
-                                className={currentTab == TabState.InDevelopment ? "selected" : ""}
+                                ariaLabel={"Open files from your computer"}
+                                title={lf("Import File")}
+                                key="import-file"
```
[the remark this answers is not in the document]
aznhassan: Key has been removed!

webapp/src/extensionsBrowser.tsx
```diff
@@ -411,108 +418,132 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
...
+                            </div>
+                        }
+                        <div className="import-button">
                             <Button
-                                key={"In Development"}
-                                title={lf("In Development")}
-                                label={lf("In Development")}
-                                onClick={() => { setCurrentTab(TabState.InDevelopment) }}
-                                className={currentTab == TabState.InDevelopment ? "selected" : ""}
+                                ariaLabel={"Open files from your computer"}
```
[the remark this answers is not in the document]
aznhassan: Done!

webapp/src/extensionsBrowser.tsx
```diff
@@ -411,108 +418,132 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
...
-                                />)}
-                        </div>
-                        {searchComplete && extensionsToShow.length == 0 &&
-                            <div aria-label="Extension search results">
-                                <p>{lf("We couldn't find any extensions matching '{0}'", searchFor)}</p>
+                <div className="extension-display">
+                    <div className="extension-header">
+                        {displayMode == ExtensionView.Search &&
+                            <div className="breadcrumbs">
+                                <span className="link" onClick={handleHomeButtonClick}>{lf("Home")}</span>
```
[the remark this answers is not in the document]
aznhassan: I've changed the breadcrumbs to use the react-common button with the "link-button" class. I added some CSS so the text size in the button would match the category breadcrumb font-size. Had to add padding as well to make sure the space between the "/" divider was the same on both sides.

webapp/src/extensionsBrowser.tsx
aznhassan: This is now a callback as suggested.

webapp/src/extensionsBrowser.tsx
```diff
@@ -412,38 +419,79 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
...
-                {displayMode == ExtensionView.Tags &&
-                    <div className="extension-display">
-                        <div className="breadcrumbs">
-                            <span className="link" onClick={handleHomeButtonClick}>{lf("Home")}</span>
-                            <span>/</span>
-                            <span>{selectedTag}</span>
+                        <div className="import-button">
+                            <Button
+                                ariaLabel={(lf("Open file from your computer"))}
+                                title={(lf("Import Fle"))}
```
riknoll, line 464:
```suggestion
                                title={(lf("Import File"))}
```

riknoll (approved):

jwunderl (approved):
(pending typo fix richard mentioned)

PR #8857, opened 2022-05-23
change install to add/remove

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/4606

Not sure if this text appears anywhere else... This is just the result of me doing a search through pxt and changing any user facing strings.


aznhassan (approved):
LGTM

PR #8861, opened 2022-05-24
Adding an autocomplete field for datalogger

riknoll:
autocomplete

Adding a blockly autocomplete field that draws values from other blocks of the same type. This field mostly extends from the field_textdropdown but I copied the rendering from field_string so that it looks like a normal string block.

In the field options on the API, you can specify an optional key that determines which fields are included in the list of potential values.


jwunderl (approved):

abchatra:
Love it. @jaustin fyi

PR #8862, opened 2022-05-24
center extension cards instead of left align

jwunderl:
The spacing on the cards / tab buttons in the extension page feels pretty off to me -- it's left aligned with some padding that ends up being ~~ 1/4 left 3/4 right or something like that, with the cards left aligned within that

current beta: 
image

having the cards be left aligned feels off, so just a quick centering of it

this change:

image

Changes spacing on the header bar with 'recommended' / 'in development' / 'import file' and the cards themselves so they're centered matching the rest of the elements in the view.

I put it up at https://makecode.microbit.org/app/ab3000df6e13e737ebefe89f0f8f1721610147d0-97a6a8050b#editor for a quick visual comparison in case anyone wants

### DIFF
--- a/theme/common.less
+++ b/theme/common.less
@@ -1024,6 +1024,7 @@ Field editors
     .extension-display {
         display: flex;
         flex-wrap: wrap;
+        justify-content: center;
         padding: 2rem;
         gap: .8rem;
 
@@ -1172,6 +1173,7 @@ Field editors
 .extension-cards {
     display: grid;
     grid-template-columns: repeat(auto-fit, 18rem);
+    justify-content: center;
     gap: 1rem;
     width: 100%;
 }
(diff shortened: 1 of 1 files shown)


dmonroym (approved):

aznhassan (approved):
LGTM. 

Nit: Since you're already here you should consider adopting the poor old media query that made the cards smaller on mobile screens: https://github.com/microsoft/pxt/blob/master/theme/common.less#1187

jwunderl:
Ah, do we want that back? I for one was glad that behavior was gone hah, the sad little cards don't give any context on the extension / I would rather have the scroll with one card at a time that gives info vs. one that doesn't even show half the extensions name

aznhassan:
> Ah, do we want that back? I for one was glad that behavior was gone hah, the sad little cards don't give any context on the extension / I would rather have the scroll with one card at a time that gives info vs. one that doesn't even show half the extensions name

Ah, fair enough. The mobile experience is currently not very optimized so I don't mind leaving the cards as they are.

PR #8864, opened 2022-05-24
Only Show In Development Extension Tab If They Exist

aznhassan:
* Use map to return local extensions
* Only show _Recommended_ when there are no extensions in development
* Don't show extension card images if the image URL is none
* Only grab the extensions in development when the component is mounted

### Validation
- Create a micro:bit extension and push it to GitHub: https://github.com/aznhassan/test-extension-three
- Open up a new project
- Go to the extension page and see tabs for recommended and in development
- In a new window delete the test extension
- Back on the old windows, hit go-back, and then hit the extensions button again
- See that there is now just a Recommended header and no tabs

#### No tabs
image

#### Tabs
image
image

Fixes microsoft/pxt-microbit/issues/4575

### DIFF
--- a/webapp/src/extensionsBrowser.tsx
+++ b/webapp/src/extensionsBrowser.tsx
@@ -525,7 +525,7 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
                                     label={pxt.isPkgBeta(scr) ? lf("Beta") : undefined}
                                 />
                             )}
-                            {currentTab == TabState.InDevelopment && local.forEach((p, index) =>
+                            {currentTab == TabState.InDevelopment && local.map((p, index) =>
                                 <ExtensionCard
                                     key={`local:${index}`}
                                     title={p.name}
--- a/webapp/src/extensionsBrowser.tsx
+++ b/webapp/src/extensionsBrowser.tsx
@@ -33,6 +33,7 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
     const [searchFor, setSearchFor] = useState("");
     const [searchComplete, setSearchComplete] = useState(true)
     const [allExtensions, setAllExtensions] = useState(fetchBundled());
+    const [extensionsInDevelopment, _] = useState(fetchLocalRepositories());
     const [extensionsToShow, setExtensionsToShow] = useState<(ExtensionMeta & EmptyCard)[]>([]);
     const [selectedTag, setSelectedTag] = useState("");
     const [currentTab, setCurrentTab] = useState(TabState.Recommended);
@@ -377,7 +378,6 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
     }
 
     const categoryNames = getCategoryNames();
-    const local = currentTab == TabState.InDevelopment ? fetchLocalRepositories() : undefined
 
     const onSearchBarChange = (newValue: string) => {
         setSearchFor(newValue || "");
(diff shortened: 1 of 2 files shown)


jwunderl (approved):

webapp/src/extensionsBrowser.tsx
```diff
@@ -33,6 +33,7 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
     const [searchFor, setSearchFor] = useState("");
     const [searchComplete, setSearchComplete] = useState(true)
     const [allExtensions, setAllExtensions] = useState(fetchBundled());
+    const [extensionsInDevelopment, _] = useState(fetchLocalRepositories());
```
pelikhan, line 36: nit: you can use `useMemo(() => fetchLocalRepositories(), [])`

webapp/src/extensionsBrowser.tsx
```diff
@@ -448,18 +448,26 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
...
-                                    onClick={() => { setCurrentTab(TabState.Recommended) }}
-                                    className={currentTab == TabState.Recommended ? "selected" : ""}
-                                />
-                                <Button
-                                    title={lf("In Development")}
-                                    label={lf("In Development")}
-                                    onClick={() => { setCurrentTab(TabState.InDevelopment) }}
-                                    className={currentTab == TabState.InDevelopment ? "selected" : ""}
-                                />
+                                {extensionsInDevelopment.length > 0 &&
```
pelikhan, line 451: nit: `!!extensionsInDevelopment?.length &&` is my fav expression because it defends against null refs

webapp/src/extensionsBrowser.tsx
```diff
@@ -448,18 +448,26 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
...
+                                    <>
+                                        <Button
+                                            key={"Recommended"}
```
pelikhan, line 454: only needed when using maps and creating tree elements in a loop

webapp/src/extensionsBrowser.tsx
```diff
@@ -448,18 +448,26 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
...
+                                            title={lf("Recommended")}
```
pelikhan, line 455: nit: title label can be more descriptive

pelikhan (approved):

webapp/src/extensionsBrowser.tsx
aznhassan: Is this syntactic sugar for what I've done here? Or does it work in a fundamentally different way?

webapp/src/extensionsBrowser.tsx
aznhassan: Ah, these were removed in and older PR and shouldn't be here. Diff isn't as good when whole blocks of code are being moved 😅.

webapp/src/extensionsBrowser.tsx
aznhassan: This uses `useMemo` now.

webapp/src/extensionsBrowser.tsx
aznhassan: I use `!!` as recommended since an non undefined/null/zero length is what we want. I use just `!` in the other case since we don't want to show tabs if length is undefined/null/zero.

webapp/src/extensionsBrowser.tsx
aznhassan: Good point, I've made them a little more descriptive!

webapp/src/extensionsBrowser.tsx
```diff
@@ -448,18 +448,24 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
...
+                                    <>
+                                        <Button
+                                            title={lf("Recommended Extensions")}
+                                            label={lf("Recommended")}
+                                            onClick={() => { setCurrentTab(TabState.Recommended) }}
+                                            className={currentTab == TabState.Recommended ? "selected" : ""}
+                                        />
+                                        <Button
+                                            title={lf("Local GitHub Projects")}
+                                            label={lf("In Development")}
```
pelikhan, line 461: I also second the comment of Mark. "In Development" is somewhat ambiguous. Why did we change it from Local?

(this is mostly hidden from all users, just curious).

aznhassan: Shoot, merged the PR before I saw this comment. I'll make a change in my next PR if we come to a conclusion here.

aznhassan: It's what we had when the new extension page was being designed. I don't mind changing it from "In Development" to "Local" if people feel that would be clearer/easier to understand.

PR #8866, opened 2022-05-25
Handling initialization of packet io

pelikhan:
The WebUSB stack is composed on multiple layer, 

- packetio: webusb transport
- wrapper: a layer on top that checks the version of codal etc...

Currently we consider the stack connected if webusb is connected; however there is a race where we haven't yet configured the wrapper layer. This change makes sure we wait till the wrapper is connected as well.

### DIFF
--- a/pxtlib/hf2.ts
+++ b/pxtlib/hf2.ts
@@ -574,11 +574,12 @@ namespace pxt.HF2 {
                         }
                     log(`Board-ID: ${this.info.BoardID} v${this.info.Parsed.Version} f${this.info.Parsed.Features}`)
                 })
-                .then(() => this.talkAsync(HF2_CMD_JDS_CONFIG, new Uint8Array([1])).then(() => {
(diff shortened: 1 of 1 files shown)


pxtlib/hf2.ts
```diff
@@ -560,13 +574,16 @@ namespace pxt.HF2 {
...
+                .then(() => this.talkAsync(HF2_CMD_JDS_CONFIG, new Uint8Array([1]))
+                    .then(() => {
+                        this.jacdacAvailable = true
+                    }, _err => {
+                        this.jacdacAvailable = false
+                    }))
                 .then(() => {
                     this.reconnectTries = 0
+                    this.initialized = true
+                    this.io.onConnectionChanged()
```
pelikhan, line 586: need to notify ui that we transitions connecting -> connected

mmoskal (approved):

PR #8869, opened 2022-05-25
Unselect tags when a search is begun

aznhassan:
* Unselect tags when a search is begun so that it's clear we are not searching within the tags

https://user-images.githubusercontent.com/6496798/170350750-0ac771af-3896-4e55-87a9-3151acb5e720.mp4

Fixes microsoft/pxt-microbit/issues/4624
Fixes microsoft/pxt-microbit/issues/4595

build: https://microbit.staging.pxt.io/app/72d3350df53901d342c82efe3bd268d8085682f3-bde050c4f3#editor

### DIFF
--- a/webapp/src/extensionsBrowser.tsx
+++ b/webapp/src/extensionsBrowser.tsx
@@ -426,16 +426,7 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
                 </div>
                 <div className="extension-display">
                     <div className="extension-header">
-                        {displayMode == ExtensionView.Search &&
-                            <Button
-                                title={lf("Home")}
-                                label={lf("Home")}
-                                onClick={handleHomeButtonClick}
-                                className="link-button"
-                            />
-                        }
-                        {displayMode == ExtensionView.Tags &&
-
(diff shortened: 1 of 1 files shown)


webapp/src/extensionsBrowser.tsx
```diff
@@ -425,25 +426,20 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
...
                                 <Button
                                     title={lf("Home")}
                                     label={lf("Home")}
                                     onClick={handleHomeButtonClick}
                                     className="link-button"
                                 />
-                                <span>/</span>
-                                <span>{selectedTag}</span>
+                                {displayMode == ExtensionView.Tags &&
+                                    <>
```
jwunderl, line 438: sidenote: Huh, I saw this in the last pr but this is a pretty nice feature -- I had assumed it just expanded to div but it actually becomes a [React.Fragment](https://reactjs.org/docs/fragments.html) which doesn't become an element in the DOM and just stays as structuring react side. Good to know.

jwunderl (approved):

PR #8870, opened 2022-05-25
fix extensions view in firefox

jwunderl:
chromium expands to full width but in this specific case firefox does not, so add an extra explicit setting fo rit


aznhassan (approved):
:shipit:

PR #8876, opened 2022-05-27
add Menu, Reset shortcut to keymap view

jwunderl:
re: https://github.com/microsoft/pxt-common-packages/pull/1348, add to the keymap view

I just added them to the end of player1/player2, can add a 'system' header above them if we want but that would take up more space and I'm not sure it would look any better:

image

### DIFF
--- a/theme/keymap.less
+++ b/theme/keymap.less
@@ -70,7 +70,7 @@
 
 .keymap.above,
 .fullscreensim .keymap {
-    top: 5.5rem;
+    top: 4.5rem;
 
     .close {
         color: @teal;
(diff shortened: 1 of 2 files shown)


Jaqster:
Yeah, it's a little confusing because it looks like "1 Menu" is for Player 1, and "2 Reset" is for Player 2.  But I'm not sure how else to show...

eanders-ms:
Not sure we want to map numbers to these functions. I'd like to reserve them for future multiplayer behavior. What about:
**back-tick**: open menu. This is a common shortcut in many games to bring up a cmd prompt. Similar enough?
**backspace**: restart.

jwunderl:
swapped away from 1/2 as mentioned to backtick for menu & backspace for reset; only real downside is that backspace is a long word for this display & not fully universal (ios/mac uses 'Delete') so I put in the the unicode character that's fairly commonly used?

image

if people don't know what that symbol means then maybe best to just drop this PR & leave it in notes on accessibility or developer notes somewhere? or maybe as hover text on the sym buttons themselves, but that would be very ugly when people use mouse controls.

PR #8879, opened 2022-06-01
Improved Python class support

riknoll:
This PR extends our Python class support. I still have a few more bugs to fix, but this adds a lot more functionality and fixes the more egregious bugs. Also this does *not* roundtrip.

Newly supported features include:
* Static methods and properties
* Floating statements in the class scope (i.e. not within a method or a constructor)
* Extending other Python classes
* Extending TypeScript classes

Intentionally unsupported features that we could conceivably support in the future:
* Calling protected properties/methods from a TypeScript class in a python class (right now the pyconverter uses the apiinfo from the language service which does not have access to those symbols; it's nontrivial to fix that)
* `__iter__`, `__str__`, and all the other "special" python methods

Also, for reference, Static Python has some additional restrictions on classes because it maps to classes in static TypeScript. These include:
* No multiple inheritance
* Constructors in subclasses must call `super().__init__` as the first statement

### DIFF
--- a/pxtpy/converter.ts
+++ b/pxtpy/converter.ts
@@ -1336,8 +1336,8 @@ namespace pxt.py {
                 }
             }
 
-            const classDefs = n.body.filter(s => s.kind === "FunctionDef");
-            const staticStmts = n.body.filter(s => classDefs.indexOf(s) === -1);
+            const classDefs = n.body.filter(s => n.isNamespace || s.kind === "FunctionDef");
+            const staticStmts = n.isNamespace ? [] : n.body.filter(s => classDefs.indexOf(s) === -1);
 
             let body = stmts(classDefs)
             nodes.push(body)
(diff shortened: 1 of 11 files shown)


pxtpy/converter.ts
```diff
@@ -733,13 +741,21 @@ namespace pxt.py {
...
-        return U.values(internalApis).filter(e => e.namespace == qn && e.kind == SK.Property)
+        return U.values(internalApis).filter(e => e.namespace == qn && ((e.kind == SK.Variable && !excludeVariables) || e.kind == SK.Property))
     }
 
     function getClassField(ct: SymbolInfo, n: string, checkOnly = false, skipBases = false): SymbolInfo | null {
-        let qid = ct.pyQName + "." + n
+        let qid: string;
+
+        if (n === "__init__") {
+            qid= ct.pyQName + ".__constructor"
```
jwunderl, line 753:
```suggestion
            qid = ct.pyQName + ".__constructor"
```

nit spacing

pxtpy/converter.ts
```diff
@@ -1307,23 +1338,71 @@ namespace pxt.py {
...
+            }
+
+            if (!n.isNamespace) {
+                let isStatic = (f: SymbolInfo) => f.kind === SK.Property && !f.isInstance || f.kind === SK.Variable;
+
+                const fieldDefs = listClassFields(n, false)
+                    .filter(f => f.kind == SK.Property || isStatic(f))
+                    .map(f => {
+                        if (!f.pyName || !f.pyRetType)
+                            error(n, 9535, lf("field definition missing py name or ret type", f.qName));
```
jwunderl, line 1389:
```suggestion
                            error(n, 9535, lf("field definition missing py name or return type", f.qName));
```

sidenote: hm I suppose that user would probably be able to infer what py name refers to by looking at the line? We should maybe go through all the errors on this page and put up a python-errors docs page that just gives a little more explanation and an example of fixing. not directly relevant to this pr tho

jwunderl (approved):

jwunderl:
Probably worth an uploadtrg for testing / future reference on the state this puts python classes in~

riknoll:
Here you go:

https://arcade.makecode.com/app/901ffb36d2b750073b7a6f84b34297626914acae-9aaf75cc0d

BTW this could definitely use some more testing... I figured it was safe to merge though because classes were basically unusable before

jwunderl:
Ah yes, I just meant basic testing -- it's nice to have uploadtrgs for prs like this so that if we have to check if something's a regression or anything it's easy to go back without having to figure out tags / etc

PR #8880, opened 2022-06-02
Support for github "main" branch

eanders-ms:
| **This change needs thorough testing** |
|-|

#### Changes
Adds support for non-"master" default GitHub branches
* Query default branch name when importing a repo
* When creating a new repo, create with "main" branch
* Compatible with old repos using "master" as default
    * (the code can handle any default branch name)

#### Out of scope
The pxt cli has some references to "master" branch. I have not investigated what these are or whether they need to be updated.

### DIFF
--- a/pxtlib/github.ts
+++ b/pxtlib/github.ts
@@ -191,8 +191,8 @@ namespace pxt.github {
 
         async loadConfigAsync(repopath: string, tag: string): Promise<pxt.PackageConfig> {
             if (!tag) {
-                pxt.debug(`dep: default to master branch`)
-                tag = "master";
+                pxt.debug(`dep: default to main branch`)
+                tag = "main";
             }
 
             // cache lookup
@@ -229,8 +229,8 @@ namespace pxt.github {
 
         async loadPackageAsync(repopath: string, tag: string): Promise<CachedPackage> {
             if (!tag) {
-                pxt.debug(`load pkg: default to master branch`)
-                tag = "master";
+                pxt.debug(`load pkg: default to main branch`)
+                tag = "main";
             }
 
             // try using github proxy first
@@ -401,7 +401,7 @@ namespace pxt.github {
         return (resp.statusCode == 200)
     }
 
-    export async function putFileAsync(repopath: string, path: string, content: string) {
+    export async function putFileAsync(repopath: string, tag: string, path: string, content: string) {
         const parsed = parseRepoId(repopath);
         await ghRequestAsync({
             url: `https://api.github.com/repos/${pxt.github.join(parsed.slug, "contents", parsed.fileName, path)}`,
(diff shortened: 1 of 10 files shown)


pxtlib/github.ts
```diff
@@ -410,7 +414,7 @@ namespace pxt.github {
             data: {
                 message: lf("Initialize empty repo"),
                 content: btoa(U.toUTF8(content)),
-                branch: "master"
+                branch: branch ?? "main"
```
pelikhan, line 417: dubious

pxtlib/github.ts
```diff
@@ -468,7 +472,7 @@ namespace pxt.github {
     }
 
     export function getRefAsync(repopath: string, branch: string) {
-        branch = branch || "master";
+        branch = branch ?? "main";
```
pelikhan, line 475: dubious

pxtlib/github.ts
```diff
@@ -722,6 +726,16 @@ namespace pxt.github {
         fork?: boolean;
     }
 
+    export function isDefaultBranch(branch: string, repo?: GitRepo) {
+        if (repo && repo.defaultBranch)
+            return branch === repo.defaultBranch;
+        return /^(main|master)$/.test(branch);
```
pelikhan, line 732: it's a hack but it'll probably work

webapp/src/workspace.ts
```diff
@@ -1488,8 +1492,8 @@ export async function initializeGithubRepoAsync(hd: Header, repoid: string, forc
 }
 
 export async function importGithubAsync(id: string): Promise<Header> {
-    // if tag is not specified, asssume master
-    const repoid = pxt.github.normalizeRepoId(id, "master").replace(/^github:/, "")
+    // if tag is not specified, asssume main
+    const repoid = pxt.github.normalizeRepoId(id, "main").replace(/^github:/, "")
```
pelikhan, line 1496: dubious, you're breaking legacy projects using master

webapp/src/db.ts
```diff
@@ -100,8 +100,13 @@ class GithubDb implements pxt.github.IGithubDb {
     }
 
     loadConfigAsync(repopath: string, tag: string): Promise<pxt.PackageConfig> {
+        if (!tag) {
+            pxt.debug(`dep: default to main`)
+            tag = "main"
```
pelikhan, line 105: dubious, we have a ton of existing repos using master

webapp/src/db.ts
```diff
@@ -124,11 +129,12 @@ class GithubDb implements pxt.github.IGithubDb {
     }
     loadPackageAsync(repopath: string, tag: string): Promise<pxt.github.CachedPackage> {
         if (!tag) {
-          pxt.debug(`dep: default to master`)
-          tag = "master"
-        }
-        // don't cache master
-        if (tag == "master")
+            pxt.debug(`dep: default to main`)
```
pelikhan, line 132: why are you changing the tag here?

PR #8881, opened 2022-06-02
add time options for localization

kimprice:
closes https://github.com/microsoft/pxt-microbit/issues/4126

Thanks for your help @jwunderl!

Should I add anything in the defining-blocks.md about localizing field options?

### DIFF
--- a/pxtblocks/fields/field_numberdropdown.ts
+++ b/pxtblocks/fields/field_numberdropdown.ts
@@ -1,5 +1,15 @@
 /// <reference path="../../localtypings/pxtblockly.d.ts" />
 
+// common time options -- do not remove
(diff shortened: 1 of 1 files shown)


jwunderl:
so that file is user / extension author facing but they wouldn't really be able to do the same thing as we did here -- if we wanted to add a note on how to do it to that file we'd probably want to make it something the user can do solely from static typescript but that's more work that I'm not sure is worth doing right now.

We can add a note if you think it's right but I don't think we'd take prs adding arbitrary lfs just to support those being added in an extension or anything, and I haven't seen any extensions that would actually make use of that in the first place.

jwunderl (approved):

PR #8882, opened 2022-06-02
Handle python class instance fields with the same name as static fields

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/3456

Python's class semantics for static fields are a bit different than TypeScript's... Basically, if you try to access a field on a class instance and it is defined as a static field, the static field is returned. If that field is then assigned on the instance, then the assigned value is returned. Luckily, it's pretty easy to replicate that behavior using getters and setters in TypeScript!

This PR updates how we emit static fields so that classes like this:

```python
class Greeter:
    greeting = ""
    def __init__(self, message):
        self.greeting = message
    def greet(self):
        return "Hello, " + self.greeting
```

now convert into TypeScript like this:

```typescript
class Greeter {
    static greeting: string
    private ___greeting_is_set: boolean
    private ___greeting: string
    get greeting(): string {
        return this.___greeting_is_set ? this.___greeting : Greeter.greeting
    }
    set greeting(value: string) {
        this.___greeting_is_set = true
        this.___greeting = value
    }
    
    public static __initGreeter() {
        Greeter.greeting = ""
    }
    
    constructor(message: any) {
        this.greeting = message
    }
    
    public greet() {
        return "Hello, " + this.greeting
    }
    
}

Greeter.__initGreeter()
```

...which is not pretty, but hey this doesn't need to go to blocks anyway :)

### DIFF
--- a/pxtpy/converter.ts
+++ b/pxtpy/converter.ts
@@ -1402,18 +1402,14 @@ namespace pxt.py {
                     });
                 const staticFieldSymbols = fieldDefs.filter(f => isStatic(f));
 
-                const instanceFields = fieldDefs.filter(f => !isStatic(f))
-                    .map((f) => {
-                        const quotedName = quoteStr(f.pyName!);
-
-                        if (staticFieldSymbols.some(s => quoteStr(s.pyName!) === quotedName)) {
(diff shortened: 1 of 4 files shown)


riknoll:
okay, now it's ready

PR #8884, opened 2022-06-03
fix css issues in tutorials

thelucasfiles:
fixes https://github.com/microsoft/pxt-arcade/issues/4643
fixes https://github.com/microsoft/pxt-arcade/issues/4327

fixes padding issues in hints as well as pointer events being captured by blur at bottom of tutorial content

### DIFF
--- a/theme/tutorial.less
+++ b/theme/tutorial.less
@@ -394,6 +394,7 @@ code.lang-filterblocks {
 .tutorial-callout > div {
     max-height: 60vh;
     overflow: auto;
+    padding-right: 3vh;
 }
 
 .tutorialhint.hidden,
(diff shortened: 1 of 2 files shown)


jwunderl (approved):

jwunderl:
BabyYodaGIF

theme/tutorial.less
```diff
@@ -394,6 +394,7 @@ code.lang-filterblocks {
 .tutorial-callout > div {
     max-height: 60vh;
     overflow: auto;
+    padding-right: 3vh;
```
jwunderl, line 397:
```suggestion
    padding-right: 2rem;
```

Thinking about it again probably want to use em instead of vh -- vh changes with the browser window (well, viewport) height so this would grow / shrink based off the height of the browser window, when we really just want it to be 'a bit larger so the 1.5rem button never touches this'

PR #8885, opened 2022-06-03
Add cache skip opt for tutorial info, fix tutorial functions in monaco

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/2160

This ended up requiring a bit more context in the codebase / block definitions that expected so it ended up being that I fixed it while figuring out what wasn't working when the first portion lucas & I were looking at only half fixed it, and then lucas and I chatted about it.

It originally looked like the issue was here: https://github.com/microsoft/pxt/blob/master/webapp/src/toolboxeditor.tsx#L186, since the blocks toolbox has a button but the monaco functions toolbox category only has two fake blocks. However, the issue ended up being over in the monacoSnippets block definition, since those two were using the old blockly style function id for the two function snippets that show up in monaco which doesn't match the blockIds used in the tutorialInfo.

I also ended up having to add a quick flag for skipping the cache for tutorial info, since that was thwarting testing

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1726,7 +1726,7 @@ export class ProjectView
 
         const t = header.tutorial;
 
-        const skipTutorialInfoCache = /notutorialinfocache=1/i.test(window?.location.href);
+        const skipTutorialInfoCache = /notutorialinfocache=1/i.test(window.location.href);
 
         if (typeof t.tutorialCode === "string") {
             t.tutorialCode = [t.tutorialCode];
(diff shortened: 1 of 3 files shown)


webapp/src/app.tsx
```diff
@@ -1726,12 +1726,14 @@ export class ProjectView
 
         const t = header.tutorial;
 
+        const skipTutorialInfoCache = /notutorialinfocache=1/i.test(window?.location.href);
```
thelucasfiles, line 1729:
```suggestion
        const skipTutorialInfoCache = /notutorialinfocache=1/i.test(window.location.href);
```
fix the typo

thelucasfiles (approved):
great job

riknoll (approved):

PR #8893, opened 2022-06-08
Bump pxt-blockly to 4.0.12

aznhassan:
Bumping pxt-blockly to 4.0.12

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,7 +131,7 @@
     "monaco-editor": "0.24.0",
     "pouchdb": "7.2.1",
     "pouchdb-adapter-memory": "7.2.1",
-    "pxt-blockly": "4.0.10",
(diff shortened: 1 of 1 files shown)


jwunderl (approved):
Build failed as the publish step in pxt-blockly is still running, should be fine if you rerun it in ~5 minutes or so

Since we don't make changes in pxt-blockly as often I like to list the changes in the bump prs:

* prevent the insertion marker from creating variables https://github.com/microsoft/pxt-blockly/pull/396
* Prevent hidden inputs from enabling connection tracking https://github.com/microsoft/pxt-blockly/pull/398
* don't show delete while debugging https://github.com/microsoft/pxt-blockly/pull/397
* only use output shape for output connections https://github.com/microsoft/pxt-blockly/pull/399

PR #8900, opened 2022-06-10
don't disconnect when deploying

pelikhan:
Fix for https://github.com/microsoft/pxt-microbit/issues/4694

While a deployment is on-going, do not automatically disconnect when page visibilty changes.

- [x] don't disconnect while deploying on page visibilty change
- [x] check page visibility upon deploy completed

Testing:

Keep 2 makecode tab open, with two different program (using different C++ to trigger full flash). Start download, switch tab, wait until download is completed, tab releases USB


abchatra:
Thanks @pelikhan ❤️

PR #8901, opened 2022-06-10
Datalogger csv simulator

jwunderl:
Adds in datalogger 

fix https://github.com/microsoft/pxt-microbit/issues/4167
close https://github.com/microsoft/pxt-microbit/issues/4492

uploadtrg with test link: https://makecode.microbit.org/app/57c0fa23ac50cdca72b24b113df6d4cf4111e670-be79ef3912#pub:_8Fy8aFbriLjJ

couple small things:

* a few css issues I'm working on but just put up the pr, e.g. probably gonna flexbox the serial header bar for centering, make it prettier
* the micro:bit side isn't sending time info as that cut muddled in with the serial mirroring, but will work on that over on micro:bit side; putting up for now as that's a separate issue.
* (I was going to do a bit bigger change with refactoring the serial view but ended up not having time with other things coming up, so this is more in line with that existing code; next time >:( )
* toggle change was pretty last minute, it should only show up if you're using both datalogger & serial in the same project. let me know if you see anything surprising there.
* I made it so the csv table view only autoscroll if you can already see the bottom -- that is, if you scroll up it won't auto scroll anymore so that you can read old data. Maybe worth playing around with a little give there

### DIFF
--- a/pxteditor/editor.ts
+++ b/pxteditor/editor.ts
@@ -96,6 +96,7 @@ namespace pxt.editor {
 
         simSerialActive?: boolean;
         deviceSerialActive?: boolean;
+        csvSerialActive?: boolean;
 
         errorListState?: ErrorListState;
 
@@ -209,6 +210,7 @@ namespace pxt.editor {
         openSimView(): void;
         openSimSerial(): void;
         openDeviceSerial(): void;
+        openCsvSerial(): void;
         openPreviousEditor(): void;
 
         switchTypeScript(): void;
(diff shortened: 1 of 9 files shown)


abchatra:
@jaustin fyi

jaustin:
I've taken a look at the preview link, it's looking great

@jwunderl do you plan to change the language from "show console" to "show log" or something

My main outstanding concern having seen this is that I think we really need a way to hint to users how and where to find the *real* log file from their device. Because this still looks somewhat like serial view, I'm worried it leads people to expect the log file to be visible in MakeCode.  

I think a single line of text on the simulated log or a new button underneath could help this.

image

@Jaqster I recall you had some thoughts about how/where to help users find the log file too?

abchatra (approved):

PR #8909, opened 2022-06-14
resolve and upgrade github versions as they are loaded

pelikhan:
As we load packages and resolve github references, always try to upgrade the referenced version to avoid version clashes. This codepath only happens on github project references

For example, we can have a project with a dependency A, B, A/C where

    A is a repo, version v2
    A/C is a subproject in repo A
    B references A/C#v1

With this program, A gets loaded first and locks to v2; then we need to make sure to all future loaded references of A or A/C are upgraded on v2.

To test this:

- create new project
- add jacdac, which should be something like v1.8.xx
- add https://github.com/pelikhan/pxt-kitronik-stopbit/tree/master/jacdac
- verify that jacdac-traffic-light is now v1.8.xx instead of v0.10.40

### DIFF
--- a/pxtlib/package.ts
+++ b/pxtlib/package.ts
@@ -701,6 +701,30 @@ namespace pxt {
                     pxt.debug(`dep: load ${from.id}.${id}${isCpp ? "++" : ""}: ${ver}`)
                     if (id == "hw" && pxt.hwVariant)
                         id = "hw---" + pxt.hwVariant
+
+                    // for github references, make sure the version is compatible with previously
+                    // loaded references, regardless of the id
+                    const ghver = github.parseRepoId(ver)
+                    if (ghver?.slug) {
+                        // let's start by resolving the maximum version
+                        // number of the parent repo already loaded
+                        const repoVersions = Object.values(from.parent.deps)
+                            .map(p =>  github.parseRepoId(p._verspec))
+                            .filter(v => v?.slug === ghver.slug)
+                            .map(v => v.tag)
+                        const repoVersion = repoVersions
+                            .reduce((v1, v2) => semver.strcmp(v1,v2) > 0 ? v1 : v2, "0.0.0")
+                        pxt.debug(`dep: common repo ${ghver.slug} version found ${repoVersion}`)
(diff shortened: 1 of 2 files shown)


pxtlib/package.ts
```diff
@@ -701,6 +701,30 @@ namespace pxt {
...
+                    const ghver = github.parseRepoId(ver)
+                    if (ghver?.slug) {
+                        // let's start by resolving the maximum version
+                        // number of the parent repo already loaded
+                        const repoVersions = Object.values(from.parent.deps)
+                            .map(p =>  github.parseRepoId(p._verspec))
+                            .filter(v => v?.slug === ghver.slug)
+                            .map(v => v.tag)
+                        const repoVersion = repoVersions
+                            .reduce((v1, v2) => semver.strcmp(v1,v2) > 0 ? v1 : v2, "0.0.0")
```
jwunderl, line 716:
```suggestion
                            .reduce((v1, v2) => semver.strcmp(v1, v2) > 0 ? v1 : v2, "0.0.0")
```

nit spacing

pxtlib/semver.ts
```diff
@@ -101,6 +101,10 @@ namespace pxt.semver {
         return aa.major - bb.major;
     }
 
+    /**
+     * Compares two semver version strings and returns -1 if a < b, 1 if a > b and 0
+     * if versions are equivalent
+     */
```
jwunderl, line 107: maybe worth a note on what happens if a / b are nullish since that seemed to cause an issue previously? but already better than not having a comment ~

jwunderl (approved):
lgtm

pelikhan:
@jwunderl is there a linter you use for pxt? I have prettier enable and it shuffles everything around.

jwunderl:
We don't actually have a standard one we've used (besides a few small rules in ci), just the quick whitespacing style we've always checked in prs; we probably should just add in prettier or something to gulp to handle almost all of these anyways~

pxtlib/semver.ts
pelikhan: updated the comment, take a look

PR #8914, opened 2022-06-15
make learn more for builtin point to /reference/{pkg}

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/4752#issuecomment-1157016629
fix https://github.com/microsoft/pxt-microbit/issues/4587

funny enough this fails locally since our cli seems to miss the `/libs/pkg/docs/` path, but the live site verison of that page should be fine

image


aznhassan (approved):
:shipit:

PR #8917, opened 2022-06-16
hide embed opt in share dialog on mobile devices

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/4760

@riknoll I added a new size to react-common-breakpoints for this; largestMobileScreen is actually pretty big to the point that it hits lots of tablets / not full screen monitors. 420 px is just above the biggest normal portrait screens (e.g. my samsung big phone is 412px) and I feel like the portrait size is probably what we intend most of the time we add 'phone size screen optimizations' since landscape can be pretty wide / fit most things

image


Jaqster (approved):
Thanks Joey!

abchatra (approved):

PR #8927, opened 2022-06-17
add test redirect to /v4

jwunderl:


abchatra (approved):

abchatra:
@pelikhan this is what you were suggesting too right?

PR #8939, opened 2022-06-23
Add support for accordion hints in tutorials

adrianwyatt:
This feature adds support for accordion hints in tutorial markdowns. 
[pxt-arcade issue 4555: Looking for a quick, simple solution to provide extra info without having to divert the coder's attention.](https://github.com/microsoft/pxt-arcade/issues/4555)

untitled


webapp/src/marked.tsx
```diff
@@ -385,7 +486,7 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
         // Set markdown options
         marked.setOptions({
             renderer: renderer,
-            sanitize: true,
+            sanitize: false,
```
abchatra, line 489: Early look, we can't set sanitize to false. This is a security issue with the third party injecting html. Why is this required?

adrianwyatt: Ah - accidentally left that in there. I will set it back to "true".

abchatra: Awesome thank you!

webapp/src/marked.tsx
```diff
@@ -350,6 +352,105 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
...
+    // -------------------------------
+    private renderAccordianHints(content: HTMLElement) {
+        const hintBeginRegex = /^\s*~hint\s*(.+)/i;
+        const hintEndRegex = /^\s*hint~.*/i;
+
+        // Processing each hint modifies the node order and count of 'content.childnodes'.
+        // Keep reprocessing content.childnodes until all hints have been processed.
+        let hintFound: boolean = false;
+
+        do {
```
abchatra, line 379: This is bit hard to read for me. Can we do this loop in two steps?

1. First loop finds all the hints
2. Second loop acts on the hints it found in the first loop.

Are we supporting nested hints here?

abchatra (changes requested):

tests/tutorial-test/cases/accordionHints.md
```diff
@@ -0,0 +1,10 @@
+# Accordion Hints
+## Step 1
+Welcome!
+
+~hint This content is hidden until the user clicks here.
```
abchatra, line 5: Some more tests with the multiple hints and nested hints would be great. If we are not supporting nested hints we should call out (I am ok not supporting nested hints as well.

abchatra:
Overall looks good to me.  I like ~hint & hint~ pair. However @kiki-lee is the expert here to confirm it is the right thing from an content author perspective.

@riknoll can you please review this as well?

theme/tutorial-sidebar.less
```diff
@@ -434,6 +434,18 @@
     top: calc(5.5rem + @mainMenuHeight);
 }
 
+/*********************************
+        Accordion Hints (details)
+*********************************/
+
+.tabTutorial details {
+    padding: .5rem .5rem .5rem 2rem;
```
jwunderl, line 442: 2rem left padding is pretty significant in the details view -- something like 10% of the whole width on my machine, which is notable when you're putting an image or block in a hint: 
image

```suggestion
    padding: .5rem;
```

feels a bit better to me personally:

image

webapp/src/marked.tsx
```diff
@@ -350,6 +352,105 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
...
+            hintFound = false;
+            // The hint summary element (i.e. <summary>text</summary>).
+            let hintSummary: HTMLElement = null;
+            // The element that signifies the beginning of the hint (i.e. <p>~hint (hint summary)</p>).
```
jwunderl, line 383: would be nice to have this logic included for docs view (e.g. localhost:3232/tutorials/chase-the-pizza) -- that logic lives over in https://github.com/microsoft/pxt/blob/master/pxtrunner/renderer.ts#L1302 . Currently it shows up as plaintext: 
image

(but this is also probably fine to handle separately / as a follow up, that view often ends up a little neglected to be honest and it's not the primary scenario)

docs/writing-docs/tutorials/basics.md
```diff
@@ -168,6 +168,22 @@ forever(function() {
 ```
 ````
 
+## Accordion/hidden hints
+If you want to provide extra information without having to divert the coder's attention, you can include content in an "accordion" style hint control.
```
jwunderl, line 172: It would maybe be worth dropping a link to https://makecode.com/writing-docs/tutorials/control-options#explicit-hints here? Or ideally images / snippets that are inside ~hints wouldn't trigger the 'hide in hint bubble' logic ever, but that may be a bit more complicated to implement.

kiki-lee: Great point, Joey ^

tests/tutorial-test/cases/accordionHints.md
kiki-lee: > Some more tests with the multiple hints and nested hints would be great. If we are not supporting nested hints we should call out (I am ok not supporting nested hints as well.

I'm perfectly fine not supporting nested hints right now…that said, if we can get them, I'm sure they'll get used!  They'd just have to have slightly differentiated CSS so people could tell what's going on.

adrianwyatt: I will add more test cases and keep nested accordion hints as not supported. I'll add a notice in the docs.

webapp/src/marked.tsx
adrianwyatt: We are not supporting nested hints. I have refactored the loops into two steps.

webapp/src/marked.tsx
adrianwyatt: I recommend handling this scenario in a separate PR.

docs/writing-docs/tutorials/basics.md
```diff
@@ -168,6 +168,24 @@ forever(function() {
...
+
+### ~ hint
+Nested accordion hints are not currently supported.
+### ~
+
+```
+~hint This content is hidden until the user clicks here.
+  - :blank: Bullet 1
+  - :mouse pointer: Bullet 2
+hint~
```
adrianwyatt, line 186: (not that a newline is no longer required)

webapp/src/marked.tsx
jwunderl: Yup, agreed

riknoll (approved):

abchatra:
Congratulations!  ❤️

PR #8941, opened 2022-06-23
Adding monaco field editor for sound effects

riknoll:
monaco-sound

Fixes https://github.com/microsoft/pxt-microbit/issues/4732

Note that the regex is dumb and weird things might happen if you use parentheses inside the arguments of the sound you're trying to edit.

### DIFF
--- a/pxteditor/monaco-fields/field_soundEffect.ts
+++ b/pxteditor/monaco-fields/field_soundEffect.ts
@@ -1,0 +1,183 @@
+/// <reference path="./monacoFieldEditor.ts" />
+/// <reference path="./field_react.ts" />
+
+namespace pxt.editor {
+    const fieldEditorId = "soundeffect-editor";
+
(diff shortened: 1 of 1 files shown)


jwunderl (approved):
lgtm; a bit unfortunate something like 

```
music.createSoundEffect(
WaveShape.Sine,
input.acceleration(Dimension.X),
0,
255,
0,
input.acceleration(Dimension.X),
SoundExpressionEffect.None,
InterpolationCurve.Linear)
```

will have trouble but matching that with regex feels like it only ends in catastrophic backtracking?

pelikhan (approved):

riknoll:
@abchatra yes

@jwunderl yeah, not sure if there is a great workaround. maybe i should restructure the monaco field editors to treat the regex as an initial pass and then do a finer-grain parsing of the code after the match

PR #8942, opened 2022-06-24
Add validate range for monaco field editors and some missing css

riknoll:
Re: conversation with @jwunderl in https://github.com/microsoft/pxt/pull/8941

This adds an optional function to validate/adjust monaco field editor ranges so that we don't have to cram more complicated logic into the regex (which becomes a performance concern). The validate range function here just counts parentheses and handles multiline, so a snippet like this:

```ts
music.playSoundEffect(
    music.createSoundEffect(
        WaveShape.Sine,
        1,
        input.acceleration(Dimension.X),
        0,
        0,
        500,
        SoundExpressionEffect.None,
        InterpolationCurve.Linear
    ),
    SoundExpressionPlayMode.UntilDone
);
```

...can now be edited. Note that the acceleration call in this example will be removed when you close the editor (better than the old behavior of breaking your program, though!)

I also accidentally left some css out of the last pr!

### DIFF
--- a/pxteditor/monaco-fields/field_soundEffect.ts
+++ b/pxteditor/monaco-fields/field_soundEffect.ts
@@ -151,6 +151,47 @@ namespace pxt.editor {
         }
     }
 
+    function validateRange(range: monaco.Range, model: monaco.editor.ITextModel) {
+        let currentLine = range.startLineNumber;
+        let currentColumn = 0;
+        let foundStart = false;
+        let parenCount = 0;
+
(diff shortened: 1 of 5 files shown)


jwunderl (approved):
Ooooh nice. Yeah, I like this model of 'find approximate location with regex & then finish off with a handler' way more than writing increasingly complex regexes hah

riknoll:
@jwunderl yeah, eventually i'll write a higher-level helper function to parse the arguments to make it even easier

PR #8948, opened 2022-06-28
Fix commenting for specific auto-declared variables.

momuno:
This fix addresses GitHub issue # 3169 in pxt-arcade:
https://github.com/microsoft/pxt-arcade/issues/3169

- Moves `getComments()` from `getStatementBlock()` private scope, updates signature.
- Renames `getComments()` to `getNodeComments()`, so to not be confused with `getComments()` in `emitter.ts`, which will also show up in `pxtworker.js`.
- Attaches comment to statement node for auto-declared variable.
- Update variable naming and commenting.
- Add test for comments associated with 'false' and "0" auto-declared variables

### DIFF
--- a/pxtcompiler/emitter/decompiler.ts
+++ b/pxtcompiler/emitter/decompiler.ts
@@ -1301,13 +1301,27 @@ ${output}</xml>`;
                         break;
                     case SK.VariableDeclaration:
                         const decl = node as ts.VariableDeclaration;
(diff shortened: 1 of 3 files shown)


riknoll (approved):

riknoll:
Thanks @momuno!

abchatra:
Thanks @momuno Wonderful!

momuno:
Thanks @abchatra!

PR #8950, opened 2022-06-30
Streams Support in Tutorials

thelucasfiles:


webapp/src/marked.tsx
```diff
@@ -284,6 +285,19 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
...
     }
 
+    private renderVideo(content: HTMLElement){
+        pxt.Util.toArray(content.querySelectorAll('Video.ams-embed'))
+            .forEach((inlineVideo: HTMLElement)=> {
+                let player = MediaPlayer().create()
+                if (inlineVideo) {
+                    player.initialize(inlineVideo, inlineVideo.getAttribute("src"));
+                }
+                return () => {
```
jwunderl, line 296:
```suggestion
                return () => {
```

this doesn't look like it's doing anything?

webapp/src/marked.tsx
```diff
@@ -4,6 +4,7 @@ import * as React from "react";
 import * as data from "./data";
 import * as marked from "marked";
 import * as compiler from "./compiler"
+import {MediaPlayer} from "dashjs"
```
jwunderl, line 7:
```suggestion
import { MediaPlayer } from "dashjs"
```

nit

abchatra:
Needs a better description and title.

webapp/src/marked.tsx
```diff
@@ -284,6 +285,19 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
         }
     }
 
+    private renderVideo(content: HTMLElement){
+        pxt.Util.toArray(content.querySelectorAll('Video.ams-embed'))
+            .forEach((inlineVideo: HTMLElement)=> {
+                let player = MediaPlayer().create()
+                if (inlineVideo) {
```
jwunderl, line 293: inlineVideo should always exist so I don't think there's a reason to have this if / it won't ever be false

react-common/components/controls/EmbedVideo.tsx
```diff
@@ -1,45 +1,67 @@
...
+            </div>
+        );
 
-    let st1 = "encrypted-media; picture-in-picture";
+    } else {
 
-    if (pxt.BrowserUtils.isIOS()){
-        st1 += "; muted=true";
-    }
+        videoURL = "https://makecode-lucas-testing-makecodetempmediaservice-usea.streaming.media.azure.net/a6dd2090-b963-490c-bc5d-cdeecdee2c6e/WIN_20220622_17_44_45_Pro.ism/manifest(format=mpd-time-cmaf)";
```
jwunderl, line 47: might as well fix this to use the ID as well here instead of just using a full string literal?

webapp/src/marked.tsx
```diff
@@ -284,6 +285,19 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
         }
     }
 
+    private renderVideo(content: HTMLElement){
+        pxt.Util.toArray(content.querySelectorAll('Video.ams-embed'))
+            .forEach((inlineVideo: HTMLElement)=> {
```
jwunderl, line 291:
```suggestion
            .forEach((inlineVideo: HTMLElement) => {
```

style nit

jwunderl (commented):
(what version of node are you using (rune `node -v)? it seems like the skillmap npm might be from a different version / would be better to remove that change form this pr?)

webapp/src/marked.tsx
thelucasfiles: yes, that makes sense. thanks for feedback!

.vscode/launch.json
```diff
@@ -11,7 +11,7 @@
 				"serve",
 				"--rebundle"
 			],
-			"cwd": "${workspaceRoot}/../pxt-microbit",
```
jwunderl, line 14: I don't know that anyone uses this launch command but typically it's better to avoid committing things unrelated to your pr like this as it can distract from the other changes / if anyone else is using it they be surprised. Same with removing the `"name": "checkdocs"`, below~

react-common/components/controls/EmbedVideo.tsx
thelucasfiles: good point, thanks for feedback!

jwunderl (approved):
LGTM

react-common/components/controls/EmbedVideo.tsx
```diff
@@ -1,45 +1,67 @@
...
+import { useRef, useEffect } from 'react';
+import * as React from 'react';
+import { MediaPlayer } from "dashjs";
 
 export interface EmbedVideoProps extends ControlProps {
-  src: "youtube";
-  id: string;
-  autoplay?: boolean;
-  allowFullScreen?: boolean;
+    src: "youtube" | "streams";
```
abchatra, line 8: Lets not call it streams. No one knows this other than us. Call it azuremedia

thelucasfiles: makes sense, thanks for feedback!

react-common/components/controls/EmbedVideo.tsx
```diff
@@ -1,45 +1,67 @@
...
+            </div>
+        );
 
-    let st1 = "encrypted-media; picture-in-picture";
+    } else {
 
-    if (pxt.BrowserUtils.isIOS()){
-        st1 += "; muted=true";
-    }
+        videoURL = `https://makecode-lucas-testing-makecodetempmediaservice-usea.streaming.media.azure.net/${id}/manifest(format=mpd-time-cmaf)`;
```
abchatra, line 47: This shouldn't be checked-in?

abchatra: This can change between various subscription or we create a new resources, This needs to be inferred. Similar to immersive reader.

thelucasfiles: fair point, i wonder if another parameter might have to get passed in to account for the specific streaming endpoint as well as locator will be used

react-common/components/controls/EmbedVideo.tsx
```diff
@@ -1,45 +1,69 @@
...
+    src: "youtube" | "azuremedia";
+    id: string;
+    autoplay?: boolean;
+    allowFullScreen?: boolean;
 }
 
-
-
 export const EmbedVideo = (props: EmbedVideoProps) => {
+    let endpointName;
```
jwunderl, line 16:
```suggestion
    const endpointName = "makecode-lucas-testing-makecodetempmediaservice-usea";
```

I believe abhijith just mean to move this to get the temp part of the endpoint name out of the built url, and see below \/

react-common/components/controls/EmbedVideo.tsx
```diff
@@ -1,45 +1,69 @@
...
+            </div>
+        );
 
-    let st1 = "encrypted-media; picture-in-picture";
+    } else {
 
-    if (pxt.BrowserUtils.isIOS()){
-        st1 += "; muted=true";
-    }
+        videoURL = `${endpointName}/${id}/manifest(format=mpd-time-cmaf)`;
```
jwunderl, line 49:
```suggestion
        videoURL = `https://${endpointName}.streaming.media.azure.net/${id}/manifest(format=mpd-time-cmaf)`;
```

since the endpoint will change in staging / prod / etc but streaming.media.azure would theoretically be the same~

thelucasfiles: oh yeah that makes sense, thanks!

pxtlib/docsrender.ts
```diff
@@ -420,9 +420,13 @@ namespace pxt.docs {
...
 
             if (href.startsWith("youtube:")) {
-                    let out = '<div class="tutorial-video-embed"><iframe src="https://www.youtube.com/embed/' + href.split(":").pop()
-                        + '" title="' + title + '" frameborder="0" ' + 'allowFullScreen ' + 'allow="autoplay; picture-in-picture"></iframe></div>';
-                    return out;
+                let out = '<div class="tutorial-video-embed"><iframe src="https://www.youtube.com/embed/' + href.split(":").pop()
+                    + '" title="' + title + '" frameborder="0" ' + 'allowFullScreen ' + 'allow="autoplay; picture-in-picture"></iframe></div>';
+                return out;
+
+            } else if (href.startsWith("streams:")) {
```
abchatra, line 427: streams: should be azuremedia: as well

pxtlib/docsrender.ts
```diff
@@ -420,9 +420,13 @@ namespace pxt.docs {
...
+                let out = '<div class="tutorial-video-embed"><video class="ams-embed" controls src="https://makecode-lucas-testing-makecodetempmediaservice-usea.streaming.media.azure.net/' + href.split(":").pop() + '/manifest(format=mpd-time-cmaf)" /></div>';
```
abchatra, line 428: Please make a variable for the resource name here as well.

abchatra (approved):

pxtlib/docsrender.ts
thelucasfiles: Will update, thanks for feedback!

abchatra (approved):

pxtlib/docsrender.ts
thelucasfiles: updated, thanks for feedback!

PR #8951, opened 2022-06-30
Fix share copy in chrome

riknoll:
Fixes https://github.com/microsoft/pxt-minecraft/issues/2184

The new clipboard API isn't supported in all older versions of chrome. Luckily, the old deprecated copy api does work.


jwunderl (approved):

jwunderl:
oh, are these all possibly the same issue too just in different spots? https://github.com/microsoft/pxt-minecraft/issues/2132  https://github.com/microsoft/pxt-minecraft/issues/2034 https://github.com/microsoft/pxt-minecraft/issues/1572

riknoll:
could be!

PR #8953, opened 2022-07-06
Don't allow multiple events with the same arguments to be decompiled

riknoll:
Fixes https://github.com/microsoft/pxt-minecraft/issues/860

I kinda can't believe i never implemented this before...

This makes it so that the decompiler no longer allows for multiple events with the same arguments to be decompiled. Previously we would decompile them and the blockly compiler would then disable them since you can't have multiple events with the same args in blocks. Now one of the offending calls will end up as a grey block

### DIFF
--- a/pxtcompiler/emitter/decompiler.ts
+++ b/pxtcompiler/emitter/decompiler.ts
@@ -609,11 +609,12 @@ ${output}</xml>`;
             if (attributes.blockAllowMultiple) return undefined;
             if (attributes.blockHandlerKey) return attributes.blockHandlerKey;
 
-            // FIXME: this might have some issues with whitespace?
             const args = callInfo.args.filter(arg => !isArrowFunction(arg) && !isFunctionExpression(arg)).map(arg => getArgKeyRecursive(arg)).join("$$");
             return attributes.blockId + "-" + args;
         }
 
+        // custom function to get the key for an argument. we do this instead of simply calling getText() because this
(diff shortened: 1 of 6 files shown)


pxtcompiler/emitter/decompiler.ts
```diff
@@ -2369,7 +2421,22 @@ ${output}</xml>`;
                 }
             }
 
-            eventStatements.map(n => getStatementBlock(n, undefined, undefined, false, topLevel)).forEach(emitStatementNode);
+            const compiledEvents: StatementNode[] = [];
+            const keyMap: {[index: string]: boolean} = {};
```
jwunderl, line 2425:
```suggestion
            const keyMap: pxt.Map<boolean> = {};
```

nit

pxtcompiler/emitter/decompiler.ts
```diff
@@ -2369,7 +2421,22 @@ ${output}</xml>`;
...
+            for (const statement of eventStatements) {
```
jwunderl, line 2426: In theory this should be going through the statements in the order of the program, right? If so, then we should be able to say that since these would get dropped in on start we'll typically maintain the correct behavior for repeated event handlers from javascript -> blocks most of the time right (/ the correct event will usually be set at the end)?

(except I guess if the first event was intentionally registered after other things in on start for some reason, or if `afterOnStart=true` is set on the block.)

jwunderl (approved):

PR #8964, opened 2022-07-12
fix special ext search params

jwunderl:
for e.g. minecraft tests


jwunderl:
for ref this used to occur here: https://github.com/microsoft/pxt/blob/fixBetaExtSearch/webapp/src/scriptsearch.tsx#L227

abchatra (approved):

abchatra:
Port to stable

PR #8965, opened 2022-07-12
css updates for video sizing in tutorials

thelucasfiles:


theme/tutorial.less
```diff
@@ -221,6 +221,24 @@ body#docs.tutorial {
...
+}
+
+.ui.modal {
+    .yt-embed {
+        width: 100%;
+        height: 30rem;
+        max-width: 80%;
+        max-height: 50vh;
+    }
+}
```
jwunderl, line 240:
```suggestion
.ui.modal {
    .ams-embed, .yt-embed {
        width: 100%;
        height: 30rem;
        max-width: 80%;
        max-height: 50vh;
    }
}
```

since you wan the same css to apply to both it just needs a comma separate list like this~ This also helps since you most likely want the size to be consistent between the two so when we need to make a change in the future we don't forget to update one

jwunderl (commented):
It's usually nice to have a few before / after screenshots in the description for changes like this~

theme/tutorial.less
thelucasfiles: good point! thanks for suggestion, should have remembered that!

theme/tutorial.less
```diff
@@ -221,6 +221,15 @@ body#docs.tutorial {
     width: 100%;
 }
 
+.ui.modal {
+    .ams-embed, .yt-embed {
+        width: 100%;
```
jwunderl, line 226:
```suggestion
```
I think this one can go too actually? or at least can set it to 80%; the 80% is less than 100% after all so having the max-width always less than this is a little confusing

thelucasfiles: oh yeah, that makes sense. thanks!

jwunderl (approved):

PR #8972, opened 2022-07-21
Add support for mixer synth engine to sound effect editor

riknoll:
This PR adds support for the mixer synth engine that @mmoskal wrote in pxt-common-packages to the sound effect editor.

In order to mimic the interpolation curves and the vibrato/tremolo effects, each sound is compiled to 20 synth instructions. Each instruction is 12 bytes long, so the end result is 240 bytes per sound which is roughly equivalent to a 16 x 32 image in arcade. If that ends up being too much (though I don't think it will) we can tweak the number of instructions.

As part of this I also rewrote the sim implementation of the synth because the old one wasn't so great at playing multiple short instructions in quick succession.

Here's a test build of arcade with the changes:

https://arcade.makecode.com/app/37e2d1ae2e70da5c0e4010586849d9ceca186a0f-895a1917a8

Please note that the test build currently ignores the global volume setting.

Some future work items:

1. Add more things to the gallery. I bet a lot of the builtin sounds would be easy to convert
2. Change the rendering of the noise wave since it's not actually noise
3. Make sounds into assets?

### DIFF
--- a/pxtblocks/fields/field_sound_effect.ts
+++ b/pxtblocks/fields/field_sound_effect.ts
@@ -13,6 +13,7 @@ namespace pxtblockly {
         waveFieldName: string;
         interpolationFieldName: string;
         effectFieldName: string;
+        useMixerSynthesizer: any;
     }
 
     const MUSIC_ICON_WIDTH = 20;
@@ -38,6 +39,7 @@ namespace pxtblockly {
             if (!this.options.waveFieldName) this.options.waveFieldName = "waveShape";
             if (!this.options.interpolationFieldName) this.options.interpolationFieldName = "interpolation";
             if (!this.options.effectFieldName) this.options.effectFieldName = "effect";
+            if (!this.options.useMixerSynthesizer) this.options.useMixerSynthesizer = false;
 
             this.redrawPreview();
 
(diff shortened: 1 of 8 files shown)


pxtblocks/fields/field_sound_effect.ts
```diff
@@ -13,6 +13,7 @@ namespace pxtblockly {
         waveFieldName: string;
         interpolationFieldName: string;
         effectFieldName: string;
+        useMixerSynthesizer: any;
```
jwunderl, line 16: from the looks of it, this is set as any because the param is usually a string but after the constructor you scope this down to boolean with isTrue? Almost feels worth adding an extend interface to avoid `any` here to me like this

```typescript
        export interface ParsedFieldSoundEffectOptions extends FieldSoundEffectParams {
            useMixerSynthesizer: boolean;
        }
```

but that's pretty nitpicky

jwunderl (approved):
I feel like I should note that my review is based on the code looking good, and I have almost no context on the implementation of the difference portions of sound effects / the differences between tremolo/vibrato/warble/etc are basically magic to me :) (I guess maybe I'll learn from playing with this field editor)

riknoll:
@jwunderl that's totally fine! but here's a little explainer just in case:

Each sound instruction has a start/end frequency and a start/end volume that it goes between using a linear function. To simulate the different interpolation curves, I'm just breaking it into multiple steps to approximate the curve.

As for the effects, they're actually all just variations on the same concept:

1. Vibrato increases the frequency on odd steps and decreases it on even steps
2. Tremolo does the same thing but with the volume instead of the frequency
3. Warble is the same as vibrato but it moves the frequency by a higher amount

The effects are really easy to hear if you play a sound with the same start/end frequency and volume and add the effect

riknoll:
Also, yes this only goes between two points and does not support the multiple point UI that I was playing with before

PR #8976, opened 2022-07-22
include language specification in captioning for youtube videos

thelucasfiles:
note: changed endpoint from staging to prod

### DIFF
--- a/pxtlib/docsrender.ts
+++ b/pxtlib/docsrender.ts
@@ -418,7 +418,7 @@ namespace pxt.docs {
 
     export function setupRenderer(renderer: marked.Renderer) {
         renderer.image = function (href: string, title: string, text: string) {
-            const endpointName="makecode-lucas-testing-makecodetempmediaservice-usea";
+            const endpointName="makecodeprodmediaeastus-usea";
             if (href.startsWith("youtube:")) {
                 let out = '<div class="tutorial-video-embed"><iframe class="yt-embed" src="https://www.youtube.com/embed/' + href.split(":").pop()
                     + '" title="' + text + '" frameborder="0" ' + 'allowFullScreen ' + 'allow="autoplay; picture-in-picture"></iframe></div>';
--- a/pxtlib/main.ts
+++ b/pxtlib/main.ts
@@ -389,7 +389,6 @@ namespace pxt {
         authcodeUrl?: string; // "/beta---authcode"
         isStatic?: boolean;
         verprefix?: string; // "v1"
-        isChinaIp?: boolean;
     }
 
     export function localWebConfig() {
--- a/pxtlib/main.ts
+++ b/pxtlib/main.ts
@@ -389,6 +389,7 @@ namespace pxt {
         authcodeUrl?: string; // "/beta---authcode"
         isStatic?: boolean;
         verprefix?: string; // "v1"
+        isChinaIp?: boolean;
     }
 
     export function localWebConfig() {
--- a/webapp/src/marked.tsx
+++ b/webapp/src/marked.tsx
@@ -294,6 +294,8 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
             const src = inlineVideo.getAttribute('src');
             var url=new URL(src);
             url.searchParams.append('hl',lang);
+            inlineVideo.setAttribute('src',url.toString());
+
             });
 
         pxt.Util.toArray(content.querySelectorAll('Video.ams-embed'))
--- a/webapp/src/marked.tsx
+++ b/webapp/src/marked.tsx
@@ -291,7 +291,7 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
         pxt.Util.toArray(content.querySelectorAll('iframe.yt-embed'))
             .forEach((inlineVideo: HTMLElement) => {
                 let lang = pxt.appTarget.appTheme?.defaultLocale ?? "en";
-            let src=content.querySelectorAll('iframe.yt-embed')[0].getAttribute('src');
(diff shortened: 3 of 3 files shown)


pxtlib/main.ts
```diff
@@ -389,6 +389,7 @@ namespace pxt {
         authcodeUrl?: string; // "/beta---authcode"
         isStatic?: boolean;
         verprefix?: string; // "v1"
+        isChinaIp?: boolean;
```
jwunderl, line 392: this looks unrelated?

webapp/src/marked.tsx
```diff
@@ -287,6 +287,16 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
 
 
     private renderVideo(content: HTMLElement){
+
+        pxt.Util.toArray(content.querySelectorAll('iframe.yt-embed'))
+        .forEach((inlineVideo: HTMLElement)=>{
+
+            let lang=pxt.appTarget.appTheme?.defaultLocale ?? "en";
```
jwunderl, line 294:
```suggestion
            .forEach((inlineVideo: HTMLElement) => {
                let lang = pxt.appTarget.appTheme?.defaultLocale ?? "en";
```

Make sure to check that the style of your code roughly matches existing code -- when spacing is inconsistent or out of place it can be harder to identify errors (e.g. the indentation of this foreach is wrong so it looks like the scope is wrong).

It may seem super small but when spacing is nice and consistent it can be much easier to parse at a glance, especially when you're like through 100s or 1000s of lines of code

webapp/src/marked.tsx
```diff
@@ -287,6 +287,16 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
...
+            let src=content.querySelectorAll('iframe.yt-embed')[0].getAttribute('src');
```
jwunderl, line 295: this looks wrong? I think you want 
```suggestion
            const src = inlineVideo.getAttribute('src');
```

? This will currently only ever get the src of the first video so if you have multiple youtube videos in the tutorial I think they'll all point to the same thing?

webapp/src/marked.tsx
```diff
@@ -287,6 +287,16 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
...
+            content.querySelectorAll('iframe.yt-embed')[0].setAttribute('src',src+"?hl="+lang);
```
jwunderl, line 296: I think you were relying on passing the start / end time as query parameters already, which would not work with this -- you'd build a url like

`youtube/embed/blah?startTime=1&endTime=3?hl=zh-CN`

-- maybe https://stackoverflow.com/a/44160941 will be helpful to look at~

pxtlib/docsrender.ts
```diff
@@ -418,7 +418,7 @@ namespace pxt.docs {
 
     export function setupRenderer(renderer: marked.Renderer) {
         renderer.image = function (href: string, title: string, text: string) {
-            const endpointName="makecode-lucas-testing-makecodetempmediaservice-usea";
+            const endpointName="makecodeprodmediaeastus-usea";
```
jwunderl, line 421: this is the new prod endpoint / not staging endpoint? pretty small but good to give a shout out to things like this in the pr description

jwunderl (commented):
Few comments \\/

pxtlib/main.ts
thelucasfiles: Ah, fair point. This was a way for me to put up incremental progress from hiding video (and/or remember where that file was), but maybe not a good idea to put mix multiple work streams in one pull request.

thelucasfiles: Thanks for feedback!

webapp/src/marked.tsx
thelucasfiles: Good point, should have run 'beautify' to make sure I had the right spacing/indentation. Thanks!

webapp/src/marked.tsx
thelucasfiles: Good catch, yeah shouldn't have assumed there would be just one video. Thanks!

pxtlib/docsrender.ts
thelucasfiles: good point, thanks for feedback!

pxtlib/main.ts
jwunderl: Ah, yes, typically best to just keep a different branch and if you get a small portion done you can even put up a 'draft' pr -- if you leave traces of separate work in a PR, it can be hard to track later and/or cause issues if a PR gets put on the backlog for some reason -- e.g. you find out you have to test on a new device and wait for that to arrive, then the next PR might be blocked waiting for the previous one to be merged as it's missing some code or something, or when someone is reviewing the history they might miss a seemingly unrelated pr adding a crucial feature.

(these things are not likely to occur with one line changes like this as much, but the habit is more important -- it's easy for one line changes to become functions or small features if you're not paying attention :) )

webapp/src/marked.tsx
thelucasfiles: i think it's fixed now, thanks for helpful reference!

pxtlib/main.ts
thelucasfiles: thanks for explanation, have removed these changes!

pxtlib/docsrender.ts
jwunderl: (quick note for next time, pr description typically refers to this portion 
image
where it looked like you changed the title -- not a big deal but short title is usually fine as long as there is further details in description)

thelucasfiles: ah, dang was having trouble finding that, thanks for letting me know!

PR #8984, opened 2022-07-28
Fix colorpicker default

riknoll:
I know there is an issue for this *somewhere* but I cannot find it for the life of me! Also reported on the forum [here](https://forum.makecode.com/t/new-version-jacdac-same-color-error/14635)

Edit: thanks @jwunderl fixes https://github.com/microsoft/jacdac/issues/1184

### DIFF
--- a/libs/pxt-common/pxt-helpers.ts
+++ b/libs/pxt-common/pxt-helpers.ts
@@ -660,13 +660,13 @@ namespace __internal {
 
     /**
      * Get the color wheel field editor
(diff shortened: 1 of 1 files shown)


jwunderl (approved):
ah, I just saw the issue https://github.com/microsoft/jacdac/issues/1184#issue-1305409249

pelikhan:
Will you push this fix in microbit as well?

riknoll:
sure, i'll make sure it goes into the next patch release we push out

PR #8985, opened 2022-07-29
Include support for specifying start/end time for Azure Videos

thelucasfiles:


webapp/src/marked.tsx
```diff
@@ -299,13 +300,24 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
...
                 let player = MediaPlayer().create()
                 player.initialize(inlineVideo, inlineVideo.getAttribute("src"));
+                const src = inlineVideo.getAttribute('src');
+                let url = new URL(src);
+                const END_TIME = url.searchParams.get("endTime");
+                player.on(
+                    dashjs.MediaPlayer.events.PLAYBACK_TIME_UPDATED,
+                    (e: dashjs.PlaybackTimeUpdatedEvent) => {
+                        if (parseInt(END_TIME) <= e.time) {
+                            player.pause();
```
jwunderl, line 313: One thing you might check for is a 'set time' function and set the current time in the video to END_TIME, so if the user tries to scrub past the end of the video section it gets 'locked' back to `END_TIME`. This is not strictly necessary though / doesn't have to be in this pr

jwunderl (approved):
YesNiceGIF

webapp/src/marked.tsx
thelucasfiles: thanks for recommendation, i'll put this on to-do for later!

PR #8990, opened 2022-08-02
Add tick events for starting a video

thelucasfiles:
This supports both YouTube and Azure Media.


jwunderl (approved):
Just to make sure, you ran and checked that the player.on(... PLAYBACK_STARTED) event ran when you started the video, right? good to double check even if the event feels self explanatory hah

thelucasfiles:
> Just to make sure, you ran and checked that the player.on(... PLAYBACK_STARTED) event ran when you started the video, right? good to double check even if the event feels self explanatory hah

sounds good, thanks for feedback!

PR #8991, opened 2022-08-02
Add an editor extension for 'program too large' errors

riknoll:
This is a workaround for https://github.com/microsoft/pxt-microbit/issues/4788

It doesn't fix the code size issues, but does allow multivariant editors to show a dialog when this happens and optionally recompile with fewer variants selected.

In other words: it lets us add a "download for V2 only" option to unblock micro:bit v2 users

Here is a gif showing how this works. Please note that the dialog is **very much not final** (that will be a different PR)

download-for-v2

### DIFF
--- a/pxtcompiler/emitter/emitter.ts
+++ b/pxtcompiler/emitter/emitter.ts
@@ -247,7 +247,7 @@ namespace ts.pxtc {
         console.log(stringKind(n))
     }
 
-    // next free error 9283
+    // next free error 9284
     function userError(code: number, msg: string, secondary = false): Error {
         let e = new Error(msg);
         (<any>e).ksEmitterUserError = true;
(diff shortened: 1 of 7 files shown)


Jaqster:
Thanks for doing this Richard.  Couple questions - do we have telemetry to see how often this gets hit?  And will compiling to v2 only always fix the problem?

riknoll:
@Jaqster I'll add telemetry.

And it will fix the problem unless they write a program that is also too large for V2 (which would be quite a feat). In that case they will get an error just like before and the compilation will fail.

jwunderl (approved):

finneyj:
hey @riknoll - Just noticed this PR during a weekly call with Abhijith. nice work, and great to have this feedback to users, but I wondered if using the word "compiling" is in line with the normal terminology we'd use with teachers/students?

riknoll:
@finneyj that was just placeholder text for that GIF! you can see the "final" dialog here:
https://github.com/microsoft/pxt-microbit/pull/4824#issuecomment-1220953203

PR #8999, opened 2022-08-05
update css to improve readability

thelucasfiles:
hey claire! wanted to get your opinion on the styling here if the readability is improved..thanks! :)

image

### DIFF
--- a/theme/greenscreen.less
+++ b/theme/greenscreen.less
@@ -36,6 +36,9 @@
     .monaco-editor .lines-content, .monaco-editor .view-line, .monaco-editor .view-lines {
         text-shadow: 1px 1px #000;
         font-weight: 500;
+        > span {
+            background: @editorToolsBackground;
+        }
     }
 
     .blocklyMainBackground {
--- a/theme/greenscreen.less
+++ b/theme/greenscreen.less
@@ -34,8 +34,7 @@
     }
 
     .monaco-editor .lines-content, .monaco-editor .view-line, .monaco-editor .view-lines {
-        text-shadow: 1px 1px #000;
-        font-weight: 500;
+
         > span {
             background: @editorToolsBackground;
         }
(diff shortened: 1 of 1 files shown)


clkantner:
Is the image on the left before and the image on the right after? I'm having trouble understanding what I need to review. On the left, the text is difficult to read on the background, and if the background is a video or moving image, it will be hard to tell if the text will always work. On the right, It's also not readable because there's a drop shadow behind the text so it appears blurry. For best results, I would suggest going with the right using a highlight, but have no drop shadow. If you did something like Netflix captions where it's white text over a black bar, that's the best scenario so the contrast is high.

thelucasfiles:
> Is the image on the left before and the image on the right after? I'm having trouble understanding what I need to review. On the left, the text is difficult to read on the background, and if the background is a video or moving image, it will be hard to tell if the text will always work. On the right, It's also not readable because there's a drop shadow behind the text so it appears blurry. For best results, I would suggest going with the right using a highlight, but have no drop shadow. If you did something like Netflix captions where it's white text over a black bar, that's the best scenario so the contrast is high.

Sorry, should have given more context! The left photo was WITH the new changes and the right photo was WITHOUT those changes.  There's a setting where you can have the camera displaying your image underneath your code.

theme/greenscreen.less
```diff
@@ -35,6 +35,7 @@
 
     .monaco-editor .lines-content, .monaco-editor .view-line, .monaco-editor .view-lines {
         text-shadow: 1px 1px #000;
+        font-weight: 500;
```
jwunderl, line 38: hm claires suggestion sounds somewhat like where we started, something like

image

```suggestion
        font-weight: 500;
        > span {
            background: @editorToolsBackground;
        }
```

(plus deleting the text shadow on the previous line too, but I don't think the suggestion feature allows making suggestions on non-modified lines)

thelucasfiles: Sounds good, this makes sense!

jwunderl (approved):

PR #9004, opened 2022-08-10
Fix Asset Create Button Formatting

thsparks:
Specify flex properties for the common-button-flex contained within asset-editor-create-button, ensuring the content of the button (image & text) will wrap as needed to fit within the shape of the button.

Fixes https://github.com/microsoft/pxt-arcade/issues/4839

Previously, these buttons relied on white-space wrapping to get this behavior, but that has since been disabled and the flex approach is hopefully more robust.

Before fix:
image

After fix:
image

### DIFF
--- a/theme/asset-editor.less
+++ b/theme/asset-editor.less
@@ -231,7 +231,6 @@
 
     .common-button-flex {
         display: flex;
-        flex-wrap: wrap;
         flex-direction: column;
         justify-content: center;
     }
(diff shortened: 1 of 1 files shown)


riknoll (approved):

theme/asset-editor.less
```diff
@@ -229,6 +229,12 @@
         margin: 1rem;
     }
 
+    .common-button-flex {
+        display: flex;
+        flex-wrap: wrap;
```
riknoll, line 234: not sure if `flex-wrap` is necessary here

eanders-ms (approved):
:shipit:

PR #9008, opened 2022-08-12
Fix Python Preferred Editor for Tutorials

thsparks:
When Python is set as the preferred editor for a tutorial, we still end up loading the blocks editor because main.py doesn't exist at the time we try to load it.

To address this issue, if the preferred file does not exist, we can create it rather than falling back on the blocks file. We also have to update config to ensure the python files are included in pxt.json. Without that, the python script won't actually be included when the app runs, so it doesn't do anything. This was also missing inside loadTutorialTemplateCodeAsync, so I've added it there, too.

Fixes https://github.com/microsoft/pxt-arcade/issues/4892

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1651,7 +1651,7 @@ export class ProjectView
                     file = main.lookupFile(fileName);
 
                     // If the preferred file does not exist, create it.
-                    if(!file) {
+                    if (!file) {
                         file = main.setFile(fileName, '\n');
 
                         pkg.mainPkg.config.files.push(fileName);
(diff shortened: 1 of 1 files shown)


webapp/src/app.tsx
```diff
@@ -1647,7 +1647,16 @@ export class ProjectView
                 // override inferred editor with tutorial editor if present in markdown
                 const tutorialPreferredEditor = h.tutorial?.metadata?.preferredEditor;
                 if (tutorialPreferredEditor) {
-                    file = main.lookupFile("this/" + filenameForEditor(tutorialPreferredEditor)) || file;
+                    let fileName  = "this/" + filenameForEditor(tutorialPreferredEditor);
+                    file = main.lookupFile(fileName);
+
+                    // If the preferred file does not exist, create it.
+                    if(!file) {
```
jwunderl, line 1654:
```suggestion
                    if (!file) {
```

small nit: we usually do spacing between keywords / operators~

jwunderl (approved):
LGTM!

riknoll (approved):

PR #9010, opened 2022-08-18
Fixed color theming in extension dialog.

srietkerk:
When clicking on extensions in the development environment of micro:bit, the extensions header was not the correct color. This code change fixed this issue and it was verified that the right color was maintained across pxt.

Closes https://github.com/microsoft/pxt-arcade/issues/4782

### DIFF
--- a/skillmap/package-lock.json
+++ b/skillmap/package-lock.json
@@ -23,7 +23,8 @@
     },
     "../built/react-common/components": {
       "name": "react-common",
-      "version": "0.0.1"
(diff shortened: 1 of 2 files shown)


jwunderl (approved):
OtterClappingGIF


.less change looks good!

I've seen similar package-lock changes locally on occasion before but I'm not clear on why it's downgrading the version number for these two? @riknoll any idea there?

riknoll:
@jwunderl nope. I'd just leave package-lock out of this pr

PR #9012, opened 2022-08-18
Add support for multiple auth providers

eanders-ms:
Update the sign-in dialog to support multiple identity providers. Additionally, the dialog can now display "sign in" and "sign up" variants. This is to help guide the user in case they might not have a Microsoft or Google account yet, or want to make a new one.

Sample:
image
and
image

The buttons in both cases take the user to the same login flow, where they have the option to create a new account.

Unfinished/Todo later:
- Identity docs need refreshed.
- It was proposed to add an additional `Sign Up` button to the header bar, next to `Sign In`. I'm deferring this work for now, as the header bar is already crowded. This needs more discussion.

### DIFF
--- a/react-common/components/controls/Link.tsx
+++ b/react-common/components/controls/Link.tsx
@@ -2,8 +2,9 @@ import * as React from "react";
 import { classList, ContainerProps } from "../util";
 
 export interface LinkProps extends ContainerProps {
-    href: string;
+    href?: string;
     target?: "_self" | "_blank" | "_parent" | "_top";
+    onClick?: () => void;
 }
 
 export const Link = (props: LinkProps) => {
@@ -29,6 +30,7 @@ export const Link = (props: LinkProps) => {
             href={href}
             target={target}
             rel={target === "_blank" ? "noopener noreferrer" : ""}
+            onClick={props.onClick}
             >
             {children}
         </a>
--- a/react-common/components/controls/Link.tsx
+++ b/react-common/components/controls/Link.tsx
@@ -2,7 +2,7 @@ import * as React from "react";
 import { classList, ContainerProps } from "../util";
 
 export interface LinkProps extends ContainerProps {
-    href?: string;
+    href: string;
     target?: "_self" | "_blank" | "_parent" | "_top";
 }
 
--- a/react-common/components/controls/Link.tsx
+++ b/react-common/components/controls/Link.tsx
@@ -4,7 +4,6 @@ import { classList, ContainerProps } from "../util";
 export interface LinkProps extends ContainerProps {
     href?: string;
     target?: "_self" | "_blank" | "_parent" | "_top";
-    onClick?: () => void;
 }
 
 export const Link = (props: LinkProps) => {
@@ -30,7 +29,6 @@ export const Link = (props: LinkProps) => {
             href={href}
             target={target}
             rel={target === "_blank" ? "noopener noreferrer" : ""}
-            onClick={props.onClick}
             >
             {children}
         </a>
--- a/react-common/components/profile/SignInModal.tsx
+++ b/react-common/components/profile/SignInModal.tsx
@@ -1,0 +1,100 @@
+/// <reference path="../types.d.ts" />
+
+import * as React from "react";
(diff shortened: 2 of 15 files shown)


react-common/components/profile/UserPane.tsx
```diff
@@ -27,8 +29,12 @@ export const UserPane = (props: UserPaneProps) => {
 
     return <div className="profile-user-pane">
         <div className="profile-portrait">
-            { picture?.dataUrl ?
-                <img src={picture?.dataUrl} alt={pxt.U.lf("Profile Picture")} />
+            { picUrl ?
+                // Google user picture URL must have referrer policy set to no-referrer
+                <div dangerouslySetInnerHTML={{__html: pxt.BrowserUtils.imgTag(picUrl, {
```
eanders-ms, line 34: Alternatively, we could add a meta tag to change the referrer policy globally:
```
<meta name="referrer" content="no-referrer" />
```
But this will affect all `fetch` calls. I'm not sure what it will break.

eanders-ms: Open to suggestions here. I don't love this solution.

react-common/components/profile/UserPane.tsx
```diff
@@ -27,8 +29,13 @@ export const UserPane = (props: UserPaneProps) => {
 
     return <div className="profile-user-pane">
         <div className="profile-portrait">
-            { picture?.dataUrl ?
-                <img src={picture?.dataUrl} alt={pxt.U.lf("Profile Picture")} />
+            { picUrl ?
+                // Google user picture URL must have referrer policy set to no-referrer
+                // eslint-disable-next-line: react/no-danger
+                <div dangerouslySetInnerHTML={{__html: pxt.BrowserUtils.imgTag(picUrl, {
```
jwunderl, line 35: Can we patch up the declaration of ImgHTMLAttributes to get around having to do this? e.g. throw this at the top of the file

```typescript
declare module 'react' {
    interface ImgHTMLAttributes<T> {
      referrerPolicy?:  "no-referrer" | "no-referrer-when-downgrade" | "origin" | "origin-when-cross-origin" | "unsafe-url";
    }
}
```

and the typing should work I believe (I just tested in a random tsx file)

image

Benefit there is that it gets rid of runtime weirdness, just leaving compile time only weirdness

jwunderl (approved):
lgtm besides avoiding those eslint disables if possible with \\/

react-common/components/profile/UserPane.tsx
jwunderl: Ah only saw this now but suggestion in my review below \\/

react-common/components/profile/UserPane.tsx
eanders-ms: Oh good suggestion, let me try that.

react-common/components/controls/Link.tsx
```diff
@@ -2,8 +2,9 @@ import * as React from "react";
 import { classList, ContainerProps } from "../util";
 
 export interface LinkProps extends ContainerProps {
-    href: string;
+    href?: string;
     target?: "_self" | "_blank" | "_parent" | "_top";
+    onClick?: () => void;
```
riknoll, line 7: instead of this, use the link-button class on the Button component. better for accessibility

react-common/components/profile/SignInModal.tsx
```diff
@@ -0,0 +1,100 @@
...
+                                isChecked={rememberMe}
+                                onChange={setRememberMe}
+                            ></Checkbox>
+                        </div>
+                    </div>
+                </div>
+                <div className='signin-footer'>
+                    {footerFragment}
+                    <div className='learn'>
+                        <Link href='/identity/sign-in' target='_blank' ariaHidden={true}>
```
riknoll, line 92: why is this aria hidden?

riknoll (approved):
LGTM except for the one comment on link-button

react-common/components/profile/SignInModal.tsx
eanders-ms: It shouldn't be. Will remove.

PR #9015, opened 2022-08-19
Fix Python destructuring for global/nonlocal variables

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/4826

When a variable is already declared, we can't convert python destructuring into the TS syntax. Instead, fall back to a series of assigments. for example,

```python
g = 0
def initValue():
    global g

    g, _ = 1, 2
```
becomes

```typescript
let g = 0
function initValue() {
    let __tempvar2 = [1, 2]
    g = __tempvar2[0]
    let _ = __tempvar2[1]
}
```

### DIFF
--- a/tests/pyconverter-test/baselines/destructuring.ts
+++ b/tests/pyconverter-test/baselines/destructuring.ts
@@ -1,0 +1,12 @@
+function initValue() {
+
+    let __tempvar2 = [1, 0, 3]
+    g_value = __tempvar2[0]
+    let _ = __tempvar2[1]
+    _ = __tempvar2[2]
+    let [a, b, c] = ["A", "B", "C"]
+    let [d, e, f] = ["D", "E", "F"]
+}
+
+let g_value = 0
+initValue()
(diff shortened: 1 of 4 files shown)


tests/pyconverter-test/cases/destructuring.py
```diff
@@ -0,0 +1,9 @@
+def initValue():
```
thsparks, line 1: Am I correct in understanding this will be picked up automatically as a test case simply because of its file path, then compared with the baseline by the same name?

tests/pyconverter-test/baselines/destructuring.ts
```diff
@@ -0,0 +1,12 @@
+function initValue() {
+
+    let ___tempvar2 = [1, 0, 3]
```
thsparks, line 3: Any particular reason this starts with tempvar2 instead of tempvar0?

thsparks (approved):
Very neat!

jwunderl (approved):

tests/pyconverter-test/cases/destructuring.py
riknoll: yup!

tests/pyconverter-test/baselines/destructuring.ts
riknoll: our python compiler makes multiple passes and i'm not resetting the variable on each pass. no one should ever see this code, so i decided not to fix it.

PR #9019, opened 2022-08-24
.blocklyNonEditableText>text.semanticIcon now has correct styling for…

srietkerk:
… font-family and fill

In the third level, **Get Animated**, of the Whackem tutorial in beta (the sixth step), there is a preview of an animation block that does not show the preview icon for the field animator. This is because the `semanticIcons` class needs to have `font-family: "Icons";` in CSS. This was not happening for the preview block in the tutorial panel, but now it is.

**Before:**
before-anim-preview-icon

**After:**
after-anim-preview-icon

Closes  https://github.com/microsoft/pxt-arcade/issues/4921 

Note: https://github.com/microsoft/pxt/blob/dd7494242ef28baf69d93d67e50e35ca2c3597e6/pxtblocks/fields/field_animation.ts#L139 this line was removed from `field_animation.ts` as, with playing, it did not seem to change anything on the blocks.

### DIFF
--- a/pxtblocks/fields/field_animation.ts
+++ b/pxtblocks/fields/field_animation.ts
@@ -136,7 +136,6 @@ namespace pxtblockly {
 
             const icon = new svg.Text("\uf008")
                 .at(X_PADDING, 5 + (TOTAL_HEIGHT >> 1))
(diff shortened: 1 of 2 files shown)


jwunderl (approved):
LGTM, nice!

Issue #8142, opened 2021-04-25
Block "Math.abs" can't load in "Guitar"


Sheila111:
**Describe the bug**
Block "**Math.abs**" can't load in "**Guitar**"

**Steps to reproduce the behavior**
1.Navigate to https://makecode.microbit.org/projects/guitar/accelerometer
2.Click "**Math.abs**" 

**Expect behavior**
It can load

Actual behavior
Block "**Math.abs**" can't load in "**Guitar**"
![1](https://user-images.githubusercontent.com/59192069/115978707-fac6e580-a5b3-11eb-8805-4677dd435c92.gif)

**Additional context**
1.OS: Windows(rs6)
2.makecode.microbit.org version: 3.1.79
3.Microsoft MakeCode version: 6.13.25

PR #9020, opened 2022-08-24
Fixes the broken link to Math.abs reference page

srietkerk:
In the Guitar Project for micro:bit, a link to the docs page for Math.abs was broken. Thanks to the edits suggested in the attached issue, the link is no longer broken.

Quick video demonstrating the fix: https://user-images.githubusercontent.com/49178322/186541056-5f59059c-70b3-4094-90e6-64ca0f4368f1.mp4

Closes: https://github.com/microsoft/pxt/issues/8142

**Updates:**
>A block now displays on the `abs` card because, as @jwunderl suggested, the real fix lied within code. When blocks are initialized, the function `mkCard()` is run for those that have the `codeCard` field in their object definition. `mkCard()` pre-pends a block's defined 'help' URL with `/reference` so the redirect takes the user to the specified block's docs page. 

block-card

This fix adds the `codeCard` field to the special case blocks `Math.abs`, `Math.min`, and `Math.max`.

_**Important note**_ 
The solution for `Math.min` and `Math.max` is not ideal. These blocks have the same special-cased **blockId** `math_op2` but have different **qNames**. A `qName` is needed with this solution as it is the entry point to retrieve the block's `SymbolInfo` which is needed for `mkCard()`. Currently, the hardcoded qName used is 'Math.min'. The hardcoded qName makes it so that, in markdown that uses `Math.min(x,y)` or `Math.max(x,y)` cards, the card displays the min block, description, and, when clicked, directs to the min doc page. The bottom of the min doc page has a link to the max doc page since they are related to each other. This is, with this fix, the only way to reach the max block and doc page through cards.  
Ideally, we would be able to grab which function is being called in markdown (min or max). However, since `codeCard` is being set on block initialization, it is a bit tricky and out of scope for this PR/issue.

### DIFF
--- a/libs/pxt-common/pxt-helpers.ts
+++ b/libs/pxt-common/pxt-helpers.ts
@@ -535,7 +535,8 @@ namespace Math {
       * For example, the absolute value of -5 is the same as the absolute value of 5.
       * @param x A numeric expression for which the absolute value is needed.
       */
-    //% help=./reference/math/abs
+    //% blockId=math_op3
+    //% help=math/abs
     export function abs(x: number): number {
         return x < 0 ? -x : x;
     }
--- a/libs/pxt-common/pxt-helpers.ts
+++ b/libs/pxt-common/pxt-helpers.ts
@@ -554,6 +554,7 @@ namespace Math {
     /**
       * Returns the larger of two supplied numeric expressions.
       */
+    //% blockId=math_op2
     //% help=math/max
     export function max(a: number, b: number): number {
         if (a >= b) return a;
@@ -563,6 +564,7 @@ namespace Math {
     /**
       * Returns the smaller of two supplied numeric expressions.
       */
+    //% blockId=math_op2
     //% help=math/min
     export function min(a: number, b: number): number {
         if (a <= b) return a;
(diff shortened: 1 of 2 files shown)


thsparks (approved):

libs/pxt-common/pxt-helpers.ts
```diff
@@ -535,7 +535,7 @@ namespace Math {
       * For example, the absolute value of -5 is the same as the absolute value of 5.
       * @param x A numeric expression for which the absolute value is needed.
       */
-    //% help=math/abs
+    //% help=./reference/math/abs
```
jwunderl, line 538: (sidenote, I don't think we would typically wan the . here, some weird stuff ends up happening at times with the pathing~)

jwunderl (commented):
hm, there shouldn't need to be a reference/ prefix here - e.g. https://github.com/microsoft/pxt-common-packages/blob/master/libs/accelerometer/shims.d.ts#L23 / the other links are all pointing at the right thing. It's also pointing at the right thing if you right click on math.abs block -> help I think? And if it needs to be fixed we'd also have to adjust the paths for all the other help paths in here and in pxt-core.d.ts right?

I would think ideally the fix could live in https://github.com/microsoft/pxt/blob/master/pxtlib/blocks.ts#L243 or so where it generates help paths in the first place as it seems like something must be getting missed there? Or possibly https://github.com/microsoft/pxt/blob/master/pxtblocks/blocklyloader.ts#L1038 / somewhere that calls it if we're special casing pxt-common somewhere

jwunderl (approved):

Issue #7907, opened 2021-02-24
Math: It's hard to see + from ÷


jfo8000:
I've had feedback that it's hard to see + from ÷.

Suggestion: use images in math dropdowns instead of text so that the math operators are more clear?   I have tried changing fonts and stuff but it doesnt make that much of a difference. 

![image](https://user-images.githubusercontent.com/16273166/109062463-117bc800-773c-11eb-8348-8241f1568165.png)

PR #9024, opened 2022-08-26
Changed division symbol to '/' for math blocks.

srietkerk:
Because of how the symbols are rendered, it is hard to tell the difference between a plus sign and the division symbol in the math blocks. This PR changes the classic division symbol to a '/' so there is a noticeable difference between addition and division. 

**Before:**
traditional-division

**After:**
division-forward-slash

Closes: https://github.com/microsoft/pxt/issues/7907


riknoll (approved):

jwunderl (approved):

thsparks (approved):

abchatra:
👍

PR #9027, opened 2022-08-29
Clean Deletable Flags Before Loading a Workspace

thsparks:
Before we load a workspace from xml, clean out any deletable flags that exist in the XML. There are currently no scenarios where we rely on this deletable flag, and it can be persisted erroneously with value "false" if the app closes unexpectedly while in debug mode, which then prevents users from being able to delete/duplicate blocks in their workspace.

This approach should also fix any existing projects that have gotten into that non-deletable state.

Fixes https://github.com/microsoft/pxt-arcade/issues/4906

### DIFF
--- a/webapp/src/blocks.tsx
+++ b/webapp/src/blocks.tsx
@@ -320,7 +320,7 @@ export class Editor extends toolboxeditor.ToolboxEditor {
     private cleanXmlForWorkspace(dom: Element) {
         // Clear out any deletable flags. There are currently no scenarios where we rely on this flag,
         // and it can be persisted erroneously (with value "false") if the app closes unexpectedly while in debug mode.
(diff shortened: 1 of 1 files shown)


riknoll (approved):

webapp/src/blocks.tsx
```diff
@@ -316,6 +317,12 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         return true;
     }
 
+    private cleanXmlForWorkspace(dom: Element) {
+        // Clear out any deletable flags. There are currently no scenarios where we rely on this flag,
+        // and it can be persisted erroneously (with value "false") if the app closes unexpectedly while in debug mode.
+        dom.querySelectorAll("block[deletable]").forEach(b => { b.removeAttribute("deletable") });
```
riknoll, line 323: oh, wait, you should also do shadow[deletable]

jwunderl (approved):

PR #9030, opened 2022-09-02
New share experience

riknoll:
new-share

This PR adds most of the new share experience. A few of the major changes from what's in beta:

1. No longer full screen
2. The choice for snapshot/live share is a checkbox instead of a dropdown
3. Added social buttons for Microsoft Teams, Google Classroom, and WhatsApp
4. Refactored the code to make the interface between react-common and the webapp (and eventually the skillmap) a bit cleaner

Still to do:

1. Also use this dialog in the skillmap instead of the separate dialog it currently has
2. Not super responsive right now. Need to handle mobile/tablet better (but still better than what's in beta)


riknoll:
Here's a test upload:

https://arcade.makecode.com/app/796d150e29fcbc2ec5b7ac3c2ce7179a048b8cd5-a2574c2c33

react-common/components/share/ShareInfo.tsx
```diff
@@ -153,124 +142,139 @@ export const ShareInfo = (props: ShareInfoProps) => {
...
-                                heading={lf("Share on Twitter")} />
-                            {navigator.share && <Button className="circle-button device-share"
-                                title={lf("Show device share options")}
-                                ariaLabel={lf("Show device share options")}
-                                leftIcon={"icon share"}
-                                onClick={handleDeviceShareClick}
+                </div>
+            }
+            <div className="project-share-content">
+                {(prePublish|| shareState === "publish") && <>
```
jwunderl, line 165:
```suggestion
                {(prePublish || shareState === "publish") && <>
```

nit

react-common/components/share/ThumbnailRecorder.tsx
```diff
@@ -0,0 +1,137 @@
...
+    removeStateChangeListener: (handler: (newState: SimRecorderState) => void) => void;
+    removeThumbnailListener: (handler: (uri: string, type: "gif" | "png") => void) => void;
+    removeErrorListener: (handler: (message: string) => void) => void;
+}
+
+
+export const ThumbnailRecorder = (props: ThumbnailRecorderProps) => {
+    const { initialUri, onApply, onCancel, simRecorder } = props;
+    const [ uri, setUri ] =  React.useState(initialUri);
+    const [error, setError] = React.useState<string>(undefined)
```
jwunderl, line 35:
```suggestion
    const [ error, setError] = React.useState<string>(undefined)
```

webapp/src/components/SimRecorder.tsx
```diff
@@ -0,0 +1,187 @@
...
+    return <div id="shareLoanedSimulator" ref={handleContainerRef}>
+    </div>
+}
+
+let ref: SimRecorderRefImpl;
+
+function createSimRecorderRef() {
+    if (ref) return ref;
+    let encoder: screenshot.GifEncoder;
+    ref = {
```
jwunderl, line 77: nit; since ref declaration is hoisted we could just move this to the very bottom without the need for an any cast and then have all the anon functions attached in the object literal definition? Doesn't really matter though I suppose

jwunderl (approved):
LGTM, a few small nitpicks that are fine to ignore or defer for future prs

future pr ideas / nonblocking: 
* would be nice to persist thumbnail locally between shares? Would be nice to avoid having to recreate nice thumbnails for bug fix shares / also could allow for a 'custom thumbnail' if you prebuild a logo cutscene but don't want that to be seen in game
* maybe an icon showing 'empty' when a thumbnail not yet set, it feels a little empty with just a slightly grayer box.

thsparks:
Not sure why this is yet, but I tried the share link and when I took a thumbnail, the sim disappeared and didn't come back when I closed the share dialog.

pxtsim/runtime.ts
```diff
@@ -989,19 +989,16 @@ namespace pxsim {
             if (Runtime.messagePosted) Runtime.messagePosted(data);
         }
 
-        static postScreenshotAsync(opts?: SimulatorScreenshotMessage): Promise<void> {
+        static async postScreenshotAsync(opts?: SimulatorScreenshotMessage): Promise<void> {
```
eanders-ms, line 992: I think we should supply a default thumbnail, or have a stronger call to action to create one. The current default behavior doesn't feel great:
image

If the user hasn't provided a screenshot, can we take one automatically three seconds into gameplay?

react-common/components/share/ThumbnailRecorder.tsx
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    const targetTheme = pxt.appTarget.appTheme;
+
+    const handleSimRecorderRef = (ref: SimRecorderRef) => {
+        setRecorderRef(ref);
+    }
+
+    const screenshotLabel = lf("Take screenshot ({0})", targetTheme.simScreenshotKey);
+    const startRecordingLabel = lf("Record game play ({0})", targetTheme.simGifKey);
```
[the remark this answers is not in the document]
eanders-ms: unsure which is more correct.

eanders-ms (approved):
LGTM, with a few comments.

abchatra:
Few minor UI nits:
- Move the continue to right aligned? Seems odd to be on the left aligned. 
- Font size is  also seems too big for me.
image

- Remove CANCEL and move the continue to right aligned here as well?

image

react-common/components/share/ShareInfo.tsx
```diff
@@ -153,124 +142,139 @@ export const ShareInfo = (props: ShareInfoProps) => {
...
+                    </div>
+                    {showDescription && <>
+                        <Input
+                            className="name-input"
+                            initialValue={name}
+                            placeholder={lf("Name your project")}
+                            onChange={setName} />
+                        {isLoggedIn && <Checkbox
+                            id="persistent-share-checkbox"
+                            label={lf("Allow people to see future changes to my project")}
```
thsparks, line 177:
```suggestion
                            label={lf("Allow people to see future changes to my project.")}
```

thsparks: Just a nit. Wasn't sure which is preferred but based this on other strings in the file which seem to include final period punctuation.

webapp/src/components/SimRecorder.tsx
```diff
@@ -0,0 +1,187 @@
...
+}
+
+export const SimRecorderImpl: SimRecorder = props => {
+    const { onSimRecorderInit } = props;
+
+    const [loanedSimulator, setLoanedSimulator] = React.useState<HTMLElement>(undefined);
+
+    React.useEffect(() => {
+        const sim = simulator.driver.loanSimulator()
+        setLoanedSimulator(sim);
```
thsparks, line 18: So React is still a bit of a mystery to me and I may be missing something obvious...but regarding the issue I mentioned in https://github.com/microsoft/pxt/pull/9030#issuecomment-1235770439, I debugged it a little and the issue seems to be that even after calling `setLoanedSimulator` here, `loanedSimulator` still isn't set. My understanding is that this is sort of expected with React since the updates don't happen immediately, but it *still* wasn't set even when the user clicks "Apply" and return is called down in line 46, which seems unexpected. As a result, `unloadSimulator` was never called, and I'm guessing that's why the sim didn't re-appear once the share dialog was closed.

No idea if this is repro'ed on other people's machines, but it's a consistent repro for me.

thsparks (changes requested):
Mostly nitpicking, but there's one error with the sim disappearing that seems a bit concerning (https://github.com/microsoft/pxt/pull/9030#issuecomment-1235770439). If no-one else is hitting it, maybe not a big deal. But probably worth a look if it's consistent for more folks than just me.

I tried to debug it myself. Seems like something weird going on with the loanedSimulator never getting set even when setLoanedSimulator is called in the SimRecorderImpl, but I haven't figured out why that's happening yet.

riknoll:
@thsparks fixed the sim issue

thsparks (approved):

pxtsim/runtime.ts
riknoll: the plan is to do exactly that, i just didn't do it as part of this PR. we used to do this but i think it broke at some point.

my basic plan is this:
1. try to take a screenshot after x seconds provided the screen isn't blank
2. when the user starts interacting with the sim, record frames for a gif (but put off encoding until they attempt to share)

riknoll:
Going to merge this and do another PR with @abchatra's feedback and mobile css

Jaqster:
Sorry, just seeing this now.  Overall looks great, couple comments:
- Do kids know what a 'thumbnail' is (aside from what's on their hand).  Should we say 'picture' instead?
- Why is part of the 'update thumbnail' dialog disconnected?  I didn't see this the first few times I clicked on it.  Can we make it part of the window?
image

Jaqster:
@unthinkmedia  - do you have thoughts on this?

riknoll:
@Jaqster I'm going to open a PR soon that addresses the buttons. @unthinkmedia gave me a design to work from.

Jaqster:
Also, have we thought through all the implications of removing the QR code?  This is a regression (removing a feature) so I want to be sure about this.  Is there any way to see how many people use/depend on it?

abchatra:
Hmm QR code was not removed.

abchatra:
There is no way to measure QR code today. However I have seen lot of folks use it

thsparks:
> There is no way to measure QR code today. However I have seen lot of folks use it

Maybe not the place for this discussion, but could we have the QR code go to the link with some additional parameter (i.e. QR=true or something) that tells us it was used?

abchatra:
Yes we should add that for future

Jaqster:
Sorry, I was reading the user study write-up where it's removal was mentioned - but I see it here... nevermind!

PR #9033, opened 2022-09-08
Add "Processing Complete" Messages to Gulp Watch

thsparks:
I was finding it difficult to know when gulp watch was finished processing a change, so I added a simple console log to let the user know when it's done. I'm not super familiar with gulp, but this seems to do the trick?

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -150,7 +150,8 @@ function initWatch() {
         webapp,
         browserifyWebapp,
         browserifyAssetEditor,
-        gulp.parallel(semanticjs, copyJquery, copyWebapp, copySemanticFonts, copyMonaco)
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -232,6 +233,10 @@ function pxtcommon() {
     return Promise.resolve();
 }
 
+function notifyProcessingComplete() {
+    console.log("-- Processing Complete --");
```
jwunderl, line 237:
```suggestion
    console.log("-- Build Complete --");
```

maybe? 'Processing' sounds slightly off to me here but it doesn't really matter either way I suppose~

jwunderl (approved):
LGTM, maybe worth adding to `initWatchCli` as well?

eanders-ms (approved):
:shipit:

PR #9035, opened 2022-09-10
Skill map intro page alt text updated to be more descriptive  and visible on dark backgrounds

srietkerk:
The changes in this PR make it so that alt text for skill map landing page backgrounds is no longer black when the background color is dark. I also added to the descriptions given by the alt text in the `InfoPanel` so it is specific to an activity and a skill map.

**Before** 

before-alt

**After** 

updated-alt-text

over-and-under-alt-text

**Note:** There is a check to see how many maps are passed into the `SkillGraphContainer` component because there can be multiple skill graphs in one landing page for a skill map (i.e. https://arcade.makecode.com/--skillmap#beginner). When this happens, we don't want the alt text to cater to a specific map as the background image applies to the overall experience. Saying `background image for skillmap x, y, and z` is a bit too much detail to give in alt text, so I thought giving reference to how many skill graphs were represented in the skill map was good detail.

**Another Note:** Accessibility Insights still gives a flag for no alt text for an image on the usabilla button that we have for feedback. This change is not something that we can access through the iframe we inject on the site. So the following issue won't be fully closed until an alt attribute is added to the image for that button.

Closes https://github.com/microsoft/pxt-arcade/issues/4809


srietkerk:
Another thing that I wanted to note: not having the image load/having a broken link _**does not**_ make the site look like it does in the screenshots that I provided. If a background image is rendered, the skill graphs render according to the size of said background image. This is why you can barely see the skill graphs in the screenshots that I provided. I removed the src of the image tag after the page was completely loaded, thus making the skill graph shrink to it's parent container size. In normal page-loading situations, the image's inability to load would happen before the skill graph renders and thus doesn't actually cause a problem when the site is loaded. I took screenshots in this state because it made it easier to showcase the change.

skillmap/src/components/InfoPanel.tsx
```diff
@@ -83,12 +83,13 @@ export class InfoPanelImpl extends React.Component<InfoPanelProps> {
         const isMap = !node;
         const isActivity = node && !isRewardNode(node);
         const tags = isActivity && (node as MapActivity).tags || undefined;
+        const thumbnailAltText: string = `Thumbnail for ${!subtitle ? "starting" : title} activity in skillmap ${!subtitle ? title : subtitle}`;
```
jwunderl, line 86: we need to pass these through our localization pipeline so the text makes sense for non-english speaking users. This is done through the 'lf' function, and would look like 

```suggestion
        const thumbnailAltText = lf("Thumbnail for {0} activity in skillmap {1}", !subtitle ? lf("starting") : title, !subtitle ? title : subtitle);
```

skillmap/src/components/SkillGraphContainer.tsx
```diff
@@ -67,6 +72,14 @@ export class SkillGraphContainerImpl extends React.Component<SkillGraphContainer
             }
         }
 
+        if (darkBackgroundColors.includes(backgroundColor) ) {
```
jwunderl, line 75: Could this just be using `strokeColor` prop instead of narrowing to a list of currently used background colors? typically I would expect that to be properly contrasting with `backgroundColor` or else we'd likely have other coloring issues.

(and if it needs special casing for high contrast mode I'd think a css rule for `.high-contrast .skill-graph-background img` would handle it for you)

thsparks:
Probably outside the scope of this PR, but @kiki-lee mentioned this in the original issue: "since skillmap thumbnail images use our special text instead of actual image links".

I don't fully understand what that means, but I was wondering what the motivation was behind doing something custom here. Is that something we've looked into?

jwunderl:
> Probably outside the scope of this PR, but @kiki-lee mentioned this in the original issue: "since skillmap thumbnail images use our special text instead of actual image links".
> 
> I don't fully understand what that means, but I was wondering what the motivation was behind doing something custom here. Is that something we've looked into?

i read that as referring to the src of the content in the skillmap being derived from markdown content that gets parsed into the graph / imgs / etc vs just having html inline or something to that effect - which is as expected, just means we have to fill in the rest of the blanks :)

kiki-lee:
> > Probably outside the scope of this PR, but @kiki-lee mentioned this in the original issue: "since skillmap thumbnail images use our special text instead of actual image links".
> > I don't fully understand what that means, but I was wondering what the motivation was behind doing something custom here. Is that something we've looked into?
> 
> i read that as referring to the src of the content in the skillmap being derived from markdown content that gets parsed into the graph / imgs / etc vs just having html inline or something to that effect - which is as expected, just means we have to fill in the rest of the blanks :)

Our code to add a level to the skillmap includes a tag for adding the thumbnail ("imageURL:") and there's no place to include alt text.

Screenshot by Dropbox Capture

thsparks:
> > > Probably outside the scope of this PR, but @kiki-lee mentioned this in the original issue: "since skillmap thumbnail images use our special text instead of actual image links".
> > > I don't fully understand what that means, but I was wondering what the motivation was behind doing something custom here. Is that something we've looked into?
> > 
> > 
> > i read that as referring to the src of the content in the skillmap being derived from markdown content that gets parsed into the graph / imgs / etc vs just having html inline or something to that effect - which is as expected, just means we have to fill in the rest of the blanks :)
> 
> Our code to add a level to the skillmap includes a tag for adding the thumbnail ("imageURL:") and there's no place to include alt text.
> 
> Screenshot by Dropbox Capture

Ah, got it. Maybe we can file a separate task to add that :) But not needed for this PR. Thanks for the clarification!

skillmap/src/components/SkillGraphContainer.tsx
srietkerk: So, `strokeColor` can work, but it won't be a fully inclusive solution. This is because `strokeColor` would be accessed through the app's theme prop which is of type `SkillGraphTheme` ([here](https://github.com/microsoft/pxt/blob/98d18583829a877a7e2ed6373fac888d0a8a1acc/skillmap/src/lib/skillMap.d.ts#L143)), making `strokeColor` the color around the skill map graph nodes. The background images create enough separation from the background color so the `strokeColor` is often just black. If I changed the condition from `if (darkbackgroundColors.includes(backgroundColor))` to something like `if (backgroundColor === strokeColor)` that would solve the problem with strictly black backgrounds but will we will have an accessibility color contrast issue on, currently, just the Shark Splash skillmap. If any more skillmaps are created with a darker background where the alt text shouldn't be black, then we'd have to re-think something. 

So, the `strokeColor` solution should be fine for now and would be more efficient than the list look-up, and we would have to be okay with not having _all_ our skillmaps have alt text that will pass accessibility for color contrast.

srietkerk:
> > > > Probably outside the scope of this PR, but @kiki-lee mentioned this in the original issue: "since skillmap thumbnail images use our special text instead of actual image links".
> > > > I don't fully understand what that means, but I was wondering what the motivation was behind doing something custom here. Is that something we've looked into?
> > > 
> > > 
> > > i read that as referring to the src of the content in the skillmap being derived from markdown content that gets parsed into the graph / imgs / etc vs just having html inline or something to that effect - which is as expected, just means we have to fill in the rest of the blanks :)
> > 
> > 
> > Our code to add a level to the skillmap includes a tag for adding the thumbnail ("imageURL:") and there's no place to include alt text.
> > Screenshot by Dropbox Capture
> 
> Ah, got it. Maybe we can file a separate task to add that :) But not needed for this PR. Thanks for the clarification!

Including the alt-text for a skill map background image seems like it would be a really nice feature. I was looking into it a bit, but I do think it will be a bit more involved and can be separated into a new PR, thanks @thsparks! I will cut an issue.

jwunderl (approved):
LGTM!

abchatra:

srietkerk:
> What is the status of this PR?

I am going to do a pull from master into the branch to make sure that everything is good still, but I should be able to merge this.

skillmap/src/components/SkillGraphContainer.tsx
```diff
@@ -67,6 +73,14 @@ export class SkillGraphContainerImpl extends React.Component<SkillGraphContainer
             }
         }
 
+        if (backgroundColor === strokeColor) {
+            altTextColor = 'white';
+        }
+
+        if (maps.length > 1) {
+            backgroundAltText = `Background image for ${maps?.length} connected skillmaps`
```
jwunderl, line 81:
```suggestion
            backgroundAltText = lf("Background image for {0} connected skillmaps", maps?.length || 0)
```

and no lf below -- the first parameter to an lf needs to be a string with " or ', not a var~

skillmap/src/components/SkillGraphContainer.tsx
```diff
@@ -83,7 +97,7 @@ export class SkillGraphContainerImpl extends React.Component<SkillGraphContainer
                     </svg>
                 </MenuBar>
                 {backgroundImageUrl && <div className="skill-graph-background">
-                    <img src={backgroundImageUrl} alt={lf("Background Image")} onLoad={this.onImageLoad} />
+                    <img src={backgroundImageUrl} alt={lf(backgroundAltText)} onLoad={this.onImageLoad} style={{ color: altTextColor }} />
```
jwunderl, line 100:
```suggestion
                    <img src={backgroundImageUrl} alt={backgroundAltText} onLoad={this.onImageLoad} style={{ color: altTextColor }} />
```

skillmap/src/components/InfoPanel.tsx
```diff
@@ -83,12 +83,13 @@ export class InfoPanelImpl extends React.Component<InfoPanelProps> {
...
         const isActivity = node && !isRewardNode(node);
         const tags = isActivity && (node as MapActivity).tags || undefined;
+        const thumbnailAltText: string = lf("Thumbnail for {0} activity in skillmap {1}", !subtitle ? lf("starting") : title, !subtitle ? title : subtitle);
 
         const hasCloudSync = pxt.auth.hasIdentity();
         return <div className="info-panel" ref={this.handleRef}>
             <div className="info-panel-image">
                 {imageUrl
-                ? <img src={imageUrl} alt={lf("Preview of activity content")} />
+                ? <img src={imageUrl} alt={lf(thumbnailAltText)} />
```
jwunderl, line 92:
```suggestion
                ? <img src={imageUrl} alt={thumbnailAltText} />
```

jwunderl (approved):
LGTM, one more small lf to remove though. the build should be checking for these but I guess that is only implemented for webapp and cli localization, not skillmap

PR #9041, opened 2022-09-13
Fix memory leaks in compile webworker

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/3931

This fixes the memory leak in pxtworker that was causing us to crash the tab (and sometimes all of chrome) by running out of memory.

During a compile we annotate the TS ast with a bunch of information for our compiler. To support incremental compilation, we also store references to functions that we can call again when re-emitting the nodes. The closures for these functions were causing a reference to the Binary object to stick around, which in turn was referencing CompileOptions (which contains the entire source of the program) and CompileResult (which includes the text of the emitted `binary.js`). All together this amounted to many mbs of strings being allocated every time the compiler was invoked.

This PR removes the binary's references to CompileResult and CompileOptions.

I also ended up hoisting some of the recordAction functions because they were causing references to the binary and compile result to stick around. I'm not sure why that is, because they don't appear to be referenced inside the function bodies, but this fixes that issue regardless.

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -4295,12 +4295,6 @@ async function testSnippetsAsync(snippets: CodeSnippet[], re?: string, pyStrictS
             return;
         }
 
-        if (++codeSnippetCount % 40 == 0) {
-            // clean up cache every once in a while to speed up / clear up mem
-            // TODO probably this should be handled in the service host as it does become a perf concern
-            cleanService();
-        }
-
         const isPy = snippet.ext === "py";
         const inFiles: pxt.Map<string> = {
             [pxt.MAIN_TS]: isPy ? "" : snippet.code,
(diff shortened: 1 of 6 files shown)


jwunderl (approved):
LGTM, should be able to remove the hack I had to ignore this issue here as well for minor checkdocs benefits (/ stress test to make sure there wasn't something else hidden deeper) https://github.com/microsoft/pxt/blob/14adcbed50fbac2d28235bb6d7de6485c35fad63/cli/cli.ts#L4130-L4134

(I checked out the branch locally and am running checkdocs without it, will edit if something is still wrong but seems totally fine)

eanders-ms (approved):
:shipit:
nice!!

riknoll:
@jwunderl removed! and yeah, we'll see! I'd totally believe there is something else going on too.

mmoskal (approved):

abchatra:
🥳

PR #9042, opened 2022-09-13
Fix snippet for fixedinstance methods that use variables_get

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/4931

If a fixedinstance type has a variable shadow block set in its block definition, then we don't really want to reference one of the exported fixed instances.


jwunderl (approved):

thsparks (approved):
:shipit:

PR #9044, opened 2022-09-15
Upgrade to React 17, TypeScript 4.8.3

eanders-ms:
Hey engineers, how do you feel about upgrading to React 17? It's good to not fall too far behind, and I had a reason: I'm adding a new CRA app for multiplayer, and I wanted to use a feature introduced in 17 (useContext, useReducer), but because of a breaking change to jsx config in tsconfig.json, I had to upgrade all the projects. It was straightforward, just a few minor code changes were required.

I did a quick run through the webapp and skillmap, all appears to work as expected.

I will open companion PR in pxt-arcade to upgrade TypeScript there.


riknoll (approved):

jwunderl (approved):
Nice!

eanders-ms:
Quick note, when you grab this change, be sure to `npm install`!

PR #9045, opened 2022-09-15
Tutorial updates for user studies / redesign

jwunderl:
This pr is mostly the layout changes, aiming to get in for testing tomorrow. Listed out remaining changes for future pr at the below \\/

build since it's probably better to just play with it than look at screenshots
* editor https://arcade.makecode.com/app/6188de4950eebb0ea0c342708b800c1a6d179d86-50f7e5f8fc
* skillmap: https://arcade.makecode.com/app/6188de4950eebb0ea0c342708b800c1a6d179d86-50f7e5f8fc--skillmap

changes included (few more than I'd like, but a lot of them are interdependent in that you would lose functionality / be in a broken state if they weren't packed together. Or were just two line fixes in separate commits)
* layout -- no sim / tutorial panel split
* take up full left side down to bottom of page for vertical pane view
* sim always in bottom right / playable (button to pop out to full screen)
* toolbox label
* toolbox search (kiki and I think jacqueline request, easy to revert or separate from pr)
* share btn in tutorial view (skillmap header is separate so would need to add there if we want it included as well, but few questions on that - namely, should it use skillmap share dialog or send message to editor to prompt editor share?)
* few random bug fixes, tried to keep contained to their own small commits
* new (closer to old) step counter -- bubbles (max 7, centered on current step) with next / last
* next button / hint prompt at bottom of scrollable section

Changes not included in this pr but still aiming to get done for release:
* horizontal smaller sim w/ toggle, pulled off this pr as it wasn't directly related
* block discoverability (click on inline block to highlight toolbox) -- I bumped this to the end of the queue as it felt like the thing that is easiest to add in a followup patch / etc since it wouldn't affect 
* few small color tweaks -- off the top of my head the bottom toolbar buttons (undo redo zoom in zoom out), collapsible snippets styling

few screenshots, but again build above it best:

image
image

### DIFF
--- a/webapp/src/components/tutorial/TutorialStepCounter.tsx
+++ b/webapp/src/components/tutorial/TutorialStepCounter.tsx
@@ -45,7 +45,7 @@ export function TutorialStepCounter(props: TutorialStepCounterProps) {
             />
             {stepsToShow.map(stepNum => <Button
                 key={`step${stepNum}`}
-                className={`ui button empty circular label clear ${stepNum === currentStep ? "active" : ""}`}
+                className={`empty circle-button ${stepNum === currentStep ? "active" : ""}`}
                 onClick={handleSetStep(stepNum)}
                 aria-label={stepButtonLabelText(stepNum)}
                 title={stepButtonLabelText(stepNum)}
--- a/webapp/src/components/tutorial/TutorialStepCounter.tsx
+++ b/webapp/src/components/tutorial/TutorialStepCounter.tsx
@@ -31,8 +31,8 @@ export function TutorialStepCounter(props: TutorialStepCounterProps) {
 
     return <div className="tutorial-step-counter">
         <div className="tutorial-step-label">
-            <span className="tutorial-step-title">{title || lf("Step")}</span>
-            <span className="tutorial-step-number">{`${currentStep + 1}/${totalSteps}`}</span>
+            {title && <span className="tutorial-step-title">{title}</span>}
+            <span className="tutorial-step-number">{lf("Step {0} of {1}", currentStep + 1, totalSteps)}</span>
         </div>
         <div className="tutorial-step-bubbles">
             <Button
--- a/webapp/src/components/tutorial/TutorialStepCounter.tsx
+++ b/webapp/src/components/tutorial/TutorialStepCounter.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-
+import { Button } from "../../../../react-common/components/controls/Button";
 interface TutorialStepCounterProps {
     tutorialId: string;
     currentStep: number;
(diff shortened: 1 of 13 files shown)


kiki-lee:
Woohoo!  So glad to have block search back!
Also very excited for the new layouts.  

Question: Limit is 7 bubbles, but that's only a limit for what's showing at one time, yes? They'll scroll as users go through more levels, yes? Most of our tutorials are 10-15 levels. 

Thought: Could we prioritize the toolbox highlight if possible?  I think I'll need to put it in the intro video and I don't want to have to re-record that too many times.

unthinkmedia:
Could we get a slight fade on overflow to give an additional signal that
there are more dots?

[image: Screen Shot 2022-09-15 at 4.36.21 PM.png]

abchatra:
Great job @jwunderl

unthinkmedia:
Also if we feel it is necessary to have Project 3/8 then we should get rid
of the 3 on the dot, or vice versa.  Otherwise it is redundant and a bit
confusing.

jwunderl:
@unthinkmedia could you give me a quick example for how that would look or a figma? It should be simple but I'm having trouble visualizing it in a way that looks good~
@kiki-lee will start looking at that now then, should be fine. And yes, if you open up the first link an go to e.g. chase the pizza with 19 steps you can see an example -- if you're on step 1 you'll be left most bubble, then when you get past 4 you'll be centered until you get to the last few.

webapp/src/components/tutorial/TutorialStepCounter.tsx
```diff
@@ -11,23 +11,54 @@ interface TutorialStepCounterProps {
...
+                disabled={currentStep == 0}
+                className={`square-button ${currentStep == 0 ? "disabled" : ""}`}
+                leftIcon="icon left chevron"
+                onClick={handleSetStep(currentStep - 1)}
+                aria-label={backButtonLabel}
+                title={backButtonLabel}
+            />
+            {stepsToShow.map(stepNum => <Button
+                key={`step${stepNum}`}
+                className={`ui button empty circular label clear ${stepNum === currentStep ? "active" : ""}`}
```
riknoll, line 48: Can we avoid using semantic ui here?

riknoll (approved):
code lgtm

riknoll:
The new design looks a little squished on tutorials with short instructions:

image


Also, I think we should get rid of the transition on the step bubbles. It looks glitchy when moving to the next step because the previous step teleports instead of smoothly transitioning

abchatra:
@unthinkmedia for design input.

unthinkmedia:
I just talked through several options with the design team and we decided
to go with this one,

- Include a slight fade in left and right overflow
- Text is a small sub-heading that scrolls away, avoiding increasing the
height of the fixed row

https://www.figma.com/proto/XVstEjPUlNnk0FNz1l2yoe/MakeCode-Arcade-Tutorials?page-id=3106%3A11848&node-id=4540%3A37280&viewport=10670%2C-6363%2C0.95&scaling=min-zoom&starting-point-node-id=4540%3A37280&show-proto-sidebar=1

[image: image.png]

kiki-lee:
> 

Weren't we going to have the number showing in the bubble of the step that we're on?

Jaqster:
Woo-hoo!!!!  Amazing stuff Joey!  :-)

jwunderl:
need upto date link with changes in from comments, I'll see about adding a fade in / out to the bubbles in another pr 

https://arcade.makecode.com/app/2a28dcfaea26e5c766abf5368ba75b9b2dd64777-c06ef6fb0a 
866ba4ca-5965-40cb-b301-bd87ef4562bb


I'll merge in a bit after build goes through so it can be ready for tomorrow testing

(I did accidentally leave some test changes I was working on for the click->highlight so just... ignore the fact that if you click an inlinesnippet it fires an alert in that build)

PR #9049, opened 2022-09-16
Inline snippet click highlight

jwunderl:
add tutorial block discoverability feature -- when user clicks on an inlineblock snippet (the little colored bars), open the associated toolbox category. (Note that the namespace needs to exactly match, so it might need some post release cleanup for existing tutorials, but should be fine for the most part):

ezgif com-gif-maker (4)


thsparks (approved):
Looks really smooth!

PR #9051, opened 2022-09-18
Add multiplayer app, adjust package dependencies

eanders-ms:
#### Background
This change began as simply adding a placeholder multiplayer app to the repo. But I wanted to use a newer version of React. This led me down a bit of a rabbit hole to update React everywhere, committed in PR #9044. Updating React everywhere was necessary due to a breaking change in tsconfig.json around how "jsx" is configured for React 17, making it impossible to take a dependency on react-common without upgrading it as well, and upgrading react-common required root level changes to everything.

#### Multiplayer app stub
The main purpose of this PR is to add a stub multiplayer application, to be served at https://arcade.makecode.com/--multiplayer. This will be the endpoint for hosting and joining multiplayer games. This initial version is very basic, it's a simple app that supports authentication, and a "message of the day" as an example of how to change application state. It doesn't do multiplayer yet.

#### Other changes in this PR
While testing the multiplayer app, I was seeing runtime errors having to do with multiple instances of React and React DOM being loaded. I'd followed the example of skillmap as closely as possible, the builds seemed identical, but I wasn't able to resolve why this was happening in the multiplayer app but not in skillmap. It seemed to be stemming from the fact that we had react dependencies at both the root of the repo and within sub-projects like multiplayer, skillmap, and authcode. I eventually looked for another solution, which led me to the fairly extensive dependency adjustments in this PR.

#### Dependency adjustments
Projects like skillmap and multiplayer are built using the `react-scripts` utility, part of a project bootstrap tool known as `create-react-app`, or CRA. The `react-scripts` build system is super convenient, but also over-opinionated (in my opinion 😸) about what is valid to include in the project. `react-common` falls on the wrong side of create-react-app's judgement, and so workarounds are necessary. Skillmap had a hack for this, but it wasn't working for multiplayer. After wrestling with that for a while, I put a different hack in place, one that seems to work for all our CRA projects: skillmap, multiplayer, and authcode. What I did:
1. In the root package.json, made `react` and `react-dom` peer dependencies rather than direct dependencies. This change requires each sub-project to supply these dependencies (which the CRA apps already did).
2. Also updated the CRA projects with workaround to allow them to reference react-common directly, even though it lies outside of their project roots. This was accomplished using a combination of `react-app-rewired` and `react-app-alias-ex` utilities.
3. The webapp project was different. This is a legacy project with a complex build process. What I ended up doing was creating a package.json for it, moving all its dependencies there, then selectively moving one or two back to the root until the entire build was working again. This was a blunt force instrument and may need further refinement later. Some tweaks to browserify were also necessary so it could find webapp's dependencies in their new location.
4. I also checked in the root `package-lock.json`. If there is a reason this was excluded, I'd like to know it. package-lock eliminates a large uncertainty: it assures that all devs are working with the same version of every dependency, and that official builds ship with those same dependencies.

*Needless to say, when you sync this change, you will need to run `npm install`. I've updated the `postinstall` step to install sub-project dependencies, so you should need only run it in the repo root.*

#### Reflection on comingled changes
I would have preferred to separate this change from the addition of the multiplayer app, but as multiplayer is purely additive, I thought it low risk to include it, especially as it was the core test of these dependency adjustments.


eanders-ms:
Please note that it is fine to let this sit until after the hackathon, and even until after the next Arcade release.

abchatra:
Changes look good to me. I actually prefer taking this sooner than later. Better to ship as part of this release ensure as we will get enough eyes on this change.

abchatra (approved):

jwunderl (approved):

eanders-ms:
@abchatra are you ok with merging this now?

abchatra:
I am good with merging. Lets do it.

PR #9054, opened 2022-09-22
Fix Image Editor Top Bar Button Sizing

thsparks:
When a tutorial is open, there's a size threshold where the text in the image editor's top buttons overruns the buttons' bounds. This happens because the tutorial pane severely limits the horizontal space allowed for the popup, but we aren't in tablet mode yet when considering the whole available screen size. Regardless, the space allowed for the popup is too small.

To address this, I've allowed the popup to overlay the tutorial if the **container's** size is smaller than the tablet threshold (as opposed to the whole window). I also increased the weight on the buttons' section of the flexbox so they aren't quite so restricted on either side. Combined, these give us enough flexibility to still look okay until we get to tablet mode, at which point the buttons change into their smaller form and we're good to go. (The tutorial pane is also moved at this point, and still covered up by the popup.)

Sample Build: https://arcade.makecode.com/app/9c9c393a96d2837ab719d34ee832dbd65d923de9-9a2bb76c07--skillmap#racer

DemoOfFixFor4862EvenSmaller4

I considered trying to change the buttons to their image-only (tablet/mobile) form earlier, rather than modifying the popup overlay behavior, but that all happens in css using media screen size, and I didn't see a good way to change that without bringing all of that into javascript, which seemed less ideal, especially since this affects a fairly narrow range of screen sizes and hopefully won't be hit too often.

This fixes https://github.com/microsoft/pxt-arcade/issues/4862

### DIFF
--- a/webapp/src/blocks.tsx
+++ b/webapp/src/blocks.tsx
@@ -693,8 +693,8 @@ export class Editor extends toolboxeditor.ToolboxEditor {
     protected resizeFieldEditorView() {
         if (!window) return;
         const blocklyDiv = this.getBlocksEditorDiv();
-        if (blocklyDiv && this.parent.isTutorial() && !pxt.BrowserUtils.isTabletSize()) {
-            const containerRect = blocklyDiv.getBoundingClientRect();
+        const containerRect = blocklyDiv.getBoundingClientRect();
+        if (blocklyDiv && this.parent.isTutorial() && containerRect.width > pxt.BREAKPOINT_TABLET) {
             blocklyFieldView.setEditorBounds({
                 top: containerRect.top,
                 left: containerRect.left,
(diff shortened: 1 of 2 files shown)


webapp/src/blocks.tsx
```diff
@@ -693,8 +693,8 @@ export class Editor extends toolboxeditor.ToolboxEditor {
     protected resizeFieldEditorView() {
         if (!window) return;
         const blocklyDiv = this.getBlocksEditorDiv();
-        if (blocklyDiv && this.parent.isTutorial() && !pxt.BrowserUtils.isTabletSize()) {
-            const containerRect = blocklyDiv.getBoundingClientRect();
+        const containerRect = blocklyDiv.getBoundingClientRect();
```
jwunderl, line 696: It's *probably* safe, but right now theres a bad check on blocklyDiv that reads as unsafe (i.e. would have crashed here here but you have it in the if on the next line still). How about `blocklyDiv?.getBoundingClientRect()` here and check `containerRect` instead of `blocklyDiv` on the next line?

jwunderl (approved):
Lgtm, one minor change requested but it's barely a functional change / just to avoid confusion in future

jwunderl (approved):

PR #9055, opened 2022-09-23
fix clickable snippets when capitalization differs

jwunderl:
I don't think there was an issue for this? but richard noticed during testing, lots of the time we have different capitalization in md vs namespace names -- in particular with `Math` as it's capitalization is a bit weird.

e.g. 

```md
``||math:pick random||``
```

instead of 

```md
``||Math:pick random||``
```

this fixes that so it checks for namespaces ignoring case


srietkerk (approved):
LGTM!

PR #9065, opened 2022-09-24
Node 16 peer deps, and make TypeScript happy

eanders-ms:
Node 16 performs a stricter peer dependency check, and we were failing it after upgrading to React 17.

Changes in this PR:
* Upgrade react-redux
* Upgrade redux
* Upgrade react-modal
* Fixed up compiler errors that came in with the updated type definitions
* Run skillmap and authcode's `npm install` in pxt's `postinstall` step

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,8 +56,8 @@
   ],
   "main": "built/pxt.js",
   "engines": {
-    "node": ">=14.0.0",
-    "npm": ">=3.0.0"
+    "node": ">=16.0.0",
+    "npm": ">=8.0.0"
   },
   "dependencies": {
     "@blockly/keyboard-navigation": "^0.1.18",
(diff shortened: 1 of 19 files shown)


package.json
```diff
@@ -157,6 +157,7 @@
     "test:fmt": "gulp testfmt",
     "test:lang": "gulp testlang",
     "update": "gulp update",
-    "watch-streamer": "cd docs/static/streamer && tsc -t es6 --watch"
+    "watch-streamer": "cd docs/static/streamer && tsc -t es6 --watch",
+    "postinstall": "cd skillmap && npm install && cd .. && cd authcode && npm install && cd .."
```
jwunderl, line 161: If i recall correctly the `npm ci` was there as it sped up builds quite a bit - is that still the case / something we should consider?

jwunderl (approved):
Question on npm ci but otherwise lgtm

package.json
eanders-ms: Oh, makes sense. I'll put it back.

eanders-ms: Would be nice if `postinstall` had a ci-specific variant.

eanders-ms: @riknoll did `npm ci` for skillmap speed up the build? Wondering whether it's worth the tradeoff of having to manually install dependencies for skillmap, authcode, and soon multiplayer.

jwunderl: One thing you can do is just have a postinstall.js that checks prod vs normal and execs the correct installs / cis, and have postinstall run `node postinstall.js` (would just have to copy over the existing code from cli, we set some env variables for ci builds)

eanders-ms: I'd like to merge this, then improve postinstall as you suggest in a separate PR.

PR #9073, opened 2022-09-26
More share fixes

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/4991
Fixes https://github.com/microsoft/pxt-arcade/issues/4960
Fixes https://github.com/microsoft/pxt-arcade/issues/4964
Fixes https://github.com/microsoft/pxt-arcade/issues/4958

Changes include:

1. Clicking record now auto-focuses the sim
2. The GIF recorder now automatically times out after ten seconds
3. The share dialog buttons are now aligned to the bottom-right
4. The first share of a project is now always persistent, and subsequent shares have a checkbox asking if they want to update the existing link (defaults to checked)

For 4, the checkbox defaults to being checked. I can easily make it unchecked if that's preferred.


jwunderl (approved):

abchatra:
I think it is the right call keep it checked. We want the latest changes to be reflected in the same page.

PR #9075, opened 2022-09-27
Change Shape of Skillmap Learning Objective Tags to Differentiate from Buttons

thsparks:
Currently, the skillmap info panel has boxes that display a learning objectives, and they look very similar to the buttons lower down on the page, which can be confusing since the learning objective boxes are not interactive. This PR changes the shape of the learning objective containers so they're more visually distinct from the buttons.

Without this change:
image

With change:
image

Fixes https://github.com/microsoft/pxt-arcade/issues/4967

### DIFF
--- a/skillmap/src/styles/infopanel.css
+++ b/skillmap/src/styles/infopanel.css
@@ -104,6 +104,7 @@
     margin: 0 0.5rem 0.5rem 0;
     white-space: nowrap;
     user-select: none;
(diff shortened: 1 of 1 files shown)


jwunderl (approved):
LGTM! Maybe wait for one extra review just for confirmation ~

kiki-lee (approved):

abchatra:
:shipit:

PR #9076, opened 2022-09-27
inline snippets: for advanced categories open advanced before toggle

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/5042

(minor note that it only ever opens advanced / doesn't toggle it back closed, as closing advanced feels aggressive)

thanks for finding @thsparks 

handle-animation

### DIFF
--- a/webapp/src/marked.tsx
+++ b/webapp/src/marked.tsx
@@ -406,18 +406,17 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
                     inlineBlock.addEventListener("click", e => {
                         // need to filter out editors that are currently hidden as we leave toolboxes in dom
                         const editorSelector = `#maineditor > div:not([style*="display:none"]):not([style*="display: none"])`;
-                        const toolboxSelector = `${editorSelector} .blocklyTreeRow[data-ns="${ns}"]`;
-                        let toolboxRow = document.querySelector<HTMLDivElement>(toolboxSelector);
-                        if (toolboxRow) {
-                            toolboxRow.click();
-                        } else if (isAdvanced) {
-                            // toggle advanced open first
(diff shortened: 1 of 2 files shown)


riknoll (approved):

webapp/src/marked.tsx
```diff
@@ -406,8 +407,17 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
...
                         const toolboxSelector = `${editorSelector} .blocklyTreeRow[data-ns="${ns}"]`;
-                        const toolboxRow = document.querySelector<HTMLDivElement>(toolboxSelector);
-                        toolboxRow?.click();
+                        let toolboxRow = document.querySelector<HTMLDivElement>(toolboxSelector);
+                        if (toolboxRow) {
+                            toolboxRow.click();
+                        } else if (isAdvanced) {
+                            // toggle advanced open first
+                            const advancedSelector = `${editorSelector} .blocklyTreeRow[data-ns="advancedcollapsed"]`;
+                            const advancedRow = document.querySelector<HTMLDivElement>(advancedSelector);
```
thsparks, line 416: Any reason we can't look up and expand the advanced row earlier (before the initial toolboxrow look up) since we already have isAdvanced, then reduce the duplication around looking up and clicking on the toolbox row?

thsparks (approved):
One question in the comments but not a big deal. Fine with whichever you decide.

webapp/src/marked.tsx
jwunderl: You're totally right, I initially had it as toggling advanced, fixed that, and forgot to refactor. Thanks for pointing out

thsparks (approved):

PR #9077, opened 2022-09-27
don't show 'saving' text in sidepanel tutorial editortoolbar

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/5036


aznhassan (approved):
:shipit:

PR #9085, opened 2022-09-29
slight adjustment to tutorial hint colors, hint icon > key, title

jwunderl:
close https://github.com/microsoft/pxt-arcade/issues/5028

\+ decision from alex + kiki on tutorial hint behavior, changing lightbulb to key + a label for 'answer key':

image

PR #9088, opened 2022-09-29
[stable8.3] cherry pick hint updates

jwunderl:
cherry pick https://github.com/microsoft/pxt/pull/9085


abchatra:
Pause on this as this is new design and needs baking in /beta

jwunderl:
@abchatra I pulled out the label / icon change from here so now it's just the portion that the collapsible hints more accessible, so we should merge this for tomorrow~

abchatra (approved):

PR #9091, opened 2022-09-29
add titles / aria labels for hint & next/done buttons

jwunderl:
Not a regression but probably want to sneak this into the release build, I just noticed that the sidepanel tutorials next / hint buttons never got titles / aria labels for accessibility~


abchatra:
Yes lets port it.

PR #9096, opened 2022-09-30
autoplay off for ams videos

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/5093, looks like this optional parameter defaults to true?


abchatra (approved):

kiki-lee:
Thanks!  This turns out to be really bad for "hint" videos that are closed up in an accordion but play anyway 😆

abchatra (approved):

PR #9099, opened 2022-10-02
Initial multiplayer client

eanders-ms:
Adds a basic multiplayer client app skeleton, with gulp build support.

Please see the `multiplayer/README.md` for more information.


eanders-ms:
@riknoll The changes I made to the build process look more extensive than they really are. I did some refactoring so that skillmap, authcode, and multiplayer share a build framework. The duplication was getting a bit much. Hopefully it all "just works" as before.

multiplayer/src/components/HostGame.tsx
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    return (
+        <>
+            <div className="mt-2 h-max">
+                <div className="text-2xl font-bold">{"Hosting a Game"}</div>
+                {netMode === "init" && (
+                    <div className="flex flex-row gap-1 items-end">
+                        <Input
+                            label="Game URL or Share Code"
```
riknoll, line 40: seems like `lf()` is missing on most the localizable strings in this file

jwunderl (approved):

multiplayer/.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
+    "arrowParens": "avoid",
+    "semi": false,
```
jwunderl, line 3: nit: I am a little curious on this one, I would say most of our codebases uses semicolons / I personally find it easier to read; any particular reason to override prettier default?

riknoll: i too prefer semicolons

riknoll (approved):
LGTM, but missing a lot of localized string in join/host game files

multiplayer/.prettierrc
eanders-ms: I can go either way, so with-semicolons it is.  I wonder how many terabytes of semicolons exist in github's storage.

PR #9102, opened 2022-10-03
[DRAFT] Resizable sim / tutorial panel

jwunderl:
Dumping this as a draft for now as I just spent an hour or two friday getting it to a demoable state and I have other things I gotta work on, will return and finish before next release :)

https://arcade.makecode.com/app/b8dfbcc37519503e713131cd7cf4a0fa3be60a6b-1ca2b8259a

todos:

* design
    * [ ] check reflow layout of sim frame / tutorial frame content in and see 
    * [ ] make it prettier
    * [ ] make it discoverable
* [ ] also support horizontal layout, for testing I only implemented it for the desktop (/ widescreen) view
* [ ] think on min / max widths
* [ ] remove collapse button on desktop, make it so dragging all the way (< 1/2 or 1/3 min width) 'collapses' and show expand tab
* [ ] probably should actually pass the size up instead of just query selector / setting the style on the div but it works fine
* [ ] don't just use pageX, use width / dx 
* [ ] also draggable tutorial top bar for < tablet size
* [ ] (maybe throttle resize a for perf)
* [ ] do something when resizing adds or removes a scrollbar -- either allocate space for them ahead of time so it never flickers or disable scrollbar when cursor-resize class is set / resize is in progress


jwunderl:
This was fleshed out and reimplemented in https://github.com/microsoft/pxt/pull/9533

PR #9108, opened 2022-10-06
Initial music editor check-in

riknoll:
The music editor is finally in a check-in-able state, so here's a PR with all of the initial code. 

Songs are represented as a series of tracks, each of which has its own associated instrument or collection of drum sounds. When the song is encoded into a hex buffer, the entire instrument is encoded with each track so this should hopefully be robust to us changing instruments in the future. The format of the hex buffer encoding is in a comment in music.ts and there will be an additional PR in pxt-common-packages that consumes the buffers.

Additionally, songs are treated as a new type of asset. They show up in the assets tab just like images/animations/tilemaps and can be named/referenced the same way.

Unfortunately, because this PR adds files to common-docs/static, I can't give an upload target link.

The treble clef SVG came from [here](https://en.wikipedia.org/wiki/File:Treble_clef.svg). The file is in the public domain so I didn't add a copyright note.

Short-term todo:
* More/better drum sounds
* Select/move/copy/cut/paste notes
* Monaco field editor
* Design team pass

Long-term todo:
* UI to define your own instruments
* Full-screen mode that gives access to a base clef
* Loop portions of the song
* Pitchbend between notes


jwunderl (approved):
LGTM, I only glanced over `pxtlib/music.ts` as I don't really have enough audio context to know where to start on giving a real review there though; it all seems internally consistent at least :)

theme/music-editor/Track.less seems empty, do we want that / is it just a placeholder for potential future use?

eanders-ms (approved):
LGTM, I looked over the touch points with existing code -- nothing jumped out.

abchatra:
Awesome work @riknoll

riknoll:
@jwunderl ah whoops... i'll delete that file in the next pr

PR #9109, opened 2022-10-06
Remove `lookup` REST call, remove heartbeat

eanders-ms:
* `lookup` call is no longer needed with redis-based message routing
* `heartbeat` message is unnecessary with websockets.

### DIFF
--- a/multiplayer/src/services/gameClient.ts
+++ b/multiplayer/src/services/gameClient.ts
@@ -10,8 +10,9 @@ import {
     playerLeftAsync,
 } from "../epics";
 
-const GAME_HOST =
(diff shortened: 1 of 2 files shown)


multiplayer/src/services/gameClient.ts
```diff
@@ -131,27 +127,10 @@ class GameClient {
 
         const authToken = await authClient.authTokenAsync();
 
-        // TODO: Send real credentials
```
thsparks, line 134: Do we still need this `send real credentials` to do? Or is that done already?

thsparks (approved):

multiplayer/src/services/gameClient.ts
eanders-ms: Yes, we're sending actual credentials now.

PR #9113, opened 2022-10-11
Sim in arcade mp

jwunderl:
Add a sim to arcade mp, start on passing msg back and forth:
image

screen seems a bit slow currently, but that is likely being exacerbated by it being clogged up on my local network with multiple connections on one device to test it.

json-ifying the msg breaks the image data as it's a uint8array. I'm converting to a hex string for now, which takes up ~20kb, ideally we could get that portion sent directly as a unit8buffer to get it back down to ~10kb which would help with network a bit.

I also added in a path for propagating query parameters to the sim through the run page without having to do them one by one -- have it here just for the theming but will also likely end up using it for another sim change I need to finish up to make sure we can lay things out consistently~


pxtsim/simdriver.ts
```diff
@@ -320,8 +321,8 @@ namespace pxsim {
                 if (parentWindow) {
                     // if message comes from parent already, don't echo
                     if (source !== parentWindow) {
-                        const parentOrigin = this.options.parentOrigin || window.location.origin
-                        parentWindow.postMessage(msg, parentOrigin);
+                        // posting sim messages to parent frame; no origin restriction.
```
jwunderl, line 324: also this one's worth noting -- we hit this when we send messages from sim to parent frames, but we really don't care about blocking those as there is nothing 'secret' in them. (for our case, clearing this up gets rid of about 3k errors / minute & is required to pass around messages nicely on localhost.)

multiplayer/src/types/index.ts
```diff
@@ -86,6 +105,20 @@ export namespace Srv2Cli {
...
+        slot: number;
+        data: {
+            button: number;
+            state: "Pressed" | "Released" | "Held";
+        };
+    };
+
+    export type ScreenMessage = MessageBase & {
+        type: "screen";
+        data: any; // pxsim.RefBuffer
```
eanders-ms, line 119: Just a note to say I'm going to make all the network messages binary.

eanders-ms (approved):
LGTM! Made some small comments.

multiplayer/src/types/index.ts
```diff
@@ -46,6 +48,19 @@ export namespace Cli2Srv {
         type: "start-game";
     };
 
+    export type InputMessage = MessageBase & {
+        type: "input";
+        data: {
```
eanders-ms, line 53: `data` scoping seems unnecessary here.

thsparks (approved):
LGTM just a few nits & questions

multiplayer/src/types/index.ts
```diff
@@ -61,8 +76,12 @@ export namespace Cli2Srv {
         | HeartbeatMessage
         | ConnectMessage
         | StartGameMessage
+        | InputMessage
+        | ScreenMessage
         | ChatMessage
         | ReactionMessage;
+
+    export type SimMessage = ScreenMessage | InputMessage;
```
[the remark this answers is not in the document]
thsparks: Spoke offline, this is mostly used for organizational purposes to differentiate sim -> sim communication from webapp-level messages. Used in the gameclient sendSimMessage function [here](https://github.com/microsoft/pxt/pull/9113/files#diff-beb8babb4b8fcf308bf7764d85ec7e6026f4b2afc25d63099d283d20d7ea312cR288).

multiplayer/src/services/gameClient.ts
```diff
@@ -54,35 +80,10 @@ class GameClient {
...
-                            case "player-joined":
-                                return await this.recvPlayerJoinedMessageAsync(
-                                    msg
-                                );
-                            case "player-left":
-                                return await this.recvPlayerLeftMessageAsync(
-                                    msg
-                                );
-                        }
+                        if (msg.type === "joined") clearTimeout(rejectTimeout);
```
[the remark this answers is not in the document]
jwunderl: Mainly a personal preference in avoiding passing around the rejectTimeout / promise resolve handler which is only used for the one-off message -- I find it easier to keep track when promise handlers / timeouts are scoped as minimally as possible. Up to changing if there is disagreement though~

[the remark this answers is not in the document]
thsparks: Ah, got it. I don't feel strongly either way, so if you prefer this let's stick with it for now.

multiplayer/src/types/index.ts
jwunderl: discussed offline: I had left it as is to avoid requiring a backend change tied up with this, we'll need to version messages eventually & can swap over with that change.

PR #9114, opened 2022-10-12
Quick guide describing what isn't translated in markdown

ganicke:
Slammed together a quick guide as to what is left as untranslated in markdown. You can add to it anything else I missed.


Jaqster (approved):
LGTM!

PR #9115, opened 2022-10-12
Add Prefix to Tailwind CSS Classes

thsparks:
This helps avoid conflicts with existing css classes, like "hidden", which can break tailwind behavior.

For example, I noticed an issue where tailwind could not properly adjust if something was shown or hidden based on screen size. The tailwind css classes to make that happen are `hidden md:block`. According to [the docs](https://tailwindcss.com/docs/responsive-design#mobile-first), this should hide the item when the screen is smaller than md and show the item when the screen is md or larger. However, the behavior I saw was that the item was always hidden, regardless of screen size. This was happening because we already have a `hidden` css class in our common theme (see [here](https://github.com/microsoft/pxt/blob/master/theme/common.less#L860-L862)). That class prevented tailwind from being able to override the display value on md+ screen sizes.

We could do a one-off workaround for `hidden`, but that may be error-prone if devs aren't aware of it, and it's possible we may encounter similar issues for other classes later on, so adding a prefix seemed like a more robust solution. And better to do it now while it's still a manageable amount to change.

Ideally, this doesn't have any functional impact, though it's possible I missed the prefix in a few places in which case some layout may look off. We may need to just update those as we notice them, though I'm hoping there aren't many.

Tailwind docs on prefixes: https://tailwindcss.com/docs/configuration#prefix

### DIFF
--- a/multiplayer/src/components/ArcadeSimulator.tsx
+++ b/multiplayer/src/components/ArcadeSimulator.tsx
@@ -82,14 +82,14 @@ export default function Render() {
     const fullUrl = `${pxt.webConfig.runUrl}?${queryParameters.join("&")}`;
     return (
         /* eslint-disable @microsoft/sdl/react-iframe-missing-sandbox */
-        <div id="sim-container" className="grow mt-5">
+        <div id="sim-container" className="tw-grow mt-5">
             <iframe
                 id="sim-iframe"
                 ref={simIframeRef}
                 src={fullUrl}
                 allowFullScreen={true}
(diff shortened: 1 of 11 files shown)


eanders-ms (approved):
Glad there was a robust fix for this. Having to type `tw-` everywhere will be a little tedious, at least until we get used to it. Bit the prefix also adds clarity to where classes are coming from, so in that way it's nice.

jwunderl (approved):

PR #9117, opened 2022-10-12
Add Multiplayer Header Bar

thsparks:
Add a header bar for multiplayer. It's loosely based on the header bar implementation in skillmaps, though I found it tricky to reuse much of it since we're using tailwind css and skillmap is not.

Signed Out:
image

Signed In (MS):
image

Signed In (Google):
image

Smaller Screen:
image
image

Drop Downs:
image
image

Currently contains stubs for reporting abuse and getting help. Not sure how those will work, yet.

Since it's my first time working in react and with tailwind css, there will inevitably be some mistakes, antipatterns, etc... If you see any, please let me know!


jwunderl (approved):

multiplayer/src/App.tsx
```diff
@@ -10,11 +13,29 @@ import "./App.css";
 function App() {
     const { state } = useContext(AppStateContext);
     const { signedIn } = state;
+    const [showSignInModal, setShowSignInModal] = useState(false);
```
eanders-ms, line 16: I can see the argument for moving the sign in/out apparatus here, but it means passing props to every component where you want to surface sign in functionality. Skillmap handles this by setting a state variable indicating which modal to show, and an app-level modal component to render whichever modal should be visible. That's not a bad system.

Let's try to render header bar as `<HeaderBar />`.

multiplayer/src/components/HeaderBar.tsx
```diff
@@ -0,0 +1,147 @@
+/// <reference path="../../../localtypings/react.d.ts" />
+
+// TODO multiplayer : Reduce duplication with skillmap HeaderBar.tsx (may require enabling tailwind css there or removing it here?)
```
eanders-ms, line 3: Interesting point. The reuse of any component authored here requires adopting tailwind everywhere... or for us to stop using it and go the route of custom class encrustation (noooo...). Note to self: Set up an engineering discussion to decide this.

multiplayer/src/components/HeaderBar.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+import { Button } from "../../../react-common/components/controls/Button";
+import { MenuBar } from "../../../react-common/components/controls/MenuBar";
+import { MenuDropdown, MenuItem } from "../../../react-common/components/controls/MenuDropdown";
+
+interface HeaderBarProps {
+    showReportAbuse?: boolean;
```
eanders-ms, line 10: showReportAbuse unused prop?

eanders-ms: I think this will pop a modal dialog eventually. Another argument to manage modals the way skillmap does.

eanders-ms (commented):
Looks good overall. I think we need to manage modals differently. Comments on that inline.

eanders-ms (approved):
Changes look good. We can rework modals in another PR!

PR #9118, opened 2022-10-13
Move auth check. Add init state.

eanders-ms:
When served from staging, it looks like the auth check can complete before the app's first render. This change moves the auth check downstream, to a point where the app is initialized and ready to receive state changes.

I also added a "Loading..." state. This is shown at startup while the auth check is running and login status of the user is undetermined.

### DIFF
--- a/multiplayer/src/components/HostGame.tsx
+++ b/multiplayer/src/components/HostGame.tsx
@@ -44,7 +44,9 @@ export default function Render() {
     return (
         <>
             <div className="tw-mt-2 tw-h-max">
-                <div className="tw-text-2xl tw-font-bold">{"Hosting a Game"}</div>
+                <div className="tw-text-2xl tw-font-bold">
+                    {"Hosting a Game"}
+                </div>
                 {netMode === "init" && (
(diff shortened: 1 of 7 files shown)


multiplayer/src/components/HostGame.tsx
```diff
@@ -44,7 +44,9 @@ export default function Render() {
     return (
         <>
             <div className="tw-mt-2 tw-h-max">
-                <div className="tw-text-2xl tw-font-bold">{"Hosting a Game"}</div>
+                <div className="tw-text-2xl tw-font-bold">
```
eanders-ms, line 47: #prettier

multiplayer/src/components/JoinGame.tsx
```diff
@@ -30,7 +30,9 @@ export default function Render() {
     return (
         <>
             <div className="tw-mt-2">
-                <div className="tw-text-2xl tw-font-bold">{"Joining a Game"}</div>
+                <div className="tw-text-2xl tw-font-bold">
```
eanders-ms, line 33: #prettier

jwunderl (approved):

thsparks (approved):

PR #9119, opened 2022-10-13
Inline the metronome worker to fix CORS issue

riknoll:
The browser didn't like my hack to put the webworker into common-docs because the JS ended up being on the CDN domain instead of the makecode.com domain.

Since it's so short, just inlining it instead.


webapp/src/components/musicEditor/playback.ts
```diff
@@ -193,4 +193,42 @@ export function addPlaybackStopListener(listener: () => void) {
 
 export function removePlaybackStopListener(listener: () => void) {
     playbackStopListeners = playbackStopListeners.filter(l => listener !== l);
-}
\ No newline at end of file
+}
+
+const workerJS = `
+/**
+ * Adpated from https://github.com/cwilso/metronome/
```
abchatra, line 200: License request done?

abchatra (approved):

webapp/src/components/musicEditor/playback.ts
riknoll: I didn't actually use any of their code, more just the idea. I can do the license request anyway to be safe.

PR #9121, opened 2022-10-13
Rework Modals for Multiplayer

thsparks:
This is a follow up from comments in https://github.com/microsoft/pxt/pull/9117. The design is based loosely on the skillmap modal design.

We want to create a simple way to show modals without having to pass around a bunch of callbacks. To do that, I've created a `modal` field in the app state which defines what type of modal (if any) is visible. The `AppModal` component handles displaying the actual modal based on this field's value. To display a modal, we can call into the dispatcher and fire the ShowModal event, passing in the type of modal to show.

### DIFF
--- a/multiplayer/src/App.tsx
+++ b/multiplayer/src/App.tsx
@@ -28,8 +28,8 @@ function App() {
 
     return (
         <div className={`${pxt.appTarget.id}`}>
-            <HeaderBar />
             {loading && <Loading />}
+            {!loading && <HeaderBar />}
             {!loading && !signedIn && <SignInPage />}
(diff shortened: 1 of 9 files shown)


thsparks:
This is a follow up from comments in https://github.com/microsoft/pxt/pull/9117. The design is based loosely on the skillmap modal design.

We want to create a simple way to show modals without having to pass around a bunch of callbacks. To do that, I've created a modal field in the app state which defines what type of modal (if any) is visible. The AppModal component handles displaying the actual modal based on this field's value. To display a modal, we can call into the dispatcher and fire the ShowModal event, passing in the type of modal to show.

PR #9123, opened 2022-10-14
fix parsing for shared scripts query params / hashes

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/5134

The regex produced by this for arcade can be tested here: https://regex101.com/r/qk2y2e/1, it's this whole thing:

```
/^(?:(?:https:\/\/)?(?:makecode\.com|arcade\.makecode\.com)\/)?(?:v[0-9]+\/)?(?:(?:api\/oembed\?url=.*%2F([^&#]*)&.*)|(?:([a-z0-9-_]+)(?:[#?&].*)?))$/i
```

The link I shared above has that (with the global and multi line flags enabled to allow for multiple test strings at once)

few quick notes;
* oembed check was already broken so I 'fixed it' but don't have a url handy to try it / don't know how a user would really end up using that as a share code. (the broken bit in particular was that it was checking `oembed?` instead of `oembed\?` in the produced regex)
* the domains check on line 246 was redundant -- we've already checked listed domains and valid domains would not match the _asdfasdfasdf / credit card number check below and shouldn't be able to match the domain portion above anyways

### DIFF
--- a/pxtlib/emitter/cloud.ts
+++ b/pxtlib/emitter/cloud.ts
@@ -244,7 +244,7 @@ namespace pxt.Cloud {
         const domainCheck = `(?:(?:https:\/\/)?(?:${domains.join('|')})\/)`;
         const versionRefCheck = "(?:v[0-9]+\/)";
         const oembedCheck = "api\/oembed\\?url=.*%2F([^&#]*)&.*";
-        const sharePageCheck = "([a-z0-9\-_]+)(?:[#?&].*)?";
+        const sharePageCheck = "([a-z0-9\\-_]+)(?:[#?&].*)?";
         const scriptIdCheck = `^${domainCheck}?${versionRefCheck}?(?:(?:${oembedCheck})|(?:${sharePageCheck}))$`;
         const m = new RegExp(scriptIdCheck, 'i').exec(uri.trim());
         const scriptid = m?.[1] /** oembed res **/ || m?.[2] /** share page res **/;
--- a/pxtlib/emitter/cloud.ts
+++ b/pxtlib/emitter/cloud.ts
@@ -233,7 +233,7 @@ namespace pxt.Cloud {
 
     export function parseScriptId(uri: string): string {
         const target = pxt.appTarget;
-        if (!uri || !target.appTheme || !target.cloud || !target.cloud.sharing) return undefined;
+        if (!uri || !target.appTheme || !target.cloud?.sharing) return undefined;
 
         let domains = ["makecode.com"];
         if (target.appTheme.embedUrl)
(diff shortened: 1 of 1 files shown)


pxtlib/emitter/cloud.ts
```diff
@@ -233,24 +233,23 @@
...
-        if (!scriptid) return undefined;
-
-        if (scriptid[0] == "_" && scriptid.length == 13)
-            return scriptid;
-
-        if (scriptid.length == 23 && /^[0-9\-]+$/.test(scriptid) || scriptid.length == 24 && /^S[0-9\-]+$/.test(scriptid))
+        const domainCheck = `(?:(?:https:\/\/)?(?:${domains.join('|')})\/)`;
+        const versionRefCheck = "(?:v[0-9]+\/)";
+        const oembedCheck = "api\/oembed\\?url=.*%2F([^&#]*)&.*";
+        const sharePageCheck = "([a-z0-9\-_]+)(?:[#?&].*)?";
```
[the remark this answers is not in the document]
jwunderl: Huh, this is a good warning -- it doesn't actually change the behavior in this case (can see in regex101 above) but I'll fix it anyways

pxtlib/emitter/cloud.ts
```diff
@@ -233,24 +233,23 @@ namespace pxt.Cloud {
...
-
-        if (!scriptid) return undefined;
-
-        if (scriptid[0] == "_" && scriptid.length == 13)
-            return scriptid;
-
-        if (scriptid.length == 23 && /^[0-9\-]+$/.test(scriptid) || scriptid.length == 24 && /^S[0-9\-]+$/.test(scriptid))
+        const domainCheck = `(?:(?:https:\/\/)?(?:${domains.join('|')})\/)`;
+        const versionRefCheck = "(?:v[0-9]+\/)";
+        const oembedCheck = "api\/oembed\\?url=.*%2F([^&#]*)&.*";
```
thsparks, line 246: In the sample regex you provided, I think there's only one backslash after oembed. Here, there's two. Was that intentional?

thsparks (approved):

pxtlib/emitter/cloud.ts
thsparks: Oh looks like you mentioned this in your description, so I assume it was by design...

jwunderl: Yes, the intention is for the constructed regex to match ? literally -- this oembed url would be something like

```
arcade.makecode.com/api/oembed?url=https%2F%2Fgarbage%2F_asdfasdfasdf&otherparams
```
(share url encoded as a uri component) and we want to match the `_asdfasdfasdf`

PR #9127, opened 2022-10-19
[multiplayer client] binary message support

eanders-ms:
* Adds support for network messages in binary format.
* Switch a few performance-sensitive messages to binary:
    * Screen updates
    * User input


jwunderl (approved):

pxtlib/auth.ts
```diff
@@ -102,7 +102,6 @@ namespace pxt.auth {
         private state$: Readonly<State>;
 
         constructor() {
-            pxt.Util.assert(!_client);
```
eanders-ms, line 105: This assert was breaking dev server hot reload.

pxtsim/simdriver.ts
```diff
@@ -398,18 +398,8 @@ namespace pxsim {
...
-                // If using the production simulator on local serve, the domain might have been
-                // redirected by the CLI server. Also send to the production domain just in case
-                try {
-                    frame.contentWindow.postMessage(msg, `https://trg-${(pxt as any)?.appTarget?.id}.userpxt.io/---simulator`);
-                }
-                catch (e) {
-                    // Ignore exceptions if the target origin doesn't match
-                }
-            }
+            const origin = U.isLocalHostDev() ? "*" : frame.dataset["origin"];
```
eanders-ms, line 401: @jwunderl can we ship this?

jwunderl: Yes, it only affects our local host serve & messages are sim messages from the host frame which are non vital / nothing sensitive in them.

PR #9129, opened 2022-10-20
Multiplayer embed sim directly (no --run)

jwunderl:
Use pxtembed.js to embed the simulator directly. 

This embeds the sim frame directly in the page instead of embedding a simulator which embed the page.

few notable benefits:

* Easier to pass messages between / get direct access to simulatordriver
    * One thing I added but have not used yet is the `setSimulatorPaused` / `setSimulatorResumed` in ArcadeSimulator.tsx - I was going to use them to have the sim loaded but stalled to start with but Eric suggested to just compile at first and only run when the game has been started by the host.  
* Easier to deal with any embed css, /--run embeds tend to be a bit of a nightmare to get sized exactly how you want.
* I actually noticed visibly better perf during gameplay, in particular when multiple windows are open  -- it appears that the extra packing / unpacking for every screen message was taking a pretty hefty toll on perf and this gets rid of 1 set of that per message.
* pxtembed is defined here: https://github.com/microsoft/pxt/blob/multiplayerUsePxtEmbedJs/gulpfile.js#L100 
    * `pxt.runner`, `pxtc`, `pxsim`, `pxt.blocks`, `pxt.py`, `LZMA` are the big ones that may be useful if you find yourself needing them~

There is downside to this; pxtrunner doesn't use a worker to run compiles at the moment, so when the compile starts (that is, when the join code or share url has been successfully entered and you've joined a game session) it will block the ui thread while it goes through the compile -- it'll pop up the 'waiting for game' or the 'invite players' screen but the buttons won't work for a second. This was previously somewhat handled because the middle iframe was the one that was being 'blocked' and not the main multiplayer page. I'll take a look at getting it connected to pxtworker for the compile, but either way I think this experience is better - worst case, to start we can make the invite players dialog say something like "loading, please wait" with the share code at first and only show the buttons after the compile has finished.


jwunderl:
One extra note, right now we don't have a sim behind the 'start game modal' (as that usually gets created in `simulateAsync`) -- I'll add support for loading the sim before as well, I've done it before (e.g. https://github.com/microsoft/pxt-arcade/blob/master/docfiles/script.html#L233) just easier to separate out from this pr.

eanders-ms (approved):

PR #9133, opened 2022-10-21
download game metadata

eanders-ms:
When the client joins a game, fetch the game's metadata (title, description, thumbnail) and add it to the app state, where it's available to the UI.

### DIFF
--- a/multiplayer/src/services/gameClient.ts
+++ b/multiplayer/src/services/gameClient.ts
@@ -234,8 +234,9 @@ class GameClient {
         );
         const { gameMode, shareCode } = msg;
 
-        // eslint-disable-next-line @typescript-eslint/no-unused-expressions
-        (await setGameMetadataAsync(shareCode)) && (await setGameModeAsync(gameMode, msg.slot));
+        if (await setGameMetadataAsync(shareCode)) {
+            await setGameModeAsync(gameMode, msg.slot);
+        }
     }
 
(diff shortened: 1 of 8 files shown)


multiplayer/src/services/gameClient.ts
```diff
@@ -231,14 +232,15 @@ class GameClient {
         console.log(
             `Server said we're joined as "${msg.role}" in slot "${msg.slot}"`
         );
-        const gameMode = msg.gameMode;
+        const { gameMode, shareCode } = msg;
 
-        setGameModeAsync(gameMode, msg.slot);
+        // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+        (await setGameMetadataAsync(shareCode)) && (await setGameModeAsync(gameMode, msg.slot));
```
jwunderl, line 238: I guess this eslint rule is trying to push you towards doing a conditional here

```suggestion
        if (await setGameMetadataAsync(shareCode)) 
            await setGameModeAsync(gameMode, msg.slot);
```

which is maybe a little easier to parse? doesn't really matter at all

fwiw it has helped me remember to delete accidental one liners where I was exploring api with intellisense a few times (e.g. https://github.com/microsoft/pxt/pull/9132/commits/2f4f33768c9f568c8446f60cacccb57ad3260595)

jwunderl (approved):

PR #9135, opened 2022-10-21
use a symlink to link to docs for multiplayer/skillmap build

jwunderl:
a little bit nicer to not have to wait for all of our docs folder to get copied over (twice for everything in static) when serving multiplayer / skillmap, and also makes it so you can do local testing of skillmap files without constantly reserving. Also removes requirement to have `--docs .` when serving multiplayer

note the 'junction' param, normal symlinks for dirs get blocked in node because they are deemed a security risk (I believe?) and restrict to admin env only.  The important thing to note is that that it's restricted to local paths only, but I doubt that affects anyone? https://stackoverflow.com/questions/9042542

Probably worth getting someone to test this real quick on mac or linux (wsl) real quick? just make sure `pxt multiplayer` looks valid / has the right icons and maybe check that `localhost:3000/?nolocalhost=1#local:skillmap/intermediate-skillmap`, but it looks like skillmap is broken locally for me right now? throws a bunch of errors for me e.g. `Uncaught Error: Invalid hook call. ` with 
image
and also 
image

but this also occurs on master~

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -2068,7 +2068,6 @@ function buildReactAppAsync(app: string, parsed: commandParser.ParsedCommand, op
     };
     // local serve
     const appRoot = `node_modules/pxt-core/${app}`;
-    const docsPath = parsed.flags["docs"];
     return rimrafAsync(`${appRoot}/public/blb`, {})
         .then(() => rimrafAsync(`${appRoot}/build/assets`, {}))
         .then(() => rimrafAsync(`${appRoot}/public/docs`, {}))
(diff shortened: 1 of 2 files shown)


eanders-ms:
I have a fix incoming for `process not defined`, which just started happening in multiplayer. Maybe it will fix things for skillmap.

eanders-ms:
I'm not seeing build errors in skillmap.

Does this mean we'd have to run shells as admin?

jwunderl:
no, "junction" type is runable on normal cli, it is "dir" type that requires admin.

jwunderl:
This code was refactored away / pr no longer relevant

PR #9136, opened 2022-10-21
Fixes bug with loading help resources (side docs) on Python methods

kimprice:
Closes https://github.com/microsoft/pxt-arcade/issues/2149

### DIFF
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -832,7 +832,7 @@ namespace ts.pxtc.service {
                 type: v.infoType
             };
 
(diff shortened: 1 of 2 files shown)


riknoll (approved):

jwunderl (approved):

abchatra:
Looks good. Add bit more description about the root cause and the fix if possible here.

kimprice:
As we build the AST from the user's Python code we make several passes to gather all the information at each node. After the last pass we mark the node with all the information like the position, symbols, etc. Before this fix, we weren't marking the node at the last pass for method names. Then when trying to load help resources, there were no symbols to iterate through to find any corresponding help URL.

@riknoll Am I explaining this correctly?

PR #9137, opened 2022-10-21
Add Multiplayer Footer with Privacy and Terms of Use

thsparks:
The code also supports a copyright section, but we don't have one in the target theme so it doesn't appear here.

This is lifted pretty directly from https://github.com/microsoft/pxt/blob/master/webapp/src/projects.tsx#L209-L216 with some tailwind-ification. I did remove the language section, though, as I suspect that may go in the gear menu. We can add it back here later if needed. I also made the text black because I think the original color wouldn't pass accessibility checks.

Links:
Microsoft MakeCode: https://makecode.com/org
Terms of Use: https://go.microsoft.com/fwlink/?LinkID=206977
Privacy: https://go.microsoft.com/fwlink/?LinkId=521839

image

Fixes https://github.com/microsoft/pxt-arcade/issues/5158

### DIFF
--- a/multiplayer/src/App.tsx
+++ b/multiplayer/src/App.tsx
@@ -15,6 +15,7 @@ function App() {
     const { state } = useContext(AppStateContext);
     const { signedIn, appMode } = state;
     const { uiMode } = appMode;
+    const targetTheme = pxt?.appTarget?.appTheme;
 
     const loading = useMemo(() => uiMode === "init", [uiMode]);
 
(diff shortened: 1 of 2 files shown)


multiplayer/src/App.tsx
```diff
@@ -27,13 +28,22 @@ function App() {
...
             {loading && <Loading />}
             {!loading && <HeaderBar />}
             {!loading && !signedIn && <SignInPage />}
             {!loading && signedIn && <SignedInPage />}
             <AppModal />
             <Toast />
+            <div className="tw-flex-grow"/>
+
+            {/* Footer */}
+            {targetTheme?.organizationUrl || targetTheme?.organizationUrl || targetTheme?.privacyUrl || targetTheme?.copyrightText ? <div className="ui horizontal small divided link list tw-text-center" role="contentinfo">
```
eanders-ms, line 41: For the sake of code organization, can you put this in a `Footer` component?

eanders-ms (approved):

PR #9143, opened 2022-10-25
sendBuilt query parameter added to send simJSBuiltInfo to the kiosk

srietkerk:
Kiosk games take awhile to load because they are getting compiled each launch. To decrease load time for the kiosk, we want to just send the game build instead of compiling each time in the runner. The addition of the `sendBuilt` query parameter provides this functionality.

The other part of this change can be found in pxt-arcade in this PR:  https://github.com/microsoft/pxt-arcade/pull/5175


webapp/public/run.html
```diff
@@ -184,6 +185,25 @@
...
+                        dependencies: deps ? decodeURIComponent(deps[1]).split(",") : undefined,
+                        builtJsInfo: data,
+                        single: single,
+                        hideSimButtons: hideSimButtons,
+                        autofocus: autofocus,
+                        additionalQueryParameters: additionalSimParams
+                    };
+                    pxt.runner.simulateAsync(sims, options).then(function() {
+                                $(loading).remove();
+                            })
```
jwunderl, line 205:
```suggestion
                    pxt.runner.simulateAsync(sims, options).then(function() {
                        $(loading).remove();
                    })
```

pxtrunner/runner.ts
```diff
@@ -24,6 +24,7 @@ namespace pxt.runner {
         additionalQueryParameters?: string;
         debug?: boolean;
         mpRole?: "server" | "client";
+        sendBuilt?: boolean;
```
jwunderl, line 27:
```suggestion
```

these ones aren't used, change in run.html should be fine on it's own~

pxtrunner/runner.ts
```diff
@@ -437,6 +438,7 @@ namespace pxt.runner {
             autofocus: simOptions.autofocus,
             queryParameters: simOptions.additionalQueryParameters,
             mpRole: simOptions.mpRole,
+            sendBuilt: simOptions.sendBuilt,
```
jwunderl, line 441:
```suggestion
```

pxtsim/simdriver.ts
```diff
@@ -74,6 +74,7 @@ namespace pxsim {
         autofocus?: boolean;
         queryParameters?: string;
         mpRole?: "server" | "client";
+        sendBuilt?: boolean;
```
jwunderl, line 77:
```suggestion
```

webapp/public/run.html
```diff
@@ -184,6 +185,25 @@
                     // propagate message packets to parent frames
                     window.parent.postMessage(msg.data, "*");
                 }
+                else if (server && data.type === "kiosk") {
```
jwunderl, line 188:
```suggestion
                else if (server && data.type === "builtjs") {
```

 might be a bit more descriptive / findable?

jwunderl (approved):
LGTM, the changes outside run.html aren't necessary though / two minor nits in run.html~

srietkerk:
> LGTM, the changes outside run.html aren't necessary though / two minor nits in run.html~

I was getting in-editor and runtime errors when I didn't have the other locations were not defined. Are they something I can ignore?

pxtrunner/runner.ts
srietkerk: I was getting in-editor and runtime errors when I didn't have the other locations were not defined. Are they something I can ignore? I also figured I should add them since the edits you made recently for multiplayer also included these types of changes.

srietkerk: Oh, I think I understand why. Is it because the existence of the query parameter doesn't really effect the runtime of the simulator at all? It effects `simulateAsync` and the simOptions paramenter, but `builtJsInfo` is already included in that simOptions, so it doesn't need to be in the larger objects themselves. Is that about right?

jwunderl: the reason we shouldn't need them is because the sendBuilt portion is handled solely in run.html -- that is, the flow is something like

```
kiosk -> run.html -> pxtrunner ->  pxtcompiler (building compile info) and pxtsim (simdriver)
```

where kiosk embeds run.html, which calls into pxtrunner, which calls into pxtcompiler to build the game and pxtsim to create a simulator / run the game. You have to add more fields depending on how deep the flag has to effect, but in this case pxtrunner already gives back the buildjsinfo due to a change I had previously made for the arcade share page, so the changes only need to exist in run.html to send that game back up for kiosk to read, and no types have to be added for pxtrunner / pxtsim

srietkerk: Okay, yes, that makes sense. Thank you for the clarification!

jwunderl (approved):

PR #9144, opened 2022-10-25
Pass join code or share url as query param

eanders-ms:
Deep link to join a game:
`https://arcade.makecode.com/--multiplayer?join=<joinCode>`

Deep link to host a game:
`https://arcade.makecode.com/--multiplayer?host=<shareUrl or shareCode>`

### DIFF
--- a/multiplayer/src/services/gameClient.ts
+++ b/multiplayer/src/services/gameClient.ts
@@ -22,8 +22,10 @@ import {
     setGameMetadataAsync,
 } from "../epics";
 
-const GAME_HOST = "https://multiplayer.staging.pxt.io";
-//const GAME_HOST = "http://localhost:8082";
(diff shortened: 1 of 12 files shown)


multiplayer/src/services/gameClient.ts
```diff
@@ -22,8 +22,10 @@ import {
     setGameMetadataAsync,
 } from "../epics";
 
-const GAME_HOST = "https://multiplayer.staging.pxt.io";
-//const GAME_HOST = "http://localhost:8082";
+const GAME_HOST_PROD = "https://mp.makecode.com";
```
jwunderl, line 25: I think for this and the --multiplayer url we should probably toss them into pxt.webConfig, but that's unrelated to this pr.

jwunderl (approved):

PR #9146, opened 2022-10-26
add button to host a multiplayer game to share dialog

jwunderl:
like this:

hostmult

the button color was picked at random to make it visually stand out from other social opts, would be happy to swap it given a good color

note dev side our share url logic needs some work, it's tangential to this so I didn't do anything to address it but 2/3 envs are broken currently:

* staging: when link is created, it is properly stored in staging but the displayed link is makecode.com/blah and does not work
* localhost: it's created on staging (which is a bit unexpected personally / would prefer to make it on prod as staging share links cannot be distributed) but also displays prod url
* prod is fine

in addition, we don't include the staging domains in the domains check here: https://github.com/microsoft/pxt/blob/master/pxtlib/emitter/cloud.ts#L243 so staging urls get rejected wherever that is used, and multiplayer pulls share urls from prod only. For this reason the games you share on localhost / staging won't end up loading in the multiplayer app (but it hits the right url so should be fine / in theory should work fine on prod site.)

### DIFF
--- a/react-common/components/share/ShareInfo.tsx
+++ b/react-common/components/share/ShareInfo.tsx
@@ -109,6 +109,19 @@ export const ShareInfo = (props: ShareInfoProps) => {
         }
     };
 
+    const handleMultiplayerShareClick = async () => {
+        // TODO: This won't work on staging (parseScriptId domains check doesn't include staging urls)
+        // but those wouldn't load anyways (as staging multiplayer is currently fetching games from prod links)
+        const shareId = pxt.Cloud.parseScriptId(shareData.url);
+        if (!shareId)
+            return;
+
+        const domain = pxt.BrowserUtils.isLocalHostDev() ? "http://localhost:3000" : "";
+        const multiplayerHostUrl = `${domain}${pxt.webConfig.relprefix}multiplayer?host=${shareId}`;
+
+        window.open(multiplayerHostUrl, "_blank");
+    }
+
     const embedOptions = [{
         name: "code",
         label: lf("Code"),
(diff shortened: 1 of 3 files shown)


react-common/components/share/ShareInfo.tsx
```diff
@@ -109,6 +109,19 @@ export const ShareInfo = (props: ShareInfoProps) => {
         }
     };
 
+    const handleMultiplayerShareClick = async () => {
+        // TODO: This won't work on staging (parseScriptId domains check doesn't include staging urls)
```
thsparks, line 113:
```suggestion
        // TODO multiplayer: This won't work on staging (parseScriptId domains check doesn't include staging urls)
```

nit: just makes it easier to search for these later.

react-common/components/share/ShareInfo.tsx
```diff
@@ -264,6 +277,12 @@ export const ShareInfo = (props: ShareInfoProps) => {
                                         leftIcon={"icon share"}
                                         onClick={handleDeviceShareClick}
                                     />}
+                                    {pxt.appTarget?.appTheme?.multiplayerShareButton && <Button className="square-button multiplayer-host"
+                                        title={lf("Start a multiplayer lobby")}
```
thsparks, line 281: Would "Host a multiplayer session" maybe be clearer? Or perhaps "Play in multiplayer"?

thsparks (approved):
Would be good to get designer feedback on the button image & color, but code looks okay to me. Just some minor nits.

We are getting a little crowded on all the buttons. I wonder if we can remove a few...but outside the scope of this change.

eanders-ms (approved):
Nice!

abchatra (approved):

abchatra:
@Jaqster fyi.

react-common/components/share/ShareInfo.tsx
jwunderl: Hm, that's a good question, all three sound good to me / I'm sure I've seen each before in full games -- @unthinkmedia could I get your opinion on this?

jwunderl:
Gonna merge this as is so we have something in for testing, and fix order / color / icon in a subsequent pr~

PR #9147, opened 2022-10-26
fix lfs that the regex will miss

jwunderl:
our lf regex is a bit picky about spacing I guess, would slightly prefer to avoid touching the cli lf scraping code as I feel like that's been the same forever / might be dragons. Just pulling out a few strings so that prettier won't mangle them.

image

Should probably make it throw in this if https://github.com/microsoft/pxt/blob/fixAuthDialogLfs/gulpfile.js#L327? ~I guess would be a little annoying for local dev, would prefer not to do in this pr either way~ just made it so build will error if env == prod so it'll error in prs like lint but won't prevent you from doing local dev work


thsparks (approved):
Weird.

eanders-ms (approved):
🤔

PR #9149, opened 2022-10-26
multiplayer conditionally set GAME_HOST

jwunderl:
these values should probably be derived from webconfig / deploy config eventually but for now this feels like it should be fine to make it so all three environments can work with one commit / build of multiplayer app?


thsparks (approved):

eanders-ms (approved):
Works for me!

PR #9150, opened 2022-10-26
Update Multiplayer Lobby Appearance 

thsparks:
This changes both host and join lobbies so they are non-modal and closer to the final designs. The sim exists and is loaded in the background, just hidden.

image

image

There are a few things I did not include in this PR, because I think they'll work better as separate changes (and will take more time). Those are:
1. Creating and adding the full join link (https://github.com/microsoft/pxt-arcade/issues/5180).
2. QR Code (https://github.com/microsoft/pxt-arcade/issues/5181).
3. Left the "join" lobby text as-is. Designs have instructions on how to send emojis with the keyboard, but I omitted those because we don't have the keyboard controls yet. (https://github.com/microsoft/pxt-arcade/issues/5182)

Fixes https://github.com/microsoft/pxt-arcade/issues/5145
Fixes https://github.com/microsoft/pxt-arcade/issues/5146

### DIFF
--- a/multiplayer/src/components/HostLobby.tsx
+++ b/multiplayer/src/components/HostLobby.tsx
@@ -19,6 +19,7 @@ export default function Render() {
     };
 
     const handleCopyClick = () => {
+        pxt.tickEvent("mp.hostlobby.copyjoinlink");
         if (pxt.BrowserUtils.isIpcRenderer()) {
             if (inputRef.current) {
                 setCopySuccessful(
--- a/multiplayer/src/components/HostLobby.tsx
+++ b/multiplayer/src/components/HostLobby.tsx
@@ -1,10 +1,10 @@
 import { useContext, useRef, useState } from "react";
 import { Button } from "react-common/components/controls/Button";
 import { Input } from "react-common/components/controls/Input";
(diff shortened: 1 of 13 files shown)


thsparks:
@eanders-ms LMK if you'd like me to change anything about the "PresenceBar" to reduce conflicts with your work. I'm not tied to current implementation.

jwunderl (approved):
Nice! LGTM

multiplayer/src/components/HostLobby.tsx
```diff
@@ -0,0 +1,69 @@
...
+    const inviteString = lf("Invite anyone to join your game instantly. Just send them a link!");
+
+    const onStartGameClick = async () => {
+        pxt.tickEvent("mp.hostlobby.startgame");
+        dispatch(clearModal());
+        await startGameAsync();
+    };
+
+    const handleCopyClick = () => {
+        if (pxt.BrowserUtils.isIpcRenderer()) {
```
eanders-ms, line 22: add tick event for code copy?

eanders-ms (approved):
Looks good! Made one small suggestion.

PR #9153, opened 2022-10-27
Use Deep Links for Copying Multiplayer Join Code, Update Host Lobby Appearance 

thsparks:
This updates the host lobby appearance to match the latest designs more closely, and it will copy a deep link rather than the code itself when the copy button is pressed (on both the host lobby and the game page).

Also sneaking in a fix to make the sim use the full screen width when on a smaller screen.

image

image

### DIFF
--- a/multiplayer/src/components/HostLobby.tsx
+++ b/multiplayer/src/components/HostLobby.tsx
@@ -4,7 +4,8 @@ import { Input } from "react-common/components/controls/Input";
 import { startGameAsync } from "../epics";
 import { clearModal } from "../state/actions";
 import { AppStateContext } from "../state/AppStateContext";
-import JoinCodeLabel from "./JoinCodeLabel";
+import CopyButton from "./CopyButton";
+import Loading from "./Loading";
 import PresenceBar from "./PresenceBar";
 
 export default function Render() {
@@ -18,35 +19,31 @@ export default function Render() {
         await startGameAsync();
     };
 
-    const handleCopyClick = () => {
-        pxt.tickEvent("mp.hostlobby.copyjoinlink");
-        if (pxt.BrowserUtils.isIpcRenderer()) {
-            if (inputRef.current) {
-                setCopySuccessful(
-                    pxt.BrowserUtils.legacyCopyText(inputRef.current)
-                );
-            }
-        } else {
-            navigator.clipboard.writeText(joinLink);
-            setCopySuccessful(true);
-        }
-    };
+    const joinCode = state.gameState?.joinCode;
+    if (!joinCode) {
+        return <Loading />;
+    }
 
-    const handleCopyBlur = () => {
-        setCopySuccessful(false);
-    };
+    const displayJoinCode = joinCode?.slice(0, 3) + " " + joinCode?.slice(3);
 
     const joinLinkBaseUrl = "aka.ms/a9";
     const inviteString = lf("Go to {0} and enter code", joinLinkBaseUrl);
+    const fullJoinLink = `${joinLinkBaseUrl}?join=${joinCode}`; // TODO multiplayer : create full link
 
-    const joinLink = `${state.gameState?.joinCode}`; // TODO multiplayer : create full link
     return (
         <div className="tw-flex tw-flex-col tw-gap-1 tw-items-center tw-justify-between tw-bg-white tw-py-[3rem] tw-px-[7rem] tw-shadow-lg tw-rounded-lg">
             <div className="tw-mt-3 tw-text-lg tw-text-center tw-text-neutral-700">
                 {inviteString}
             </div>
-            <div className="tw-text-4xl tw-mt-4">
-                <JoinCodeLabel />
+            <div className="tw-text-4xl tw-mt-4 tw-flex tw-flex-row tw-items-center">
+                {displayJoinCode}
+                <div className="tw-ml-2 tw-text-[75%]">
+                    <CopyButton
+                        copyValue={joinCode}
+                        title={lf("Copy join link")}
+                        eventName="mp.hostlobby.copyjoinlink"
+                    />
+                </div>
             </div>
             <Button
                 className={"primary tw-mt-5 tw-mb-7"}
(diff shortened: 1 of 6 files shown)


eanders-ms:
We should not show a space in the code:

image

Typing a space in the middle will be seen as an invalid join code.

thsparks:
> We should not show a space in the code:
> 
> image
> 
> Typing a space in the middle will be seen as an invalid join code.

I updated the join code parsing to account for that. I think it's much easier to read with the space, personally...

multiplayer/src/components/HostLobby.tsx
```diff
@@ -1,65 +1,63 @@
...
+    const joinDeepLink = makeJoinLink(joinCode);
 
-    const joinLink = `${state.gameState?.joinCode}`; // TODO multiplayer : create full link
     return (
         <div className="tw-flex tw-flex-col tw-gap-1 tw-items-center tw-justify-between tw-bg-white tw-py-[3rem] tw-px-[7rem] tw-shadow-lg tw-rounded-lg">
             <div className="tw-mt-3 tw-text-lg tw-text-center tw-text-neutral-700">
-                {inviteString}
+                {inviteStringSegments[0]}
+                {
+                    <a
```
eanders-ms, line 39: Use `Link` component from react-common, rather than `a`.

multiplayer/src/util/index.ts
```diff
@@ -80,7 +80,7 @@ export function cleanupJoinCode(
     joinCode: string | undefined
 ): string | undefined {
     if (!joinCode) return undefined;
-    joinCode = joinCode.trim();
+    joinCode = joinCode.replaceAll(" ", "");
```
eanders-ms, line 83: Oh I see, this will remove the space. Maybe it's fine.

multiplayer/src/components/HostLobby.tsx
```diff
@@ -1,65 +1,63 @@
...
-                );
-            }
-        } else {
-            navigator.clipboard.writeText(joinLink);
-            setCopySuccessful(true);
-        }
-    };
+    const joinCode = state.gameState?.joinCode;
+    if (!joinCode) {
+        return <Loading />;
```
eanders-ms, line 22: Why show loading?

thsparks: It was just something so it wasn't empty, but this case shouldn't ever really happen. It can be null instead if you think that's better.

multiplayer/src/components/HostLobby.tsx
```diff
@@ -1,65 +1,63 @@
...
+    }
 
-    const handleCopyBlur = () => {
-        setCopySuccessful(false);
-    };
+    // To get a link in the middle of the invite string, we actually preserve the {0} and insert the link manually as an html element later.
+    const inviteString = lf("Go to {0} and enter code", "{0}");
+    const inviteStringSegments = inviteString.split("{0}");
```
eanders-ms, line 27: nice :)

multiplayer/src/util/index.ts
jwunderl: Should we maybe just move line 85 up and delete this / line 84? That should handle this just fine as it's already clearing out all non alphanumeric (which would fix if there's a weird nbsp or something)

jwunderl (approved):
LGTM besides using link component instead of anchor directly as eric mentioned~

multiplayer/src/components/HostLobby.tsx
eanders-ms: null is better, but not worth holding this PR

eanders-ms (approved):

multiplayer/src/components/HostLobby.tsx
```diff
@@ -1,65 +1,64 @@
...
+    // Insert a space to make the join code easier to read.
+    const displayJoinCode = joinCode?.slice(0, 3) + " " + joinCode?.slice(3);
+    const joinDeepLink = makeJoinLink(joinCode);
 
-    const joinLink = `${state.gameState?.joinCode}`; // TODO multiplayer : create full link
     return (
         <div className="tw-flex tw-flex-col tw-gap-1 tw-items-center tw-justify-between tw-bg-white tw-py-[3rem] tw-px-[7rem] tw-shadow-lg tw-rounded-lg">
             <div className="tw-mt-3 tw-text-lg tw-text-center tw-text-neutral-700">
-                {inviteString}
+                {inviteStringSegments[0]}
```
eanders-ms, line 38: Assuming the number of elements in inviteStringSegments feels brittle to me. And assuming the link is in the middle segment isn't a safe assumption. I think you will need to iterate over the segments, and replace the one matching "{0}" with the link.

thsparks: Hmm, currently we split on {0} so there's no actual segment containing it. The only scenario where there wouldn't be 2 segments is if it was at the beginning or the end, so would it be better if we simply checked startsWith and endsWith {0} and reacted accordingly?

thsparks: I guess if {0} ended up in multiple places that'd also mess things up but I don't expect that would happen as a result of localization.

jwunderl: split will leave an empty string at beginning / end if the the string starts with / ends with it, e.g. 
image
I guess if the translation doubled up it would be an issue, but that would be pretty weird / we do this in a few other locations already

eanders-ms: I think I would split on whitespace and find the {0}. You could optimize by rejoining the contiguous word sequences before and after it. Do this in a useMemo.

eanders-ms: But there a ways of solving it.

thsparks: What's the benefit? Feels like unnecessary complexity to me.

eanders-ms (approved):

multiplayer/src/components/HostLobby.tsx
thsparks: Leaving this as-is for now b/c time crunch, but we can follow up later to understand concerns and address as needed.

PR #9156, opened 2022-10-28
fix color order on presence bar, yellow -> orange to match sim

jwunderl:
A little pumpkiny, but swapping from yellow to orange to fix contrast issues & match the player 3 theme in arcade game engine (e.g. when you set player 3 health)

image

also fix order on presence bar to match player color order on sim and in game

### DIFF
--- a/multiplayer/src/index.css
+++ b/multiplayer/src/index.css
@@ -6,6 +6,6 @@
     --slot-0-color: 77, 77, 77;
     --slot-1-color: 255, 0, 0;
     --slot-2-color: 0, 56, 255;
-    --slot-3-color: 77, 166, 77;
-    --slot-4-color: 228, 204, 0;
+    --slot-3-color: 255, 154, 20;
+    --slot-4-color: 77, 166, 77;
 }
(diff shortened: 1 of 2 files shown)


multiplayer/src/index.css
```diff
@@ -6,6 +6,6 @@
     --slot-0-color: 77, 77, 77;
     --slot-1-color: 255, 0, 0;
     --slot-2-color: 0, 56, 255;
-    --slot-3-color: 77, 166, 77;
-    --slot-4-color: 228, 204, 0;
+    --slot-3-color: 255, 154, 20;
+    --slot-4-color: 77, 166, 77;
```
thsparks, line 10: Intentional to change slot 4 as well?

thsparks (approved):
❤️ 🎃

multiplayer/src/index.css
jwunderl: Yes, orange (previously yellow) is play 3 color and green is p4, so swapping green /orangeyellow

eanders-ms (approved):
I'll remove the headerbar bg color, which should reduce the pumpkin factor.

PR #9157, opened 2022-10-28
Add Beta Indicator to Multiplayer Pages

thsparks:
This adds a little "beta" tag to the home page, lobbies, and game page. Margins on the sim were a bit weird, so I just put it in the corner of the page in that scenario for now. If needed, we can revisit this later, but it may require css updates on the sim-side of things.

Also, I don't know the "right" way to position stuff in css/html so feel free to offer alternate suggestions.

Fixes https://github.com/microsoft/pxt-arcade/issues/5162

image

image

image

image

image

### DIFF
--- a/multiplayer/src/components/BetaTag.tsx
+++ b/multiplayer/src/components/BetaTag.tsx
@@ -1,8 +1,8 @@
 export default function Render() {
     return (
         <div className="tw-text-sm">
-            <span className="tw-bg-white tw-shadow-sm tw-text-primary-color tw-rounded-md tw-font-bold tw-px-1 tw-pt-[0.05rem] tw-pb-[0.15rem] tw-mr-1">
-                {lf("BETA")}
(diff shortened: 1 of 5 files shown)


eanders-ms:
Seeing this implemented, I don't love it. Especially that we lose simulator space:
image

Can you try moving it to the lower left? I think it might fit better in the footer.

multiplayer/src/components/BetaTag.tsx
```diff
@@ -0,0 +1,10 @@
+export default function Render() {
+    return (
+        <div className="tw-text-sm">
+            <span className="tw-bg-white tw-shadow-sm tw-text-primary-color tw-rounded-md tw-font-bold tw-px-1 tw-pt-[0.05rem] tw-pb-[0.15rem] tw-mr-1">
+                {lf("BETA")}
```
jwunderl, line 5: Could consider using
```css
    text-transform: uppercase;
```
css rule -- looks like `tw-uppercase`? https://tailwindcss.com/docs/text-transform

Crowdin can get confusing if you have multiple text entries with the same words but different capitalization and no context on why~

jwunderl (approved):
LGTM, one suggestion on translations for beta~

eanders-ms (approved):
Let's go with this for now, but want to circle back with design to place it somewhere that doesn't impact simulator size.

PR #9162, opened 2022-10-31
Don't allow signin dialog to be dismissed

eanders-ms:
Fixes https://github.com/microsoft/pxt-arcade/issues/5220

### DIFF
--- a/multiplayer/src/components/SignedOutPage.tsx
+++ b/multiplayer/src/components/SignedOutPage.tsx
@@ -1,0 +1,16 @@
+import { useEffect, useContext } from "react";
+import { showModal } from "../state/actions";
+import { AppStateContext } from "../state/AppStateContext";
+import { useAuthDialogMessages } from "../hooks/useAuthDialogMessages";
+
+export default function Render() {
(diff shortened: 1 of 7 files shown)


multiplayer/src/components/SignedOutPage.tsx
```diff
@@ -0,0 +1,16 @@
...
+export default function Render() {
+    const { dispatch } = useContext(AppStateContext);
+
+    const dialogMessages = useAuthDialogMessages();
+
+    useEffect(() => {
+        dispatch(showModal("sign-in", { dialogMessages }));
+    }, [dialogMessages]);
+
+    return <></>;
```
thsparks, line 15: Is this any different from returning null? (Just curious, no need to update or change anything).

thsparks (approved):

multiplayer/src/components/SignedOutPage.tsx
eanders-ms: It's effectively the same as returning null. Fragments themselves don't add anything to the DOM, so it's a no-op.

PR #9164, opened 2022-11-01
Added title html tag for the blocks playground page

kimprice:
Added the title so it displays on the browser tab instead of the url. Closes https://github.com/microsoft/pxt-arcade/issues/3613

image

### DIFF
--- a/docs/playground.html
+++ b/docs/playground.html
@@ -9,6 +9,7 @@
         href="https://microsoft.github.io/monaco-editor/node_modules/monaco-editor/min/vs/editor/editor.main.css">
     <link rel="stylesheet" href="/static/shared/iframeeditor.css">
     <link rel="stylesheet" href="/static/playground/playground.css">
+    <title>MakeCode Blocks Playground</title>
 </head>
 
 <body>


docs/playground.html
```diff
@@ -9,6 +9,7 @@
         href="https://microsoft.github.io/monaco-editor/node_modules/monaco-editor/min/vs/editor/editor.main.css">
     <link rel="stylesheet" href="/static/shared/iframeeditor.css">
     <link rel="stylesheet" href="/static/playground/playground.css">
+    <title>MakeCode Blocks Playground</title>
```
abchatra, line 12: It is better to call it "MakeCode Playground". This also supports testing "javascript\python" as well.

abchatra (approved):

docs/playground.html
kimprice: Yea that is what I was going to originally call it, but then I thought it should be consistent with the header title. Should we change both?

abchatra: Oh, I see. I will take back my comment. Looks like even though you can write "python" the focus of this tool is blocks, with even help pointing to "Defining blocks". Your change looks good.

abchatra (approved):

PR #9165, opened 2022-11-01
multiplayer propagate audio

jwunderl:
send and receive messages between pxt-c-p and pxt-backend, and add a little hook in sim sound that calls back on sound events. 

@riknoll if you see any events I need to propagate besides just mute channels & play buffer let me know, it seems fine as far as I can tell.

here's the bad test sound 'pad' I did: https://arcade.makecode.com/S05411-77245-32862-42860 (I should have asked richard for one of his cool sounds)

### DIFF
--- a/multiplayer/src/services/gameClient.ts
+++ b/multiplayer/src/services/gameClient.ts
@@ -27,7 +27,7 @@ import {
 const GAME_HOST_PROD = "https://mp.makecode.com";
 const GAME_HOST_STAGING = "https://multiplayer.staging.pxt.io";
 const GAME_HOST_LOCALHOST = "http://localhost:8082";
-const GAME_HOST_DEV = GAME_HOST_LOCALHOST;
+const GAME_HOST_DEV = GAME_HOST_STAGING;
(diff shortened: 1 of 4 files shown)


multiplayer/src/services/gameClient.ts
```diff
@@ -25,7 +27,7 @@ import {
 const GAME_HOST_PROD = "https://mp.makecode.com";
 const GAME_HOST_STAGING = "https://multiplayer.staging.pxt.io";
 const GAME_HOST_LOCALHOST = "http://localhost:8082";
-const GAME_HOST_DEV = GAME_HOST_STAGING;
+const GAME_HOST_DEV = GAME_HOST_LOCALHOST;
```
eanders-ms, line 30: Probably want to change this back to staging.

eanders-ms (approved):
LGTM! One minor comment.

riknoll:
that should be it! if we ever want to support microbit sound expressions, that would be something to consider but let's cross that bridge when we come to it...

riknoll (approved):

PR #9167, opened 2022-11-02
Added new asset tags/categories for localization

kimprice:
We updated arcade gallery asset tags (https://github.com/microsoft/pxt-arcade/pull/5219) so we need to localize them as well.

We still need to extract the strings, so I've filed an issue here: https://github.com/microsoft/pxt-arcade/issues/5197


srietkerk (approved):
LGTM

jwunderl (approved):

PR #9171, opened 2022-11-03
Add QR Code to Multiplayer Host Lobby

thsparks:
This adds a QR Code (containing the join deep link) to the host's lobby. I've brought in the [qrcode.react](https://www.npmjs.com/package/qrcode.react) package to generate the qr code itself.

image

image

Fixes https://github.com/microsoft/pxt-arcade/issues/5181


jwunderl (approved):
LGTM, but i think you need to update packagelock before checking in?

thsparks:
> LGTM, but i think you need to update packagelock before checking in?

Hmm you're probably right. Is that auto-generated, or do I need to update it manually?

jwunderl:
I believe it should update if you run `npm i` in multiplayer folder

eanders-ms (approved):
:shipit:

PR #9177, opened 2022-11-04
Fix Remix on Joined Games

thsparks:
We weren't setting the game ID in state when a player joined a game, so we couldn't show the remix game button. With this change, it should now get saved in the GameInfo. I've also consolidated gameId variables, since we had one directly in state and one in state.gameState.


multiplayer/src/components/ArcadeSimulator.tsx
```diff
@@ -9,7 +9,8 @@ let builtSimJsInfo: Promise<pxtc.BuiltSimJsInfo> | undefined;
 
 export default function Render() {
     const { state } = useContext(AppStateContext);
-    const { gameId, playerSlot, gameState, clientRole } = state;
+    const { playerSlot, gameState, clientRole } = state;
+    const gameId = state.gameState?.gameId;
```
eanders-ms, line 13:
```suggestion
    const { gameId } = gameState;
```
minor nit.

multiplayer/src/components/RemixGameButton.tsx
```diff
@@ -4,9 +4,10 @@ import { AppStateContext } from "../state/AppStateContext";
 
 export default function Render() {
     const { state } = useContext(AppStateContext);
+    const gameId = state.gameState?.gameId;
```
eanders-ms, line 7:
```suggestion
    const { gameId } = gameState;
```

eanders-ms (approved):
LGTM! Made a small suggestion.

jwunderl (approved):

multiplayer/src/components/RemixGameButton.tsx
thsparks: This doesn't work since gameState can be undefined.

multiplayer/src/components/ArcadeSimulator.tsx
thsparks: This doesn't work since gameState can be undefined.

thsparks:
Bypassing check-in requirements so we can get this into beta asap for testing...

PR #9180, opened 2022-11-04
Make it so that only one share action shows up in the completion dialog

riknoll:
As discussed in testing today, making it so that the multiplayer share button replaces the regular share button when enabled in the skillmap


eanders-ms (approved):

abchatra:
Possible to remove dynamic buttons if the showMultiplayerShare  is set?

riknoll:
@abchatra you can just change the authoring

jwunderl (approved):

PR #9181, opened 2022-11-04
Move multiplayer share button to previous step in the share dialog

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/5258

image

Opening as a draft PR because this is currently going through testing and I still need to update the icon. This moves the share call-to-action out of the social buttons and into the previous step in the dialog.

Here's an upload target (updated with multiplayer detection and confirmation dialog):

https://arcade.makecode.com/app/07c9cb2c8c08ac86ae6888a100b1c67518a30dd6-5cbe4d31c8


eanders-ms (approved):

jwunderl (approved):

thsparks (approved):

abchatra:
Will this be shown for every game? We don't want to everyone who shares a link to go to the multiplayer page yet.
We need lobby and viewer support etc before it is ready for stream.

riknoll:
@abchatra it now detects if the game is multiplayer or not. Keep in mind that this means that folks might get confused as to why the button isn't always present.
I'll update the build link

abchatra:
Thanks for adding the logic to detect.

>> Keep in mind that this means that folks might get confused as to why the button isn't always present.
 What is the point of sending folks to multiplayer if there is nothing to do there? It will add to confusion.

Question on design: Why is this part of this dialog and the not the next dialog after you share? There looks like 3 call of action here: 
1. Share Project (prominent link)
2. Update project thumbnail (link)
3. Host multiplayer game (link)
If I host multiplayer game, do I still need to click share to get the share link? 

I would think making it prominent link after you share is better for the flow.

riknoll:
@abchatra putting it after the share step did not test well. People only notice the copy link button and nothing else on that page.

Hosting a multiplayer game automatically does a share. This is the same as our skillmap behavior.

abchatra:
It is because it was not prominent. What if we make it prominent?

abchatra:
We have goals for share, and we shouldn't count automatic share. I worry putting these buttons will confuse users on the primary action. We need a meeting to discuss. Can you schedule one?

abchatra:
From data I see QR code is heavily used, so I find it hard to believe multiplayer button wont' be used if we make it prominent.

riknoll:
As discussed offline, I added a confirmation dialog. Here's a gif showing the new flow:

multi-share


The build link has also been updated

abchatra (approved):

PR #9183, opened 2022-11-04
Multiplayer use worker to run compiles, sim more consistent

jwunderl:
Compiles a bit faster, does not freeze ui. closes https://github.com/microsoft/pxt-arcade/issues/5257. No longer uses full pxtrunner, just pxtsim/pxtlib and worker

build: https://arcade.makecode.com/app/591bd0c0f9818e2bbd31c4b078b0f26ebae81665-457c2b8bf5--multiplayer (remember that join code button will no point you to the build so have to navigate in multiplayer)

`simHost.ts` is currently a bit... messy. A significant reason for this is that it is largely copied / massaged from `pxtrunner/runner.ts` and just a bit of `webapp/compiler.tsx`, with many portions of that being 5-7 years old with some dragons. I'd prefer to get this pr in close to how it currently looks so we can test it thoroughly and I can do a follow up later to clean it up more, probably at least ~20% ish can be cut as there are special cases or things not directly relevant that need a little more looking

couple notes:
* the worker msgs rely on a few types listed in pxtcompiler.d.ts so it's included for now, I'll see about moving them to the right spot in pxtlib in a future pr
* have to copy workers as webworkers must spawn from same domain (localhost:3232 + localhost:3000 does not work)


multiplayer/src/components/ArcadeSimulator.tsx
```diff
@@ -96,68 +100,70 @@ export default function Render() {
...
         }
-    }, [gameState]);
+    }, [gameMode]);
 
     useEffect(() => {
-        const codeReadyToCompile =
-            playerSlot! > 1 || (playerSlot == 1 && gameId);
-        if (codeReadyToCompile && gameState?.gameMode !== "playing") {
-            preloadSim().then(compileSimCode);
+        if (gameState?.gameMode !== "playing" && simContainerRef.current) {
```
eanders-ms, line 145:
```suggestion
        if (gameMode !== "playing" && simContainerRef.current) {
```
it's the smallest suggestion I could come up with

eanders-ms (approved):
Really glad someone understands this ancient system.

PR #9186, opened 2022-11-06
pause game on visibility change

eanders-ms:
When page loses visibility, the host will pause the game. Everyone sees a pause modal.

One improvement we can make later: Sending reactions would work fine on this screen if the presence bar were interactable. It is blocked by the modal in this design.

**Host's view**
image

**Guest's view**
image

Also added a few missing telemetry ticks: "mp.sendreaction", "mp.kickplayer", "mp.leavegame".

I added a TODO where the simulator needs to be paused/resumed. Making this a draft PR until that part is in.


multiplayer/src/epics/pauseGameAsync.ts
```diff
@@ -0,0 +1,18 @@
...
+import { showModal } from "../state/actions";
+
+export async function pauseGameAsync() {
+    const { clientRole } = state;
+
+    try {
+        pxt.tickEvent("mp.pausegame", { role: clientRole! });
+        if (clientRole === "host") {
+            await gameClient.pauseGameAsync();
+            // TODO: pause simulator
```
eanders-ms, line 12: @jwunderl what message should be sent here to pause the sim? And similarly in `resumeGameAsync.ts`.

jwunderl: Should be these guys https://github.com/microsoft/pxt/blob/master/multiplayer/src/components/ArcadeSimulator.tsx#L50-L56, I had left them in for testing but they're currently unused

jwunderl: And worth a note, should only really need to pause host, as the clients will just 'pause' by virtue of the screen not receiving updates / buttons doing nothing

jwunderl (approved):

multiplayer/src/epics/pauseGameAsync.ts
eanders-ms: And the simulator render in grayscale when paused, like we do in the editor? Is that just a css effect?

eanders-ms: Oh we can do this on on the sim container. Easy.

jwunderl: Right, if you want to match what it looks like in editor it should be whatever applies when you do simdriver().stop() - we don't really use a 'pause' state in the editor outside of debug mode so just need to dupe it here

eanders-ms:
It was killing me that reactions weren't available during pause, so I made it not use an actual modal, just show a message overlaid and grayscale the simulator.
image

PR #9188, opened 2022-11-07
pause all sims. tweak pause visuals

eanders-ms:
Pause guest sims as well as host, otherwise their controls are still active and it felt odd.

Also adjusted pause state visuals a bit:
* Leave some color rather than go pure grayscale.
* Add a blur affect to help the pause notification stand out from background detail, and really drive the message that the game isn't playable right now. Maybe it won't be performant on low end machines, needs testing.

image

### DIFF
--- a/multiplayer/src/components/ArcadeSimulator.tsx
+++ b/multiplayer/src/components/ArcadeSimulator.tsx
@@ -173,7 +173,7 @@ export default function Render() {
                 "tw-h-[calc(100vh-16rem)] tw-w-screen md:tw-w-[calc(100vw-6rem)]"
             }
             style={{
-                filter: gamePaused ? "grayscale(1)" : "none",
+                filter: gamePaused ? "grayscale(60%) blur(8px)" : "none",
                 transition: "filter .25s",
             }}
         />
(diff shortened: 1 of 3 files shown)


multiplayer/src/components/ArcadeSimulator.tsx
```diff
@@ -173,7 +173,7 @@ export default function Render() {
                 "tw-h-[calc(100vh-16rem)] tw-w-screen md:tw-w-[calc(100vw-6rem)]"
             }
             style={{
-                filter: gamePaused ? "grayscale(1)" : "none",
+                filter: gamePaused ? "grayscale(60%) blur(8px)" : "none",
```
jwunderl, line 176: Can we maybe try like half this? 8px is pretty intense, my eyes hurt a little looking at the screenshot

jwunderl (approved):

multiplayer/src/components/ArcadeSimulator.tsx
eanders-ms: brought it down to 3px. Maybe this is better?
image

eanders-ms: Actually, let's go with 1px.
image

eanders-ms: Probably not feasible, but going to the opposite extreme gives a nice effect. Here's a blur of 60px:
image

jwunderl: 1 and 3 both look fine to me,

PR #9193, opened 2022-11-07
Minor fixes to defining blocks documentation and playground samples

kimprice:
Fixed some typos, added missing 'export' keywords to block definitions, and made parameters optional for the expandable blocks sample.


riknoll (approved):
we should also remove all the examples that use the % argument syntax in the block definition

jwunderl (approved):

kimprice:
> we should also remove all the examples that use the % argument syntax in the block definition

@riknoll why should we remove those? And by % argument, do you mean % parameter?

kimprice:
I'll create a new issue for removing the examples with the % argument/parameter syntax.

PR #9195, opened 2022-11-07
Null check for the share tick

abchatra:


jwunderl (approved):

eanders-ms (approved):

react-common/components/share/ShareInfo.tsx
```diff
@@ -46,7 +46,7 @@ export const ShareInfo = (props: ShareInfoProps) => {
 
     React.useEffect(() => {
         if (isLoggedIn) {
-            pxt.tickEvent("share.open.loggedIn", { state: shareState, anonymous: isAnonymous.toString(), persistent: hasProjectBeenPersistentShared.toString() });
+            pxt.tickEvent("share.open.loggedIn", { state: shareState, anonymous: isAnonymous?.toString(), persistent: hasProjectBeenPersistentShared?.toString() });
```
jwunderl, line 49: Might make sense to just do (!!isAnonymous).toString so values are always true / false for data wrangling?

abchatra: I think it is ok as is. We get tristate information? appinsights it's all strings, so doesn't matter so much?

abchatra: I will fix later if you think otherwise.

PR #9201, opened 2022-11-08
Multiplayer show starter games

jwunderl:
close https://github.com/microsoft/pxt-arcade/issues/5275

can't make a meaningful build till https://github.com/microsoft/pxt-arcade/pull/5301 is in

One thing to note, I currently have it just put you immediately into the hosting game state when you click, do we want a confirmation dialog or anything? I think it's probably fine to merge in for now / for testing tomorrow without the dialog but not sure.

image

image

image

and truncates any long titles:
image

(something something it's just too perfect to not fit)


eanders-ms (approved):
Looks good. Let's merge it in without a confirmation dialog and see how it feels in testing.

PR #9206, opened 2022-11-08
multiplayer reactions keyboard ui updates from testing

jwunderl:
Feedback in testing this am was that reactions pane looked a little worse with all the keyboard hint, so took a stab at rearranging to hopefully look a little more inviting again.

* make the numbers on the emojis look a little 'like keys' as some weren't sure that users would understand the circled numbers:
image

* move from a text description under the reactions to the 'keyboard controls dialog'
image

took the same icon as the one on the reactions button to hopefully guide them to press that if they see it and haven't noticed the reactions yet.

@unthinkmedia let me know what you think~

### DIFF
--- a/multiplayer/src/components/KeyboardControlsInfo.tsx
+++ b/multiplayer/src/components/KeyboardControlsInfo.tsx
@@ -36,7 +36,7 @@ export default function Render() {
             })}
             <div className="keymap-row tw-flex tw-flex-row tw-items-center">
                 <div className="keymap-key tw-align-middle tw-min-w-[2.5rem] tw-text-center tw-pb-[2.1rem]">
-                    1-6
+                    1 - 6
                 </div>
                 <span className="keymap-name tw-font-semibold tw-text-center">
                     <ReactionsIcon />
--- a/multiplayer/src/components/KeyboardControlsInfo.tsx
+++ b/multiplayer/src/components/KeyboardControlsInfo.tsx
@@ -1,3 +1,5 @@
+import ReactionsIcon from "./icons/ReactionsIcon";
+
 export default function Render() {
     const keyData = {
         [lf("up")]: ["↑", "W"],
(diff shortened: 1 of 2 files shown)


multiplayer/src/components/KeyboardControlsInfo.tsx
```diff
@@ -32,6 +34,14 @@ export default function Render() {
                     </div>
                 );
             })}
+            <div className="keymap-row tw-flex tw-flex-row tw-items-center">
+                <div className="keymap-key tw-align-middle tw-min-w-[2.5rem] tw-text-center tw-pb-[2.1rem]">
+                    1-6
+                </div>
+                <span className="keymap-name tw-font-semibold tw-text-center">
+                    <ReactionsIcon className="hover:tw-scale-125 tw-ease-linear tw-duration-[50ms]" />
```
thsparks, line 42: So does this actually work if they click on it?

multiplayer/src/components/KeyboardControlsInfo.tsx
```diff
@@ -32,6 +34,14 @@ export default function Render() {
                     </div>
                 );
             })}
+            <div className="keymap-row tw-flex tw-flex-row tw-items-center">
+                <div className="keymap-key tw-align-middle tw-min-w-[2.5rem] tw-text-center tw-pb-[2.1rem]">
+                    1-6
```
thsparks, line 39: nit, either way is fine, but personally it'd have a slight preference for spaces around the dash (1 - 6).

eanders-ms (approved):

thsparks (approved):
LGTM just some minor comments

multiplayer/src/components/KeyboardControlsInfo.tsx
jwunderl: No, this one's just the icon that's used as a label for the button in `Reactions.tsx`, not the button itself (I did forget to save after removing the hover state though, so removed that~)

PR #9213, opened 2022-11-09
multiplayer make hosting a game always pull latest ver

jwunderl:
few small fixes:

* keep gh cache so we don't keep re-requesting extensions
* clean up compile path a bit
* cache the user project as the load project path ends up attempting to load it multiple times while making sure all deps are available.
* clear cache of user project so when someone tries to host the same persistent link twice in a round without refreshing we get the up to date game.

### DIFF
--- a/multiplayer/src/services/simHost.ts
+++ b/multiplayer/src/services/simHost.ts
@@ -1,7 +1,7 @@
 let _mainPkg: pxt.MainPackage;
 let mainPkg = (reset?: boolean) => {
     if (!_mainPkg || reset) {
-        _mainPkg = new pxt.MainPackage(new PkgHost());
+        _mainPkg = new pxt.MainPackage(_mainPkg?.host() || new PkgHost());
     }
     return _mainPkg;
 };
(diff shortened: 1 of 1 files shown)


multiplayer/src/services/simHost.ts
```diff
@@ -18,32 +18,25 @@ if (!pxt.react.getTilemapProject) {
...
+        // we want this to be cached only within the scope of a single call to loadpackageasync,
+        // as the file can be requested multiple times while loading.
+        mp.host().cacheStoreAsync(verspec, undefined!);
         mp._verspec = verspec;
-        downloadPackagePromise = mp
-            .host()
-            .downloadPackageAsync(mainPkg(), dependencies);
-        installPromise = mainPkg().installAllAsync();
+        await mp.host().downloadPackageAsync(mainPkg());
+        await mainPkg().installAllAsync();
```
jwunderl, line 35: technically we were running these at the same time before, but it shouldn't matter as installAllAsync recursively grabs deps so it was already waiting for the script text network request anyways -- with this that will be available immediately  when downloadPackageAsync gets it

eanders-ms (approved):

PR #9215, opened 2022-11-10
make simulator a bit more greedy for space, fix mobile reactions going off screen

jwunderl:
Easiest to see with a build: https://arcade.makecode.com/app/99710e6c974bf62478eff348d40682c1d5dca5cc-3935cb69d8--multiplayer

make the sim a bit more greedy for space to fill out the page a bit more:

image

vs

image

also should fix the issue where the reactions occasionally go off screen by about 1em on  mobile too (at least it fixes it for me >>)

(also a one line fix for compiler so that it doesn't throw off a spurious error message, a debug message was assuming that optional parameter existed)

### DIFF
--- a/multiplayer/src/components/GamePage.tsx
+++ b/multiplayer/src/components/GamePage.tsx
@@ -34,7 +34,7 @@ export default function Render(props: GamePageProps) {
                 className="tw-flex tw-flex-col tw-items-center tw-grow tw-pb-4"
                 style={
                     state.gameState?.gameMode !== "playing"
-                        ? { visibility: "hidden" }
+                        ? { display: "none" }
                         : undefined
                 }
             >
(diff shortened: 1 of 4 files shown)


multiplayer/src/components/GamePage.tsx
```diff
@@ -31,9 +31,12 @@ export default function Render(props: GamePageProps) {
                 </div>
             )}
             <div
-                className={`tw-flex tw-flex-col tw-items-center ${
-                    state.gameState?.gameMode === "playing" ? "" : "hidden"
-                }`}
+                className="tw-flex tw-flex-col tw-items-center tw-grow tw-pb-4"
+                style={
```
jwunderl, line 35: reason to move this to a style is because tailwind jit does not work with non-string literals like template strings

thsparks (approved):
LGTM!

PR #9219, opened 2022-11-10
don't duplicate persistent share links when importing

jwunderl:
When loading a persistent share link, check if we already have the header for it, and if so load it. Here's a build:

https://arcade.makecode.com/app/1542f0098ae8cd9614e77e373590f0623ff8129d-57379544fb

@eanders-ms question, when you first login to a browser do we sync all the headers? I tried to check if this would have trouble immediately following logging in or cache being cleared, but the project I was testing always ended up being loaded while the page reloaded so I couldn't test if there was a case where loading old projects could run into an issue or anything of the sort.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -5387,9 +5387,18 @@ async function importGithubProject(repoid: string, requireSignin?: boolean) {
 
 function loadHeaderBySharedId(id: string) {
     core.showLoading("loadingheader", lf("loading project..."));
+    let previousHeader: pxt.workspace.Header;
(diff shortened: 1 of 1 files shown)


abchatra (approved):

eanders-ms:
@jwunderl yes, when a user logs in we sync all cloud headers. It's asynchronous, of course, so there might be a bit of a race condition in the scenario where:
- headers aren't synced
- the user clicks "remix" from multiplayer
- the editor loads before the headers are synced.

abchatra:
Probably should wait for the headers to load then?

eanders-ms:
@jwunderl we fixed this a different way, with the no-cache header, correct? Can we close this?

PR #9220, opened 2022-11-10
end lobby if game fails to compile

jwunderl:
When a game fails to compile kick them back to the lobby with an error msg

here's a test game https://arcade.makecode.com/S53442-66845-63450-74981 and test build https://arcade.makecode.com/app/ec3bc1503ea06af8b35cea5989a15893d50ba80c-6817c2d7b1--multiplayer?host=S53442-66845-63450-74981

### DIFF
--- a/multiplayer/src/epics/notifyGameDisconnected.ts
+++ b/multiplayer/src/epics/notifyGameDisconnected.ts
@@ -41,7 +41,7 @@ export function notifyGameDisconnected(reason: GameOverReason | undefined) {
                         type: "warning",
                         text: lf("Game is full"),
                         timeoutMs: 5000,
-                        icon: "😤"
+                        icon: "😤",
                     })
                 );
             case "rejected":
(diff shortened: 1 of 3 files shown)


multiplayer/src/epics/notifyGameDisconnected.ts
```diff
@@ -63,6 +63,16 @@ export function notifyGameDisconnected(reason: GameOverReason | undefined) {
                         icon: "😟",
                     })
                 );
+            case "compile-failed":
+                pxt.tickEvent("mp.compilefailed");
+                return dispatch(
+                    showToast({
+                        type: "error",
+                        text: lf("Game failed to compile"),
```
thsparks, line 71: nit: "Unable to compile game" (or something to that effect) might be a slightly less user-blame-y phrasing.

thsparks (approved):
LGTM, just one small nit-pick.

eanders-ms (approved):

multiplayer/src/epics/notifyGameDisconnected.ts
eanders-ms: "compile" may not be a familiar term for many. How about "There was a problem loading this game"

jwunderl: Yeah, I was thinking about that as well, I find this message a little challenging to craft as the problem most likely does come from the game itself, so ideally we would point them in the direction of fixing that so they can progress (vs. just trying to host the same game again or something) -- if the message is ambiguous it might seem like a fault in multiplayer and then they might not end up being able to play or something. I think your wording is enough to drive people but this probably just needs testing to get the right wording?

PR #9231, opened 2022-11-14
Extension search fixes

jwunderl:
* fix https://github.com/microsoft/pxt-arcade/issues/5367
* fix https://github.com/microsoft/pxt-arcade/issues/5361
* also lowered a delay that was waiting 1 second after the dependency was added until it reloaded headers, I'm a little suspicious of it but 1 second is quite a while if it's not strictly necessary / fighting against a known race condition somehow

left a few todos spread around as I didn't want to go to poke around too much besides fixing the known regressions as this will probably be ported to arcade, microbit, maybe minecraft (I'm guessing content authors could be using the share script extensions to test at least?) ~

Here's a build to double check: https://arcade.makecode.com/app/5dc92e4c8357413a5065b97411a4157ffd9c89ec-0aa863d716

image

### DIFF
--- a/webapp/src/extensionsBrowser.tsx
+++ b/webapp/src/extensionsBrowser.tsx
@@ -428,19 +428,34 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
         }
     }
 
-    function extensionMetaCard(scr: ExtensionMeta & EmptyCard, baseClass: string, ind: number) {
+    function ExtensionMetaCard(props: {
+        extensionInfo: ExtensionMeta & EmptyCard,
+        className: string
+    }) {
+        const { className, extensionInfo } = props;
+        const {
+            description,
+            fullName,
+            imageUrl,
+            learnMoreUrl,
+            loading,
+            name,
+            repo,
+            type,
+        } = extensionInfo;
+
         return <ExtensionCard
-            key={classList(baseClass, ind + "", scr.loading && "loading")}
-            title={scr.name ?? `${ind}`}
-            description={scr.description}
-            imageUrl={scr.imageUrl}
-            extension={scr}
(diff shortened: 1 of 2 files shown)


abchatra:
Question, is the logic ported from old extension code?

jwunderl:
Maybe closer to call it inspired by, closest port is here with the logic for setting the dependency here https://github.com/microsoft/pxt/blob/extensionSearchFixes/webapp/src/scriptsearch.tsx#L253 which is the main portion on using `setDepAsync` 

I didn't look at how the cards were displayed in scriptsearch.tsx as the ui implementation is just different between the two, but it was pretty straightforward to extract script card info -> card contents -- I looked at arcade.makecode.com/v1.5 and it doesn't show the thumbnail as the image or the "we're not endorsing this you know" warning, but those feel like pretty good additions to me

image

jwunderl:
I swapped it over to make it so loading share scripts uses the same `addDepIfNoConflict` path as adding extensions, new build here: https://arcade.makecode.com/app/9fd9d109f9c8727d9b30d6f8f04be9dc6e5b89c3-1eac5059bc

webapp/src/extensionsBrowser.tsx
```diff
@@ -377,6 +428,21 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
         }
     }
 
+    function extensionMetaCard(scr: ExtensionMeta & EmptyCard, baseClass: string, ind: number) {
```
riknoll, line 431: nit: make this a functional reach component

riknoll (approved):

abchatra (approved):

abchatra:
Once merged please port to microbit and arcade stable channels.

PR #9233, opened 2022-11-16
Cherry-Pick Downloading Multiple Projects as Zip to Stable7.0

thsparks:
Cherrypick 3656df448d33176855bfba0935b422fbc5b9896b. I had to make a few changes so it could build on the older branch. Namely:
1. Use index instead of id to select headers.
2. Use sui.Modal instead of react-common Modal, since react-common Modal does not exist here.

This is needed so we can support the functionality in Microbit v4, which is what the old WinJS windows app uses. We're asking users to migrate to a new app, and this functionality allows them to transfer all of their projects over relatively easily.


abchatra (approved):

webapp/src/scriptmanager.tsx
```diff
@@ -274,6 +293,133 @@ export class ScriptManagerDialog extends data.Component<ScriptManagerDialogProps
...
+        pxt.tickEvent("scriptmanager.downloadZip", undefined, { interactiveConsent: true });
+
+        await loadZipAsync();
+
+        const { selected } = this.state;
+        const zipWriter = new zip.ZipWriter(new zip.Data64URIWriter("application/zip"));
+
+        const allHeaders = this.getSortedHeaders();
+        const selectedHeaders: pxt.workspace.Header[] = [];
+        Object.keys(selected).forEach((selectedIndex) =>  {
```
thsparks, line 306: Calling attention to this section, where I had to change the lookup logic since we don't have id's in this branch.

webapp/src/scriptmanager.tsx
```diff
@@ -451,6 +603,10 @@ export class ScriptManagerDialog extends data.Component<ScriptManagerDialogProps
                         </table>
                     </div>
                     : undefined}
+
+                {download && <sui.Modal header={lf("Preparing your zip file...")} isOpen={!!download} onClose={this.handleDownloadProgressClose}>
```
thsparks, line 607: Calling attention to this section, where I used sui.Modal instead of react-common Modal, since react-common Modal doesn't exist on the branch

webapp/src/scriptmanager.tsx
```diff
@@ -274,6 +293,133 @@ export class ScriptManagerDialog extends data.Component<ScriptManagerDialogProps
...
+
+        await loadZipAsync();
+
+        const { selected } = this.state;
+        const zipWriter = new zip.ZipWriter(new zip.Data64URIWriter("application/zip"));
+
+        const allHeaders = this.getSortedHeaders();
+        const selectedHeaders: pxt.workspace.Header[] = [];
+        Object.keys(selected).forEach((selectedIndex) =>  {
+            if(selected[selectedIndex]){
```
riknoll, line 307:
```suggestion
            if (selected[selectedIndex]) {
```

riknoll (approved):

PR #9238, opened 2022-11-17
Update Microbit App Deprecation Notifications

thsparks:
This updates the top-bar notification and the no-text-editor modal that both appear in the WinJS app to signal deprecation. Now, they both point to the new windows app. The documentation link hasn't changed, but we'll probably want to update the docs to explain the new bulk download/upload options.

I wanted to be able to inject JSX into an LF statement, but the "jsxLF" function doesn't exist in stable7.0 so I just copied it in. Hopefully that's alright.

image

image


abchatra:
Its ok to copy as long it works.

abchatra (approved):

PR #9239, opened 2022-11-18
Fix Deprecation Notification Links for Microbit WinJS App

thsparks:
These links worked in the browser but not in the web app, and my hypothesis is that this is because they were using window.open instead of the link's href. (I looked at "Support" in the help menu, which works, and that used href). Unfortunately, I can't test inside the app locally, so this is a best guess. Validated in the browser, but hard to say if it will be 1:1.

I've also updated the link to go directly to the windows store rather than the web version.


thsparks:
Bypassing check-in gates so we can test/iterate more quickly

PR #9242, opened 2022-11-18
Update defining blocks page and playground samples

### LINKED ISSUE #9175: Defining blocks documentation not rendering code snippet as blocks
Go to https://makecode.com/defining-blocks
The first code section should be in blocks, but instead it appears blank. The code text can be seen upon highlighting.

image
image

Browser: Edge, Chrome

### LINKED ISSUE #9211: Update Playground samples and defining blocks documentation to use '$' instead of '%' for parameters
All of our new block definitions follow the '$' syntax, which allow you to put the parameters in any order on the block. The '%' forced you to have the parameters in the same order and caused breaking changes to the block definition when people changed parameter names. 

We should also replace any inline parameter definitions.

For example:
```
//% blockId="read_until" block="read until %del=delimiter_conv"
export function readUntil(del: string) : string {
    ...
}
```
would be rewritten as
```
//% blockId="read_until" block="read until $del"
//% del.shadow=delimiter_conv
export function readUntil(del: string) : string {
    ...
}
```

kimprice:
- Replaced all '%' parameter annotations with '$' (closes #9211)
- Removed 'eg:' from JSDoc comments in favor of setting the default with PARAM.defl. The eg notation automatically sets the first value listed as the default value for the parameter. This indirect way of setting the default may cause confusion, so we are no longer advertising it.
- Added a sample for fixed instances of a class
- Added a sample for default array values
- Updated the classes sample to include an example with blockCombine
- Removed all the information on Blockly mutators. The mutator blocks do not fit with the look and feel of our other blocks, so we no longer want to advertise it.

Also closes #9175 because the corresponding section was removed as unnecessary.


abchatra:
Nice set up changes.

docs/defining-blocks.md
```diff
@@ -105,16 +99,16 @@ will be organized to create the block.
 ```
 block = field, { '|' field }
 field := string
-    | string `%` parameter [ `=` type ]
+    | string `$` parameter [ `=` type ]
```
riknoll, line 102: should probably get rid of the `= type` bit since `param.shadow` is preferred

docs/defining-blocks.md
```diff
@@ -285,12 +292,14 @@ enum AlarmSound {
...
-     * @param sound of the alarm to play, eg: AlarmSound.Annoy
-     * @param duration of the alarm sound, eg: 2000
+     * @param sound of the alarm to play
+     * @param duration of the alarm sound
      */
-    //% block="play an alarm sound || of %sound for %duration ms"
+    //% block="play an alarm sound || of $sound for $duration ms"
     //% duration.shadow=timePicker
     //% expandableArgumentMode="toggle"
+    //% sound.defl=AlarmSound.Annoy
```
riknoll, line 301: you might need to put quotes around `AlarmSound.Annoy` here... I'm not actually sure. The `.` might confuse our parser

riknoll (approved):

jwunderl (approved):

docs/defining-blocks.md
kimprice: I checked it in the playground and it works fine.

PR #9245, opened 2022-11-22
fix little scrollbar that shows up in headless sim

jwunderl:
fix little sim scrollbar that is showing up when running minecraft with latest pxt due to sim -- this came from a combo of my arcade tutorial layout change + this special casing headless layout for minecraft~

image


srietkerk (approved):
LGTM

PR #9249, opened 2022-11-23
fix nested repos direct add

jwunderl:
In https://github.com/microsoft/pxt/pull/9231 I forgot to account for 'nested' repos in the add logic so this is a quick fix for that

Just noticed this, doesn't matter too much as the only thing currently using nested repos is jacdac and that only encourages people to add the base package but good to fix early. If you search `microsoft/pxt-jacdac/accelerometer` and click after that patch it would just add `microsoft/pxt-jacdac`, where it was working before as the package names should be different (though have to double check that, it's very possible a few of the many subpackages had a typo or something.)


abchatra (approved):

abchatra:
Needs to port to microbit & arcade?

jwunderl:
Probably should yes, as it'll be a little weird even if that's not how jacdac is normally used (as it directly adds the extensions within the sim) and the other one is still waiting to be released anyways

PR #9254, opened 2022-12-01
Auto rebuild css in pxt serve

jwunderl:
in pxt serve also watch for css changes and rebuild when they occur. little noisey in the serve window if you start making changes, but that's already super noisy if you're e.g. making common packages changers so doesn't feel like a big deal.

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -1394,14 +1394,16 @@ export function ghpPushAsync(builtPackaged: string, minify = false) {
         .then(() => ghpGitAsync("push"))
 }
 
-function maxMTimeAsync(dirs: string[]) {
+async function maxMTimeAsync(dirs: string[], maxDepth: number) {
     let max = 0
-    return U.promiseMapAll(dirs, dn => readDirAsync(dn)
-        .then(files => U.promiseMapAll(files, fn => statAsync(path.join(dn, fn))
-            .then(st => {
-                max = Math.max(st.mtime.getTime(), max)
-            }))))
-        .then(() => max)
+    await U.promiseMapAll(dirs, async dir => {
+        const files = nodeutil.allFiles(dir, { allowMissing: true, maxDepth });
+        await U.promiseMapAll(files, async file => {
+            const st = await statAsync(file);
+            max = Math.max(st.mtime.getTime(), max);
+        });
+    });
+    return max;
 }
 
 export interface BuildTargetOptions {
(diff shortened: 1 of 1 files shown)


cli/cli.ts
```diff
@@ -2611,7 +2616,11 @@ function buildAndWatchTargetAsync(includeSourceMaps: boolean, rebundle: boolean)
         simDirectories = simDirectories.filter(fn => fs.existsSync(fn));
     }
 
-    return buildAndWatchAsync(() => buildCommonSimAsync()
+    const lessFiles = lessFilePaths().filter(p => fs.existsSync(p));
+    // css build already occurs midway through internalBuildTargetAsync, so skip first rerun
+    let skipFirstCssBuild = true;
+
+    const buildTarget = () => buildCommonSimAsync()
```
riknoll, line 2624: async/awaitify?

riknoll (approved):
finally!

cli/cli.ts
```diff
@@ -2611,17 +2617,39 @@ function buildAndWatchTargetAsync(includeSourceMaps: boolean, rebundle: boolean)
...
+            console.log("rebuilding css")
+            await buildSemanticUIAsync();
+            console.log("css build complete")
+        }
+        return lessFiles;
+    };
+
+    await buildAndWatchAsync(buildTarget, 1)
+    await buildAndWatchAsync(buildCss, 6);
 }
```
kimprice, line 2653: How do you determine the max depth needed?

thsparks (approved):

cli/cli.ts
jwunderl: The first one I left at 1 mainly to avoid having to dig into the existing watch logic too much - if you look at where `dirsToWatch` gets added to, there are lots of places where we aggregate, and it looks like at least a few had subdirectories with generated files (e.g. `libs/` folder for game library code can have locale data get saved as part of `pxt serve`, so if we rebuilt based off that there could be constant reloads) - I'll add a comment here

The secone one for the css watch was check me taking the deepest path we currently have (which ends up being something like a nested theme folder with a site.variables file in a few of its subdirs) and adding a few for spacing - this tends to be how all the maxDepths in here that aren't 1 tend to end up being set

kimprice: So it's possible if someone had a deeply nested css/less file that it wouldn't pick up the changes. I guess that's just something to keep in mind.

kimprice (approved):

cli/cli.ts
jwunderl: right -- as far as I've seen, the files that are actually currently used typically max out at around folder + 2 (e.g. `[react-styles]/control/blah.less`, `[theme]/site/elements/blah.variables`, so there's plenty of wiggle room. This issue is present for most of our watchers though, good to be aware of as you mentioned~

PR #9255, opened 2022-12-01
Update --asseteditor page to support all asset types

riknoll:
This PR rewrites the --asseteditor page to support all asset types.

The messaging protocol for editing/creating/duplicating assets now takes the entire project filesystem as an argument. This way, all the logic for managing assets and emitting TS files can stay inside the iframe and the outside window doesn't need to take a direct dependency on pxtlib. As a bonus, this also means we can send all of the pxt_modules for a project so the galleries are actually populated now instead of empty.


webapp/src/assetEditor.tsx
```diff
@@ -19,179 +16,409 @@ function init() {
...
+}
+
+interface CreateAssetEditorRequest extends BaseAssetEditorRequest {
+    type: "create";
+    assetType: pxt.AssetType;
 }
 
-export interface Message {
-    data: MessageData;
+interface SaveAssetEditorReqeust extends BaseAssetEditorRequest {
```
jwunderl, line 40: SaveAssetEditorReq**ue**st
```suggestion
interface SaveAssetEditorRequest extends BaseAssetEditorRequest {
```

jwunderl (approved):

PR #9257, opened 2022-12-06
no more uwp

jwunderl:
Arcade seems to be fine with these changes, will test micro:bit tomorrow and see what else can be culled over there~ Cuts off the ~8.2mb .d.ts for the uwp api, which is fun

I added over the redirect too so we don't have to remember that whenever we do the next big micro:bit update, from https://github.com/microsoft/pxt/pull/8927


common-docs/windows-app.md
```diff
@@ -10,8 +10,8 @@ is an alternative to the web version for offline and portability purposes.
 
 #### Windows app deprecation
 
-MakeCode Windows app is being deprecated. Support for the Windows app
-will end after June 2022.
+MakeCode Windows app is deprecated. Support for the Windows app
```
abchatra, line 13: Please do a callout to the new PWA app. Otherwise it is confusing. We still have new app, for user it is one and the same.

abchatra (approved):

common-docs/windows-app.md
jwunderl: Do we have a docs page pointing at / describing that? I poked around for a second when updating this and didn't see one so I left it off

abchatra: Sounds good to remove base file

abchatra (approved):

Issue #9268, opened 2022-12-13
Color palette: add telemetry


kimprice:
Add telemetry so we can keep track of how many people use the color palette editor and how often it's used.

Issue #9263, opened 2022-12-08
Color palette; number and color spacing in dialogue 


clkantner:
When opening the color palette edit dialogue, the space between the number and color is a bit wide, and could benefit to make use of the space. 

<img width="723" alt="Screenshot 2022-12-08 at 2 09 00 PM" src="https://user-images.githubusercontent.com/106267801/206552905-f352f98d-f7f5-4bbd-9b5b-eb73c6c90b27.png">

Issue #9269, opened 2022-12-14
Color palette: update UI


kimprice:
Tasks:
- [ ] Add modal buttons (save and reset)
- [ ] Palette name input (either ask on save or at the top right)
- [ ] Add palette swatch for selected palette in the dropdown instead of just the palette name
- [ ] Add palette icon in top left corner within the modal title
- [ ] Add functionality/UI to delete custom palettes
- [ ] Make mobile friendly

The modal buttons are disabled when a built-in palette is selected and there are no color changes.
![image](https://user-images.githubusercontent.com/15070078/207705865-b913cc71-f22d-49a0-ba36-43f575bb4863.png)
Once the user edits the colors, the modal buttons are enabled, and they are able to add a name to their custom palette.
![image](https://user-images.githubusercontent.com/15070078/207705935-b47c712f-e80a-4db3-816d-33f82546e868.png)
Delete option in the dropdown for custom palettes:
![image](https://user-images.githubusercontent.com/15070078/207709624-db65f45c-323e-42a7-9852-9658a5e0cfdb.png)

PR #9270, opened 2022-12-14
Color palette editor: add reset and save modal buttons, add telemetry

kimprice:
This PR adds reset and save buttons to the color palette editor.  Reset reverts the colors to the previously saved colors. 

I also added telemetry for when the color palette editor is opened and when users leave the palette editor with saved changes  (closes https://github.com/microsoft/pxt/issues/9268). 

I made a minor UI change to fix the spacing between the number and the color picker (closes https://github.com/microsoft/pxt/issues/9263).

Refer to https://github.com/microsoft/pxt/issues/9269 and https://github.com/microsoft/pxt-arcade/issues/5327

(Disregard commits from before Dec 13 - these were captured in my first color palette PR and already merged)

You can try it out here: https://arcade.makecode.com/app/4e350193157b215a7d487850a4d7c9b44c06ab16-f2ddffd761

Buttons are disabled when no changes have been made:
image
Buttons are enabled once a change is made:
image
Exit modal if closing without saving:
image


srietkerk (commented):
The css/less variables in the only suggestion I have, everything else looks good!

theme/asset-editor.less
```diff
@@ -242,4 +242,18 @@
...
     }
+}
+
+.palette-transparent-button.common-button {
+    background: none transparent;
+    border: none;
+    &.disabled {
+        background: none transparent;
+        border: none;
+        color: gray;
```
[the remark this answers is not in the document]
kimprice: yea, I was wondering about this. Thanks!

theme/asset-editor.less
```diff
@@ -242,4 +242,18 @@
     i.fas {
         transform: rotate(90deg);
     }
+}
+
+.palette-transparent-button.common-button {
+    background: none transparent;
```
jwunderl, line 248: doesn't matter much but `background: none` is sufficient / clears out other all the other background-* sub properties for you so transparent shouldn't be needed~

```suggestion
    background: none;
```

jwunderl (approved):

theme/asset-editor.less
kimprice: thanks! I'll make that change in both spots.

PR #9272, opened 2022-12-30
Music editor design overhaul and keyboard navigation

riknoll:
This PR does the following for the music editor:

1. Overhauls the design and adds a bass clef
2. Makes it full screen by default just like the other field editors
3. Adds the ability to drag to select, move, and transpose notes
4. Adds copy/paste support
5. Adds keyboard navigation and editing
6. Improves rendering perf for long songs

The keyboard support is currently documented in `keyboardNavigation.ts`, but eventually there'll be a docs page and possibly some UI element that links out to it. Everything is supported via keyboard, but I have not added screen reader support yet. That'll be its own PR.

Apologies for the big PR! There were some rather fundamental changes that needed to be made to the way the coordinate system works and it ended up being easier to do everything at once.

The bass clef SVG came from wikimedia (it was in the public domain so no thirdpartynotice update).

Here's an upload target: https://arcade.makecode.com/app/3a70b8fda716e48f4196afa5be2893a78d6ea1bb-48ef05ea6a

Still to do:

1. Hide bass clef on smaller screens (or make it toggleable?)
2. Better drum sounds
3. Screen reader support for music workspace
4. Better support for enharmonic spelling (store the sharp/flat state with the note)


Jaqster:
Looks great!  Couple requests -
1.  Can you implement a expand/collapse toggle for the bass line?  Or maybe just provide another build version with just the treble line?  I have a hypothesis that 2 staff signatures/lines may be overwhelming - so wanted to test out 2 different versions in user testing.
2. Few small bugs I saw - Bass clef svg missing, Grid not updating, Looping toggle not updating on block, Erase doesn't seem to work

kimprice:
This is super cool!

I also found some bugs. Same as Jacqueline: bass clef svg missing, loop toggle not working, and erase doesn't work on click but it works on drag. Other bugs: if I have the same instrument starting at the same time on different notes, they automatically have the same holding time and if I drag one to change the holding time, both are changed. If I try to add drums to the bass, they show up in the treble. It's very hard to hear the drums preview when selecting it at the top.

I expected the play button to toggle as a stop or pause rather than playing again from the beginning. If there's time in the future, it would be nice for the play button to toggle as a pause and also to be able to select where to start the play from. I think these would be really important for anyone trying to create long songs. It would also be great if the view could follow the animation as it's playing.

eanders-ms:
Code changes are pretty extensive, but look fine upon cursory review.

Feedback on keyboard navigation:
* This is probably by design, but feels like a bug: when two different notes appear on the same vertical bar in the song, I can't traverse between them using alt + up/down. After some time trying to understand what's not working, I figured out I can only navigate notes that match the currently active note selected in the toolbar. When burger is active, I can only navigate burger notes in the song, for example. Furthermore, when a note is active that doesn't appear in the song, I can't navigate any of the notes and it feels stuck.
* I can select "erase" with keyboard navigation and it becomes the active cursor, but I have no way to control it.
* Navigating within the song track, an arrow-based directional cursor might work better here, paired with tab navigation to snap the cursor to the next note in the song. Mapping non-intuitive keyboard shortcuts to every possible action does not lead to a discoverable interface, and doesn't scale well.

jwunderl (commented):
Just adding a few notes on typos since others already said my other thoughts~ (the one in button.less is actually a meaningful typo / bug, others just nits) otherwise code changes lgtm

riknoll:
@eanders-ms the "one track at a time" paradigm is indeed by design. i get that it's confusing for sighted users who can see the tracks all at once, but for people with low/no vision i thought it might make more sense to go 1 track at a time. i'm willing to discuss options here, though!

and yeah, i'm not really sure what to do with erase in keyboard navigation

riknoll:
I fixed the erase bug! if it's alright with everyone, i'll merge this is as-is and we can talk about design/strategy for the keyboard interactions later

eanders-ms (approved):

PR #9273, opened 2023-01-03
Locales use "fr-FR" format, just "fr" doesn't work

johan-gorter:


johan-gorter:
The format seems to have changed back to  "fr" instead of "fr-FR" again now... If this is the desired format, this PR should not be merged.

Issue #9267, opened 2022-12-13
Color palette: add functionality to save/delete custom palettes


kimprice:
Currently, you can have one custom palette, but if you select a built-in palette from the dropdown, you will lose the custom palette.

We should support adding and deleting custom palettes.

PR #9277, opened 2023-01-05
Color palette editor: support for multiple custom palettes

kimprice:
Users can now create unlimited custom palettes, allowing them to more easily test out how their assets look with different palettes. Before, users were only able to have one custom palette at a time and if they swapped to one of the built-in palettes, they would lose their custom palette. Now, their custom palettes are stored in a hidden file and users can access them through the dropdown.

I also changed the 'save' button to 'apply' which now applies the changes and closes the modal instead of saving and keeping the modal open. With this change, there is no intermediary palette and therefore 'reset' always sets the palette back to what it was when the palette editor was opened.

When a user edits a built-in palette, it automatically becomes a custom palette and prompts the user to name the palette.

Try it out here: https://arcade.makecode.com/app/640a2320e33b39231cea8193bf4cb6ba72e56ace-e43fec8b3b

image
image
image

Still To Do:
- add functionality/UI for deleting custom palettes
- add scroll bar for dropdown (if you add many custom palettes, the dropdown becomes long and gets cut off)
- shorten appearance of long custom palette names within the dropdown (use '...' to indicate the name has been shortened). This will prevent the dropdown from becoming too wide and getting cut off.
- add some separation between the custom palettes and the built-in palettes within the dropdown (ex. horizontal line)
- add ability to rename custom palettes
- add a button to create a new custom palette, either within the dropdown or directly on the modal. Currently, custom palettes are created by making an edit to a built-in palette.

Refer to https://github.com/microsoft/pxt/issues/9267, https://github.com/microsoft/pxt/issues/9269, and https://github.com/microsoft/pxt-arcade/issues/5327

### DIFF
--- a/webapp/src/components/assetEditor/assetPalette.tsx
+++ b/webapp/src/components/assetEditor/assetPalette.tsx
@@ -3,8 +3,14 @@ import { useEffect, useRef, useState } from "react";
 import { Modal, ModalAction } from "../../../../react-common/components/controls/Modal";
 import { PalettePicker } from "../../../../react-common/components/palette/PalettePicker";
 import { PaletteEditor } from "../../../../react-common/components/palette/PaletteEditor";
-import { AllPalettes, Palette } from "../../../../react-common/components/palette/Palettes";
+import { AllPalettes as BuiltinPalettes, Palette } from "../../../../react-common/components/palette/Palettes";
+import { Input } from "../../../../react-common/components/controls/Input";
 
+export interface CustomPalettes {
+    nextPaletteID: number;
+    initialPalette: Palette; // could be custom or built-in
+    palettes: Palette[];
+}
 export interface AssetPaletteProps {
     onClose: (paletteChanged: boolean) => void;
 }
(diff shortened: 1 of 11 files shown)


webapp/src/components/assetEditor/assetPalette.tsx
```diff
@@ -1,139 +1,228 @@
...
+            if (colorSet1[i].toLowerCase() !== colorSet2[i].toLowerCase()) {
                 isEqual = false;
                 break;
             }
         }
         return isEqual;
     }
 
-    return renderPaletteModal();
+    const initiatePalettes = () => {
```
jwunderl, line 148: nit: probably `initializePalettes`? When I see initiate I would think of kicking off some process vs instantiating initial state or something

webapp/src/components/assetEditor/assetPalette.tsx
```diff
@@ -1,139 +1,228 @@
...
+}
 export interface AssetPaletteProps {
     onClose: (paletteChanged: boolean) => void;
 }
 
 export const AssetPalette = (props: AssetPaletteProps) => {
     const { onClose } = props;
-
+    const [customPalettes, setCustomPalettes] = useState<CustomPalettes>(undefined);
+    const [initialPalette, setinitialPalette] = useState<Palette | undefined>(undefined);
```
jwunderl, line 21: nit: camel casing for `setInitialPalette`

jwunderl (approved):
Nice!

PR #9279, opened 2023-01-06
multiplayer: add event tick "mp.remixgame"

eanders-ms:


multiplayer/src/components/RemixGameButton.tsx
```diff
@@ -12,6 +12,7 @@ export default function Render() {
 
     function handleRemixGameClick() {
         if (remixUrl) {
+            pxt.tickEvent("mp.remixgame", { gameId: gameId! })
```
abchatra, line 15: Isn't this private info?

eanders-ms: It's the share id of the game. Is that private?

abchatra: Until now we don't emit gameId in our telemetry: See this: https://github.com/microsoft/pxt/blob/8178b4ba05f2d9d1dfc8f48a18bf2186e98e665a/docfiles/tracking.html#L60

We can discuss though; these are games published for public. Take it offline?

benvillalobos (approved):

PR #9281, opened 2023-01-07
Adding lang as customDimension for every tick

abchatra:
When the telemetry is enabled, we set the lang as custom dimension and if the user changes the language to new one, we observe setcookielang and update this.
This is dependent on language cookie which all our properties use to determine the locale.

Joey noted this won't work for forcelang and in context translation as they don't set the cookie. However, these are not the primary user scenarios, and I am ok not being right in these cases. Piggy backing cookie seemed cleaner to me.

### DIFF
--- a/multiplayer/src/index.tsx
+++ b/multiplayer/src/index.tsx
@@ -12,7 +12,7 @@ import App from "./App";
 import { AppStateProvider } from "./state/AppStateContext";
 
 function enableAnalytics() {
-    pxt.analytics.enable();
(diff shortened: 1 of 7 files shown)


jwunderl (approved):

eanders-ms:
Should this apply to skillmap too?

abchatra:
@eanders-ms this should be applicable to all.
@jwunderl I also changed to the API you mentioned, I tested it and worked well too. Better to cover all scenarios.

jwunderl (approved):

eanders-ms (approved):

PR #9282, opened 2023-01-09
Small refactor for multiplayer utils (was: Add "play with friends" button to home page)

eanders-ms:
Update! We aren't going to add a play with friends button, but go with a banner instead. Still, this PR has some changes I want to keep. I've removed the header button and kept the other changes.


Jaqster:
I thought we were going to have a row of Multiplayer games on the home page, and that was how we were going to promote the experience?  I'm a little hesitant to clutter up the top nav bar with more buttons.  Can we take some time to discuss?

eanders-ms:
> I thought we were going to have a row of Multiplayer games on the home page, and that was how we were going to promote the experience? I'm a little hesitant to clutter up the top nav bar with more buttons. Can we take some time to discuss?

We plan to add a row of multiplayer games as well. We settled on this design some time ago, but it isn't too late to reopen the conversation. Will schedule it.

In the meantime you may like to review the list of multiplayer changes planned for this release, in case there are any more concerns: https://github.com/microsoft/pxt-arcade/issues?q=is%3Aissue+is%3Aopen+label%3Amultiplayer+milestone%3AJan-winter-Release

eanders-ms:
Update! We aren't going to add a play with friends button, but go with a banner instead. Still, this PR has some changes I want to keep. I've removed the header button and kept the other changes.

thsparks (approved):
Wasn't sure if you still wanted the change to multiplayer.svg. Otherwise looks okay to me!

pxtlib/cloud.ts
```diff
@@ -1,4 +1,19 @@
 namespace pxt.cloud {
+    export type DevBackendType = "default" | "prod" | "staging" | "localhost";
+
+    const DEV_BACKEND_PROD = "https://www.makecode.com";
+    const DEV_BACKEND_STAGING = "https://staging.pxt.io";
+    const DEV_BACKEND_LOCALHOST = "http://localhost:8080";
+
+    export const DEV_BACKEND = DEV_BACKEND_PROD;
```
jwunderl, line 8: nit; if this is being exported we might as well give it a proper type to avoid having to `<any>` like below / false errors that might show up elsewhere when it's type is super narrowed by the const; e.g.

```typescript
type BackendUrls = typeof DEV_BACKEND_PROD | typeof DEV_BACKEND_STAGING | typeof DEV_BACKEND_LOCALHOST;
export const DEV_BACKEND: BackendUrls = DEV_BACKEND_PROD;

export function devBackendType(): DevBackendType {
    if (DEV_BACKEND === DEV_BACKEND_PROD) return "prod";
    if (DEV_BACKEND === DEV_BACKEND_STAGING) return "staging";
    if (DEV_BACKEND === DEV_BACKEND_LOCALHOST) return "localhost";
    return "prod";
}
```

theme/common.less
```diff
@@ -2297,22 +2295,18 @@ div.signin-button {
...
     font-family: @segoeUIFont;
     font-weight: 500;
 
     .text {
         margin-right: .5rem;
     }
-}
-
-div.signin-button:hover {
-    transform: scale(1.1);
```
jwunderl, line 2310: 👍

jwunderl (approved):
since * can't comment on svgs directly here:

for `svgicons/multiplayer.svg`, should that remain 1000x1000? It's mentioned as a requirement here https://github.com/microsoft/pxt#icons but I haven't touched the svg icons path much so I'm not sure if that actually matters or if it's just a leftover comment from 6 years ago

PR #9287, opened 2023-01-11
Multiplayer post recieve play icons

jwunderl:
Support for passing up / down icons (imgs up to 64x64) from host -> all for use in reactions, player icons

image

closes https://github.com/microsoft/pxt-arcade/issues/5432
closes https://github.com/microsoft/pxt-arcade/issues/5431

I might take a pass at the sizing for the `pixel-art-image`s again after I add the player chooser sim / other things are ready, as sizing them correctly to fit in within rounded edges was slightly tricky to get right / might have missed some sizes if someone makes extra wide / tall images -- will be easier to test when it's in /beta as with just local host it's a bit tricky (e.g. have to push up a broken project as all user projects are pulled from prod and localhost can't push share scripts to prod)

image

### DIFF
--- a/multiplayer/src/components/ReactionParticle.tsx
+++ b/multiplayer/src/components/ReactionParticle.tsx
@@ -1,9 +1,13 @@
-import { useMemo } from "react";
+import { useContext, useMemo } from "react";
+import { AppStateContext } from "../state/AppStateContext";
 import { Reactions, Particle } from "../types/reactions";
 
 export default function Render(props: { particle: Particle }) {
     const { particle } = props;
     const { index } = particle;
+    const { state } = useContext(AppStateContext);
+    const { gameState } = state;
+    const reactionIconOverrides = gameState?.reactionIconOverrides;
 
     const emoji = Reactions[index].emoji;
 
@@ -25,12 +29,25 @@ export default function Render(props: { particle: Particle }) {
         };
     }, []);
 
+    const override = reactionIconOverrides?.[index + 1];
+    const display = override ? (
+        <img
+            className="pixel-art-image tw-w-[100%]"
+            src={override}
+            alt={lf(`Game reaction image ${index + 1}`)}
+        />
+    ) : (
+        emoji
+    );
+
     return (
         <div
             className="tw-absolute tw-select-none tw-pointer-events-none tw-z-[100]"
             style={{
                 animation: `reaction-fly-up linear ${parms.flyUpDuration}s forwards,
                 ${parms.horzAnim} linear ${parms.flyHorzDuration}s forwards`,
+                width: override ? "1.5rem" : undefined,
+                height: override ? "1.5rem" : undefined,
             }}
         >
             <div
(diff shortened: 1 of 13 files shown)


multiplayer/src/services/gameClient.ts
```diff
@@ -428,6 +437,58 @@ class GameClient {
...
+            dispatchHandler = setReactionIconOverride;
+        } else {
+            // unhandled icon type, ignore.
+            return;
+        }
+
+        if (iconBuffer) {
+            const unzipped = await gunzipAsync(iconBuffer);
+            const imgConv = new pxt.ImageConverter();
+            const iconPalette = unzipped.slice(0, 48);
```
eanders-ms, line 461: Would it make sense to use `PALETTE_BUFFER_SIZE`, defined above?

multiplayer/src/components/ReactionParticle.tsx
```diff
@@ -25,12 +29,25 @@ export default function Render(props: { particle: Particle }) {
         };
     }, []);
 
+    const override = reactionIconOverrides?.[index + 1];
```
kimprice, line 32: Why do you need the `+ 1`?

jwunderl: reactions are 0 indexed but displayed / used as 1 index by users (see `/Reaction.tsx`); `reactionIconOverrides` come directly from user code so I left the indexing there as 1 indexed there to make it a bit easier to match state vs user action (and also to be consistent with the player slots which are 1 indexed), so need to

multiplayer/src/epics/setCustomIconAsync.ts
```diff
@@ -0,0 +1,23 @@
...
+    setPresenceIconOverride,
+    setReactionIconOverride,
+} from "../state/actions";
+
+export async function setCustomIconAsync(
+    iconType: IconType,
+    slot: number,
+    pngDataUri?: string
+) {
+    try {
```
eanders-ms, line 14: Would it make sense to have a telemetry tick here? It might not, if we're doing this automatically, but if we knew whether or not it was done explicitly by the user we could just log those.

multiplayer/src/components/icons/UserIcon.tsx
```diff
@@ -1,5 +1,11 @@
-export default function Render(props: { slot: number }) {
-    const { slot } = props;
+export default function Render(props: { slot: number; datauri?: string }) {
```
kimprice, line 1: should you camel case `datauri`?

multiplayer/src/epics/setCustomIconAsync.ts
jwunderl: the implicit / explicit bit isn't passed around right now -- I see it more like an `!important` where most of the time the default is what you want, and it's just for the special case where you want a little more control that you would use another block, so outside the simulator they're treated the same.

That said happy to add a tick here but it would be a bit noisy - a standard, full multiplayer game would usually send 16 ticks per play of the game (i.e. would resend them on restart, as it stems from user code that will typically be in the on start event)

eanders-ms: Makes sense, thanks.

eanders-ms (approved):

kimprice (approved):

PR #9288, opened 2023-01-11
Add "Host multiplayer game" button below simulator

eanders-ms:
Adds a "host multiplayer game" button below the simulator when the game uses multiplayer features.

image

Clicking this button takes you to the share dialog, working in a slightly different mode. Instead of a Share Project button, it promotes the "Host a multiplayer game" button to be the primary (and only) one:
image

Clicking that:
image


jwunderl (approved):
🎮🕹️

react-common/components/share/ShareInfo.tsx
```diff
@@ -251,17 +270,18 @@ export const ShareInfo = (props: ShareInfoProps) => {
                         <div className="project-share-publish-actions">
                             {shareState === "share" &&
                             <>
-                                {pxt.appTarget?.appTheme?.multiplayer && isMultiplayerGame &&
-                                    <Button className="primary inverted text-only share-publish-button share-host-button"
+                                {pxt.appTarget?.appTheme?.multiplayer && (isMultiplayerGame || forMultiplayer) &&
```
kimprice, line 273: Why do we need to also check `forMultiplayer` here? If `forMultiplayer` is true, wouldn't `isMultiplayerGame` always be true?

kimprice (approved):

react-common/components/share/ShareInfo.tsx
eanders-ms: We don't have a UI flow for the `forMultiplayer: true, isMultiplayerGame: false` scenario , but I didn't want to exclude it. I can see a future where we have UI support for hosting single player games online. It would be a part of spectating games, and branded something like "stream your gameplay to your friends".

PR #9289, opened 2023-01-11
Color palette editor: delete, rename, add new palette

kimprice:
The color palette editor functionality is complete. You can now delete and rename custom palettes. You can also create a new palette from the `+` icon.

Try it out here: https://arcade.makecode.com/app/1695b56f178734461bfa309891b588452b196135-a3a6a16858

image

Closes https://github.com/microsoft/pxt-arcade/issues/5486, closes https://github.com/microsoft/pxt-arcade/issues/5327, closes https://github.com/microsoft/pxt-arcade/issues/5488. Refer to https://github.com/microsoft/pxt-arcade/issues/5485.


eanders-ms:
I believe svgicons should be 1000x1000 (but not sure why!), and run through the `svgo` utility, according to https://github.com/microsoft/pxt#icons.

riknoll:
if the icons aren't 1000x1000, then the svg to font library we have sometimes resizes them weirdly. making them all 1000x1000 results in a much more consistent icon size in the font (it's still not perfect though 😅)

webapp/src/components/assetEditor/assetPalette.tsx
```diff
@@ -15,13 +18,20 @@ export interface AssetPaletteProps {
     onClose: (paletteChanged: boolean) => void;
 }
 
+const enum NameModal {
```
eanders-ms, line 21: Just curious: Why does this enum need to be const?

react-common/components/palette/ColorPickerField.tsx
```diff
@@ -25,6 +25,7 @@ export const ColorPickerField = (props: ColorPickerFieldProps) => {
     }
 
     const onTextInputChanged = (newValue: string) => {
+        newValue = newValue.trim();
```
jwunderl, line 28: Just to confirm, this is the fix for what I mentioned yesterday where it was ignoring input if you typed in ` #ffffff `, right?

react-common/styles/palette/PaletteSwatch.less
```diff
@@ -25,4 +34,13 @@
 
 .common-palette-picker > .common-dropdown > .common-button .common-button-flex {
     display: flex;
+}
+
+.common-palette-picker > .common-dropdown > .common-menu-dropdown-pane {
+    max-height: 35rem;
+    overflow-y: auto;
+    overflow-x: hidden;
+    ul > li > .common-button.common-dropdown-item{
```
jwunderl, line 43:
```suggestion
    ul > li > .common-button.common-dropdown-item {
```

nit

jwunderl (approved):
LGTM!

eanders-ms (approved):

react-common/components/palette/ColorPickerField.tsx
kimprice: yes thanks for catching it!

webapp/src/components/assetEditor/assetPalette.tsx
jwunderl: It doesn't make a real difference here but const enums do compile slightly differently (i.e. you can't index into them with a value to get the enum member string it came from) https://www.typescriptlang.org/docs/handbook/enums.html#const-enums

kimprice: I guess it doesn't need to be. Should I drop the `const`?

PR #9293, opened 2023-01-16
Sort non-builtin assets at top of gallery

kimprice:
closes https://github.com/microsoft/pxt-arcade/issues/5078

### DIFF
--- a/pxtlib/tilemap.ts
+++ b/pxtlib/tilemap.ts
@@ -17,6 +17,7 @@ namespace pxt {
         tags?: string[];
         blockIDs?: string[];
         temporaryInfo?: TemporaryAssetInfo;
+        package?: string;
     }
 
     export interface TemporaryAssetInfo {
@@ -987,6 +988,7 @@ namespace pxt {
                 const images = this.readImages(dep.parseJRes(), isProject);
 
                 for (const image of images) {
+                    image.meta.package = dep.id;
                     if (image.type === AssetType.Tile) {
                         if (isProject) {
                             this.state.tiles.add(image);
--- a/pxtlib/tilemap.ts
+++ b/pxtlib/tilemap.ts
@@ -1022,20 +1022,22 @@ namespace pxt {
                         }
                     }
                 }
+
+                for (const tm of getTilemaps(dep.parseJRes())) {
+                    this.state.tilemaps.add({
+                        internalID: this.getNewInternalId(),
+                        type: AssetType.Tilemap,
+                        id: tm.id,
+                        meta: {
+                            // For tilemaps, use the id as the display name for backwards compat
+                            displayName: tm.displayName || tm.id,
+                            package: dep.id
+                        },
+                        data: decodeTilemap(tm, id => this.resolveTile(id))
+                    })
+                }
             }
 
-            for (const tm of getTilemaps(pack.parseJRes())) {
-                this.state.tilemaps.add({
-                    internalID: this.getNewInternalId(),
-                    type: AssetType.Tilemap,
-                    id: tm.id,
-                    meta: {
-                        // For tilemaps, use the id as the display name for backwards compat
-                        displayName: tm.displayName || tm.id
-                    },
-                    data: decodeTilemap(tm, id => this.resolveTile(id))
-                })
-            }
             this.committedState = this.cloneState();
             this.undoStack = [];
             this.redoStack = [];
--- a/pxtlib/tilemap.ts
+++ b/pxtlib/tilemap.ts
@@ -1022,20 +1022,20 @@ namespace pxt {
                         }
                     }
                 }
+            }
 
-                for (const tm of getTilemaps(dep.parseJRes())) {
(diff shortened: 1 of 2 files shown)


webapp/src/assets.ts
```diff
@@ -70,6 +70,13 @@ export function getAssets(gallery = false, firstType = pxt.AssetType.Image, temp
         }
     }
 
+    // sort so that non-builtin assets appear at the top of the gallery
+    images.sort(comparePackage);
+    tiles.sort(comparePackage);
+    tilemaps.sort(comparePackage);
```
kimprice, line 76: Do we support extensions for tilemap packs and/or do we have builtin tilemaps? Wondering if it's necessary for me to sort here. And if so, then I think I'd need to add the package info as well like in L1032 of tilemap.ts.

jwunderl: songs and tilemaps don't currently have galleries so they couldn't be pulled in from extensions, but I'd probably support it if it's easy (which it should be when it's just adding a field when loading) as otherwise this sort of thing will be easily forgotten in the future if we do choose to add galleries for them.

webapp/src/assets.ts
```diff
@@ -118,4 +125,10 @@ export function assetToGalleryItem(asset: pxt.Asset, imgConv = new pxt.ImageConv
 
 function compareInternalId(a: pxt.Asset, b: pxt.Asset) {
     return a.internalID - b.internalID;
+}
+
+function comparePackage(a: pxt.Asset, b: pxt.Asset) {
+    const aPack = a.meta.package === "device" ?  1 : 0;
+    const bPack = b.meta.package === "device" ?  1 : 0;
```
jwunderl, line 132:
```suggestion
    const aPack = a.meta.package === "device" ? 1 : 0;
    const bPack = b.meta.package === "device" ? 1 : 0;
```

nit

webapp/src/assets.ts
```diff
@@ -70,6 +70,13 @@ export function getAssets(gallery = false, firstType = pxt.AssetType.Image, temp
         }
     }
 
+    // sort so that non-builtin assets appear at the top of the gallery
```
jwunderl, line 73: can we maybe combine this with the sorting we do for in `compareInternalId` -- that'd just basically be `return comparePackage if that value is non-zero, otherwise do current compareInternalId`. We shouldn't have to worry about the assets that get added in between that sort (L46-51) and here as those are all temporary assets which go into myAssets not the gallery, and combining the sorts is nice as it makes reasoning about the sorting easier (that way you can see the full sorting logic in one place & you don't have to know that javascript sort is guaranteed stable by default as of es2019)

pxtlib/tilemap.ts
```diff
@@ -1020,20 +1022,22 @@ namespace pxt {
...
-                })
+                for (const tm of getTilemaps(dep.parseJRes())) {
+                    this.state.tilemaps.add({
+                        internalID: this.getNewInternalId(),
+                        type: AssetType.Tilemap,
+                        id: tm.id,
+                        meta: {
+                            // For tilemaps, use the id as the display name for backwards compat
+                            displayName: tm.displayName || tm.id,
+                            package: dep.id
```
jwunderl, line 1034: Ah hm, I meant just adding in the package field so it exists (like `package: pack.id`), not necessarily expanding it to read in tilemaps from extensions right now; it probably shouldn't do any harm as if they had tilemaps we'd already error on the generated files conflicting but I'm not certain / don't think there's a benefit to pulling them in at this time; @riknoll any opinion here?

riknoll: yeah, i'd stick with pack.parseJRes()


if we're going to change that behavior, let's do it in a different pr

webapp/src/assets.ts
```diff
@@ -118,4 +118,13 @@ export function assetToGalleryItem(asset: pxt.Asset, imgConv = new pxt.ImageConv
 
 function compareInternalId(a: pxt.Asset, b: pxt.Asset) {
     return a.internalID - b.internalID;
+}
+
+function comparePackage(a: pxt.Asset, b: pxt.Asset) {
+    const aPack = a.meta.package === "device" ? 1 : 0;
```
riknoll, line 124: instead of hardcoding device, we could probably detect what the  default package is:

```ts
const blocksprj = pxt.appTarget.bundledpkgs[pxt.BLOCKS_PROJECT_NAME];
            const config = blocksprj?.[pxt.CONFIG_NAME];
            if (config) {
                const pkgConfig: pxt.PackageConfig = JSON.parse(config);
                // cache this and check for the package id in this list
                const dependencies = Object.keys(pkgConfig.dependencies);
            }
```
obviously you'd want to cache the result from this instead of reparsing the config for each comparison
i haven't tested this snippet, so it might need tweaking

riknoll (changes requested):
LGTM besides small comments

riknoll (approved):

jwunderl (approved):

PR #9296, opened 2023-01-17
fix bug where 'exit the game' button does nothing

jwunderl:
fixes https://github.com/microsoft/pxt-arcade/issues/5508

Interesting gotcha bug I hadn't thought of. The issue was that `PlayerMenuPopup` was declared inside the `Presence` component, so the function was getting redeclared every render. This feels like it should be innocuous but it's not; any time ANY state was changing in the app it was causing the component to rerender every frame, as `PlayerMenuProp` was getting counted as a separate component each time (you couldn't even expand div in the dev tools as it kept replacing itself).

This didn't matter before as the calls to render didn't occur quite so frequently (so if you clicked on it and it didn't do anything you'd default to think that you misclicked or something), but the state was changing much more often as a result of the rapidly animating reactions changing so often in my sample game, so this bug became more prominent for that reason - it must've started hitting some internal clean up logic of 'don't run click event as the component is being rerendered / doesn't exist on dom anymore' after the popup `useClickedOutside` intercepted the event for just long enough to make it not be the same button anymore.

### DIFF
--- a/multiplayer/src/components/Presence.tsx
+++ b/multiplayer/src/components/Presence.tsx
@@ -6,6 +6,27 @@ import Popup from "./Popup";
 import { Button } from "react-common/components/controls/Button";
 import { showModal } from "../state/actions";
 
+function PlayerMenuPopup(
+    props: React.PropsWithChildren<{
+        slot: number;
+        setShowPlayerMenu: (p: number) => void;
+        currentlyDisplayed: number;
+    }>
+) {
+    const { slot, children, currentlyDisplayed, setShowPlayerMenu } = props;
+    return (
+        <Popup
+            className="tw-absolute tw-z-50 tw-translate-y-[-110%] tw-translate-x-[-50%] tw-rounded-lg tw-border-2 tw-border-gray-100"
+            visible={currentlyDisplayed === slot}
+            onClickedOutside={() => setShowPlayerMenu(0)}
+        >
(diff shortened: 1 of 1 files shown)


kimprice (approved):
Nice fix!

multiplayer/src/components/Presence.tsx
```diff
@@ -6,6 +6,27 @@ import Popup from "./Popup";
 import { Button } from "react-common/components/controls/Button";
 import { showModal } from "../state/actions";
 
+function PlayerMenuPopup(
```
eanders-ms, line 9: This approach seems fine, but did you consider trying to memoize it with `useMemo`?

jwunderl: I didn't, I've only used `useMemo` with computed values / pure functions vs components so I'm not certain I see how it could be applied in this case -- would it be to memoize the `PlayerMenuPopup` function so that the jsx element it returns out can be compared nicely? This component seems to rely on the `setShowPlayerMenu` being current (& `children` which I'm not sure how they would be compared in the memo) so I'm not sure how that would be handled with `useMemo` without getting stale function references after react rerenders

eanders-ms (approved):
looks good!

multiplayer/src/components/Presence.tsx
eanders-ms: Just a thought I had after a quick look. Your way seems more correct.

PR #9297, opened 2023-01-18
Kiosk: added a button in the share modal to send game to a kiosk

srietkerk:
We want sharing to the kiosk in a classroom to be easier, so we added the ability for students to share directly from their games to the kiosk in the share modal.

When you click on the kiosk button in the smorgasbord of share options, an extra input field appears to enter the kiosk id. The visibility of the input follows the same pattern as the embed button. 

Here's what it looks like:
image

TODO: add a tooltip to help the user understand what the kiosk is. I don't have designs or a link to point to for this, so it is left out for now.
We will also want some error messaging if the submission was not successful.

I will upload a target tomorrow so y'all can play with it. 

Steps to take to play with it: 
1. Go to kiosk https://arcade.makecode.com/kiosk
2. Go to add a game ( so you get a kiosk id)
3. Copy the kiosk id
4. Open a game in arcade
5. Share the game
6. Click on the new kiosk button
7. Paste the kiosk code in the new input that shows up and click submit
8. Go back to the kiosk, the game should be launched there. IF the game is not launched, check your arcade editor for errors or just wait a couple of seconds. The upload can take a couple of seconds based on the time that we are waiting for the poll.

Closes https://github.com/microsoft/pxt-arcade/issues/5447


react-common/components/share/ShareInfo.tsx
```diff
@@ -100,21 +104,54 @@ export const ShareInfo = (props: ShareInfoProps) => {
         setCopySuccessful(false);
     }
 
+    const handleKioskSubmitBlur = () => {
+        setKioskSubmitSuccessful(false);
+    }
+
+    const handleKioskSubmitClick = async () => {
+        const gameId = /([^\/]+$)/.exec(shareData.url)?.[1];
```
kimprice, line 112: I don't know reg expressions well, but why do you take `[1]` here instead of `[0]`?

srietkerk: I honestly have no clue why I decided to use 1 here. I can change it to 0. I'm also not all that great with regex, but `.exec()` returns an array whose first entry is the text that matches the given regular expression and the second entry holds the first string that fit defined capturing groups. Since I wrapped the regex in a capturing group, the entries at index 0 and 1 are the same. I'm not sure if there's a benefit to referencing one over the other. I can also be missing something because I do have limited understanding and experience with regex.

jwunderl: https://regex101.com/r/lSUeQc/1 is a helpful site for evaluating regexes -- the `[1]` here grabs the capture group (stuff in parentheses) rather than the match (portion of the string matching the entire regex), but in this case those two will be the same thing as it's just grabbing everything after the last `/` (that is, if you wanted you could remove the parentheses and just grab `[0]`, but again, it really doesn't matter~).

That said, we should just use `pxt.Cloud.parseScriptId` here to extract shareid~

react-common/components/share/ShareInfo.tsx
```diff
@@ -311,6 +352,10 @@ export const ShareInfo = (props: ShareInfoProps) => {
                             </div>
                             <div className="project-share-actions">
                                 <div className="project-share-social">
+                                <Button className="square-button gray mobile-portrait-hidden"
```
jwunderl, line 355: This needs to go behind a pxtarget flag, something like `pxt.appTarget.appTheme.shareToKiosk` - we won't be wanting this in micro:bit, for example~

abchatra:
It would be great to have a simple question mark (?) or Help or "How do I get the kiosk code" below this which sends the user to the documentation page.
image

react-common/components/share/ShareInfo.tsx
```diff
@@ -100,21 +104,54 @@ export const ShareInfo = (props: ShareInfoProps) => {
...
+    const handleKioskSubmitClick = async () => {
+        const gameId = /([^\/]+$)/.exec(shareData.url)?.[1];
+        if (kioskInputRef?.value) {
+            const validKioskId = /^[a-zA-Z0-9]{6}$/.exec(kioskInputRef.value)?.[0];
+            if (validKioskId) {
+                setKioskSubmitSuccessful(true);
+                try {
+                    await addGameToKioskAsync(validKioskId, gameId);
+                } catch (error) {
+                    console.log("there was an error sending game to kiosk");
```
eanders-ms, line 120: It might be helpful to log the exception message here.

react-common/components/share/Kiosk.ts
```diff
@@ -0,0 +1,20 @@
...
+            },
+            body: JSON.stringify({
+                "kioskCode": kioskId,
+                "shareId": gameId
+            }),
+        });
+        await response.json();
+    }
+    catch (error) {
+        throw new Error("Failed to post game to the kiosk");
```
eanders-ms, line 18: It might be helpful to include the error message in the new Error being thrown.

react-common/components/share/ShareInfo.tsx
eanders-ms, line 120:
```suggestion
                    pxt.log("there was an error sending game to kiosk");
```

react-common/components/share/ShareInfo.tsx
```diff
@@ -361,6 +406,27 @@ export const ShareInfo = (props: ShareInfoProps) => {
                             rows={5}
                             initialValue={shareData?.embed[embedState]} />
                     </div>}
+                    {kioskState &&
+                        <div>
+                            <div className="project-share-label">
+                                Enter Kiosk Code
```
eanders-ms, line 412:
```suggestion
                                {lf("Enter Kiosk Code")}
```

srietkerk:
> It would be great to have a simple question mark (?) or Help or "How do I get the kiosk code" below this which sends the user to the documentation page. image

I ended up just implementing the component side of this. Claire gave me the recommended design with the question mark next to the title, so it's in the changes. This won't do anything for the time being, though, while the documentation for kiosk is still getting done.

react-common/components/share/ShareInfo.tsx
```diff
@@ -146,6 +146,13 @@ export const ShareInfo = (props: ShareInfoProps) => {
         }
     }
 
+    const handleKioskHelpClick = () => {
+        pxt.log("this will lead the user to documentation");
+        // TODO: have this actually link to the documentation when it's done
+        //const kioskDocumentationUrl = ....
```
jwunderl, line 152: It's usually easier to just point at a page here and fill that in later, since md is live / can be updated whenever and worst case is seeing a 404 -- occasionally you'll run into issues like cors or browsers not allowing pop ups for one reason or another so at least seeing it's going to a reasonable page is typically good :) (that said it looks like https://github.com/microsoft/pxt-arcade/pull/5514 is almost ready so easiest to just plug that in)

jwunderl (approved):

srietkerk:
I know it's basically too late, but if anyone wanted to play with the new button, here's a target: 
https://arcade.makecode.com/app/75fe7d1e5301ad5f71741b541162f136a43cb884-ce828dcb58#editor

And the steps:
Steps to take to play with it:

Go to kiosk https://arcade.makecode.com/kiosk
Go to add a game ( so you get a kiosk id)
Copy the kiosk id
Open a game in arcade
Share the game
Click on the new kiosk button
Paste the kiosk code in the new input that shows up and click submit
Go back to the kiosk, the game should be launched there. IF the game is not launched, check your arcade editor for errors or just wait a couple of seconds. The upload can take a couple of seconds based on the time that we are waiting for the poll.

kimprice:
If you send the same game to the kiosk twice, does it keep both or only the most up to date?

srietkerk:
> If you send the same game to the kiosk twice, does it keep both or only the most up to date?

It depends on if you are using a persistent share link or not. If you use a persistent share link, nothing will happen if you try to send it twice. Whenever you go to re-share the game, if you refresh kiosk, kiosk will get the most up-to-date changes.

If you are using a non-peristent share link and reshare the game, then two entries will show up in kiosk for the same game. You have to reshare, though. 

Kiosk checks the ids that are being added. If the game id already exists in the list, it will not add it to the list.

PR #9298, opened 2023-01-18
add help button for how to pop out video in firefox

jwunderl:
Possible quick fix for https://github.com/microsoft/pxt-arcade/issues/5490, since firefox doesn't support picture in picture api / have there own way to do it. 

image

We'd have to write a quick markdown page at `/firefox-picture-in-picture` which I didn't do, but would likely largely be similar to https://support.mozilla.org/en-US/kb/about-picture-picture-firefox

### DIFF
--- a/webapp/src/marked.tsx
+++ b/webapp/src/marked.tsx
@@ -350,6 +350,18 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
                     pipButton.textContent = lf("Pop out video");
                     pipButton.ariaLabel = lf("Open video in picture-in-picture mode");
                     pipButton.title = pipButton.ariaLabel;
+                } else if (pxt.BrowserUtils.isFirefox()) {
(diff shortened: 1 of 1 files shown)


thsparks:
LGTM, but should we wait until we have the docs to check in?

kiki-lee:
Could we do an anigif to quickly show how instead? Something that appears in a div when you click that button?

abchatra:
Primary object for the students is to watch the video and popping out is secondary action only if required.  This long text makes me think, students will try to click or play with the button instead of watching the video. Just keep the old "Pop out" and take them to the documentation page for firefox? It is also not worthwhile to spending engineering effort on this as Firefox is less than 1% of our use base.

kiki-lee:
I'm kind of worried that the pop-out markdown would cause more trouble than the video.  Even with the button, they're probably pretty likely to click that full-screen icon. 

This is a tough one,

PR #9299, opened 2023-01-19
Add toggle for music editor bass clef and better support for sharps/flats

riknoll:
This PR does three things:

1. Adds a "show bass clef" checkbox to the music editor
2. Revamps how sharps/flats are handled to better support changing the enharmonic spelling of a note
3. Reorders the instruments so that the dog is first (sorry duck)

(2) is the big one because I had to change the way notes are represented. Now you can change any note to be sharp or flat by holding ctrl and clicking it. You can also use the **j** keyboard shortcut to cycle through the options on the selected note.

The unfortunate side effect of this PR is that any existing songs will no longer sound right after this is merged. ah, the dangers of living on the bleeding edge.

I will be opening a corresponding pxt-common-packages pr momentarily

### DIFF
--- a/pxtblocks/fields/field_utils.ts
+++ b/pxtblocks/fields/field_utils.ts
@@ -161,7 +161,7 @@ namespace pxtblockly {
                 if (col > cellsShown) break;
 
                 for (const note of noteEvent.notes) {
-                    const row = 12 - (note % 12);
+                    const row = 12 - (note.note % 12);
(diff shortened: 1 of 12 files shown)


riknoll:
and here's the promised pxt-common-packages pr: https://github.com/microsoft/pxt-common-packages/pull/1391

jwunderl (approved):

PR #9300, opened 2023-01-19
Code Validation Infrastructure + Blocks Exist Validator

thsparks:
This change adds a new infrastructure for code validation in tutorials, and includes one validator, which checks if the highlighted blocks in a tutorial step's answer key exist in the workspace, and that they're connected to the code.

Note that the new `MarkdownMetadataParser` file is all just cut/paste code from `skillMapParser`. I moved it to pxtlib so we can access it to parse the validation markdown, too.

Sample syntax for specifying a validator in tutorial markdown:
````
```validation.global
# BlocksExistValidator
```
````

You could also enable it for a single step using validation.local, much like blockConfig, and if you wanted to enable it globally but disable it for a single step, then you can enable it as above and add this in the local step you wish to disable it on:
````
```validation.local
# BlocksExistValidator
* enabled: false
```
````

If validation fails, a popup will appear (it does not close unless the user selects "Continue Anyway" or the close button):
image

If the user clicks "Show Hint" it will update to display the answer key and some descriptive text:
image

I'm still trying to figure out why there's a border around the blocks, but didn't want to block the PR on it. Also, in the future, it would be best to show just the missing/detached blocks instead of the whole answer key, but that wasn't in scope for this MVP.

### CODE REVIEW

File: webapp/src/components/tutorialValidators.tsx [hunk 1]
```diff
@@ -0,0 +1,171 @@
...
+
+        if (missingBlocks.length > 0 && disabledBlocks.length > 0) {
+            isValid = false;
+            errorDescription = lf("Make sure you see blocks that look like this and that they're connected to the rest of your code.")
+        } else if (missingBlocks.length > 0) {
+            isValid = false;
+            errorDescription = lf("Make sure you see blocks that look like this in your workspace.");
+        } else if (disabledBlocks.length > 0) {
+            isValid = false;
+            errorDescription = lf("Make sure your blocks are connected to the rest of your code like this.");
```
thsparks, line 127: @kiki-lee Would love to get your thoughts on the three different error descriptions here. One is for when there are both missing and detached blocks, one is when there's only missing blocks, and one is when there's only detached blocks.

cli/cli.ts
```diff
@@ -6136,6 +6136,7 @@ function internalCacheUsedBlocksAsync(): Promise<Map<pxt.BuiltTutorialInfo>> {
                                 if (!builtInfo.snippetBlocks[snippetHash]) builtInfo.snippetBlocks[snippetHash] = {};
                                 builtInfo.snippetBlocks[snippetHash][m1] = 1;
                                 builtInfo.usedBlocks[m1] = 1;
+                                //TODO: Fill builtInfo.HighlightedBlocks
```
thsparks, line 6139: Created https://github.com/microsoft/pxt-arcade/issues/5516 to track

jwunderl: for reference for other reviewers -- this cli fn fills out the tutorial info cache, which is used to cache used blocks / etc so first load of our 'hero' tutorials is faster. The reason it's okay that we're not fixing it in this pr is because currently micro:bit doesn't use this (no `cacheusedblocksdir` in that pxtarget) and while arcade tries to use this feature, it crashes at build time and doesn't produce anything (you can see `pxt.appTarget.tutorialInfo` is empty) -- minecraft does so we'll have to fix this eventually (and should fix for arcade too) but fine for now~

pxtlib/markdownMetadataParser.ts
```diff
@@ -0,0 +1,144 @@
+namespace pxt {
+    export interface MarkdownSection {
+        headerKind: "single" | "double" | "triple";
+        header: string;
+        attributes: { [index: string]: string };
+        listAttributes?: { [index: string]: MarkdownList };
```
jwunderl, line 6: nit: probably use `pxt.Map<string>` / `pxt.Map<MarkdownList>` helper types~

webapp/src/components/tutorial/TutorialContainer.tsx
```diff
@@ -106,9 +109,77 @@ export function TutorialContainer(props: TutorialContainerProps) {
...
     const hintMarkdown = steps[visibleStep].hintContentMd;
 
+    const validateTutorialStep = async () => {
+        const globalValidators = tutorialOptions.globalValidationConfig?.validators
+                                    ?? PopulateValidatorCache(tutorialOptions.globalValidationConfig);
+        const localValidators = currentStepInfo.localValidationConfig?.validators
+                                    ?? PopulateValidatorCache(currentStepInfo.localValidationConfig);
+
+        let validators: pxt.Map<CodeValidator> = {}
+        if(localValidators) {
```
jwunderl, line 119:
```suggestion
        if (localValidators) {
```

nit, several of these throughout file

webapp/src/components/tutorial/TutorialContainer.tsx
```diff
@@ -106,9 +109,77 @@ export function TutorialContainer(props: TutorialContainerProps) {
...
+            Object.entries(localValidators).forEach(v => validators[v[0]] = v[1]);
+        }
+        if (globalValidators) {
+          Object.entries(globalValidators).forEach(v => {
```
jwunderl, line 123: nit: looks like there's a bit of a mix of 4 space tabs and 2 space tabs in this file, should be 4 across the board~

webapp/src/components/tutorial/TutorialContainer.tsx
```diff
@@ -106,9 +109,77 @@ export function TutorialContainer(props: TutorialContainerProps) {
...
+        const globalValidators = tutorialOptions.globalValidationConfig?.validators
+                                    ?? PopulateValidatorCache(tutorialOptions.globalValidationConfig);
+        const localValidators = currentStepInfo.localValidationConfig?.validators
+                                    ?? PopulateValidatorCache(currentStepInfo.localValidationConfig);
+
+        let validators: pxt.Map<CodeValidator> = {}
+        if(localValidators) {
+            Object.entries(localValidators).forEach(v => validators[v[0]] = v[1]);
+        }
+        if (globalValidators) {
```
jwunderl, line 122: could we just declare validators as 

```typescript
const validators: pxt.Map<CodeValidator> = {
    ...(globalValidators ?? {}),
    ...(localValidators ?? {}),
};
```

? should get the behavior you want of localValidators overriding globals and maybe a little easier to read.

(will defer to you though as they'll ultimately do the same thing~)

webapp/src/components/tutorial/TutorialValidationErrorMessage.tsx
```diff
@@ -0,0 +1,81 @@
+import * as React from "react";
+import { Button } from "../../../../react-common/components/controls/Button";
+import { Button as SuiButton } from "../../sui";
+import CodeValidationResult = pxt.tutorial.CodeValidationResult;
+
+interface TutorialValidationErrorMessageProps {
+  onContinueClicked: () => void;
```
jwunderl, line 7: nit: same as above, we use 4 space tabs~

webapp/src/components/tutorial/TutorialValidationErrorMessage.tsx
```diff
@@ -0,0 +1,81 @@
...
+    </div>
+  ) : undefined;
+
+  const showHintLabel = lf("Show Hint");
+  const continueAnywayLabel = lf("Continue Anyway");
+  const closeLabel = lf("Close");
+  return (
+    <div className="tutorial-validation-error-container">
+      {/* Using SuiButton to get close icon consistent with other popups */}
+      <SuiButton
```
jwunderl, line 43: can we just match modal and use normal react-common button here instead of sui? https://github.com/microsoft/pxt/blob/master/react-common/components/controls/Modal.tsx#L100

webapp/src/components/tutorialValidators.tsx
```diff
@@ -0,0 +1,171 @@
...
+        const {parent, tutorialOptions} = options;
+        const stepInfo = tutorialOptions.tutorialStepInfo
+            ? tutorialOptions.tutorialStepInfo[tutorialOptions.tutorialStep]
+            : null;
+        if (!stepInfo) return { isValid: true, hint: "" };
+
+        // Currently, this validator only supports useHintHighlight, so this flag/check is redundant,
+        // but there are plans to support manual specification of blocks in the future.
+        // As such, it felt prudent to still require this flag so tutorial authors don't have to go back and add it later on.
+        if (this.useHintHighlight) {
```
jwunderl, line 90: `useHintHighlight` does feel like a default case to me -- it might make sense to negate this so you opt out of hintHintlight instead of opt in when this validator is enabled so there's less 'standard' configuration needed?

jwunderl (commented):
Overall looks great, some spacing nits & suggestions

abchatra:
:medal_sports:

webapp/src/components/tutorialValidators.tsx
thsparks: @kiki-lee Thoughts?

jwunderl: just to clarify here, what I mean is that instead of having to have

````md
```validation.global
# BlocksExistValidator
* UseAnswerKeyHighlight: true
```
````

would just be

````md
```validation.global
# BlocksExistValidator
```
````

to turn it on, then when we add more features to BlocksExistValidator and you want to set specific blocks would be

````md
```validation.global
# BlocksExistValidator
* DoNotUseAnswerKeyHighlight: true
* CheckFor: spriteSay, myCoolBlock
```
````

Comment (thsparks) [reply webapp/src/components/tutorialValidators.tsx:90]: Makes sense to me, and since there's currently no reason to have this validator with the flag disabled, I'll just remove it altogether for now. We can add one later, once we have other ways to specify blocks to validate.

File: webapp/src/components/tutorial/TutorialContainer.tsx (hunk 6 above)
Comment (thsparks) [reply webapp/src/components/tutorial/TutorialContainer.tsx:122]: Much cleaner, thanks.

File: webapp/src/components/tutorialValidators.tsx (hunk 1 above)
Comment (thsparks) [reply webapp/src/components/tutorialValidators.tsx:117]: Checked with Kiki offline.

File: webapp/src/components/tutorial/TutorialContainer.tsx (hunk 5 above)
Comment (thsparks) [reply webapp/src/components/tutorial/TutorialContainer.tsx]: lol good catch. I had prettier set to 2 spaces and VS Code's default formatter set to 4 🙄

File: webapp/src/components/tutorial/TutorialContainer.tsx [hunk 10]
```diff
@@ -106,9 +109,75 @@ export function TutorialContainer(props: TutorialContainerProps) {
...
+            tutorial: tutorialId,
+            step: currentStep,
+        });
+    };
+
+    const handleStepCounterSetStep = (step: number) => {
+        // Only validate if we're moving to the next step.
+        // If going backwards, user could be trying to fix something from a previous step and this message could be annoying/confusing.
+        // If going forwards multiple steps, then the user has fully skipped entire steps, so validation doesn't make much sense.
+        if (step == currentStep+1) {
```
jwunderl, line 168:
```suggestion
        if (step == currentStep + 1) {
```

webapp/src/tutorial.tsx
```diff
@@ -97,16 +100,25 @@ function getUsedBlocksInternalAsync(code: string[], id: string, language?: strin
                             snippetBlocks[snippetHash][blk.type] = snippetBlocks[snippetHash][blk.type] + 1;
                             usedBlocks[blk.type] = 1;
                         }
+
+                        const comment = blk.getCommentText();
+                        if(comment && /@highlight/.test(comment)) {
+                            if(!highlightBlocks[snippetHash][blk.type]) {
```
jwunderl, line 106:
```suggestion
                        if (comment && /@highlight/.test(comment)) {
                            if (!highlightBlocks[snippetHash][blk.type]) {
```

jwunderl (approved):

abchatra:
If you merge this I will produce a build to test tomorrow.

Also possible to add tests for the markdown syntax? (Can be done post merge)

PR #9302, opened 2023-01-19
Don't use cryptic share code in UI string

eanders-ms:
Fixes https://github.com/microsoft/pxt-arcade/issues/5204


kimprice:
Is it too much work to include the game title?

eanders-ms:
It would require some reworking of how/when game metadata is fetched so that we're not fetching it multiple times. That puts it out of scope for this release, but would be a nice improvement. Will file an issue to track it.

kimprice (approved):

eanders-ms:
Logged https://github.com/microsoft/pxt-arcade/issues/5519

PR #9303, opened 2023-01-19
Confirmation should be rightmost button

eanders-ms:
Fixes https://github.com/microsoft/pxt-arcade/issues/5368


kimprice:
The issue states the "report abuse dialog" also needs to be switched.

eanders-ms:
Report abuse is in the share page. I think it's in a different repo? I will log a separate bug for that.

kimprice (approved):

eanders-ms:
Logged: https://github.com/microsoft/pxt-arcade/issues/5518

PR #9304, opened 2023-01-19
Put multiplayer debug output behind a flag

eanders-ms:
Fixes https://github.com/microsoft/pxt-arcade/issues/5334

### DIFF
--- a/multiplayer/src/epics/hostGameAsync.ts
+++ b/multiplayer/src/epics/hostGameAsync.ts
@@ -30,7 +30,7 @@ export async function hostGameAsync(shareCode: string | undefined) {
         dispatch(connectingToast);
 
         const hostResult = await gameClient.hostGameAsync(shareCode);
-        console.log(hostResult);
+        pxt.debug(hostResult);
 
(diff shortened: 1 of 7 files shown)


kimprice:
What's the difference between using `console.log` vs `pxt.log`? Do you need to have a flag to convert `pxt.log` to `console.log` like you do for debug?

What's the logic behind when to use `debug` vs `log`?

eanders-ms:
`pxt.log` routes to `console.log`, if it exists. Or pxt.log can be reassigned to a different method, something one might want to do if using pxtlib in a command line tool.

`pxt.debug` is the same, but behind a flag. Useful when you want to debug an issue a user is experiencing and want to get all the debug data that was useful during development. Rather than strip out those logs, they can be put behind pxt.debug.

Here are the default implementations: https://github.com/microsoft/pxt/blob/7d75808e5cb4731bbbd495fcfb5881bd761da35f/pxtlib/main.ts#L335

kimprice (approved):

jwunderl (approved):

PR #9305, opened 2023-01-19
When a multiplayer popup is expanded, clicking the button again should hide it

eanders-ms:
Fixes https://github.com/microsoft/pxt-arcade/issues/5313

### DIFF
--- a/multiplayer/src/hooks/useClickedOutside.ts
+++ b/multiplayer/src/hooks/useClickedOutside.ts
@@ -1,20 +1,22 @@
 import { useEffect, RefObject } from "react";
 
 export function useClickedOutside(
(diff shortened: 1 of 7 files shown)


multiplayer/src/hooks/useClickedOutside.ts
```diff
@@ -1,20 +1,22 @@
 import { useEffect, RefObject } from "react";
 
 export function useClickedOutside(
-    ref: RefObject<Element | undefined>,
+    refs: RefObject<Element | null>[],
```
eanders-ms, line 4: The essential piece is this change. We can provide additional refs considered inside the click area.

jwunderl (approved):

PR #9313, opened 2023-01-21
multiplayer: register both join handlers separately to avoid dropping msgs

jwunderl:
fix for https://github.com/microsoft/pxt-arcade/issues/5531

I suspect this is the issue we were hitting with late joiners not getting the initial screen causing screen compression / partial screen updates to fail

here's a build: https://arcade.makecode.com/app/7b49972cc23786e9b7696c213f27dd03ddef93e7-82f76e9ff6--multiplayer (remember to not use the join link it gives you directly as that won't point you to the right build)

### DIFF
--- a/multiplayer/src/services/gameClient.ts
+++ b/multiplayer/src/services/gameClient.ts
@@ -60,7 +60,7 @@ class GameClient {
     screen: Buffer | undefined;
     clientRole: ClientRole | undefined;
     gameOverReason: GameOverReason | undefined;
-    joinMessageInTimeHandler: ((a: any) => void) | undefined;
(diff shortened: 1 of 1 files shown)


jwunderl:
As a sidenote do you remember the bug you were seeing that required this? https://github.com/microsoft/pxt/blob/multiplayerDontDropJoiningMsgs/multiplayer/src/services/gameClient.ts#L192-L193 -- it feels a little like it might be related, I'm not seeing any problems in local serve / staging testing when dropping the settimeout there -- I'll test a bit more and produce a build with that change and then I can drop that too if we don't notice anything?

eanders-ms (approved):
Nice that seems to have fixed it! Excellent sleuthing. I wouldn't have suspected this code, as wonky as it is.

PR #9314, opened 2023-01-21
Fix Tutorial Next Button After Refresh

thsparks:
We were seeing an issue with code validation, where after refreshing the page, the Next button in a tutorial no longer worked. To fix it, I've removed the caching of tutorial validators on the TutorialInfo and TutorialStepInfo objects.

When a tutorial is reloaded, these objects get serialized and then deserialized, but the "execute" method is not able to be stored in the serialized format, so it's lost in the process. As a result, when we try to call it, we get an error that it doesn't exist.

This cache isn't saving us from much work right now anyway (validation construction is relatively cheap), so I've opted to simply remove it. If we decide we need it later on, we could cache it elsewhere (like on the TutorialContainer itself) and make sure it's cleared/refreshed as needed.

Fixes https://github.com/microsoft/pxt-arcade/issues/5533

### DIFF
--- a/webapp/src/components/tutorial/TutorialContainer.tsx
+++ b/webapp/src/components/tutorial/TutorialContainer.tsx
@@ -114,7 +114,7 @@ export function TutorialContainer(props: TutorialContainerProps) {
 
         currentStepInfo.localValidationConfig?.validatorsMetadata?.forEach(v => validators[v.validatorType] = GetValidator(v));
         tutorialOptions.globalValidationConfig?.validatorsMetadata?.forEach(v => {
-            if(!validators[v.validatorType]) {
+            if (!validators[v.validatorType]) {
                 validators[v.validatorType] = GetValidator(v);
             }
         })
--- a/webapp/src/components/tutorial/TutorialContainer.tsx
+++ b/webapp/src/components/tutorial/TutorialContainer.tsx
@@ -7,7 +7,7 @@ import { TutorialHint } from "./TutorialHint";
 import { TutorialResetCode } from "./TutorialResetCode";
 import { classList } from "../../../../react-common/components/util";
 import { TutorialValidationErrorMessage } from "./TutorialValidationErrorMessage";
-import { PopulateValidatorCache } from "../tutorialValidators";
+import { GetValidator } from "../tutorialValidators";
 import CodeValidator = pxt.tutorial.CodeValidator;
 import CodeValidationResult = pxt.tutorial.CodeValidationResult;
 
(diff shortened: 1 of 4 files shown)


jwunderl (approved):

webapp/src/components/tutorial/TutorialContainer.tsx
```diff
@@ -110,17 +110,14 @@ export function TutorialContainer(props: TutorialContainerProps) {
...
-        const localValidators =
-            currentStepInfo.localValidationConfig?.validators ??
-            PopulateValidatorCache(currentStepInfo.localValidationConfig);
-
-        let validators: pxt.Map<CodeValidator> = {
-            ...(globalValidators ?? {}),
-            ...(localValidators ?? {})
-        };
+        let validators: pxt.Map<CodeValidator> = {};
+
```
thsparks, line 114: If there's a more elegant way to populate `validators` here, I'm open to suggestions. I figured this way, we at least don't unnecessarily create global validators that are just going to be replaced by local ones, but the code itself seems a little messy.

jwunderl: Eh, I think it's fine here, think the alternatives are typically pretty ugly and it's mapping an array to an object anyways.

If you really wanted you could do something like `(currentStepInfo.localValidationConfig.validatorsMeta ?? []).concat(tutorialOptions.globalValidationConfig ?? [])` and keep the same `if (!validators[blah])`, and possible use an `Array.reduce`, but neither of those typically end up particularly pretty anyways

webapp/src/components/tutorial/TutorialContainer.tsx
```diff
@@ -110,17 +110,14 @@ export function TutorialContainer(props: TutorialContainerProps) {
...
+        currentStepInfo.localValidationConfig?.validatorsMetadata?.forEach(v => validators[v.validatorType] = GetValidator(v));
+        tutorialOptions.globalValidationConfig?.validatorsMetadata?.forEach(v => {
+            if(!validators[v.validatorType]) {
```
jwunderl, line 117:
```suggestion
            if (!validators[v.validatorType]) {
```
nit

thsparks: One day I'll remember to check for these before sending the PR :)

webapp/src/components/tutorial/TutorialContainer.tsx
thsparks: Hmm fair point. Suppose I'll leave it as-is for now, then. Thanks for the info!

webapp/src/components/tutorial/TutorialContainer.tsx
jwunderl: I mean really we should just automatically run prettier or something so we never have to think about formatting, just have to find a way to get prettier to ignore our hacky `lf` logic (or support lfs going to new lines in the cli). I'll try to find some free time to do it someday haha

PR #9320, opened 2023-01-23
add a flag to allow tutorials to show blocks collapsed in hints, template code

jwunderl:
e.g. so if you have a template section in a tutorial like the following it starts off collapsed in order to 'hide' it from view at the start (or if you want to show it exists in a hint, but not in detail)

No bug associated, just a very quick feature that came out of a chat with kiki~

image

Maybe we could even emit this so collapsed / non collapsed state persists through compile / decompile ;)

### DIFF
--- a/pxtblocks/blocklyimporter.ts
+++ b/pxtblocks/blocklyimporter.ts
@@ -40,18 +40,21 @@ namespace pxt.blocks {
             .filter(b => !!b.getCommentText())
             .forEach(b => {
                 const initialCommentText = b.getCommentText();
+                if (/@hide/.test(initialCommentText) && opts.applyHideMetaComment) {
+                    b.dispose(true);
+                    return;
+                }
+
                 let newCommentText = initialCommentText;
                 if (/@highlight/.test(newCommentText)) {
                     newCommentText = newCommentText.replace(/@highlight/g, '').trim();
                     (workspace as Blockly.WorkspaceSvg).highlightBlock?.(b.id, true)
                 }
-                if (/@hide/.test(newCommentText) && opts.applyHideMetaComment) {
-                    b.dispose(true);
-                }
                 if (/@collapsed/.test(newCommentText) && !b.getParent()) {
                     newCommentText = newCommentText.replace(/@collapsed/g, '').trim();
                     b.setCollapsed(true);
                 }
+
                 if (initialCommentText !== newCommentText) {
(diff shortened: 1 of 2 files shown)


kiki-lee:
This is going to be so helpful in trying to wean off of magic blocks.  Thanks so much, @jwunderl !

eanders-ms (approved):

pxtblocks/blocklyimporter.ts
```diff
@@ -35,21 +33,28 @@ namespace pxt.blocks {
...
-                    const cc = c.replace(/@highlight/g, '').trim();
-                    b.setCommentText(cc || null);
+                const initialCommentText = b.getCommentText();
+                let newCommentText = initialCommentText;
+                if (/@highlight/.test(newCommentText)) {
+                    newCommentText = newCommentText.replace(/@highlight/g, '').trim();
                     (workspace as Blockly.WorkspaceSvg).highlightBlock?.(b.id, true)
                 }
-                if (/@hide/.test(c)) {
+                if (/@hide/.test(newCommentText) && opts.applyHideMetaComment) {
```
srietkerk, line 48: Could there ever be a case where there is both the `@hide` comment and the `@collapsed` comment on the same block? Would this cause a problem for the `if collapsed` or the following condition if we dispose of the block?

jwunderl: There would never be a meaningful situation where they would both apply (as the end result of hide & collapsed would just be hide), but it's a fair point / I'll move the hide to the bottom just in case (it wouldn't ever throw but better safe than sorry / etc)

thsparks (approved):
Nice!

pxtblocks/blocklyimporter.ts
srietkerk: So hiding a block makes it so it doesn't make sense to apply the other flags because we get rid of it, am I understanding that correctly?

jwunderl: hiding a block is basicaly equivalent to just deleting it (in fact, that's exactly what the code is doing with the call to dispose, and why it's only enabled with the flag when the blocks are being decompiled for screenshots) - so highlighting, collapsing, or really any other mutation to a block that is going to be deleted doesn't make too much sense, though we might come up with a reason to fiddle with it when we add more flags in the future :)

srietkerk (approved):
Looks good!

PR #9322, opened 2023-01-25
keep comments on blocks when loading blocks for tutorial info

jwunderl:
Last pr made the comments get stripped when decompiling (which was the old behavior / only went away as a sideeffect of the small 'hide blocks' pr back in dec).

Headless workspaces actually don't have a concept of highlighting (`(workspace as Blockly.WorkspaceSvg).highlightBlock?.(b.id, true)` in blocklyimporter bails out on the `?.`), so we have to persist the comments when loading up the workspace so we can read them again when getting usedblocks: https://github.com/microsoft/pxt/blob/master/webapp/src/tutorial.tsx#L104 .


thsparks (approved):
LGTM! Thanks!

PR #9327, opened 2023-01-25
Make Block Errors Easier to Read

thsparks:
I think white is a safe bet, since I believe the bubble's background is chosen based on the block color and it looks like we use white text for all the regular block text, too.

Before:
image

After:
image

Fixes https://github.com/microsoft/pxt-arcade/issues/4993


eanders-ms (approved):
nice.

jwunderl (approved):

PR #9329, opened 2023-01-26
Select active player in multiplayer mode

jwunderl:
Close https://github.com/microsoft/pxt-arcade/issues/5469

Adds a multiplayer bar underneath simulator when you're writing a multiplayer game so you can pretend to be someone else / preview what presence icons look like

image

Because there is lacking space in tutorials, I made a quick decision for now to just show it to the left

image

There was a little bit that I'm gonna try and rearrange, but I already rewrote where it was rendered (out of iframe to accommodate tutorials better) & where the data was stored (treating it similar to console logs / etc ends up way too clunky just to propagate a few images, and don't want to send out significant, root of react tree updates up to 120s and second and slow full editor perf). The current semi hacky version of latching it as a style var on root ends up feeling okay to me, but I'm gonna see about adding a few message listeners in `SimulatorPresenceBar.tsx` to keep track of the images there nicely. (I'll see how that turns out and fix the typings todo I mentioned at the same time)

relies on https://github.com/microsoft/pxt-arcade-sim/pull/85 (well, not strictly relies on, but the buttons won't do anything if it's not there)

build: https://arcade.makecode.com/app/0c63e39b191ab5a18f69ed9bef12099bde2e8db7-ebce63eee6#pub:_cmhaDMf2XHX6

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -251,6 +251,15 @@ export class ProjectView
             }
             else if (msg.type === "thumbnail") {
                 if (this.shareEditor) this.shareEditor.setThumbnailFrames((msg as pxsim.SimulatorAutomaticThumbnailMessage).frames);
+            } else if (msg.type === "multiplayer") {
+                // todo move over types from multiplayer app to pxsim/embed.ts
+                const multiplayerMessage = msg as any;
+                if (multiplayerMessage.content === "Icon"
+                    && multiplayerMessage.iconType === 0 /** IconType.Player **/
+                ) {
+                    const { palette, icon, slot } = multiplayerMessage;
+                    this.handleSetPresenceIcon(slot, palette, icon.data);
+                }
             }
         }, false);
     }
@@ -314,6 +323,36 @@ export class ProjectView
             core.handleNetworkError(e);
         } finally {
             core.hideLoading("addextensions")
+        }
+    }
+
+    handleSetPresenceIcon = (slot: number, palette?: Uint8Array, icon?: Uint8Array) => {
+        slot = slot | 0;
+        if (slot < 1 || slot > 4)
+            return;
+
+        const root = document.getElementById("root");
+
+        const slotCssVarName = `--multiplayer-presence-icon-${slot}`;
+
+        if (!palette || !icon) {
+            root.style.removeProperty(slotCssVarName);
+            return;
+        } else {
+            const iconPngDataUri = pxt.convertUint8BufferToPngUri(
+                palette,
+                icon,
+            );
+            if (!iconPngDataUri) {
+                // failed to parse icon, bail out
+                return;
+            }
+            const asUrl = `url("${iconPngDataUri}")`;
+            root.style.setProperty(
+                slotCssVarName,
+                asUrl
+            );
+            // TODO: set with `content: url(var(--multiplayer-presence-icon-1))` in css later
         }
     }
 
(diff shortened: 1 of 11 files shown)


webapp/src/app.tsx
```diff
@@ -251,6 +251,15 @@ export class ProjectView
             }
             else if (msg.type === "thumbnail") {
                 if (this.shareEditor) this.shareEditor.setThumbnailFrames((msg as pxsim.SimulatorAutomaticThumbnailMessage).frames);
+            } else if (msg.type === "multiplayer") {
+                // todo move over types from multiplayer app to pxsim/embed.ts
+                const multiplayerMessage = msg as any;
```
jwunderl, line 256: (noting here that I'll be trying to move this over to `simulatorPresenceBar` without losing nice updates / clearing on restart in a followup pr)

eanders-ms:
This is awesome!

Just a usability question, would it be possible to focus the simulator after clicking on a player slot? Right now when I click a player, it unfocuses the sim and I have to move the mouse back over it to continue playing the game with the keyboard.

eanders-ms:
> This is awesome!
> 
> Just a usability question, would it be possible to focus the simulator after clicking on a player slot? Right now when I click a player, it unfocuses the sim and I have to move the mouse back over it to continue playing the game with the keyboard.

Though, I wouldn't want this if it breaks tab navigation.

jwunderl:
I think that makes sense to do, you are clicking a button that is 'part' of the simulator experience and the general intention is that you want to interact with the simulator as that player.

webapp/src/app.tsx
```diff
@@ -4887,7 +4925,8 @@ export class ProjectView
             this.editor != this.blocksEditor ? "editorlang-text" : "",
             this.editor == this.textEditor && this.state.errorListState,
             'full-abs',
-            pxt.appTarget.appTheme.embeddedTutorial ? "tutorial-embed" : ""
+            pxt.appTarget.appTheme.embeddedTutorial ? "tutorial-embed" : "",
+            // isMultiplayerGame ? "multiplayer-preview" : "",
```
eanders-ms, line 4929: needed?

eanders-ms (approved):
Looks amazing.

webapp/src/app.tsx
jwunderl: Ah, good catch, I had set this class for just a minute while making fullscreensim layout correctly but was able to move it down to just be on sidepanel, will remove

jwunderl:
Here's a build with the latest commit that makes focus transition over to sim when you change characters, feels a lot better to me: https://arcade.makecode.com/app/db769e73b2b39c139fe3e810c39835b546914d3b-39eefacf5a

thsparks:
While playing around with the build, I noticed the colors don't appear on the d-pad for players 2 - 4 when movinga round with WASD. Is that known / by design?

jwunderl:
> While playing around with the build, I noticed the colors don't appear on the d-pad for players 2 - 4 when moving around with WASD. Is that known / by design?

Huh, weird css bug, thanks for pointing out / will fix it (though maybe in follow up pr if it takes more than a minute)

jwunderl:
Okay, fixed that dpad visual not updating issue in https://arcade.makecode.com/app/1087befeeee68a6cfb227218ade8906eae9a862e-3b3dc5507a, [this line in arcade sim](https://github.com/microsoft/pxt-arcade-sim/pull/85/commits/29a95db76acb3d5c9a7c9985665a9f64cba64dfa) was expecting one of the face buttons to update correctly so just had to pass in the un-adjusted key id there

thsparks (approved):

PR #9331, opened 2023-01-27
Better drum sounds and add support for bass clef to drums

riknoll:
This improves the array of drum sounds that are available.

If you want to hear them, here's an arcade project that has 'em all:

https://arcade.makecode.com/beta#pub:52073-79957-62825-57855

Also it makes it so that you can add drum sounds to the bass clef, so 24 sounds are now available!

### DIFF
--- a/pxtlib/music.ts
+++ b/pxtlib/music.ts
@@ -1020,7 +1020,8 @@ namespace pxt.assets.music {
                         }
                     },
                     drums: [
-                        { /* kick drum */
+                        {
+                            /* neutral kick */
                             startFrequency: 100,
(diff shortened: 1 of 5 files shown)


riknoll:
oh also it changes the lemon sound so that it's no longer the same as the computer sound.

jwunderl (approved):
I'm just believing in you on the `pxtlib/music.ts` side but the rest LGTM

eanders-ms (approved):
We need another tech talk on this :)

riknoll:
@eanders-ms i'm always happy to talk about sound synthesis 😄

PR #9332, opened 2023-01-27
hook up player join / leave events to multiplayer

jwunderl:
Hook up player events connected / disconnected events, see https://github.com/microsoft/pxt-common-packages/pull/1407 for common packages event registration


eanders-ms (approved):
W00t!

PR #9333, opened 2023-01-27
Expand bass clef for drums as well

riknoll:
Fixing a little bug I just noticed. The bass clef wasn't automatically showing when there were notes in the bass clef for drums


jwunderl (approved):

abchatra:
Port to stable.

PR #9334, opened 2023-01-27
Fix serviceworker race condition

riknoll:
Finally fixes the problem where the serviceworker isn't reloading; I think at some point the initialization of the load event was moved below the auth code initialization, which was causing it to miss the window's load event

This puts us in a bit of a pickle.... Since the old "bad" payload was released in the last release there is no way I know of to force all of the existing clients to unregister the old serviceworker. As a result, nobody will get the updated payload from this release. That means everyone will be stuck on an old version of arcade unless they clear their browser storage (bad idea) or manually unregister the serviceworker (multi-step-process that involves the devtools). Not sure what to do here.

@abchatra it is ABSOLUTELY CRITICAL that this gets merged before release

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -5545,6 +5545,9 @@ document.addEventListener("DOMContentLoaded", async () => {
     }
     pxt.perf.measureStart("setAppTarget");
     pkg.setupAppTarget((window as any).pxtTargetBundle);
+
+    // DO NOT put any async code before this line! The serviceworker must be initialized before
+    // the window load event fires
     appcache.init(() => theEditor.reloadEditor());
     pxt.setBundledApiInfo((window as any).pxtTargetBundle.apiInfo);
     pxt.perf.measureEnd("setAppTarget");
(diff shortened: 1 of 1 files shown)


eanders-ms:
Looks like it's been in the wrong place for multiple releases. Things will start to get better after this release.

webapp/src/app.tsx
```diff
@@ -5545,6 +5545,7 @@ document.addEventListener("DOMContentLoaded", async () => {
     }
     pxt.perf.measureStart("setAppTarget");
     pkg.setupAppTarget((window as any).pxtTargetBundle);
+    appcache.init(() => theEditor.reloadEditor());
```
eanders-ms, line 5548: Worth putting a comment here explaining the reason this needs to happen in this location. Something like `// don't place code above this line unless you know why it needs to happen before app refresh`

eanders-ms (approved):

jwunderl (approved):

webapp/src/app.tsx
jwunderl: Probably worth tagging this pr explicitly to give a trail in case it gets moved / git blame changes it for some reason too

abchatra:
I think if folks close the browser and reopen it will load the latest version. It is only if they never close the browser.

riknoll:
@abchatra no, that's not how it works. the serviceworker will only get reloaded if they window load event happens to run  before the auth initialization works. otherwise, they'll have to wait for the browser cache to expire

PR #9336, opened 2023-01-27
propagate presence on host resuming pause game

jwunderl:
since messages can be dropped while sim is paused, propagate current user state.

(we only fire events when connected state changes https://github.com/microsoft/pxt-common-packages/blob/sendConnectionEventsOnScreenPop/libs/game/controller.ts#L257 so no worries on the event coming in twice)

### DIFF
--- a/multiplayer/src/epics/resumeGameAsync.ts
+++ b/multiplayer/src/epics/resumeGameAsync.ts
@@ -2,9 +2,10 @@ import { state, dispatch } from "../state";
 import * as gameClient from "../services/gameClient";
 import { setGamePaused } from "../state/actions";
 import { simDriver } from "../services/simHost";
+import { SimMultiplayer } from "../types";
 
 export async function resumeGameAsync() {
-    const { clientRole } = state;
+    const { clientRole, presence } = state;
 
     try {
         if (clientRole === "host") {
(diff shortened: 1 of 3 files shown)


multiplayer/src/epics/resumeGameAsync.ts
```diff
@@ -13,6 +14,18 @@ export async function resumeGameAsync() {
...
+            for (let i = 1; i <= 4; i++) {
+                const nowConnected = !!presence.users.find(el => el.slot === i);
+                simDriver()?.postMessage({
+                    type: "multiplayer",
+                    content: "Connection",
+                    slot: i,
+                    connected: nowConnected,
+                } as SimMultiplayer.MultiplayerConnectionMessage);
+            }
+            await gameClient.sendCurrentScreenAsync();
```
jwunderl, line 27: this is a quick 'side' fix, I noticed that the bug I was seeing before with xor diffs was occurring once in a while when switching tabs caused it to pause -- maybe the backend is rejecting a posted diff as it's already in the pause state or something. This felt like a pretty safe / reasonable place to put in a 'make sure everyone is synced now that we're resuming' either way

eanders-ms (approved):
Coming in hot!

PR #9341, opened 2023-01-28
Kiosk share: input strings converted to upper case

srietkerk:
I noticed that if I tried to type in a kiosk id not in upper case, I would get a "Kiosk code not found" error. If I typed the same code in with upper case, it would successfully submit. It shouldn't matter if the user types in upper or lowercase, if it is the right sequence of characters, it should be a successful submit. 

image


jwunderl (approved):
Maybe better fixed in backend to ignore case (/ coerce to uppercase) on the endpoint? Doesnt

PR #9348, opened 2023-01-30
Fix Answer Key Causing Code Validation to Detect Error Incorrectly

thsparks:
Fix Answer Key Causing Code Validation to Detect Error Incorrectly (#9346)

Not 100% consistent, but sometimes, if the user opens the answer key and then tries to proceed to the next step, validation will flag an error, even if the user has the correct code. This happens because, in this scenario, the call to Blockly.getMainWorkspace() sometimes doesn't return the user's actual workspace. Instead, it returns a workspace with no blocks in it.

To fix this, I've gotten rid of the call to Blockly.getMainWorkspace and instead grab it directly from the ProjectView editor (using an existing getBlocksEditor() function)

### DIFF
--- a/webapp/src/components/tutorialValidators.tsx
+++ b/webapp/src/components/tutorialValidators.tsx
@@ -1,7 +1,7 @@
 import { MarkedContent } from "../marked";
+import { getBlocksEditor } from "../app";
 import TutorialOptions = pxt.tutorial.TutorialOptions;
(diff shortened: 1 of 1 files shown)


thsparks:
Cherry pick of https://github.com/microsoft/pxt/commit/a043c2cf41645aa09f5a2b7ac561c6130dee5af1

abchatra (approved):

PR #9349, opened 2023-01-30
Use SVG File for Code Validation Ladybug

thsparks:
Hopefully this will look less scary on mac.

The image comes from https://github.com/microsoft/fluentui-emoji/blob/main/assets/Lady%20beetle/Flat/lady_beetle_flat.svg and I ran it through svgo.

With the changes:
image

For reference, here's what it looked like on iOS before:
image

Not totally sure if I put the svg file in the right place. Feel free to suggest alternatives.


jwunderl (approved):
That is certainly much less horrifying!

Actually since it's an svg you might consider just inlining it into a component (e.g. like this: https://github.com/microsoft/pxt/blob/master/multiplayer/src/components/icons/UserIcon.tsx#L17) but should be fine either way

srietkerk (approved):

thsparks:
> That is certainly much less horrifying!
> 
> Actually since it's an svg you might consider just inlining it into a component (e.g. like this: https://github.com/microsoft/pxt/blob/master/multiplayer/src/components/icons/UserIcon.tsx#L17) but should be fine either way

Thanks for the heads up. I'm going to check it in as-is for now so this can make it to the hotfix. If we find this is causing trouble later on, we can inline it then.

PR #9351, opened 2023-01-31
send connection message when user selects presence icon

jwunderl:
requires https://github.com/microsoft/pxt-arcade-sim/pull/87, https://github.com/microsoft/pxt-common-packages/pull/1422

sends connected state re: https://github.com/microsoft/pxt-arcade/issues/5625. Also got the types for multiplayer messages in pxsim itself -- I'll dedup them with multiplayer app and pxt-common-packages separately as there are some slightly tedious null / undefined differences that will take up a lot more swaps.

This one just adds the connect message sending, as I think we still need to talk more on how to handle testing disconnection -- but implementing it in a way that makes it hard to test player 2 interacting with player 3 feels wrong / kindly handling disconnection & reconnection are much more 'advanced' scenarios than typical multiplayer stuff

quick test https://arcade.makecode.com/app/06c407421e523ff12f8d7cfeb171a147bd3dc9fb-fc9250afff#pub:_9KpLiMCxoFmP

### DIFF
--- a/webapp/src/components/SimulatorPresenceBar.tsx
+++ b/webapp/src/components/SimulatorPresenceBar.tsx
@@ -19,7 +19,14 @@ function PlayerPresenceIcon(props: React.PropsWithoutRef<{slot: 1 | 2 | 3 | 4}>)
             type: "setactiveplayer",
             playerNumber: slot,
         };
+        const connectionMsg: pxsim.multiplayer.ConnectionMessage = {
+            type: "multiplayer",
+            content: "Connection",
+            slot: slot,
+            connected: true,
+        };
         simulator.driver.postMessage(setSlotMsg);
+        simulator.driver.postMessage(connectionMsg);
         simulator.driver.focus();
     }
     return (<Button
--- a/pxtsim/embed.ts
+++ b/pxtsim/embed.ts
@@ -274,6 +274,60 @@ namespace pxsim {
         color: string;
     }
 
+    export namespace multiplayer {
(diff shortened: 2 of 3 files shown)


webapp/src/components/SimulatorPresenceBar.tsx
```diff
@@ -19,7 +19,14 @@ function PlayerPresenceIcon(props: React.PropsWithoutRef<{slot: 1 | 2 | 3 | 4}>)
...
             playerNumber: slot,
         };
+        const connectionMsg: pxsim.multiplayer.ConnectionMessage = {
+            type: "multiplayer",
+            content: "Connection",
+            slot: slot,
+            connected: true,
+        };
         simulator.driver.postMessage(setSlotMsg);
+        simulator.driver.postMessage(connectionMsg);
```
eanders-ms, line 29: If the player is already connected then I don't think we should raise the event again. Connects and disconnects should be reciprocal (like React's mount/unmount, etc.)

jwunderl: That's handled in the controller connection -- the game engine listens for this message and sets https://github.com/microsoft/pxt-common-packages/blob/master/libs/game/controller.ts#L256, which raises the controller event only when it has actually changed

eanders-ms: Oh gotcha! Thanks.

eanders-ms (approved):

PR #9352, opened 2023-01-31
multiplayer: responsive layout improvements

eanders-ms:
Made targeted fixes to the multiplayer header bar so that the org logo and profile pic don't get squished on mobile.

image
image

I would have liked to tighten up the spacing between menu items, but doing so would require changes in react-common, and the effects would be felt in webapp, skillmap, etc. We can tackle this in another pass.

Fixes https://github.com/microsoft/pxt-arcade/issues/5372


thsparks (approved):

jwunderl (approved):
LGTM, though tagged in design folks just in case there's any concern with having text next to logo

eanders-ms:
@unthinkmedia, @clkantner merging this now, but please let me know if you have feedback or need anything changed.

PR #9357, opened 2023-02-01
Add open in vs code button

riknoll:
Currently does nothing! Only shows up in JavaScript.

image

### DIFF
--- a/pxteditor/editor.ts
+++ b/pxteditor/editor.ts
@@ -354,7 +354,7 @@ namespace pxt.editor {
 
         showReportAbuse(): void;
         showLanguagePicker(): void;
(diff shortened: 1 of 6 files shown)


riknoll:
also, ideally this will have a vscode icon but i didn't have time to add one! the visual studio code brand guidelines also mention that we should use the full name "visual studio code" when referring to it for the first time on the page, but it made the button too wide so i used the approved abbreviation. if we want to adhere to those guidelines, we will likely need to reconsider this design.

jwunderl (approved):

eanders-ms (approved):
Would it be simpler to add it to the gear menu?

abchatra:
I will merge this for now. We can change this later.

PR #9359, opened 2023-02-01
Documentation for Code Validation in Tutorials

thsparks:
Figured I'd throw something together quickly, just so we have this information down somewhere. If we add more validators/properties in the future, it may be worth pulling this out of "Basics" and into its own file, but I didn't think that was necessary right out of the gate. I don't feel strongly, though, so if people would prefer to put it in its own section, let me know.

### DIFF
--- a/docs/writing-docs/tutorials/basics.md
+++ b/docs/writing-docs/tutorials/basics.md
@@ -277,7 +277,7 @@ Within a validation section, you may specify which validators you want to enable
 
 Currently, only one validator exists: the `BlocksExistValidator`. This validator looks at highlighted blocks in the answer key and confirms that, for each highlighted block, the user's code contains at least one block of the same type. It does *not* validate the parameters passed into the block.
 
-The only property currently available on the `BlocksExistValidator` is `Enabled`, which determines whether or not the validator runs.
+The only property currently available on the `BlocksExistValidator` is `Enabled`, which determines whether or not the validator runs. This is `true` by default whenever you specify the validator but can be set to `false` if you wish to disable it on a single step.
 
 ### Examples
 **Enable the `BlocksExistValidator` globally**
--- a/docs/writing-docs/tutorials/basics.md
+++ b/docs/writing-docs/tutorials/basics.md
@@ -269,7 +269,7 @@ If you don't see an error message, try adding the `dbg=1` URL parameter and relo
 
 ## Code Validation (`validation.local` and `validation.global` sections)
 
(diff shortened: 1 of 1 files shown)


docs/writing-docs/tutorials/basics.md
```diff
@@ -267,6 +267,33 @@ Gray blocks will be generated when not all datatypes are known (e.g. `mySprite`
...
 
+## Code Validation (`validation.local` and `validation.global` sections)
+
+If you want to enable code validation in your tutorial, you can do so by adding a validation section. As with [`blockconfig` above](#reconfiguring-blocks-in-the-toolbox-blockconfiglocal-and-blockconfigglobal-sections), there are two types of sections: `validation.global` and `validation.local`. Global can be anywhere in the markdown and applies to all steps in the tutorial. Local must appear within a specific step, and applies only to that step. Local takes precedence over global.
+
+Within a validation section, you may specify which validators you want to enable and properties for those validators using the same syntax we use for writing Skillmaps (see [skillmap structure](..\skillmaps.md#skillmap-structure)).
+
+### Validators
+
+Currently, only one validator exists: the `BlocksExistValidator`. This validator looks at highlighted blocks in the answer key and confirms that, for each highlighted block, the user's code contains at least one block of the same type. It does *not* validate the parameters passed into the block.
```
thsparks, line 278: I couldn't find any docs on highlights in the answer key. Do we have those? If so, I'd like to link to them here.

kiki-lee: I don't know if we do.  We should probably add //@highlight (and if it does exist, update to explain that you can now highlight multiple blocks). 

Related, do we have documentation on @hide and @collapse?

docs/writing-docs/tutorials/basics.md
```diff
@@ -267,6 +267,33 @@ Gray blocks will be generated when not all datatypes are known (e.g. `mySprite`
...
+
+The only property currently available on the `BlocksExistValidator` is `Enabled`, which determines whether or not the validator runs.
+
+### Examples
+**Enable the `BlocksExistValidator` globally**
+````
+```validation.global
+# BlocksExistValidator
```
kimprice, line 286: Do you need to include `Enabled: true` or is that set by default?

thsparks: Set by default. I updated the docs to make that clearer, thanks!

docs/writing-docs/tutorials/basics.md
jwunderl: looks like highlight is mentioned here https://makecode.com/writing-docs/snippets#highlight, I should add hide /collapse there as well

jwunderl (approved):

kimprice (approved):

docs/writing-docs/tutorials/basics.md
thsparks: Nice, thanks! I've added a link to that section.

PR #9361, opened 2023-02-02
[Stable 8.5] Kiosk share: input strings converted to upper case (#9341)

srietkerk:


react-common/components/share/ShareInfo.tsx
```diff
@@ -112,8 +112,9 @@ export const ShareInfo = (props: ShareInfoProps) => {
     const handleKioskSubmitClick = async () => {
         const gameId = pxt.Cloud.parseScriptId(shareData.url);
         if (kioskInputRef?.value) {
-            const validKioskId = /^[a-zA-Z0-9]{6}$/.exec(kioskInputRef.value)?.[0];
+            let validKioskId = /^[a-zA-Z0-9]{6}$/.exec(kioskInputRef.value)?.[0];
```
eanders-ms, line 115: Just for reference, here's how multiplayer cleans up join codes:
```ts
    joinCode = joinCode.toUpperCase().replace(/[^A-Z0-9]/g, "");
    if (joinCode.length !== 6) return undefined;
```
This handles trimming the string and removal of things like dashes and spaces in a user-friendly way.

PR #9367, opened 2023-02-03
Option to fill in name for skillmap completion certificates

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/3784

Here's a starter build: https://arcade.makecode.com/app/5f2cbe11bf76ab229a92909763a0dd91e5caeb2e-dbc7b4a85a--skillmap?debugCompleted=1#mole

completion-cert-choosing

need to upload a test file in pxt-arcade with a name entry form field set in the pdf to get the right bounding box (kiki said it would be very quick to update all the existing ones) as right now it is just falling back to the existing behavior of 'make put the name just below the mid area where most signatures go', so that part is partially untested but basically just mirrored over from my test repl https://whitesmokerustyyottabyte.jwunderl.repl.co/

If you're signed in, it fills in first name by default, otherwise it's empty. Either way you can type in there, and it does a quick size check to find max-ish size we can render without going out of the bounding box for field entry.


webapp/public/index.html
```diff
@@ -108,7 +108,32 @@
...
+            "music-editor/fish.png": "/blb/music-editor/fish.png",
+            "music-editor/ice-cream.png": "/blb/music-editor/ice-cream.png",
+            "music-editor/lemon.png": "/blb/music-editor/lemon.png",
+            "music-editor/snake.png": "/blb/music-editor/snake.png",
+            "music-editor/star.png": "/blb/music-editor/star.png",
+            "music-editor/strawberry.png": "/blb/music-editor/strawberry.png",
+            "music-editor/taco.png": "/blb/music-editor/taco.png",
+            "music-editor/bass-clef.svg": "/blb/music-editor/bass-clef.svg",
+            "music-editor/treble-clef.svg": "/blb/music-editor/treble-clef.svg",
+>>>>>>> bc1f32ada9dfe758a4b900efbab595986fae3534
```
kimprice, line 136: Looks like you need to clean up this merge.

webapp/public/pdf-lib/pdf-lib.min.js
```diff
@@ -0,0 +1,15 @@
+!function(t,e){"object"==typeof exports&&"undefined"!=typeof module?e(exports):"function"==typeof define&&define.amd?define(["exports"],e):e((t=t||self).PDFLib={})}(this,(function(t){"use strict";
+/*! *****************************************************************************
+    Copyright (c) Microsoft Corporation. All rights reserved.
+    Licensed under the Apache License, Version 2.0 (the "License"); you may not use
```
thsparks, line 4: Just double checking: Apache License 2.0 is okay? Weirdly the github for pdf-lib says MIT license but this indicates a different one. A quick google makes it seem like they're pretty comparable, but thought it might be worth calling out, just in case.

Also, do we need a LICENSE file? I see that in some of the other dependencies we've pulled into this location, but not sure if it's actually required.

skillmap/src/components/AppModal.tsx
```diff
@@ -487,13 +539,34 @@ export class AppModalImpl extends React.Component<AppModalProps, AppModalState>
...
+                <Input
+                    className="cert-name-input"
+                    placeholder={"Put your name on it!"}
+                    label={lf("Enter your name")}
+                    type="text"
+                    initialValue={firstName || undefined}
+                    handleInputRef={handleInputRef}
+                    preserveValueOnBlur={true}
+                    onEnterKey={onCertificateClick}
+                    title={lf("Enter a custom name to put on the certificate")}
```
thsparks, line 565:
```suggestion
                    title={lf("Enter your name to customize the certificate")}
```

"Custom name" sounded a bit weird to me, but I don't feel strongly either way :)

skillmap/src/components/AppModal.tsx
```diff
@@ -487,13 +539,34 @@ export class AppModalImpl extends React.Component<AppModalProps, AppModalState>
...
+                    ariaLabel={lf("Enter a custom name to put on the certificate")}
```
thsparks, line 566:
```suggestion
                    ariaLabel={lf("Enter your name to customize the certificate")}
```

thsparks (approved):
A few notes, nothing blocking. Neat feature!

webapp/public/pdf-lib/pdf-lib.min.js
jwunderl: Hm I didn't copy it over as I didn't see it in zip-js, but I'll throw one in if we want?

Apache's fine, the reason this is different than pdf-lib is because this is the declaration for the dependency of zip-js (notice it's for a msft dependency in the first place, where pdf-lib is from a github employee https://github.com/Hopding/pdf-lib ~)

skillmap/src/components/AppModal.tsx
jwunderl: Woops I meant to do a pass on the wording on these and totally forgot, thanks!

kimprice (approved):

webapp/public/pdf-lib/pdf-lib.min.js
thsparks: Eh, no worries if we don't need the LICENSE file. Sounds like a non-issue. Thanks for the info!

eanders-ms (approved):
LGTM

jwunderl:
Okay, and here's the test version of shark that's using the `name-entry` input field: https://arcade.makecode.com/app/5f2cbe11bf76ab229a92909763a0dd91e5caeb2e-dbc7b4a85a--skillmap?debugCompleted=1#docs:/test/skillmap/shark-with-input from https://github.com/microsoft/pxt-arcade/pull/5658

PR #9369, opened 2023-02-03
Add description on how to use hide, collapsed

jwunderl:
re: https://github.com/microsoft/pxt/pull/9359#discussion_r1093739108

### DIFF
--- a/docs/writing-docs/snippets.md
+++ b/docs/writing-docs/snippets.md
@@ -49,7 +49,7 @@ This will appear in tutorials as the `variables` category, but clicking it will 
 
 ### highlight
 
-Although used in snippets, the renderer will higlight the next line of code or block following a comment containing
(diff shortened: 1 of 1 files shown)


docs/writing-docs/snippets.md
```diff
@@ -61,6 +61,60 @@ console.log(":(")
...
+mySprite.x += 50;
+myCoolFunction();
+```
+````
+
+### collapsed
+
+When used in snippets, the following function or event will show up collapsed in the hint. This
+can also be used in template code in order to have code start off collapsed in the user's workspace,
+in case there is code that should exist but not be focused on immediately / that you will call out a few steps in.
```
thsparks, line 90: nit:

```suggestion
in case there is code that should exist but not be focused on immediately, or that you will call out in subsequent steps.
```

thsparks (approved):
👍

PR #9377, opened 2023-02-07
add default .pxt git ignore, set excludeGitIgnore setting

jwunderl:
ignore generated .pxt files, and default 'on' for the `explorer.excludeGitIgnore` setting which makes explorer not show built files can drop that second one if we don't like that or think it's potentially confusing? I guess the thing to note is that it'll hide pxt_modules unless you e.g. go to definition which might be controversial:

image

test repo: https://github.com/jwunderl/ignore-me-testing-2
image

### DIFF
--- a/pxtlib/initprj.ts
+++ b/pxtlib/initprj.ts
@@ -80,6 +80,7 @@ node_modules
 yotta_modules
 yotta_targets
 pxt_modules
+.pxt
 _site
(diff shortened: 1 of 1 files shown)


abchatra:
I like hiding pxt-modules

abchatra (approved):

PR #9382, opened 2023-02-09
Add option to change language in Skillmap

thsparks:
This adds an option to the skillmap gear menu to change the language. The UI is based off of the main webapp's UI, but it's been reimplemented in react-common, but for now, I have not added the "Translate the Editor" button:

image

image

Eventually, it would be good to consolidate to a single language selector implementation, but I've just used it in skillmap for now. Once we've built confidence in it, we can go through and replace the other implementations with the react-common one. (Follow up task filed, see below.)

Link to test: https://arcade.makecode.com/app/297ee0520765e0b30cea2e7043c492983b7c9352-108cd3ee03--skillmap

Links to follow-up items:
Show language option when an activity is open: https://github.com/microsoft/pxt-arcade/issues/5682
Unify language selectors: https://github.com/microsoft/pxt-arcade/issues/5684

Fixes https://github.com/microsoft/pxt-arcade/issues/5427


react-common/components/language/LanguageSelector.tsx
```diff
@@ -0,0 +1,79 @@
...
+                                langId={langId}
+                                name={lang.localizedName}
+                                ariaLabel={lang.englishName}
+                                description={lang.englishName}
+                                onClick={this.changeLanguage}/>
+                        );
+                    })}
+                </div>
+                {targetTheme.crowdinProject ? (
+                    <Link
```
thsparks, line 69: I've noticed I cannot tab to this link using the keyboard. I didn't want to block the PR on it, but if anyone knows what might cause that, I'm certainly open to recommendations.

eanders-ms (approved):
Looks good!

react-common/components/language/LanguageSelector.tsx
thsparks: Fixed this by setting tabindex explicitly. Apparently the Modal has some custom handling to keep focus, and that was tripping up the usual behavior. Tabindex seems to fix it.

jwunderl (approved):

react-common/components/language/LanguageSelector.tsx
```diff
@@ -0,0 +1,79 @@
...
+}
+
+export class LanguageSelector extends React.Component<LanguageSelectorProps> {
+    constructor(props: LanguageSelectorProps) {
+        super(props);
+        this.changeLanguage = this.changeLanguage.bind(this);
+    }
+
+    languageList(): string[] {
+        if (pxt.appTarget.appTheme.selectLanguage && pxt.appTarget.appTheme.availableLocales && pxt.appTarget.appTheme.availableLocales.length) {
```
kimprice, line 23: nit:
```suggestion
        if (pxt.appTarget.appTheme.selectLanguage && pxt.appTarget.appTheme.availableLocales?.length) {
```

react-common/components/language/LanguageSelector.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { Link } from "../controls/Link";
+import { Modal } from "../controls/Modal";
+import { LanguageCard } from "./LanguageCard";
+
+
+const defaultLanguages = ["en"];
+
+interface LanguageSelectorProps {
+    visible: boolean;
+    activityOpen: boolean;
```
kimprice, line 11: Do you need this `activityOpen` prop? I don't see it being used.

skillmap/src/App.tsx
```diff
@@ -407,6 +419,12 @@ class AppImpl extends React.Component<AppProps, AppState> {
                 <MakeCodeFrame forcelang={forcelang} onWorkspaceReady={this.onMakeCodeFrameLoaded}/>
                 <AppModal />
                 <UserProfile />
+                <LanguageSelector
+                    visible={this.props.showSelectLanguage}
```
kimprice, line 423: I'm wondering why we would pass `visible` as a prop. Why not make this a conditional div instead? 
Like `{ this.props.showSelectLanguage && <LanguageSelector`

react-common/components/language/LanguageSelector.tsx
thsparks: Ah good point, leftover from my various experiments. I'll remove it.

skillmap/src/App.tsx
thsparks: tl;dr: because that's what UserProfile does and I wanted to be consistent, but I'm open to switching it if you feel strongly about doing it the other way

longer:
Mostly I was just following the pattern I saw in similar controls on this page (namely UserProfile above), where you include the element and manage visibility from inside the control itself. But I don't feel strongly either way.

The reason I actually have to pass it in here (whereas UserProfile doesn't have the parameter), is because I'm not using the React Redux pattern UserProfile uses (with the whole `connect` method, etc...). And the reason I'm not using redux is because LanguageSelector is in react-common, where I don't see it getting used much. I don't know much about it TBH, but I get the impression redux would have coupled the control pretty tightly to some skillmap state and made it difficult to reuse in other areas of the code (like multiplayer).

kimprice: Hmm yea I don't feel strongly, I just thought this way might be simpler, but it also makes sense to keep things consistent.

kimprice: The more I think about it though, it does seem kind of unnecessary to have it as a prop if there's no other need for it within the component.

thsparks: Moved it out, as you suggested

PR #9389, opened 2023-02-15
Focus image canvas and support ctrl-s in asset page

riknoll:
I would not be surprised if this change fixes all of our "blockly is swallowing our keyboard events" bugs for the image/tilemap/animation editor.

This PR does 2 things:
1. Makes the image canvas focusable
2. In the --asseteditor page, map ctrl+s to the same callback as the done button

### DIFF
--- a/webapp/src/assetEditor.tsx
+++ b/webapp/src/assetEditor.tsx
@@ -164,8 +164,15 @@ export class AssetEditor extends React.Component<{}, AssetEditorState> {
         })
     }
 
+    handleKeydown = (e: KeyboardEvent) => {
+        if (e.ctrlKey && (e.key === "s" || e.key === "S")) {
+            this.callbackOnDoneClick();
+        }
+    }
+
     componentDidMount() {
         window.addEventListener("message", this.handleMessage, null);
+        window.addEventListener("keydown", this.handleKeydown, null);
         this.sendEvent({
             type: "event",
             kind: "ready"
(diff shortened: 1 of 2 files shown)


jwunderl (approved):

webapp/src/assetEditor.tsx
```diff
@@ -164,8 +164,15 @@ export class AssetEditor extends React.Component<{}, AssetEditorState> {
         })
     }
 
+    handleKeydown = (e: KeyboardEvent) => {
+        if (e.ctrlKey && (e.key === "s" || e.key === "S")) {
+            this.callbackOnDoneClick();
```
eanders-ms, line 169: In the web app, will pressing ctrl+s close the asset editor?

riknoll: nope! assetEditor.tsx is only used by the ---asseteditor page

eanders-ms: In the vscode context, would it be possible to:
- honor the user's custom key mapping for the "save" command?
- save when vscode triggers an auto save?
- save when user issues "save all" command.
- use vscode's visual indicator that the doc has unsaved edits (the dot in the tab)?

riknoll: apparently we are moving to auto-save... but this is all theoretically possible. i don't think vscode has hooks for when save is triggered on a custom editor (not that i'm aware of anyways) but we could read it from the shortcuts file and pass it into the webview...

riknoll: actually, we aren't even a custom editor so we definitely don't get save hooks.

PR #9390, opened 2023-02-15
Email checkbox on profile page now a set size

srietkerk:
The size of the email checkbox would change based on the size of the window. A checkbox like that is so small that it doesn't need to be responsive, so I just set the size to a fixed rem. It will always look the size shown in the picture. 

image

Closes https://github.com/microsoft/pxt-arcade/issues/5687

### DIFF
--- a/react-common/styles/controls/Checkbox.less
+++ b/react-common/styles/controls/Checkbox.less
@@ -1,6 +1,6 @@
 .common-checkbox {
     position: relative;
-    padding: 0.25rem;
+    // padding: 0.25rem;
 
     display: flex;
     align-items: center;
@@ -14,6 +14,10 @@
 .common-checkbox:focus-within {
     outline: @checkboxFocusOutline;
     border-radius: 0.2em;
+}
+
+#profile-email-checkbox {
+    width: 2rem;
 }
 
 /****************************************************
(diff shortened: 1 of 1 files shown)


react-common/styles/controls/Checkbox.less
```diff
@@ -1,6 +1,6 @@
 .common-checkbox {
     position: relative;
-    padding: 0.25rem;
+    // padding: 0.25rem;
```
eanders-ms, line 3: With this change, do other checkboxes render as expected?

eanders-ms, line 3: Does the checkbox scale when zoomed in? I think it should, but just curious.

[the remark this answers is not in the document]
srietkerk: Oh, shoot. I just forgot to remove that comment. I'll put it back in. From what I saw, it didn't make a difference, but I am not sure if it will impact other checkboxes.

srietkerk: Yes, it scales with zoom.

eanders-ms (approved):

jwunderl (approved):

abchatra:
Thanks port to stable.

PR #9391, opened 2023-02-15
Prevent Mini-Playable Sim from Overlapping Scrollbars

thsparks:
Noticed this while touching-up layout for microbit, the mini-playable sim would overlap the scrollbar in desktop mode. Then in mobile and table mode, it was actually a bit far away from them.

## Screenshots

| Scenario | Before Fix  | After Fix |
| ------------- | ------------- | ------------- |
| Desktop Arcade | image | image  |
| Desktop Microbit | image | image  |
| Mobile/Tablet Arcade | image |  image |
| Mobile/Tablet Microbit | image |  image |

### DIFF
--- a/theme/common.less
+++ b/theme/common.less
@@ -1721,7 +1721,7 @@ p.ui.font.small {
             width: 10rem;
         }
         .simPanel .simtoolbar {
-            margin: 0.5em 0;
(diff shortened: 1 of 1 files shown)


jwunderl (approved):
Awesome, nice clean up

PR #9392, opened 2023-02-16
Adjust code validation popup position based on sidebar

thsparks:
The sidebar for microbit is larger than the sidebar for arcade. TBD if it should stay that way, but regardless, I think it's better to make the code validation popup position reactive to changes in the sidebar width.

Without change:
image

With change:
image


jwunderl (approved):
Nice, better to use the const anyways~

PR #9394, opened 2023-02-17
auto save assets for vscode ext

jwunderl:
was this about how you'd want this to happen? Instead of a `setInterval` I suppose I add could add a field in the image editor itself to make it happen on each edit that goes in the undo/redo stack but not sure there'd be much difference there in the end.

~(I suppose might be worth renaming the `onDoneClick` now that it's gone from that to 'on done click, ctrl+s clicked, or we noticed it updated')~ cut out done, probably need to look at a throttle on the vscode ext side when saving

### DIFF
--- a/webapp/src/assetEditor.tsx
+++ b/webapp/src/assetEditor.tsx
@@ -226,7 +226,9 @@ export class AssetEditor extends React.Component<{}, AssetEditorState> {
                 ref={this.refHandler}
                 singleFrame={this.state.editing.type !== "animation"}
                 isMusicEditor={this.state.editing.type === "song"}
-                doneButtonCallback={this.callbackOnDoneClick} />
+                doneButtonCallback={this.callbackOnDoneClick}
+                hideDoneButton={true}
+            />
         }
 
         return <div></div>
--- a/webapp/src/assetEditor.tsx
+++ b/webapp/src/assetEditor.tsx
@@ -73,9 +73,9 @@ interface DuplicateAssetEditorResponse extends BaseAssetEditorResponse {
 
 type AssetEditorResponse = OpenAssetEditorResponse | CreateAssetEditorResponse | SaveAssetEditorResponse | DuplicateAssetEditorResponse;
 
-interface AssetEditorDoneClickedEvent {
+interface AssetEditorRequestSaveEvent {
     type: "event";
-    kind: "done-clicked"
+    kind: "done-clicked";
 }
 
 interface AssetEditorReadyEvent {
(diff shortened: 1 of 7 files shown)


webapp/src/components/ImageEditor/ImageEditor.tsx
```diff
@@ -35,6 +35,7 @@ export interface ImageEditorProps {
     onTileEditorOpenClose?: (open: boolean) => void;
     nested?: boolean;
     lightMode?: boolean;
+    hideDoneButton?: boolean;
```
riknoll, line 38: not for this pr, but next time we add a prop like this we should gather all the switches into a property bag. would save us a lot of passing props down in the image editor and music editor

webapp/src/assetEditor.tsx
```diff
@@ -178,11 +179,38 @@ export class AssetEditor extends React.Component<{}, AssetEditorState> {
...
+        }
+        if (this.state?.editing) {
+            this.tilemapProject.addChangeListener(
+                this.state.editing,
+                this.callbackOnDoneClick
+            );
+        }
     }
 
     callbackOnDoneClick = () => {
```
riknoll, line 216: should we rename this callback now that it's not really called on done click?

webapp/src/assetEditor.tsx
```diff
@@ -178,11 +179,38 @@ export class AssetEditor extends React.Component<{}, AssetEditorState> {
             kind: "ready"
         });
         tickAssetEditorEvent("asset-editor-shown");
+        this.pollingInterval = setInterval(this.pollForUpdates, 200);
```
riknoll, line 182: if we implemented an "onchange" function in the music editor and image editor, it might pick up changes faster and prevent the need for an unload event handler. i don't *think* it would need to be throttled at all...

riknoll (approved):

webapp/src/assetEditor.tsx
riknoll: iirc we already listen for changes to the store in the image editor and the music editor has an on song change callback

jwunderl: Yeah, that's probably the better way to do it, at first that's what I was thinking `tilemapProject.addChangeListener` would do until I looked a bit deeper; I think it's probably fine for now as even when I was attempting to induce it to save the wrong thing I was struggling to actual close the window in time after making a change such that this setInterval would miss (though if you have a pointer to those two events would be good to note down)

PR #9395, opened 2023-02-17
Fix Code Validation Mobile Layout

thsparks:
Still not perfect, but I think it's a big improvement from before.

Before:
image

After:
image

### DIFF
--- a/theme/tutorial-sidebar.less
+++ b/theme/tutorial-sidebar.less
@@ -1027,4 +1027,33 @@
     .tutorial-title {
         max-width: 100%;
     }
+
+    /*******************************
+        Validation
+    *******************************/
+
+    .tutorial-validation-error-container {
+        max-width: unset;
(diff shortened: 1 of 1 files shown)


kimprice:
Great improvement! Can you move the text so that it's left-aligned with the buttons? And maybe a little more padding on the top?

jwunderl (approved):
LGTM but aligning buttons / text as kim mentioned would be nice! (though maybe the other way around, reducing button margin so we take up as much space as possible)

thsparks:
> Great improvement! Can you move the text so that it's left-aligned with the buttons? And maybe a little more padding on the top?

> LGTM but aligning buttons / text as kim mentioned would be nice! (though maybe the other way around, reducing button margin so we take up as much space as possible)

Good suggestions. Made the change and updated the screenshot in the description.

kimprice (approved):

PR #9400, opened 2023-02-24
Make Tutorial Background Change with Green Screen Mode

thsparks:
The new tutorial backgrounds weren't being set to transparent when in greenscreen mode.

When I made the switch, I also noticed green screen mode was adding a scroll bar that allowed the user to scroll the entire site horizontally. This was happening because the video element that gets created was larger than the site itself (it gets scaled up to fill the screen, but this usually means `width` becomes larger than 100% so `height` can be 100% while maintaining the correct aspect ratio). To avoid this behavior, I set 100% `width` in the `video` element itself rather than the css, then cap the `width` to 100% in css and set `object-fit` to crop the `video` element appropriately, so it doesn't shrink down again when css limits its size.

I also noticed that, since the tutorial pane extends 100% vertically on desktop and the editor tools pane (with the download button) is pushed to the right, it looked weird to have the green background on the tutorial pane but not the editor tools bar. In non-tutorial layouts, however, the editor tools bar keeps its white background, and I didn't want to change that. Instead, I decided to leave the editor tools bar at 100% width when green screen is enabled. I'm not sure that this is the best way to do it, but it's relatively cheap and I think a reasonable trade off. Could revisit if we get user feedback, or if any team members feel strongly that another approach would be preferable.

Fixes https://github.com/microsoft/pxt-arcade/issues/5699

image

image

image

### DIFF
--- a/theme/greenscreen.less
+++ b/theme/greenscreen.less
@@ -34,6 +34,10 @@
         box-shadow: none !important;
     }
 
+    .toolbox-title {
+        background-color: @white;
+    }
+
     .monaco-editor .lines-content, .monaco-editor .view-line, .monaco-editor .view-lines {
 
         > span {
--- a/theme/greenscreen.less
+++ b/theme/greenscreen.less
@@ -17,8 +17,9 @@
         background-color: @greenScreenColor !important;
     }
     .videoContainer video {
-        min-width:100%;
-        min-height:100%;
+        max-width:100%;
+        max-height:100%;
+        object-fit: cover;
     }
     .videoContainer video.flipx {
         transform: scaleX(-1);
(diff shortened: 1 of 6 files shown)


jwunderl:
Can you show a screenshot with the camera actually on? It really feels like it'll end up very, very hard to read at all in any where your background is not a brightly lit whiteboard -- this section is much closer to the being similar to the toolbox (which is given a white background) than the workspace IMO

thsparks:
> Can you show a screenshot with the camera actually on? It really feels like it'll end up very, very hard to read at all in any where your background is not a brightly lit whiteboard -- this section is much closer to the being similar to the toolbox (which is given a white background) than the workspace IMO

Yeah it just depends on what the camera is actually looking at. I agree it may make more sense to just close the bug as by design.

image

jwunderl:
Yeah. It'd make a bit more sense to me if we trimmed it (e.g. put a white background on everything above and to the left of the bottom-right corner of the next button), but as is I think that is gonna be unusable for lots of set ups (in particular with any sort of compression / loss of quality from streaming) and harder than you would like at best. (we should probably throw a white background on the `Toolbox` label above the toolbox as well)

jwunderl (approved):

PR #9403, opened 2023-03-01
Fix sim hang after the sound editor is dismissed

riknoll:
Fixes the bug I mentioned in standup today.

to repro:

1. Open a project
2. Drag out a sound effect block
3. Open the sound effect editor and close it
4. Note that the sim hangs and won't restart


jwunderl (approved):

abchatra (commented):
port to stable

riknoll:
@abchatra I thought you'd say that: https://github.com/microsoft/pxt/pull/9404

PR #9405, opened 2023-03-01
Code Validation Telemetry Improvements

thsparks:
1. Added a field to several events which records how many times the user has tried to click next and encountered an error on the step. Basically, how many failed attempts have they made. This will simplify queries and I think it will be helpful in understanding if users are actually getting the error popup and fixing things, or if they just see the popup a few times and continue anyway.
2. Remove a redundant event that wasn't firing anyway due to a bug.
3. Fix a bug where two tutorial.next events were firing when the step counter was used.


thsparks:
> Nice fixes, will have to have a hotfix soon for #9404 and #9401 so can probably sneak it in the same one?

Yup, I was hoping so as well: https://github.com/microsoft/pxt/pull/9406

PR #9408, opened 2023-03-07
don't hide asset name in asseteditor page

riknoll:
Fixes https://github.com/microsoft/vscode-makecode/issues/100

checks to see if the done button is hidden and then doesn't hide the asset name

also fixed a random exception while i was in there; we weren't checking if assetname was defined before calling trim


kimprice (approved):

thsparks (approved):

abchatra:
Port to stable

PR #9411, opened 2023-03-08
add tags field so .g.jres can have gallery tags

jwunderl:
@kiki-lee wanted to put some gallery images for a tutorial but they don't show up in the background image gallery; currently the only way to sneak em in without running the cli tool would be via the assetsjres but that ends up creating ~40-200k long lines which can break things.

Basically, make https://github.com/jwunderl/arrow-images/blob/master/images.g.jres#L11 work to define tags for assets (I editted this by hand (don't tell anyone >.>) but we should probably just add the option to list out tags somewhere in the arcade-sprite-pack tool)

test build https://arcade.makecode.com/app/b7117e402ed06a7aaeda7ead979300400be20fee-8a1cae52b6#pub:_F3kbef0rbhzE

### DIFF
--- a/localtypings/pxtpackage.d.ts
+++ b/localtypings/pxtpackage.d.ts
@@ -205,6 +205,7 @@ declare namespace pxt {
         displayName?: string;
         tilemapTile?: boolean;
         tileset?: string[];
+        tags?: string[];
     }
 
     type SnippetOutputType = 'blocks'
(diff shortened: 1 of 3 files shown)


riknoll (approved):

riknoll:
>I editted this by hand

the comment is VERY CLEAR

kiki-lee:
> > I editted this by hand
> 
> the comment is VERY CLEAR

I warned him!

kiki-lee:
It looks like this also allows for animations to join the animation library?

jwunderl:
Animations in assetjson look supported already https://arcade.makecode.com/#pub:_ib9ACERscHf6 from https://github.com/jwunderl/arrow-images/blob/master/images.g.jres#L67 but this would support adding tags to them yes

PR #9414, opened 2023-03-09
Refinements to cloud status indicator

eanders-ms:
* Don't show "saving..." in the status bar while waiting/debouncing before cloud save start. This adds to the perception of lag.
* Make the cloud status color themeable.
* Normalize cloud status icons (make square, run through svgo)


pxtlib/cloud.ts
```diff
@@ -73,9 +73,9 @@ namespace pxt.cloud {
         ["localEdits"]: {
             value: "localEdits",
             icon: "cloud-saving-b",
-            tooltip: lf("Saving project to the cloud..."),
-            shortStatus: lf("saving..."),
-            longStatus: lf("Saving to cloud..."),
+            tooltip: lf("Project has local changes and will save to cloud soon."),
+            shortStatus: lf(""),
+            longStatus: lf(""),
```
jwunderl, line 78: I think build is complaining about these two, we don't want to upload empty strings to crowdin to translate :)

jwunderl (approved):
lgtm besides removing those two lfs in pxtlib/cloud.ts

pxtlib/cloud.ts
eanders-ms: yep fixed :)

PR #9415, opened 2023-03-13
Initial onboarding feature / editor tour

kimprice:
This is the initial onboarding feature for micro:bit. It lives in the Help menu, under Tour. The tour consists of several teaching bubbles.

Here's an upload target: https://makecode.microbit.org/app/0cda70b2f9538cff79a5aff199e632256c763801-b65db7415c

**Before merging, I need to set a flag so the tour is only available in micro:bit.**

To Do:
- Markdown parsing of content (instead of content from .tsx file)
- Query selector validation - this is what is used to identify the element to highlight and point to
- Screen-size issues - we may need to limit the tour depending on whether the screen is big enough
- Automated positioning / checks for space for teaching bubbles - currently we specify the relative location
- Styling: support high-contrast mode
- Styling: differentiate primary and secondary buttons
- (Future goal) Trigger tours automatically rather than just self-selecting the tour

EditorTour-short

image

image


webapp/src/container.tsx
```diff
@@ -66,6 +77,7 @@ export class DocsMenu extends data.PureComponent<DocsMenuProps, {}> {
         const targetTheme = pxt.appTarget.appTheme;
         return <sui.DropdownMenu role="menuitem" icon="help circle large"
             className="item mobile hide help-dropdown-menuitem" textClass={"landscape only"} title={lf("Help")} >
+            {(this.props.editor === "Blocks") && getTourItem(parent)}
```
kimprice, line 80: Should I just add a check here for the target?
```suggestion
            {(this.props.editor === "Blocks") && (pxt.appTarget.id === "microbit") && getTourItem(parent)}
```

eanders-ms: Add a feature flag to microbit's pxtarget.json.

react-common/components/controls/TeachingBubble.tsx
```diff
@@ -0,0 +1,291 @@
...
+    let top: number;
+    let left: number;
+    let arrowTop: number;
+    let arrowLeft: number;
+
+    useEffect(() => {
+        positionBubbleAndCutout();
+        window.addEventListener("resize", positionBubbleAndCutout);
+        return () => {
+            window.removeEventListener("resize", positionBubbleAndCutout);
```
kimprice, line 69: I know the event listener is being added every time the stepNumber changes and I'm not sure whether this return will take care of cleaning all of them up.

I tried to add the event listener for resizing the window in a useEffect with an empty dependency array, but that resulted in the teaching bubbles and cutouts taking the shape of the first teaching bubble and cutout whenever the window was resized. It seems the event listener needs to be in this useEffect so that the position of the bubble and cutout are calculated with respect to the current step in the tour.

Please let me know if there might be an easier/better way to do this.

webapp/src/container.tsx
kimprice: Do you have an example I can reference?

abchatra:
Awesome to see this. Great work!

webapp/src/container.tsx
abchatra: +1 Something like this: https://github.com/microsoft/pxt-microbit/blob/master/pxtarget.json#L590
Then you can enable this for any target and see how it works. The feature becomes feature flag driven instead of target driven.

eanders-ms: In code, feature flags are defined in the `AppTheme` interface: https://github.com/microsoft/pxt/blob/a44f003cb5d80c5fd9a77c09659cc0c3c378336b/localtypings/pxtarget.d.ts#L287

At runtime this interface is instantiated as `pxt.appTarget.appTheme`. Search for that to see how it is used to control feature UI. Here's an example of choosing to show immersive reader based on the feature flag: https://github.com/microsoft/pxt/blob/a44f003cb5d80c5fd9a77c09659cc0c3c378336b/webapp/src/components/tutorial/TutorialContainer.tsx#L50

react-common/components/controls/TeachingBubble.tsx
jwunderl: Yes, that's the way to do this - useEffect runs whenever those values change (including trivially changing the first time it renders) & the intention is to have it rerun whenever things need to be recalculated (e.g. the dimensions of the region changing per step).

The return value here is used whenever it is 'cleaning up' before running the handler again or discarding the element, so yes, handlers should all end up cleaned up. If you wanna check this, you can see what event listeners are registered in the dev tools by clicking into the 'Event Listeners' tab and finding the event you're looking for:

image

which can be helpful if you're looking for memory / event listener leaks~

srietkerk:
Haven't looked at the code yet, just playing with the target, but this happens on the simulator step on smaller screen sizes... It's possible this might be triggered on smaller iPads. I don't know how easy it would be, but we want to have a media query or something that puts the dialog on the other side of the simulator when the screen is this size. Super cool that the focus keeps on it!
image

srietkerk:
This one is has a similar situation, though this screen size is a bit smaller, more on the mobile device side. I'm imagining that we don't care about a use case of a screen this small. And since the "info" button doesn't show up on this form factor then we actually probably don't need to do more work here, because they won't be able to select the tutorial. Just putting this here in case it's something we want to consider.

image

jwunderl:
> This one is has a similar situation, though this screen size is a bit smaller, more on the mobile device side. I'm imagining that we don't care about a use case of a screen this small. And since the "info" button doesn't show up on this form factor then we actually probably don't need to do more work here, because they won't be able to select the tutorial. Just putting this here in case it's something we want to consider.
> 
> image

Ideally, we should probably move it over to the cogwheel when in small views like we do the other 'help' links, and point out that the share button is in that dropdown as well. Feels fine to leave as a follow up pr.

(if we want to be fancy, it's probably easiest to expand that menu with just `document.querySelector(".more-dropdown-menuitem").click()` before highlighting / showing the step. Maybe made as an optional field for mobile screens where it shows a different description, since we'd need the separate query anyway)

react-common/components/controls/TeachingBubble.tsx
```diff
@@ -0,0 +1,291 @@
...
+    const hasSteps = totalSteps > 1;
+
+    const classes = classList(
+        "teaching-bubble-container",
+        className
+    );
+
+    return ReactDOM.createPortal(<FocusTrap className={classes} onEscape={onClose}>
+        <div className="teaching-bubble-cutout"></div>
+        <div className="teaching-bubble-arrow"></div>
```
jwunderl, line 233:
```suggestion
        <div className="teaching-bubble-cutout" />
        <div className="teaching-bubble-arrow" />
```

nit; defining as self closing can be a little helpful to indicate 'nothing is intended to be inside here' / that it intentionally doesn't have any child divs; fine either way though

react-common/components/controls/TeachingBubble.tsx
```diff
@@ -0,0 +1,291 @@
...
+                        {!hasNext && <Button
+                            className="primary-button"
+                            onClick={onClose}
+                            title={finishLabel}
+                            ariaLabel={finishLabel}
+                            label={finishLabel}
+                        />}
+                    </div>
+                </div>}
+                {(!hasSteps) && <div className="teaching-bubble-footer">
```
jwunderl, line 279: nit; this looks pretty similar to the previous `teaching-bubble-footer`; if we made `teaching-bubble-steps` not show up when `!hasSteps` could we avoid splitting this into 2 footer paths (can always do `const finishLabel = hasSteps ? lf("Finish") : lf("Got it")` to make the label match what you want)?

It's not a big deal to split them up if we have to, but it is an extra place to forget to update something / have to test / reason about when we come back to fix something in the future.

```tsx
{stepNumber && hasSteps && <div className="teaching-bubble-steps">
```

webapp/src/components/onboarding/EditorTour.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+    const onNext = () => {
+        setCurrentStep(currentStep + 1);
+    };
+
+    const onBack = () => {
+        setCurrentStep(currentStep - 1);
+    };
+
+    return <TeachingBubble id="teachingBubble"
```
srietkerk, line 63: This is super clean, I really like how you did this.

react-common/components/controls/TeachingBubble.tsx
```diff
@@ -0,0 +1,291 @@
...
+                        onClick={onClose}
+                        title={confirmLabel}
+                        ariaLabel={confirmLabel}
+                        label={confirmLabel}
+                    />
+                </div>}
+            </div>
+        </div>
+    </FocusTrap>, parentElement || document.body)
+}
```
srietkerk, line 291: Oh, I also wanted to note here, and this may be more work than what it's worth. There's a lot of functionality in here that directly relates to the bubble arrow itself. Something that we may want to to do make the logic in `TeachingBubble.tsx` more simple is to break the bubble arrow into a separate component and then pass in, as a prop, which way the arrow is facing, and then do the computation from there.

react-common/components/controls/TeachingBubble.tsx
```diff
@@ -0,0 +1,291 @@
...
+        window.addEventListener("resize", positionBubbleAndCutout);
+        return () => {
+            window.removeEventListener("resize", positionBubbleAndCutout);
+        }
+    }, [stepNumber]);
+
+    useEffect(() => {
+        if (!hasSteps) {
+            const footer = document.querySelector(".teaching-bubble-footer") as HTMLElement;
+            footer.style.flexDirection = "row-reverse";
```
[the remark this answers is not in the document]
kimprice: It's to change the style in the case that the tour consists of just one teaching bubble. Since the steps won't be shown in this case, I want to reverse the direction so that the button shows on the bottom right of the bubble. Joey gave a good suggestion below of how to do this without using a useEffect,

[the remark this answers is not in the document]
srietkerk: Oh, cool! Thanks for the information!

webapp/src/container.tsx
kimprice: Thanks!

react-common/components/controls/TeachingBubble.tsx
```diff
@@ -0,0 +1,291 @@
...
+        cutout.style.left = `${cutoutLeft}px`;
+        cutout.style.width = `${cutoutWidth}px`;
+        cutout.style.height = `${cutoutHeight}px`;
+
+        if (activeTarget) {
+            cutout.style.pointerEvents = "none";
+        }
+    }
+
+    const clearCutout = () => {
```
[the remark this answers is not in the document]
kimprice: Good suggestion with making this based on state. Instead of adding a class, I'll just make that div conditional depending on the state.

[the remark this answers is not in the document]
kimprice: Actually, the div needs to always exist, so I'll do it with classes.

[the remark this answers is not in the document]
kimprice: Sadly, it won't work with classes because when the styles are set in code, they are set at the element level, which overrides anything at the class level. So, it looks like I will need to leave it as is.

react-common/components/controls/TeachingBubble.tsx
kimprice: oh yea. This is much cleaner. Thanks!

react-common/components/controls/TeachingBubble.tsx
```diff
@@ -0,0 +1,291 @@
...
+
+    useEffect(() => {
+        positionBubbleAndCutout();
+        window.addEventListener("resize", positionBubbleAndCutout);
+        return () => {
+            window.removeEventListener("resize", positionBubbleAndCutout);
+        }
+    }, [stepNumber]);
+
+    useEffect(() => {
```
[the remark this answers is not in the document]
kimprice: Good point. Thanks!

react-common/components/controls/TeachingBubble.tsx
kimprice: Yea, I actually thought about making the cutout a separate component. I could also make the bubble arrow its own component. Not sure if I'll do it for this PR, but might be worth doing in the future. Thanks for the suggestion.

react-common/components/controls/TeachingBubble.tsx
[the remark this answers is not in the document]
srietkerk: Ah, yeah, definitely a frustrating thing with CSS. Sounds good!

jwunderl (approved):
LGTM, lets get it in beta :)

srietkerk (approved):
Looks good to me! Super exciting stuff, and awesome work! It sounded frustrating, but everything looks awesome!

PR #9417, opened 2023-03-14
Don't emit songs in the 'myImages' namespace

riknoll:
Fixes https://github.com/microsoft/vscode-makecode/issues/130

The issue, in a nutshell:

1, We create a song (mySongs.Song1)
2. We emit that with the wrong namespace (myImages.Song1)
3. The user creates a new song
4. Since mySongs.Song1 is not used, we use that id
5. We emit again, the new asset goes to the wrong namespace and overwrites the original


jwunderl (approved):

eanders-ms (approved):

abchatra:
Port to stable

PR #9423, opened 2023-03-15
Fixes for filters in asset editor page (mobile view, read out tags from ts file)

jwunderl:
fix https://github.com/microsoft/vscode-makecode/issues/119

Also fixes some mobile view / thin desktop view sizing issues in gallery filter as it didn't look good in the thin webciew, and makes the section scrollable when it doesn't fit so it's still possible to select things when you have  a lot of categories. Probably looking at pr as separate commits is easiest as two of them were just moving some css rules below the ones they're overriding / removing unneeded importants now that the cascading order was correct

and adds a flag to the image field editor that makes it so background / tile / dialog will show up as options in the filter for vscode extension. Was just thinking again and might be worth throwing that flag on when it's loading the monacofieldeditor as well?

here's a build https://arcade.makecode.com/app/140dc9f3eb2d99ae039ee07fd466af633a638491-a23c652e99 (I was thinking maybe vscode extension should use mkc.json to set the --asseteditor page we point at, but that's probably not worth it when we just update the const locally anyway)

### DIFF
--- a/theme/image-editor/imageEditor.less
+++ b/theme/image-editor/imageEditor.less
@@ -354,15 +354,15 @@
 // Tablet + Mobile
 @media only screen and (max-width: @largestTabletScreen) {
     .filter-panel-container {
-        margin: 0 0 0 0.5rem !important;
-        padding: 0 0.5rem !important;
+        margin: 0 0 0 0.5rem;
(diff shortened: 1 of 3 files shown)


kimprice (approved):
Nice CSS clean up!

PR #9426, opened 2023-03-16
one more small adjustment for very thin phones

jwunderl:
quick follow up to https://github.com/microsoft/pxt/pull/9423 -- I was playing around on very thin views and the words got a little closer than I wanted, so just give it a wrap / some gap

build: https://arcade.makecode.com/app/3a4976d9a846afecd7a9248ba447046535e8b16b-4d237bd7dd

image

and anywhere it's wide enough display is normal: 
image


kimprice (approved):
Looks good!

PR #9427, opened 2023-03-17
allow passing display name for createAsset command

jwunderl:
for https://github.com/microsoft/vscode-makecode/pull/116, make this work:

I'm gonna port this over to the stable branch real quick so we can sneak it into this hotfix for testing tomorrow / it doesn't change anything if the value is just not passed anyways so pretty safe


jwunderl:
Oh, I'll just merge this as it already got reviewed in the stable branch and released to live / don't want to forget this one and have vscode extension break next release :)

PR #9430, opened 2023-03-17
Sim-Sidebar Layout for Tutorials

thsparks:
### Overview
This adds an optional feature flag that allows us to switch tutorials to a layout with the sim on the left-hand side and instructions at the top. It's somewhat reminiscent of the older tutorial layout, mixed with the tablet/mobile layout in the current tutorial infrastructure. The changes only really affect desktop view. In mobile and tablet, the layout is relatively unchanged.

### Notable changes
1. Removing the tabs in `sidepanel.tsx` and having both the instructions and the sim visible at the same time.
2. Switching the editorSidebar css styling to use a class (editor-sidebar) instead of an id, so more than one component can use the styles.
3. Taking a lot of the tablet-specific css styling and moving it into a shared style (using [lesscss mixins](https://lesscss.org/features/#mixins-feature))

### Thoughts / Future Work
Gut feeling is that, ultimately, the tutorial content and the sim sidepanel should be moved into fully separate components instead of keeping them both inside `sidepanel.tsx`, but that refactor felt out of scope for this initial change.

A separate change will have to go through pxt-microbit to enable the flag, but I'll do that after this goes in.

### Upload Target
Try it here: https://makecode.microbit.org/app/a73d12439a9e7b3c03aec6bcaa2b90db696f3d38-d8b5430834

#### Desktop
image

#### Tablet
image

#### Mobile
image

#### Arcade unchanged (flag is disabled):
image


theme/tutorial-sidebar.less
```diff
@@ -973,11 +904,124 @@
...
+                    padding-bottom: unset !important;
+                    height: 100%;
+                    min-height: @tutorialSimulatorMinHeight;
+                }
+            }
+        }
+    }
+
+    .topInstructionsWrapper {
+        .topInstructionsSettings();
```
kimprice, line 981: These mixins are really neat! Thanks for introducing them to me.

thsparks: Yeah, very handy! I'm not sure I'm using them 100% correctly, but it seems to work? So I'm happy with it :)

webapp/src/sidepanel.tsx
```diff
@@ -187,28 +171,27 @@ export class Sidepanel extends data.Component<SidepanelProps, SidepanelState> {
...
-                        onTutorialStepChange={onTutorialStepChange}
-                        onTutorialComplete={onTutorialComplete}
-                        setParentHeight={this.setComponentHeight} />
-                </TabContent>}
-            </TabPane>
+                </div>
+            </div>
+            {tutorialOptions &&
+                <div id="tutorialWrapper" className={this.props.tutorialSimSidebar ? "topInstructionsWrapper" : ""} style={tutorialHeightStyle}>
+                    <div className={outerTutorialContainerClassName} style={tutorialHeightStyle}>
```
kimprice, line 178: Would the style not be inherited from the parent div above? I'm wondering why we need to define the style again here.

webapp/src/sidepanel.tsx
```diff
@@ -187,28 +171,27 @@ export class Sidepanel extends data.Component<SidepanelProps, SidepanelState> {
...
+                        <TutorialContainer
+                            parent={parent}
+                            tutorialId={tutorialOptions.tutorial}
+                            name={tutorialOptions.tutorialName}
+                            steps={tutorialOptions.tutorialStepInfo}
+                            currentStep={tutorialOptions.tutorialStep}
+                            tutorialOptions={tutorialOptions}
```
kimprice, line 185: I know this code hasn't changed from before, but I'm wondering what the benefit is to passing so many props in like this rather than just extracting all this info from `tutorialOptions` from within the component?
Maybe it's just style/preference, but I think it would be better to pass in the minimum that is necessary.

thsparks: @jwunderl can probably say more on this than I can, but I think you're correct that it's mostly just style/preference.

By some metrics, doing it this way is actually more "minimal" than just passing in the tutorialOptions object would be. It's more lines of code, but less actual data than the entire tutorialOptions object. It also means the TutorialContainer is not dependent on the structure of the tutorialOptions object, so it could theoretically be used elsewhere (by something that doesn't have a tutorialOptions object) without the need to refactor. I don't know how realistic that scenario is, but it's one reason you might avoid just passing in the entire object.

kimprice:
There's a lot of movement of the hint and next buttons from step to step. I don't think this is new behavior, but it feels more pronounced when the instructions are at the top. I think it would look better if we could keep their positions consistent. Also for the longer instructions, they seem a little too close to the edge.
tutorialSteps

webapp/src/sidepanel.tsx
kimprice: Oh yea that's a good point. I think the biggest thing that surprised me in this case is that we also pass in the entire `tutorialOptions` object.

kimprice (approved):
Looks good!

thsparks:
> There's a lot of movement of the hint and next buttons from step to step. I don't think this is new behavior, but it feels more pronounced when the instructions are at the top. I think it would look better if we could keep their positions consistent. Also for the longer instructions, they seem a little too close to the edge. tutorialSteps [ tutorialSteps ](https://user-images.githubusercontent.com/15070078/226065647-def17222-937d-4b2b-a075-9471d79a179d.gif) [ ](https://user-images.githubusercontent.com/15070078/226065647-def17222-937d-4b2b-a075-9471d79a179d.gif)

Yeah I mentioned this in one of our design meetings (specifically, I thought it was weird that the next and hint buttons can go out of the screen and you have to scroll to get to them). My suggestion was to keep them in the same position and only scroll the text of the tutorial, but apparently that had already been considered and this was chosen as the better design--the main reason being, if "Next" was always right there, students wouldn't scroll the instructions at all; they would just click next without seeing everything.

I think the same issue applies to your point as well. If we lock the buttons in place, we'll have to scroll the instructions separately, and students just won't do that a lot of the time and will click next without seeing it. The fact that they're close to the edge is also just happening because they're getting pushed down but haven't quite hit the point where a scrollbar is necessary. I could see about trying to add the scrollbar in those scenarios too but I'm not sure if that's better or worse...

webapp/src/sidepanel.tsx
jwunderl: Yeah that's basically the intention, e.g. this leftover comment with that as the intention https://github.com/microsoft/pxt/blob/stable8.5/webapp/src/components/tutorial/TutorialContainer.tsx#L24

We have a few huge interfaces that end up making refactoring clunky at times (e.g. `IProjectView`) and *ideally* passing just what you need down to the components can make refactoring / component reuse a bit easier. as noted this one is a bit of a stretch as we are unpacking a decent amount of it here, but for what it's worth it is also sometimes nice to have the things the component cares about explicitly listed out.

In theory there are other benefits - e.g. perf improvements if `tutorialOptions` is changing a lot but the props you care about is not as react won't have to call rerender on every change, and avoiding bugs from missed renders if e.g. `tutorialOptions` is mutated in place as the props you care about are viewed explicitly (though that would typically just be considered masking the problem as the mutation would be the true issue), but they're all relatively minor.

jwunderl (approved):
Looks great, and some nice clean up along the way. Agreed on future work section as well (as that's basically what I intended to do last time but ran out of time)

webapp/src/sidepanel.tsx
thsparks: I hadn't even noticed that we also pass in `tutorialOptions`. I agree that's a bit strange, but I suppose Joey's point about explicitly listing out the requirements still applies to an extent.

jwunderl: Yeah, passing in the full object was because it was tedious to unpack completely when a few calls deeply nested in required it (which, I suppose, kinda proves the point on making refactors harder in a slightly roundabout way)

kimprice:
Re: buttons moving
A scrollbar probably won't be much better. Not sure there is an easy solution, but something to think about - nothing to hold this PR up for, though.

webapp/src/sidepanel.tsx
thsparks: Weirdly, it isn't inherited, but while looking into it I realized it can be simplified a little by just putting the style on the inner div and moving the id there too, so I'll make that change.

PR #9431, opened 2023-03-17
Enable Logging Tick Events to Console with URL Parameter

thsparks:
Add a url parameter (`consoleticks`), which will send all tick events to the console. I find this makes it much easier to test and debug telemetry events.

It comes with a two "modes". By default (`consoleticks=1` or `consoleticks=verbose`), it will print output like so:
```15:34:29 - Tick - activity.autorun {"editor":"blocksArea"} <no opts>```

You can also specify `consoleticks=2` or `consoleticks=short` to get a more abbreviated output:
```activity.autorun {"editor":"blocksArea"} <no opts>```

### DIFF
--- a/pxtlib/analytics.ts
+++ b/pxtlib/analytics.ts
@@ -89,17 +89,12 @@ namespace pxt.analytics {
 
     function consoleLogTick(id: string, data: Map<string | number>, opts: TelemetryEventOptions) {
         const query = Util.parseQueryString(window.location.href);
-
-        // Optionally specify consoletickfilter to limit which ticks get printed.
-        if (!query["consoletickfilter"] || id.startsWith(query["consoletickfilter"])) {
-            const tickInfo = `${id} ${data ? JSON.stringify(data) : "<no data>"} ${opts ? JSON.stringify(opts) : "<no opts>"}`;
-
-            if (query["consoleticks"] == "1" || query["consoleticks"] == "full") {
-                const timestamp = new Date().toLocaleTimeString(undefined, { hour12: false });
-                console.log(`${timestamp} - Tick - ${tickInfo}`);
-            } else if (query["consoleticks"] == "2" || query["consoleticks"] == "short") {
-                console.log(tickInfo);
-            }
+        const tickInfo = `${id} ${data ? JSON.stringify(data) : "<no data>"} ${opts ? JSON.stringify(opts) : "<no opts>"}`;
+        if (query["consoleticks"] == "1" || query["consoleticks"] == "full") {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pxtlib/analytics.ts [hunk 1]
```diff
@@ -33,6 +33,13 @@ namespace pxt.analytics {
 
         const te = pxt.tickEvent;
         pxt.tickEvent = function (id: string, data?: Map<string | number>, opts?: TelemetryEventOptions): void {
+            // Log tick to console if "consoleticks" url param is present.
+            // We do this simple check first (rather than the full parse) to minimize overhead when it's not present, which is most of the time.
+            if(window && /consoleticks=/.test(window.location.href))
```
jwunderl, line 38:
```suggestion
            if (window && /consoleticks=/.test(window.location.href))
```

jwunderl (approved):
lgtm!

kimprice (approved):
Nice! This will be very useful. :)

pxtlib/analytics.ts
```diff
@@ -33,6 +33,13 @@ namespace pxt.analytics {
 
         const te = pxt.tickEvent;
         pxt.tickEvent = function (id: string, data?: Map<string | number>, opts?: TelemetryEventOptions): void {
+            // Log tick to console if "consoleticks" url param is present.
+            // We do this simple check first (rather than the full parse) to minimize overhead when it's not present, which is most of the time.
+            if (window && /consoleticks=/.test(window.location.href))
```
eanders-ms, line 38: I'd prefer this were controlled by a flag set by the application, e.g. `pxt.analytics.consoleTicks = true`. How the flag is set is up to the app.

jwunderl: Here's where we have a few similar flags set to this:

https://github.com/microsoft/pxt/blob/stable8.5/webapp/src/app.tsx#L5539

thsparks: @eanders-ms - just wanted to clarify, do you mean it should still be enabled/disabled via a query parameter, but that query parameter should then set a flag (like `pxt.analytics.consoleTicks`), or do you mean we should just have a flag hardcoded to false which devs can set to true if they want to debug? Or something else entirely?

eanders-ms: pxtlib should be agnostic to how the flag is set. Leave that to the app. The webapp would parse the query param and set the flag. The cli would read a command line arg.

eanders-ms (approved):
Looks good!

jwunderl (approved):

PR #9433, opened 2023-03-21
Onboarding feature: add styling and automatic repositioning of teaching bubbles

kimprice:
Included in this PR:
- Styling: support high-contrast mode
- Styling: differentiate primary and secondary buttons
- Automated positioning
  - Checks whether teaching bubbles fit and repositions if necessary
  - This takes care of many screen-size issues

Here's an upload target: https://makecode.microbit.org/app/2f7a21a2fc5d243c9b32a0659fbceebb3fbb547b-48763d0d80

Still To Do:
- Add more telemetry ticks
- Screen-size issues
  - Move "Tour" to the cogwheel for small screens (just like other items from the help menu)
  - Have secondary query selector for elements that disappear/move for smaller screens (i.e. Share button)
- Refactoring
  - Make cutout its own React component
  - Make the teaching bubble arrow its own React component
  - clean up code in general (feel free to offer suggestions)
- Markdown parsing of content (instead of content from .tsx file)

EditorTour
editorTourHC

### DIFF
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -469,6 +469,7 @@ declare namespace pxt {
         multiplayer?: boolean; // enable multiplayer features
         shareToKiosk?: boolean; // enable sharing to a kiosk
         editorTour?: boolean; // enable the editor tour
+        tutorialSimSidebarLayout?: boolean; // Enable tutorial layout with the sim in the sidebar (desktop only)
     }
 
     interface DownloadDialogTheme {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pxt-core",
-  "version": "8.6.13",
+  "version": "8.6.14",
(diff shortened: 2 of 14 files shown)


jwunderl (approved):
To confirm my reading, the flow is:

1. attempt to use whatever position tour author (tsx but eventually md) says
2. if it succeeds, go ahead
3. else, try again in following order until it fits
    1. above
    2. below
    3. left
    4. right
    5. give up and just center modal it

Seems good to me. I could maybe see a step between checking right and center screen that would be 'centered inside element' (e.g. for if it is describing the workspace) but that's an edge case that probably won't come up often / shouldn't matter much.

kimprice:
re: positioning attempt
@jwunderl, yea the reason I don't include the center over element as an option is because I think that would really only be viable when it's being specified by the author. Otherwise, the bubble will likely cover most or all of the element. 

Another thing to consider is whether it would be better to not include the bubble at all if it can't fit or the element is not being displayed. Although I think it would be better to have the info than not in most cases.

PR #9434, opened 2023-03-22
Fix high contrast issues for palette editor and extensions page

kimprice:
Fixes https://github.com/microsoft/pxt-arcade/issues/5701
Fixes https://github.com/microsoft/pxt-microbit/issues/4997

At first glance, high contrast seemed to be working, but that was only the case if you had just turned high contrast on. If instead, you already had high contrast on from a previous session or reloaded the page, the palette editor and extensions page would not be in high contrast mode.

### DIFF
--- a/react-common/components/controls/Modal.tsx
+++ b/react-common/components/controls/Modal.tsx
@@ -127,5 +127,5 @@ export const Modal = (props: ModalProps) => {
                 </div>
             }
         </div>
-    </FocusTrap>, parentElement || document.body)
+    </FocusTrap>, parentElement || document.getElementById("root") || document.body)
 }
--- a/react-common/styles/controls/Input.less
+++ b/react-common/styles/controls/Input.less
@@ -125,7 +125,7 @@
  *                 High Contrast                    *
  ****************************************************/
 
-.high-contrast {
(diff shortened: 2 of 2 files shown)


react-common/components/controls/Modal.tsx
```diff
@@ -127,5 +127,5 @@ export const Modal = (props: ModalProps) => {
                 </div>
             }
         </div>
-    </FocusTrap>, parentElement || document.body)
+    </FocusTrap>, parentElement || document.getElementById("root") || document.body)
```
jwunderl, line 130: to clarify a bit in case anyone's confused what this actually changes; `parentElement` is an optional prop that determines where this this modal should be rooted (as modals are things that should 'show up' above other things and so they need to get positioned at a different place in the dom). When we were looking at it yesterday we noticed that not all modals were passing a value for this and were instead falling back to being placed under the body instead of root where `.hc` gets applied (`#root` is the `ProjectView`'s id in `app.tsx`).

If we're in the webapp almost everything is intended to be under root anyway which is why I suggested this vs going back and making `parentElement` required~

riknoll (approved):

PR #9435, opened 2023-03-22
Add missing third party licenses

BenVillalobos:
Fixes https://github.com/microsoft/pxt-backend/issues/776

Done manually by searching for each of our dependencies in our current ThirdPartyNotice. There's a lot more that we don't have, see below for details.

These are (roughly) the npm libraries we don't have licenses for. They're all under the `devDependencies` element in package.json. Is there a reason we don't already have licenses for many of these? Seems weird that so many would have slipped through.
```
    "@microsoft/eslint-plugin-sdl": "^0.1.5",
    "@types/chai": "4.0.6",
    "@types/fuse": "2.5.2",
    "@types/highlight.js": "9.12.2",
    "@types/jquery": "3.3.29",
    "@types/marked": "0.3.0",
    "@types/mocha": "2.2.44",
    "@types/node": "10.14.2",
    "@types/react": "17.0.50",
    "@types/react-dom": "17.0.17",
    "@types/react-modal": "3.13.1",
    "@types/react-redux": "7.1.24",
    "@types/react-tooltip": "3.3.6",
    "@types/request": "2.48.5",
    "@types/resize-observer-browser": "^0.1.5",
    "@types/web-bluetooth": "0.0.4",//
    "@typescript-eslint/eslint-plugin": "^4.20.0", //
    "@typescript-eslint/parser": "^4.20.0", //
    "@vusion/webfonts-generator": "^0.7.1",
    "crypto-js": "^3.1.9-1",
    "envify": "4.1.0",
    "eslint": "^7.23.0", ///
    "eslint-plugin-import": "^2.22.1",
    "eslint-plugin-jsx-a11y": "^6.4.1",
    "eslint-plugin-react": "^7.23.1", ///
```

Side note, the script to generate our TPN doesn't work out of the box after giving it an API key.

### DIFF
--- a/ThirdPartyNotice
+++ b/ThirdPartyNotice
@@ -28365,3 +28365,587 @@ The copyright in this software is being made available under the BSD License, in
  
 =========================================
 END OF dashjs 4.4.0 NOTICES AND INFORMATION
+
+%% @fortawesome/fontawesome-free 5.15.4 NOTICES AND INFORMATION BEGIN HERE (https://www.npmjs.com/package/@fortawesome/fontawesome-free/v/5.15.4)
+=========================================
+Fonticons, Inc. (https://fontawesome.com)
+
+--------------------------------------------------------------------------------
+
+Font Awesome Free License
+
+Font Awesome Free is free, open source, and GPL friendly. You can use it for
+commercial projects, open source projects, or really almost whatever you want.
+Full Font Awesome Free license: https://fontawesome.com/license/free.
+
+--------------------------------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


jwunderl (approved):
unless something has changed anything in dev dependencies should not be included in third party notices, as we aren't shipping code from those / they're only used locally (e.g. the `@types` are all only used at build time as types are elided in the final result of a typescript compile)

BenVillalobos:
Glad I stopped adding licenses before going through the devDependencies. Thanks for the context!

PR #9437, opened 2023-03-23
Tighten Up Sim-Sidebar & Tablet Tutorial Layout

thsparks:
## Overview
This change makes some adjustments to try and reduce empty space in our sim-sidebar and tablet tutorial layouts. Notabel adjustments include:
1. Moving the tutorial name and step to the header bar
2. Moving the step counter to the bottom and combining it with the next button
3. Reduce some of the padding
4. Scroll-bar if the sim is taller than the screen

This change affects the desktop view when sim-sidebar is enabled and the tablet and mobile views, even when the sim-sidebar flag is disabled.

Try it here: https://makecode.microbit.org/app/9da527050b955d5e99842495ee9bbb28752bb8a5-fc35b1c6da

### Desktop, sim-sidebar view
image

### Desktop, No Sim-Sidebar View
image

### Tablet
|  Microbit | Arcade |
| ----------- | ----------- |
| image | image |

### Mobile
|  Microbit | Arcade |
| ----------- | ----------- |
| image | image |

### DIFF
--- a/webapp/src/components/tutorial/TutorialStepCounter.tsx
+++ b/webapp/src/components/tutorial/TutorialStepCounter.tsx
@@ -5,6 +5,8 @@ interface TutorialStepCounterProps {
     currentStep: number;
     totalSteps: number;
     title?: string;
+    customNext?: JSX.Element; // TODO thsparks : This feels weird, but alternative (I can think of) is to duplicate a bunch of next/done logic from TutorialContainer, so maybe this is preferable?
+    isHorizontal?: boolean;
     setTutorialStep: (step: number) => void;
 }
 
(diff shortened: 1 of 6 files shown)


jwunderl (approved):
Looks great

webapp/src/components/tutorial/TutorialStepCounter.tsx
```diff
@@ -37,7 +39,27 @@ export function TutorialStepCounter(props: TutorialStepCounterProps) {
 
     const stepButtonLabelText = (step: number) => lf("Go to step {0} of {1}", step + 1, totalSteps);
     const backButtonLabel = lf("Go to the previous step of the tutorial.");
-    const nextButtonLabel = lf("Go to the next step of the tutorial.");
+
+    const lastStep = currentStep == totalSteps - 1;
+    const nextButtonTitle = lastStep ? lf("Finish the tutorial.") : lf("Go to the next step of the tutorial.");
+    const nextButtonAction = lastStep ? props.onDone : handleNextStep;
+
+    const nextButton = props.isHorizontal ? (
```
srietkerk, line 47: Curious if we need to complicate it like this. If the screen is a bit smaller anyway, I'm wondering if we should just make it so that the button to the far right for progressing or finishing the tutorial/skillmap node should just be the icon without text.

jwunderl: We need this to be as visible as possible, as its the most important button in the tutorial ui. We've had lots of issues in the past with kids not finding the next button because it was just an arrow and not understanding how to progress

srietkerk (approved):

PR #9442, opened 2023-03-24
Hide Tutorial Instructions in Fullscreen Mode and Fix CSS Specificity 

thsparks:
This change fixes https://github.com/microsoft/pxt-microbit/issues/5003. The fix for this happens in `common.less` around line 1419.

It also goes through and adds `#simulator` to the places where I used .editor-sidebar so that the css specificity is closer to what it was before my changes in https://github.com/microsoft/pxt/pull/9430. It used to be id-level and I changed it to class level, which had layout side-effects because the priority of certain styles was now different. The hope is that this will reduce the likelihood of those bugs.

### DIFF
--- a/theme/common.less
+++ b/theme/common.less
@@ -1416,6 +1416,9 @@ p.ui.font.small {
         display: none !important;
         z-index: -10 !important;
     }
+    #tutorialWrapper {
+        display: none;
+    }
     .sandboxfooter {
         z-index: @sandboxFooterZIndex;
         bottom: 1rem;
--- a/theme/common.less
+++ b/theme/common.less
@@ -55,18 +55,18 @@ pre {
     position: relative;
 }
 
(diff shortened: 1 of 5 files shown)


jwunderl (approved):
LGTM, though I'm not sure I see what you were referring to with the app.tsx line 1419 ref? Is it existing code or a change that wasn't included in this pr? https://github.com/microsoft/pxt/blob/master/webapp/src/app.tsx#L1419

thsparks:
> LGTM, though I'm not sure I see what you were referring to with the app.tsx line 1419 ref? Is it existing code or a change that wasn't included in this pr? https://github.com/microsoft/pxt/blob/master/webapp/src/app.tsx#L1419

Oh whoops, I meant `common.less` not `app.tsx`

PR #9449, opened 2023-03-28
Allowing setting a theme for sim (& palette) from tutorial md

jwunderl:
Allow setting a theme in tutorial / pxt.json, and send that to the sim: https://arcade.makecode.com/app/777ae82cacd76f735a2b99ee0ee53328ed69906c-7731d4169b

w/ tutorial loading bubblegum theme:
https://arcade.makecode.com/app/777ae82cacd76f735a2b99ee0ee53328ed69906c-7731d4169b#tutorial:44182-86552-65080-62109

loading normal project w/ theme set:
https://arcade.makecode.com/app/777ae82cacd76f735a2b99ee0ee53328ed69906c-7731d4169b#pub:_gmeV5DWmEfKf

re: https://github.com/microsoft/pxt-arcade/issues/5819, I took a look around and this should be very simple to implement, but I didn't attach anything for that in this pr / I'll take a look again for next release -- the code is easy but the ux flow / user testing will take a bit and we can't fit that in for a hotfix. I'll find some time to put in the 'glue' for it so we can put something in beta, but before release there would be lots of things to consider, e.g. ;
* making sure user understands when a skin is unlocked
* making sure user understands difference between it being unlocked & being used (e.g. a skillmap might want to use a skin while 'in' it and allow the user to unlock it at the end along with a badge)
* ui for picking skin unlocked skins -- a dropdown is easy but maybe a preview or something?
* how much we care about making sure a user is 'allowed' to share a game with an unlocked skin
    * clientside just not listing in a dropdown, if they edit into pxt.json manually then it works
    * ignore field in pxt.json if not unlocked
    * stripping it from pxt.json when sharing if not unlocked
    * stripping it in backend when sharing if not unlocked

Similarly, I left off any handling for displaying which user you are playing as in the sim when a user sets there skin & edits a multiplayer project; on the sim side I have it set that explicitly defined skin overrides the player # based one to ignore this for now. Since this is only sneaking in the feature for tutorial authors (and manual-pxt-json-editors) at the moment I'd consider this as a fine 'known future enhancement'.

one last thing, I snuck in a fix for the 'not being able to define user palette' issue from https://github.com/microsoft/pxt-arcade/issues/5804 as it wasn't any extra work / fell into the same object. So this provides a way to apply a theme for a tutorial that you've selected in the asset editor.

https://arcade.makecode.com/app/777ae82cacd76f735a2b99ee0ee53328ed69906c-7731d4169b#tutorial:46259-14675-05160-20703

````
```simtheme
{
    "theme": "bubblegum",
    "palette": [
        "#000000",
        "#ffffff",
        "#d45362",
        "#e8958b",
        "#cc8945",
        "#f5dc8c",
        "#417d53",
        "#5dd48f",
        "#5162c2",
        "#6cadeb",
        "#b56edd",
        "#8f3f29",
        "#612431",
        "#c0fac2",
        "#24325e",
        "#1b1221"
    ]
}
```
````

### DIFF
--- a/pxtlib/gallery.ts
+++ b/pxtlib/gallery.ts
@@ -12,6 +12,7 @@ namespace pxt.gallery {
         filesOverride: pxt.Map<string>;
         dependencies: pxt.Map<string>;
         features?: string[];
+        simTheme?: Partial<pxt.PackageConfig>;
     }
 
     export function parsePackagesFromMarkdown(md: string): pxt.Map<string> {
@@ -63,6 +64,15 @@ namespace pxt.gallery {
 
         return {};
     }
+    export function parseSimThemeJSON(md: string): Partial<pxt.PackageConfig> {
+        const pm = /```simtheme\s+((.|\s)+?)\s*```/i.exec(md);
+
+        if (pm) {
+            return pxt.tutorial.parseSimThemeJson(pm[1]);
+        }
+
+        return {};
+    }
 
     export function parseExampleMarkdown(name: string, md: string): GalleryProject {
         if (!md) return undefined;
@@ -75,6 +85,7 @@ namespace pxt.gallery {
         const source = m[2];
         const features = parseFeaturesFromMarkdown(md);
         const jres = parseJResFromMarkdown(md);
+        const simTheme = parseSimThemeJSON(md);
 
         const prj: GalleryProject = {
             name,
@@ -85,7 +96,8 @@ namespace pxt.gallery {
             dependencies,
             features,
             snippetType,
-            source
+            source,
+            simTheme
         };
 
         prj.filesOverride = {
@@ -97,6 +109,7 @@ namespace pxt.gallery {
             prj.filesOverride[pxt.TILEMAP_JRES] = jres.jres;
             prj.filesOverride[pxt.TILEMAP_CODE] = jres.ts;
         }
+
         return prj;
     }
 
--- a/pxtlib/gallery.ts
+++ b/pxtlib/gallery.ts
@@ -65,7 +65,7 @@ namespace pxt.gallery {
         return {};
     }
     export function parseSimThemeJSON(md: string): Partial<pxt.PackageConfig> {
-        const pm = /```simtheme\s+((.|\s)+?)\s*```/i.exec(md);
+        const pm = /```simtheme\s+([\s\S]*?)```/i.exec(md);
 
         if (pm) {
             return pxt.tutorial.parseSimThemeJson(pm[1]);
--- a/webapp/src/simulator.ts
+++ b/webapp/src/simulator.ts
@@ -329,7 +329,8 @@ export function run(pkg: pxt.MainPackage, debug: boolean,
         autoRun,
         ipc: isIpcRenderer,
         dependencies,
-        activePlayer: playerNumber
+        activePlayer: playerNumber,
+        theme: pkg.config.theme,
     }
     //if (pxt.options.debug)
     //    pxt.debug(JSON.stringify(opts, null, 2))
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -1119,6 +1119,7 @@ declare namespace pxt.tutorial {
         tutorialValidationRules?: pxt.Map<boolean>; //a map of rules used in a tutorial and if the rules are activated
         globalBlockConfig?: TutorialBlockConfig; // concatenated `blockconfig.global` sections. Contains block configs applicable to all tutorial steps
         globalValidationConfig?: CodeValidationConfig; // concatenated 'validation.global' sections. Contains validation config applicable to all steps
+        simTheme?: Partial<pxt.PackageConfig>;
     }
 
     interface TutorialMetadata {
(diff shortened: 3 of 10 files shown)

### CODE REVIEW

File: pxtlib/gallery.ts [hunk 1]
```diff
@@ -63,6 +64,15 @@
 
         return {};
     }
+    export function parseSimThemeJSON(md: string): Partial<pxt.PackageConfig> {
+        const pm = /```simtheme\s+((.|\s)+?)\s*```/i.exec(md);
```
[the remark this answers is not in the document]
jwunderl: code ql is upset at this line, I'll fix it here for now but it is an existing regex (in this file, `parseTemplateProjectJSON`, `parseJResFromMarkdown`, `parseFeaturesFromMarkdown`, `parsePackagesFromMarkdown`) and in theory we should probably fix in all of those as well separately. Here's a build with the fix for this case => `/```simtheme\s+([\s\S]*?)```/i`, since we don't really care about trailing whitespace here as JSON.parse will just ignore it (though the worst case we should hit in this path would be the tutorial not loading, which there are already plenty of ways to make happen :) )

https://arcade.makecode.com/app/841da135762055afde2037ec09857edab2cc65ae-e8f3c30598#tutorial:44182-86552-65080-62109

kimprice:
For the last link, is it expected for me to get an "unapproved content" warning:
> https://arcade.makecode.com/app/777ae82cacd76f735a2b99ee0ee53328ed69906c-7731d4169b#tutorial:46259-14675-05160-20703

image

webapp/src/simulator.ts
```diff
@@ -329,7 +329,8 @@ export function run(pkg: pxt.MainPackage, debug: boolean,
         autoRun,
         ipc: isIpcRenderer,
         dependencies,
-        activePlayer: playerNumber
+        activePlayer: playerNumber,
```
srietkerk, line 332: Couple quesitons
1. Are there any special things that we might want to do with custom themes and multiplayer selection to make sure that they work well together?
2. Will one setting override the other? For example, if someone is trying to make a multiplayer game with a theme that has the simulator be a different color and have a different configuration, will the theme make it so the multiplayer selection is not applied? Are there edge cases here where if you select a specific active player and then apply a theme the theme might not apply?

jwunderl:
Yes, that is intneded to show up whenever you load a non- builtin, not explicitly approved tutorial

thsparks (approved):

webapp/src/simulator.ts
```diff
@@ -329,7 +329,8 @@ export function run(pkg: pxt.MainPackage, debug: boolean,
...
+        theme: pkg.config.theme,
```
kimprice, line 333: Should we have the field in the config be `simTheme` rather than just `theme` since it only applies to the simulator?

pxtlib/gallery.ts
```diff
@@ -63,6 +64,15 @@ namespace pxt.gallery {
 
         return {};
     }
+    export function parseSimThemeJSON(md: string): Partial<pxt.PackageConfig> {
+        const pm = /```simtheme\s+([\s\S]*?)```/i.exec(md);
```
srietkerk, line 68: Just curious, what does "pm" stand for?

kimprice (approved):
Awesome stuff, Joey! 🥇

pxtlib/gallery.ts
kimprice: is it parse markdown??

webapp/src/simulator.ts
srietkerk: Oh, I see my second question was answered in your PR comment here:
> Similarly, I left off any handling for displaying which user you are playing as in the sim when a user sets there skin & edits a multiplayer project; on the sim side I have it set that explicitly defined skin overrides the player # based one to ignore this for now.

Sorry about that!

pxtlib/gallery.ts
srietkerk: That's what I was thinking, but wasn't sure.

webapp/src/simulator.ts
jwunderl: Since it's in pxtJson it feels similar level to me as  `palette` in that it's assumed it applies to the output for the project (simulator/hw), but if it feels confusing we can change -- only thing is that the snippet that goes in the markdown would also change to `{ "simTheme": "bubblegum" }` which is a little more room for error (e.g. not camel casing, etc)

webapp/src/simulator.ts
jwunderl: Good question to have though! And yeah, for next release we'll probably want to consider it more carefully (e.g. maybe dedicate a little area to indicate which player you are if the theme is explicitly set) but for now I just left it as theme field taking priority

pxtlib/gallery.ts
jwunderl: it's just from copy / paste of the few functions above it, no thinking involved :) if I had to guess the `package` function was probably added first, and `m` is commonly used as a variable name for `Regex.exec` results as a shorthand for `match` (similar to `i` in loops, etc) so it probably was originally `package match`, but just not changed to match previous, similar code.

PR #9450, opened 2023-03-28
Fix Tutorial Layout Bugs

thsparks:
This addresses the following tutorial layout issues:

1. Makes Next and Back buttons the same height (fixes https://github.com/microsoft/pxt-microbit/issues/5029).
2. Aligns top of hint and next/back buttons, and shrinks hint button to be the same size as the next/back buttons (fixes https://github.com/microsoft/pxt-microbit/issues/5030).
3. Reduce space between hint button and popup (fixes https://github.com/microsoft/pxt-microbit/issues/5031).

Try it here on Micro:bit: https://makecode.microbit.org/app/1b7457b8c6c200710558a2c93bb057d5a46513fc-e3db915f82

And here on Arcade: https://arcade.makecode.com/app/1cff0a1e2c5efd69dd0811a54ec7ccb282cbfb0e-24044b6328

### DIFF
--- a/webapp/src/components/tutorial/TutorialStepCounter.tsx
+++ b/webapp/src/components/tutorial/TutorialStepCounter.tsx
@@ -44,6 +44,24 @@ export function TutorialStepCounter(props: TutorialStepCounterProps) {
     const nextButtonTitle = lastStep ? lf("Finish the tutorial.") : lf("Go to the next step of the tutorial.");
     const nextButtonAction = lastStep ? props.onDone : handleNextStep;
 
+    const previousButton = props.isHorizontal ? (
+        <Button
+            disabled={currentStep == 0}
+            className="ui button counter-previous-button"
+            leftIcon="icon arrow circle left"
+            onClick={handlePreviousStep}
+            aria-label={backButtonLabel}
+            title={backButtonLabel} />
+    ) : (
+        <Button
+            disabled={currentStep == 0}
+            className="square-button"
+            leftIcon="left chevron"
+            onClick={handlePreviousStep}
+            aria-label={backButtonLabel}
+            title={backButtonLabel} />
+    );
+
     const nextButton = props.isHorizontal ? (
         <Button
             className="ui button counter-next-button"
@@ -67,14 +85,7 @@ export function TutorialStepCounter(props: TutorialStepCounterProps) {
             <span className="tutorial-step-number">{lf("Step {0} of {1}", currentStep + 1, totalSteps)}</span>
         </div>
         <div className="tutorial-step-bubbles">
-            <Button
-                disabled={currentStep == 0}
-                className="square-button"
-                leftIcon={`icon ${props.isHorizontal ? "arrow circle left" : "left chevron"}`}
-                onClick={handlePreviousStep}
-                aria-label={backButtonLabel}
-                title={backButtonLabel}
-            />
+            {previousButton}
             {stepsToShow.map(stepNum => {
                 const isCurrentStep = stepNum === currentStep;
                 return <Button
--- a/webapp/src/components/tutorial/TutorialStepCounter.tsx
+++ b/webapp/src/components/tutorial/TutorialStepCounter.tsx
@@ -56,7 +56,7 @@ export function TutorialStepCounter(props: TutorialStepCounterProps) {
         <Button
             disabled={currentStep == 0}
             className="square-button"
-            leftIcon="left chevron"
+            leftIcon="icon left chevron"
             onClick={handlePreviousStep}
             aria-label={backButtonLabel}
             title={backButtonLabel} />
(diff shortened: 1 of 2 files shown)


webapp/src/components/tutorial/TutorialStepCounter.tsx
```diff
@@ -44,6 +44,24 @@ export function TutorialStepCounter(props: TutorialStepCounterProps) {
     const nextButtonTitle = lastStep ? lf("Finish the tutorial.") : lf("Go to the next step of the tutorial.");
     const nextButtonAction = lastStep ? props.onDone : handleNextStep;
 
+    const previousButton = props.isHorizontal ? (
```
srietkerk, line 47: I don't know if this suggestion would be better/beneficial/more readable, but one thing that can be done instead of making the ternary operator between two button elements would be just to have the ternary be on `className` and `leftIcon` . I notice that you had the ternary on `leftIcon` before, would love to know the context on this change. I know this would make those lines kinda long and clunky and I'm not sure what the performance implications are, either, so take it with a grain of salt.

thsparks: I'm open to either. I actually originally implemented it with the ternary operator on the classname and icon fields, then switched to this because I thought it was a little easier to read/understand/debug if someone is just looking at the code, but it's all a matter of preference and I don't feel strongly. If there's a preference to do it inside the strings instead, I'm open to switching it.

As far as performance, I haven't checked but I would be surprised if there was any measurable impact either way.

srietkerk:
Playing with the window size, I noticed that when the screen is in a more tablet-like size, the hints popup aligned to the right. I think this looks okay, but I just wanted to check to see if that's what we want.

image

image

webapp/src/components/tutorial/TutorialStepCounter.tsx
srietkerk: I have a slight preference for doing it in the strings because there is so much that is similar between the two buttons and I've seen a lot of other components in our code base change class name with ternary operators, so it's a bit more consistent with code style.

jwunderl: Yeah, this is not a place where we have to be considerate of performance, it'll never matter (especially in the context of something that would be run once or twice a second -- e.g. it's like 1 second for a billion computations including ternaries :)

```javascript
(() => {
  let sum = 0;
  let start = Date.now()
  for (let i = 0; i < 1000000000; ++i) {
    sum += i % 2 === 0 ? i : 0;
  }
  console.log(`${Date.now() - start}ms to calculate ${sum}`);
})();
```
image

if you wanna make it so just the icon / classname are set off a ternary it'd probably be just as readable to set them as a variable `backButtonClass` / `backButtonIcon` just above, similar to how `nextButtonTitle` / `nextButtonAction` are currently being set

jwunderl (approved):
lgtm

kimprice (approved):
LGTM! 
But +1 to Sarah's comment about the hint. I think we should be consistent and that the hint should show up near the hint button. Seems like something to be addressed in a future PR, though.

webapp/src/components/tutorial/TutorialStepCounter.tsx
thsparks: I made the change Sarah recommended here. Seemed simplest to just make it all inline rather than create a bunch of single-use variables, so I went with that approach. I don't love the nested ternaries but we're making compromises either way.

webapp/src/components/tutorial/TutorialStepCounter.tsx
```diff
@@ -87,7 +68,13 @@ export function TutorialStepCounter(props: TutorialStepCounterProps) {
...
                 />
             })}
-            {nextButton}
+            <Button
+                className={props.isHorizontal ? "ui button counter-next-button" : "square-button"}
+                leftIcon={`icon ${lastStep ? "check" : props.isHorizontal ? "arrow circle right" : "right chevron"}`}
+                onClick={lastStep ? props.onDone : handleNextStep}
+                aria-label={nextButtonTitle}
+                title={nextButtonTitle}
+                label={props.isHorizontal ? lastStep ? lf("Done") : lf("Next") : ""} />
```
srietkerk, line 77: Is it possible to simplify this? Since we don't want to give a label if it's not horizontal, I'm wondering if we can just do something like 
`props.isHorizontal && lastStep ? lf("Done") : lf("Next")`

Typescript might get angry or it might cause bugs if `label` equals `null`. I'm not sure.

thsparks: I don't think there's much more simplification we can do here, unfortunately. In the sample you provided, we would end up with a "Next" label on the button when it's not-horizontal mode (i.e. in "Chevron" mode), but we actually want it to be no label in that case. We basically want no label in non-horizontal mode, then either "Done" or "Next" based in horizontal mode based on whether or not it's the last step.

I guess we could try setting the label as "Next" in chevron-mode but hide it through CSS instead...

srietkerk: Oh, right. No, no. The way you have it is the most simple, you're right. I don't know why my brain defaulted to "if `isHorizontal` is false then nothing will happen". I guess I was trying to apply short circuiting here, but that's not a thing with ternaries.

srietkerk (approved):
Looks good! Nice work!!

PR #9451, opened 2023-03-29
client-side auth cleanup

eanders-ms:
This set of changes looks big, but the bulk is contained in `pxtlib/auth.ts`. The main purpose here is twofold:
- be tidier about cleaning up local auth state when logging out.
- be more resilient against stale auth state that might have been left over from a previous session.

This change also contains several important bug fixes:
- Fixes https://github.com/microsoft/pxt-arcade/issues/3878
- Fixes an issue where if a cloud project is synced down and any error occurs while processing it, some number of other cloud projects will not be listed in My Projects.
- Fixes the infinite reload issue @srietkerk saw yesterday in microbit.

Once merged, I will bump beta and test skillmap and multiplayer in the beta channel. Skillmap + auth is especially difficult to test effectively on localhost.

### DIFF
--- a/authcode/src/services/auth.ts
+++ b/authcode/src/services/auth.ts
@@ -13,13 +13,12 @@ class AuthClient extends pxt.auth.AuthClient {
     protected onSignInFailed(): Promise<void> {
         return Promise.resolve();
     }
-    protected onUserProfileChanged(): Promise<void> {
-        const state = this.getState();
-        if (state.profile) {
+    protected async onUserProfileChanged(): Promise<void> {
+        const state = await pxt.auth.getUserStateAsync();
+        if (state?.profile) {
             pxt.auth.generateUserProfilePicDataUrl(state.profile);
+            store.dispatch(authStore.setUserProfile(state.profile));
         }
-        store.dispatch(authStore.setUserProfile(state.profile));
-        return Promise.resolve();
     }
     protected onUserPreferencesChanged(diff: ts.pxtc.jsonPatch.PatchOperation[]): Promise<void> {
         return Promise.resolve();
@@ -27,7 +26,8 @@ class AuthClient extends pxt.auth.AuthClient {
     protected onStateCleared(): Promise<void> {
         return Promise.resolve();
     }
-    protected async onProfileDeleted(userId: string): Promise<void> {
+    protected onProfileDeleted(userId: string): Promise<void> {
+        return Promise.resolve();
     }
     protected onApiError(err: any): Promise<void> {
         return Promise.resolve();
--- a/pxtlib/auth.ts
+++ b/pxtlib/auth.ts
@@ -71,7 +71,7 @@ namespace pxt.auth {
     /**
      * Cloud-synced user state.
      */
-    export type State = {
+    export type UserState = {
         profile?: UserProfile;
         preferences?: UserPreferences;
     };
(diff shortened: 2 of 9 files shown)


pxtlib/auth.ts
```diff
@@ -674,7 +684,7 @@ namespace pxt.auth {
         const hash = callbackState.hash.startsWith('#') ? callbackState.hash : `#${callbackState.hash}`;
         const params = pxt.Util.stringifyQueryString('', callbackState.params);
         const pathname = loginState.callbackPathname.startsWith('/') ? loginState.callbackPathname : `/${loginState.callbackPathname}`;
-        const redirect = `${pathname}${hash}${params}`;
+        const redirect = `${pathname}${params}${hash}`;
```
jwunderl, line 687: Oh, this is a nice tiny side fix :)

skillmap/src/lib/authClient.ts
```diff
@@ -16,13 +16,12 @@ class AuthClient extends pxt.auth.AuthClient {
         // Show a notification?
         return Promise.resolve();
     }
-    protected onUserProfileChanged(): Promise<void> {
-        const state = this.getState();
-        if (state.profile) {
+    protected async onUserProfileChanged(): Promise<void> {
+        const state = await pxt.auth.getUserStateAsync();
+        if (state?.profile) {
```
jwunderl, line 21: I saw there was a change to multiplayers `firstName` logic, is that something that would need to be handled in skillmap as well (for my 'name on cert' i.e. https://github.com/microsoft/pxt/pull/9367/files#diff-b64bd726c7c8a82e62558b9df9a313c6c5eea061222664a197d9d6dfe3631c33R546)

jwunderl (approved):
LGTM, one quick question on a place I used profile / firstName

pxtlib/auth.ts
```diff
@@ -84,29 +84,70 @@ namespace pxt.auth {
...
+    // Preference hasAuthTokenAsync() over taking a dependency on this cached value.
+    export let cachedHasAuthToken = false;
+
+    export async function getAuthTokenAsync(): Promise<string> {
+        let token: string;
+        try { token = await pxt.storage.shared.getAsync(AUTH_CONTAINER, CSRF_TOKEN_KEY); } catch {}
+        cachedHasAuthToken = !!token;
+        return token;
+    }
+    async function setAuthTokenAsync(token: string): Promise<void> {
```
srietkerk, line 101: Are we worried that this can be used maliciously? If token is a string that shouldn't be there, will there be security problems?

pxtlib/auth.ts
```diff
@@ -84,29 +84,70 @@ namespace pxt.auth {
...
+        cachedHasAuthToken = !!token;
+        return await pxt.storage.shared.setAsync(AUTH_CONTAINER, CSRF_TOKEN_KEY, token);
+    }
+    export async function hasAuthTokenAsync(): Promise<boolean> {
+        return !!(await getAuthTokenAsync());
```
srietkerk, line 106: Is the function call necessary? Could we just use `cachedHasAuthToken` or does it not get set yet? Or is it that we don't want to look at the cached token here?

pxtlib/auth.ts
```diff
@@ -201,18 +229,17 @@ namespace pxt.auth {
...
+            // Clear local auth state so we can no longer make authenticated requests.
+            await delAuthTokenAsync();
+            await delUserStateAsync();
 
-            // Redirect to home screen, or skillmap home screen
+            // Redirect to home screen
             if (pxt.BrowserUtils.hasWindow()) {
-                window.location.href = `${window.location.origin}${window.location.pathname}${hash}`;
+                const hash = continuationHash ? continuationHash.startsWith('#') ? continuationHash : `#${continuationHash}` : "";
+                window.location.href = `${window.location.origin}${window.location.pathname}${window.location.search}${hash}`;
```
srietkerk, line 242: What does the `location.search` do?

authcode/src/services/auth.ts
```diff
@@ -13,21 +13,21 @@ class AuthClient extends pxt.auth.AuthClient {
     protected onSignInFailed(): Promise<void> {
         return Promise.resolve();
     }
-    protected onUserProfileChanged(): Promise<void> {
-        const state = this.getState();
-        if (state.profile) {
+    protected async onUserProfileChanged(): Promise<void> {
```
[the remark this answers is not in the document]
eanders-ms: I had to mark the function `async`, which made an explicit return statement unnecessary in this case. async methods always return a promise, but the semantics are such that code can just return a value directly (e.g. `return true`) rather than a promise-resolved value (e.g. `return Promise.resolve(true)`). When an async method returns `Promise<void>`, an explicit return is unnecessary because it isn't returning a value and the promise is implicit.

authcode/src/services/auth.ts
```diff
@@ -13,21 +13,21 @@ class AuthClient extends pxt.auth.AuthClient {
...
-        return Promise.resolve();
     }
     protected onUserPreferencesChanged(diff: ts.pxtc.jsonPatch.PatchOperation[]): Promise<void> {
         return Promise.resolve();
     }
     protected onStateCleared(): Promise<void> {
         return Promise.resolve();
     }
-    protected async onProfileDeleted(userId: string): Promise<void> {
+    protected onProfileDeleted(userId: string): Promise<void> {
```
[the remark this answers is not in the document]
eanders-ms: Just stubbing out all the callbacks available from the base class. Having them here serves as a kind of documentation for future devs, letting the reader know these callbacks are available.

pxtlib/auth.ts
eanders-ms: This value by itself isn't useful. Additional information is needed for it to work, and that info isn't visible to code.

pxtlib/auth.ts
eanders-ms: See comment above `cachedHasAuthToken`. `hasAuthTokenAsync` is the reliable way to get this value. `cachedHasAuthToken` may not be correct, or not initialized at all. It is there for legacy compatibility only.

pxtlib/auth.ts
```diff
@@ -178,7 +203,7 @@ namespace pxt.auth {
                 pxt.tickEvent('auth.login.start', { 'provider': idp });
                 window.location.href = apiResult.resp.loginUrl;
             } else {
-                await this.onSignInFailed();
+                try { await this.onSignInFailed(); } catch {}
```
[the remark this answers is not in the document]
eanders-ms: The exception is just ignored in this case and the code continues to execute. The application callback needs to handle its errors and not propagate them. This is just the auth lib protecting itself from callbacks that don't honor this contract.

pxtlib/auth.ts
eanders-ms: It returns the query string of the current url, e.g. `?identity=1&dbg=1`.

pxtlib/auth.ts
```diff
@@ -221,24 +248,18 @@ namespace pxt.auth {
...
 
-                    try {
-                        await this.onProfileDeleted(userId);
-                    } catch {
-                        pxt.tickEvent('auth.profile.cloudToLocalFailed');
-                    }
-
-                    // Update state and UI to reflect logged out state.
-                    this.clearState();
+                    try { await this.onProfileDeleted(userId); } catch {}
```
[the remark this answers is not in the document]
eanders-ms: I moved it. This was an application-specific tick, so I moved it to the webapp's `onProfileDeleted` callback.

authcode/src/services/auth.ts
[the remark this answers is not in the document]
srietkerk: I always forget that fact that async functions automatically return a Promise, thanks!

webapp/src/auth.ts
```diff
@@ -66,43 +67,42 @@ class AuthClient extends pxt.auth.AuthClient {
         return Promise.resolve();
     }
     protected async onProfileDeleted(userId: string): Promise<void> {
-        // Convert cloud-saved projects to local projects.
-        await cloud.convertCloudToLocal(userId);
+        try {
+            // Convert cloud-saved projects to local projects.
+            await cloud.convertCloudToLocal(userId);
```
kimprice, line 72: Do you want a tick event for the successful case as well?

eanders-ms: We were more interested in just knowing whether if this ever failed, but having a success tick would make it easier to get the rate of failure. We have other ticks we can use in its place - notably `auth.profile.deleted` - so I think we're covered if that question comes up.

pxtlib/auth.ts
kimprice: What did this fix exactly?

jwunderl: in a correctly formatted url parameters (`?abc=1`) come before the has (`tutorial:ajskdlf`) -- if it is set after the hash they are interpreted as part of the hash. e.g.:

image

vs

image

In a lot of locations we handle this as it's easy for users to make mistakes when constructing the url (e.g. using a regex on `window.location` instead of `URLSearchParams` to check when a parameter is set) but it's nice to keep it correctly formatted url when possible.

PR #9452, opened 2023-03-29
More Tutorial Layout Fixes

thsparks:
## Overview
This addresses the following issues:

1. Increase tutorial height so it doesn't change as much (fixes https://github.com/microsoft/pxt-microbit/issues/5016).
2. Even out padding around tutorial controls (fixes https://github.com/microsoft/pxt-microbit/issues/5032).
3. Border under instructions in high contrast (fixes https://github.com/microsoft/pxt-microbit/issues/5019).

## Some Weird Specifics
The best way I could find to then distribute the instruction content evenly throughout the space (once I'd given it a more-fixed height) was to switch the div containing the instructions (`tutorial-content-bkg`) from a `block` display to `flex`, which worked fairly well, but it also necessitated that I switch the immersive reader button to use absolute positioning, since it doesn't fit well into the flex layout.

While I was looking into adding the bottom border for instructions in high contrast mode, I realized the border that's supposed to be there normally was not getting applied either, because the css selector was looking for `.topInstructionsWrapper` > `#simulator` > `.editor-sidebar` but the actual hierarchy is `#simulator` > `.topInstructionsWrapper` > `.editor-sidebar`. In this case, `#simulator` is unnecessary anyway, so I just removed it.

## Upload Targets
Try it here on Micro:bit: https://makecode.microbit.org/app/2d29195101272fd7e50ee924b15f5bd9f540fb97-d100f373bc

And here on Arcade: https://arcade.makecode.com/app/baaf3d3ee58fc27bed1b133951ec1def2e73f4a4-fb60c7a42d


jwunderl (approved):

srietkerk (approved):
Looks good!

theme/highcontrast.less
```diff
@@ -877,6 +877,14 @@
         border: 1px solid @HCtextColor;
     }
 
+    #simulator .editor-sidebar:not(.topInstructions) {
+        border-right: 1px solid @HCtextColor;
+    }
+
+    #simulator .topInstructionsWrapper .editor-sidebar {
```
kimprice, line 884: > In this case, #simulator is unnecessary anyway, so I just removed it.

Did you leave `#simulator` in by accident?

thsparks: In this case it's helpful to add specificity to the selector. I only needed to remove it in a couple of spots in tutorial-sidebar.less where it was throwing off the selector ordering.

thsparks: Sorry, PR description was confusing :)

I noticed the selector-ordering issue while looking into high contrast mode, but it's actually unrelated to the HC changes. It was related to the border when *not* in high contrast mode.

PR #9455, opened 2023-04-03
Onboarding tour accessibility fixes, animation improvement

kimprice:
This PR improves:
- animation by making it faster and including width and height in transitions
- visibility of tabbing / focus on buttons (closes https://github.com/microsoft/pxt-microbit/issues/5012)
- screen reading (closes https://github.com/microsoft/pxt-microbit/issues/5014)
  - tested with Narrator and NVDA
  - it's not perfect, but it should read all the content now

Try it out: https://makecode.microbit.org/app/ad2675c5dac61c2bed8b9b8816708b290fda6d92-5997b1ced2

tabButton

### DIFF
--- a/react-common/styles/onboarding/TeachingBubble.less
+++ b/react-common/styles/onboarding/TeachingBubble.less
@@ -32,6 +32,16 @@
     border-radius: 0.5rem;
     padding: 1rem;
     z-index: @modalDimmerZIndex;
+
+    .common-button {
+
+        &:focus {
+            outline: .1rem solid @white;
+        }
+        &:focus::after {
+            outline: none;
+        }
+    }
 }
 
 .teaching-bubble-arrow {
--- a/react-common/styles/onboarding/TeachingBubble.less
+++ b/react-common/styles/onboarding/TeachingBubble.less
@@ -37,8 +37,9 @@
 
         &:focus {
             outline: .1rem solid @white;
+            filter: grayscale(.15) brightness(.85) contrast(1.3);
         }
-        &:focus::after {
+        &.blue:focus::after, &:focus::after {
             outline: none;
         }
     }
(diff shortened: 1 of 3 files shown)


react-common/styles/onboarding/TeachingBubble.less
```diff
@@ -32,6 +32,16 @@
     border-radius: 0.5rem;
     padding: 1rem;
     z-index: @modalDimmerZIndex;
+
+    .common-button {
+
+        &:focus {
```
jwunderl, line 38: can we make this darken the button a bit as well? It's pretty subtle with just a small extra outline, especially on the white button

react-common/styles/onboarding/TeachingBubble.less
```diff
@@ -32,6 +32,16 @@
...
+            outline: .1rem solid @white;
+        }
+        &:focus::after {
```
jwunderl, line 41: do we normally set outline using an ::after pseudoelement that this is overriding?

jwunderl (approved):

thsparks (approved):
I agree with Joey's comment about darkening the button (or something similar to make button selection more apparent), otherwise LGTM!

eanders-ms (approved):

react-common/styles/onboarding/TeachingBubble.less
kimprice: Found it. I think it's this: https://github.com/microsoft/pxt/blob/master/react-common/styles/controls/Button.less#L158

kimprice:
> Not blocking, but at 125% zoom, I see a thin white border strike across the bubble tab: image

Yea, I've noticed this happens on smaller screens, too. I'll file an issue for it.

react-common/styles/onboarding/TeachingBubble.less
kimprice: @jwunderl @thsparks our buttons generally only darken on hover, do you want me to add this effect for on focus for all buttons or just for the teaching bubble?

kimprice: I did it for the teaching bubble only, but using the same darkening method we use for on hover:
tabButton2

kimprice: I'm going to merge, but feel free to file an issue if I should make any other change.

PR #9456, opened 2023-04-04
Editor tour: add congrats with confetti

kimprice:
This PR adds:
- Confetti as a react common component (this code comes from Shannon's implementation within skillmaps)
- Final congrats step in tour with confetti
- Initial transition animation for cutout (full screen to toolbox)

Try it out: https://makecode.microbit.org/app/7ce1ca36335e9776ca7ab2161f3747d9689808da-3c95e373c1
editorTourCongrats

### DIFF
--- a/react-common/components/animations/Confetti.tsx
+++ b/react-common/components/animations/Confetti.tsx
@@ -1,0 +1,14 @@
+import * as React from "react";
+
+export const Confetti = () => {
+    const density = 100;
+    return <div className="confetti-container">
+        {Array(density).fill(0).map((el, i) => {
+            const style = {
+                animationDelay: `${0.1 * (i % density)}s`,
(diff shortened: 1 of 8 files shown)


Jaqster:
This tour is a delightful experience!!  So cool!  @kiki-lee - what do you think about having a Badge for this too?

jwunderl (commented):
code lgtm, but we should remove the implementation that's over in skillmap if we're moving confetti to react common or else it'll probably stay duplicated forever (https://github.com/microsoft/pxt/blob/master/skillmap/src/components/AppModal.tsx#L197, https://github.com/microsoft/pxt/blob/master/skillmap/src/styles/modal.css#L136)

jwunderl (approved):

PR #9460, opened 2023-04-05
Editor tour: remove project area from toolbox cutout

kimprice:
fixes https://github.com/microsoft/pxt-microbit/issues/5028
Before, when the toolbox was expanded, the cutout was including the project area directly below it. Now, I check if the cutout will collide with the project area and then adjust the cutout, if needed.
image

### DIFF
--- a/webapp/src/components/onboarding/EditorTour.tsx
+++ b/webapp/src/components/onboarding/EditorTour.tsx
@@ -13,6 +13,8 @@ const Toolbox: TargetContent = {
     title: lf("Toolbox"),
     description: lf("Drag out blocks of code from the Toolbox categories into the Workspace."),
     targetQuery: "#blocksEditorToolbox",
+    sansQuery: "#projectNameArea",
+    sansLocation: Location.Below,
     location: Location.Right,
 };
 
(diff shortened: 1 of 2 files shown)


webapp/src/components/onboarding/EditorTour.tsx
```diff
@@ -13,6 +13,8 @@ const Toolbox: TargetContent = {
     title: lf("Toolbox"),
     description: lf("Drag out blocks of code from the Toolbox categories into the Workspace."),
     targetQuery: "#blocksEditorToolbox",
+    sansQuery: "#projectNameArea",
```
jwunderl, line 16: Code seems good but would it be easier in this case to just move the targetQuery here to `.blocklyToolboxDiv`? I would see the sansquery as more of a last resort of possible (e.g. in the other case the toolbox gets counted as part of the blocksEditor & the way the workspace svg is set up there's not an immediately obviously delineation on what the workspace is, but most of the time when you have an element that is 'too big' it's because there is a parent element constraining it / with `overflow: auto` on it)

jwunderl (approved):
Code looks good (one suggestion for falsey safety), but probably moving the selector for this specific case would be better (can merge in the positioning logic still, just that this case we know exactly the size the element will be / don't have to go through the new logic)

webapp/src/components/onboarding/EditorTour.tsx
kimprice: Yea this works. I think originally I selected the other query because it doesn't include the extra whitespace in the toolbox when it isn't expanded. But it is simpler this way, so I made the change. Thanks for the suggestion!

PR #9461, opened 2023-04-10
Update App Insights

thsparks:
## Overview
This updates our application insights version from "applicationinsights-js" version 1.0.20 to "applicationinsights-web" version 2.8.11. 

We don't use the API much (from what I can see), but I aligned our two calls to trackEvent and trackException to the API defined in the [App Insights API](https://github.com/microsoft/ApplicationInsights-JS/blob/master/API-reference.md). I also updated how we load in the package inside `apptracking.html` and `tracking.html`.

## Specifics
The big chunks of code in `apptracking.html` and `tracking.html` are copied directly from the [Snippet Setup](https://github.com/microsoft/ApplicationInsights-JS/#snippet-setup-ignore-if-using-npm-setup) section of the app-insights getting started page.

I've kept our settings consistent. The only change needed was to rename `isCookieUseDisabled` to `disableCookiesUsage` per the readme: [Configuration](https://github.com/microsoft/ApplicationInsights-JS/#:~:text=isCookieUseDisabled%0AdisableCookiesUsage)

Technically, usage of `measures` in `trackEvent` is deprecated, but it still works, and I'm hesitant to move away from it, since it looks like it would break any queries that currently rely on measures (in other words, any numeric property). With that in mind, I've continued using it for now.

For exceptions, we have a `kind` that gets passed in, but it wasn't clear to me how this was getting used, even in the older package. There's also no clear parallel to it in the app insights [IExceptionTelemetry](https://github.com/microsoft/ApplicationInsights-JS/blob/master/API-reference.md#trackexception), so I've just opted to include it as a property. Another option was to use it to determine severity, but it didn't look like one kind was inherently more/less severe than the other.

## Important Note
After this change goes in, developers will need to re-run npm install on their repos to get the new package.


abchatra:
Great job @thsparks

PR #9464, opened 2023-04-13
Prioritize webusb downloads when available

jwunderl:
When pressing download, show pairing flow first to encourage webusb use.

* if they exit the flow (click outside, x) it will bail out and generate a hex file
* if they go through the pairing flow as normal and end up not connecting, there is a secondary option for 'download as file' which will generate the hex file && not show pairing prompt automatically again (in memory) (similar to a "don't show this again" toggle, but in this case we have to give an indication that clicking it will still download as well)
image

    * Above two points will probably be fiddled with after testing in beta to get something that feels right without making it too easy to accidentally not pair
* Minor tweaking of the download completed dialog to ideally make webusb pairing a bit more appealing
image
    * label from new to describing why you'd want it -- it's not really "new" anymore and new might make people think it's less stable / in testing / etc
    * plain anchor to button styling for pair to make it obvious it's an action you take
* pushing webusb first is enabled by target flag
*  `...`  menu next to download button has same options (connect device, download as file)

build https://makecode.microbit.org/app/ce711df396e087f68c1a6b552ca1c2c7a20f9fd5-088b06d19f

### DIFF
--- a/webapp/src/cmds.ts
+++ b/webapp/src/cmds.ts
@@ -447,12 +447,19 @@ export function maybeReconnectAsync(pairIfDeviceNotFound = false, skipIfConnecte
     return reconnectPromise;
 }
 
-export function pairAsync(): Promise<void> {
+export function pairAsync(implicitlyCalled?: boolean): Promise<void> {
     pxt.tickEvent("cmds.pair")
-    return pxt.commands.webUsbPairDialogAsync(pxt.usb.pairAsync, core.confirmAsync)
+    return pxt.commands.webUsbPairDialogAsync(pxt.usb.pairAsync, core.confirmAsync, implicitlyCalled)
         .then(res => {
-            if (res) return maybeReconnectAsync();
-            else return core.infoNotification("Oops, no device was paired.")
+            switch (res) {
+                case pxt.commands.WebUSBPairResult.Success:
+                    return maybeReconnectAsync();
+                case pxt.commands.WebUSBPairResult.Failed:
+                    return core.infoNotification("Oops, no device was paired.")
+                case pxt.commands.WebUSBPairResult.UserRejected:
+                    // User exited pair flow intentionally
+                    return;
+            }
         });
 }
 
--- a/webapp/src/cmds.ts
+++ b/webapp/src/cmds.ts
@@ -14,10 +14,6 @@ function log(msg: string) {
 }
 
 let extensionResult: pxt.editor.ExtensionResult;
-// This can be overidden by the extension result
-pxt.commands.renderBrowserDownloadInstructions = dialogs.renderBrowserDownloadInstructions;
-pxt.commands.renderIncompatibleHardwareDialog = dialogs.renderIncompatibleHardwareDialog;
-
 
 function browserDownloadAsync(text: string, name: string, contentType: string): Promise<void> {
     pxt.BrowserUtils.browserDownloadBinText(
@@ -314,6 +310,9 @@ function applyExtensionResult() {
     if (res.renderBrowserDownloadInstructions) {
         log(`extension upload renderBrowserDownloadInstructions`);
         pxt.commands.renderBrowserDownloadInstructions = res.renderBrowserDownloadInstructions;
+    } else {
+        // default
+        pxt.commands.renderBrowserDownloadInstructions = dialogs.renderBrowserDownloadInstructions;
     }
     if (res.renderUsbPairDialog) {
         log(`extension renderUsbPairDialog`)
@@ -322,6 +321,9 @@ function applyExtensionResult() {
     if (res.renderIncompatibleHardwareDialog) {
         log(`extension renderIncompatibleHardwareDialog`)
         pxt.commands.renderIncompatibleHardwareDialog = res.renderIncompatibleHardwareDialog;
+    } else {
+        // default
+        pxt.commands.renderIncompatibleHardwareDialog = dialogs.renderIncompatibleHardwareDialog;
     }
     if (res.showUploadInstructionsAsync) {
         log(`extension upload instructions async`);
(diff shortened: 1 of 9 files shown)


abchatra:
@Jaqster @jaustin

abchatra:
Just wondering, if the websub flow fails, can we remember it for that session and just do download file when the click again on download? Otherwise going through this flow for every time seems too much of a hassle. For example: You have really old microbit but want to keep flashing in the classroom without upgrading it to new firmware.

jwunderl:
If it fails and you end up on the modal where it gives you the prompt to try again, if you press `download as file` it won't prompt again until you refresh the page. Easy to make it so it does the same if you just close out, but then it's VERY easy to make a mistake (click too fast outside, see a pop up and reflexively bail out, escape from browser webusb dialog without selecting something) and not get the pairing suggested again.
image

jwunderl:
Itd be easy to add that button to the first step of the flow (just adding an extra boolean flag on the modal call) though if we want an immediate "gimme a file" button on the start pairing screen

abchatra:
Ok sounds good. Let keep the current flow for beta and test and see if we need to tweak later.

kimprice:
The flow feels weird to me. I think there should be an explicit button for downloading it as a file on the initial dialog. It feels unnatural that closing out of the dialog or clicking away still results in the file being downloaded. Instead, I would expect that to cancel the download.

jwunderl:
@kimprice yeah, that's a similar point to Abhijith's -- I'm not against adding a button at the start of the dialog / already made it easy with the, but the problem is that adding one will make a certain percentage of users just click it by accident and not get the pairing experience at all -- we want users using webusb if available as it is a much smoother experience / doesn't rely on students dealing with file system and they end up much more successful in flashing overall, so ideally we want to minimize points of failure to do so in the flow.

I'd describe this as something that's super easy to add if people start asking for it during the beta, but if we add it pre-emptively it's not something any user would think to ask to be taken away, even if it ends up as a net negative -- so better to leave out until we see it's a problem in practice, similar to how we are slow in adding new default blocks.

re: file downloading when you escape the dialog -- I agree it does feel *slightly* weird when I think about it all as one flow, but that button as been in the exact same location, always resulting in a flashed / flashable device (outside of network c++ fetching errors or compile errors, and in most cases requiring user interacting to drag over the file / etc) since the very beginning of MakeCode -- they pressed 'Download' in the first place so something should be produced, the pairing is providing an alternative way of making the flashing happen but if they escape it is just continuing with the normal download flow.

martinwork:
Related: https://github.com/microsoft/pxt-microbit/issues/4770

webapp/src/cmds.ts
```diff
@@ -447,12 +447,19 @@ export function maybeReconnectAsync(pairIfDeviceNotFound = false, skipIfConnecte
...
         .then(res => {
-            if (res) return maybeReconnectAsync();
-            else return core.infoNotification("Oops, no device was paired.")
+            switch (res) {
+                case pxt.commands.WebUSBPairResult.Success:
+                    return maybeReconnectAsync();
+                case pxt.commands.WebUSBPairResult.Failed:
+                    return core.infoNotification("Oops, no device was paired.")
+                case pxt.commands.WebUSBPairResult.UserRejected:
+                    // User exited pair flow intentionally
```
kimprice, line 460: Does this happen for both hitting the close button and for clicking outside the dialog?

webapp/src/dialogs.tsx
```diff
@@ -771,17 +772,16 @@ export function renderBrowserDownloadInstructions(saveonly?: boolean) {
                                         </div>
                                         {webUSBSupported &&
                                             <div className="download-callout">
-                                                <label className="ui purple ribbon large label">{lf("New!")}</label>
-                                                <div className="ui two column grid">
+                                                <label className="ui purple ribbon label">{lf("Want faster downloads?")}</label>
```
kimprice, line 775: I like this label!

kimprice (approved):
Looks good!

abchatra:
No browser needs user approval to detect any usb hardware. You don't want websites to enumerate the usb hardware you have.

Jaqster:
Here are my thoughts on user flow - 
[WebUSB download flow.pdf](https://github.com/microsoft/pxt/files/11253911/WebUSB.download.flow.pdf)
I'll set up some time for us to go through and discuss.

jaustin:
...well I was drafting the above while Jacqueline was writing her PDF. There's a lot of similarity to the dialog suggestions and feedback (numbers, less text, centred images, etc :D )

@Jaqster do you think it's worth giving people the option of pairing? It feels to me more like we should just be encouraging them down that flow? 

I also like that this leads the user through to a possible pairing in the app. @martinwork can we trigger pairing separately to a download with the iOS app today?

I think in this flow we should consider a future possible implementation where people pair by Web Bluetooth, too.

webapp/src/cmds.ts
jwunderl: This one triggers when we decide not to show the pairing automatically in the future -- so right now this will only be if the user clicks the 'download as file' button on the last screen, if they press close or click outside it will show the 'no device paired' warning in the case above, but we have some updates for another pr to clean up the flow / prettify it and we'll need to play around with how much we want to keep showing this dialog in the future

martinwork:
> I also like that this leads the user through to a possible pairing in the app. @martinwork can we trigger pairing separately to a download with the iOS app today?

The app does do pairing separate from flashing. MakeCode could send a message to the app to kick of the pairing process.  I would worry that a shared UI might create difficulties because, compared with WebUSB being connected, it's not so clear that BLE is paired, the app avoids remaining connected (and usually can't if not in pairing mode), etc.

IMHO, I'm not sure "Pair" and "Pairing" are the right terms for the WebUSB connection. As I understand it, it's about the user giving MakeCode permission to connect directly to the micro:bit. There's no mutual pairing bond like BLE. When the "Connect to device" menu choice was added, I expected "Pair now" to change to "Connect now". Saying "Connect" seems OK because MakeCode does connect and reconnect, once it has permission. The browser dialog also says "connect". On the other hand, it is awkward that it's the same verb for the physical USB cable.

abchatra:
FYI - we are planning to merge this PR and do the subsequent change to gif\text etc in a later PR. This is just so everyone can try and do one change at a time.

PR #9469, opened 2023-04-14
Missing extension telemetry

abchatra:
When we moved to new extensions UI we lost all telemetry points. Adding them back.


abchatra:
Probably good idea to port to Arcade for hot fix as well.

jwunderl (approved):

PR #9472, opened 2023-04-19
Make unlink github open in new tab

kimprice:
closes https://github.com/microsoft/pxt-microbit/issues/4977

The menu continues to stay open after clicking "unlink GitHub". Should I do something about that? Any easy fix?
unlinkGH


eanders-ms:
It looks like the sui.DropdownMenu component will auto-close on loss of focus, but in this case you might have to explicitly call its `hide()` method.

jwunderl:
it looks like the same applies to the forum link in arcade (which is just an outgoing link as well)

image

kimprice:
So should I make it close and also do the same for the forum link? Or should I leave it? @jwunderl @eanders-ms

webapp/src/container.tsx
```diff
@@ -249,6 +249,7 @@ export class SettingsMenu extends data.Component<SettingsMenuProps, SettingsMenu
 
     signOutGithub() {
         pxt.tickEvent("menu.github.signout");
+        (document.querySelector(".more-dropdown-menuitem") as HTMLElement).blur();
```
eanders-ms, line 252: Mixing direct DOM manipulation with React should be avoided if what you want to do can be done in pure React. In this case you should be able to get a reference to React's instantiation of the dropdown menu and call `hide` on that. There are a bunch of ways to get and maintain a ref. Here's one way:

```
return <sui.DropdownMenu ..... ref={ref => this.dropdown = ref}>
```

Then, add a `hide` method to `SettingsMenu`:

```ts
    hide() {
        this.dropdown?.hide();
    }
```

eanders-ms (approved):
Looks good!

jwunderl (approved):

PR #9475, opened 2023-04-20
Allow projects to be imported as asset packs

riknoll:
This PR adds a field to pxt.json that allows you to turn a project into an asset pack.

When the field is set, we do the following things:
1. Remove all files except for *.jres
2. Remap all of the asset ids to be under a new namespace
3. Add a gallery.ts file that declares variables for all of the assets, allowing you to reference them in TS as well

The namespace we generate is based on the key for this dependency in the project pxt.json. When referencing a package in a tutorial it is very important that you make the id in the `````` ```packages`````` block match the id field in the pxt.json of the package to avoid weird bugs.

For example, if the imported project is named "my asset pack", then the namespace for referencing the assets will be "my_asset_pack".

### DIFF
--- a/localtypings/pxtpackage.d.ts
+++ b/localtypings/pxtpackage.d.ts
@@ -92,6 +92,7 @@ declare namespace pxt {
         utf8?: boolean; // force compilation with UTF8 enabled
         disableTargetTemplateFiles?: boolean; // do not override target template files when commiting to github
         theme?: string | pxt.Map<string>;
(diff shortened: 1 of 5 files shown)


jwunderl (approved):

kiki-lee:
I am unreasonably excited for this.

kiki-lee:
One more question. I have a 160px x 360px background in Galaxy. How would I tag this as a background in the new system so that it shows up in the correct gallery?  If I add tags to my jres, will it honor them?

PR #9476, opened 2023-04-21
Add markdown parsing for tours

kimprice:
The markdown file is located in the target repo.
I also changed the name of `EditorTour.tsx` to `Tour.tsx` since this can be used for more than just the editor tour.


pxtlib/markdownMetadataParser.ts
```diff
@@ -12,7 +12,7 @@ namespace pxt {
     }
 
     export function getSectionsFromMarkdownMetadata(text: string): MarkdownSection[] {
-        const lines = text.split("\n");
+        const lines = text.split(/\r?\n/);
```
jwunderl, line 15: for ref, this caused issues parsing text when it was saved and not yet committed in windows so that's why I suggested this simple fix, seems fine to me~

webapp/src/onboarding.ts
```diff
@@ -0,0 +1,76 @@
...
+        "everything" : {
+            targetQuery: "#root",
+        }
+    }
+    return targetMap[target];
+}
+
+export async function loadTourStepsAsync(name: string): Promise<pxt.MarkdownSection[]> {
+    return pxt.Cloud.markdownAsync(name)
+        .then(md => pxt.getSectionsFromMarkdownMetadata(md));
```
jwunderl, line 35:
```suggestion
export async function loadTourStepsAsync(name: string): Promise<pxt.MarkdownSection[]> {
    const md = pxt.Cloud.markdownAsync(name);
    return pxt.getSectionsFromMarkdownMetadata(md);
```

prefer async/await

jwunderl (approved):
LGTM!

webapp/src/onboarding.ts
```diff
@@ -0,0 +1,76 @@
+export interface querySelector {
+    targetQuery: string;
+    sansQuery?: string;
+    sansLocation?: pxt.tour.BubbleLocation;
+}
+
+function getTargetMap(target: string): querySelector {
+    const targetMap: pxt.Map<querySelector> = {
+        "simulator": {
```
jwunderl, line 9: Since it might not be obvious, the reason I suggested doing it this way instead of directly putting the query / etc in the markdown was to avoid the content authors getting too deep into the weeds / boxing us in -- query structures rely on the structure of the dom which is basically an implementation detail, so this way we have a fixed set we can maintain compatibility with (so we don't e.g. get prevented from refactoring because we don't want to check every existing md file)

kimprice: Thanks for adding this info, Joey!

webapp/src/onboarding.ts
kimprice: Is `await` not needed on line 34?

jwunderl: whoops, I got distracted in a call while typing this comment, it is definitely needed! edited

PR #9477, opened 2023-04-21
Allow projects to declare dependencies to be asset packs and emit tags

riknoll:
This is an extension to my asset pack pr that allows for projects that import dependencies to declare those dependencies to be asset packs. it looks like this:

```
{
    "name": "Untitled",
    "description": "",
    "dependencies": {
        "device": "*",
        "tagged-assets": "pub:37366-48747-84953-13766"
    },
    "files": [
        "main.blocks",
        "main.ts",
        "README.md",
        "assets.json"
    ],
    "preferredEditor": "tsprj",
    "assetPacks": {
        "tagged-assets": true
    }
}

```

Also added tag emitting to the declaration file generated for asset packs. If tags are not specified in the jres, i check the name of the asset for "dialog" or "background" and add those tags if missing.

### DIFF
--- a/pxtlib/package.ts
+++ b/pxtlib/package.ts
@@ -168,6 +168,7 @@ namespace pxt {
             config.files = config.files.filter(f => f.endsWith(".jres"));
             config.files.push("gallery.ts");
             config.dependencies = {};
+            config.assetPack = true;
             this.config = config;
             this.saveConfig();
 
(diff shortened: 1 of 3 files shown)


pxtlib/package.ts
```diff
@@ -823,63 +830,50 @@ namespace pxt {
...
+                if (mainTs) {
+                    const missingPackages = this.getMissingPackages(this.config, mainTs);
+                    let didAddPackages = false;
+                    for (const missing of Object.keys(missingPackages)) {
+                        const conflicts = await this.findConflictsAsync(missing, missingPackages[missing]);
+                        if (conflicts.length) {
+                            const conflictNames = conflicts.map((c) => c.pkg0.id).join(", ");
+                            const settingNames = conflicts.map((c) => c.settingName).filter((s) => !!s).join(", ");
+                            pxt.log(`skipping missing package ${missing} because it conflicts with the following packages: ${conflictNames} (conflicting settings: ${settingNames})`);
+                            continue;
```
jwunderl, line 856: delete I think?
```suggestion
```

jwunderl (approved):

PR #9479, opened 2023-04-25
Add "Disconnect GitHub" button to GitHub screen

eanders-ms:
Add a button at the bottom of the screen to sign out of GitHub:

Arcade
image

Microbit
image

I changed the wording from "Unlink" to "Disconnect" because it made more sense to me. Curious what others think.
image

Resolves https://github.com/microsoft/pxt-microbit/issues/4398

### DIFF
--- a/webapp/src/gitjson.tsx
+++ b/webapp/src/gitjson.tsx
@@ -784,7 +784,6 @@ class GithubComponent extends data.Component<GithubProps, GithubState> {
                         {!isBlocksMode && isOwner &&
                             <sui.Link className="ui item button desktop only" icon="user plus" href={`https://github.com/${githubId.slug}/settings/collaboration`} target="_blank" title={lf("Invite others to contributes to this GitHub repository.")} />}
                         <sui.Link className="ui button" icon="external alternate" href={url} title={lf("Open repository in GitHub.")} target="_blank" onKeyDown={fireClickOnEnter} />
-                        {user && <sui.Button className="ui button" icon="fas fa-sign-out-alt" text={lf("Disconnect GitHub")} title={lf("Log out of GitHub")} onClick={this.handleSignoutGithub} onKeyDown={fireClickOnEnter} />}
                     </div>
                 </div>
                 <MessageComponent parent={this} needsToken={needsToken} githubId={githubId} master={master} gs={gs} isBlocks={isBlocksMode} needsCommit={needsCommit} user={user} pullStatus={pullStatus} pullRequest={pr} />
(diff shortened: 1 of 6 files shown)


abchatra:
Looks good to me too. @Jaqster

webapp/src/gitjson.tsx
```diff
@@ -801,6 +807,11 @@ class GithubComponent extends data.Component<GithubProps, GithubState> {
                     {!isBlocksMode && <ExtensionZone parent={this} needsToken={needsToken} githubId={githubId} master={master} gs={gs} isBlocks={isBlocksMode} needsCommit={needsCommit} user={user} pullStatus={pullStatus} pullRequest={pr} />}
                     <div></div>
                 </div>
+                <div className="ui serialHeader">
+                    <div className="rightHeader">
+                        {user && <sui.Button className="ui button" icon="fas fa-sign-out-alt" text={lf("Disconnect GitHub")} title={lf("Log out of GitHub")} onClick={this.handleSignoutGithub} onKeyDown={fireClickOnEnter} />}
```
kimprice, line 812: Are we using the `sui.Button` because it's already used within this file? I thought we preferred to use the react common components because they have better defaults for accessibility.

jwunderl: Yeah, I wouldn't want to mix one react-common button in with 51 other sui elements on the page, nor swap over everything to react common just to add a single button on this page. Whenever we do a larger refactor on this page it should be done, though

Jaqster:
In general, the GitHub auth experience is a little weird to me.  Right now, the only way you can sign into GitHub is through the little GitHub button at the bottom of the editor, correct?  

Even though I know our GitHub auth is different from our Cloud Synch auth stuff, I feel like as a user I would expect all my services/account connection status and actions to be in one place.

I know this is meant to be a quick fix, but I do think it's a disjointed experience, and this little button doesn't really help too much with that.  Ideally, would want to find a way to connect/disconnect from GitHub through the top right Profile/Sign in UI component.

abchatra:
Unfortunately, we can't logout from the profile because github is local sign-in only. For example, profile should be identical if you open from any machine. It won't be in case of github if we put the thing in the profile.

Jaqster:
I don't understand.  Maybe we can have a quick call and you can explain?

eanders-ms:
This was meant to be a quick fix specifically to address the issue mentioned in the description. I agree the GitHub integration, especially as it intersects with cloud identity, is disjoint. Redesigning it is out of scope for this release as much as I'd like to tackle that!

eanders-ms:
This button fixes this scenario: when the user is not signed in to their profile, but they are signed into github, there is currently no UI to sign out of github.

PR #9483, opened 2023-04-25
Update App Insights

thsparks:
## Notes on Revert & Fixes
I originally attempted this with https://github.com/microsoft/pxt/pull/9461, however I needed to rollback after noticing a few errors:

1. Common (shared) properties were not being added to events correctly.
2. Page views were not getting tracked correctly.
3. Docs telemetry was broken with the following error: image 

This change includes the original update plus the following fixes for the errors above:
1. Update how we include the telemetry initializer to fit the new API.
2. Update trackPageView calls to fit the new API, and prevent the default trackPageView event (called by the app insights setup snippet) from being sent.
3. Add ai.2.min.js to pxt-deployment-config: https://github.com/microsoft/pxt-deployment-config/pull/212

## Overview
This updates our application insights version from "applicationinsights-js" version 1.0.20 to "applicationinsights-web" version 2.8.11. 

We don't use the API much (from what I can see), but I aligned our two calls to trackEvent and trackException to the API defined in the [App Insights API](https://github.com/microsoft/ApplicationInsights-JS/blob/master/API-reference.md). I also updated how we load in the package inside `apptracking.html` and `tracking.html`.

Fixes https://github.com/microsoft/pxt-microbit/issues/4992

## Specifics
The big chunks of code in `apptracking.html` and `tracking.html` are copied directly from the [Snippet Setup](https://github.com/microsoft/ApplicationInsights-JS/#snippet-setup-ignore-if-using-npm-setup) section of the app-insights getting started page.

I've kept our settings consistent. The only change needed was to rename `isCookieUseDisabled` to `disableCookiesUsage` per the readme: [Configuration](https://github.com/microsoft/ApplicationInsights-JS/#:~:text=isCookieUseDisabled%0AdisableCookiesUsage)

Technically, usage of `measures` in `trackEvent` is deprecated, but it still works, and I'm hesitant to move away from it, since it looks like it would break any queries that currently rely on measures (in other words, any numeric property). With that in mind, I've continued using it for now.

For exceptions, we have a `kind` that gets passed in, but it wasn't clear to me how this was getting used, even in the older package. There's also no clear parallel to it in the app insights [IExceptionTelemetry](https://github.com/microsoft/ApplicationInsights-JS/blob/master/API-reference.md#trackexception), so I've just opted to include it as a property. Another option was to use it to determine severity, but it didn't look like one kind was inherently more/less severe than the other.

Had to update calls to `trackPageView`, move where we setup the `telemetryInitializer`, and adjust `telemetryInitializer` function to account for new APIs.

## Important Note
After this change goes in, developers will need to re-run npm install on their repos to get the new package.


docfiles/tracking.html
```diff
@@ -2,54 +2,68 @@
...
-                }
-            });
-        });
         var location = window.location.toString();
         // Use generic title for script share page, otherwise use current page title
         var pageTitle = isScriptPage(location) ? "Share Page" : null;
-        appInsights.trackPageView(pageTitle, scrubUrl(location), {urlReferrer: scrubUrl(document.referrer.toString())});
+        appInsights.trackPageView({
+            name: pageTitle,
+            uri: scrubUrl(location),
```
abchatra, line 62: Can we send existing name and uri as part of properties and set the top level name and uri to something like following? 

pageTitle: "Share Page"
uri: "arcade.makecode.com"  (Just the domain name).

This is to avoid explosion of properties to help data team. I can explain

abchatra (changes requested):
Looks great. Thank you so much!

docfiles/apptracking.html
```diff
@@ -2,25 +2,33 @@
...
-            appInsights.context.addTelemetryInitializer(telemetryInitializer);
-        });
+
         var location = window.location.toString();
         // Use generic title for script share page, otherwise use current page title
         var pageTitle = isScriptPage(location) ? "Share Page" : null;
-        appInsights.trackPageView(pageTitle, scrubUrl(location), {urlReferrer: scrubUrl(document.referrer.toString())});
+        appInsights.trackPageView({
+            name: pageTitle,
+            uri: scrubUrl(location),
```
abchatra, line 27: { 
 name: "arcade",
 uri: "arcade.makecode.com"
 properties: {
       urlReferrer: scrubUrl(document.referrer.toString())
       uri: scrubUrl(location),
 }

abchatra (approved):

docfiles/tracking.html
thsparks: Updated it to use window.location.hostname for the pageTitle and window.location.origin for the uri, which leads to names like "makecode.microbit.org" and urls like "https://makecode.microbit.org". It's a little redundant but I'm hopeful it'll at least stop the property "explosion" issue. Does that seem sufficient?

image

Microbit target with this change: https://makecode.microbit.org/app/9bc086c4340afea1f95376a4d20278faac5e4839-cdf103f54d
Arcade target with this change: https://arcade.makecode.com/app/bd62aa9e2785e9a3effbb7957c7ad33cb29ae27c-74d4bebe50

thsparks: Also visible in that screenshot: original url moved into properties

webapp/src/serviceworker.ts
```diff
@@ -96,6 +96,7 @@ function initWebappServiceWorker() {
 
         // ai
         `/blb/ai.0.js`,
+        `/blb/ai.2.min.js`,
```
thsparks, line 99: @riknoll Can you confirm if it's safe to append rather than replace the new file here? I wasn't 100% sure based on the code in the rest of the file.

riknoll: hmmm.... why not replace it? is the old file still being served? i think the serviceworker install might fail if one of the requests 404s, so if the old file is no longer present then it's entry should be removed

jwunderl: ^^ I wasn't sure on this either -- isn't this serviceworker shared between all tabs? How does it work if we open beta in one tab and live in another, but beta needs `ai.0.js` and live needs `ai.2.min.js`?

jwunderl: Oh, I think I see, is the way it behaves that https://github.com/microsoft/pxt/blob/master/webapp/src/appcache.ts#L29 which runs from `DOMContentLoaded`  will always end up causing it to use the last loaded page's serviceworker being the one that is 'active' as it'll overwrite if the files are different, and then we use the ref (beta / index / etc) to cache the files here https://github.com/microsoft/pxt/blob/master/webapp/src/serviceworker.ts#L182, so they won't stomp all over each other even if file set is diff.

Makes sense for handling sequential loads, I'm a little iffy on the behavior if you load them both 'simultaneously' (/index in one page and /beta in a separate tab while the first one is still loading) but that's not a real scenario.

thsparks: Hmm okay, I'll switch this over to replace rather than append, then.

PR #9486, opened 2023-04-26
Webusb pairing simplification

jwunderl:
Simplifying webusb pair flow -- starting from discussion from https://github.com/microsoft/pxt/pull/9464#issuecomment-1511822423 down. image changes are in an associated pxt-microbit pr~

No build because images have to be checked in, so extra details on flow below

microbit-pairing-success-path

* when the download button will started you into the pairing flow, don't show any icon; icon will fill in when the user either successfully pairs or selects to download file in pairing flow
image
* remove text from first modal, expand title, large gif front and center. Also makes this one **not** have an escape button, as first reaction a lot of users have is accidentally clicking out and getting lost / missing things; they have to click 'next'
image
* pairing step: simplify text, make image focus only on computer where action is happening. Download as file option offset in bottom left. Do not mention exact board names, there will only be valid things in the list as we filter on it / students get rabbit-holed trying to understand the jargon when really they just need. Change 'next' to 'pair' 
image
* success case: text was fine, change button to 'download' in the implicit pairing flow (though exiting will still trigger download, just acts as a hint), changed image to what was previously the 'firmware update needed' image as it actually serves as a pretty good "you're all set and ready to go" -- paintbucket pouring in a rainbow is basically just pouring in the multi-colored blocks :)
image
* failure case: students focus too much on the images as they're pretty / not text, and the suggestions aren't super needed ("try connecting again" suggestion is what will show up when they press try again and start over in step 1, "update firmware" is not actionable for many users (hard flow to go through / needs instructions)) -- just give the three options:
image
* for failure case we might add back in an image, though keeping as just text is somewhat nice as there's nothing drawing away their attention when they need to make a choice.
* most of the buttons in the webusb pairing flow get icons to make them a little more appealing / hint what they do

### DIFF
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -482,7 +482,6 @@ declare namespace pxt {
         deviceSuccessIcon?: string;
         downloadMenuHelpURL?: string;
         downloadHelpURL?: string;
-        firmwareHelpURL?: string;
         troubleshootWebUSBHelpURL?: string;
         incompatibleHardwareHelpURL?: string;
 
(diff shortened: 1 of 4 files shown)


jwunderl:
Oh and a note; I'll probably do a follow up pr swapping over from sui to react common modals/buttons/etc next week, but wanted to get in the 'flow changes' for testing tomorrow where those changes are mainly just refactoring based

abchatra (approved):

PR #9487, opened 2023-04-27
elide hidden and deprecated apis from ts completions

jwunderl:
Currently, we're showing deprecated apis in ts completions; e.g. `tiles.setTilemap` is shown which was deprecated in favor of `tiles.setCurrentTilemap`: 

image

vs with this:

image

I think there is some race condition where we don't attach the info correctly so sometimes we don't actually apply the filters, but it seems rare / would affect this whole fn anyways (most likely just before the project is fully loaded / etc type thing)


kimprice (approved):
LGTM!

riknoll (approved):

thsparks (approved):

PR #9488, opened 2023-04-28
Update URL Scrubbing in Telemetry

thsparks:
This change does two things:
1. Moves where `scriptIdRegex` is defined. Previously, the scope was such that `scriptIdRegex` was actually `undefined` when `scrubUrl` was called.
2. Adds a block-list of keywords and filters out any query parameters that contain those keywords before logging the telemetry. It's not the most efficient check in the world, but it's two very short lists, so I don't think it'll be an issue.

I would love to reduce the duplication here, but I believe that's already been tried before and I didn't want to delay sending the PR. That said, if anyone has ideas, I'm open to suggestions.

### DIFF
--- a/docfiles/apptracking.html
+++ b/docfiles/apptracking.html
@@ -1,4 +1,5 @@
 <script type="text/javascript">
+    window.scriptIdRegex = /(?:S?\d{5}-\d{5}-\d{5}-\d{5})|(?:_[0-9a-zA-Z]{12})/g;
     window.loadAppInsights = function (includeCookie, telemetryInitializer) {
         //Backend will patch / + blb + / with CDN url and the check that URL ends with the sha would pass
         var isProduction = includeCookie && (/[0-9a-f]{40}\/$/.test("/blb/"));
@@ -29,14 +30,29 @@
             }
         });
 
-        var scriptIdRegex = /(?:S?\d{5}-\d{5}-\d{5}-\d{5})|(?:_[0-9a-zA-Z]{12})/g;
         // Check if the current page contains a share URL
         function isScriptPage(url) {
-            return !!url.match(scriptIdRegex);
+            return !!url.match(window.scriptIdRegex);
         }
-        // Scrub the key (if any) from the URL.
+
+        // Scrub potential PII from the URL.
         function scrubUrl(url) {
-            return url.replace(scriptIdRegex, "xxxxx-xxxxx-xxxxx-xxxxx");
+            debugger;
+            if(!url) return url;
+
+            var scrubbedUrl = new URL(url.toLocaleLowerCase().replace(window.scriptIdRegex, "xxxxx-xxxxx-xxxxx-xxxxx"));
+
+            // Just as a precaution, strip any suspicious query parameters.
+            let blockedParamKeywords = ["username", "password", "token", "pwd"];
+            let searchParams = Array.from(scrubbedUrl.searchParams.keys());
+            for(let param of searchParams) {
+                console.log(param);
+                if(blockedParamKeywords.some(k => param.indexOf(k) != -1)) {
+                    scrubbedUrl.searchParams.delete(param);
+                }
+            }
+
+            return scrubbedUrl.toString();
         }
 
         return isProduction;
--- a/docfiles/tracking.html
+++ b/docfiles/tracking.html
@@ -1,5 +1,6 @@
 <script type="text/javascript" src="/doccdn/pxtweb.js"></script>
 <script type="text/javascript">
+    window.scriptIdRegex = /(?:S?\d{5}-\d{5}-\d{5}-\d{5})|(?:_[0-9a-zA-Z]{12})/g;
     window.loadAppInsights = function (includeCookie) {
         //Backend will patch / + doccdn + / with CDN url and the check that URL ends with the sha would pass
         var isProduction = includeCookie && (/[0-9a-f]{40}\/ai\.2\.min\.js$/.test("/doccdn/ai.2.min.js"));
@@ -64,14 +65,29 @@
             }
         });
 
-        var scriptIdRegex = /(?:S?\d{5}-\d{5}-\d{5}-\d{5})|(?:_[0-9a-zA-Z]{12})/g;
         // Check if the current page contains a share URL
         function isScriptPage(url) {
-            return !!url.match(scriptIdRegex);
+            return !!url.match(window.scriptIdRegex);
         }
-        // Scrub the key (if any) from the URL.
+
+        // Scrub potential PII from the URL.
         function scrubUrl(url) {
-            return url.replace(scriptIdRegex, "xxxxx-xxxxx-xxxxx-xxxxx");
+            debugger;
+            if(!url) return url;
+
+            var scrubbedUrl = new URL(url.toLocaleLowerCase().replace(window.scriptIdRegex, "xxxxx-xxxxx-xxxxx-xxxxx"));
+
+            // Just as a precaution, strip any suspicious query parameters.
+            let blockedParamKeywords = ["username", "password", "token", "pwd"];
+            let searchParams = Array.from(scrubbedUrl.searchParams.keys());
+            for(let param of searchParams) {
+                console.log(param);
+                if(blockedParamKeywords.some(k => param.indexOf(k) != -1)) {
+                    scrubbedUrl.searchParams.delete(param);
+                }
+            }
+
+            return scrubbedUrl.toString();
         }
 
         return isProduction;


docfiles/apptracking.html
```diff
@@ -29,14 +30,29 @@
             }
         });
 
-        var scriptIdRegex = /(?:S?\d{5}-\d{5}-\d{5}-\d{5})|(?:_[0-9a-zA-Z]{12})/g;
```
jwunderl, line 32: Can we just bump this up to the beginning of `loadAppInsights` right next to `isProduction` instead of storing it on window? Typically better to avoid polluting global namespace when possible

docfiles/apptracking.html
```diff
@@ -29,14 +30,29 @@
...
         function isScriptPage(url) {
-            return !!url.match(scriptIdRegex);
+            return !!url.match(window.scriptIdRegex);
         }
-        // Scrub the key (if any) from the URL.
+
+        // Scrub potential PII from the URL.
         function scrubUrl(url) {
-            return url.replace(scriptIdRegex, "xxxxx-xxxxx-xxxxx-xxxxx");
+            debugger;
```
jwunderl, line 40:
```suggestion
```

docfiles/tracking.html
```diff
@@ -64,14 +65,29 @@
...
+            debugger;
+            if(!url) return url;
+
+            var scrubbedUrl = new URL(url.toLocaleLowerCase().replace(window.scriptIdRegex, "xxxxx-xxxxx-xxxxx-xxxxx"));
+
+            // Just as a precaution, strip any suspicious query parameters.
+            let blockedParamKeywords = ["username", "password", "token", "pwd"];
+            let searchParams = Array.from(scrubbedUrl.searchParams.keys());
+            for(let param of searchParams) {
+                console.log(param);
```
jwunderl, line 84:
```suggestion
```

docfiles/tracking.html
```diff
@@ -64,14 +65,29 @@
...
         function isScriptPage(url) {
-            return !!url.match(scriptIdRegex);
+            return !!url.match(window.scriptIdRegex);
         }
-        // Scrub the key (if any) from the URL.
+
+        // Scrub potential PII from the URL.
         function scrubUrl(url) {
-            return url.replace(scriptIdRegex, "xxxxx-xxxxx-xxxxx-xxxxx");
+            debugger;
```
jwunderl, line 75:
```suggestion
```

docfiles/apptracking.html
```diff
@@ -29,14 +30,29 @@
...
+            if(!url) return url;
+
+            var scrubbedUrl = new URL(url.toLocaleLowerCase().replace(window.scriptIdRegex, "xxxxx-xxxxx-xxxxx-xxxxx"));
+
+            // Just as a precaution, strip any suspicious query parameters.
+            let blockedParamKeywords = ["username", "password", "token", "pwd"];
+            let searchParams = Array.from(scrubbedUrl.searchParams.keys());
+            for(let param of searchParams) {
+                console.log(param);
```
jwunderl, line 49:
```suggestion
```

docfiles/apptracking.html
```diff
@@ -29,14 +30,29 @@
             }
         });
 
-        var scriptIdRegex = /(?:S?\d{5}-\d{5}-\d{5}-\d{5})|(?:_[0-9a-zA-Z]{12})/g;
         // Check if the current page contains a share URL
         function isScriptPage(url) {
```
jwunderl, line 34: this function looks not used? can probably just delete and move the regex into scrubUrl then

docfiles/apptracking.html
```diff
@@ -29,14 +30,29 @@
...
-            return !!url.match(scriptIdRegex);
+            return !!url.match(window.scriptIdRegex);
         }
-        // Scrub the key (if any) from the URL.
+
+        // Scrub potential PII from the URL.
         function scrubUrl(url) {
-            return url.replace(scriptIdRegex, "xxxxx-xxxxx-xxxxx-xxxxx");
+            debugger;
+            if(!url) return url;
```
kimprice, line 41:
```suggestion
            if (!url) return url;
```

docfiles/apptracking.html
```diff
@@ -29,14 +30,29 @@
...
+
+            var scrubbedUrl = new URL(url.toLocaleLowerCase().replace(window.scriptIdRegex, "xxxxx-xxxxx-xxxxx-xxxxx"));
+
+            // Just as a precaution, strip any suspicious query parameters.
+            let blockedParamKeywords = ["username", "password", "token", "pwd"];
+            let searchParams = Array.from(scrubbedUrl.searchParams.keys());
+            for(let param of searchParams) {
```
kimprice, line 48:
```suggestion
            for (let param of searchParams) {
```

docfiles/apptracking.html
```diff
@@ -29,14 +30,29 @@
...
+        // Scrub potential PII from the URL.
         function scrubUrl(url) {
-            return url.replace(scriptIdRegex, "xxxxx-xxxxx-xxxxx-xxxxx");
+            debugger;
+            if(!url) return url;
+
+            var scrubbedUrl = new URL(url.toLocaleLowerCase().replace(window.scriptIdRegex, "xxxxx-xxxxx-xxxxx-xxxxx"));
+
+            // Just as a precaution, strip any suspicious query parameters.
+            let blockedParamKeywords = ["username", "password", "token", "pwd"];
```
jwunderl, line 46: are these coming from somewhere? I'd be more concerned about the parameters themselves in the first place as the query parameters are part that go to the server / would probably show up in backend logs? I'm not sure what scrubbing them would do for us really, as it would be just like going to `makecode.com/{my-credit-card-number}` or the user explicitly calling window.pxtTickEvent from the dev tools

docfiles/apptracking.html
```diff
@@ -29,14 +30,29 @@
...
+            let searchParams = Array.from(scrubbedUrl.searchParams.keys());
+            for(let param of searchParams) {
+                console.log(param);
+                if(blockedParamKeywords.some(k => param.indexOf(k) != -1)) {
```
kimprice, line 50:
```suggestion
                if (blockedParamKeywords.some(k => param.indexOf(k) != -1)) {
```

docfiles/tracking.html
```diff
@@ -64,14 +65,29 @@
...
+            if(!url) return url;
```
kimprice, line 76:
```suggestion
            if (!url) return url;
```

docfiles/tracking.html
```diff
@@ -64,14 +65,29 @@
...
+
+            var scrubbedUrl = new URL(url.toLocaleLowerCase().replace(window.scriptIdRegex, "xxxxx-xxxxx-xxxxx-xxxxx"));
+
+            // Just as a precaution, strip any suspicious query parameters.
+            let blockedParamKeywords = ["username", "password", "token", "pwd"];
+            let searchParams = Array.from(scrubbedUrl.searchParams.keys());
+            for(let param of searchParams) {
```
kimprice, line 83:
```suggestion
            for (let param of searchParams) {
```

docfiles/tracking.html
```diff
@@ -64,14 +65,29 @@
...
+                console.log(param);
+                if(blockedParamKeywords.some(k => param.indexOf(k) != -1)) {
```
kimprice, line 85:
```suggestion
                if (blockedParamKeywords.some(k => param.indexOf(k) != -1)) {
```

docfiles/tracking.html
abchatra, line 75: remove debugger?

docfiles/tracking.html
abchatra, line 84: remove console.log?

docfiles/tracking.html
thsparks: whoops :)

docfiles/apptracking.html
thsparks: This function gets used to clean the urlReferrer property (among other things), which could be any website with any params that ultimately linked to our site. We don't really know or control it in that case. And frankly I'd hope they aren't including PII in the url anyway...but this is meant to be an extra precaution just in case (and it might help reduce false positives on automated checks, like one we received the other day from something that looked sort of like a password but wasn't).

jwunderl: can we just cut off all the query parameters from the referrer then? I doubt we care about anything more than the domain and maybe path? Any sequence of unicode characters with length > ~6 'looks' like a possible password to some extent

thsparks: I considered that, but apparently some of them are important.

thsparks: @abchatra might be able to provide more context on which ones

abchatra: Minecraft has bunch of query parameters like Minecraft version and protocol version, tutorial md etc. it is very useful to query those

jwunderl: Maybe allowlist known good ones domains vs trying to denylist arbitrary input? It sounds like the issue came from referrer which is a bit different from our url (I was suggesting more aggressive scrubbing of urlReferrer than url, as url is 'explicitly intended for us', and minecraft params are all queryParams on url)

I'll defer to you though @abchatra , probably not something that really comes up too much in reality

abchatra: I agree this may not be very useful. However, I am ok with this change as it would atleast avoid these false positives.

abchatra (approved):

docfiles/tracking.html
```diff
@@ -64,14 +64,23 @@
...
+            var scrubbedUrl = new URL(url.toLocaleLowerCase().replace(scriptIdRegex, "xxxxx-xxxxx-xxxxx-xxxxx"));
+
+            // Just as a precaution, strip any suspicious query parameters.
+            var blockedParamKeywords = ["username", "password", "token", "pwd"];
+            var searchParams = Array.from(scrubbedUrl.searchParams.keys());
+            for (var param of searchParams) {
+                if (blockedParamKeywords.some(k => param.indexOf(k) != -1)) {
+                    scrubbedUrl.searchParams.delete(param);
+                }
+            }
```
jwunderl, line 81:
```suggestion
            for (var blockedKW of blockedParamKeywords) {
                if (scrubbedUrl.searchParams.has(blockedKW)) {
                    scrubbedUrl.searchParams.delete(blockedKW);
                }
            }
```

maybe a bit clearer, since searchParams is already a map

jwunderl (approved):
Probably let's just keep in mind the idea of wiping all searchParams from the referrer for if this comes up again then~ one suggestion on simplifying a bit

docfiles/tracking.html
thsparks: This doesn't quite have the same behavior, since it only checks for exact, full-word matches, whereas the intent is to check for params _containing_ any of the keywords. For example, the recent false positive was on a parameter named "force_modify_password". If I'm understanding your suggestion correctly, that would not get filtered, but if we stick to the indexOf check it would. The downside is, of course, we can't just use the convenient `.has()` function, but I think this will be better at weeding out potential issues.

jwunderl: Ah, I see, makes sense

PR #9491, opened 2023-05-02
Expand on the failure step text a bit

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/5102

expanded on the text a bit: 
image

open to any suggestions if it's still not clear; would prefer to avoid making it any longer / adding pictures if possible as kids in the user testing were very much distracted on this step when there were images /  didn't read the lengthy spots of text.

### DIFF
--- a/webapp/src/webusb.tsx
+++ b/webapp/src/webusb.tsx
@@ -200,11 +200,17 @@ function showConnectionSuccessAsync(confirmAsync: ConfirmAsync, willTriggerDownl
 
 function showConnectionFailureAsync(confirmAsync: ConfirmAsync, showDownloadAsFileButton?: boolean) {
     const boardName = getBoardName();
+    const tryAgainText = lf("Try Again");
+    const helpText = lf("Help");
+    const downloadAsFileText = lf("Download as file");
 
     const jsxd = () => (
         <div>
             <div className="ui content download-troubleshoot-header">
-                {lf("We couldn't find your {0}. Here's a few ways to fix that:", boardName)}
+                {lf("We couldn't find your {0}.", boardName)}
+                <br />
+                <br />
+                {lf("Click \"{0}\" for more info, \"{1}\" to try to pair again, or \"{2}\" for drag and drop pairing.", helpText, tryAgainText, downloadAsFileText)}
             </div>
         </div>
     );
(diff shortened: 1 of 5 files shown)


webapp/src/webusb.tsx
```diff
@@ -200,11 +200,17 @@ function showConnectionSuccessAsync(confirmAsync: ConfirmAsync, willTriggerDownl
...
+    const downloadAsFileText = lf("Download as file");
 
     const jsxd = () => (
         <div>
             <div className="ui content download-troubleshoot-header">
-                {lf("We couldn't find your {0}. Here's a few ways to fix that:", boardName)}
+                {lf("We couldn't find your {0}.", boardName)}
+                <br />
+                <br />
+                {lf("Click \"{0}\" for more info, \"{1}\" to try to pair again, or \"{2}\" for drag and drop pairing.", helpText, tryAgainText, downloadAsFileText)}
```
thsparks, line 213: Is "pairing" the right word for the drag-and-drop scenario? I wonder if "drag-and-drop install" or something would be more applicable.

webapp/src/webusb.tsx
```diff
@@ -200,11 +200,17 @@ function showConnectionSuccessAsync(confirmAsync: ConfirmAsync, willTriggerDownl
 
 function showConnectionFailureAsync(confirmAsync: ConfirmAsync, showDownloadAsFileButton?: boolean) {
     const boardName = getBoardName();
+    const tryAgainText = lf("Try Again");
+    const helpText = lf("Help");
+    const downloadAsFileText = lf("Download as file");
```
thsparks, line 205: Was the intent to use `downloadAsFileText` and `helpText` elsewhere in the file? Seems like `downloadAsFileText` would need to go in `showPairStepAsync` which would require a broader scope, and I'm guessing that's why you opted not to use it there...but figured I'd check.

thsparks (approved):
A couple questions/nits but looks good on the whole!

webapp/src/webusb.tsx
kimprice: I was wondering the same thing. "Pairing" doesn't sound quite right to me.

webapp/src/webusb.tsx
jwunderl: Yeah, I just pulled them out so the lf on the rest of the description didn't get nested, not attempting to reduce redundancy there as it's a little messy (e.g. can't put it as top level as lf requires translations to be loaded )

kimprice: Do we have any styling rules for when/how to capitalize the button text? Seems a little inconsistent now with the `Download as file` capitalization, but maybe it's fine to use sentence case since it's a secondary button?

kimprice (approved):

webapp/src/webusb.tsx
jwunderl: we're a little inconsistent yeah; I had copied it form 
image
but I'd say we typically lean more towards Title Casing buttons
image
I'll swap these two to title case here as well

jwunderl:
Okay, cleaned up a bit: 
image
image
image
image

PR #9493, opened 2023-05-02
webusb try pair again: return the recursive call

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/5099

wasn't using the result from the recursive call, a little interesting this has stuck around for > a year without being noticed but I guess not that many people were using webusb and failing the first time / succeeding the second.


eanders-ms (approved):
😮

kimprice (approved):

PR #9495, opened 2023-05-03
Fix toolboxParent attr when no shadow is given in the parent block

riknoll:
Fixing a bug that @thsparks ran into while using the `//% toolboxParent` comment attribute.

If a block doesn't specify a shadow block id or default value for a parameter, then we don't actually create a value tag for that parameter in the toolbox block XML. That's a problem when that block is used as the toolboxParent for another block.

If that value tag doesn't exist in the parent XML, the query selector we use when applying the toolboxParent attribute comes up empty and the attribute is ignored completely.

If you want to see the bug for yourself, you can use this code:

```
enum Bar {
    A,
    B
}

//% weight=100 color=#0fbc11 icon=""
namespace custom {
    export class Foo {
    }

    //% block="create foo"
    //% toolboxParent=use_foo
    export function createFoo(): Foo {
        return new Foo();
    }

    //% block="create foo 2"
    //% toolboxParent=use_foo
    //% toolboxParentArgument=foo
    export function createFoo2(): Foo {
        return new Foo();
    }

    //% block="use $bar foo $foo"
    //% blockId=use_foo
    export function useFoo(bar: Bar, foo: Foo) {
    }
}
```

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -342,11 +342,34 @@ namespace pxt.blocks {
             const parentFn = info.blocksById[fn.attributes.toolboxParent];
 
             if (parentFn) {
-                parent = createToolboxBlock(info, parentFn, pxt.blocks.compileInfo(parentFn));
+                const parentInfo = pxt.blocks.compileInfo(parentFn);
+                parent = createToolboxBlock(info, parentFn, parentInfo);
 
-                parentInput = fn.attributes.toolboxParentArgument ?
-                    parent.querySelector(`value[name=${fn.attributes.toolboxParentArgument}]`) :
-                    parent.querySelector(`value`);
+                if (fn.attributes.toolboxParentArgument) {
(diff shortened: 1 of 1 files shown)


pxtblocks/blocklyloader.ts
```diff
@@ -342,11 +342,34 @@ namespace pxt.blocks {
...
+                        // try looking for the first parameter that isn't a field
+                        for (const param of parentInfo.parameters) {
+                            if (parent.querySelector(`field[name=${param.definitionName}]`)) continue;
 
-                parentInput = fn.attributes.toolboxParentArgument ?
-                    parent.querySelector(`value[name=${fn.attributes.toolboxParentArgument}]`) :
-                    parent.querySelector(`value`);
+                            parentInput = document.createElement("value");
+                            parentInput.setAttribute("name", param.definitionName);
+                            parent.appendChild(parentInput);
```
thsparks, line 369: Should we break out of the loop here?

thsparks (approved):
One question. Otherwise LGTM.

jwunderl (approved):

PR #9500, opened 2023-05-03
add button when webusb is disabled in chrome/edge

jwunderl:
re: https://github.com/microsoft/pxt-microbit/issues/4914, I did a quick check and it looks like about 3-3.5% of our chrome / edge users end up with webusb being unavailable. There are several known reasons for that, which will cause us to cut webusb features from being displayed (same as behavior in firefox / safari), but most versions of chrome / edge *should* support it so debugging can be hard.

This ads in a download menu item when 1. on chrome or edge and 2. webusb features are disabled for some reason.
image

This pops up a quick modal with a short explanation of the issue 
image

and just to note explicitly, this only shows up in chromium edge / chrome, safari and firefox will not get this option as the browsers explicitly do not support webusb. If anyone feels that I should put it in for all 'webusb not working' browsers let me know, I'm a bit indecisive on it (the menu is just a repeat of the main button + help on safari / firefox anyway, so maybe it is alright to have a small suggestion to try a different browser. At least exclude ios though)
image

** updated images above re: wording change + returning the button to look like normal

### DIFF
--- a/webapp/src/editortoolbar.tsx
+++ b/webapp/src/editortoolbar.tsx
@@ -11,7 +11,7 @@ import * as identity from "./identity";
 import { ProjectView } from "./app";
 import { clearDontShowDownloadDialogFlag } from "./dialogs";
 import { userPrefersDownloadFlagSet } from "./webusb";
-import { dialogAsync } from "./core";
+import { dialogAsync, hideDialog } from "./core";
 
 type ISettingsProps = pxt.editor.ISettingsProps;
 
@@ -232,6 +232,13 @@ export class EditorToolbar extends data.Component<ISettingsProps, EditorToolbarS
             header: lf("WebUSB unavailable"),
             body: modalBody,
             hasCloseIcon: true,
+            buttons: [
+                {
+                    label: lf("Okay"),
+                    className: "primary",
+                    onclick: hideDialog
+                }
+            ]
         });
     }
 
--- a/webapp/src/editortoolbar.tsx
+++ b/webapp/src/editortoolbar.tsx
@@ -11,6 +11,7 @@ import * as identity from "./identity";
 import { ProjectView } from "./app";
 import { clearDontShowDownloadDialogFlag } from "./dialogs";
 import { userPrefersDownloadFlagSet } from "./webusb";
+import { dialogAsync } from "./core";
 
 type ISettingsProps = pxt.editor.ISettingsProps;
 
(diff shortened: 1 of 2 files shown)


webapp/src/editortoolbar.tsx
```diff
@@ -208,6 +209,39 @@ export class EditorToolbar extends data.Component<ISettingsProps, EditorToolbarS
         this.props.parent.pairAsync();
     }
 
+    protected onCannotPairClick = async () => {
+        pxt.tickEvent("editortools.pairunsupported", undefined, { interactiveConsent: true });
+        const reasonUnsupported = await pxt.usb.getReasonUnavailable();
+        let modalBody: string;
+        switch (reasonUnsupported) {
+            case "security":
+                modalBody = lf("WebUSB appears to have been disabled by browser policy; check your browser's policy page at chrome://policy for more info.");
```
jwunderl, line 218: Opinions on phrasing here would be helpful; my guess is that the number of users who are actively applying policies (vs having a policy applied by a school district) is... hand countable, so maybe this could just be phrased as "Check with your technical administrators if WebUSB can be enabled" or something of the sort

For reference, links to `chrome://whatever` from sites (anchor tags, window.open, etc) are not allowed, user has to type it in (paste it in) themselves. Also, `chrome://policy` does work in edge as well (redirects to `edge://policy`)

kimprice:
It seems weird to me that one of the menu items would say "Cannot Connect Device" since that's not really an option. I think it should still say "Connect Device" but then the popup says it can't connect for whatever reason. Are users able to enable WebUSB from the policy page? When I went to the policy page I didn't see anything about WebUSB.

abchatra:
+1 for what Kim said. Seems weird to show the error always (even though we know it programmatically)

webapp/src/editortoolbar.tsx
jwunderl: image
image

Updated re comments~

webapp/src/editortoolbar.tsx
```diff
@@ -208,6 +209,39 @@ export class EditorToolbar extends data.Component<ISettingsProps, EditorToolbarS
...
+                break;
+            case "oldwindows":
+                modalBody = lf("WebUSB is not available on Windows devices with versions below 8.1.");
```
kimprice, line 221: It looks like the version number above was listed as 6.3. (Line 573 of webusb.ts)

jwunderl: The version listed in the userAgent is misleading, see the comment above it -- Windows NT 6.3 maps to windows8.1 consumer (it looks like with 10+ they made it just map how you would think it does)

webapp/src/editortoolbar.tsx
```diff
@@ -208,6 +209,39 @@ export class EditorToolbar extends data.Component<ISettingsProps, EditorToolbarS
...
+            case "electron":
+                modalBody = lf("WebUSB is not supported in electron.");
+                break;
+            case "notimpl":
+                modalBody = lf("WebUSB is not supported by this browser; please check for updates.");
+                break;
+        }
+
+        dialogAsync({
+            header: lf("WebUSB unavailable"),
```
kimprice, line 232: I think the header should be "Cannot Connect Device" and then in the body it should say something about WebUSB being unavailable. Otherwise, I'm not sure people will make the connection.

kimprice (approved):

webapp/src/editortoolbar.tsx
```diff
@@ -208,6 +209,39 @@ export class EditorToolbar extends data.Component<ISettingsProps, EditorToolbarS
         this.props.parent.pairAsync();
     }
 
+    protected onCannotPairClick = async () => {
+        pxt.tickEvent("editortools.pairunsupported", undefined, { interactiveConsent: true });
+        const reasonUnsupported = await pxt.usb.getReasonUnavailable();
+        let modalBody: string;
+        switch (reasonUnsupported) {
+            case "security":
+                modalBody = lf("WebUSB appears to have been disabled by browser policy; check with your technical adminstrator for details.");
```
eanders-ms, line 218: Can this language be simplified for clarity? It felt wordy to me.

```suggestion
                modalBody = lf("WebUSB is disabled. Check with your admin for help.");
```

jwunderl: I do like simplifying it like this, but I'd think some mention of 'browser policy' somewhere is still probably useful -- web usb is not a particularly common feature so it's very possible it won't be top of mind for the admin if they're asked about it - e.g. maybe it was disabled a few years back due to the chrome cve (I think 2018ish?) and it's been forgotten about since then, or another person set it up. How about:

```suggestion
                modalBody = lf("WebUSB is disabled by browser policies. Check with your admin for help.");
```

maybe?

eanders-ms: makes sense.

eanders-ms (approved):

PR #9501, opened 2023-05-05
patch /static/* images in pxtarget to point to cdnUrl

jwunderl:
Walk through pxtarget and replace all `/static/etc/*.[media-file]` links with cdn urls when we upload.

~For some reason it looks like we must special case appTheme replacements only when replacing `@cdnUrl@` for pxtarget.json, so this will require us to use `pxt.BrowserUtils.patchCdn(image defined in pxtarget.json)`, but that's a pretty small cost / we can figure out the back end replacement weirdness if we really want to. All that function does is replace `@cdnUrl@` with `pxt.webConfig.cdnUrl` though so it's probably fine to leave as is.~ fixed / patching in rest of the code not necessary, see https://github.com/microsoft/pxt/pull/9501#issuecomment-1536758664

Updated so now cloud provider logos, hero banner, web usb connection images, badges, and blockly grid img (minecraft) will all get cdn cached. We probably oughta move https://github.com/microsoft/pxt/pull/9362/files over to a pxtarget field so it's defined per target / doesn't get packed into micro:bit for the release as well, but I can go ahead and do that as a follow up pr probably?

Also fixed it so the uploaded pages get the direct cdn links for e.g. cardLogo, which should fix the excessive amounts of requests we end up getting for these (@aznhassan, i think it covers all the /static/ ones here https://github.com/microsoft/pxt-backend/issues/791, I think we just need to move favicon into /static/ & update all the places it's located for the other one, not sure why sim.webmanifest is missing though / that is pointing at a blob that's already cdnified anyways so maybe a typo / missing `@cdnUrl@` somewhere?)

Side bonus, this fixes paths for uploadtargets in cases where we have a leading `./` (e.g. https://github.com/microsoft/pxt-microbit/blob/master/pxtarget.json#L383), which causes issues when it starts looking in `makecode.microbit.org/app/static/icons/icon.png`

One other subtle difference is that the images linked to in pxtarget will be 'versioned' when referred to using that link (meaning uploadtrg will upload them / if you change the file in docs/static it won't update the things that are linking to it as the urls are 'baked in' at build time.). I'll double check I have this path working by making a few local changes and seeing them, but this is generally a good thing as things linked in pxtarget are typically part of our base ui (logos, icons, explanatory images) where them changing underneath the web app isn't helpful (e.g. when I was updating the webusb flow, I had to upload new files with different names to avoid changing the existing experience before next release.)

a build: https://makecode.microbit.org/app/3802e20b8f923ab25b41ef9e6d07982f30a6d77d-419d74c97d

~I'll make an arcade one to test badges and a minecraft one to test grid tomorrow~

* arcade: https://arcade.makecode.com/app/73b18875fe6d549293dfe3e786417a085515111d-ab1dd5ccd1 ~looks like there's somewhere else I'll need to add in extra patching for --multiplayer's index.html as it still has a few stray "./static"'s, will do before merging this~ fixed
* https://minecraft.makecode.com/app/9a11f105bced8c66713bfb5f12b6a69c1e67a22b-d55ef48319
* and one extra micro:bit build showing me making a local change to hero banner first image and having it show up in uploadtrg: https://makecode.microbit.org/app/71c05979923bd98d318dd6044d7f48399e58bf9c-3bb1f9637b -- this'll be the case for all imgs listed in pxtarget.json, so e.g. if we add another provider to the login screen it'll "just work" in uploadtrg / be versioned with the editor so we don't have docs files that can break editor ui

One thing to note is the replacements in `pxtlib/main.ts` -- as written we do walk the entire pxtarget blob matching strings (pretty efficiently since the regex just ends up as a  `startsWith("@cdnUrl@")`) -- it doesn't appear to be noticeable but if we're concerned I can trim it down to only running on a few subpaths of pxtarget. I'd prefer to leave it as is though, as it seems to behave fine & this way we don't have to keep any extra weird special cases in mind

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -946,8 +946,7 @@ function uploadedArtFileCdnUrl(fn: string): string {
     if (!fn || /^(https?|data):/.test(fn)) return fn; // nothing to do
 
     fn = fn.replace(/^\.?\/*/, "/")
-    const cdnBlobUrl =  "@cdnUrl@/blob/" + gitHash(fs.readFileSync("docs" + fn)) + "" + fn;
-    // console.log(`uploading art file ${fn} to ${cdnBlobUrl}`);
+    const cdnBlobUrl = "@cdnUrl@/blob/" + gitHash(fs.readFileSync("docs" + fn)) + "" + fn;
     return cdnBlobUrl;
 }
 
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -957,9 +957,9 @@ function gitHash(buf: Buffer) {
 }
 
 function uploadCoreAsync(opts: UploadOptions) {
-    let targetConfig = readLocalPxTarget();
-    let defaultLocale = targetConfig.appTheme.defaultLocale;
-    let hexCache = path.join("built", "hexcache");
+    const targetConfig = readLocalPxTarget();
+    const defaultLocale = targetConfig.appTheme.defaultLocale;
+    const hexCache = path.join("built", "hexcache");
     let hexFiles: string[] = [];
 
     if (fs.existsSync(hexCache)) {
@@ -1065,7 +1065,7 @@ function uploadCoreAsync(opts: UploadOptions) {
         }
     }
 
-    let replFiles = [
+    const replFiles = [
         "index.html",
         "embed.js",
         "run.html",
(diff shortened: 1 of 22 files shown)


jwunderl:
Yeah, the files get passed in through https://github.com/microsoft/pxt/blob/cachifyPxtargetJsonFields/cli/cli.ts#L1072 which gets patched in here https://github.com/microsoft/pxt/blob/cachifyPxtargetJsonFields/webapp/src/serviceworker.ts#L137

jwunderl:
fixed --multiplayer refs, and at the same time found out where appTheme was getting special cased / made it so patchCdn calls are not necessary for any new calls / /static/ urls in pxtarget 'just work™️' by patching them when install pxt.appTarget. see https://github.com/microsoft/pxt/pull/9501/commits/b8b09cb93a440fe43222728acceec18ec237e653

riknoll (approved):

PR #9506, opened 2023-05-10
when spawning sim from broadcast, defer msg till spawned

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/4986

~If we want I can make it a little more extensible by tagging each sim with an id and including that in `deferredMessages`, so if we extend to > 2 sims in the future it'd still work as expected, but that feels pretty low priority / it's been years with just 2 in the first place. But it wouldn't be hard / would make the code a little more explicit so just let me know what you prefer, whoever ends up reviewing this 😸~ I just went ahead and did it

here's the build without the fix: https://makecode.microbit.org/beta#pub:_KPJbgURehgML - press a and see nothing happens on first press

here's the build with the fix; https://makecode.microbit.org/app/40bfbd4a142932c6ba83c231537276246a59f908-2dfbec65d1#pub:_KPJbgURehgML press a and see it is happy as soon as it comes in

### DIFF
--- a/pxtsim/simdriver.ts
+++ b/pxtsim/simdriver.ts
@@ -314,6 +314,8 @@ namespace pxsim {
 
             if (frameID) frames = frames.filter(f => f.id === frameID);
 
+            let isDeferrableBroadcastMessage = false;
+
             const broadcastmsg = msg as pxsim.SimulatorBroadcastMessage;
             if (source && broadcastmsg?.broadcast) {
                 // if the editor is hosted in a multi-editor setting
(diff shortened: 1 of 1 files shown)


pxtsim/simdriver.ts
```diff
@@ -367,6 +369,7 @@ namespace pxsim {
                                 this.startFrame(messageFrame);
                             }
                         } else {
+                            isDeferrableBroadcastMessage = true;
```
kimprice, line 372: How come this doesn't need to be in the start second simulator condition? Are two messages sent on the initial radio send?

jwunderl: if you modify the program I wrote to send two radio messages it would do so -- it could take a moment for the frame to start up so passing it through the case where it'll check whether it's been loaded before sending feels right to me~

eanders-ms (approved):

PR #9507, opened 2023-05-11
don't extension search on blur + fix for blur losing input on repeated blur

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/4959; see https://github.com/microsoft/pxt-microbit/issues/4959#issuecomment-1542583536 for an easier repro.

~The root of the issue is that onBlur is kinda a messy place to start the search from -- it running means you've already clicked / navigated elsewhere (e.g. the scrollbar or one of the extension buttons), so the results will be changing from underneath you (sort of like if you're reading a website and about to click a link, and then a big add pops in and pushes down the link you were going to click; it feels bad), and typically navigating away can mean that something has caught your eye in the first place -- so the update will potentially cause that to disappear.~

~Tagging in hassan / abhijith here as well as a check on if this feels fine for hitting our search endpoint with; I could bump it up to be a little longer (1.5seconds? 2 seconds?) if we want?~

If we don't wanna do the search on change for any reason, I'd lean towards making this pr just delete `onBlur={onSearchBarChange}`, add in the `preserveValueOnBlur={true}`, and leave the rest as is. << decided to do this, to avoid unnecessary network traffic

~Build with change: https://makecode.microbit.org/app/71e42671356a4958e8509ec38d919eb348f21005-3d945dca17#editor~

Also small fix for losing input after clicking in and out of the input a few times in arow, I'm a little surprised `preserveValueOnBlur` isn't the default behavior / feels like it should be?

### DIFF
--- a/webapp/src/extensionsBrowser.tsx
+++ b/webapp/src/extensionsBrowser.tsx
@@ -47,11 +47,6 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
         setSearchFor(newValue || "");
     }
 
-    const onSearchBarChangeDebounced: (newValue: string) => void = React.useMemo(
-        () => pxt.Util.debounce(onSearchBarChange, 1000),
-        []
-    );
-
     useEffect(() => {
         updateExtensionTags();
         updatePreferredExts();
@@ -522,7 +517,6 @@ export const ExtensionsBrowser = (props: ExtensionsProps) => {
                         placeholder={lf("Search or enter project URL...")}
                         ariaLabel={lf("Search or enter project URL...")}
                         onEnterKey={onSearchBarChange}
-                        onChange={onSearchBarChangeDebounced}
                         preserveValueOnBlur={true}
                         icon="fas fa-search"
                     />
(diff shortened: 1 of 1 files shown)


kimprice:
I slightly prefer the search to happen only on enter, but I don't feel strongly.

kimprice (approved):

jwunderl:
Yeah, we could choose to make it on just enter. We do have a few other search inputs that build up the 'will apply the search after a moment automatically' though (blocks toolbox, project search in 'my projects' view, minecraft's searchable 'blocks' field editor), and vscode's extension search behaves this way too (I think maybe a 500ms delay or so only? But 1-1.5 seconds feels okay ish to me when the audience likely includes students in the ~10-15wpm range where 500ms would potentially catch them each character):

extension-search-in-vscode

kimprice:
In that case, I think we should be consistent with the way we do search inputs elsewhere. Do we have a delay for our other search inputs?

jwunderl:
The other existing ones are all local searches which are quick enough they don't really need the debounce for the most part; I suppose closest would be typechecking / compiling which I think is stuck at .5-1 seconds

That said, chatted with @abchatra and we decided to keep it to 'on enter only' to ease network traffic / unnecessary backend work

PR #9508, opened 2023-05-11
Add "Duplicate with Parent" option for blocks in the toolbox.

thsparks:
This option allows blocks to specify that they should be duplicated in the toolbox, and the duplicate should be wrapped in a parent. It functions much like `toolboxParent` (and `toolboxParentArgument`) but adds a new block instead of wrapping the existing one.

Needed for https://github.com/microsoft/pxt-microbit/pull/5123 so we can include the sound effect block by itself *and* include the sound effect block wrapped in the play block parent without having to create a whole new play block. (See comment: https://github.com/microsoft/pxt-microbit/pull/5123#issuecomment-1544424908)


kimprice (approved):
LGTM!

jwunderl (approved):

PR #9512, opened 2023-05-16
add block attribute for grid that emits to string literal

jwunderl:
Basically, just the grid field editor that doesn't require explicit sim / cpp side handling / emits to a string template

```typescript
//% color="#AA278D" weight=100
namespace hello {
    //% weight=75
    //% blockId=buildagrid block="my special grid"
    //% imageLiteralColumns=7
    //% imageLiteralRows=4
    //% gridLiteral=1
    export function createImage(leds: string) {
        console.log(leds)

    }

    //% weight=70
    //% blockId=gridthatreturns block="my special grid that returns"
    //% imageLiteralColumns=7
    //% imageLiteralRows=2
    //% gridLiteral=1
    export function createImageAndReturn(leds: string) {
        return leds;
    }
}
```

ends up as 

image

and doesn't go through the all the special cases in the -> binary js / -> cpp compilers, so it's useable for defining grids that aren't explicitly micro:bit images (and therefore doesn't just immediately cause a runtime error anywhere outside of micro:bit) - in particular where the bitpacking just doesn't make sense like in sim only editors.

(I'll add compile / decompile tests / a note in the docs if this feels okay, just easier to justify writing them after it's approved :) )

### DIFF
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -806,6 +806,7 @@ declare namespace ts.pxtc {
         jresURL?: string;
         iconURL?: string;
         imageLiteral?: number;
+        gridLiteral?: number;
         imageLiteralColumns?: number; // optional number of columns
         imageLiteralRows?: number; // optional number of rows
         imageLiteralScale?: number; // button sizing between 0.6 and 2, default is 1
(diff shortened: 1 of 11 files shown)


riknoll (approved):
add tests please!

jwunderl:
added a test case to both decompile test & the karma tests, tagged everyone else in more as an FYI / pointer that this is a part of the code that is interesting but hasn't changed too much recently / will merge after tests go through~

PR #9513, opened 2023-05-16
allow large images in webusb dialog

jwunderl:
Allow slightly larger images in download flow; used it in the build for https://github.com/microsoft/pxt-microbit/pull/5144, might make some more minor adjustments on this branch if we go with that change

### DIFF
--- a/theme/common.less
+++ b/theme/common.less
@@ -1636,7 +1636,7 @@ p.ui.font.small {
     margin: 0 auto;
 }
 
-.download-dialog-image img {
+.download-dialog-image img:not(.large) {
     max-height: 200px;
     object-fit: contain;
 }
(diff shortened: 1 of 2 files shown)


theme/common.less
```diff
@@ -1636,7 +1636,7 @@ p.ui.font.small {
     margin: 0 auto;
 }
 
-.download-dialog-image img {
+.download-dialog-image img:not(.large) {
```
kimprice, line 1639: Is `.download-dialog-image img` used anywhere else? If not, should we just remove this?

jwunderl: The other images in the browser download + webusb pairing flow all hit this

kimprice (approved):

PR #9521, opened 2023-05-20
Add Duration to Sound Effect Editor Duration Dropdown

thsparks:
Add a dropdown menu to the sound effect editor's duration field:

image

To do this, I added an optional `options` property that can be passed into a react-common `Input`. When present, a new button will appear that allows you to expand a dropdown list of options and select a value from there. Options are passed in as a map, where the text that gets displayed in the dropdown (i.e. "5 seconds") is mapped to the value that gets put in the `Input` when the option is selected (i.e. "5000").

I'm not in love with this approach. I think there's a lot of redundancy between what I'm adding inside `Input` and what already exists inside `Dropdown`, but I toyed with a few other approaches and this one ended up feeling like the least of the evils. The others I tried were:

1. Use built-in input/value dropdown from browser: I didn't go with this approach because it looked confusing and really inconsistent with the rest of our UI (and different per browser).

2. Add an optional `Input` component inside the existing `Dropdown` component: When I tried this approach, I added a flag to optionally replace the button inside the react-common `Dropdown` component with an `Input`, but I struggled to connect the value of the `Input` to the value in the `Dropdown` without some messy refactoring, and some of the arrow-key behavior felt pretty unintuitive.

3. Add an optional `Dropdown` component inside the existing `Input` component: When I tried this approach, I added a `Dropdown` inside the `Input` component and used css to hide everything but the expand/collapse button in the `Dropdown`. When a `Dropdown` option was chosen, I used that to change the value of the `Input`. With some css fiddling, this looked okay, but it didn't work well with screen readers, which still "saw" the `Dropdown`'s selected value when it was different from the `Input` (for example, when the user typed in a value that wasn't in the `Dropdown`'s set of options).

~~Upload Target: https://makecode.microbit.org/app/5d1921aebf06395f1f5676532b71038a771bed9b-dab60284d3~~
New Upload Target: https://makecode.microbit.org/app/bd1d631130421a362f294c6d2ac128bea6c1beba-f7e7245a4d

Fixes https://github.com/microsoft/pxt-microbit/issues/4702

### DIFF
--- a/react-common/components/controls/Input.tsx
+++ b/react-common/components/controls/Input.tsx
@@ -74,6 +74,10 @@ export const Input = (props: InputProps) => {
         if (selectOnClick) {
             (evt.target as any).select()
         }
+
+        if (options && !expanded) {
+            setExpanded(true);
+        }
     }
 
     const changeHandler = (e: React.ChangeEvent<any>) => {
@@ -129,6 +133,8 @@ export const Input = (props: InputProps) => {
         if (onOptionSelected) { 
             onOptionSelected(value);
         }
+
+        document.getElementById(id)?.focus();
     }
 
     const getDropdownOptionId = (option: string) => {
--- a/react-common/components/controls/Input.tsx
+++ b/react-common/components/controls/Input.tsx
@@ -98,6 +98,9 @@ export const Input = (props: InputProps) => {
 
     const iconClickHandler = () => {
         if (onIconClick) onIconClick(value);
+    }
+
+    const expandButtonClickHandler = () => {
         if(options) {
             setExpanded(!expanded);
         }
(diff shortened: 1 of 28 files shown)


react-common/components/controls/Input.tsx
```diff
@@ -87,6 +100,12 @@ export const Input = (props: InputProps) => {
         if (onIconClick) onIconClick(value);
     }
 
+    const expandButtonClickHandler = () => {
+        if(options) {
```
jwunderl, line 104:
```suggestion
        if (options) {
```

nit

react-common/components/controls/Input.tsx
```diff
@@ -96,8 +115,28 @@ export const Input = (props: InputProps) => {
         }
     }
 
+    const containerBlurHandler = (e: React.FocusEvent) => {
+        if(expanded && !container.contains(e.relatedTarget as HTMLElement)) {
```
jwunderl, line 119:
```suggestion
        if (expanded && !container.contains(e.relatedTarget as HTMLElement)) {
```

nit

react-common/components/controls/Input.tsx
```diff
@@ -96,8 +115,28 @@ export const Input = (props: InputProps) => {
...
+            setExpanded(false);
+        }
+    }
+
+    const optionClickHandler = (option: string) => {
+        setExpanded(false);
+
+        const value = options[option];
+        setValue(value);
+        if(onOptionSelected) {
```
jwunderl, line 129:
```suggestion
        if (onOptionSelected) { 
```
nit

jwunderl (commented):
It took me a bit to realize I had to click into the arrow itself to get the dropdown (/ if I didn't know that it was there, I might have just thought it wasn't working); I expected it to show the dropdown automatically like the datalist field, and move focus down to the dropdown when I moved with arrow keys, rather than being another focus target to toggle, which then makes you tab again to get to the list itself -- this would match browser behavior in 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist (for what it's worth, for buttons like these I always click the item rather than the arrow itself as a habit)

Also minor thing but I think it ought to move focus back onto the text input on something being selected from the field (probably fine as part of `ondurationchange` as I could see it having other options for non text input?) -- that also is the behavior of the default datalist. Right now it looks like it softly 'drops' the focus at the current location (after selecting tab moves to effect / shift tab moves back to the toggle)

jwunderl (approved):

PR #9523, opened 2023-05-20
Webusb: catch errors more errors, stop saying success when pairing not complete

jwunderl:
https://makecode.microbit.org/app/ffbdd7bc4924b3b24a8d03c86d3dd5b68a77b717-cc763a8701

I'll be double checking this on monday, I'm sure I messed something up a bit, but this should overall catch the errors more consistently, route to 'congrats you succeeded' screen more accurately (both confirming connection was actually successful before congratulating user, and ignoring false negatives when e.g. escaping from browser pair screen with an already pair micro:bit), giving download dialog consistently when downloading to device was unsucessfully, and deploy automatically when 'download success -> pair now for faster downloads -> successful pairing' path occurs

Should:
* fix https://github.com/microsoft/pxt-microbit/issues/5146
* fix https://github.com/microsoft/pxt-microbit/issues/5090
* fix a few other slight variants on the above that other team members have sent me that didn't get their own bugs filed ^^
* fix https://github.com/microsoft/pxt-microbit/issues/4425 (remaining portion of the issue is that it still shows micro:bit connection when you have two paired browsers open at once; I can look at that but there are several other features waiting / I would be very surprised if a significant amount of users actually hit this, since it requires 2 browsers open / attempting to connect concurrently which seems... rare. We don't have any control at all over 'taking focus' or even really notifying them of it, as the behavior would be the same as if they e.g. unplugged mid flash / or there was any other hardware error.)
* fix https://github.com/microsoft/pxt-microbit/issues/5157 (this makes the 'downloaded -> saw suggestion in dialog complete modal -> paired' flow now deploy immediately, but left the `...` menu as 'only pairing no forced download' for the same 'probably nobody ends up seeing that anyway' type of reason I mentioned in standup)
* fix for https://github.com/microsoft/pxt-microbit/issues/5130
* When the failure occurs and it downloads as a file, consistently show 'file downloaded' instead of showing it half the time & just downloading without display half the time

(sorry the code is a little uglier / more complicated now, so many edge cases D:)

### DIFF
--- a/pxteditor/editor.ts
+++ b/pxteditor/editor.ts
@@ -377,7 +377,7 @@ namespace pxt.editor {
         showPackageDialog(query?: string): void;
         showBoardDialogAsync(features?: string[], closeIcon?: boolean): Promise<void>;
         checkForHwVariant(): boolean;
-        pairAsync(): Promise<void>;
+        pairAsync(): Promise<boolean>;
 
         createModalClasses(classes?: string): string;
         showModalDialogAsync(options: ModalDialogOptions): Promise<void>;
--- a/pxteditor/editorcontroller.ts
+++ b/pxteditor/editorcontroller.ts
@@ -574,7 +574,7 @@ namespace pxt.editor {
                                         .then(() => projectView.printCode());
                                 }
                                 case "pair": {
-                                    return projectView.pairAsync();
+                                    return projectView.pairAsync().then(() => {});
                                 }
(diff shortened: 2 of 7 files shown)


eanders-ms (approved):
Nice! I now get a "failed to connect" dialog when I try to pair, which is expected because WebUSB has never quite worked on my machine.

PR #9525, opened 2023-05-22
Fix Tutorial Scroll Gradient on Resize

thsparks:
There were a few different places that were all checking if the scrollGradient should exist, and two of them didn't take into account the current scroll position. I updated both of those places to point to the third approach, which does take scroll position into account.

Fixes https://github.com/microsoft/pxt-microbit/issues/5066

Upload target: https://makecode.microbit.org/app/20bac495db713761f5ed13aec44b985873d249be-c357ca8155


jwunderl (approved):

kimprice (approved):
Looks good!

PR #9527, opened 2023-05-23
Auto start the editor tour on first project

kimprice:
Closes https://github.com/microsoft/pxt-microbit/issues/5149

This change will auto-start the editor tour when a user creates a new project and they have zero other projects. So, if you delete all your projects, the next project you open will trigger the tour again. The tour is also still available in the help menu.

This change depends on a change in pxt-microbit to add a welcome step to the editor tour markdown.

Here's a build: https://makecode.microbit.org/app/ae6a0123d79e5e04105f1a55ce497865d6942e07-a402fe3471
Note: The welcome step won't show up in the build since it's from the markdown in pxt-microbit and docs are live.

autoStartTour

### DIFF
--- a/pxteditor/editor.ts
+++ b/pxteditor/editor.ts
@@ -119,6 +119,7 @@ namespace pxt.editor {
         extensionUnderTest?: string; // workspace id of the extension under test
         skillmapProject?: boolean;
         simTheme?: Partial<pxt.PackageConfig>;
+        firstProject?: boolean;
     }
 
     export interface ExampleImportOptions {
--- a/pxteditor/editor.ts
+++ b/pxteditor/editor.ts
@@ -239,7 +239,7 @@ namespace pxt.editor {
         exportAsync(): Promise<string>;
 
         newEmptyProject(name?: string, documentation?: string, preferredEditor?: string): void;
-        newProject(options?: ProjectCreationOptions): void;
+        newProject(options?: ProjectCreationOptions, isFirstProject?: boolean): void;
         createProjectAsync(options: ProjectCreationOptions): Promise<void>;
         importExampleAsync(options: ExampleImportOptions): Promise<void>;
         showScriptManager(): void;
--- a/pxteditor/editor.ts
+++ b/pxteditor/editor.ts
@@ -240,7 +240,7 @@ namespace pxt.editor {
         exportAsync(): Promise<string>;
 
         newEmptyProject(name?: string, documentation?: string, preferredEditor?: string): void;
-        newProject(options?: ProjectCreationOptions, isFirstProject?: boolean): void;
+        newProject(options?: ProjectCreationOptions): void;
         createProjectAsync(options: ProjectCreationOptions): Promise<void>;
         importExampleAsync(options: ExampleImportOptions): Promise<void>;
         showScriptManager(): void;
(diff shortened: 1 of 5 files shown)


webapp/src/app.tsx
```diff
@@ -2787,13 +2787,18 @@ export class ProjectView
         })
     }
 
-    newProject(options: ProjectCreationOptions = {}) {
+    newProject(options: ProjectCreationOptions = {}, isFirstProject?: boolean) {
```
jwunderl, line 2790: we don't need to add a new parameter for this one, just add it into `ProjectCreationOptions` here https://github.com/microsoft/pxt/blob/master/pxteditor/editor.ts#L121 and pass in the blob (e.g. how `name` and `languageRestriction` are passed in down in projects.tsx)

jwunderl (approved):

pxteditor/editor.ts
```diff
@@ -239,7 +240,7 @@ namespace pxt.editor {
         exportAsync(): Promise<string>;
 
         newEmptyProject(name?: string, documentation?: string, preferredEditor?: string): void;
-        newProject(options?: ProjectCreationOptions): void;
+        newProject(options?: ProjectCreationOptions, isFirstProject?: boolean): void;
```
jwunderl, line 243:
```suggestion
        newProject(options?: ProjectCreationOptions): void;
```

webapp/src/app.tsx
```diff
@@ -2793,7 +2793,12 @@ export class ProjectView
         return this.createProjectAsync(options)
             .then(() => this.autoChooseBoardAsync())
             .then(() => Util.delay(500))
-            .finally(() => core.hideLoading("newproject"));
+            .finally(() => {
+                core.hideLoading("newproject");
+                if (options?.firstProject) {
```
jwunderl, line 2798: Ah wait, one more thing; this should probably go behind a target flag, in case a target doesn't have editor tours (or has an editor tour but doesn't want it automatic)

kimprice: Oh yea good catch!

webapp/src/projects.tsx
```diff
@@ -570,7 +570,6 @@ export class ProjectsCarousel extends data.Component<ProjectsCarouselProps, Proj
         this.closeDetail = this.closeDetail.bind(this);
         this.closeDetailOnEscape = this.closeDetailOnEscape.bind(this);
         this.reload = this.reload.bind(this);
-        this.newProject = this.newProject.bind(this);
```
thsparks, line 573: Is the reason we no longer need this because of the change in how the function get scalled in `onClick` below? This kind of binding is still a bit of a mystery to me...

thsparks (approved):
LGTM! I had a question, but more for my own information than any recommended change.

webapp/src/projects.tsx
jwunderl: Yes -- the issue with passing this.newProject is that javascript basically loses the 'this` context for explicit methods when passing (here's an article on it / the weirdness https://www.freecodecamp.org/news/this-is-why-we-need-to-bind-event-handlers-in-class-components-in-react-f7ea1a6f93eb/) -- for what it's worth, my preference is just to switch to arrow function (`newProject = () => {`), which does a similar job. Or, well, using functional components in new code, which sidesteps this issue completely.

thsparks: Thanks for the info! Very helpful article.

Jaqster (approved):
Looks awesome!  Thank you Kim!

PR #9531, opened 2023-05-25
Ignore invalid locales files in extensions

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/5186; we should not crash when loading extensions with invalid localization files; just ignore them. (if we want I can walk up the loading path here a bit and find a place to place to catch more broadly, but I don't see errors here often / feels like something we can do later as clean up, probably go to an "extension failed to load" dialog of something?).

@abchatra not sure if we'd care about ticks for this, as it should only really apply to third party extensions? I left it as just a log for now similar to the existing 'ignored translations' logic without the `reportError` https://github.com/microsoft/pxt/blob/ignoreInvalidLocalesFiles/pxtlib/service.ts#L767 so extension developers can check that without a tick for that reason, happy to add one in though if you want it tracked.

test build: https://makecode.microbit.org/app/8b8692a266bacc35344bdf582a3e39fa2cd76f22-9a85c1acb4

### DIFF
--- a/pxtlib/package.ts
+++ b/pxtlib/package.ts
@@ -969,11 +969,11 @@ namespace pxt {
             const files = this.config.files;
 
             let fn = `_locales/${initialLang}/${filename}-strings.json`;
-            if (files.indexOf(fn) === -1){
+            if (initialLangLowerCase && files.indexOf(fn) === -1) {
                 fn = `_locales/${initialLangLowerCase}/${filename}-strings.json`;
-                if (files.indexOf(fn) === -1) {
-                    fn = `_locales/${baseLang}/${filename}-strings.json`;
-                }
+            }
+            if (baseLang && files.indexOf(fn) === -1) {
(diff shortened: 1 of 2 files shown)


eanders-ms (approved):
+1 for de-nesting.

abchatra:
I would add a tick with the extension slug. Atleast we could see how many are facing this issue.

We should also log error message to console. I am assuming pxt.log does that. That is sufficient for the extension author to ensure their translation is working well.

jwunderl:
Okay, updated to send a tick. Also changed the log over to an error reporting, and made the other mismatched params one send a tick as well (I thought it was sending one already but `.reportError` only logs to console, not a tick) -- so updated the logging of both.

Updated build w/ thomas's log tick events flag turned on https://makecode.microbit.org/app/ec4035b887de2789bced79936d4af2d568b2eee6-5fccd8d85a?consoleticks=1#editor, if you switch to catalan & import cutebot you'll see both types of exceptions: 
image

PR #9532, opened 2023-05-27
core.startLoading add option for a percentage loaded to display progress bar

jwunderl:
will clean up css / make it a little prettier / maybe add a percent completed label / etc on tuesday, just putting this up before weekend / it's functional anyways

build: https://makecode.microbit.org/app/d260c65383df540eae0131ead16171faeb3923db-3df33463f1

flash-progress-bar

### DIFF
--- a/pxtlib/packetio.ts
+++ b/pxtlib/packetio.ts
@@ -16,7 +16,7 @@ namespace pxt.packetio {
         isConnected(): boolean
         isConnecting(): boolean
         // flash the device, does **not** reconnect
-        reflashAsync(resp: pxtc.CompileResult): Promise<void>;
+        reflashAsync(resp: pxtc.CompileResult, progressCallback?: (percentageComplete: number) => void): Promise<void>;
 
         onCustomEvent: (type: string, payload: Uint8Array) => void;
         sendCustomEventAsync(type: string, payload: Uint8Array): Promise<void>;
--- a/react-common/components/controls/ProgressBar.tsx
+++ b/react-common/components/controls/ProgressBar.tsx
@@ -28,7 +28,7 @@ export const ProgressBar = (props: ProgressBarProps) => {
 
     return (
         <div className={classList("common-progressbar-wrapper", className)}>
-            {label && <label className="common-textarea-label">
+            {label && <label className="common-progressbar-label">
                 {label}
             </label>}
             <progress
--- a/react-common/components/controls/ProgressBar.tsx
+++ b/react-common/components/controls/ProgressBar.tsx
@@ -1,0 +1,48 @@
+import * as React from "react";
+import { classList, ControlProps } from "../util";
+
+export interface ProgressBarProps extends ControlProps {
+    value: number;
+    max?: number; // default: 1
+    title?: string;
+    label?: string;
+
+    ariaValueText?: string;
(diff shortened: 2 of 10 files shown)


skillmap/src/components/makecodeFrame.tsx
```diff
@@ -166,7 +164,7 @@ class MakeCodeFrameImpl extends React.Component<MakeCodeFrameProps, MakeCodeFram
 
     protected onMessageReceived = (event: MessageEvent) => {
         const data = event.data as pxt.editor.EditorMessageRequest;
-        if (this.state.frameState === "opening-project") this.setState({ loadPercent: Math.min((this.state.loadPercent || 0) + 4, 95) });
+        if (this.state.frameState === "opening-project") this.setState({ loadPercent: Math.min((this.state.loadPercent || 0) + 7, 95) });
```
jwunderl, line 167: This change is completely unnecessary and unrelated, but we never reach past ~half and for consecutive loads often not even a quarter  of the progress bar, and that felt bad to me :( with this we get 'close' to completion right around when it disappears and shows the editor.

webapp/src/simulator.ts
```diff
@@ -307,6 +307,8 @@ export function run(pkg: pxt.MainPackage, debug: boolean,
         }
     }
 
+    const theme = pkg.config.theme || (pxt.appTarget.appTheme.matchWebUSBDeviceInSim && pxt.packetio.isConnected() && pxt.packetio.deviceVariant());
```
jwunderl, line 310: I added this deviceVariant check in a previous pr and just noticed it was occasionally giving some spurious warnings in the console. very minor adjustment to check connected state here (basically no functional difference besides not logging when the result would be undefined anyways), didn't feel worth extracting from the webusb related pr :)

jwunderl (commented):
Updated theming to match rest of default loader:
flash-progress-bar-themed

pxtlib/packetio.ts
```diff
@@ -16,7 +16,7 @@ namespace pxt.packetio {
         isConnected(): boolean
         isConnecting(): boolean
         // flash the device, does **not** reconnect
-        reflashAsync(resp: pxtc.CompileResult): Promise<void>;
+        reflashAsync(resp: pxtc.CompileResult, progressCallback?: (percentageComplete: number) => void): Promise<void>;
```
thsparks, line 19: Where does this actually get implemented? I'm not seeing it in the PR but could just be missing something...

thsparks: Oh nvm looks like https://github.com/microsoft/pxt-microbit/pull/5200/files

thsparks (approved):

pxtlib/packetio.ts
jwunderl: Ah yeah good point, would have been nice if I had added an annotation here; there is a packetio implementation here https://github.com/microsoft/pxt/blob/coreStartLoadingSpinner/pxtlib/hf2.ts#L185 which is the one used for arcade, I can look at making use of it there when we get back to arcade but arcade is still using the old pairing flow / didn't feel worth doubling (or quadrupling if I wanted to test the different arcade variants) for this pr / focused on micro:bit for now.

PR #9533, opened 2023-05-30
Make Top-Instructions Resizable

thsparks:
This makes the instructions resizable (when in the top-instructions view). I created a new `VerticalResizeContainer` so it will _theoretically_ be easier to make similar changes easier in the future, if we decide to make other components resizable. (I did consider adding horizontal resizing too, but since it's not needed for this change, I decided to hold off. Can be added later if/when we need it.)

The change is primarily based on @jwunderl's https://github.com/microsoft/pxt/pull/9102.

Try it here: https://makecode.microbit.org/app/c07cbb23d6fd576dccb1afa91e15a1566e0a2c68-80c4cf5b23

Fixes https://github.com/microsoft/pxt-microbit/issues/5024

ResizeInstructions

### DIFF
--- a/react-common/components/controls/VerticalResizeContainer.tsx
+++ b/react-common/components/controls/VerticalResizeContainer.tsx
@@ -1,0 +1,81 @@
+import * as React from "react";
+import { classList, ContainerProps } from "../util";
+
+export interface VerticalResizeContainerProps extends ContainerProps {
+    minHeight?: string;
+    maxHeight?: string;
+    heightProperty?: string;
+    style?: any;
+    onResizeDrag?: () => void;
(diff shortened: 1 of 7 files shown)


jwunderl (approved):
Nice to see this sneak in :) looks good to me. For next arcade release we should extend this to the arcade tutorials as well like in my pr (and maybe sim frame, with auto collapse when dragged < 10 px or something), but that seems easy enough to extend here (e.g. eventually rename the class `ResizeContainer` & accept horizontal / vertical resizes / maybe even both if we're feeling fancy) but good as is / no need to take that on at this time.

PR #9537, opened 2023-05-31
Use forget device api to directly disconnect when available

jwunderl:
close https://github.com/microsoft/pxt-microbit/issues/5109
also part of https://github.com/microsoft/pxt-microbit/issues/5164, I'll close that one in a separate pr over in micro:bit adding the two images in this pr.

Could use wording critique on the two dialogs I'm updating:

New short dialog for 'we successfully disconnected your device' case, where forget worked:

image
gif is just the 'connected device' one reversed. This dialog is purely informational, besides a suggestion to unplug (but that only really matters if they wish to reconnect right away, which is a bit of a niche use case.)

The other one is the existing dialog, just updated the wording a bit to accommodate upcoming browser changes (getting rid of lock icon). This one shouldn't be hit for up to date browsers that support the api, just older chrome / edge, or if somehow something errored out in the forget api (I induced it here by just causing a null pointer exception by setting something to undefined in `forgetAsync`):

image 

build: https://makecode.microbit.org/app/b4c4c2fcdeff73191ccb9a3216f8ee199d47aef2-da9e91d6ab

### DIFF
--- a/pxtlib/webusb.ts
+++ b/pxtlib/webusb.ts
@@ -166,6 +166,8 @@ namespace pxt.usb {
         isochronousTransferIn(endpointNumber: number, packetLengths: number[]): Promise<USBIsochronousInTransferResult>;
         isochronousTransferOut(endpointNumber: number, data: BufferSource, packetLengths: number[]): Promise<USBIsochronousOutTransferResult>;
         reset(): Promise<void>;
+        // chromium 101+
+        forget?(): Promise<void>;
     }
 
     class WebUSBHID implements pxt.packetio.PacketIO {
@@ -262,6 +264,17 @@ namespace pxt.usb {
                     this.clearDev()
                     return U.delay(500)
                 })
+        }
+
+        async forgetAsync(): Promise<boolean> {
+            if (!this.dev?.forget)
+                return false;
+            try {
+                await this.dev.forget();
+                return true;
+            } catch (e) {
+                return false;
+            }
         }
 
         reconnectAsync() {
@@ -492,7 +505,7 @@ namespace pxt.usb {
             })
     }
 
-    async function tryGetDevicesAsync(): Promise<USBDevice[]> {
+    export async function tryGetDevicesAsync(): Promise<USBDevice[]> {
         log(`webusb: get devices`)
         try {
             const devs = await ((navigator as any).usb.getDevices() as Promise<USBDevice[]>);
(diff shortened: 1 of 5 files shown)


webapp/src/webusb.tsx
```diff
@@ -470,18 +470,67 @@ function theme() {
...
+    const jsxd = () => (
+        <div className={`ui ${columns} column grid padded download-dialog`}>
+            <div className="column">
+                <div className="ui">
+                    <div className="content">
+                        <div className="description">
+                            {lf("Your {0} has been disconnected.", boardName)}
+                            <br />
+                            <br />
+                            {lf("Unplug it from your computer and disconnect any battery to fully reset it.")}
```
kimprice, line 507: Should they just click the reset button instead of disconnecting the battery?

webapp/src/webusb.tsx
```diff
@@ -470,18 +470,67 @@ function theme() {
...
-        {lf("You can unpair your {0} if the WebUSB download is malfunctioning. Click on the lock icon and uncheck your device.", boardName)}
-    </p>
-        <img className="ui image centered medium" src={"./static/webusb/unpair.gif"} alt={lf("A gif showing how to unpair the {0}", boardName)} />
+    const unpairImg = theme().browserUnpairImage;
+    const jsx = <div>
+        <p>
+            {lf("You can unpair your {0} if the WebUSB download is malfunctioning.", boardName)}
+        </p>
+        <p>
+            {lf("Click on the icon on the left side of the address bar and uncheck your device.")}
```
kimprice, line 479: Should it be "unmount your device" or will it actually be a check box?

pxtlib/webusb.ts
```diff
@@ -166,6 +166,8 @@ namespace pxt.usb {
         isochronousTransferIn(endpointNumber: number, packetLengths: number[]): Promise<USBIsochronousInTransferResult>;
         isochronousTransferOut(endpointNumber: number, data: BufferSource, packetLengths: number[]): Promise<USBIsochronousOutTransferResult>;
         reset(): Promise<void>;
+        // chromium 101+
```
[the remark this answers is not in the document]
jwunderl: the api was added in chromium v101 https://developer.mozilla.org/en-US/docs/Web/API/USBDevice/forget

webapp/src/webusb.tsx
jwunderl: No, this is pushing towards a power cycle, not a soft reset. Most probably won't read this dialog anyway, but if they only do a power cycle it e.g. would probably be in an inconsistent state on the device connection (daplink) side if they immediately try to pair again which may or may not work (depending on whether the device firmware handles it, e.g. not storing any bad addresses / state / etc in a way our connection logic can't reset)

webapp/src/webusb.tsx
jwunderl: It's a bit different depending on both browser and version, e.g. an relatively recent version of chrome I have looks like this: 
image
with chrome canary looking similar but with a little more emphasis on reset
image

I mostly left the wording how it originally was there since this would typically only be shown to people on older chrome installs anyways, and I'm assuming 'check' was closer to what the dialog looked like when it was originally written (chrome 101 was apr 2022)

kimprice (approved):

PR #9538, opened 2023-06-01
Only show micro:bit board (no external parts) for thumbnail recorder

kimprice:
Closes https://github.com/microsoft/pxt-microbit/issues/5155

This also fixes a lingering issue where the thumbnail had two simulators duplicated. It was fixed for screenshots before, but it became apparent during this bug fix that it was lingering for gifs. (See https://github.com/microsoft/pxt-microbit/issues/4988)

This change does not impact Arcade.

thumbnail

### DIFF
--- a/pxtlib/browserutils.ts
+++ b/pxtlib/browserutils.ts
@@ -470,15 +470,19 @@ namespace pxt.BrowserUtils {
     }
 
     export function scaleImageData(img: ImageData, scale: number): ImageData {
-        const cvs = document.createElement("canvas");
-        cvs.width = img.width * scale;
-        cvs.height = img.height * scale;
-        const ctx = cvs.getContext("2d")
+        const inputCanvas = document.createElement("canvas");
+        const outputCanvas = document.createElement("canvas");
+        inputCanvas.width = img.width;
+        inputCanvas.height = img.height;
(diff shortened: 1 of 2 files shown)


pxtlib/browserutils.ts
```diff
@@ -470,15 +470,19 @@ namespace pxt.BrowserUtils {
     }
 
     export function scaleImageData(img: ImageData, scale: number): ImageData {
-        const cvs = document.createElement("canvas");
-        cvs.width = img.width * scale;
-        cvs.height = img.height * scale;
-        const ctx = cvs.getContext("2d")
+        const inputCanvas = document.createElement("canvas");
+        const outputCanvas = document.createElement("canvas");
+        inputCanvas.width = img.width;
```
thsparks, line 475: Do the input canvas dimensions need to be multiplied by scale in the same way the output canvas dimensions are?

thsparks (approved):
LGTM. Had one question but may be a non-issue. Not sure.

pxtlib/browserutils.ts
kimprice: No, they don't. See the same process below in `imageDataToPNG`.

jwunderl (approved):

PR #9543, opened 2023-06-07
Error with 'groups'-array

OffByOne37:
The old documentation states, that the order of the 'groups'-function is determing in which order the groups will apeaer in the toolbox-flyout. But this is not true, it depends on which function, with which group, is defined first in the namespace. This can easily be seen in the playground (https://makecode.com/playground#basic-groups). According to the old documentation the 'Fruits'-group should be the first one in the toolbox-flyout, but this is wrong.


kimprice:
@OffByOne37 Thanks for pointing out this error. The problem with the sample in the playground is that the syntax is off. Try it with `//% groups='["Fruits", "Veggies"]'` and see that the blocks do follow the group order.

The syntax is also wrong in the groups example in defining-blocks.md. Additionally, the 'other' group should not be plural when listed in the groups definition. For example, this is a valid group definition placing 'other' at the top of the list: 
`//% groups='["other", "Servo", "i2c"]'`

PR #9545, opened 2023-06-07
Prevent Sim from Shrinking on Short Screens 

thsparks:
Fixes https://github.com/microsoft/pxt-microbit/issues/5214

On short screens, the sim was shrinking and looked very small. In non-tutorial mode, this doesn't happen. Instead, it just scrolls. By removing `padding-bottom: unset !important;`, we get the same behavior in tutorials.

Also did a little cleanup to reduce redundancy.

Try it here: https://makecode.microbit.org/app/1bd5e26f5481b85e4000d211b5641976886e78d5-51943d1273 (recommend using morse-chat to test, since that tutorial has audio blocks)

### DIFF
--- a/theme/tutorial-sidebar.less
+++ b/theme/tutorial-sidebar.less
@@ -452,6 +452,12 @@
             .simtoolbar, #miniSimOverlay { display: block; }
             .play-button { display: none !important; }
             .expand-button, .fullscreen-button, .tutorial.mute-button { display: block !important; }
+
+            display: flex;
+            align-items: flex-end;
+            aside.simtoolbar {
+                float: unset;
+            }
         }
 
         .tutorial-controls { display: none; }
@@ -470,14 +476,6 @@
 
         aside.simtoolbar {
             z-index: 5;
-        }
-
-        .simPanel {
-            display: flex;
-            align-items: flex-end;
-            aside.simtoolbar {
-                float: unset;
-            }
         }
     }
     .simulator-container:not(.hidden) {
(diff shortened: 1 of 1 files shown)


theme/tutorial-sidebar.less
```diff
@@ -1030,7 +1028,6 @@
                     min-height: @tutorialSimulatorMinHeight;
                 }
                 .simframe {
-                    padding-bottom: unset !important;
```
jwunderl, line 1033: Can you double check arcade looks fine without this? we might need to have this in the non `#root.tutorialSimSidebar` case (I guess in it's own `#root.tabTutorial:not(.fullscreensim):not(.tutorialSimSideBar) .simulator-container:not(.hidden) .simframe {`)

jwunderl (approved):

kimprice (approved):
Looks good! +1 to Joey's comment about checking Arcade.

theme/tutorial-sidebar.less
thsparks: I took a look and didn't notice any unexpected behavior. Seems like, on short screens, it's doing the same thing the live site is doing...

Target here if you want to double check: https://arcade.makecode.com/app/a5958ad6297dd96d3b68623bb7e827895f34b074-52a96d1738

(I did notice a different bug, but I think that's related to the resizable instructions change, not this one. Filed that here: https://github.com/microsoft/pxt-arcade/issues/5929)

PR #9546, opened 2023-06-08
Fix settings in iframe editor embed

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/5209

Here's a forked build from the one in that issue pointing at a build with this change: https://wyttlr.csb.app/ (there will be an overlay button in the bottom right to see the code of the sandbox)
microbit build: https://makecode.microbit.org/app/4e857c4bdf435b028015828dbd8d4af76e3d2c48-72860080ee
arcade build: https://arcade.makecode.com/app/15ee7b4f6abf540a00d30fa275c70925a5bdceeb-35abd09842--skillmap

I believe the change in `pxtlib/auth.ts` should technically be enough to cover this regression (+ https://github.com/microsoft/pxt/commit/a673adfed979c0356fcd270d331e0c4ed54e2012 to make it so language / high contrast persist through refreshes), but I ran into some other things while trying to figure out why it wasn't working so poked around a bit (e.g. we have some logic to cut auth from embed scenarios, but it was running after some initialization code that was leaving around some half initialized login state).

### DIFF
--- a/pxtlib/auth.ts
+++ b/pxtlib/auth.ts
@@ -77,7 +77,10 @@ namespace pxt.auth {
     };
 
     let _client: AuthClient;
-    export function client(): AuthClient { return _client; }
+    export function client(): AuthClient {
+        if (authDisabled) return undefined;
+        return _client;
+    }
 
     const PREFERENCES_DEBOUNCE_MS = 1 * 1000;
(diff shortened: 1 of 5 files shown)


eanders-ms (approved):
LGTM

PR #9551, opened 2023-06-13
Profile pages: added option for smaller dialog

srietkerk:
For micro:bit that doesn't have any badges, or any other targets that might not have something to fill the profile page space, we don't want the profile page to be the full screen. With these changes, a target can have a smaller profile page. 

image

EDIT: Here's an upload target: https://makecode.microbit.org/app/29277b08cefbec40fdc2f89cd0afd2933226581d-321f1764ea

Part of the solution for: https://github.com/microsoft/pxt-microbit/issues/5084

### DIFF
--- a/react-common/components/profile/Profile.tsx
+++ b/react-common/components/profile/Profile.tsx
@@ -22,6 +22,7 @@ export const Profile = (props: ProfileProps) => {
     const userBadges = user?.preferences?.badges || { badges: [] };
     const showBadges = pxt.appTarget?.cloud?.showBadges || false;
     const profileSmall = pxt.appTarget.appTheme?.condenseProfile;
+    const profileIcon = pxt.appTarget.appTheme?.cloudProfileIcon;
 
     const onBadgeClick = (badge: pxt.auth.Badge) => {
         showModalAsync({
@@ -47,9 +48,21 @@ export const Profile = (props: ProfileProps) => {
         />}
 
         {profileSmall &&
-            <div>
-               Now that you're logged in, your projects will be automatically saved to the cloud so you can
-                access them from any device! Learn more at Cloud Sync or Identity or Privacy
+            <div className="profile-info-container">
+                <p className="profile-info">
+                    {lf("Now that you're logged in, your projects will be automatically saved to the cloud so you can access them from any device! ")}
+                    {lf("Learn more at ")}
+                    <a href="https://arcade.makecode.com/identity/cloud-sync" target="_blank" rel="noopener noreferrer" tabIndex={0}>{lf("Cloud Sync ")}</a>
+                    {lf("or ")}
+                    <a href="https://makecode.com/privacy-faq" target="_blank" rel="noopener noreferrer" tabIndex={0}>{lf("Privacy ")}</a>
+                    {lf("or ")}
+                    <a href="https://arcade.makecode.com/identity/sign-in" target="_blank" rel="noopener noreferrer" tabIndex={0}>{lf("Identity.")}</a>
+                </p>
+                {profileIcon && <img
+                    className="ui image centered medium"
+                    src={profileIcon}
+                    alt={lf("Image of microbit microcontroller surrounded by clouds")}
+                />}
             </div>
         }
     </div>
--- a/react-common/components/profile/Profile.tsx
+++ b/react-common/components/profile/Profile.tsx
@@ -50,18 +50,12 @@ export const Profile = (props: ProfileProps) => {
         {profileSmall &&
             <div className="profile-info-container">
                 <p className="profile-info">
-                    {lf("Now that you're logged in, your projects will be automatically saved to the cloud so you can access them from any device! ")}
-                    {lf("Learn more at ")}
-                    <a href="https://arcade.makecode.com/identity/cloud-sync" target="_blank" rel="noopener noreferrer" tabIndex={0}>{lf("Cloud Sync ")}</a>
-                    {lf("or ")}
-                    <a href="https://makecode.com/privacy-faq" target="_blank" rel="noopener noreferrer" tabIndex={0}>{lf("Privacy ")}</a>
-                    {lf("or ")}
-                    <a href="https://arcade.makecode.com/identity/sign-in" target="_blank" rel="noopener noreferrer" tabIndex={0}>{lf("Identity.")}</a>
+                    {lf("Now that you're logged in, your projects will be automatically saved to the cloud so you can access them from any device!")}
                 </p>
                 {profileIcon && <img
                     className="ui image centered medium"
                     src={profileIcon}
-                    alt={lf("Image of microbit microcontroller surrounded by clouds")}
+                    alt=""
                 />}
             </div>
         }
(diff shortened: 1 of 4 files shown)


srietkerk:
Here's an upload target: https://makecode.microbit.org/app/29277b08cefbec40fdc2f89cd0afd2933226581d-321f1764ea

react-common/components/profile/Profile.tsx
```diff
@@ -44,5 +46,24 @@ export const Profile = (props: ProfileProps) => {
...
+                    <a href="https://arcade.makecode.com/identity/cloud-sync" target="_blank" rel="noopener noreferrer" tabIndex={0}>{lf("Cloud Sync ")}</a>
+                    {lf("or ")}
+                    <a href="https://makecode.com/privacy-faq" target="_blank" rel="noopener noreferrer" tabIndex={0}>{lf("Privacy ")}</a>
+                    {lf("or ")}
+                    <a href="https://arcade.makecode.com/identity/sign-in" target="_blank" rel="noopener noreferrer" tabIndex={0}>{lf("Identity.")}</a>
+                </p>
+                {profileIcon && <img
+                    className="ui image centered medium"
+                    src={profileIcon}
+                    alt={lf("Image of microbit microcontroller surrounded by clouds")}
```
jwunderl, line 64: Alt text would have to be more generic / not micro:bit specific, but this feels like a scenario where the image is purely decorative & as such alt="" would be preferred (e.g. https://html.spec.whatwg.org/multipage/images.html#a-purely-decorative-image-that-doesn't-add-any-information)

react-common/components/profile/Profile.tsx
```diff
@@ -44,5 +46,24 @@ export const Profile = (props: ProfileProps) => {
...
             onBadgeClick={onBadgeClick}
         />}
+
+        {profileSmall &&
+            <div className="profile-info-container">
+                <p className="profile-info">
+                    {lf("Now that you're logged in, your projects will be automatically saved to the cloud so you can access them from any device! ")}
+                    {lf("Learn more at ")}
+                    <a href="https://arcade.makecode.com/identity/cloud-sync" target="_blank" rel="noopener noreferrer" tabIndex={0}>{lf("Cloud Sync ")}</a>
+                    {lf("or ")}
```
jwunderl, line 56: we shouldn't be pointing at target specific docs pages; either need to go under makecode.com/docs or be under common-docs with a relative link here so they get mirrored in each target.

react-common/components/profile/Profile.tsx
```diff
@@ -44,5 +46,24 @@ export const Profile = (props: ProfileProps) => {
...
+                    <a href="https://makecode.com/privacy-faq" target="_blank" rel="noopener noreferrer" tabIndex={0}>{lf("Privacy ")}</a>
```
jwunderl, line 57: Privacy link is at the bottom of every page already, I don't see a reason to put it here as well?

react-common/components/profile/Profile.tsx
```diff
@@ -44,5 +46,24 @@ export const Profile = (props: ProfileProps) => {
             userState={userBadges}
             onBadgeClick={onBadgeClick}
         />}
+
+        {profileSmall &&
+            <div className="profile-info-container">
+                <p className="profile-info">
+                    {lf("Now that you're logged in, your projects will be automatically saved to the cloud so you can access them from any device! ")}
+                    {lf("Learn more at ")}
+                    <a href="https://arcade.makecode.com/identity/cloud-sync" target="_blank" rel="noopener noreferrer" tabIndex={0}>{lf("Cloud Sync ")}</a>
```
abchatra, line 55: We shouldn't hardcode target in the PXT.

react-common/components/profile/Profile.tsx
```diff
@@ -44,5 +46,24 @@ export const Profile = (props: ProfileProps) => {
             userState={userBadges}
             onBadgeClick={onBadgeClick}
         />}
+
+        {profileSmall &&
+            <div className="profile-info-container">
+                <p className="profile-info">
+                    {lf("Now that you're logged in, your projects will be automatically saved to the cloud so you can access them from any device! ")}
+                    {lf("Learn more at ")}
```
jwunderl, line 54: We can't concat strings like this to form a sentence, it becomes untranslatable due to languages having different sentence structures. there is a helper here https://github.com/microsoft/pxt/blob/glueInDepsForHex/react-common/components/util.tsx#L15 that can be used like this https://github.com/microsoft/pxt/blob/glueInDepsForHex/skillmap/src/components/AppModal.tsx#L581

react-common/components/profile/Profile.tsx
srietkerk: As in this link shouldn't be the full arcade.makecode.com?

abchatra: Yes, there is a way to get the full URL.

abchatra: @jwunderl ?

react-common/components/profile/Profile.tsx
```diff
@@ -44,5 +46,24 @@ export const Profile = (props: ProfileProps) => {
...
+                    <a href="https://arcade.makecode.com/identity/cloud-sync" target="_blank" rel="noopener noreferrer" tabIndex={0}>{lf("Cloud Sync ")}</a>
+                    {lf("or ")}
+                    <a href="https://makecode.com/privacy-faq" target="_blank" rel="noopener noreferrer" tabIndex={0}>{lf("Privacy ")}</a>
+                    {lf("or ")}
+                    <a href="https://arcade.makecode.com/identity/sign-in" target="_blank" rel="noopener noreferrer" tabIndex={0}>{lf("Identity.")}</a>
```
jwunderl, line 59: Can we just make a single docs page at `/identity` that links out to whatever else we want here? having a bunch of links under an or feels a little messy to me. Probably should go in common-docs/ so it's available per target

jwunderl (changes requested):

react-common/components/profile/Profile.tsx
jwunderl: Ah I was writing out a not for it down below, we ought to add the ref pages to /common-docs so they're available per target (e.g. see common-docs/faq and how it is available or overridden in any target

react-common/components/profile/Profile.tsx
srietkerk: Is this to avoid CORS issues?

jwunderl: It's not a cors issue, it's that targets should be relatively independent and should not point at each other without explicitly meaning to do so. few reasons;

* docs have common macro replacements that e.g. insert the target into the documentation which could be confusing
* school districts might choose to only allow a single target (maybe they really don't like the idea of multiplayer or making games, but are fine with programming micro:bits) so this page wouldn't be available
* if they get to the search page and poke around by searching or using the side menu to find more info they'll be on a page that doesn't relate to what they're learning (e.g. incorrect api references, different simulator, screenshots of arcade features that could be in different places not implemented in micro:bit, etc)

react-common/components/profile/Profile.tsx
srietkerk: Oh, I actually just realized that when you sign-in, you can hit the `identity/sign-in` page with the learn more link in the lower right hand corner of the modal. And with the privacy link everywhere, I don't think that's necessary in this pop-up either. If we were going to have any link here, I would consider https://arcade.makecode.com/identity/cloud-sync, but I would make an argument against the other two.

srietkerk:
This is the modal without the links on the info side 
image

jwunderl (approved):
🦖

PR #9565, opened 2023-06-19
Fix Step Counter on Tutorial Window Resize

thsparks:
Fixes https://github.com/microsoft/pxt-microbit/issues/5250

We were persisting horizontal/vertical layout in state, but it was getting out of sync with the actual screen size and there's not really a need for it to be persisted. We can just recompute isHorizontal every time, and it works as intended.

Upload target: https://makecode.microbit.org/app/fd87f2fd28a4aeb6f3efd140372ea3dbcbeab38c-0c65b5e47b


benvillalobos:
Tested out the upload target, it no longer repros 👍

benvillalobos (approved):

PR #9569, opened 2023-06-21
Support for silent SSO

eanders-ms:
Right now MakeCode always prompts the user to select an account when signing in. This is inconvenient in schools, where kids only have one account and there is never a need to select an account. This PR brings in support for silent single-sign-on (SSO), enabling zero-prompt authentication if it is available. SSO can fail if the provider requires user interaction (like getting consent). When this happens we'll fallback to an interactive login.

Additionally, if the user explicitly signs out, the next sign in will be interactive. This is to handle the case where SSO signed in with the wrong account and the user wants to log in with a different account.


jwunderl (approved):

eanders-ms:
Will hold off merging this until the requisite backend updates are deployed.

PR #9570, opened 2023-06-21
Improve Share Page Sandbox Display on Small Screens

thsparks:
This partially addresses the issues seen in https://github.com/microsoft/pxt-microbit/issues/4828. While it's not a perfect fix, it's a step in the right direction, and it's at least usable now. I wanted to get this out in time for the hotfix, since I suspect some of the remaining improvements will take me longer to implement.

This change fixes the sim sizing and stops it from disappearing on smaller screens, and it simply removes the sim toolbar for now (which is currently offscreen anyway).

SharePageResizeSmaller

Future improvements that would be worth doing, perhaps in a separate change, would be:
1. Fix weird gap on the right-hand side where background doesn't fully extend (looking into this now, will add to PR if I figure it out but didn't want to block on it).
2. Bring back the sim toolbar, but probably remove "debug" from it.
3. Make the download bar blend a bit more seamlessly into the page.

### DIFF
--- a/theme/common.less
+++ b/theme/common.less
@@ -1471,6 +1471,21 @@ p.ui.font.small {
     }
 }
 
+.sandbox {
+    .simPanel .simtoolbar {
+        // position: fixed;
+        // bottom: 4rem;
+        // left: auto;
+        // right: auto;
+        // display: block !important;
+        display: none;
+    }
+
+    // .simtoolbar .ui.button {
+    //     font-size: 1.7rem;
+    // }
+}
+
 /***********************************
    Headless Simtoolbar (Minecraft)
 ************************************/
@@ -1693,7 +1708,7 @@ p.ui.font.small {
            Mini sim view
 *******************************/
 
-#root.miniSim:not(.fullscreensim) {
+#root.miniSim:not(.fullscreensim, .sandbox) { // todo thsparks : check for other places fullscreensim is excluded and also exclude sandbox?
     /* Layout */
     .simPanel.ui.items {
         position: fixed;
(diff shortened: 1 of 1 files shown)


theme/common.less
```diff
@@ -2393,12 +2398,12 @@ button.ui.button.hostmultiplayergame-button {
...
 }
 
-#root:not(.fullscreensim) .multiplayer-presence {
+#root:not(.fullscreensim, .sandbox) .multiplayer-presence {
     margin: 0;
     margin-bottom: -.5rem;
 }
 
-.miniSim:not(.fullscreensim):not(.tabTutorial) .multiplayer-presence {
+.miniSim:not(.fullscreensim, .sandbox):not(.tabTutorial) .multiplayer-presence {
```
jwunderl, line 2406: oh, you can pass multiple values to not? neat, probably should switch over the `:not(.tabTutorial)`s to that then, but doesn't have to happen now

jwunderl (approved):
Huh, on the live site looks like it's there in some breakpoints just positioned above the sim / underneath the toggle 
image

I'd personally be inclined to cut the download button from this view as there's the other download button that's always there, but seems like a step in the right direction.

thsparks:
> I'd personally be inclined to cut the download button from this view as there's the other download button that's always there, but seems like a step in the right direction.

Yeah, I considered this too, but the bar isn't there in older versions so I was assuming it was added as an intentional design choice...Then again, I suppose it could just have easily been an accident. I think it may be worth having a quick discussion with the team around what we actually want this page to look like - but for now, I just want to get it functional :)

PR #9572, opened 2023-06-22
Fix signin button font on linux

eanders-ms:
Removal of font-family attribute fixes the issue on Linux. Still looks fine on Windows (tested Arcade and microbit). It seems this value was unnecessary. Unsure why Chrome on Linux would resolve the css differently than on other platforms.

Fixes https://github.com/microsoft/pxt-microbit/issues/5272

### DIFF
--- a/theme/common.less
+++ b/theme/common.less
@@ -2356,7 +2356,6 @@ div.simframe.ui.embed {
     border-radius: 6px;
     height: 3rem;
     padding: 0 .6rem;
-    font-family: @segoeUIFont;
     font-weight: 500;
 
     .icon {


jwunderl (approved):
That's real weird, we use this font in several different places so might be worth a check here and adding in whatever font we expect to resolve to on linux to the fallback list? https://github.com/microsoft/pxt/blob/master/theme/themes/pxt/globals/site.variables#L42

I believe we use it for debugger, share, and tutorial text by default so maybe worth a quick look, I don't have a linux based setup handy though :(

jwunderl (approved):

PR #9579, opened 2023-06-27
Fix audio recording offline compilation

jwunderl:
This pr https://github.com/microsoft/pxt/pull/7320 is some context you'll probably need to understand why this actually matters; when compiling we alphabetically sort to generate a stable order so the produced sha is consistent even when packages are imported in different orders (e.g. if radio happens to be before or after audio-recording in the list). Probably worth revisiting that logic and sorting in a way that takes into account the source better, but since these are all builtin extensions the behavior I used in the cli should be consistent with webapp (e.g. for builtin packages can use id since it's "*" at runtime / behaves like this change, but make sure github packages use slug / version / etc). Several of the other changes were just minor style nits or fixing logging so I'll point at the main fix

I tested this by packaging and running electron app locally and the package still working with internet turned off after build time, can't produce a signed app to fully test but I think it should be good.

Added in a side fix that I noticed while testing, was getting some null pointer exceptions when you try and search while offline that prevented builtin packages from showing up (was trying to test bluetooth)

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -2390,7 +2390,7 @@ async function buildTargetCoreAsync(options: BuildTargetOptions = {}) {
         }
 
         // For the projects, we need to save the base HEX file to the offline HEX cache
-        if (isPrj && pxt.appTarget.compile && pxt.appTarget.compile.hasHex) {
+        if (isPrj && pxt.appTarget.compile?.hasHex) {
             if (!pkgOptions) {
                 pxt.debug(`Failed to extract native image for project ${dirname}`);
                 return;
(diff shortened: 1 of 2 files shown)


cli/cli.ts
```diff
@@ -2417,26 +2417,26 @@ async function buildTargetCoreAsync(options: BuildTargetOptions = {}) {
...
-            pxt.log(`Dependencies of pkg: ${allDeps}`);
-            pxt.log(`Attemping to bundle necessary hexfiles to compile with: ${pkgsToBuildWith}`);
+            pxt.log(`Dependencies of base pkgs: ${allDeps.map(el => path.basename(el)).join(", ")}`);
             for (const extraPackage of pkgsToBuildWith) {
+                const extraPathBaseName = path.basename(extraPackage);
                 process.chdir(path.join(rootDir, dirname));
                 const deps: pxt.Map<string> = {
                     ...config.dependencies,
-                    extra: "file:" + path.relative(path.resolve("."), extraPackage)
+                    [extraPathBaseName]: "file:" + path.relative(path.resolve("."), extraPackage)
```
jwunderl, line 2433: this line is the fix (well, I guess this+the const just above)

jwunderl:
Merging so I can port to stable branch / build for testing tomorrow am but please still leave comments / change requests if you have any~

PR #9580, opened 2023-06-27
Hide SimToolbar in Arcade's Sandbox Mode

thsparks:
In microbit, this worked without !important, but in arcade it gets overridden, so the !important is necessary.


jwunderl (approved):

thsparks:
For posterity: this was a follow-up on https://github.com/microsoft/pxt/pull/9571

PR #9590, opened 2023-06-30
[stable8.5] only mutate kinds of the same type on delete / rename

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/5965; we may want to think on prohibiting names from duplicating across kinds, but that seemed a little unnecessary as it wasn't a regression / arguable in the first place.

This one's targeting the release branch since I was already on that accidentally (which I suppose is a good thing), will reverse port to master.

build (off stable1.12 in arcade, stable8.5 here): https://arcade.makecode.com/app/bf6c7d16c202bec13b3afdab05cdd54b7d5a5093-6edb864625

with the sample from that issue: https://arcade.makecode.com/app/bf6c7d16c202bec13b3afdab05cdd54b7d5a5093-6edb864625#pub:S63783-19068-25507-81035

notouchyspritekindfood

### DIFF
--- a/pxtblocks/fields/field_kind.ts
+++ b/pxtblocks/fields/field_kind.ts
@@ -30,11 +30,11 @@ namespace pxtblockly {
                     lf("Rename '{0}':", oldName),
                     newName => {
                         // Update the values of all existing field instances
-                        const allFields = getAllFields(ws, field => field instanceof FieldKind);
+                        const allFields = getAllFields(ws, field => field instanceof FieldKind
+                            && field.getValue() === oldName
+                            && field.opts.blockId === this.opts.blockId);
                         for (const field of allFields) {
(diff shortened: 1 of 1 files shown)


kimprice (approved):

jwunderl:
Made a minor change, the rest of the file keys off the name field (which comes from kindNamespace e.g. here https://github.com/jwunderl/pxt-status-bar/blob/master/status-bar.ts#L45), so keying off that instead of blockId (in case for some reason we want to make another block with the same namespace but different implementation... that'd be a bit weird but easier to be consistent.)

updated build: https://arcade.makecode.com/app/1e0d6138bcbf3f9798bfcea181faa2f89e7d0674-c0ba07b08b#pub:S63783-19068-25507-81035

PR #9603, opened 2023-07-12
do not use comma sepd pseudo-selectors

jwunderl:
Should fix https://github.com/microsoft/pxt-microbit/issues/5305. The pxt diff between v6.0.5 and v6.0.15 is https://github.com/microsoft/pxt/compare/v9.0.5...v9.0.11, I just had a gut feeling this might be related as I remembered being pleasantly surprised that it worked in the first place :)

@vsbogd if you have time could you check if this link works on your device: https://makecode.microbit.org/app/be0eecc7a88c77bfe1125aeeb3fc7c6d4286f4bd-a7b9522d32 (just in browser should be fine, as they share a webview) that build is just changes that will be released to live site soon + this

### DIFF
--- a/theme/common.less
+++ b/theme/common.less
@@ -1697,7 +1697,7 @@ p.ui.font.small {
            Mini sim view
 *******************************/
 
-#root.miniSim:not(.fullscreensim, .sandbox) {
+#root.miniSim:not(.fullscreensim):not(.sandbox) {
     /* Layout */
     .simPanel.ui.items {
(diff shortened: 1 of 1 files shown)


vsbogd:
I am sorry fo a late reply, but I can confirm, the link above works for me.
Thanks!

jwunderl:
@vsbogd thank you for confirming! I will be pushing out a release with this fix and a few more in the next few days

PR #9605, opened 2023-07-12
Branding: Add Microsoft logo to first loading screen

kimprice:
The following loading screen is displayed on first load for all targets instead of a target-specific icon.
microsoftLoad

### DIFF
--- a/theme/themes/pxt/elements/loader.overrides
+++ b/theme/themes/pxt/elements/loader.overrides
@@ -21,7 +21,7 @@
 }
 
 .ui.loader.main.msft:after {
-  background: transparent @loaderMsftImage no-repeat center center;
+  background-image: @loaderMsftImage;
   background-size: 99%;
 }
 
--- a/theme/themes/pxt/elements/loader.overrides
+++ b/theme/themes/pxt/elements/loader.overrides
@@ -22,7 +22,7 @@
 
 .ui.loader.main.msft:after {
   background-image: @loaderMsftImage;
-  background-size: 99%;
+  background-size: 99%; // Must be less than 100% to avoid being cut off
 }
 
 .ui.loader.avatar:after {
--- a/theme/themes/pxt/elements/loader.overrides
+++ b/theme/themes/pxt/elements/loader.overrides
@@ -3,7 +3,7 @@
 *******************************/
 
 
-.ui.loader.main:before, 
+.ui.loader.main:before,
 .ui.loader.avatar:before {
    border: none;
    border-radius: 0px;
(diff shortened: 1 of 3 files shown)


Jaqster:
Will we still keep the theme colors?  i.e. blue for micro:bit and dark grey for minecraft? 
image
image

kimprice:
@Jaqster great question
Yes, it will keep the theme color. I just checked micro:bit:
image

theme/themes/pxt/elements/loader.overrides
```diff
@@ -15,11 +15,16 @@
...
   background: transparent @loaderImage no-repeat center center;
-  background-size: 100%;
+  background-size: @loaderImageSize;
  -webkit-animation: @loaderAnimation @loaderSpeed infinite linear;
  animation: @loaderAnimation @loaderSpeed infinite linear;
 }
 
+.ui.loader.main.msft:after {
+  background: transparent @loaderMsftImage no-repeat center center;
+  background-size: 99%;
```
jwunderl, line 25: does the 1% here make a difference / what's the change here handling?

theme/themes/pxt/elements/loader.overrides
```diff
@@ -15,11 +15,16 @@
...
   border-radius: 0px;
   background: transparent @loaderImage no-repeat center center;
-  background-size: 100%;
+  background-size: @loaderImageSize;
  -webkit-animation: @loaderAnimation @loaderSpeed infinite linear;
  animation: @loaderAnimation @loaderSpeed infinite linear;
 }
 
+.ui.loader.main.msft:after {
+  background: transparent @loaderMsftImage no-repeat center center;
```
jwunderl, line 24:
```suggestion
  background-image: @loaderMsftImage;
```

should work fine with a little less duped? doesn't really matter though, either way works

jwunderl (approved):
1 question / one possible suggestion but otherwise lgtm

theme/themes/pxt/elements/loader.overrides
jwunderl: Ah okay makes sense, might be worth a quick comment above it so if we run into it again we know why it's special cased a bit~

jwunderl (approved):

THEb0nny:
@kimprice you need to make @loaderAnimation separate, as I understand it, if you override @loaderAnimation, then the animation for the Microsoft logo will be the same as for the project logo.

kimprice:
> @kimprice you need to make @loaderAnimation separate, as I understand it, if you override @loaderAnimation, then the animation for the Microsoft logo will be the same as for the project logo.

Yes, this change is just to change the initial loading screen. All other loading screens will be target-specific. I override the `@loaderImageUrl` in Arcade to change the target-specific loading screen (https://github.com/microsoft/pxt-arcade/pull/5987/files#diff-a71b80d431a659366ab01f6d02dc9b1fc7a5f3a6cfe4cccf65df4450839aafc8).

THEb0nny:
@kimprice https://github.com/microsoft/pxt-ev3/pull/1030/commits/bf193b4e1ee319ebbd15869c668071c083c58988

PR #9602, opened 2023-07-12
clean up PageviewPerformanceData tick a bit

jwunderl:
Here's an example of what the tick looks like when it goes through this fn:

into ->

```json
{
    "name": "Microsoft.ApplicationInsights.{0}.PageviewPerformance",
    "time": "2023-07-12T16:44:43.715Z",
    "iKey": "9801ed01-c40f-46ec-aa40-2a1742a9e71c",
    "ext": {
        "app": {
            "sesId": "{b64ish}"
        },
        "device": {
            "localId": "browser",
            "deviceClass": "Browser"
        },
        "trace": {
            "traceID": "{b16ish}",
            "name": "/75618-11489-17671-85517"
        },
        "user": {
            "id": "{b64ish}"
        }
    },
    "tags": [],
    "data": {
        "duration": 3642.60000000149
    },
    "baseType": "PageviewPerformanceData",
    "baseData": {
        "name": "world_race",
        "uri": "http://localhost:3232/75618-11489-17671-85517",
        "isValid": true,
        "durationMs": 3642.60000000149,
        "duration": "00:00:03.643",
        "perfTotal": "00:00:03.643",
        "networkConnect": "00:00:00.319",
        "sentRequest": "00:00:00.876",
        "receivedResponse": "00:00:00.002",
        "domProcessing": "00:00:02.446"
    },
    "ver": "4.0"
}
```

end of fn

```json
{
    "name": "Microsoft.ApplicationInsights.{0}.PageviewPerformance",
    "time": "2023-07-12T16:44:43.715Z",
    "iKey": "9801ed01-c40f-46ec-aa40-2a1742a9e71c",
    "ext": {
        "app": {
            "sesId": "{b64ish}"
        },
        "device": {
            "localId": "browser",
            "deviceClass": "Browser"
        },
        "trace": {
            "traceID": "{b16ish}",
            "name": "/xxxxx-xxxxx-xxxxx-xxxxx"
        },
        "user": {
            "id": "{b64ish}"
        }
    },
    "tags": [],
    "data": {
        "duration": 3642.60000000149
    },
    "baseType": "PageviewPerformanceData",
    "baseData": {
        "name": "http://localhost:3232",
        "uri": "http://localhost:3232/xxxxx-xxxxx-xxxxx-xxxxx",
        "isValid": true,
        "durationMs": 3642.60000000149,
        "duration": "00:00:03.643",
        "perfTotal": "00:00:03.643",
        "networkConnect": "00:00:00.319",
[...]

PR #9609, opened 2023-07-18
one more fix for pageviews; include pathname

jwunderl:
quick follow up on https://github.com/microsoft/pxt/pull/9602 before I port it to other editors, include pathname so we can compare those as well (vs. just page titles); share pages come in as xxxxx-xxxxx-xxxxx-xxxxx e.g. 
image


srietkerk (approved):
LGTM

PR #9611, opened 2023-07-18
[stable8.5] Check definition name before accepting translation (#9502)

jwunderl:
cherry pick https://github.com/microsoft/pxt/pull/9502 for arcade. Spanish translations seem to have about half a dozen of these issues, which is causing issues for some teachers; gonna try to push this out in next hotfix with a few upgrade rules to clean it up, will try those now~.

For ref:  https://forum.makecode.com/t/error-when-using-makecode-arcade-in-google-chrome/21210/4


kimprice (approved):

srietkerk (approved):
LGTM

PR #9617, opened 2023-07-25
tooltip hides on very small screens so it doesn't get cut off

emoltz:
https://github.com/microsoft/pxt-microbit/issues/3130

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,4 @@ tests/ui-test/chromedriver.exe
 tests/ui-test/screenshot/**
 report.*.json
 *.tsbuildinfo
+.vscode/settings.json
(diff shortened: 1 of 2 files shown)


emoltz:
@microsoft-github-policy-service agree [company="{ethan shafran moltz}"]

emoltz:

jwunderl:
Hey, thanks for the pr!

I took a look at it and it doesn't appear to fully solve the issue :( the download toolbar is currently switching to the mini-icon only mode for mobile and tablet dimensions (basically, any screen <= 991px), so the issue remains for those (e.g. this is at 990px width): 

image

and it looks like the dimensions that this would affect (< 500px) already have the pop up hidden / not show up, so this wouldn't change that behavior unfortunately. I think ideally the fix for this would either cap the width of the tooltip so it stays fully in screen, or shift it over so it's `left` / `right` are always at least 0 (while still keeping the arrow in the correct position).

emoltz:
So sorry, meant to just draft a PR. I will complete this and get back to you!

PR #9621, opened 2023-07-29
Re: Sim missing at exactly 991px width from MicroBit repo

emoltz:
This is concerning issue #5328 in the Microbit repo (https://github.com/microsoft/pxt-microbit/issues/5328). The issue was that the simulator was not showing up when the window width was exactly 991px wide. After a lot of digging I was able to pinpoint that the issue was stemming from the `isTabletSize()` function within the `browserutils.ts` file. 

```
return window?.innerWidth = pxt.BREAKPOINT_TABLET;
```

to

```
return window?.innerWidth <= pxt.BREAKPOINT_TABLET;
```

### DIFF
--- a/pxtlib/browserutils.ts
+++ b/pxtlib/browserutils.ts
@@ -166,7 +166,7 @@ namespace pxt.BrowserUtils {
     }
 
     export function isTabletSize(): boolean {
-        return window?.innerWidth < pxt.BREAKPOINT_TABLET;
+        return window?.innerWidth <= pxt.BREAKPOINT_TABLET;
     }
 
     export function isComputerSize(): boolean {
--- a/pxtlib/browserutils.ts
+++ b/pxtlib/browserutils.ts
@@ -170,7 +170,7 @@ namespace pxt.BrowserUtils {
     }
 
     export function isComputerSize(): boolean {
-        return window?.innerWidth >= pxt.BREAKPOINT_TABLET;
+        return window?.innerWidth > pxt.BREAKPOINT_TABLET;
     }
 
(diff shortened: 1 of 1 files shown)


pxtlib/browserutils.ts
```diff
@@ -166,7 +166,7 @@ namespace pxt.BrowserUtils {
     }
 
     export function isTabletSize(): boolean {
-        return window?.innerWidth < pxt.BREAKPOINT_TABLET;
+        return window?.innerWidth <= pxt.BREAKPOINT_TABLET;
     }
 
     export function isComputerSize(): boolean {
```
jwunderl, line 172: this function already has it `>=`, so we need to fix that so we never end up with both `isTabletSize() === true` & `isComputerSize() === true` at the same time

jwunderl (approved):
Thanks for the PR! Fix looks good to me, but could you also remove the `=` from `isComputerSize` just below so they don't both return true for `innerWidth == 991`?

emoltz:
Certainly!

jwunderl (approved):
Perfect, thank you so much!

emoltz:
Sounds great. I will look at other bugs to fix.

jwunderl:
(if you want a related one, I believe there is a minor spacing issue still with the download button having extra left margin when in a tutorial 991px width -- this fixed most of the issues related to tutorials at the breakpoint though! This one should be in the less though / probably a bit annoying to nail down)

emoltz:
I'll take a look. I'm in Colombia at the moment with limited internet but will fix once I am back in the states.

PR #9622, opened 2023-08-07
Add `--no-webapps` gulp option

eanders-ms:
Optional flag to skip building authcode, skillmap, multiplayer, and soon kiosk. Usage:
```bash
gulp --no-webapps
```

```bash
gulp watch --no-webapps
```

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -17,7 +17,7 @@ const isWin32 = os.platform() === "win32";
 
 const clean = () => rimraf("built").then(() => rimraf("temp"));
 const update = () => exec("git pull", true).then(() => exec("npm install", true))
-
+const noop = () => Promise.resolve();
 
 /** onlineline */
 const onlinelearning = () => {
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -653,6 +653,26 @@ const copyMultiplayerHtml = () => rimraf("webapp/public/multiplayer.html")
 
 const multiplayer = gulp.series(cleanMultiplayer, buildMultiplayer, gulp.series(copyMultiplayerCss, copyMultiplayerJs, copyMultiplayerHtml));
 
+/********************************************************
+                 Webapp build wrappers
+*********************************************************/
+
+const shouldBuildWebapps = () => (process.argv.indexOf("--no-webapps") === -1);
```
jwunderl, line 660: Nit; making this an async function would remove the need for an explicit returned noop / inverted logic, but fine either way

jwunderl (approved):
Nice

kimprice (approved):
This is awesome!

gulpfile.js
eanders-ms: Something we could improve later for sure. Gulp's expectations here are very specific and I wrestled with it more than I'd expected to.

thsparks (approved):

PR #9632, opened 2023-08-09
Fix Minecraft (Headless) Tutorial Layout

thsparks:
This is one part of the fix needed for Minecraft's tutorial layout after the PXT bump. The other part is in the pxt-minecraft repo: https://github.com/microsoft/pxt-minecraft/pull/2314

Tutorials with fixes applied can be seen here: https://minecraft.makecode.com/app/8ef76a97727b2cdaf01a9b01e0703b80e23721a8-8e733a6100

Note that this does **not** fix https://github.com/microsoft/pxt-minecraft/issues/2312, which I'll investigate and address in a separate change.

Before:
image

image

After:
image

image

### DIFF
--- a/theme/tutorial-sidebar.less
+++ b/theme/tutorial-sidebar.less
@@ -514,8 +514,6 @@
 
 #root.headless.tabTutorial {
     #simulator .editor-sidebar {
-        position: relative;
-        display: block !important;
         width: @simulatorWidth;
         height: calc(~'100%' - @mainMenuHeight);
         top: @mainMenuHeight;
@@ -658,6 +656,7 @@
         }
 
         #simulator .editor-sidebar {
+            display: block !important;
             height: @defaultTutorialHeight;
             width: 100%;
         }


srietkerk:
I found a bug that I'm not sure is super important to address and might be something we want to address later, but if you shrink the screen down so it's the horizontal instruction view and then enlarge the screen again, the tutorial instructions seem to be stuck in a smaller div and scrollable.
image

thsparks:
> I found a bug that I'm not sure is super important to address and might be something we want to address later, but if you shrink the screen down so it's the horizontal instruction view and then enlarge the screen again, the tutorial instructions seem to be stuck in a smaller div and scrollable. image

Yeah, I have that filed here: https://github.com/microsoft/pxt-minecraft/issues/2312

Will fix separately, as it's not new from these changes and exists in other targets too.

srietkerk:
Oh, crap, sorry. I saw that issue but just didn't watch the GIF all the way through. Sounds good. Sorry for wasting your time there.

thsparks:
> Oh, crap, sorry. I saw that issue but just didn't watch the GIF all the way through. Sounds good. Sorry for wasting your time there.

Hahah it's no problem :)

srietkerk (approved):
Looks good to me!

PR #9634, opened 2023-08-10
Null check instead of falsey check for storedState value

kimprice:
fixes https://github.com/microsoft/pxt-arcade/issues/6003

This fixes a bug that was happening when an arcade game was storing a setting as an empty array. When saving the settings on restart, it was getting serialized as an empty string, which resulted in the value not being set since it evaluates to falsey. Then later the game referenced the setting, causing a "dereferencing null/undefined value" error.

Now we check for null/undefined instead of falsey values when setting the storedState.

Big thanks to @jwunderl for walking me through this bug.


srietkerk (approved):
LGTM!

jwunderl (approved):

PR #9636, opened 2023-08-10
On/off colors added as block params for grids

srietkerk:
In an addition to https://github.com/microsoft/pxt/pull/9512, for the cybersecurity curriculum, they are trying to make rules that match a specific pattern of a grid (color and shape). However, we didn't have a way to allow for customizations for the cell colors. By default, the cell colors of a grid for "off" is black with an opacity and "on" is white. 

With these changes, if a user specifies on and off colors in comment attributes for block definitions, when clicking on a box in the grid, you can toggle between the two colors specified. 

image

image

Closes https://github.com/microsoft/pxt-minecraft/issues/2313

### DIFF
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -813,6 +813,8 @@ declare namespace ts.pxtc {
         iconURL?: string;
         imageLiteral?: number;
         gridLiteral?: number;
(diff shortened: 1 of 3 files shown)


jwunderl (approved):
Nice!

PR #9639, opened 2023-08-14
More Minecraft Tutorial Fixes

thsparks:
Fixes up a few additional tutorial layout bugs, including:

1. Fixes https://github.com/microsoft/pxt-minecraft/issues/2312 (need to ensure the vertical resize container is turned off and reset once the instructions are re-expanded)

2. Fix navigation buttons on mobile screen size (more headless-mode css adjustments)

Upload Target: https://minecraft.makecode.com/app/2d4dd01b4eb9f96d87fb427cf3ccf4d91e0af147-cb16e8b4bb

### DIFF
--- a/webapp/src/components/tutorial/TutorialContainer.tsx
+++ b/webapp/src/components/tutorial/TutorialContainer.tsx
@@ -85,7 +85,7 @@ export function TutorialContainer(props: TutorialContainerProps) {
         } else {
             setParentHeight();
         }
-    })
+    }, [props.hasBeenResized, document.body])
 
     React.useEffect(() => {
         setCurrentStep(props.currentStep);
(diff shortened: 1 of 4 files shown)


webapp/src/components/tutorial/TutorialContainer.tsx
```diff
@@ -85,7 +85,7 @@ export function TutorialContainer(props: TutorialContainerProps) {
         } else {
             setParentHeight();
         }
-    })
+    }, [props.hasBeenResized, document.body])
```
kimprice, line 88: Why do you need `props.hasBeenResized` here if you just return at the beginning of the `useEffect`?

thsparks: This is to make sure the function is run in the scenario where it changes from true to false.

kimprice: oh. That makes sense, thanks!

thsparks: (Admittedly, that's more of a theoretical scenario than anything else. It shouldn't really be changing in that direction...but I figured it didn't hurt to cover it, just in case something changes down the line)

kimprice (approved):

PR #9642, opened 2023-08-16
Remove old old code validation, move out some logic to pxtlib/pxtblocks

jwunderl:
Mostly removing old, unused version of code validation, and moving some of the code validation logic to pxtblocks / pxtlib to reduce business logic in webapp layer / make it so we can add tests & expose on editor interface / etc.

I played around with a few approaches of splitting up up so the validator portion lives in pxt / only rendering related bits lived in webapp, but I'm gonna put that off till we have a few more validators / expose it elsewhere; it felt messy to pull the parts out for now, will be more clear / just as easy to separate out when we have 2 or 3 working examples.


pxtblocks/code-validation/validateBlocksExist.ts
```diff
@@ -0,0 +1,40 @@
...
+
+        for (const [requiredBlockId, requiredCount] of Object.entries(requiredBlockCounts)) {
+            const countForBlock = userBlocksEnabledByType[requiredBlockId];
+            if (countForBlock === undefined) {
+                // user did not use a specific block
+                missingBlocks.push(requiredBlockId);
+            } else if (!countForBlock) {
+                // all instances of block are disabled
+                disabledBlocks.push(requiredBlockId);
+            } else if (countForBlock < requiredCount) {
```
jwunderl, line 27: Added this branch, though not yet used; to allow us to check if a block is used enough when the same block is used multiple times

webapp/src/components/tutorialValidators.tsx
```diff
@@ -1,19 +1,17 @@
...
 import { getBlocksEditor } from "../app";
-import TutorialOptions = pxt.tutorial.TutorialOptions;
-import TutorialStepInfo = pxt.tutorial.TutorialStepInfo;
 import CodeValidator = pxt.tutorial.CodeValidator;
 import CodeValidatorMetadata = pxt.tutorial.CodeValidatorMetadata;
 import CodeValidationResult = pxt.tutorial.CodeValidationResult;
 import CodeValidationExecuteOptions = pxt.tutorial.CodeValidationExecuteOptions;
 
-const defaultResult: CodeValidationResult = {
+const defaultResult: () => CodeValidationResult = () => ({
```
jwunderl, line 8: nit; avoiding returning the same instance, as that could be mutated accidentally / lead to issues. Also thought to just Object.freeze it but we don't use it much / that can be hard to debug afterwards as it doesn't error or anything on attempted mutations

webapp/src/components/tutorialValidators.tsx
```diff
@@ -1,19 +1,17 @@
...
     isValid: true,
     hint: null,
-};
+});
 
 export function GetValidator(metadata: CodeValidatorMetadata): CodeValidator {
```
jwunderl, line 13: Might still move these over and split up later on, but as mentioned in description putting that off till we have a few more use cases to see how much we need to reuse it.

pxtlib/tutorial.ts
```diff
@@ -563,4 +527,37 @@ ${code}
...
+     * need to cover getMetadataRegex + stripHiddenSnippets types, maybe more if any happen to be around.
+     **/
+    function getBlockSnippetCode(mdSnippet: string): string[] {
+        let hintCode: string[] = [];
+
+        mdSnippet?.replace(/((?!.)\s)+/g, "\n")
+            .replace(
+                /``` *(block|blocks)\s*\n([\s\S]*?)\n```/gim,
+                function (m0, m1, m2) {
+                    hintCode.push(`{\n${m2}\n}`);
```
jwunderl, line 556: minor diff, we were only accounting for first snippet in a hint, probably affects no one but probably good to confirm. We might also want to include package keys in future to avoid packages changing but that's probably fine

thsparks (approved):
LGTM! Thanks :)

localtypings/pxtarget.d.ts
```diff
@@ -1190,9 +1178,13 @@ declare namespace pxt.tutorial {
         execute(options: CodeValidationExecuteOptions): Promise<CodeValidationResult>;
     }
 
+    interface CodeValidatorBaseProperties {
+        enabled?: string;
```
kimprice, line 1182: Why is this a string and not a boolean? Does it have to do with getting the value from markdown?

jwunderl: Yes, just matching the existing types for now; this is part of the other change I played around with but didn't commit, which would be to make it so basically all the webapp side does is link up parsing these into proper types and pass them to a validator object over in pxtlib; deferring doing that bit till we have a few more examples of validator types to work with so we can scope out what possible options / needs better and make sure it has the right api

pxtlib/tutorial.ts
```diff
@@ -135,7 +124,7 @@ namespace pxt.tutorial {
                         m2 = "";
                         break;
                     case "tutorialValidationRules":
```
kimprice, line 126: why not remove the case here?

kimprice: Or remove "tutorialValidationRules" from getMetadataRegex?

jwunderl: I suppose can remove them now, this was the first file I started cleaning things up from so I hadn't made up my mind to fully clean it out yet and I guess I didn't come back to do the last few lines since they didn't really change anything

kimprice (approved):

PR #9645, opened 2023-08-21
Sync kiosk changes

eanders-ms:
Pulling in changes from pxt-arcade/kiosk/src


srietkerk (approved):
Thanks for porting, Eric!

PR #9646, opened 2023-08-21
Quick change display name for custom namespace

THEb0nny:
Add a custom.ts property to the example so that you can quickly change the display name for the specified namespace.
The user will always be in front of how to change the display name of the category.


THEb0nny:
@kimprice You can view? It seems to me that this option should initially be specified so that all users know how to change the name of the created namespace...

kimprice (approved):

jwunderl (approved):

PR #9647, opened 2023-08-21
Add insufficient blocks case to code validation

kimprice:
The insufficient blocks case is when the tutorial step requires more than one of the same type of block and the user has at least one of that type of block, but not all of the blocks. For example, in the picture below the user needs two "variables_set" blocks but only has one. Note: this example is a made-up tutorial step as I don't think we have any current tutorials that require more than one block of the same type in a single step.

Joey already added a check for insufficient blocks here: https://github.com/microsoft/pxt/blob/master/pxtblocks/code-validation/validateBlocksExist.ts#L27
This PR just uses that check to create the appropriate code validation message to the user.
image

### DIFF
--- a/webapp/src/components/tutorialValidators.tsx
+++ b/webapp/src/components/tutorialValidators.tsx
@@ -84,7 +84,7 @@ export class BlocksExistValidator extends CodeValidatorBase {
             errorDescription = lf("Make sure your blocks are connected to the rest of your code like this.");
         } else if (insufficientBlocks.length > 0) {
             isValid = false;
-            errorDescription = lf("Make sure you have enough blocks in your workspace. It should look like this.");
+            errorDescription = lf("Make sure you have enough blocks and that they're connected to the rest of your code. It should look like this.");
         }
 
         const blockImages = stepInfo?.hintContentMd ? (<div>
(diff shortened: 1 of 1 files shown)


kimprice:
We don't currently have a separate case for insufficient blocks that are disabled - it'll show the same message. For example:
image

webapp/src/components/tutorialValidators.tsx
```diff
@@ -81,6 +82,9 @@ export class BlocksExistValidator extends CodeValidatorBase {
         } else if (disabledBlocks.length > 0) {
             isValid = false;
             errorDescription = lf("Make sure your blocks are connected to the rest of your code like this.");
+        } else if (insufficientBlocks.length > 0) {
+            isValid = false;
+            errorDescription = lf("Make sure you have enough blocks in your workspace. It should look like this.");
```
thsparks, line 87: Since it will send this message if the block exists but is disconnected, perhaps we could rephrase slightly: "Make sure you have enough blocks connected to your workspace. It should look like this.

jwunderl (approved):

eanders-ms (approved):

thsparks (approved):
One nit, otherwise LGTM!

jwunderl:
(leaving a note here for other reviewers that we'll be looking at adding other authoring for specifying required blocks vs just from highlighting in a separate pr)

PR #9651, opened 2023-08-23
Switch Editor Modes When Changing Between Py/TS in File Explorer

thsparks:
Fixes https://github.com/microsoft/pxt-arcade/issues/6028

With this change, when a user selects `main.py` (or any other python file) while in typescript mode, the editor will fully switch to python mode, which means any updated javascript will be translated into python before the file switch occurs, preventing code loss. The same will also happen in reverse when selecting a typescript file while in python mode. This behavior aligns with what happens when you click on `main.blocks` (it will switch to the blocks editor).

Upload target: https://arcade.makecode.com/app/962a0b823b61325c162a2fed0995f02a87f2aec4-d0020cc65b

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1283,10 +1283,19 @@ export class ProjectView
     setSideFile(fn: pkg.File, line?: number) {
         let fileName = fn.name;
         let currFile = this.state.currFile.name;
+
         if (fileName != currFile && pxt.editor.isBlocks(fn)) {
-            // Going from ts -> blocks
+            // Going from ts/py -> blocks
             pxt.tickEvent("sidebar.showBlocks");
             this.openBlocks();
+        } else if (fileName.endsWith(".py") && currFile.endsWith(".ts")) {
+            // Going from ts -> py
+            pxt.tickEvent("sidebar.showPython");
+            this.openPythonAsync();
+        } else if (fileName.endsWith(".ts") && currFile.endsWith(".py")) {
+            // Going from py -> ts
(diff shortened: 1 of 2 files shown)


webapp/src/app.tsx
```diff
@@ -1283,10 +1283,19 @@ export class ProjectView
     setSideFile(fn: pkg.File, line?: number) {
         let fileName = fn.name;
         let currFile = this.state.currFile.name;
+
         if (fileName != currFile && pxt.editor.isBlocks(fn)) {
-            // Going from ts -> blocks
+            // Going from ts/py -> blocks
             pxt.tickEvent("sidebar.showBlocks");
             this.openBlocks();
+        } else if (fileName.endsWith(".py") && currFile.endsWith(".ts")) {
```
jwunderl, line 1291: do we need to special case ts->py / py->ts here vs. just 'going from not py -> py' / 'going from not ts -> ts'? I might be missing something but it feels like there'd be edge cases with 'ts -> json/jres file -> py' here like the repro steps if it's specifically checking curr file & newly set file?

thsparks: This turned out to be somewhat involved, as there are quite a few places in the language-switching code that assume you have main.* selected. I went in and tried to update the critical locations to remove this assumption, so I think it should be working now and my own testing seems to have gone okay, but totally possible I missed something.

New upload target: https://arcade.makecode.com/app/962a0b823b61325c162a2fed0995f02a87f2aec4-d0020cc65b

abchatra:
Nice work!

jwunderl (approved):
Nice, LGTM!

Issue #9653, opened 2023-08-24
Cannot set the dropdown's value to an unavailable option.


jp-96:
# PXT Bug - FieldUserEnum

**Describe the bug**

Console message:
> Cannot set the dropdown's value to an unavailable option.
> Block type: radioMessageCode, Field name: MEMBER, Value: add1

The `radio send` block's dropdown item is unselected!

![image](https://github.com/microsoft/pxt/assets/2490958/3eeb8de2-3786-4c2a-b8c7-e964c536fef8)

**To Reproduce**
1. Create new project on MakeCode for micro:bit
2. Select `radio-broadcast` extention
3. `radio send` block
4. See console log

**Expected behavior**

`message1` is selected and there is no warning message.

**Screenshots**

![image](https://github.com/microsoft/pxt/assets/2490958/bc0b85ee-80b6-460f-89f3-bbd80145065e)

Edit source code on JavaScript
![image](https://github.com/microsoft/pxt/assets/2490958/9c5b7266-d65c-411c-b364-87356623d75d)

When convert to Blocks, item is unselected.
![image](https://github.com/microsoft/pxt/assets/2490958/47f3d11d-fd91-4644-a72b-c256c07a6cc4)

jp-96:
transfered:
https://github.com/microsoft/pxt-arcade/issues/6037

PR #9654, opened 2023-08-24
Update field_userenum.ts

jp-96:
Data binding fails.
https://github.com/microsoft/pxt/issues/9653

Console warning:
Cannot set the dropdown's value to an unavailable option. Block type: radioMessageCode, Field name: MEMBER, Value: add1

### DIFF
--- a/pxtblocks/fields/field_userenum.ts
+++ b/pxtblocks/fields/field_userenum.ts
@@ -21,6 +21,18 @@ namespace pxtblockly {
         }
 
         doClassValidation_(value: any) {
+            // The format of the name is 10mem where "10" is the value and "mem" is the enum member
+            if (this.sourceBlock_ && this.sourceBlock_.workspace) {
+                const options = this.sourceBlock_.workspace.getVariablesOfType(this.opts.name);
+                options.some(model => {
+                    const [name, ] = parseName(model);
+                    if (name == value) {
+                        value = model.name
+                        return true
+                    }
+                    return false
+                });
+            }
             // update cached option list when adding a new kind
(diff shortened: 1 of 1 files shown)


pxtblocks/fields/field_userenum.ts
```diff
@@ -21,6 +21,18 @@ namespace pxtblockly {
         }
 
         doClassValidation_(value: any) {
+            // The format of the name is 10mem where "10" is the value and "mem" is the enum member
+            if (this.sourceBlock_ && this.sourceBlock_.workspace) {
```
jwunderl, line 25:
```suggestion
            if (this.sourceBlock_?.workspace) {
```

tiny nit, we tend to prefer safe navigation operator nowadays for ease of reading~

jwunderl (dismissed):
Thanks for the PR! Looks good to me; this also fixes https://github.com/microsoft/pxt-arcade/issues/6037, I believe you re-filed it this morning after I moved it over to arcade last night (we typically track / prioritize bugs we can see in the editors themselves as they make it easier to repro / identify how many people are affected by them, so when possible it's best to file an issue in pxt-arcade or pxt-microbit even if it affects all editors.)

Here's a test project for anyone else who wants to review, swap around the order of red / yellow / blue in typescript then switch to blocks to induce the error: https://makecode.microbit.org/_gPUFdTT0s1zu

jp-96:
> we typically track / prioritize bugs we can see in the editors themselves as they make it easier to repro / identify how many people are affected by them, so when possible it's best to file an issue in pxt-arcade or pxt-microbit even if it affects all editors.

I understand,
Thank you.

thsparks (approved):
Thanks!

abchatra:
@thsparks @jwunderl can we merge this PR?

PR #9655, opened 2023-08-24
Allow for smaller share dialog when we don't want previews

srietkerk:
The new share dialog allows for the recording and screenshotting of the simulators in micro:bit and arcade, but we don't have something to easily preview in Minecraft at the moment. That means that the "record gif" or "take screenshot" area of the share dialog is not useful and is also buggy. 

If one of our targets specifies that it doesn't want to record GIF previews or allow screenshots of a simulator in `pxtarget.json`, we now get rid of that part of the share modal. When getting rid of the preview area of the share modal, the input for the project name was awkwardly long. So I've also updated the styling so that if we don't have something to preview for the project, we use a smaller share modal.

**Before**
image

**After**
image
image

Closes https://github.com/microsoft/pxt-minecraft/issues/2317

### DIFF
--- a/webapp/src/share.tsx
+++ b/webapp/src/share.tsx
@@ -195,9 +195,7 @@ export class ShareEditor extends auth.Component<ShareEditorProps, ShareEditorSta
         const { visible, projectName: newProjectName, title, screenshotUri } = this.state;
         const { simScreenshot, simGif } = pxt.appTarget.appTheme;
         const hasIdentity = auth.hasIdentity() && this.isLoggedIn();
-        const light = !!pxt.options.light;
-        const thumbnails = pxt.appTarget.cloud && pxt.appTarget.cloud.thumbnails
-            && (simScreenshot || simGif);
+        const thumbnails = simScreenshot || simGif;
 
         const hasProjectBeenPersistentShared = parent.hasHeaderBeenPersistentShared();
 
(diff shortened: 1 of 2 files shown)


jwunderl (approved):

webapp/src/share.tsx
```diff
@@ -209,7 +207,7 @@ export class ShareEditor extends auth.Component<ShareEditorProps, ShareEditorSta
         return visible
             ? <Modal
                 title={lf("Share Project")}
-                className="sharedialog wide"
+                className={`sharedialog${thumbnails ? " wide" : ""}`}
```
jwunderl, line 210: Ahah, was my guesstimate of 40rem just the default when it's not `.wide`? Looks same

srietkerk: Yep! I was a bit mind blown by that!

image

webapp/src/share.tsx
```diff
@@ -195,9 +195,7 @@ export class ShareEditor extends auth.Component<ShareEditorProps, ShareEditorSta
         const { visible, projectName: newProjectName, title, screenshotUri } = this.state;
         const { simScreenshot, simGif } = pxt.appTarget.appTheme;
         const hasIdentity = auth.hasIdentity() && this.isLoggedIn();
-        const light = !!pxt.options.light;
-        const thumbnails = pxt.appTarget.cloud && pxt.appTarget.cloud.thumbnails
```
thsparks, line 199: Was `pxt.appTarget.cloud.thumbnails` superfluous in some way? I don't really know what it represents, but was curious as to why it was removed.

thsparks (approved):
One question, but just a curiosity. Not blocking.

webapp/src/share.tsx
srietkerk: I'm not sure as to what it did, honestly. But the `light` and `thumbnails` variables were never used, and for the case of share dialog, if it's specified in `pxtarget.json` that we do or don't want the thumbnails,  I think that's enough to either supply or ignore them.

thsparks: Ah, got it. I missed that it was unused before your change.

PR #9657, opened 2023-08-26
High contrast: Icons for block categories now have colors in Minecraft

srietkerk:
All of our other targets still have the colors on the icons for the block categories, so we want Minecraft to match that. 

Most of the categories are super straightforward because you can just change the icon color via code. However, the `Agent` category is a bit tricky because the icon that we're using doesn't have an associated icon in Semantic; we're using a static png. Because of this, changing the color of the `Agent` icon via code isn't really possible. That's where the new `Agent_icon_orange` file comes in. I just alternate between the two files to make the color changes work. I wasn't sure if referencing files across the repos like this would cause problems in production, so we might have to find another solution, but it works locally. One thing to note, though, is that when you first switch to high contrast mode, it takes a second for the agent icon to load in, so there is definitely room for improvements.

image

Will get an upload target on Monday.

Closes: https://github.com/microsoft/pxt-minecraft/issues/2062

### DIFF
--- a/theme/toolbox.less
+++ b/theme/toolbox.less
@@ -260,6 +260,50 @@ div.blocklyTreeIcon span {
 }
 
 /*******************************
+    Inverted Toolbox
+*******************************/
+#root.hc .invertedToolbox div div {
+    div.blocklyTreeRow .blocklyTreeIcon {
+        color: var(--block-meta-color);
+    }
+
+    div.blocklyTreeRow.blocklyTreeSelected .blocklyTreeIcon {
+        color: #fff;
+    }
+}
+
+.invertedToolbox .blocklyTreeRoot div div {
+    div.blocklyTreeRow {
+        background-color: var(--block-meta-color);
+        color: #fff;
+    }
+
+    div.blocklyTreeRow.blocklyTreeSelected {
+        background-color: var(--block-meta-color);
+    }
+
+    div.blocklyTreeRow:hover {
+        background-color: var(--block-faded-color);
+        color: #fff;
+    }
+}
+
+/*******************************
+    Colored Toolbox
+*******************************/
+.coloredToolbox .blocklyTreeRoot div div {
+    div.blocklyTreeRow {
+        color: var(--block-meta-color);
+        border-left: var(--block-category-border);
+    }
+
+    div.blocklyTreeRow.blocklyTreeSelected {
+        background-color: var(--block-meta-color);
+        color: #fff;
+    }
+}
+
+/*******************************
         Media Adjustments
 *******************************/
 
--- a/theme/toolbox.less
+++ b/theme/toolbox.less
@@ -229,6 +229,14 @@ div.blocklyTreeIcon span {
 .blocklyFlyoutLabelIcon.blocklyFlyoutIconfunctions {
     font-family: xicon !important;
 }
+.blocklyTreeIcon.image-icon {
+    background-image: var(--image-icon-url);
+    width: 30px;
+    height: 100%;
+    background-size: 20px !important;
+    background-repeat: no-repeat !important;
+    background-position: 50% 50% !important;
+}
 
 /*******************************
       Toolbox Animation
@@ -262,7 +270,7 @@ div.blocklyTreeIcon span {
 /*******************************
     Inverted Toolbox
 *******************************/
-#root.hc .invertedToolbox div div {
+#root.hc .invertedToolbox {
     div.blocklyTreeRow .blocklyTreeIcon {
         color: var(--block-meta-color);
     }
@@ -272,7 +280,7 @@ div.blocklyTreeIcon span {
     }
 }
 
-.invertedToolbox .blocklyTreeRoot div div {
+.invertedToolbox .blocklyTreeRoot {
     div.blocklyTreeRow {
         background-color: var(--block-meta-color);
         color: #fff;
@@ -291,10 +299,10 @@ div.blocklyTreeIcon span {
 /*******************************
     Colored Toolbox
 *******************************/
-.coloredToolbox .blocklyTreeRoot div div {
+.coloredToolbox .blocklyTreeRoot {
     div.blocklyTreeRow {
         color: var(--block-meta-color);
-        border-left: var(--block-category-border);
+        border-left: 8px solid var(--block-meta-color);
     }
 
     div.blocklyTreeRow.blocklyTreeSelected {
--- a/webapp/src/toolbox.tsx
+++ b/webapp/src/toolbox.tsx
@@ -747,7 +747,6 @@ export class TreeRow extends data.Component<TreeRowProps, {}> {
         const { nameid, advancedButtonState, subns, name, icon } = this.props.treeRow;
         const appTheme = pxt.appTarget.appTheme;
         const metaColor = this.getMetaColor();
-        const hc = core.getHighContrastOnce();
 
         const invertedMultipler = appTheme.blocklyOptions
             && appTheme.blocklyOptions.toolboxOptions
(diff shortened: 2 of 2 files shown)


webapp/src/toolbox.tsx
```diff
@@ -776,11 +777,24 @@ export class TreeRow extends data.Component<TreeRowProps, {}> {
...
         }
+        let treeIconStyle: React.CSSProperties = {
+            color: '#fff', display: 'inline-block'
+        }
         let treeRowClass = 'blocklyTreeRow';
         if (appTheme.invertedToolbox) {
             // Inverted toolbox
             treeRowStyle.backgroundColor = (metaColor || '#ddd');
             treeRowStyle.color = '#fff';
+            if (hc) {
```
jwunderl, line 788: Can we do this over in pxt-minecraft by uri embedding png like https://github.com/microsoft/pxt-minecraft/blob/master/theme/site/globals/site.variables#L61 & selecting off existing hc class, rather than changing any js?

srietkerk: Just to clarify, are you saying to do this for all of the icons or just the agent icon?

jwunderl: hm ideally if it were in less that would feel a little nicer (e.g. could set border-color / metaColor / icon img / etc as css variables), but this class is already a little explicit that way so it's good.

One thing to confirm is that behavior is as expected when you toggle on / off from within editor (vs. already being in high contrast mode when entering editor) since `getHighContrastOnce` means it relies on a forced render from elsewhere to pick up the high contrast change. I think we might already get that from editor interface itself / if you noticed the image pop in in the first place you probably already tested this though~

jwunderl (approved):

webapp/src/toolbox.tsx
srietkerk: Yeah, toggling high contrast before entering or after entering the editor behave the same.

srietkerk:
**UPDATE**

I've decided to go the CSS route so it's a bit easier to interpret going forward. There are still a couple things that I want to fiddle with, but I wanted to push up my current changes.

The following has changed on this first pass:
1. Hover and not hovering for inverted toolbox is no longer handled in JS, but is handled by the `:hover` property in CSS
2. Styling for inverted and colored toolboxes is moved to `toolbox.less`
3. CSS variables made for block meta color, the fade color that applies to toolbox categories for inverted toolbox hover, and the border properties that we use for our colored toolbox. Looking at it now, I will likely just move  this fully into CSS and pass in the `--block-meta-color` variable for it's color property.
4. I removed the logic for flipping the border side when the editor changes between right to left and left to right orientation for colored toolboxes. From what I can tell, we sort this out programmatically. When adding the `coloredToolbox` styling to CSS, I noticed that the styling in CSS that specified `border-left` automatically turned into `border-right`.

What I want to change still:
1. The icon image style. Right now, we are using a style element to apply more specific styling to blocks categories that use images as their icons. I think a majority of this styling can just live in CSS, and in the case of the `backgroundImage` url, I'll use a CSS variable as was done with the meta color. 

Things that need clarification:
1. The animation stuff that was in the styling. Thanks to the git blame, I was able to find this PR: https://github.com/microsoft/pxt/pull/6774, showing what the animation stuff should be giving us. This is broken as far as I can tell. I'm wondering if this is something that I should open a bug for or if I should just get rid of the animation logic.

webapp/src/toolbox.tsx
```diff
@@ -775,49 +754,18 @@ export class TreeRow extends data.Component<TreeRowProps, {}> {
...
-                }
-            } else {
-                treeRowStyle.backgroundColor = (metaColor || '#ddd');
-            }
-            treeRowStyle.color = '#fff';
-        }
+            paddingLeft: '0px',
+            "--block-meta-color": metaColor,
+            "--block-faded-color": pxt.toolbox.fadeColor(metaColor || '#ddd', invertedMultipler, false),
+            "--block-category-border": `8px solid ${metaColor}`
```
jwunderl, line 760: I'd lean towards eliding this and just referencing `--block-meta-color` directly; i.e. just replace any `var(--block-category-border)` in the `.less` with `8px solid var(--block-meta-color`).

theme/toolbox.less
```diff
@@ -259,6 +259,50 @@ div.blocklyTreeIcon span {
     background-color: #e4e4e4;
 }
 
+/*******************************
+    Inverted Toolbox
+*******************************/
+#root.hc .invertedToolbox div div {
```
jwunderl, line 265: probably don't need the `div div` here, and if we do we should figure out why (selectivity fight?) -- default behavior in css selectors is that it will go down as far as needed no matter how many divs are between (if you want to explicitly say "only divs that are directly descendants" it requires an > operator like `.parent > .direct-child`)

theme/toolbox.less
```diff
@@ -259,6 +259,50 @@ div.blocklyTreeIcon span {
...
+    }
+}
+
+/*******************************
+    Colored Toolbox
+*******************************/
+.coloredToolbox .blocklyTreeRoot div div {
+    div.blocklyTreeRow {
+        color: var(--block-meta-color);
+        border-left: var(--block-category-border);
```
jwunderl, line 297: I think this will still need rtl handling so that it "looks right"; check arabic language in micro:bit. I manually set border-left on basic category as an example vs. current behavior, you can see it is on the opposite side from the icon: 
image

when in an rtl language we set the rtl class on body root so you can start the selector with `.main.rtl` to set the special handling for it -- e.g.

```
.main.rtl .coloredToolbox .blocklyTreeRoot .blocklyTreeRow {
    border-left: none;
    border-right: 8px solid var(--block-meta-color);
}
```

jwunderl (commented):
Awesome, nice clean up! Thanks for listening to me ramble on this for a bit & then fixing everything!

left a few comments since you with minor things since you mentioned still wanting to clean up a little more, Also, in case it's helpful, here's a pointer to the place I can remember using a css variable for an image url , I remember it being a little finicky / having to play with it but that was probably just cause I was doing a datauri + css variable instead of just a url https://github.com/microsoft/pxt/pull/9329/files#diff-7047859ba8a434609145f1795eb0ee90ad1991dcdb4e22fc986bc181f2a55862R350

theme/toolbox.less
srietkerk: Are we worried about an override of sorts? Even though the code is `border-left`, in the live site, when I choose Arabic, it gets changed to `border-right`.
image

thsparks:
> Hover and not hovering for inverted toolbox is no longer handled in JS, but is handled by the :hover property in CSS

I might be missing something obvious, but I couldn't find where this is happening. Can you point me to it?

> The animation stuff that was in the styling. Thanks to the git blame, I was able to find this PR: https://github.com/microsoft/pxt/pull/6774, showing what the animation stuff should be giving us. This is broken as far as I can tell. I'm wondering if this is something that I should open a bug for or if I should just get rid of the animation logic.

Personally, I'd recommend just opening a bug for now. No need to remove the code; it could be helpful for whoever ends up debugging.

srietkerk:
> I might be missing something obvious, but I couldn't find where this is happening. Can you point me to it?

Yeah, no problem, https://github.com/microsoft/pxt/blob/8aa26877869c51363dd0e3d8b77f71ee7a301e22/theme/toolbox.less#L285

thsparks:
> > Hover and not hovering for inverted toolbox is no longer handled in JS, but is handled by the :hover property in CSS
> 
> I might be missing something obvious, but I couldn't find where this is happening. Can you point me to it?
> 
> > The animation stuff that was in the styling. Thanks to the git blame, I was able to find this PR: #6774, showing what the animation stuff should be giving us. This is broken as far as I can tell. I'm wondering if this is something that I should open a bug for or if I should just get rid of the animation logic.
> 
> Personally, I'd recommend just opening a bug for now. No need to remove the code; it could be helpful for whoever ends up debugging.

Unless it's causing a crash or something, that is. In that case, I'd say we should remove it and reference the commit that removed it in the bug. If it's just not doing anything then I think it's fine to leave it in.

thsparks (approved):

srietkerk:
> > Personally, I'd recommend just opening a bug for now. No need to remove the code; it could be helpful for whoever ends up debugging.
> 
> Unless it's causing a crash or something, that is. In that case, I'd say we should remove it and reference the commit that removed it in the bug. If it's just not doing anything then I think it's fine to leave it in.

Yep, not causing any issues, just not adding anything.

theme/toolbox.less
srietkerk: Yeah, it's what I'm getting on local serve with my new changes. When I added the special case to handle it, because I also figured we'd need a special case to ensure it was working, the styling was actually messed up. I'll retry this, though and see what happens.

srietkerk: This is what it looks like with the specific style: 
image

image

jwunderl: I think the -right and -left are reversed in that css rule screenshot? default / non-rtl version should be setting `border-left`, and `.main.rtl` should be `none`ing `border-left` and setting `border-right`

jwunderl: ah, okay, chatted online, I didn't realize that our less preprocessor step actually swaps border-left to border-right and vice versa automatically when making `rtlsemantic.css`. Fun / something to remember, nice to know we don't have to edge case things

webapp/src/toolbox.tsx
```diff
@@ -774,57 +752,29 @@ export class TreeRow extends data.Component<TreeRowProps, {}> {
...
+            treeRowClass += ' blocklyTreeAnimate';
         }
 
         // Icon
-        const iconClass = `blocklyTreeIcon${subns ? 'more' : icon ? (nameid || icon).toLowerCase() : 'Default'}`.replace(/\s/g, '');
+        let iconClass = `blocklyTreeIcon${subns ? 'more' : icon ? (nameid || icon).toLowerCase() : 'Default'}`.replace(/\s/g, '');
         let iconContent = subns ? pxt.toolbox.getNamespaceIcon('more') : icon || pxt.toolbox.getNamespaceIcon('default');
-        let iconImageStyle: JSX.Element;
+        let iconImageStyle: React.CSSProperties = {
+            "--image-icon-url": `url("${Util.pathJoin(pxt.webConfig.commitCdnUrl, encodeURI(icon))}")!important`,
```
jwunderl, line 771:
```suggestion
            "--image-icon-url": iconContent.length > 1 ? `url("${Util.pathJoin(pxt.webConfig.commitCdnUrl, encodeURI(icon))}")!important` : undefined,
```

nit: might be worth only setting this when `iconContent.length > 1` just for future debugging sake, since as written the variable will be set even when it won't be used / doesn't actually point to a valid url. Wouldn't cause a behavioral difference though, just to prevent people from getting weirded out on why they're seeing seemingly invalid urls in css. Can hoist that `iconContent.length > 1` out to a `const isImageIcon = iconContent.length > 1` as well if you want to dedup / simplify a little more

jwunderl (approved):
Again nice job :)

PR #9658, opened 2023-08-28
Time machine

riknoll:
Opening a PR to check in the time machine code as it stands now. It's currently hidden behind a flag.

The final UI for this is TBD, but the current UI works. For diffing/patching, I'm using Google's diff-match-patch library which is Apache 2.0 licensed. Currently every single change creates a history entry which is way too much. Once we have the UI finalized I'll have a better idea of what interval we want to use.

Here's an upload target link:

https://arcade.makecode.com/app/49743c2e211ad4a2542a4d91d0a5a3a059db044c-f4a46a5ba2

When using that link, I recommend starting with a fresh project. Also, be sure to only edit said project in the above target link; it will work in other editors but the history will get corrupted. It's on my todo list to figure out a way to detect when this happens


pxteditor/workspace.ts
```diff
@@ -22,6 +22,37 @@ namespace pxt.workspace {
         version: Version;
     }
 
+    export interface HistoryFile {
+        entries: HistoryEntry[];
+    }
+
+    export interface HistoryEntry {
+        timestamp: number;
```
jwunderl, line 30: Maybe an editor version stamp here, or we could push a HistoryEntry whenever we load in a new version I suppose. (In theory we could scrape from pxt.json but that's messy / a little inconsistent)

jwunderl (approved):
One thing to think about, how do we wanna have editor updates here? E.g. if we need a block patch like https://github.com/microsoft/pxt-arcade/blob/master/editor/extension.ts#L5 after a given update, I guess we would have to apply that when reverting to an old 'commit'?

eanders-ms (commented):
Related to Joey's question about block patching, if a patch were applied to the latest version, does this render the stored diff to the previous snapshot corrupt? If the user then reverts, would that result in a corrupt project?

riknoll:
@eanders-ms because we're always just taking a straight diff, whenever the project is saved again that patch will just be included in the user changes. Shouldn't cause any corruption.

I will add the editor version to the history entries, though.

jwunderl (approved):

PR #9665, opened 2023-08-30
Removing stable reference from kiosk

abchatra:


eanders-ms (approved):
Seems right to me. @srietkerk?

srietkerk (commented):
There was a reason why we were pointing to stable instead of live, but I can't remember at the moment. Is there a reason why we want to point it to live from now on?

PR #9668, opened 2023-09-01
kiosk: increase gamepad polling frequency

eanders-ms:
Faster polling to reduce odds of missing an input event.


kimprice:
How did you decide on the number?

eanders-ms:
I looked at the behavior of Xbox's `@gaming/react-gamepad-navigation`​ library and matched their polling interval. I'd link to it here, but it isn't open source.

kimprice (approved):

PR #9669, opened 2023-09-02
Kiosk navigation improvements

eanders-ms:
* Added UI sound effects. Actual sounds are placeholder. To be team reviewed.
* Decoupled gamepad polling from event generation. Gamepad polling at 50ms required a "cooldown" to be added so that multiple pressed events wouldn't fire in rapid succession.
* Added the ability to block the "is A pressed" event until A is released, so that this event doesn't bleed across UI transitions.
* Added a "volume" url parameter to specify sound effects volume. We might want to forward this to the simulator. Default value is 0.5.


kiosk/src/Components/MainMenu.tsx
```diff
@@ -43,13 +50,16 @@ const MainMenu: React.FC<IProps> = ({ kiosk }) => {
...
             tickEvent("kiosk.addGamePageLoaded");
             kiosk.launchAddGame();
+            playSoundEffect("select");
         }
         if (
             deleteButtonSelected &&
             (kiosk.gamepadManager.isAButtonPressed() ||
                 kiosk.gamepadManager.isBButtonPressed())
         ) {
+            kiosk.gamepadManager.blockAPressUntilRelease();
```
srietkerk, line 60: Would we want to block for other cases as well? Was the problem that if people selected delete game and with the increased polling that the pressed would stay triggered and just cancel the delete modal really quickly?

kiosk/src/Models/GamepadManager.ts
```diff
@@ -55,6 +71,27 @@ export class GamepadManager {
...
+        );
+    }
+
+    setCooldownTimer(control: GamepadControl) {
+        this.cooldownTimers[control] =
+            Date.now() + configData.GamepadCooldownMilli;
+    }
+
+    isPressedWithCooldown(control: GamepadControl, isPressed: () => boolean): boolean {
+        if (this.isCoolingDown(control)) return false;
```
srietkerk, line 87: Just a clarification question. We have this check because if the control is cooling down, then the control has already been pressed, so we don't want to trigger a press again until the cool down is done?

kiosk/src/Components/AddingGame.tsx
```diff
@@ -48,6 +50,7 @@ const AddingGame: React.FC<IProps> = ({ kiosk }) => {
         if (menuButtonSelected && kiosk.gamepadManager.isAButtonPressed()) {
             tickEvent("kiosk.toMainMenu");
             kiosk.showMainMenu();
+            playSoundEffect("select");
```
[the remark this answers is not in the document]
eanders-ms: Done

kiosk/src/Models/GamepadManager.ts
eanders-ms: That is correct. It's similar to repeat delay on a keyboard. A keyboard key may be held down continuously, but generation of "key pressed" events is throttled to a max frequency.

kiosk/src/Components/MainMenu.tsx
eanders-ms: I wanted to be careful about where to add this. There may be a few more places where it's useful. It's a short term fix. I have a new game controller navigation lib incoming that handles this a different way.

kimprice:
Are the sound effect names just carried over from whatever library they are from? Also, should we have a comment in the SoundEffectService that tells where we found the sounds?

eanders-ms:
> Are the sound effect names just carried over from whatever library they are from? Also, should we have a comment in the SoundEffectService that tells where we found the sounds?

The filenames are unmodified from their source library. But these are placeholder sounds just to get something out for testing. As such, I don't want to spend time attributing them just yet. My plan forward is to author the final sounds from scratch. We only need a few of them, so I think this will be easiest and quickest path. On the other hand, if we end up using 3rd party sounds I will certainly attribute them and sort out licensing.

kimprice (approved):

srietkerk (approved):
Looks great! Excited to play with all the sounds!

PR #9670, opened 2023-09-05
Kiosk: retry on initial fetch if there's a problem

srietkerk:
There are times when kiosk will throw an error on launch/visit that it "Failed to fetch", referring to the initial fetch that we perform for the games from `GameList.json`. The error that we get from this is usually nothing serious and gets fixed with retry logic. That's what this PR is for. 

Closes https://github.com/microsoft/pxt-arcade/issues/6013


eanders-ms (approved):

kimprice (approved):
LGTM!

PR #9675, opened 2023-09-06
Code validation: validate-exists marker

kimprice:
This PR adds validation for blocks which have a `// @validate-exists` comment. This allows validation for blocks without requiring the block to be highlighted. In the tutorial markdown, tutorial authors will specify at the global scope which markers to use for validation: `validate-exists`, `highlight`, or both (`validate-exists, highlight`). If no markers are specified, it defaults to using both.

`````
```validation.global
# BlocksExistValidator
* markers: validate-exists, highlight
```
`````

Note that if both `validate-exists` and `highlight` are specified as markers and one block has both `// @validate-exists` and `// @highlight`, that block will be counted twice.

This also sets up future work to support multiple validate comments on one block. For example, you might have a `// @validate-sequence` and `// @validate-exists` on one block.

### CODE REVIEW

**Review by eanders-ms — APPROVED** [review]

File: pxtlib/tutorial.ts [hunk 1]
```diff
@@ -530,9 +530,27 @@ ${code}
...
+                requiredBlocks[block] = (requiredBlocks[block] || 0) + 1;
+            });
+        }
+        return requiredBlocks;
+    }
+
     export function getTutorialStepHash(tutorial: TutorialOptions): string {
         const { tutorialStepInfo, tutorialStep } = tutorial;
-        const body = tutorialStepInfo[tutorialStep].hintContentMd;
+        const body = tutorialStepInfo[tutorialStep].contentMd;
```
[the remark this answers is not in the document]
kimprice: Actually, I need to double check this. I needed to change it when I was working with one of the tutorial markdowns but I'm not positive on the change.

thsparks (approved):
The change seems okay if this is the approach we want to take (and I'm not strictly against it), but I do wonder: is there ever a scenario where the user would have //@validate-exists comments but *not* specify the exists marker? If not, would it be better to simply have that comment always get checked and optionally have a flag to disable looking at highlight blocks (rather than the whole marker system)?

I know you mentioned we might extend the marker system later on for things like sequence. I'm not totally sure I see how that would work without a concrete example, but I do wonder if it'd be better to wait and only add the complexity once we need it. I suppose the downside is, if we went that route, we'd have the `ignoreHighlights` flag separate from the markers, which could be a bit confusing...

Main takeaway - do we have a concrete example of how these markers will be extended in the future? I think that'd be helpful in understanding the goals of the marker system as a whole.

thsparks:
> The change seems okay if this is the approach we want to take (and I'm not strictly against it), but I do wonder: is there ever a scenario where the user would have //@validate-exists comments but _not_ specify the exists marker? If not, would it be better to simply have that comment always get checked and optionally have a flag to disable looking at highlight blocks (rather than the whole marker system)?
> 
> I know you mentioned we might extend the marker system later on for things like sequence. I'm not totally sure I see how that would work without a concrete example, but I do wonder if it'd be better to wait and only add the complexity once we need it. I suppose the downside is, if we went that route, we'd have the `ignoreHighlights` flag separate from the markers, which could be a bit confusing...
> 
> Main takeaway - do we have a concrete example of how these markers will be extended in the future? I think that'd be helpful in understanding the goals of the marker system as a whole.

Alternatively we could keep all of this the same but not expose markers to the user (yet), then just have "exists, highlight" as the default and expose an ignoreHighlights flag they can set which, behind the scenes, removes the marker. That way we keep the flexibility of the system without locking ourselves to it completely should we decide to do things differently down the road.

localtypings/pxtarget.d.ts
```diff
@@ -551,6 +551,7 @@ declare namespace pxt {
         usedBlocks: Map<number>;
         snippetBlocks: Map<Map<number>>;
         highlightBlocks: Map<Map<number>>;
+        validateBlocks: Map<Map<string[]>>;
```
jwunderl, line 554: Not in this pr, but we'll want to solve the issue of having a block with `// @highlight @validate-exists` 'counting twice' eventually as it would feel buggy for that to require the same block twice. One way we discussed in a sync earlier would be to make this a `Map<Map<BlockData[]>>` , where `BlockData` would be something like

```ts
interface BlockData {
    type: string; // the pxtc blockId that is currently stored in validateBlocks
    id: string; // the id attached to the block in the blockly workspace
}
```

e.g. in https://github.com/microsoft/pxt/pull/9675/files#diff-cae476038f191f5305fbc82c2c9f96c74216375046a125e25dafe2bb12be86b1R119 `id` would come from `blk.id`. The ID is just a random value generated when going from xml -> blockly workspace but can be useful to e.g. identify when the tags came from separate (similar) blocks

localtypings/pxtarget.d.ts
```diff
@@ -551,6 +551,7 @@ declare namespace pxt {
         usedBlocks: Map<number>;
         snippetBlocks: Map<Map<number>>;
         highlightBlocks: Map<Map<number>>;
```
jwunderl, line 553: Not necessarily in this pr, but this could probably be merged in as just a field in `validateBlocks` (maybe would wanna rename it a bit more generic e.g. `taggedBlocks["highlight"]`) -- we'd have to do the `list -> map of counts` when accessing it but that is being done for `validateBlocks["exists"]` anyway / the calculation is quick enough, and we can dedup the logic a bit.

webapp/src/components/tutorialValidators.tsx
```diff
@@ -53,21 +61,43 @@ export class BlocksExistValidator extends CodeValidatorBase {
             return defaultResult();
         }
 
+        // if no valid markers are specified, default to "highlight"
+        if (!this.markers || (!this.markers.includes("highlight") && !this.markers.includes("exists"))) {
+            this.markers = ["highlight"];
```
jwunderl, line 66: I think I agree starting with `validate-exists` defaulting on as well is fine / just leaving this as an escape hatch if we don't want highlight.

webapp/src/components/tutorialValidators.tsx
```diff
@@ -53,21 +61,43 @@ export class BlocksExistValidator extends CodeValidatorBase {
...
         const stepHash = pxt.tutorial.getTutorialStepHash(tutorialOptions);
-        const stepHighlights = allHighlightedBlocks[stepHash];
+        let highlightBlocksCount: pxt.Map<number> = {};
+        let validateBlocksCount: pxt.Map<number> = {};
+
+        if (allHighlightedBlocks && this.markers.includes("highlight")) {
+            highlightBlocksCount = allHighlightedBlocks[stepHash];
+        }
+
+        if (allValidateBlocks && this.markers.includes("exists")) {
```
jwunderl, line 83: if we refactor out the separate `highlightBlocks` field and just stored in one, this portion would get a little simpler / just a loop over `this.markers` and summing up

webapp/src/tutorial.tsx
```diff
@@ -100,13 +103,21 @@ function getUsedBlocksInternalAsync(code: string[], id: string, language?: strin
...
+                        let comment = blk.getCommentText();
                         if (comment && /@highlight/.test(comment)) {
                             if (!highlightBlocks[snippetHash][blk.type]) {
                                 highlightBlocks[snippetHash][blk.type] = 0;
                             }
                             highlightBlocks[snippetHash][blk.type] = highlightBlocks[snippetHash][blk.type] + 1;
                         }
+                        while (comment && /@validate-\S+/.test(comment)) {
+                            const marker = comment.match(/@validate-(\S+)/)[1];
+                            comment = comment.replace(/@validate-\S+/, "");
```
jwunderl, line 115: I think cutting out just the `@` would be fine (so it's in there as key `validate-exists` instead of just `exists`) -- avoids us having to remember that the value changes slightly from markdown -> map, and also makes it so we don't have any weirdness if we need to track other meta fields later on (e.g. including highlight)

jwunderl (approved):
Looks good to me, I'd probably add in the exists marker being on by default in `tutorialValidator.ts` but every other comment is just ideas for internal structure / implementation details that we can follow up on in separate prs.

webapp/src/tutorial.tsx
jwunderl: Actually just thinking on this again, would `check-` as the prefix maybe be easier to read (so like `check-exists` / `check-sequence` / ...)? `validate` feels mildly fancy / technical to me. This doesn't matter much / we can defer the decision till a bit later if we want since it'll be a little while before we start shipping this code / markdown making use of it

jwunderl (approved):
:shipit:

webapp/src/tutorial.tsx
kimprice: I'll just leave validate for now, but I did change this regex match to basically grab all the keys in the comment regardless of the prefix. So it's getting highlight now too and I can build on that in a future PR.

PR #9687, opened 2023-09-18
Kiosk: Use pxt.tickEvent

eanders-ms:
Following multplayer's implementation.


kiosk/src/index.tsx
```diff
@@ -18,19 +22,43 @@ function enableAnalytics() {
...
+    if (pxt.options.debug) pxt.debug = console.debug;
+
+    pxt.setupWebConfig((window as any).pxtConfig || pxt.webConfig);
+    pxt.setAppTarget(bundle);
+    // todo: handle this better?
+    if (pxt.BrowserUtils.isLocalHostDev()) {
+        // patch webconfig to refer to pxt serve instead of multiplayer serve
+        const wc = pxt.webConfig as any;
+
+        for (const key of Object.keys(wc)) {
```
jwunderl, line 48: Now that I'm thinking of this we might be able to clear this out now with your localhost serving changes (or handle there)? have to test / see

jwunderl (approved):

kiosk/src/index.tsx
eanders-ms: Oh you're probably right. Let me try that.

PR #9688, opened 2023-09-19
localization for webusb connection failed msg

jwunderl:
re: https://github.com/microsoft/pxt-microbit/issues/5357

The extension labels would need to be extracted from `targetconfig.json` and uploaded at build time (along with an `rlf` if that isn't already there). We do similar things here https://github.com/microsoft/pxt/blob/master/cli/cli.ts#L1852 for gallery, so it'd be walking through all approved extensions, aggregating list of labels, and including in translation. Maybe worth considering packaging in translations for `targetConfig.multiplayer.games[?].title` as well (suggestion cards underneath multiplayer host screen), since those aren't translatable either right now

### DIFF
--- a/webapp/src/cmds.ts
+++ b/webapp/src/cmds.ts
@@ -479,11 +479,11 @@ export async function pairAsync(implicitlyCalled?: boolean): Promise<boolean> {
                 return true;
             } catch (e) {
                 // Device
-                core.infoNotification("Oops, connection failed.");
+                core.infoNotification(lf("Oops, connection failed."));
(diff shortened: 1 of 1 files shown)


srietkerk (approved):
LGTM! For the extensions label part, do we want to make a new issue or are we just going to continue to reference the issue that you linked?

jwunderl:
I'm not closing that one with this pr so probably fine to leave it on that one :) Normally I'd just fix it just timing with a vacation / other things to do / bleh

PR #9691, opened 2023-09-19
kiosk: refactor add game polling

eanders-ms:
* Moved game polling logic out of UI component in prep for port to react-common.
* Moved toast notifications to global space.
* Games can be added any time now, if:
    * A game is not being played
    * Kiosk code is valid


kiosk/src/Services/AddingGamesService.ts
```diff
@@ -0,0 +1,80 @@
...
+import { KioskState } from "../Types";
+import * as Actions from "../State/Actions";
+import * as Backend from "./BackendRequests";
+import { saveNewGamesAsync } from "../Transforms/saveNewGamesAsync";
+import { postNotification } from "../Transforms/postNotification";
+
+function initialize() {
+    const pollOnce = async () => {
+        const { state, dispatch } = stateAndDispatch();
+        if (state.kioskState === KioskState.PlayingGame) {
```
jwunderl, line 11: Might want to remove this case eventually, it'd feel a little bad if you played a game in the last few minutes and it missed a few uploads,

jwunderl (approved):
LGTM

kiosk/src/Services/AddingGamesService.ts
eanders-ms: I was on the fence about this. Will remove.

PR #9692, opened 2023-09-21
Kiosk: Remove direct dom manipulation

eanders-ms:
* Move simulator iframe to a React component
* Removed direct DOM manipulation
* Move footer into React, reduced bottom margin
* Some compensating css tweaks


kiosk/src/Components/PlayingGame.tsx
```diff
@@ -0,0 +1,26 @@
...
+export default function PlayingGame() {
+    const { state: kiosk, dispatch } = useContext(AppStateContext);
+    const { launchedGameId: gameId } = kiosk;
+
+    const playUrl = useMemo(() => {
+        if (gameId) {
+            const playUrlBase = `${configData.PlayUrlRoot}?id=${gameId}&hideSimButtons=1&noFooter=1&single=1&fullscreen=1&autofocus=1`;
+            const playQueryParam = kiosk.builtGamesCache[gameId]
+                ? "&server=1"
+                : "&sendBuilt=1";
```
srietkerk, line 15: I might be misunderstanding this, but would we want to change the order for this ternary operator? If there is a game in the `builtGamesCache`, we would want the `sendBuilt=1` flag, right?

kiosk/src/Components/PlayingGame.tsx
```diff
@@ -0,0 +1,26 @@
...
+            return playUrlBase + playQueryParam;
+        }
+    }, [gameId, kiosk.builtGamesCache]);
```
srietkerk, line 18: Can you explain why we want the `kiosk.builtGamesCache` to be a dependency for the useMemo?

eanders-ms: Good call. We probably don't. We wouldn't want to re-launch the game if it changed.

kiosk/src/Components/PlayingGame.tsx
eanders-ms: You might be correct! Code is unchanged from the original: https://github.com/microsoft/pxt-arcade/blob/0b1189cbc8abc015854cc940e63f6ae9dc192de4/kiosk/src/Models/Kiosk.ts#L343

eanders-ms: I'd prefer not to change it in this PR, but will file an issue to investigate.

kiosk/src/Components/PlayingGame.tsx
eanders-ms: Removed

kiosk/src/Components/PlayingGame.tsx
srietkerk: Oh, interesting.. Yeah, this might be a bug. If we don't see performance improvements for subsequent game launches, then the order is wrong. If we do see those load time improvements, then there's something that I'm reading incorrectly. Like you said, something to investigate.

srietkerk (approved):
Looks good!

PR #9694, opened 2023-09-26
Add snapshots and refactor history code

riknoll:
This adds snapshots to the version history, which are complete copies of the project (minus the autogenerated/internal files) that can be restored as a failsafe for any weird corruption that my occur due to classic webapp race conditions. Snapshots currently are taken every 15 minutes for the past 24 hours and once per day before that. I haven't done any measuring on what this does to project size just yet; next time I'm making a big arcade program I'll use this and see how huge it gets.

I also refactored all of the history stuff into its own file in pxteditor and added some end-to-end style tests.

I still have to add entries for shares from the project; that'll be the next PR.


webapp/src/timeMachine.tsx
```diff
@@ -145,82 +158,45 @@ export const TimeMachine = (props: TimeMachineProps) => {
...
 
     const onGoPressed = React.useCallback(() => {
-        if (selected === -1) {
-            hideDialog();
-        }
-        else {
-            const previewFiles = applyUntilTimestamp(text, history, selected);
-            onProjectLoad(previewFiles, history.find(e => e.timestamp === selected).editorVersion, selected);
-        }
+        (async () => {
```
jwunderl, line 180: I'm missing why the iife's here? shouldn't 

```typescript
const onGoPressed = React.useCallback(async () => {
```

be fine or is there some typing weirdness from there? Or does `React.useCallback` actually look at returned value and do something if promise-like?

jwunderl (approved):
LGTM

webapp/src/timeMachine.tsx
riknoll: oh... i guess that's true

riknoll: i'll fix it in the next pr

PR #9695, opened 2023-09-26
Adding project shares to the version history

riknoll:
These have a share icon next to them:

image

We already store shares in the project header, so share entries will actually appear for pre-time machine projects (albeit, all with the same timestamp).

### DIFF
--- a/pxtlib/emitter/cloud.ts
+++ b/pxtlib/emitter/cloud.ts
@@ -116,6 +116,15 @@ namespace pxt.Cloud {
         })
     }
 
+    export function downloadScriptMetaAsync(id: string): Promise<JsonScriptMeta> {
+        return privateRequestAsync({
+            url: id + (id.startsWith("S") ? `?time=${Date.now()}` : ""),
+            forceLiveEndpoint: true,
+        }).then(resp => {
+            return JSON.parse(resp.text).meta;
+        })
+    }
+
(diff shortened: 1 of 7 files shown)


pxtlib/emitter/cloud.ts
```diff
@@ -116,6 +116,15 @@ namespace pxt.Cloud {
         })
     }
 
+    export function downloadScriptMetaAsync(id: string): Promise<JsonScriptMeta> {
+        return privateRequestAsync({
+            url: id + (id.startsWith("S") ? `?time=${Date.now()}` : ""),
+            forceLiveEndpoint: true,
```
eanders-ms, line 122: This is a good flag to know about.

webapp/src/timeMachine.tsx
```diff
@@ -40,12 +40,13 @@ type FrameState = "loading" | "loaded" | "loading-project" | "loaded-project";
 export const TimeMachine = (props: TimeMachineProps) => {
     const { text, history, onProjectLoad, onProjectCopy } = props;
 
-    // -1 here is a standin for "now"
-    const [selected, setSelected] = React.useState<number>(-1);
+    // undefind here is a standin for "now"
```
eanders-ms, line 43:
```suggestion
    // undefined here is a standin for "now"
```

eanders-ms (approved):
LGTM

jwunderl (approved):

PR #9696, opened 2023-09-26
Pull hw options from target config

jwunderl:
needs https://github.com/microsoft/pxt-arcade/pull/6089, we want to remove dependency on that hardware page and move live content to be rooted from targetconfig

fwiw I removed the `reportError` case as that seems like a valid use case when we have `experimentalHardware`; can add it back if we want


srietkerk (approved):
Nice!

abchatra (approved):

PR #9698, opened 2023-09-27
extract localization for multiplayer hero games & tags

jwunderl:
re: https://github.com/microsoft/pxt/pull/9688, should fix https://github.com/microsoft/pxt-microbit/issues/5357 (the tags bit)

I'll merge https://github.com/microsoft/pxt/pull/9696 first & then add in localization for descriptions of hardware options, but I'm not certain whether we want names since they're actually products? Will check in stand up tomorrow & update this pr. Also need to check I formatted it correctly for multiplayer, just gotta spin up multiplayer serve tomorrow

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -1849,7 +1849,7 @@ function saveThemeJson(cfg: pxt.TargetBundle, localDir?: boolean, packaged?: boo
     walkDocs(theme.docMenu);
     if (nodeutil.fileExistsSync("targetconfig.json")) {
         const targetConfig = nodeutil.readJson("targetconfig.json") as pxt.TargetConfig;
-        if (targetConfig && targetConfig.galleries) {
+        if (targetConfig?.galleries) {
             const docsRoot = nodeutil.targetDir;
             let gcards: pxt.CodeCard[] = [];
             let tocmd: string =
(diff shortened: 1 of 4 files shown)


cli/cli.ts
```diff
@@ -1900,6 +1900,32 @@ ${gcards.map(gcard => `[${gcard.name}](${gcard.url})`).join(',\n')}
 
 `, { encoding: "utf8" });
         }
+        const multiplayerGames = targetConfig?.multiplayer?.games;
+        for (const game of (multiplayerGames ?? [])) {
+            if (game.title) targetStrings[`{id:game-title}${game.title}`] = game.title;
```
srietkerk, line 1905: I'm a bit confused by the template literal. Is the entry in `targetStrings` going to look like `{id:game-title}hello-world: hello-world`?

jwunderl: Taking "perfect fit" as an example [from here](https://github.com/microsoft/pxt-arcade/blob/master/targetconfig.json#L230), this would set the entry in the json blob as `"{id:game-title}Perfect Fit": "Perfect Fit"`, so that the translator has context on what the string is referring to (e.g. that it is a game-title and not an indication of how things snap together or something). This gets stripped off before rendering via https://github.com/microsoft/pxt/blob/master/pxtlib/commonutil.ts#L214, and shows up in crowdin something like this: 
image
from `field_note.ts`

srietkerk: Oh, interesting! Thanks for the context!

srietkerk (approved):
LGTM

PR #9700, opened 2023-09-28
Kiosk: Support event-based navigation

eanders-ms:
This PR is huge, I apologize. It contains two main changesets: the switch to **event-based navigation** and some **aesthetic changes** to align more closely to Claire's design.

**Event-based navigation**
Switching to event-based navigation puts Kiosk where it needs to be in order to integrate react-common and gain the accessibility benefits to be had there. I introduced three main services to managed multi-modal input (tabbing, mousing, and gamepad(ing?)):
* **GamepadManager** - modified the existing GamepadManager to emit events from gamepad input.
* **NavGrid** - Figures which DOM element to focus based on directional navigation (gamepad or keyboard).
* **RectCache** - NavGrid likes to query elements repeatedly for their DOMRect, so it made sense to cache them. This is purely for perf.

With these services in place, I reimplemented how UI components are navigated. No more direct polling of gamepad state to switch focus. Events emitted from gamepad activity will direct NavGrid to move focus in that direction. Components can hint NavGrid how they want to be treated when navigating, like specifying valid exit directions, or indicating they can be automatically focused. Components can override NavGrid if the generic approach just doesn't work for the use case. This is the case with the 3rd party carousel component we're using (swiper). It has it's own internal navigation system.

As part of this work I had to reimplement modal dialogs. To not stray too far from the goal of moving closer to compatibility with react-common, I stole react-common's `FocusTrap` and `Modal` components. Once react-common is pulled in, these will go away.

**Aesthetic changes**
I also made some changes to Kiosk's look & feel to align with Claire's design, and unfortunately included them in this PR.
~~If nobody is too opposed, I'd like to keep them here, and get them out to beta for group testing and feedback. `pxt uploadtrg` doesn't work with webapps at the moment, so I'm unsure of another way to easily publish these changes for testing.~~ 
**Link to test target in comments below!** Please give it a test and let me know what you love/hate.
image
image

**Loose ends**
- There's a bug in the carousel with tab-based navigation when navigating away from the delete game button where the carousel gets all messed up. Seems like a bug in the control, but need to investigate it more.
- There's a bug in the carousel with swipe-based navigation that results in the selected game mismatching what appears to be the current game in the carousel.
- I'd like to avoid rewriting the carousel, but it's sort of on my chopping block at this point. I wrestled with it a lot.


eanders-ms:
I was wrong about uploadtrg not supporting webapps! Here is a test build:
https://arcade.makecode.com/app/e65a1da46ae1b3ef5f92dc000c36783fa5989d28-454ef2e16d--kiosk

kiosk/src/Kiosk.css
```diff
@@ -197,15 +289,12 @@ h2 {
...
     line-height: 200%;
 }
 
 .addGame h1 {
     font-size: 2.25rem;
     font-weight: 400;
-    margin: 1em 0;
 }
 
 .addGame {
```
eanders-ms, line 300: Lost `justify-content: center;` here. Readding.

eanders-ms:
New test target: https://arcade.makecode.com/app/f3c1c62da85f595eabd552826c0a25438da78ed2-3556e555b6--kiosk

kiosk/src/Components/Background.tsx
```diff
@@ -0,0 +1,100 @@
...
+                }
+            }
+        }, 750);
+    }, [selectedGame, kiosk.kioskState, myRef]);
+
+    const className = useMemo(() => {
+        if (kiosk.kioskState === KioskState.PlayingGame) {
+            return "background-flat";
+        } else {
+            return "background-acrylic";
```
srietkerk, line 93: I really like this addition. It adds a really nice polish to kiosk!!

kiosk/src/Components/GameList.tsx
```diff
@@ -1,133 +1,119 @@
...
-    addButtonSelected,
-    deleteButtonSelected,
-}) => {
+const GameList: React.FC<IProps> = ({}) => {
     const { state: kiosk } = useContext(AppStateContext);
-    const buttonSelected = addButtonSelected || deleteButtonSelected;
     const localSwiper = useRef<SwiperClass>();
+    const [userInitiatedTransition, setUserInitiatedTransition] =
+        React.useState(false);
+    const [pageInited, setPageInited] = React.useState(false);
```
srietkerk, line 27: Not quite sure what "inited" is.. Is this to check initialization ?

srietkerk (commented):
I haven't made it through everything yet, but I wanted to get these comments out so I don't lose those thoughts.

kiosk/src/Components/AddingGame.tsx
```diff
@@ -127,13 +111,12 @@ const AddingGame: React.FC<IProps> = ({}) => {
                         </li>
                     </ol>
                 </div>
-
-                <div className="QRCodeHolder">{qrDivContent()}</div>
+                <div style={{ flexGrow: 1 }} />
```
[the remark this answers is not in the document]
eanders-ms: The new div adds a spacer between the instructions and the QR code that will grow to fill the available space, pushing them apart as the viewport gets wider and wider. I'm sure there's a pure CSS way to accomplish this, but this is what came to mind first.

kiosk/src/Components/Background.tsx
```diff
@@ -0,0 +1,100 @@
...
+                if (dataUrlCache.current.has(selectedGame.id)) {
+                    if (myRef) {
+                        myRef.style.backgroundImage = `url(${dataUrlCache.current.get(
+                            selectedGame.id
+                        )})`;
+                    }
+                    return;
+                }
+                image.current.crossOrigin = "anonymous";
+                image.current.src = `https://makecode.com/api/${selectedGame.id}/thumb`;
```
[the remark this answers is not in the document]
eanders-ms: I don't see a `thumbnail` member on `GameData`, but I get your meaning. Caching fetched thumbnails is a great idea.

`dataUrlCache` is a little different from a thumbnail cache. It caches a 80x60 rendering of the thumbnail, or the first frame of the thumbnail if it's a gif, as a base-64 encoded png. I have a TODO to apply the acrylic effect to this data rather than having CSS do it. So, `dataUrlCache` is (or will be) a cache of low-res, blurry, dark versions of thumbnails.

kiosk/src/Components/Background.tsx
```diff
@@ -0,0 +1,100 @@
...
+                image.current.onload = () => {
+                    canvas.current.width = 80;
+                    canvas.current.height = 60;
+                    const ctx = canvas.current.getContext("2d");
+                    if (ctx) {
+                        // TODO: Apply blur and opacity filters to image here, not in CSS
```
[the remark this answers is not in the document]
eanders-ms: CSS filters suffer performance issues on low-end hardware. They need a GPU to perform well. Right now, every time we switch the background, CSS reapplies the acrylic effect to the image. I want to avoid this by applying it once and caching that result.

kiosk/src/Components/GameList.tsx
eanders-ms: This is me fighting with Swiper for who gets to decide which game is selected on component mount. I don't like it. This carousel is feature rich, but surprisingly difficult to control.

kiosk/src/Components/AppModal.tsx
```diff
@@ -0,0 +1,64 @@
...
+    };
+
+    const selectedGame = state.allGames.find(
+        g => g.id === state.selectedGameId
+    );
+
+    const deleteGame = () => {
+        const userAddedGames = Storage.getUserAddedGames();
+        const gameId = state.selectedGameId;
+        if (gameId && gameId in userAddedGames) {
```
jwunderl, line 32: typically prefer to avoid `in` as it's got confusing / mildly unexpected semantics at times (`"toString" in {} === true`) -- prefer using `userAddedGames.hasOwnProperty(gameId)` for that reason but not a big deal either way

kiosk/src/Services/BackendRequests.ts
```diff
@@ -1,47 +1,55 @@
...
+                throw new Error("No code returned from the request.");
+            }
         }
+    } catch (error) {
+        console.error(error);
     }
 };
 
-export const addGameToKioskAsync = async (
+export const canthrow_addGameToKioskAsync = async (
```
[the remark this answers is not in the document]
eanders-ms: I made a pass to standardize how backend request failures are handled, but this one remains an "exception" to the pattern (harhar).

PR #9711, opened 2023-10-09
Fix the home screen's project row card size on mobile

riknoll:
Fixes https://github.com/microsoft/pxt-minecraft/issues/2243

I also had to make the (+) icon smaller on the new project card and nudge the blocks file image 4 pixels to the left because the svg for it has margins baked in

image


kimprice:
Looks good, but what about the issue when scaling up to full screen?

riknoll:
@kimprice that issue doesn't repro for me...

kimprice:
> @kimprice that issue doesn't repro for me...

hmm me neither actually

kimprice (approved):

PR #9712, opened 2023-10-09
kiosk: support gamepad dpad for navigation

eanders-ms:
Also made tweaks to control focus logic for more reliable behavior on Xbox.

### DIFF
--- a/kiosk/src/Components/GameList.tsx
+++ b/kiosk/src/Components/GameList.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, useContext, useCallback } from "react";
+import React, { useEffect, useState, useContext, useCallback } from "react";
 import { Swiper, SwiperSlide } from "swiper/react";
 import { Swiper as SwiperClass, Pagination } from "swiper";
 import "swiper/css";
@@ -8,11 +8,7 @@ import { playSoundEffect } from "../Services/SoundEffectService";
 import { AppStateContext } from "../State/AppStateContext";
 import { selectGameByIndex } from "../Transforms/selectGameByIndex";
 import { launchGame } from "../Transforms/launchGame";
-import {
-    getHighScores,
-    getSelectedGameIndex,
-    getSelectedGameId,
-} from "../State";
+import { getSelectedGameIndex, getSelectedGameId } from "../State";
 import * as GamepadManager from "../Services/GamepadManager";
 import * as NavGrid from "../Services/NavGrid";
 import { useOnControlPress, useMakeNavigable } from "../Hooks";
@@ -21,12 +17,18 @@ interface IProps {}
 
 const GameList: React.FC<IProps> = ({}) => {
     const { state: kiosk } = useContext(AppStateContext);
-    const localSwiper = useRef<SwiperClass>();
+    const [localSwiper, setLocalSwiper] = useState<SwiperClass | undefined>(
+        undefined
+    );
     const [userInitiatedTransition, setUserInitiatedTransition] =
         React.useState(false);
     const [pageInited, setPageInited] = React.useState(false);
 
-    useMakeNavigable(localSwiper?.current?.el, {
+    const handleLocalSwiper = (swiper: SwiperClass) => {
+        setLocalSwiper(swiper);
+    };
+
+    useMakeNavigable(localSwiper?.el, {
         exitDirections: [NavGrid.NavDirection.Down, NavGrid.NavDirection.Up],
         autofocus: true,
     });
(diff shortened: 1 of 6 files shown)


kiosk/src/Components/GameList.tsx
```diff
@@ -95,12 +101,12 @@ const GameList: React.FC<IProps> = ({}) => {
...
         () => {
-            if (NavGrid.isActiveElement(localSwiper?.current?.el)) {
+            if (!localSwiper || localSwiper.destroyed) return;
+            if (NavGrid.isActiveElement(localSwiper.el)) {
                 launchSelectedGame();
             }
         },
-        GamepadManager.GamepadControl.AButton,
-        GamepadManager.GamepadControl.BButton
+        GamepadManager.GamepadControl.AButton
```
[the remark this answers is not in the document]
kimprice: Oh, is this just setting the button for scrolling through the games?

kiosk/src/Services/GamepadManager.ts
```diff
@@ -258,6 +265,15 @@ class GamepadManager {
         }
     }
 
+    mergeControlValues(...values: ControlValue[]): ControlValue {
+        for (const value of values) {
+            if (value === ControlValue.Down) {
+                return ControlValue.Down;
+            }
+        }
+        return ControlValue.Up;
```
kimprice, line 274: I'm a little confused. Do the left and right controls become an up control?

kiosk/src/Components/GameList.tsx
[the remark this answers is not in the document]
eanders-ms: That's correct, this is the code that maps which button (or buttons) launch the selected game. This is a spot where the expected behavior of keyboard and gamepad differ. With a keyboard, you would typically expect that pressing `Enter` would launch the game. But, in kiosk's setup, both the keyboard's `Enter` key and the gamepad's `B` button are mapped to the same functionality: they both represent the simulator's `B` button. In other areas of Kiosk, `B` is used to navigate back within the UI, like closing a dialog. It felt like a bug to be able to launch a game with the gamepad's `B` button. For that reason, and for consistency with the "navigate back" behavior elsewhere, I decided to remove it here.

kiosk/src/Services/GamepadManager.ts
eanders-ms: This code just returns `Down` if any of the passed in values are `Down`. Otherwise it returns `Up`. It's a helper function that is useful when multiple gamepad buttons are mapped to the same control. For example, both left-stick-left and dpad-left represent the `DPadLeft` control. It either one are active, we want to represent the `DPadLeft` control as being `Down`.

kiosk/src/Utils/domUtils.ts
```diff
@@ -43,3 +43,9 @@ export const isInteractable = (
         isElementValid(element)
     );
 };
+
+export const focusElement = (el: HTMLElement | null | undefined): void => {
+    setTimeout(() => {
+        el?.focus({ focusVisible: true, preventScroll: true } as any);
+    }, 0);
```
kimprice, line 50: Why is the setTimeout needed here?

kiosk/src/Services/GamepadManager.ts
```diff
@@ -497,15 +513,11 @@ class GamepadManager {
...
-        if (axisIndex < 0 || axisIndex >= gamepad.axes.length) {
-            throw new Error(
-                `Gamepad at index ${gamepad.index} does not have an axis at index ${axisIndex}`
-            );
-        }
+        const axisConf = configData.GamepadAxes[gamepadControlToAxis[control]];
+        const axisIndex = axisConf.Pin;
+        if (axisIndex < 0 || axisIndex >= gamepad.axes.length)
+            return ControlValue.Up;
+        const threshold = axisConf.Sign * axisConf.Threshold;
```
kimprice, line 520: What does threshold mean?

kiosk/src/Utils/domUtils.ts
eanders-ms: It probably isn't needed. I was struggling with getting focus to work reliably on Xbox, but turns out the problem was elsewhere. This was a workaround I found for a similar problem seen in earlier browsers like IE. I will make a note to remove it.

kiosk/src/Services/GamepadManager.ts
eanders-ms: This value represents the threshold at which a gamepad axis is considered active. Gamepad sticks, even when not touched, rarely report (0,0) as their position due to inherent noise. They're always changing and reporting small, off-center values, like (0.01, -0.003). Furthermore, small stick movements report erratic positions and aren't reliable. The threshold allows us to filter out this noise and only respond to meaningful movements.

kimprice (approved):
Looks good! Thanks for answering all my questions! :)

PR #9713, opened 2023-10-10
Update the 'Identity' and 'share' base pages

ganicke:
An update for the Identity and Share pages. Images are sourced from the target's static asset path for target specific context.

RE: https://github.com/microsoft/pxt-microbit/issues/5366 and https://github.com/microsoft/pxt-arcade/issues/6105


Jaqster (approved):
Thanks Galen!

PR #9716, opened 2023-10-10
Prevent simulator start on the home screen

riknoll:
Fixes https://github.com/microsoft/pxt-minecraft/issues/2358

Another self explanatory fix! I also async/await-ified the code while i was in there.


thsparks (approved):
Nice :)

PR #9717, opened 2023-10-12
kiosk: setup react-common dependency

eanders-ms:
Swap React 18 for react-common, and concomitantly, React 17.

I'm very surprised that this "just worked".


srietkerk (approved):
Exciting stuff! Nice!

PR #9718, opened 2023-10-12
Force version history to update when restoring a version or sharing

riknoll:
Fixes https://github.com/microsoft/pxt-minecraft/issues/2356

Three version history fixes:

1. Filter the history file out of the file explorer and github diffs
2. Explicitly save a history snapshot before restoring an old version of the project so that we definitely save the old version of the project
3. Save history whenever we do a share so that we don't have to wait for the user to make a change before it shows up in the version history

### DIFF
--- a/pxteditor/history.ts
+++ b/pxteditor/history.ts
@@ -375,6 +375,51 @@ namespace pxt.workspace {
         toWrite[pxt.HISTORY_FILE] = JSON.stringify(history);
     }
 
+    export function pushSnapshotOnHistory(text: ScriptText, currentTime: number) {
+        let history: pxt.workspace.HistoryFile;
+
+        if (text[pxt.HISTORY_FILE]) {
+            history = pxt.workspace.parseHistoryFile(text[pxt.HISTORY_FILE]);
+        }
+        else {
+            history = {
+                entries: [],
+                snapshots: [],
+                shares: []
+            };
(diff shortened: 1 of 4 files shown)


pxteditor/history.ts
```diff
@@ -375,6 +375,51 @@ namespace pxt.workspace {
         toWrite[pxt.HISTORY_FILE] = JSON.stringify(history);
     }
 
+    export function pushSnapshotOnHistory(text: ScriptText, currentTime: number) {
```
thsparks, line 378: Where is this used?

thsparks (approved):
One question but otherwise LGTM!

pxteditor/history.ts
```diff
@@ -375,6 +375,51 @@ namespace pxt.workspace {
...
+        let history: pxt.workspace.HistoryFile;
```
srietkerk, line 379: Not super necessary, and there might also be something that I'm missing, but have you considered putting this `let history .. if .. else.. ` chunk in a separate function to limit repeated code?

riknoll: it's slightly different in each function, sadly

pxteditor/history.ts
riknoll: in saveSnapshotAsync in workspace.ts

PR #9719, opened 2023-10-12
react-common: support for button children

eanders-ms:
This is to ease porting Kiosk to react common. Kiosk displays button content as JSX children.

### DIFF
--- a/react-common/components/controls/Button.tsx
+++ b/react-common/components/controls/Button.tsx
@@ -1,7 +1,7 @@
 import * as React from "react";
-import { classList, ControlProps, fireClickOnEnter } from "../util";
+import { classList, ContainerProps, fireClickOnEnter } from "../util";
 
-export interface ButtonViewProps extends ControlProps {
+export interface ButtonViewProps extends ContainerProps {
     buttonRef?: (ref: HTMLButtonElement) => void;
     title: string;
     label?: string | JSX.Element;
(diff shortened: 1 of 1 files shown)


react-common/components/controls/Button.tsx
```diff
@@ -101,13 +102,15 @@ export const Button = (props: ButtonProps) => {
...
             aria-describedby={ariaDescribedBy}
             aria-selected={ariaSelected}>
-                <span className="common-button-flex">
-                    {leftIcon && <i className={leftIcon} aria-hidden={true}/>}
-                    <span className="common-button-label">
-                        {label}
-                    </span>
-                    {rightIcon && <i className={"right " + rightIcon} aria-hidden={true}/>}
-                </span>
+                {(leftIcon || rightIcon || label) && (
```
srietkerk, line 105: I'm not sure why this is needed. Since we check for the left and right icon within the span, the check before the span feels a bit much since the label logic hasn't changed.

eanders-ms: I can embed the label check.

eanders-ms: Oh I recall now, this was to avoid rendering the outer "common-button-flex" div container if none of the inner components would render.

srietkerk: I see. So there's a possibility that label could be nothing?

kimprice (approved):
LGTM!

react-common/components/controls/Button.tsx
eanders-ms: Yes, label is an optional prop.

PR #9720, opened 2023-10-13
kiosk: use button control from react common

eanders-ms:
Swap out kiosk's button control for react-common's.

### DIFF
--- a/kiosk/src/Components/AddingGame.tsx
+++ b/kiosk/src/Components/AddingGame.tsx
@@ -73,14 +73,14 @@ const AddingGame: React.FC<IProps> = ({}) => {
                     <h1 className="kioskCode">{kiosk.kioskCode}</h1>
                     <QRCodeSVG value={kioskUrl} />
                     <div className="kioskLink">
-                        <a
+                        <GenericButton
+                            title={kioskUrl}
+                            label={kioskUrl}
                             target="_blank"
-                            rel="noopener noreferrer"
                             onClick={kioskLinkClicked}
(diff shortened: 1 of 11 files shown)


kiosk/src/Components/GameMenu.tsx
```diff
@@ -30,6 +28,7 @@ export const GameMenu: React.FC<IProps> = () => {
         <div className="gameMenu">
             {game?.userAdded && !kiosk.locked && (
                 <GenericButton
+                    title={lf("Delete Game")}
```
kimprice, line 31: Do you need a label here?

kimprice (approved):
Other than that missing label, looks good!

kiosk/src/Components/GameMenu.tsx
eanders-ms: In this case button content is provided by the nested svg and other child elements, so we don't want to send a label.

kiosk/src/Components/AddingGame.tsx
```diff
@@ -112,8 +114,12 @@ const AddingGame: React.FC<IProps> = ({}) => {
                 <div style={{ flexGrow: 1 }} />
                 {qrDivContent()}
             </div>
-            <GenericButton autofocus={true} onClick={gotoMainMenu}>
-                {lf("Return to menu")}
+            <GenericButton
```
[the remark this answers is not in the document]
eanders-ms: yep!

PR #9721, opened 2023-10-13
Fix indices check for pop operation in Python

kimprice:
fixes https://github.com/microsoft/pxt-minecraft/issues/2024

We were throwing an Index Error when trying to pop the 0th element because our check didn't include 0 as a valid index.


kimprice:
Whoops, accidentally linked the above.

PR #9729, opened 2023-10-19
Scale Tutorial Text Alongside Workspace

thsparks:
Fixes https://github.com/microsoft/pxt-minecraft/issues/2050 by making tutorial text size increase when the workspace is zoomed in.

I went through a few different approaches to this. Initially, I hooked into the zoom button events directly, but you could get out of sync if you used "ctrl + mouse wheel" to zoom the workspace, so I decided to track change-in-scale events instead.

I also opted to scale the text relative to zoom % of the workspace (as opposed to increasing/decreasing by a fixed amount on every zoom in/out). I did this because different ways of zooming (mouse wheel vs workspace buttons vs "ctrl +/-" in monaco) would zoom in different amounts, and it looked strange to always use a fixed change to the tutorial-text size regardless. It also led to some unexpected behavior if you zoomed in with one technique but out with another.

Lastly, I prevent the tutorial text from ever shrinking beyond its initial size, just because it gets very hard to read and looks a little silly when that happens. I'm not sure there's a realistic use case for it.

Upload target: https://minecraft.makecode.com/app/bfd4e1e83bcc22c2b26d7062d853a80e4555fb59-9dbaf7a159

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -211,6 +211,7 @@ export class ProjectView
         this.exitTutorial = this.exitTutorial.bind(this);
         this.setEditorOffset = this.setEditorOffset.bind(this);
         this.resetTutorialTemplateCode = this.resetTutorialTemplateCode.bind(this);
+        this.onScaleChanged = this.onScaleChanged.bind(this);
         this.initSimulatorMessageHandlers();
 
         // add user hint IDs and callback to hint manager
@@ -1029,6 +1030,8 @@ export class ProjectView
         }
         this.allEditors = [this.pxtJsonEditor, this.gitjsonEditor, this.blocksEditor, this.serialEditor, this.assetEditor, this.textEditor]
         this.allEditors.forEach(e => e.changeCallback = changeHandler)
+        this.allEditors.forEach(e => e.onScaleChanged = this.onScaleChanged)
+
         this.editor = this.allEditors[this.allEditors.length - 1]
     }
 
@@ -1529,6 +1532,20 @@ export class ProjectView
                         break;
                 }
                 break;
+        }
+    }
+
+    zoomIncrement = 0.25; // TODO thsparks : move up.
+    initialFontSize = 1.125; // TODO thsparks : move up. Can this reference the css?
+
+    onScaleChanged(oldScale: number, newScale: number) {
+        if (this.isTutorial && oldScale !== newScale) {
+            const change = newScale > oldScale ? this.zoomIncrement : -this.zoomIncrement;
+            if (!this.state.tutorialFontSize) {
+                this.setState({ tutorialFontSize: this.initialFontSize + change });
+            } else {
+                this.setState({ tutorialFontSize:  this.state.tutorialFontSize + change });
+            }
         }
     }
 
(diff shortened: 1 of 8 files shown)


webapp/src/app.tsx
```diff
@@ -1532,6 +1542,26 @@ export class ProjectView
...
+
+            const root = document.getElementById("root");
+            if (root && newScale <= this.initialEditorScale) {
+                // Do not shrink the text beyond its initial size.
+                root.style.setProperty("--tutorialFontSize", `${this.tutorialInitialFontSize}rem`);
+            } else if (root) {
+                // Increase font size to match the editor's % zoom.
+                const maxEditorScale = this.editor.getMaxScale();
+                const zoomAmt = (newScale - this.initialEditorScale) / (maxEditorScale - this.initialEditorScale);
+                const newTutorialFontSize = this.tutorialInitialFontSize + (this.tutorialMaxFontSize - this.tutorialInitialFontSize) * zoomAmt;
```
srietkerk, line 1559: Can you explain this math a bit? I'm having a hard time understanding why we want to multiply zoom amount with the difference of `tutorialMaxFontSize` and `tutorialInitialFontSize`.

webapp/src/blocks.tsx
```diff
@@ -636,6 +640,15 @@ export class Editor extends toolboxeditor.ToolboxEditor {
                     }
                 }
             }
+            else if (ev.type === Blockly.Events.VIEWPORT_CHANGE) {
+                const viewportChangeEvent = ev as Blockly.Events.ViewportChange;
+                if (this.initialScaleSet && viewportChangeEvent.oldScale !== viewportChangeEvent.scale) {
+                    if (this.onScaleChanged) {
```
srietkerk, line 646: Is it possible that the function doesn't get initialized?

webapp/src/app.tsx
srietkerk: Yeah, this explanation was extremely helpful! Thank you for that clarification!!

srietkerk (approved):
Nice!

webapp/src/app.tsx
```diff
@@ -1108,6 +1116,8 @@ export class ProjectView
         // subscribe to user preference changes (for simulator or non-render subscriptions)
         data.subscribe(this.highContrastSubscriber, auth.HIGHCONTRAST);
         data.subscribe(this.cloudStatusSubscriber, `${cloud.HEADER_CLOUDSTATE}:*`);
+
+        document.getElementById("root").style.setProperty("--tutorialFontSize", `${this.tutorialInitialFontSize}rem`);
```
riknoll, line 1120: i would define this in css instead of adding it here and just override it when the scale changes. that way it'll always have at least some value... anything set on the root directly will override whatever is in the css styles.

webapp/src/app.tsx
```diff
@@ -1532,6 +1542,26 @@ export class ProjectView
         }
     }
 
+    onScaleChanged(oldScale: number, newScale: number) {
```
riknoll, line 1545: passing in the oldScale here to set the initial scale seems kinda fragile. is the initial scale not always the same value? and failing that, can't we just read it from blockly when we load the workspace?

riknoll:
not sure if this should be behind a target flag or not. are we sure we'll want this in all targets?

webapp/src/app.tsx
thsparks: Yeah I went back and forth on this. Decided to put it here to keep it all in once place, but happy to move it back to css.

webapp/src/app.tsx
thsparks: It's different between Monaco and Blockly. I'll see about reading it when we load the workspace. Should be do-able.

thsparks:
Putting this on hold. Sounds like we probably don't want this - or at the very least, we may only want to scope it to HoC. But even then, browser zoom would be preferable.

PR #9730, opened 2023-10-19
tweak toggle simulator settings menu option in headless mode

riknoll:
Fixes https://github.com/microsoft/pxt-minecraft/issues/2366
Fixes https://github.com/microsoft/pxt-minecraft/issues/2325

Tweaks the simulator collapse toggle when in headless mode so that:

1. It doesn't show up in blocks
2. In JavaScript/Python, it says "Toggle the File Explorer" rather than "Toggle the Simulator"
3. When switching back to blocks the simulator pane automatically collapses if toggled on

This only affects headless mode which is only used by Minecraft.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -654,6 +654,10 @@ export class ProjectView
             return;
         }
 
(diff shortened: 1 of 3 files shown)


thsparks (approved):
LGTM. Thanks!

PR #9731, opened 2023-10-20
kiosk: fixes for carousel touch nav, storage exceptions

eanders-ms:
* The compiled js for games can exceed local storage quota. Added protection for that.
* Fixed a bug where carousel touch/scroll could result in app state confusion around which game is the currently selected game.


eanders-ms:
Update: I disabled saving of compiled js to local storage. We were hitting storage quotas with only a few games. A few options:
* LZ compress the games. They're text and probably very compressible
* Save to indexdb instead. indexdb allows for much larger blobs.

PR #9733, opened 2023-10-20
Fixing a couple of forced reflows in our tutorial components

riknoll:
Fixes two perf issues for tutorials:

First, the tutorial callout was running it's resize observer even when not visible. The callback for the observer was checking the height of an element, which forces the browser to rerender elements to get an accurate number. I guarded this behind a check to see if the callout is visible and also cleaned up the event listeners in that component a little bit while I was there.

Second, the tutorial container was running a querySelector to get a reference to its container so that it could run a resize observer on it. Query selectors also force reflows, so I changed this to instead use a ref of the container element.

Finally, I fixed a small CSS bug that I noticed when resizing tutorials where the editor sidebar wasn't taking up the whole height of the editor.

### DIFF
--- a/theme/tutorial-sidebar.less
+++ b/theme/tutorial-sidebar.less
@@ -520,8 +520,8 @@
         min-width: inherit;
         max-width: inherit;
         width: @simulatorWidth;
-        height: calc(~'100%' - @mainMenuHeight);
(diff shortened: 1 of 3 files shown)


thsparks (approved):
Nice, ty!

PR #9734, opened 2023-10-22
kiosk: fix for skipped letters when entering high score

eanders-ms:
Fixes https://github.com/microsoft/pxt-arcade/issues/6134

### DIFF
--- a/kiosk/src/Components/HighScoreInitial.tsx
+++ b/kiosk/src/Components/HighScoreInitial.tsx
@@ -21,7 +21,7 @@ const HighScoreInitial: React.FC<IProps> = ({
         (index + 1) % configData.HighScoreInitialAllowedCharacters.length;
 
     const previousInitial = () => {
-                const newIndex = getPreviousIndex();
+        const newIndex = getPreviousIndex();
         setIndex(newIndex);
         onCharacterChanged(
             configData.HighScoreInitialAllowedCharacters[newIndex]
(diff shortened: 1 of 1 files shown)


kiosk/src/Components/HighScoreInitial.tsx
```diff
@@ -38,14 +38,14 @@ const HighScoreInitial: React.FC<IProps> = ({
 
     // Handle DPadUp button press
     useOnControlPress(
-        [onCharacterChanged, isSelected, index],
+        [onCharacterChanged, isSelected, setIndex],
```
srietkerk, line 41: I don't see how this change will fix the double skip issue. Wouldn't we want the `useEffect` dependency to be `index`? Is it possible that it's skipping two because of the fact that in `Strict` mode, React will run `useEffect` twice?

eanders-ms: I see your point about React strict mode though. In this case it's fine. The underlying useEffect can be called many times without issue. It has no side effects.

srietkerk: Ah, that makes sense. So should we take `setIndex` out or is it a good reference to keep in to say "this will cause the index to change"?

eanders-ms: Because it is referenced, I'd prefer to keep it in. I think an eslint exhaustive deps check would complain otherwise. I'll take a todo here is to figure out why the inclusion of index is causing side effects.

srietkerk (approved):
Looks good!

PR #9735, opened 2023-10-22
kiosk: remap esc to the controller's back button

eanders-ms:
Exit a game using the controller's back button (view button on newer controllers), not the start button (menu button on newer controllers). Start/view is used by the simulator to open the in-game settings menu.

Fixes https://github.com/microsoft/pxt-arcade/issues/6145

### DIFF
--- a/kiosk/src/Components/PlayingGame.tsx
+++ b/kiosk/src/Components/PlayingGame.tsx
@@ -19,8 +19,7 @@ export default function PlayingGame() {
             playSoundEffect("select");
             escapeGame();
         },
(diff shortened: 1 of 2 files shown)


thsparks (approved):
Looks okay to me! We had also discussed making the Y button function as a back button - is that still something we're considering, or did you decide against it?

eanders-ms:
> Looks okay to me! We had also discussed making the Y button function as a back button - is that still something we're considering, or did you decide against it?

Oh yes! I will handle this in a separate PR. Thanks for the reminder.

PR #9736, opened 2023-10-22
kiosk: store builtjs in indexeddb, not local storage

eanders-ms:
We were hitting quota limits in local storage. Indexed db is much more forgiving.

### DIFF
--- a/kiosk/package-lock.json
+++ b/kiosk/package-lock.json
@@ -11,6 +11,7 @@
         "@types/node": "^16.11.33",
         "howler": "^2.2.3",
         "html5-qrcode": "^2.2.6",
+        "idb": "^7.1.1",
         "nanoid": "^4.0.2",
         "qrcode.react": "^3.1.0",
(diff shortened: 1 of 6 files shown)


srietkerk:
Just a quick question. Do we know that the built games are causing the most problems for hitting the quota? I know that user-added games can get long.

eanders-ms:
> Just a quick question. Do we know that the built games are causing the most problems for hitting the quota? I know that user-added games can get long.

I am certain the built games were the cause of the quota exceptions I saw, but it's a good idea to store anything of any appreciable size in index db. I'll file an issue for user-added games.

eanders-ms:
> Just a quick question. Do we know that the built games are causing the most problems for hitting the quota? I know that user-added games can get long.

Actually, it should be fine to store user-added games in local storage. We're not storing the content of the games, just metadata. You'd have to upload many thousands of games to approach the quota.

srietkerk (approved):

PR #9737, opened 2023-10-23
kiosk: fix for shoebox controller

eanders-ms:
I also bound the reset button to exit game. There isn't a way to do this using only the shoebox controller otherwise.

Fixes https://github.com/microsoft/pxt-arcade/issues/6143

### DIFF
--- a/kiosk/src/Services/GamepadManager.ts
+++ b/kiosk/src/Services/GamepadManager.ts
@@ -242,8 +242,7 @@ class GamepadManager {
             gamepad.axes &&
             gamepad.axes.length >= this.minAxisRequired &&
             gamepad.buttons &&
-            gamepad.buttons.length >= this.minButtonPinRequired &&
-            !!gamepad.mapping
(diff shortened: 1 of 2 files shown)


kiosk/src/Services/GamepadManager.ts
```diff
@@ -242,8 +242,7 @@ class GamepadManager {
             gamepad.axes &&
             gamepad.axes.length >= this.minAxisRequired &&
             gamepad.buttons &&
-            gamepad.buttons.length >= this.minButtonPinRequired &&
-            !!gamepad.mapping
```
srietkerk, line 246: Why can this check get taken away?

eanders-ms: I borrowed this condition from gamepad filtering code I found in an internal React library developed by the Xbox team. I wasn't entirely sure of its purpose, and turns out doesn't suit our needs because it filters out shoebox controllers.

PR #9738, opened 2023-10-23
kiosk: set color of "loading" text in run.html

eanders-ms:
.. and added support setting color of "loading" text in run.html

Fixes https://github.com/microsoft/pxt-arcade/issues/5407
Fixes https://github.com/microsoft/pxt-arcade/issues/6137


srietkerk (approved):
LGTM

PR #9742, opened 2023-10-23
Enable GitHub Release Zone in Main Branches.

thsparks:
This is a very targeted fix. I tried going through and making "main" and "master" equivalent throughout the github integration code ([thsparks/fix_main_in_github_integration](https://github.com/microsoft/pxt/tree/thsparks/fix_main_in_github_integration)), but I struggled with how to determine which one is used by a repo when on a non-main/master branch and I didn't think it was worth spending a ton of time on, so I paused that and focused on this quick fix instead. At least now, you can push releases and create branches from a Main branch.

### DIFF
--- a/webapp/src/gitjson.tsx
+++ b/webapp/src/gitjson.tsx
@@ -235,8 +235,8 @@ class GithubComponent extends data.Component<GithubProps, GithubState> {
             }
         }))
 
-        // only branch from master...
-        if (gid.tag == "master") {
+        // only branch from main and master...
+        if (gid.tag === "master" || gid.tag === "main") {
             branchList.unshift({
                 name: lf("Create new branch"),
                 description: lf("Based on {0}", gid.tag),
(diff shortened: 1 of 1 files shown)


srietkerk (approved):
Nice!!

webapp/src/gitjson.tsx
```diff
@@ -235,8 +235,8 @@ class GithubComponent extends data.Component<GithubProps, GithubState> {
             }
         }))
 
-        // only branch from master...
-        if (gid.tag == "master") {
+        // only branch from main and master...
+        if (gid.tag === "master" || gid.tag === "main") {
```
abchatra, line 239: There is a _default_branch_ in github repo which you should query and do this?

abchatra: _default_branch_ is configurable and need not be main.

abchatra: Feel free to be not blocked on this.

thsparks: Yeah this is not a fool-proof fix by any means. It's just a targeted change to unblock a bit of functionality. I think making us more resilient to different default branches would be nice, but a larger change.

thsparks: Filed this as a follow-up, though I expect it's somewhat low-pri: https://github.com/microsoft/pxt/issues/9744

PR #9749, opened 2023-10-26
Fix Release Zone on Main Branch

thsparks:
Forgot to include this change in https://github.com/microsoft/pxt/pull/9742


srietkerk (approved):
Nice!

PR #9751, opened 2023-10-26
Add info message after saving a copy of a project in time machine

riknoll:
image

Fixes https://github.com/microsoft/pxt-minecraft/issues/2382

Now when you save a copy of a project from version history, a toast shows up telling you it happened.


srietkerk (approved):
Nice!

Issue #9684, opened 2023-09-17
Remove scaling when hovering over the button to exit debug mode


THEb0nny:
I suggest removing the scaling when hovering over the button to exit debug mode, because... this is not visually beautiful... the size of the button increases so much that it extends beyond the right side of the screen, and visual bugs also appear in the editor. It seems to me that changing the color of the button will be sufficient.

@kimprice what do you think of this proposal?

![2023-09-17_21-05-29](https://github.com/microsoft/pxt/assets/13646226/da0ab726-20f4-4aeb-9ef7-4cd39be59c17)

kimprice:
yes, I agree.

THEb0nny:
> yes, I agree.
All other buttons do not have such a scaling animation, only a color change.

THEb0nny:
> > yes, I agree.
> > All other buttons (similar) do not have such a scaling animation, only a color change.

PR #9755, opened 2023-10-26
Remove scaling at hover over the button to exit debug mode

THEb0nny:
Remove the scaling of the button to exit debug mode when hover.

https://github.com/microsoft/pxt/issues/9684

### DIFF
--- a/theme/themes/pxt/collections/menu.overrides
+++ b/theme/themes/pxt/collections/menu.overrides
@@ -33,10 +33,6 @@
     background: @invertedBackground;
 }
 
-#mainmenu > .menu > .ui.item:hover > .icon {
-    transform: scale(1.2);
-}
-
 @media only screen and (max-width: @largestMobileScreen) {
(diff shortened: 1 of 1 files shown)


THEb0nny:
@kimprice you can view?

theme/themes/pxt/collections/menu.overrides
```diff
@@ -33,10 +33,6 @@
     background: @invertedBackground;
 }
 
-#mainmenu > .menu > .ui.item:hover > .icon {
-    transform: scale(1.2);
-}
```
kimprice, line 38: This change is going to impact more than just the exit debug button. I think we'll need to do something targeted for ".exit-debugmode-btn". Possibly using the "not" css attribute.

THEb0nny: I did it using the not selector, I don’t know why I didn’t do it right away...

kimprice (approved):
Thanks for your contribution!

PR #9762, opened 2023-10-30
kiosk: Fix for clean=1 param

eanders-ms:
More fun with Swiper idiosyncrasies. When Swiper's `loop` option is enabled, it will duplicate slides until there are three, but it won't distinguish between them when rendering, so you won't know which is actually selected. Transitioning between these duplicated slides doesn't reliably trigger a React re-render. I had to work around this behavior to get the carousel rendering correctly when there were less than three games uploaded and clean=1. @srietkerk this probably sounds familiar to you! :)

Fixes https://github.com/microsoft/pxt-arcade/issues/6169
Fixes https://github.com/microsoft/pxt-arcade/issues/6136, incidentally

### DIFF
--- a/kiosk/src/App.tsx
+++ b/kiosk/src/App.tsx
@@ -46,11 +46,11 @@ function App() {
                 if (cfg) {
                     dispatch(Actions.setTargetConfig(cfg));
                     if (cfg.kiosk) {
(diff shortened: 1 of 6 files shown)


kimprice (approved):
LGTM!

PR #9771, opened 2023-11-02
Editor tour: add play button for Minecraft

kimprice:
To enable the editor tour in Minecraft, we need to include the play button which is specific to Minecraft.

### DIFF
--- a/webapp/src/onboarding.ts
+++ b/webapp/src/onboarding.ts
@@ -29,6 +29,9 @@ function getTargetMap(target: string): querySelector {
         "download": {
             targetQuery: "#downloadArea",
         },
+        "play button" : {
+            targetQuery: ".big-play-button",
+        },
         "everything" : {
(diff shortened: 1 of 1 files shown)


jwunderl (approved):

webapp/src/onboarding.ts
```diff
@@ -29,6 +29,9 @@ function getTargetMap(target: string): querySelector {
         "download": {
             targetQuery: "#downloadArea",
         },
+        "play button" : {
```
srietkerk, line 32: Will there be any problems with having a space in this name? Not sure how these are parsed.

jwunderl: ^ I had the same thought but the ones just above also have spaces and are working fine~

kimprice: Yea good question, it works fine with the space.

PR #9772, opened 2023-11-02
Update Code Validation Docs to Include Validate-Exists Marker

thsparks:
Wanted to update the docs to include the changes made in https://github.com/microsoft/pxt/pull/9675.

These can now be utilized in Minecraft to validate non-highlighted blocks, which is nice!

### DIFF
--- a/docs/writing-docs/tutorials/basics.md
+++ b/docs/writing-docs/tutorials/basics.md
@@ -275,7 +275,7 @@ Within a validation section, you may specify which validators you want to enable
 
 ### Validators
 
-Currently, only one validator exists: the `BlocksExistValidator`. This validator looks at blocks tagged with `//@validate-exists` or `//@highlight` comments in the answer key and confirms that, for each tagged block, the user's code contains at least one block of the same type. It does *not* validate the parameters passed into the block. For more information on highlighting, see [highlighted blocks](../snippets.md#highlight).
+Currently, only one validator exists, the `BlocksExistValidator`, for [highlight](/writing-docs/snippets.md#highlight) and [validate-exists](/writing-docs/snippets#validate-exists) blocks. This validator looks at blocks tagged with `//@validate-exists` or `//@highlight` comments in the answer key and confirms that, for each tagged block, the user's code contains at least one block of the same type. It does *not* validate the parameters passed into the block.
 
 You can specify whether the `BlocksExistValidator` checks for `//@validate-exists` or `//@highlight` using the `markers` property on the validator. If you specify only `validate-exists`, then highlighted blocks will not be validated. Similarly, if you specify only `highlight`, then blocks marked with `validate-exists` will not be checked. By default, both checks are enabled.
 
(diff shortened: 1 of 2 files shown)


thsparks:
Hmm it just occurred to me, this will only work in targets that have had a pxt bump since the original change (just Minecraft at the moment), so we may need to wait until we've had microbit and arcade releases too...

docs/writing-docs/tutorials/basics.md
```diff
@@ -275,9 +275,11 @@ Within a validation section, you may specify which validators you want to enable
 
 ### Validators
 
-Currently, only one validator exists: the `BlocksExistValidator`. This validator looks at [highlighted blocks](../snippets.md#highlight) in the answer key and confirms that, for each highlighted block, the user's code contains at least one block of the same type. It does *not* validate the parameters passed into the block.
+Currently, only one validator exists: the `BlocksExistValidator`. This validator looks at blocks tagged with `//@validate-exists` or `//@highlight` comments in the answer key and confirms that, for each tagged block, the user's code contains at least one block of the same type. It does *not* validate the parameters passed into the block. For more information on highlighting, see [highlighted blocks](../snippets.md#highlight).
```
srietkerk, line 278: I feel like the "for more information on highlighting" bit is misplaced. It's a bit confusing to me since the paragraph seems to be talking about both those tags, but the resources only links to one. I'm not really sure of a better place for it, though...

thsparks: I agree. I wasn't sure where to place it either. I could just remove it :) I only kept it because we had the link earlier.

ganicke: I think you could remove it.

ganicke: Could add another section in `snippets.md` for `validate-exists` and then refer to them:

```suggestion
Currently, only one validator exists: the `BlocksExistValidator`. This validator looks at blocks tagged with `//@validate-exists` or `//@highlight` comments (see [highlight](/writing-docs/snippets.md#highlight) and [validate-exists](/writing-docs/snippets#validate-exists) in code blocks) in the answer key and confirms that, for each tagged block, the user's code contains at least one block of the same type. It does *not* validate the parameters passed into the block.
```

ganicke: Or, maybe:

```suggestion
Currently, only one validator exists, the `BlocksExistValidator`, for [highlight](/writing-docs/snippets.md#highlight) and [validate-exists](/writing-docs/snippets#validate-exists) blocks. This validator looks at blocks tagged with `//@validate-exists` or `//@highlight` comments in the answer key and confirms that, for each tagged block, the user's code contains at least one block of the same type. It does *not* validate the parameters passed into the block.
```

thsparks: Good idea. Thanks for making that change!

ganicke (approved):

thsparks:
> Hmm it just occurred to me, this will only work in targets that have had a pxt bump since the original change (I think just Minecraft at the moment), so we may need to wait until we've had microbit and arcade releases too...

Discussed in stand-up and decided it's okay to commit this, since we're really the only ones using it at the moment anyway.

PR #9778, opened 2023-11-10
Before compiling, try to fetch built js from backend

eanders-ms:
In pxtrunner, before building the project from source, attempt to fetch the compiled game from the backend.

Tested with run.html and works locally. I'm unsure this is the best approach though, and I don't know if this will work with the share page.

### DIFF
--- a/pxtlib/emitter/cloud.ts
+++ b/pxtlib/emitter/cloud.ts
@@ -125,6 +125,17 @@ namespace pxt.Cloud {
         })
     }
 
+    export function downloadBuiltSimJsInfoAsync(id: string): Promise<pxtc.BuiltSimJsInfo> {
+        const targetVersion = pxt.appTarget.versions && pxt.appTarget.versions.target || "";
+        const url = pxt.U.stringifyQueryString(id + "/js", { v: "v" + targetVersion }) + (id.startsWith("S") ? `&time=${Date.now()}` : "");
+        return privateRequestAsync({
+            url,
+            forceLiveEndpoint: true,
+        }).then(resp => {
+            return resp.json
+        });
+    }
+
     export async function markdownAsync(docid: string, locale?: string, propagateExceptions?: boolean): Promise<string> {
         // 1h check on markdown content if not on development server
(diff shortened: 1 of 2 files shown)


pxtlib/emitter/cloud.ts
```diff
@@ -125,6 +125,17 @@ namespace pxt.Cloud {
         })
     }
 
+    export function downloadBuiltSimJsInfoAsync(id: string): Promise<pxtc.BuiltSimJsInfo> {
```
riknoll, line 128: async/await?

riknoll (approved):

pxtrunner/runner.ts
```diff
@@ -466,6 +466,14 @@ namespace pxt.runner {
         simDriver?.postMessage(msg);
     }
 
+    export async function fetchSimJsInfo(simOptions: SimulateOptions): Promise<pxtc.BuiltSimJsInfo> {
+        try {
+            return await pxt.Cloud.downloadBuiltSimJsInfoAsync(simOptions.id);
+        } catch {
+            return undefined;
```
riknoll, line 473: maybe add a console warn here

jwunderl (approved):

pxtrunner/runner.ts
eanders-ms: Sure. Since most games won't be precompiled, the exception path will execute here virtually every time this is called, but logging it could be useful for debugging.

pxtlib/emitter/cloud.ts
eanders-ms: This function modeled after similar ones above, but I'll bring this one into the modern age :)

eanders-ms:
@aznhassan please feel free to merge this when you're ready.

abchatra:
Will this change work without related backend changes?

eanders-ms:
The backend has an existing `/js` endpoint [here](https://github.com/microsoft/pxt-backend/blob/1522a0ed943f8d98cf0ebb24a0fecf23a604c8bc/node/backend/src/sharing.ts#L542). If it returns something unexpected here, that could be a problem.

eanders-ms:
It would be safer to change the endpoint to a new one. Something like `/builtjs` would work.

PR #9783, opened 2023-11-15
[stable9.2] Make normalize language code which is not only 2-2 / Add langs (#9763)

sae220:
cherry picking #9763


abchatra:
@sae220 can you explain what you want to achieve?

sae220:
@abchatra 
Other pull-requests are created for both latest and stable version, so I created this for stable version.

abchatra:
We don't need to port to stable channels unless there is an urgent need to hotfix this. I will close this PR for now.

PR #9786, opened 2023-11-27
Add pxtarget.json options for tuning time machine intervals

riknoll:
I'm a little worried about accidentally making projects too big in arcade and messing with cloud sync. Adding a couple of pxtarget.json options to increase the project history interval so that we can start with a higher interval and lower it if all goes well.


eanders-ms:
Max project size for cloud sync is 32mb, [(here)](https://github.com/microsoft/pxt-backend/blob/7ce14ad66788c972e8de8c745a1b6d4ca62d9048/node/backend/src/project-store.ts#L11). Can the history system do things to stay under this line?

riknoll:
@eanders-ms I will definitely look into doing that, but i want to merge this first so that i can get some version of it into beta for the vendor team

eanders-ms (approved):
Sorry thought this was approved!

PR #9789, opened 2023-12-06
Check for qualified name in sprite editor

kimprice:
The sprite editor will now check for qualified names such as "sprites.food.smallBurger" when converting to blocks from JS and maintain those names when converting back to JS.

This change in combination with https://github.com/microsoft/pxt-common-packages/pull/1459 is a fix for the image picker problem that @kiki-lee was having in which using a qualified name resulted in the block having an image picker instead of the sprite editor.

Before:
image

After:
image
image

### DIFF
--- a/pxtblocks/fields/field_asset.ts
+++ b/pxtblocks/fields/field_asset.ts
@@ -38,6 +38,8 @@ namespace pxtblockly {
         protected pendingEdit = false;
         protected isEmpty = false;
 
+        protected qName?: string;
+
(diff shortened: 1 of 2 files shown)


kimprice:
I also checked this change by importing an old project that used the image picker for a qualified name and it successfully converted to a sprite editor.

abchatra:
Lets evaluate the risk of the change & testing required. If low risk, port to stable channels

PR #9790, opened 2023-12-07
Add index of the source iframe to the broadcast message

eanders-ms:
In the radio scenario, the robot simulator (https://github.com/microsoft/microbit-robot) will receive messages from multiple microbits. The botsim needs know if a message is coming from the primary or secondary microbit.

image


kimprice (approved):

jwunderl (approved):
might be worth doing this on line 315 so it's stable even if `frameID` is passed, but I suppose we shouldn't ever hit that case with existing code as we saw that `frameID` isn't passed when propagating messages from one frame to another / probably that won't ever change.

pelikhan (approved):

PR #9794, opened 2023-12-08
Track js compilation time

aznhassan:
Enables telemetry in `runner.ts` to track the time it takes to either download the compiled JS or download and compile the text. These events will help know the impact the pre-compilation step is having.

### Validation
Ran pxt-arcade locally with a local `pxt` on this branched linked. I then changed the InstrumentationKey in [apptracking.html](../blob/master/docfiles/apptracking.html) and [tracking.html](../blob/master/docfiles/tracking.html) to the PPE app insights instance. 
- Visited http://localhost:3232/68336-13291-41085-56601 in an incognito tab a few times. I was then able to confirm the new `runner.fetchSimJsInfo` event was present.
- Visited http://localhost:3232/56776-01920-30861-75662 in an incognito tab a few times. Confirmed that the new `runner.buildSimJsInfo` event was present.

Both these links are the same game just shared twice. For the first one I precompiled the JS to the `fetchSimJsInfo` event would be recorded. After loading each page five times in a new icognito tab each time here is what I found:
- buildSimJsInfo had an average of 1463.4 milliseconds, a min of 1171 milliseconds, and a max of 1748 milliseconds.
- fetchSimJsInfo had an average of 148 milliseconds, a min of 29 milliseconds, a max of 723 milliseconds.

>Note: It's important to note that for fetching the pre-compiled JS the initial page load was the longest. After that, the request was cached by the browser. While the request to grab the share page's text is also cached when compiling the JS, the actual compilation still takes time. Regardless, fetching the compressed pre-compiled JS is faster than fetching the text and then compiling client-side.

### DIFF
--- a/pxtrunner/runner.ts
+++ b/pxtrunner/runner.ts
@@ -471,7 +471,10 @@ namespace pxt.runner {
         try {
             const start = Date.now();
             const result = await pxt.Cloud.downloadBuiltSimJsInfoAsync(simOptions.id);
-            pxt.tickEvent("runner.fetchSimJsInfo", { durationMs: Date.now() - start });
+            pxt.tickEvent("perfMeasurement", {
+              durationMs: Date.now() - start,
+              operation: "fetchSimJsInfo",
+            });
             return result;
         } catch (e) {
             // This exception will happen in the majority of cases, so we don't want to log it unless for debugging.
@@ -544,7 +547,10 @@ namespace pxt.runner {
 
         const res = pxtc.buildSimJsInfo(compileResult);
         res.parts = compileResult.usedParts;
-        pxt.tickEvent("runner.buildSimJsInfo", { durationMs: Date.now() - start });
+        pxt.tickEvent("perfMeasurement", {
+          durationMs: Date.now() - start,
+          operation: "buildSimJsInfo",
+        });
         return res;
     }
 
--- a/pxtrunner/runner.ts
+++ b/pxtrunner/runner.ts
@@ -471,7 +471,7 @@ namespace pxt.runner {
         try {
             const start = Date.now();
             const result = await pxt.Cloud.downloadBuiltSimJsInfoAsync(simOptions.id);
-            pxt.tickEvent("runner.fetchSimJsInfo", { durationMs: Date.now() - start});
+            pxt.tickEvent("runner.fetchSimJsInfo", { durationMs: Date.now() - start });
             return result;
         } catch (e) {
             // This exception will happen in the majority of cases, so we don't want to log it unless for debugging.
@@ -544,7 +544,7 @@ namespace pxt.runner {
 
         const res = pxtc.buildSimJsInfo(compileResult);
         res.parts = compileResult.usedParts;
-        pxt.tickEvent("runner.buildSimJsInfo", { durationMs: Date.now() - start});
+        pxt.tickEvent("runner.buildSimJsInfo", { durationMs: Date.now() - start });
         return res;
     }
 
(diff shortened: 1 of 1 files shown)


pxtrunner/runner.ts
```diff
@@ -213,6 +213,7 @@ namespace pxt.runner {
 
     function initInnerAsync() {
         pxt.setAppTarget((window as any).pxtTargetBundle)
+        pxt.analytics.enable(pxt.Util.userLanguage());
```
jwunderl, line 216: Not sure about this change? We have other ticks coming in so feels unexpected to have to change this, `enableAnalytics` call should be getting in thru here https://github.com/search?q=repo%3Amicrosoft%2Fpxt%20analytics.enable&type=code

aznhassan: If I don't include this line, the `tickEvents` do not make it to Application Insights. See how there are only `share.isMultiplayer` ticks most recently where I commented out this line? Whereas in around 3:04pm I had still had this line enabled and we get the same events mixed in with the new `runner` events.
```suggestion
        pxt.analytics.enable(pxt.Util.userLanguage());
```

image

abchatra: Adding @thsparks. I have observed that pageview ticks stopped flowing recently as well.

thsparks (approved):
👍

pxtrunner/runner.ts
```diff
@@ -468,7 +469,10 @@ namespace pxt.runner {
 
     export async function fetchSimJsInfo(simOptions: SimulateOptions): Promise<pxtc.BuiltSimJsInfo> {
         try {
-            return await pxt.Cloud.downloadBuiltSimJsInfoAsync(simOptions.id);
+            const start = Date.now();
+            const result = await pxt.Cloud.downloadBuiltSimJsInfoAsync(simOptions.id);
+            pxt.tickEvent("runner.fetchSimJsInfo", { durationMs: Date.now() - start });
```
thsparks, line 474: Food for thought: do we want 2 different events here, or is it more helpful/convenient to have one "perfMeasurement" event with a property for the event? In some cases it's helpful to have the latter, so you can easily query all of them and sort by which events are taking the longest. Trickier to do that if they all have different names.

pxtrunner/runner.ts
aznhassan: @jwunderl There are five files where `pxt.analytics.enable` are called in the linked you provided:

1. Kiosk does not apply in this case since I tested it by opening a share-page after running Arcade locally.
2. I don't believe the share page loads the web app, so telemetry should not be enabled via that route. Though if I'm incorrect, please let me know!
3. Multiplayer page separately enables telemetry, but that does not apply here.
4. Same with the teachertool page.

I can see that argument that if do enable additional telemetry for share pages, then we should match what we've done before in these other examples that you've linked.

pxtrunner/runner.ts
aznhassan: Fetching the already built JS info and building the JS info from the script text (which we request from the backend as well) are mutually exclusive. Though I agree that using something like `perfMeasurement` vs `runner` will make it easier to keep track of these. When you say property, you mean adding a customDimension with operation name? Just like we do with `duration` now?

pxtrunner/runner.ts
aznhassan: > Adding @thsparks. I have observed that pageview ticks stopped flowing recently as well.

These changes are unrelated to tracking page views, which are making it. @abchatra

jwunderl (approved):

pxtrunner/runner.ts
aznhassan: Ended up implementing this!

PR #9795, opened 2023-12-08
Add React App for Teacher Tool

thsparks:
This is largely just copy/paste from Kiosk with a few substitutions from multiplayer and skillmap. Build & run instructions in the readme.

Project folder structure mirror's Kiosk's. Inside src, we have:
1. Components: React UI components
2. Hooks: Custom hooks for react
3. Services: Manage continuous background tasks
4. State: Contains logic for managing redux-like app state
5. Transforms: Create transforms to handle big/complex state change events
6. Types: Custom type definitions
7. Utils: Helper functions, etc...


teachertool/src/App.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+    useEffect(() => {
+        // Init subsystems.
+        NotificationService.initialize();
+    }, [ready]);
+
+    // Test notification
+    useEffect(() => {
+        if (ready && !didNotify) {
+            postNotification(makeNotification("🎓", 2000));
```
eanders-ms, line 25: nice :)

teachertool/src/config.json
```diff
@@ -0,0 +1,3 @@
+{
+    "SoundEffectsVolume": 0.5
```
eanders-ms, line 2: It's ok to keep this file in the project, but want to avoid putting debug-only settings here. Let's find a better way to do that through the environment.

teachertool/src/state/state.ts
```diff
@@ -0,0 +1,12 @@
+import { Notifications } from "../types";
+
+export type AppState = {
+    targetConfig?: pxt.TargetConfig;
+    notifications: Notifications;
+    volume?: number; // volume level of UI sounds, in [0..1]
```
eanders-ms, line 6: I imagine we'll remove this `volume` setting.

eanders-ms (approved):
Very nice! :boom:

gulpfile.js
```diff
@@ -863,6 +895,8 @@ exports.skillmap = skillmap;
 exports.authcode = authcode;
 exports.multiplayer = multiplayer;
 exports.kiosk = kiosk;
+exports.teacherTool = teacherTool;
+exports.tt = teacherTool;
```
srietkerk, line 899: Is this safe to do? Just wanted to double check since it looks like all other exports are set to something different.

package.json
```diff
@@ -164,6 +166,6 @@
     "test:lang": "gulp testlang",
     "update": "gulp update",
     "watch-streamer": "cd docs/static/streamer && tsc -t es6 --watch",
-    "prepare": "cd skillmap && npm install && cd .. && cd authcode && npm install && cd .. && cd multiplayer && npm install && cd .. && cd kiosk && npm install && cd .."
+    "prepare": "cd skillmap && npm install && cd .. && cd authcode && npm install && cd .. && cd multiplayer && npm install && cd .. && cd kiosk && npm install && cd ..  && cd teachertool && npm install && cd .."
```
srietkerk, line 169: Not something for this PR, but I wonder if we can clean this up a bit somehow since there are a lot of repeated operations... Could we maybe make a script that this npm script points to? And it performs the directory change and install per whatever webapps we have configured?

teachertool/src/state/state.ts
thsparks: I just went ahead and removed it now

gulpfile.js
thsparks: I believe so. It just means both "gulp teacherTool" and "gulp tt" will run the same command. (At least, as far as I'm aware.)

teachertool/src/components/HeaderBar.tsx
```diff
@@ -0,0 +1,69 @@
...
+            }
+            window.open(pxt.appTarget.appTheme.homeUrl + rel);
+        }
+        else {
+            window.open(pxt.appTarget.appTheme.homeUrl);
+        }
+    }
+
+    render() {
+        const hasIdentity = pxt.auth.hasIdentity();
```
srietkerk, line 50: Oh, one last thing, will this cause an ESLint error because it doesn't get used?

thsparks: Checked passed, so apparently not :)

srietkerk (approved):
LGTM!

package.json
thsparks: Agree. There's a lot of shared bits and pieces, plus some repetitive things like what we have here. Would be nice to go through and simplify the whole process of creating new apps at some point, but I'm not sure if that'll happen anytime soon :)

eanders-ms:
woot!

PR #9797, opened 2023-12-12
add tick events for perf measurements

riknoll:
This adds two tick events that contain all of the info from the performance report that we print out to the console:

* `performance.milestones` - a map of event names to the elapsed time since page load for when that event occurred. For example, if blockly loaded 2 seconds after page load you'd get an entry like this: `loadBlockly: 2000`
* `performance.durations` - a map of task names to the amount of elapsed time that task took to complete

These ticks are only sent once per session! That makes them most useful for activities that are loaded directly from a URL, as the minecraft tutorials are.

I also added three new duration measures:

* `startActivity` - this is the overall time it takes for a tutorial or example to load into the editor
* `downloadMarkdown` - this is the time it takes to download markdown for a tutorial
* `downloadGitHubTutorial` - this is the time it takes to fetch a GitHub repo that contains a tutorial

and one new milestone:

* `editorContentLoaded` - the time at which the tutorial has fully loaded (you can interact with the Blockly workspace)

Other things I did:

1. Fixed a bug where the simulator ready event was never firing for the first time the editor was loaded (we weren't registering the message event listener in simdriver for preloaded iframes)
2. Converted loadActivityFromMarkdownAsync to async/await
3. Changed where the reporting happens for targets that do not autorun the simulator. Instead of in the simulator ready event, it fires once the content has loaded in the editor

### DIFF
--- a/docfiles/pxtweb/cookieCompliance.ts
+++ b/docfiles/pxtweb/cookieCompliance.ts
@@ -106,14 +106,21 @@ namespace pxt {
             }
         }
         export function report(filter: string = null) {
+            perfReportLogged = true;
+
             if (enabled) {
+                const milestones: {[index: string]: number} = {};
+                const durations: {[index: string]: number} = {};
+
                 let report = `performance report:\n`
                 for (let [msg, time] of stats.milestones) {
                     if (!filter || msg.indexOf(filter) >= 0) {
                         let pretty = prettyStr(time)
                         report += `\t\t${msg} @ ${pretty}\n`
+                        milestones[msg] = time;
                     }
                 }
+
                 report += `\n`
                 for (let [msg, start, duration] of stats.durations) {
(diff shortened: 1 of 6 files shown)


aznhassan:
I just added a `perfMeasurement` tick to [runner.ts](../blob/master/pxtrunner/runner.ts#L474) where I included the duration and operation. Do we want these to be unified? We could change `runner.ts` to use `performance.durations` instead of `perfMeasurement` like it is now.

thsparks (approved):
LGTM! Thanks!

riknoll:
@aznhassan i don't know if they need to be unified given that they are from different scenarios (the editor vs the script page). that being said, the payload should certainly be in the same format

abchatra (approved):

abchatra:
Looks good. Lets test in beta of minecraft and then hotfix it tomorrow.

PR #9801, opened 2023-12-20
Fix typo in math.md

abchatra:


ganicke:
Ok, that was incorrect by a certain degree. Thanks.

PR #9803, opened 2023-12-20
Add URL Parameter to Show Block IDs in Tooltip

thsparks:
This adds the tooltipBlockIds URL parameter. When set to 1 in the URL (?tooltipBlockIds=1), blocks will include their block id in the tooltip when you hover over them.

I'm doubtful that an exported bool directly inside pxt.blocks is the best way to store this, but I figured I'd send the PR anyway and get feedback. If anyone has thoughts on a better approach, please let me know!

### DIFF
--- a/pxtblocks/blocklyloader.ts
+++ b/pxtblocks/blocklyloader.ts
@@ -716,7 +716,8 @@ namespace pxt.blocks {
         block.setPreviousStatement(!(hasHandlers && !fn.attributes.handlerStatement) && fn.retType == "void");
         block.setNextStatement(!(hasHandlers && !fn.attributes.handlerStatement) && fn.retType == "void");
 
-        block.setTooltip(/^__/.test(fn.namespace) ? "" : fn.attributes.jsDoc);
+        block.setTooltip((/^__/.test(fn.namespace) ? "" : fn.attributes.jsDoc) + (pxt.blocks.showBlockIdInTooltip ? " (id: '" + fn.attributes.blockId + "')" : ""));
+
         function buildBlockFromDef(def: pxtc.ParsedBlockDef, expanded = false) {
             let anonIndex = 0;
             let firstParam = !expanded && !!comp.thisParameter;
--- a/pxtlib/blocks.ts
+++ b/pxtlib/blocks.ts
@@ -3,6 +3,8 @@
 namespace pxt.blocks {
     const THIS_NAME = "this";
 
+    export let showBlockIdInTooltip: boolean = false;
+
     // The JS Math functions supported in the blocks. The order of this array
     // determines the order of the dropdown in the math_js_op block
     export const MATH_FUNCTIONS = {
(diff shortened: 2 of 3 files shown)


pxtblocks/blocklyloader.ts
```diff
@@ -716,7 +716,8 @@ namespace pxt.blocks {
         block.setPreviousStatement(!(hasHandlers && !fn.attributes.handlerStatement) && fn.retType == "void");
         block.setNextStatement(!(hasHandlers && !fn.attributes.handlerStatement) && fn.retType == "void");
 
-        block.setTooltip(/^__/.test(fn.namespace) ? "" : fn.attributes.jsDoc);
+        block.setTooltip((/^__/.test(fn.namespace) ? "" : fn.attributes.jsDoc) + (pxt.blocks.showBlockIdInTooltip ? " (id: '" + fn.attributes.blockId + "')" : ""));
```
srietkerk, line 719: This is a question about validation since I know that this tooltip and the flag are largely for help with investigation/debugging there. For our blocks exist validation currently, are we only looking at block id? Just wanted to make sure that there weren't other identifiers that we were using to look for blocks because I remember in some docs work that I did early on, that qName was used, but I might be misremembering it's significance.

pxtlib/blocks.ts
```diff
@@ -806,5 +808,20 @@ namespace pxt.blocks {
                 )
             })
         }
+
+        if (pxt.blocks.showBlockIdInTooltip) {
+            for (const id of Object.keys(_blockDefinitions)) {
+                const tooltip = _blockDefinitions[id].tooltip;
+                if (typeof tooltip === "object" && tooltip !== null) {
```
srietkerk, line 815: Wondering if these conditions should be reversed for short circuiting or if the second check is needed. If the tooltip has type object, is it possible for it to be null?

pxtlib/blocks.ts
```diff
@@ -806,5 +808,20 @@ namespace pxt.blocks {
...
+                    for (const innerKey in tooltip) {
+                        if (tooltip.hasOwnProperty(innerKey)) {
```
srietkerk, line 817: Not sure what this check will do.. If we're iterating through the `innerKey` of the specific tooltip, it should have that `innerKey`, right?

pxtlib/blocks.ts
thsparks: Weirdly, I believe typeof null will return object, so `typeof tooltip === "object"` would be true. I think it'd be fine to check for null first, but I don't think it'll make much difference.

pxtlib/blocks.ts
thsparks: It's to protect against any fields tooltip may have inherited which aren't directly on tooltip itself. I don't think it's strictly necessary here, since tooltip is defined pretty explicitly when it isn't a string, but I thought better safe than sorry.

pxtblocks/blocklyloader.ts
thsparks: I believe we check against the id and nothing else, though we could potentially change that in the future if needed.

srietkerk (approved):
LGTM

PR #9806, opened 2023-12-22
Teacher tool iframe

srietkerk:
I wanted to get out a foundation for the iframe since our work is growing and it will be useful to have the iframe for the  teacher tool. 

There are a couple of things that I would like to expand upon for iframe to be more generic and usable across any situation in the future:
- Move it to react-common so it's in one place and easily accessible
- Make the props more extensible to allow an app's react-redux store to be hooked into the iframe. If we do this, we can refactor skillmap's iframe.
- Make it so the iframe can respond to a wider range of event listeners.
- Investigate more on different architectures. I think having the component the way it is right now is fine, but I've also considered having a redux store for the iframe itself because then it can be extended on via actions and dispatch functions, but I'm uncertain if that is the best way forward.


teachertool/src/App.tsx
```diff
@@ -30,13 +32,40 @@ function App() {
...
+        return url;
+    }
+
     return (
         <>
             <HeaderBar />
-            <div className="appContainer">
-                <DebugInput />
+            <div className="app-container">
+                <div className="debug-container">
```
thsparks, line 58: Can we move the debug-container div inside the DebugInput component itself? Just thinking it might be good to keep it all contained there.

teachertool/src/components/makecodeFrame.tsx
```diff
@@ -0,0 +1,170 @@
+/// <reference path="../../../built/pxteditor.d.ts" />
+import * as React from "react";
+import { useState, useEffect, useRef } from "react";
+import { ProgressBar } from "react-common/components/controls/ProgressBar";
+
+// This is something I want to do eventually. The iframe should be able to work with a redux store
+// import { connect } from 'react-redux';
```
thsparks, line 7: Not a huge concern but I'd recommend removing the commented-out code before checking in. (Could move it to a different branch, if you're worried about losing it).

teachertool/src/components/makecodeFrame.tsx
```diff
@@ -0,0 +1,170 @@
...
+    let frameRef = useRef<HTMLIFrameElement>(null);
+
+    useEffect(() => {
+        // logic we want to do when the iframe is loaded
+        const root = document.getElementById("root");
+        if (ref && ref.contentWindow) {
+            window.addEventListener("message", onMessageReceived);
+            ref.addEventListener("load", handleFrameReload)
+
+            if (root) pxt.BrowserUtils.addClass(root, "editor");
```
thsparks, line 61: What's the motivation for adding the editor class to our root element?

teachertool/src/utils/browserUtils.ts
```diff
@@ -0,0 +1,12 @@
+export const isLocal = () => {
+    return window.location.hostname === "localhost";
+}
+
+export const getEditorUrl = (embedUrl: string) => {
```
thsparks, line 5: I think a comment explaining what this returns (maybe with an example) would be super helpful.

teachertool/src/components/makecodeFrame.tsx
```diff
@@ -0,0 +1,170 @@
...
+        if (frameState === "opening-project") setLoadPercent(Math.min((loadPercent || 0) + 7, 95));
+
+        if (data.type === "pxteditor" && data.id && pendingMessages[data.id]) {
+            const pending = pendingMessages[data.id];
+            pending.handler(data);
+            delete pendingMessages[data.id];
+            return;
+        }
+
+        switch (data.action) {
```
thsparks, line 107: Similar to removing the commented-out code, I'd recommend taking the unimplemented sections of this switch statement and moving them to a different branch for now. I would uncomment and keep the default statement though, which prints out the message. I think that could be quite helpful.

teachertool/src/App.tsx
```diff
@@ -30,13 +32,40 @@ function App() {
...
 
+    const onIframeLoaded = () => {
+        console.log("iframe loaded");
+    }
+
+    const onIframeClosed = () => {
+        console.log("iframe closed");
+    }
+
+    const createIFrameUrl = (): string => {
```
thsparks, line 43: No need to block on this, but as a nit-pick, I think app.tsx may not be the right place for this function. I'd recommend maybe putting it inside a util function or perhaps even inside the MakeCodeFrame file...but since you're trying to make the MakeCodeFrame more generic, I'd lean towards util (or even creating an additional component that wraps the MakeCodeFrame...)

teachertool/src/components/makecodeFrame.tsx
```diff
@@ -0,0 +1,170 @@
+/// <reference path="../../../built/pxteditor.d.ts" />
```
thsparks, line 1: For this file (and components in general), it might be good to keep the capitalize-the-first-letter naming convention (i.e. MakeCodeFrame.tsx). It's a little weird/inconsistent with our naming for other files, but personally I think it's nice for the file to map to the component name exactly.

teachertool/src/utils/browserUtils.ts
```diff
@@ -0,0 +1,12 @@
+export const isLocal = () => {
```
thsparks, line 1: Personally, I don't see an issue with just putting these inside the existing index.ts file in utils rather than creating a separate file for them, but it's not something I feel strongly about.

thsparks (approved):
A few smaller comments I think we should address, but nothing extreme. Feel free to push back if you disagree with any of them!

teachertool/src/components/makecodeFrame.tsx
srietkerk: Ah, good question. I thought this was important, but investigated and it looks like this was added for skillmap to work around some weirdness with the header, but I don't think that will apply in this situation. If there are special cases like this, I think it's something we should address in the future, so I'll take it out for now. Thanks, Thomas!

teachertool/src/components/makecodeFrame.tsx
srietkerk: Yeah, good point. I was wanting to leave these in as a reminder, but I have notes and the skillmap example to look back on for extensibility things. I'll take the comments out.

teachertool/src/components/makecodeFrame.tsx
srietkerk: Will do. As I said in your other comment about removing the comments, I don't think it's necessary to put this in a separate branch. The skillmap iframe holds a good reference for everything here that had no-ops, so we can look there.

teachertool/src/utils/browserUtils.ts
srietkerk: Sure, makes sense. I'll move the functions.

teachertool/src/App.tsx
srietkerk: Yeah, that's a good point. I'll put the frame in a separate component. I think this will be better for future use, anyway, since I would like to lift the generic frame to react-common in the near future.

teachertool/src/components/makecodeFrame.tsx
srietkerk: I will do this as a last step since when I do this, it will say there's a new file. I think that would be annoying to try to read through when looking at my other new updates.

teachertool/src/components/EditorContainer.tsx
```diff
@@ -0,0 +1,31 @@
...
+export const EditorContainer: React.FC<{}> = () => {
+    const onIframeLoaded = () => {
+        console.log("iframe loaded");
+    }
+
+    const onIframeClosed = () => {
+        console.log("iframe closed");
+    }
+
+    return <>
```
thsparks, line 24: nit: I don't think you need the <> </> tags here, but I don't think they're harming anything either.

thsparks (approved):
Nice!

srietkerk:
I'm gonna wait to merge this. I still need to change the file name for the frame, and let this rebuild, but I need to log off.

teachertool/src/components/makecodeFrame.tsx
```diff
@@ -0,0 +1,138 @@
...
+import * as React from "react";
+import { useState, useEffect, useRef } from "react";
+import { ProgressBar } from "react-common/components/controls/ProgressBar";
+interface MakeCodeFrameProps {
+    pageSourceUrl: string;
+    highContrast?: boolean;
+    onFrameOpen: () => void;
+    onFrameClose: () => void;
+    tutorialEventHandler?: (event: pxt.editor.EditorMessageTutorialEventRequest) => void;
```
kimprice, line 10: What is tutorialEventHandler for?

teachertool/src/components/makecodeFrame.tsx
```diff
@@ -0,0 +1,138 @@
...
+        // logic we want to do when the iframe is loaded
+        if (ref && ref.contentWindow) {
+            window.addEventListener("message", onMessageReceived);
+            ref.addEventListener("load", handleFrameReload)
+
+        }
+
+        // logic we want when the iframe unmounts
+        return () => {
+            window.removeEventListener("message", onMessageReceived);
```
kimprice, line 46: Do you want to remove the event listener for ref too?

teachertool/src/components/makecodeFrame.tsx
```diff
@@ -0,0 +1,138 @@
...
+        sendMessageAsync({
+            type: "pxteditor",
+            action: "sethighcontrast",
+            on: highContrast
+        }  as pxt.editor.EditorMessageSetHighContrastRequest);
+    }, [highContrast])
+
+
+    useEffect(() => {
+        if (frameState === "project-open" /*&& this.props.save*/) {
```
kimprice, line 60: remove comment?

teachertool/src/utils/index.ts
```diff
@@ -11,4 +11,21 @@ export function makeNotification(
...
         expiration: Date.now() + duration,
     };
+}
+
+export const isLocal = () => {
+    return window.location.hostname === "localhost";
+}
+
+// example: embedUrl for arcade is: https://arcade.makecode.com/
+export const getEditorUrl = (embedUrl: string) => {
```
thsparks, line 21: It just occurred to me, I believe this is built to essentially detect which target we're on and use the same target for the link we send to the iframe. This might end up being fine, but we also have the target in the shared project's metadata, so it may make more sense to construct the URL using that instead (see targetToUrl function in https://github.com/microsoft/pxt/compare/srietkerk-eval-iframe...thsparks/get_blocks_for_validation)

Not necessary for checking this in, but thought I'd mention it.

kimprice (approved):
Looks good! :) Just had a couple comments.

teachertool/src/components/makecodeFrame.tsx
```diff
@@ -0,0 +1,138 @@
...
+                if (frameState === "loading") {
+                    setFrameState("no-project");
+                }
+                break;
+            default:
+                console.log(JSON.stringify(data, null, 4));
+        }
+    }
+
+    const sendMessageAsync = (message?: any) => {
```
thsparks, line 101: Something to think about is how will we actually call into this function when we want to send a message? I could certainly be missing something, but at the moment I don't think it's exposed in any way, and I'm not sure it'll be trivial to do so without coupling this class quite tightly to Teacher Tool state.

Did you already have a specific plan in mind around how we'll call this, or is it still TBD?

teachertool/src/components/makecodeFrame.tsx
srietkerk: This is something that I want to have here for making the frame more generic.  The generic MakeCodeFrame is the thing that needs to listen for the tutorial event messages from the editor, but the way that tutorial messages are processed I think are going to vary. For example, skillmap handles tutorial events in ways that are very specific to skillmaps. It would be great if the generic MakeCodeFrame replaced, or could replace, all of our uses of the iframe, so this function would be needed as a prop if and when the frame would be used by skillmap so the function that processes tutorial events on skillmap's side can do so. 

The function that I'm referencing from skillmap is here: https://github.com/microsoft/pxt/blob/master/skillmap/src/components/makecodeFrame.tsx#L282-L309

You can see that each of the individual events in this function call functions that are specific for skillmap.

teachertool/src/components/makecodeFrame.tsx
srietkerk: Yeah, good point.

teachertool/src/utils/index.ts
srietkerk: Yeah, interesting. I think we'd need both of these functions for things to work as expected, but maybe not. Or at least a merging of their functionalities. If there are any trailing details to the url, if we just used the `targetToUrl` function, those details would get lost. The functionality in `getEditorUrl` is definitely important for skillmaps, to preserve which skillmap is being visited, so that's something to keep in mind. 

I won't address this in this PR, but I agree that it's something to discuss.

teachertool/src/components/makecodeFrame.tsx
srietkerk: Yeah, that's a good point. It's something that I haven't thought about much, so definitely still TBD. I'll play with this a bit to see what we can do.

teachertool/src/utils/index.ts
```diff
@@ -11,4 +11,21 @@ export function makeNotification(
         duration,
         expiration: Date.now() + duration,
     };
+}
+
+export const isLocal = () => {
```
eanders-ms, line 16: `isLocal` -> fyi we have this method already in pxt.util.

teachertool/src/components/EditorContainer.tsx
```diff
@@ -0,0 +1,32 @@
+import { MakeCodeFrame } from "./MakecodeFrame";
+import { isLocal, getEditorUrl } from "../utils";
+
+const createIFrameUrl = (): string => {
+    const editorUrl: string = isLocal() ? "http://localhost:3232/index.html#editor" : getEditorUrl((window as any).pxtTargetBundle.appTheme.embedUrl);
```
eanders-ms, line 5: Noting here we need to open the editor in read-only mode. A url param may exist for this.

teachertool/src/components/MakecodeFrame.tsx
```diff
@@ -0,0 +1,141 @@
...
+type FrameState = "loading" | "no-project" | "opening-project" | "project-open" | "closing-project";
+
+interface PendingMessage {
+    original: pxt.editor.EditorMessageRequest;
+    handler: (response: any) => void;
+}
+
+export const MakeCodeFrame: React.FC<MakeCodeFrameProps> =
+    ( { pageSourceUrl,
+        highContrast,
```
eanders-ms, line 22: highContrast should exist in app state, not passed in props.

teachertool/src/components/EditorContainer.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+    let url = editorUrl
+    if (editorUrl.charAt(editorUrl.length - 1) === "/" && !isLocal()) {
+        url = editorUrl.substr(0, editorUrl.length - 1);
+    }
+    url += `?controller=1&ws=browser&nocookiebanner=1`;
+    return url;
+}
+
+export const EditorContainer: React.FC<{}> = () => {
```
eanders-ms, line 15: Not sure what value the `EditorContainer` container brings. Can this be rolled into `MakeCodeFrame`?

eanders-ms (commented):
Great starting point. Made a few comments.

teachertool/src/components/EditorContainer.tsx
srietkerk: The idea here is that there's something that wraps around the `MakeCodeFrame` for the teacher tool in particular to use, especially with the hope in mind that the `MakeCodeFrame` can be moved into react-common at some point. I'm under the impression that this is likely where we would handle the receiving blocks logic and things like that, as well. From playing with the skillmap iframe and thinking about generic cases, I imagined that the parent app should be able to control how the iframe is opened and closed, so the wrapper also provides a good place to define those functions.

The alternative would be to just put the functions and url-defining logic in the app base, but it felt a bit out of place.

teachertool/src/utils/index.ts
srietkerk: Ah, I forgot that I can just call it that way. I think I found it there and just copied it over, but I'll change it to be the function call in pxt.util instead. Thanks!

teachertool/src/components/MakecodeFrame.tsx
```diff
@@ -0,0 +1,141 @@
...
+        onFrameClose,
+        onFrameOpen
+    } ) => {
+    let ref: HTMLIFrameElement | undefined;
+    const messageQueue: pxt.editor.EditorMessageRequest[] = [];
+    let nextId: number = 0;
+    let pendingMessages: {[index: string]: PendingMessage} = {};
```
eanders-ms, line 29: `messageQueue` and related state should be moved to a service layer (e.g. `/src/services/makecodeEditor.ts`). Having it here in the UI introduces a maintainability problem long term.

PR #9807, opened 2023-12-23
Teacher Tool: initial rubric, rubricCriteria classes and parsing

kimprice:
This is the initial work for rubrics. We'll probably add an evaluate method which takes in the project data. We can also add meta data to the rubric, but for now it just has the criteria.

Valid JSON for a rubric looks like this:

```JSON
{
    "criteria": [
        {
            "displayText": "One or more custom functions that must be called",
            "criteriaId": "blockCheck",
            "blockRequirements": [
                {
                    "blocks": ["function_call", "function_call_output"],
                    "count": 1
                },
                {
                    "blocks": ["function_def"],
                    "count": 1
                }
            ]
        },
        {
            "displayText": "Two different kinds of loops used",
            "criteriaId": "blockCheck",
            "blockRequirements": [
                {
                    "blocks": ["controls_repeat_ext", "device_while", "pxt_controls_for", "pxt_controls_for_of"],
                    "count": 2
                }
            ]
        },
        {
            "displayText": "Four or more comments",
            "criteriaId": "comment",
            "count": 4
        }
    ]
}
```


kimprice:
I'll wait for Sarah to merge her PR and update accordingly.

pxtblocks/code-validation/rubric.ts
```diff
@@ -0,0 +1,30 @@
+namespace pxt.blocks {
+    export class Rubric {
```
srietkerk, line 2: I like the architecture that you've set up for the rubric and the criteria. It's easy to understand and it will be pretty easy to expand on, too. Nice!

pxtblocks/code-validation/rubric.ts
```diff
@@ -0,0 +1,30 @@
...
+        }
+
+        if (!rubricObj.criteria) {
+            console.error(`No criteria found in rubric`);
+            return null;
+        }
+
+        const criteria: RubricCriteria[] = rubricObj.criteria.map((c: CriteriaData) => {
+            return getCriteria(c);
+        }).filter((r: RubricCriteria) => !!r);
```
srietkerk, line 26: Just wanting to clarify, this filter is here to get rid of any entries that couldn't be processed? 

This might be something that we want to talk about.. I would generally like the sort of system that doesn't fail early if there are entries that are formatted incorrectly, but that might be somewhat frustrating for the rubric case. If a teacher submitted a rubric to our tool with a criteria that isn't defined or formatted correctly, and then they ran that rubric against a student file and only got results from the criteria that we could accept, would that be more work for them? Would they want those results from the few that worked or would it be better to have them re-submit the rubric where everything is formatted correctly?

I don't think there's anything to change right now, but it's just something to keep in mind. This will likely be a more UI-centered solution anyway. Maybe having something where we inform the user that some of the criteria in the rubric couldn't be parsed and ask them if they would like to run the rubric or if they would want to reformat the rubric and resubmit or something like that.

thsparks (approved):
💯

pxtblocks/code-validation/rubricCriteria.ts
```diff
@@ -0,0 +1,56 @@
...
+            case "comment":
+                return new CommentCriteria(data.friendlyText, data.count);
+            default:
+                console.error(`Unrecognized criteriaID: ${data.criteriaID}`);
+                return null;
+        }
+    }
+
+    export abstract class RubricCriteria {
+        friendlyText: string;
```
eanders-ms, line 27: Re `friendlyText`: How will localization work for rubric content? Now I'm curious what we do for tutorials (I don't know off hand). We might want to follow that pattern.

pxtblocks/code-validation/rubricCriteria.ts
```diff
@@ -0,0 +1,56 @@
...
+        abstract criteriaID: string;
```
eanders-ms, line 28: the smallest of nits: use `Id`, not `ID`:

```suggestion
        abstract criteriaId: string;
```

pxtblocks/code-validation/rubricCriteria.ts
```diff
@@ -0,0 +1,56 @@
...
+
+        constructor(friendlyText: string) {
```
eanders-ms, line 30: Would `displayText` be a more descriptive name?

```suggestion
        constructor(displayText: string) {
```

teachertool/src/components/DebugInput.tsx
```diff
@@ -22,6 +24,8 @@ const DebugInput: React.FC<IProps> = ({}) => {
```
eanders-ms, line 24: Can you move `runEvaluate` into a transform (i.e. to a method defined in `src/transforms/runEvaluate.ts`)? We want to keep application logic out of UI components.

pxtblocks/code-validation/rubricCriteria.ts
kimprice: I have no idea...

thsparks: Vaguely, I think the markdown files for tutorials are checked in, then those are parsed and somehow sent for translation? But if these aren't going to be checked into our repo, that may not work.

For teacher-generated rubrics, I don't think translation should be a goal.

For rubrics we provide, I'm wondering if we could just provide different files/links for the different languages, then teachers select the ones that are most appropriate for their students? It depends on how we end up storing the rubrics, I suppose. I wonder if our existing curriculum content would be a better pattern to follow, rather than tutorials. Seems like that might be a slightly more similar scenario?

pxtblocks/code-validation/rubric.ts
kimprice: Yea, all good points. My thought process was that we would be formatting the rubric behind the scenes based on the teacher's selections in the UI - so there shouldn't be any rubric criteria that is ill-formatted. But yes, in the future we would want some kind of indication if any of the rubric criteria wasn't successfully parsed.

teachertool/src/components/DebugInput.tsx
```diff
@@ -1,27 +1,19 @@
...
-            console.log("Invalid share link!");
-            return;
-        }
-        const projText = await getProjectTextAsync(scriptId);
-        if (projText) console.log(projText["main.blocks"] || "Failed to get blocks xml!");
-
-        const projMeta = await getProjectMetaAsync(scriptId);
-        if (projMeta) console.log(projMeta);
+    const evaluate = async () => {
+        runEvaluate(shareLink, rubric);
```
eanders-ms, line 16: Needs await

teachertool/src/transforms/runEvaluate.ts
```diff
@@ -0,0 +1,17 @@
+import { getProjectMetaAsync, getProjectTextAsync } from "../services/BackendRequests";
+
+export async function runEvaluate(shareLink: string, rubric: string) {
```
eanders-ms, line 3: Per convention, append `Async` to async methods.

```suggestion
export async function runEvaluateAsync(shareLink: string, rubric: string) {
```

teachertool/src/components/DebugInput.tsx
```diff
@@ -1,27 +1,19 @@
+/// <reference path="../../../built/pxtblocks.d.ts"/>
+
 import { useState } from "react";
 import { Button } from "react-common/components/controls/Button";
 import { Input } from "react-common/components/controls/Input";
 import { Textarea } from "react-common/components/controls/Textarea";
-import { getProjectMetaAsync, getProjectTextAsync } from "../services/BackendRequests";
+import { runEvaluateAsync } from "../transforms/runEvaluate";
```
eanders-ms, line 7: Can you please rename the file to match the function name? `runEvaluateAsync.ts`

pxtblocks/code-validation/rubricCriteria.ts
```diff
@@ -0,0 +1,56 @@
...
+            this.displayText = displayText;
+        }
+    }
+
+
+    class BlockCheckCriteria extends RubricCriteria {
+        criteriaId: "blockCheck";
+        blockRequirements: BlockSet[];
+
+        constructor(displayText: string, blockRequirements: BlockSet[]) {
```
eanders-ms, line 40: Not for this PR, but member assignment could be simplified by declaring class variables inline in the constructor, e.g.:
```suggestion
        constructor(displayText: string, public blockRequirements: BlockSet[]) {
```

eanders-ms (approved):
Looks good!

PR #9809, opened 2024-01-03
Move npm prepare to a script

eanders-ms:
Moved for maintainability. It was getting very long and unwieldly.

### DIFF
--- a/scripts/npm-prepare.sh
+++ b/scripts/npm-prepare.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+cd skillmap && npm install --no-update-notifier && cd ..
+cd authcode && npm install --no-update-notifier && cd ..
+cd multiplayer && npm install --no-update-notifier && cd ..
+cd kiosk && npm install --no-update-notifier && cd ..
+cd teachertool && npm install --no-update-notifier && cd ..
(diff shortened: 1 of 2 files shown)


thsparks (approved):

scripts/npm-prepare.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+cd skillmap && npm install --no-update-notifier && cd ..
+cd authcode && npm install --no-update-notifier && cd ..
```
srietkerk, line 4: Is this something that we can possibly set in an npm config? I looked up what the flag meant and it looks like it's also a variable that can be set to false in the npm config. I found that information from this Stack Overflow post, so it might be outdated: https://stackoverflow.com/questions/46800717/eliminate-npm-update-available-message. If it's possible, it might be nice to go that route instead of including the flag each time.

eanders-ms: The notice was being printed for every entry here, very repetitive. It might still be useful to see it once at the top level call to `npm install`.

srietkerk: Right, that makes sense. I was more saying to reduce repetition in the script itself, we could maybe move the `no-update` flag into an npm config, but I'm not sure if that is doable/would be applied to everyone's device if we put it in an npm config.

PR #9810, opened 2024-01-03
Stop using PouchDB

riknoll:
This PR deprecates the old PouchDB database in favor of just using indexedDB directly. PouchDB has caused us a lot of trouble, particularly on ios and minecraft, and there isn't much reason for us to use it anymore now that indexedDB is well supported across the browsers we target.

The things we store in PouchDB include:
* headers and projects
* cached github repos
* cached shared projects we downloaded
* cached hexfiles
* headers and projects for previous editor versions (i.e. for targets that use browser prefixes like micro:bit, but not arcade)

All of this was stored in one giant table.

The new structure has separate tables for each of those things, and separate DBs for each browser prefix (i.e. a v1 editor and v2 editor will have completely separate databases). When you open a new major version of an editor everything from the previous major version's DB is copied over to the new DB, same as before.

I migrate projects from the old PouchDB by reading directly from the IndexedDB it uses. I did it that way so that we wouldn't have to use PouchDB at all and I figured out the structure by looking at the browser storage in the dev tools. This definitely needs to be well tested on other platforms to make sure that this structure is consistent cross-platform.

This does delete the PouchDB contents after migration is complete, but only if the migration is successful. The old pouchDB workspace is still accessible via a query parameter, however new projects will get migrated if you then visit the editor without the query parameter.


webapp/src/idbworkspace.ts
```diff
@@ -17,47 +11,166 @@ interface StoredText {
...
+async function getDbAsync(prefix = "__default") {
+    if (_dbPromises[prefix]) return _dbPromises[prefix];
+
+    _dbPromises[prefix] = createDbAsync();
+
+    return _dbPromises[prefix];
 
     async function createDbAsync(): Promise<pxt.BrowserUtils.IDBWrapper> {
-        const idbDb = new pxt.BrowserUtils.IDBWrapper("__pxt_idb_workspace", 1, (ev, r) => {
+        const idbDb = new pxt.BrowserUtils.IDBWrapper(`__pxt_idb_workspace_${pxt.storage.storageId()}_${prefix}`, 1, (ev, r) => {
```
eanders-ms, line 167: Just wondering, why was it necessary to append these things to the db name?

eanders-ms (approved):
Terrifying, but LGTM.

webapp/src/idbworkspace.ts
riknoll: the storage id is just for localhost so that different local served targets don't use the same db (we did the same thing for pouch).

the prefix is just to separate the different browser prefix versions of storage. in micro:bit we maintain separate dbs for different major versions of the editor (e.g. v0, v1, v2, etc.)

abchatra:
Can we get rid of npm dependency on pouchdb?

riknoll:
@abchatra it's still there for two reasons:

1. You can still access the pouchDB using a query param (for debug purposes)
2. In micro:bit if a project from an earlier major version has errors, we show a button to open it in the older version of the editor. In order for that to work I need to copy the header into the old DB, which requires using pouchDB

It's not used beyond those two scenarios

abchatra:
Sounds good. One other testing thought I had.

You could write a simple comparison validator in debug flag which runs after the migration is done to ensure byte by byte copy is correct and then delete the old db entries. 

Also what happens if the migration fails and it runs again next time? Is the process idempotent?

riknoll:
@abchatra if the migration fails, it will just try again next time. I check for existing entries before writing new ones so there's no chance of accidentally overwriting anything.

I'm not particularly worried that anything will get corrupted during the actual copy (i'm literally just reading it from one db and immediately writing it to the other), so i'm not sure that comparison validator is necessary

abchatra:
Sounds good.

webapp/src/browserworkspace.ts
```diff
@@ -121,20 +121,13 @@ function setCoreAsync(headers: db.Table, texts: db.Table, h: Header, prevVer: an
     return headerRes
 }
 
-export function copyProjectToLegacyEditor(h: Header, majorVersion: number): Promise<Header> {
+export async function copyProjectToLegacyEditor(header: Header, script: pxt.workspace.ScriptText, majorVersion: number): Promise<void> {
```
kimprice, line 124: should this function name end in 'async'?

riknoll:
@eanders-ms @abchatra @kimprice I modified this PR to no longer delete the old entries in the pouchDB, instead it just marks them as migrated. This is to prevent you from losing your projects in previous versions of the editor (e.g. v1 of microbit) by visiting a more recent version.

PR #9811, opened 2024-01-04
Teacher tool: set up infrastructure to send messages to iframe

srietkerk:
First off, I just wanted to give a HUGE thanks to @eanders-ms and @thsparks for the help with this PR. Your explanation of infrastructure and review of how the code works was crucial to my understanding for what it will look like when the pieces fit together. 

**Important**

- The changes that I made in `DebugInput.tsx` are temporary and should be removed. I wanted to leave the logic in to reflect a proposed rudimentary flow for the app. The transform function should just take a string (probably, but maybe something else would work better?) and should hook into state so that the blocks data that we get back from the iframe is stored in state.
- I wasn't really sure what we would want that part of state to look like, so the transform isn't connected to state yet.
- I wanted to get this PR out because the key logic is there, and I think it will attach nicely to work that Kim and Thomas are doing currently. 

**Some clarifying notes**
All of the logic in the iframe for handling messages has been extracted into a service that is accessed by both the `MakeCodeFrame` UI to hook into the iframe and the transform so that the app can communicate to the iframe through the service. This is why the `MakeCodeFrame` can be so clean. For the future work of making the `MakeCodeFrame` into an independent entity, it will be bundled with the `makecodeEditorService. 

There are a lot of things that I was able to remove from the iframe logic. As noted, @thsparks helped me realize that a lot of the logic that still lived in the `MakeCodeFrame` was logic that was specific to skillmap, so those parts were cut out. This includes the `frameState` and it's associated `useEffect` and the switch statement in `onMessageReceived`. Thomas helped me better understand the logic of how sending messages works, and because of this, I realized that the switch statement wasn't really doing anything. All of the logic that's important for our use case is everything that happens in the condition in `onMessageReceived`. This will handle events regardless of their types, and really helps the `MakeCodeFrame` be super generic.  If a parent app needs special logic, I think a transform would be able to handle that.


srietkerk:
Also, I made an upload target that should technically have the changes that I made, but I have no idea how to access eval from the upload target.

If anyone knows how to get there, here's the upload target: https://arcade.makecode.com/app/4c7378997bc31a422e9822167822e2f35982712b-45cc08b47b

eanders-ms:
Eval tool should be accessible here: https://arcade.makecode.com/app/4c7378997bc31a422e9822167822e2f35982712b-45cc08b47b--eval, but we need to update the backend to serve this endpoint.

teachertool/src/services/makecodeEditorService.ts
```diff
@@ -0,0 +1,52 @@
...
+            messageQueue.push(message);
+            while (messageQueue.length) {
+                sendMessageCore(messageQueue.shift());
+            }
+            if (message) sendMessageCore(message);
+        }
+    });
+}
+
+export const getIframeRef = (ref: HTMLIFrameElement) => {
```
eanders-ms, line 43: My recommendation is to combine `getIFrameRef` and `removeIFrameRef` into single method: `function setEditorRef(ref: HTMLIFrameElement | undefined)`. The method would unhook from any existing iframe, then if `ref` is not undefined, hook into the new one.

teachertool/src/services/makecodeEditorService.ts
```diff
@@ -0,0 +1,52 @@
...
+    const data = event.data as pxt.editor.EditorMessageRequest;
+    if (data.type === "pxteditor" && data.id && pendingMessages[data.id]) {
+        const pending = pendingMessages[data.id];
+        pending.handler(data);
+        delete pendingMessages[data.id];
+        return;
+    }
+}
+
+export const sendMessageAsync = (message?: any) => {
```
eanders-ms, line 21: We shouldn't expose `sendMessageAsync`. It is too low level. Instead, create methods for specific messages, like `setHighContrast(on: boolean)`. For methods that expect a reply, return a Promise (strongly typed to the reply message). The obvious benefit to this approach is that it keeps application code very clean. Another is that within makecodeEditorService, you can decide which messages need to be kept in PendingMessages. Only the messages that expect a response need to be tracked this way (though you would need another queue to pend messages sent before the iframe is ready, but I think you need this anyway).

teachertool/src/services/makecodeEditorService.ts
```diff
@@ -0,0 +1,52 @@
...
+            while (messageQueue.length) {
+                sendMessageCore(messageQueue.shift());
+            }
+            if (message) sendMessageCore(message);
+        }
+    });
+}
+
+export const getIframeRef = (ref: HTMLIFrameElement) => {
+    makecodeEditorRef = ref;
```
eanders-ms, line 44: When the iframe is set, if there are pending messages in the queue, will they be sent to it?

teachertool/src/services/makecodeEditorService.ts
```diff
@@ -0,0 +1,52 @@
...
+let makecodeEditorRef: HTMLIFrameElement | undefined;
+const messageQueue: pxt.editor.EditorMessageRequest[] = [];
+let nextId: number = 0;
+let pendingMessages: {[index: string]: PendingMessage} = {};
+
+const onMessageReceived = (event: MessageEvent) => {
+    const data = event.data as pxt.editor.EditorMessageRequest;
+    if (data.type === "pxteditor" && data.id && pendingMessages[data.id]) {
+        const pending = pendingMessages[data.id];
+        pending.handler(data);
```
eanders-ms, line 15: Are there any messages the editor might send unbidden? As it is now, it seems `onMessageReceived` can only handle replies.

teachertool/src/services/makecodeEditorService.ts
```diff
@@ -0,0 +1,52 @@
...
+            message.response = true;
+            message.id = nextId++ + "";
+            pendingMessages[message.id] = {
+                original: message,
+                handler: resolve
+            };
+            makecodeEditorRef!.contentWindow!.postMessage(message, "*");
+        }
+
+        if (makecodeEditorRef) {
```
eanders-ms, line 33: I'm not sure this check is sufficient to ensure the message doesn't get sent to the editor before it is ready.

teachertool/src/components/MakecodeFrame.tsx
```diff
@@ -1,140 +1,31 @@
...
+        if (frameRef.current) {
+            getIframeRef(frameRef.current);
         }
 
-        // logic we want when the iframe unmounts
         return () => {
-            window.removeEventListener("message", onMessageReceived);
-            if (ref && ref.contentWindow) {
-                ref.removeEventListener("load", handleFrameReload)
+            if (frameRef.current) {
```
eanders-ms, line 21: This check is unnecessary. It should be safe to always call `removeIFrameRef` here.

teachertool/src/components/MakecodeFrame.tsx
```diff
@@ -1,140 +1,31 @@
...
-    let ref: HTMLIFrameElement | undefined;
-    const messageQueue: pxt.editor.EditorMessageRequest[] = [];
-    let nextId: number = 0;
-    let pendingMessages: {[index: string]: PendingMessage} = {};
-    const [frameState, setFrameState] = useState<FrameState>("loading");
-    const [loadPercent, setLoadPercent] = useState(0);
-    const [workspaceReady, setWorkspaceReady] = useState(false);
+    ( { pageSourceUrl} ) => {
 
     let frameRef = useRef<HTMLIFrameElement>(null);
```
eanders-ms, line 13: I strongly recommend using a method to receive the ref, like this:
```
const handleIFrameRef = (el: HTMLIFrameElement | null) => {
    EditorService.setEditorRef(el ?? undefined)
}
```

then:
```
<iframe ... ref={handleIFrameRef}></iframe>
```


This creates an airtight coupling in case the ref changes at any point during rerender.

thsparks, line 13: I have mixed feelings on this. On the one hand, passing the ref does keep all the logic very contained within the service, but hanging onto a ref like that feels less safe to me than passing a callback to send a message to the iframe instead. The function passed would essentially just be something like:

```ts
function sendMessageToIFrame(message) {
   frameRef!.contentWindow!.postMessage(message, "*");`)
}
```

Then the service calls into that instead of calling `makecodeEditorRef!.contentWindow!.postMessage(message, "*");`

thsparks: Does using React's useRef handle this?

teachertool/src/components/MakecodeFrame.tsx
```diff
@@ -1,140 +1,31 @@
...
 
     useEffect(() => {
-        // logic we want to do when the iframe is loaded
-        if (ref && ref.contentWindow) {
-            window.addEventListener("message", onMessageReceived);
-            ref.addEventListener("load", handleFrameReload)
-
+        if (frameRef.current) {
```
thsparks, line 16: If `frameRef.current` is unset here for whatever reason, should we call `removeIframeRef`?

teachertool/src/components/EditorContainer.tsx
```diff
@@ -22,10 +22,7 @@ export const EditorContainer: React.FC<{}> = () => {
     }
```
thsparks, line 22: I think you can remove `onIframeLoaded` and `onIframeClosed` now, too

thsparks: (And possibly move createIFrameUrl into a util function)

teachertool/src/services/makecodeEditorService.ts
```diff
@@ -0,0 +1,52 @@
...
+            messageQueue.push(message);
```
thsparks, line 34: If I understand the goal correctly, we want the messageQueue to be used to store messages that came in before the iframe was ready. With that in mind, I think we want to push to the queue when the iframe _is not_ ready, then clear the queue when it is. I think the code as written would just send the message twice when the iframe is ready and then lose any messages that came in while the iframe was not ready.

teachertool/src/services/makecodeEditorService.ts
thsparks: Agree. We'll probably need to listen for some "loaded" event (not sure what that is yet) and then set a flag, rather than checking the ref directly.

teachertool/src/services/makecodeEditorService.ts
thsparks: +1 this will keep all of the "nitty gritty" details (communicating with the iframe, what messages to send, etc...) contained within this class, then the rest of the code doesn't have to worry about it.

teachertool/src/services/makecodeEditorService.ts
srietkerk: If there are unbidden messages, are they things that we need to handle? Also, if there are messages that don't need give a response, is there anything else that we need to do? I don't know what those scenarios would look like.

teachertool/src/components/EditorContainer.tsx
srietkerk: I think you're right. With moving these things, I'm thinking that we actually won't need the editor container at all, then, and can just plug the `MakeCodeFrame` straight into the app.

teachertool/src/components/MakecodeFrame.tsx
srietkerk: This way is a bit confusing to me. If we have the function be the ref for the iframe, the function wouldn't have anything as an argument so the argument for `setEditorRef` would always be undefined, wouldn't it?

srietkerk: Hmm, interesting. So you would suggest to have a function in the service of, let's say, `connectToFrame(func: () => void)`, save that passed-in function to something inside the service, and then call that function when we need to send a message to the iframe. Am I following that correctly?

I kinda like this infrastructure. However, following this can get kinda convoluted. Because if we don't have the ref somewhere in the service, then we would need to have the adding and removing event listeners in the `MakeCodeFrame`. This is fine, but that would then mean that we also need to expose `onMessageReceived` from the service for the iframe to use, if we want to keep the `onMessageReceived` logic in the service. So, in the iframe, you would be sending, to the service, a way for the service to communicate to the iframe and receiving, from the service, a function for what to do when messages are received. 

I think this infrastructure could work, but I'm not a huge fan of the circular logic that's involved. I like just having one entry point for the service and iframe to communicate.

teachertool/src/components/MakecodeFrame.tsx
srietkerk: Sure, I don't see a harm in doing that at least.

teachertool/src/services/makecodeEditorService.ts
srietkerk: Is there a benefit of doing `undefined` over `null` here? I find that passing `undefined` into a function bizarre and `null` to be more natural for some reason.

The other option we could do is have ref be an optional parameter and if there is nothing passed into the function, we just unset it, but I think going with a null set is more natural to me.

abchatra: I don't know the context here. But throwing out the difference between null & undefined spec wise.

Undefined is a non-availability of the property (expando) on an object, Null is non availability of a value for the property on an object.

let abc = {}
console.log(abc.xyz === undefined) //true as property doesn't exist
console.log(abc.xyz === null) //false as this is strict equals
console.log(abc.xyz == undefined) //true
console.log(abc.xyz == null) //true

abc.xyz = null;

console.log(abc.xyz === undefined) //false as peroperty exists but not the false
console.log(abc.xyz === null)  //true
console.log(abc.xyz == undefined) //true
console.log(abc.xyz == null) //true

Lots of people use it interchangeably and that is ok. == resolves to same anyways.

.

teachertool/src/components/EditorContainer.tsx
srietkerk: I'm going to do this. I think any logic we'll possibly want to plug into the frame, it should be handled in the transform or the service.

teachertool/src/services/makecodeEditorService.ts
thsparks: An example of an "unbidden" message would be one for when the workspace & content has loaded in the iframe, it'd be nice for it to send us a message saying so without us having to poll for it (I'm trying to do this now).

That said, I think this function would still receive those messages? We'd just need to make sure we handle the ones we're expecting even if they aren't in the pending messages queue. Since we don't currently have any we expect, I'm not sure if there's much to do here just yet. (Unless I'm missing something 🙂)

teachertool/src/services/makecodeEditorService.ts
```diff
@@ -0,0 +1,52 @@
...
+            }
+            if (message) sendMessageCore(message);
+        }
+    });
+}
+
+export const getIframeRef = (ref: HTMLIFrameElement) => {
+    makecodeEditorRef = ref;
+    window.addEventListener("message", onMessageReceived);
+    makecodeEditorRef.addEventListener("load", () => {console.log("loaded")});
```
thsparks, line 46: We could probably remove this, if all it's doing is a console log.

teachertool/src/components/MakecodeFrame.tsx
thsparks: It seems like the event listener is actually on the window, not the iframe ref, so I'm wondering if it's actually necessary to have the ref for that. If we care about the iframe loaded event, then you're right, the component would have to bubble that up in addition to the "sendMessageToIFrame" function it sends (or it could only send that callback when the iframe is ready).

I don't think it makes a huge difference either way, as long as we're careful to be resilient against the ref changing. One of Eric's comments suggests another approach to doing that too, which might also work.

teachertool/src/services/makecodeEditorService.ts
srietkerk: Makes sense, thanks. I don't really see a case where we'll need to do an "equals" comparison, just a truthy check. Since undefined and null are both falsey, I think using null will be fine.

teachertool/src/components/MakecodeFrame.tsx
thsparks: This is using some React functionality called a "callback ref". When the ref is set this way, React will automatically call the function you provide with the `HTMLIFrameElement` parameter set to the actual element in question. So in this case, `el` would point to the iframe node, and that's what you'd end up passing into `setEditorRef`.

https://legacy.reactjs.org/docs/refs-and-the-dom.html#callback-refs

teachertool/src/services/makecodeEditorService.ts
srietkerk: Oh, thank you for that example, that makes sense. I do think that the unbidden case, then, should be handled when those cases arise. Or, I guess we could just have a notification dispatched saying that there was a message from the editor? Being that I'm not sure _what_ we'd want to do with those kinds of messages, I did nothing for now. The nice thing, though, is that it can be added easily. I am also unsure if there's much to do here currently, but I can add some default behavior like dispatching a notification or log something if it would be valuable for testing.

teachertool/src/components/MakecodeFrame.tsx
srietkerk: Oh, cool, thanks!

teachertool/src/services/makecodeEditorService.ts
thsparks: I would lean towards undefined in this case for two reasons:
1. It's more consistent with the rest of this code base
2. In some cases (not sure about this one specifically) "null" is a valid value for certain types/variables, whereas "undefined" specifically exists to indicate that no value exists. It's a little weird conceptually, but the idea is the difference between "there is a value, the value is null" and "there is no value set". In this case you might be able to get away with either approach, but then see #1. When something can go one way or another, I would tend towards staying consistent with the rest of the codebase. Makes things more predictable and easier to understand.

teachertool/src/components/MakecodeFrame.tsx
srietkerk: Oh, that's a good point. I feel like a combination of your suggestion and Eric's suggestion would be really solid, but I might be misunderstanding something, and TypeScript might also not like the fact that the "ref" variable wouldn't be used. Let me know what you think.

Basically, we have the `handleIframeRef` function that Eric suggested and, instead of setting the ref for the service in the body of that function, we send the message function that you suggested through to the service. This way, we have a really solid way of knowing that the iframe is initialized and we don't have to pass a ref to the service.

teachertool/src/services/makecodeEditorService.ts
eanders-ms: Let's log unhandled messages to the console. See what surfaces.

teachertool/src/services/makecodeEditorService.ts
eanders-ms: I think we'll need to do one of:
A. queue the messages until the iframe is ready to receive them.
B. during app init, keep the UI in a loading state until the iframe is ready.

teachertool/src/services/makecodeEditorService.ts
```diff
@@ -40,13 +40,13 @@ export const sendMessageAsync = (message?: any) => {
...
-    makecodeEditorRef = ref;
-    window.addEventListener("message", onMessageReceived);
-    makecodeEditorRef.addEventListener("load", () => {console.log("loaded")});
-}
-
-export const removeIframeRef = () => {
-    makecodeEditorRef = undefined;
-    window.removeEventListener("message", onMessageReceived);
+export const setEditorRef = (ref: HTMLIFrameElement | null) => {
+    if (ref) {
```
eanders-ms, line 44: If `makecodeEditorRef` is already set, it needs to be torn down. Listeners removed, etc.

teachertool/src/services/makecodeEditorService.ts
eanders-ms: > Is there a benefit of doing `undefined` over `null` here? I find that passing `undefined` into a function bizarre and `null` to be more natural for some reason.
 

The project has strict null checks enabled, so TypeScript will complain when sending a value of null to a parameter that takes a value or undefined. As a convention, I prefer undefined over null and will exchange null for undefined as quickly as possible when there isn't a meaningful distinction given the use case. In this case the callback ref will receive an HTML element or null. Following that convention, I switch null to undefined before passing it to our systems.

> The other option we could do is have ref be an optional parameter and if there is nothing passed into the function, we just unset it, but I think going with a null set is more natural to me.

Optional parameter is fine too (it resolves to undefined), but calling `setEditorRef` without a parameter reads funny.

teachertool/src/components/MakecodeFrame.tsx
thsparks: I think if you're going with the handleIframeRef approach, you can just pass the ref around. Passing the function instead of the ref would probably be a little redundant at that point...but Eric's approach may be better in this case, since the service would have a clearer sense for when the iframe gets cleared for whatever reason.

teachertool/src/services/makecodeEditorService.ts
thsparks: I think we'd want to do both A and B, ideally

thsparks: Though B could be done in a separate change, IMO

teachertool/src/components/MakecodeFrame.tsx
eanders-ms: > Does using React's useRef handle this?

The ref will receive the updated value, but it won't propagate anywhere. For propagation, the defined `useEffect` would need to be triggered when the ref changes. You might think this could be done by adding ref.current as a dependency, but this is problematic due to the ordering of updates in the component. ref values are updated after effect dependencies are evaluated. Because of this ordering, the React team considers refs in dependency lists an anti-pattern and it will be flagged by react/exhaustive-deps.

Solution is to use a callback ref (thanks for the doc link!). Here's a similar example from the React docs: https://legacy.reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node

And I see they're memoizing the callback ref with `useCallback`. Nice optimization. We could safely do that.

teachertool/src/services/makecodeEditorService.ts
thsparks: For a more "sufficient" check, we may need the event I'm working on adding to detect when the editor is done loading. If that's the case, I wouldn't block on this. I'm building off of your changes, so I can add it once you check in if you're otherwise ready to go. If you're still working on other things though, I can try to separate it out and send it in a new PR. (I probably should've done that to begin with, sorry...)

teachertool/src/services/makecodeEditorService.ts
srietkerk: So, Thomas had a good point that the event listeners are on the window rather than the on the ref.. Would those listeners still need to be removed? 

Also, if we do this, it kinda sounds like we should just set `makeCodeEditorRef` to undefined each time we set the editor ref. As in, the first thing we do in the function is `makecodeEditorRef = undefined` since no matter if there's a ref passed in or not, we'll want the `makecodeEditorRef` to be clean.

teachertool/src/services/makecodeEditorService.ts
srietkerk: So the suggestion here is that if `pendingMessages` is not empty, flush the message queue to the frame when we get a new ref?

teachertool/src/services/makecodeEditorService.ts
srietkerk: Yeah, I was mulling on this and waiting to respond because I kinda figured that was the case. I also think that the if statement is a pretty weak protection, but I don't know if there's anything right now that I can do to make this better.

Maybe we could also have some sort of async function to wait for the editor to load? But that sounds a bit messy to me.

teachertool/src/services/makecodeEditorService.ts
srietkerk: Makes sense. I just did some testing with this. The frame gets ready pretty quickly, so there's not much wait time, but I put a five second timeout before attaching the ref to the service and was able to stack up a lot of requests. I just wanted to double check that what the desired clear mechanism is. Because with the playing that I did within `sendMessageAsync`, all of the message wouldn't be processed until another message was sent to the iframe. I think this would be the job of an event listener that we were chatting about before. So we have the window listen for when the frame is ready, and when it is, send the messages over.

srietkerk: Kinda related to this work stream, I don't see the value in `if (message) sendMessageCore(message)` it seems like it duplicates calls..  We push the latest message to the queue and then clear the queue, so shouldn't the most recent message already have called `sendMessageCore()`?

teachertool/src/services/makecodeEditorService.ts
thsparks: Yup (unless I misunderstood Eric's original comment)! You can see skillmap doing the same thing here, after it sets the workspace to ready: https://github.com/microsoft/pxt/blob/master/skillmap/src/components/makecodeFrame.tsx#L180

teachertool/src/services/makecodeEditorService.ts
thsparks: That's the reason behind my note that this would just send the message twice, which is not what we want. I think you could keep `if (message) sendMessageCore(message)` but remove the part where it adds it to the queue first. So you clear the queue (without adding this message), then send this message.

As for your note about "all of the message wouldn't be processed until another message was sent to the iframe", I believe that's the reasoning behind this comment: https://github.com/microsoft/pxt/pull/9811#discussion_r1442057193 Basically, we want to flush the queue the moment we learn that the iframe is ready, in addition to anytime a new message is sent.

teachertool/src/services/makecodeEditorService.ts
srietkerk: I think we might need a rename of `pendingMessages` because I was a bit confused by your question, but I think that you're meaning that if there were any messages sent to the iframe before the iframe was loaded, we would flush those messages once the iframe's loaded. Is that right? Thomas' reply below and some experimenting gave me a bit more  context on this. The "pending messages" to the iframe will exist in the `messageQueue`, or at least should be based on the conversation we have/had in a thread below. Pending messages definitely can get sent to the iframe once it is loaded.  I think this ties into the stream of what should we do when the iframe is ready.

teachertool/src/services/makecodeEditorService.ts
srietkerk: Right, makes sense. 

I think I would prefer to keep the pushing to the queue logic and get rid of that `if (message)...` stuff because of this comment of yours:

>With that in mind, I think we want to push to the queue when the iframe is _not_ ready, then clear the queue when it is.

Because I can move `messageQueue.push(message)` outside of the `if (makecodeEditorRef)` check and allow for pushing messages whether or not the iframe is ready. This way, the logic stays consistent. When the iframe is ready, we can flush the queue with another function/event listener and when the iframe is ready. Separately, when a message is sent, the queue will just get cleared immediately. 

I guess the possible problem here would be a race condition, though, in the queue.

teachertool/src/services/makecodeEditorService.ts
thsparks: Assuming the intention is to match skillmap's behavior, there are two distinct queues here:

`pendingMessages` is the list of messages that _have already been sent_ to the editor and we're waiting for a response
`messageQueue` is the list of messages that _have not yet been sent_ and are waiting for the editor to be ready to receive messages before we send them.

I think we want to clear the `messageQueue` here but not `pendingMessages`. The terminology is a bit confusing, yeah. Perhaps we can think of better names...

teachertool/src/services/makecodeEditorService.ts
thsparks: Spoke offline, but to reiterate here: I don't think we really need to worry about race conditions in this scenario. Your proposed approach seems fine, as long as we still make sure to send all pending messages once the iframe is ready and before any new messages get sent.

thsparks: Oh but small gotcha: make sure you only add the message to the queue if it's not undefined. (In skillmap, at least, calling sendMessageAsync with undefined clears the queue without sending a new message. If we're following that pattern, that case is important to account for)

teachertool/src/services/makecodeEditorService.ts
```diff
@@ -0,0 +1,62 @@
...
+        makecodeEditorRef = ref;
+        window.addEventListener("message", onMessageReceived);
+    } else {
+        makecodeEditorRef = undefined;
+        window.removeEventListener("message", onMessageReceived);
+    }
+}
+
+//  an example of events that we want to/can send to the editor
+export const setHighContrastAsync = async (on: boolean) => {
```
eanders-ms, line 55: Just wondering, why not make these functions?
```
export async function setHighContrastAsync(...)
```

kimprice: What's the difference between doing it this way and declaring it as a function like you suggested? Is it just styling or are there some kind of logical differences?

thsparks: I had the same question. Depending on how much you trust copilot, it seems like `export async function setHighContrastAsync(...)` would be more fitting to this scenario:

image

teachertool/src/services/makecodeEditorService.ts
```diff
@@ -0,0 +1,62 @@
...
+                sendMessageCore(messageQueue.shift());
+            }
+        }
+    });
+}
+
+export const setEditorRef = (ref: HTMLIFrameElement | undefined) => {
+    if (ref) {
+        makecodeEditorRef = ref;
+        window.addEventListener("message", onMessageReceived);
```
thsparks, line 47: I thought there was a comment on this already but maybe it's considered outdated now. Just to reiterate, we should clear the message queue here now that the ref is set (which, as far as this change is concerned, is indicating that the iframe is "ready")

thsparks: If following skillmap's convention, you can do that by calling sendMessageAsync with no parameter passed in, but you'll need to update your sendMessageAsync function to handle the message being undefined (just a check before adding it to the queue)

teachertool/src/services/makecodeEditorService.ts
eanders-ms: Regular functions are generally preferred as there is more flexibility around when they can be called (you can call them from a location in the file above where they're declared).

I generally try to use regular functions unless there's a specific reason otherwise.

A few scenarios where I've found function expressions useful:
1. You need to determine at runtime which implementation of a function to make active (function assignment)
2. Similarly, you want to reassign a function after its been called, so that it can perform its original purpose only once.
3. In classes, functions as member variables are useful when passing the function as an event handler (e.g. `this.handleButtonClick`) because they're automatically bound to the class's `this` pointer, obviating the need to call `bind`. When passing a member function, the value `this` when the handler is called might be the calling context rather than the containing class!

thsparks (approved):
A few nits (clearing message queue when iframe is ready, maybe remove some of the debug code, and I agree with Eric's function declaration comment) but on the whole LGTM 🎉

eanders-ms (approved):
LGTM as well! Thanks for getting this in place Sarah! We can iterate on this. I'm not sure we'll want to clear the message queue when we set the ref. We may want to send it messages while it's loading. Let's use it and see.

teachertool/src/services/makecodeEditorService.ts
srietkerk: Ah, those are good to know, thanks!

I didn't realize you had those benefits. I like having functions defined as anonymous functions because it helps me remember that functions are first-class citizens in TypeScript. I will change all of these to declarative functions.

PR #9813, opened 2024-01-05
Send Notification to Parent Frame when Editor Content has Loaded

thsparks:
This change adds an event that gets sent to the parent when the editor content loads. We need this for the teacher tool app, so we know once the project has finished loading inside the iframe.


kimprice (approved):
LGTM! :)

srietkerk (approved):
I don't know much about our messaging system, but this looks standard with everything else, so this looks good to me!

eanders-ms (approved):
LGTM!

PR #9820, opened 2024-01-09
Fix pouch db migration edge cases

riknoll:
This fixes a couple of pouch db migration bugs.

First, I was writing back to the pouchDB db to mark each entry as migrated so that we didn't accidentally re-import them. This was causing problems due to the internal structure of the pouchDB which has some uniqueness constraints that were being violated. Now I just sidestep that issue by storing the ids of the projects that have already been migrated in a separate db. This fixes the issue we were seeing with github projects in testing.

Secondly, I now check to see if the pouchDB actually exists before doing any migrating. This fixes an issue that prevented accessing old versions of the editor if you were in a browser with no pre-existing pouchDB. Figuring out if the db already exists was a little tricky; see the comment in the function for an explanation.

### DIFF
--- a/webapp/src/idbworkspace.ts
+++ b/webapp/src/idbworkspace.ts
@@ -48,14 +48,42 @@ async function performMigrationsAsync() {
     return _migrationPromise;
 }
 
+const POUCH_OBJECT_STORE = "by-sequence";
+const POUCH_DB_VERSION = 5;
+
+async function checkIfPouchDbExistsAsync() {
+    // Unfortunately, there is no simple cross-browser way to check
+    // if an indexedDb already exists. This works by requesting the
+    // db with a version lower than the current version. If it
+    // throws an exception, then the db must already exist with the
+    // higher version. If it tries to upgrade, then the db doesn't
+    // exist. We abort the transaction to avoid poisoning pouchdb
+    // if anyone ever visits an old version of the editor.
+    let result = true;
+    try {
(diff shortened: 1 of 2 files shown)


webapp/src/idbworkspace.ts
```diff
@@ -303,6 +334,29 @@ export async function copyProjectToLegacyEditorAsync(header: Header, script: pxt
...
+    const idbDb = new pxt.BrowserUtils.IDBWrapper(`__pxt_idb_migration_${pxt.storage.storageId()}`, 1, (ev, r) => {
+        const db = r.result as IDBDatabase;
+        db.createObjectStore(TEXTS_TABLE, { keyPath: KEYPATH });
+        db.createObjectStore(HEADERS_TABLE, { keyPath: KEYPATH });
+        db.createObjectStore(SCRIPT_TABLE, { keyPath: KEYPATH });
+        db.createObjectStore(HOSTCACHE_TABLE, { keyPath: KEYPATH });
+        db.createObjectStore(GITHUB_TABLE, { keyPath: KEYPATH });
+    }, async () => {
+        await pxt.BrowserUtils.clearTranslationDbAsync();
+        await pxt.BrowserUtils.clearTutorialInfoDbAsync();
```
thsparks, line 347: Why are we clearing the TutorialInfoDb here?

thsparks: Oh, I see. This is just to free up space if needed...

thsparks (approved):

PR #9821, opened 2024-01-10
Have Teacher Tool Request Iframe to Run Evaluation for Specified Share Link

thsparks:
### At a high level, this PR does the following:
1. Connects the iframe to the share link that has been provided (except in localhost, because cross-domain localhost <-> makecode.com messaging doesn't work)
2. Saves project metadata and rubric info in app state
3. Adds a request to the iframe to get evaluation results. This request includes the rubric json as a parameter.

### A few notable points:
**Logging Service**
I added a logging service for convenient, consistent output and telemetry. There's `LogError`, which sends a tick event and does a console error output, `LogInfo` which does a tick event and non-error console output, and `LogDebug`, which sends console output but only if you're running the app locally.

**State in MakeCode Frame**
I went back and forth on this, but in the end, in order to be consistent with our other component usage, I pushed the state for the project metadata into the `MakecodeFrame`. I can pull this out and have it passed in as props instead if desired, but I wasn't sure if there was all that much value in doing so at the moment.

**Fix for Promise Resolution in makecodeEditorService**
This one had me scratching my head for a while. Basically, without the fix, when a message got queued up by Caller A but not sent immediately (editor not ready), then another caller has to come in and clear the queue once the editor is ready (Caller B). This was working okay, except it would assign the message's `handler` to the `resolve` function from the promise associated with Caller B's request, not Caller A's. As such, when a response comes back for the message, we'd end up resolving Caller B's promise instead of Caller A's, so Caller A would never return. To fix this, I assign the message metadata (including the handler) when we queue the message rather than when we dequeue it. This way, the `handler` will always get set to the `resolve` function for the promise returned to the caller that tried to send the message initially.

**Renamed loadProjectAsync to loadProjectMetadataAsync**
Wanted to avoid confusion between loading the metadata and the project "loading" in the iframe.

**Placeholders**
The `EvaluationResult` class and editor-side handler for the runeval request are just placeholders until we have actual evaluation running. I expect these to change in the near future.

### Try it Here
<del>Upload target: https://arcade.makecode.com/app/63812800ecd17f960039d2bee64091ef1cf07921-8b553c6ff2--eval</del>
Newer target: https://arcade.makecode.com/app/e620e5c33cf3af44b631781f582f4ebcc34223c7-c7c3f7a5ad--eval


teachertool/src/components/DebugInput.tsx
```diff
@@ -15,9 +15,8 @@ const DebugInput: React.FC<IProps> = ({}) => {
     const [bools, setBools] = useState(true);
 
     const evaluate = async () => {
-        setBools(!bools);
-        await runEvaluateAsync(shareLink, rubric);
-        await loadProjectAsync("hi", bools);
+        await loadProjectMetadataAsync(shareLink);
+        await runEvaluateAsync(rubric);
```
srietkerk, line 19: I'm thinking that `loadProjectMetadataAsync` should be called in `runEvaluateAsync` rather than here. I don't think the debug container needs context on what the project metadata is, and we can just have `runEvaluateAsync(shareLink, rubric)`, which I think better shows that `runEvaluteAsync` is going to run a rubric on a specific sharelink.

teachertool/src/components/EvalResultDisplay.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { AppStateContext } from "../state/appStateContext";
+
+interface IProps {}
+
+const EvalResultDisplay: React.FC<IProps> = ({}) => {
+    const { state: teacherTool } = useContext(AppStateContext);
+
+    return (
+        <>
+            {teacherTool.projectMetadata && (
```
srietkerk, line 13: I'm not a huge fan of having this whole component be conditional. Can we have this check in `App.tsx` instead and just have `EvalResultDisplay` return the div?

teachertool/src/components/MakecodeFrame.tsx
```diff
@@ -1,21 +1,28 @@
...
     /* eslint-disable @microsoft/sdl/react-iframe-missing-sandbox */
-    return <div className="makecode-frame-outer" style={{ display: "block" }}>
-        <iframe className="makecode-frame" src={pageSourceUrl} title={"title"} ref={handleIFrameRef}></iframe>
-    </div>
+    return (
+        <div className="makecode-frame-outer" style={{ display: "block" }}>
+            {teacherTool.projectMetadata && (
+                <iframe
+                    className="makecode-frame"
+                    src={createIFrameUrl(teacherTool.projectMetadata.id)}
```
srietkerk, line 21: I know you commented about this in your PR description, but I think this can be put into a transform to keep the `MakeCodeFrame` as clean as possible. You can have the transform access the app state and then just return the url, so instead of the util `createIFrameUrl` being called in source, it would the the transform getting called.

teachertool/src/components/MakecodeFrame.tsx
```diff
@@ -1,21 +1,28 @@
...
         setEditorRef(el ?? undefined);
     }
 
     /* eslint-disable @microsoft/sdl/react-iframe-missing-sandbox */
-    return <div className="makecode-frame-outer" style={{ display: "block" }}>
-        <iframe className="makecode-frame" src={pageSourceUrl} title={"title"} ref={handleIFrameRef}></iframe>
-    </div>
+    return (
+        <div className="makecode-frame-outer" style={{ display: "block" }}>
+            {teacherTool.projectMetadata && (
```
srietkerk, line 18: Similar to my comment above, I think you can just have this check in `App.tsx`.

teachertool/src/services/loggingService.ts
```diff
@@ -0,0 +1,26 @@
+import { isLocal } from "../utils";
```
srietkerk, line 1: Oh, this is something I forgot to change. It's not something that needs to be done in this PR, but Eric pointed out to me that we have this function in `pxt.util`, so we can technically remove it from the teachertool util.

https://github.com/microsoft/pxt/pull/9806#discussion_r1439775765

teachertool/src/state/actions.ts
```diff
@@ -41,8 +53,19 @@ const removeNotification = (notificationId: string): RemoveNotification => ({
     notificationId,
 });
 
+const setProjectMetadata = (meta: pxt.Cloud.JsonScript | undefined): SetProjectMetadata => ({
```
srietkerk, line 56: nit: not a fan of `meta` as a parameter name, is there a reason we don't have it be `metadata`? Then we could just have the object have `type: "xyz", metadata`, right?

teachertool/src/components/DebugInput.tsx
thsparks: I was hesitant to make these calls tightly coupled. I thought there might be scenarios where you want to do one and not the other (like adjust the rubric and re-run evaluate without reloading the share link, or loading a share link in the background while we're still not sure what the rubric will be), and there might also be cases where we just want to simply rely on the share link that's already in state and don't have trivial access to it through some input like we do here (remember that the debug input has it all in one place like this, but that may not always be the case if you're providing share links and then constructing a rubric in separate stages of the UI flow).

It might make sense to combine them down the road if these scenarios never materialize, but until we know a bit more about how that's all going to flow, I personally think it's better to keep them as distinct operations.

eanders-ms: I see your point @srietkerk. I think this is probably fine as is for the time being. It's an interim implementation. The rubric will move into the app state eventually.

teachertool/src/components/MakecodeFrame.tsx
thsparks: I'm not sure it's worth creating a transform that just replaces one one-line call with another one-line call, simply redirecting into a util function. My understanding is that it's reasonable for components to read state and use it as needed for the UI components like this, whereas a transform is more useful for larger operations that actually affect state. (And in terms of code-shareability, calling into a transform ties the component to the teachertool in much the same way that this does.)

In this case, since we're really just reading state (not changing it) and calling a util function, I'm skeptical that a transform is the right approach...unless there's something I'm missing, of course.

teachertool/src/components/EvalResultDisplay.tsx
eanders-ms: I wouldn't worry too much about how this component works just yet. This is placeholder while we get a design worked out.

eanders-ms:
> (except in localhost, because cross-domain localhost <-> makecode.com messaging doesn't work)

We shouldn't be encountering this scenario in local development. The editor should always load from localhost. Is that not the case?

thsparks:
> > (except in localhost, because cross-domain localhost <-> makecode.com messaging doesn't work)
> 
> We shouldn't be encountering this scenario in local development. The editor should always load from localhost. Is that not the case?

Ah, I think my comment was confusing. You're correct, in local development, the editor always loads from localhost. I just meant that, since the editor is loading from localhost, and I didn't think we'd be able to load or display projects with normal arcade share links in localhost...

That said, your comment made me second guess my assumption, so I just tried it out manually and we totally can load normal share links in localhost, so I'll change the code to support that.

teachertool/src/components/MakecodeFrame.tsx
eanders-ms: My inclination is the opposite of @srietkerk's here. I'd suggest moving `createIFrameUrl` into this file as a local function. I don't think it needs to be called from anywhere else.

Transforms are for code that affects the state of the app in some way.

teachertool/src/services/loggingService.ts
```diff
@@ -0,0 +1,26 @@
...
+    console.error(formatMessageForConsole(`${name}: ${details}`));
+}
+
+export const logInfo = (name: string, message: string) => {
+    pxt.tickEvent(`teachertool.${formatName(name)}`, { message: message });
+    console.log(formatMessageForConsole(message));
+}
+
+export const logDebug = (message: string) => {
+    if (isLocal()) {
```
eanders-ms, line 23: A logging service is great! Can you make it also respect the `pxt.options.debug` flag?

teachertool/src/components/DebugInput.tsx
srietkerk: For the record, I was not suggesting to get rid of `loadProjectMetadataAsync`, so it would still be available to be called in the scenarios where we might want to load the metadata outside of the evaluate function. 

Otherwise, I understand what you're saying, so I'm okay with keeping it as is, but I do have a bit of a hard time picturing a case where we'd run evaluate and not reload a share link, but maybe I'm thinking of reload too generally.. If we re-evaluate a project, you would still want to know that the project has been loaded and it's in the iframe so you can send the message and get the most accurate response. Whether that be from a backend fetch request or from a fetch in state, you'd still want to make sure that the thing that you're evaluating is the most recent thing, right? I guess I'm saying that I think we _should_ always process the metadata that's being used to make sure that we're evaluating the right project, but maybe that's not necessary.

teachertool/src/components/MakecodeFrame.tsx
srietkerk: Oh, right. Thank you for that clarification @thsparks, @eanders-ms. 

I think I'd be more inclined, then, to do as you said in your comment and pass the `teachertool.projectMetadata.id` into the `MakeCodeFrame` as a prop. I think this is a clean alternative and still allows for the `MakeCodeFrame` to be decoupled from  app state. I feel like having the prop be named `pageSourceUrl` would still be fine, as well, but maybe there's a better name.

teachertool/src/components/DebugInput.tsx
eanders-ms: I think you're right. Eventually runEvaluateAsync will pull the rubric from app state.

teachertool/src/utils/index.ts
```diff
@@ -13,10 +13,6 @@ export function makeNotification(
     };
 }
 
-export const isLocal = () => {
```
srietkerk, line 16: Thanks for doing this!

teachertool/src/components/MakecodeFrame.tsx
eanders-ms: > I think I'd be more inclined, then, to do as you said in your comment and pass the `teachertool.projectMetadata.id` into the `MakeCodeFrame` as a prop.

State is available to all components globally. This makes passing state as props largely unnecessary. Why does this component need this state passed as a prop?

srietkerk: So we can keep the `MakeCodeFrame` disconnected from state; but I guess if it's only one place of attachment, and we want to move the frame to someplace more general in the future, refactoring this won't be difficult.

teachertool/src/components/DebugInput.tsx
srietkerk: Oh, I wasn't talking about the rubric, I was talking about a project's metadata. I think we will likely pull a rubric from app state, but that's unrelated to my comments.

eanders-ms (approved):
Excellent progress!

teachertool/src/components/MakecodeFrame.tsx
thsparks: FWIW, I think the code that would be more valuable to share/generalize in the future would be the makecodeEditorService rather than this component. I'm inclined to agree with Eric here. It makes sense to pull the function into this component, so no-one else should know or care about the iframe specifics.

teachertool/src/components/DebugInput.tsx
thsparks: To Sarah's point: I can see the case for calling `loadProjectMetadataAsync` at the start of evaluation, but it does take a while and I'd be hesitant to do it unnecessarily. I'd prefer to leave this as-is for now and make sure we're intentional about each time we call into that transform. If it becomes a pain point later, we can see to it then.

To Eric's point about reading rubric from app state: yeah I could see that happening down the road. Currently the evaluate transform is doing a bit of double-duty, in that it sets the rubric in state and also uses it. Once we have more of a rubric "builder" then I expect these will get pulled out.

For now, I'm inclined to leave this as-is. LMK if that's an issue for anyone.

thsparks: To clarify why updating the metadata take a while, it causes the  iframe to update its url which means we have to wait for the editor to load everything again fresh. We could short-circuit the update if the url is the same, but I think that defeats the purpose you were referring to of fetching any last-minute changes.

teachertool/src/components/MakecodeFrame.tsx
srietkerk: Agreed, but if you have the component generic enough you can just ship the component and the service as a packaged deal. That said, as I mentioned before, it's likely something that can easily be removed if we want to rethink it, so I'm fine with leaving it as is since I'm in the minority.

For what it's worth, I'm also in favor of having the `createIFrameUrl` in the component. I don't think what I'm talking about and what Eric's suggesting is mutually exclusive.

teachertool/src/components/DebugInput.tsx
srietkerk: Sure, that sounds good to me.

srietkerk (approved):

teachertool/src/components/MakecodeFrame.tsx
thsparks: My hesitation with this one is that if we ever end up using this component in more than one place, we'd have to put the check in every time we specify it. Whereas inside the component, we have full context, knowing we're dependent specifically on this field in the state and that it doesn't make sense to render the component without it (in fact, the app would look like it was broken).

A sample scenario where this sort of thing might be relevant, imagine we decide to throw in a spinner or some kind of loading/placeholder message when projectMetadata isn't set. If we keep this check inside the component, we only have to update it here. However, if we push it out, then we have to update all the places that have used the component.

In this context, it's a little academic since I don't think we'll be using MakecodeFrame in more than one place in the app, but it's sticking to a convention that I think makes sense on the whole...

teachertool/src/components/MakecodeFrame.tsx
eanders-ms: From a reuse perspective, makecodeEditorService.ts is of more value than this UI component, and is already quite portable.

teachertool/src/services/loggingService.ts
```diff
@@ -0,0 +1,24 @@
...
+    return name.toLowerCase().replace(/ /g, "_");
+}
+
+export const logError = (name: string, details: string) => {
+    pxt.tickEvent("teachertool.error", { name: formatName(name), message: details });
+    console.error(formatMessageForConsole(`${name}: ${details}`));
+}
+
+export const logInfo = (name: string, message: string) => {
+    pxt.tickEvent(`teachertool.${formatName(name)}`, { message: message });
```
eanders-ms, line 16: Just noticed we're sending tick events from here. Logging info to telemetry will be overly chatty. Can we please remove it?

teachertool/src/services/loggingService.ts
```diff
@@ -0,0 +1,24 @@
+const formatMessageForConsole = (message: string) => {
+    const time = new Date();
+    return `[${time.getHours()}:${time.getMinutes()}:${time.getSeconds()}] ${message}`;
+}
+
+const formatName = (name: string) => {
+    return name.toLowerCase().replace(/ /g, "_");
+}
+
+export const logError = (name: string, details: string) => {
```
eanders-ms, line 10: Can we just log a message instead of name + message? I think logging to telemetry from here is too low level. Better to log errors where they happen.

teachertool/src/services/loggingService.ts
thsparks: I'll remove this in a future change.

teachertool/src/services/loggingService.ts
thsparks: I actually think it's helpful to have a unified error log like this. My assumption is we'd call this when the error happens, and the "details" would be the relevant context from there (i.e. a stack trace, description, etc...). Maybe details should be a property bag instead of a string...but in essence I suppose I don't see this as "not logging errors when they happen" but rather, this could be the mechanism through which you log an error when it happens.

thsparks: But I think it's helpful to simplify queries around "What are the kinds of errors we're seeing in the teacher tool?" "Have there been any spikes in errors after a release?" and so on without having to query an every-changing set of different error events.

PR #9823, opened 2024-01-11
Teacher tool: Initial validation logic

kimprice:
Big thanks to @thsparks for helping me get this set up.

Currently, the validation works for `BlockCheckCriteria` only. It takes the rubric and sets the required block counts for each block id. Then it runs the `validateBlocksExist` function for each set of blocks in the rubric.

Upload target: https://arcade.makecode.com/app/5df94cd2b924c898ff9960973789eff8048cc4c5-2776605799--eval
Sample share link: https://arcade.makecode.com/S42128-59682-77192-60950
Sample rubric:
```JSON
{
    "criteria": [
        {
            "displayText": "One or more custom functions that must be called",
            "criteriaId": "blockCheck",
            "blockRequirements": [
                {
                    "blocks": ["function_call", "function_call_output"],
                    "count": 1
                },
                {
                    "blocks": ["function_def"],
                    "count": 1
                }
            ]
        },
        {
            "displayText": "Two different kinds of loops used",
            "criteriaId": "blockCheck",
            "blockRequirements": [
                {
                    "blocks": ["controls_repeat_ext", "device_while", "pxt_controls_for", "pxt_controls_for_of"],
                    "count": 2
                }
            ]
        }
    ]
}
```

### DIFF
--- a/pxtblocks/code-validation/rubricCriteria.ts
+++ b/pxtblocks/code-validation/rubricCriteria.ts
@@ -23,6 +23,55 @@ namespace pxt.blocks {
         }
     }
 
+    function blockSetToRequiredBlockCounts(blockSet: BlockSet): pxt.Map<number> {
+        const requiredBlockCounts: pxt.Map<number> = {};
+        blockSet.blocks.forEach((block) => {
+            requiredBlockCounts[block] = blockSet.count;
+        });
+        return requiredBlockCounts;
+    }
+
+    export function validateProject(usedBlocks: Blockly.Block[], rubric: string): EvaluationResult {
+        const rubricData = parseRubric(rubric);
+        const finalResult: pxt.Map<boolean> = {};
+        rubricData.criteria.forEach((criteria: RubricCriteria) => {
+            (criteria as BlockCheckCriteria).blockRequirements.forEach((blockSet) => {
+                const result = validateBlockSet(usedBlocks, blockSet);
+                Object.keys(result).forEach((blockId) => {
+                    finalResult[blockId] = result[blockId];
+                });
+            });
+        });
+        return { blockIdResults: finalResult} as EvaluationResult;
+    }
+
+
+    function validateBlockSet(usedBlocks: Blockly.Block[], blockSet: BlockSet): pxt.Map<boolean> {
+        const requiredBlockCounts = blockSetToRequiredBlockCounts(blockSet);
+        const blockResults: pxt.Map<boolean> = {};
+        Object.keys(requiredBlockCounts).forEach((blockId) => {
+            blockResults[blockId] = true;
+        });
+        const {
+            missingBlocks,
+            disabledBlocks,
+            insufficientBlocks
+        } = pxt.blocks.validateBlocksExist({
+            usedBlocks: usedBlocks,
+            requiredBlockCounts: requiredBlockCounts,
+        });
+        missingBlocks.forEach((blockId) => {
+            blockResults[blockId] = false;
+        });
+        disabledBlocks.forEach((blockId) => {
+            blockResults[blockId] = false;
+        });
+        insufficientBlocks.forEach((blockId) => {
+            blockResults[blockId] = false;
+        });
+        return blockResults;
+    }
+
     export abstract class RubricCriteria {
         displayText: string;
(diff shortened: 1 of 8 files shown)


teachertool/src/components/EvalResultDisplay.tsx
```diff
@@ -14,8 +14,15 @@ const EvalResultDisplay: React.FC<IProps> = ({}) => {
                 <div className="eval-results-container">
                     <h3>{lf("Project: {0}", teacherTool.projectMetadata.name)}</h3>
                     {teacherTool.currentEvalResult === undefined && <div className="common-spinner" />}
-                    {teacherTool.currentEvalResult?.passed === true && <h4 className="positive-text">Passed!</h4>}
-                    {teacherTool.currentEvalResult?.passed === false && <h4 className="negative-text">Failed!</h4>}
+                    {Object.keys(teacherTool.currentEvalResult?.blockIdResults??{}).map((id) => {
```
eanders-ms, line 17: whitespace nit
```suggestion
                    {Object.keys(teacherTool.currentEvalResult?.blockIdResults ?? {}).map((id) => {
```

pxtblocks/code-validation/rubricCriteria.ts
```diff
@@ -23,6 +23,55 @@ namespace pxt.blocks {
...
+        const finalResult: pxt.Map<boolean> = {};
+        rubricData.criteria.forEach((criteria: RubricCriteria) => {
+            (criteria as BlockCheckCriteria).blockRequirements.forEach((blockSet) => {
+                const result = validateBlockSet(usedBlocks, blockSet);
+                Object.keys(result).forEach((blockId) => {
+                    finalResult[blockId] = result[blockId];
+                });
+            });
+        });
+        return { blockIdResults: finalResult} as EvaluationResult;
```
eanders-ms, line 45: whitespace nit
```suggestion
        return { blockIdResults: finalResult } as EvaluationResult;
```

eanders-ms (approved):
Looks good!

webapp/src/app.tsx
```diff
@@ -4038,6 +4038,14 @@ export class ProjectView
             });
     }
 
+    async getBlocksAsync(): Promise<Blockly.Block[]> {
+        if (!this.isBlocksActive()) {
+            await this.openBlocksAsync();
```
thsparks, line 4043: I wonder if we should just throw an error in this case rather than switch to blocks. Having that switch be a side-effect of another call feels like it could lead to some unwanted behavior...plus, if we throw an error instead of switching, this function doesn't have to be async :)

teachertool/src/components/EvalResultDisplay.tsx
```diff
@@ -14,8 +14,15 @@ const EvalResultDisplay: React.FC<IProps> = ({}) => {
...
                     <h3>{lf("Project: {0}", teacherTool.projectMetadata.name)}</h3>
                     {teacherTool.currentEvalResult === undefined && <div className="common-spinner" />}
-                    {teacherTool.currentEvalResult?.passed === true && <h4 className="positive-text">Passed!</h4>}
-                    {teacherTool.currentEvalResult?.passed === false && <h4 className="negative-text">Failed!</h4>}
+                    {Object.keys(teacherTool.currentEvalResult?.blockIdResults ?? {}).map((id) => {
+                        const result = teacherTool.currentEvalResult?.blockIdResults[id];
+                        return (
+                            <div className="result-block-id" key={id}>
+                                <p className="block-id-label">{id}:</p>
+                                <p className={result?"positive-text":"negative-text"}>{result?"passed":"failed"}</p>
```
thsparks, line 22: Another whitespace nit :)

```suggestion
                                <p className={result ? "positive-text" : "negative-text"}>{result ? "passed" : "failed"}</p>
```

thsparks (approved):
One thought about perhaps throwing an error rather than having a switch-to-blocks side effect on that `getBlocksAsync` call. Otherwise LGTM!

webapp/src/app.tsx
kimprice: Good point!

thsparks (approved):
👍

PR #9829, opened 2024-01-23
Load Teacher Tools Catalog from Docs, Store in State

thsparks:
This change includes the following:
1. Loading the catalog from docs files (shared and target-specific, with "live" and "test" files in each).
2. Storing catalog in state once loaded
3. Selecting catalog criteria items to add to your rubric
4. Storing selected criteria instances in state
5. Basic UI for selecting criteria & displaying selected criteria
6. A few new types needed to read/parse criteria json

Notably, it does **not** include:
1. Any linking between the selected criteria and what gets sent to the editor for validation
2. Generating validator plans from the selected criteria (would be a prerequisite to sending to the editor)
3. Any target-specific catalogs (those will need to be checked into their respective pxt-target repos)

A few notes:
1. Test catalog entries are only loaded if the URL contains dbg=1. We could also show this in localhost builds, but I decided not to since I thought that could lead to confusion if someone saw a new catalog entry in localhost and didn't realize it needed to be moved to be visible in live.
2. In order to generate GUIDS, I added the uuid package: https://www.npmjs.com/package/uuid

Unfortunately upload targets don't work well with docs, but here are some screenshots:
image

image

image


teachertool/src/components/CatalogModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { Checkbox } from "react-common/components/controls/Checkbox";
+import { Modal } from "react-common/components/controls/Modal";
+import { hideCatalogModal } from "../transforms/hideCatalogModal";
+import { addCriteriaToRubric } from "../transforms/addCriteriaToRubric";
+
+interface IProps {}
+
+const CatalogModal: React.FC<IProps> = ({}) => {
+    const { state: teacherTool, dispatch } = useContext(AppStateContext);
+    const [ checkedCriteriaIds, setCheckedCriteria ] = useState<string[]>([]);
```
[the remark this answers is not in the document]
srietkerk: Would it be possible to make this into a set? I think we could make `handleCheckboxChange` simpler if so.

teachertool/src/services/loggingService.ts
```diff
@@ -7,13 +7,12 @@ const formatName = (name: string) => {
     return name.toLowerCase().replace(/ /g, "_");
 }
 
-export const logError = (name: string, details: string) => {
-    pxt.tickEvent("teachertool.error", { name: formatName(name), message: details });
-    console.error(formatMessageForConsole(`${name}: ${details}`));
+export const logError = (name: string, details: string, props: pxt.Map<string | number> = {}) => {
```
srietkerk, line 10: nit: rename `props` to `data` or something similar. When I think of props, I think of React props..

teachertool/src/transforms/addCriteriaToRubric.ts
```diff
@@ -0,0 +1,39 @@
...
+
+        const params = catalogCriteria.parameters?.map(
+            param =>
+                ({
+                    name: param.name,
+                    value: undefined,
+                } as pxt.blocks.CriteriaParameterValue)
+        );
+
+        const instanceId = uuidV4();
```
srietkerk, line 26: This means that there is no chances of duplicate criteria, right? So for now, you can add the same criteria to a rubric as many times if you re-open the modal and select it?

teachertool/src/components/CatalogModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+    function handleCheckboxChange(criteria: pxt.blocks.CatalogCriteria, newValue: boolean) {
```
srietkerk, line 16: Something I was thinking about.. What if we added a `selected` field to the Catalog Criteria Schema? Since we parse the catalog from docs into state, we can manage whether or not something is checked from the catalog directly. The look-up for an object like that would be faster, right? And then when the checkbox is changed, we can just toggle that field. The done and close modal would have to be modified, though. Also, I'm not a huge fan of the implications that this has on state.. it feels kinda icky that a modal would have so much power over the state of the catalog, but I also feel like this handler should be simpler than having to do a look-up or filter.

srietkerk (commented):
Have lots of questions and comments, but this is super exciting stuff. Great work!!

teachertool/src/transforms/addCriteriaToRubric.ts
```diff
@@ -0,0 +1,39 @@
+import { stateAndDispatch } from "../state";
+import { v4 as uuidV4 } from 'uuid';
+import * as Actions from "../state/actions";
+import { logDebug, logError } from "../services/loggingService";
+
+export async function addCriteriaToRubric(catalogCriteriaIds: string[]) {
+    const { state: teacherTool, dispatch } = stateAndDispatch();
+
+    // Create instances for each of the catalog criteria.
+    const instances = catalogCriteriaIds.reduce((accumulator, catalogCriteriaId) => {
```
thsparks, line 10: >reduce [](http://example.com/codeflow?start=41&length=6)

I should probably just use a for...of loop here instead of reduce. Would be cleaner & easier to read.

pxtblocks/code-validation/types.ts
```diff
@@ -0,0 +1,35 @@
+namespace pxt.blocks {
+    // A criteria defined in the catalog of all possible criteria for the user to choose from when creating a rubric.
+    export interface CatalogCriteria {
+        id: string; // A unique id (GUID) for the catalog criteria
+        use: string; // Refers to the validator plan this criteria relies upon
+        template: string; // A (mostly) human-readable string describing the criteria. May contain parameters
+        description: string | undefined; // More detailed description
```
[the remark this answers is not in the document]
thsparks: Based on our designs, I believe the "template" is what gets displayed for every criteria in the list. As currently defined, even when there are no parameters, there's still a "template", it's just all filled-in. Open to alternative name suggestions, but I don't have any off the top of my head. Description is more long-winded and, arguably, could be completely hidden.

teachertool/package-lock.json
```diff
@@ -9,11 +9,13 @@
       "version": "0.1.0",
       "dependencies": {
         "@types/node": "^16.11.33",
+        "@types/uuid": "^9.0.7",
```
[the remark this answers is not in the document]
thsparks: I looked and wasn't able to find one. I was also surprised that we hadn't already needed to do this elsewhere, but searching for guid/uuid didn't turn up much at all.

docs/teachertool/catalog-shared-test.json
```diff
@@ -0,0 +1,19 @@
...
+            "use": "check_for_comments",
+            "template": "At least ${count} comments",
+            "description": "The project contains at least the specified number of comments.",
+            "docPath": "/teachertool",
+            "params": [
+                {
+                    "name": "count",
+                    "type": "number",
+                    "default": 1,
+                    "path": "checks[0].count"
```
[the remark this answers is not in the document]
thsparks: I think we'd need to install it, but I don't want to do that until we actually implement the processing of this info. No point in having an unused package hanging around.

teachertool/src/transforms/addCriteriaToRubric.ts
thsparks: Correct

teachertool/src/components/CatalogModal.tsx
thsparks: We want the user to be able to have multiple instances of the same catalog criteria, so just having `selected` on catalog criteria doesn't quite work (if it did, there'd be no real need for criteria instances at all).

I'm not sure I quite understand the concern around the handler. Could you elaborate on that a bit?

teachertool/src/components/CatalogModal.tsx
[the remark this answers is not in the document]
thsparks: I add it when the type is not easily inferred from the default value. So it's not really needed for a number or a string, but in this case it's not clear what type is expected in the list, so I think it's helpful to include.

Set might work well. I'll take a look.

teachertool/src/components/ActiveRubricDisplay.tsx
```diff
@@ -0,0 +1,38 @@
...
+interface IProps {}
+
+const ActiveRubricDisplay: React.FC<IProps> = ({}) => {
+    const { state: teacherTool, dispatch } = useContext(AppStateContext);
+
+    return (
+        <div className="rubric-display">
+            <h3>{lf("Rubric")}</h3>
+            {teacherTool.selectedCriteria?.map(criteriaInstance => {
+                const catalogCriteria = getCatalogCriteriaWithId(criteriaInstance.catalogCriteriaId);
```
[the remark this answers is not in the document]
thsparks: I should actually remove it from below, I think. Shouldn't be needed. Thanks!

teachertool/src/transforms/removeCriteriaFromRubric.ts
```diff
@@ -0,0 +1,12 @@
+import { stateAndDispatch } from "../state";
+import { v4 as uuidV4 } from 'uuid';
+import * as Actions from "../state/actions";
+import { logDebug } from "../services/loggingService";
+
+export async function removeCriteriaFromRubric(instance: pxt.blocks.CriteriaInstance) {
```
eanders-ms, line 6: Does this need to be async?

teachertool/src/utils/index.ts
```diff
@@ -1,5 +1,6 @@
 import { nanoid } from "nanoid";
 import { NotificationWithId } from "../types";
+import { stateAndDispatch } from "../state";
```
eanders-ms, line 3: Let's keep app state out of the utils file. I suggest moving `getCatalogCriteriaWithId` to /src/state/helpers.ts. We'll probably accumulate more of these helpers, and I'd like to keep utils uncomplicated.

teachertool/src/transforms/loadCatalog.ts
```diff
@@ -0,0 +1,51 @@
...
+            // Empty file.
+            continue;
+        }
+
+        try {
+            const catalogInfoParsed = JSON.parse(catalogContent);
+            const catalogInfo = catalogInfoParsed as CatalogInfo;
+            fullCatalog = fullCatalog.concat(catalogInfo.criteria ?? []);
+        } catch (e) {
+            logError("parse_catalog_failed", e as string, {catalogFile});
```
eanders-ms, line 45: It might be nice to organize these error codes in one place and refer to them here like `Errors.parse_catalog_failed`. This would simplify looking up error codes when building a Kusto query.

pxtblocks/code-validation/types.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    // An instance of a criteria in a rubric.
+    export interface CriteriaInstance {
+        catalogCriteriaId: string;
+        instanceId: string;
+        params: CriteriaParameterValue[] | undefined;
+    }
+
+    // Possible values for CriteriaParameterPicker, these are different ways a user may enter parameter values.
+    export type CriteriaParameterPicker = "blocksPicker" | "numericInput";
```
eanders-ms, line 20: > "numericInput"

I don't think we need a picker for numbers. We can make the param value directly editable.

pxtblocks/code-validation/types.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    // Represents a parameter definition in a catalog criteria.
+    export interface CriteriaParameter {
+        name: string;
+        type: string;
+        picker: CriteriaParameterPicker;
```
eanders-ms, line 26: Now I'm thinking we don't need a `picker` field at all. We can use the `type` field to infer the picker. `type: "block_id"` for the block picker.

eanders-ms (commented):
Still reviewing, but wanted to send out my comments so far.

teachertool/src/components/CatalogModal.tsx
srietkerk: Yeah, my idea would be that the `selected` state on the catalog would be reset every time the modal is closed so when the user re-opens the modal, they have a fresh state. This might be more work than it's worth. 

In regards to the handler, I'm a bit worried about a list operation every time the box is checked or unchecked. To be fair, they shouldn't be very computationally taxing, but the list operations plus conditions when we want to toggle a condition for something seems like a lot of logic.

teachertool/src/components/CatalogModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+    function handleDoneClicked() {
+        addCriteriaToRubric(checkedCriteriaIds)
+        closeModal();
+    }
+
+    function closeModal() {
+        hideCatalogModal();
+
+        // Clear for next open.
+        setCheckedCriteria([]);
```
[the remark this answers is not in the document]
srietkerk: Makes sense! I don't think the "don't allow duplicates" is something that needs to be included in this PR.

pxtblocks/code-validation/types.ts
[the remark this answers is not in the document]
srietkerk: You're totally right. Makes sense!

teachertool/src/components/ActiveRubricDisplay.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+                return criteriaInstance?.catalogCriteriaId && (
+                        <div className="criteria-instance-display" id={`criteriaInstance${criteriaInstance.instanceId}`}>
```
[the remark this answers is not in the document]
srietkerk: For what it's worth, I think what you have is really close to what we're wanting, at least in terms of what we have for the scoped-down v1, but I understand what you're saying about keeping it contained, so I'm okay with leaving it as is for now.

teachertool/src/transforms/loadCatalog.ts
```diff
@@ -0,0 +1,51 @@
...
+    "/teachertool/catalog-shared-test.json"
+]
+
+interface CatalogInfo {
+    criteria: pxt.blocks.CatalogCriteria[];
+}
+
+export async function loadCatalog() {
+    const { dispatch } = stateAndDispatch();
+    const catalogFiles = pxt.options.debug ? prodFiles.concat(testFiles) : prodFiles;
```
thsparks, line 22: >pxt.options.debug [](http://example.com/codeflow?start=25&length=17)

Use a different flag or beta?

teachertool/src/App.tsx
```diff
@@ -20,8 +23,13 @@ function App() {
...
 
     useEffect(() => {
-        // Init subsystems.
-        NotificationService.initialize();
+        if (ready) {
+            // Init subsystems.
+            NotificationService.initialize();
+
+            // Load criteria catalog
+            loadCatalog();
```
thsparks, line 31: >loadCatalog [](http://example.com/codeflow?start=12&length=11)

Additional flag for when all this is done?

thsparks: Separate PR :)

teachertool/src/transforms/loadCatalog.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    let fullCatalog: pxt.blocks.CatalogCriteria[] = [];
+    for (const catalogFile of catalogFiles) {
+        let catalogContent = "";
+        try {
+            const catalogResponse = await fetch(catalogFile);
+            catalogContent = await catalogResponse.text();
```
thsparks, line 29: >text [](http://example.com/codeflow?start=51&length=4)

json accessor?

teachertool/src/components/CatalogModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+            onClick: closeModal,
+        },
+        {
+            label: lf("Add Selected"),
+            className: "primary",
+            onClick: handleDoneClicked,
+        },
+    ]
+
+    return teacherTool.modal ? (
```
thsparks, line 53: >teacherTool.modal [](http://example.com/codeflow?start=11&length=17)

teacherTool.modal == "catalog-modal"

teachertool/src/components/CatalogModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+        <Modal className="catalog-modal" title={lf("Select the criteria you'd like to include")} onClose={closeModal} actions={modalActions}>
+            <div className="catalog-container" title={lf("Select the criteria you'd like to include")}>
```
thsparks, line 55: ><div className="catalog-container" title={lf("Select the criteria you'd like to include")}> [](http://example.com/codeflow?start=12&length=91)

redundant now?

teachertool/src/components/CatalogModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+                {teacherTool.catalog?.map(criteria => {
+                    return criteria?.template && (
+                            <Checkbox
+                                id={`chk${criteria.id}`}
```
thsparks, line 59: >id={`chk${criteria.id}`} [](http://example.com/codeflow?start=32&length=24)

We need key but not id and just use an integer

teachertool/src/components/CatalogModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+        } else if (!newValue && checkedCriteriaIds.includes(criteria.id)) {
+            setCheckedCriteria(checkedCriteriaIds.filter(id => id !== criteria.id));
+        }
+    }
+
+    function isCheckboxChecked(criteriaId: string): boolean {
+        return checkedCriteriaIds.includes(criteriaId);
+    }
+
+    function handleDoneClicked() {
```
thsparks, line 28: >handleDoneClicked [](http://example.com/codeflow?start=13&length=17)

Rename

teachertool/src/components/CatalogModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+        addCriteriaToRubric(checkedCriteriaIds)
```
thsparks, line 29: ;

teachertool/src/transforms/addCriteriaToRubric.ts
```diff
@@ -0,0 +1,39 @@
...
+        logDebug(`Adding criteria with ID ${catalogCriteriaId}`);
+        const catalogCriteria = teacherTool.catalog?.find(cat => cat.id === catalogCriteriaId);
```
thsparks, line 12: >teacherTool.catalog?.find(cat => cat.id === catalogCriteriaId); [](http://example.com/codeflow?start=32&length=63)

helper function

teachertool/src/transforms/addCriteriaToRubric.ts
thsparks, line 26: >uuidV4 [](http://example.com/codeflow?start=27&length=6)

nanoid?

thsparks: wrap in util function getId()

pxtblocks/code-validation/types.ts
```diff
@@ -0,0 +1,35 @@
+namespace pxt.blocks {
+    // A criteria defined in the catalog of all possible criteria for the user to choose from when creating a rubric.
+    export interface CatalogCriteria {
```
thsparks, line 3: >CatalogCriteria [](http://example.com/codeflow?start=21&length=15)

move criteria types to teacher tool

teachertool/src/state/reducer.ts
```diff
@@ -43,6 +43,36 @@ export default function reducer(state: AppState, action: Action): AppState {
                 currentEvalResult: action.result,
             };
         }
+        case "SET_CATALOG": {
+            return {
+                ...state,
+                catalog: action.catalog,
+            };
+        }
+        case "ADD_CRITERIA_INSTANCES": {
```
thsparks, line 52: >ADD_CRITERIA_INSTANCES [](http://example.com/codeflow?start=14&length=22)

Fully set, not append

teachertool/src/components/ActiveRubricDisplay.tsx
thsparks, line 22: >div [](http://example.com/codeflow?start=25&length=3)

key here

teachertool/src/transforms/hideCatalogModal.ts
```diff
@@ -0,0 +1,13 @@
+import { logDebug } from "../services/loggingService";
+import { stateAndDispatch } from "../state";
+import * as Actions from "../state/actions";
+
+export async function hideCatalogModal() {
```
thsparks, line 5: >hideCatalogModal [](http://example.com/codeflow?start=22&length=16)

hideModal(ModalType)

teachertool/src/components/CatalogModal.tsx
thsparks: Turns out our react common checkbox does actually require an id, so I'll leave that in but also add a key.

teachertool/src/components/CatalogModal.tsx
thsparks: Per our discussion in person, I think something like that might work (we'd need a new component like a CriteriaCheckbox to store `selected`, since I don't think we want UI-related fields to leak out into non-UI-specific classes), but it would probably just move the complexity elsewhere rather than actually simplifying things, so I'm inclined to leave this as-is at the moment.

teachertool/src/App.tsx
```diff
@@ -37,9 +45,11 @@ function App() {
             <HeaderBar />
```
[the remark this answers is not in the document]
thsparks: I'll leave this be since you've added one in https://github.com/microsoft/pxt/pull/9830

teachertool/src/App.tsx
```diff
@@ -28,6 +34,10 @@ function App() {
                 const cfg = await downloadTargetConfigAsync();
                 dispatch(Actions.setTargetConfig(cfg || {}));
                 pxt.BrowserUtils.initTheme();
+
+                // Load criteria catalog
+                await loadCatalogAsync();
```
thsparks, line 39: @eanders-ms I wasn't totally sure how to plug into this initialization code. Is this suitable?

eanders-ms: Yep, looks good! If we end up with a lot of async tasks here we should try to parallelize some of them. Nothing to worry about yet.

eanders-ms (approved):
Looks good!

teachertool/src/types/errorCode.ts
```diff
@@ -0,0 +1,8 @@
+export enum ErrorCode {
```
eanders-ms, line 1: Oh nice. We might want to strongly type logError to this. `logError(errorCode: ErrorCode,`

PR #9830, opened 2024-01-25
teacher tool: initial theming from target

eanders-ms:
Initial support for styling from target theme. Only tested with microbit. It will take another PR or two before we have theming nicely packaged, I think.

I updated app startup to download targetconfig.json, but the local pxt server puts it behind a gate. This is for legacy reasons that I doubt apply anymore, but can't know for sure. As a workaround, run pxt serve with the `--noauth` option. We don't actually use targetconfig.json yet, so you won't notice anything if you forget to add --noauth.

I also updated loggingService.ts to accept a message as `any`, because I like to log JSON objects to the console and they'll be expandable there. @thsparks I'm curious what you think of my changes.

I tried to update the name of `BackendRequests.ts` --> `backendRequests.ts`, but I'm unsure if it worked.

image


eanders-ms:
I've temp renamed `BackendRequests.ts` -> `ackendRequests.ts`. Will fix it in the next PR.

teachertool/src/services/loggingService.ts
```diff
@@ -1,24 +1,32 @@
...
+    });
+    console.error(timestamp(), name, message, data);
+};
 
-export const logInfo = (name: string, message: string) => {
-    pxt.tickEvent(`teachertool.${formatName(name)}`, { message: message });
-    console.log(formatMessageForConsole(message));
-}
+export const logInfo = (message: any) => {
+    console.log(timestamp(), message);
```
thsparks, line 25: `message: any` change LGTM. It appears there's more to these console functions that I realized :) I didn't even know you could pass in multipe parameters like that!

teachertool/src/components/HeaderBar.tsx
```diff
@@ -3,67 +3,112 @@ import * as React from "react";
...
 
-    protected getOrganizationLogo(targetTheme: pxt.AppTheme) {
-        const logoUrl = targetTheme.organizationWideLogo;
-        return <div className="header-logo">
-            {logoUrl
-                ? <img src={logoUrl} alt={lf("{0} Logo", targetTheme.organization)}/>
-                : <span className="name">{targetTheme.organization}</span>}
-        </div>
-    }
+    const brandIconClick = () => {};
```
thsparks, line 11: Are there any functional changes in this file, or mostly formatting? Just want to know where to focus :)

teachertool/src/state/actions.ts
```diff
@@ -28,6 +28,11 @@ type SetEvalResult = ActionBase & {
     result: pxt.blocks.EvaluationResult | undefined;
 };
 
+type SetTargetConfig = ActionBase & {
```
thsparks, line 31: Is this expected to be hooked up to anything in the state yet?

teachertool/src/components/HeaderBar.tsx
eanders-ms: Mostly adjusting css classes. I ran prettier on it and the reformatting was significant. I plan to do a prettier-only PR after this to fix formatting across the project. Then we can make a practice of running it as part of each PR.

teachertool/src/state/actions.ts
eanders-ms: Oops I missed that part. That should have been caught by the linter. Will check settings.

eanders-ms: The reducer was returning the state unmodified when there was no matching case statement. Removed that.

teachertool/src/services/loggingService.ts
```diff
@@ -1,24 +1,45 @@
...
+    message?: any,
+    data: pxt.Map<string | number> = {}
+) => {
+    errorCode = formatName(errorCode);
+    let dataObj = { ...data };
+    if (message) {
+        if (typeof message === "object") {
+            dataObj = { ...dataObj, ...message };
+            // Look for non-enumerable properties found on Error objects
+            ["message", "stack", "name"].forEach(key => {
```
eanders-ms, line 21: @thsparks this will log the callstack if `message` is of type `Error`.

eanders-ms: I don't know why these fields are non-enumerable. Makes no sense to me.

teachertool/src/services/loggingService.ts
```diff
@@ -1,24 +1,45 @@
...
     return name.toLowerCase().replace(/ /g, "_");
-}
+};
 
-export const logError = (name: string, details: string) => {
-    pxt.tickEvent("teachertool.error", { name: formatName(name), message: details });
-    console.error(formatMessageForConsole(`${name}: ${details}`));
-}
+export const logError = (
+    errorCode: string,
```
eanders-ms, line 11: I renamed this `errorCode` because the Error type already has a field called `name`, and I wanted to include that in the data payload.

teachertool/src/components/HeaderBar.tsx
thsparks: If we must use prettier, I will do my best. But to my personal taste, I think it actually makes the code harder to read...

teachertool/src/services/loggingService.ts
thsparks: hmm, would it work if you did an `if message instanceof Error` check?

teachertool/src/components/HeaderBar.tsx
eanders-ms: I don't love prettier, but I do feel it's important to have consistent styling. I wonder how configurable prettier is. I don't like how it stacks lines. Maybe we can increase its max line width to compensate that.

teachertool/src/services/loggingService.ts
eanders-ms: Nope, they'd still be non-enumerable and excluded from the spread operator.

teachertool/src/components/HeaderBar.tsx
thsparks: Yeah, that would be a big help. I vote we investigate that before doing any big prettier-only PR :) Seems like it ought to be possible: https://prettier.io/docs/en/options.html#print-width

PR #9831, opened 2024-01-25
teacher tool - run prettier

eanders-ms:
I know everyone hates prettier. I don't like it either. But I do think style consistency is important. I'm open to other solutions.

I'm hoping we will remember to run prettier as part of each PR.

There is one thing be alert for: calls to `lf(...)` cannot be broken into multiple lines. Our gulp script cannot handle this currently and will error out. TODO to fix this.


teachertool/src/state/actions.ts
```diff
@@ -62,24 +62,12 @@ type HideModal = ActionBase & {
...
-    | RemoveNotification
-    | SetProjectMetadata
-    | SetEvalResult
-    | SetTargetConfig
-    | SetCatalog
-    | SetSelectedCriteria
-    | RemoveCriteriaInstance
-    | ShowModal
-    | HideModal;
+export type Action = PostNotification | RemoveNotification | SetProjectMetadata | SetEvalResult | SetTargetConfig | SetCatalog | SetSelectedCriteria | RemoveCriteriaInstance | ShowModal | HideModal;
```
thsparks, line 65: Could we `// prettier-ignore` both of the export statements in this file? I realize these might fit within our expanded max line length, but I feel like it's okay to make an exception here...

eanders-ms: Certainly. There is a tension with increased line width vs what we'd prefer to have stacked. We may want to bring the max line width in a bit. 200 is very wide.

eanders-ms: Setting width to 120 corrected this. And the results seemed more reasonable to me in general.

thsparks: Seems like you found the sweet spot!

thsparks (approved):

PR #9832, opened 2024-01-26
teacher tool - set module css pattern

eanders-ms:
The main reason for this PR is to introduce "module css". This is a feature we get with create-react-app that scopes css class names to their module, making it easier to use simple names for css classes and they won't clash with one another.

Module css docs:
* https://create-react-app.dev/docs/adding-a-css-modules-stylesheet/
* https://css-tricks.com/css-modules-part-1-need/

This PR contains some React changes too, but it is work in progress and will likely change a lot.

### DIFF
--- a/react-common/components/util.tsx
+++ b/react-common/components/util.tsx
@@ -46,7 +46,7 @@ export function fireClickOnEnter(e: React.KeyboardEvent<HTMLElement>) {
     }
 }
 
-export function classList(...classes: string[]) {
+export function classList(...classes: (string | undefined)[]) {
     return classes
         .filter(c => typeof c === "string")
         .reduce((prev, c) => prev.concat(c.split(" ")), [] as string[])
--- a/teachertool/src/App.tsx
+++ b/teachertool/src/App.tsx
@@ -10,11 +10,8 @@ import { downloadTargetConfigAsync } from "./services/backendRequests";
 import { logDebug } from "./services/loggingService";
 
 import HeaderBar from "./components/HeaderBar";
+import MainPanel from "./components/MainPanel";
 import Notifications from "./components/Notifications";
-import DebugInput from "./components/DebugInput";
-import { MakeCodeFrame } from "./components/MakecodeFrame";
-import EvalResultDisplay from "./components/EvalResultDisplay";
-import ActiveRubricDisplay from "./components/ActiveRubricDisplay";
 import CatalogModal from "./components/CatalogModal";
 
 import { postNotification } from "./transforms/postNotification";
@@ -54,17 +51,12 @@ function App() {
             <div className="ui large main loader msft"></div>
         </div>
     ) : (
-        <div className="app-container">
+        <>
             <HeaderBar />
-            <div className="inner-app-container">
-                <DebugInput />
-                <ActiveRubricDisplay />
-                <EvalResultDisplay />
-                <MakeCodeFrame />
-            </div>
+            <MainPanel />
             <CatalogModal />
             <Notifications />
-        </div>
+        </>
     );
 }
 
--- a/teachertool/src/components/MainPanel.tsx
+++ b/teachertool/src/components/MainPanel.tsx
@@ -1,0 +1,31 @@
+import * as React from "react";
+import css from "./styling/MainPanel.module.css";
+
+import DebugInput from "./DebugInput";
+import MakeCodeFrame from "./MakecodeFrame";
+import EvalResultDisplay from "./EvalResultDisplay";
+import ActiveRubricDisplay from "./ActiveRubricDisplay";
+import SplitPane from "./SplitPane";
+
+interface IProps {}
+
+const MainPanel: React.FC<IProps> = () => {
+    return (
+        <div className={css["main-panel"]}>
+            <SplitPane split={"vertical"} defaultSize={"80%"} primary={"left"}>
+                {/* Left side */}
+                <>
+                    <DebugInput />
+                    <ActiveRubricDisplay />
+                    <EvalResultDisplay />
+                </>
+                {/* Right side */}
+                <>
+                    <MakeCodeFrame />
+                </>
+            </SplitPane>
+        </div>
+    );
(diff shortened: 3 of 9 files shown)


eanders-ms:
I wanted to get these changes out so I can switch gears back to wrangling target theme colors into css variables.

teachertool/src/components/SplitPane.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+interface IProps {
+    className?: string;
+    split: "horizontal" | "vertical";
+    defaultSize: number | string;
+    primary: "left" | "right";
+    children: React.ReactNode;
+}
+
+const VerticalSplit: React.FC<IProps> = ({ className, split, children }) => {
```
eanders-ms, line 14: I want to switch away from the default export pattern for react components. Makes it too easy for actual component names to go stale, like this one. Something for a future PR.

teachertool/src/teacherTool.css
```diff
@@ -44,6 +44,10 @@
     --high-contrast-hyperlink: #807FFF;
 }
 
+* {
```
eanders-ms, line 47: This is to reset some global styling coming from the target css that was messing up default div layout.

teachertool/src/components/SplitPane.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+interface IProps {
+    className?: string;
+    split: "horizontal" | "vertical";
+    defaultSize: number | string;
+    primary: "left" | "right";
+    children: React.ReactNode;
+}
+
+const SplitPane: React.FC<IProps> = ({ className, split, children }) => {
```
eanders-ms, line 14: react-common has a splitter already. i will combine them eventually.

react-common/components/util.tsx
```diff
@@ -46,7 +46,7 @@ export function fireClickOnEnter(e: React.KeyboardEvent<HTMLElement>) {
     }
 }
 
-export function classList(...classes: string[]) {
+export function classList(...classes: (string | undefined)[]) {
```
srietkerk, line 49: What does allowing an entry to be undefined allow us to do?

teachertool/src/App.tsx
```diff
@@ -54,17 +51,12 @@ function App() {
             <div className="ui large main loader msft"></div>
         </div>
     ) : (
-        <div className="app-container">
+        <>
```
srietkerk, line 54: What's the benefit of making this an empty tag rather than a div?

teachertool/src/components/MainPanel.tsx
```diff
@@ -0,0 +1,32 @@
...
+import SplitPane from "./SplitPane";
+
+interface IProps {}
+
+const MainPanel: React.FC<IProps> = () => {
+    return (
+        <div className={css["main-panel"]}>
+            <SplitPane split={"vertical"} defaultSize={"80%"} primary={"left"}>
+                {/* Left side */}
+                <>
```
srietkerk, line 18: Similar question to as above. Why do we want empty tags in components? Would having something like this cause problems for screen readers?

teachertool/src/components/SplitPane.tsx
srietkerk: I'm a little confused about this comment because all of the components in this PR are using the `export default ..` pattern at the bottom of the file. Were we using the `export default` in the component declaration and you're wanting to use the `export default` pattern separate from the component declaration?

teachertool/src/App.tsx
eanders-ms: The empty tag is shorthand for React Fragment. I moved the functionality this div was providing into MainPanel. But we still need a container here as a workaround for React's limitation of being unable to render a list of nodes at the top level of a component. React Fragments well-suited for this, because they make clear that the purpose of the node isn't to express UI, but to act solely as a container.

teachertool/src/components/MainPanel.tsx
eanders-ms: This is another instance of needing a logical container that effectively disappears at runtime. In this case, each fragment acts as an envelope to deliver content to each side of the split pane. After that, their job is done.

react-common/components/util.tsx
eanders-ms: This was always the intended functionality of this method, but teachertool uses strict typing and so I had to make it explicit.

eanders-ms: The method filters out undefined elements.

teachertool/src/components/SplitPane.tsx
eanders-ms: I think we shouldn't `export default` components at all, so that we have to import them by their given name. It's the difference between this:
```ts
import SplitPane from "./SplitPane";
```
and this:
```ts
import { SplitPane } from "./SplitPane";
```

I'm suggesting we go with the second option, because it will enforce component name consistency. With the first option, this is equally valid:
```
import Button from "./SplitPane";
```
The result is you've imported SplitPane and called it Button. We should close off this vector for confusion.

thsparks (approved):
Intrigued by this css modules thing. Will be interesting to try out.

thsparks:
> Intrigued by this css modules thing. Will be interesting to try out.

Slight concern it will make debugging css a bit more difficult since the actual class names are different from the ones in the code, but willing to try it out and see if that's actually the case.

eanders-ms:
> > Intrigued by this css modules thing. Will be interesting to try out.
> 
> Slight concern it will make debugging css a bit more difficult since the actual class names are different from the ones in the code, but willing to try it out and see if that's actually the case.

The generated class names are usually readable enough to figure out where they came from. css sourcemaps would be great. I don't think they exist. Here what `.main-panel` from MainPanel.module.css looks like: ```MainPanel_main-panel__F7Z6A```

teachertool/src/App.tsx
srietkerk: Oh, interesting, thanks for the information!

PR #9833, opened 2024-01-26
teacher tool - don't default export components

eanders-ms:
In the interest of long-term code maintainability, prefer not to use default export unless there is a compelling reason for it. Default exports have a few gotchas that I'd like to avoid:
* The name assigned to the import can be anything, and doesn't reflect what you're actually importing.
* They're not discoverable by IntelliSense. You have to open the file to see what you're actually importing. The filename isn't a reliable source of information.
* During the inevitable refactor/renaming, you must search the code and manually update every import to match the new component name. The compiler can't help you.

Sometimes a default export is the appropriate pattern to employ, but let's not make them the default export pattern we employ. (Did I break your parser? :))

### DIFF
--- a/teachertool/src/App.tsx
+++ b/teachertool/src/App.tsx
@@ -9,15 +9,15 @@ import * as NotificationService from "./services/notificationService";
 import { downloadTargetConfigAsync } from "./services/backendRequests";
 import { logDebug } from "./services/loggingService";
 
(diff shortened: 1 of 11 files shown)


srietkerk (approved):
Thanks for the explanations of the downfalls of `export default`, it's super helpful!

PR #9834, opened 2024-01-26
allow pxt cli to survive css build errors

eanders-ms:
When `pxt serve` is running and I make the slightest typo in a .less file, the less compiler errors and the cli dies. This adds a lot of friction during development due to the overhead of restarting `pxt serve`.

With this change, the cli will log the build failure and move on.


thsparks (approved):
💯

PR #9835, opened 2024-01-29
Add Validator Plans

thsparks:
This PR adds validator plans and sends those to the editor for evaluation (rather than a rubric). It also adjusts evaluation logic slightly so that, instead of sending all requirements at once, we send one plan at a time. This allows us to get results for each check separately, so fast-running checks can finish and be displayed while long-running checks are still in progress.

Validation plans are loaded into state very similarly to how the catalog gets loaded, with shared and target-specific files as well as additional test files. The main change here is in how we handle test catalog/plans. I've now put them in a test folder instead of having files with -test suffixes, which matches how we do it for tutorials and skillmaps. It's also a little tidier in the directory.

The general flow is:
1. Validator plans are defined in a json file and loaded into app state on startup
2. The user selects criteria from the catalog, which creates a criteria instance in their selected criteria list
3. When the user runs evaluate, we take each individual criteria instance and use it to create (and populate) a validator plan
4. The validator plan for that criteria instance is then sent to the editor for evaluation
5. The plan is parsed by the editor and executed accordingly
6. The request returns to the teacher tool, where the result is displayed

image


teachertool/src/utils/fileSystemHelpers.ts
```diff
@@ -0,0 +1,31 @@
...
+// convert teachertool/somefile.json to teachertool/test/somefile.json
+function getTestFileName(fileName: string) {
+    const parts = fileName.split("/");
+    parts.splice(parts.length - 1, 0, "test");
+    return parts.join("/");
+}
+
+// This allows us to take multiple lists stored in json docs files and concatenate them into
+// a single unified list. It also automatically adds entries from test files if the test catalog flag is set.
+export async function loadTestableCollectionFromFileAsync<T>(fileNames: string[], rootName: string): Promise<T[]> {
```
thsparks, line 14: I feel this code didn't end up generalizing very cleanly. I was torn on whether or not to keep the helper function or just have two transforms (loadCatalogAsync and loadValidatorPlansAsync) with very similar code. I ended up keeping it, but I'm happy to switch back if folks feel this is too messy and/or troublesome to maintain. I don't anticipate it being shared outside of those two scenarios.

pxtblocks/code-validation/runValidatorPlan.ts
```diff
@@ -0,0 +1,43 @@
...
+                            break;
+
+                        default:
+                            console.error(`Unrecognized validator: ${check.validator}`);
+                            resolve(false);
+                    }
+                })
+        );
+
+        const results = await Promise.all(checkRuns);
```
eanders-ms, line 23: Consider using `promisePoolAsync` here to control concurrency.

teachertool/src/utils/fileSystemHelpers.ts
```diff
@@ -0,0 +1,31 @@
...
+    const { state: teacherTool } = stateAndDispatch();
+
+    const files = teacherTool.flags.testCatalog ? fileNames.concat(fileNames.map(getTestFileName)) : fileNames;
+    let allResults: T[] = [];
+    for (const planFile of files) {
+        try {
+            const response = await fetch(planFile);
+            const content = await response.json();
```
eanders-ms, line 22: If it is acceptable to skip any 404s or other errors, check `response.ok` before calling `response.json()`.

teachertool/src/utils/fileSystemHelpers.ts
eanders-ms, line 14: This function belongs in services/backendRequests.ts

teachertool/src/transforms/runEvaluateAsync.ts
```diff
@@ -1,18 +1,79 @@
...
+    const evalRequests = teacherTool.selectedCriteria.map(
+        criteriaInstance =>
+            new Promise(async resolve => {
+                dispatch(Actions.setEvalResult(criteriaInstance.instanceId, CriteriaEvaluationResult.InProgress));
+
+                const plan = generateValidatorPlan(criteriaInstance);
+
+                if (!plan) {
+                    dispatch(Actions.clearEvalResult(criteriaInstance.instanceId));
+                    resolve(false);
```
eanders-ms, line 52: It may not be necessary, but I like to include a `return` when resolving the promise just to make it really clear the code will not continue executing from this point.
```suggestion
                    return resolve(false);
```

teachertool/src/transforms/runEvaluateAsync.ts
```diff
@@ -1,18 +1,79 @@
...
+                }
+
+                const planResult = await runValidatorPlanAsync(plan!);
```
eanders-ms, line 55: A return on the resolve above makes the non-null assertion unnecessary.
```suggestion
                const planResult = await runValidatorPlanAsync(plan);
```

pxtblocks/code-validation/runValidatorPlan.ts
```diff
@@ -0,0 +1,43 @@
...
+            (check) =>
+                new Promise<boolean>(async (resolve) => {
+                    switch (check.validator) {
+                        case "blocksExist":
+                            const result = runBlocksExistValidation(usedBlocks, check as BlocksExistValidatorCheck);
+                            resolve(result);
+                            break;
+
+                        default:
+                            console.error(`Unrecognized validator: ${check.validator}`);
```
eanders-ms, line 17: Avoid writing to console directly.
```suggestion
                            pxt.debug(`Unrecognized validator: ${check.validator}`);
```

eanders-ms (commented):
Looks good overall! Made a few minor suggestions.

eanders-ms (approved):
Looks great.

srietkerk (commented):
I know this is already merged, but I was looking through it, so I figured I'd add my comments.

docs/teachertool/validator-plans-shared.json
```diff
@@ -0,0 +1,35 @@
+{
+    "validatorPlans": [
+        {
+            ".desc": "Two different kinds of loops used",
```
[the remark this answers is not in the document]
thsparks: I don't think we've discussed it in detail, but I'd be against it personally. That said, it's the kind of thing we might need to poll teachers about (or examine our own curriculum). @kiki-lee might also have some insight.

All that said, we still need to find some time to finalize our criteria catalog for this v1...until then, I wouldn't consider any of this to be finalized content.

pxtblocks/code-validation/rubricCriteria.ts
```diff
@@ -31,21 +31,6 @@ namespace pxt.blocks {
...
-                Object.keys(result).forEach((blockId) => {
-                    finalResult[blockId] = result[blockId];
-                });
-            });
-        });
-        return { blockIdResults: finalResult } as EvaluationResult;
-    }
-
-
     function validateBlockSet(usedBlocks: Blockly.Block[], blockSet: BlockSet): pxt.Map<boolean> {
```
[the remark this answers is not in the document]
thsparks: Yes we probably want to separate them to some extent in future changes, but I think a lot of this file may be unused code anyway. Probably worth a pass just trimming out the content we no longer need.

pxtblocks/code-validation/runValidatorPlanAsync.ts
```diff
@@ -0,0 +1,39 @@
+namespace pxt.blocks {
+    export async function runValidatorPlanAsync(usedBlocks: Blockly.Block[], plan: ValidatorPlan): Promise<boolean> {
+        // Each plan can have multiple checks it needs to run.
+        // Run all of them in parallel, and then check if the number of successes is greater than the specified threshold.
+        // TBD if it's faster to run in parallel without short-circuiting once the threshold is reached, or if it's faster to run sequentially and short-circuit.
+        const startTime = Date.now();
+
+        const checkRuns = pxt.Util.promisePoolAsync(4, plan.checks, async (check: ValidatorCheckBase): Promise<boolean> => {
```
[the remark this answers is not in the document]
thsparks: Honestly I just looked at other callers into this function and saw that it's referenced in runtime.ts with max concurrency of 4. I thought that seemed reasonable for mid-low end machines (as opposed to 15, which is seen in the cli), but it's just a gut feeling. 

I can pull it out into a more descriptive variable in a separate change.

teachertool/src/components/EvalResultDisplay.tsx
```diff
@@ -1,27 +1,46 @@
...
                         return (
-                            <div className="result-block-id" key={id}>
-                                <p className="block-id-label">{id}:</p>
-                                <p className={result ? "positive-text" : "negative-text"}>
-                                    {result ? "passed" : "failed"}
+                            <div className="result-block-id" key={criteriaInstanceId}>
+                                <p className="block-id-label">
+                                    {getTemplateStringFromCriteriaInstanceId(criteriaInstanceId)}:
                                 </p>
+                                {result === CriteriaEvaluationResult.InProgress && <div className="common-spinner" />}
```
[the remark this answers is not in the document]
thsparks: This was another case where this UI is all going to change dramatically once we start working on the results page, so I just did something quick for now.

But to answer your immediate question: since InProgress is just another status for the result, I don't think it could be in both "InProgress" and "Fail" at the same time.

pxtblocks/code-validation/rubricCriteria.ts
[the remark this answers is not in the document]
srietkerk: For visibility, the code in this file was no longer needed, removed here: https://github.com/microsoft/pxt/pull/9838

PR #9836, opened 2024-01-29
teacher-tool: initial support for universal app theming

eanders-ms:
This PR introduces a unified CSS theming mechanism, shared between the main editor, react-common, and CRAs.

The immediate goal for this change is to enable teacher tool to receive its theming from the target it was launched from, but its benefits stretch beyond this. This system gives us a clear and simple way to implement new themes or fine-tune existing ones, and it provides a generalized solution to the problem of duplicated styling in CRAs, an area where we've been accumulating technical debt.

The approach involves:
1. Defining a well-known set of CSS style variables in the core web app.
2. Overriding these variables in targets.
3. Using these variables for theming in the editor, react-common, and CRAs.
4. Providing a utility function to set the current theme of the application.

This is just the initial commit to put the system in place for teacher tool to leverage. TODO:
- [ ] Define the actual set of CSS theme variables we want to support as our theme schema. What is there currently is placeholder.
- [ ] For the controller scenario: create an iframe message to set the editor theme.
- [ ] Propagate the use of the new variables across existing style definitions in core editor, targets, and react-common.
- [ ] Extract high contrast theme colors to `.theme-highcontrast`.
- [ ] Later: propagate the variables to multiplayer and skillmap when necessity arises.


eanders-ms:
Also switched to using sass/scss for teacher tool styling. We get it for free from create-react-app.

theme/themepacks.less
```diff
@@ -0,0 +1,62 @@
+/**************************************************************************
+    CSS theme variables
+    These are overridden by targets and used in: react-common, CRA webapps.
+    To add new theme variable:
+        1. Add variable to :root and .high-contrast below
+        2. Set the variable value in each target's theme/themepacks.less
+***************************************************************************/
+
+// TODO: Define the actual set of theme variables we want to support
```
eanders-ms, line 9: The CSS variables defined here are placeholder. I have a TODO to comb the editor css and extract the needed variables to here.

eanders-ms: While combing, I will also be referencing vscode's theme system to see what we can adopt from their schema. We probably have overlap, might be good to align where we can.

theme/themepacks.less
```diff
@@ -0,0 +1,62 @@
...
+    --pxt-c2-foreground: #FFFFC7;
+    --pxt-c2-border: #548687;
+    // color combo 3 (eg: secondary button)
+    --pxt-c3-background: #F2CEE6;
+    --pxt-c3-background-alpha-25: rgba(var(--pxt-c3-background), 0.25);
+    --pxt-c3-foreground: #0D1B1E;
+    --pxt-c3-border: #D8D8D8;
+}
+
+// TODO: Don't make bad choices here. Base shigh-contrast theme can be shared across targets and CRAs
```
thsparks, line 37:
```suggestion
// TODO: Don't make bad choices here. Base high-contrast theme can be shared across targets and CRAs

```

theme/themepacks.less
```diff
@@ -0,0 +1,62 @@
...
+    CSS theme variables
+    These are overridden by targets and used in: react-common, CRA webapps.
+    To add new theme variable:
+        1. Add variable to :root and .high-contrast below
+        2. Set the variable value in each target's theme/themepacks.less
+***************************************************************************/
+
+// TODO: Define the actual set of theme variables we want to support
+:root {
+    // Default to terrible choices so we can easily see when we're missing a theme variable in a target
```
thsparks, line 11: I'm fine with this for now, but I think long-term it'd be nice to have a "default theme" that actually looked nice, just to establish a baseline so that target themes don't have to override *everything* if they don't want to.

teachertool/src/global.scss
```diff
@@ -0,0 +1,347 @@
...
+.noclick {
+    pointer-events: none;
+    cursor: default;
+}
+
+/**********************************************/
+/********* THINGS THAT NEED HOMES *************/
+/**********************************************/
+
+.btn-primary {
```
thsparks, line 99: So is the idea that ultimately these kinds of styles (referencing the theme variables) will move into react common, so this doesn't have to be repeated in every app's global css?

eanders-ms: Yes, it means that classes like "primary" can move to react common.

teachertool/src/components/HeaderBar.tsx
```diff
@@ -12,10 +13,10 @@ export const HeaderBar: React.FC<HeaderBarProps> = () => {
 
     const getOrganizationLogo = () => {
         return (
-            <div className="ui item logo organization">
+            <div className={css["org"]}>
```
thsparks, line 16: Not quite following what this does. How is it different from just `className="org"`? Is it purely a intellisense/refactoring/maintainability thing, or does it have different functionality?

thsparks:
> Also switched to using sass/scss for teacher tool styling. We get it for free from create-react-app.

Are we using this anywhere else in pxt? And will it only affect teacher tool or will this become necessary for any app hooking into this theming system?

teachertool/src/global.scss
thsparks: Isn't primary already in react common?

teachertool/src/components/HeaderBar.tsx
eanders-ms: This is how module css classes are referenced. Higher up in the file you'll see the associated import:

```ts
import css from "./styling/HeaderBar.module.scss";
```

theme/themepacks.less
eanders-ms: Maybe a monochrome default.

teachertool/src/global.scss
eanders-ms: If so it will need to be updated use the color vars, so it can be changed at runtime.

thsparks (approved):

teachertool/src/global.scss
```diff
@@ -0,0 +1,334 @@
...
+    border-style: solid;
+    border-color: var(--pxt-content-foreground);
+    &:hover {
+        filter: none;
+        background-color: var(--pxt-headerbar-background-glass);
+    }
+}
+
+/*******************************/
+/******* CATALOG STYLING  ******/
```
srietkerk, line 131: Should the styling sections from here down (other than mobile, small desktop view) be separate modules or is there a reason that we left them in global styles?

teachertool/src/components/HeaderBar.tsx
```diff
@@ -29,28 +30,28 @@ export const HeaderBar: React.FC<HeaderBarProps> = () => {
...
                         </span>,
-                        <span className="name-short" key="org-name-short">
+                        <span className={css["name-short"]} key="org-name-short">
                             {appTheme.organizationShortText || appTheme.organizationText}
                         </span>,
                     ]
                 ) : appTheme.logo || appTheme.portraitLogo ? (
                     <img
-                        className={`ui ${appTheme.logoWide ? "small" : ""} logo`}
+                        className={css["logo"]}
```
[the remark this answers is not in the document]
eanders-ms: It may be necessary to support the a special layout for `logoWide` in the future. But if so, we should implement it using responsive classes defined in react-common, and avoid use of classes defined in core webapp css. They're not well-suited to use outside the webapp.

teachertool/src/components/HeaderBar.tsx
```diff
@@ -72,24 +73,20 @@ export const HeaderBar: React.FC<HeaderBarProps> = () => {
     };
 
     return (
-        <header className="menubar" role="banner">
```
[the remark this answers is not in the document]
eanders-ms: It was copy/pasted from somewhere else. I don't think wrapping a navbar in a header tag is an appropriate use of a header tag. We should add a role assignment to the navbar div though. `role="navigation"` may be right one.

PR #9838, opened 2024-01-31
Teacher tool: Removed old criteria logic in validation

srietkerk:
With Thomas' work with feeding the validator plans to the validation and evaluating one validator plan at a time, the old infrastructure that we had in place that parsed a rubric and criteria were no longer needed. I wanted to remove it before adding more validators.


eanders-ms (approved):
👍

PR #9840, opened 2024-01-31
Store Rubric in IndexedDB

thsparks:
This change stores the rubric in the indexed db and loads the last-active rubric automatically when the page loads. Currently only one rubric is really stored at a time, but the schema can support multiple if we want to evolve this over time.

Indexed DB has two tables:
1. Rubrics - where we store the rubrics, with the key set to the rubric name
2. Metadata - internal settings we may want to store, currently just lastActiveRubricName, but I imagine we may want to use this later to preserve preferences like auto-save and auto-run.

**A note on code reuse:**
The indexed db code is borrowed from Kiosk. I considered moving some of it into a shared class, but noticed we already have IDBWrapper in BrowserUtils. Then I tried switching to that, but I ultimately needed more code to manage the error handling than I saved by re-using the wrapper code, so I moved away from that approach as well. (It also wasn't working cleanly with custom key fields.) I think the right long-term move here is to evolve the IDBWrapper class to work better for these scenarios, but I think it'd require quite a lot of testing outside of the teachertool, so I was hesitant to make that change now.


pxtblocks/code-validation/runValidatorPlanAsync.ts
```diff
@@ -1,11 +1,14 @@
 namespace pxt.blocks {
+
```
thsparks, line 2: @srietkerk The changes in this file are responding to your comments in https://github.com/microsoft/pxt/pull/9835. They're not related to the indexed db change, I actually forgot I'd made them in this branch, but I suppose here they are 😅

teachertool/src/transforms/setRubricName.ts
```diff
@@ -0,0 +1,18 @@
...
+
+    if (oldName === name) {
+        return;
+    }
+
+    const newRubric = {
+        ...teacherTool.rubric,
+        name,
+    };
+    dispatch(Actions.setRubric(newRubric));
```
eanders-ms, line 17: May want to provide feedback here that the name was saved. Post a notification?

teachertool/src/transforms/tryLoadLastActiveRubricAsync.ts
```diff
@@ -0,0 +1,56 @@
...
+    const lastActiveRubric = await getLastActiveRubricAsync();
+
+    if (lastActiveRubric) {
+        logDebug(`Loading last active rubric '${lastActiveRubric.name}'...`);
+
+        const initialCriteriaCount = lastActiveRubric.criteria.length;
+        lastActiveRubric.criteria = lastActiveRubric.criteria.filter(validateCriteriaInstance);
+
+        if (lastActiveRubric.criteria.length !== initialCriteriaCount) {
+            postNotification(makeNotification("Some criteria could not be loaded.", 2000));
```
eanders-ms, line 49:
```suggestion
            postNotification(makeNotification(lf("Some criteria could not be loaded."), 2000));
```

teachertool/src/components/ActiveRubricDisplay.tsx
```diff
@@ -1,21 +1,36 @@
...
+
+    function handleConfirmName() {
+        setRubricName(inProgressName);
+    }
 
     return (
         <div className="rubric-display">
-            <h3>{lf("Rubric")}</h3>
-            {teacherTool.selectedCriteria?.map(criteriaInstance => {
+            <Input
```
eanders-ms, line 25: needs `ariaLabel`

teachertool/src/services/indexedDbService.ts
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    public deleteRubric(name: string): Promise<void> {
+        return this.deleteAsync(rubricsStoreName, name);
+    }
+}
+
+const db = new TeacherToolDb();
+
+let initializeAsync = async () => {
```
eanders-ms, line 103: This code actually doesn't work as intended. Due to the async, the function can be called multiple times before the reassignment happens. This is my bug :). I introduced it in Kiosk. One fix would be to have a global promise that does the db init and then resolves. getLastActiveRubricAsync, etc. would wait on that.

teachertool/src/services/indexedDbService.ts
```diff
@@ -0,0 +1,128 @@
...
+    const name = await db.getLastActiveRubricNameAsync();
+    if (name) {
+        rubric = await db.getRubric(name);
+    }
+
+    return rubric;
+}
+
+export async function saveRubricAsync(rubric: Rubric) {
+    await db.saveRubric(rubric);
```
eanders-ms, line 122: Need to await the db init here.

teachertool/src/services/indexedDbService.ts
eanders-ms: You could create a global promise that is the only source of the db instance. API calls have to await it to get the db instance. This would ensure no calls can be done on `db` before it's initialized.

teachertool/src/state/reducer.ts
```diff
@@ -59,16 +60,11 @@ export default function reducer(state: AppState, action: Action): AppState {
                 catalog: action.catalog,
             };
         }
-        case "SET_SELECTED_CRITERIA": {
+        case "SET_RUBRIC": {
+            updateStoredRubricAsync(state.rubric, action.rubric); // fire and forget, we don't need to wait for this to finish.
```
eanders-ms, line 64: For even more clarity of intention here:
```suggestion
            /*await*/ updateStoredRubricAsync(state.rubric, action.rubric); // fire and forget, we don't need to wait for this to finish.
```

teachertool/src/components/ActiveRubricDisplay.tsx
```diff
@@ -1,21 +1,36 @@
...
+                label={lf("Rubric Name")}
+                onChange={setInProgressName}
+                placeholder={lf("Rubric Name")}
+                initialValue={inProgressName}
+                onEnterKey={handleConfirmName}
+                onBlur={handleConfirmName}
```
eanders-ms, line 31: Normally you would need to include the `preserveValueOnBlur={true}` prop here, but you might be getting around it by setting `initialValue={inProgressName}`

thsparks: Ah, yeah. I think that was masking the issue. I'll still add it.

thsparks (commented):
PR Suggestions

teachertool/src/transforms/setRubricName.ts
thsparks: Hmm, a notification feels a little too heavy-weight here to me. Ideally, I think the user would just feel like the name is always saved once they type it, similarly to how criteria adding/removing works, no real need for a "saved" notification. Maybe I should just save the name whenever the text is changed and debounce it a little, rather than requiring them to hit enter and/or click away from the Input?

eanders-ms: Probably fine as is for now. Let's see what kind of feedback we get.

thsparks: I did end up adding a debounced input to handle this instead...but I can change it back if you want :)

thsparks: (also considered just wrapping the onChange call in pxt.Util.debounce but I got worried it wouldn't play nicely with React re-rendering things and possibly resetting fields)

teachertool/src/components/DebouncedInput.tsx
```diff
@@ -0,0 +1,28 @@
...
+// so they only fire once every `interval` milliseconds (defined in props)
+export const DebouncedInput: React.FC<DebouncedInputProps> = props => {
+    const timerId = useRef<NodeJS.Timeout | undefined>(undefined);
+
+    const onChangeDebounce = (newValue: string) => {
+        if (timerId.current) {
+            clearTimeout(timerId.current);
+        }
+
+        timerId.current = setTimeout(() => {
```
eanders-ms, line 18: What is the expected behavior if the component unmounts while a timeout is pending?

teachertool/src/components/DebouncedInput.tsx
```diff
@@ -0,0 +1,28 @@
+import { useRef } from "react";
+import { Input, InputProps } from "react-common/components/controls/Input";
+
+export interface DebouncedInputProps extends InputProps {
+    intervalMs?: number | undefined;
```
eanders-ms, line 5: Consider assigning the default value here so it's less buried and available to intellisense.
```ts
    intervalMs = 500
```

teachertool/src/components/DebouncedInput.tsx
thsparks: Hmm good question. I'll add some code to fire the onChange immediately if we're umounting and the timeout is pending.

thsparks:
This change stores the rubric in the indexed db and loads the last-active rubric automatically when the page loads. Currently only one rubric is really stored at a time, but the schema can support multiple if we want to evolve this over time.

Indexed DB has two tables:
1. Rubrics - where we store the rubrics, with the key set to the rubric name
2. Metadata - internal settings we may want to store, currently just lastActiveRubricName, but I imagine we may want to use this later to preserve preferences like auto-save and auto-run.

PR #9841, opened 2024-02-01
Teacher tool: added block comments validators

srietkerk:
I added two block comments validators to be used in the teacher tool but should also be usable in tutorials. 

criteria-comments

The first is a very rudimentary validator. It's for the use cases that users might not care where or how the comments are used, just that comments exist in the project.

The second is for when a users want to check that a specific type of block has a block comment on it. For example, a common ask I had from my professors was to add comments to every function definition. Users of the teacher tool might want to do the same, so if there is a block type specified, the validator will make sure that all blocks of that type has a comment on it. For teacher tool purposes, this is pre-defined only for function definitions at the moment, but an expansion would be when we introduce user-parameterized criteria, a user can allow the user to choose a block they would like to have comments on.

Some quick notes about decisions for how I wrote this code:
- I really like the idea of having this sort of mapping of one validator to one function run. As in, I like that even though the "any block comments" and "block comments on all this type of block" are pretty different, I like that they both live under the `blockCommentExists` validator and that running the `runValidateBlockCommentsExist` will sort out the more nitty-gritty details. I would love to hear other's thoughts on this, though.
- For validators, I think that we should maintain a common parameter pattern. An object is used in the `validateBlocksExist` function and I asked Joey why this was the choice and he explained that using an object as a parameter lends to "self-documenting" and can make it easier to add flags and the like into the validator down the road. I think these benefits are valuable, so I used this `object parameter` pattern in both of the new validators.
- In the comments above the new validators, I note why I returned the type that I did. Note that the "any block comments" validator returns all blocks that have comments in them. I do this because, in the future of the teacher tool, if we want to return more detail to the user as to why a validator passed or failed (and in this case in particular, passed), we can return all of the blocks where comments exist as an easy access point. The "block comments on all this type of block" scenario returns a list of the blocks that DO NOT have comments on them. This will not only be helpful for the teacher tool scenario, but also a tutorial scenario. If there is a tutorial step that includes having a user put a comment on a block, we would be able to give the hint of what block they're missing the comment on.

In action:
image

### DIFF
--- a/docs/teachertool/catalog-shared.json
+++ b/docs/teachertool/catalog-shared.json
@@ -15,13 +15,13 @@
         {
             "id": "76D86387-E3CF-4BEB-B62C-B811F5997631",
             "use": "block_comment_used",
-            "template": "At least one comment was left on a block in the code",
+            "template": "At least one block has comments",
             "docPath": "/teachertool"
         },
         {
             "id": "8F97C9A6-CF16-48B4-A84F-3105C24B20DE",
             "use": "functions_have_comments",
-            "template": "All function definitions have block comments",
+            "template": "All function definitions have comments",
             "docPath": "/teachertool"
         }
     ]
--- a/docs/teachertool/catalog-shared.json
+++ b/docs/teachertool/catalog-shared.json
@@ -13,9 +13,15 @@
             "docPath": "/teachertool"
         },
         {
-            "id": "49262A2B-C02A-43A2-BAD5-FCAC6AE9D465",
+            "id": "76D86387-E3CF-4BEB-B62C-B811F5997631",
             "use": "block_comment_used",
             "template": "At least one comment was left on a block in the code",
+            "docPath": "/teachertool"
+        },
+        {
+            "id": "8F97C9A6-CF16-48B4-A84F-3105C24B20DE",
+            "use": "functions_have_comments",
+            "template": "All function definitions have block comments",
             "docPath": "/teachertool"
         }
     ]
--- a/docs/teachertool/catalog-shared.json
+++ b/docs/teachertool/catalog-shared.json
@@ -11,6 +11,12 @@
             "use": "custom_function_called",
             "template": "A custom function exists and gets called",
             "docPath": "/teachertool"
+        },
+        {
+            "id": "49262A2B-C02A-43A2-BAD5-FCAC6AE9D465",
+            "use": "block_comment_used",
+            "template": "At least one comment was left on a block in the code",
+            "docPath": "/teachertool"
         }
     ]
 }
--- a/pxtblocks/code-validation/runValidatorPlanAsync.ts
+++ b/pxtblocks/code-validation/runValidatorPlanAsync.ts
@@ -15,7 +15,7 @@ namespace pxt.blocks {
                 case "blockCommentsExist":
                     return runValidateBlockCommentsExist(usedBlocks, check as BlockCommentsExistValidatorCheck);
                 case "specificBlockCommentsExist":
-                    return runValidateSpecificBlockCommentsExist(usedBlocks, check as SpecificBlockCommentExistsValidatorCheck);
+                    return runValidateSpecificBlockCommentsExist(usedBlocks, check as SpecificBlockCommentsExistValidatorCheck);
                 default:
                     pxt.debug(`Unrecognized validator: ${check.validator}`);
                     return false;
@@ -49,7 +49,7 @@ namespace pxt.blocks {
         return blockResults.passed;
     }
 
-    function runValidateSpecificBlockCommentsExist(usedBlocks: Blockly.Block[], inputs: SpecificBlockCommentExistsValidatorCheck): boolean {
+    function runValidateSpecificBlockCommentsExist(usedBlocks: Blockly.Block[], inputs: SpecificBlockCommentsExistValidatorCheck): boolean {
         const blockResults = validateSpecificBlockCommentsExist({ usedBlocks, blockType: inputs.blockType });
         return blockResults.passed;
     }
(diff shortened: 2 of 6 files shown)


pxtblocks/code-validation/runValidatorPlanAsync.ts
```diff
@@ -36,4 +37,13 @@ namespace pxt.blocks {
             blockResults.insufficientBlocks.length === 0;
         return success;
     }
+
+    function runValidateBlockCommentsExist(usedBlocks: Blockly.Block[], inputs: BlockCommentExistsValidatorCheck): boolean {
```
thsparks, line 41: I know you mentioned this in your PR description...but I don't think I fully understand the motivation of combining these two forms of validation into the same validator-plan-check "validator" label. I'm a little worried that having both of these utilize the same "validator" (meaning, the string we specified in the validator plan check) is going to lead to confusion down the road, especially since they're not actually using the same "validator" (meaning, the function getting called that actually runs the validation).

From a dev perspective, I worry the mismatch of "validator specified in validator plan check" and "actual validation functions" could be a point of confusion.

From a user perspective, if we imagine someone ever trying to author their own criteria, I think it's more confusing to have optional parameters that dramatically change the nature of the validation than it is to simply have two different validator choices. In this case, it's just the one parameter so it's potentially not too confusing, but as a pattern, I'd be concerned it could become untenable fairly quickly.

pxtblocks/code-validation/validateCommentsExist.ts
```diff
@@ -0,0 +1,20 @@
+
+namespace pxt.blocks {
+    // validates that one or more blocks comments are in the project
+    // returns the blocks that have comments for teacher tool scenario
+    export function validateBlockCommentsExist({ usedBlocks }: { usedBlocks: Blockly.Block[] }): Blockly.Block[] {
+        const commentBlocks = usedBlocks.filter((block) => !!block.getCommentText());
+        return commentBlocks;
```
thsparks, line 7: For similar reasons that we accept an object as the parameter, I think it'd make sense to return an object as well, so the returned information has a built-in label. And I also think having a "passed: boolean" field on that object would be very helpful. I wish we'd added one to the blocksExistValidator too, but that's not something to worry about in this change.

Ditto on the function below.

pxtblocks/code-validation/validateCommentsExist.ts
```diff
@@ -0,0 +1,20 @@
...
+    }
+
+    // validates that all of a specific block type have comments
+    // returns the blocks that do not have comments for a tutorial validation scenario
+    export function validateCommentsOnSpecificBlocksExist({ usedBlocks, blockType }: {
```
thsparks, line 12: nit: I'd recommend putting these into two separate files

docs/teachertool/catalog-shared.json
```diff
@@ -11,6 +11,18 @@
...
+        {
+            "id": "76D86387-E3CF-4BEB-B62C-B811F5997631",
+            "use": "block_comment_used",
+            "template": "At least one comment was left on a block in the code",
+            "docPath": "/teachertool"
+        },
+        {
+            "id": "8F97C9A6-CF16-48B4-A84F-3105C24B20DE",
+            "use": "functions_have_comments",
+            "template": "All function definitions have block comments",
```
thsparks, line 24: I worry the term "block comment" is going to be confusing for teachers, and in this case it's redundant anyway.
```suggestion
            "template": "All function definitions have comments",
```

docs/teachertool/catalog-shared.json
```diff
@@ -11,6 +11,18 @@
             "use": "custom_function_called",
             "template": "A custom function exists and gets called",
             "docPath": "/teachertool"
+        },
+        {
+            "id": "76D86387-E3CF-4BEB-B62C-B811F5997631",
+            "use": "block_comment_used",
+            "template": "At least one comment was left on a block in the code",
```
thsparks, line 18: Perhaps: "At least one block of code has comments"?

I wish we could just say "Program contains at least one comment" but I understand the hesitation since it doesn't actually check for workspace comments yet. Would be great if we could figure that out and update this accordingly.

pxtblocks/code-validation/runValidatorPlanAsync.ts
srietkerk: Yeah, that's a good point. I'll separate them. Thinking about it a bit, it's likely that the parameterized scenario will be similar to the "only check functions" scenario, so there might be some re-use there, but like you said, just checking for any and all comments is very different.

docs/teachertool/catalog-shared.json
thsparks: Or even shorter: "At least one block has comments"

pxtblocks/code-validation/runValidatorPlanAsync.ts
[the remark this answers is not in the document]
srietkerk: Are you saying to have something like the `blockCounts` object for this or just an optional `count` field? I though about using `blockCounts` but didn't because I wasn't sure how I would represent all of a block type in number form, but the optional field would work.

docs/teachertool/catalog-shared.json
srietkerk: Yeah, the thing I worry about is that if a teacher looks at a project that has workspace comments next to a function definition that pertains to the function and we fail the validation, it would seem broken.

thsparks: I'm not sure we preserve workspace comment locations well enough for that to work regardless. In this case I'd personally opt for the simpler wording and see if we get feedback later on.

I think the term "block comment" might just make teachers confused. To me, it sort of sounds like some comment that's an actual block in the code.

pxtblocks/code-validation/runValidatorPlanAsync.ts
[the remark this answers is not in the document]
srietkerk: Yeah, cool, that's what I was thinking about, too.

thsparks (approved):
LGMT, thanks!

PR #9842, opened 2024-02-01
teacher tool: add tabbed view

eanders-ms:
Adds tabbed view panel.

TODO in a future PR:
- Rubric scrolling needs attention.
- Deferring Home tab until we have an updated design.

image

### DIFF
--- a/teachertool/src/components/TabGroup.tsx
+++ b/teachertool/src/components/TabGroup.tsx
@@ -1,0 +1,37 @@
+import * as React from "react";
+// eslint-disable-next-line import/no-internal-modules
+import css from "./styling/TabGroup.module.scss";
+
+import { useContext } from "react";
+import { classList } from "react-common/components/util";
+import { Button } from "react-common/components/controls/Button";
+import { AppStateContext } from "../state/appStateContext";
+import { TabName } from "../types";
+import { setActiveTab } from "../transforms/setActiveTab";
+
+interface ITabProps extends React.PropsWithChildren<{}> {
+    name: TabName;
+}
+
+export const Tab: React.FC<ITabProps> = ({ children, name }) => {
+    const { state: teacherTool } = useContext(AppStateContext);
+
+    const onClick = () => {
+        setActiveTab(name);
+    };
+
+    const active = teacherTool.activeTab === name;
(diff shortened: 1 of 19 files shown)


teachertool/src/components/TabGroup.tsx
```diff
@@ -0,0 +1,37 @@
...
+    const onClick = () => {
+        setActiveTab(name);
+    };
+
+    const active = teacherTool.activeTab === name;
+    const classes = classList("tt-tabgroup-tab", active ? "tt-tabgroup-tab-active" : undefined);
+
+    return (
+        <Button className={classes} title={name} onClick={onClick}>
+            {children}
```
srietkerk, line 28: Why would a button/tab have children? I understand the use case for the TabGroup and the TabPanel, but I'm a little confused about the use here.

srietkerk (approved):
This looks good to me, I just had one question, but it's not blocking.

teachertool/src/components/styling/RubricWorkspace.module.scss
```diff
@@ -0,0 +1,17 @@
+.panel {
+    width: 100%;
+    height: 100%;
+    display: flex;
+    flex-direction: column;
+    overflow: hidden;
+    background-color: var(--pxt-page-background);
+    color: var(--pxt-page-foreground);
+
+    div[class*="tt-tabgroup"] {
```
thsparks, line 10: nit (and arguably not really an issue at all): is the tt prefix necessary with the module scoping this to just the RubricWorkspace anyway?

thsparks (approved):
LGTM

teachertool/src/components/TabGroup.tsx
eanders-ms: This is how react-common is designed. It can receive button content like label as a child element. We may want to utilize this flexibility to show a progress indicator on the results tab while the eval is running, or show an indicator that the results have updated.

teachertool/src/components/styling/RubricWorkspace.module.scss
eanders-ms: This is a case where the parent control wants to modify a child component's style. It is selecting the child element by class name, so we're not using a module css-generated name.

PR #9843, opened 2024-02-01
Add permission block for Code QL workflow

aznhassan:
Adding the `security-events: write` permission to the code ql workflow. This should fix the [errors we've seen in the workflow](https://github.com/microsoft/pxt/actions/runs/7746005421/job/21123225546#step:9:55).

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -22,6 +22,10 @@ on:
   schedule:
     - cron: '34 6 * * 0'
 
+  # Permissions required by OpenID Connect to access Azure.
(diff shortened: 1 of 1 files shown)


thsparks (approved):
Thanks for looking into this! Do we know why it needs write access? Seems odd to me for Code QL...

srietkerk (approved):
Thanks, Hassan!! I think we need to do this for all our repos, though rip.

aznhassan:
> Thanks for looking into this! Do we know why it needs write access? Seems odd to me for Code QL...

Probably because it's making a PUT request: 
image

PR #9844, opened 2024-02-02
Teacher tool: block in set validator added

srietkerk:
This PR establishes the block in set validator. I thought the `blocksInSetExist` terminology made more sense than `blockSetExists` since we're not checking that a group of blocks exist. With this validator, we're checking that some kinds of blocks from a set exist, so I'm hoping `blocksInSetExist` is more descriptive. I'm up for changing it, though, if people feel strongly.

For its use for now, I plugged it into the `custom_function_call` validator plan we have in an example in our OneNote. It's a good example of how we'd want to use two different validators together where we want to make sure that a function is made and it is either called and the return function is accessed or it's just called. The `blocksInSetExist` validator is meant to handle this "or" scenario. 

Here are some other possible scenarios that I thought:
- Creating and manipulating a sprite (as in, a teacher wants their student to move the sprite, scale  the sprite, remove the sprite, add effects to the sprite, and it can be any of those)
- Creating and using a variable, although this might be better situated with the `blocksExist` validator
- Maybe something with extensions, like we ensure that an extension was installed and a certain number of blocks from that extension was used, but it doesn't matter which of those blocks were used
- Something with arrays similar to the sprite scenario that I explained above

### DIFF
--- a/pxtblocks/code-validation/validateBlocksInSetExist.ts
+++ b/pxtblocks/code-validation/validateBlocksInSetExist.ts
@@ -1,0 +1,20 @@
+namespace pxt.blocks {
+    // validates that a combination of blocks in the set satisfies the required count
+    // returns the blocks that make the validator pass
+    export function validateBlocksInSetExist({ usedBlocks, blocks, count }: {
+        usedBlocks: Blockly.Block[],
+        blocks: string[],
+        count: number,
+    }): {
+        successfulBlocks: Blockly.Block[],
+        passed: boolean
+    } {
+        const successfulBlocks = [];
+        const enabledBlocks = usedBlocks.filter((block) => block.isEnabled());
+        for (const block of blocks) {
+            const blockInstances = enabledBlocks.filter((b) => b.type === block);
+            successfulBlocks.push(...blockInstances);
+        }
+        return { successfulBlocks, passed: successfulBlocks.length >= count };
+    }
+}
--- a/pxtblocks/code-validation/validateBlocksInSetExist.ts
+++ b/pxtblocks/code-validation/validateBlocksInSetExist.ts
@@ -1,19 +1,24 @@
 namespace pxt.blocks {
     // validates that a combination of blocks in the set satisfies the required count
     // returns the blocks that make the validator pass
-    export function validateBlocksInSetExist({ usedBlocks, blocks, count }: {
+    export function validateBlocksInSetExist({ usedBlocks, blockIdsToCheck, count, requireUnique }: {
         usedBlocks: Blockly.Block[],
-        blocks: string[],
+        blockIdsToCheck: string[],
         count: number,
+        requireUnique?: boolean
     }): {
         successfulBlocks: Blockly.Block[],
         passed: boolean
     } {
         const successfulBlocks = [];
         const enabledBlocks = usedBlocks.filter((block) => block.isEnabled());
-        for (const block of blocks) {
+        for (const block of blockIdsToCheck) {
             const blockInstances = enabledBlocks.filter((b) => b.type === block);
-            successfulBlocks.push(...blockInstances);
+            if (requireUnique && blockInstances.length >= 1) {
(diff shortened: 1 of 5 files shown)


pxtblocks/code-validation/validateBlocksInSetExist.ts
```diff
@@ -0,0 +1,20 @@
+namespace pxt.blocks {
+    // validates that a combination of blocks in the set satisfies the required count
+    // returns the blocks that make the validator pass
+    export function validateBlocksInSetExist({ usedBlocks, blocks, count }: {
+        usedBlocks: Blockly.Block[],
+        blocks: string[],
```
thsparks, line 6: nit: perhaps a slightly more descriptive name could be helpful here, i.e. 'blockIdsToCheck'

pxtblocks/code-validation/validateBlocksInSetExist.ts
```diff
@@ -0,0 +1,20 @@
+namespace pxt.blocks {
+    // validates that a combination of blocks in the set satisfies the required count
+    // returns the blocks that make the validator pass
+    export function validateBlocksInSetExist({ usedBlocks, blocks, count }: {
```
thsparks, line 4: Perhaps worth some discussion around whether or not we want to add a `requireUnique` flag here, which would make it so that each block in the set can only be counted once towards the desired total count. (i.e. instead of pushing all the blockInstances that match the id onto successfulBlocks, you push at most one, or zero if there are no matches).

Did you have any thoughts on this?

thsparks: The kinds of scenarios where this would be helpful for teacher-tool are very edge-case-y, since you can (usually) just do it with blocksExist and a threshold. The only scenarios I can think of where that is not the case would be if you wanted to check something like "two different kinds of loops" AND a third criteria that's not directly tied to that, all in the same criteria for some reason.

But it could be quite useful for tutorials, which don't have the same check/threshold architecture.

thsparks (approved):
One nit and some food for thought, but all-in-all, LGTM!

pxtblocks/code-validation/validateBlocksInSetExist.ts
srietkerk: Quick question, if I change the name here, should I update the name in the validator plans and the interface definition as well? I know changing the parameter in the function is a bit more pertinent since that's where we'd interface with the validator elsewhere, and having the name be more descriptive is helpful when using the validator. I'm wondering if there should be consistency underneath as well.

thsparks: I'd aim for consistency wherever possible :)

thsparks: Actually I take it back in this case. I think blockIdsToCheck is too wordy for the actual validator plan. For that, I'd go with blocks (as you have it) or blockIds. The "toCheck" is a little less necessary, since there's no notion of the "usedBlocks" in that scenario, which is part of what causes the possible-confusion in this function.

pxtblocks/code-validation/validateBlocksInSetExist.ts
thsparks: Yeah that's reasonable. Some functionality, like requiring different counts for different blocks, is still done better with blocksExist, but it's possible that as this validator becomes more fleshed out, its functionality could become a superset of what blocksExist can do, then the blocksExist could head towards deprecation...but I wouldn't worry about that too much right now. For now, I think having the flag as an option but not using it in teacherTool is fine. If we want to use it later as a shorthand so we're not typing out a bunch of individual blocksExist checks, then that's great too. But no need for it to be a part of this change.

srietkerk: Sounds good, I'll just add it as a parameter for the validator itself, then.

PR #9850, opened 2024-02-06
Convert pxteditor into modules

riknoll:
Yet another scary looking PR that is not actually changing anything!

This converts pxteditor into modules. Well, some of it. Most of the stuff in pxteditor is actually a bunch of types/interfaces. These have been moved into a new file called `localtypings/pxteditor.d.ts`.

The two functions that are patched by targets (`initExtensionsAsync` and `initFieldExtensionsAsync`) have been moved into pxtlib. The only change required in targets is to switch the `///` in their files from pointing to `built/pxteditor.d.ts` to `localtypings/pxteditor.d.ts`

As I was fixing all of the type references, I also cleaned up a lot of unused imports in the webapp.


eanders-ms (approved):
Looks good!

THEb0nny:
@riknoll I made the following changes as here and here...
https://github.com/microsoft/pxt-microbit/pull/5474/files
https://github.com/microsoft/pxt-arcade/pull/6343/files

image
image
image

but I get an error
image

eanders-ms:
@THEb0nny can you try running `gulp clean` in your pxt folder? You may have some old typings lying around.

riknoll:
@THEb0nny that error looks like you still have a type reference to the old pxteditor.d.ts. Make sure you update all type references to point to `localtypings/pxteditor.d.ts` instead of `built/pxteditor.d.ts`

PR #9851, opened 2024-02-07
teacher tool: add auto-run toggle

eanders-ms:
Added an auto-run toggle to the toolbar. Autorun will trigger an eval whenever the ~~catalog~~ rubric changes or a new project is loaded.

The toggle control introduced here is a quick-and-dirty solution using a button + icon. I intend to circle back to improve it, but making a nicely animated toggle that is also accessible is actually a pretty nuanced problem. Something to tackle later.

image


eanders-ms:
Test target: https://makecode.microbit.org/app/50500ab70fb7a019c7fb7e6c2356a818faa4ff3c-9b90f5510f--eval

eanders-ms:
UI needs to provide feedback when eval is running. I will address this in a separate PR. Some ideas:
- [ ] Change the play icon to a spinner while eval us running.
- [ ] Show current run status on the Results tab button (icons for pending, running, finished)

teachertool/src/components/RubricWorkspace.tsx
```diff
@@ -109,7 +124,7 @@ interface IProps {}
 export const RubricWorkspace: React.FC<IProps> = () => {
     return (
         <div className={css.panel}>
-            <Toolbar left={<WorkspaceTabButtons />} right={<WorkspaceToolbarButtons />} />
+            <Toolbar left={<WorkspaceTabButtons />} center={<WorkspaceAutorun />} right={<WorkspaceToolbarButtons />} />
```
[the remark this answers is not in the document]
eanders-ms: When right-justified, its position will shift left and right when switching tabs due to the changing number of buttons to the right of it, felt jarring. Always happy to run by designers.

react-common/components/controls/Button.tsx
```diff
@@ -21,6 +21,7 @@ export interface ButtonViewProps extends ContainerProps {
     ariaPosInSet?: number;
     ariaSetSize?: number;
     ariaSelected?: boolean;
+    ariaChecked?: boolean | "mixed";
```
thsparks, line 24: Hm, I thought mixed was more for those scenarios where you have a parent check box and some of the children are checked (it often gets filled in instead of checked). Should we default to undefined instead? Or are we just relying on the fact that aria-checked isn't normally supported on the button role anyway?

teachertool/src/components/RubricWorkspace.tsx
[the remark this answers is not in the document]
thsparks: Hmm, that's fair, but left-justified it looked like it should be an actual selected tab to me. Could we position it on the right such that any optional buttons are to the left of it, so it won't shift? Or we can leave as-is and get feedback.

react-common/components/controls/Button.tsx
eanders-ms: Thanks for calling this out. I'd misremembered. It was `aria-pressed` I wanted to expose: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#aria_state_information

teachertool/src/services/autorunService.ts
```diff
@@ -0,0 +1,18 @@
...
+import { isProjectLoaded } from "../state/helpers";
+import { runEvaluateAsync } from "../transforms/runEvaluateAsync";
+
+let autorunTimer: NodeJS.Timeout | null = null;
+
+export function poke() {
+    if (autorunTimer) {
+        clearTimeout(autorunTimer);
+    }
+    autorunTimer = setTimeout(() => {
```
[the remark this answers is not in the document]
thsparks: Another option would be to just manually call into this when the link changes and when the rubric changes.

react-common/components/controls/Button.tsx
eanders-ms: Fixed

teachertool/src/services/autorunService.ts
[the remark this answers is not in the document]
eanders-ms: It doesn't run every second. setTimeout runs the function just once.

teachertool/src/services/localStorage.ts
```diff
@@ -0,0 +1,48 @@
+import { ErrorCode } from "../types/errorCode";
+import { logError } from "./loggingService";
+
+const KEY_PREFIX = "teachertool";
+const AUTORUN_KEY = [KEY_PREFIX, "autorun"].join("/");
+
+function getValue(key: string, defaultValue?: string): string | undefined {
+    return localStorage.getItem(key) || defaultValue;
```
thsparks, line 8: Huh. Didn't know this existed. Wonder if I should switch the indexedDb meatdata table over to localstorage instead...

teachertool/src/transforms/removeCriteriaFromRubric.ts
```diff
@@ -14,6 +15,7 @@ export function removeCriteriaFromRubric(instance: CriteriaInstance) {
     };
 
     dispatch(Actions.setRubric(newRubric));
+    AutorunService.poke();
```
thsparks, line 18: This is fine, but if you wanted, I think you could also just put this in the reducer when the rubric is changed.

teachertool/src/services/localStorage.ts
eanders-ms: Capacity of local storage is severely limited. We should only use it to store small small things like strings and bools. In fact, I should remove the APIs for reading/writing JSON objects.

teachertool/src/services/autorunService.ts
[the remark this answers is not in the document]
thsparks: Oh I misunderstood, this isn't running every second, it's debouncing. You can disregard this comment :)

teachertool/src/transforms/removeCriteriaFromRubric.ts
thsparks: You're also already calling it in setRubric so it's fairly redundant here. But if you want to be explicit, I don't think it's harming anything.

eanders-ms: I tried that, but was getting very bizarre startup errors from react. The app would load to a blank page, with an error message in the console stating something about `stateAndDispatch` being accessed before initialization. I couldn't catch it in the debugger or figure out exactly what was happening. AFAICT, stateAndDispatch was not being referenced at all at startup. autorun does reference it in its timeout callback, and that is where the error was stemming from, but the function wasn't executing at startup. Bewildered, I moved the calls out to where they are now.

teachertool/src/services/localStorage.ts
thsparks: Yeah rubrics themselves should be in indexedDb, but the "lastActiveRubric" is just a key value pair to the name of the most recent rubric, which I think would be a good fit for this? Currently it's in its own metadata table in the indexedDb. I was originally going to say we could put autoRun in the same metadata table, but I think it actually makes more sense for me to move lastActiveRubric(Name) to localstorage.

thsparks (approved):
Some comments but nothing blocking

teachertool/src/services/localStorage.ts
eanders-ms: That makes sense.

teachertool/src/transforms/removeCriteriaFromRubric.ts
thsparks: Fair enough...setRubric is probably still sufficient, but since you have the debounce anyway, I think it's fine regardless.

eanders-ms: This codepath doesn't call the setRubric transform, but maybe it should.

thsparks: Ahh. Yeah, my bad. I think originally, I didn't do that because I was doing validation in setRubric, but I moved that out and never updated this code. I can do it in a separate change.

teachertool/src/components/RubricWorkspace.tsx
```diff
@@ -87,20 +102,20 @@ const WorkspaceToolbarButtons: React.FC = () => {
     const actionItems = getActionMenuItems(activeTab);
 
     return (
-        <ToolbarControlGroup>
+        <Toolbar.ControlGroup>
             {activeTab === "results" && (
-                <ToolbarButton icon="fas fa-print" title={lf("Print")} onClick={() => console.log("Print")} />
+                <Toolbar.Button icon="fas fa-print" title={lf("Print")} onClick={() => console.log("Print")} />
```
srietkerk, line 107: What's the decision for changing the architecture here?

teachertool/src/components/RubricWorkspace.tsx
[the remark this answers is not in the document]
srietkerk: Yeah, I think having the toggle be between the action menu and the play button is a good place for it, it also has association that way, too. The "play" button is for running the validation, and you can decide whether or not you want the validation to run automatically right next to it.

Another thought, what would people think of putting the toggle in the actions menu? I know it wouldn't be the greatest to keep something like that hidden, but I feel no matter where you put the toggle in the tab bar, it will make the bar look a bit crowded. 

Fine to wait for design feedback on this, just wanted to share my thoughts here.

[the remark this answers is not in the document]
eanders-ms: I hadn't considered putting it before the variable button set. That's a nice solution. Let me put up a new PR with that change....

PR #9852, opened 2024-02-07
Move code validation into modules

riknoll:
Moves all of the code validation into modules under pxteditor.

The typings that are shared with the teacher tool have been moved into `localtypings/codeValidation.d.ts`.

I haven't tested anything, but it all builds!


localtypings/validatorPlan.d.ts
```diff
@@ -1,4 +1,6 @@
-namespace pxt.blocks {
+/// <reference path="./pxtpackage.d.ts" />
+
+declare namespace pxt.blocks {
```
srietkerk, line 3: Why does the types definition still need a namespace?

riknoll: all of the other typings in localtypings used namespaces, not modules. i just matched the style. might be possible to make it a module, but you'll run into trouble when referencing it in the teachertool since it's outside the src folder

thsparks (approved):
I pulled the branch and tested it out, seems to be working well!

PR #9855, opened 2024-02-07
Do not allow duplicate criteria in a rubric

thsparks:
Do not allow the user to add duplicate criteria to their rubric. The general rule is: is the rubric has not been added, or if it has parameters (thus making it customizable), then it is available to be added. Otherwise, we hide it. If there are no available criteria left to add, we disable the Add Criteria button.

Upload target: https://makecode.microbit.org/app/65c45049c205f8e5185373296dd9acb3dca889aa-e2856c7f06--eval


eanders-ms (approved):
Looks great!

teachertool/src/components/AddCriteriaButton.tsx
```diff
@@ -0,0 +1,26 @@
...
+        () => getSelectableCatalogCriteria(teacherTool).length > 0,
+        [teacherTool.catalog, teacherTool.rubric]
+    );
+    return <Button
+            className="primary"
+            label={lf("Add Criteria")}
+            onClick={() => showModal("catalog-display")}
+            title={lf("Add Criteria")}
+            leftIcon="fas fa-plus-circle"
+            disabled={!hasAvailableCriteria}
```
srietkerk, line 24: Styling nit: I think the "disabled" version of the button should have the text and circle dimmed as well. Having the text and symbol be such a dark black makes it feel like I should still be able to click it.

srietkerk (approved):
One styling thing, but not blocking. Looks good!

teachertool/src/components/AddCriteriaButton.tsx
eanders-ms: Another styling nit we can fix later: It looks like button lost its "inline" styling. The design of this button is for it to look like it's part of the page, not a primary button:
image


This is consistent with the styling of the "Add notes" buttons on the Result page. On this page there may be many such buttons, so it would be visual overload to make them all primary:

image

PR #9857, opened 2024-02-08
Teacher Tool: Update Notifications

thsparks:
This swaps out our Teacher Tool notification system for one based on the multiplayer toast system. Since multiplayer uses tailwind, it would've been a pain to move everything into shared components. Instead, I've more-or-less copied everything over, but the functionality is the same.

image
image

Upload target: https://makecode.microbit.org/app/3b61276871d4d0bb9bcb84449a66303d9e161331-cc11238b8e--eval (For what it's worth, the way I've been triggering notifications is by including a correctly-formatted but non-existent share link into the project link input field.)

Note: the latest version of framer-motion broke the build, so I imported the same version as multiplayer.


srietkerk (approved):
Nice!

teachertool/src/utils/index.ts
```diff
@@ -1,13 +1,13 @@
 import { nanoid } from "nanoid";
-import { NotificationWithId } from "../types";
+import { ToastType, ToastWithId } from "../types";
 import { classList } from "react-common/components/util";
 
-export function makeNotification(message: string, duration: number): NotificationWithId {
+export function makeNotification(type: ToastType, text: string, timeoutMs: number = 5000): ToastWithId {
```
eanders-ms, line 5: ?
```suggestion
export function makeToast(type: ToastType, text: string, timeoutMs: number = 5000): ToastWithId {
```

eanders-ms (approved):

teachertool/src/utils/index.ts
eanders-ms: For toast colors, a nice touch might be to use the rgb values from the microbit's four colors:
```css
{
    green: #3AFFB3;
    cyan: #3ADCFF;
    yellow: #FFD43A;
    red: #FF3A54;
}
```

eanders-ms:
For toast colors, a nice touch might be to use the rgb values from the microbit's four colors:
```css
{
    green: #3AFFB3;
    cyan: #3ADCFF;
    yellow: #FFD43A;
    red: #FF3A54;
}
```

teachertool/src/components/styling/Toasts.module.scss
```diff
@@ -0,0 +1,133 @@
...
+    flex-direction: column;
+    margin-right: 0;
+    border: none;
+    border-radius: 0.25rem;
+    box-shadow: 0 1px 3px 0 rgba(0, 0, 0, 0.1), 0 1px 2px 0 rgba(0, 0, 0, 0.06);
+    overflow: hidden;
+    pointer-events: none;
+
+    &.success {
+        background-color: #81E6D9;
```
eanders-ms, line 26: These colors should be declared as vars in `theme/themepacks.less`, and overridden in targets.

eanders-ms: Eventually we will want to define color vars for toast border/background/text so that we can implement high-contrast toasts. But background color only is fine for this iteration.

eanders-ms: Also fine to push this to a future PR, track it as a todo.

thsparks:
> For toast colors, a nice touch might be to use the rgb values from the microbit's four colors:
> 
> ```css
> {
>     green: #3AFFB3;
>     cyan: #3ADCFF;
>     yellow: #FFD43A;
>     red: #FF3A54;
> }
> ```

I'm worried the red might be a little aggressive, but I could use that as the accent perhaps. The others look pretty good, though I'll probably change the yellow accent as well:
image

eanders-ms:
> > For toast colors, a nice touch might be to use the rgb values from the microbit's four colors:
> > ```css
> > {
> >     green: #3AFFB3;
> >     cyan: #3ADCFF;
> >     yellow: #FFD43A;
> >     red: #FF3A54;
> > }
> > ```
> 
> I'm worried the red might be a little aggressive, but I could use that as the accent perhaps. The others look pretty good, though I'll probably change the yellow accent as well: image

Wow those are bright! Maybe add a css filter to desaturate them a little?

thsparks:
> > > For toast colors, a nice touch might be to use the rgb values from the microbit's four colors:
> > > ```css
> > > {
> > >     green: #3AFFB3;
> > >     cyan: #3ADCFF;
> > >     yellow: #FFD43A;
> > >     red: #FF3A54;
> > > }
> > > ```
> > 
> > 
> > I'm worried the red might be a little aggressive, but I could use that as the accent perhaps. The others look pretty good, though I'll probably change the yellow accent as well: image
> 
> Wow those are bright! Maybe add a css filter to desaturate them a little?

I'll have to send a microbit PR but this will be the updated colors:
image

thsparks:
Follow-up PR: https://github.com/microsoft/pxt-microbit/pull/5478

PR #9858, opened 2024-02-08
Move Last Active Rubric Name to Browser Local Storage

thsparks:
I originally created a metadata table inside the Indexed DB to store this kind of info, but I didn't know about localstorage back then. Since we now have a service to handle that, I wanted to move the last active rubric name into the same area and get rid of that metadata table.

This did create a slight wrinkle: we now have some operations (like saving a rubric and getting the last active rubric) that require both localstorage and indexed db operations, and I didn't really want to call into one service from another, so I created a storageHelpers class to hold functions that handle those sorts of multi-step operations. I also gave the indexedDB service functions some arguably overly-specific names to avoid anyone accidentally calling into those instead of the helper.


teachertool/src/utils/storageHelpers.ts
```diff
@@ -0,0 +1,21 @@
+import { Rubric } from "../types/rubric";
+import * as LocalStorage from "../services/localStorage";
+import * as IndexedDb from "../services/indexedDbService";
+
+export async function getLastActiveRubricAsync(): Promise<Rubric | undefined> {
+    const lastActiveRubricName = LocalStorage.getLastActiveRubricName();
+    return await IndexedDb.getRubricFromIndexedDbAsync(lastActiveRubricName);
+}
+
+export async function saveRubricAsync(rubric: Rubric) {
```
eanders-ms, line 10: I think these should just be transforms.

teachertool/src/services/indexedDbService.ts
```diff
@@ -104,25 +78,23 @@ const getDb = (async () => {
     return db;
 })();
 
-export async function getLastActiveRubricAsync(): Promise<Rubric | undefined> {
+export async function getRubricFromIndexedDbAsync(name: string): Promise<Rubric | undefined> {
```
srietkerk, line 81: I saw that you mentioned this in your PR comment. I don't think this needs to have the `fromIndexedDb` in the name since you have to access the function from the indexedDbService. And the functions that you have for localstorage only involve the name, so I don't think there's really room for confusion there.

thsparks: For this one, I think it's less necessary since there isn't a helper for it, but I think it is necessary for the others (we don't want someone calling saveRubric here instead of the helper for it), and then it's just a matter of consistency.

teachertool/src/utils/storageHelpers.ts
thsparks: I can move it. I wasn't sure since they don't really touch app state and I don't think transforms usually return things...but I'm happy to move them if you think that's a better fit. I'm pretty ambivalent.

teachertool/src/services/indexedDbService.ts
srietkerk: Gotcha, that makes sense.

srietkerk (approved):
LGTM

teachertool/src/utils/storageHelpers.ts
eanders-ms: I guess these are a bit different than transforms. I want to avoid the unfettered propagation of helper methods around the codebase. We have them in state, and now in utils. It will become messier over time. Perhaps move them all to a `helpers` folder? I like that solution. Must contain the chaos. Fine as is for now. We can move them in a separate pr.

thsparks: I was under the impression that's sort of what `utils` was :)

eanders-ms: Do these helpers represent a storage service? Maybe that's the right place for them.

srietkerk: I agree; I view things in utils folders as things that are supposed to help in some way. I think the `utils` folder is the right place for this.

eanders-ms: In my mind, the definitions go like this:
* **util**: is a very basic utility function. It shouldn't have to call into services. Example: `debounce`.
* **helper**: makes reading or writing data easier, handling common edge cases internally. It can access services. Example: `getSafeRubricName`.
* **transform**: modifies application state.

srietkerk: I'm fine with any of these solutions.

eanders-ms: I'm partial to 3!

PR #9859, opened 2024-02-08
Teacher tool: added two simple variable validator rules

srietkerk:
I added these just really quick to have a quick validator for variables_set and variables_get, as a variable validator will be important for rubrics for tutorials. I think they might need to be a bit more nuanced  than this, but it's a start.


eanders-ms (approved):

docs/teachertool/catalog-shared.json
```diff
@@ -23,6 +23,18 @@
             "use": "functions_have_comments",
             "template": "All function definitions have comments",
             "docPath": "/teachertool"
+        },
+        {
+            "id": "D21D76A2-D9FD-4F9B-B0AC-973CB870EA78",
+            "use": "variable_set",
+            "template": "At least one variable's value is set",
```
thsparks, line 30:
```suggestion
            "template": "At least one custom variable is set",
```

docs/teachertool/catalog-shared.json
```diff
@@ -23,6 +23,18 @@
...
+            "docPath": "/teachertool"
+        },
+        {
+            "id": "0173898D-8A48-4266-AAB9-CE934471A734",
+            "use": "variable_accessed",
+            "template": "At least one variable is accessed",
```
thsparks, line 36:
```suggestion
            "template": "At least one variable is accessed",
```

docs/teachertool/validator-plans-shared.json
```diff
@@ -70,6 +70,32 @@
...
+        },
+        {
+            ".desc": "A variable's value is set",
+            "name": "variable_set",
+            "threshold": 1,
+            "checks": [
+                {
+                    "validator": "blocksExist",
+                    "blockCounts": {
+                        "variables_set": 1
```
thsparks, line 82: We may want to add an additional check here for variables_change, which will also work as a set in some cases (but keep threshold as 1).

thsparks (approved):
One nit and one suggestion, but LGTM on the whole!

docs/teachertool/validator-plans-shared.json
srietkerk: I would prefer to keep that as a separate check and wait for feedback on it. It makes sense that we assign a variable a value when the variable is created so there are less errors. However, allowing a `variables_change`  to act as a setter in a validation scenario feels like a dangerous thing to allow. The act of distinctively setting a variable value is a pretty important learning point for computer science.

srietkerk: I can add a validator for `variables_change`, too, though.

thsparks: Hm, I don't think we need to add a whole separate validator for it. Feel free to leave as-is for now.

thsparks: Validating that the program changes the value of a variable is actually non-trivial, since both "change" or "set" could do it, but "set" would have to be called twice on the same variable to achieve the intuitive result, which might be tricky to check. I'd only go down that rabbit hole if we get feedback asking for it.

srietkerk: Sounds good. I'll leave as is, then. Will definitely want to revisit the "variable changed" validation in the future.

PR #9860, opened 2024-02-08
Fix Shared Teacher Tool Docs

thsparks:
We weren't able to see any shared catalog entries in beta, and apparently this is because the /docs/ folder is for actual pxt docs (makecode.com/docs) and they're not considered shared. For shared docs, we need to move them into /common-docs/.

If we try to go to one of the pages now (https://makecode.microbit.org/teachertool/catalog-shared.json), you can go to more info and see that it does search common-docs:
image


srietkerk (approved):
Thanks!!!

eanders-ms (approved):
😎

PR #9861, opened 2024-02-09
Teacher Tool: Consolidate Calls to SetRubric

thsparks:
This change sends all `setRubric` calls through the `setRubric` transform (https://github.com/microsoft/pxt/blob/master/teachertool/src/transforms/setRubric.ts), so they don't call into the action directly. This is just a refactor; there is no change in functionality.

I didn't do this originally because the setRubric transform was doing a bunch of validation that would have been redundant, but I refactored that out and forgot to go in and change these calls. Now with this change, we can rely on the transform to poke autorun and we don't need to call that separately every time.


srietkerk (approved):
Nice!

PR #9863, opened 2024-02-12
teacher tool: initial homescreen

eanders-ms:
Adds initial home screen, with two cards: New Rubric and Import Rubric. Home screen is set up for additional carousels.

Also did some tidying and refinement:
- Extracted ticks to a central area
- Extracted shared `lf` strings to a central area
- Disabled the lint warnings about unassigned imports
- Added rubric name to header bar
- Made org and brand logos in header bar clickable
- Added project name to project toolbar
- Sorted out scrolling issues in rubric panel
- Check loaded project matches target
- Made the startup 🎓 toast a bit bigger

image

### DIFF
--- a/teachertool/src/components/HeaderBar.tsx
+++ b/teachertool/src/components/HeaderBar.tsx
@@ -1,11 +1,11 @@
 import * as React from "react";
 import { useContext } from "react";
-// eslint-disable-next-line import/no-internal-modules
 import css from "./styling/HeaderBar.module.scss";
 import { Button } from "react-common/components/controls/Button";
 import { MenuBar } from "react-common/components/controls/MenuBar";
 import { AppStateContext } from "../state/appStateContext";
 import { getSafeRubricName } from "../state/helpers";
+import { Ticks } from "../constants";
 
 interface HeaderBarProps {}
 
@@ -15,7 +15,7 @@ export const HeaderBar: React.FC<HeaderBarProps> = () => {
     const appTheme = pxt.appTarget?.appTheme;
 
     const brandIconClick = () => {
-        pxt.tickEvent("teachertool.brandicon");
+        pxt.tickEvent(Ticks.BrandIcon);
         // TODO: Link this
     };
 
@@ -75,7 +75,7 @@ export const HeaderBar: React.FC<HeaderBarProps> = () => {
     };
 
     const onHomeClicked = () => {
-        pxt.tickEvent("teachertool.home");
+        pxt.tickEvent(Ticks.HomeTab);
 
         // relprefix looks like "/beta---", need to chop off the hyphens and slash
         let rel = pxt.webConfig?.relprefix.substr(0, pxt.webConfig.relprefix.length - 3);
--- a/teachertool/src/components/HeaderBar.tsx
+++ b/teachertool/src/components/HeaderBar.tsx
@@ -14,7 +14,7 @@ export const HeaderBar: React.FC<HeaderBarProps> = () => {
 
     const appTheme = pxt.appTarget?.appTheme;
 
-    const brandIconClick = () => {
+    const onBrandIconClick = () => {
         pxt.tickEvent(Ticks.BrandLink);
         if (appTheme?.logoUrl) {
             window.open(appTheme.logoUrl);
@@ -50,7 +50,7 @@ export const HeaderBar: React.FC<HeaderBarProps> = () => {
                 className={css["brand"]}
                 aria-label={lf("{0} Logo", appTheme.boardName)}
                 role="menuitem"
-                onClick={brandIconClick}
+                onClick={onBrandIconClick}
             >
                 {appTheme.useTextLogo ? (
                     [
--- a/teachertool/src/components/HeaderBar.tsx
+++ b/teachertool/src/components/HeaderBar.tsx
@@ -15,13 +15,22 @@ export const HeaderBar: React.FC<HeaderBarProps> = () => {
     const appTheme = pxt.appTarget?.appTheme;
 
     const brandIconClick = () => {
-        pxt.tickEvent(Ticks.BrandIcon);
-        // TODO: Link this
+        pxt.tickEvent(Ticks.BrandLink);
+        if (appTheme?.logoUrl) {
+            window.open(appTheme.logoUrl);
+        }
     };
+
+    const onOrgClick = () => {
+        pxt.tickEvent(Ticks.OrgLink);
+        if (appTheme?.organizationUrl) {
+            window.open(appTheme.organizationUrl);
+        }
+    }
 
     const getOrganizationLogo = () => {
         return (
-            <div className={css["org"]}>
+            <div className={css["org"]} onClick={onOrgClick}>
                 {appTheme.organizationWideLogo || appTheme.organizationLogo ? (
                     <img
                         className={css["logo"]}
@@ -75,7 +84,7 @@ export const HeaderBar: React.FC<HeaderBarProps> = () => {
     };
 
     const onHomeClicked = () => {
-        pxt.tickEvent(Ticks.HomeTab);
+        pxt.tickEvent(Ticks.HomeLink);
 
         // relprefix looks like "/beta---", need to chop off the hyphens and slash
         let rel = pxt.webConfig?.relprefix.substr(0, pxt.webConfig.relprefix.length - 3);
@@ -101,7 +110,7 @@ export const HeaderBar: React.FC<HeaderBarProps> = () => {
                 <Button
                     className="menu-button"
                     leftIcon="fas fa-home large"
-                    title={lf("Return to the editor homepage")}
+                    title={lf("Open the MakeCode editor")}
                     onClick={onHomeClicked}
                 />
             </div>
--- a/teachertool/src/components/HeaderBar.tsx
+++ b/teachertool/src/components/HeaderBar.tsx
@@ -1,15 +1,23 @@
 import * as React from "react";
+import { useContext } from "react";
 // eslint-disable-next-line import/no-internal-modules
 import css from "./styling/HeaderBar.module.scss";
 import { Button } from "react-common/components/controls/Button";
 import { MenuBar } from "react-common/components/controls/MenuBar";
+import { AppStateContext } from "../state/appStateContext";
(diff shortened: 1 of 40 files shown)


teachertool/src/state/helpers.ts
```diff
@@ -43,10 +44,24 @@ export function verifyRubricIntegrity(rubric: Rubric): {
...
-export function isProjectLoaded(state: AppState) {
+export function isProjectLoaded(state: AppState): boolean {
     return !!state.projectMetadata;
 }
 
+export function isRubricLoaded(state: AppState): boolean {
+    return !!(state.rubric.criteria.length || state.rubric.name);
+}
+
+export function getSafeProjectName(state: AppState): string | undefined {
```
srietkerk, line 55: Is the "safe" addition to the function name just because even if we don't have anything in state, we'll serve the type that's needed back?

teachertool/src/state/helpers.ts
```diff
@@ -43,10 +44,24 @@ export function verifyRubricIntegrity(rubric: Rubric): {
...
+    if (state.projectMetadata) {
+        return state.projectMetadata.name ?? Strings.UntitledProject;
+    }
+}
+
+export function getSafeRubricName(state: AppState): string | undefined {
```
srietkerk, line 61: How would the return type here be undefined?

teachertool/src/transforms/confirmAsync.ts
```diff
@@ -0,0 +1,7 @@
+export async function confirmAsync(prompt: string) {
```
srietkerk, line 1: We had a discussion about helpers versus transforms in Thomas's PR and I feel like this is a similar stream.. This feels more like a helper to me than a state manager. Did we want to make a helpers folder or put something like this in utils? Or am I misunderstanding and this does manipulate state?

teachertool/src/transforms/resetRubricAsync.ts
```diff
@@ -0,0 +1,24 @@
+import { stateAndDispatch } from "../state";
+import { isRubricLoaded } from "../state/helpers";
+import * as Actions from "../state/actions";
+import { setRubric } from "./setRubric";
+import { confirmAsync } from "./confirmAsync";
+import { makeRubric } from "../utils";
+import { Strings } from "../constants";
+
+export async function resetRubricAsync(fromUserInteraction: boolean) {
```
srietkerk, line 9: What's the case that a rubric can be changed without user interaction?

teachertool/src/components/HeaderBar.tsx
```diff
@@ -77,13 +103,14 @@ export const HeaderBar: React.FC<HeaderBarProps> = () => {
             <div className={css["left-menu"]}>
                 {getOrganizationLogo()}
                 {getTargetLogo()}
+                {getRubricName()}
```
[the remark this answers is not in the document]
eanders-ms: This is per the current design. Let's address this design feedback in Figma. Maybe we can put the rubric name somewhere else, like on the Rubric tab itself, similar to vscode tabs. I'd like to iterate on it in Figma and come up with options.

teachertool/src/components/HomeScreen.tsx
```diff
@@ -0,0 +1,122 @@
...
+import "swiper/scss";
+import "swiper/scss/navigation";
+import "swiper/scss/mousewheel";
+
+const Welcome: React.FC = () => {
+    return (
+        <div className={css.welcome}>
+            <h1>{lf("Welcome to MakeCode Project Insights!")}</h1>
+            <p>
+                {lf("This tool is designed to help you evaluate student projects using a rubric.")}{" "}
```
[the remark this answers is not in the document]
eanders-ms: prettier added it. It enables the line break without loosing the trailing space character in the rendered html.

teachertool/src/components/ShareLinkInput.tsx
```diff
@@ -45,6 +43,7 @@ export const ShareLinkInput: React.FC<IProps> = () => {
                 onChange={onTextChange}
                 onEnterKey={onEnterKey}
                 preserveValueOnBlur={true}
+                autoComplete={false}
```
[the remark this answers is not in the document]
eanders-ms: Disables auto-correct. We never type dictionary words into this input box, and I was getting the red squiggley below the entered url.

teachertool/src/state/helpers.ts
eanders-ms: The addition of "safe" is to indicate the caller will get back a value that is always friendly to display. For untitled projects, it returns "Untitled Project" so that a falsey check on the result will never pass when a project is loaded.

teachertool/src/state/helpers.ts
eanders-ms: Future-proofing for when we have a "My Rubrics" carousel.

teachertool/src/transforms/confirmAsync.ts
eanders-ms: This is clearly not a helper in my view. A helper makes reading/writing data easier. I put it this transforms because it shows UI. I don't want to get to fine-grained or precise with these categories.

teachertool/src/transforms/resetRubricAsync.ts
eanders-ms: Good question! They can't, but it didn't feel safe to assume this transform would always be called in response to a user interaction. I was being defensive. Maybe overly so?

teachertool/src/components/HomeScreen.tsx
```diff
@@ -0,0 +1,122 @@
...
+};
+
+interface CardProps {
+    title: string;
+    className?: string;
+    icon?: string;
+    onClick: () => void;
+}
+
+const Card: React.FC<CardProps> = ({ title, className, icon, onClick }) => {
```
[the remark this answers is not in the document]
eanders-ms: pxt.CodeCard relies heavily on the old `sui` UI framework that we're trying to move away from with react-common.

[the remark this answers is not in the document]
eanders-ms: Also, this is just a portion of pxt.CodeCard's render method!
image

teachertool/src/components/HomeScreen.tsx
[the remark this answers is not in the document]
srietkerk: Oh, cool!

teachertool/src/state/helpers.ts
srietkerk: I still don't understand. Even if a rubric was created and listed in the "My Rubrics" carousel, won't they always have either a user-created name or a name that we generated to make sure all rubrics have names?

teachertool/src/state/helpers.ts
srietkerk: Makes sense. So this function returning `undefined` works in the case where there is no project loaded, though, because in that case, nothing would be displayed anyway.

teachertool/src/state/helpers.ts
eanders-ms: If it works like the editor then they won't always have a rubric loaded. There's also the first-time user experience to consider.

teachertool/src/transforms/resetRubricAsync.ts
eanders-ms: Removed. We can always bring it back if necessary.

teachertool/src/components/HeaderBar.tsx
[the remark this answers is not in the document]
eanders-ms: Added a work item to the board to track this.

teachertool/src/state/helpers.ts
srietkerk: Sure, but in the current design of this function, there are one of two things getting returned:
1. The name of the rubric that the user made
2. "Untitled Rubric" 
So there is no possibility that this would return undefined unless we changed how this function worked. Or is there something I'm misunderstanding?

eanders-ms: In the very near future there will be a case where it returns undefined. This is future proofing for that eminent scenario.

teachertool/src/components/ProjectWorkspace.tsx
```diff
@@ -1,17 +1,22 @@
...
+    const projectName = getSafeProjectName(teacherTool);
+    return projectName ? <div className={css.projectName}>{projectName}</div> : null;
+};
 
-export const ProjectWorkspace: React.FC<IProps> = () => {
+export const ProjectWorkspace: React.FC = () => {
     return (
         <div className={css.panel}>
-            <Toolbar />
+            <Toolbar center={<ProjectName />} />
```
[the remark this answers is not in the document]
eanders-ms: Ya, if/when the toggle goes in I'll move it over. Trivial to move!

teachertool/src/components/RubricWorkspace.tsx
```diff
@@ -108,9 +130,7 @@ const WorkspaceToolbarButtons: React.FC = () => {
     );
 };
 
-interface IProps {}
-
-export const RubricWorkspace: React.FC<IProps> = () => {
+export const RubricWorkspace: React.FC = () => {
```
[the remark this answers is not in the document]
eanders-ms: I actually want to make a pass at removing all the empty IProps interfaces. We can easily add them when we need them.

teachertool/src/components/RubricWorkspace.tsx
```diff
@@ -87,6 +108,7 @@ const WorkspaceToolbarButtons: React.FC = () => {
     const actionItems = getActionMenuItems(activeTab);
 
     const onAutorunChange = (checked: boolean) => {
+        pxt.tickEvent(Ticks.Autorun, { checked: checked ? "true" : "false" });
```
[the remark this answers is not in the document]
eanders-ms: I removed a different tick because it seemed extraneous, but I'd like to keep this one. I don't imagine teachers will be toggling it a lot, but I am curious to see how often it is toggled, and which toggle state is more popular over time.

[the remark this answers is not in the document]
srietkerk: Sounds good!

srietkerk (approved):

PR #9864, opened 2024-02-12
teacher tool: add scrollbar styling

eanders-ms:
Just for fun: match app scrollbar style to the blocks editor scrollbar.

image

### DIFF
--- a/teachertool/src/global.scss
+++ b/teachertool/src/global.scss
@@ -93,6 +93,34 @@ code {
 }
 
 /*******************************/
+/********** SCROLLBAR  *********/
+/*******************************/
+
+// Modeled after the scrollbar in the MakeCode blocks editor
+
+::-webkit-scrollbar {
+    width: 14px;
+    height: 14px;
+    background-color: rgba(0, 0, 0, 0);
+    margin: 2px;
+}
+
+::-webkit-scrollbar-thumb {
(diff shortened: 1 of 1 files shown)


teachertool/src/global.scss
```diff
@@ -92,6 +92,33 @@ code {
...
+
+::-webkit-scrollbar-thumb {
+    height: 0;
+    border: 2px solid rgba(0, 0, 0, 0);
+    border-radius: 99px;
+    background-clip: padding-box;
+    background-color: rgb(206, 205, 206);
+    &:hover {
+        // Style matching the standalone editor
+        //background-color: rgb(187, 187, 187);
```
srietkerk, line 115: Do we want to keep these commented styles?

srietkerk (approved):
Just one comment about some of the commented stuff, but other than that looks good!

I'm very excited about having nice-looking scroll bars!!

teachertool/src/global.scss
eanders-ms: Yes, I'd like to keep them because I don't know why the embedded editor's scrollbars look slightly different. If they ever end up updating to look like the standalone editor, we can just uncomment these lines. It took me longer than I'd like to admit to figure out the border/inset solution. :)

PR #9867, opened 2024-02-14
Teacher tool: field value validator added

srietkerk:
A validator that checks the field values of a block has been added to our validator library. Field values are a specific type of input for a block. The easiest type that I can demonstrate is the different operators that are associated with the `math_arithmetic` block. In this case, the field value _type_ is "OP" or, I'm guessing, the shorthand for operation. Some others that I've seen in `blocklycompiler.ts` are: `TEXT, EXPRESSION, VAR, BOOL, NAME, VALUE` and more. The "OP" field value for the `math_arithmetic` block can be `ADD, MINUS, MULTIPLY, DIVIDE` and more.

You'll see that the validation I made for `logic_compare` is looking for something very specific. The `logic_compare` blocks in the following program don't pass because the one that has the `EQ` "OP" type doesn't have two `math_number` blocks, and the other that has two numbers has a different "OP" type.

nums-equal

Edit: And to show that it can pass lol
nums-equal-pass

One thing to note about this approach is that if we want to support any type of field value, we need to know that field value's name and possible values it can be before we can create the validator for it. @riknoll chatted with me and let me know that this will require quite a bit of work because all of our custom blocks have field value names that are going to be tricky to find. He recommended a way for me to find the different field values, which isn't hard, just tedious. I'm planning on making a table in OneNote that will have the relationship between a block, it's field value names, and the variations that a specific field value can have. For now, I'm planning to build this table as I go, but we might reach a time when we'll want to just fill the table as earnestly as possible.

### DIFF
--- a/common-docs/teachertool/catalog-shared.json
+++ b/common-docs/teachertool/catalog-shared.json
@@ -51,7 +51,7 @@
         {
             "id": "DC564060-F177-46CC-A3AC-890CD8545972",
             "use": "num_compare_eq",
-            "template": "Check that two numbers are equal",
+            "template": "Compare two numbers for equality",
             "docPath": "/teachertool"
         }
     ]
--- a/common-docs/teachertool/catalog-shared.json
+++ b/common-docs/teachertool/catalog-shared.json
@@ -47,6 +47,12 @@
             "use": "device_random_used",
             "template": "The math random number generator block is used",
             "docPath": "/teachertool"
+        },
+        {
+            "id": "DC564060-F177-46CC-A3AC-890CD8545972",
+            "use": "num_compare_eq",
+            "template": "Check that two numbers are equal",
+            "docPath": "/teachertool"
         }
     ]
 }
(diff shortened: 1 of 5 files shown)


common-docs/teachertool/catalog-shared.json
```diff
@@ -47,6 +47,12 @@
             "use": "device_random_used",
             "template": "The math random number generator block is used",
             "docPath": "/teachertool"
+        },
+        {
+            "id": "DC564060-F177-46CC-A3AC-890CD8545972",
+            "use": "num_compare_eq",
+            "template": "Check that two numbers are equal",
```
eanders-ms, line 54:
```suggestion
            "template": "Compare two numbers for equality",
```

eanders-ms (approved):
Very nice! Made one small suggestion.

I agree building the table as we go is fine for now. We will want to put this info in a publicly visible doc at some point as part of documenting the process for writing validator plans.

Issue #9868, opened 2024-02-14
generalize hardcoded support for "Image" type in pxt


tballmsft:
There are a few places in emitter.ts where there is a dependence on the user-visible 
"Image" type (and some other helper functions related to arcade images).   Need to make these parameterizable by target, defaulting to values used by Arcade.

tballmsft:
https://github.com/microsoft/pxt/pull/9869

PR #9869, opened 2024-02-15
start on generalizing treatment of image

tballmsft:
there's quite a bit of hardcoding of arcade type "Image" which is user visible (not hidden in runtime) in pxt. This prevents renaming of the type, which is needed for use in pxt-microbit.  Various examples:

- image.ofBuffer
    - buildsprites.md
    - tilemap.ts
- "_Image"
   - emitter.ts
- "pxsim.image*"
   - backjs.ts


tballmsft:
https://github.com/microsoft/pxt/issues/9868

PR #9870, opened 2024-02-16
teacher tool: misc tidy stuff

eanders-ms:
Miscellaneous cleanup items I've been sitting on. I didn't want these included with other changes.


cli/cli.ts
```diff
@@ -2047,7 +2047,6 @@ async function buildSemanticUIAsync(parsed?: commandParser.ParsedCommand) {
         generateReactCommonCss("skillmap"),
         generateReactCommonCss("authcode"),
         generateReactCommonCss("multiplayer"),
-        generateReactCommonCss("teachertool")
```
thsparks, line 2050: I'm surprised. Did something change, or has this always been unnecessary?

teachertool/public/index.html
```diff
@@ -7,17 +7,16 @@
     <meta name="theme-color" content="#000000" />
     <meta
       name="description"
-      content="MakeCode Teacher Tool. Automatically analyze and evaluate projects."
+      content="MakeCode Project Insights. Designed to help teachers evaluate student projects using a rubric."
```
thsparks, line 10: FYI, don't need to block on this, but in our sync yesterday, there was some disagreement around the name Project Insights. We probably need to have a meeting with branding folks to hash this out.

cli/cli.ts
eanders-ms: It's always been unnecessary, and was causing the react-common css to be loaded twice. Duplicates of every class.

thsparks (approved):

eanders-ms:
I merged this rather quickly after opening it. @srietkerk please feel free to leave feedback if you had anything.

PR #9871, opened 2024-02-21
Teacher Tool: Update Rubric Display

thsparks:
Upload target: https://makecode.microbit.org/app/fd0f6bded821712f4dae4e3b350ae30256029d11-c72fad3c88--eval
<del>https://makecode.microbit.org/app/01cbf512a2c5c800b62a49e07045fb6b0021bcc0-892453ee9f--eval</del>

Screenshots:
image

image

image

image


eanders-ms:
It's a small layout thing, but can you adjust the rubric name input field so that it aligns with the other one?
image

eanders-ms:
When tabbing, the border/outline looks funny. Can you improve it?
image

teachertool/src/components/CriteriaTable.tsx
```diff
@@ -0,0 +1,69 @@
...
+            </div>
+        </div>
+    ) : null;
+};
+
+interface CriteriaTableProps {}
+const CriteriaTableControl: React.FC<CriteriaTableProps> = ({}) => {
+    const { state: teacherTool } = useContext(AppStateContext);
+
+    return teacherTool.rubric.criteria && teacherTool.rubric.criteria.length > 0 ? (
```
eanders-ms, line 45:
```suggestion
    return teacherTool.rubric.criteria?.length > 0 ? (
```

teachertool/src/components/CriteriaTable.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { Button } from "react-common/components/controls/Button";
+import css from "./styling/CriteriaTable.module.scss";
+import React from "react";
+
+interface CriteriaInstanceDisplayProps {
+    criteriaInstance: CriteriaInstance;
+}
+const CriteriaInstanceRow: React.FC<CriteriaInstanceDisplayProps> = ({ criteriaInstance }) => {
+    const catalogCriteria = getCatalogCriteriaWithId(criteriaInstance.catalogCriteriaId);
+    if (!catalogCriteria) {
```
eanders-ms, line 19: This check is redundant with the conditional render below.

teachertool/src/components/CriteriaTable.tsx
```diff
@@ -0,0 +1,69 @@
...
+        return null;
+    }
+
+    return catalogCriteria ? (
+        <div className={css["criteria-instance-display"]} role="row" tabIndex={0}>
```
eanders-ms, line 24: Why is this tabbable?

teachertool/src/components/CriteriaTable.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+    return teacherTool.rubric.criteria && teacherTool.rubric.criteria.length > 0 ? (
+        <div className={css["criteria-table"]} role="table" aria-label={Strings.Criteria}>
+            <div className={css["criteria-header"]} role="row">
+                <div className={classList(css["cell"], css["criteria-text-cell"])} role="columnheader">
+                    {Strings.Criteria}
+                </div>
+                <div className={classList(css["cell"], css["criteria-action-menu-cell"])} role="columnheader">
+                    {/* Intentionally left empty */}
+                </div>
```
srietkerk, line 53: Do we need this closing tag here? Wondering if we can just have <div .... /> for this cell. Or do we want to leave this for now because we might have an action menu here?

teachertool/src/components/CriteriaTable.tsx
```diff
@@ -0,0 +1,69 @@
...
+            </div>
+            <div className={css["criteria-table-body"]}>
+                {teacherTool.rubric.criteria.map(criteriaInstance => {
+                    if (!criteriaInstance) return null;
```
srietkerk, line 57: Is this possible? I know it's good to be careful, but I just wanted to check in about the possible scenario that a rubric entry is falsey.

teachertool/src/style.overrides/Input.scss
```diff
@@ -0,0 +1,50 @@
+.common-input-wrapper {
+
+    div[class*="common-input-group"] {
+        border-radius: 99px;
```
srietkerk, line 4: I know this is just a border radius, but 99px seems like a high value, like we should be using something like rem with pixel values that high.

eanders-ms: This is a common hack to put circular borders on rectangular elements. The browser will adjust the actual border radius to match the element size.

teachertool/src/components/CriteriaTable.tsx
thsparks: Originally thought it was needed for table navigation, but I looked into it a little more and it seems like the expectation is to use ctrl+alt_arrows to navigate a table which works without this, so I'll remove.

teachertool/src/components/CriteriaTable.tsx
thsparks: It's not strictly necessary, I just wanted it so I could have my intentionally left empty comment.

teachertool/src/components/CriteriaTable.tsx
thsparks: Removed (but kept check in CriteriaInstanceRow in case it gets reused somewhere else).

thsparks:
> When tabbing, the border/outline looks funny. Can you improve it? image

This is no longer focus-able now that I've removed the tab index.

srietkerk (approved):
LGTM!

teachertool/src/components/CriteriaTable.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+interface CriteriaTableProps {}
+const CriteriaTableControl: React.FC<CriteriaTableProps> = ({}) => {
+    const { state: teacherTool } = useContext(AppStateContext);
+
+    return teacherTool.rubric.criteria?.length > 0 ? (
+        <div className={css["criteria-table"]} role="table" aria-label={Strings.Criteria}>
+            <div role="rowgroup">
+                <div className={css["criteria-header"]} role="row">
+                    <div className={classList(css["cell"], css["criteria-text-cell"])} role="columnheader" tabIndex={0}>
```
eanders-ms, line 53: The "Criteria" header is still navigable by tab.
image

PR #9875, opened 2024-02-22
Allow a Rubric with an Empty Name to be Loaded

thsparks:
Fixes https://github.com/orgs/microsoft/projects/1046/views/1?pane=issue&itemId=53913156 ("bug: criteria don't persist for unnamed rubric.")

### DIFF
--- a/teachertool/src/services/storageService.ts
+++ b/teachertool/src/services/storageService.ts
@@ -152,7 +152,7 @@ export async function getRubric(name: string): Promise<Rubric | undefined> {
     const db = await getDb;
 
     let rubric: Rubric | undefined = undefined;
-    if (name) {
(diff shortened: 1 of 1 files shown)


teachertool/src/services/storageService.ts
```diff
@@ -152,7 +152,7 @@ export async function getRubric(name: string): Promise<Rubric | undefined> {
     const db = await getDb;
 
     let rubric: Rubric | undefined = undefined;
-    if (name) {
+    if (name !== undefined) {
```
eanders-ms, line 155: I think this conditional is unnecessary because `name` cannot be undefined. Maybe just remove it entirely?

eanders-ms (approved):

PR #9878, opened 2024-02-22
Fix asset editor for non-arcade targets

riknoll:
This fixes the image editor for non-arcade targets like micro:bit and minecraft (the pixel art extension)

Includes three fixes:
1. Removes the name field from the bottom bar of the asset editor when assetEditor isn't set in pxtarget.json
2. Fixes palettes loading from pxt.json when using compile variants (i.e. in micro:bit)
3. Hides the gallery tab in the image editor when there are no gallery assets


tballmsft (approved):
Looks good to me. Thanks, Richard!

srietkerk (approved):
Nice!

PR #9879, opened 2024-02-22
Teacher tool: Added hidden flag for catalog criteria

srietkerk:
As we start to make rubrics for some tutorials, it's going to come up that we don't want those very specific rubric-based criteria to show up in the catalog list. This flag will make it so if you specify `"hideInCatalog": true` in the `catalog__.json` entry, then that criteria will not show up in the modal.

### DIFF
--- a/teachertool/src/types/criteria.ts
+++ b/teachertool/src/types/criteria.ts
@@ -6,6 +6,7 @@ export interface CatalogCriteria {
     description: string | undefined; // More detailed description
     docPath: string | undefined; // Path to documentation
     parameters: CriteriaParameter[] | undefined; // Any parameters that affect the criteria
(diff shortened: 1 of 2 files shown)


teachertool/src/types/criteria.ts
```diff
@@ -6,6 +6,7 @@ export interface CatalogCriteria {
     description: string | undefined; // More detailed description
     docPath: string | undefined; // Path to documentation
     parameters: CriteriaParameter[] | undefined; // Any parameters that affect the criteria
+    hideInCatalog?: boolean; // Whether the criteria should be hidden in the catalog
```
thsparks, line 9:
```suggestion
    hideInCatalog?: boolean; // Whether the criteria should be hidden in the user-facing catalog
```

eanders-ms (approved):

PR #9885, opened 2024-02-23
teacher tool: add home screen carousels

eanders-ms:
Adds additional carousels to the teacher tool home screen. I will also open a companion PR in the microbit repo.

There are a few TODOs remaining that I'll address in a future PR:
- [ ] Refactor HomeScreen.tsx. I will pull the Cards out into their own files.
- [ ] backendRequests.fetchJsonDocAsync may need additional work to point at the correct CDN origin in production. But maybe not. The backend may already route it properly.
- [ ] I added a lot of cards to the carousels! I did this to get a sense of what it will feel like when filled out. We will need to whittle them down to the ones we can implement in time.

image

### DIFF
--- a/teachertool/src/components/HomeScreen.tsx
+++ b/teachertool/src/components/HomeScreen.tsx
@@ -1,17 +1,22 @@
+import "swiper/scss";
+import "swiper/scss/navigation";
+import "swiper/scss/mousewheel";
+
 import * as React from "react";
+import { useContext, useState } from "react";
 import css from "./styling/HomeScreen.module.scss";
 import { Link } from "react-common/components/controls/Link";
 import { Button } from "react-common/components/controls/Button";
 import { classList } from "react-common/components/util";
 import { showModal } from "../transforms/showModal";
 import { resetRubricAsync } from "../transforms/resetRubricAsync";
+import { loadRubricAsync } from "../transforms/loadRubricAsync";
 import { Constants, Strings, Ticks } from "../constants";
 import { Swiper, SwiperSlide } from "swiper/react";
 import { Mousewheel, Navigation } from "swiper";
-
-import "swiper/scss";
-import "swiper/scss/navigation";
-import "swiper/scss/mousewheel";
+import { AppStateContext } from "../state/appStateContext";
+import { CarouselCardSet, RequestStatus, CarouselRubricResourceCard } from "../types";
+import { useJsonDocRequest } from "../hooks/useJsonDocRequest";
 
 const Welcome: React.FC = () => {
     return (
@@ -27,14 +32,14 @@ const Welcome: React.FC = () => {
     );
 };
 
-interface CardProps {
+interface IconCardProps {
     title: string;
     className?: string;
     icon?: string;
     onClick: () => void;
 }
 
-const Card: React.FC<CardProps> = ({ title, className, icon, onClick }) => {
+const IconCard: React.FC<IconCardProps> = ({ title, className, icon, onClick }) => {
     return (
         <div className={css.cardContainer}>
             <Button className={classList(css.cardButton, className)} title={title} onClick={onClick}>
@@ -46,6 +51,60 @@ const Card: React.FC<CardProps> = ({ title, className, icon, onClick }) => {
                     )}
                     <div className={css.cardTitle}>
                         <h3>{title}</h3>
+                    </div>
+                </div>
+            </Button>
+        </div>
+    );
+};
+
+interface LoadingCardProps {
+    delay?: boolean;
+}
+
+const LoadingCard: React.FC<LoadingCardProps> = ({ delay }) => {
+    return (
+        <div className={css.cardContainer}>
+            <Button
+                className={classList(css.cardButton, css.loadingGradient, delay ? css.loadingGradientDelay : undefined)}
+                title={""}
+                onClick={() => {}}
+            >
+                <div className={css.cardDiv}>
+                    <div className={css.loadingGradient}></div>
+                </div>
+            </Button>
+        </div>
+    );
+};
+
+interface RubricResourceCardProps {
+    cardTitle: string;
+    imageUrl: string;
+    rubricUrl: string;
+}
+
+const RubricResourceCard: React.FC<RubricResourceCardProps> = ({ cardTitle, imageUrl, rubricUrl }) => {
+    const onRubricClickedAsync = async () => {
+        pxt.tickEvent(Ticks.LoadRubric, { rubricUrl });
+        await loadRubricAsync(rubricUrl);
+    };
+    return (
+        <div className={css.cardContainer}>
+            <Button
+                className={classList(css.cardButton, css.rubricResource)}
+                title={cardTitle}
+                onClick={onRubricClickedAsync}
+            >
+                <div
+                    className={classList(css.cardDiv)}
+                    style={{
+                        backgroundImage: `url("${window.location.origin}${imageUrl}")`,
+                        backgroundSize: "cover",
+                    }}
+                >
+                    <div className={classList(css.cardTitle, css.rubricResourceCardTitle)}>
+                        <h3>{cardTitle}</h3>
                     </div>
                 </div>
             </Button>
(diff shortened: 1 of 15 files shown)


teachertool/src/components/HomeScreen.tsx
```diff
@@ -27,14 +32,14 @@ const Welcome: React.FC = () => {
     );
 };
 
-interface CardProps {
+interface IconCardProps {
```
srietkerk, line 35: Just wondering if we should have some inheritance for these cards. Are there going to be any shared properties between all these different props that we want to extract? It may be unnecessary more work, but I figured I'd start a conversation about it. For example, it seems like all cards should have titles, which two out of the three do, and I'm wondering if we should extract that into a more generic class.

teachertool/src/components/HomeScreen.tsx
```diff
@@ -53,6 +58,60 @@ const Card: React.FC<CardProps> = ({ title, className, icon, onClick }) => {
     );
 };
 
+interface LoadingCardProps {
+    delay?: boolean;
+}
+
+const LoadingCard: React.FC<LoadingCardProps> = ({ delay }) => {
+    return (
+        <div className={css.cardContainer}>
```
srietkerk, line 67: Is this just a different way of doing something like `css["notice-label-background"]`?

teachertool/src/components/HomeScreen.tsx
```diff
@@ -53,6 +58,60 @@ const Card: React.FC<CardProps> = ({ title, className, icon, onClick }) => {
...
+};
+
+interface RubricResourceCardProps {
+    cardTitle: string;
+    imageUrl: string;
+    rubricUrl: string;
+}
+
+const RubricResourceCard: React.FC<RubricResourceCardProps> = ({ cardTitle, imageUrl, rubricUrl }) => {
+    const onRubricClickedAsync = async () => {
```
srietkerk, line 88: I feel like this function can be renamed. I find it a bit confusing to read `onRubricClicked` when the user really isn't clicking on a rubric, but on the card to load a rubric. Being that the function is in the `RubricResourceCard` component, can it just be `onCardClicked`?

teachertool/src/components/HomeScreen.tsx
```diff
@@ -53,6 +58,60 @@ const Card: React.FC<CardProps> = ({ title, className, icon, onClick }) => {
...
+        <div className={css.cardContainer}>
+            <Button
+                className={classList(css.cardButton, css.rubricResource)}
+                title={cardTitle}
+                onClick={onRubricClickedAsync}
+            >
+                <div
+                    className={classList(css.cardDiv)}
+                    style={{
+                        backgroundImage: `url("${window.location.origin}${imageUrl}")`,
```
srietkerk, line 102: Is there a way to point at the right image without using `window.location.origin`?

teachertool/src/components/HomeScreen.tsx
```diff
@@ -53,6 +58,60 @@ const Card: React.FC<CardProps> = ({ title, className, icon, onClick }) => {
...
+    return (
+        <div className={css.cardContainer}>
+            <Button
+                className={classList(css.cardButton, css.rubricResource)}
+                title={cardTitle}
+                onClick={onRubricClickedAsync}
+            >
+                <div
+                    className={classList(css.cardDiv)}
+                    style={{
```
srietkerk, line 101: Is it better to have the style here instead of in an scss file?

teachertool/src/components/HomeScreen.tsx
```diff
@@ -112,11 +173,76 @@ const GetStarted: React.FC = () => {
...
+
+    useJsonDocRequest(cardsUrl, setFetchStatus, setCardSet);
+
+    return (
+        <>
+            <div className={css.carouselRow}>
+                <div className={css.rowTitle}>
+                    <h2>{title}</h2>
+                </div>
+                {(fetchStatus === "loading" || fetchStatus === "error") && (
```
srietkerk, line 193: Is there a plan to split this out in the future? Or maybe have retry logic? I feel like it would be a bit frustrating if you think that a card is going to load and it just never does.

teachertool/src/hooks/useJsonDocRequest.ts
```diff
@@ -0,0 +1,29 @@
+import { useEffect, useState } from "react";
+import { RequestStatus } from "../types";
+import { fetchJsonDocAsync } from "../services/backendRequests";
+
+export function useJsonDocRequest<T>(
+    url: string,
+    statusCb: (status: RequestStatus) => void,
```
thsparks, line 7: Is "Cb" callback? If so, nit request to simply spell it out. Took me a little while to figure out what it was.

teachertool/src/services/backendRequests.ts
```diff
@@ -2,6 +2,21 @@ import { stateAndDispatch } from "../state";
 import { ErrorCode } from "../types/errorCode";
 import { logError } from "./loggingService";
 
+export async function fetchJsonDocAsync<T = any>(url: string): Promise<T | undefined> {
+    try {
+        // TODO: Prepend CDN origin if not localhost
+        const response = await fetch(url);
```
thsparks, line 8: I'm wondering if we should update loadTestableCollectionFromDocsAsync to call into this as well on line 74. Can be done in a separate change, if preferred.

teachertool/src/services/backendRequests.ts
```diff
@@ -2,6 +2,21 @@ import { stateAndDispatch } from "../state";
 import { ErrorCode } from "../types/errorCode";
 import { logError } from "./loggingService";
 
+export async function fetchJsonDocAsync<T = any>(url: string): Promise<T | undefined> {
+    try {
+        // TODO: Prepend CDN origin if not localhost
```
thsparks, line 7: So far in beta or uploaded targets, I don't think we've seen any CDN-related issues for the doc files w/ the catalog and validator plans (though I could be missing something). Are you anticipating issues specific to prod?

teachertool/src/components/HomeScreen.tsx
```diff
@@ -112,11 +173,76 @@ const GetStarted: React.FC = () => {
...
+                    <Carousel>
+                        <LoadingCard />
+                        <LoadingCard delay={true} />
+                    </Carousel>
+                )}
+                {fetchStatus === "success" && (
+                    <Carousel>
+                        {cardSet?.cards.map((card, index) => {
+                            switch (card.cardType) {
```
srietkerk, line 202: This feels messy to me. I recently learned about type guards and I'm wondering if this situation would be a good use case for it. https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

teachertool/src/components/styling/HomeScreen.module.scss
```diff
@@ -104,6 +105,38 @@ div.page {
                 outline-color: var(--pxt-button-primary-foreground);
             }
         }
+        &.loadingGradient {
+            background: linear-gradient(45deg, var(--pxt-content-background), var(--pxt-content-foreground));
+            opacity: 0.2;
+            background-size: 400% 200%;
```
srietkerk, line 111: Why are both of theses greater than 100?

teachertool/src/components/HomeScreen.tsx
srietkerk: Makes sense. I'm just wondering if we should keep it consistent. It seems like in most other components, we doing it the `css["..."]` way, but if it's best to use dot notation when we can, then this is fine.

teachertool/src/components/HomeScreen.tsx
eanders-ms: Moving the typecast out of the render does make the code more readable!

teachertool/src/hooks/useJsonDocRequest.ts
```diff
@@ -0,0 +1,31 @@
...
+                    } else {
+                        setStatus("success");
+                        setJson(json as T);
+                    }
+                })
+                .catch(() => {
+                    setStatus("error");
+                });
+        }
+    }, [fetching]);
```
srietkerk, line 30: Before, when the state variable was `status`, there were no dependencies in the `useEffect`, why did this change?

eanders-ms: While reworking the function I noticed I had neglected to include this dependency. As it was, the value of `fetching` was captured and would never change within useEffect handler. Probably harmless, but the code is corrected now.

srietkerk (approved):
Looks great!

thsparks (approved):

PR #9887, opened 2024-02-23
Teacher Tool: Make Splitter Functional

thsparks:
This makes it so the splitter can actually be used to resize the windows. I've also made it so you can double-click the split to reset it to its original position.

I intentionally did not tie the split position into app state, so the control will be more reusable in the future, if we move it to react common.

It'd be nice to merge this and the existing VerticalResizeContainer, but they've approached the problem from two somewhat different angles that makes the refactor a bit messy. The VerticalResizeContainer is a single window that can have its height changed, but it doesn't care what's shrinking/expanding below it. Meanwhile, this is a container for two different panels that manages both sides. In the future, I think it'd make sense to move this SplitPanel into React Common and take anything that's using the VerticalResizeContainer and have it use the SplitPanel instead, but I wanted to save that for another change.

Upload Target: https://makecode.microbit.org/app/50d4b45ccb39db5b42fee1e5cacc60776082aa3c-32ad1744d7--eval

### DIFF
--- a/teachertool/src/components/MainPanel.tsx
+++ b/teachertool/src/components/MainPanel.tsx
@@ -16,8 +16,7 @@ export const MainPanel: React.FC<IProps> = () => {
 
     const lastSavedSplitPosition = getLastSplitPosition() ?? defaultSize;
 
-    // TODO still:
-    // Min and Max sizes.
+    // TODO : fix horizontal (offset within the view?)
 
     return (
         <div className={css["main-panel"]}>
@@ -28,6 +27,8 @@ export const MainPanel: React.FC<IProps> = () => {
                 primary={"left"}
                 left={<RubricWorkspace />}
                 right={<ProjectWorkspace />}
+                leftMinSize="5rem"
+                rightMinSize="5rem"
                 onResizeEnd={handleResizeEnd}
             />
         </div>
--- a/teachertool/src/components/MainPanel.tsx
+++ b/teachertool/src/components/MainPanel.tsx
@@ -8,13 +8,15 @@ import { getSplitPosition as getLastSplitPosition, setSplitPosition as setLastSp
 interface IProps {}
 
 export const MainPanel: React.FC<IProps> = () => {
-    function handleResize(size: number | string) {
+    function handleResizeEnd(size: number | string) {
         setLastSplitPosition(size.toString());
     }
 
     const lastSavedSplitPosition = getLastSplitPosition() ?? "50%";
 
-    // TODO still - fix arrow mouse display, figure out why it isn't working with the iframe loaded. Maybe test local storage more (does it redraw a bunch because of this changing after the change in SplitPane?)...
+    // TODO still:
+    // figure out why it isn't working with the iframe loaded.
+    // Double click to restore defaults.
 
     return (
         <div className={css["main-panel"]}>
@@ -24,7 +26,7 @@ export const MainPanel: React.FC<IProps> = () => {
                 primary={"left"}
                 left={<RubricWorkspace />}
                 right={<ProjectWorkspace />}
-                onResize={handleResize}
+                onResizeEnd={handleResizeEnd}
             />
         </div>
     );
--- a/teachertool/src/components/MainPanel.tsx
+++ b/teachertool/src/components/MainPanel.tsx
@@ -3,7 +3,7 @@ import css from "./styling/MainPanel.module.scss";
 import { SplitPane } from "./SplitPane";
 import { RubricWorkspace } from "./RubricWorkspace";
 import { ProjectWorkspace } from "./ProjectWorkspace";
-import { getSplitPosition as getLastSplitPosition, setSplitPosition as setLastSplitPosition } from "../services/storageService";
+import { getLastSplitPosition, setLastSplitPosition } from "../services/storageService";
 
 interface IProps {}
 
--- a/teachertool/src/components/MainPanel.tsx
+++ b/teachertool/src/components/MainPanel.tsx
@@ -15,7 +15,6 @@ export const MainPanel: React.FC<IProps> = () => {
     const lastSavedSplitPosition = getLastSplitPosition() ?? "50%";
 
     // TODO still:
-    // figure out why it isn't working with the iframe loaded.
     // Double click to restore defaults.
 
     return (
--- a/teachertool/src/components/MainPanel.tsx
+++ b/teachertool/src/components/MainPanel.tsx
@@ -11,7 +11,7 @@ export const MainPanel: React.FC<IProps> = () => {
         <div className={css["main-panel"]}>
             <SplitPane
                 split={"vertical"}
-                defaultSize={"80%"}
+                defaultSize={"50%"}
                 primary={"left"}
                 left={<RubricWorkspace />}
                 right={<ProjectWorkspace />}
--- a/teachertool/src/components/MainPanel.tsx
+++ b/teachertool/src/components/MainPanel.tsx
@@ -16,8 +16,6 @@ export const MainPanel: React.FC<IProps> = () => {
 
     const lastSavedSplitPosition = getLastSplitPosition() ?? defaultSize;
 
-    // TODO : fix horizontal (offset within the view?)
-
     return (
         <div className={css["main-panel"]}>
             <SplitPane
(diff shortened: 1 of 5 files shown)


teachertool/src/components/MainPanel.tsx
```diff
@@ -3,18 +3,31 @@ import css from "./styling/MainPanel.module.scss";
 import { SplitPane } from "./SplitPane";
 import { RubricWorkspace } from "./RubricWorkspace";
 import { ProjectWorkspace } from "./ProjectWorkspace";
+import { getLastSplitPosition, setLastSplitPosition } from "../services/storageService";
 
 interface IProps {}
 
 export const MainPanel: React.FC<IProps> = () => {
+    const defaultSize = "50%";
```
srietkerk, line 11: Do we want to keep this at 50%? I know in the designs we have the split with the rubric side of things wider than the iframe. Wondering if we want to have that wider design be the default.

teachertool/src/components/SplitPane.tsx
```diff
@@ -5,20 +5,115 @@ import { classList } from "react-common/components/util";
 interface IProps {
     className?: string;
     split: "horizontal" | "vertical";
-    defaultSize: number | string;
+    defaultSize: number | string; // The size to reset to when double clicking the splitter.
+    startingSize?: number | string; // The size to use initially when creating the splitter. Defaults to `defaultSize`.
```
srietkerk, line 9: I'm not sure I understand the benefit of having this extra parameter. I can understand that the parent of the `SplitPane` should be able to declare the default size of the panes, but I think starting size is something that can be determined inside `SplitPane`.

I think just having `defaultSize` is all we need, and then when first setting the size in the `useState`, you can just have the `getLastSplitPosition` call as the `lastSavedPosition` variable and the `useState` are essentially doing the same thing right now.

srietkerk: Oh, I see. You're trying to keep the state/service stuff outside of `SplitPane` to make it easier to move in the future. I wonder if saving the size in localstorage is something you just want to bring with this, though.

teachertool/src/components/SplitPane.tsx
```diff
@@ -5,20 +5,115 @@ import { classList } from "react-common/components/util";
...
+
+    function handleResize(clientX: number, clientY: number) {
+        const containerRect = containerRef.current?.getBoundingClientRect();
+        if (containerRect) {
+            const newSize =
+                split === "vertical"
+                    ? `${((clientX - containerRect.left) / containerRect.width) * 100}%`
+                    : `${((clientY - containerRect.top) / containerRect.height) * 100}%`;
+
+            setIsResizing(true); // Do this here rather than inside startResizing to prevent interference with double click detection.
```
srietkerk, line 57: It probably doesn't make that much of a difference, but is there a reason you don't do this at the start of the `if` statement?

teachertool/src/components/SplitPane.tsx
```diff
@@ -5,20 +5,115 @@ import { classList } from "react-common/components/util";
...
+    }
+
+    function setToDefaultSize() {
+        setSize(defaultSize);
+        onResizeEnd?.(defaultSize);
+    }
+
+    const leftStyle: React.CSSProperties = { flexBasis: size };
+    if (split === "vertical") {
+        leftStyle.minWidth = leftMinSize;
```
srietkerk, line 86: Do we need to make the `leftMinSize` and `rightMinSize` props required? If they aren't set, I think these calculations will fail sine we don't have default values for `leftMinSize` and `rightMinSize`.

teachertool/src/components/SplitPane.tsx
```diff
@@ -5,20 +5,115 @@ import { classList } from "react-common/components/util";
...
+        leftStyle.maxHeight = `calc(100% - ${rightMinSize})`;
+    }
+
     return (
-        <div className={classList(css[`split-pane-${split}`], className)}>
-            <div className={css[`left-${split}`]}>{left}</div>
-            <div className={css[`splitter-${split}`]}>
-                <div className={css[`splitter-${split}-inner`]} />
+        <div ref={containerRef} className={classList(css["split-pane"], css[`split-${split}`], className)}>
+            <div className={css[`left`]} style={leftStyle}>
```
srietkerk, line 98: I like keeping style outside of components as much as possible; I'm wondering if a pattern like this would work in this case. https://github.com/microsoft/pxt/blob/16f4fb291a47583c317410125fe40a30ef7e1ed4/webapp/src/toolbox.tsx#L758
Where we just define css variables with the values we need and then leverage these css variables in the scss modules. I did this in this PR a while back https://github.com/microsoft/pxt/pull/9657, but the pattern might not be the best fit for every scenario.

teachertool/src/components/MainPanel.tsx
thsparks: I don't feel strongly, but I think 50% is a reasonable starting point and we can observe if/how teachers tend to resize it when they use it. If there is consistent expansion in one direction, we can make that the default.

teachertool/src/components/SplitPane.tsx
thsparks: I don't think we should have this class interface with local storage. The caller would still need to pass in the local storage key to use, which would feel a little weird, but more importantly, if we re-use it somewhere else, that behavior may not be desirable. We may not want to save it at all, or we may want to save it somewhere else, like an account-level setting for example.

teachertool/src/components/SplitPane.tsx
thsparks: I just thought it looked nice to keep the set-state calls together. Do you think we should move it up?

teachertool/src/components/SplitPane.tsx
thsparks: Hmm, I don't think we should allow them to be fully unset, since it's possible to get into a state where you the splitter goes offscreen or something and then you can never get out of that state, even when you refresh the page.

We could have a default value, but when in doubt, I think it's simpler to just have the caller tell us what these values should be. The caller will always have more context than this component does, and it's not too burdensome to include.

teachertool/src/components/SplitPane.tsx
thsparks: Hmm, we could do it that way, but I'm not sure I fully understand the benefit? We're setting css inside the component either way. Do the variables provide any practical value, or is it just unnecessary complexity? If I understand correctly, I think it made more sense in the HC change because those variables are sometimes set through css and sometimes js. HC is kind of an override. That's not quite the case here, where it's basically always set through JS all the time. I could be missing/mis-understanding something, though.

teachertool/src/components/SplitPane.tsx
srietkerk: Yes, I agree, but right now these props are optional: https://github.com/microsoft/pxt/pull/9887/files#diff-5c44e5a8b91e45e13e21ff9f1a12da3634d9282699fb7a757190de64abcff78dR13-R14. I think they should just be required props to avoid possible problems.

teachertool/src/components/MainPanel.tsx
srietkerk: Makes sense. I just wanted to bring it up because I think on most monitors, having the splitter at 50% would show the large simulator instead of the small simulator in the corner and I wasn't sure if we were okay with that.

teachertool/src/components/SplitPane.tsx
srietkerk: My initial thought is yes, we should move it up because we want to make sure that `onResizeEnd` doesn't get called while the container is actively resizing, but I don't know if that case could happen or if it would be the worst thing if it did. It also might be the case that putting `setResizing` above the `newSize` definition would cause more problems with the `useEffect`. It's just one of those React things that I'm a bit fuzzy on.

teachertool/src/components/MainPanel.tsx
thsparks: Yeah, that's certainly the case on my screen, but let's see how it plays out.

teachertool/src/components/SplitPane.tsx
eanders-ms: It's common to use ResizeObserver in these scenarios. Did you consider using ResizeObserver here before going with this solution?

thsparks: I moved it up just in case.

As for the ResizeObserver, I'm not sure that'd help much in this case. I think it's actually preferable to react to the mouse up/down/move events, which feel closer to user's intent than trying to deduce from resize observations.

srietkerk (approved):
Awesome, thanks!!

PR #9889, opened 2024-02-26
Teacher tool: fixed function for getting block children, added one validator for blow away

srietkerk:
This PR is basically just a bug fix for nested validators. Using the built-in `Blockly.Block.getChildren()` was too limiting for our uses. I've split out the logic of getting the nested child blocks into a helper function that uses `getDescendants()` and `getSurroundParent()`. I also added a small validator plan to be used in the rubric that I made for the Blow Away micro:bit tutorial.

**Small note**
There's a comment above the use of `getDescendants()` that explains why the parameter that we're passing into the function is `false`. I just wanted to reiterate that here and post the documentation. By using `getDescendants(false)`, Blockly is filtering out disabled and collapsed blocks. The fact that we don't have to check for disabled blocks is nice, but I'm unsure if filtering out collapsed blocks will be a problem. 

Collapsed blocks are not something I've experimented with for the validation, and might be something that we want to discuss.


common-docs/teachertool/validator-plans-shared.json
```diff
@@ -150,6 +150,19 @@
                     }
                 }
             ]
+        },
+        {
+            ".desc": "a number exists // no catalog entry",
```
thsparks, line 155: nit: I don't see a need to specify "no catalog entry" here. I think that'll end up being true for a lot of validator plans.

pxteditor/code-validation/getNestedChildBlocks.ts
```diff
@@ -0,0 +1,10 @@
+/// <reference path="../../localtypings/validatorPlan.d.ts" />
+
+import * as Blockly from "blockly";
+
+export function getNestedChildBlocks(parentBlock: Blockly.Block): Blockly.Block[] {
```
thsparks, line 5: Can we add a comment above this function explaining specifically which types of blocks this returns (for example, I'm unclear if it will return only immediate children or if it will also contain children of children)

thsparks: From the code I'd guess it's only immediate children, but would be nice to have documentation confirming

pxteditor/code-validation/getNestedChildBlocks.ts
```diff
@@ -0,0 +1,10 @@
...
+    // want param to be false to filter out disabled blocks
+    const descendants = parentBlock.getDescendants(false);
```
thsparks, line 7: From the docs, I thought this flag was just an ordering parameter.

https://developers.google.com/blockly/reference/js/blockly.block_class.getdescendants_1_method

Even if it does have a side effect of removing disabled blocks, I don't think we should rely on undocumented behavior. I'd still recommend adding a check for disabled inside the filter condition.

common-docs/teachertool/validator-plans-shared.json
```diff
@@ -150,6 +150,19 @@
...
+            "name": "num_exists",
```
thsparks, line 156: nit: perhaps "math_number_exists" for a more descriptive name?

pxteditor/code-validation/validateBlockFieldValueExists.ts
```diff
@@ -15,8 +15,13 @@ export function validateBlockFieldValueExists({ usedBlocks, fieldType, fieldValu
         block.isEnabled() && block.type === specifiedBlock
     );
 
-    const successfulBlocks = enabledSpecifiedBlocks.filter((block) =>
-        block.getFieldValue(fieldType) === fieldValue
+    const successfulBlocks = enabledSpecifiedBlocks.filter((block) => {
+        if (fieldType === "VAR") {
```
thsparks, line 19: Is this reacting to a special case in blockly, or is it a special case we're creating ourselves? I'd love a comment here explaining what's up.

pxteditor/code-validation/getNestedChildBlocks.ts
thsparks: As a side note, idk if it's even possible for a child to be disabled when the parent is not disabled, so maybe a check on the parent block is sufficient? But that may require a bit more investigation.

srietkerk: You're right, it is. I have no idea why I thought it filtered disabled blocks. I thought I read it somewhere, but I can't find it lol; maybe I dreamt it. 😆

pxteditor/code-validation/validateBlockFieldValueExists.ts
srietkerk: I'll make a comment. It's reacting to a special case in blockly. Performing `block.getFieldValue(fieldType)` for a variable field block will return the `id` of that variable rather than it's name, so the comparison with `fieldValue` will never be truthy. For example, in the case that we're checking that the variable being used is named `row`, the comparison for the return would be `row === -2lX#Jei3u)pzbSEnrMB`.

thsparks: Is that just a signal that we should create a separate validator for variable names rather than using field value and type?

srietkerk: Yeah, I thought about that, too. I'm on the fence about it. Yes, the VAR field type uses a completely different function to reach the same thing as a different field value, _**but**_ the inputs to that validator will be the exact same. Or, I guess since we'd split it out, the variable validator wouldn't need the `fieldType` and `fieldValue` could be renamed to `varName` or something like that. 

I think it would be extremely reasonable to split it out, but being that we're still unsure of the value of validating variable names, I'm leaning towards leaving it for now. If it's something that teachers say they want, I think we can change it then.

pxteditor/code-validation/getNestedChildBlocks.ts
srietkerk: > As a side note, idk if it's even possible for a child to be disabled when the parent is not disabled, so maybe a check on the parent block is sufficient? But that may require a bit more investigation.

I agree; my first impression would be that it wouldn't be possible to have disabled child blocks if the parent is not, but I'm also not sure. 

On another note, I don't think we need to filter out disabled blocks here. All of the validators filter out disabled blocks (except comment validators), so we'd be doing extra work. We can take out the filtering in the different validators and put it in `runValidatorPlan`, but if we do that, the tutorial validation might get different results from the teacher tool validation.

srietkerk: And, I'm going to change this from false to true. I think ordered children will generally be more helpful if we want to take advantage of the ordered capability.

thsparks: As a helper function, I don't think we can make assumptions about what the caller will/won't have done regarding disabled blocks, however you could add a parameter filterDisabled and validators could pass false if they think it's unnecessary. That said, I _think_ the check is pretty trivial so I wouldn't expect any significant impact by running it.

pxteditor/code-validation/getNestedChildBlocks.ts
srietkerk: Yeah, immediate children. I'll still add a comment, but I won't be able to have a visual in the comment, and I think a visual will help show what's happening. 

For example, if you have this if statement:
image
`getNestedChildBlocks` will return the condition, the two variable assignments, and  the music block.

pxteditor/code-validation/getNestedChildBlocks.ts
```diff
@@ -0,0 +1,13 @@
...
+
+import * as Blockly from "blockly";
+
+// returns the directly nested children of a block
+// for blocks with a mouth, it returns the blocks inside the mouth and its inputs, if any
+// for something like pick random, it would return the two number blocks
+export function getNestedChildBlocks(parentBlock: Blockly.Block): Blockly.Block[] {
+    const descendants = parentBlock.getDescendants(true);
+    const enabledDescendants = descendants.filter((block) => block.isEnabled());
+    const nestedChildren = enabledDescendants.filter((block) => block.getSurroundParent() === parentBlock);
```
thsparks, line 11: What you have is fine, but I think you could avoid the duplicate list iteration with:

`const nestedChildren = enabledDescendants.filter((block) => block.isEnabled() && block.getSurroundParent() === parentBlock);`

PR #9890, opened 2024-02-27
Restore the reindeer connection previewer

riknoll:
The latest version of Blockly added the ConnectionPreviewer extension point. This adds one to render our reindeer connections and increases the snap radius to match what we had in pxt-blockly


thsparks (approved):
🦌🦌🦌

PR #9891, opened 2024-02-28
Update Import Flow to Match Designs

thsparks:
This updates our import flow (and confirmation dialog) to match our v1 designs.

I tried reusing https://github.com/microsoft/pxt/blob/master/webapp/src/draganddrop.ts, by moving it into browserutils and adding extra events for when drag enters and exits/leaves, but it didn't quite work out. The enter and exit events caused a lot of flickering when hovering over child elements, which I believe would require a workaround somewhat like the one I've described in the DragAndDropFileSurface in this PR. I wasn't comfortable adding that workaround at such a global scale (and it isn't necessary when you don't care about enter/exit), so I decided not to reuse.

Upload target: <del>https://makecode.microbit.org/app/5fc5d7fc1489c936a8bdc40422b6a5b3a5926855-6b68fdda25--eval</del> https://makecode.microbit.org/app/3743a0ed15ed73ae921ee8bf75533603348b2665-b7ed2b5339--eval

### Screenshots
Import:
image

File hover:
image

Confirm:
image

Invalid file:
image

Small screen:
image


teachertool/src/components/DragAndDropFileSurface.tsx
```diff
@@ -0,0 +1,103 @@
...
+    To solve this, we add a transparent div (droppable-surface) over everything and use that for most drag-related event handling.
+    However, we don't want the transparent droppable-surface to intercept pointer events when there is no drag occurring, so
+    we still use the drag-and-drop-file-surface to detect dragEnter events and only intercept pointer events on the droppable-surface
+    after that has happened.
+    */
+    return (
+        <div className={css["drag-and-drop-file-surface"]} onDragEnter={handleDragEnter}>
+            <div className={css["instruction-container"]}>
+                <i className={classList("fas fa-file-upload", css["upload-icon"])}></i>
+                <div className="no-select">{fileIsOverSurface ? Strings.ReleaseToUpload : Strings.DragAndDrop}</div>
```
srietkerk, line 66: Do we need to wrap `no-select` in `css["no-select"]`?

srietkerk: Same for the "hidden" classname for the input below.

teachertool/src/components/DragAndDropFileSurface.tsx
```diff
@@ -0,0 +1,103 @@
...
+import css from "./styling/DragAndDropFileSurface.module.scss";
+import { Button } from "react-common/components/controls/Button";
+
+export interface DragAndDropFileSurfaceProps {
+    onFileDroppedAsync: (file: File) => void;
+    errorMessage?: string;
+}
+export const DragAndDropFileSurface: React.FC<DragAndDropFileSurfaceProps> = ({ onFileDroppedAsync, errorMessage }) => {
+    const [fileIsOverSurface, setFileIsOverSurface] = useState(false);
+    const [errorKey, setErrorKey] = useState(0); // Used to reset animations on the error label.
```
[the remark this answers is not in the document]
thsparks: By changing the key of the component, we force react to unmount the component and mount a new one, which resets the animation state. If we don't do this, then the animation only runs on the first error and no subsequent ones, since from the component's perspective, the animation has already run.

teachertool/src/components/DragAndDropFileSurface.tsx
```diff
@@ -0,0 +1,103 @@
...
+        event.preventDefault();
+        setFileIsOverSurface(false);
+    }
+
+    function handleDrop(event: React.DragEvent<HTMLDivElement>) {
+        event.preventDefault();
+
+        setFileIsOverSurface(false);
+
+        const file = event.dataTransfer.files[0];
```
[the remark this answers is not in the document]
thsparks: Yeah, I'm not sure what uploading multiple files would mean in terms of import, so my inclination would be to disallow it and provide an error message, but we can discuss as a team and follow up in a separate change.

teachertool/src/components/DragAndDropFileSurface.tsx
thsparks: Nope! These classes come from react-common which means the names haven't been modified by the scss stuff, so no lookup is necessary.

teachertool/src/components/ConfirmationModal.tsx
```diff
@@ -0,0 +1,58 @@
+import { useContext, useEffect, useState } from "react";
+import { AppStateContext } from "../state/appStateContext";
+import { Modal } from "react-common/components/controls/Modal";
+import { hideModal } from "../transforms/hideModal";
+import { classList } from "react-common/components/util";
+import { CautionLevel } from "../types";
+import css from "./styling/ConfirmationModal.module.scss";
+
+export type ConfirmationModalProps = {
```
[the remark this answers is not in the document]
srietkerk: Yeah, I think it should be moved to stay consistent.

teachertool/src/components/DragAndDropFileSurface.tsx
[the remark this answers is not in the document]
srietkerk: Ohh, gotcha. Makes sense, so errorKey could technically be any type, as long as we can make a quick change to it to cause an unmount and mount, and that's why using numbers is convenient?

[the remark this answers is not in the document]
thsparks: As long as it's a valid key, then yes (though I think that's really only strings and numbers).

eanders-ms:
I tried dropping tsconfig.json and got this error in the console:
image

teachertool/src/types/index.ts
```diff
@@ -44,3 +46,11 @@ export type CarouselCardSet = {
 };
 
 export type RequestStatus = "init" | "loading" | "error" | "success";
+
+export type ConfirmationModalProps = {
```
eanders-ms, line 50: The inclusion of `Props` in the name tells me this configures a React component. Props is uncommon for non-React configuration. I suggest using something else here. Suggestions:
`ConfirmationModalOptions`
`ConfirmationModalParams`

teachertool/src/components/DragAndDropFileSurface.tsx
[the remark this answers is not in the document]
eanders-ms: Just importing the first file is fine. I'd prefer not to treat this as an error or warning. The fewer alarm states we create, the better. If we can make it "just work", that's good. The user will learn how it behaves through observation.

teachertool/src/components/styling/ConfirmationModal.module.scss
```diff
@@ -0,0 +1,5 @@
+.confirmation-modal {
+    button.caution {
+        background-color: rgb(255, 71, 71);
```
eanders-ms, line 3: Can this color be derived from the toast error color? I think it will be the same color for all themes, so not a big issue to appear as a literal here. Though high-contrast might be a special case.

teachertool/src/components/styling/DragAndDropFileSurface.module.scss
```diff
@@ -0,0 +1,88 @@
...
+        right: 0;
+        bottom: 0;
+
+        pointer-events: none;
+        &.dragging {
+            pointer-events: all;
+        }
+    }
+
+    @media screen and (max-height: 720px) {
```
eanders-ms, line 74: We should be using breakpoints from react-common, not creating our own.

eanders-ms: Probably better to just remove this for now. We have an issue on the board tracking the need to implement responsive layout app-wide. Something we can address post v1.

teachertool/src/transforms/loadRubricAsync.ts
```diff
@@ -23,10 +21,5 @@ export async function loadRubricAsync(rubricUrl: string) {
         return;
     }
 
-    if (!(await confirmAsync(Strings.ConfirmReplaceRubricMsg))) {
-        return;
-    }
-
-    setRubric(json);
-    setActiveTab("rubric");
+    replaceActiveRubricAsync(json);
```
eanders-ms, line 24:
```suggestion
    await replaceActiveRubricAsync(json);
```

teachertool/src/components/styling/ConfirmationModal.module.scss
thsparks: I tried that at first, but the button looked disabled:
image


That said, the toast's accent color might work. I'll switch to that:
image

thsparks:
> I tried dropping tsconfig.json and got this error in the console: image

Good catch. Added a check in `verifyRubricIntegrity` to ensure criteria is not undefined.

teachertool/src/components/DragAndDropFileSurface.tsx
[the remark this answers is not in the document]
srietkerk: I think this might just be something we want to cover/ask in user testing. My thoughts are that because we have the modal where we tell the user that importing will override their current rubric, especially with the red coloring, we are putting the user in an alarm state, so I think we should give them the most piece of mind that we possibly can. Personally, when I saw the modal, I didn't even read the rubric title in the top bar because most of my focus was on the button and the details outlining what importing is going to do, so, when I dragged multiple rubrics in, I got nervous not knowing which rubric was going to override my current rubric. In my mind, a "just work" scenario if I drag multiple files in would be that all of my rubrics are now on the site and accessible. I know this isn't what we're going for right now, but I still think it's a valid scenario to consider. For now, I think we can take the first file, but I think we should make an issue to supporting uploading multiple rubrics at once.

teachertool/src/state/actions.ts
```diff
@@ -55,6 +55,11 @@ type SetRubric = ActionBase & {
     rubric: Rubric;
 };
 
+type SetConfirmationProps = ActionBase & {
```
eanders-ms, line 58: Can we change this name (and various related identifiers) to match?
```suggestion
type SetConfirmationOptions = ActionBase & {
```

teachertool/src/components/DragAndDropFileSurface.tsx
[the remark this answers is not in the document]
eanders-ms: On that note, I don't think we should use a red/alarm color for the Continue button. The primary button color should be sufficient.

eanders-ms (approved):
Looks great!

srietkerk (approved):
Nice!

PR #9894, opened 2024-02-29
Fix Readonly Mode

thsparks:
Readonly mode wasn't being honored in the blockly workspace because we were resetting the flag based on whether or not we were in debug mode. With this change, we will also account for the original readonly mode in that check.

### DIFF
--- a/webapp/src/blocks.tsx
+++ b/webapp/src/blocks.tsx
@@ -1889,7 +1889,7 @@ export class Editor extends toolboxeditor.ToolboxEditor {
             }
         });
 
-        this.editor.options.readOnly = debugging;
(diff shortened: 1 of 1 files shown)


eanders-ms (approved):
Looks good to me. Is this a recent regression?

thsparks:
> Looks good to me. Is this a recent regression?

Yeah, looks like it was introduced with the [blockly upgrade](https://github.com/microsoft/pxt/commit/16f4fb291a47583c317410125fe40a30ef7e1ed4#diff-3c724788360b53dfd2a1ea4b5e044d12ea802ad504a9fa5567da60d7bde7f9c1R1886)

riknoll (approved):

PR #9895, opened 2024-02-29
tt: enable vertical scrolling in modal dialog

eanders-ms:
We should consider porting this to react-common.

### DIFF
--- a/teachertool/src/style.overrides/Modal.scss
+++ b/teachertool/src/style.overrides/Modal.scss
@@ -1,0 +1,13 @@
+
+.common-modal {
+    max-height: 90vh;
+    display: flex;
+    flex-direction: column;
+}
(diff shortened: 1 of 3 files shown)


srietkerk (approved):
Nice!

thsparks (approved):

teachertool/src/style.overrides/Modal.scss
```diff
@@ -0,0 +1,13 @@
+
+.common-modal {
+    max-height: 90vh;
+    display: flex;
```
thsparks, line 4: Is it not already a flex? I'm surprised.

eanders-ms: I was shocked!

PR #9897, opened 2024-03-01
Teacher tool results page

srietkerk:
It's finally here! The teacher tool results page!

Features added in this PR:
- Styling that better reflect the designs for the results page
- Ability to change the evaluated result outcome. I.E, if a criteria evaluated to "needs work" but the teacher thought it was fine, they could change it to "looks good"
- Ability to add notes to a criteria result
- A header for the page that includes a QR Code for student's project
- If a project is loaded, the results page will have the criteria listed, but will have the eval result set to "not started"

TODO:
- Hook up the page to the print button so teachers can actually use the whole pane as a document to be handed out

image

Upload target: ~https://makecode.microbit.org/app/7f8be11839a9cb6975327e946a3c16b72055d1fa-78e9ccfd2a--eval~
https://makecode.microbit.org/app/b05f767340a077fc626fce120d9a539b4f21cfcb-5d8f328e64--eval


teachertool/src/state/actions.ts
```diff
@@ -84,6 +84,11 @@ type SetAutorun = ActionBase & {
     autorun: boolean;
 };
 
+type SetEvalResultsBatch = ActionBase & {
```
eanders-ms, line 87: `SET_EVAL_RESULTS` vs `SET_EVAL_RESULT` is a very subtle difference. Recommend calling this one `SET_EVAL_RESULTS_BATCH`.

teachertool/src/style.overrides/Button.scss
```diff
@@ -39,4 +39,16 @@
...
 .common-button.disabled {
     opacity: 0.8;
+}
+
+.rounded > .common-button.common-dropdown-button {
+    border-radius: 1rem;
+}
+
+.fail > .common-button.common-dropdown-button {
+    background-color: #FFD15C;
```
eanders-ms, line 49: No color literals please. Use vars. They will need to be themed for high contrast.

teachertool/src/transforms/initNewProjectResults.ts
```diff
@@ -0,0 +1,12 @@
+import { stateAndDispatch } from "../state";
+import { setEvalResultsPending } from "./setEvalResultsPending";
+import * as Actions from "../state/actions";
+
+
+export function initNewProjectResults() {
+    const { state: teacherTool, dispatch } = stateAndDispatch();
+    setEvalResultsPending({ clearAllEntries: true });
+    for (const result of Object.keys(teacherTool.evalResults)) {
+        delete teacherTool.evalResults[result].notes
```
eanders-ms, line 10: **Important**: Never modify the state object directly. Changes must be made in the reducer only. The UI will not update in response to this change. It may be ignored entirely.

eanders-ms: This is a ship blocker.

eanders-ms: Can you make an action for this instead?

theme/themepacks.less
```diff
@@ -14,6 +14,7 @@
     --pxt-page-background: white;
     --pxt-page-foreground: black;
     --pxt-page-foreground-light: #767676;
+    --pxt-page-foreground-lighter: #888888;
```
eanders-ms, line 17: Ugh. I need to revisit our theme vars soon! Great to capture this requirement though, and thank you for adding it as a var.

teachertool/src/components/DebouncedTextarea.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { Textarea, TextareaProps } from "react-common/components/controls/Textarea";
+
+export interface DebouncedTextareaProps extends TextareaProps {
+    intervalMs?: number; // Default 500 ms
+}
+
+// This functions like the React Common Input, but debounces onChange calls,
+// so if onChange is called multiple times in quick succession, it will only
+// be executed once after a pause of the specified `interval` in milliseconds.
+export const DebouncedTextarea: React.FC<DebouncedTextareaProps> = ({ intervalMs = 500, ...props }) => {
```
eanders-ms, line 11: I believe textarea has a prop to enable a resize handler on the corner of the control. If so, it would be great to enable that here:
image

teachertool/src/components/EvalResultDisplay.tsx
```diff
@@ -1,11 +1,39 @@
...
+    function getProjectLink(): string {
+        const inputText = teacherTool.projectMetadata?.inputText!;
+        const hasMakeCode = inputText?.indexOf("makecode") !== -1;
+        return hasMakeCode ? inputText : `https://makecode.com/${inputText}`;
+    }
+    // TODO: change the headers to be the correct thing for html (not h3, h4)
+    return (
+        <div className={css["header"]}>
+            <div className={css["rubric-name"]}>
+                <h2>{lf("{0}",teacherTool.rubric.name)}</h2>
```
eanders-ms, line 22: This is not a localizable string.
```suggestion
                <h2>{teacherTool.rubric.name}</h2>
```

teachertool/src/components/EvalResultDisplay.tsx
```diff
@@ -1,11 +1,39 @@
...
+            </div>
+            <div className={css["project-details"]}>
+                <div className={css["project-text"]}>
+                    <h3>{lf("{0}", teacherTool?.projectMetadata?.name)}</h3>
```
eanders-ms, line 26: Also non-localizable.
```suggestion
                    <h3>{teacherTool?.projectMetadata?.name}</h3>
```

eanders-ms (commented):
Looking really good! I had some feedback that can be addressed post-testing. But I did find one ship blocker: an instance where state is being modified outside the reducer.

teachertool/src/transforms/initNewProjectResults.ts
srietkerk: Can I iterate through the notes like in the for loop in the reducer?

eanders-ms: Iterating is ok, but you must not modify the current `state` object. You can build a new set of `evalResults` and assign it to the copy being generated in the reducer.

teachertool/src/transforms/setEvalResultOutcome.ts
```diff
@@ -0,0 +1,13 @@
+import { stateAndDispatch } from "../state";
+import * as Actions from "../state/actions";
+import { CriteriaEvaluationResult } from "../types/criteria";
+
+export function setEvalResultOutcome(criteriaId: string, result: CriteriaEvaluationResult) {
```
thsparks, line 5: I think we should rename the `CriteriaEvaluationResult` enum to something like `EvaluationStatus`. I keep getting confused because I think `CriteriaEvaluationResult` is the type for the status + notes object, but then it's actually just the status. And in some places (like `newCriteriaEvalResult` below), we kind of are using the term `CriteriaEvalResult` to refer to the whole object. So just to minimize confusion, I think it'd help to rename the enum.

teachertool/src/transforms/setEvalResultsPending.ts
```diff
@@ -0,0 +1,22 @@
+import { stateAndDispatch } from "../state";
+import { CriteriaEvaluationResult, CriteriaResult } from "../types/criteria";
+import { Rubric } from "../types/rubric";
+import * as Actions from "../state/actions";
+
+
+export function setEvalResultsPending({ clearAllEntries, rubric }: {
+    clearAllEntries?: boolean,
```
thsparks, line 8: perhaps `overwriteExistingEntries`?

teachertool/src/transforms/setEvalResultsPending.ts
```diff
@@ -0,0 +1,22 @@
...
+    const { state: teachertool, dispatch } = stateAndDispatch();
+    let allEvalResults: pxt.Map<CriteriaResult> = {};
+    const usedRubric = rubric || teachertool.rubric;
+    for (const criteria of usedRubric.criteria) {
+        const instanceId = criteria.instanceId;
+        if (!teachertool.evalResults[instanceId] || clearAllEntries) {
+            allEvalResults[instanceId] = { result: CriteriaEvaluationResult.Pending };
+        }
+    }
+    allEvalResults = { ...teachertool.evalResults, ...allEvalResults };
```
thsparks, line 20: If `clearAllEntries` is set, should we refrain from including the `...teachertool.evalResults` in this? Wondering about the scenario where a criteria exists in the old eval results but not in the new rubric, I think this would preserve it.

theme/themepacks.less
thsparks: For the criteria table, we use --pxt-content-accent. Would that color work for this use case as well?

teachertool/src/state/reducer.ts
```diff
@@ -40,6 +40,22 @@ export default function reducer(state: AppState, action: Action): AppState {
...
+        case "SET_EVAL_RESULTS_BATCH": {
+            return {
+                ...state,
+                evalResults: action.criteriaResults,
+            };
+        }
+        case "CLEAR_ALL_EVAL_RESULT_NOTES": {
+            const evalResults = { ...state.evalResults };
+            for (const result of Object.keys(evalResults)) {
+                delete evalResults[result].notes
```
thsparks, line 52: Do we really want to delete here, or would setting it to undefined be sufficient?

thsparks (approved):
Have a few comments, but nothing strictly blocking. Could be done in a follow-up PR if you're trying to check this in asap.

teachertool/src/components/DebouncedTextarea.tsx
srietkerk: Here are the props for textarea https://github.com/microsoft/pxt/blob/697fe51d4c563df3f78508f75fb8eb2a8c99934d/react-common/components/controls/Textarea.tsx#L4-L20 I have `resize` set to vertical, but it doesn't really seem to do anything.. Maybe it's because I've set a `max-height`?

srietkerk: Yeah, I was thinking that, too @thsparks, and I thought the resize field would do that, but it didn't grow as more text was input. Maybe it has something to do with `cols`?

teachertool/src/state/reducer.ts
eanders-ms: Another alternative... would setting `notes` to an empty string be preferred?

teachertool/src/components/DebouncedTextarea.tsx
thsparks: Hmm, did you explore doing it through css rather than props? There may be a way through that.

srietkerk: No, I hadn't... I'll look into it.

teachertool/src/components/AddCriteriaButton.tsx
```diff
@@ -17,7 +17,7 @@ export const AddCriteriaButton: React.FC<IProps> = ({}) => {
     );
     return (
         <Button
-            className={classList("inline", "add-criteria-button")}
+            className={classList("inline", "add-button")}
```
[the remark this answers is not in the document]
srietkerk: Yeah, I think if there will be more use diverse cases for the button, we should move it out, but I don't think it's super needed yet. 

But I can rename it if that will clarify things.

theme/themepacks.less
```diff
@@ -45,6 +46,10 @@
     --pxt-toast-accent-info: #3182CE;
     --pxt-toast-accent-warning: #DD6B20;
     --pxt-toast-accent-error: #E53E3E;
+
+    // Eval dropdown
```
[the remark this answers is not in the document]
srietkerk: I don't think the colors from the toasts work well for these dropdowns. 
image

When you say add variations, do you mean adding new variables but just specifying them as a toast color?

theme/themepacks.less
srietkerk: Yeah, I actually like that!! I'm just wondering if the header's bottom border should be darker, though. I played with it and I'm not sure between these:

**With header bottom border darker**
Screenshot 2024-03-01 150057

**With all bottom borders being the same**
Screenshot 2024-03-01 150235

eanders-ms: 2nd one (same borders) seems cleaner to me.

theme/themepacks.less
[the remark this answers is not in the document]
eanders-ms: These colors came from microbit, but were too strong for the toast, so we softened them by lowered their intensity. You can find the originals in the Figma, on the Components page: 
image

I believe these are what I used in the design of the results page. I wonder if these could be integrated into the theme as:

`--pxt-error` (red)
`--pxt-error-soft`  (red softened)
`--pxt-warning`  (yellow)
`--pxt-warning-soft`  (yellow softened)
`--pxt-success`  (green)
`--pxt-success-soft`  (green softened)
`--pxt-info`  (blue)
`--pxt-info-soft`  (blue softened)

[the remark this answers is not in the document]
eanders-ms: More consistent naming:

`--pxt-error` (red)
`--pxt-error-accent` (red softened)

teachertool/src/components/DebouncedTextarea.tsx
srietkerk: It looks like this is going to be tricky-ish. Definitely more work that shouldn't be added to this PR, I'll make an item for it on our board. 

@eanders-ms, in regards to the resize question, it looks like we can change set it, but it looks really funky and is not functioning in the way that I would expect, so I'm going to leave it out. I agree with Thomas, though, that I think the better approach here would just have this auto-resize. I think it would be cool for users to be able to change the size of the textarea for a bit more customization, but again, I think it can be something I look into at a later time.

theme/themepacks.less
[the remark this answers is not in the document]
srietkerk: I'm unsure of the accent/soft use for us since dropdown doesn't use colors like that and we already have those "softened" colors in the toast coloring, but I can add these colors in as

`--pxt-warning`
`--pxt-error`
`--pxt-success`
`--pxt-info`

if that's preferred.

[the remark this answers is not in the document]
eanders-ms: Sorry, I should have mentioned... these would replace the toast color vars. Toast would used the the `-accent` vars.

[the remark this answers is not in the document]
srietkerk: The toasts also have their own accents, though... Would we just keep those the same?
image

teachertool/src/transforms/setEvalResultsPending.ts
srietkerk: Yeah, you're totally right.

teachertool/src/components/AddCriteriaButton.tsx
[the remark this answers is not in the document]
thsparks: Renaming is good for now

eanders-ms:
Fine to leave as a rough edge for now, but I don't think "evaluating..." should be an option in the dropdown: 
image

teachertool/src/components/DebouncedTextarea.tsx
eanders-ms: Auto resize would be a fine solution. Let's track it as followup work.

theme/themepacks.less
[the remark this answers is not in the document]
eanders-ms: We should remove the toast vars entirely, rolling them into the warning/error/success/info categories. There will be background and accent variations of each, other variations if necessary. Whatever is needed to support toast and the result dropdown should be defined in these categories. It looks like accents will be necessary, so include those:

```
--pxt-warning-background
--pxt-warning-accent
```

If you'd like assistance working up the complete set of variable changes, please ping me on teams and we'll go through them.

eanders-ms (approved):
Looks great! Thanks for tidying up those color vars.

PR #9899, opened 2024-03-01
re-cache json docs on request

eanders-ms:
I was getting old versions of validator-plans and catalog.json. Turns out they're getting cached by the browser. Specifying `no-cache` on the fetch request so that the files, if cached, will still work offline, but the cache will be refreshed after each new request.


eanders-ms:
https://developer.mozilla.org/en-US/docs/Web/API/Request/cache

srietkerk (approved):

eanders-ms:
Note to future self: We will need to add this flag to the fetch of persistent share metadata.

thsparks (approved):

PR #9900, opened 2024-03-01
Fixes for image editor in micro:bit

riknoll:
This is an addendum to my earlier pr https://github.com/microsoft/pxt/pull/9878 which neglected to fix the palette when it was in the pxt.json of a dependency rather than the top package.

Also adds the ability to configure the image and animation editors to compile/parse different tagged template literals since micro:bit can't use `img`. To use it, simply add it to the field options on the function defining the field editor like so:

```
//% bmp.fieldOptions.taggedTemplate="bmp"
```

it might already be there since this is the same field option we use for configuring the decompiler.

### DIFF
--- a/pxtblocks/fields/field_animation.ts
+++ b/pxtblocks/fields/field_animation.ts
@@ -13,6 +13,8 @@ export interface FieldAnimationOptions {
 
     filter?: string;
     lightMode: boolean;
+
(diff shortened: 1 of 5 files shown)


tballmsft (approved):

tballmsft:
Thanks, Richard!

PR #9901, opened 2024-03-04
Teacher tool: icon in sharelink input now clickable

srietkerk:
https://makecode.microbit.org/app/8064d90149396080f7266d4f8935b133be8e7173-7d530bca36--eval

Just a note that the styling is a bit off in the upload target, even though I built the css, the upload target is using `-0.75rem` for the margin instead of `-1rem` to get the button hugging the right side more like in our design right now. 

This is what it looks like with `margin-right: -1rem` 
image

### DIFF
--- a/teachertool/src/components/ShareLinkInput.tsx
+++ b/teachertool/src/components/ShareLinkInput.tsx
@@ -42,6 +42,7 @@ export const ShareLinkInput: React.FC<IProps> = () => {
                 icon={icon}
                 onChange={onTextChange}
                 onEnterKey={onEnterKey}
+                onIconClick={onEnterKey}
                 preserveValueOnBlur={true}
                 autoComplete={false}
             ></Input>
--- a/teachertool/src/components/ShareLinkInput.tsx
+++ b/teachertool/src/components/ShareLinkInput.tsx
@@ -40,6 +40,7 @@ export const ShareLinkInput: React.FC<IProps> = () => {
                 placeholder={lf("Enter Project Link or Share ID")}
                 ariaLabel={lf("Project Link or Share ID")}
                 icon={icon}
+                iconTitle={lf("Submit project link or share ID")}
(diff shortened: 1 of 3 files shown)


teachertool/src/components/ShareLinkInput.tsx
```diff
@@ -42,6 +42,7 @@ export const ShareLinkInput: React.FC<IProps> = () => {
                 icon={icon}
                 onChange={onTextChange}
                 onEnterKey={onEnterKey}
+                onIconClick={onEnterKey}
```
thsparks, line 45: Screen reader doesn't seem to be announcing what the button does (it just says "button"). Does setting iconTitle fix that?

teachertool/src/style.overrides/Button.scss
```diff
@@ -39,4 +39,9 @@
 
 .common-button.disabled {
     opacity: 0.8;
+}
+
+.common-input-group>.common-button {
```
eanders-ms, line 44: I don't think we want this styling applied to all instances of .common-input-group>.common-button. This is a special case, so it should go in ShareLinkInput.module.scss, as `button[class~="common-button"]`.

teachertool/src/components/ShareLinkInput.tsx
srietkerk: I can add it, but I don't know how to test screen readers.

thsparks: Simplest way is to open narrator, which should be installed by default on windows, then tab over the element.

eanders-ms: You could pass iconTitle for the Button's ariaLabel prop.

eanders-ms (approved):
😎

thsparks (approved):

PR #9902, opened 2024-03-04
Fix Comparison in Splitter Default Size

thsparks:
When the user hasn't set a size directly, we were seeing the panel resize depending on the tab. This was because the startingSize was an empty string and we were doing a nullish check instead of a falsey one for the initial splitter position, so it was using that empty string for the initial flexBasis instead of the default 50%.


srietkerk (approved):
Nice!

PR #9904, opened 2024-03-04
Fix the keyboard navigation plugin support

riknoll:
Fixing our keyboard navigation. Three small bugs:

1. The dummy workspace was never initialized in the multi-flyout and caused an exception during init
2. We were immediately closing the flyout when the cursor moved because of the workspace event being fired
3. The dropdown portion of the text dropdown field was being focused, which broke typing into the field.


eanders-ms (approved):
👍

PR #9905, opened 2024-03-05
Update Blocks Exist Json Schema for JsonPath Friendliness

thsparks:
The structure of the `blockCounts` input for `blocksExist` validator plans was not JsonPath friendly, as the customizable `blockId` was treated as a property and it was also a part of the path for count. This change adjusts the schema so both `blockId` and `count` can be referenced more easily by [JsonPath](https://github.com/json-path/JsonPath), which we will need for parameterized criteria.

No user-facing functionality changes.


eanders-ms (approved):
👍

PR #9906, opened 2024-03-06
Teacher Tool: Run Prettier

thsparks:
Noticed we had quite a large list of prettier issues. Wanted to clean those up. No functional changes, just ran "npm run prettier" from the teachertool folder and confirmed it built/ran fine.


eanders-ms (approved):
😎

PR #9907, opened 2024-03-06
Refactor all of our various cra build steps

riknoll:
I will soon be adding another CRA for the tutorial tool, so I figured it was about time to refactor all of this build code to make it easier to add new apps and generally unify things.

@thsparks since you did this most recently, did I miss anything here?


gulpfile.js
```diff
@@ -863,11 +787,6 @@ exports.watch = initWatch;
 exports.watchCli = initWatchCli;
 exports.testlanguageservice = testlanguageservice;
 exports.onlinelearning = onlinelearning;
-exports.skillmap = skillmap;
-exports.authcode = authcode;
-exports.multiplayer = multiplayer;
-exports.kiosk = kiosk;
-exports.teacherTool = teacherTool;
 exports.tt = teacherTool;
```
thsparks, line 790: FWIW there's also `exports.tt` here (just a shortcut to teachertool), but I do actually use that on occasion when I only want to build teacher tool. Is there a reason we're removing these?

gulpfile.js
```diff
@@ -568,109 +576,25 @@ const testSkillmap = gulp.series(buildSkillmapTests, copySkillmapTests, runSkill
...
-    .pipe(gulp.dest(`${teacherToolOut}/js`));
-
-const copyTeacherToolHtml = () => rimraf("webapp/public/teachertool.html")
-    .then(() => gulp.src(`${teacherToolRoot}/build/index.html`)
-                    .pipe(replace(/="\/static\//g, `="/blb/teachertool/`))
-                    .pipe(concat("teachertool.html"))
-                    .pipe(gulp.dest("webapp/public")));
-
-const teacherTool = gulp.series(cleanTeacherTool, buildTeacherTool, gulp.series(copyTeacherToolCss, copyTeacherToolJs, copyTeacherToolHtml));
+const teacherTool = createWebappTasks("teachertool");
```
thsparks, line 597: Down below here, we have the "webapp build wrapper" section which calls into the update*Strings functions. That could maybe be made slightly simpler with a unified updateWebappStrings function so we don't have to create a new one for each app, but not a huge deal.

package.json
```diff
@@ -34,16 +34,8 @@
...
-    "built/web/authcode/css/*.css",
-    "built/web/authcode/js/*.js",
-    "built/web/multiplayer/css/*.css",
-    "built/web/multiplayer/js/*.js",
-    "built/web/kiosk/css/*.css",
-    "built/web/kiosk/js/*.js",
-    "built/web/teachertool/css/*.css",
-    "built/web/teachertool/js/*.js",
+    "built/web/*/css/*.css",
+    "built/web/*/js/*.js",
```
thsparks, line 38: It'd be nice if we could update the npm-prepare.js script (gets called as a part of prepare at the bottom of this file) to also reference subwebapp.ts but I'm not sure if that's possible...

thsparks (approved):
Huge improvement! I think only the `webappNames` issue in `server.ts` is likely to cause actual problems. Other comments are just potential opportunities for further simplification, though they may not be worth the trouble. This is a big help either way!

thsparks:
Maybe not very helpful at this point, but FWIW, I saved a branch which just had the bare-bones updates I needed to make a new CRA. Figured I'd mention it, in case it's a useful reference: https://github.com/microsoft/pxt/compare/master...thsparks/new_react_app#diff-d7ffbddb3a36042f0033638e9e21f556c8873232b040c3f6579488d74adec3daR6

gulpfile.js
riknoll: they're still there! just being set inside the function defining the tasks

thsparks: Ahh, I see now. Makes sense!

cli/server.ts
```diff
@@ -1184,19 +1184,11 @@ export function serveAsync(options: ServeOptions) {
...
-
-        if (pathname == "/--authcode") {
-            sendFile(path.join(publicDir, 'authcode.html'));
-            return
-        }
-
-        if (pathname == "/--multiplayer") {
-            sendFile(path.join(publicDir, 'multiplayer.html'));
-            return
+        for (const subapp of SUB_WEBAPPS) {
```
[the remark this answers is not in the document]
riknoll: ah, good catch. lemme see what i can do here...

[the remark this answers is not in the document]
riknoll: fixed!

gulpfile.js
riknoll: done!

package.json
riknoll: done! I moved this config into a json file that gets used in npm-prepare.js, in the cli, and in the gulpfile

thsparks: Nice 😎

PR #9908, opened 2024-03-06
Teacher tool: fixed successful blocks for blocks exist new blocks exist schema

srietkerk:
The way successful blocks are populated for the blocks exist validator is a bit different from the others, and when we updated the blocks exist schema, the successful blocks list that we get from running the blocksExist validation broke. You'd have only run into this bug with blocksExist validation with children.

### DIFF
--- a/pxteditor/code-validation/runValidatorPlan.ts
+++ b/pxteditor/code-validation/runValidatorPlan.ts
@@ -72,8 +72,7 @@ function runBlocksExistValidation(usedBlocks: Blockly.Block[], inputs: pxt.block
     const blockResults = validateBlocksExist({ usedBlocks, requiredBlockCounts });
     let successfulBlocks: Blockly.Block[] = [];
     if (blockResults.passed) {
-        const blockIdsFromValidator = Object.keys(inputs.blockCounts)
-        const blockId = blockIdsFromValidator?.[0];
+        const blockId = inputs.blockCounts[0].blockId;
         successfulBlocks = blockResults.successfulBlocks.length ? blockResults.successfulBlocks[0][blockId] : [];
     }
     return [successfulBlocks, blockResults.passed];


pxteditor/code-validation/runValidatorPlan.ts
```diff
@@ -72,8 +72,7 @@ function runBlocksExistValidation(usedBlocks: Blockly.Block[], inputs: pxt.block
     const blockResults = validateBlocksExist({ usedBlocks, requiredBlockCounts });
     let successfulBlocks: Blockly.Block[] = [];
     if (blockResults.passed) {
-        const blockIdsFromValidator = Object.keys(inputs.blockCounts)
-        const blockId = blockIdsFromValidator?.[0];
+        const blockId = inputs.blockCounts[0].blockId;
```
thsparks, line 75: blockCounts can have more than one entry, so I don't think we can assume index 0 here. Rather, we need to iterate over these ids. I also don't think we can assume index 0 in the line below.

srietkerk: Yeah, this is something I thought of when I was first making this. I just kept it at the zeroth index because I hadn't seen a validator that we were building for the teacher tool that used multiple blocks inside the blocks exist. I thought that it would either go in a separate check or we would just use the block set validator instead. I can change this, though.

Your comment about the successful blocks makes me think, though. I'm not really sure why I made successful blocks a list.. Maybe just to keep with the pattern of the other things that we return from blocksExist since they all return lists. I'm thinking that it might be better to make successful blocks an Object instead, that way we don't _need_ to do any indexing, we can just do a look-up of `blockResults.successfulBlocks[blockId]` and this will return the list of successful blocks of that block id.

thsparks (approved):
Thanks!

PR #9910, opened 2024-03-06
[Stable8.5] Cherry pick codeql fix

riknoll:
This isn't actually blocking anything (the action that handles pxt bumps is still running fine) but this PR fixes the codeql action on stable8.5 because it was annoying me.


srietkerk (approved):
Nice!

PR #9911, opened 2024-03-07
Initial Parameterized Input Support

thsparks:
This adds support for string and number parameters in criteria. I've tried to make the UI passable, but it's not perfect (particularly, I think the text boxes should expand to fit contents, it lacks signal when there's an issue, like an empty parameter, and the results page should probably put boxes around the parameter values). I'd like to follow up with a separate change to polish all that off, since this one is a bit big already.

One notable adjustment, I made the parameter `path` field into a list so that one parameter can affect multiple values in a validator plan (useful if the thing you're validating requires multiple checks, like custom functions are created and called).

###Screenshots
Add Criteria Modal
image

In the Criteria List:
image

Filled in:
image

Results Taking Into Account Custom Values:
image

Upload target: https://makecode.microbit.org/app/010e736468338d36b5870860ad598f1de57e8e6f-f52a28b536--eval?tc=1

TODO (in future PRs):
1. Finalize UI
2. Block selection


eanders-ms:
Seeing an error when I try to run the "comment used 1 times" criteria:
image

thsparks:
> Seeing an error when I try to run the "comment used 1 times" criteria: image

Ah, that's because it's using the live docs file so it doesn't have this change:
image

teachertool/src/transforms/runEvaluateAsync.ts
```diff
@@ -29,7 +33,42 @@ function generateValidatorPlan(criteriaInstance: CriteriaInstance): pxt.blocks.V
...
+                });
+                showToast(
+                    makeToast(
+                        "error",
+                        // prettier-ignore
+                        lf("Unable to evaluate criteria: missing '{0}' in '{1}'", param.name, getReadableCriteriaTemplate(catalogCriteria))
+                    )
+                );
+            }
+            return undefined;
```
srietkerk, line 65: What would happen if we don't return here or above when we get an error? Would it show all of the problems with that one criteria if there are multiple? Would it be too overwhelming for the user?

Just curious about these things, not saying that we need to change anything here.

teachertool/src/utils/index.ts
```diff
@@ -44,3 +45,24 @@ export function getProjectLink(inputText: string): string {
...
+    // Split by the regex, which will give us an array where every other element is a parameter.
+    const paramRegex = /\$\{([\w\s]+)\}/g;
+    const parts = template.split(paramRegex);
+
+    const segments: CriteriaTemplateSegment[] = [];
+    for (let i = 0; i < parts.length; i++) {
+        const part = parts[i];
+
+        if (part) {
+            segments.push({ type: i % 2 === 0 ? "plain-text" : "param", content: part });
```
eanders-ms, line 59: This loop is making several assumptions I'm not sure are safe to make:
* The `parts` array begins with a `plain-text` element.
* Two consecutive `param` elements will be separated by a `plain-text` element.

Are you certain these will hold true for all possible input?

thsparks: I did wonder about these when I was working on it, but the split with a regex separator inserts empty strings even between consecutive matches, [and at the start if the first element is a separator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split#:~:text=If%20separator%20appears%20at%20the%20beginning%20(or%20end)%20of%20the%20string%2C%20it%20still%20has%20the%20effect%20of%20splitting%2C%20resulting%20in%20an%20empty%20(i.e.%20zero%20length)%20string%20appearing%20at%20the%20first%20(or%20last)%20position%20of%20the%20returned%20array.), so I am pretty confident this is a safe assumption.

For good measure, I tested out with "${one} ${two}${three}" and here's the result:
image
image
image

teachertool/src/types/criteria.ts
```diff
@@ -17,10 +17,14 @@ export interface CriteriaInstance {
 }
 
 // Represents a parameter definition in a catalog criteria.
+export type CriteriaParameterType = "string" | "number" | "block";
+export type CriteriaParameterPicker = "longString" | "blockSelector";
```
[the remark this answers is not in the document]
thsparks: Hmm maybe we could combine them. I was thinking type was more like the type of input we're expecting and both string (short) and longString are just expecting strings, but I guess we could combine them and that info is just baked in. I'll do that.

teachertool/src/types/index.ts
```diff
@@ -50,9 +50,15 @@ export type RequestStatus = "init" | "loading" | "error" | "success";
...
 };
+
 export type ConfirmationModalOptions = {
     title: string;
     message: string;
     onCancel: () => void;
     onContinue: () => void;
 };
+
+export type CriteriaTemplateSegment = {
```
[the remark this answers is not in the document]
eanders-ms: For better or worse, I've been advocating for centralizing shared types here.

teachertool/src/utils/index.ts
eanders-ms: I am surprised it works that well, but ok!

eanders-ms: You may want to trim the parsed param names. Consider the input `"${ param1 }   ${ param2 }". I think you'll end up with param names in in the `segments` array that include the padded whitespace.

common-docs/teachertool/test/catalog-shared.json
```diff
@@ -11,30 +11,74 @@
...
-                    "path": "checks[0].count"
+                    "paths": ["checks[0].count"]
                 }
             ]
         },
         {
             "id": "59AAC5BA-B0B3-4389-AA90-1E767EFA8563",
             "use": "block_used_n_times",
-            "template": "${block_id} used ${count} times",
+            "template": "${Block} used ${count} times",
```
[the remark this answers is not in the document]
thsparks: I've tried to make these case insensitive, though I just noticed an issue if the param name defined below has capitalized letters so I'll look into that. But the intent, at least, is for the case to not matter. We could change that if we wanted to, though.

teachertool/src/utils/index.ts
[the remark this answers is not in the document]
thsparks: I believe it is, because `string.split()` will insert an empty string in the returned list if the string starts with the separator. (Documented [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split#:~:text=If%20separator%20appears%20at%20the%20beginning%20(or%20end)%20of%20the%20string%2C%20it%20still%20has%20the%20effect%20of%20splitting%2C%20resulting%20in%20an%20empty%20(i.e.%20zero%20length)%20string%20appearing%20at%20the%20first%20(or%20last)%20position%20of%20the%20returned%20array.)). Included a little more info in this comment: https://github.com/microsoft/pxt/pull/9911#discussion_r1517038306

teachertool/src/components/CriteriaInstanceDisplay.tsx
```diff
@@ -0,0 +1,109 @@
...
+                    />
+                );
+            case "block": // TODO
+            default:
+                return null;
+        }
+    }
+
+    function getPlainTextSegmentDisplay(text: string): JSX.Element | null {
+        return text ? <div className={css["text-segment"]}>{text}</div> : null;
```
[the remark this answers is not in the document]
thsparks: I think it's useful so we don't create an empty div when the text is an empty string, and in this case, it felt slightly tidier to do that check here rather than before calling this function.

teachertool/src/types/index.ts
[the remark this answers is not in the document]
thsparks: Yeah, my internal logic here was to use `criteria.ts` for types that "compose" a criteria object, so `CatalogCriteria` which uses `CriteriaParameter` for a field inside it, which in turn uses `CriteriaParameterType` for a field inside it, and so on... The template segment exists outside of that, so I didn't put it in the file.

Admittedly, I was lazy and put the `CriteriaInstance` types inside `criteria.ts` as well, so it's all a bit muddled. I honestly don't feel that strongly, but the `index.ts` feels like a better "default" to me than `criteria.ts`.

common-docs/teachertool/test/catalog-shared.json
```diff
@@ -11,30 +11,74 @@
                     "name": "count",
                     "type": "number",
                     "default": 1,
-                    "path": "checks[0].count"
+                    "paths": ["checks[0].count"]
```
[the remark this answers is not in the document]
thsparks: Another scenario might be "A variable named ${name} is set and used". We need one check to validate the "Set" exists and one check to validate the "used" exists, but ${name} has to be inserted into both checks, so that's one parameter being put into multiple paths.

There could probably be work-arounds for this by defining more validator plans and adding the complexity there instead, but I think it's cleaner and much easier to use if we just allow the parameter to be inserted into multiple places in the json.

teachertool/src/transforms/runEvaluateAsync.ts
thsparks: As implemented currently, it would actually end up running the check, but it'd use the values we have stored in the validatorPlan pre-replacement (often empty strings or 0), so it would look like the check ran but what it really checked would be something different.

We could expand this to have a more detailed validation where we parse the whole thing in advance and make a toast with all errors combined, but I'm not sure that's worth the trouble. Ideally any errors would be immediately apparent from the rubric display regardless, though I'm planning to do that in a follow-up change.

teachertool/src/utils/index.ts
thsparks: > I am surprised it works that well, but ok!

Happy to try out other test cases if you can think of any. I think I've verified all the ones that came to my mind, at least (begin with param, end with param, params next to each other, no params, one parameter and nothing else, and of course, params sprinkled throughout with plain text)

common-docs/teachertool/test/catalog-shared.json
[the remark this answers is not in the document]
srietkerk: Yeah, makes sense. Thanks for the example!

teachertool/src/types/index.ts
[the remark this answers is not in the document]
srietkerk: Yeah, that's fine, I don't feel strongly about this. I feel like if that's the pattern we want to follow, then maybe the `EvaluationStatus` enum and `CriteriaResult` interface should be moved out then, too. I put them there to start because I thought anything revolving around Criteria should be in `criteria.ts`, but it sounds like we don't want that to be the case. It isn't something to do for this PR, but something we can do in the future maybe.

srietkerk (approved):
LGTM! Exciting!! 🥳

teachertool/src/utils/index.ts
eanders-ms: Two that I can think of:
* Nested params: `${this is ${nested}} ${this_is_a_valid_param}`.
* Dot notation: `${user.name} ${user.email} ${this_is_a_valid_param}`.

Fine if nested/dotted params don't parse out as parameters, but would be good to know the parser can recover and successfully pull out a valid param appearing later.

eanders-ms (approved):
Great start on this feature!

teachertool/src/utils/index.ts
thsparks: Hmm interesting. They definitely don't parse out as parameters, but it also doesn't crash and it is able to get the valid parameter afterwards.

Testing `${this is ${nested}} ${count}` and `some text ${user.name} ${user.email} ${count} some more text`, we get:
image
image

Unless there are concerns from your end, I'm inclined to leave it be for now and worry about supporting those scenarios if/when we need them.

eanders-ms: Ya that's fine. Looks like it was able to find the last param. There are libraries that do this kind of thing really well, though probably overkill for us at this time. The [classic "handlbars" package](https://handlebarsjs.com/), comes to mind.

PR #9912, opened 2024-03-11
Add pxtservices package and iframe driver

riknoll:
This PR adds a new top-level package called pxtservices to be shared between our many webapps. Currently it just contains a single file that defines a nifty class for driving iframes that embed the makecode editor. Features include:

* All currently defined requests now have type checked methods (no more calling postMessage directly)
* An optional host object parameter that lets you implement all of the methods required by the iframe workspace (no need to figure out how that protocol works)
* All editor events/messages can be accessed via `addEventListener()` (also type checked)

As part of this, I also fixed three editor events that were being incorrectly fired with the "pxteditor" type instead of "pxthost". The "pxteditor" versions still fire so as not to break existing implementations.

I also updated the makecodeEditorService to use this.

### DIFF
--- a/pxtservices/iframeDriver.ts
+++ b/pxtservices/iframeDriver.ts
@@ -1,0 +1,553 @@
+/// <reference path="../localtypings/pxteditor.d.ts" />
+
+type MessageHandler = (response: any) => void;
+
+const MessageReceivedEvent = "message";
+const MessageSentEvent = "sent";
+
+interface PendingMessage {
+    original: pxt.editor.EditorMessageRequest;
+    resolve: MessageHandler;
+    reject: MessageHandler;
+}
+
+export interface IframeWorkspaceStatus {
+    projects: pxt.workspace.Project[];
+    editor?: pxt.editor.EditorSyncState;
+    controllerId?: string;
+}
+
+export interface IFrameWorkspaceHost {
+    saveProject(project: pxt.workspace.Project): Promise<void>;
+    getWorkspaceProjects(): Promise<IframeWorkspaceStatus>;
+    resetWorkspace(): Promise<void>;
+    onWorkspaceLoaded?(): Promise<void>;
+}
+
+/**
+ * Manages communication with the editor iframe.
+ *
+ * TODO: Currently only supports a single iframe as all incoming messages
+ * are posted on the window. Would be nice in the future to refactor the
+ * iframe messages to use a MessageChannel.
+ */
+export class IframeDriver {
+    protected readyForMessages = false;
+    protected messageQueue: pxt.editor.EditorMessageRequest[] = [];
+    protected nextId = 0;
+    protected pendingMessages: { [index: string]: PendingMessage } = {};
+    protected editorEventListeners: { [index: string]: MessageHandler[] } = {};
+
+    constructor(public iframe: HTMLIFrameElement, public host?: IFrameWorkspaceHost) {
+        window.addEventListener("message", this.onMessageReceived);
+    }
+
+    dispose() {
+        window.removeEventListener("message", this.onMessageReceived);
+    }
+
+    async switchEditorLanguage(lang: "typescript" | "blocks" | "python") {
+        let action;
+        switch (lang) {
+            case "blocks":
+                action = "switchblocks";
+                break;
+            case "typescript":
+                action = "switchjavascript";
+                break;
+            case "python":
+                action = "switchpython";
+                break;
+        }
+
+        await this.sendRequest(
+            {
+                type: "pxteditor",
+                action
+            } as pxt.editor.EditorMessageRequest
+        );
+    }
+
+    async setLanguageRestriction(restriction: pxt.editor.LanguageRestriction) {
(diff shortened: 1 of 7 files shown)


teachertool/src/services/makecodeEditorService.ts
```diff
@@ -3,92 +3,37 @@
...
 
 //  an example of events that we want to/can send to the editor
 export async function setHighContrastAsync(on: boolean) {
-    const result = await sendMessageAsync({
-        type: "pxteditor",
-        action: "sethighcontrast",
-        on: on,
-    });
-    console.log(result);
+    await driver!.setHighContrast(on)
```
eanders-ms, line 36: Is driver guaranteed to be set here? I can imagine a scenario where we send the teacher tool's high contrast setting to the editor during app load, but the iframe isn't loaded yet.

jwunderl (approved):

pxtservices/iframeDriver.ts
```diff
@@ -0,0 +1,553 @@
...
+    addEventListener(event: "simevent", handler: (ev: pxt.editor.EditorSimulatorEvent) => void): void;
+    addEventListener(event: "tutorialevent", handler: (ev: pxt.editor.EditorMessageTutorialEventRequest) => void): void;
+    addEventListener(event: "editorcontentloaded", handler: (ev: pxt.editor.EditorContentLoadedRequest) => void): void;
+    addEventListener(event: "workspacesave", handler: (ev: pxt.editor.EditorWorkspaceSaveRequest) => void): void;
+    addEventListener(event: "workspaceevent", handler: (ev: pxt.editor.EditorWorkspaceEvent) => void): void;
+    addEventListener(event: "workspacereset", handler: (ev: pxt.editor.EditorWorkspaceSyncRequest) => void): void;
+    addEventListener(event: "workspacesync", handler: (ev: pxt.editor.EditorWorkspaceSyncRequest) => void): void;
+    addEventListener(event: "workspaceloaded", handler: (ev: pxt.editor.EditorWorkspaceSyncRequest) => void): void;
+    addEventListener(event: "workspacediagnostics", handler: (ev: pxt.editor.EditorWorkspaceDiagnostics) => void): void;
+    addEventListener(event: "editorcontentloaded", handler: (ev: pxt.editor.EditorContentLoadedRequest) => void): void;
```
eanders-ms, line 412: duplicate declaration of "editorcontentloaded" handler.

pxtservices/iframeDriver.ts
```diff
@@ -0,0 +1,553 @@
...
+                } as pxt.editor.EditorMessageResponse);
+            }
+        }
+    }
+
+    protected sendRequest(message: any) {
+        return new Promise((resolve, reject) => {
+            message.response = true;
+            message.id = this.nextId++ + "";
+            this.pendingMessages[message.id] = {
```
eanders-ms, line 531: Noting a potential leak: All sent messages are wrapped in a promise and added to `pendingMessages`, which resolve once a reply is received. Is a reply received for every message sent?

pxtservices/iframeDriver.ts
```diff
@@ -0,0 +1,553 @@
...
+    sendMessage(message: pxt.editor.EditorMessageRequest): Promise<pxt.editor.EditorMessageResponse> {
+        return this.sendRequest(message) as Promise<pxt.editor.EditorMessageResponse>;
+    }
+
+    protected onMessageReceived = (event: MessageEvent) => {
+        const data = event.data as pxt.editor.EditorMessageRequest;
+        if (!data || !/^pxt(host|editor|pkgext|sim)$/.test(data.type)) return;
+
+        if (data.type === "pxteditor") {
+            if (data.id && this.pendingMessages[data.id]) {
```
eanders-ms, line 441: Noting a potential leak: The promise associated with sent messages that never receive a their expected reply could hang forever. This driver may want to run a pending message checker in the background (on a timeout), to clear pending messages that haven't received a reply in a certain amount of time.

thsparks (approved):
Agree with Eric's comments, but otherwise LGTM!

teachertool/src/services/makecodeEditorService.ts
riknoll: good point. fixed!

pxtservices/iframeDriver.ts
riknoll: only requests are wrapped in a promise. all requests with `message.response` set to true should indeed receive a response from the editor. i guess we could time out requests, but i think it's a minor scenario and i worry it might affect some long running requests (like starting a tutorial on a low end chromebook)

i guess i could implement some sort of heartbeat mechanism where the iframe reports back what messages are still active, but that might be overkill... thoughts?

eanders-ms: Ok, that seems fine then. If we expect every message sent this way to receive a response, then promises won't leak except when something else is already broken. Adding a heartbeat seems like overkill :)

eanders-ms (approved):

PR #9915, opened 2024-03-12
Teacher Tool: Improve Customizable Criteria UI

thsparks:
This change puts parameters into a clear error-state when they're empty, updates the results view to box in parameters per the design, and contains a few other minor tweaks throughout.

The boxes around parameter values in the results view seemed very similar to the boxes in the catalog view, so I refactored that into a shareable component. The main difference between the two is whether or not you show the parameter name vs the parameter value, which is indicated by whether or not you pass a criteria instance or just a catalog criteria.

image

image

image

Upload target: <del>https://makecode.microbit.org/app/5656ba85c36837d663cb6609d998ee5eb395949f-f1d049b96c--eval?tc=1</del> https://makecode.microbit.org/app/6d8a16dc031a6cba000e6c00781331f5a6da05cb-d358cf977b--eval?tc=1

### DIFF
--- a/teachertool/src/components/CriteriaInstanceDisplay.tsx
+++ b/teachertool/src/components/CriteriaInstanceDisplay.tsx
@@ -43,13 +43,12 @@ const InlineInputSegment: React.FC<InlineInputSegmentProps> = ({
                 isEmpty ? css["error"] : undefined
             )}
             icon={isEmpty ? "fas fa-exclamation-triangle" : undefined}
-            iconTitle={isEmpty ? lf("Missing value") : undefined}
             initialValue={initialValue}
             onDebouncedChange={onDebouncedChange}
             onChange={onChange}
             preserveValueOnBlur={true}
             placeholder={numeric ? "0" : param.name}
-            title={param.name}
+            title={isEmpty ? lf("{0}: value required", param.name) : param.name}
             autoComplete={false}
             type = {numeric ? "number" : "text"}
         />
--- a/teachertool/src/components/CriteriaInstanceDisplay.tsx
+++ b/teachertool/src/components/CriteriaInstanceDisplay.tsx
@@ -34,24 +34,27 @@ const InlineInputSegment: React.FC<InlineInputSegmentProps> = ({
         setIsEmpty(!newValue);
     }
 
+    const tooltip = isEmpty ? lf("{0}: value required", param.name) : param.name;
     return (
-        <DebouncedInput
-            className={classList(
-                css["inline-input"],
-                numeric ? css["number-input"] : css["string-input"],
-                shouldExpand ? css["long"] : undefined,
-                isEmpty ? css["error"] : undefined
-            )}
-            icon={isEmpty ? "fas fa-exclamation-triangle" : undefined}
-            initialValue={initialValue}
-            onDebouncedChange={onDebouncedChange}
-            onChange={onChange}
-            preserveValueOnBlur={true}
-            placeholder={numeric ? "0" : param.name}
-            title={isEmpty ? lf("{0}: value required", param.name) : param.name}
-            autoComplete={false}
-            type = {numeric ? "number" : "text"}
-        />
+        <div title={tooltip} className={css["inline-input-wrapper"]}>
+            <DebouncedInput
+                className={classList(
+                    css["inline-input"],
+                    numeric ? css["number-input"] : css["string-input"],
+                    shouldExpand ? css["long"] : undefined,
+                    isEmpty ? css["error"] : undefined
+                )}
+                icon={isEmpty ? "fas fa-exclamation-triangle" : undefined}
+                initialValue={initialValue}
+                onDebouncedChange={onDebouncedChange}
+                onChange={onChange}
+                preserveValueOnBlur={true}
+                placeholder={numeric ? "0" : param.name}
+                title={tooltip}
+                autoComplete={false}
+                type={numeric ? "number" : "text"}
+            />
+        </div>
     );
 };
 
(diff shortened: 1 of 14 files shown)


thsparks:
> I'm getting this error on the comment validator right now.. image I cleared indexeddb and local storage just in case as well.
> 
> This might be something with local serves, though.. I think I remember having problems with seeing new criteria in the catalog when I made new ones in the shared area.

Hmm, that's odd. I updated it in https://github.com/microsoft/pxt/commit/a3c462caabaf3de9b3129c86cbca903163d672ca so it should no longer be using that nonexistent plan (https://github.com/microsoft/pxt/blob/master/common-docs/teachertool/test/catalog-shared.json#L5). I'm not sure why it's not updated, but I don't think it's related to this change.

teachertool/src/components/CriteriaInstanceDisplay.tsx
```diff
@@ -22,22 +23,32 @@ const InlineInputSegment: React.FC<InlineInputSegmentProps> = ({
     shouldExpand,
     numeric,
 }) => {
-    function onChange(newValue: string) {
+    const [isEmpty, setIsEmpty] = useState(!initialValue);
+
+    function onDebouncedChange(newValue: string) {
```
srietkerk, line 28: I understand the want to have the debounce to limit the amount of state changes, but I'm not sure about introducing and ability for something to trigger on a debounce and something else fires immediately. It feels like an anti-pattern to have logic on a debounce component that doesn't get "debounced". Maybe just setting the `intervalMs` lower than 500ms could be a solution?

teachertool/src/components/CriteriaInstanceDisplay.tsx
```diff
@@ -22,22 +23,32 @@ const InlineInputSegment: React.FC<InlineInputSegmentProps> = ({
...
+                isEmpty ? css["error"] : undefined
             )}
+            icon={isEmpty ? "fas fa-exclamation-triangle" : undefined}
             initialValue={initialValue}
+            onDebouncedChange={onDebouncedChange}
             onChange={onChange}
             preserveValueOnBlur={true}
             placeholder={numeric ? "0" : param.name}
-            title={param.name}
+            title={isEmpty ? lf("{0}: value required", param.name) : param.name}
```
srietkerk, line 51: The hit box for the input is a bit weird when there's an error.. If you hover over the placeholder, you get this tooltip, but if you hover over the icon, you get the tooltip associated with the criteria instance.

teachertool/src/components/CriteriaInstanceDisplay.tsx
thsparks: I think even 500ms is going to be noticeable. I'm not sure I agree with the idea that it's an anti-pattern. The debounce is purely internal, not really relevant to the user, but the visual feedback of an error is something they see and interact with. After all, we don't debounce the characters appearing in the input field, and I think this error display is closer to that scenario than it is to us saving the input internally (which is what gets debounced).

teachertool/src/components/CriteriaInstanceDisplay.tsx
thsparks: Fixed. At first, I tried just setting iconTitle but that didn't work as expected when it's not a button, and I was a little worried about messing up existing scenarios if I changed that, so I just wrapped it in a div with the desired tooltip.

With fix: https://makecode.microbit.org/app/6d8a16dc031a6cba000e6c00781331f5a6da05cb-d358cf977b--eval?tc=1

teachertool/src/components/CriteriaInstanceDisplay.tsx
srietkerk: I'm meaning anti-pattern as a code smell. Having the props `onChange` and `onDebouncedChange` are a bit hard to understand. If I'm importing a debounced input, I'm expecting all of the things that I input to be debounced, so if I use `onChange` and it doesn't get debounced, I would consider it a bug. 

I agree that the error is different. In this scenario, we're just checking whether or not the input is empty whenever the input is interacted with. "onChange" is a bit misleading because it could be that no text was changed, that the box was just clicked on.

I propose getting rid of `onDebouncedChange`, keeping `onChange` as it was, and adding a prop like `checkIsEmpty` or something and that logic can be added to that if statement that you added under the `setTimeout` in DebouncedInput.

thsparks: In this scenario it's checking isEmpty, but there could be scenarios where it's more involved (like putting a letter in a number field, exceeding some length, or even reaching some minimum length to enter a "valid" state instead of an error one). I don't think we want to make the component that situationally specific.

If this is a sticking point, I would sooner just get rid of the debouncing altogether and have it react in real time.

thsparks: I think checkInputIsValid is probably still too specific and would imply that the DebouncedInput component itself is going to have some styling / reaction to the invalid state.

Would it be preferable if we had `onChange` which is debounced and `onChangeImmediate` which is not? That feels less intuitive to me personally, but perhaps works as a compromise?

thsparks: On second thought, I think I'll just remove the debouncing for now.

thsparks: On second thought, I think I will just remove the debounce.

srietkerk (approved):
LGTM!

eanders-ms (approved):
Looks great. The error background color is a bit dark. Something we can tweak wholistically later.

PR #9916, opened 2024-03-13
Teacher tool: print results page

srietkerk:
When you click the print button that shows up when you're on the results page, it will now actually print the results view. There are a couple of style changes when printing:
- Add notes button goes away
- Notes area's textbox is no longer rounded
- The criteria entry is a bit taller if there are notes inserted
- The dropdown does not have color and the icon is no longer showing

I also wanted to note that I changed the label for the notes area to be "Feedback" I thought this was a bit more indicative of what the notes actually meant when students receive this report back, and I think it's a bit more descriptive for teachers as well. 

Upload target: ~https://makecode.microbit.org/app/4681c0a9e9db964679b2d6f2ed8ee9173643af5b-a6dc22f71f--eval~
https://makecode.microbit.org/app/c1e324f715b0650c746f40817a3ef2cbf6fe72eb-6117cfded2--eval
Note about the upload target... I keep getting errors about being "unable to run criteria". I'm not sure the problem here, but will look into this after this change. I was getting problems with this before this change as well.


thsparks:
I think it'd be great if we could remove some of the empty space when there's no feedback:
image

thsparks:
To my eye, the pill-shaped outline still looks very UI-ish to me and gives off an "I printed out a web page" feeling. I wonder if maybe removing the outline and making it bold and underlined might look better? Something like:

image

teachertool/src/components/CriteriaResultEntry.tsx
```diff
@@ -95,7 +95,7 @@ export const CriteriaResultEntry: React.FC<CriteriaResultEntryProps> = ({ criter
                             criteriaId={criteriaId}
                         />
                     </div>
-                    <div className={css["result-notes"]}>
+                    <div className={classList(css["result-notes"], css[`${teacherTool.evalResults[criteriaId]?.notes ? '': 'no-print'}`])}>
```
thsparks, line 98:
```suggestion
                    <div className={classList(css["result-notes"], teacherTool.evalResults[criteriaId]?.notes ? undefined : css['no-print'])}>
```

teachertool/src/components/EvalResultDisplay.tsx
```diff
@@ -27,13 +27,13 @@ const ResultsHeader: React.FC = () => {
     );
 };
 
-export const EvalResultDisplay: React.FC<{}> = () => {
+export const EvalResultDisplay: React.FC<{printRef: React.RefObject<HTMLDivElement>}> = ({ printRef }) => {
```
thsparks, line 30: nit: For our other components, we've been defining an interface for the props (i.e. `EvalResultDisplayProps`). For consistency, it'd be nice to do that here too. Functionally, I'm not sure if it makes any difference.

srietkerk:
> I think it'd be great if we could remove some of the empty space when there's no feedback: image

I can play with it, especially with the idea of just having a label rather than the pill for the evaluation result. I left it larger because it feels a bit cramped without it. I found myself getting overwhelmed and unsure of where to look when it was tighter.

teachertool/src/components/EvalResultDisplay.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { useContext } from "react";
+import { useContext, useRef } from "react";
```
thsparks, line 2: Is this useRef import still needed?

teachertool/src/components/PrintButton.tsx
```diff
@@ -2,16 +2,21 @@ import * as React from "react";
...
+    onHandlePrint?: () => void;
 }
 
 export const PrintButton: React.FC<PrintButtonProps> = ({ printRef, onHandlePrint }) => {
+    const { state: teacherTool } = stateAndDispatch();
     const handlePrint = useReactToPrint({
         content: () => printRef.current,
-        onAfterPrint: onHandlePrint,
+        onPrintError: () => showToast(makeToast("error", lf("Unable to print evaluation results"), 2000)),
+        documentTitle: `MakeCode Project Evaluation Results - ${teacherTool.projectMetadata?.name}`,
```
thsparks, line 19: I'm guessing we'll want to sanitize the `teacherTool.projectMetadata.name` here in case it contains invalid file name characters. We already have a `sanitizeFileName` helper function for this in `pxt.Util`. We use it when exporting rubrics too: https://github.com/microsoft/pxt/blob/thsparks/teachertool/better_ui_for_customizable_criteria/teachertool/src/services/fileSystemService.ts#L8

teachertool/src/components/RubricWorkspace.tsx
```diff
@@ -53,7 +54,11 @@ const WorkspaceTabButtons: React.FC = () => {
     );
 };
 
-const WorkspaceTabPanels: React.FC = () => {
+interface WithRefProps {
```
thsparks, line 57: nit: perhaps `PrintRefProps` would be a clearer name?

teachertool/src/components/PrintButton.tsx
srietkerk: Yeah, I can add this. 

Just want some quick brainstorming for this. Since `sanitizeFileName` needs a string, I need  to make sure that the project metadata exists before passing in `teacherTool.projectMetadata.name`, but it should be impossible for a user to print without project metadata. That said, if they do get to a case where they can see the print button without a project loaded, I figure we should error and say something like "cannot print results if there is no project loaded". If I do error, I figure I should move the "sanitize" logic out of the component, but I'm unsure of the best place to put it. Maybe as a transform or in utils?

teachertool/src/components/RubricWorkspace.tsx
```diff
@@ -101,7 +106,7 @@ function getActionMenuItems(tab: TabName): MenuItem[] {
     return items;
 }
 
-const WorkspaceToolbarButtons: React.FC = () => {
+const WorkspaceToolbarButtons: React.FC<PrintRefProps> = ({ printRef }) => {
```
eanders-ms, line 109: Lets keep with this naming pattern for consistency.

```suggestion
const WorkspaceToolbarButtons: React.FC<WorkspaceToolbarButtonsProps> = ({ printRef }) => {
```

teachertool/src/components/RubricWorkspace.tsx
```diff
@@ -53,7 +54,11 @@ const WorkspaceTabButtons: React.FC = () => {
     );
 };
 
-const WorkspaceTabPanels: React.FC = () => {
+interface PrintRefProps {
```
eanders-ms, line 57: Lets stick with bespoke Props interfaces, for consistency.
```suggestion
interface WorkspaceTabPanelsProps {
```

teachertool/src/components/EvalResultDisplay.tsx
```diff
@@ -27,13 +27,17 @@ const ResultsHeader: React.FC = () => {
     );
 };
 
-export const EvalResultDisplay: React.FC<{}> = () => {
+interface EvalResultDisplayProps {
+    printRef: React.RefObject<HTMLDivElement>;
```
eanders-ms, line 31: The name `printRef` seems overly specific here. The ref is just a generic ref, could be used for anything.

teachertool/src/components/RubricWorkspace.tsx
srietkerk: So, I didn't want to make different prop interfaces for these because they have the same prop. I don't know the benefit of giving both the toolbarbuttons and tabpanels different prop interfaces if we aren't going to make them unique in the future, which I can't imagine a scenario where either of them would need other props.

teachertool/src/components/EvalResultDisplay.tsx
srietkerk: That's true, it can be used for anything, but I think it's nice to have the specificity here. You can look at the code and know that the ref that is being defined will be used for printing.

teachertool/src/components/RubricWorkspace.tsx
eanders-ms: If, down the line, we needed to add a new property to one of these components, the developer responsible for making this change would need to pause and understand the reasoning behind sharing this interface, and evaluate whether it is ok to split it off into a new one. They will think through this carefully. Is this future cognitive load justified here? I'm concerned we're degrading the orthogonality of the design without good reason. Sharing a property name isn't a good justification for a base class.

teachertool/src/components/EvalResultDisplay.tsx
eanders-ms: There's nothing enforcing this usage. It's just a ref. What about naming it for what it points to, maybe `resultsRef`?

srietkerk: So, would it be preferred that `printRef` gets renamed to `resultsRef` everywhere?

eanders-ms: Yes, I think that would work fine.

teachertool/src/components/PrintButton.tsx
```diff
@@ -1,17 +1,23 @@
...
-export const PrintButton: React.FC<PrintButtonProps> = ({ printRef, onHandlePrint }) => {
+export const PrintButton: React.FC<PrintButtonProps> = ({ printRef }) => {
+    const { state: teacherTool } = stateAndDispatch();
     const handlePrint = useReactToPrint({
         content: () => printRef.current,
-        onAfterPrint: onHandlePrint,
+        onPrintError: () => showToast(makeToast("error", lf("Unable to print evaluation results"), 2000)),
+        documentTitle: `MakeCode Project Evaluation Results - ${pxt.Util.sanitizeFileName(
+            teacherTool.projectMetadata?.name || Strings.UntitledProject
+        )}`,
```
eanders-ms, line 20: Can this be localized?

teachertool/src/components/PrintButton.tsx
```diff
@@ -1,17 +1,23 @@
...
 }
 
-export const PrintButton: React.FC<PrintButtonProps> = ({ printRef, onHandlePrint }) => {
+export const PrintButton: React.FC<PrintButtonProps> = ({ printRef }) => {
+    const { state: teacherTool } = stateAndDispatch();
     const handlePrint = useReactToPrint({
         content: () => printRef.current,
-        onAfterPrint: onHandlePrint,
+        onPrintError: () => showToast(makeToast("error", lf("Unable to print evaluation results"), 2000)),
+        documentTitle: `MakeCode Project Evaluation Results - ${pxt.Util.sanitizeFileName(
```
eanders-ms, line 18: Including the rubric name in the document title would be helpful, I think.

```suggestion
        documentTitle: `MakeCode - ${rubric name} - 
```

srietkerk: Yeah, I was torn about this. I think there are two scenarios to consider with the naming of the file:
1. What will make the teacher's job easier to find a specific student's submission results in a whole slew of files?
2. If students receive the file online, with the associated file name, do they need to know the rubric name?

Is the suggestion to include both the rubric and file name? So something like `MakeCode - My Rubric - Rietkerk Rock Paper Scissors`? I do think having something like that would help teachers, but it also seems really long. To be fair, what I had before was long, too. And if they gave the electronic file back to students, I would be under the impression that teachers would take out the rubric name, but maybe that wouldn't happen. I think I'd prefer to get user feedback on this and change it to something more permanent then, but maybe have the rubric and project name in for the time being?

teachertool/src/components/PrintButton.tsx
srietkerk: I can definitely put this into `lf`, but I figure we'll likely not want all of it localized. I can definitely localize whatever we decide to put as the leading descriptor (the `MakeCode Project Evaluation Results` here), but I figured we wouldn't want to localize the name of the project.

teachertool/src/components/PrintButton.tsx
thsparks: I think you could get rid of the MakeCode and just have "Rubric - Project", personally. But I also don't think it's a big deal either way.

teachertool/src/components/PrintButton.tsx
thsparks: Typically with translations like this, you'd do lf("MakeCode Project Evaluation Results - {0}", pxt.Util.sanitizeFileName(...))

The stuff passed into the format won't get translated, but the initial string will be formatted with the additional context that something is going to get put in place of that {0}.

teachertool/src/components/PrintButton.tsx
srietkerk: So if I go this route, I shouldn't localize anything, right?

eanders-ms: "Rubric - Project" seems like a good doc name. No localization needed either.

eanders-ms (approved):
Looks great.

thsparks (approved):

PR #9929, opened 2024-03-21
Include prefix when marking scripts as already migrated

riknoll:
When going to pxt-microbit beta the other day, I noticed a few of my scripts being left behind. Turns out I didn't account for the possibility of the same header existing multiple times in different versions of the pouch db!

This corrects that by prepending the browser prefix to the id we use for checking if something has been migrated.

### DIFF
--- a/webapp/src/idbworkspace.ts
+++ b/webapp/src/idbworkspace.ts
@@ -132,7 +132,7 @@ async function migratePouchAsync() {
             continue;
         }
 
-        alreadyMigratedList.setAsync(table, { id: migrationDbKey(prefix, id) });
+        await alreadyMigratedList.setAsync(table, { id: migrationDbKey(prefix, id) });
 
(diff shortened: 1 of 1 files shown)


webapp/src/idbworkspace.ts
```diff
@@ -128,11 +128,11 @@ async function migratePouchAsync() {
                 continue;
         }
 
-        if (await alreadyMigratedList.getAsync(table, id)) {
+        if (await alreadyMigratedList.getAsync(table, migrationDbKey(prefix, id))) {
             continue;
         }
 
-        alreadyMigratedList.setAsync(table, { id });
+        alreadyMigratedList.setAsync(table, { id: migrationDbKey(prefix, id) });
```
eanders-ms, line 135: await this?

eanders-ms (approved):

PR #9936, opened 2024-03-22
Teacher Tool: Block Picker

thsparks:
This change adds a "Block Picker" which can be used for customizable block parameters. Doing this involved:

1. Making it so the iframe always loads something, even when no sharelink has been provided. We need this so we can call into it to get the blocks. It's invisible when there is no share link, but it's still loaded.
2. Adding commands we can send to the iframe to get all the toolbox categories & blocks, then to get images of blocks (as xml, when we have it, or simply by id)
3. Adding a new modal for the block picker, which displays categories and blocks to choose from
4. Some adjustments to the LazyImage to make the loading look a little smoother (I think)

Upload Target: <del>https://makecode.microbit.org/app/400c169359d2defd8af3c8aa838f740f3a932406-c40b446680--eval?tc=1</del> https://makecode.microbit.org/app/a60b885ddbfa1cf3cc0f4407e122347ce0c1ebf9-d07aa77cc2--eval?tc=1

Screenshots

(In order to prevent modal resizing, the modal is large even when all categories are collapsed)
image

Expand to see blocks
image

I change the color of the button but do not insert an image of the block in the rubric. I think the actual blocks would be too big a lot of the time:
image


teachertool/src/state/state.ts
```diff
@@ -15,6 +15,9 @@ export type AppState = {
     validatorPlans: pxt.blocks.ValidatorPlan[] | undefined;
     autorun: boolean;
     confirmationOptions: ConfirmationModalOptions | undefined;
+    blockPickerOptions: BlockPickerOptions | undefined;
```
eanders-ms, line 18: At some point we may want to have a single `modalOptions` field as a compound type, e.g.: `modalOptions: ConfirmationModalOptions | BlockPickerOptions | undefined`

teachertool/src/transforms/loadToolboxCategoriesAsync.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    // Create a map so categories can be looked up by their name.
+    const mappedCategories: pxt.Map<pxt.editor.ToolboxCategoryDefinition> = categories.reduce((map, category) => {
+        if (shouldIncludeCategory(category)) {
+            // Remove blocks with duplicate ids
+            const filteredBlocks = category.blocks?.filter((block, index, self) => {
+                return self.findIndex(b => b.blockId === block.blockId) === index;
+            });
+
+            map[category.name!] = { ...category, blocks: filteredBlocks };
```
eanders-ms, line 27: What if `category.name` is null? I know this is safe because I noticed `shouldIncludeCategory` checks it. Maybe put a comment to that effect here?

eanders-ms (commented):
Looking great. Made some minor comments/suggestions.

teachertool/src/services/makecodeEditorService.ts
```diff
@@ -28,6 +28,9 @@ export function setEditorRef(ref: HTMLIFrameElement | undefined) {
         });
         driver.addEventListener("editorcontentloaded", ev => {
             AutorunService.poke();
+
+            // Reload all blocks.
+            loadToolboxCategoriesAsync();
```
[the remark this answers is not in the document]
thsparks: I think it's okay to just let this run in the background without awaiting. I'll update comments to indicate that's intentional.

teachertool/src/state/state.ts
thsparks: I went ahead and combined them, though I took a slightly different approach. May have gone a little overboard with it, but I wanted to avoid blindly casting to the desired types inside the modals themselves, so I ended up making an interface for the options with a modal field we can check against. LMK if you think it's too much.

eanders-ms: Your changes look fine, but I think it could be simplified a bit:
* Remove `modal` from `AppState`. The presence of a `modalOptions` is enough to know if a modal is showing, and which one.
* Use a union of types for the different modal options, with no base:
```ts
export interface ConfirmationModalOptions {
    modal: "confirmation";
    title: string;
    message: string;
    onCancel: () => void;
    onContinue: () => void;
}

export interface BlockPickerOptions {
    modal: "block-picker";
    criteriaInstanceId: string;
    paramName: string;
}

export type ModalOptions = ConfirmationModalOptions | BlockPickerOptions;
```

* For showing a modal, implement a single `showModal` transform that takes a `ModalOptions` parameter:
```ts
export function showModal(modalOptions: ModalOptions) {
   // dispatch the action
    ...
}
```

eanders-ms (approved):
Made one more suggestion for modal options, but looks great.

PR #9945, opened 2024-03-28
Refactor iframe communication and use MessageChannel

riknoll:
This PR does a few things:

1. Refactors the IFrameDriver into a base class and adds a new EditorDriver that takes the place of the old class
2. Adds another IFrameDriver that can be used to drive the asset editor embed (arcade.makecode.com/--asseteditor)
3. Adds a new IFrameEmbeddedClient that can be used within the embedded iframe to handle communication with the parent frame
4. Refactors our messaging protocol to use a MessageChannel so that we can support multiple iframe embeds at the same time

This is all 100% backwards compatible with old embeds. To use the new MessageChannel stuff, you need to add `?frameid=<some-unique-id>` to the iframe's src url, then everything will just work and it will automatically switch to using the MessageChannel. If you don't add that, the new code falls back to using the good old `postMessage` for communication. I went ahead and added it to the embed for the teacher tool and everything is working fine.

We still have a few iframe embeds that I'll want to eventually switch over to this new stuff:
* pxtrunner (both --docs and --run)
* the simulator (this will involve refactoring the existing pxsim SimDriver class)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -48,6 +48,7 @@ const cli = () => compileTsProject("cli", "built", true);
 const webapp = () => compileTsProject("webapp", "built", true);
 const reactCommon = () => compileTsProject("react-common", "built/react-common", true);
 const pxtblocks = () => compileTsProject("pxtblocks", "built/pxtblocks", true);
+const pxtservices = () => compileTsProject("pxtservices", "built/pxtservices", true);
 
 const pxtapp = () => gulp.src([
     "node_modules/lzma/src/lzma_worker-min.js",
@@ -118,7 +119,7 @@ function initWatch() {
         pxtlib,
         gulp.parallel(pxtcompiler, pxtsim, backendutils),
         pxtpy,
-        gulp.parallel(pxtblocks, pxteditor),
+        gulp.parallel(pxtblocks, pxteditor, pxtservices),
         gulp.parallel(pxtrunner, cli, pxtcommon),
         gulp.parallel(updatestrings, browserifyEmbed),
         gulp.parallel(pxtjs, pxtdts, pxtapp, pxtworker, pxtembed),
@@ -139,6 +140,7 @@ function initWatch() {
 
     gulp.watch("./pxtpy/**/*", gulp.series(pxtpy, ...tasks.slice(3)));
     gulp.watch("./pxtblocks/**/*", gulp.series(pxtblocks, ...tasks.slice(4)));
+    gulp.watch("./pxtservices/**/*", gulp.series(pxtservices, ...tasks.slice(4)));
 
     gulp.watch("./pxteditor/**/*", gulp.series(pxteditor, ...tasks.slice(4)));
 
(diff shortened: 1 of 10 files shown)


gulpfile.js
```diff
@@ -48,6 +48,7 @@ const cli = () => compileTsProject("cli", "built", true);
 const webapp = () => compileTsProject("webapp", "built", true);
 const reactCommon = () => compileTsProject("react-common", "built/react-common", true);
 const pxtblocks = () => compileTsProject("pxtblocks", "built/pxtblocks", true);
+const pxtservices = () => compileTsProject("pxtservices", "built/pxtservices", true);
```
eanders-ms, line 51: This is eerie. I just made all these exact changes locally.

gulpfile.js
```diff
@@ -139,6 +140,7 @@ function initWatch() {
 
     gulp.watch("./pxtpy/**/*", gulp.series(pxtpy, ...tasks.slice(3)));
     gulp.watch("./pxtblocks/**/*", gulp.series(pxtblocks, ...tasks.slice(4)));
+    gulp.watch("./pxtservices/**/*", gulp.series(pxtservices, ...tasks.slice(4)));
```
eanders-ms, line 143: Does the task index need to increment?
```suggestion
    gulp.watch("./pxtservices/**/*", gulp.series(pxtservices, ...tasks.slice(5)));
```

pxtservices/iframeEmbeddedClient.ts
```diff
@@ -0,0 +1,79 @@
...
+};
+
+type IFrameClientReadyMessage = {
+    type: "iframeclientready";
+};
+
+export type IframeClientMessage = IframeClientSetMessagePortRequest | IFrameClientReadyMessage;
+
+export class IFrameEmbeddedClient {
+    protected frameId: string;
```
eanders-ms, line 12: `frameId()` can return undefined. Teacher tool and other CRAs use strict typechecks and would trip on this.

```suggestion
    protected frameId: string | undefined;
```

pxtservices/iframeEmbeddedClient.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    protected sendReadyMessage() {
+        this.postMessage({
+            type: "iframeclientready"
+        });
+    }
+}
+
+function frameId(): string {
```
eanders-ms, line 71:
```suggestion
function frameId(): string | undefined {
```

pxtservices/iframeEmbeddedClient.ts
```diff
@@ -0,0 +1,79 @@
...
+export type IframeClientMessage = IframeClientSetMessagePortRequest | IFrameClientReadyMessage;
+
+export class IFrameEmbeddedClient {
+    protected frameId: string;
+    protected port: MessagePort;
+
+    constructor(protected messageHandler: (message: MessageEvent) => void) {
+        this.frameId = frameId();
+
+        window.addEventListener("message", this.onMessageReceived);
```
eanders-ms, line 18: Provide a mechanism to remove this listener? If the containing object is unmounted, for instance, this service should be able to be stopped.

gulpfile.js
riknoll: nope, it's the same because pxtblocks and pxtservices run at the same step

eanders-ms (approved):
Looks good. (Lint disagrees, but I'm not sure why).

PR #9953, opened 2024-04-09
Copilot Criteria Support

thsparks:
# Overview
This change adds the front-end support for copilot criteria. In essence, we call into the backend with the Share ID of the project and the question being asked, then the backend forwards that information to DeepPrompt and returns the result to us.

The backend changes will not be checked in until we get more feedback and a better sense for funding, so this change has a few workarounds to compensate while still enabling easy demos and experimentation. Notably, there is a new "copilot" flag that can be set to point the copilot backend request to a staging slot. For example, `http://localhost:3232/eval?copilot=thsparks` sets it to the thsparks staging slot.

# Adjustments to Evaluation Flow
This change necessitated a few adjustments to how we run evaluation of the rule set.

Firstly, it introduces the idea of **System Parameters**. These are parameters that get passed into a validator plan, but which the user doesn't necessarily need to input as a part of the criteria template. Instead, the teacher tool auto-populates the value during evaluation. For this scenario, it's used to pass along the ShareId, though hypothetical future scenarios could involve things like the target or student vs teacher modes.

Secondly, I've introduced **Teacher Tool Validator Plan Overrides** which allow the teacher tool to "intercept" specific validator plans and run its own validation instead of calling into the iframe. This is used to keep the ai validation contained purely in the teacher tool (for now) because I'm not actually sure we _want_ to expose it for anyone with a MakeCode iframe to call. These overrides are designed to mirror the context of the editor evaluation quite closely, so it's easy to move code from one to the other if needed.

# Additional Changes
Since the "evaluating..." state is much longer with AI Eval, I made some adjustments to what that state looks like. Specifically, I removed the dropdown and the notes while the result is still pending. Instead, there is a three-dot loading animation.

I made some minor adjustments to our catalog loading order so that test criteria gets placed at the top (and copilot is at the top of those). This was based on feedback for demo-ing, but I think it also makes sense to put testing criteria at the top, since that is presumably what someone cares about, if they specifically included test criteria.

I made some changes to CriteriaResultNotes and Textarea so that they resize when the content is set outside of user input, and so they re-check their height when the width changes.

# Follow Up Items
There are a few additional considerations I think we'll need to address in future changes, but I didn't want to include them here since it's already a sizable change. Those include:
1. Changes to auto-run so we don't send requests to the AI on every change
2. We need to tidy up the catalog, either with categories or an overhauled flow. It's pretty cluttered at the moment.

# Try It
Upload Target: https://makecode.microbit.org/app/2b32c0f2978128bb706a8eae85cfe3c680a8515c-8b3d39bfa7--eval?copilot=thsparks


teachertool/src/transforms/runEvaluateAsync.ts
```diff
@@ -94,11 +104,21 @@ export async function runEvaluateAsync(fromUserInteraction: boolean) {
...
+                let planResult = await runValidatorPlanOverrideAsync(plan);
+                if (!planResult) {
+                    planResult = await runValidatorPlanAsync(plan, loadedValidatorPlans);
+                }
 
                 if (planResult) {
-                    const result = planResult.result ? EvaluationStatus.Pass : EvaluationStatus.Fail;
-                    setEvalResultOutcome(criteriaInstance.instanceId, result);
+                    const result =
+                        planResult.result === undefined
```
srietkerk, line 115: Does this need to be a nested ternary? This will always boil down to the result being pass or fail, right? Can the update here then be planResult.result === undefined?

teachertool/src/services/backendRequests.ts
```diff
@@ -88,3 +88,30 @@ export async function loadTestableCollectionFromDocsAsync<T>(fileNames: string[]
...
+
+        if (!request.ok) {
+            throw new Error("Unable to reach Copilot");
+        }
+        result = await request.json();
+    } catch (e) {
+        logError(ErrorCode.askCopilotQuestion, e);
+    }
+
+    return result;
```
srietkerk, line 116: Is there any way to indicate progress with the thinking? Because the other evaluations are so fast, I found myself impatient as I waited. Not something needed for this PR and I don't feel strongly about it, but might be something interesting to investigate.

teachertool/src/components/CriteriaEvalResultDropdown.tsx
```diff
@@ -3,55 +3,48 @@ import { useMemo } from "react";
...
-        title: lf("not evaluated"),
-        label: lf("not evaluated"),
+        title: lf("not applicable"),
+        label: lf("N/A"),
     },
     {
         id: "fail",
         title: lf("needs work"),
-        label: lf("needs work"),
+        label: lf("Needs work"),
```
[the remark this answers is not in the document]
thsparks: I'm not aware of any pattern, but I think it looks cleaner with capitalized first letters when shown in the page.

teachertool/src/state/appStateContext.tsx
```diff
@@ -38,13 +38,20 @@ export function AppStateProvider(props: React.PropsWithChildren<{}>): React.Reac
     const url = window.location.href;
     const testCatalog = !!/testcatalog(?:[:=])1/.test(url) || !!/tc(?:[:=])1/.test(url);
 
+    const copilotSlot = url.match(/copilot=([^&]+)/);
+    const copilotEndpoint =
+        copilotSlot && copilotSlot[1]
+            ? `https://makecode-app-backend-ppe-${copilotSlot[1]}.azurewebsites.net/api/`
```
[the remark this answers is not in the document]
thsparks: Since this is all just temporary code until we can check in backend changes, I personally think it's okay to hardcode in this way.

teachertool/src/transforms/mergeEvalResult.ts
```diff
@@ -0,0 +1,18 @@
+import { stateAndDispatch } from "../state";
+import { EvaluationStatus } from "../types/criteria";
+import { setEvalResult } from "./setEvalResult";
+
+// This will set the outcome and notes for a given criteria instance id, but if the provided value is undefined, it will not change that value.
+export function mergeEvalResult(criteriaInstanceId: string, outcome?: EvaluationStatus, notes?: string) {
```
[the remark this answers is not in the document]
thsparks: I think you could call into the setEvalResult transform for that code (or if you're clearing all results, make a new transform that just clears all of them). But yeah, we wouldn't want to use this transform for that.

teachertool/src/transforms/runEvaluateAsync.ts
```diff
@@ -47,6 +50,13 @@ function generateValidatorPlan(
             return undefined;
         }
 
+        if (catalogParam.type === "system" && catalogParam.key) {
+            param.value = getSystemParameter(catalogParam.key, teacherTool);
+            if (!param.value) {
+                param.value = catalogParam.default;
```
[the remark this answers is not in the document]
thsparks: It's an optional field you can set on a parameter value in the json to define the initial (i.e. default) value when the criteria is added to the rubric. It's not guaranteed to be set, and if that's the case, it'll go into the if statement on line 60 and log an error, the same as if any other parameter is missing a value.

teachertool/src/transforms/runEvaluateAsync.ts
thsparks: It does not always boil down to pass or fail. If planResult.result is undefined, then result is CompleteWithNoResult. If planResult.result is set, then we decide pass/fail.

teachertool/src/services/backendRequests.ts
thsparks: Not at present. Deep Prompt just tells us if it's "in progress" or not. We could see if streaming the response text as it comes in or if getting some kind of % completion is in their plans, but I don't think it's possible currently.

teachertool/src/services/backendRequests.ts
```diff
@@ -88,3 +88,30 @@ export async function loadTestableCollectionFromDocsAsync<T>(fileNames: string[]
 
     return allResults;
 }
+
+export async function askCopilotQuestion(shareId: string, question: string): Promise<string | undefined> {
+    const { state: teacherTool } = stateAndDispatch();
+
+    const url = `${
+        teacherTool.copilotEndpointOverride ? teacherTool.copilotEndpointOverride : pxt.Cloud.apiRoot
```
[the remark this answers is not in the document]
thsparks: Hmm, maybe not. Changed it to remove this assumption.

teachertool/src/transforms/runEvaluateAsync.ts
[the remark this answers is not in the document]
srietkerk: Ah, makes sense. Thanks for clarifying that.

teachertool/src/transforms/runEvaluateAsync.ts
srietkerk: Oh, gotcha. Thanks!

teachertool/src/services/backendRequests.ts
```diff
@@ -88,3 +88,30 @@ export async function loadTestableCollectionFromDocsAsync<T>(fileNames: string[]
...
+    }copilot/question`;
+
+    const data = { id: shareId, question };
```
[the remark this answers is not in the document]
srietkerk: I think just wanting to make sure that SQL injection can't happen, but if DeepPrompt has protections, then I would imagine that nothing bad can happen.

srietkerk (approved):
Looks good to me! Exciting stuff!

teachertool/src/services/backendRequests.ts
srietkerk: Sounds good. It would be interesting to find out if that's something they plan to do, but it can probably be something we inquire about after demoing.

teachertool/src/components/CriteriaResultEntry.tsx
```diff
@@ -36,23 +38,29 @@ const AddNotesButton: React.FC<AddNotesButtonProps> = ({ criteriaId, setShowInpu
 
 interface CriteriaResultNotesProps {
     criteriaId: string;
-    notes?: string;
 }
 
-const CriteriaResultNotes: React.FC<CriteriaResultNotesProps> = ({ criteriaId, notes }) => {
+const CriteriaResultNotes: React.FC<CriteriaResultNotesProps> = ({ criteriaId }) => {
     const { state: teacherTool } = useContext(AppStateContext);
+    const [value, setValue] = useState(teacherTool.evalResults[criteriaId]?.notes ?? "");
```
eanders-ms, line 45: I'm having difficulty understanding why `value` state was introduced. Was it something to do with initializing DebouncedTextArea?

eanders-ms (approved):
Looks good. Just one small question.

teachertool/src/components/CriteriaResultEntry.tsx
thsparks: Hm. I genuinely can't recall. I'll undo it.

Issue #9539, opened 2023-06-02
Upgrade Crowdin API's


Jaqster:
Message from Crowdin - 
Hello,
You've received this email because you're currently using Crowdin API v1 or services built upon it (e.g., CLI, SDK, etc.).
In our continuous effort to provide you with the best experience and improved functionality, we will be removing API v1 from our platform. Effective from November 30, 2023, API v1 will no longer be supported.
This change means that any integrations, tools like CLI, SDK, or other services relying on API v1 will cease to function. To ensure uninterrupted service, we recommend migrating to API v2, which offers enhanced features and compatibility.
Learn more:
- https://developer.crowdin.com/api/v2
- https://crowdin.github.io/crowdin-cli

From Joey - The docs upload path in the CI builds still is using v1 api, I had given it a quick shot and backed off upgrading it back when I changed how the backend read out the translations from the server making requests to our backup repo- it was very early in their v2 api and it was missing lots of functionality that appear to have been added since then (also, that upload path in our cli is a little clunky at the best of times, e.g. the way it handles making sure folders exist is to just send a "make folder" request for each subfolder and checking the error code to see if the folder already existed). While we're making changes we should give it another look to see how we can trim it down to only upload changed files, as it's taking a big chunk of our build time currently.

sae220:
I'd like to help for this issue but how can I test for my code?
I think I need crowdin API key for development.

sae220:
I noticed I can generate crowdin API key by myself.
Now I'm updating some code for crowdin API v2 with the test by my API key.
But some API endpoints (/projects/{projectId}/files) return 403 response.
I want to test how it work with `recursion` parameter.
I guess it is needed that crowdin manager make some treatment for my key.

sae220:
It is still in the middle but I updated some code into using crowdin api v2.
https://github.com/sae220/pxt/commit/38d6fed0922ba03420f38d2b2cf1f5488a24b600
I'll update the rest of code and test some.
After that I'll create pull request.

ganicke:
@sae220 - some upgrade work was done to update to the V2 API in #9954.

PR #9954, opened 2024-04-09
Upgrade to Crowdin V2 APIs

riknoll:
Upgrading our Crowdin support to use the V2 APIs.

I've done a little bit of testing and things seems to be running as they should. I was able to download strings from arcade and upload a few (unchanged) files to pxt-ev3. In any case, this is safe to merge since we don't have any updated API tokens in place. I plan to test this with pxt-ev3 first since that target has fewer eyes on it.

This PR also removes support for crowdin branches. AFAIK we don't actually use them anywhere (@pelikhan correct me if I'm wrong) and they are not compatible with the directory based structure in the V2 api, so I don't expect this to be an issue. I still need to pull out a lot of references to crowdin branches in the code but I'll do that in a separate PR.

Also, I have temporarily disabled the `pxt crowdin stats` command but I've got a plan to bring it back.


pelikhan:
Yup not used

pelikhan:
My scripts use their cli

cli/crowdinApi.ts
```diff
@@ -0,0 +1,371 @@
...
+    const storageResponse = await uploadStorageApi.addStorage(fileName, fileContent);
+
+    await sourceFilesApi.updateOrRestoreFile(projectId, fileId, {
+        storageId: storageResponse.data.id,
+    });
+}
+
+function getClient() {
+    if (!client) {
+        client = new crowdin(crowdinCredentials());
```
jwunderl, line 345: might want to pass opts with a retryConfig as there have historically been a decent number of spurious errors (pass that in pxt-translations along with ignoring 304's and it was very necessary at the time at least)

jwunderl (approved):

ganicke:
Gonna tag this one to close #9539

PR #9956, opened 2024-04-10
Bug fix: Microsoft profile pic not showing

srietkerk:
The data that we receive from the Microsoft auth service is does not have a `dataUrl` anymore, but a whole picture object where the image is encoded. I added a function to handle an encoded image for the avatar picture.

Fixes https://github.com/microsoft/pxt-microbit/issues/5477


eanders-ms (approved):
💥

Issue #9926, opened 2024-03-19
No localized strings now when setting language to 'ja-HIRA'


ganicke:
It seems that something isn't fully connected with the addition of "ja-HIRA". Not seeing any strings at all in the response payload when `ja-hira` is selected. The response should at least have the default preload of "ja":

![image](https://github.com/microsoft/pxt/assets/27789908/67c691e8-dd0b-4f4e-b0e5-7712be8cb48e)

sae220:
I guess each editor uses `{host}/api/translations` API and this API uses crowdin API internally.
And now This API doesn't work well.
We should check whether crowdin API which is used internally works well.
I want someone to check this or generate crowdin API key for my checking.

sae220:
I couldn't find where the code of `{host}/api/translations` is.
I want someone to tell me where.

sae220:
I noticed I can generate crowdin API key by myself.
I generated my key and made some tests.
I found the locale of "ja-hira" edited into upper case "ja-HIRA" but the id is still lower case "ja-hira".

Now I'm trying to update the crowdin API version, so I will be able to solve this problem after that.

sae220:
I found it works well with query string `translate=1`.
With query string `translate=1`, crowdin JIPT is used, I guess.

PR #9961, opened 2024-04-10
Bring crowdinStatsCommand back

sae220:
This command is temporarily disabled by @riknoll in #9954 .
I haven't done any testing because I can't use some API with my key.

### DIFF
--- a/cli/crowdin.ts
+++ b/cli/crowdin.ts
@@ -5,7 +5,7 @@ import * as path from 'path';
 import Map = pxt.Map;
 
 import * as commandParser from './commandparser';
-import { downloadFileTranslationsAsync, listFilesAsync, uploadFileAsync } from './crowdinApi';
+import { downloadFileTranslationsAsync, getAllFiles, getLanguageProgressAsync, getProjectInfoAsync, listFilesAsync, normalizePath, uploadFileAsync } from './crowdinApi';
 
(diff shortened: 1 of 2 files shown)


riknoll:
ah, @sae220 I actually just opened a PR fixing this! Check it our here:

https://github.com/microsoft/pxt/pull/9962

Interesting that it doesn't work with your API key... do you happen to know which API was failing? what permissions did you give it?

riknoll:
Also, were you using this command? If so, can I ask what for? I might be interested in your use case to see if I could make something more comprehensive for you.

sae220:
Thanks a lot for your great help.

> do you happen to know which API was failing? what permissions did you give it?
SorceFiles API (like List Files). This returns 403 Forbidden. I gave all permissons.

> Also, were you using this command? If so, can I ask what for?
I only did some test to try to solve #9926. I'm sorry that I can't give you interesting use case.

riknoll:
No worries! Thanks for the PR in any case.

I'll have to investigate if Crowdin has a way to get file IDs without list files, it's a really slow request and the permissions on it do seem weird.

sae220:
I asked crowdin support team why I can't access some API. The answer is below.

> Yes, to access all API capabilities you need to be a manager and have a manager token. Also developer role would work

Now I ask ganicke to give me in crowdin conversations.

PR #9964, opened 2024-04-11
Remove all traces of crowdin branch support

riknoll:
As mentioned in https://github.com/microsoft/pxt/pull/9954, I removed support for crowdin branches with the V2 upgrade. This PR removes the remaining traces of support we had throughout the webapp.

There was surprising amount of code for a crowdin feature that AFAIK we have never used.


srietkerk (approved):
LGTM.

Just curious if you have any context on why we implemented this if we weren't really using it?

riknoll:
@srietkerk nope. this code was checked in 8 years ago

PR #9965, opened 2024-04-11
Teacher Tool: Catalog Clean-Up

thsparks:
Now that parameterized criteria are supported in beta, I moved those out of test, then I reduced some of the duplication in our catalog.

Kept:
- All function definitions have comments
- At least one custom variable is set
- At least one variable is accessed
- At least ${count} comments
- ${Block} used ${count} times
- At least ${count} loops used
- At least ${count} custom functions exist and get called
- [Test] Ask Copilot: ${question}

Removed:
- Two different kinds of loops used (in favor of `At least ${count} loops used`. They have slightly different functionality but I think it's ok to just have one)
- A custom function exists and gets called (in favor of `At least ${count} custom functions exist and get called`)
- At least one block has comments (in favor of `At least ${count} comments`)
- A custom variable's value is set to a random number (seemed too specific)
- The 'pick random' block is used (in favor of `${Block} used ${count} times`)
- Compare two numbers for equality (seemed too specific)

Related micro:bit change: https://github.com/microsoft/pxt-microbit/pull/5592

While testing, I noticed a bug where parameters seem to be getting overwritten when we have multiple instances of the same catalog criteria, but I'll look into that separately.


common-docs/teachertool/catalog-shared.json
```diff
@@ -43,25 +22,69 @@
             "description": "At least one variable's value is read."
         },
         {
-            "id": "8E6F9A92-2D22-4BB0-A77A-BD7490D3CEF7",
-            "use": "variable_set_random",
-            "template": "A custom variable's value is set to a random number",
+            "id": "7AE7EA2A-3AC8-42DC-89DB-65E3AE157156",
+            "use": "block_comment_used",
+            "template": "At least ${count} comments",
```
srietkerk, line 27: Not for this change, but wondering if we're still planning on updating this validator since it's just looking at block comments right now.

srietkerk (approved):

common-docs/teachertool/catalog-shared.json
thsparks: Yeah, it'd be good to do. Not highest pri at the moment, though.

thsparks: When I get around to moving all the work items from our task board over to issues/bugs, I'll make sure we include that one.

PR #9967, opened 2024-04-12
Teacher Tool: Fix Parameters on Criteria Instances that Reference the Same Validator Plans

thsparks:
While simplifying the catalog, I noticed parameters were overwriting each other when we had multiple criteria instances that referenced the same validator plan. This was happening because all of them were pointing to the same plan object. To fix it, we can make a deep copy of the validator plan before we modify it.

I also updated the logDebug function to accept a data parameter to help keep our logs a little tidier.

Upload Target: https://makecode.microbit.org/app/cb0809faf5a812b45067c743d1bf2d0c7a3bc799-65bd000c7b--eval?tc=1

### DIFF
--- a/teachertool/src/transforms/loadProjectMetadataAsync.ts
+++ b/teachertool/src/transforms/loadProjectMetadataAsync.ts
@@ -34,5 +34,5 @@ export async function loadProjectMetadataAsync(inputText: string, shareLink: str
     };
     dispatch(Actions.setProjectMetadata(projectData));
     initNewProjectResults();
-    logDebug(`Loaded project metadata: ${JSON.stringify(projMeta)}`);
+    logDebug("Loaded project metadata", projMeta);
 }
--- a/teachertool/src/services/loggingService.ts
+++ b/teachertool/src/services/loggingService.ts
@@ -32,8 +32,9 @@ export const logInfo = (message: any) => {
     console.log(timestamp(), message);
 };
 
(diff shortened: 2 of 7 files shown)


teachertool/src/transforms/loadProjectMetadataAsync.ts
```diff
@@ -34,5 +34,5 @@ export async function loadProjectMetadataAsync(inputText: string, shareLink: str
     };
     dispatch(Actions.setProjectMetadata(projectData));
     initNewProjectResults();
-    logDebug(`Loaded project metadata: ${JSON.stringify(projMeta)}`);
+    logDebug("Loaded project metadata", projMeta);
```
srietkerk, line 37: Why is the `JSON.stringify` no longer applied to the `projMeta`?

thsparks: It's now passed into the log as an object, which allows the console to display it in a more user-friendly fashion (you can expand/collapse fields instead of having just one long string)

thsparks: Like this:
image

srietkerk: Oh, cool! Thanks for clarifying!

srietkerk (approved):

PR #9971, opened 2024-04-15
Fix TOC header and card list roles in docs

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/5475
Fixes https://github.com/microsoft/pxt-arcade/issues/4869

Also fixes the roles for the search box in the menubar @srietkerk FYI (sorry, I thought this was part of my bug. hopefully you weren't working on it)

While I was at it, I also removed some semantic ui stuff (but not all of it)


srietkerk (approved):
Wasn't working on this, so no problem! Nice!

PR #9973, opened 2024-04-16
Automatically adjust block colors to have a contrast ratio of 4.5

riknoll:
Motivated by: https://github.com/microsoft/pxt-arcade/issues/4870

This PR adds a little code to automatically adjust all block background colors to have a contrast ratio of at least 4.5 with white foreground text. The algorithm I'm using is pretty dumb: I just lower the luminosity until a color with the correct contrast ratio is generated and cache the result.

I placed this behind a feature flag for now because the general consensus is that this makes all the blocks look duller/uglier and creates a breaking change with old screenshots. The goal should be for us to eventually turn this on in all targets; we might have to adjust the base colors of some categories to make them more visually appealing though (arrays is particularly ugly).

Here is an upload target in micro:bit with the feature turned on:

https://makecode.microbit.org/app/36f327a321d404020e3804ed55f5a3567b2e929f-3d8dd326ca

### DIFF
--- a/pxtblocks/builtins/misc.ts
+++ b/pxtblocks/builtins/misc.ts
@@ -8,6 +8,12 @@ export function initOnStart() {
     const onStartDef = pxt.blocks.getBlockDefinition(ts.pxtc.ON_START_TYPE);
     Blockly.Blocks[ts.pxtc.ON_START_TYPE] = {
         init: function () {
+            let colorOverride = pxt.appTarget.runtime?.onStartColor;
+
+            if (colorOverride) {
+                colorOverride = pxt.toolbox.getAccessibleBackground(colorOverride);
+            }
+
             this.jsonInit({
                 "message0": onStartDef.block["message0"],
                 "args0": [
@@ -19,7 +25,7 @@ export function initOnStart() {
                         "name": "HANDLER"
                     }
                 ],
-                "colour": (pxt.appTarget.runtime ? pxt.appTarget.runtime.onStartColor : '') || pxt.toolbox.getNamespaceColor('loops')
+                "colour": colorOverride || pxt.toolbox.getNamespaceColor('loops')
             });
 
             setHelpResources(this,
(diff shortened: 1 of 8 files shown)


riknoll:
also, should be noted that turning this flag on will fix https://github.com/microsoft/pxt-arcade/issues/4870

pxtlib/color.ts
```diff
@@ -0,0 +1,115 @@
...
+
+        const r2 = (r <= 0.03928) ? r / 12.92 : Math.pow((r + 0.055) / 1.055, 2.4);
+        const g2 = (g <= 0.03928) ? g / 12.92 : Math.pow((g + 0.055) / 1.055, 2.4);
+        const b2 = (b <= 0.03928) ? b / 12.92 : Math.pow((b + 0.055) / 1.055, 2.4);
+
+        return 0.2126 * r2 + 0.7152 * g2 + 0.0722 * b2;
+    }
+
+    function contrastRatio(fg: string, bg: string) {
+        return (relativeLuminance(fg) + 0.05) / (relativeLuminance(bg) + 0.05)
```
srietkerk, line 73: Why do you need to add 0.05 to each of these values? Doesn't adding it to both the top and bottom cancel the effects?

srietkerk:
Also, is there a reason that `on start`'s color doesn't change?

pxtlib/color.ts
riknoll: this is just the formula (I didn't come up with it), https://www.w3.org/TR/WCAG20/#contrast-ratiodef

adding to bottom and top doesn't cancel out, though! that's multiplication

riknoll:
ah, that would be because i missed on start! will fix

pxtlib/color.ts
srietkerk: Ah, right. Thanks!

pxtblocks/builtins/misc.ts
```diff
@@ -8,6 +8,12 @@ export function initOnStart() {
     const onStartDef = pxt.blocks.getBlockDefinition(ts.pxtc.ON_START_TYPE);
     Blockly.Blocks[ts.pxtc.ON_START_TYPE] = {
         init: function () {
+            let colorOverride = pxt.appTarget.runtime?.onStartColor;
```
srietkerk, line 11: So we special-case the onStart block? Why?

srietkerk (approved):
Just a question for my own understanding, otherwise looks good.

pxtblocks/builtins/misc.ts
riknoll: on-start is kind of a special builtin block because we give targets a lot of control over where it shows up. For example, in micro:bit it's in the basic category but everywhere else it's in loops. iirc we added this option for cue because they had an on-start that doesn't show up in any category, it was just always on the workspace and couldn't be deleted.

PR #9974, opened 2024-04-16
Fix various docs/homepage accessibility bugs

riknoll:
This PR fixes a bunch of small accessibility bugs. Each commit is a single fix, so you can also review it one by one if that helps.

Fixes https://github.com/microsoft/pxt-microbit/issues/5495
Fixes https://github.com/microsoft/pxt-arcade/issues/4865
Fixes https://github.com/microsoft/pxt-arcade/issues/4868 
Fixes https://github.com/microsoft/pxt-microbit/issues/5466
Fixes https://github.com/microsoft/pxt-arcade/issues/3969
Fixes https://github.com/microsoft/pxt-microbit/issues/5473

Does the following:

1. Adds underlines to links within paragraphs in docs pages
2. Adds a tile attribute to simulator frames embedded in docs pages
3. Sets the lang on docs html pages to "en" instead of `@locale@`
4. Gives the carousel arrows on the homepage an outline when focused
5. Marks landmark roles on the docs pages
6. Adds a `gulp pxtrunner` shortcut for compiling just pxtrunner (much like `gulp webapp`)

In regards to the `@locale@` change, the backend wasn't patching this with anything (it was returning the empty string) so I'm not sure how this was supposed to work. We also use this on index.html, but it seems like it always returns `en` regardless of what my browser locale is or what I pass as a query parameter. Rather than debug this, I opted to just hardcode it to `en` which is what we do in most of our other webapps anyhow and seems to be the only thing the backend ever returns.

### DIFF
--- a/docfiles/docs.html
+++ b/docfiles/docs.html
@@ -36,7 +36,7 @@
 
 				<!-- @include header.html -->
 
-				<div class="main ui container fluid mainbody">
+				<div class="main ui container fluid mainbody" role="main">
 					<button id="printbtn" class="circular ui icon right floated button hideprint hidemobile"
 						title="Print" aria-label="Print">
 						<i class="icon print"></i>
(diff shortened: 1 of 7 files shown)


thsparks (approved):
LGTM!

PR #9978, opened 2024-04-18
Fix github accessibility bugs and add an accordion component

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/5498
Fixes https://github.com/microsoft/pxt-microbit/issues/5503
Fixes https://github.com/microsoft/pxt-microbit/issues/5500

This PR contains two small accessibility fixes for the github integration and one larger one that also adds a new react-common accordion component.

The "History" portion of the GitHub integration is where the accordion is used. This overhauls that section with react-common components and fixes https://github.com/microsoft/pxt-microbit/issues/5500 in the process. The UI is basically unchanged.

### DIFF
--- a/react-common/components/controls/Accordion/Accordion.tsx
+++ b/react-common/components/controls/Accordion/Accordion.tsx
@@ -29,7 +29,6 @@ export const Accordion = (props: AccordionProps) => {
         role,
     } = props;
 
-
     return (
         <AccordionProvider>
             <div
@@ -74,7 +73,6 @@ export const AccordionItem = (props: AccordionItemProps) => {
         }
     }, [isExpanded]);
 
-
     return (
         <div
             className={classList("common-accordion", className)}
@@ -108,7 +106,7 @@ export const AccordionItem = (props: AccordionItemProps) => {
             <div
                 id={panelId}
                 className="common-accordion-panel-outer"
-                style={{ display: expanded ? "block" : "none" }}
+                style={{ display: isExpanded ? "block" : "none" }}
             >
                 {isExpanded && mappedChildren[1]}
             </div>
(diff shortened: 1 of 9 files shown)


webapp/src/gitjson.tsx
```diff
@@ -18,6 +18,9 @@ import * as pxtblockly from "../../pxtblocks";
 import IProjectView = pxt.editor.IProjectView;
 import UserInfo = pxt.editor.UserInfo;
 
+import { Accordion, AccordionItem, AccordionHeader, AccordionPanel } from "../../react-common/components/controls/Accordion/Accordion";
```
eanders-ms, line 21: As a design, it might be nice to scope the child components to the Accordion parent type, e.g.:
```jsx
    <Accordion>
        <Accordion.Item>
            <Accordion.Header>
                ...
            </Accordion.Header>
            <Accordion.Panel>
                ...
            </Accordion.Panel>
    </Accordion>
```

eanders-ms: Example here: https://github.com/microsoft/pxt/blob/master/teachertool/src/components/Toolbar.tsx

react-common/components/controls/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,153 @@
...
+        id,
+        className,
+        ariaLabel,
+        ariaHidden,
+        ariaDescribedBy,
+        role,
+    } = props;
+
+    return (
+        <AccordionProvider>
```
eanders-ms, line 33: Wonderful use of a local reducer here!

eanders-ms (approved):
A lot of changes here. Looks fine overall.

webapp/src/gitjson.tsx
riknoll: done!

PR #9983, opened 2024-04-23
Teacher Tool: Make Catalog Non-Modal

thsparks:
## What
This change makes it so you can keep the catalog open, add multiple criteria, and interact with other elements on the page without having to close the catalog.

## Why
The process of having to open the catalog, check a box (or boxes), then click a button to confirm worked fairly well for single-use criteria, but it feels cumbersome when you want to add multiple instances of the same criteria (like copilot questions or block-exists checks), since you can only add one at a time.

## How
I created a new Catalog Overlay that looks similar to a modal, but it is contained within the right side of the screen. When the user clicks on a criteria, it now gets added immediately. If it's a single-use criteria, the button is then disabled and shows a "Max" indicator. Otherwise, a quick "checkmark" appears to acknowledge the addition, then fades out.

### Screen Reader Announcer
In order to provide screen reader information when adding criteria, I added a Screen Reader Announcer component which can be used to send messages to the screen reader. This is kind of similar to the FluentUI [Announced](https://developer.microsoft.com/en-us/fluentui#/controls/web/announced) component.

At some point, we should also hook this up to our toasts so they announce whatever they're displaying, but I didn't want to lump that into this change. Same for the checklist remove buttons.

### "Allow Multiple" in Json
We were relying on the presence of parameters to determine if a criteria should allow multiple instances, but this doesn't work very well. Some parameterized criteria (like "At least X comments") should really only allow one, still. The correct thing to do is probably just to make a field in json, but I didn't want to do that in this change. As such, I didn't spend much time on the `criteriaIsSelectable` checks here. That should all be tidied up in the json changes later on.

## Screenshots
image

image

## Try It
Upload Target: <del>https://makecode.microbit.org/app/dcf2a8c4e75daf42961544654658683ecd329fd1-955a2d1201--eval</del>  https://makecode.microbit.org/app/9f9ad3a8ac1fa0be1f265c4052fa1471087f9901-44ac4a4284--eval


teachertool/src/components/CatalogOverlay.tsx
```diff
@@ -0,0 +1,149 @@
...
+    const criteria = useMemo<CatalogCriteria[]>(
+        () => getCatalogCriteria(teacherTool),
+        [teacherTool.catalog, teacherTool.rubric]
+    );
+
+    function onItemClicked(c: CatalogCriteria) {
+        addCriteriaToRubric([c.id]);
+        setRecentlyAddedIds([...recentlyAddedIds, c.id]);
+
+        setTimeout(() => {
```
thsparks, line 92: This seems unreliable when adding many in quick succession. May need to adjust approach.

Jaqster:
Feels nice!  I like the little Plus icon to add a rule.  I also like that I can add a blocks exist rule and then select which block I want without having to close my rule window... didn't realize how jarring that was before.

teachertool/src/components/CatalogOverlay.tsx
srietkerk: Hmm, I see what you're saying. Maybe try a queue? The adding the id can stay the same since it's basically enqueuing the entry in a JavaScript way, but then when removing the id from the list, instead of filtering, just pop the first item off the list. I'm not fully certain how this will look, especially since I've seen `setTimeout` behave buggy before, but it might help.

thsparks: I'm working on this now. Best approach I can think of is to use a map, but the real issue is not so much the data structure. It's the async nature of React state updates, so sometimes one change ends up overwriting another.

Turns out there's a React feature to help with this, where you can pass "[updater functions](https://react.dev/reference/react/useState#passing-the-updater-function)" into set state calls, which accept the previous state as a parameter and modify that instead of acting on the state var directly. This seems to work well. Now I'm just tidying things up before pushing.

teachertool/src/components/CatalogOverlay.tsx
```diff
@@ -0,0 +1,149 @@
...
+    recentlyAdded: boolean;
+}
+const CatalogItemLabel: React.FC<CatalogItemLabelProps> = ({
+    catalogCriteria,
+    allowsMultiple,
+    existingInstanceCount,
+    recentlyAdded,
+}) => {
+    const canAddMore = allowsMultiple || existingInstanceCount === 0;
+    const showRecentlyAddedIndicator = recentlyAdded && canAddMore;
```
srietkerk, line 46: In relation to the "adding multiple criteria in quick succession" problem, I'm wondering if this might also cause problems being static. It might be worth trying wrapping this in a `useEffect` to see if anything changes.

teachertool/src/components/CatalogOverlay.tsx
```diff
@@ -0,0 +1,149 @@
...
+    return (
+        <div className={css["catalog-item-label"]}>
+            <div className={css["action-indicators"]}>
+                {canAddMore ? (
+                    <>
+                        <i
```
srietkerk, line 52: Still related, I was thinking that the icons could also be rendered via a ternary operator. I'm not sure which would actually be faster, but I feel like it might take more time to change the css class and have the css apply to the icon to hide it rather than just saying if this is true, render this, other wise render the other one.

teachertool/src/components/CatalogOverlay.tsx
srietkerk: Oh, nice! My other comments might not be super helpful, then, just a heads-up.

teachertool/src/components/CatalogOverlay.tsx
thsparks: I think moving these into a useEffect wouldn't help (and would actually be less efficient, since I'd have to make state vars and whatnot for them). They're just quick calculations based directly on the props. I think it's fine to re-do them. "adding multiple criteria in quick succession" should be fixed now with changes in c7b41141294206f7e462b8460e3774cf4d16f91a and 798107748afa97c45399c033c5630fbc2bb7940f

teachertool/src/components/CatalogOverlay.tsx
thsparks: The reason it's done this way is to allow us to "fade" between the two by changing opacity over time (with css transitions). If we just render one or the other conditionally, the immediate switch can look a little jarring.

teachertool/src/components/CatalogOverlay.tsx
```diff
@@ -78,19 +78,27 @@ const CatalogList: React.FC = () => {
...
+    const [recentlyAddedIds, setRecentlyAddedIds] = useState<pxsim.Map<boolean>>({}); // Id -> bool, True if recently added. False (or not in list) otherwise.
 
     const criteria = useMemo<CatalogCriteria[]>(
         () => getCatalogCriteria(teacherTool),
         [teacherTool.catalog, teacherTool.rubric]
     );
 
+    function updateRecentlyAddedValue(id: string, value: boolean) {
+        setRecentlyAddedIds(prevState => {
+            const newState = { ...prevState };
```
srietkerk, line 90: Do you need to make this newState variable? If the updater function is going to modify the previous state, adding/updating the `id` entry of that state should work, right?

srietkerk: Oh, right, makes sense!

srietkerk (approved):
Looks great! And the `ScreenReaderAnnouncer` is awesome!!

PR #9984, opened 2024-04-24
Fix bug causing project conflicts when migrating scripts from pouch db

riknoll:
I finally figured out what's causing all of the project conflicts in beta! As I was debugging things today, I happened to notice that PouchDB was sometimes keeping multiple entries for the same document. I'm guessing this is part of some sort of internal redundancy mechanism that we don't use, but it was thwarting our project migration because we would sometimes end up migrating an older version of a script instead of the more recent version. When we go to sync this older version, we get an exception and it causes the conflict.

Luckily, it appears the revision number is stored in the pouch db document id, so it's easy to parse.

### DIFF
--- a/webapp/src/idbworkspace.ts
+++ b/webapp/src/idbworkspace.ts
@@ -74,6 +74,14 @@ async function checkIfPouchDbExistsAsync() {
     return result;
 }
 
+interface MigrationEntry {
+    prefix?: string;
+    table: string;
+    id: string;
+    rev: number;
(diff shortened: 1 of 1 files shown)


abchatra:
Do we need to port this to Minecraft?

riknoll:
@abchatra yep. I doubt it's causing any noticeable bugs over there but better safe than sorry

thsparks (approved):
Nice find!

PR #9989, opened 2024-04-29
Accessibility compliance: Improve share dialog accessibility

srietkerk:
I've made a couple of changes with this PR, but I've done so much fiddling and changing and testing across these commits that it's hard all of what's changed. The code in this PR does fix the following issues:

Fixes https://github.com/microsoft/pxt-microbit/issues/5468
Fixes https://github.com/microsoft/pxt-microbit/issues/5472
Fixes https://github.com/microsoft/pxt-microbit/issues/5488
Fixes https://github.com/microsoft/pxt-microbit/issues/5471

~~I also did some work for this issue . I finagled with this a bit and got it to a place that looked okay, but it really made the other zoom and screen experiences just bad, so I scratched the progress made. I did make a change, though, that if a screen is small enough, I change the height of the modal overlay to be 95% of the screen so more of the modal is usable.~~
^ this no longer applies as making the modal scrollable fixes the problem

It will be easier to see these changes just by playing with difference zooms, screen sizes, and resolutions. Let me know if there are other things that I should adjust.
~~Upload target: https://makecode.microbit.org/app/44c2bdbef7436f5a3215b42b34cf672cb13298ac-2ae3b742d0#editor~~
New upload target: https://makecode.microbit.org/app/9cc334c6852c269cc944ea1241973ece3d4a12a6-950d2c09e5#


react-common/components/share/ShareInfo.tsx
```diff
@@ -32,6 +32,19 @@ export interface ShareInfoProps {
     onClose: () => void;
 }
 
+function useWindowWidth() {
```
riknoll, line 35: can this not be done with CSS?

react-common/styles/react-common-breakpoints.less
```diff
@@ -1,17 +1,22 @@
 @mobileBreakpoint: 320px;
+@shareModalBreakpoint: 440px;
```
riknoll, line 2: I don't like using non-standard breakpoints like this. can we look into altering the design to works with the existing breakpoints? if not, lets put this breakpoint in the share's less files and not the common one.

react-common/styles/share/share.less
```diff
@@ -314,7 +315,46 @@
     }
 }
 
-@media @mobileAndBelow {
+@media only screen and (max-width: 854px) {
```
riknoll, line 318: these numbers seem really arbitrary. what are they based on?

riknoll (changes requested):

react-common/components/share/ShareInfo.tsx
srietkerk: No, this is needed in JavaScript. We are already conditionally rendering the embed button with css and that's causing problems with tab navigation. On live, if you shrink the screen to make the embed button go away and then try to tab past the `copy` button, you can't. You can get to the lower buttons with shift tabbing, but that shouldn't be something that's necessary. That's because the embed button is not displayed and the tab index isn't set correctly, causing problems. I tried just checking the width at the rendering of the button, but that means that if the width of the screen gets larger, the embed button's `enabled` state won't change, thus, this listener needs to be here.

riknoll: ah, I see. then the right fix here should actually be in the focus trap in react-common, it should skip elements that have display: none.

riknoll: [this function](https://github.com/microsoft/pxt/blob/master/react-common/components/controls/FocusTrap.tsx#L29) needs to filter out elements with display: none. that should fix this bug and not require any changes to the css

react-common/styles/react-common-breakpoints.less
srietkerk: Yeah, I'll remove these for now, and hopefully we can get a solution for everything working together for smaller screens in a later change.

react-common/styles/share/share.less
srietkerk: The 854 was an observed breakpoint, but on further testing, it looks like the more accurate measurement was the 991 breakpoint that we have for tablets.

The 513 is just for 512px screen widths; this is particularly for https://github.com/microsoft/pxt-microbit/issues/5468. When a screen reaches 512px in width on a 1280x720px resolution screen, making the modal break into the mobile view is not ideal and hard to work with. There needs to be an in-between before the content on the share modal orients as a column and the styling here is meant to help bridge that. This might be something that can also be removed in favor of a re-design, though.

react-common/styles/react-common-breakpoints.less
srietkerk: I do need to make a new rule, though, because there needs to be an in-between for smaller tablet sizes and the like. From playing with it, I think that the modal should switch flex-directions at smaller screens than we currently have it.

react-common/components/share/ShareInfo.tsx
srietkerk: I was having trouble with finding the right query selector, I will make this change when I get back.

react-common/components/controls/FocusList.tsx
```diff
@@ -57,42 +59,6 @@ export const FocusList = (props: FocusListProps) => {
         }
     }
 
-    const isFocusable = (e: HTMLElement) => {
-        return e.getAttribute("data-isfocusable") === "true"
```
riknoll, line 61: the data-isfocusable attribute still needs to be checked for the focuslist. Maybe findNextFocusableElement should optionally take in a predicate function?

react-common/components/util.tsx
```diff
@@ -91,4 +91,25 @@ export function screenToSVGCoord(ref: SVGSVGElement, coord: ClientCoordinates) {
     screenCoord.x = coord.clientX;
     screenCoord.y = coord.clientY;
     return screenCoord.matrixTransform(ref.getScreenCTM().inverse());
+}
+
+export function findNextFocusableElement(elements: HTMLElement[], focusedIndex: number, index: number, forward: boolean): HTMLElement {
```
riknoll, line 96: nit: you can eliminate the focusedIndex argument by using the iterative implementation above instead of the recursive one

react-common/components/share/ShareInfo.tsx
```diff
@@ -323,7 +324,7 @@ export const ShareInfo = (props: ShareInfoProps) => {
             {showSimulator && shareState !== "gifrecord" &&
                 <div className="project-share-thumbnail">
                     {thumbnailUri
-                        ? <img src={thumbnailUri} />
+                        ? <img src={thumbnailUri} alt={lf("Preview of your code running on a micro:bit")} aria-label={lf("Micro:bit simulator preview")}/>
```
riknoll, line 327: This alt-text is very micro:bit specific. You can use the boardname property of the apptarget to get the name of the hardware for the target. Or just use the more generic term "simulator" without mentioning the board at all.

react-common/components/controls/FocusList.tsx
srietkerk: This is something that I didn't fully grasp. In `handleRef` where the `focusableElements` list is populated, the query selector looks for `data-isfocusable`. Are both really needed?

react-common/styles/share/share.less
```diff
@@ -312,6 +313,11 @@
     .gif-recorder-content .thumbnail-controls {
         padding: 0 2rem;
     }
+
+    .embed.mobile-portrait-hidden {
+        color: #323130 !important;
```
riknoll, line 318: why the use of `!important` here? can it be avoided by making the selector more specific?

important makes maintaining css harder (though i know we use it a little too often in our codebase already 😆)

react-common/components/controls/FocusList.tsx
riknoll: data-isfocusable can be set to false programatically. for example, in Tree.tsx

react-common/styles/share/share.less
srietkerk: It looks like I can not have the important for the background color, but the text color needs to be important because of semantic ui stuff: 
image

react-common/components/controls/FocusList.tsx
srietkerk: Gotcha, I'll add it back.

react-common/components/util.tsx
srietkerk: I prefer using recursion for this, but would like to hear more of your thoughts if you feel strongly about implementing it with iteration.

riknoll (approved):

react-common/components/util.tsx
riknoll: the iterative implementation doesn't require passing in focusedIndex. I just also generally prefer loops to tail recursion, personally (but feel free to ignore)

PR #9990, opened 2024-04-29
Make a troubelshooting page for tutorial authoring

ganicke:
Here's is a start on a troubleshooting page for tutorial writing. Suggest other tips to include.

Closes https://github.com/microsoft/pxt-arcade/issues/6408

### DIFF
--- a/docs/writing-docs/tutorials/troubleshooting.md
+++ b/docs/writing-docs/tutorials/troubleshooting.md
@@ -1,0 +1,25 @@
+# Troubleshooting
+
+If your tutorial is not behaving properly or you receive an error message, you can check some of these common trouble situations to help resolve the problem with your custom tutorial.
+
+### ~ hint
+
+#### Tutorial format or content problems
+
+If MakeCode can't load your tutorial, you will likely see the message:
+
+```
+Please check your internet connection and check the tutorial is valid
+```
+
+This may indicate that your tutorial has a problem with it's content or it can't be located at the path you provided.
+
+### ~
+
+## Tutorial file is too large
+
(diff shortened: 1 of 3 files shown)


Jaqster (approved):
Looks like a great start!  I'm going to try and crowdsource contributions for this...

docs/writing-docs/tutorials/troubleshooting.md
kiki-lee, line 1: Some other thoughts:
- Dashes '-' in your image pack repo name need to disappear when you import the pack into a tutorial.  For instance "my-images" becomes "my images" but "my_images" stays the same. 
- If you added a "kind" the way that our compiler does, it won't work correctly in the tutorial. You need to add //% isKind  before each declaration.
- If the game screen on the tutorial seems to spin and never load, there's probably an error. Start by opening the JS console and trying to re-run the tutorial. It should give you an idea of what's making it hang.

ganicke: Thanks! I added those in.

kiki-lee (changes requested):
No spaces. That would turn it into three separate variables in the code.  It will just be `myassetpack`.

ganicke:
> No spaces. That would turn it into three separate variables in the code. It will just be `myassetpack`.

OK. I was confused cuz you mentioned earlier:

Dashes '-' in your image pack repo name need to disappear when you import the pack into a tutorial. For instance "my-images" becomes "my images" but "my_images" stays the same.

kiki-lee (approved):

PR #9992, opened 2024-04-30
Teacher Tool: Reorder catalog so copilot and "block used count times" are at the top.

thsparks:
Got feedback that we should move these to the top.

image

@Jaqster FYI


srietkerk (approved):
LGTM

PR #10003, opened 2024-05-02
Updating to Blockly v11

riknoll:
~Opening this as a draft because it relies on a pre-release version of Blockly.~

This PR upgrades us to v11 of Blockly and in doing so restores the last of the features we lost with the great blockly swap. That includes:

* The ability to duplicate shadow blocks on drag
* Updated rendering for Workspace comments
* Updated rendering for Block comments to match workspace comments

The biggest chunk of code comes from re-implementing the block comments. The bubble code is mostly the same as Blockly's with the rendering re-implemented to resemble the workspace comment rendering. Unfortunately, the classes in Blockly make extensive use of the `private` keyword so I had to copy the whole implementations just to swap the rendering.


riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/5433

jwunderl (approved):

PR #10005, opened 2024-05-06
Add partner telemetry flag

riknoll:
Adds a query parameter that lets you add a custom "partner" property to all of the telemetry items that are logged to app insights. The purpose of this feature is to let us differentiate between traffic generated by partners that link to or embed us; think of it as a manual referrer property that can be embedded in a URL. This parameter is added to all tick events.

url should look like this:

```
arcade.makecode.com/?partner=partnerName
```

Right now I am not removing the query parameter from the URL. I can do that, though, if we want.

Here is a test upload:

https://arcade.makecode.com/app/088c007a20a09fcee83221eeda1df6c9dd2a1f5b-590b47925d

### DIFF
--- a/docfiles/pxtweb/cookieCompliance.ts
+++ b/docfiles/pxtweb/cookieCompliance.ts
@@ -9,6 +9,7 @@ namespace pxt {
     let analyticsLoaded = false;
     let interactiveConsent = false;
     let isProduction = false;
+    let partnerName: string;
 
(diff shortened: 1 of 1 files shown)


thsparks (approved):
👍

abchatra (approved):

abchatra:
Lets hotfix to microbit.

PR #10007, opened 2024-05-07
Add Tags to Catalog Criteria

thsparks:
This change adds tags to criteria in the catalog. The schema is setup to support criteria having multiple tags, but for now we only adjust our display based on the first tag. This may change if we add more criteria and start using more granular tags, but it's not necessary given the current catalog.

If a criterion doesn't have any tags, it gets assigned an "Other" tag. Currently, all visible catalog criteria have tags, so the other tag is never shown.

Initially, the first tag in the catalog is expanded, but the tool remembers which tags the user expands/collapses and preserves those for the next time the catalog is opened.

I made a few adjustments to the accordion display to support this change:
1. A flag to allow multiple items to be expanded
2. The ability to specify an initial "open" item set
3. The optional ability to specify ids for each item (to support number 2)

Upload target is here, but since it's looking at production docs, everything is just placed into the "Other" category for now: https://makecode.microbit.org/app/13ea77ed2e1da12176c4c12a9907981375709c43-71c65280c6--eval

Screenshots (from localhost with updated docs)
image
image
image
image

Fixes https://github.com/microsoft/pxt-microbit/issues/5597

### DIFF
--- a/react-common/components/controls/Accordion/context.tsx
+++ b/react-common/components/controls/Accordion/context.tsx
@@ -8,11 +8,11 @@ interface AccordionState {
 const AccordionStateContext = React.createContext<AccordionState>(null);
 const AccordionDispatchContext = React.createContext<(action: Action) => void>(null);
 
-export const AccordionProvider = ({ multiExpand, children }: React.PropsWithChildren<{multiExpand?: boolean}>) => {
+export const AccordionProvider = ({ multiExpand, defaultExpandedIds, children }: React.PropsWithChildren<{multiExpand?: boolean, defaultExpandedIds?: string[]}>) => {
     const [state, dispatch] = React.useReducer(
         accordionReducer,
         {
-            expanded: [],
+            expanded: defaultExpandedIds ?? [],
             multiExpand
         }
     );
--- a/react-common/components/controls/Accordion/context.tsx
+++ b/react-common/components/controls/Accordion/context.tsx
@@ -1,16 +1,20 @@
 import * as React from "react";
 
 interface AccordionState {
-    expanded?: string;
+    multiExpand?: boolean;
+    expanded: string[];
 }
 
 const AccordionStateContext = React.createContext<AccordionState>(null);
 const AccordionDispatchContext = React.createContext<(action: Action) => void>(null);
 
-export const AccordionProvider = ({ children }: React.PropsWithChildren<{}>) => {
+export const AccordionProvider = ({ multiExpand, children }: React.PropsWithChildren<{multiExpand?: boolean}>) => {
     const [state, dispatch] = React.useReducer(
         accordionReducer,
-        {}
+        {
+            expanded: [],
+            multiExpand
+        }
     );
 
     return (
@@ -27,15 +31,27 @@ type SetExpanded = {
     id: string;
 };
 
+type RemoveExpanded = {
+    type: "REMOVE_EXPANDED";
+    id: string;
+};
+
 type ClearExpanded = {
     type: "CLEAR_EXPANDED";
 };
 
-type Action = SetExpanded | ClearExpanded;
+type Action = SetExpanded | RemoveExpanded | ClearExpanded;
 
 export const setExpanded = (id: string): SetExpanded => (
     {
         type: "SET_EXPANDED",
+        id
+    }
+);
+
+export const removeExpanded = (id: string): RemoveExpanded => (
+    {
+        type: "REMOVE_EXPANDED",
         id
     }
 );
@@ -59,7 +75,12 @@ function accordionReducer(state: AccordionState, action: Action): AccordionState
         case "SET_EXPANDED":
             return {
                 ...state,
-                expanded: action.id
+                expanded: state.multiExpand ? [...state.expanded, action.id] : [action.id]
+            };
+        case "REMOVE_EXPANDED":
+            return {
+                ...state,
+                expanded: state.expanded.filter(id => id !== action.id)
             };
         case "CLEAR_EXPANDED":
             return {
--- a/react-common/components/controls/Accordion/context.tsx
+++ b/react-common/components/controls/Accordion/context.tsx
@@ -8,14 +8,15 @@ interface AccordionState {
 const AccordionStateContext = React.createContext<AccordionState>(null);
 const AccordionDispatchContext = React.createContext<(action: Action) => void>(null);
 
-export const AccordionProvider = ({ multiExpand, defaultExpandedIds, children }: React.PropsWithChildren<{multiExpand?: boolean, defaultExpandedIds?: string[]}>) => {
-    const [state, dispatch] = React.useReducer(
-        accordionReducer,
-        {
-            expanded: defaultExpandedIds ?? [],
-            multiExpand
-        }
-    );
+export const AccordionProvider = ({
+    multiExpand,
+    defaultExpandedIds,
+    children,
+}: React.PropsWithChildren<{ multiExpand?: boolean; defaultExpandedIds?: string[] }>) => {
+    const [state, dispatch] = React.useReducer(accordionReducer, {
+        expanded: defaultExpandedIds ?? [],
+        multiExpand,
+    });
 
     return (
         <AccordionStateContext.Provider value={state}>
@@ -63,7 +64,7 @@ export const clearExpanded = (): ClearExpanded => (
 );
 
 export function useAccordionState() {
-    return React.useContext(AccordionStateContext)
+    return React.useContext(AccordionStateContext);
 }
 
 export function useAccordionDispatch() {
@@ -75,17 +76,17 @@ function accordionReducer(state: AccordionState, action: Action): AccordionState
         case "SET_EXPANDED":
             return {
                 ...state,
-                expanded: state.multiExpand ? [...state.expanded, action.id] : [action.id]
+                expanded: state.multiExpand ? [...state.expanded, action.id] : [action.id],
             };
         case "REMOVE_EXPANDED":
             return {
                 ...state,
-                expanded: state.expanded.filter(id => id !== action.id)
+                expanded: state.expanded.filter((id) => id !== action.id),
             };
         case "CLEAR_EXPANDED":
             return {
                 ...state,
-                expanded: undefined
+                expanded: undefined,
             };
     }
 }
--- a/react-common/components/controls/Accordion/context.tsx
+++ b/react-common/components/controls/Accordion/context.tsx
@@ -37,11 +37,7 @@ type RemoveExpanded = {
     id: string;
 };
 
-type ClearExpanded = {
-    type: "CLEAR_EXPANDED";
-};
-
-type Action = SetExpanded | RemoveExpanded | ClearExpanded;
+type Action = SetExpanded | RemoveExpanded;
 
 export const setExpanded = (id: string): SetExpanded => (
     {
@@ -54,12 +50,6 @@ export const removeExpanded = (id: string): RemoveExpanded => (
     {
         type: "REMOVE_EXPANDED",
         id
-    }
-);
-
-export const clearExpanded = (): ClearExpanded => (
-    {
-        type: "CLEAR_EXPANDED"
     }
 );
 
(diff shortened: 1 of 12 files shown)


teachertool/src/services/storageService.ts
```diff
@@ -192,3 +193,45 @@ export async function deleteChecklistAsync(name: string) {
...
+        expandedTags = [];
+    }
+    expandedTags.push(tag);
+    await setExpandedCatalogTags(expandedTags);
+}
+
+export async function removeExpandedCatalogTagAsync(tag: string) {
+    let expandedTags = getExpandedCatalogTags();
+    if (!expandedTags) {
+        await setExpandedCatalogTags([]);
```
srietkerk, line 229: Why is it needed that we set the expanded tags to an empty list if `getExpandedCatalogTags()` returns undefined?

teachertool/src/components/CatalogOverlay.tsx
```diff
@@ -106,34 +127,72 @@ const CatalogList: React.FC = () => {
         announceToScreenReader(lf("Added '{0}' to checklist.", getReadableCriteriaTemplate(c)));
     }
 
+    function getItemIdForTag(tag: string) {
+        return `accordion-item-${tag}`;
+    }
+
+    function onTagExpandToggled(tag: string, expanded: boolean) {
+        if (expanded) {
+            /* await */ addExandedCatalogTagAsync(tag);
```
srietkerk, line 136: If we don't need these to be awaited, can we removed the commented `await`?

teachertool/src/components/CatalogOverlay.tsx
```diff
@@ -106,34 +127,72 @@ const CatalogList: React.FC = () => {
...
+        } else {
+            /* await */ removeExpandedCatalogTagAsync(tag);
+        }
+    }
+
+    const tags = Object.keys(criteriaGroupedByTag);
+    if (tags.length === 0) {
+        return null;
```
srietkerk, line 144: To clarify, does this mean that the overlay will be empty if there were no tags defined?

teachertool/src/services/storageService.ts
```diff
@@ -192,3 +193,45 @@ export async function deleteChecklistAsync(name: string) {
...
+
+export async function setExpandedCatalogTags(tags: string[]) {
+    try {
+        setValue(EXPANDED_CATALOG_TAGS_KEY, JSON.stringify(tags));
+    } catch (e) {
+        logError(ErrorCode.localStorageWriteError, e);
+    }
+}
+
+export async function addExandedCatalogTagAsync(tag: string) {
```
srietkerk, line 217:
```suggestion
export async function addExpandedCatalogTagAsync(tag: string) {
```

teachertool/src/components/CatalogOverlay.tsx
```diff
@@ -106,34 +127,72 @@ const CatalogList: React.FC = () => {
...
+    }
+
+    let expandedTags = getExpandedCatalogTags();
+    if (!expandedTags) {
+        // If we haven't saved an expanded set, default expand the first one.
+        addExandedCatalogTagAsync(tags[0]);
```
srietkerk, line 150:
```suggestion
        addExpandedCatalogTagAsync(tags[0]);
```

react-common/components/controls/Accordion/context.tsx
```diff
@@ -27,11 +32,16 @@ type SetExpanded = {
     id: string;
 };
 
+type RemoveExpanded = {
+    type: "REMOVE_EXPANDED";
+    id: string;
+};
+
 type ClearExpanded = {
```
riknoll, line 40: is this used anymore? can probably be removed

teachertool/src/components/CatalogOverlay.tsx
```diff
@@ -106,34 +127,72 @@ const CatalogList: React.FC = () => {
...
-                    )
+                    <Accordion.Item
+                        itemId={getItemIdForTag(tag)}
+                        onExpandToggled={expanded => onTagExpandToggled(tag, expanded)}
+                        key={getItemIdForTag(tag)}
+                    >
+                        <Accordion.Header>{tag}</Accordion.Header>
+                        <Accordion.Panel>
+                            {criteriaGroupedByTag[tag].map(c => {
+                                const existingInstanceCount = teacherTool.checklist.criteria.filter(
```
srietkerk, line 167: I'm wondering if we should break out these catalog entries into its own component like `CatalogOverlayEntry` or something like that to keep the accordion area clean.

teachertool/src/components/CatalogOverlay.tsx
```diff
@@ -106,34 +127,72 @@ const CatalogList: React.FC = () => {
         announceToScreenReader(lf("Added '{0}' to checklist.", getReadableCriteriaTemplate(c)));
     }
 
+    function getItemIdForTag(tag: string) {
+        return `accordion-item-${tag}`;
+    }
+
+    function onTagExpandToggled(tag: string, expanded: boolean) {
```
riknoll, line 134: this being async seems dangerous. looking at the implementation below of these two functions, it seems like hammering on this button could cause some issues with multiple copies of the same tag being added to the state

teachertool/src/components/styling/CatalogOverlay.module.scss
```diff
@@ -77,7 +90,7 @@
                     .action-indicators {
                         position: relative;
                         padding: 0 1rem 0 0;
-                        width: 3rem;
+                        width: 3rem; // Match common-accordion-chevron
```
srietkerk, line 93: Is this comment needed?

teachertool/src/components/CatalogOverlay.tsx
```diff
@@ -106,34 +127,72 @@ const CatalogList: React.FC = () => {
...
+                        <Accordion.Panel>
+                            {criteriaGroupedByTag[tag].map(c => {
+                                const existingInstanceCount = teacherTool.checklist.criteria.filter(
+                                    i => i.catalogCriteriaId === c.id
+                                ).length;
+                                const isMaxed = c.maxCount !== undefined && existingInstanceCount >= c.maxCount;
+                                return (
+                                    c.template && (
+                                        <Button
+                                            id={`criteria_${c.id}`}
```
riknoll, line 174: is this id used anywhere?

react-common/components/controls/Accordion/context.tsx
```diff
@@ -27,11 +32,16 @@ type SetExpanded = {
...
     type: "CLEAR_EXPANDED";
 };
 
-type Action = SetExpanded | ClearExpanded;
+type Action = SetExpanded | RemoveExpanded | ClearExpanded;
```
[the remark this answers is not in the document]
thsparks: Yeah, I just thought it might be useful later on.

teachertool/src/components/CatalogOverlay.tsx
aznhassan: Do we have a linter in this repo? We usually comment this or `async` to let the linter and the reader know that we explicitly and not accidentally running this in the background.

teachertool/src/services/storageService.ts
thsparks: If someone is trying to remove a saved expanded tag but there were none saved to begin with, it's a pretty bizarre scenario to begin with, but I decided to switch it over to an empty list (to save state as "all categories collapsed") rather than undefined, which is mostly just the initial "nothing has happened yet" state. It's pretty arbitrary handling of a shouldn't-ever-happen edge case.

teachertool/src/components/styling/CatalogOverlay.module.scss
thsparks: I think it's a nice reminder if someone ever changes it, they should consider changing it there too (which may or may not look good depending on the new value, up to the engineer at that point). Could go and do a whole var for it but that seemed like overkill to me.

teachertool/src/components/CatalogOverlay.tsx
thsparks: This is a pattern we followed in other areas of the code to make it clear that we're intentionally *not* awaiting even though it's async (as opposed to an oversight). I believe it was initially a recommendation by @eanders-ms in another PR. I'm fine to remove or keep either way, but I wanted to be consistent.

teachertool/src/components/CatalogOverlay.tsx
thsparks: Yes. It'd mean there were no criteria, since every criteria (once loaded into state) should have at least one tag. Should never happen, but better this than some random error, imo.

teachertool/src/components/CatalogOverlay.tsx
```diff
@@ -106,34 +127,72 @@ const CatalogList: React.FC = () => {
...
-                            onClick={() => onItemClicked(c)}
-                            disabled={isMaxed}
-                        />
-                    )
+                    <Accordion.Item
+                        itemId={getItemIdForTag(tag)}
+                        onExpandToggled={expanded => onTagExpandToggled(tag, expanded)}
+                        key={getItemIdForTag(tag)}
+                    >
+                        <Accordion.Header>{tag}</Accordion.Header>
```
srietkerk, line 164: Should the tag get `lf`-ed here?

teachertool/src/components/CatalogOverlay.tsx
srietkerk: Gotcha! Yeah, I'm on board to keep it in to be consistent.

teachertool/src/components/CatalogOverlay.tsx
thsparks: It should be translated before we get here. We actually can't lf variables, as I recal.

teachertool/src/components/CatalogOverlay.tsx
srietkerk: Gotcha, makes sense. Maybe we should log something, too? Makes sense that it should never happen, but I feel like if there was a blank overlay and someone looked at the console and no problems were reported, it would feel very buggy. Although I'm guessing there should be other errors in the console if that were to happen?

teachertool/src/components/CatalogOverlay.tsx
srietkerk: Ah, yes, you're totally right.

riknoll: there's an rlf function for variables, which does a runtime lookup for the translation. not necessary if already translated at this point, but FYI.

this is also why you sometimes see comments like this:

```
// DO NOT REMOVE!
// lf("Loops")
```

it's just to make it so that "Loops" get's added to the json file for lookup later with rlf

react-common/components/controls/Accordion/context.tsx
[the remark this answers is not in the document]
srietkerk: I'm actually on board with keeping this in for the component's sake. I think it's a very common use case that a user wants to collapse all the expanded categories when there's too much going on, I just wanted to check in that that was the reason why it was left in.

teachertool/src/components/CatalogOverlay.tsx
thsparks: Hmm, I don't think it is, actually. I'll remove it.

teachertool/src/components/styling/CatalogOverlay.module.scss
srietkerk: Hmm, I'm not sure if that warrants a comment, personally. I feel like there is a lot of styling that is made in order to match with styling elsewhere, and leaving that up to experimentation is fine.

teachertool/src/components/CatalogOverlay.tsx
thsparks: Ah true...I'll just remove the async stuff. Most of our local storage stuff isn't async anyway.

teachertool/src/components/CatalogOverlay.tsx
thsparks: Added error log

teachertool/src/components/styling/CatalogOverlay.module.scss
thsparks: If I were the engineer making the change, I would appreciate the comment because it's the kind of thing I'd otherwise miss :) But if you feel very strongly I can remove it.

srietkerk: Nope, don't feel strongly; fine to keep it in.

srietkerk (approved):
Looks great!

PR #10008, opened 2024-05-08
Fix the help link button in the monaco flyout 

riknoll:
This is not fixing the accessibility issues that this button has, just the functionality.

Fixes https://github.com/microsoft/pxt-minecraft/issues/2524

Also changes it so that the docs open in the sidedocs and not a new tab.


riknoll:
the entire monaco flyout could really do with an accessibility pass...

thsparks (approved):
✔️

PR #10010, opened 2024-05-09
Remove the decompiler block limit

riknoll:
People run into this all the time on the forum, most recently here:

https://forum.makecode.com/t/my-code-seems-to-big-to-be-decompile-in-block/28637/4

If you have a very large blocks project, you can get stuck if you switch to text because we have a limit in place for the maximum size of a project we decompile. This limit was added many years ago in response to an issue where someone tried to decompile a gigantic script (they had copy+pasted code over and over again) and it basically hung the browser when Blockly tried to load the gigantic workspace file we generated.

I think this made sense at the time, but nowadays Blockly perf has improved and I don't think we need the restriction anymore. I was able to decompile a 9000 block program locally just now (on my admittedly nice laptop) and it was fine, it just took a while for Blockly to render the workspace.

### DIFF
--- a/pxtcompiler/emitter/decompiler.ts
+++ b/pxtcompiler/emitter/decompiler.ts
@@ -48,7 +48,6 @@ namespace ts.pxtc.decompiler {
 
     export type Comment = MultiLineComment | SingleLineComment;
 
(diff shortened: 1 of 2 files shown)


eanders-ms (approved):
yes!

PR #10014, opened 2024-05-14
Accessibility compliance: fixed focus for closed modals

srietkerk:
If a user were to close a modal with keyboard navigation, their focus would be put on the root rather than the most recent element they were focused on, which is a poor user experience. This PR fixes that and, upon closing a modal, will put focus on the DOM element that had focus before the modal was opened.

I want to give a quick thanks to @thsparks for this idea! Going with this approach greatly limited the amount of work that was needed for this fix.

EDIT: I wanted to add a quick note that the "Experiments" full-screen modal isn't performing correctly.. I want to look into this in a follow-up since the other full-screen modals are working as expected.

Follow-up: The "experiments" modal does not restore focus to the "settings" menuitem as expected because the last item that was focused upon opening the "experiments" modal is the experiments button in the "about" modal. The "about" modal is closed and thus does not have anything to focus on and thus just restores focus to the base. This is something that should be fixed, either by moving experiments or making the experiments area not a modal, but I don't think needs to be addressed in this change.

fixes https://github.com/microsoft/pxt-microbit/issues/5476

Upload target: https://makecode.microbit.org/app/02245e3b8695a39a2cf0053d656c5064581af07e-f2b0cbf1a5

### DIFF
--- a/react-common/components/controls/FocusTrap.tsx
+++ b/react-common/components/controls/FocusTrap.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { classList, nodeListToArray, findNextFocusableElement } from "../util";
+import { classList, nodeListToArray, findNextFocusableElement, isFocusable } from "../util";
 
 export interface FocusTrapProps extends React.PropsWithChildren<{}> {
     onEscape: () => void;
@@ -8,6 +8,7 @@ export interface FocusTrapProps extends React.PropsWithChildren<{}> {
     arrowKeyNavigation?: boolean;
     dontStealFocus?: boolean;
     includeOutsideTabOrder?: boolean;
+    dontRestoreFocus?: boolean;
 }
 
 export const FocusTrap = (props: FocusTrapProps) => {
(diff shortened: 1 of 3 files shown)


react-common/components/controls/FocusTrap.tsx
```diff
@@ -18,13 +19,26 @@ export const FocusTrap = (props: FocusTrapProps) => {
...
+        return () => {
+            if (!dontRestoreFocus) {
+                let toFocus = previouslyFocused.current as HTMLElement
+                while (!isFocusable(toFocus)) {
+                    toFocus = toFocus.parentElement;
+                }
+                toFocus.focus()
+            }
+        }
+    }, [previouslyFocused])
```
riknoll, line 40: you should remove this from the dependency array. also, probably need to make sure that previouslyFocused.current is actually defined.

webapp/src/sui.tsx
```diff
@@ -1212,6 +1225,18 @@ export class Modal extends data.Component<ModalProps, ModalState> {
 
     componentWillUnmount() {
         cancelAnimationFrame(this.animationRequestId);
+        if (!this.props.dontRestoreFocus) {
+            let toFocus = this.state.previouslyFocused as HTMLElement;
+            while (toFocus && !this.isFocusable(toFocus)) {
```
riknoll, line 1230: You could refactor both this parent-finding logic and isFocusable into utils.tsx in react-common and import it here and in the focus trap

riknoll (approved):

webapp/src/sui.tsx
srietkerk: Okay, cool. I wasn't certain about putting importing something from react-common into here, so that's why I made the duplicate in `sui.tsx`, I'll move and import.

webapp/src/sui.tsx
```diff
@@ -1212,6 +1216,10 @@ export class Modal extends data.Component<ModalProps, ModalState> {
 
     componentWillUnmount() {
         cancelAnimationFrame(this.animationRequestId);
+        if (!this.props.dontRestoreFocus) {
+            let toFocus = this.state.previouslyFocused as HTMLElement;
+            focusLastActive(toFocus);
```
riknoll, line 1221: could be undefined here

PR #10019, opened 2024-05-22
Accessibility compliance: arcade color changes

srietkerk:
Related to the changes with this PR: https://github.com/microsoft/pxt-arcade/pull/6425
There are some parts of the colors that needed to be changed in pxt rather than just arcade. The asset editor and skillmap styling is defined in pxt, so I needed to change colors here for those things.

### DIFF
--- a/webapp/src/codecard.tsx
+++ b/webapp/src/codecard.tsx
@@ -136,7 +136,7 @@ export class CodeCardView extends data.Component<CodeCardProps, CodeCardState> {
                     }
                 </div> : undefined}
             {card.time ? <div className="meta">
-                {card.tutorialLength ? <span className={`ui tutorial-progress ${tutorialDone ? "green" : "orange"} left floated label`}><i className={`${tutorialDone ? "trophy" : "circle"} icon`}></i>&nbsp;{lf("{0}/{1}", (card.tutorialStep || 0) + 1, card.tutorialLength)}</span> : undefined}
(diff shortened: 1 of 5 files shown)


webapp/src/codecard.tsx
```diff
@@ -136,7 +136,7 @@ export class CodeCardView extends data.Component<CodeCardProps, CodeCardState> {
                     }
                 </div> : undefined}
             {card.time ? <div className="meta">
-                {card.tutorialLength ? <span className={`ui tutorial-progress ${tutorialDone ? "green" : "orange"} left floated label`}><i className={`${tutorialDone ? "trophy" : "circle"} icon`}></i>&nbsp;{lf("{0}/{1}", (card.tutorialStep || 0) + 1, card.tutorialLength)}</span> : undefined}
+                {card.tutorialLength ? <span className={`ui tutorial-progress ${tutorialDone ? "green" : "purple"} left floated label`}><i className={`${tutorialDone ? "trophy" : "circle"} icon`}></i>&nbsp;{lf("{0}/{1}", (card.tutorialStep || 0) + 1, card.tutorialLength)}</span> : undefined}
```
thsparks, line 139: I'm trying to figure out where this component is. A little surprised we had it hard-coded to orange before and not theme-specific. Is it only on skillmaps?

thsparks (approved):
One question, I don't think it's blocking

webapp/src/codecard.tsx
srietkerk: This is the progress indicator in the cards on the home page. 
tutorial-progress

I was also shocked that this was hard coded to orange. It was just handled in micro:bit to have a style override for the .orange class and set the color to purple. It looks like Minecraft does something a bit different and doesn't rely on the class being orange to set the style. I looked into accessing the "secondaryColor" less variable and then plugging the calculated color into the class here, but I couldn't find a nice way to do it, so I just put purple here instead.

PR #10020, opened 2024-05-23
Add warning when trying to carry over code without a template code block

riknoll:
Fixes https://github.com/microsoft/pxt-arcade/issues/6386

When you try to carry code between skillmap steps, we refuse to import the old code if the destination tutorial doesn't have a template code block. This behavior was confusing to content authors, so adding an explicit warning to the console to hopefully make everything a little more discoverable. Check out the forum thread in the linked issue for more info.


thsparks (approved):

kiki-lee:
Great find!

PR #10024, opened 2024-05-31
Accessibility compliance: inline blocks styling updates

srietkerk:
Because not all of our block colors are accessible, we've come up with an alternative solution for displaying inline blocks. With this update, we are using underlines and namespace icons to depict the blocks. For block highlights in tutorials and docs that are not clickable, we just use an underline. 

Fixes https://github.com/microsoft/pxt-arcade/issues/4870

**IMPORTANT** 
This will also affect micro:bit, so if, after testing, users respond positively to this, we might need to go through more steps to get approval from the micro:bit team.

Upload targets:
micro:bit ~~https://makecode.microbit.org/app/5ad32a6d20c70ec72517ad979b75b21a90250a27-45f381e4aa#~~
https://makecode.microbit.org/app/b01d093aa836e1a318f38be16d589f5d608836f9-3fe55a0f54#
arcade ~~https://arcade.makecode.com/app/d601caf565ccfe8cfae4809ba8436fe9749c2316-82ab3f31e0#~~
https://arcade.makecode.com/app/1403937f656f6e50637c8a8cb7cca026690a6ca4-14c598d835#editor

Make sure to look at both tutorials and docs.

### DIFF
--- a/theme/docs.less
+++ b/theme/docs.less
@@ -378,11 +378,16 @@
     }
 
     span.docs.inlineblock {
-        border-bottom: 2px solid var(--inline-namespace-color);
+        border-bottom: 3px solid var(--inline-namespace-color);
         border-radius: 0;
         color: black;
         font-weight: 600;
         background-color: transparent;
+        padding-bottom: 0.1rem;
+        
+        i {
+            margin-right: 6px;
+        }
     }
 }
 
--- a/theme/docs.less
+++ b/theme/docs.less
@@ -376,6 +376,14 @@
             padding: 0;
         }
     }
+
+    span.docs.inlineblock {
+        border-bottom: 2px solid var(--inline-namespace-color);
+        border-radius: 0;
+        color: black;
+        font-weight: 600;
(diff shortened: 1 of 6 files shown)


riknoll (changes requested):

theme/docs.less
```diff
@@ -376,6 +376,14 @@
...
         }
     }
+
+    span.docs.inlineblock {
+        border-bottom: 2px solid var(--inline-namespace-color);
+        border-radius: 0;
+        color: black;
+        font-weight: 600;
+        background-color: transparent;
+    }
```
riknoll, line 386:
```suggestion
    span.docs.inlineblock {
        border-bottom: 3px solid var(--inline-namespace-color);
        border-radius: 0;
        color: black;
        font-weight: 600;
        background-color: transparent;
        padding-bottom: 0.1rem;
        
        i {
            margin-right: 6px;
        }
    }
```

riknoll (approved):

Issue #8970, opened 2022-07-19
Adding a Variable Start Value in a for Loop


THEb0nny:
Please add the ability to specify the starting value of the for loop.

![image](https://user-images.githubusercontent.com/13646226/179748011-667d72d0-0ca9-460f-9e56-6a573d6d147e.png)

THEb0nny:
I would ask to add another loop iteration index step. There is a step in the standard google blockly.

THEb0nny:
@riknoll Please make an adequate for loop, which was originally in blockly. With initial value and step value. Makecode spoils the impression of programming in blocks.
![image](https://user-images.githubusercontent.com/13646226/227879757-897b87a8-c062-4abf-871f-585050e4913c.png)

THEb0nny:
![image](https://user-images.githubusercontent.com/13646226/228554292-d8b9736a-6e47-4179-ab77-ed30a96f9926.png)

PR #10025, opened 2024-06-02
Add start value in "for" loop

sae220:
https://github.com/microsoft/pxt/issues/8970
https://github.com/microsoft/pxt-microbit/issues/3837

### DIFF
--- a/pxtblocks/builtins/loops.ts
+++ b/pxtblocks/builtins/loops.ts
@@ -58,6 +58,11 @@ export function initLoops() {
                     },
                     {
                         "type": "input_value",
+                        "name": "FROM",
+                        "check": "Number"
+                    },
+                    {
+                        "type": "input_value",
                         "name": "TO",
                         "check": "Number"
                     }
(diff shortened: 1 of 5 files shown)


THEb0nny:
We need such an update!

sae220:
@riknoll 
Could you review this if you have time?

riknoll:
@sae220 I don't think we can merge this PR. it would break too many existing screenshots and documentation that uses the old for-loop. we've done a lot of iterating over the for-loop internally over the years and we've always run into walls with feedback from either partners or teachers when trying to update it.

riknoll:
@sae220 that might be doable, but we'd probably want to put it behind an experiment flag to start. do you know how to add something to the experiments dialog?

THEb0nny:
@sae220 change does not support setting the step?

sae220:
@riknoll 
I don't know but I'll refer to https://github.com/microsoft/pxt/pull/6809. Is it OK?

sae220:
@THEb0nny 
This pull request doesn't support the step but I'll make a new pull request support it.

PR #10038, opened 2024-07-09
Dont break decompilation when blocks file is empty

riknoll:
This was hit by a forum user here:

https://forum.makecode.com/t/oops-something-went-wrong-trying-to-convert-your-code/1159/6

I'm not sure how they got their project into the state of having an empty blocks file, but whatever the case we shouldn't refuse to decompile if that happens.

The issue was that we were passing the empty string to the xml parser which caused an exception. Fix is to just check if the file is empty before parsing


srietkerk (approved):
Nice!

PR #10039, opened 2024-07-10
Fall back to cache when fetching index.html with a query param

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/5687

When we load the serviceworker for the first time we cache index.html, but that doesn't cover requests that also have a query parameter in the url.

This PR caches responses when a query parameter is present and uses those as a fallback when we're offline. Note that this is only used as a fallback; we will still fetch live whenever possible.

At some point we should probably inventory every possible query parameter we pass on requests to index.html; a lot of them could probably be moved into the URL hash so that they don't bust the service worker cache. For those that we have already shipped (e.g. minecraft) we can create an allowlist that lets us serve the cached index.html sans query parameters if the cache misses. I'll create an issue for this in the pxt-minecraft repo

test upload: https://makecode.microbit.org/app/0490e8531c8c225154ffe2a11e81ea23f7004ef4-dc66e1bc83

### DIFF
--- a/webapp/src/serviceworker.ts
+++ b/webapp/src/serviceworker.ts
@@ -26,25 +26,26 @@ enum DisconnectResponse {
     Waiting,
     TimedOut
 }
+// Empty string for released, otherwise contains the ref or version path
+const ref = `@relprefix@`.replace("---", "").replace(/^\//, "");
+const pageUrl =  `@targetUrl@/` + ref;
+
+// pxtRelId is replaced with the commit hash for this release
+const refCacheName = "makecode;" + ref + ";@pxtRelId@";
 
 initWebappServiceWorker();
 initWebUSB();
 
 function initWebappServiceWorker() {
-    // Empty string for released, otherwise contains the ref or version path
-    const ref = `@relprefix@`.replace("---", "").replace(/^\//, "");
 
     // We don't do offline for version paths, only named releases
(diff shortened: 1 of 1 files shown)


webapp/src/serviceworker.ts
```diff
@@ -197,12 +198,49 @@ function initWebappServiceWorker() {
...
+
+            // If this is just the main page with a query parameter, attempt
+            // to fetch from the network and fall back to the cache if it fails
+            if (path.charAt(0) === "?") {
+                let response: Response;
+                try {
+                    response = await fetch(ev.request);
+
+                    // Store this response in the cache in case the user tries
+                    // to visit this same query param offline
```
srietkerk, line 217: Just a clarifying question. This just means that because we are caching the site with query parameters, the user _can_ get to the site with those configurations, right? Like if the user removes the query param they will still be able to use the site without those added features from the query parameters because our base targets work offline?

srietkerk: Oh this might be the answer already that you had in the PR comment:
>This PR caches responses when a query parameter is present and uses those as a fallback when we're offline. Note that this is only used as a fallback; we will still fetch live whenever possible.

riknoll: yup! if you look at the init function above, the first thing it caches is the main index.html with no query parameters

PR #10041, opened 2024-07-15
Kiosk: user-added games can now have high scores

srietkerk:
Because it was tricky to know what type of scoring a user-added game had, we didn't give user-added games the option for high score entries on the creation of the kiosk. When a new game was added, we would set the "highScoreMode" to "None" and never change it. There are two important pieces that needed to happen in order to accept user's games' high scores:

1. Modify the game over block's event that gets recognized by the kiosk to include the scoring type. This change is made in common packages. We utilize this "scoring type" data in kiosk to sort the high score lists properly.
2. Update the game's scores and scoring type on the kiosk if the game supplies it upon playing said game. This is to allow for user-added games to be sorted from lowest scores at the top or highest scores at the top.


jwunderl (approved):

abchatra:
Nice change.

PR #10048, opened 2024-07-23
Kiosk: now using the config run url if it exists

srietkerk:
This change to kiosk is to make sure that we are using the correct run url in our different environments. For example, in stable, we should be using stable in order to test the changes that we're trying to ship or change. We can make sure that's the case by using the webConfig's run url. However, when testing locally, webConfig was null. While I am already specializing local host, I also wanted to keep that case in mind for the live/built scenarios, so I'm keeping the hard-coded "PlayUrlRoot" as a fall back just in case webConfig is null.

Edit: Here's an upload target (this upload target includes some console logs for testing but those console logs are now removed): https://arcade.makecode.com/app/b864d38cd277079aa0d82d37340a2a5c08132a96-51ba117ade--kiosk. You can see in those logs and the network request for when you play a game that the run url being used is according to the environment, which, in this case is the upload target url.

### DIFF
--- a/kiosk/src/Components/PlayingGame.tsx
+++ b/kiosk/src/Components/PlayingGame.tsx
@@ -17,9 +17,7 @@ export default function PlayingGame() {
     }, [kiosk.launchedGameId]);
 
     const makeRunUrl = () => {
-        return pxt.webConfig?.runUrl
-        ? "https://arcade.makecode.com" + pxt.webConfig.runUrl
-        : configData.PlayUrlRoot;
+        return pxt.webConfig?.runUrl ?? configData.PlayUrlRoot;
     }
 
     // Handle Back and Start button presses
--- a/kiosk/src/Components/PlayingGame.tsx
+++ b/kiosk/src/Components/PlayingGame.tsx
@@ -16,6 +16,12 @@ export default function PlayingGame() {
         return getLaunchedGame();
     }, [kiosk.launchedGameId]);
 
+    const makeRunUrl = () => {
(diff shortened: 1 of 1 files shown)


kimprice (approved):
LGTM!

kiosk/src/Components/PlayingGame.tsx
```diff
@@ -16,6 +16,12 @@ export default function PlayingGame() {
         return getLaunchedGame();
     }, [kiosk.launchedGameId]);
 
+    const makeRunUrl = () => {
+        return pxt.webConfig?.runUrl
+        ? "https://arcade.makecode.com" + pxt.webConfig.runUrl
```
jwunderl, line 21: this will make it so staging still points at prod sim, which isn't ideal -- if leaving it as just a relative url (that is, just using .runUrl with no host domain) works that'd be great, otherwise `window.location.origin` instead of `https://arcade.makecode.com` would be good as well

srietkerk: So, I've been testing it with just using `pxt.webConfig.runUrl`, and it looks like it works. I'm not sure I understand why, though. `pxt.Webconfig.runUrl` only has the end path of the url, not the domain. Does the browser just resolve and say "just use the same root/domain if one isn't explicitly supplied"?

jwunderl: (discussed in standup but basically answer is 'yes' / paths get treated as relative to origin)

jwunderl (approved):

PR #10049, opened 2024-07-24
Fix all critical vulnerabilities

riknoll:
I had to manually fix three dependencies in pxt-core:
* `mocha` - confirmed that tests still run
* `@vusion/webfonts-generator` - non-breaking change
* `crypto-js` - doesn't seem like a breaking change

I was able to fix all of the sub-webapps by just running `npm audit fix` (no breaking changes)

I only tested by doing a local serve of the main webapp for sanity's sake. If anyone has concerns about the sub-webapps we could test those too


multiplayer/src/services/gameClient.ts
kimprice, line 1: How does this change relate to fixing vulnerabilities?

riknoll: I think one of the updated dependencies had an unmarked breaking change in their typescript types. had to make this change to fix the build

kimprice (approved):

abchatra (approved):

jwunderl (approved):

PR #10054, opened 2024-07-26
Add support for different codal message bus listener flags

riknoll:
This adds simulator support for the various CODAL message bus listener modes. These modes are:

* **MESSAGE_BUS_LISTENER_QUEUE_IF_BUSY** - the default mode that queues events if an event handler is busy. this has been the only supported mode in our simulator until now
* **MESSAGE_BUS_LISTENER_DROP_IF_BUSY** - events are dropped if fired while the event handler is already executing
* **MESSAGE_BUS_LISTENER_REENTRANT** - events are always fired immediately, even if the event handler is already executing
* **MESSAGE_BUS_LISTENER_IMMEDIATE** - this is used by the scheduler and throws an error if used elsewhere

this is part of the fix for https://github.com/microsoft/pxt-microbit/issues/5709, i'll have two more prs (one in pxt-common-packages and one in pxt-microbit)

### DIFF
--- a/pxtsim/runtime.ts
+++ b/pxtsim/runtime.ts
@@ -576,7 +576,7 @@ namespace pxsim {
         constructor(public handler: RefAction, public flags: number) {}
 
         async runAsync(eventValue: EventIDType, runtime: Runtime, valueToArgs?: EventValueToActionArgs) {
-            // The default behavior can technically be configujred in codal, but we always set it to queue if busy
+            // The default behavior can technically be configured in codal, but we always set it to queue if busy
             const flags = this.flags || MessageListenerFlags.MESSAGE_BUS_LISTENER_QUEUE_IF_BUSY;
 
             if (flags === MessageListenerFlags.MESSAGE_BUS_LISTENER_IMMEDIATE) {
--- a/pxtsim/runtime.ts
+++ b/pxtsim/runtime.ts
@@ -3,6 +3,14 @@
 namespace pxsim {
     const MIN_MESSAGE_WAIT_MS = 200;
     let tracePauseMs = 0;
+
+    enum MessageListenerFlags {
+        MESSAGE_BUS_LISTENER_REENTRANT = 8,
+        MESSAGE_BUS_LISTENER_QUEUE_IF_BUSY = 16,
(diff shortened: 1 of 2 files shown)


pxtsim/runtime.ts
```diff
@@ -563,13 +571,43 @@ namespace pxsim {
 
     export type EventIDType = number | string;
 
+    class EventHandler {
+        private busy = 0;
+        constructor(public handler: RefAction, public flags: number) {}
```
srietkerk, line 576: This is a bit pedantic, but wondering if the `flags` parameter will be able to pass on multiple flags or just one at a time since it has type number?

pxtsim/runtime.ts
```diff
@@ -3,6 +3,14 @@
 namespace pxsim {
     const MIN_MESSAGE_WAIT_MS = 200;
     let tracePauseMs = 0;
+
+    enum MessageListenerFlags {
```
srietkerk, line 7: Just wanted to put this here for reference: https://github.com/lancaster-university/codal-core/blob/509086cc8590465041b15493ab52b56e7071c110/inc/core/CodalListener.h#L33-L41

Also, just wanted to know why we're not including all of the defined flags here?

pxtsim/runtime.ts
```diff
@@ -563,13 +571,43 @@ namespace pxsim {
 
     export type EventIDType = number | string;
 
+    class EventHandler {
+        private busy = 0;
+        constructor(public handler: RefAction, public flags: number) {}
+
+        async runAsync(eventValue: EventIDType, runtime: Runtime, valueToArgs?: EventValueToActionArgs) {
+            // The default behavior can technically be configujred in codal, but we always set it to queue if busy
```
srietkerk, line 579:
```suggestion
            // The default behavior can technically be configured in codal, but we always set it to queue if busy
```

pxtsim/runtime.ts
riknoll: those other ones are internal

pxtsim/runtime.ts
riknoll: i'm matching the codal naming here, but actually numbers can hold up to 32 flags (one for each bit)

jwunderl (approved):

PR #10056, opened 2024-07-30
Use github access token when running docker locally

riknoll:
One annoying thing about using our CLI to build locally w/ docker images is that yotta is constantly getting throttled by GitHub. We even have a GITHUB_ACCESS_TOKEN environment variable that we advertise as something you can configure, but it only gets used by PXT and not yotta which is a bummer.

Well, not anymore! This PR passes that GITHUB_ACCESS_TOKEN variable to the docker container and runs a script to setup the local `~/.yotta/config.json` and `~/.netrc` files to prevent throttling. This token doesn't actually need to have any permissions at all, GitHub just wants to use it for authentication.

Also, this PR fixes the `--local` flag which I had accidentally disabled in my last pr.

Completely unrelated to this, I also converted a function in `cli.ts` to use async await because it was bothering me.

### DIFF
--- a/cli/buildengine.ts
+++ b/cli/buildengine.ts
@@ -387,14 +387,14 @@ function runDockerAsync(args: string[], flags?: string[]) {
         if (process.platform == "darwin")
             mountArg += ":delegated"
 
-        let fullArgs = ["--rm", "-v", mountArg, "-w", "/src"].concat(dargs).concat([cs.dockerImage]).concat(args);
+        let fullArgs = ["--rm", "-v", mountArg, "-w", "/src", ...dargs, cs.dockerImage, ...args];
         if (flags) {
-            fullArgs = flags.concat(fullArgs)
+            fullArgs = [...flags, ...fullArgs];
         }
 
         return nodeutil.spawnAsync({
             cmd: "docker",
-            args: ["run"].concat(fullArgs),
+            args: ["run", ...fullArgs],
             cwd: thisBuild.buildPath
         })
     }
--- a/cli/buildengine.ts
+++ b/cli/buildengine.ts
@@ -58,10 +58,10 @@ export const buildEngines: Map<BuildEngine> = {
 
     dockeryotta: {
         id: "dockeryotta",
-        updateEngineAsync: () => runDockerAsync(["yotta", "update"]),
-        buildAsync: () => runDockerAsync(["yotta", "build"]),
+        updateEngineAsync: () => runDockerYottaAsync(["yotta", "update"]),
+        buildAsync: () => runDockerYottaAsync(["yotta", "build"]),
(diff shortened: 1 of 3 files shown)


cli/buildengine.ts
```diff
@@ -387,14 +387,29 @@ function runDockerAsync(args: string[]) {
         if (process.platform == "darwin")
             mountArg += ":delegated"
 
+        let fullArgs = ["--rm", "-v", mountArg, "-w", "/src"].concat(dargs).concat([cs.dockerImage]).concat(args);
```
srietkerk, line 390: All of these `concat`s look a little funky to me. Can we use spread syntax here instead?

cli/buildengine.ts
```diff
@@ -387,14 +387,29 @@ function runDockerAsync(args: string[]) {
...
+        if (flags) {
+            fullArgs = flags.concat(fullArgs)
+        }
+
         return nodeutil.spawnAsync({
             cmd: "docker",
-            args: ["run", "--rm", "-v", mountArg, "-w", "/src"].concat(dargs).concat([cs.dockerImage]).concat(args),
+            args: ["run"].concat(fullArgs),
```
srietkerk, line 397: Why wouldn't we just put this in `fullArgs`?

cli/buildengine.ts
```diff
@@ -58,10 +58,10 @@ export const buildEngines: Map<BuildEngine> = {
 
     dockeryotta: {
         id: "dockeryotta",
-        updateEngineAsync: () => runDockerAsync(["yotta", "update"]),
-        buildAsync: () => runDockerAsync(["yotta", "build"]),
+        updateEngineAsync: () => runDockerYottaAsync(["yotta", "update"]),
+        buildAsync: () => runDockerYottaAsync(["yotta", "build"]),
```
srietkerk, line 62: If we are special casing Yotta, do we need to include "yotta" in the args?

cli/buildengine.ts
riknoll: because of the prepending that happens with the flags parameter. run always needs to come first

srietkerk: Ah, makes sense.

cli/buildengine.ts
riknoll: I think it's easier to understand with yotta there since that's the command you would be running if you were doing all this manually.

cli/buildengine.ts
riknoll: sure

cli/buildengine.ts
srietkerk: Makes sense!

srietkerk (approved):

PR #10058, opened 2024-07-30
Attempt to read the DAL values for sound effect editor effect constants

riknoll:
Part of the fix for https://github.com/microsoft/pxt-microbit/issues/5569

Reads the DAL constants for the default effect parameters if possible.

### DIFF
--- a/webapp/src/components/soundEffectEditor/soundUtil.ts
+++ b/webapp/src/components/soundEffectEditor/soundUtil.ts
@@ -26,18 +26,18 @@ export function soundToCodalSound(sound: pxt.assets.Sound): pxsim.codal.music.So
     switch (sound.effect) {
         case "vibrato":
             codalSound.fx = pxsim.codal.music.Effect.Vibrato;
-            codalSound.fxnSteps = 512;
-            codalSound.fxParam = 2;
+            codalSound.fxnSteps = readDalConstant("SFX_DEFAULT_VIBRATO_STEPS", 512);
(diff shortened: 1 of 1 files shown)


webapp/src/components/soundEffectEditor/soundUtil.ts
```diff
@@ -289,4 +289,19 @@ export function getMixerGallerySounds(): SoundGalleryItem[] {
...
+    const file = pxt.appTarget.bundledpkgs?.["core"]?.["dal.d.ts"];
+
+    if (!file) return defaultValue;
+
+    const regEx = new RegExp(`${name}\\s*=\\s*(\\d+),`);
+
+    const match = regEx.exec(file);
+
+    if (match) {
+        return parseInt(match[1]);
```
thsparks, line 304: Should we protect against non-numeric strings here and return defaultValue?

thsparks (approved):
One nit, not major

webapp/src/components/soundEffectEditor/soundUtil.ts
thsparks: Oh nvm, I guess regex does this already

thsparks (approved):

webapp/src/components/soundEffectEditor/soundUtil.ts
riknoll: yeah, also these are always numbers

PR #10059, opened 2024-07-31
Kiosk: make sure exit condition for low score as best score is correctly handled

srietkerk:
A quick fix for a bug that evaded me until today for user high scores. Whether we would exit to entering a high score or just exit to game over had a condition that was reliant on the high score mode being high to low. If all the possible high scores slots were filled for a low to high score game, and you got a lower score than the last value, the game would exit to game over instead of enter a high score. 

I've fixed this now with having exitToHighScore use the high score mode to determine if the score has to be lower or higher than the last recorded score.

### DIFF
--- a/kiosk/src/Transforms/exitToEnterHighScore.ts
+++ b/kiosk/src/Transforms/exitToEnterHighScore.ts
@@ -3,7 +3,7 @@ import configData from "../config.json";
 import { KioskState } from "../Types";
 import { exitGame } from "./exitGame";
 
-export function exitToEnterHighScore(): void {
+export function exitToEnterHighScore(highScoreMode: string): void {
     const { state } = stateAndDispatch();
 
     if (!state.mostRecentScores?.length) {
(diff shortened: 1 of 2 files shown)


kiosk/src/Transforms/exitToEnterHighScore.ts
```diff
@@ -20,11 +20,12 @@ export function exitToEnterHighScore(): void {
     launchedGameHighs = launchedGameHighs || [];
     const currentHighScore = state.mostRecentScores[0];
```
kimprice, line 21: Is this getting the best score? Would you want to be getting the third-place score (or however many places are listed)?

srietkerk: This is getting the score that was sent from the simulator from the state, so it doesn't know what place the score will go yet.

kiosk/src/Transforms/gameOver.ts
```diff
@@ -27,7 +27,7 @@ export function gameOver(skipHighScore?: boolean): void {
             selectedGame?.highScoreMode !== "None" &&
             state.mostRecentScores?.length
         ) {
-            exitToEnterHighScore();
+            exitToEnterHighScore(selectedGame?.highScoreMode!);
```
thsparks, line 30: hmm, if `selectedGame` is undefined here, this would cause an issue. I'm not sure what would cause it to end up null, but should we just have a check for that case above and handle it there, rather than having this optional chaining followed by a non-null assertion, which seems sort of contradictory.

srietkerk: Yeah, that's a good idea. This was something I was thinking about yesterday as well, but I figured since we weren't doing the explicit check before that it would be fine still, but it's definitely better to be safer here. And it's very true that it will be cleaner to not have to do the optional chaining and assertion business here.

thsparks: Before, it looks like we didn't rely on selectedGame to actually provide any required information, but with this change we need highScoreMode to be defined. I'd recommend just putting a `selectedGame !== undefined` before `selectedGame?.highScoreMode !== "None"` and removing the `?` from the selectedGame references following that. That way, it'll just default to exitGame when selectedGame is undefined.

thsparks (approved):
👍

PR #10065, opened 2024-08-01
Teacher Tool: Require Auth

thsparks:
Fixes https://github.com/microsoft/pxt-microbit/issues/5731

This change makes it so that users must be signed in to use the teacher tool. If they are not signed in, the sign-in modal will open automatically.

A lot of the code here is based on code in multiplayer and skillmap. In effect, we're just adding a userProfile item to state, then we add an authClient to handle signing in and out, which sets the profile appropriately. The header bar has been changed to display the usual profile icon and dropdown.

We don't actually do anything with the profile yet, but it'd be nice to save checklists to the account at some point. Out of scope for this change, though.

Upload Target: https://makecode.microbit.org/app/7d1462ee65b490779c5c30575f4310b00316e379-55b853f737--eval


srietkerk:
This might be a better question for the managers, but I just wanted to ask if all of the teacher tool is always going to be behind auth or will we limit it in the future? I just wanted to ask to make sure that the infrastructure that we put in place here can be moved around in the case that we want to shift it to only require it for teacher-only features in the future.

thsparks:
> This might be a better question for the managers, but I just wanted to ask if all of the teacher tool is always going to be behind auth or will we limit it in the future? I just wanted to ask to make sure that the infrastructure that we put in place here can be moved around in the case that we want to shift it to only require it for teacher-only features in the future.

@abchatra  can confirm, but I suspect certain features (like copilot) would always be behind an auth wall. Others don't have to be, but I think it's simpler now to just say it's fully required to use the tool.

From an implementation perspective, it would be easy to filter things out more granularly if we decide to go that way. But for the initial "beta", I think it's best to start with just a blanket requirement.

teachertool/src/App.tsx
```diff
@@ -35,28 +38,37 @@ export const App = () => {
...
+                await authClient.authCheckAsync()
+            } catch (e) {
+                // Log error but continue
+                // Don't include actual error in error log in case PII
+                logError(ErrorCode.authCheckFailed);
+                logDebug("Auth check failed details", e);
+            }
+            setAuthCheckComplete(true)
+        }
+        checkAuth();
```
srietkerk, line 60: Should this be awaited?

teachertool/src/App.tsx
```diff
@@ -35,28 +38,37 @@ export const App = () => {
...
-
                 setInited(true);
                 logDebug("App initialized");
             });
         }
     }, [ready, inited]);
 
-    return !inited ? (
+    useEffect(() => {
+        async function checkAuth() {
```
srietkerk, line 48:
```suggestion
        async function checkAuthAsync() {
```

srietkerk: Oh, did you want to keep this without the `async` since you have the `async` one in the auth client?

teachertool/src/components/HeaderBar.tsx
```diff
@@ -73,20 +77,66 @@ export const HeaderBar: React.FC<HeaderBarProps> = () => {
         );
     };
 
-    const onHomeClicked = () => {
```
srietkerk, line 76: Do we want to remove this?

teachertool/src/components/HeaderBar.tsx
```diff
@@ -73,20 +77,66 @@ export const HeaderBar: React.FC<HeaderBarProps> = () => {
...
-        pxt.tickEvent(Ticks.HomeLink);
+    function avatarPicUrl(): string | undefined {
+        return teacherTool.userProfile?.idp?.pictureUrl ?? teacherTool.userProfile?.idp?.picture?.dataUrl;
```
srietkerk, line 81: There was a bug fix that I did for the webapp a while back where I found that dataUrl was no longer giving a photo for Mcrosoft auth. I'm not sure if this still applies, though, because on the upload target I logged in with my Microsoft account and it populated fine.. Not sure what's up with all that, but I wanted to drop this here just in case. https://github.com/microsoft/pxt/pull/9956

teachertool/src/components/SignInModal.tsx
```diff
@@ -0,0 +1,25 @@
+import { useContext, } from "react";
+import { AppStateContext } from "../state/appStateContext";
+import { SignInModal as RcSignInModal } from "react-common/components/profile/SignInModal";
```
srietkerk, line 3: nit: also capitalize the C to better show that it's an abbreviation

teachertool/src/App.tsx
thsparks: useEffects can't be async, so we can't await here, but the important thing is just to kick it off.

teachertool/src/components/HeaderBar.tsx
thsparks: Yeah, we're not using it anymore. We can add it back if we end up needing the functionality again, but since I'm not confident that we will, I didn't want to have the unused code sitting here.

thsparks: If you mean the home button in general, I thought it looked better (and more consistent with out other pages) without it.

teachertool/src/App.tsx
thsparks: Either way is fine, as far as I'm concerned. It's only defined in the context of this useEffect anyway. I'll update it for consistency.

teachertool/src/components/HeaderBar.tsx
srietkerk: Hmm, yeah, that's fair. I was thinking it might be nice to have it there for a teacher scenario. Like "oh, I wanted to look at this one thing in the tutorial, let me go back to the home page." However, since we don't have an easy way to _get_ to the eval tool from the home page I can see how it wouldn't be needed.

srietkerk (approved):
Looks great!

jwunderl (approved):

PR #10066, opened 2024-08-01
Add option to have an unset gridpicker dropdown width

riknoll:
Part of the fix for https://github.com/microsoft/pxt-microbit/issues/5730

In live, that width parameter was just being ignored by the browser. I have no idea how that works (it was being set on the element, just had no effect), but for the pins dropdown we want it to just expand to wrap the whole list anyhow.

Also fixing the font/size of the menu items.

### DIFF
--- a/pxtblocks/fields/field_gridpicker.ts
+++ b/pxtblocks/fields/field_gridpicker.ts
@@ -60,7 +60,7 @@ export class FieldGridPicker extends Blockly.FieldDropdown implements FieldCusto
 
         this.columns_ = parseInt(options.columns) || 4;
         this.maxRows_ = parseInt(options.maxRows) || 0;
-        this.width_ = parseInt(options.width) || 200;
+        this.width_ = parseInt(options.width) || undefined;
 
         this.backgroundColour_ = parseColour(options.colour);
         this.borderColour_ = pxt.toolbox.fadeColor(this.backgroundColour_, 0.4, false);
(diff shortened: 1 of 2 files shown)


pxtblocks/fields/field_gridpicker.ts
```diff
@@ -60,7 +60,7 @@ export class FieldGridPicker extends Blockly.FieldDropdown implements FieldCusto
 
         this.columns_ = parseInt(options.columns) || 4;
         this.maxRows_ = parseInt(options.maxRows) || 0;
-        this.width_ = parseInt(options.width) || 200;
+        this.width_ = parseInt(options.width) || undefined;
```
srietkerk, line 63: Do you know if there is any scenario where we should have the default be 200 if there isn't a width defined? Is it possible to have the width not be defined but there be elements in the gridpicker and thus introduce weird behavior?

theme/fieldeditors.less
```diff
@@ -143,6 +143,8 @@
 
 .blocklyWidgetDiv .blocklyGridPickerMenu .goog-menuitem-content {
     color: #fff;
+    font-size: 13px;
```
thsparks, line 146: 13px seems quite small. Does it pass a11y testing?

thsparks (approved):

theme/fieldeditors.less
riknoll: none of our blocks pass a11y testing, sadly! this is what we have in live

jwunderl (approved):

PR #10067, opened 2024-08-05
Teacher Tool: Add Optional Feedback on Criteria

thsparks:
Fixes https://github.com/microsoft/pxt-microbit/issues/5742

This change adds an optional flag on catalog criteria which indicates that we should request feedback from the user on the tool's response. It's currently only set to true on the copilot criteria. When feedback is provided, it simply sends a tick event (collecting actual prompt and response data will require a more thorough privacy review). 

There's one slightly odd note about this change: it felt weird to lock a user into their first feedback selection they make, so I've allowed the response to be changed or cleared, but we can't "take back" a tick event. I considered adding a delay before sending the event, but that seemed flaky, like there could always be scenarios where the delay wasn't long enough or the user closed the page before it fired, so instead, each feedback event includes the previous value as well as the new one. We'll have to do a bit of math in our queries to account for that, but it shouldn't be too difficult, something like `actual helpful = count(helpful == true) - count(previousValue == true)`

Because I wanted it for my own debugging here, I also added teacher tool support for the consoleticks flag (uses same infrastructure that the main webapp uses).

Screenshot:
image

I also considered adding the feedback option for every response, but I thought it looked a little cluttered...open to making that change though, if folks feel differently. It would be useful data...It looked like this:
<details>
<summary>Screenshot of alternate ui, not ui in change</summary>

</details>

### DIFF
--- a/teachertool/src/components/CriteriaFeedback.tsx
+++ b/teachertool/src/components/CriteriaFeedback.tsx
@@ -1,0 +1,49 @@
+import { Button } from "react-common/components/controls/Button";
+import css from "./styling/CriteriaFeedback.module.scss";
+import { Ticks } from "../constants";
+import { useState } from "react";
+import { showToast } from "../transforms/showToast";
+import { makeToast } from "../utils";
+import { classList } from "react-common/components/util";
+
+interface ICriteriaFeedbackProps {
+    catalogCriteriaId: string;
+    className?: string;
+}
+export const CriteriaFeedback: React.FC<ICriteriaFeedbackProps> = ({className, catalogCriteriaId}) => {
+    const [helpful, setHelpful] = useState<boolean | undefined>(undefined);
+
+    function onResponseClicked(isHelpful: boolean) {
+        const previousHelpful = helpful;
+
+        // If they click the same button again, unset the feedback.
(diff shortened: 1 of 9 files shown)


teachertool/src/components/styling/EvalResultDisplay.module.scss
```diff
@@ -19,6 +19,11 @@
         padding-top: 0;
         flex-grow: 1;
     }
+
+    .criteria-feedback {
+        align-self: flex-end;
+        margin-top: 0.5rem;
```
thsparks, line 25: Remove from print view

teachertool/src/components/CriteriaFeedback.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { classList } from "react-common/components/util";
+
+interface ICriteriaFeedbackProps {
+    catalogCriteriaId: string;
+    className?: string;
+}
+export const CriteriaFeedback: React.FC<ICriteriaFeedbackProps> = ({className, catalogCriteriaId}) => {
+    const [helpful, setHelpful] = useState<boolean | undefined>(undefined);
+
+    function onResponseClicked(isHelpful: boolean) {
```
riknoll, line 16: React.useCallback?

riknoll (approved):

PR #10070, opened 2024-08-05
Restore workspace search extensions

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/5629

This restores all of our extensions to Blockly's workspace search plugin; I had disabled them temporarily when we upgraded Blockly. The changes include:

* Pressing Shift+Enter navigates backwards through the list of search matches (default plugin can only move forwards)
* The content of the searchbox is selected when the search box is opened, which matches the behavior of most ctrl+f search boxes
* Custom rendering (selected blocks are outlined, all non-matching blocks in the workspace are faded out)

I had to do some hacks to customize the keydown behavior and get a reference to the text input element. I'll probably try to contribute the first two things back to Blockly since that's how all search boxes work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
   "dependencies": {
     "@blockly/field-colour": "5.0.5",
     "@blockly/keyboard-navigation": "0.6.5",
-    "@blockly/plugin-workspace-search": "9.0.5",
+    "@blockly/plugin-workspace-search": "9.1.0",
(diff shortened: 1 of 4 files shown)


riknoll:
Blockly merged my PR, so i'm converting this into a draft until that change gets released (which should be on thursday morning)

jwunderl (approved):

riknoll:
@jwunderl updated the PR with a new commit if you could take a brief look

PR #10076, opened 2024-08-07
Teacher Tool: Add Auth Headers to AI Requests

thsparks:
We're putting this API behind an auth wall in the backend too, so we need to include authentication headers in the request.


abchatra:
@eanders-ms please double check this when you are back.

teachertool/src/services/backendRequests.ts
```diff
@@ -99,19 +100,34 @@ export async function askCopilotQuestionAsync(shareId: string, question: string)
...
         result = await request.json();
     } catch (e) {
-        logError(ErrorCode.askCopilotQuestion, e);
+        logError(ErrorCode.askAIQuestion, e);
     }
 
     return result;
 }
+
+async function getAuthHeadersAsync(): Promise<pxt.Map<string>> {
```
eanders-ms, line 124: To avoid code duplication, let's move `getAuthHeadersAsync` to pxtlib/auth.ts. Also update the method `staticApiAsync` in pxtlib/auth.ts to call this function.

eanders-ms (approved):

teachertool/src/state/appStateContext.tsx
```diff
@@ -14,8 +14,8 @@ const copilotSlot = url.match(/copilot=([^&]+)/);
 const copilotEndpoint =
     copilotSlot && copilotSlot[1]
         ? copilotSlot[1] === "local"
-            ? `${DEV_BACKEND_LOCALHOST}/api`
-            : `https://makecode-ppe-app-backend-eastus-${copilotSlot[1]}.azurewebsites.net/api`
```
abchatra, line 18: Hmm shouldn't this point to prod\staging depending on the URL?

thsparks: By default, it points to the same api root used by the teacher tool for all its requests (https://github.com/microsoft/pxt/blob/master/teachertool/src/index.tsx#L53). This only takes effect if the ?copilot=<alias> query parameter is present in the URL. It's just what allowed us to use the ?copilot=thsparks flag for all those demos and testing scenarios, where we wanted to go through our backend but hadn't shipped the backend apis yet. It's useful for testing, but we will probably want to get rid of it at some point.

abchatra: Sounds good.

PR #10079, opened 2024-08-08
Optionally pass user language to sim extensions

microbit-matt-hillsdon:
Closes https://github.com/microsoft/pxt-microbit/issues/5746

Just a possible approach, very happy with any way the achieve the same.

You can see it working by adding &language=$LANGUAGE$ to the jacdac messageSimulator config in pxt-microbit.


microbit-matt-hillsdon:
@abchatra this is the sim language PR we discussed on the call yesterday
Also tagging @microbit-robert to deal with any queries while I'm on leave next week

abchatra:
@riknoll  please take a look and see if we don't already have the mechanism to do this in a different way.

riknoll (approved):

PR #10082, opened 2024-08-09
iframe controller mode that additionally uses native/app cmds

microbit-matt-hillsdon:
This allows an iframe embedder to take the same level of control as an app:

- intercept downloads
- intercept save
- benefit from the "back" messages

Also adds an iframe message to trigger compile.

To avoid separately disabling WebUSB this mode also forces WebUSB off. If we didn't do this a queryVariant like "androidapp" would be needed.

@abchatra this enables [this version of the ML tool MakeCode embedding experience](https://review-ml.microbit.org/controller-mode/). Video below. We'll take more meaningful advantage of intercepting the download later to help the users manage the complexity around having an "input" micro:bit for streaming bluetooth training data and then using the same or different micro:bit to run their MakeCode program that uses the ML model.

https://github.com/user-attachments/assets/300af591-4057-48ed-b95e-bf25170f5128

@jaustin @microbit-lucy this is what I would have liked to demo on the MakeCode call - any feedback to Rob.
CC @microbit-robert to help with any queries while I'm out next week.

Ahead of the GitHub actions CI build you can also try it [via this controller.html file](https://ml-tool.pxt-microbit.pages.dev/controller) - hit "toggle mode" to switch to the new mode and view the dev tools console for the new messages.

Closes https://github.com/microsoft/pxt-microbit/issues/5747

### DIFF
--- a/pxtservices/editorDriver.ts
+++ b/pxtservices/editorDriver.ts
@@ -141,6 +141,15 @@ export class EditorDriver extends IframeDriver {
         );
     }
 
+    async compile() {
+        await this.sendRequest(
+            {
+                type: "pxteditor",
+                action: "compile"
+            } as pxt.editor.EditorMessageRequest
+        );
+    }
+
     async undo() {
         await this.sendRequest(
             {


webapp/src/app.tsx
```diff
@@ -5866,7 +5866,7 @@ document.addEventListener("DOMContentLoaded", async () => {
     pxt.perf.measureEnd("setAppTarget");
 
     let theme = pxt.appTarget.appTheme;
-    const isControllerIFrame = (theme.allowParentController || pxt.shell.isControllerMode()) && pxt.BrowserUtils.isIFrame();
+    const isControllerIFrame = theme.allowParentController || pxt.shell.isControllerMode()
```
microbit-matt-hillsdon, line 5869: This isn't a change in behaviour as the iframe check is already done by the pxt.shell code.

pxteditor/editorcontroller.ts
```diff
@@ -77,6 +76,7 @@ export function bindEditorMessages(getEditorAsync: () => Promise<IProjectView>)
                             case "closeflyout": return Promise.resolve().then(() => projectView.closeFlyout());
                             case "unloadproject": return Promise.resolve().then(() => projectView.unloadProjectAsync());
                             case "saveproject": return projectView.saveProjectAsync();
+                            case "compile": return projectView.compile();
```
riknoll, line 79: please add this message in two places:
* the typing for `EditorMessageRequest` in `localtypings/pxteditor.d.ts`.
* as a method in the `EditorDriver` class in `pxtservices/editorDriver.ts`.

i'll open a pr that fixes this switch statement to disallow strings outside of the typing; looks like the call to toLowerCase() broke the type inference.

pxtlib/shell.ts
```diff
@@ -20,7 +27,11 @@ namespace pxt.shell {
                 // in iframe
                 || pxt.BrowserUtils.isIFrame();
             const nosandbox = /nosandbox=1/i.test(window.location.href);
-            const controller = /controller=1/i.test(window.location.href) && pxt.BrowserUtils.isIFrame();
+            const controllerMatch = /controller=(0|1|2)/i.exec(window.location.href)
+            if (controllerMatch && pxt.BrowserUtils.isIFrame()) {
+                controllerMode = parseInt(controllerMatch[1], 10)
```
riknoll, line 32:
```suggestion
                controllerMode = parseInt(controllerMatch[1])
```

riknoll (changes requested):
LGTM besides the comments

riknoll (approved):

jwunderl (approved):

PR #10090, opened 2024-08-13
store variants attempted before compiling

jwunderl:
When I was stepping through here in a program that was too big for v2 microbit, it started getting stuck in a retry loop; the putting a breakpoint after `compiler.compileAsync` would reset `pxt.appTarget.multiVariants` to the original value with 2 variants while the `if (res?.recompile) {` block was going / before it got to the finally. Just storing prior to attempting compile so `pxt.commands.showProgramTooLargeErrorAsync` gets the correct value for now, worth looking into a bit more later.

(big program https://makecode.microbit.org/_6oaTMe08Yc7b)

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -3277,6 +3277,7 @@ export class ProjectView
             if (simRestart) this.stopSimulator();
             let state = this.editor.snapshotState()
             this.syncPreferredEditor()
+            const attemptedVariants = pxt.appTarget.multiVariants;
 
             try {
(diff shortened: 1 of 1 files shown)


srietkerk (approved):

jwunderl:
After checking, in webusb mode this actually *does* show up as an infinite recompile loop when program is too big for v2

image

so this also fixes that, but i want to look into this a bit more later on to see if there's anything else to fix along this path

abchatra:
Great find!

PR #10093, opened 2024-08-14
Adding char code at api to blocks

abchatra:
image

Fixes https://github.com/microsoft/pxt-microbit/issues/5430

### DIFF
--- a/libs/pxt-common/pxt-core.d.ts
+++ b/libs/pxt-common/pxt-core.d.ts
@@ -284,7 +284,7 @@ declare interface String {
      * Return the Unicode value of the character at the specified location.
      * @param index The zero-based index of the desired character. If there is no character at the specified index, NaN is returned.
      */
-    //% shim=String_::charCodeAt
+    //% shim=String_::charCodeAt weight=46
     //% help=text/char-code-at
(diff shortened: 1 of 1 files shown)


libs/pxt-common/pxt-core.d.ts
```diff
@@ -284,7 +284,10 @@ declare interface String {
      * Return the Unicode value of the character at the specified location.
      * @param index The zero-based index of the desired character. If there is no character at the specified index, NaN is returned.
      */
-    //% shim=String_::charCodeAt
+    //% shim=String_::charCodeAt weight=46
+    //% help=text/char-code-at
+    //% blockId="string_code_at" block="char code from %this=text|at %pos" blockNamespace="text"
```
srietkerk, line 289:
```suggestion
    //% blockId="string_charcode_at" block="char code from $this=text|at $pos" blockNamespace="text"
```

I don't remember why, but dollar signs for argument positions is better to use than the percent if we have the ability to use them.

abchatra: Oh ya. I copy pasted the old API. thanks!

srietkerk (approved):

PR #10096, opened 2024-08-15
teacher tool: combine Checklist and Results tabs (pass 1 of 2)

eanders-ms:
Combined the Results and Checklist tabs. This is pass 1 of 2. In this pass:
* Combined the tabs
* Made QR Code and project link only show up in print
* Removed the Add Notes button and always show feedback textarea
* Always show eval status dropdown
* Checklist entries are individually evaluatable. A toolbar tray appears below the status dropdown on hover/focus.
* Checklist name is editable by clicking on it directly, or clicking the stylus tool button
* For convenience: Added URL parameter `project`, which takes a project link or share Id, allowing a project to be automatically loaded on startup. Vastly speeds up development time when you need a project loaded.

In pass 2 I will make some adjustments to the tab bar's toolbar buttons, moving them out of the tab bar entirely.

Test target: https://microbit.staging.pxt.io/app/861566127366c67af4241842c3d2e483e6211e28-84f916a8b1--eval

Combined tabs, with toolbar showing for the first criteria:
image

Print preview:
image


teachertool/src/components/CriteriaResultEntry.tsx
```diff
@@ -65,61 +44,90 @@ const CriteriaResultNotes: React.FC<CriteriaResultNotesProps> = ({ criteriaId })
...
-                            )}
-                        >
-                            {!showInput && <AddNotesButton criteriaId={criteriaId} setShowInput={setShowInput} />}
-                            {showInput && <CriteriaResultNotes criteriaId={criteriaId} />}
+                        <div className={classList(css["result-notes"], !hasFeedback ? "no-print" : undefined)}>
+                            <CriteriaResultNotes criteriaId={criteriaId} />
                         </div>
                     )}
 
+                    <div className={css["separator"]}></div>
```
thsparks, line 123: Should this come after the CriteriaFeedback below? As is, it ends up looking like the thumbs up and down are at the top of the following feedback instead of the bottom of the current feedback:
image

teachertool/src/components/CriteriaResultEntry.tsx
```diff
@@ -65,61 +44,90 @@ const CriteriaResultNotes: React.FC<CriteriaResultNotesProps> = ({ criteriaId })
...
+
                     {/* Criteria Response Feedback (For us, not student feedback) */}
                     {!isInProgress && catalogCriteria.requestFeedback && (
-                        <CriteriaFeedback catalogCriteriaId={catalogCriteria.id} className={classList(css["no-print"], css["criteria-feedback"])}/>
+                        <CriteriaFeedback
```
thsparks, line 127: We should probably hide this if the status is Not Started

thsparks (approved):
Just a few notes, nothing major. Super excited to have this change in!

teachertool/src/components/CriteriaResultEntry.tsx
eanders-ms: good catch!

PR #10098, opened 2024-08-15
Make runtime yield delay configurable

eanders-ms:
When developing the pxt-bedrock target, I was seeing major performance problems when our code ran within Minecraft's JavaScript runtime. The culprit turned out to be this fixed value of 5ms for the yield delay. Lowering the value to zero completely resolved the issue. Would we see similar improvement in pxt-minecraft? The runtime environments are very different. Experimentation will tell.

Making this value configurable so we can adjust it in the pxt-minecraft target.

### DIFF
--- a/pxtsim/embed.ts
+++ b/pxtsim/embed.ts
@@ -29,6 +29,7 @@ namespace pxsim {
         traceDisabled?: boolean;
         activePlayer?: 1 | 2 | 3 | 4 | undefined;
         theme?: string | pxt.Map<string>;
+        yieldDelay?: number;
     }
 
     export interface SimulatorInstructionsMessage extends SimulatorMessage {
--- a/pxtsim/runtime.ts
+++ b/pxtsim/runtime.ts
@@ -1199,6 +1199,7 @@ namespace pxsim {
             let userGlobals: string[];
             let __this = this // ex
             this.traceDisabled = !!msg.traceDisabled;
+            const yieldDelay = msg.yieldDelay !== undefined ? msg.yieldDelay : 5;
 
             // this is passed to generated code
             const evalIface = {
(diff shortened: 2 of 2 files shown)


abchatra:
@mmoskal to take a quick look. I am surprised 5ms to 0ms makes a difference.

eanders-ms:
> @mmoskal to take a quick look. I am surprised 5ms to 0ms makes a difference.

This codepath is hit a lot during script execution, and Minecraft's JS runtime seems to be very lax in trying to honor the value. It seems to be interpreting it as "not less than". Chromium's JS runtime will probably behave better even when hosted in Minecraft. Still, I believe @riknoll experimented setting this to zero in Arcade and saw a slight improvement.

jwunderl (approved):
lgtm, but yeah good to wait on michal thoughts here

eanders-ms:
> lgtm, but yeah good to wait on michal thoughts here

Happy to wait, no urgency, but just want to clarify that this PR doesn't change current behavior.

thsparks (approved):

jwunderl:
> > lgtm, but yeah good to wait on michal thoughts here
> 
> Happy to wait, no urgency, but just want to clarify that this PR doesn't change current behavior.

Oh of course, the default value is the same as it previously was, I meant more in the 'are there any dragons hidden in changing this value that we should track' sense.

mmoskal:
I think it should be OK. This code should only kick in when the user script is running without yielding for at least 20ms. You could increase that instead, to say 50ms. But 0 ms yield also sounds good!

eanders-ms:
> I think it should be OK. This code should only kick in when the user script is running without yielding for at least 20ms. You could increase that instead, to say 50ms. 

Nice info. Good to know of this behavior, thank you!

PR #10099, opened 2024-08-15
add importexternalproject editor command

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/5748

Adds a command for importing projects for the embedded editor scenario. It works pretty similarly to the skillmap codepath, but it doesn't require the projects to be stored on the same domain.

The flow goes like this:

1. Outside frame sends the `importexternalproject` command to the iframe with a project
2. The inner frame saves the project to a temporary DB (not the workspace DB)
3. The inner frame responds with a URL that the outer frame can open in a new tab to import the project
4. When the URL gets opened, the project is removed from the temporary DB and stored in whatever workspace is active (e.g. the indexedDB, the cloud, etc.)

The temporary db gets cleared whenever a new project is written or removed, so this can't be used to import multiple projects at once. If that's a scenario, I can loosen this restriction. Right now, it's really meant to be an "Open in MakeCode" button for a single project.

@microbit-matt-hillsdon @microbit-carlos @jaustin let me know if you have any feedback on this

### DIFF
--- a/localtypings/pxteditor.d.ts
+++ b/localtypings/pxteditor.d.ts
@@ -53,6 +53,7 @@ declare namespace pxt.editor {
         | "closeflyout"
         | "newproject"
         | "importproject"
+        | "importexternalproject"
         | "importtutorial"
         | "openheader"
         | "proxytosim" // EditorMessageSimulatorMessageProxyRequest
@@ -256,6 +257,19 @@ declare namespace pxt.editor {
         // (optional) filtering argument
         filters?: ProjectFilters;
         searchBar?: boolean;
+    }
+
+    export interface EditorMessageImportExternalProjectRequest extends EditorMessageRequest {
+        action: "importexternalproject";
+        // project to load
+        project: pxt.workspace.Project;
+    }
+
+    export interface EditorMessageImportExternalProjectResponse extends EditorMessageResponse {
+        action: "importexternalproject";
+        resp: {
+            importUrl: string;
+        };
     }
 
     export interface EditorMessageSaveLocalProjectsToCloud extends EditorMessageRequest {
--- a/pxteditor/projectImport.ts
+++ b/pxteditor/projectImport.ts
@@ -28,7 +28,7 @@ export async function removeProjectAsync(importId: string): Promise<pxt.workspac
     const stored = await db.getAsync<StoredProject>(PROJECT_TABLE, importId);
     await db.deleteAllAsync(PROJECT_TABLE);
 
-    return stored.project;
+    return stored?.project;
 }
 
 async function initAsync() {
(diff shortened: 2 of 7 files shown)


localtypings/pxteditor.d.ts
```diff
@@ -53,6 +53,7 @@ declare namespace pxt.editor {
         | "closeflyout"
         | "newproject"
         | "importproject"
+        | "importexternalproject"
```
abchatra, line 56: exportproject may be a better name?

eanders-ms (approved):

localtypings/pxteditor.d.ts
riknoll: exportproject sound more like a message that would take an existing project in the makecode workspace and send a file back

jwunderl (approved):

microbit-matt-hillsdon:
Thanks @riknoll. We'll try this out today and give feedback.

microbit-matt-hillsdon:
Thanks, I think this works for us in Chrome/Edge and Firefox. 

I don't think the current implementation can be made to work in Safari as it won't allow the storage to be shared at all. IndexDB isn't supported by Safari in iframes and it seems to apply stricter rules to local storage sharing than other browsers. Please let me know if you've seen it working in Safari. So far as I can tell, the only option here would be a server-side data exchange approach. 

It also doesn't work unless the domains meet the [same site definition](https://html.spec.whatwg.org/multipage/browsers.html#sites) used to determine whether to [partition storage/state](https://developer.mozilla.org/en-US/docs/Web/Privacy/State_Partitioning#state_partitioning). If I try with two unrelated TLDs (by adding local DNS entries) then the IndexedDB store is empty when opening the MakeCode URL and the code crashes when trying to delete the entry from the store:

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'project')
    at Object.removeProjectAsync (projectImport.js:22:1)
    at async ProjectView.importEmbedProjectAsync (app.js:2338:1)
```

This isn't an issue for classroom.microbit.org/ml.microbit.org/makecode.microbit.org but is worth being aware of.

pxteditor/projectImport.ts
```diff
@@ -0,0 +1,49 @@
...
+    return toStore.importId;
+}
+
+export async function removeProjectAsync(importId: string): Promise<pxt.workspace.Project> {
+    const db = await initAsync();
+
+    const stored = await db.getAsync<StoredProject>(PROJECT_TABLE, importId);
+    await db.deleteAllAsync(PROJECT_TABLE);
+
+    return stored.project;
```
microbit-matt-hillsdon, line 31: I think this should return null or undefined if `stored` doesn't exist, otherwise we error on `stored.project` rather than running the error handling in importEmbedProjectAsync (which checks project is truthy).

riknoll:
@microbit-matt-hillsdon yeah, this was only intended to work for embedders on the same domain

i wasn't aware of that restriction on safari! i tried it on localhost and everything worked fine, but i wouldn't put it past safari to be inconsistent between localhost and another domain. we might just have to test it out and see!

PR #10103, opened 2024-08-16
teacher tool: move toolbar buttons into page

eanders-ms:
I moved the toolbar buttons into the checklist page and gave them labels:

image

The resulting button row is pretty wide. When the checklist workspace gets too narrow, the labels stop rendering:
image

Other changes:
* When a new checklist is created, it gets a default name of "New Checklist"
* Updated result toolbar button styling to match checklist header button styling (visible in second screenshot)


srietkerk:
A couple of design nits.
- I think the rename should still be next to the checklist name. All the other actions in the list refer to something that will impact every item on the page, so I think having the rename button in that menu is a bit confusing.
- Ordering. Can we put `Evaluate` and `Auto Run` at the beginning or end of the list since they are the most important actions to take? And then have printing and export next to each other since they are similar actions.
- I'm wondering what it would look like to have a bar underneath the actions. The header looks a bit busy with all the buttons expanded and I'm wondering if a bar under that section will create some separation.
- Just curious if the checklist side shrinks down enough will all of the items collapse into a menu?
- Also wanted to bring up the merging play and auto run idea again. We talked about this a bit a couple of months ago, and wasn't sure if that was still something we'd want to pursue or consider. Not a change for this PR, but maybe a follow-up. The idea would be that the play would be configurable, like how we have with our download buttons. You have the main "play/evaluate" but then the menu or configuration would be to toggle auto run or not. 
image

thsparks:
> A couple of design nits.
> 
> * I think the rename should still be next to the checklist name. All the other actions in the list refer to something that will impact every item on the page, so I think having the rename button in that menu is a bit confusing.
> * Ordering. Can we put `Evaluate` and `Auto Run` at the beginning or end of the list since they are the most important actions to take? And then have printing and export next to each other since they are similar actions.
> * I'm wondering what it would look like to have a bar underneath the actions. The header looks a bit busy with all the buttons expanded and I'm wondering if a bar under that section will create some separation.
> * Just curious if the checklist side shrinks down enough will all of the items collapse into a menu?
> * Also wanted to bring up the merging play and auto run idea again. We talked about this a bit a couple of months ago, and wasn't sure if that was still something we'd want to pursue or consider. Not a change for this PR, but maybe a follow-up. The idea would be that the play would be configurable, like how we have with our download buttons. You have the main "play/evaluate" but then the menu or configuration would be to toggle auto run or not.

I agree with Sarah about keeping rename next to the checklist name (I also think we can get away with it just being the pencil and not having the full "Rename" text in there).

Also agree about ordering. I would recommend (from left to right): Export -> Print -> Auto-run -> Evaluate. I'd also be interested in seeing how Evaluate looks as a primary button.

Lastly, did you experiment with putting the buttons on the same line as the title (example below)? At least experimentally, I think it looks more pleasing to the eye, though I realize space might be a bit tight in that scenario, especially with longer checklist names. I think removing "rename" and maybe changing "Evaluate" to "Run" or "Check" could help.
image

srietkerk:
> Also agree about ordering. I would recommend (from left to right): Export -> Print -> Auto-run -> Evaluate. I'd also be interested in seeing how Evaluate looks as a primary button.

Yeah, I like that ordering and agree that Evaluate should stand out since it's the primary action. 

>Lastly, did you experiment with putting the buttons on the same line as the title (example below)? At least experimentally, I think it looks more pleasing to the eye, though I realize space might be a bit tight in that scenario, especially with longer checklist names. I think removing "rename" and maybe changing "Evaluate" to "Run" or "Check" could help.

Ooh, I like having the actions on the same line as the name!!

eanders-ms:
> > Also agree about ordering. I would recommend (from left to right): Export -> Print -> Auto-run -> Evaluate. I'd also be interested in seeing how Evaluate looks as a primary button.
> 
> Yeah, I like that ordering and agree that Evaluate should stand out since it's the primary action.
> 
> > Lastly, did you experiment with putting the buttons on the same line as the title (example below)? At least experimentally, I think it looks more pleasing to the eye, though I realize space might be a bit tight in that scenario, especially with longer checklist names. I think removing "rename" and maybe changing "Evaluate" to "Run" or "Check" could help.
> 
> Ooh, I like having the actions on the same line as the name!!

Good feedback! I will experiment with these suggestions.

thsparks (approved):
Left some optional feedback, but discoverability wins are good regardless. Feel free to check-in as-is if experimentation is too time consuming!

srietkerk (approved):
Looks great! Thanks Eric!

PR #10106, opened 2024-08-19
Don't convert indexOf to python's .index

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/5703

This removes the mapping between JavaScript's `Array.indexOf` and Python's `List.index` function. Python's index function throws an exception if the value is missing instead of returning -1, so this isn't a safe conversion to make.

This means that if you are using `Array.indexOf` in Blocks or TypeScript, when you switch to Python you will now get the non-standard `.index_of`, which will maintain the same behavior at the cost of being a function that doesn't exist.

### DIFF
--- a/cli/pyconv.ts
+++ b/cli/pyconv.ts
@@ -412,7 +412,7 @@ namespace py {
     }
     export interface Constant extends Expr {
         kind: "Constant";
-        value: any; // ??? 
+        value: any; // ???
     }
 
     // the following expression can appear in assignment context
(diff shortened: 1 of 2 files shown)


cli/pyconv.ts
```diff
@@ -1443,7 +1443,6 @@ let funMap: Map<FunOverride> = {
     "ustruct.calcsize": { n: "pins.packedSize", t: tpNumber },
     "pins.I2CDevice.read_into": { n: ".readInto", t: tpVoid },
     "bool": { n: "!!", t: tpBoolean },
-    "Array.index": { n: ".indexOf", t: tpNumber },
```
srietkerk, line 1446: I see that this is a mapping of TS to Python, but curious what "n" and "t" stand for in this object?

srietkerk (approved):
Looks good!

cli/pyconv.ts
riknoll: well `t` is definitely type, not sure what the etymology of `n` is but it's the text that gets replaced. bad names i agree

PR #10107, opened 2024-08-19
Blocks with toolbox parents will have correct snippets and toolbox previews in JS/PY

srietkerk:
Fixes: https://github.com/microsoft/pxt-microbit/issues/5349

As described in the bug, the problem we were having is that the music play blocks, with the new Playable infrastructure that we have in place, have a `toolboxParent`. This parent is the `play(toPlay, playbackMode)` block. We handle this `toolboxParent` in blocks, but it was not getting handled in JS/Python snippets, so when the play blocks would get dragged into the Monaco editor from the toolbox, the inner block would be generated, but not the outer parent block.

Big thank you to @riknoll for all your help getting situated in the right files and implementation guidance.

This PR does the following:
1. Fixes the snippet drag out problems that are the root of the linked issue. These changes can be found in `snippet.ts` and has to do with the `parameterOverride` object added to the SnippetContext.
2. Adds an `includeParentSnippet` flag to `getSnippet` to dictate whether or not we actually want the snippet to include that `toolboxParent`. We create snippets both when we drag the JS/Py snippets from the toolbox _and_ when a user makes completions (as in, presses tab to fill in code). In the "drag-out-of-the-toolbox" scenario, we want the parent block to exist in order to get the correct block that will convert smoothly when switching back to blocks. However, in the "completions" scenario, we shouldn't be getting the parent because it could introduce nesting that we don't want. Thus, the flag is passed in order to say "only bring the parent with the toolbox drag, not on completions". Changes in `snippet.ts` and `pxtcompiler/emitter/service.ts` reflect this fix. 
**Before**
image
**After**
image

3. Fixes a bug with completions that came apparent to me when trying to complete `music.play`. When tabbing on `music.play`, it would complete to `music.playable` and include the rest of the inner part of `music.play`. @riknoll, (thank you!!) noticed that this was because we weren't indexing correctly and would essentially cut out the first left parenthesis that we would find, so we just needed to account for finding that parenthesis when completing. This is the one line change in `monaco.tsx`.
4. Fixes the snippet preview in the toolbox. The snippet previews are handled separately from actually generating the snippet code in the editor. The "block" that you see for JS/Python snippets is generated in `monacoFlyout.tsx`, so I've added `toolboxParent` handling logic there. This fix also needed some changes in `monaco.tsx`, where we actually get/generate the block definitions for the toolbox, specifically in `partitionBlocks()`.

Upload target: https://makecode.microbit.org/app/674c0781c279b633915e3c16177d0232799eacb2-262379068a#editor you can play with the `music.play` completions in JS and Python and also take a look at the toolbox improvements.

### DIFF
--- a/pxtcompiler/emitter/snippets.ts
+++ b/pxtcompiler/emitter/snippets.ts
@@ -125,7 +125,9 @@ namespace ts.pxtc.service {
         let overrideSnippet: SnippetNode;
         if (parameterOverride) {
             overrideParamLabel = Object.keys(parameterOverride)?.[0];
-            overrideSnippet = parameterOverride[overrideParamLabel];
+            if (overrideParamLabel) {
+                overrideSnippet = parameterOverride[overrideParamLabel];
+            }
         }
 
 
--- a/pxtcompiler/emitter/snippets.ts
+++ b/pxtcompiler/emitter/snippets.ts
@@ -104,12 +104,14 @@ namespace ts.pxtc.service {
         takenNames: pxt.Map<SymbolInfo>;
         checker: ts.TypeChecker;
         screenSize?: pxt.Size;
+        parameterOverride?: { [key: string]: SnippetNode };
+        includeParentSnippet?: boolean;
     }
 
     export function getSnippet(context: SnippetContext, fn: SymbolInfo, decl: ts.FunctionLikeDeclaration, python?: boolean, recursionDepth = 0): SnippetNode {
         // TODO: a lot of this is duplicate logic with blocklyloader.ts:buildBlockFromDef; we should
         //  unify these approaches
-        let { apis, takenNames, blocksInfo, screenSize, checker } = context;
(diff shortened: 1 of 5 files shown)


pxtcompiler/emitter/snippets.ts
```diff
@@ -118,6 +120,14 @@ namespace ts.pxtc.service {
         let addNamespace = false;
         let namespaceToUse = "";
         let functionCount = 0;
+        let parentSnippet: SnippetNode;
+        let overrideParamLabel: string;
+        let overrideSnippet: SnippetNode;
+        if (parameterOverride) {
+            overrideParamLabel = Object.keys(parameterOverride)?.[0];
+            overrideSnippet = parameterOverride[overrideParamLabel];
```
kimprice, line 128: Should we check that `overrideParamLabel` is nonnull?

srietkerk: Sure!

jwunderl (approved):
lgtm

PR #10110, opened 2024-08-21
code eval: add beta tag

eanders-ms:
This change adds a beta tag to the code evaluation tool. To give the tag a place to anchor, I included the app title in the header bar, which also aligns with previous feedback. However, since the title is wide and doesn't fit well on small screens, I defined standard media breakpoints to allow the header layout to be responsive. If the teacher tool already has a set of generic breakpoints, I'll use those; otherwise, these breakpoints can be adopted across the app when we address responsiveness.

The design of these breakpoints is somewhat unique in that they're composable.

Test target: https://makecode.microbit.org/app/cf32d5a03e863c227a7d4bee10b3b9193dc75b08-0e9e16aba4--eval

Header layouts at different widths:
image
image
image

Fixes https://github.com/microsoft/pxt-microbit/issues/5743

### DIFF
--- a/teachertool/src/global.scss
+++ b/teachertool/src/global.scss
@@ -139,78 +139,104 @@ code {
 /****  BREAKPOINT QUERIES  *****/
 /*******************************/
 
-// Composable breakpoints
+// Composable breakpoints, aligned to react-common breakpoint sizes (see https://github.com/microsoft/pxt/blob/master/react-common/styles/react-common-breakpoints.less)
 
+// Become visible at the given breakpoint
 .min-xl,
-.max-xl,
 .min-lg,
-.max-lg,
 .min-md,
-.max-md,
 .min-sm,
-.max-sm,
+.min-2sm,
 .min-xs,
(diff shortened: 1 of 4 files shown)


teachertool/src/global.scss
```diff
@@ -135,6 +135,86 @@ code {
     }
 }
 
+/*******************************/
```
thsparks, line 138: I guess the breakpoints in https://github.com/microsoft/pxt/blob/master/react-common/styles/react-common-breakpoints.less are inaccessible to us since it's less, but I wonder if it's worth mirroring that pattern for easy portability between our components and react-common...not something we need to block on, though.

thsparks (approved):
Looks great! I have one comment which I don't feel strongly about. Feel free to check in as-is.

teachertool/src/global.scss
eanders-ms: Makes sense to mirror the breakpoints for portability.

PR #10112, opened 2024-08-21
Show Error on Failed Criteria Evaluation

thsparks:
Fixes https://github.com/microsoft/pxt-microbit/issues/5819

If a criterion fails to evaluate through a non-user error (i.e. we can't reach the AI service), we don't really provide any indication of the issue to the user, it just shows no result.

With this change, we show an error window:
image

Upload target (easy to get an error by using "Ask AI", since ai endpoint doesn't exist yet): https://makecode.microbit.org/app/d51851820874badb8277c7a4bf73ce82ff4bf5f9-eb19b40cfa--eval?tc=1

### DIFF
--- a/teachertool/src/components/CriteriaResultEntry.tsx
+++ b/teachertool/src/components/CriteriaResultEntry.tsx
@@ -48,10 +48,9 @@ interface CriteriaResultErrorProps {
     error: string;
 }
 const CriteriaResultError: React.FC<CriteriaResultErrorProps> = ({ error }) => {
-    // Display an X icon and the error message in a faint red box
     return (
         <div className={css["result-error"]}>
-            <i className="fas fa-times-circle"></i>
+            <i className="fas fa-exclamation-circle"></i>
             <div className={css["error-info-container"]}>
                 <span className={css["error-title"]}>{Strings.UnableToEvaluate}</span>
                 <span className={css["error-details"]}>{error}</span>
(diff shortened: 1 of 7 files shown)


jwunderl (approved):

teachertool/src/components/CriteriaResultEntry.tsx
```diff
@@ -44,6 +45,31 @@ const CriteriaResultNotes: React.FC<CriteriaResultNotesProps> = ({ criteriaId })
...
 };
 
+interface CriteriaResultErrorProps {
+    criteriaInstanceId: string;
+    error: string;
+}
+
+const CriteriaResultError: React.FC<CriteriaResultErrorProps> = ({ criteriaInstanceId, error }) => {
+    return (
+        <div className={css["result-error"]}>
```
eanders-ms, line 55: I believe we want it excluded from the printed copy.
```suggestion
        <div className={classList(css["result-error"], "no-print")}>
```

eanders-ms: image

teachertool/src/components/styling/EvalResultDisplay.module.scss
```diff
@@ -102,6 +102,49 @@
...
+        border: 1px solid var(--pxt-error);
+        border-radius: 1rem;
+        background-color: var(--pxt-error-background);
+        gap: 0.5rem;
+
+        .error-info-container {
+            display: flex;
+            flex-direction: column;
+            align-items: flex-start;
+            justify-content: space-between;
```
eanders-ms, line 122: very minor suggestion. clip text to the error box when layout is very narrow:
```suggestion
            justify-content: space-between;
            overflow: hidden;
```

eanders-ms (approved):
Looks good. Made a few minor suggestions.

PR #10113, opened 2024-08-22
Poll for compile variant hex files in parallel

riknoll:
One benefit of the new compile service is that it can now execute compile jobs in parallel. This PR takes advantage of that feature and makes it so that when hitting the compile service for micro:bit, we request the compiling of both the V1 and V2 hexes at the same time.

In other words, it should cut the CPP compile time in half (but only for micro:bit, since it's the only "fat hex" target).

I'm planning on doing the same thing with our CLI builds in the near future, which should speed up `pxt buildtarget` significantly for targets with multiple compile variants.

### DIFF
--- a/pxtlib/package.ts
+++ b/pxtlib/package.ts
@@ -1167,35 +1167,36 @@ namespace pxt {
                 }
                 const prevVariant = pxt.appTargetVariant
 
-                if (variant)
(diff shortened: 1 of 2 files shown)


abchatra:
Clever change.

jwunderl (approved):

PR #10116, opened 2024-08-23
Add the template code to the first custom file created, no matter the name

srietkerk:
Fixes https://github.com/microsoft/pxt-microbit/issues/5367

It's frustrating to not have something to work off of if it's your first time creating custom blocks. Including our "creating blocks" template should not depend on that file's name being `custom.ts`. However, for those that are making a lot of custom blocks, having a template in every new file that you make would be frustrating to deal with. Thus, we've changed it so that we are no longer special casing `custom.ts`. For the first custom file that you make, no matter the name, the template will be included in that file. Any subsequent custom files made will have the `Add your code here` comment as is the current behavior.

Upload target: https://makecode.microbit.org/app/6b84da6b59b4491ed7055b51c377ce774d5774e0-f63aae8ed5

### DIFF
--- a/webapp/src/filelist.tsx
+++ b/webapp/src/filelist.tsx
@@ -362,11 +362,13 @@ export class FileList extends data.Component<ISettingsProps, FileListState> {
                 return Promise.resolve()
             }
             let fileText = "";
-            if (fileName == customFile) {
+            const tsFiles = pkgCfg.files.filter(f => f.endsWith(".ts"));
+            // if the length is 1, the only ts file is main.ts, thus
+            // we are adding the first custom file to the project
+            if (tsFiles.length === 1) {
                 fileText = customFileHeader(pxt.appTarget.appTheme.homeUrl) + customFileText;
             } else if (comment) {
-                fileText = `${comment} ${commentText}
-`;
+                fileText = `${comment} ${commentText}`;
             }
 
             return this.props.parent.updateFileAsync(fileName, fileText, true);
--- a/webapp/src/filelist.tsx
+++ b/webapp/src/filelist.tsx
@@ -362,13 +362,14 @@ export class FileList extends data.Component<ISettingsProps, FileListState> {
                 return Promise.resolve()
             }
             let fileText = "";
-            const tsFiles = pkgCfg.files.filter(f => f.endsWith(".ts"));
+            const tsFiles = pkgCfg.files.filter(f => f.endsWith(".ts") && !f.endsWith(".g.ts"));
             // if the length is 1, the only ts file is main.ts, thus
             // we are adding the first custom file to the project
             if (tsFiles.length === 1) {
                 fileText = customFileHeader(pxt.appTarget.appTheme.homeUrl) + customFileText;
             } else if (comment) {
-                fileText = `${comment} ${commentText}`;
+                fileText = `${comment} ${commentText}
+`;
             }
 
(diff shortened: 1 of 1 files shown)


riknoll (approved):

webapp/src/filelist.tsx
```diff
@@ -362,11 +362,13 @@ export class FileList extends data.Component<ISettingsProps, FileListState> {
...
             }
             let fileText = "";
-            if (fileName == customFile) {
+            const tsFiles = pkgCfg.files.filter(f => f.endsWith(".ts"));
+            // if the length is 1, the only ts file is main.ts, thus
+            // we are adding the first custom file to the project
+            if (tsFiles.length === 1) {
                 fileText = customFileHeader(pxt.appTarget.appTheme.homeUrl) + customFileText;
             } else if (comment) {
-                fileText = `${comment} ${commentText}
```
thsparks, line 368: nit: Was this newline here for stylistic (empty new line at the end of the file) reasons? If so, should we keep it?

thsparks (approved):

eanders-ms (approved):

webapp/src/filelist.tsx
srietkerk: Ah, yeah, it is. Good call, I'll leave it in so we have that extra new line in the file.

jwunderl (approved):
LGTM; quick question out of curiosity, did you ever find out what the cause of that issue where the toolbox wasn't updating for non custom.ts files was?

webapp/src/filelist.tsx
```diff
@@ -362,11 +362,13 @@ export class FileList extends data.Component<ISettingsProps, FileListState> {
                 return Promise.resolve()
             }
             let fileText = "";
-            if (fileName == customFile) {
+            const tsFiles = pkgCfg.files.filter(f => f.endsWith(".ts"));
```
jwunderl, line 365: Ah wait, have you checked this in arcade? We have some generated files over there with the `.g.ts` file extension (e.g. add  a tilemap to your project in blocks and then convert to ts)

srietkerk:
@jwunderl, yeah, it was just a me thing/possibly a race condition that I was hitting. As you suggested, when I switched over to blocks, the category was there as expected. It also looks like I'm not running into the same problem with the upload target, so I'm guessing it was just some weirdness I saw locally.

webapp/src/filelist.tsx
srietkerk: I haven't. I'll test it there with a tilemap now.

jwunderl:
> @jwunderl, yeah, it was just a me thing/possibly a race condition that I was hitting. As you suggested, when I switched over to blocks, the category was there as expected. It also looks like I'm not running into the same problem with the upload target, so I'm guessing it was just some weirdness I saw locally.

Interesting, weird that it sounded like it was working nicely for `custom.ts`. maybe we have some special cased logic for ignoring caches related to custom.ts or something along those lines, would have to check; thanks for info!

webapp/src/filelist.tsx
srietkerk: Yeah, it looks like tilemaps break this. Good catch. Is it okay to just add a case to filter out `.g.ts` files? Or are there other main package files that I should look out for?

srietkerk: Talked offline about this, just going to filter out `.g.ts` files from the included files.

PR #10117, opened 2024-08-26
add "eval on load" toggle

eanders-ms:
Removes "auto-run" toggle in favor of "eval on load". We'd previously decided to call it "run on load", but this seemed ambiguous once I saw it in the UI. What would run exactly? The code eval or the project itself? Open to feedback, as always.

image

Fixes https://github.com/microsoft/pxt-microbit/issues/5617


teachertool/src/transforms/runEvaluateAsync.ts
```diff
@@ -6,6 +6,11 @@ import { runSingleEvaluateAsync } from "./runSingleEvaluateAsync";
 
 export async function runEvaluateAsync(fromUserInteraction: boolean) {
     const { state: teacherTool } = stateAndDispatch();
+    const { projectMetadata } = teacherTool;
+
+    if (!projectMetadata) {
```
srietkerk, line 11: Do we also need to make a check that there are criteria created? If a user loads a project without creating any criteria, will we error?

teachertool/src/constants.ts
```diff
@@ -17,8 +17,8 @@ export namespace Strings {
     export const ChecklistNamePlaceholder = lf("Enter Checklist Name");
     export const AddCriteria = lf("Add Criteria");
     export const Actions = lf("Actions");
-    export const AutoRun = lf("Auto-run");
-    export const AutoRunDescription = lf("Automatically re-evaluate when the checklist or project changes");
+    export const RunOnLoad = lf("Eval on load");
```
srietkerk, line 20: Will changing this to "Evaluate on Load" be too long? It's a bit of a nitpick, but I almost regard "eval" as short for "evaluation" rather than "evaluate", so it feels a bit off to be used here.

teachertool/src/constants.ts
```diff
@@ -17,8 +17,8 @@ export namespace Strings {
...
+    export const RunOnLoadDescription = lf("Automatically evaluate the project when it is loaded");
```
srietkerk, line 21:
```suggestion
    export const RunOnLoadDescription = lf("Automatically evaluate a project when it is loaded");
```

teachertool/src/transforms/runEvaluateAsync.ts
eanders-ms: I don't see why that would be an error, but it we could interpret it as an indication that the user is exploring the UI to learn how it works. It might make sense to view this as an opportunity to suggest adding criteria.

teachertool/src/constants.ts
eanders-ms: Sure, I can change it to "Evaluate". It seems clearer. Later when we address responsiveness we might want to dynamically shorten it when the panel gets narrow.

teachertool/src/transforms/runEvaluateAsync.ts
eanders-ms: Oh, it works fine without any criteria defined. We don't throw any errors.

teachertool/src/components/ProjectWorkspace.tsx
```diff
@@ -7,18 +7,41 @@ import { Toolbar } from "./Toolbar";
...
     const projectName = getSafeProjectName(teacherTool);
     return projectName ? <div className={css.projectName}>{projectName}</div> : null;
 };
 
+const ProjectToolbar: React.FC = () => {
+    const { state: teacherTool } = useContext(AppStateContext);
+    const { runOnLoad } = teacherTool;
+
+    const onRunOnLoadToggled = (checked: boolean) => {
+        pxt.tickEvent(Ticks.RunOnLoad, { checked: checked ? "true" : "false" });
```
thsparks, line 24:
```suggestion
        pxt.tickEvent(Ticks.RunOnLoad, { newValue: checked ? "true" : "false" });
```

Just for clarity. When I see "checked" my immediate next thought is "is that the before checked or the after checked"

teachertool/src/constants.ts
thsparks: +1 for Evaluate (or "Run Checks on Load", but I think Evaluate on Load is more concise)

thsparks (approved):
LGTM!

srietkerk (approved):
Looks good!

PR #10119, opened 2024-08-26
don't error when a hint is missing from a tutorial step

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/5845
fixes https://github.com/microsoft/pxt-microbit/issues/5854

i'm not sure how this broke but fixing this fixes the issue and closing issues is my job not solving mysteries


thsparks (approved):
Not sure I understand how this fixes the linked bugs, but change seems fine in general.

kimprice (approved):

riknoll:
@thsparks the exception was preventing the block compile from running

PR #10122, opened 2024-08-27
fix snapshot rendering of comments

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/5850

there were three issues here:
* the blockly css sets the comment colors on the injectionDiv class, which wasn't present on the svg we were rendering into a png
* the new comment icons link to images rather than using data uris so it was skipping our step that encodes images as pngs
* the blocklyCommentTextarea class was changed to blocklyTextarea

also converted everything to async/await

### DIFF
--- a/pxtblocks/layout.ts
+++ b/pxtblocks/layout.ts
@@ -391,7 +391,7 @@ async function expandImagesAsync(xsg: Document): Promise<void> {
     const dataUriPromises = images
         .filter(image => {
             const href = image.getAttributeNS(XLINK_NAMESPACE, "href");
-            return !!href
+            return href && !/^data:/.test(href);
         })
         .map(async image => {
             const href = image.getAttributeNS(XLINK_NAMESPACE, "href");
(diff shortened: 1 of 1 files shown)


pxtblocks/layout.ts
```diff
@@ -384,65 +384,96 @@ export function documentToSvg(xsg: Node): string {
...
-    return Promise.all(p).then(() => { })
+
+    const linkedIconPromises = images
+        .filter(image => /^\/.*.svg$/.test(image.getAttribute("href")))
+        .map(async image => {
+            const svgUri = image.getAttribute("href");
+
+            // Don't love hard coding these icon values, but the comment icons set their
+            // width/height using CSS so there isn't a great means of detecting it
+            let width = 24;
```
jwunderl, line 429: would `window.getComputedStyle(image).getPropertyValue("width")` work? Not a big deal to hardcode this really / computed style has it's own downsides, but mainly curious.

jwunderl (approved):

srietkerk (approved):

pxtblocks/layout.ts
riknoll: hmm does that work with svg elements? i'll give it a shot!

riknoll: does not work, not sure if it's because it's svg or because it's not in the document

PR #10123, opened 2024-08-27
Fix 'save a copy' time machine bugs

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/5853

fixes three tiny bugs with the time machine dialog:
* when trying to save a copy from "Now", the project name would contain "invalid date" instead of a timestamp
* projects saved using the "save a copy" button weren't keeping around the project history from before the selected time
* opening the time machine dialog for a project with no history (e.g. you clicked the button too fast) would cause an exception


srietkerk (approved):
LGTM

PR #10124, opened 2024-08-27
don't prevent block icons from being clicked on full block fields

riknoll:
this fixes a bug i noticed while testing today.

blockly's dropdown field treats the entire block as a click target when the block has only a single field in it. that generally works fine except when the block also has an icon on it (e.g. a warning or a comment) since it causes the field to swallow up the click events on those icons. this pr overrides the event binding to fix that.

i should probably open an issue against blockly for this

also annoying is the fact that i have to duplicate this code in two places. i briefly tried to get this working with mixin classes, but typescript doesn't like mixin classes that have private or protected properties.


srietkerk (approved):
Don't know much about this area, but I don't see anything that would cause problems!

PR #10127, opened 2024-08-28
print layout improvements

eanders-ms:
**Improvements to print layout**
- Animation-based resize of notes textarea wasn't reflowing properly in print preview. Fixed by showing a div in this view instead of the textarea.
- Prevent insertion of page breaks before notes if feedback runs over a page, as this was causing insertion of blank pages.

**Other tweaks**
- Format criteria description in italics in print preview. Without this distinction it looked like it was part of the feedback notes.
- Don't hide label of the Evaluate button on narrow panels. The removal of autorun allowed for this button to remain prominent.
- prettier

image

image

Fixes https://github.com/microsoft/pxt-microbit/issues/5611

### DIFF
--- a/teachertool/src/components/CriteriaInstanceDisplay.tsx
+++ b/teachertool/src/components/CriteriaInstanceDisplay.tsx
@@ -1,11 +1,10 @@
+import css from "./styling/CriteriaInstanceDisplay.module.scss";
 import { getCatalogCriteriaWithId } from "../state/helpers";
 import { CriteriaInstance, CriteriaParameterValue } from "../types/criteria";
 import { logDebug } from "../services/loggingService";
 import { setParameterValue } from "../transforms/setParameterValue";
 import { classList } from "react-common/components/util";
 import { getReadableBlockString, splitCriteriaTemplate } from "../utils";
-// eslint-disable-next-line import/no-internal-modules
-import css from "./styling/CriteriaInstanceDisplay.module.scss";
 import { useContext, useMemo, useState } from "react";
 import { Input } from "react-common/components/controls/Input";
 import { Button } from "react-common/components/controls/Button";
@@ -158,7 +157,10 @@ export const CriteriaInstanceDisplay: React.FC<CriteriaInstanceDisplayProps> = (
                     </span>
                 ))}
             </div>
-            <div className={css["criteria-description"]}>{catalogCriteria.description}</div>
+            <div className={classList(css["criteria-description"], "no-print")}>{catalogCriteria.description}</div>
+            <div className={classList(css["criteria-description"], css["for-print"], "only-print")}>
+                {catalogCriteria.description}
+            </div>
         </div>
     ) : null;
 };
--- a/teachertool/src/components/CriteriaResultEntry.tsx
+++ b/teachertool/src/components/CriteriaResultEntry.tsx
@@ -23,7 +23,7 @@ interface CriteriaResultNotesProps {
 
 const CriteriaResultNotes: React.FC<CriteriaResultNotesProps> = ({ criteriaId }) => {
     const { state: teacherTool } = useContext(AppStateContext);
-    const notesInitialValue = teacherTool.evalResults[criteriaId]?.notes;
+    const notes = teacherTool.evalResults[criteriaId]?.notes;
 
     const onTextChange = (str: string) => {
         setEvalResultNotes(criteriaId, str);
@@ -37,16 +37,16 @@ const CriteriaResultNotes: React.FC<CriteriaResultNotesProps> = ({ criteriaId })
                     ariaLabel={lf("Feedback regarding the criteria result")}
                     label={lf("Feedback")}
                     title={lf("Write your notes here")}
-                    initialValue={notesInitialValue}
+                    initialValue={notes}
                     autoResize={true}
                     onChange={onTextChange}
                     autoComplete={false}
                     intervalMs={500}
                 />
             </div>
-            {notesInitialValue && (
+            {notes && (
                 <div className={classList(css["notes-container"], css["for-print"], "only-print")}>
-                    {notesInitialValue}
+                    {notes}
                 </div>
             )}
         </>
--- a/teachertool/src/components/CriteriaResultEntry.tsx
+++ b/teachertool/src/components/CriteriaResultEntry.tsx
@@ -23,25 +23,33 @@ interface CriteriaResultNotesProps {
 
 const CriteriaResultNotes: React.FC<CriteriaResultNotesProps> = ({ criteriaId }) => {
     const { state: teacherTool } = useContext(AppStateContext);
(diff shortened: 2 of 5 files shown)


teachertool/src/components/CriteriaInstanceDisplay.tsx
```diff
@@ -1,11 +1,10 @@
+import css from "./styling/CriteriaInstanceDisplay.module.scss";
 import { getCatalogCriteriaWithId } from "../state/helpers";
 import { CriteriaInstance, CriteriaParameterValue } from "../types/criteria";
 import { logDebug } from "../services/loggingService";
 import { setParameterValue } from "../transforms/setParameterValue";
 import { classList } from "react-common/components/util";
 import { getReadableBlockString, splitCriteriaTemplate } from "../utils";
-// eslint-disable-next-line import/no-internal-modules
```
srietkerk, line 7: Why does moving this import to the top of the file allow for getting rid of this?

teachertool/src/components/CriteriaInstanceDisplay.tsx
```diff
@@ -158,7 +157,10 @@ export const CriteriaInstanceDisplay: React.FC<CriteriaInstanceDisplayProps> = (
                     </span>
                 ))}
             </div>
-            <div className={css["criteria-description"]}>{catalogCriteria.description}</div>
+            <div className={classList(css["criteria-description"], "no-print")}>{catalogCriteria.description}</div>
+            <div className={classList(css["criteria-description"], css["for-print"], "only-print")}>
```
srietkerk, line 161: I don't know much about making webpages printable. Is this a common pattern to an element defined twice, one with classes for printing and one without?

teachertool/src/components/CriteriaResultEntry.tsx
```diff
@@ -23,25 +23,33 @@ interface CriteriaResultNotesProps {
...
+                    label={lf("Feedback")}
+                    title={lf("Write your notes here")}
+                    initialValue={notesInitialValue}
+                    autoResize={true}
+                    onChange={onTextChange}
+                    autoComplete={false}
+                    intervalMs={500}
+                />
+            </div>
+            {notesInitialValue && (
```
srietkerk, line 47: In this case, `notesInitialValue` is checking that notes exist on this criteria, correct? If that's true, I think we can just have the `notesInitialValue` variable just be `notes`.

srietkerk: To clarify, I don't think having `InitialValue` in this variable name adds clarity, so I think having `notes` is good, an alternative could be `currentNotes`.

teachertool/src/components/CriteriaInstanceDisplay.tsx
eanders-ms: We disabled this eslint warning in config some time ago, so this line wasn't needed anymore.

teachertool/src/components/CriteriaInstanceDisplay.tsx
```diff
@@ -158,7 +157,10 @@ export const CriteriaInstanceDisplay: React.FC<CriteriaInstanceDisplayProps> = (
                     </span>
                 ))}
             </div>
-            <div className={css["criteria-description"]}>{catalogCriteria.description}</div>
+            <div className={classList(css["criteria-description"], "no-print")}>{catalogCriteria.description}</div>
```
thsparks, line 160: For the criteria-description, is the only difference between the two the italics? If so, could that be done with a single control inside a `@media print` section of the css?

teachertool/src/components/CriteriaInstanceDisplay.tsx
eanders-ms: Just seemed like a clean way to implement, given how we've setup the media queries. Open to other suggestions!

thsparks (approved):
👍

teachertool/src/components/CriteriaResultEntry.tsx
eanders-ms: I'm not quite sure what you're saying. Is this in regard to the name `notesInitialValue` itself? I chose this name to mesh cleanly with its original use: passing it DebouncedTextarea's `initialValue` prop. I factored the value out of that assignment for use in multiple places.

teachertool/src/components/CriteriaInstanceDisplay.tsx
thsparks: If italics is the only difference, putting it on the same control and just using a print media query in the CriteriaInstanceDisplay.module.scss would feel slightly cleaner to me, but I don't feel strongly.

eanders-ms: I generally shy away from sprinkling media queries around css, as they introduce hidden behaviors not apparent in code.

srietkerk: I agree with Thomas that a media query feels cleaner

eanders-ms: Are the "only-print" and "no-print" css classes not clear?

teachertool/src/components/CriteriaResultEntry.tsx
srietkerk: Yes, sorry, just the name. I should've put my comment where the variable was first assigned. 

I like the fact that the name meshes with its original use, but I think because the notes are also being used in a spot (here) where we actually don't care if it's an "initialValue" or not, I think it should be a more general name to reflect that, thus the suggestion to rename the variable name to `currentNotes`.

teachertool/src/components/CriteriaInstanceDisplay.tsx
thsparks: I'd say it's clear enough. Just a personal stylistic preference. If you prefer this approach, that's fine by me!

srietkerk: Ah, I see your point around media queries. 
The "only-print" and "no-print" classes are definitely clear, but it just feels a bit funky to have two divs defined here where the inner content is the same. That said, I think your stance about media queries makes sense, so I'm fine with leaving it as is.

eanders-ms: I see your points. I think I'd rather leverage the system we have rather than make an exception here.

srietkerk: Sounds good!

teachertool/src/components/CriteriaResultEntry.tsx
eanders-ms: `currentNotes` seems less clear to me, as the value will change as the user edits the field. The name `initialNotesValue` makes sense in the screen rendering context where the value is editable, but not so much in the print context, where it can't change. I guess I can split the difference and just call it `notes` as you originally suggested.

srietkerk (approved):
:shipit:

PR #10128, opened 2024-08-28
Teacher Tool: Relative URL for AI Eval (also remove copilot query param)

thsparks:
This is important for ensuring we hit microbit.org instead of makecode.com for authenticated requests (where the auth cookie is associated with microbit.org). It also means we can reuse AuthClient.staticApiAsync(), which is nice.

I did not end up using relative URLs for fetching projects, as they would typically try to access staging instead of prod (depending on how the dev has configured `pxt.cloud.DEV_BACKEND`), but testing with staging projects isn't well supported at the moment. We could change this later, but for now, I think just using `pxt.Cloud.apiRoot` is simplest.

I also removed the copilot query param. It was helpful for debugging and demos, but with auth required, it no longer really works.

### DIFF
--- a/teachertool/src/services/backendRequests.ts
+++ b/teachertool/src/services/backendRequests.ts
@@ -2,7 +2,6 @@ import { Strings } from "../constants";
 import { stateAndDispatch } from "../state";
 import { ErrorCode } from "../types/errorCode";
 import { logError } from "./loggingService";
-import * as auth from "./authClient";
 
 export async function fetchJsonDocAsync<T = any>(url: string): Promise<T | undefined> {
     try {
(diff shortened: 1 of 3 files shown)


teachertool/src/services/backendRequests.ts
```diff
@@ -92,29 +91,15 @@ export async function loadTestableCollectionFromDocsAsync<T>(fileNames: string[]
...
-            method: "POST",
-            headers,
-            body: JSON.stringify(data),
-            credentials: "include",
-        });
-
-        if (!request.ok) {
-            throw new Error(Strings.UnableToReachAI);
+        const request = await pxt.auth.AuthClient.staticApiAsync(url, data, "POST");
+        if (request.statusCode != 200) {
```
srietkerk, line 99:
```suggestion
        if (request.statusCode !== 200) {
```
This might be fine in this scenario, but I always give pause when I see the single equals in JS.

srietkerk (approved):
One nit, but otherwise looks good!

teachertool/src/services/backendRequests.ts
thsparks: Some days I just really miss C# 🙂

eanders-ms: I'm surprised we don't have an `isHttpSuccess(code)` util. You could do something like `Math.floor(request.statusCode / 100) === 2`. :)

eanders-ms (approved):

teachertool/src/services/backendRequests.ts
thsparks: Just went to double check there wasn't one and saw the object returned by staticApiAsync has a success field. I'll switch to that instead.

srietkerk: It looks like `staticApiAsync` actually does that check, @eanders-ms.. So we could technically also just use a check to see if the request was a success or not? I think this check is fine, but just wanted to put that out there. 

https://github.com/microsoft/pxt/blob/26c79ec926881e3b378008c0d37c1aa4925436bc/pxtlib/auth.ts#L616

srietkerk (approved):

PR #10132, opened 2024-08-28
teacher tool: persist user feedback

eanders-ms:
Store user feedback in state so that it isn't cleared when the feedback control unmounts for any reason.
Feedback state will be programmatically cleared when:
* The criteria evaluation returns a different value for `notes`
* The criteria fails to evaluate.

When feedback is given, the text label now changes to "Thanks for your feedback!". Because of this, I removed the toast that appears with the same message.

image

Fixes https://github.com/microsoft/pxt-microbit/issues/5815
Fixes https://github.com/microsoft/pxt-microbit/issues/5831

### DIFF
--- a/teachertool/src/transforms/mergeEvalResult.ts
+++ b/teachertool/src/transforms/mergeEvalResult.ts
@@ -1,6 +1,7 @@
 import { stateAndDispatch } from "../state";
 import { EvaluationStatus } from "../types/criteria";
 import { setEvalResult } from "./setEvalResult";
+import { setUserFeedback } from "./setUserFeedback";
 
 // This will set the outcome and notes for a given criteria instance id, but if the provided value is undefined, it will not change that value.
 export function mergeEvalResult(criteriaInstanceId: string, outcome?: EvaluationStatus, notes?: string) {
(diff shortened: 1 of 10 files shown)


srietkerk (approved):
LGTM!

eanders-ms:
@thsparks I changed the behavior of the control to disallow multiple clicks. If the user clicks "helpful" they will no longer be able to change their response. Though it's fairly easy to reset manually. Running the AI criteria again will likely return slightly different results. Let me know if you preferred the other way. I can put it back.

jwunderl (approved):

thsparks:
> @thsparks I changed the behavior of the control to disallow multiple clicks. If the user clicks "helpful" they will no longer be able to change their response. Though it's fairly easy to reset manually. Running the AI criteria again will likely return slightly different results. Let me know if you preferred the other way. I can put it back.

I think it's okay. I have a habit of accidentally clicking buttons like this and then wanting to reset/change it, which is why I kept the functionality. But it's probably not a universal experience :)

teachertool/src/transforms/mergeEvalResult.ts
```diff
@@ -15,6 +16,10 @@ export function mergeEvalResult(criteriaInstanceId: string, outcome?: Evaluation
         newCriteriaEvalResult.result = outcome;
     }
     if (notes !== undefined) {
+        if (newCriteriaEvalResult.notes !== notes) {
```
thsparks, line 19: I'm wondering if we should do this when the outcome changes too and not notes. Not currently a scenario we have, but if we ever enabled feedback for different criteria it might become applicable.

thsparks: If we do want that, simplest approach is probably just to do this in `setEvalResult` anytime it changes, instead.

thsparks (approved):

PR #10135, opened 2024-08-29
Teacher Tool: Handle Reload on ServiceWorkerRegistered Events

thsparks:
With https://github.com/microsoft/pxt-microbit/pull/5875, we disabled the automatic webapp reload that occurs when a new service worker is registered. We did that because the unexpected reload was causing https://github.com/microsoft/pxt-microbit/issues/5824 and https://github.com/microsoft/pxt-microbit/issues/5822.

Now, however, we have to handle that refresh ourselves. In this change, we listen for the event added in https://github.com/microsoft/pxt/pull/10120 and refresh the iframe (via changing the src) when we get it.

A few details:
1. I had to change the event type for the serviceworkerregistered event. Since this comes from the iframe unprompted, we want "pxthost" rather than "pxteditor" (which is more for responding to requests)

2. I moved the iframe url into the component's state. I'm not sure why I didn't do this in the first place, but I couldn't find or think of a reason not to and it simplified the code a bit.

3. I'm forcing the reload by changing the iframe id. This is a little bit arbitrary as a way to modify the url and trigger a refresh, but it does have the benefit that the iframe driver handling communication will filter out any weird lingering messages from the old id and only focus on the new one, should that ever occur. It creates something of a "clean slate" for the messaging infrastructure.

Fixes https://github.com/microsoft/pxt-microbit/issues/5824
Fixes https://github.com/microsoft/pxt-microbit/issues/5822

### DIFF
--- a/localtypings/pxteditor.d.ts
+++ b/localtypings/pxteditor.d.ts
@@ -453,6 +453,10 @@ declare namespace pxt.editor {
         advanced?: boolean;
     }
 
+    export interface EditorMessageServiceWorkerRegisteredRequest extends EditorMessageRequest {
+        action: "serviceworkerregistered";
+    }
+
     export interface EditorMessageGetToolboxCategoriesResponse {
         categories: pxt.editor.ToolboxCategoryDefinition[];
     }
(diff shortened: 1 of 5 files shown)


srietkerk (approved):
LGTM

jwunderl (approved):

riknoll (approved):

PR #10137, opened 2024-08-29
fix warning highlight resizing

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/5877

when setting the warning message on a block it changes the size of the block (the icon adds width).

this pr makes sure to resize the highlight border post render once the icon is visible.

### DIFF
--- a/webapp/src/blocks.tsx
+++ b/webapp/src/blocks.tsx
@@ -1110,7 +1110,7 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         // clear previous warnings on non-disabled blocks
         this.editor.getAllBlocks(false).filter(b => b.isEnabled()).forEach((b: Blockly.BlockSvg) => {
             b.setWarningText(null, pxtblockly.PXT_WARNING_ID);
-            setHighlightWarning(b, false);
+            setHighlightWarningAsync(b, false);
         });
         let tsfile = file && file.epkg && file.epkg.files[file.getVirtualFileName(pxt.JAVASCRIPT_PROJECT_NAME)];
         if (!tsfile || !tsfile.diagnostics) return;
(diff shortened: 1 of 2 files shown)


webapp/src/blocks.tsx
```diff
@@ -1956,9 +1956,13 @@ function clearTemporaryAssetBlockData(workspace: Blockly.Workspace) {
     forEachImageField(workspace, field => field.clearTemporaryAssetData());
 }
 
-function setHighlightWarning(block: Blockly.BlockSvg, enabled: boolean) {
+async function setHighlightWarning(block: Blockly.BlockSvg, enabled: boolean) {
```
srietkerk, line 1959:
```suggestion
async function setHighlightWarningAsync(block: Blockly.BlockSvg, enabled: boolean) {
```

webapp/src/blocks.tsx
```diff
@@ -1956,9 +1956,13 @@ function clearTemporaryAssetBlockData(workspace: Blockly.Workspace) {
...
     (block.pathObject as PathObject).setHasError(enabled);
     block.setHighlighted(enabled);
```
srietkerk, line 1961: Does setting the highlight and then resizing make it so there's a snapping effect for the highlight? Or is it fast enough that there's no visual difference?

riknoll: there's no visual difference

webapp/src/blocks.tsx
kimprice, line 1959: Should we follow convention here and name it `setHighlightWarningAsync`?

riknoll: i renamed it

jwunderl (approved):

PR #10140, opened 2024-08-30
Teacher Tool: User-Friendly Error Messages

thsparks:
Fixes https://github.com/microsoft/pxt-microbit/issues/5873

While fixing this, I decided to avoid relying on Error throwing/catching to communicate these kinds of errors. Instead, I added a field inside our evaluation result.  This will hopefully help make it clear that the error messages are user-visible and should be understandable text.

I considered just using a blanket error message for all criteria evaluation failures which would essentially just say "Evaluation failed" with no details at all, but as a user, I'd find that frustrating, especially when it's not clear if it's a "me" issue or a product issue. This change is trying to strike a middle-ground, where different criteria can surface different errors if they want to, but we will never end up displaying random, unfiltered exception messages (defaults to a generic "unexpected error occurred" message if it isn't properly handled).

image

I wasn't sure why there are two different definitions of `EvaluationResult`, but I updated them both to be in-sync. @riknoll, is this a  modules thing? Or should I just try to unify them?

### DIFF
--- a/teachertool/src/transforms/runSingleEvaluateAsync.ts
+++ b/teachertool/src/transforms/runSingleEvaluateAsync.ts
@@ -14,6 +14,7 @@ import { setEvalResultOutcome } from "./setEvalResultOutcome";
 import { mergeEvalResult } from "./mergeEvalResult";
 import { setEvalResult } from "./setEvalResult";
 import { setUserFeedback } from "./setUserFeedback";
+import { Strings, Ticks } from "../constants";
 
 function generateValidatorPlan(
     criteriaInstance: CriteriaInstance,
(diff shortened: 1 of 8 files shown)


jwunderl (approved):

teachertool/src/transforms/runSingleEvaluateAsync.ts
```diff
@@ -131,7 +132,7 @@ export async function runSingleEvaluateAsync(criteriaInstanceId: string, fromUse
                 planResult = await runValidatorPlanAsync(plan, loadedValidatorPlans);
             }
 
-            if (planResult) {
+            if (planResult && planResult.executionSuccess) {
```
eanders-ms, line 135: tiny suggestion.
```suggestion
            if (planResult?.executionSuccess) {
```

eanders-ms (approved):

PR #10141, opened 2024-08-30
Teacher Tool: Ensure Run-All Re-Runs Everything

thsparks:
The user interaction flag is what determines if we should bypass the "run if unchanged / modified" check. We just need to pass it through.

I've pushed the toast display code back up to the callers of runSingleEvaluateAsync since we want different behavior in a run-all scenario vs a single run scenario that can't be totally decided via user interaction. 

Fixes https://github.com/microsoft/pxt-microbit/issues/5829

### DIFF
--- a/teachertool/src/transforms/runEvaluateAsync.ts
+++ b/teachertool/src/transforms/runEvaluateAsync.ts
@@ -19,7 +19,7 @@ export async function runEvaluateAsync(fromUserInteraction: boolean) {
     // EvalRequest promises will resolve to true if evaluation completed successfully (regarless of pass/fail).
     // They will only resolve to false if evaluation was unable to complete.
     const evalRequests = teacherTool.checklist.criteria.map(criteriaInstance =>
-        runSingleEvaluateAsync(criteriaInstance.instanceId, false)
+        runSingleEvaluateAsync(criteriaInstance.instanceId, fromUserInteraction)
     );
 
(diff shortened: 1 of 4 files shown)


teachertool/src/transforms/runEvaluateAsync.ts
```diff
@@ -19,7 +19,7 @@ export async function runEvaluateAsync(fromUserInteraction: boolean) {
     // EvalRequest promises will resolve to true if evaluation completed successfully (regarless of pass/fail).
     // They will only resolve to false if evaluation was unable to complete.
     const evalRequests = teacherTool.checklist.criteria.map(criteriaInstance =>
-        runSingleEvaluateAsync(criteriaInstance.instanceId, false)
+        runSingleEvaluateAsync(criteriaInstance.instanceId, fromUserInteraction)
```
eanders-ms, line 22: This will cause multiple toasts to appear, one for each individual criteria evaluated, and then a final toast once all evaluates.

thsparks: It doesn't do that on success. But I do see that happening when there's an error in the criteria. Will address...

thsparks: Pushed the toast-display code back up to the callers of runSingleEvaluateAsync to handle this

eanders-ms (approved):
👍

PR #10144, opened 2024-08-30
teacher tool: add feedback button

eanders-ms:
Added a feedback button, following vscode edu's design:

image

Partially addresses https://github.com/microsoft/pxt-microbit/issues/5835


teachertool/src/components/HeaderBar.tsx
```diff
@@ -151,27 +151,42 @@ export const HeaderBar: React.FC<HeaderBarProps> = () => {
...
+                        labelClassName="min-sm"
+                        leftIcon="xicon feedback"
+                        title={Strings.GiveFeedback}
+                        label={Strings.GiveFeedback}
                         onClick={() => {
-                            pxt.tickEvent(Ticks.UserMenuSignIn);
-                            showModal({ modal: "sign-in" });
+                            pxt.tickEvent(Ticks.FeedbackForm);
                         }}
+                        href="https://aka.ms/teachertool-feedback"
```
srietkerk, line 165: Does this go anywhere yet? I was trying to visit it just in browser and just kept hitting bing.

teachertool/src/components/HeaderBar.tsx
```diff
@@ -151,27 +151,42 @@ export const HeaderBar: React.FC<HeaderBarProps> = () => {
...
-                {teacherTool.userProfile ? (
-                    <MenuDropdown
-                        id="profile-dropdown"
-                        items={items}
-                        label={avatarElem || initialsElem}
-                        title={lf("Profile Settings")}
-                    />
-                ) : (
+            <>
+                <div>
```
srietkerk, line 155: Is this div here in case we want to add stuff in the future?

srietkerk (approved):
Couple comments, not blocking.

teachertool/src/components/HeaderBar.tsx
eanders-ms: It wraps the button element so that reflow uses div-centric layout semantics. I just find it simpler to use divs for all layout. Mixing elements introduces a mishmash of different behavior, leading to more complicated css.

teachertool/src/components/HeaderBar.tsx
eanders-ms: Nope! We still have to create the survey form.

teachertool/src/components/HeaderBar.tsx
srietkerk: Makes sense, thanks!

PR #10146, opened 2024-09-03
async/awaitify remaining logic in webusb.ts, tweak error message on locked device

jwunderl:
Was walking through this code line by line on friday to try and figure out what was up with chromebook webusb, so async/await-ified the remaining bits along the way. Also added some type info to navigator to remove some unnecessary casting, & made the 'connection failure' dialog a bit more explicit in cases where user gave a device but it didn't work (as that most often comes down to them having it already open in another browser / application, e.g. incognito + normal browser at once).

### DIFF
--- a/pxtlib/webusb.ts
+++ b/pxtlib/webusb.ts
@@ -1,3 +1,23 @@
+// Declare the subset of usb interface we depend on; definitely typed has full but
+// better to list it as optional / use our existing types where possible.
+// https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/w3c-web-usb/index.d.ts
+interface Navigator {
+    readonly usb?: {
+        getDevices(): Promise<pxt.usb.USBDevice[]>;
+        requestDevice(options?: pxt.usb.USBDeviceRequestOptions): Promise<pxt.usb.USBDevice>;
+        addEventListener(
+            type: "connect" | "disconnect",
+            listener: (ev: pxt.usb.USBConnectionEvent) => any,
+            useCapture?: boolean,
+        ): void;
+        removeEventListener(
+            type: "connect" | "disconnect",
+            callback: (ev: pxt.usb.USBConnectionEvent) => any,
+            useCapture?: boolean,
+        ): void;
+    }
+}
+
 namespace pxt.usb {
 
     /**
@@ -39,6 +59,15 @@ namespace pxt.usb {
         serialNumber?: string;
     }
 
+    export interface USBDeviceRequestOptions {
+        filters: USBDeviceFilter[];
+        exclusionFilters?: USBDeviceFilter[] | undefined;
+    }
+
+    export interface USBConnectionEvent extends Event {
+        device: USBDevice;
+    }
+
     // this is for HF2
     export let filters: USBDeviceFilter[] = [{
         classCode: 255,
@@ -60,7 +89,6 @@ namespace pxt.usb {
     export type USBDirection = "in" | "out";
 
     export type BufferSource = Uint8Array;
-
 
     export interface USBConfiguration {
         configurationValue: number;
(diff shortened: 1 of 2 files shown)


pxtlib/webusb.ts
```diff
@@ -197,40 +225,38 @@ namespace pxt.usb {
...
         }
 
         disable() {
             if (!this.enabled) return;
 
             this.enabled = false;
             this.log(`unregistering webusb events`);
-            (navigator as any).usb.removeEventListener('disconnect', this.handleUSBDisconnected);
-            (navigator as any).usb.removeEventListener('connect', this.handleUSBConnected);
+            navigator?.usb.removeEventListener('disconnect', this.handleUSBDisconnected);
```
eanders-ms, line 237: `usb` is declared optional in the interface. Do you want to honor that here?
```suggestion
            navigator?.usb?.removeEventListener('disconnect', this.handleUSBDisconnected);
```

pxtlib/webusb.ts
```diff
@@ -421,104 +447,101 @@ namespace pxt.usb {
...
+                this.log("using ctrl pipe");
+            }
+            this.log("claim interface");
+            await dev.claimInterface(iface.interfaceNumber);
+            this.log("device ready");
+            this.lastKnownDeviceSerialNumber = this.dev.serialNumber;
+            this.ready = true;
+            if (isHF2) {
+                // just starting, not waiting on it.
+                this.readLoop();
```
eanders-ms, line 510: For extra clarity on intent here:
```suggestion
                /*await*/ this.readLoop();
```

pxtlib/webusb.ts
```diff
@@ -421,104 +447,101 @@ namespace pxt.usb {
...
-            .then(dev => !!dev)
-            .catch(e => {
-                // user cancelled
-                if (e.name == "NotFoundError")
-                    return undefined;
-                throw e;
-            })
+    export async function pairAsync(): Promise<boolean> {
+        try {
+            const dev = await navigator.usb.requestDevice({
```
eanders-ms, line 518: Could safe continuation be needed here? There are a few other instances in the file as well.
```suggestion
            const dev = await navigator?.usb?.requestDevice({
```

eanders-ms (approved):
Nice cleanup. Made a few minor suggestions.

pxtlib/webusb.ts
jwunderl: Whoops, i intended to put them there yeah, usb feature will be disabled overall if `navigator.usb` is unset so this was for clarity but fixing, thanks!

pxtlib/webusb.ts
```diff
@@ -197,40 +225,38 @@ namespace pxt.usb {
 
             this.enabled = true;
             this.log("registering webusb events");
-            (navigator as any).usb.addEventListener('disconnect', this.handleUSBDisconnected, false);
-            (navigator as any).usb.addEventListener('connect', this.handleUSBConnected, false);
+            navigator.usb?.addEventListener('disconnect', this.handleUSBDisconnected, false);
```
srietkerk, line 228: Similar to Eric's questions about safe continuation. Does this need 
```suggestion
            navigator?.usb?.addEventListener('disconnect', this.handleUSBDisconnected, false);
```
like there are in other places or is just having `.usb` fine in this case? Same question applies to the line below.

pxtlib/webusb.ts
jwunderl: https://github.com/microsoft/pxt/blob/dev/jwunderl/async-ify-webusb-cleanup/pxtlib/webusb.ts#L608 is where it'll be checked overall, but adding in here for clarity sake ~

pxtlib/webusb.ts
jwunderl: https://github.com/microsoft/pxt/blob/dev/jwunderl/async-ify-webusb-cleanup/pxtlib/webusb.ts#L608 is where we check for it being enabled for the `pxt.usb.isEnabled()` which we use to gate all other calls; but adding in for clarity sake~

srietkerk (approved):

PR #10149, opened 2024-09-03
Teacher tool docs link updated

srietkerk:
Closes https://github.com/microsoft/pxt-microbit/issues/5622

I wanted to bring up a couple of points here, as well. 
1. We have the TODO comment on this line. I'm wondering if we actually want to have an aka.ms link here instead or since the docs are complete we can remove the comment now?
2. The docs link is "teacher tool" even though the tool to the public and even in the title is "Code Evaluation Tool". Should we rename the file to something like `codeevaltool` to accurately reflect it or is having the endpoint as `teachertool` fine?
3. This doesn't have to be addressed now, but I'm also wondering the future of these docs. If we're hoping to ship the code eval tool with other targets in the future, are we going to have different docs in each target so we have the accurate theming in each or are we going to have the docs in a more global space like https://makecode.com/docs


ganicke (approved):
I imagine this should migrate to `./common-docs/teachertool` with images overridden in the target...eventually.

thsparks (approved):
To your questions:
1. TODO comment: I think we can just remove this comment for now.
2. "Teacher tool" vs "Code Evaluation Tool" in link: We probably should adjust this eventually, but it can be a separate change.
3. Future of these docs: I agree with @ganicke's comment on this.

jwunderl (approved):

PR #10157, opened 2024-09-05
fix workspace search plugin CSS in Firefox

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/5902

The workspace search plugin's buttons just use the padding from the user agent's button style. Whoever implemented this the first time around was probably testing with Chrome, which has 6px padding on its buttons by default. Firefox has 4px padding, which results in the error in the issue.

This was an issue in the plugin itself. I have also [filed it on the Blockly team](https://github.com/google/blockly-samples/issues/2458) and [opened a PR](https://github.com/google/blockly-samples/pull/2459) to fix it.


srietkerk (approved):
Nice! Does this mean that when the Blockly team integrates your changes on their side that we can remove this here?

riknoll:
yup

jwunderl (approved):

PR #10158, opened 2024-09-05
detect and warn on webusb stall

jwunderl:
re: https://github.com/microsoft/pxt-microbit/issues/5530#issuecomment-2329825860; with some fiddling you can get your micro:bit into a state where you cannot read from it, and in that scenario the device appears to just be stalled / unrecoverable without power cycling it. This detects the cases where we call into connections from the webapp side -- when refreshing the page, downloading, or just 'pairing'.

The main case this doesn't cover is when you just plug in the micro:bit without downloading; for now it'll still wobble on the status & only show error when they press download, because that path happens entirely over on the micro:bit flash.ts side & it'll take a little more work to get that to pop this dialog properly.

### DIFF
--- a/webapp/src/cmds.ts
+++ b/webapp/src/cmds.ts
@@ -141,10 +141,8 @@ export function showReconnectDeviceInstructionsAsync(
     const boardName = pxt.appTarget.appTheme.boardName || lf("device");
     const helpUrl = pxt.appTarget.appTheme.usbDocs;
     const jsx = webusb.renderDisconnectDeviceDialog();
-    const body = lf("Your {0} appears to have stalled; please disconnect any battery and usb connection, and try again.", boardName);
     return confirmAsync({
         header: lf("{0} Connection failed...", boardName),
-        body,
         jsx,
         hasCloseIcon: true,
         hideAgree: true,
--- a/webapp/src/cmds.ts
+++ b/webapp/src/cmds.ts
@@ -135,6 +135,35 @@ function showUploadInstructionsAsync(
     }).then(() => { });
 }
 
+export function showReconnectDeviceInstructionsAsync(
+    confirmAsync: (options: core.PromptOptions) => Promise<number>
+): Promise<void> {
+    const boardName = pxt.appTarget.appTheme.boardName || lf("device");
+    const helpUrl = pxt.appTarget.appTheme.usbDocs;
+    const jsx = webusb.renderDisconnectDeviceDialog();
+    const body = lf("Your {0} appears to have stalled; please disconnect any battery and usb connection, and try again.", boardName);
+    return confirmAsync({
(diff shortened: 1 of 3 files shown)


webapp/src/cmds.ts
```diff
@@ -135,6 +135,33 @@ function showUploadInstructionsAsync(
     }).then(() => { });
 }
 
+export function showReconnectDeviceInstructionsAsync(
+    confirmAsync: (options: core.PromptOptions) => Promise<number>
+): Promise<void> {
+    const boardName = pxt.appTarget.appTheme.boardName || lf("device");
+    const helpUrl = pxt.appTarget.appTheme.usbDocs;
+    const jsx = webusb.renderDisconnectDeviceDialog();
+    return confirmAsync({
```
jwunderl, line 144: image
happy for feedback on wording, just put together quick one to start with

srietkerk: It looks good to me!!

srietkerk (approved):

eanders-ms (approved):

abchatra:
@Jaqster to give feedback on the dialog. We can take the fix later

Jaqster:
Just for comparison sake, this is another fail dialog - 
image

We may want to be consistent in using the ! icon with the "Failed to Connect" title.  And maybe change the wording to something more like "There was a problem downloading to your micro:bit.  Please disconnect your micro:bit from your computer, remove any batteries from your micro:bit, and click to Try Again.  Click "Help" for more information or "Download as File" for drag-and-drop flashing."
And then have the "Try Again" button instead of the Done button.  Do we also want to add the Download File button too?

microbit-carlos:
If I'm not mistaken, the image shown in pxt-microbit when this error is encountered only shows the USB cable being disconnected:
- https://pxt.azureedge.net/blob/dd759c7e39f810c4304c711547c77232fd96b250//static/download/device-forgotten.gif
- https://github.com/microsoft/pxt-microbit/blob/v7.0.40/pxtarget.json#L584



As it is very important for the user to understand the battery pack must be removed as well (as battery pack usage is the main way to trigger this problem) we suggested in https://github.com/microsoft/pxt-microbit/issues/5530#issuecomment-2050146377 an animation like this one:

https://github.com/microsoft/pxt-microbit/assets/29712657/d9c210c3-aede-4e93-9854-05f6a76da0f1

Would it be possible to update the image/animation to show battery removal as well?

jwunderl:
> 
>  web.usb.reset.short.small.mp4 
> Would it be possible to update the image/animation to show battery removal as well?

https://github.com/microsoft/pxt-microbit/pull/5939 added, should be in next patch release

microbit-carlos:
That's great, thanks @jwunderl!

abchatra:
@carlosperate  released to live. Please validate.

PR #10164, opened 2024-09-09
store block comment position in block data

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/5889

one of the additions we made in pxt-blockly was the ability for block comments to serialize their position to the XML when saving workspaces. unfortunately, we can't do the same thing with unforked blockly because there are no extension points in the XML serialization of comments. however, blockly _does_ have a data element on blocks that you can stick arbitrary data into and will get saved in the XML (we already use this for persisting data for some of our field editors)

this pr brings back the saving of the comment position by shoving it into the block's data element. note that this will not fix existing projects from before the blockly upgrade, but i really don't think anyone will notice.

### DIFF
--- a/pxtblocks/plugins/comments/blockComment.ts
+++ b/pxtblocks/plugins/comments/blockComment.ts
@@ -1,6 +1,7 @@
 import * as Blockly from "blockly";
 
 import { TextInputBubble } from "./textinput_bubble";
+import { getBlockDataForField, setBlockDataForField } from "../../fields";
 
 const eventUtils = Blockly.Events;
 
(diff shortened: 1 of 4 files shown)


pxtblocks/plugins/comments/blockComment.ts
```diff
@@ -246,6 +266,15 @@ export class CommentIcon extends Blockly.icons.Icon {
         }
     }
 
+    onPositionChange(): void {
+        if (this.textInputBubble) {
+            const coord = this.textInputBubble.getPositionRelativeToAnchor();
+
+            setBlockDataForField(this.sourceBlock, COMMENT_OFFSET_X_FIELD_NAME, coord.x + "");
```
srietkerk, line 273: Why is the `+ ""` needed for the coordinate?

srietkerk (approved):
One small question, otherwise looks good.

pxtblocks/plugins/comments/blockComment.ts
riknoll: it's a number, needs to be a string

srietkerk: Woah, I didn't know that was a thing, thanks for the info!

PR #10165, opened 2024-09-09
Code eval tool: use toast upon deletion instead of alert

srietkerk:
Closes https://github.com/microsoft/pxt-microbit/issues/5911

I found the alert experience to be really clunky, so, as per @eanders-ms's suggestion, I've changed it so when a user deletes a criteria, they can undo it via a toast that pops up. 

As part of this change, I made it so deleting a criteria just performs a soft deletion so undoing a deletion was possible.

image

Upload target: https://makecode.microbit.org/app/1e6a09bea0071a2dc2a1090b02ea5bce60eb59cf-1d05a9bc99--eval#

### DIFF
--- a/teachertool/src/components/CriteriaResultEntry.tsx
+++ b/teachertool/src/components/CriteriaResultEntry.tsx
@@ -84,6 +84,7 @@ export const UndoDeleteCriteriaButton: React.FC<{ criteriaId: string }> = ({ cri
 
     return (
         <Button
+            className="undo-button"
             title={Strings.Undo}
             label={Strings.Undo}
             onClick={handleUndoClicked}
@@ -107,7 +108,7 @@ const CriteriaResultToolbarTray: React.FC<{ criteriaId: string }> = ({ criteriaI
 
     async function handleDeleteClickedAsync() {
         removeCriteriaFromChecklist(criteriaId);
-        showToast(makeToast("info", Strings.CriteriaDeleted, 5000, <UndoDeleteCriteriaButton criteriaId={criteriaId} />));
+        showToast(makeToast("info", Strings.CriteriaDeleted, 500000, <UndoDeleteCriteriaButton criteriaId={criteriaId} />));
     }
 
     return (
--- a/teachertool/src/components/CriteriaResultEntry.tsx
+++ b/teachertool/src/components/CriteriaResultEntry.tsx
@@ -18,6 +18,7 @@ import { Button } from "react-common/components/controls/Button";
 import { setEvalResult } from "../transforms/setEvalResult";
 import { showToast } from "../transforms/showToast";
 import { makeToast } from "../utils";
+import { readdCriteriaToChecklist } from "../transforms/readdCriteriaToChecklist";
 
 interface CriteriaResultNotesProps {
     criteriaId: string;
@@ -76,6 +77,20 @@ const CriteriaResultError: React.FC<CriteriaResultErrorProps> = ({ criteriaInsta
     );
 };
 
+export const UndoDeleteCriteriaButton: React.FC<{ criteriaId: string }> = ({ criteriaId }) => {
+    const handleUndoClicked = () => {
+        readdCriteriaToChecklist(criteriaId);
+    }
+
+    return (
+        <Button
+            title={Strings.Undo}
+            label={Strings.Undo}
+            onClick={handleUndoClicked}
+        />
+    )
+}
+
 const CriteriaResultToolbarTray: React.FC<{ criteriaId: string }> = ({ criteriaId }) => {
     const { state: teacherTool } = useContext(AppStateContext);
 
@@ -91,9 +106,8 @@ const CriteriaResultToolbarTray: React.FC<{ criteriaId: string }> = ({ criteriaI
     }
 
     async function handleDeleteClickedAsync() {
-        if (confirm(Strings.ConfirmDeleteCriteriaInstance)) {
-            removeCriteriaFromChecklist(criteriaId);
-        }
+        removeCriteriaFromChecklist(criteriaId);
+        showToast(makeToast("info", Strings.CriteriaDeleted, 5000, <UndoDeleteCriteriaButton criteriaId={criteriaId} />));
     }
 
     return (
--- a/teachertool/src/components/CriteriaResultEntry.tsx
+++ b/teachertool/src/components/CriteriaResultEntry.tsx
@@ -19,6 +19,7 @@ import { setEvalResult } from "../transforms/setEvalResult";
 import { showToast } from "../transforms/showToast";
 import { makeToast } from "../utils";
 import { readdCriteriaToChecklist } from "../transforms/readdCriteriaToChecklist";
+import { dismissToast } from "../state/actions";
 
 interface CriteriaResultNotesProps {
     criteriaId: string;
@@ -77,9 +78,13 @@ const CriteriaResultError: React.FC<CriteriaResultErrorProps> = ({ criteriaInsta
     );
 };
 
-export const UndoDeleteCriteriaButton: React.FC<{ criteriaId: string }> = ({ criteriaId }) => {
+const UndoDeleteCriteriaButton: React.FC<{ criteriaId: string, toastId: string }> = ({ criteriaId, toastId }) => {
+    const { dispatch } = useContext(AppStateContext);
     const handleUndoClicked = () => {
         readdCriteriaToChecklist(criteriaId);
+        if (toastId) {
+            dispatch(dismissToast(toastId));
+        }
     }
 
     return (
@@ -108,7 +113,9 @@ const CriteriaResultToolbarTray: React.FC<{ criteriaId: string }> = ({ criteriaI
 
     async function handleDeleteClickedAsync() {
         removeCriteriaFromChecklist(criteriaId);
-        showToast(makeToast("info", Strings.CriteriaDeleted, 500000, <UndoDeleteCriteriaButton criteriaId={criteriaId} />));
+        const toast = makeToast("info", Strings.CriteriaDeleted);
+        toast.jsx = <UndoDeleteCriteriaButton criteriaId={criteriaId} toastId={toast.id} />;
+        showToast(toast);
     }
 
     return (
--- a/teachertool/src/components/EvalResultDisplay.tsx
+++ b/teachertool/src/components/EvalResultDisplay.tsx
@@ -122,6 +122,7 @@ const CriteriaWithResultsTable: React.FC = () => {
         <div className={css["results-list"]}>
             {Object.values(criteria).map(criteriaInstance => {
                 return (
+                    !criteriaInstance.deleted &&
                     <CriteriaResultEntry criteriaId={criteriaInstance.instanceId} key={criteriaInstance.instanceId} />
                 );
             })}
--- a/teachertool/src/style.overrides/Button.scss
+++ b/teachertool/src/style.overrides/Button.scss
@@ -75,3 +75,14 @@
 .pass > .common-button.common-dropdown-button {
     border: 3px solid var(--pxt-success-accent);
 }
+
+.undo-button {
+    width: 3rem;
+    height: 2rem;
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    justify-content: center;
+    background-color: var(--pxt-button-secondary-background);
+    color: var(--pxt-button-secondary-foreground);
+}
--- a/teachertool/src/style.overrides/Button.scss
+++ b/teachertool/src/style.overrides/Button.scss
@@ -83,6 +83,6 @@
     flex-direction: column;
     align-items: center;
     justify-content: center;
-    background-color: var(--pxt-button-secondary-background);
+    background-color: var(--pxt-info-accent-darkened);
     color: var(--pxt-button-secondary-foreground);
 }
--- a/teachertool/src/transforms/readdCriteriaToChecklist.ts
+++ b/teachertool/src/transforms/readdCriteriaToChecklist.ts
@@ -4,7 +4,7 @@ import { setChecklist } from "./setChecklist";
 import { Ticks } from "../constants";
 import { getCriteriaInstanceWithId } from "../state/helpers";
 
-export function removeCriteriaFromChecklist(criteriaInstanceId: string) {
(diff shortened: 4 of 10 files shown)


eanders-ms:
When I delete a criteria, if that criteria was at max (and therefore not able to be added), it doesn't become available again. To repro:

1. Add "uses at least <count> variables" to your checklist
2. Once added, delete that criteria from your checklist
3. Try to add it back again

It shows as Max in the catalog, and cannot be added.

image

teachertool/src/transforms/readdCriteriaToChecklist.ts
```diff
@@ -0,0 +1,28 @@
+import { stateAndDispatch } from "../state";
+import { logDebug } from "../services/loggingService";
+import { setChecklist } from "./setChecklist";
+import { Ticks } from "../constants";
+import { getCriteriaInstanceWithId } from "../state/helpers";
+
+export function readdCriteriaToChecklist(criteriaInstanceId: string) {
```
eanders-ms, line 7: Over in `addCriteriaToChecklist` do you want to look for an existing deleted matching criteria and restore it if found? Would that make the distinction between adding and re-adding unnecessary?

eanders-ms: An alternative might be to permanently delete the criteria instance once the toast disappears.

thsparks, line 7: nit: I saw this as a typo of "read" instead of "re-add". Perhaps "reAdd" or "undelete" would be clearer?

teachertool/src/components/CriteriaResultEntry.tsx
```diff
@@ -91,9 +112,10 @@ const CriteriaResultToolbarTray: React.FC<{ criteriaId: string }> = ({ criteriaI
     }
 
     async function handleDeleteClickedAsync() {
-        if (confirm(Strings.ConfirmDeleteCriteriaInstance)) {
-            removeCriteriaFromChecklist(criteriaId);
-        }
+        removeCriteriaFromChecklist(criteriaId);
+        const toast = makeToast("info", Strings.CriteriaDeleted);
+        toast.jsx = <UndoDeleteCriteriaButton criteriaId={criteriaId} toastId={toast.id} />;
```
thsparks, line 117: As much as I like the sleekness of this design, something to keep in mind is from an accessibility perspective, how can a user navigate to this button without a mouse? If this means we need to treat toasts like modals and have them steal focus, that could be somewhat confusing/irritating. Maybe we could do that only if the toast is "interactive" in some way? Or there may be other workarounds.

Not necessarily a blocker, but good to keep it in mind.

teachertool/src/transforms/readdCriteriaToChecklist.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    const newChecklist = {
+        ...teacherTool.checklist,
+        criteria: allCriteria,
+    };
+
+    setChecklist(newChecklist);
+
+    if (catalogCriteriaId) {
+        pxt.tickEvent(Ticks.RemoveCriteria, { catalogCriteriaId });
```
thsparks, line 26: Is removeCriteria the correct tick to send in this case?

teachertool/src/transforms/removeCriteriaFromChecklist.ts
```diff
@@ -11,10 +11,14 @@ export function removeCriteriaFromChecklist(criteriaInstanceId: string) {
...
     const catalogCriteriaId = instance?.catalogCriteriaId;
+    const allCriteria = [...teacherTool.checklist.criteria];
+    const criteriaIndex = allCriteria.findIndex(c => c.instanceId === criteriaInstanceId);
+    allCriteria[criteriaIndex].deleted = true;
+
 
     const newChecklist = {
         ...teacherTool.checklist,
-        criteria: teacherTool.checklist.criteria.filter(c => c.instanceId !== criteriaInstanceId),
+        criteria: allCriteria,
```
thsparks, line 21: So since we're just setting this flag but never removing the criteria, it this means that when a teacher exports their checklist, it will contain all the deleted checklist items which is somewhat wasteful since there's no way to get them back after the toast is dismissed.

I like Eric's suggestion of removing the checklist once the toast is gone. I also think we should move them out of the main criteria and into a separate `deletedCriteria` list, rather than add a delete flag. This means that, even if they export while the toast is open, the export will only contain the un-deleted criteria. It also means we don't need all the deleted checks within the code (which would be easy to forget when implementing future features).

thsparks: Also, if we do a separate list, it's probably still good to remove items once the toast is dismissed, but not quite as important (assuming it clears when the user closes the session).

teachertool/src/transforms/readdCriteriaToChecklist.ts
thsparks: I think permanently deleting is probably for the best with this approach. Though see https://github.com/microsoft/pxt/pull/10165#discussion_r1752355733 for some thoughts on a slightly different approach that I think would lessen the importance of it (would still ultimately be good to have, though).

teachertool/src/transforms/removeCriteriaFromChecklist.ts
eanders-ms: A downside to a separate list for deleted items is that undoing the deletion wouldn't preserve original location in the active criteria list (unless we stored the original index). If undeletion appended to the active set, it may confuse the user, especially if the end of the list is off the page.

thsparks: Hmm, that's true, but I still think it'd be simpler to store a separate list of { criteria + index } objects than have to filter based on deleted everywhere (or remember to call the appropriate helper function)

teachertool/src/components/CatalogOverlay.tsx
```diff
@@ -80,7 +80,7 @@ const CatalogItem: React.FC<CatalogItemProps> = ({ catalogCriteria, recentlyAdde
     const { state: teacherTool } = useContext(AppStateContext);
 
     const existingInstanceCount = teacherTool.checklist.criteria.filter(
-        i => i.catalogCriteriaId === catalogCriteria.id
+        i => i.catalogCriteriaId === catalogCriteria.id && !i.deleted
```
eanders-ms, line 83: For long-term code maintainability, I recommend adding a helper (in state/helpers.ts) for getting active criteria.

teachertool/src/components/CriteriaResultEntry.tsx
eanders-ms: Good point. I wonder what Outlook does here.

teachertool/src/components/EvalResultDisplay.tsx
```diff
@@ -122,6 +122,7 @@ const CriteriaWithResultsTable: React.FC = () => {
         <div className={css["results-list"]}>
             {Object.values(criteria).map(criteriaInstance => {
```
eanders-ms, line 123: For code clarity: Use helper to get active criteria, then render them.

teachertool/src/transforms/removeCriteriaFromChecklist.ts
eanders-ms: A helper to get active criteria should mitigate this. I agree we don't want to include deleted criteria in the exported JSON.

thsparks: Even with a helper, you have to remember to call it :) Separate list makes it a non-issue as long as we can sort out the index thing. I don't feel that strongly, so I'm happy for Sarah to decide. But personally, I still think a separate list would be slightly cleaner.

teachertool/src/style.overrides/Button.scss
```diff
@@ -75,3 +75,14 @@
...
 }
+
+.undo-button {
+    width: 3rem;
+    height: 2rem;
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    justify-content: center;
+    background-color: var(--pxt-info-accent-darkened);
```
thsparks, line 86: We should probably change these colors based on the type of toast containing them. I believe there are different css classes for each type of toast we can reference.

teachertool/src/components/CriteriaResultEntry.tsx
thsparks, line 117: In `makeToast`, it looks like you added a parameter for jsx, but it seems like we're not using it here. Was there a specific reason for that?

teachertool/src/transforms/removeCriteriaFromChecklist.ts
srietkerk: I would like to continue with the flag approach and use the "get active criteria" helper function. I have been thinking A LOT about the pros and cons of each approach, and I think both have really strong merit. The selling point for me is for the removing and readding scenario of criteria. Both approaches have the same amount of state operations for full removing and newly adding criteria. However, in the new scenario that we are supporting here (removing and readding), the "having a separate list for deleted criteria" would need four state operations along with list operations (two for removing, two for readding) while the flag approach only needs two state operations and no list operations are needed.

teachertool/src/components/CriteriaResultEntry.tsx
srietkerk: Yes. I originally just had the `<UndoDeleteCriteriaButton/>` inside the make toast. However, in order to have clicking the undo button dismiss the toast, I needed to have access to the toast id so I switched to setting the toast's jsx to the button after making the toast. That said, I could see future scenarios where passing in jsx when making the toast would be helpful so I left it in. I can remove it, though, since it's not getting used with this change.

thsparks: I see. I would probably lean towards removing it in that case, at least for now. Either way is fine though.

teachertool/src/transforms/removeCriteriaFromChecklist.ts
thsparks: That's fine, as long as we can get the helper in place and filter deleted criteria out of the serialization.

teachertool/src/style.overrides/Button.scss
srietkerk: Yeah, I think that's a good idea. Can this be done in a follow-up?

teachertool/src/transforms/readdCriteriaToChecklist.ts
srietkerk: Just for clarity, we are now permanently deleting the criteria after the toast goes away. I also wanted to add since this is related to the `max` problem that @eanders-ms expressed above, I removed the check for whether a criteria is deleted on the `CatalogOverlay`. I did so because having the check would allow a user to add and also readd a criteria, thus possibly having two instances of a criteria that is supposed to only have one. Getting rid of the check in the overlay makes it so it still looks like I can't add the "one only" criteria until the criteria is permanently deleted, thus avoiding the ability for a user to add and readd a criteria that can only be used once.

teachertool/src/style.overrides/Button.scss
thsparks: Sure

teachertool/src/transforms/removeCriteriaFromChecklist.ts
```diff
@@ -14,7 +14,10 @@ export function removeCriteriaFromChecklist(criteriaInstanceId: string) {
 
     const newChecklist = {
         ...teacherTool.checklist,
-        criteria: teacherTool.checklist.criteria.filter(c => c.instanceId !== criteriaInstanceId),
+        criteria: teacherTool.checklist.criteria.filter(c =>
+            c.instanceId !== criteriaInstanceId ||
+            c.instanceId === criteriaInstanceId && !c.deleted
```
thsparks, line 19: Hmm, I'm wondering if there's a scenario where we would want this action to work even for non-deleted criteria (in which case we could have a separate action `removeDeletedCriteriaFromChecklist` that checks for deleted before running remove, or just an `onlyIfDeleted` parameter passed into this action). But I can't think of a realistic scenario at the moment so maybe it's fine...

thsparks, line 19: I think the `c.instanceId === criteriaInstanceId` check is redundant. Could just be `c.instanceId !== criteriaInstanceId || !c.deleted`

teachertool/src/types/criteria.ts
```diff
@@ -20,6 +20,7 @@ export interface CriteriaInstance {
     instanceId: string;
     params: CriteriaParameterValue[] | undefined;
     userFeedback?: UserFeedback;
+    deleted?: boolean;
```
thsparks, line 23: I think we should still ensure deleted criteria get filtered out of export/serialization. It's not as big of an issue since it gets removed once the toast goes away, but the user could still export while the toast is around (or more likely, close the browser/tab, so it gets serialized and stored in browser storage), at which point the deleted criteria would be there forever since the delete code won't run on it anymore.

teachertool/src/transforms/reAddCriteriaToChecklist.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    const newChecklist = {
+        ...teacherTool.checklist,
+        criteria: allCriteria,
+    };
+
+    setChecklist(newChecklist);
+
+    if (catalogCriteriaId) {
+        pxt.tickEvent(Ticks.RemoveCriteria, { catalogCriteriaId });
```
thsparks, line 26: My original comment on this is out of date now, but just a note that I think we need to update this tick from Remove to Re-Add.

eanders-ms:
When I open the Add Criteria panel and add a singleton criteria (limited to a single instance), it correctly appears the catalog unselectable and tagged with "Max". But after deleting the criteria, it remains unselectable in the catalog until the Undo toast disappears. This behavior seems incorrect. The catalog entry should immediately become available to add again.

eanders-ms:
I think that soft deletion is complicating things here. Rather than soft delete, if the undo state were to hold onto the deleted criteria (alongside the toast id), it could simply re-add it at its original index when the user clicks Undo. Validation would need to run in the `addCriteria` transform to enforce constraints (Max instances, etc.). This is to handle the scenario where the user adds a new copy of the criteria before triggering Undo. Removing soft delete in favor of holding the deleted criteria in memory for a short time would simplify things a lot.

PR #10166, opened 2024-09-09
fix interleaving of save/sync operations

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/5926

this fixes a bug caused by interleaving save/sync operations when duplicating a project from within the code editor.

we maintain a global array of all the available headers in the current workspace that we pull projects from. currently, that array gets updated in two places:
1.  when we do a sync operation
2. when we start to save a new project (but before the project is actually saved)

if a sync operation gets triggered after we start saving a new project but before that save has actually completed, it can accidentally overwrite the change that was made when we saved a new project.

luckily, it's an easy fix. we just need to wait to add the project to the header array until after it's actually saved.


srietkerk (approved):
Nice!

PR #10168, opened 2024-09-10
Teacher tool: Restrict numerical inputs to no more than two digits

kimprice:
This PR restricts the numerical inputs so that only non-negative integers are allowed. Thanks to @srietkerk and @eanders-ms for all their help!
Fixes https://github.com/microsoft/pxt-microbit/issues/5659, https://github.com/microsoft/pxt-microbit/issues/5652, https://github.com/microsoft/pxt-microbit/issues/5651, https://github.com/microsoft/pxt-microbit/issues/5649, https://github.com/microsoft/pxt-microbit/issues/5648

Upload target: https://makecode.microbit.org/app/39245c391c49d14a19f37fcdfe2949bbb3c6a7aa-dddb50c050--eval#


react-common/components/controls/Input.tsx
```diff
@@ -83,7 +85,11 @@ export const Input = (props: InputProps) => {
     }
 
     const changeHandler = (e: React.ChangeEvent<any>) => {
-        const newValue = (e.target as any).value;
+        let newValue = (e.target as any).value;
+        if (newValue && filter) {
+            const pat = new RegExp(filter);
```
eanders-ms, line 90: I'd suggest moving this RegExp to a state variable so it isn't recreated on each change.

thsparks:
In firefox in the upload target, I'm still able to type in non-numeric characters (behavior seems unchanged from current version). Not sure if the changes just didn't make it into the target or if it's not working for firefox...

image

eanders-ms:
> In firefox in the upload target, I'm still able to type in non-numeric characters (behavior seems unchanged from current version). Not sure if the changes just didn't make it into the target or if it's not working for firefox...
> 

It's possible the target was built from out of date code. With subapps like teacher tool there's a gotcha: before creating the target, you have to rebuild pxt (gulp) to build the production version of the subapp.

react-common/components/controls/Input.tsx
kimprice: Moved it. Does it look okay?

kimprice:
It's working locally. I'll try fixing the upload target with Eric's tip.

kimprice:
@thsparks try it now. I updated the upload target.

react-common/components/controls/Input.tsx
```diff
@@ -59,11 +60,13 @@ export const Input = (props: InputProps) => {
...
         preserveValueOnBlur,
-        options
+        options,
+        filter
     } = props;
 
-    const [value, setValue] = React.useState(undefined);
+    const [value, setValue] = React.useState(initialValue || "");
     const [expanded, setExpanded] = React.useState(false);
+    const [pattern, setPattern] = React.useState(filter ? new RegExp(filter) : undefined);
```
eanders-ms, line 69: Looks good, except I would avoid using the name `pattern` as the input element has an attribute by this name and we may want to expose it at some point. You could call it `filter` if you didn't destructure that value from props. Something like:
```suggestion
    const [filter] = React.useState(opts.filter ? new RegExp(opts.filter) : undefined);
```

srietkerk:
Oh, one thing I just thought about is that this change will allow `0` as the number entered. That said, we say that some criteria aren't evaluated with a 0 entered, but not sure if we should also have that extra lower bound check.

kimprice:
> Oh, one thing I just thought about is that this change will allow `0` as the number entered. That said, we say that some criteria aren't evaluated with a 0 entered, but not sure if we should also have that extra lower bound check.

Our current placeholder is 0, so we'd need to change that too if we wanted to disallow 0. Maybe the placeholder should be 1?

thsparks (approved):
LGTM!

eanders-ms (approved):
Looks good!

PR #10170, opened 2024-09-11
More Parameter Validation and Limitations (Length Limit for AI Prompts)

thsparks:
## The Change
This change started with just adding a limit to the AI prompt length, but as I was adding yet another field to the `CriteriaParameter` type definition (`maxCharacters`) which only really applies to a subset of parameter types, I decided it was time to tidy that up. Hence, a larger change.

Now, each parameter type has a corresponding class which can have its own fields and validate logic. Currently, I've added `maxCharacters` for strings (used for the AI prompt), and `min` and `max` for numbers (tested manually but not used; I mostly wanted a proof of concept around 2 different classes with 2 different types of validation).

Since there's now 100+ lines of "parameter stuff," I've moved them into their own file instead of keeping them in the shared criteria.ts types file.

## The Not-So-Great Parts
There are two icky parts to this change. I've tried to explain via comments, but I'll call them out here as well:
1. We allow parameters to be set with invalid values. This seemed like the least-bad option I could think of that still worked in generalized scenarios. Other ideas I explored were:
   - **Preventing the input**: but imagine a number with min value 10 and you can't type the first digit "1" because it's below the min. Similarly, what if you have 20 already in the box and delete the 0, do we prevent deletion too?
   -  **Allowing the text to be typed, but clearing it afterwards**: this is consistent with how we do it in blocks, but there's no way to see what's causing the error, and we could accidentally lose data if the user writes up a long string and clicks away without realizing that it's invalid.
   - **Allowing the text to be typed and retained in the textbox, but not setting the value behind the scenes**: this causes the UI and state to get out of sync and leads to confusing behavior.

2. We have to re-instantiate the parameter objects after deserializing so they have the correct types (and validate functionality). I don't really know of a way around this when trying to deserialize into variable classes like this, but I'm open to suggestions!

## Also Note...
I'll also note that I considered wrapping the recently-added numeric validation (https://github.com/microsoft/pxt/pull/10168) into this system, but I think they work best separately. That change works well to prevent non-numeric inputs entirely (which makes sense in that context), and it can be shared effectively across users of the `Input` component. The validation added here is fairly teacher-tool specific and doesn't prevent input (per the reasoning above), so there's just not much overlap.

## Out of Scope
Lastly, this change intentionally **does not** modify how we display these errors to the user. It's already a bigger change than I'd like, so for now, this just relies on what is already in place for when parameters have invalid values. But we might want to consider adjusting this in the future so the actual error messages are more immediately visible.

## Try It (or not...)
I would love to do an upload target, but sadly any uploaded target references prod docs so we don't actually get the maxCharacters setting 🙁 Instead, here's a gif:

MaxCharacterSample

### DIFF
--- a/teachertool/src/types/criteriaParameters.ts
+++ b/teachertool/src/types/criteriaParameters.ts
@@ -1,0 +1,84 @@
+import { logError } from "../services/loggingService";
+import { CriteriaParameter, CriteriaParameterType } from "./criteria";
+import { ErrorCode } from "./errorCode";
+
+export class BaseStringParameter extends CriteriaParameter {
+    maxCharacters: number | undefined;
+
+    constructor(name: string, type: CriteriaParameterType, paths: string[], defaultValue?: string, maxCharacters?: number) {
+        super(name, type, paths, defaultValue);
+        this.maxCharacters = maxCharacters;
+    }
+
+    override validate(value: any): boolean {
+        return typeof value === "string" && (!this.maxCharacters || value.length <= this.maxCharacters);
+    }
+}
+
+export class StringParameter extends BaseStringParameter {
+    constructor(template: CriteriaParameter, maxCharacters?: number) {
+        super(template.name, "string", template.paths, template.default, maxCharacters);
+    }
+}
+
+export class LongStringParameter extends BaseStringParameter {
+    constructor(template: CriteriaParameter, maxCharacters?: number) {
+        super(template.name, "longString", template.paths, template.default, maxCharacters);
+    }
+}
+
+export class NumberParameter extends CriteriaParameter {
+    min: number | undefined;
+    max: number | undefined;
+
+    constructor(template: CriteriaParameter, min?: number, max?: number) {
+        super(template.name, "number", template.paths, template.default);
+        this.min = min;
+        this.max = max;
+    }
+
+    override validate(value: any): boolean {
+        // Ensure the value is numeric and within the specified range.
+        const num = Number(value);
+        return !isNaN(num) && (!this.min || num >= this.min) && (!this.max || num <= this.max);
+    }
+}
+
+export class BlockParameter extends CriteriaParameter {
+    constructor(template: CriteriaParameter) {
+        super(template.name, "block", template.paths, template.default);
+    }
+}
+
+export class SystemParameter extends CriteriaParameter {
+    key?: string;
+
+    constructor(template: CriteriaParameter, key?: string) {
+        super(template.name, "system", template.paths, template.default);
+        this.key = key;
+    }
+}
+
+export function createSpecificParameter(param: CriteriaParameter, catalogCriteriaId: string): CriteriaParameter {
+    switch (param.type) {
+        case "string":
+            return new StringParameter(param, (param as StringParameter)?.maxCharacters);
+        case "longString":
+            return new LongStringParameter(param, (param as LongStringParameter)?.maxCharacters);
+        case "number":
+            const tempCast = param as NumberParameter;
+            return new NumberParameter(param, tempCast?.min, tempCast?.max);
+        case "block":
+            return new BlockParameter(param);
+        case "system":
+            return new SystemParameter(param, (param as SystemParameter)?.key);
+        default:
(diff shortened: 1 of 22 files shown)


teachertool/src/types/criteriaParameters.ts
```diff
@@ -0,0 +1,139 @@
...
+    paths: string[]; // The json path(s) to update with the parameter value in the catalog criteria.
+
+    constructor(name: string, type: CriteriaParameterType, paths: string[], defaultValue?: string) {
+        this.name = name;
+        this.type = type;
+        this.default = defaultValue;
+        this.paths = paths;
+    }
+
+    validate(value: string): CriteriaParameterValidationResult {
```
kimprice, line 26: Why do we have `value` if we're not using it?

kimprice: Oh I see, we're using it in the subclasses.

kimprice (approved):
LGTM!

teachertool/src/types/criteriaParameters.ts
```diff
@@ -0,0 +1,139 @@
...
+    override validate(value: any): CriteriaParameterValidationResult {
+        // Ensure the value is numeric and within the specified range.
+        const num = Number(value);
+        if (isNaN(num)) {
+            return {
+                valid: false,
+                message: Strings.MustBeANumber,
+            };
+        }
+        if (this.min && num < this.min) {
```
eanders-ms, line 86: Did you want to support a min value of zero here?

thsparks: Ah good catch

teachertool/src/types/criteriaParameters.ts
```diff
@@ -0,0 +1,139 @@
...
+    }
+}
+
+export class BlockParameter extends CriteriaParameter {
+    constructor(template: CriteriaParameter) {
+        super(template.name, "block", template.paths, template.default);
+    }
+}
+
+export class SystemParameter extends CriteriaParameter {
```
eanders-ms, line 108: Just wondering, what is a SystemParameter? Maybe a doc comment would be helpful here.

teachertool/src/types/criteriaParameters.ts
```diff
@@ -0,0 +1,139 @@
...
+// Represents a parameter definition in a catalog criteria.
+export class CriteriaParameter {
+    name: string;
+    type: CriteriaParameterType;
+    default: string | undefined;
+    paths: string[]; // The json path(s) to update with the parameter value in the catalog criteria.
+
+    constructor(name: string, type: CriteriaParameterType, paths: string[], defaultValue?: string) {
+        this.name = name;
+        this.type = type;
```
eanders-ms, line 21: The way inheritance is setup here feels very C#-like. For example, `type` is assigned at runtime. This doesn't allow the code to leverage TypeScript's type system. Assigning a literal value to `type` in the declaration itself would alleviate the need for all the typecasting in `createSpecificParameter`.

eanders-ms: Something like this should get you going in that direction:
```ts
export abstract class CriteriaParameter {
    name: string;
    abstract type: CriteriaParameterType;
```

then

```ts
export class StringParameter extends BaseStringParameter {
    type = "string";
```

eanders-ms: (edited to add `readonly`)

thsparks: Per our conversation, I've removed the classes and just gone with types and a big switch/case within a global function for validation. I also moved validation functions into a separate file, so the types file is just types again.

teachertool/src/types/criteriaParameters.ts
thsparks: Added a comment to explain, but basically, it's a variable value that is set by the system rather than the user. The main scenario for this currently is when we pass the share id as a parameter to the AI criteria. There's no need for the user to enter that into the criteria instance since we have that data already in the state, so we just auto-populate it behind the scenes at eval time.

eanders-ms (approved):
Looks great!

PR #10180, opened 2024-09-12
Teacher Tool: Assorted Small UI Fixes

thsparks:
This change makes a few small UI fixes:

1. Fixes width of criteria on firefox (webkit not supported)

2. Remove "Max" and display a permanent check instead when the user reaches the max allowed count for a criteria.

3. Fix line wrapping for criteria

Fixes https://github.com/microsoft/pxt-microbit/issues/5927
Fixes https://github.com/microsoft/pxt-microbit/issues/5934
Fixes https://github.com/microsoft/pxt-microbit/issues/5851

Upload target: https://makecode.microbit.org/app/38bf04a7d1443044cd34d10b3a6edd51761140f9-09bd2a25d5--eval

### DIFF
--- a/teachertool/src/components/CatalogOverlay.tsx
+++ b/teachertool/src/components/CatalogOverlay.tsx
@@ -40,26 +40,28 @@ interface CatalogItemLabelProps {
     recentlyAdded: boolean;
 }
 const CatalogItemLabel: React.FC<CatalogItemLabelProps> = ({ catalogCriteria, isMaxed, recentlyAdded }) => {
-    const showRecentlyAddedIndicator = recentlyAdded && !isMaxed;
     return (
         <div className={css["catalog-item-label"]}>
             <div className={css["action-indicators"]}>
                 {isMaxed ? (
-                    <span>{Strings.Max}</span>
+                    <i
+                        className="fas fa-check"
+                        title={Strings.MaxReached}
+                    />
                 ) : (
                     <>
                         <i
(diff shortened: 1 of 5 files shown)


teachertool/src/components/CatalogOverlay.tsx
```diff
@@ -40,26 +40,28 @@ interface CatalogItemLabelProps {
...
+                        title={Strings.MaxReached}
+                    />
                 ) : (
                     <>
                         <i
                             className={classList(
                                 "fas fa-check",
                                 css["recently-added-indicator"],
-                                showRecentlyAddedIndicator ? undefined : css["hide-indicator"]
+                                recentlyAdded ? undefined : css["hide-indicator"]
```
kimprice, line 57: I'm confused as to what this is doing if we already have the check mark above.

thsparks: The check mark above is only shown in the "maxed" scenario and has slightly different styling. The check mark here is traded off with the + button when a criteria is not maxed (it's what you see when there's a temporary green check that swaps out with the plus sign).

We could probably combine them into one single check mark component, but all we'd be saving there is lines of code, and I think it'd make things more confusing overall trying to reason about maxed and/or recently added and different styles in each scenario, etc...so I just kept them separate.

kimprice (approved):
Looks good!

eanders-ms (approved):

PR #10184, opened 2024-09-13
[Stable8.5] Port time machine to arcade stable

riknoll:
Ports time machine to arcade's stable branch of pxt-core.

I had to do this by hand because stable8.5 doesn't have the change where pxteditor got converted into modules. That being said, it was a very easy merge since most of this code is new/self contained. This includes all of the various fixes I've made.

Below is an upload target link that is based off of the stable1.12 branch of pxt-arcade (the current live version). The only change between this and live is the time machine code in this PR.

https://arcade.makecode.com/app/c1ced1df4e6cf7b38de44793fc64173aa2a8ab89-95b1257bf1


abchatra:
Should we try to do the entire release instead of just porting this?

riknoll:
@abchatra eh, that makes me a bit more nervous because of the blockly change. arcade has all of the asset stuff in blocks which is much more complicated than anything in the micro:bit or minecraft editors...

i think i'll switch to using beta on stream next week to get some more testing/validation in. if it's looking good, then yeah i say we do a release.

PR #10186, opened 2024-09-14
Add generated name mirrored from vscode-edu

jwunderl:
pxt side glue

Guess one question, we don't fill in on this screen

image

but I could if we wanted? Don't really mind behavior as is, if they click through it generates a name, but don't get in their way if they actually plan to fill it out

image


eanders-ms (approved):
Could we prefill the name but select all text as well? Making it easier to clear with a single keypress.

jwunderl:
implemented that

PR #10188, opened 2024-09-16
support blocks generated from static class methods

riknoll:
Support blocks defined as static methods on classes.

This works fine in the editor, but for some reason when I attempted to add a test case to the decompiler tests I was getting errors. I'll need to do some more investigation to figure out what's different between those two scenarios; this isn't the first time that the decompiler tests have given me grief on things that worked fine. I'm guessing it's an issue with the annotate function.


eanders-ms (approved):
Thank you!

jwunderl (approved):

PR #10192, opened 2024-09-17
Build d.ts for target

jwunderl:
Simple version for now since we just want a prototype, but easy to add list of packages to include / from share link support / etc if we want.

Should look at stripping it so hidden / deprecated apis etc are removed as well, but spent too long staring at this before realizing my local arcade build was just very broken so need to look at other stuff that's more useful for now, basically just stashing this here.

it outputs to separate d.ts files but also combines them into a 'combined.d.ts' at the end. This doesn't kick off a build or anything so need to have served first (which does mean it only takes about a second to run

[combined.d.zip](https://github.com/user-attachments/files/17035102/combined.d.zip)

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -5438,40 +5438,57 @@ export async function buildShareSimJsAsync(parsed: commandParser.ParsedCommand) 
 }
 
 export async function buildCoreDeclarationFiles(parsed: commandParser.ParsedCommand) {
-    const rootPackage = parsed.args[0]
     const cwd = process.cwd();
-    const builtFolder = path.join(cwd, "temp");
+    const builtFolder = path.join(cwd, "temp", "dts");
     nodeutil.mkdirP(builtFolder);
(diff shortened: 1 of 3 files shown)


cli/cli.ts
```diff
@@ -5425,6 +5433,69 @@ export async function buildShareSimJsAsync(parsed: commandParser.ParsedCommand)
...
+            if (!data?.trim()) return;
+            writeDts(fileName, data);
+        },
+        /** cancellation token **/ undefined,
+        /** emitOnlyDtsFiles **/ true,
+        /** customTransformers -> I believe where we should apply culling of deprecated blocks */
+    );
+
+    console.log(`writing combined.d.ts`)
+    fs.writeFileSync(path.join(builtFolder, "combined.d.ts"), combined);
```
eanders-ms, line 5496: If this is a single file for all d.ts, maybe call it `index.d.ts`?

eanders-ms (approved):

PR #10194, opened 2024-09-19
fileworkspace: handle fully null case as well

jwunderl:
one more patch re: version history, wasn't seeing this in local build but installer version seems to show an error sometimes, leaves history without a first entry. will port to microbit stable / rebuild


jwunderl:
merging as it's pretty trivial / need to produce build

thsparks (commented):
Retroactive approve :)

PR #10196, opened 2024-09-20
Teacher Tool: More Telemetry

thsparks:
This change adds a few telemetry events and some code changes to support them:

1. Changing the eval result (including a field indicating whether or not it's manual and whether the previous result was manual, which necessitated adding resultIsManual to the result type. I think this could be helpful in the future too, if we decide not to overwrite manual results when doing bulk evaluate)
2. Changing eval notes (debounced)
3. Run single eval (including hash of the checklist which should help with understanding evals / checklist and usage of pre-made checklists)
4. Run bulk eval
5. Importing a checklist (whether there's an invalid file, a successful import, or closed without doing anything)
6. Loading in new projects
7. Block picker opened & block selected
8. Adjusted logging of opening pre-built checklists so we can include checklist hash (and error reporting if a pre-built checklist is invalid)

There's also a one-line bug fix so we don't show the "Replace existing checklist" warning when the user clicks new checklist for the first time (only consider having an "existing checklist" if there is criteria, regardless of name).

Try it here (check console for tick events): https://makecode.microbit.org/app/8599694db2dc091609db8744031e60b1b68348cd-27cb4c68cb--eval?consoleticks=1

### DIFF
--- a/teachertool/src/App.tsx
+++ b/teachertool/src/App.tsx
@@ -19,6 +19,7 @@ import { SignedOutPanel } from "./components/SignedOutPanel";
 import * as authClient from "./services/authClient";
 import { ErrorCode } from "./types/errorCode";
 import { loadProjectMetadataAsync } from "./transforms/loadProjectMetadataAsync";
+import { Ticks } from "./constants";
 
 export const App = () => {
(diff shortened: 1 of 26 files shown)


eanders-ms (approved):
nice

PR #10199, opened 2024-09-23
[ipad] Fix unreachable content on scroll

kimprice:
Fixes https://github.com/microsoft/pxt-microbit/issues/5706

Upload target: https://makecode.microbit.org/app/85162957a6b556f452ce00deebb0d4425fb22de7-7f0c9fae0c--eval

Note: the behavior is different when using the dev tools to simulate an ipad. To see the true behavior, try the upload target on an actual ipad.

### DIFF
--- a/teachertool/src/components/styling/CatalogOverlay.module.scss
+++ b/teachertool/src/components/styling/CatalogOverlay.module.scss
@@ -112,5 +112,19 @@
                 }
             }
         }
+
+                /* Styles for iPads (portrait mode) */
+        @media (min-width: 768px) and (max-width: 1024px) and (orientation: portrait) {
+            .catalog-list {
+                height: 80%;
+            }
+        }
+
+        /* Styles for iPads (landscape mode) */
+        @media (min-width: 1024px) and (max-width: 1366px) and (orientation: landscape) {
+            .catalog-list {
+                height: 70%;
+            }
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


riknoll:
@kimprice if you have access to a mac, you can also use the xcode ipad simulator to test bugs like this! works well if you don't happen to have an ipad (as i do not)

teachertool/src/components/styling/CatalogOverlay.module.scss
```diff
@@ -111,6 +111,20 @@
                     }
                 }
             }
+        }
+
+                /* Styles for iPads (portrait mode) */
+        @media (min-width: 768px) and (max-width: 1024px) and (orientation: portrait) {
```
riknoll, line 117: where did these numbers come from? is this something that couldn't be done using the breakpoints in `react-common-breakpoints.less`?

eanders-ms: Teacher tool defines a set of breakpoints that mirrors react-common's, with a few added. Please use them: https://github.com/microsoft/pxt/blob/1e22ec7a07a5979d06b050bff407741a6e987c08/teachertool/src/global.scss#L144

Teacher tool doesn't have orientation breaks defined. Does react-common have these available? If not that would something useful to add (either in react-common or here in teacher tool). Something like:

```css
.landscape-only {
    display: initial;
}
.portrait-only {
    display: initial;
}
@media (orentation: landscape) {
    .portrait-only {
        display: none !important;
    }
}
@media (orientation: portrait) {
    .landscape-only {
        display: none !important;
    }
}
```

eanders-ms: @kimprice in case it got missed... I made a suggestion for integrating the orientation breakpoints.

PR #10202, opened 2024-09-25
Teacher Tool: Change block display

thsparks:
### Overview
This changes the inline block display for the "Block used count times" criteria in the teacher tool. I've reused the approach taken by the monaco editor to generate the javascript toolbox entries (meaning some of that code has been refactored into a shared helper method). This approach looks at the attributes for a block (mostly the block string), then breaks it up into labels, parameters, and breaks. Those sections are returned and can be displayed however the caller sees fit.

I cache the information to minimize any loading time (fewer messages to the iframe) and to reduce the likelihood of ending up with a block where we can't get a display (i.e. a rubric with a block loaded from an extension, opened later from a project without that extension installed).

### The part where I explain some weird decisions
There are a few things I don't love about this change, mostly revolving around built-in blocks which lack the attributes used to generate the different "parts". In these cases, the monaco editor uses snippetNames defined on the blocks in [monacoSnippets.ts](https://github.com/microsoft/pxt/blob/master/webapp/src/monacoSnippets.ts). I've opted for something similar, but since the monaco toolbox lacks some of the blocks that the block-based toolbox has (on start, repeat, etc...), I've gone ahead and updated the [blockSnippets.ts](https://github.com/microsoft/pxt/blob/master/webapp/src/blocksSnippets.ts) file instead of trying to reuse the monaco ones. There's some duplication here, but I think this feels like the right place for it regardless.

### The part where I explain some weirder decisions
Some blocks (specifically, some built-in blocks) have multiple entries in the toolbox which result in blocks of the same id. For example, we have different toolbox entries for plus, minus, times, and divide, but they're all the same math_arithmatic block, so we only show one entry in the block picker and it passes on any of those blocks (since BlockExists validation is based on blockId). In an ideal world, I'd change the BlockExist validator to allow us to check for these separately, but that's out of scope for the current teacher tool work. Instead, I've created a workaround where multiple matches are simply combined into one string, concatenated with " | ":
image

Similarly, there are some blocks that only appear in the toolbox once but still have multiple different dropdown options that make it tricky to find a descriptive overall snippet name. (Notably, the "Square Root" and "Round" blocks.) In these, I've gone with a similar approach of just piping together the different operations in the snippet name. I look forward to the day I can change BlockExist and remove all this code 🙂

image

### Free Trial
Upload Target: https://makecode.microbit.org/app/c07bd3857b9afa5e604d68f1d16dcbe2bd536dcd-bb5f0ab19d

### Screenshots

**Before**
image

**After**
image

### DIFF
--- a/teachertool/src/components/styling/CriteriaInstanceDisplay.module.scss
+++ b/teachertool/src/components/styling/CriteriaInstanceDisplay.module.scss
@@ -53,9 +53,10 @@
                     .block-name-segment {
                         margin-left: 0;
                         margin-right: 0.2rem;
+                        line-height: normal;
                     }
                     .block-name-param {
-                        border-radius: 0.5rem;
+                        border-radius: 1rem;
                         padding: 0.2rem 0.4rem;
                         background-color: rgba(0, 0, 0, 0.3);
                     }
--- a/teachertool/src/components/styling/CriteriaInstanceDisplay.module.scss
+++ b/teachertool/src/components/styling/CriteriaInstanceDisplay.module.scss
@@ -49,6 +49,18 @@
                 border: solid 1px var(--pxt-page-foreground);
                 padding: 0.4rem;
 
+                .block-readable-name {
+                    .block-name-segment {
+                        margin-left: 0;
+                        margin-right: 0.2rem;
+                    }
+                    .block-name-param {
+                        border-radius: 0.5rem;
+                        padding: 0.2rem 0.4rem;
+                        background-color: rgba(0, 0, 0, 0.3);
+                    }
+                }
+
                 &.error {
                     border: solid 1px var(--pxt-error-accent);
                     background-color: var(--pxt-error-background);
--- a/teachertool/src/components/styling/CriteriaInstanceDisplay.module.scss
+++ b/teachertool/src/components/styling/CriteriaInstanceDisplay.module.scss
@@ -58,7 +58,7 @@
                     .block-name-param {
                         border-radius: 1rem;
                         padding: 0.2rem 0.4rem;
-                        background-color: rgba(0, 0, 0, 0.3);
+                        background-color: var(--pxt-page-dark-shadow);
                     }
                 }
 
--- a/teachertool/src/transforms/loadReadableBlockName.ts
+++ b/teachertool/src/transforms/loadReadableBlockName.ts
@@ -5,15 +5,13 @@ export async function loadBlockTextParts(blockId: string): Promise<pxt.editor.Bl
     // Check for cached version.
     let readableBlockName = getCachedBlockTextParts(blockId);
     if (readableBlockName) {
-        // TODO thsparks : Uncomment after testing and verify this works.
-        // return Promise.resolve(readableBlockName);
+        return Promise.resolve(readableBlockName);
     }
 
     // Call into editor service & cache result
     readableBlockName = await getBlockTextParts(blockId);
     if (readableBlockName) {
-        // TODO thsparks : Uncomment after testing and verify this works.
-        // cacheBlockTextParts(blockId, readableBlockName);
+        cacheBlockTextParts(blockId, readableBlockName);
     }
 
     return readableBlockName;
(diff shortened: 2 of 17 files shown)


teachertool/src/components/styling/CriteriaInstanceDisplay.module.scss
```diff
@@ -49,6 +49,19 @@
...
+                .block-readable-name {
+                    .block-name-segment {
+                        margin-left: 0;
+                        margin-right: 0.2rem;
+                        line-height: normal;
+                    }
+                    .block-name-param {
+                        border-radius: 1rem;
+                        padding: 0.2rem 0.4rem;
+                        background-color: rgba(0, 0, 0, 0.3);
```
srietkerk, line 61: Should we make or use one of the color variables here? I don't have a strong preference, but am curious if there might be a use case elsewhere for a slightly translucent black.

teachertool/src/transforms/loadReadableBlockName.ts
```diff
@@ -0,0 +1,18 @@
+import { getBlockAsText } from "../services/makecodeEditorService";
+import { cacheBlockAsText, getCachedBlockAsText } from "../services/storageService";
+
+export async function loadBlockAsText(blockId: string): Promise<pxt.editor.BlockAsText | undefined> {
+    // Check for cached version.
+    let readableBlockName = getCachedBlockAsText(blockId);
+    if (readableBlockName) {
+        return Promise.resolve(readableBlockName);
```
srietkerk, line 8: Is this `Promise.resolve` needed here? If a truthy value is returned from this function, doesn't that mean the promise has resolved?

webapp/src/monacoFlyout.tsx
```diff
@@ -253,48 +254,26 @@ export class MonacoFlyout extends data.Component<MonacoFlyoutProps, MonacoFlyout
...
         const isPython = this.props.fileType == pxt.editor.FileType.Python;
 
-        if (parts) {
-            if (params &&
-                parts.filter((p: any) => p.kind == "param").length > params.length) {
-                // add empty param when first argument is "this"
-                params.unshift(null);
-            }
-            parts.forEach((part, i) => {
+        const blockAsText = getBlockAsText(block, params, isPython);
```
srietkerk, line 260: Ooh, this is a very nice improvement!

webapp/src/app.tsx
```diff
@@ -4142,6 +4143,51 @@ export class ProjectView
         return { categories };
     }
 
+    getBlocksWithId(blockId: string): BlockDefinition[] {
+        const matches: BlockDefinition[] = [];
+        for (const advanced of [false, true]) {
```
srietkerk, line 4148: Can you explain what the `advanced` is doing?

srietkerk:
Unrelated to code or anything, I just wanted to say your PR description is one of my favorites that I've seen. In addition to being very readable, the subcategory titles are awesome. I think I'm going to start using "Free Trial" with my upload targets now :smile:.

teachertool/src/transforms/loadReadableBlockName.ts
jwunderl: async functions don't need Promise.resolve like this, yeah

jwunderl (approved):
Nice cleanup as well

webapp/src/app.tsx
thsparks: When getting toolbox categories with the existing getToolboxCategories function, you can either get the "Advanced" categories (i.e. the ones under the Advanced header) or the non-advanced ones. This is just a way to check both.

Jaqster:
> Unrelated to code or anything, I just wanted to say your PR description is one of my favorites that I've seen. In addition to being very readable, the subcategory titles are awesome. I think I'm going to start using "Free Trial" with my upload targets now 😄.

This is the Writer in Thomas coming out.  I like the parts where weird decisions are being explained ;-)

webapp/src/app.tsx
srietkerk: Sweet, thanks!

srietkerk (approved):
LGTM!

PR #10210, opened 2024-09-27
Port to Stable - teacher tool: Change block display

thsparks:
Port of #10202

@jwunderl I'll wait to check this in until after the offline app ships, since I didn't get it in for testing today and it does touch non-teacher-tool code.


jwunderl:
you'll wanna rebase on https://github.com/microsoft/pxt/tree/stable11.1, that's the new stable branch

PR #10212, opened 2024-09-30
Improve sig snippet to display alias statement

sae220:
Before @ganicke 's PR (#10078), alias statements have same reference (or help doc) as original statements have.

old_remove_at

Thanks to his PR, `_removeAtStatement` (alias for `removeAt`) has an original page, but it not altogether good, I guess.
For blocks users, `removeAt` and `_removeAtStatement` looks different statements, but they are essentially the same.
And of course, JavaScript/Python users shouldn't use `_removeAtStatement`.

old_remove_at_statement

So, I suggest to improve sig snippet to display alias statement.
new_remove_at

And anyway `_popStatement`, `_shiftStatement` and `_unshiftStatement` (as `_removeAtStatement`) should have thier reference.

@ganicke @abchatra


riknoll (changes requested):
code looks good! however, please leave the remove-at-statement.md file.

our docs are served "live-ish" from this repo and so changing this now will break existing targets that don't have your change to the help annotation in pxt-core.d.ts. this file can be removed once this change has been propagated to other editors.

ganicke (commented):
The problem I have with combining these is that `pop()` returns a value and `_popStatement()` does not. This will either need to be noted in the common reference page or the seperate reference pages will need to remain.

abchatra:
@ganicke is this good to merge?

ganicke (approved):
I've added my edits for the additional 'statement only' info.

PR #10213, opened 2024-09-30
Support for 3rd-party simulator extensions

eanders-ms:
This change introduces a more robust system for supporting simulator extensions, making it possible to support third-party extensions (though such extensions must be approved by the team).

### Summary of changes

* **New config section in targetconfig.json**
  - Entries in the existing `approvedRepoLib` map have a new subsection called `simx`. This is used by both the client and the backend to specify and serve simulator extensions. See `SimulatorExtensionConfig` in `localtypings/pxtarget.d.ts` for specifics.

* **New config section in pxtarget.json**
  - To support testing new or updated 3rd party simulator extensions before they go live, I've added a `testSimulatorExtensions` section to `pxtarget.json`. This allows simulator extensions to be configured in an upload target.

* **New URL parameter for local testing**
  - To support local simulator extension development, I've added a new URL parameter called `simxdev`. Including this parameter will cause simulator extensions to be loaded from their configured `devUrl` (only applicable when serving the target locally).

* **`pxt serve` changes**
  - Added a new optional parameter to `pxt serve` called `backport`. It takes a port number where the locally running backend is listening (typically 8080). This is used to proxy simulator extension requests to the locally running backend. Simulator extensions are not loadable through `pxt serve` directly due to the complexities of syncing the extension's repo/branch/commit sha. This approach was simpler, and it enabled simultaneous local testing of frontend and backend changes. It works in concert with a new backend setting called `LOCAL_SIM_PORT`, which I will document in the backend repo. I doubt anyone will ever need to use this :)

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -2855,6 +2855,7 @@ export function serveAsync(parsed: commandParser.ParsedCommand) {
             browser: parsed.flags["browser"] as string,
             serial: !parsed.flags["noSerial"] && !globalConfig.noSerial,
             noauth: parsed.flags["noauth"] as boolean || false,
+            backport: parsed.flags["backport"] as number || 0,
         }))
 }
 
@@ -6966,6 +6967,11 @@ ${pxt.crowdin.KEY_VARIABLE} - crowdin key
             noauth: {
                 description: "disable localtoken-based authentication",
                 aliases: ["na"],
+            },
+            backport: {
+                description: "port where the locally running backend is listening.",
+                argument: "backport",
+                type: "number",
             }
         }
     }, serveAsync);
--- a/pxtsim/runtime.ts
+++ b/pxtsim/runtime.ts
@@ -285,11 +285,13 @@ namespace pxsim {
             return isPxtElectron() || isIpcRenderer();
         }
 
+        export function testLocalhost(url: string): boolean {
+            return /^http:\/\/(?:localhost|127\.0\.0\.1|192\.168\.\d{1,3}\.\d{1,3}|[a-zA-Z0-9.-]+\.local):\d+\/?/.test(url) && !/nolocalhost=1/.test(url);
+        }
+
         export function isLocalHost(): boolean {
             try {
-                return typeof window !== "undefined"
-                    && /^http:\/\/(localhost|127\.0\.0\.1):\d+\//.test(window.location.href)
-                    && !/nolocalhost=1/.test(window.location.href);
+                return typeof window !== "undefined" && testLocalhost(window.location.href);
             } catch (e) { return false; }
         }
 
--- a/pxtsim/simdriver.ts
+++ b/pxtsim/simdriver.ts
@@ -21,13 +21,24 @@ namespace pxsim {
         nestedEditorSim?: boolean;
         parentOrigin?: string;
         mpRole?: string;    // multiplayer role: "client", "server", or undefined
+        // `messageSimulators` is @DEPRECATED. Use `simulatorExtensions` instead.
         messageSimulators?: pxt.Map<{
             url: string;
             localHostUrl?: string;
             aspectRatio?: number;
             permanent?: boolean;
         }>;
-        // needed when messageSimulators are used to provide to their frame
+        // Simulator extensions distilled from targetconfig.json's `approvedRepoLib` entry.
+        simulatorExtensions?: pxt.Map<{
+            // Subset of pxt.SimulatorExtensionConfig
+            aspectRatio?: number;
+            permanent?: boolean;
+            index?: string;
+            devUrl?: string;
+            // Additional fields outside of pxt.SimulatorExtensionConfig
+            url?: string; // Computed URL
+        }>;
+        // needed when simulatorExtensions are used to provide to their frame
         userLanguage?: string;
     }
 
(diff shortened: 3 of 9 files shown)


cli/cli.ts
```diff
@@ -2855,6 +2855,7 @@ export function serveAsync(parsed: commandParser.ParsedCommand) {
             browser: parsed.flags["browser"] as string,
             serial: !parsed.flags["noSerial"] && !globalConfig.noSerial,
             noauth: parsed.flags["noauth"] as boolean || false,
+            backport: parsed.flags["backport"] as number || 0,
```
srietkerk, line 2858: Can you give an example of how the backport flag works? I read in your description that it allows for the simulator extensions to talk to the locally running backend, but I'm confused how you can just cast the flag as a number here. It looks like `port` and `wsport` do the same thing.. I guess I'm just not familiar enough with the cli 😅 . Is the format something like `--backport=3000`?

cli/cli.ts
```diff
@@ -6966,6 +6967,11 @@ ${pxt.crowdin.KEY_VARIABLE} - crowdin key
             noauth: {
                 description: "disable localtoken-based authentication",
                 aliases: ["na"],
+            },
+            backport: {
```
srietkerk, line 6971: Should we include an alias off the bat, like `bp`?

pxtsim/simdriver.ts
```diff
@@ -21,13 +21,24 @@ namespace pxsim {
         nestedEditorSim?: boolean;
         parentOrigin?: string;
         mpRole?: string;    // multiplayer role: "client", "server", or undefined
+        // `messageSimulators` is @DEPRECATED. Use `simulatorExtensions` instead.
```
srietkerk, line 24: Just curious if the `@` here does something for documentation?

webapp/src/simulator.ts
```diff
@@ -66,6 +69,28 @@ export function init(root: HTMLElement, cfg: SimulatorConfig) {
         }
     }
 
+    // Map simulator extensions from approved repos
+    const simulatorExtensions: pxt.Map<pxt.SimulatorExtensionConfig> = {};
+    Object.entries(trgConfig?.packages?.approvedRepoLib || {})
+        .map(([k, v]) => ({ k: k, v: v.simx }))
+        .filter(e => !!e.v)
```
srietkerk, line 76: What does the `e` stand for here?

webapp/src/simulator.ts
```diff
@@ -66,6 +69,28 @@ export function init(root: HTMLElement, cfg: SimulatorConfig) {
...
+    Object.entries(pxt.appTarget?.simulator?.testSimulatorExtensions || {})
+        .map(([k, v]) => ({ k: k, v: v as pxt.SimulatorExtensionConfig }))
+        .filter(e => !!e.v)
+        .forEach(e => simulatorExtensions[e.k] = {
+            index: "index.html", // default to index.html
+            aspectRatio: pxt.appTarget.simulator.aspectRatio || 1.22, // fallback to 1.22
+            permanent: true, // default to true
+            ...e.v
+        });
+
```
srietkerk, line 93: Is it okay to do this inline replacement for these? If something bad happens is there a way to recover?

webapp/src/app.tsx
```diff
@@ -1118,11 +1118,12 @@ export class ProjectView
                 pkg.mainEditorPkg().setSimState(k, v)
             },
             editor: this.state.header ? this.state.header.editor : ''
-        })
-        this.forceUpdate(); // we now have editors prepared
-
-        // start blockly load
-        this.loadBlocklyAsync();
+        }).then(() => {
```
srietkerk, line 1121: Can we make this use async/await instead?

pxtsim/runtime.ts
```diff
@@ -285,11 +285,13 @@ namespace pxsim {
             return isPxtElectron() || isIpcRenderer();
         }
 
+        export function testLocalhost(url: string): boolean {
```
srietkerk, line 288: Why can't we just use the browserutils function for this here?

pxtsim/simdriver.ts
```diff
@@ -120,16 +131,45 @@ namespace pxsim {
...
+                    !simx.index ||
+                    !simx.aspectRatio ||
+                    simx.permanent === undefined
+                ) {
+                    return;
+                }
+                // Compute the effective URL
+                if (simXDevMode && simx.devUrl) {
+                    // Use the dev URL if the dev flag is set (and we're on localhost)
+                    simx.url = new URL(simx.index, simx.devUrl).toString();
```
srietkerk, line 159: This is interesting! It looks like `URL` gives some nice safeguards. I wonder if we should use this in more places.

cli/cli.ts
eanders-ms: It handles an uncommon dev scenario, so probably not worth allocating an alias.

pxtsim/runtime.ts
eanders-ms: I tried that! browserutils is not available in pxtsim. You'll find many functions and types are duplicated in pxtsim.

pxtsim/simdriver.ts
eanders-ms: Nothing specific, Some syntax coloring algorithms look for jsDoc-like formatting and will visually accentuate it.

webapp/src/app.tsx
eanders-ms: This is happening in React's `componentDidMount` callback and so I assumed it didn't support async, but some research reveals otherwise! I will try it out.

webapp/src/simulator.ts
eanders-ms: It could stand for `element`, I suppose. There are other names that would have worked equally well (e.g.: `v`). In short callbacks like this where the meaning of the parameter is obvious, concise code tends to be clearer and more readable. It probably has historical origin in for loop index naming: `i`, `x`, etc. There's a practicality to using single-character names as long as the callback method itself is very simple. If the function is complex, spanning many lines and containing conditional code blocks, then descriptive naming would be preferred.

Was something unclear about the use of `e` here?

webapp/src/simulator.ts
eanders-ms: It is correct behavior for test registrations replace existing entries because we may be testing a new version of an existing extension. We must never ship extension registrations in the `testSimulatorExtensions` field. It is only for making upload targets and should always be empty in PRs.

webapp/src/simulator.ts
srietkerk: I think just all of the single-letter variables are a kinda tricky to follow. I also usually associate `e` with events, so that's what I originally thought it was. Could this maybe be `el` instead?

eanders-ms: I'd like to stick with one-letter names in the single-line handlers. How about `v`?

eanders-ms: Hm, `v` doesn't work well due to `v.v`. I'll go with `el`. `el.k` and `el.v` are both descriptive and a little amusing.

srietkerk: Ha, instead of a bug in the code, we have an elk in our code!

eanders-ms: `el` doesn't actually work well either due to it being used in many places elsewhere in the file to refer to a document element. Will try `x` I guess.

eanders-ms:
Test target: https://microbit.staging.pxt.io/app/6682aff4715e69b108f5b555a1949d807a7ca082-832f04f750
Test project (must import URL): https://makecode.microbit.org/_XHCbXc53mH5X

srietkerk (approved):
LGTM

PR #10214, opened 2024-10-01
wait for render before setting blockly theme

riknoll:
Fixes the exception currently reproduceable on minecraft beta


srietkerk (approved):
Nice!

abchatra:
Does this not impact other targets like microbit?

abchatra (approved):

PR #10215, opened 2024-10-02
Some app.tsx cleanup

riknoll:
as i was tracing our tutorial loading code, i went ahead and fixed a few tiny bugs and converted some functions to async/await. i know the async/await changes make it a little difficult to tell what changed, so here's a summary of the few actual behavior changes:

1. removed a redundant call to `compiler.newProjectAsync()` in `internalLoadHeaderAsync`
2. removed a redundant call to `pxt.github.cacheProjectDependenciesAsync()` in `createProjectAsync()`
3. made the github repo db keys in `idbworkspace.ts` case insensitive so that we don't miss the cache when a markdown author is inconsistent with their capitalization (this was happening in the HOC minecraft tutorials)

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1560,7 +1560,7 @@ export class ProjectView
         return this.loadHeaderAsync(this.state.header, this.state.editorState)
     }
 
-    tryCheckTargetVersionAsync(targetVersion: string): Promise<void> {
+    tryCheckTargetVersionAsync(targetVersion: string): Promise<void> | undefined {
         const htv = targetVersion || "0.0.0";
         // a legacy script does not have a version -- or has a major version less
         // than the current version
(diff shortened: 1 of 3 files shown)


webapp/src/app.tsx
```diff
@@ -1677,148 +1683,168 @@ export class ProjectView
...
+                await workspace.restoreFromBackupAsync(h);
+            }
+            await pkg.loadPkgAsync(h.id);
+
+            if (!this.state || this.state.header != h) {
+                this.showPackageErrorsOnNextTypecheck();
+            }
+            simulator.setDirty();
+
+            await this.loadTutorialJresCodeAsync();
```
eanders-ms, line 1699: Do you need to call ` await compiler.applyUpgradesAsync()` here?

riknoll: whoops, good catch!

webapp/src/app.tsx
```diff
@@ -2951,24 +2980,25 @@ export class ProjectView
...
+                cloudUserId: this.getUserProfile()?.id,
+                temporary: options.temporary,
+                tutorial: options.tutorial,
+                extensionUnderTest: options.extensionUnderTest,
+                isSkillmapProject: options.skillmapProject
+            },
+            files
+        );
+
+        await this.loadHeaderAsync(hd, { filters: options.filters }, false);
```
eanders-ms, line 3003: Just curious... does disabling the cloud sync check fix something?

eanders-ms (approved):
Nice cleanup! Had one question inline.

webapp/src/app.tsx
riknoll: oh right, i forgot to list that in the changes! this is a new project, so no need to do a down sync here (it doesn't exist yet)

PR #10217, opened 2024-10-07
Consume new ghtutorial route when loading tutorials from github

riknoll:
This consumes the newly proposed /api/ghtutorial/ endpoint. Opening as a draft PR because this endpoint isn't actually released yet.

When loading a tutorial from github, this takes the response from /api/ghtutorial/ and uses it to populate all of the caches that we maintain in indexeddb. The rest of the markdown loading pipeline is unchanged, the requests it makes simply hit the cache instead of actually making requests to the backend.

This PR also makes it so that we now cache the latest versions of github repos inside indexeddb just like we do everything else. Unlike the other things we cache, the entries for latest versions expire after an hour. I might end up increasing this period, though, since that isn't very much time at all.

I also added a `skipgithubcache=1` query parameter that bypasses all of our local github repo caches to make it easier for content creators to verify their changes.

also tacked on an unrelated change making it so that the `saveblocks` query param is now case insensitive and accepts `saveasblocks` since I always accidentally try that first.

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -256,6 +256,10 @@ class FileGithubDb implements pxt.github.IGithubDb {
     loadTutorialMarkdown(repopath: string, tag?: string): Promise<pxt.github.CachedPackage> {
         return this.loadAsync(repopath, tag, "tutorial", (r, t) => this.db.loadTutorialMarkdown(r, t));
     }
+
+    cacheReposAsync(resp: pxt.github.GHTutorialResponse) {
+        return this.db.cacheReposAsync(resp);
+    }
 }
 
 function searchAsync(...query: string[]) {
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -252,6 +252,10 @@ class FileGithubDb implements pxt.github.IGithubDb {
     loadPackageAsync(repopath: string, tag: string): Promise<pxt.github.CachedPackage> {
         return this.loadAsync(repopath, tag, "pkg", (r, t) => this.db.loadPackageAsync(r, t));
     }
+
+    loadTutorialMarkdown(repopath: string, tag?: string): Promise<pxt.github.CachedPackage> {
+        return this.loadAsync(repopath, tag, "tutorial", (r, t) => this.db.loadTutorialMarkdown(r, t));
+    }
 }
 
 function searchAsync(...query: string[]) {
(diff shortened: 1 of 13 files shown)


riknoll:
updated with the changes I made to the API route. this now also works with tutorials in the docs folder of targets, uses etags, and proxies requests through the CDN.

here's an upload target you can use to test (micro:bit):

https://makecode.microbit.org/app/de79fa186a7b3225ed5aeb161ea70ec7796af70e-ac898cc414

and here's a URL that includes a docs tutorial that has a GitHub dependency:

https://makecode.microbit.org/app/de79fa186a7b3225ed5aeb161ea70ec7796af70e-ac898cc414?nocdn=1#tutorial:/projects/micro-coin

the `?nocdn=1` is a new query param i added to bypass the CDN. required for this example because the CDN hits our production backend which does not have the new API deployed yet. I recommend opening the tutorial link in an incognito window to make sure you aren't hitting any local caches

riknoll:
@microsoft/makecode-engineers this is now ready for review! here is an updated upload target:

https://makecode.microbit.org/app/36c4b10f729c193f650f763ba2e37a23adefdd90-bcbbdbd05c

cli/cli.ts
```diff
@@ -252,6 +252,14 @@ class FileGithubDb implements pxt.github.IGithubDb {
     loadPackageAsync(repopath: string, tag: string): Promise<pxt.github.CachedPackage> {
         return this.loadAsync(repopath, tag, "pkg", (r, t) => this.db.loadPackageAsync(r, t));
     }
+
+    loadTutorialMarkdown(repopath: string, tag?: string): Promise<pxt.github.CachedPackage> {
```
srietkerk, line 256: Just curious why we need a function like this defined in so many places? I see that the one defined in the github namespace is getting used along with db.loadTutorialMarkdown, but what do each of these allow for us to do?

srietkerk: Same question for the `cacheReposAsync` function.

srietkerk (approved):
Just a question for my own understanding. Looks good!

cli/cli.ts
riknoll: it's defined on the githubdb interface so all of the classes that implement that interface need to have it as well! i just propagated it to all of the existing ones we have, i don't actually know what this one in the CLI is used for.

PR #10218, opened 2024-10-07
don't fire type checks on blockly theme changes

riknoll:
fixes https://github.com/microsoft/pxt-minecraft/issues/2557

minecraft is our only editor that actually customizes the blockly theme. setting the theme was causing blockly to fire an event which in turn triggered a type check on the home screen. that type check fails because there is no project open when you're on the home screen


srietkerk (approved):

abchatra:
Thanks!

PR #10220, opened 2024-10-07
Add monkey patch to restore background to minecraft workspace

riknoll:
Fixes https://github.com/microsoft/pxt-minecraft/issues/2552

This rather heinous monkey patch brings back our old image background for the minecraft editor workspace. Currently, this isn't possible without a monkey patch like this or something even hackier like observing the DOM.

I'll open an issue on Blockly to see if they can surface an extension point for customizing the Grid class. Seems like a reasonable thing to want to do!


eanders-ms (approved):
Fun!

riknoll:
issue filed here: https://github.com/google/blockly/issues/8614

PR #10221, opened 2024-10-08
Match punctuation for code card descriptions

kimprice:
Thanks to @riknoll for all the help.
Fixes https://github.com/microsoft/pxt-minecraft/issues/2508

### DIFF
--- a/pxtblocks/codecardRenderer.ts
+++ b/pxtblocks/codecardRenderer.ts
@@ -100,7 +100,13 @@ export function renderCodeCard(card: pxt.CodeCard, options: CodeCardRenderOption
         }
         if (card.description) {
             const descr = div(ct, 'ui description');
-            const shortenedDescription = card.description.split('.')[0] + '.';
+            const regex = /((?:\.\.\.)|[\!\.\?…])/;
+            const match = regex.exec(card.description);
+            let shortenedDescription = card.description + ".";
+            if (match) {
+                const punctuation = match[1];
+                shortenedDescription = card.description.split(punctuation)[0] + punctuation;
+            }
 
             descr.appendChild(document.createTextNode(shortenedDescription));
         }
--- a/pxtblocks/codecardRenderer.ts
+++ b/pxtblocks/codecardRenderer.ts
@@ -100,7 +100,7 @@ export function renderCodeCard(card: pxt.CodeCard, options: CodeCardRenderOption
         }
         if (card.description) {
             const descr = div(ct, 'ui description');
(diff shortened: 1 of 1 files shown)


pxtblocks/codecardRenderer.ts
```diff
@@ -100,7 +100,13 @@ export function renderCodeCard(card: pxt.CodeCard, options: CodeCardRenderOption
         }
         if (card.description) {
             const descr = div(ct, 'ui description');
-            const shortenedDescription = card.description.split('.')[0] + '.';
+            const regex = /((?:\.\.\.)|[\!\.\?…])/;
```
srietkerk, line 103: Nit:
```suggestion
            const regex = /((?:\.{1,3})|[\!\?…])/;
```
However, I don't feel strongly about it. Going with the above suggestion would also mean that we would replace `..` if it was used as punctuation. Not sure if we want that or we want that explicit three periods check.

pxtblocks/codecardRenderer.ts
```diff
@@ -100,7 +100,13 @@ export function renderCodeCard(card: pxt.CodeCard, options: CodeCardRenderOption
         }
         if (card.description) {
             const descr = div(ct, 'ui description');
-            const shortenedDescription = card.description.split('.')[0] + '.';
+            const regex = /((?:\.{1,3})|[\!\?…])/;
```
eanders-ms, line 103: This is a nice improvement, also somewhat English-centric. Are there additional characters we should consider here to support all languages?

eanders-ms: (totally fine to defer additional character support if it isn't straight forward)

kimprice: I'll look at the languages we support in the editor and see what else needs to be added.

kimprice: Yea I wasn't able to find a collection of end-of-sentence punctuation for all languages. This would require going through each language one by one. I'm going to defer this work for now.

riknoll: if we are concerned about this, we could always only enable this behavior for languages we know it works for and then just use css text-overflow for the rest

srietkerk (approved):

pxtblocks/codecardRenderer.ts
kimprice: @riknoll How would I do that?

eanders-ms (approved):
This is a good improvement and fixes the most glaring cases. Multi-language concerns can we addressed separately, if they're an actual concern.

PR #10222, opened 2024-10-09
simulator extensions: apply aspect ratio to simx iframe

eanders-ms:
* Simulator extensions can be configured with a custom `aspectRatio`, but it was getting ignored. Modified `startSimulatorExtension` method to take an `aspectRatio` parameter and apply it to the simulator extension's iframe.
* Sanitize message channel string before using it as a class name in the simulator extension frame.

### DIFF
--- a/pxtsim/runtime.ts
+++ b/pxtsim/runtime.ts
@@ -148,7 +148,7 @@ namespace pxsim {
             return Promise.all(values.map(v => mapper(v)));
         }
 
-        export  function promiseMapAllSeries<T, V>(values: T[], mapper: (obj: T) => Promise<V>): Promise<V[]> {
+        export function promiseMapAllSeries<T, V>(values: T[], mapper: (obj: T) => Promise<V>): Promise<V[]> {
             return promisePoolAsync(1, values, mapper);
         }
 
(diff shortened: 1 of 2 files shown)


pxtsim/runtime.ts
```diff
@@ -311,6 +311,13 @@ namespace pxsim {
             })
             return v;
         }
+
+        // Copied from pxtlib/util.ts
+        export function sanitizeFileName(name: string): string {
+            /* eslint-disable no-control-regex */
+            return name.replace(/[()\\\/.,?*^:<>!;'#$%^&|"@+=«»°{}\[\]¾½¼³²¦¬¤¢£~­¯¸`±\x00-\x1F]/g, '').trim().replace(/\s+/g, '-');
```
srietkerk, line 318: Two questions for this regex. Is there not a way to capture all special characters without something like this? Also, how were you able to type some of these?

srietkerk (approved):
Just a question about the regex, but otherwise looks good

pxtsim/runtime.ts
eanders-ms: This method was copy/pasted from pxtlib/util.ts, because that module is not available in the simulator. I don't know the history of this regex, but wanted to use it for consistency.

eanders-ms: I agree the regex could be written a lot better. It should take the approach of defining what is allowed rather than what is not.

eanders-ms: Your comment made me realize that css has slightly different validation rules than filenames. css names can't start with a number, for example. Pushed an update to fix this.

PR #10225, opened 2024-10-10
inline blockly sourcemaps in development builds

riknoll:
something changed in blockly that broke my old local debugging setup; i used to npm link my local clone of blockly and change their build files to inline the sourcemaps so that i could actually debug the compressed blockly code.

so i decided to solve the issue directly by writing my own script to inline the source maps from the released blockly npm package. i went ahead and added this to the gulpfile for non-production builds so that everyone can enjoy actually debugging blockly in a sane way.

one note: this does make it so that you correctly see the blockly source when stepping through files but it doesn't fix the names of variables that show up in the debugger; you still see the minified variable names. still it's a huge improvement over the old setup which didn't even point you to the right spot in the compressed file when sourcemaps were enabled.

also, in the past inline sourcemaps have caused the chrome debugger to be much slower when we tried using them for the entire app. i haven't noticed a problem with this but if it becomes an issue i can make this an optional thing for local builds.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -336,6 +336,14 @@ function runUglify() {
     return Promise.resolve();
 }
 
+function inlineBlocklySourcemaps() {
+    if (process.env.PXT_ENV === 'production') {
+        return;
+    }
+
+    return exec("node ./scripts/inlineBlocklySourceMaps.js");
+}
+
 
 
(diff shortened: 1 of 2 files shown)


thsparks (approved):
Neat-o ✔️

PR #10233, opened 2024-10-15
Use queueRender rather than render

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/5989

replaces all of our calls to `block.render()` with `block.queueRender()`.

for the linked bug, here's a timeline of what was happening:

1. the duplicate function is invoked
2. appendPrivate is called in Blockly to paste the blocks that are duplicated
3. the root block (in this case a forever block) starts to initialize which causes its children to be added to the render queue
4. each of the child blocks is initialized
5. as part of the function call block's initialization, `block.render()` is called which immediately flushes the render queue
6. the blocks in the render queue other than the function call are not actually rendered because they are marked as still initializing, but they are still removed from the queue
7. the child initialization continues and all of the other blocks except for the math_number are re-added to the queue
8. initialization completes and the final render takes place

the reason for step 7 where only math_number is left out of the render is pure coincidence caused by the order the blocks are initialized in; i'm sure there are many other setups that would have resulted in the same broken behavior. in any case, the fix is simple: queue the render instead of flushing the queue. i went ahead and made the same change everywhere we call `.render()` to prevent other similar bugs from cropping up (and hopefully improve perf slightly)

this definitely deserves some extra testing but i did a cursory pass over the affected blocks and fixed the only issue i encountered (the change to the focusLastEditor function in this PR).

### DIFF
--- a/pxtblocks/composableMutations.ts
+++ b/pxtblocks/composableMutations.ts
@@ -253,7 +253,7 @@ export function initExpandableBlock(info: pxtc.BlocksInfo, b: Blockly.Block, def
 
         updateButtons();
         if (variableInlineInputs) b.setInputsInline(visibleOptions < inlineInputModeLimit);
-        if (!skipRender) (b as Blockly.BlockSvg).render();
+        if (!skipRender) (b as Blockly.BlockSvg).queueRender();
     }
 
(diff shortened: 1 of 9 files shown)


eanders-ms (approved):
nice

PR #10234, opened 2024-10-15
Hide Toolbox in Minecraft Python Hour of Code

thsparks:
### Fixes
Fixes https://github.com/microsoft/pxt-minecraft/issues/2571
This adds a hideToolbox metadata field for tutorials (only supported for monaco at the moment, not sure it makes much sense in a blocks scenario). When present, the toolbox is not shown and the user actually has to type things out.

Fixes https://github.com/microsoft/pxt-minecraft/issues/2595
It also fixes a separate issue (that I initially thought was related to these changes but actually wasn't) with the workspace header, where it was always assuming we were in blocks mode. Now it accepts a property for the workspace id, which allows it to size everything correctly (thus, fixing the border).

### flyoutOnly CSS
In our Minecraft css, we've evidently started using flyoutOnly as a de facto indicator for HOC tutorials, so I've had to include it even when hideToolbox is true (and flyoutOnly is false), otherwise all the styling disappears. Ideally, this css class would be refactored/renamed to something more generic, but I worry doing so will be a big, cross-repo change and could have unexpected side-effects, so I've left it for now.

### Patch to avoid content changes
After chatting with Abhijith, it sounds like we don't want to require any content updates for the existing 2024 HOC tutorials, so I've added a quick patch setting hideToolbox to true automatically for python tutorials with the legacy layout and flyoutOnly set to true (actual flyoutOnly mode is not supported anyway).

### Screenshot
Python:
image

Blocks mode unaffected:
image

Related change: https://github.com/microsoft/pxt-minecraft/pull/2596

### DIFF
--- a/pxtlib/tutorial.ts
+++ b/pxtlib/tutorial.ts
@@ -22,6 +22,12 @@ namespace pxt.tutorial {
             simThemeJson
         } = computeBodyMetadata(body);
 
+        // For python HOC, hide the toolbox (we don't support flyoutOnly mode).
+        if (pxt.BrowserUtils.useOldTutorialLayout() && language === "python" && metadata.flyoutOnly) {
+            metadata.flyoutOnly = false;
+            metadata.hideToolbox = true;
+        }
+
         // noDiffs legacy
         if (metadata.diffs === true // enabled in tutorial
             || (metadata.diffs !== false && metadata.noDiffs !== true // not disabled
@@ -406,7 +412,7 @@ ${code}
         if (metadata.explicitHints !== undefined
             && pxt.appTarget.appTheme
             && pxt.appTarget.appTheme.tutorialExplicitHints)
-            metadata.explicitHints = true;
+                metadata.explicitHints = true;
 
         return { metadata, body };
     }
--- a/webapp/src/marked.tsx
+++ b/webapp/src/marked.tsx
@@ -384,7 +384,7 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
     // Renders inline blocks, such as "||controller: Controller||".
     private renderInlineBlocks(content: HTMLElement) {
         const { parent } = this.props;
-        const hasCategories = !parent.state.tutorialOptions?.metadata?.flyoutOnly;
+        const hasCategories = !parent.state.tutorialOptions?.metadata?.flyoutOnly && !parent.state.tutorialOptions?.metadata?.hideToolbox;
         const inlineBlocks = pxt.Util.toArray(content.querySelectorAll(`:not(pre) > code`))
             .map((inlineBlock: HTMLElement) => {
                 const text = inlineBlock.innerText;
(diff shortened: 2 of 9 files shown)


abchatra:
Nice. Thank you!

abchatra (approved):

pxtlib/tutorial.ts
```diff
@@ -22,6 +22,12 @@ namespace pxt.tutorial {
             simThemeJson
         } = computeBodyMetadata(body);
 
+        // For python HOC, hide the toolbox (we don't support flyoutOnly mode).
+        if (pxt.BrowserUtils.useOldTutorialLayout() && language === "python" && metadata.flyoutOnly) {
+            metadata.flyoutOnly = false;
```
srietkerk, line 27: I think this is breaking the change that I made recently for getting rid of the icons for block snippets for `flyoutOnly` tutorials, based on one of the screenshots you included. To get rid of the icon, I'm checking that the metadata has `flyoutOnly` to true.. 
https://github.com/microsoft/pxt/blob/4f68e8bf51adffe42186b353a01004833e9748d3/webapp/src/marked.tsx#L387
Maybe add, on this line, a check for if `hideToolbox` is true as well, because that would also mean that the tutorial wouldn't have categories.

srietkerk (changes requested):

pxtlib/tutorial.ts
thsparks: Good catch. Made your suggested change and updated the screenshot.

webapp/src/marked.tsx
```diff
@@ -384,7 +384,7 @@ export class MarkedContent extends data.Component<MarkedContentProps, MarkedCont
     // Renders inline blocks, such as "||controller: Controller||".
     private renderInlineBlocks(content: HTMLElement) {
         const { parent } = this.props;
-        const hasCategories = !parent.state.tutorialOptions?.metadata?.flyoutOnly;
+        const hasCategories = !parent.state.tutorialOptions?.metadata?.flyoutOnly && !parent.state.tutorialOptions?.metadata?.hideToolbox;
```
srietkerk, line 387: Thanks for making the change, I think this might fail in the blocks scenario now, though, since you said that the `hideToolbox` flag is only being used in Monaco 😅. 
```suggestion
        const hasCategories = !parent.state.tutorialOptions?.metadata?.flyoutOnly || !parent.state.tutorialOptions?.metadata?.hideToolbox;
```

thsparks: I think it only **has** categories when both of these checks are false, right?

thsparks: Can confirm, blocks look as expected (I think):
image

srietkerk: Oh you are totally right. Sorry about that!

srietkerk: Yep, that looks good!

srietkerk (approved):
Nice change!

PR #10245, opened 2024-10-24
Fix Gray Bar on Tutorial Resize

thsparks:
Fixes https://github.com/microsoft/pxt-minecraft/issues/2612

In the tutorial view, we were using the same `VerticalResizeContainer` component regardless of whether the tutorial was at the top (where you can actually resize the container) or the side (where resizing was disabled), and when switching between the two modes, that container was retaining state. As a result, if the user switched from top-bar to side-bar and then back to top-bar, the resize container would immediately pop back to its former height, but it didn't signal that change to the rest of the elements, so things were not positioning themselves to account for it.

I toyed with a few ways to fix this, but ultimately, I think it's odd for us to keep the `VerticalResizeContainer` around at all when the tutorial is in the side-bar (non-resizable) view. Feels like asking for weird disabled-resize-related bugs in the future. So I've gone with an approach that fully swaps out the container depending on whether or not resizing is enabled. The downside to this is that we actually _do_ want to remember the height the user set when we swap back, so I've added a new state var to track the last user-set height, and then we can set the initial height accordingly if we re-enter resize mode. This required pushing the shouldResize check into state as well.

Finally, since I now have two possible "tutorialWrapper" components, I changed the id to a className. I'm not sure if it's _actually_ an antipattern to give both components the same id when it's guaranteed that only one will be present at any given time, but it seemed like it'd be safer to switch it, just in case.

Upload Target: https://minecraft.makecode.com/app/f9a50f1535f3ae09120d070d8d6a3f68e4e9121f-c606faa4a3
Also tested micro-bit since it's using the same code path: https://makecode.microbit.org/app/967bfc5309fa70bba2e64f4eeb10e21d451746a4-42823891fa

### DIFF
--- a/webapp/src/sidepanel.tsx
+++ b/webapp/src/sidepanel.tsx
@@ -104,7 +104,7 @@ export class Sidepanel extends data.Component<SidepanelProps, SidepanelState> {
         const shouldResize = pxt.BrowserUtils.isTabletSize() || this.props.tutorialSimSidebar;
         if (shouldResize != this.state.shouldResize) {
             let height = this.state.height;
-            if (shouldResize && this.state.height === undefined && this.state.lastResizeHeight !== undefined) {
+            if (shouldResize && !this.state.height && this.state.lastResizeHeight) {
                 height = this.state.lastResizeHeight;
             }
 
@@ -219,7 +219,7 @@ export class Sidepanel extends data.Component<SidepanelProps, SidepanelState> {
             this.props.tutorialSimSidebar && "tutorial-sim"
         );
 
-        const editorSidebarHeight = this.state.shouldResize && this.state.height !== undefined ? `${this.state.height}px` : undefined; // Should we use this regardless of shouldResize?
+        const editorSidebarHeight = this.state.shouldResize && this.state.height ? `${this.state.height}px` : undefined;
 
         const tutorialContainer = tutorialOptions ? <TutorialContainer
             parent={parent}
--- a/webapp/src/sidepanel.tsx
+++ b/webapp/src/sidepanel.tsx
@@ -169,7 +169,7 @@ export class Sidepanel extends data.Component<SidepanelProps, SidepanelState> {
         const shouldResize = pxt.BrowserUtils.isTabletSize() || this.props.tutorialSimSidebar;
         const editorSidebarHeight = shouldResize ? `${this.state.height}px` : undefined;
 
-        const tutorialContainer = <TutorialContainer
+        const tutorialContainer = tutorialOptions ? <TutorialContainer
             parent={parent}
             tutorialId={tutorialOptions.tutorial}
             name={tutorialOptions.tutorialName}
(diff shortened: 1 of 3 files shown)


webapp/src/sidepanel.tsx
```diff
@@ -32,6 +34,7 @@ interface SidepanelProps extends ISettingsProps {
     simSerialActive?: boolean;
     deviceSerialActive?: boolean;
     tutorialOptions?: pxt.tutorial.TutorialOptions;
+    /** An optional tutorial layout mode where the sim is on the left and instructions are at the top, regardless of screen size. */
```
srietkerk, line 37: Is this comment regarding `tutorialSimSidebar`? Just wanted to check in since there isn't other code edited here.

webapp/src/sidepanel.tsx
```diff
@@ -70,8 +78,38 @@ export class Sidepanel extends data.Component<SidepanelProps, SidepanelState> {
...
+        if (this.simPanelResizeObserver && this.simPanelRef) {
+            this.simPanelResizeObserver.unobserve(this.simPanelRef);
+        }
+    }
+
+    private updateShouldResize() {
+        const shouldResize = pxt.BrowserUtils.isTabletSize() || this.props.tutorialSimSidebar;
+        if (shouldResize != this.state.shouldResize) {
+            let height = this.state.height;
+            if (shouldResize && this.state.height === undefined && this.state.lastResizeHeight !== undefined) {
```
srietkerk, line 107: Is  there a reason we want to do the explicit checks for undefined here?

webapp/src/sidepanel.tsx
```diff
@@ -108,20 +146,33 @@ export class Sidepanel extends data.Component<SidepanelProps, SidepanelState> {
...
                     this.simPanelRef?.classList.add("invisibleScrollbar");
             })
-            observer.observe(c);
+            this.simPanelResizeObserver.observe(c);
+        }
+    }
+
+    protected handleSimRef = (c: HTMLDivElement) => {
+        this.simRef = c;
+        if (c && typeof ResizeObserver !== "undefined") {
```
srietkerk, line 162: I notice we do the same check above. Is this something special with ResizeObservers? I would figure the type of a ResizeObserver would always be an object, but I don't know a whole lot about them.

webapp/src/sidepanel.tsx
thsparks: It is. I kept reading the code and going "What the heck is tutorialSimSidebar again?" so I added this. Formatting is weird, but that's because it's a doc comment so it appears nicely in intellisense:

image

webapp/src/sidepanel.tsx
thsparks: I thought height being set to 0 was kind of a weird edge case and figured it was safest to just leave it be, if that ever happened.

webapp/src/sidepanel.tsx
thsparks: I thought it was weird too... I'm not sure if there's actually an environment where they wouldn't exist, but I did the check here since it was done above, and I find the engineers of days past usually know more about this stuff than I do 😄

webapp/src/sidepanel.tsx
srietkerk: Makes sense!

srietkerk: Would we need to account for null here as well, though?

thsparks: aargh okay I may just do a nullish check instead

thsparks: thanks!

srietkerk: No problem! I just wanted to make sure we wouldn't hit some edge case. 

For what it's worth as well, I think this would have been a case where `==` and `!=` would be safe to use.. The different types of equality is something that is still kinda hand wavey to me, though, so I find the truthy check to be a bit easier to read. That said, if you wanted to not cover the 0 case, we could go that route. I wanted to note it in general because I feel like I'm always asking about `===` and I find it ironic that the `==` is needed here.. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#loose_equality_using (the second point in the list)

srietkerk (approved):

PR #10246, opened 2024-10-25
Re-enable intellisense on android

thsparks:
We had disabled this because it didn't work, but now we need it, so I hope it just works? I'm afraid I can only test on my phone. If anyone can try to put it through its paces on a chromebook, that'd be helpful.

Upload target:<s> https://minecraft.makecode.com/app/e42873906d2ffabc19ec09ae999dfd28c7beeedf-7a6367562b</s> https://minecraft.makecode.com/app/8429027c1a6da2fe4ee6f7e271b5680a162dfc7e-cd7ec17b08

The change that disabled it was https://github.com/microsoft/pxt/pull/7099, which references bug https://github.com/microsoft/pxt-microbit/issues/3170. Thankfully, things are clearly better now than they were then, so there's a good chance this is fine. Will still need some testing, though.


riknoll (approved):
should we delete the whatever css the android class is referring to?

Jaqster:
:heart_eyes:

PR #10248, opened 2024-10-29
Add "Hide Done" Option for Tutorials

thsparks:
Addresses https://github.com/microsoft/pxt-minecraft/issues/2539 (though that tutorial will still need a content update).

This adds a "hideDone" flag that can be set on a tutorial to indicate that the Done button should not be shown on the final step.

Also added tests for hideDone and hideToolbox (which I added in https://github.com/microsoft/pxt/commit/9e6dd9e5e5ac372d6a043265dc79af48c0850e55). Not sure how much value they bring beyond parsing validation, but I suppose that's better than nothing!

Vertical Sidebar:
NoDoneVertical

Horizontal Top-bar:
NoDoneHorizontal

### DIFF
--- a/webapp/src/components/tutorial/TutorialContainer.tsx
+++ b/webapp/src/components/tutorial/TutorialContainer.tsx
@@ -23,7 +23,6 @@ interface TutorialContainerProps {
     hasTemplate?: boolean;
     preferredEditor?: string;
     hasBeenResized?: boolean;
-    hideDone?: boolean;
 
     tutorialOptions?: pxt.tutorial.TutorialOptions; // TODO (shakao) pass in only necessary subset
     tutorialSimSidebar?: boolean;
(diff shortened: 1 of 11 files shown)


webapp/src/components/tutorial/TutorialContainer.tsx
```diff
@@ -23,6 +23,7 @@ interface TutorialContainerProps {
     hasTemplate?: boolean;
     preferredEditor?: string;
     hasBeenResized?: boolean;
+    hideDone?: boolean;
```
srietkerk, line 26: I'm a bit torn about this. I see that we created shortcuts for other `tutorialOptions.metadata` fields before, but I'm not sure of the benefit here. Since we're adding `hideDone` to the metadata, could we just grab it from the tutorial options? The `hideDone` flag definitely belongs in the metadata, but it feels a bit different to me than the other metadata fields that we decided to pass directly into the `TutorialContainer`. I don't feel strongly about this, especially for this case, but I wanted to bring it up to have a conversation of if there are things that we want to customize about tutorials in the future, should we add an optional prop to the `TutorialContainer`, or should we leave it in the `tutorialOptions`.

thsparks: I'm okay to just read it from metadata (which is already passed in), so I'll make that change.

I could see a theoretical argument towards making TutorialContainer useable without all our tutorial & tutorial metadata classes, but I don't see that being useful in practicality and it's a moot point anyway since we *also* already pass in the metadata itself.

srietkerk (approved):

PR #10249, opened 2024-10-29
Refactor all logging to go through a logger object

riknoll:
This PR is part one of my effort to add a backdoor for getting console logs from the ingame experience in minecraft.

i guess this is technically a breaking change, because `pxt.log` used to be a function that could be overwritten, but AFAIK we never actually did that so i think we should be good.

### DIFF
--- a/pxtblocks/compiler/util.ts
+++ b/pxtblocks/compiler/util.ts
@@ -25,11 +25,11 @@ export function forEachStatementInput(block: Blockly.Block, cb: (block: Blockly.
 export function printScope(scope: Scope, depth = 0) {
     const declared = Object.keys(scope.declaredVars).map(k => `${k}(${scope.declaredVars[k].id})`).join(",");
     const referenced = scope.referencedVars.join(", ");
-    console.log(`${mkIndent(depth)}SCOPE: ${scope.firstStatement ? scope.firstStatement.type : "TOP-LEVEL"}`)
+    pxt.log(`${mkIndent(depth)}SCOPE: ${scope.firstStatement ? scope.firstStatement.type : "TOP-LEVEL"}`)
     if (declared.length) {
-        console.log(`${mkIndent(depth)}DECS: ${declared}`)
+        pxt.log(`${mkIndent(depth)}DECS: ${declared}`)
     }
-    // console.log(`${mkIndent(depth)}REFS: ${referenced}`)
+    // pxt.log(`${mkIndent(depth)}REFS: ${referenced}`)
     scope.children.forEach(s => printScope(s, depth + 1));
 }
 
--- a/pxtsim/logger.ts
+++ b/pxtsim/logger.ts
@@ -1,0 +1,67 @@
+namespace pxt {
+    export interface Logger {
+        info(...args: any[]): void;
+        log(...args: any[]): void;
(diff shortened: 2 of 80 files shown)


abchatra:
This was quick and awesome.

srietkerk:
> i guess this is technically a breaking change, because `pxt.log` used to be a function that could be overwritten, but AFAIK we never actually did that so i think we should be good.

I don't understand why this is a breaking change. The new logger stuff that you made is completely separate from `pxt.log`, right? And since it's only in pxtsim and will only be used in Minecraft shouldn't the other, generic `pxt.log` be uneffected?

pxtsim/logger.ts
```diff
@@ -0,0 +1,67 @@
+namespace pxt {
```
thsparks, line 1: I take it there's no nice, shared location both pxtsim and pxtlib can consume this from as a single file?

pxtblocks/compiler/util.ts
```diff
@@ -25,11 +25,11 @@ export function forEachStatementInput(block: Blockly.Block, cb: (block: Blockly.
...
     const declared = Object.keys(scope.declaredVars).map(k => `${k}(${scope.declaredVars[k].id})`).join(",");
     const referenced = scope.referencedVars.join(", ");
-    console.log(`${mkIndent(depth)}SCOPE: ${scope.firstStatement ? scope.firstStatement.type : "TOP-LEVEL"}`)
+    pxt.log(`${mkIndent(depth)}SCOPE: ${scope.firstStatement ? scope.firstStatement.type : "TOP-LEVEL"}`)
     if (declared.length) {
-        console.log(`${mkIndent(depth)}DECS: ${declared}`)
+        pxt.log(`${mkIndent(depth)}DECS: ${declared}`)
     }
-    // console.log(`${mkIndent(depth)}REFS: ${referenced}`)
+    // pxt.log(`${mkIndent(depth)}REFS: ${referenced}`)
```
srietkerk, line 32: Why not get rid of these commented out logs? Will we ever reenable them? Or are they just used in dev scenarios and we want to leave them here for quick access?

webapp/src/serial.tsx
```diff
@@ -200,7 +200,7 @@ export class Editor extends srceditor.Editor {
         }
         if (this.isVisible) {
             if (this.isCsvView && !this.receivedCsv) {
-                // have not received a csv yet but in csv view; swap back to console.
+                // have not received a csv yet but in csv view; swap back to pxt.
```
thsparks, line 203: Not 100% sure what this comment is referring to, but my guess is it should still be "console" here?

thsparks (approved):
Nice!

pxtsim/logger.ts
riknoll: nope...

pxtblocks/compiler/util.ts
riknoll: i dunno! i didn't add them, so i'm not touching them :)

riknoll:
@srietkerk this redefines pxt.log

it was previously a variable that pointed to a function, so you could overwrite it. with this change, it's now just a function. you can still overwrite with monkey patching but TS will complain

PR #10252, opened 2024-10-31
let editor extensions contribute experiments

riknoll:
part two of my master plan to allow you to copy logs in minecraft.

allows targets to add experiments to the experiments dialog. in this case, we'll be adding an experiment that copies logs to your clipboard.

also adds support for experiment cards that act like buttons rather than toggles, and passes the showNotification function to editor extensions so that we can show toasts

### DIFF
--- a/cli/buildengine.ts
+++ b/cli/buildengine.ts
@@ -762,12 +762,12 @@ export async function compileWithLocalCompileService(extinfo: pxtc.ExtensionInfo
     const resp = await runDockerCompileAsync(extinfo.compileData);
 
     if (resp.hexfile) {
-        console.log("Compile successful");
+        pxt.log("Compile successful");
     }
     else {
-        console.log("Compile failed");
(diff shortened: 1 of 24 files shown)


srietkerk (approved):
Looks good to me, just have a couple of questions. 

This change is only providing the infrastructure to let the editors add extensions, correct?
When you say that you're planning on copying logs to the clipboard, will it be every log? Will the extension just overload your clipboard?

thsparks (approved):

riknoll:
@thsparks @srietkerk I accidentally left out a commit adding the log level to the Logger interface. please take a look!

@srietkerk yes, this does not include the actual experiment stuff (that pr will be on pxt-minecraft). as for copying every log, let's discuss this in the pxt-minecraft pr.

srietkerk (approved):

thsparks (approved):
LGTM!

PR #10256, opened 2024-11-01
Replace a few buttons with links

riknoll:
fixes https://github.com/microsoft/pxt-minecraft/issues/2483
fixes https://github.com/microsoft/pxt-minecraft/issues/2482

programmatically opening links in minecraft on macos doesn't work for whatever reason. clicking on `<a />` elements does, so converting some buttons over to links. frankly, they should have been links in the first place for accessibility reasons.

this also fixes a css bug that was bothering me in the extension dialog where extension card titles were wrapping.


react-common/components/extensions/ExtensionCard.tsx
```diff
@@ -3,6 +3,7 @@ import { Button } from "../controls/Button";
 import { Card } from "../controls/Card";
 import { LazyImage } from "../controls/LazyImage";
 import { classList } from "../util";
+import { Link } from "../controls/Link";
```
srietkerk, line 6: Should we remove the Button from the import since it's no longer being used here?

srietkerk (approved):

PR #10257, opened 2024-11-01
Only download zips if you have file access

srietkerk:
I forgot that there were parts of this PR (https://github.com/microsoft/pxt/pull/10255) that were actually good fixes that I wanted to keep. I realize now that I could've just reopened  that PR, but here we are. 

We shouldn't let the user download a zip of their projects if they don't have access to the file system, so I added the check to the `Download Zip` button. I moved the `hasFileAccess` function to `BrowserUtils` because of this. If there are more places where we might want to check for file access, we now have the function in a more accessible place. 

The addition of the `inGame` check to `BrowserUtils` is no longer being used, however, there were a couple of scenarios where I almost needed a check like this and I think it's just nice to have in case there are needs for it.

Upload target: https://minecraft.makecode.com/app/d06e69d2392ce4bcc73cf7790a8632c0b38adc0a-8bc3309a1d

### DIFF
--- a/pxtblocks/layout.ts
+++ b/pxtblocks/layout.ts
@@ -226,9 +226,7 @@ export function flow(ws: Blockly.WorkspaceSvg, opts?: FlowOptions) {
 }
 
 export function screenshotEnabled(): boolean {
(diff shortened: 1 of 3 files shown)


eanders-ms (approved):

thsparks (approved):
👍

PR #10261, opened 2024-11-08
prevent workspace from re-importing projects on db reset

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/6002

when we detect that there is no database for the current storage prefix, we attempt to migrate all of the projects from the previous prefix to the new one. this is meant to happen on the first time you load an editor but is also triggered if you happen to reset the workspace using the reset option in the settings gearwheel.

luckily, we already maintain a database that keeps track of which projects have been migrated from pouch db to indexed db. this pr extends that database to also track projects that are migrated from one prefix to another.

### DIFF
--- a/webapp/src/idbworkspace.ts
+++ b/webapp/src/idbworkspace.ts
@@ -202,19 +202,23 @@ async function migrateOldIndexedDbAsync() {
         await pxt.BrowserUtils.clearTutorialInfoDbAsync();
     });
 
+    const previousDbPrefix = "legacy";
+    const currentDbPrefix = getCurrentDBPrefix() || "default";
+
     try {
         await legacyDb.openAsync();
         const currentDb = await getCurrentDbAsync();
 
-        await copyTableEntriesAsync(legacyDb, currentDb, HEADERS_TABLE, true);
-        await copyTableEntriesAsync(legacyDb, currentDb, TEXTS_TABLE, true);
+        await copyTableEntriesAsync(legacyDb, currentDb, HEADERS_TABLE, true, previousDbPrefix, currentDbPrefix);
+        await copyTableEntriesAsync(legacyDb, currentDb, TEXTS_TABLE, true, previousDbPrefix, currentDbPrefix);
     } catch (e) {
         pxt.reportException(e);
     }
 }
 
 async function migratePrefixesAsync() {
-    if (!getCurrentDBPrefix()) return;
+    const currentDbPrefix = getCurrentDBPrefix();
+    if (!currentDbPrefix) return;
 
     const currentVersion = pxt.semver.parse(pxt.appTarget.versions.target);
     const currentMajor = currentVersion.major;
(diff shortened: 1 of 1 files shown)


webapp/src/idbworkspace.ts
```diff
@@ -267,12 +288,36 @@ async function getDbAsync(prefix = "__default") {
     }
 }
 
-async function copyTableEntriesAsync(fromDb: pxt.BrowserUtils.IDBWrapper, toDb: pxt.BrowserUtils.IDBWrapper, storeName: string, dontOverwrite: boolean) {
+async function copyTableEntriesAsync(fromDb: pxt.BrowserUtils.IDBWrapper, toDb: pxt.BrowserUtils.IDBWrapper, storeName: string, dontOverwrite: boolean, fromPrefix: string, toPrefix: string) {
+    const migrationDb = await getMigrationDbAsync();
+
     for (const entry of await fromDb.getAllAsync<any>(storeName)) {
+        const key = migrationDbPrefixUpgradeKey(fromPrefix, toPrefix, entry.id);
+        if (await migrationDb.getAsync(storeName, key)) continue;
```
srietkerk, line 296: Why not have 
```suggestion
        if (await !migrationDb.getAsync(storeName, key)) 
```
and wrap everything underneath it? I feel like I didn't put the `!` in the right place, but just wondering if we can avoid using continue here.

webapp/src/idbworkspace.ts
```diff
@@ -267,12 +288,36 @@ async function getDbAsync(prefix = "__default") {
...
+
+async function populatePrefixMigrationDb(fromDb: pxt.BrowserUtils.IDBWrapper, toDb: pxt.BrowserUtils.IDBWrapper, storeName: string, fromPrefix: string, toPrefix: string) {
+    const migrationDb = await getMigrationDbAsync();
+
+    for (const entry of await fromDb.getAllAsync<any>(storeName)) {
+        const key = migrationDbPrefixUpgradeKey(fromPrefix, toPrefix, entry.id);
+        if (await migrationDb.getAsync(storeName, key)) continue;
+
+        // If this header id was actually migrated, add an entry for it in the migration db
+        if (await toDb.getAsync(storeName, entry.id)) {
```
srietkerk, line 317: Could we combine this and the check above and use short circuiting to avoid this check if we get something from the `migrationDb`? Or do we want to keep it separate in case we want to do something else if the `migrationDb` exists?

riknoll: this check is already skipped if there is an entry from the migrationdb thanks to the continue

webapp/src/idbworkspace.ts
riknoll: this is mostly a stylistic preference, but i think the code is cleaner with the continue

eanders-ms: As a diehard never-nester, I agree the continue is clearer.

srietkerk (approved):

PR #10265, opened 2024-11-13
workaround for minecraft pointerevent bug on chromeos

riknoll:
fixes https://github.com/microsoft/pxt-minecraft/issues/2622

the issue here was that minecraft on chromeos was dropping pointerup events whenever a multitouch gesture was used on the trackpad. blockly tracks what pointers are currently still "down", so if a pointerup event goes missing then it will get stuck forever thinking that it is within a multitouch gesture even when there are no fingers still on the trackpad.

to fix this, i had to monkey patch blockly's gesture class to prevent it from tracking multiple pointers on chromeos. essentially, this patch filters all pointer events for all fingers except whichever went down first. all pointerup events are let through, however, which allows the user to get out of a stuck gesture by simply doing another click or drag on the workspace. hopefully that's the first thing they will intuitively try to do when the workspace gets stuck to the cursor.

a proper fix for this issue would be to investigate why the pointer events are being dropped by minecraft (or the android webview). but that would take longer and require a minecraft release, so this hack will have to do for now. blockly could also stand to be a little more fault-tolerant when it comes to gestures.


pxtlib/browserutils.ts
```diff
@@ -126,6 +126,10 @@ namespace pxt.BrowserUtils {
         return hasNavigator() && /Epiphany/i.test(navigator.userAgent);
     }
 
+    export function isChromeOS(): boolean {
```
abchatra, line 129: Does this check pass for Android webview inside ChromeOS?

pxtblocks/monkeyPatches/gesture.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    Blockly.Gesture.prototype.bindMouseEvents = function (this: PatchedGesture, e: PointerEvent) {
+        if (!this.boundEvents_) this.boundEvents_ = [];
+
+        const createSyntheticEvent = (e: PointerEvent) => {
+            const syntheticEvent = new PointerEvent(e.type, {
+                ...e,
+                clientX: e.clientX,
+                clientY: e.clientY,
```
thsparks, line 38: Why to clientX and clientY have to be respecified here?

thsparks (approved):
One question, otherwise looks okay to me!

pxtblocks/monkeyPatches/gesture.ts
riknoll: pointer events are native objects and i was seeing some strange behavior where certain properties weren't being assigned when using the spread operator. as a safety measure, i went ahead and directly specified everything that is used by Blockly's gesture implementation today

pxtlib/browserutils.ts
riknoll: switched to check for android

srietkerk (approved):

PR #10271, opened 2024-11-16
Inline blocks for tutorial content: agent icon now showing up

srietkerk:
Fixes https://github.com/microsoft/pxt-minecraft/issues/2655

There are blocks categories, in this case the agent category, that use images for their category icon rather than an icon within our font library. This change allows us to support these in the tutorial inline block styles. 

image

Upload target: https://minecraft.makecode.com/app/7e56fe5acb243e068f9ed4aa6a11126b94008019-2870f4b5c3#

### DIFF
--- a/theme/tutorial.less
+++ b/theme/tutorial.less
@@ -281,6 +281,15 @@ span.docs.inlineblock {
         margin-right: 6px;
     }
 
+    span.image-icon {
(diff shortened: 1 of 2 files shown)


thsparks (approved):
Nice 👍

PR #10276, opened 2024-11-19
fix in game safari webview gridpicker height

riknoll:
for whatever reason, the safari webview in minecraft is returning 0 for `window.outerHeight` but a correct number for `window.innerHeight`. this is definitely a browser bug given that the outerHeight should always be a bigger number than innerHeight as the name suggests.

this was causing both the block picker being too tall error and the animal picker being too short error. the difference between the two was that the block picker has the search bar, which was causing the div to be given a negative height whereas the animal picker was getting a height of 0

### DIFF
--- a/pxtblocks/fields/field_gridpicker.ts
+++ b/pxtblocks/fields/field_gridpicker.ts
@@ -367,11 +367,11 @@ export class FieldGridPicker extends Blockly.FieldDropdown implements FieldCusto
             width: paddingContainer.offsetWidth,
             height: paddingContainer.offsetHeight
         };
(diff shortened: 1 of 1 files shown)


thsparks (approved):

srietkerk (approved):

abchatra:
@riknoll  just curious, how does it work in live?

riknoll:
good question! i'll try and debug it

riknoll:
the old one was using one of the old closure APIs, `goog.dom.getViewportSize()`, which is no longer included in blockly. looks like that function was using the document element and getting the client height/width instead of checking the window.

PR #10277, opened 2024-11-19
use anchor element for social links when in game

riknoll:
fixes https://github.com/microsoft/pxt-minecraft/issues/2668

the in-game webview on macos doesn't support any of the window open APIs that the browser provides.`<a>` elements work though, so this PR replaces the social buttons with `<a>` elements.

normally i would just replace these for all platforms since `<a>` elements are usually better for accessibility, but the only way to open a pop up window is with js and i think that's the expected behavior for social buttons. doesn't matter for the in-game experience given that those all open in the browser outside the game anyhow.

### DIFF
--- a/react-common/components/controls/Button.tsx
+++ b/react-common/components/controls/Button.tsx
@@ -2,7 +2,7 @@ import * as React from "react";
 import { classList, ContainerProps, fireClickOnEnter } from "../util";
 
 export interface ButtonViewProps extends ContainerProps {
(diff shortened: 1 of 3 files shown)


abchatra:
@riknoll build failing

srietkerk (approved):

riknoll:
@abchatra i'll fix it, but leave this one out of the release anyhow. i'd like more people to test it first so let's save it for the hotfix

abchatra:
Sounds good.

PR #10281, opened 2024-11-20
fix cdn check for node

riknoll:
@thomasjball brought it to my attention that mkc was failing with a weird error when trying to fetch github repo dependencies. turns out it was crashing on this code because window isn't defined in node.

this will need to be ported over to micro:bit stable


riknoll:
actually, might not need to go to stable. not sure if this code ever got there in the first place

srietkerk (approved):

Issue #10279, opened 2024-11-20
Unable to load tutorials from GitHub in Minecraft editor


amg-12:
**Describe the bug**
It is currently not possible to load Minecraft MakeCode tutorials from a GitHub link.

**To Reproduce**
1. Load a MakeCode tutorial from a GitHub URL in the Minecraft editor:
https://minecraft.makecode.com/?ipc=1&noRunOnX=1#tutorial:github:mojang/educationcontent/computing/unit-2/lesson-1
2. The tutorial will not load and the editor shows an error: "Please check your internet connection and check the tutorial is valid."
3. The browser console shows an error:
```
NotFoundError: Failed to execute 'transaction' on 'IDBDatabase': One of the specified object stores was not found.
Error: Bad HTTP status code: 500 at https://pxt.azureedge.net/api/ghtutorial/https://github.com/mojang/educationcontent/computing/unit-2/lesson-1?cdn=20241120; message: {}
[object Object]
```

**Expected behavior**
The Minecraft editor should load the tutorial.

**Additional context**
The issue started sometime in the last 18 hours. This issue breaks worlds in the Lesson Library (e.g. "Computing: Unit 1").
<img src="https://github.com/user-attachments/assets/7f9f36f9-c7fd-406b-bcbd-c8142c6bf17b" width="203">

amg-12:
Thank you for the fast response!

Jaqster:
Thank you for reporting @amg-12

Ashalls:
@Jaqster @riknoll  Any idea on when this fix will be released? Thank you for being so responsive!

Jaqster:
@Ashalls - fix should be out now.  Can you test and see?

amg-12:
Yes, it is working!

However, it seems to be a partial fix.
This URL works correctly:
https://minecraft.makecode.com/#tutorial:github:mojang/educationcontent/computing/unit-2/lesson-1
However this one does not:
https://minecraft.makecode.com/#tutorial:https://github.com/mojang/educationcontent/computing/unit-2/lesson-1
They are the same tutorial, just with the URLs formatted differently. Previously, both formats would have worked.

riknoll:
can you reply with a complete list of the URLs you're using so that we can make sure they're all covered?

riknoll:
also, i'm not sure where the url in that second example came from

this URL is a 404: https://github.com/mojang/educationcontent/computing/unit-2/lesson-1
[...]

PR #10282, opened 2024-11-20
handle repopaths with github prefix

riknoll:
fixes https://github.com/microsoft/pxt/issues/10279

not sure how we didn't catch this in testing


eanders-ms (approved):

Jaqster:
yeah.  we need to add more 3rd party tutorials in our testing cycles.  I think we only tested tutorials in our repo.

srietkerk (approved):

PR #10284, opened 2024-11-20
move github init after app target setup

riknoll:
noticed this exception in the console; initGithubDb is called before the app target is being set which was throwing an exception. the exception happens in an async context so it wasn't actually causing any errors but it does prevent the indexeddb from purging  expired entries


eanders-ms (approved):

abchatra:
@riknoll does this need to be ported for Minecraft?

PR #10285, opened 2024-11-21
Reroute cloud through IPC channel

riknoll:
no need to review this just yet; mostly just opening this for visibility. adds an alternate auth client that proxies cloud requests to the parent frame. in the future, i'll swap out the frame implementation with one that talks over the IPC channel to minecraft


webapp/src/headerbar.tsx
```diff
@@ -239,7 +239,7 @@ export class HeaderBar extends data.Component<ISettingsProps, {}> {
         const { home, header, tutorialOptions } = this.props.parent.state;
         const isController = pxt.shell.isControllerMode();
         const isNativeHost = cmds.isNativeHost();
-        const hasIdentity = auth.hasIdentity();
+        const hasIdentity = auth.hasIdentity() && !pxt.auth.proxyIdentityThroughIPC();
```
eanders-ms, line 242: The webapp calls auth.hasIdentity() in a lot of places. Should this check be done at a lower level so the behavior is consistent?

riknoll: all this is doing is turning off the login/profile button in the header bar. we want all of the other cloud ui/behavior to be the same as usual, but there's no login/profile because this isn't a makecode account

PR #10291, opened 2024-11-26
allow packages to filter the toolbox

riknoll:
this adds the ability for projects and extensions to filter the toolbox. inspired by https://forum.makecode.com/t/creating-micro-bit-sandbox-restrict-available-blocks/32008

an extension or project can add a "toolboxFilter" entry in `pxt.json` like so:

```json
{
    "toolboxFilter": {
        "blocks": {
            "device_show_leds": "hidden"
        },
        "namespaces": {
            "pins": "hidden"
        }
    }
}
```

the possible values for each block or category are "hidden", "visible", and "disabled", though i don't think disabled actually does anything? i'm just matching the values we have available in our current toolbox filtering for tutorials and afaik disabled is just ignored.

toolbox filters are merged starting with the leaf extensions in the dependency tree and moving upwards to the main package. in other words, the top-level project overrides its dependencies, which override their dependencies, etc.

these filters can all be overridden by the filters that exist today. e.g., tutorial blocks filters will always trump project blocks filters.

### DIFF
--- a/webapp/src/package.ts
+++ b/webapp/src/package.ts
@@ -1001,3 +1001,52 @@ export function getExtensionOfFileName(filename: string) {
 async function getHostCacheAsync(): Promise<pxt.BrowserUtils.IDBObjectStoreWrapper<{id: string, val: string}>> {
     return getObjectStoreAsync(HOSTCACHE_TABLE)
 }
+
+export function getProjectToolboxFilters() {
+    const filters: pxt.editor.ProjectFilters = {};
+
+    const deps = mainPkg.sortedDeps();
+    // sort the dependencies by level and then id. lower level overrides
+    // higher level
+    deps.sort((a, b) => {
+        if (a.level === b.level) {
+            return pxt.U.strcmp(a.id, b.id);
+        }
+        return b.level - a.level;
+    });
+
+    const applyProjectFilter = (projectFilter: pxt.PackageConfig["toolboxFilter"], key: keyof pxt.PackageConfig["toolboxFilter"]) => {
+        if (!projectFilter[key]) {
+            return
+        }
+
+        if (!filters[key]) {
+            filters[key] = {};
+        }
+
+        for (const entry of Object.keys(projectFilter[key])) {
+            const value = String(projectFilter[key][entry]).toLowerCase();
+            if (value === "hidden") {
+                filters[key][entry] = pxt.editor.FilterState.Hidden;
+            }
+            else if (value === "visible") {
+                filters[key][entry] = pxt.editor.FilterState.Visible;
+            }
+            else if (value === "disabled") {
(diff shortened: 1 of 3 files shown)


riknoll:
also, not sure where to document this. maybe [here](https://makecode.com/extensions/pxt-json)? i'm not sure if anyone reads that page though

abchatra:
What does the extension author need to do to filter the toolbox? Can you share a sample extension which does this?

riknoll:
@abchatra it's exactly the same as how you do it for a project! simply add the "toolboxFilter" object in the extension's pxt.json.

here's a project that does it:

https://makecode.microbit.org/_JcxUkW3vAcpT

webapp/src/package.ts
```diff
@@ -1000,4 +1000,53 @@ export function getExtensionOfFileName(filename: string) {
...
+    // sort the dependencies by level and then id. lower level overrides
+    // higher level
+    deps.sort((a, b) => {
+        if (a.level === b.level) {
+            return pxt.U.strcmp(a.id, b.id);
+        }
+        return b.level - a.level;
+    });
+
+    const applyProjectFilter = (projectFilter: pxt.PackageConfig["toolboxFilter"], key: keyof pxt.PackageConfig["toolboxFilter"]) => {
```
srietkerk, line 1018: I just wanted to ask about this because I don't think I've seen `keyof` being used before. I looked at the documentation, but just wanted some clarification. 

So using `keyof` here is basically saying, "hey, the type we are expecting has to be the keys defined within this object. If the string (in our case) or type does not exist within this object, we can't accept that parameter". And the benefit of using something like this is that we can use the object itself as a growing type definition if we ever have any fields that we want to add to the `toolboxFilter` object rather than having to create a separate type definition.

webapp/src/toolboxeditor.tsx
```diff
@@ -17,11 +18,32 @@ export abstract class ToolboxEditor extends srceditor.Editor {
...
+
         if (filters) {
-            // block-level filters should not apply to shadow blocks (nested)
-            const blockFilter = filters.blocks && (filters.blocks[blockId] || (this.blockIdMap && this.blockIdMap[blockId]?.some(id => filters.blocks[id])));
+            let blockFilter: pxt.editor.FilterState | boolean;
+            if (filters.blocks) {
+                if (filters.blocks[blockId] !== undefined) {
+                    blockFilter = filters.blocks[blockId];
+                }
+                else {
```
srietkerk, line 41: Thank you for this change! This is much more readable!

srietkerk (approved):
One small nit and a learning question. Looks good!

webapp/src/package.ts
riknoll: yup!

PR #10303, opened 2024-12-04
Teacher tool: Show placeholder text when no project is loaded

thsparks:
This was some low hanging fruit we discussed after the customer interviews which I didn't get around to fixing until just now. Since I don't want to spend too long on it, I just done a "bare minimum" implementation. Haven't put together any nice graphics or anything like that, but I think it's still an improvement.

I've also increased the contrast for placeholder text in the input field (a11y issue and also came up multiple times in the interviews) and fixed a null ref exception that came up in testing.

Upload target: https://makecode.microbit.org/app/910ba06ba32792294020a356752fa127895f5f68-329df429fe--eval

image

### DIFF
--- a/teachertool/src/components/styling/MakecodeFramePlaceholder.module.scss
+++ b/teachertool/src/components/styling/MakecodeFramePlaceholder.module.scss
@@ -1,0 +1,22 @@
+.makecode-frame-placeholder {
+    width: 100%;
+    height: 100%;
+    margin: 0;
+    padding: 0;
+
+    align-items: center;
+    justify-content: center;
+    display: flex;
+    flex-direction: column;
+    gap: 1rem;
+
+    color: var(--pxt-page-foreground);
+    opacity: 0.65;
+    font-size: 2rem;
+
+    text-align: center;
+
+    i {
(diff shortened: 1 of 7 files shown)


abchatra:
Port to stable

teachertool/src/components/styling/MakecodeFramePlaceholder.module.scss
```diff
@@ -0,0 +1,22 @@
...
+
+    align-items: center;
+    justify-content: center;
+    display: flex;
+    flex-direction: column;
+    gap: 1rem;
+
+    color: var(--pxt-page-foreground);
+    opacity: 0.65;
+    font-size: 2rem;
```
srietkerk, line 15: The text size is larger than the "Welcome to MakeCode Code Evaluation" text. I'm fine with leaving this as is if we want this text to be the first thing users look at, but if not, then I think we should make the text a bit smaller.

thsparks: Good point. Reduced to 1.5rem:

image

srietkerk (approved):

PR #10305, opened 2024-12-05
add docs for help pages in github extensions

riknoll:
adding detailed docs on how to add help pages for blocks in github hosted makecode extensions.

i wrote this mostly from memory so @tballmsft i would appreciate it if you gave it a once over since you recently went through this process.

PR #10307, opened 2024-12-05
Add the ability to localize help pages for docs

riknoll:
i'll add docs to https://github.com/microsoft/pxt/pull/10305 after this is merged.

adds the ability to localize help pages for extensions. simply put a `_locales` directory next to the markdown and put the translations inside just like how our other localization stuff works. see https://github.com/riknoll/arcade-character-animations/tree/main/docs/_locales/es for an example


eanders-ms (approved):
👍

PR #10309, opened 2024-12-06
perf milestone and measurement updates for minecraft

eanders-ms:
Adds three new extension points for sharing telemetry with the target:
* onPerfMilestone
* onPerfMeasurement
* onPostHostMessage (for sharing ticks and tutorial events)

In addition:
* Updated milestones and measurements to accept parameters.
* Factored out milestone and measurement string literals.
* Added a few more measurements, most notably we now measure pxt.U.requestAsync.


riknoll (approved):

webapp/src/app.tsx
```diff
@@ -4979,11 +4980,11 @@ export class ProjectView
...
             pxt.tickEvent("tutorial.editorLoaded")
             this.postTutorialLoaded();
+        }
 
-            pxt.perf.recordMilestone("editorContentLoaded");
-            if (!this.autoRunOnStart()) {
-                pxt.analytics.trackPerformanceReport();
-            }
+        pxt.perf.recordMilestone(Milestones.EditorContentLoaded);
+        if (!this.autoRunOnStart()) {
```
srietkerk, line 4986: Why is this no longer nested?

docfiles/pxtweb/cookieCompliance.ts
```diff
@@ -108,34 +141,44 @@ namespace pxt {
...
+                            for (let k of Object.keys(params || {})) {
+                                report += `\n\t\t\t${k}: ${params[k]}`
+                            }
                         }
                         report += `\n`
                     }
                     durations[msg] = duration;
                 }
                 console.log(report)
+                enabled = false; // stop collecting milestones and measurements after report
```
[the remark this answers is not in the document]
eanders-ms: I made this change because I believe it is undesirable to continue collecting perf measurements after the perf report is generated. These measurements accumulate in memory over time and never flush, which is bad if left unbounded (which it was). These metrics are most useful for gauging startup performance, so I think we don't want to continue collecting them once the editor is fully loaded. Let me know if I'm mistaken!

webapp/src/app.tsx
eanders-ms: I wanted to send this milestone to minecraft regardless of whether the project being loaded is a tutorial.

docfiles/pxtweb/cookieCompliance.ts
```diff
@@ -38,20 +38,50 @@ namespace pxt {
...
     // performance measuring, added here because this is amongst the first (typescript) code ever executed
     export namespace perf {
         let enabled: boolean;
 
+        export const onMilestone = createEventSource<{ milestone: string, time: number, params?: Map<string> }>();
+        export const onMeasurement = createEventSource<{ name: string, start: number, duration: number, params?: Map<string> }>();
+
         export let startTimeMs: number;
         export let stats: {
             // name, start, duration
```
thsparks, line 76:
```suggestion
            // name, start, duration, params
```

docfiles/pxtweb/cookieCompliance.ts
```diff
@@ -38,20 +38,50 @@ namespace pxt {
...
-            durations: [string, number, number][],
+            durations: [string, number, number, Map<string>?][],
             // name, event
```
thsparks, line 78:
```suggestion
            // name, event, params
```

docfiles/pxtweb/cookieCompliance.ts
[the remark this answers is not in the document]
thsparks: If I understand correctly, these are pretty generic methods. I worry someone might want to use them down the line (to measure issues with perf when hitting the play button, for example) and get confused as to why they're not working.

Maybe we could have a flag to control this behavior?

thsparks (approved):
A few comments, nothing major. Looks good on the whole!

docfiles/pxtweb/cookieCompliance.ts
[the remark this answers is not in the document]
eanders-ms: @thsparks I agree it could be useful beyond startup. Supporting this would require coming up with a flushing strategy beyond what exists today. That's out of scope of this change.

PR #10310, opened 2024-12-09
Fix localisation of help docs for langs like 'fr'

microbit-matt-hillsdon:
For French the code is 'fr' which just returns a single part. We can just use whichever parts are returned for the paths.

@riknoll spotted this testing out #10307, we were lucky/unlucky enough to pick French to try it out.


riknoll (approved):
ah, i messed up when copying that check from https://github.com/microsoft/pxt/blob/master/pxtlib/tutorial.ts#L516

PR #10313, opened 2024-12-10
Pxt Cli Command for Validating Block Translations

thsparks:
# Overview
This creates a pxt cli command (`validatetranslatedblocks`) which takes a baseline file and a translated file and checks the syntax of translated block strings to ensure they won't break the block.

I wanted to re-use [`pxtc.parseBlockDefinition`](https://github.com/microsoft/pxt/blob/master/pxtlib/service.ts#L1053) which is why this is in the pxt cli and not mkc. If we want to expose it in mkc, we could implement it as a service operation in [pxtcompiler/emitter/service.ts](https://github.com/microsoft/pxt/blob/master/pxtcompiler/emitter/service.ts), which mkc can call into, but it felt out of place in that file, and the user would have needed to run the command from inside a makecode project to work, which felt weird.

My initial approach was to actually run a build, apply the translated string to the block api info, and then check for differences from the original, but doing so added a lot of complexity with little benefit. As such, I've gone with this approach instead, which parses the strings and just manually checks parameters against the original.

# Testing
We don't have a unit test project for the pxt cli (I may try to add one in the future...) but for now, my test cases are attached as files:

[Baseline.json](https://github.com/user-attachments/files/18070238/Baseline.json)
[Output.json](https://github.com/user-attachments/files/18070239/Output.json)
[Translate.json](https://github.com/user-attachments/files/18070240/Translate.json)

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -5178,10 +5178,6 @@ export function validateTranslatedBlocks(parsed?: commandParser.ParsedCommand): 
     const translationMap = JSON.parse(fs.readFileSync(translatedFilePath, 'utf8'));
     const translationKeys = Object.keys(translationMap);
 
-    if (originalKeys.length !== translationKeys.length) {
-        U.userError(`Original and translation files have different number of keys. Original: ${originalKeys.length} Translation: ${translationKeys.length}`);
-    }
-
     const results: { [translationKey: string]: BlockStringValidationResult} = {};
     for (const translationKey of translationKeys) {
         if (!(translationKey in originalMap)) {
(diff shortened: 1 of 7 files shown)


pelikhan:
- to validate these commands, should you be running them through the existing translations?

pelikhan (approved):
Looks like a good start!

thsparks:
> * to validate these commands, should you be running them through the existing translations?

I did some of that, and I made some test-case files (just added them to the PR description). But yes, I'll run it through a few more real ones tomorrow just to make sure I'm not missing anything. Would be nice to callout any live broken ones, too :)

riknoll (approved):

PR #10315, opened 2024-12-10
clean up shadow blocks in insertion markers for expandable blocks

riknoll:
not an ideal solution, but this fixes the insertion marker bug i mentioned in standup. here's what the bug looks like:

insertion-marker-expanded

### DIFF
--- a/pxtblocks/composableMutations.ts
+++ b/pxtblocks/composableMutations.ts
@@ -350,6 +350,17 @@ export function initExpandableBlock(info: pxtc.BlocksInfo, b: Blockly.Block, def
                 // use domToBlockInternal so that we don't trigger a render while
                 // the block is still being initialized
                 newBlock = Blockly.Xml.domToBlockInternal(shadow, b.workspace);
+
+                // we don't know at this time whether the parent block is an insertion marker
+                // or not. doing this check lets us clean up the block in the case that it is,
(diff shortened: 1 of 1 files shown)


pxtblocks/composableMutations.ts
```diff
@@ -350,6 +350,17 @@ export function initExpandableBlock(info: pxtc.BlocksInfo, b: Blockly.Block, def
                 // use domToBlockInternal so that we don't trigger a render while
                 // the block is still being initialized
                 newBlock = Blockly.Xml.domToBlockInternal(shadow, b.workspace);
+
+                // we don't know at this time whether the parent block is an insertion marker
+                // or not. doing this check lets us clean up the block in the case that it is,
+                // though we get an annoying flicker
```
srietkerk, line 356: I don't think it looks too bad, but I was wondering if there is any sort of css that we could also use/override to make the shadow blocks not render in the insertion marker?

srietkerk: Oh, wait, is the GIF in the comment the way the insertion marker looks with the bug or how it looks with the fix for it?

riknoll: @srietkerk that's the bug! the fix renders it correctly

thsparks (approved):

PR #10316, opened 2024-12-12
TeacherTool path rename

ganicke:
Rename doc paths of `./teachertool` to `./code-eval-tool`. This is to support the changes for https://github.com/microsoft/pxt-microbit/issues/6036.


abchatra (approved):

thsparks (changes requested):
I would not recommend changing the paths on these files. At least not yet. We have it hard-coded in https://github.com/microsoft/pxt/blob/master/teachertool/src/transforms/loadCatalogAsync.ts#L7-L10

We probably shouldn't hard code them, but until we fix that, changing the location of these files will break our ability to load the catalogs.

That said, these URLs shouldn't be user-facing anyway. Not sure if there's much reason to change them. Is it necessary for updating the https://makecode.microbit.org/teachertool URL?

abchatra:
@ganicke may be sufficient to rename this md? https://github.com/Microsoft/pxt-microbit/blob/master/docs/teachertool.md

thsparks:
If we do move forward with this, we'll need to change those hard-coded paths and hotfix

PR #10317, opened 2024-12-13
Update tilemap/image/animation editor to use react common components

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/5924

this replaces all of the custom buttons and dropdowns that we have in the tilemap/image/animation editors with the buttons from react-common, which are better for accessibility.

also re-implements the tilemap palette page selector so that it now has a maximum of 5 pages

also adds a "FocusTrapRegion" subcomponent that lets you turn tab navigation on/off for regions within a focus trap. you can also use it to control the order of tab navigation when it differs from what's actually in the dom


webapp/src/components/ImageFieldEditor.tsx
```diff
@@ -621,23 +624,44 @@ export class ImageFieldEditor<U extends pxt.Asset> extends React.Component<Image
...
-        </div>
+        let { items, hidden, onEscape } = this.props;
+
+        return (
+            <FocusTrapRegion
+                className={classList("image-editor-gallery", items && !hidden && "visible")}
+                enabled={!hidden}
+                onEscape={onEscape}
+            >
+                {!hidden && items && items.map((item, index) =>
```
jwunderl, line 660:
```suggestion
                {!hidden && items?.map((item, index) =>
```

jwunderl (approved):
LGTM, nice clean up

eanders-ms (approved):
looks great

Nolan0027:
5 page limit? Is that normal

riknoll:
@Nolan0027 it's not an actual limit, it's just the max number of dots that will be displayed. you can still use the arrows to move further forward/back in the list

Nolan0027:
@riknoll ok

PR #10346, opened 2025-01-21
[Stable11.1] Cherry pick various fixes to micro:bit stable

riknoll:
cherry picking some recent fixes of mine over to micro:bit's stable branch. will open a pr doing the same thing for minecraft shortly.


abchatra:
Do these need bake time in arcade before we release?

PR #10348, opened 2025-01-23
Add new animation editor keyboard shortcuts

riknoll:
this tweaks a lot of existing shortcuts and adds a few more that are specific to the animation editor. all of the shortcuts that are affected by the marquee tool can now be combined with the `shift` key to affect all animation frames at the same time. for example:

new-anim-shortcuts

look at the timeline in that GIF to see all of the frames editing at once.

this also means that all of the shortcuts that were previously tied to uppercase letters have now been moved to the lowercase versions of those letters. the shifted versions will still work, but will affect all frames in animations. they are unchanged in the image/tilemap editor. also the pan tool shortcut was moved from `h` to `q` so as not to conflict with the horizontal flip shortcut. this breaks from the standard set by most image editors for pan, but i'm guessing most people use the spacebar anyhow.

finally, two new shortcuts were added for advancing the animation frame forwards and backwards

### DIFF
--- a/webapp/src/components/ImageEditor/keyboardShortcuts.ts
+++ b/webapp/src/components/ImageEditor/keyboardShortcuts.ts
@@ -1,7 +1,7 @@
 
 import { Store } from 'redux';
 import { ImageEditorTool, ImageEditorStore, TilemapState, AnimationState, CursorSize } from './store/imageReducer';
-import { dispatchChangeZoom, dispatchUndoImageEdit, dispatchRedoImageEdit, dispatchChangeImageTool, dispatchSwapBackgroundForeground, dispatchChangeSelectedColor, dispatchImageEdit, dispatchChangeCursorSize} from './actions/dispatch';
+import { dispatchChangeZoom, dispatchUndoImageEdit, dispatchRedoImageEdit, dispatchChangeImageTool, dispatchSwapBackgroundForeground, dispatchChangeSelectedColor, dispatchImageEdit, dispatchChangeCursorSize, dispatchChangeCurrentFrame, dispatchSetFrames} from './actions/dispatch';
 import { mainStore } from './store/imageStore';
 import { EditState, flipEdit, getEditState, outlineEdit, replaceColorEdit, rotateEdit } from './toolDefinitions';
 let store = mainStore;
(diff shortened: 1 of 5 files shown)


riknoll:
todo: make the page up and page down keys move the current animation frame up and down in the timeline

webapp/src/components/ImageEditor/keyboardShortcuts.ts
```diff
@@ -235,4 +278,142 @@ export function replaceColor(fromColor: number, toColor: number) {
     const [ editState, type ] = currentEditState();
     const replaced = replaceColorEdit(editState, fromColor, toColor);
     dispatchAction(dispatchImageEdit(replaced.toImageState()));
+}
+
+function doReplace() {
```
srietkerk, line 283:
```suggestion
function doColorReplace() {
```

webapp/src/components/ImageEditor/keyboardShortcuts.ts
```diff
@@ -235,4 +278,142 @@ export function replaceColor(fromColor: number, toColor: number) {
...
+
+    const fromColor = state.editor.backgroundColor;
+    const toColor = state.editor.selectedColor;
+
+    const [ editState ] = currentEditState();
+    const replaced = replaceColorEdit(editState, fromColor, toColor);
+    dispatchAction(dispatchImageEdit(replaced.toImageState()));
+}
+
+function doReplaceAllFrames() {
```
srietkerk, line 294:
```suggestion
function doColorReplaceAllFrames() {
```

webapp/src/components/ImageEditor/keyboardShortcuts.ts
```diff
@@ -235,4 +278,142 @@ export function replaceColor(fromColor: number, toColor: number) {
...
+    const fromColor = state.editor.backgroundColor;
+    const toColor = state.editor.selectedColor;
+
+    editAllFrames(doReplace, editState => replaceColorEdit(editState, fromColor, toColor))
+}
+
+export function advanceFrame(forwards: boolean) {
+    const state = store.getState();
+
+    if (state.editor.isTilemap) return;
```
srietkerk, line 306: Do we need to check if it's the image editor as well?

srietkerk (approved):
Couple small suggestions, LGTM!

webapp/src/components/ImageEditor/keyboardShortcuts.ts
riknoll: that's what the frames length check does! the image editor uses the same state as the animation editor, it just has only one frame

PR #10349, opened 2025-01-23
need to start the simulator after create to avoid race

tballmsft:
A very infrequent race condition occurs where the extension's simulator appears and the micro:bit simulator never appears. This is very curious, as the extension simulator can only be invoked by static typescript (user-level code), which must be executing, which should bring up the micro:bit simulator.  The problem is that the start function of simulatorDriver does not start the simulator in the case when it is being created the first time.  Most often, some other path then creates the simulator, and the start function is called, which does start the simulator. But this change ensures the the micro:bit simulator will always appear first.

Image

### DIFF
--- a/pxtsim/simdriver.ts
+++ b/pxtsim/simdriver.ts
@@ -810,8 +810,8 @@ namespace pxsim {
                 let wrapper = this.createFrame();
                 this.container.appendChild(wrapper);
                 frame = wrapper.firstElementChild as HTMLIFrameElement;
-            } else // reuse simulator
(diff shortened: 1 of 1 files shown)


riknoll (changes requested):
this is adding pxt-microbit as a submodule. i'm guessing that's accidental, but definitely needs to be fixed

tballmsft:
thanks for that catch, richard... i had mistakenly done a git clone of pxt-microbit inside pxt!

riknoll (approved):

PR #10350, opened 2025-01-23
Add animation editor shortcuts for moving frames in the timeline

riknoll:
adds four new shortcuts:

* **PageUp** - Swap the current animation frame with the previous frame in the timeline
* **PageDown** - Swap the current animation frame with the next frame in the timeline
* **shift + PageUp** - Rotate all frames in the animation backwards one frame
* **shift + PageDown** - Rotate all frames in the animation forwards one frame

also fixes the right click handler on react-common buttons which i guess i never tested


srietkerk (approved):
Nice!

PR #10355, opened 2025-01-28
OCV first pass

srietkerk:
This is the work to plug in the One Customer Voice iframe into our products. Right now, it's only getting triggered from our settings menus, but we can pop this modal up in any context.

**Things of note in this PR**
- I likely should have split this into a separate PR, but I added a flag to the CLI to use https rather than always using http. This was needed because the OCV iframe only shows up for domains in an allow list, and the local development options they have for that allow list use https.
- I added the `giveFeedback` option to the appTheme object as a starting point for only showing the feedback menu item for our targets.
- `FeedbackListener.ts` is where the logic to get OCV to work lives. There is a lot of generic code in there and console logs that I figure we'll want to update to something more clean/helpful, but I kept it in for now for simplicity. A note about the event listener: the ones that I kept in are the ones I figured would be useful for us. There were some others, but I removed them because for our use case, they wouldn't really apply. If you would like to see the other events, I would be happy to showcase them. 

**To Dos**
- Programmatically changing the theme of the feedback for high contrast and so we can have different colors based on the target we're serving from
- Use the "kind" flag that I've added to the feedback object to change the feedback config. I'm planning on following a similar pattern that the share modal uses with ShareInfo.
- Get rating feedback to show upon finishing tutorials

**Testing**
Unfortunately, I cannot make an upload target that will be useful for these changes because of the allow list that I talked about before. I've pinged the OCV teams a couple of times to get the makecode domain in their allow list but I'm not sure of the progress. If I made an upload target, you would be able to see the "Give Feedback" option in the setting menu, but clicking on it will give you a modal with a broken iframe. If you want to actually try out the forms, you can test locally with https. I'm going to write up a doc on how to test OCV locally.

In the meantime, here's how it looks:
image

image

image


webapp/src/container.tsx
```diff
@@ -349,7 +355,7 @@ export class SettingsMenu extends data.Component<SettingsMenuProps, SettingsMenu
             {
                 // we always need a way to clear local storage, regardless if signed in or not
             }
-            {targetTheme.feedbackUrl ? <a className="ui item" href={targetTheme.feedbackUrl} role="menuitem" title={lf("Give Feedback")} target="_blank" rel="noopener noreferrer" >{lf("Give Feedback")}</a> : undefined}
```
riknoll, line 352: do we want to get rid of this? is this the URL we use for partner feedback portals (like in minecraft/micro:bit)?

react-common/components/controls/Feedback/FeedbackEventListener.ts
```diff
@@ -0,0 +1,125 @@
+import { appId, feedbackFrameUrl } from './configs';
+interface FeedbackRequestEventPayload<T> {
+  Event: string
+  EventArgs: string
+}
+type FeedbackRequestPayloadType = FeedbackRequestEventPayload<any>
```
riknoll, line 6: instead of using `any` here, would be nice to instead have a union of types that contain all of the response objects. then the compiler can figure out the typings based on the `event` field in the switch statement you have in `feedbackCallbackEventListener `

react-common/components/controls/Feedback/FeedbackEventListener.ts
```diff
@@ -0,0 +1,125 @@
...
+export const removeFeedbackEventListener = () => {
+    window.removeEventListener('message', feedbackCallbackEventListener);
+}
+
+/**
+ * The function that listens for the feedback events.
+ * The events here are the ones that seemed most useful to log or respond to
+ * @param {MessageEvent<FeedbackRequestPayloadType>} event: the event received from OCV
+ */
+const feedbackCallbackEventListener = (event: MessageEvent<FeedbackRequestPayloadType>) => {
```
riknoll, line 63: whitespace in this function seems messed up. the body isn't indented and neither is the switch statement

react-common/components/controls/Feedback/FeedbackEventListener.ts
```diff
@@ -0,0 +1,125 @@
...
+        currentTheme = 'WindowsDark'
+    }
+    const response: FeedbackResponsePayloadType = {
+        event: 'OnFeedbackHostAppThemeChanged',
+        data: {
+            baseTheme: currentTheme,
+        },
+    }
+    const iFrameEle = document.getElementById(FEEDBACK_FRAME_ID) as HTMLIFrameElement
+    iFrameEle!.contentWindow!.postMessage(response, feedbackFrameUrl)
```
riknoll, line 108:
```suggestion
    const iFrameElement = document.getElementById(FEEDBACK_FRAME_ID) as HTMLIFrameElement
    iFrameElement!.contentWindow!.postMessage(response, feedbackFrameUrl)
```

react-common/components/controls/Feedback/FeedbackEventListener.ts
```diff
@@ -0,0 +1,125 @@
...
+const sendFeedbackInitOptions = () => {
+    type FeedbackResponsePayloadType = FeedbackResponseEventPayload<any>
+    feedbackData.callbackFunctions = undefined
+    let response: FeedbackResponsePayloadType = {
+        event: 'InAppFeedbackInitOptions',
+        data: feedbackData,
+    }
+    response = JSON.parse(JSON.stringify(response))
+    const iFrameEle = document.getElementById(FEEDBACK_FRAME_ID) as HTMLIFrameElement
+    iFrameEle!.contentWindow!.postMessage(response, feedbackFrameUrl)
```
riknoll, line 124:
```suggestion
    const iFrameElement = document.getElementById(FEEDBACK_FRAME_ID) as HTMLIFrameElement
    iFrameElement!.contentWindow!.postMessage(response, feedbackFrameUrl)
```

react-common/components/controls/Feedback/configs.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export const ratingFeedbackConfig: any = {
+    ...baseConfig,
+    initialFeedbackType: "Unclassified",
+    scenarioConfig: {
+      isScenarioEnabled: true,
+      scenarioType: "Custom",
+      questionDetails: {
+        questionUiType: "Rating",
+        questionInstruction: {
```
riknoll, line 26: all of these question strings should be wrapped in `lf`. also, instead of declaring this as a variable you should create a function that returns it. otherwise, the language might not have been set before this file is imported

react-common/styles/controls/Feedback.less
```diff
@@ -0,0 +1,3 @@
+.feedback-modal iframe{
```
riknoll, line 1:
```suggestion
.feedback-modal iframe {
```

webapp/src/container.tsx
```diff
@@ -349,7 +355,7 @@ export class SettingsMenu extends data.Component<SettingsMenuProps, SettingsMenu
...
+            {targetTheme.giveFeedback ? <sui.Item role="menuitem" text={lf("🙂 Give Feedback")} onClick={this.showFeedbackDialog} /> : undefined}
```
riknoll, line 358: this icon shouldn't be included in the `lf` string, otherwise a translator might move it so it's not left aligned like the rest of the menu items. same for the home menu item below

riknoll (changes requested):
would be nice to have typings for these configs! they don't have to be exhaustive, but should include the fields you're referencing. we can expand them later as we add more code

abchatra:
Nice work

webapp/src/container.tsx
srietkerk: As far as I could tell, no. The only places where `feedbackUrl` is used is not from the appTheme. Across all the workspaces, there are instances of `feedbackUrl` for experiments and code cards but that's separate. The feedback url from the appTheme should only be used for cases in the settings menu which should all get replaced with OCV. This feedback option is also not in the settings menu on live, so I think it's safe to get rid of. If there are still concerns about it, I can keep it in, but it should be safe.

One thing I forgot to do was remove `feedbackUrl` from the appTheme interface in `pxtarget.d.ts`, I was planning on doing removing it, but I can also just leave it in for posterity.

react-common/components/controls/Feedback/configs.ts
riknoll: any string that is shown to the user should be localized. strings that are just settings, like the "Rating" above shouldn't be.

react-common/components/controls/Feedback/FeedbackEventListener.ts
```diff
@@ -0,0 +1,125 @@
...
+    const payload: FeedbackRequestPayloadType = event.data
+    switch (payload.Event) {
+    case 'InAppFeedbackInitOptions': //This is required to initialise feedback
+        sendFeedbackInitOptions()
+        break
+    case 'InAppFeedbackOnError': //Invoked when an error occurrs on feedback submission - would be nice to log something to the user
+        console.log('Error Message: ', payload.EventArgs)
+        break
+    case 'InAppFeedbackInitializationComplete': //Invoked when feedback form is fully initialised and displays error/warning if any - nice to have a log for this
+        console.log('InAppFeedbackInitializationComplete: ', payload.EventArgs)
```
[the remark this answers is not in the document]
srietkerk: So I changed the console logs to pxt.debug, but I no longer see the logs and my console is configured to output all log levels. Is this expected?

react-common/components/controls/Feedback/configs.ts
```diff
@@ -0,0 +1,57 @@
+export const appId = 50315;
+export const feedbackFrameUrl = 'https://admin-ignite.microsoft.com';
```
thsparks, line 2: Are feedbackFramUrl and appId things we should store in targetconfig?

react-common/components/controls/Feedback/FeedbackEventListener.ts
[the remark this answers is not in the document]
eanders-ms: add `dbg=1` to the url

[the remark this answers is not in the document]
srietkerk: Ah, thank you!

webapp/src/container.tsx
thsparks, line 358: IIRC, you mentioned in stand-up that the smiley face icon is jarring. I'm guessing that's why it's in the string here instead of the icon property on the Item. Per Richard's comment, we probably want to use the icon property. Did you consider using a speech bubble icon or bullhorn icon instead?

react-common/components/controls/Feedback/configs.ts
srietkerk: I actually think this is something OCV will do on their end.. I'd have to ask how they do it, but I might want to leave as is for now. 
image

react-common/components/controls/Feedback/configs.ts
```diff
@@ -0,0 +1,57 @@
...
+          {
+            displayedStringInEnglish: "Boring",
+            displayedString: "Boring"
+          },
+          {
+            displayedStringInEnglish: "Not fun",
+            displayedString: "Not fun"
+          },
+          {
+            displayedStringInEnglish: "Kinda fun",
```
thsparks, line 40: nit: I'd recommend "kind of fun" instead of "kinda fun"

thsparks: Mostly because I wonder if "kinda" will bother picky teachers

srietkerk: Yeah, definitely. I'm not expecting we keep these. These are basically placeholders as I haven't gotten PM feedback for this yet. The rating one really isn't being used yet, either, so I would like to save this change when I get the rating feedback working if that's alright.

react-common/components/controls/Feedback/configs.ts
srietkerk: This was definitely something I was considering, but these are going to be the same for each of our targets, so it feels redundant to put these in the target config of each target.

webapp/src/container.tsx
srietkerk: Ooh, I hadn't considered that! I was trying to align as close as possible to Microsoft's icon patterns, but I like the idea of the comment bubble. I'll see how that looks!

react-common/components/controls/Feedback/configs.ts
riknoll: i don't think they could be translating it on the fly... looks like you need to pass the parameter both as a localized and non localized version. i bet that the english param is just what shows up in whatever UI they have for looking at the feedback folks send in.

srietkerk: Ah, I see. I will update it, then, thanks!

webapp/src/container.tsx
srietkerk: Alright, here's the different options, let me know what you think.

With emoji
image

With comment bubble
image

thsparks: Hmm, personally I would vote for the comment icon (bonus points if there's a filled-in version to match the other icons). I feel like the color in the emoji contrasts strangely with what's already there.

react-common/components/controls/Feedback/configs.ts
thsparks: I suppose that's true. Ideally, I think we'd want to put this somewhere that wouldn't require a release to change (i.e. what happened with our CDN) but I'm not sure if we have a good option for that without going through our backend.

eanders-ms: I agree this url should go in pxtarget.json, even if it is the same for each target. This is what we do for `privacyUrl`, `termsOfUseUrl`, etc.

eanders-ms: Somewhat related question: is the OCV/feedback feature only enabled for 1st party targets?

srietkerk: Yes, we should only be showing feedback in our self-hosted targets. There is more work to really verify this, which will be my next priority, but for now, it's just something that gets set in the apptheme as something to include in the target.

eanders-ms: Were you able to move these hard-coded config values to pxtarget.json?

react-common/components/controls/Feedback/FeedbackEventListener.ts
```diff
@@ -0,0 +1,158 @@
...
+    } else {
+        currentTheme = 'WindowsDark'
+    }
+    const response: FeedbackResponsePayloadType = {
+        event: 'OnFeedbackHostAppThemeChanged',
+        data: {
+            baseTheme: currentTheme,
+        },
+    }
+    const iFrameElement = document.getElementById(FEEDBACK_FRAME_ID) as HTMLIFrameElement
```
eanders-ms, line 142: Code duplication here. Move these two lines to a helper function. In that function, check for missing iFrameElement, as document.getElementById can return undefined.

react-common/components/controls/Feedback/FeedbackEventListener.ts
```diff
@@ -0,0 +1,158 @@
...
+        }
+    }
+}
+
+// ***************** Helper Functions *****************
+
+// TODO
+// haven't implemented yet with events, but this will be needed in order to update to high contrast
+// general changes need to be made as well use the correct theme. the windows ones were just the defaults.
+const sendUpdateTheme = () => {
```
eanders-ms, line 130: Do you need to update `themeOptions` somewhere in this method, to keep it in sync?

react-common/components/controls/Feedback/FeedbackEventListener.ts
```diff
@@ -0,0 +1,158 @@
...
+                break
+            case 'InAppFeedbackInitializationComplete': //Invoked when feedback form is fully initialised and displays error/warning if any - nice to have a log for this
+                pxt.debug('InAppFeedbackInitializationComplete: ', payload.EventArgs)
+                break
+            case 'InAppFeedbackOnSuccess': //Invoked when feedback submission is successful - would be useful to have telemetry/something else on this event
+                pxt.debug('InAppFeedbackOnSuccess: ', payload.EventArgs)
+                break
+            case 'InAppFeedbackDismissWithResult': //Invoked when feedback is dismissed - the big important one for us to be able to close the feedback modal
+                pxt.debug('InAppFeedbackDismissWithResult: ', payload.EventArgs);
+                if (feedbackCallbacks.onDismiss) {
```
eanders-ms, line 117:
```suggestion
                if (feedbackCallbacks?.onDismiss) {
```

eanders-ms: Or even better: `feedbackCallbacks?.onDismiss?.()`

react-common/components/controls/Feedback/FeedbackEventListener.ts
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * Actually initializes the feedback session. This is called when the feedback modal opens.
+ */
+const sendFeedbackInitOptions = () => {
+    let response: FeedbackResponsePayloadType = {
+        event: 'InAppFeedbackInitOptions',
+        data: initfeedbackOptions,
+    }
+    response = JSON.parse(JSON.stringify(response))
```
eanders-ms, line 155: Is there a strong reason for keep copy here?

localtypings/pxtarget.d.ts
```diff
@@ -524,6 +524,7 @@ declare namespace pxt {
         timeMachineDiffInterval?: number; // An interval in milliseconds at which to take diffs to store in project history. Defaults to 5 minutes
         timeMachineSnapshotInterval?: number; // An interval in milliseconds at which to take full project snapshots in project history. Defaults to 15 minutes
         adjustBlockContrast?: boolean; // If set to true, all block colors will automatically be adjusted to have a contrast ratio of 4.5 with text
+        giveFeedback?: boolean; // Show the give feedback button in the settings menu
```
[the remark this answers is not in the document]
srietkerk: I think the comment that I gave is a bit misleading. This is used to indicate whether or not we can show feedback in the given target, so it's not just limited to a button or option in a menu. Per your second recommendation, I'm thinking of changing this to `feedbackEnabled` instead.

react-common/components/controls/Feedback/FeedbackEventListener.ts
srietkerk: Nope!

eanders-ms (approved):
Looks good!

riknoll (approved):

thsparks (approved):

PR #10356, opened 2025-01-28
filter serialized tilemaps to only include used ones

riknoll:
my riknoll/arcade-overworld extension has a block with a ton of tilemaps in it that you can expand using +/- mutator buttons. because of how our tilemap field editor works, an asset gets created for every one of these fields regardless of whether it's actually expanded in the project or not. as a result, your project gets flooded with hundred of tilemaps that you aren't even using.

this change isn't a proper fix for that issue, but it does at least prevent those tilemaps from being included in the compiled binary. the proper fix is to change how fields work so that those maps are lazily allocated. in any case, this is a good change even if we do eventually fix that root cause.

### DIFF
--- a/pxtlib/spriteutils.ts
+++ b/pxtlib/spriteutils.ts
@@ -736,6 +736,19 @@ namespace pxt.sprite {
         pxt.sprite.trimTilemapTileset(result);
     }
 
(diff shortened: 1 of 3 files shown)


srietkerk (approved):
Nice!

PR #10361, opened 2025-01-29
fix measures field in playback controls

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/6637

also fixes the react warning about uncontrolled components that appears in the console

### DIFF
--- a/react-common/components/controls/Input.tsx
+++ b/react-common/components/controls/Input.tsx
@@ -69,6 +69,10 @@ export const Input = (props: InputProps) => {
 
     let container: HTMLDivElement;
 
+    React.useEffect(() => {
+        setValue(initialValue || "");
+    }, [initialValue]);
+
     const handleContainerRef = (ref: HTMLDivElement) => {
         if (!ref) return;
         container = ref;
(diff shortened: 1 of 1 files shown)


react-common/components/controls/Input.tsx
```diff
@@ -137,7 +141,7 @@ export const Input = (props: InputProps) => {
             onBlur(value);
         }
         if (!preserveValueOnBlur) {
-            setValue(undefined);
+            setValue("");
```
srietkerk, line 144: Just for a sanity check, is this change just to stay consistent with the string type? Like changing this to the empty string should still be falsey and not cause problems, right?

react-common/components/controls/Input.tsx
```diff
@@ -69,6 +69,10 @@ export const Input = (props: InputProps) => {
 
     let container: HTMLDivElement;
 
+    React.useEffect(() => {
+        setValue(initialValue || "");
+    }, [initialValue]);
```
srietkerk, line 74: Is this behavior something we should support? I feel like `initialValue` should be something that should not change. Can the fix here be that the playback controls use onChange?

riknoll: `initialValue` probably isn't the best name for this prop, it should probably just be called `value`. i don't think there's any reason to prevent it from changing though. the child component should respect the props passed in from the parent.

PR #10363, opened 2025-01-30
Tweaks to performance metrics handling

eanders-ms:
Changes in this PR:
* **Caching change**: Allow previously logged performance events to be sent to listeners that subscribe after the fact.
* **Configurable duration threshold**: Made the threshold for measurement reporting dynamic and configurable by the target extension. This preserves the default behavior while allowing targets to change it once they're initialized.
* **Include project id**: When available, include project header id in perf events.


riknoll (approved):
LGTM but do we ever actually use the filtering stuff?

eanders-ms:
> LGTM but do we ever actually use the filtering stuff?

I couldn't find any reference to it! And the exported signature for the `report` method didn't expose it.

PR #10365, opened 2025-01-30
Fix key handling after simulator is made full screen

microbit-robert:
After triggering the full screen simulator, the "escape" key does nothing. After minimising the simulator using the relevant button, the "escape" key now triggers the full screen simulator.

The logic in the code appears to be inverted. The call to focus the simulator also seems dubious as it feels like the best thing to do is leave focus on the fullscreen toggle button in both the expand and collapse case, however, I'd welcome feedback on this.

This PR fixes the inverted logic in the code ~~and removes the dubious focus call~~.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -3651,9 +3651,8 @@ export class ProjectView
         if (this.state.collapseEditorTools) {
             this.expandSimulator();
         }
-        if (!enabled) {
+        if (enabled) {
             document.addEventListener('keydown', this.closeOnEscape);
-            simulator.driver.focus();
         } else {
             document.removeEventListener('keydown', this.closeOnEscape);
         }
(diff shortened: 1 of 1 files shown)


riknoll:
hmm... not sure about removing the focus stealing here. seems to me that the fullscreen simulator should be treated like a modal, since that's essentially what it is, and with modals we usually move focus to the content (or at least the container of the content). i think we should match that experience as closely as possible.

rest of the pr LGTM!

microbit-robert:
> hmm... not sure about removing the focus stealing here. seems to me that the fullscreen simulator should be treated like a modal, since that's essentially what it is, and with modals we usually move focus to the content (or at least the container of the content). i think we should match that experience as closely as possible.
> 
> rest of the pr LGTM!

That makes sense, thanks. I've pushed a commit to restore the focus call.

riknoll (approved):

PR #10368, opened 2025-02-04
perform tilemap upgrade on xml instead of workspace

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/6602
fixes https://github.com/microsoft/pxt-arcade/issues/6635
fixes https://github.com/microsoft/pxt-arcade/issues/6601

our old method of upgrading tilemaps doesn't seem to be working with the new blockly; when blockly sets the values of the created fields the field validation code is causing their values to be overwritten before we can perform our upgrades.

to bypass that issue, i converted the upgrade code to be performed directly on the blocks xml instead of the workspace.

also, since people might not be familiar with this ancient code, this is the upgrade logic that i put into place ages ago when we started having a concept of "assets" in projects. previously we stored the tileset in the variables of the blockly workspace and that's what the tileset field drew its options from. tilemap blocks also just inlined all of the tilemap data instead of compiling down to an asset reference

### DIFF
--- a/pxtblocks/fields/field_utils.ts
+++ b/pxtblocks/fields/field_utils.ts
@@ -238,28 +238,46 @@ export function needsTilemapUpgrade(ws: Blockly.Workspace) {
     return !!allTiles.length;
 }
 
-export function upgradeTilemapsInWorkspace(ws: Blockly.Workspace, proj: pxt.TilemapProject) {
-    const allTiles = ws.getVariablesOfType(pxt.sprite.BLOCKLY_TILESET_TYPE).map(model => pxt.sprite.legacy.blocklyVariableToTile(model.name));
-    if (!allTiles.length) return;
+export function updateTilemapXml(dom: Element, proj: pxt.TilemapProject) {
+    let needsUpgrade = false;
 
-    try {
-        Blockly.Events.disable();
-        let customMapping: pxt.Tile[] = [];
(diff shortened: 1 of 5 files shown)


abchatra:
Nice

pxtblocks/fields/field_utils.ts
```diff
@@ -238,28 +238,46 @@ export function needsTilemapUpgrade(ws: Blockly.Workspace) {
...
-
-        for (const tile of allTiles) {
-            if (tile.qualifiedName) {
-                customMapping[tile.projectId] = proj.resolveTile(tile.qualifiedName);
+export function updateTilemapXml(dom: Element, proj: pxt.TilemapProject) {
+    let needsUpgrade = false;
+
+    const upgradedTileMapping: pxt.Map<pxt.Tile> = {};
+
+    for (const element of dom.children) {
```
srietkerk, line 246: Just curious if iterating over everything in the dom is the most efficient way to do this.. Would this or would filtering for elements with the variable tag and then iterating through those entries be faster?

riknoll: i don't think it would make a huge difference either way. this is only looping over the top-level elements in the xml file so it should only be the variables and any top-level events like on-start

srietkerk (approved):

PR #10369, opened 2025-02-05
Fix keyboard navigation for the dropdown menu component

microbit-robert:
This PR fixes the following bugs in keyboard handling for the dropdown menu component:

- There is currently a bug where key handlers are added to the menu items on `componentDidMount`. Any items that are added after `componentDidMount` (such as "Connect Device" in the settings menu) do not get key handling.
- On narrow screens, additional items are shown in the settings menu, however, none of these have key handling. This is because `getChildren` checks for mobile via `userAgent` and not whether the menu item is visible based on screen width.

All key handling has been moved up to the parent dropdown to address the first bug and the use of `offsetParent` replaces the call to `pxt.BrowserUtils.isMobile()` to address the second bug. This PR maintains the existing keyboard and mouse behaviour except for focussing the first menu item when the menu is opened via the keyboard using "Enter" or "Space" which is [considered best practice](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/).

This screen recording demonstrates both bugs:
https://github.com/user-attachments/assets/76c7d84c-5e48-42cd-a6be-43a9c67a00b9

@microbit-matt-hillsdon

### DIFF
--- a/webapp/src/sui.tsx
+++ b/webapp/src/sui.tsx
@@ -91,10 +91,13 @@ export class DropdownMenu extends UIElement<DropdownProps, DropdownState> {
         this.setState({ open: false });
     }
 
-    toggle() {
+    toggle(focusFirst?: boolean) {
         if (this.state.open) {
             this.hide();
         } else {
+            if (focusFirst) {
+                this.focusFirst = true;
+            }
             this.show();
         }
     }
@@ -104,7 +107,7 @@ export class DropdownMenu extends UIElement<DropdownProps, DropdownState> {
         el.focus();
     }
 
-    private blur(el: HTMLElement) {
+    private setInactive(el: HTMLElement) {
         if (this.isActive(el)) {
             pxt.BrowserUtils.removeClass(el, "active");
         }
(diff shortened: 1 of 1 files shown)


webapp/src/sui.tsx
```diff
@@ -127,8 +130,8 @@ export class DropdownMenu extends UIElement<DropdownProps, DropdownState> {
             const child = menu.childNodes[i] as HTMLElement;
             // Remove separators
             if (pxt.BrowserUtils.containsClass(child, "divider")) continue;
-            // Check if item is intended for mobile only views
-            if (pxt.BrowserUtils.containsClass(child, "mobile") && !pxt.BrowserUtils.isMobile()) continue;
+            // Check if item is visible. Some items are intended for mobile only views.
+            if (!child.offsetParent) continue;
```
microbit-robert, line 134: We considered using `checkVisibility()` here, but this appears to be too recent for the tsconfig settings.

microbit-robert:
@microsoft-github-policy-service agree company="Micro:bit Educational Foundation"

riknoll (approved):
eventually we should move these over to the react-common implementation of menus, which handles all of this correctly. still this change is good for the meantime!

PR #10371, opened 2025-02-05
don't decompile non-shadow blocks within shadows

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/6592

mainline blockly strictly enforces that shadow blocks cannot have non-shadow children, but there are still some cases where code like that is generated by the decompiler. it doesn't matter for project code since that code gets patched to fix this issue, but the block config decompiled code doesn't get patched.

rather than just apply the patches to the decompiled code here, i figured it was easy enough to just fix it from the source.


eanders-ms (approved):
👍

PR #10377, opened 2025-02-08
OCV: add support for high contrast toggling

srietkerk:
As the title says, this makes it so when we switch to high contrast, we use a high contrast/dark themed version of OCV.

This is what it looks like toggled on:
(Image updated)
image


react-common/components/controls/Feedback/FeedbackEventListener.ts
```diff
@@ -129,17 +130,17 @@ const getIFrameAndSend = (payload: FeedbackResponsePayloadType) => {
 // TODO
```
eanders-ms, line 130: Can this TODO comment be deleted now?

eanders-ms (approved):

thsparks (approved):

PR #10379, opened 2025-02-10
add copy/paste to context menu and restrict tutorial pasting

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/6657
fixes https://github.com/microsoft/pxt-arcade/issues/6654

adds copy/paste to the context menu and tweaks where you can paste stuff. tutorials are now restricted so that you can only paste blocks that are copied from the same project.

i did not add "cut" to the context menu, but can if people feel strongly about it


eanders-ms (approved):
👍

thsparks (approved):

PR #10382, opened 2025-02-12
Align new project skip link behaviour with new project button

microbit-robert:
The new project skip link on the home page launches a new project without showing the "Create a Project" dialog (if  available depending on the `appTheme` settings) where as the "New Project" button does. This PR aligns these behaviours. 

@microbit-matt-hillsdon

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -2914,13 +2914,11 @@ export class ProjectView
             });
     }
 
-    newProjectMaybeWithProjectCreationOptions(firstProject: boolean) {
+    async newUserCreatedProject(firstProject: boolean): Promise<void> {
         if (pxt.appTarget.appTheme.nameProjectFirst || pxt.appTarget.appTheme.chooseLanguageRestrictionOnNewProject) {
-            this.askForProjectCreationOptionsAsync()
-                .then(projectSettings => {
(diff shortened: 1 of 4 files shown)


webapp/src/app.tsx
```diff
@@ -2914,6 +2914,18 @@ export class ProjectView
             });
     }
 
+    newProjectMaybeWithProjectCreationOptions(firstProject: boolean) {
```
microbit-robert, line 2917: We'd absolutely welcome a better name for this method. An alternative could be just to duplicate this code in both `accessibility.tsx` and `projects.tsx` in their own `newProject` methods that then call `this.props.parent.newProject`. The latter approach does make it more likely that this bug will reoccur if a new entry point (button) to create a new project is introduced in the app.

riknoll: oh yeah, that one's a mouthful... perhaps `newUserCreatedProject`? would also be nice to convert this into async/await while you're here

microbit-robert: That sounds good. Both changes made here [aea20a3](https://github.com/microsoft/pxt/pull/10382/commits/aea20a3cefe29c01ac7bb7d82456cf07ad3fa184).

riknoll (approved):

PR #10384, opened 2025-02-12
A few blockly tutorial fixes

riknoll:
each commit is a different fix!

fixes https://github.com/microsoft/pxt-arcade/issues/6630
the issue here was that the min/max properties on the field were unset, which was causing it fail to parse values passed in as the default

fixes https://github.com/microsoft/pxt-arcade/issues/6643
this issue was that we weren't busting the flyout cache properly for tutorials where the contents of a flyout can change for each step

fixes https://github.com/microsoft/pxt-arcade/issues/6627
there are a ton of different ways to decompile to the tileset field and each one has slightly different syntax. adding a fallback case to the if statement that just tries to find the asset by name even if the value passed in isn't in the full asset reference format

### DIFF
--- a/pxtblocks/fields/field_numberdropdown.ts
+++ b/pxtblocks/fields/field_numberdropdown.ts
@@ -108,7 +108,12 @@ class BaseFieldNumberDropdown extends BaseFieldTextDropdown {
             return null;
         }
         // Get the value in range.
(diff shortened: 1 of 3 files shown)


thsparks (approved):
:shipit:

PR #10386, opened 2025-02-13
OCV: add web config flag client side

srietkerk:
This makes it so we can only get feedback for targets that we serve.

### DIFF
--- a/pxtlib/main.ts
+++ b/pxtlib/main.ts
@@ -395,7 +395,7 @@ namespace pxt {
         teachertoolUrl?: string; // "/beta---eval"
         isStatic?: boolean;
         verprefix?: string; // "v1"
-        ocvEnabled: boolean;
+        ocvEnabled?: boolean;
     }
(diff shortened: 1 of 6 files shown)


pxtlib/main.ts
```diff
@@ -395,6 +395,7 @@ namespace pxt {
         teachertoolUrl?: string; // "/beta---eval"
         isStatic?: boolean;
         verprefix?: string; // "v1"
+        ocvEnabled: boolean;
```
thsparks, line 398: I think if you define this as an optional field (ocvEnabled?) then you won't need to go in and add it to all those other files  (like test.spec.ts)

thsparks (approved):
For my own education, is the idea that this is now a field we control from the backend?

Change itself looks good!

pxtlib/main.ts
riknoll: +1

PR #10388, opened 2025-02-14
OCV refactor

srietkerk:
Added some device telemetry points to see in OCV.

Refactored the feedback it so it's possible to have the feedback without the modal.

Made it so feedback would only show up if we're getting the right data from webconfig.


react-common/components/controls/Feedback/Feedback.tsx
```diff
@@ -6,48 +6,55 @@ import { Modal } from "../Modal";
...
+interface IFeedbackModalProps {
   kind: ocv.FeedbackKind;
   onClose: () => void;
 }
 
 // Wrapper component of the feedback modal so kind can determine what feedback actually shows in the modal
-export const Feedback = (props: IFeedbackProps) => {
+export const FeedbackModal = (props: IFeedbackModalProps) => {
   const { kind, onClose } = props;
+  const title = kind === "generic" ? lf("Leave Feedback") : lf("Rate this activity");
```
thsparks, line 25: nit: I would swap this to check if kind === "rating" so the fallback is generic (i.e. "Leave Feedback") in the event of an unrecognized/new kind being introduced.

react-common/components/controls/Feedback/Feedback.tsx
```diff
@@ -59,12 +66,10 @@ export const FeedbackModal = (props: IFeedbackModalProps) => {
     }
   }, [])
   return (
-    <Modal className="feedback-modal" title={title} onClose={onClose}>
       <iframe
       title="feedback"
       id={frameId}
-      src={`${pxt.appTarget.appTheme.ocvFrameUrl}/centrohost?appname=ocvfeedback&feature=host-ocv-inapp-feedback&platform=web&appId=${pxt.appTarget.appTheme.ocvAppId}#/hostedpage`}
+      src={`${pxt.webConfig.ocv?.iframeEndpoint}/centrohost?appname=ocvfeedback&feature=host-ocv-inapp-feedback&platform=web&appId=${pxt.webConfig.ocv?.appId}#/hostedpage`}
```
thsparks, line 72: I'm wondering if, rather than `ocv?.`, we should we check `src={ocv ? "..." : undefined}`. What actually happens if ocv is null here and this becomes a malformed URL?

webapp/src/projects.tsx
```diff
@@ -307,7 +307,7 @@ export class ProjectSettingsMenu extends data.Component<ProjectSettingsMenuProps
         const githubUser = !hasIdentity && this.getData("github:user") as UserInfo;
         const reportAbuse = pxt.appTarget.cloud && pxt.appTarget.cloud.sharing && pxt.appTarget.cloud.importing;
         const showDivider = targetTheme.selectLanguage || targetTheme.highContrast || githubUser;
-        const showFeedbackOption = pxt.webConfig.ocvEnabled && targetTheme.feedbackEnabled && targetTheme.ocvFrameUrl && targetTheme.ocvAppId;
+        const showFeedbackOption = pxt.webConfig.ocv?.appId && pxt.webConfig.ocv?.iframeEndpoint;
```
thsparks, line 310: Since we do this check, I wonder if it'd be nice to have a little helper function for `ocvEnabled()` somewhere, maybe in the ocv namespace? Not a big deal either way.

react-common/components/controls/Feedback/Feedback.tsx
```diff
@@ -6,48 +6,55 @@ import { Modal } from "../Modal";
...
-        feedbackConfig={ratingFeedbackConfig}
-        frameId="activity-feedback-frame"
-        title={lf("Rate this activity")}
-        onClose={onClose} />
-      }
-    </>
+    <Modal className="feedback-modal" title={title} onClose={onClose}>
+      <>
+        { kind === "generic" &&
+          <Feedback
```
thsparks, line 31: If the differences between these two feedback components are fully based on the `kind` field, and if that `kind` field has to be passed into the Feedback component anyway, what's the motivation to split them here rather than having the `kind`-based checks inside the Feedback component itself? (In other words, could we set `frameId` inside Feedback based on kind? Then I don't think we'd need this split creating different Feedback components based on it).

Another approach would just be to have:
```
const frameId = kind === "rating" ? "activity-feedback-frame" : "menu-feedback-frame";
<Feedback kind={kind} frameId={frameId} onClose={onClose} />
```

react-common/components/controls/Feedback/Feedback.tsx
srietkerk: Oh, that's an interesting idea. Technically, in order to have this component be active at all it has to go through the menus where we already have checks to make sure that the ocv object exists. We might just want to do a check for the ocv object before the iframe in this case, because I agree, we wouldn't want a malformed url here.

webapp/src/projects.tsx
srietkerk: Ah, yeah, good idea. I definitely found it annoying to type this out every where, I'll add that in.

eanders-ms (approved):

thsparks (approved):
:shipit:

PR #10390, opened 2025-02-18
Implement keyboard and screen reader accessibility for LED matrix field

microbit-robert:
This change is compatible with the current [keyboard navigation experiment](https://www.npmjs.com/package/@blockly/keyboard-navigation) and the [new work-in-progress one](https://www.npmjs.com/package/@blockly/keyboard-experiment).

To test it with the current experience having created a showLeds block:
1. enable the experiment via About
2. enable Accessible Blocks in the cog menu
3. focus the workspace
4. use D to navigate to the block and press enter to edit

When the new experiment is integrated we'll remove the WASD support.

This field is unusual in that it is immediately editable with a stub showEditor implementation.

Detail:
- Adds grid, row and gridcell roles
- Adds switch role for LED rect elements with aria-checked
- Adds aria-labels
- Applies CSS to clear the Blockly cursor fill when the field 'shown' and focus indicators to the selected LED
- Adds keyboard controls

Controls - see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/grid_role:
- Use "Enter" to start keyboard navigation of the individual LEDs when the parent field is focussed
- Use arrow keys to change selected LED
- The left and right arrow keys overflow into the next/prev column
- Use Home to go to the start of the row and Ctrl/Cmd+Home to go to the first cell in the grid
- Use End to go to the end of the row and Ctrl/Cmd+End to go to the last cell in grid
- Use "Enter" or "Space" to toggle the LED state
- Use "Escape" to stop editing and hand focus back to the workspace

We expect to use similar navigation for other 2D structures.

This change also widens the showLeds block (and other blocks that use this field) slightly to include margin after the last column of LEDs. This centers the entire LED matrix and was required to correctly position the Blockly cursor outline.

@microbit-matt-hillsdon @riknoll

### DIFF
--- a/pxtblocks/fields/field_ledmatrix.ts
+++ b/pxtblocks/fields/field_ledmatrix.ts
@@ -142,7 +142,7 @@ export class FieldMatrix extends Blockly.Field implements FieldCustom {
                 break;
             }
             case "Escape": {
-                (Blockly.getMainWorkspace() as Blockly.WorkspaceSvg).markFocused();
+                (this.sourceBlock_.workspace as Blockly.WorkspaceSvg).markFocused();
                 return;
             }
             default: {
@@ -238,8 +238,10 @@ export class FieldMatrix extends Blockly.Field implements FieldCustom {
             }
 
             this.fieldGroup_.replaceChild(this.elt, this.fieldGroup_.firstChild);
-            this.elt.addEventListener("keydown", this.keyHandler.bind(this));
-            this.elt.addEventListener("blur", this.blurHandler.bind(this));
+            if (!this.sourceBlock_.isInFlyout) {
+                this.elt.addEventListener("keydown", this.keyHandler.bind(this));
+                this.elt.addEventListener("blur", this.blurHandler.bind(this));
+            }
         }
     }
 
--- a/pxtblocks/fields/field_ledmatrix.ts
+++ b/pxtblocks/fields/field_ledmatrix.ts
@@ -45,6 +45,7 @@ export class FieldMatrix extends Blockly.Field implements FieldCustom {
     private elt: SVGSVGElement;
 
     private currentDragState_: boolean;
+    private selected: number[] | undefined = undefined;
 
     constructor(text: string, params: any, validator?: Blockly.FieldValidator) {
         super(text, validator);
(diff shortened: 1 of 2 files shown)


pxtblocks/fields/field_ledmatrix.ts
```diff
@@ -287,10 +421,9 @@ export class FieldMatrix extends Blockly.Field implements FieldCustom {
             this.initMatrix();
         }
 
-
         // The height and width must be set by the render function
         this.size_.height = this.scale * Number(this.matrixHeight) * (FieldMatrix.CELL_WIDTH + FieldMatrix.CELL_VERTICAL_MARGIN) + FieldMatrix.CELL_VERTICAL_MARGIN * 2 + FieldMatrix.BOTTOM_MARGIN + this.getXAxisHeight()
-        this.size_.width = this.scale * Number(this.matrixWidth) * (FieldMatrix.CELL_WIDTH + FieldMatrix.CELL_HORIZONTAL_MARGIN) + this.getYAxisWidth();
+        this.size_.width = this.scale * Number(this.matrixWidth) * (FieldMatrix.CELL_WIDTH + FieldMatrix.CELL_HORIZONTAL_MARGIN) + FieldMatrix.CELL_HORIZONTAL_MARGIN + this.getYAxisWidth();
```
microbit-robert, line 426: This adds trailing margin to center the field matrix LEDs (more obvious with the Blockly cursor present), which makes this block slightly wider. 

Before:
image

After:
image

pxtblocks/plugins/renderer/css.ts
```diff
@@ -4,4 +4,8 @@ Blockly.Css.register(`
 .blocklyDropdownMenu .blocklyMenuItemCheckbox.goog-menuitem-checkbox {
     filter: contrast(0) brightness(100);
 }
+
+.blocklyVerticalMarker {
+    fill: none;
+}
```
microbit-robert, line 10: This removes the fill from the Blockly cursor to improve contrast inside the block. This normally covers a smaller area such as an input or a dropdown, but it is unhelpful for this unique, immediately editable field.

The Blockly cursor with fill:
image

microbit-robert:
The focus indicator we have used for keyboard navigation changes depending on the state of the LEDs. We need to do this to keep the focus indicator contrast acceptable against blocks of different colours, and the colour of the LED. If the LED is off, we use white, otherwise we use black and white together. See the example below with some CSS classes manually applied:

image

riknoll (changes requested):
LGTM besides the comments!

pxtblocks/fields/field_ledmatrix.ts
```diff
@@ -132,6 +238,8 @@ export class FieldMatrix extends Blockly.Field implements FieldCustom {
             }
 
             this.fieldGroup_.replaceChild(this.elt, this.fieldGroup_.firstChild);
+            this.elt.addEventListener("keydown", this.keyHandler.bind(this));
```
riknoll, line 241: check to make sure the sourceblock isn't within a flyout before registering these handlers. the outer if statement already handles insertion markers, but we don't want this messing with keyboard navigation inside the toolbox.

pxtblocks/fields/field_ledmatrix.ts
```diff
@@ -179,20 +287,42 @@ export class FieldMatrix extends Blockly.Field implements FieldCustom {
...
+    private createRow() {
+        return pxsim.svg.child(this.elt, "g", { 'role': 'row' });
+    }
+
+    private createCell(x: number, y: number, row: SVGElement) {
         const tx = this.scale * x * (FieldMatrix.CELL_WIDTH + FieldMatrix.CELL_HORIZONTAL_MARGIN) + FieldMatrix.CELL_HORIZONTAL_MARGIN + this.getYAxisWidth();
         const ty = this.scale * y * (FieldMatrix.CELL_WIDTH + FieldMatrix.CELL_VERTICAL_MARGIN) + FieldMatrix.CELL_VERTICAL_MARGIN;
 
-        const cellG = pxsim.svg.child(this.elt, "g", { transform: `translate(${tx} ${ty})` }) as SVGGElement;
+        const cellG = pxsim.svg.child(row, "g", { transform: `translate(${tx} ${ty})`, 'role': 'gridcell' });
```
riknoll, line 298: does this need some sort of disabled aria indicator if the workspace is readonly? you can test a readonly workspace by turning on the debugger

microbit-robert: Ideally, the `cellRect` one line below with `role="switch"` should have an `aria-readonly` attribute that gets updated based on the workspace being readonly. It looks like this attribute should be updated in `updateEditable`(?), however, this method isn't called when toggling the debugger, though it is called when Blockly is initialized. The difference in the cursor appearance can be noted when hovering over an LED in debugging mode vs a program loaded into the Teacher Tool, for example.

We thought it better not to include `aria-readonly` if we can't guarantee its value is going to be correct. We would welcome your input on this.

In practice, we think this does no harm, as the field editor isn't focusable in readonly mode, including in the debugger.

pxtblocks/fields/field_ledmatrix.ts
```diff
@@ -80,17 +81,121 @@ export class FieldMatrix extends Blockly.Field implements FieldCustom {
...
+                }
+                break;
+            }
+            case "Enter":
+            case "Space": {
+                this.toggleRect(x, y, !this.cellState[x][y]);
+                break;
+            }
+            case "Escape": {
+                (Blockly.getMainWorkspace() as Blockly.WorkspaceSvg).markFocused();
```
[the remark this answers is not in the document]
microbit-robert: Yes, it should! Changed here [09c91ce](https://github.com/microsoft/pxt/pull/10390/commits/09c91ce2c561f1565799792bc74cab14955626d2).

pxtblocks/fields/field_ledmatrix.ts
microbit-robert: I've made this change as it does no harm [09c91ce](https://github.com/microsoft/pxt/pull/10390/commits/09c91ce2c561f1565799792bc74cab14955626d2). However, I don't think other Blockly core field editors have any checks against this. There is a specific flyout cursor for the both the [old](https://github.com/google/blockly-samples/blob/master/plugins/keyboard-navigation/src/flyout_cursor.js#L48-L55) and [new](https://github.com/google/blockly-keyboard-experimentation/blob/main/src/flyout_cursor.ts#L46-L53) versions of the keyboard navigation plugin that does not allow the user to move `in` to the block, so it's not possible to focus the field on which these event handlers are attached.

It would be good to decide what pattern we should follow going forward as we look at more custom fields. Should we always do this check, or rely on the flyout cursor to do the right thing?

riknoll (approved):

pxtblocks/fields/field_ledmatrix.ts
riknoll: @microbit-robert if other fields don't do it, then it's fine with me!

PR #10391, opened 2025-02-19
Add Theme Support 🎨

thsparks:
# Overview
This is a first pass at adding color theme support for arcade (and other targets in the future).

See the theme README.md file (`theme\color-themes\README.md`) for details on what each theme variable is for. They are loosely based on [fluentUI](https://react.fluentui.dev/?path=/docs/theme-colors--docs), though of course, we have a much smaller set to deal with. 

Upload Target: https://arcade.makecode.com/app/8ff7826fe1b820f3793c301c11dabf0ce5bcfdc0-0927214d31

# Key Areas
Most of this change is just updating colors, but here are a few spots that deserve special attention:
1. `react-common\components\theming\themeManager.ts` - this is what handles actual theme-related operations
2. `cli\cli.ts` - this is where theme json is added to the target bundle
3. `webapp\src\app.tsx` & `pxtlib\auth.ts` -> loading / saving themes
4. `react-common\components\theming\ThemePickerModal.tsx` & related elements

# How It Works
In short, when a user changes themes, we swap out the variable definitions and apply any override css directly on a `theme-override` style element in the document head (we create it if it does not exist). See the `ThemeManager` for implementation.

# A Few Shortcuts I Took
1. I considered going through and scanning all uses of @ color variables, but it was too much. I'm planning to just address these as bug fixes if/when we notice issues.
2. I considered doing away with the color css classes too (red/green/teal/etc…) just because they seem so tightly integrated with semantic ui and run vaguely counter to the idea of theming in general…but I think they are actually quite helpful in some scenarios, so I've left them in and overridden them instead. (Also, it'd be infeasible and likely undesirable to go through each one and create bespoke classes related to the specific component.)
3. A proper high contrast theme in this new infrastructure will require taking all the .hc and .high-contrast css adjustments and putting them into the overrides file. This will be a significant chunk of work, so for now I've added a workaround that just add those css classes when switching to the hc theme specifically.

# Additional Notes
1. The term "theme" is used all over the codebase, so I've used "color theme" when I felt there was a risk of confusion. 
2. We decided to go with all css theme variables (as opposed to less) so they can be swapped out at runtime, no need for less compilation.
3. Since semantic UI has its own way of theming that is less flexible for switching at runtime, I had to override quite a lot of semantic UI styling. I've tried to encapsulate all of this in a single `react-common\styles\semantic-ui-overrides.less` file. I *think* this was still less work than removing semantic UI altogether. But it certainly isn't tidy. I didn't want to use `!important` all over the place, so I've opted for using weirdly specific css selectors instead.
4. I don't love adding a "tertiary" type alongside primary/secondary, but there were enough places where one version of the app used primary and another used secondary that it made sense to create a third option which accounted for that. The alternative was a adding bunch of css overrides to maintain consistency with existing styles, which seemed less ideal to me.
5. I haven't made any changes for kiosk, but I also didn't notice any issues testing locally, so I'm planning to leave it as-is for now.

# Remaining Work (after PR)
1. Better dark theme. Toolbox color contrast will also be a challenge here...
2. Make high contrast theme work without hc/high-contrast root classes


theme/color-themes/microbit-light.json
```diff
@@ -0,0 +1,118 @@
+{
```
thsparks, line 1: I plan to remove this theme before checking in, but it's been useful for testing

react-common/components/theming/ThemePickerModal.tsx
```diff
@@ -0,0 +1,17 @@
...
+    onThemeClicked(them: pxt.ColorThemeInfo): void;
+    onClose(): void;
+}
+export const ThemePickerModal = (props: ThemePickerModalProps) => {
+    return (
+        <Modal id="theme-picker-modal" title={lf("Choose a Theme")} onClose={props.onClose} className="theme-picker-modal">
+            <div className="ui cards centered theme-picker" role="list" aria-label={lf("List of available themes")}>
+                {props.themes && props.themes.map((theme) => <ThemeCard key={theme.id} theme={theme} onClick={props.onThemeClicked} />)}
+            </div>
+        </Modal>
```
riknoll, line 15: formatting nitpick:

```suggestion
        <Modal 
            id="theme-picker-modal" 
            title={lf("Choose a Theme")}
            onClose={props.onClose}
            className="theme-picker-modal"
        >
            <div
                className="ui cards centered theme-picker"
                role="list"
                aria-label={lf("List of available themes")}
            >
                {props.themes && props.themes.map(theme => 
                    <ThemeCard
                        key={theme.id}
                        theme={theme}
                        onClick={props.onThemeClicked}
                    />
                )}
            </div>
        </Modal>
```

react-common/components/theming/ThemeCard.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export class ThemeCard extends React.Component<ThemeCardProps> {
+    render() {
+        const { onClick, theme } = this.props;
+
+        return (
+            <div key={theme.id} className="theme-picker-item">
+                <Button
+                    className="ui card link card-selected theme-card"
```
riknoll, line 18: can we remove the semantic ui references? we've been doing a good job of keeping semantic ui out of react-common

react-common/components/theming/themeManager.ts
```diff
@@ -0,0 +1,121 @@
...
+        return pxt.appTarget?.colorThemeMap ? Object.values(pxt.appTarget.colorThemeMap) : [];
+    }
+
+    // This is a workaround to ensure we still get all the special-case high-contrast styling
+    // until we fully support high contrast via color themes (requires a lot of overrides).
+    // TODO : this should be removed once we do fully support it.
+    private performHighContrastWorkaround(themeId: string) {
+        let setBodyClass = (add:boolean, className: string) => {
+            const body = this.document.body;
+            const hasClass = body.classList.contains(className);
```
riknoll, line 43: nitpick: no need to check contains here, classList.add already takes care of duplicates and classList.remove doesn't throw if missing

react-common/styles/semantic-ui-overrides.less
```diff
@@ -0,0 +1,448 @@
+
+/*
+--- Override Semantic UI ---
+
+This file overrides the default semantic ui theming with our own theme variables.
+
+*/
+
+// Mixin allows easy setting for background, foreground, hover, and inverted colors for a button
+.buttonVariant(@bg; @fg; @hoverBg; @hoverFg) {
```
riknoll, line 10: i think this file should be in theme and not react-common, since react-common doesn't depend on semantic ui.

react-common/styles/theming/base-theme.less
```diff
@@ -0,0 +1,111 @@
+.pxt-theme-root {
```
riknoll, line 1: should this file be built from a json file by the cli when a target is built?

react-common/components/theming/ThemeCard.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from "react";
+import { ThemePreview } from "./ThemePreview";
+import { Card } from "../controls/Card";
+
+interface ThemeCardProps {
+    theme: pxt.ColorThemeInfo;
+    onClick?: (theme: pxt.ColorThemeInfo) => void;
+}
+
+export class ThemeCard extends React.Component<ThemeCardProps> {
```
eanders-ms, line 10: Just curious, why did you choose to make this a React component?

thsparks: Was just on autopilot, didn't really realize there was an alternative.

After some searching, am I correct in understanding that since it doesn't use state, it can just be something like `const ThemeCard = (props: ThemeCardProps) => { ... }`?

thsparks: Wow, I already did it in my other components and didn't realize 😅 Will change this one to match.

abchatra:
Awesome to see this. Ok to assign copilot to review this change (I am curious what it will do) ?

thsparks:
> Awesome to see this. Ok to assign copilot to review this change (I am curious what it will do) ?

Sure, I'll assign it

react-common/components/theming/ThemePreview.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+// Programmatically generate a preview of the theme using theme colors.
+export const ThemePreview = (props: { theme: pxt.ColorThemeInfo }) => {
+    const { theme } = props;
+    const styleRef = React.useRef<HTMLStyleElement | null>(null);
+    const uniqueContainerClassName = `theme-preview-container-${theme.id}`;
+    const uniqueInnerClassName = `theme-preview-${theme.id}`; // Useful for override css adjusting previews
+
+    const miniLogo = <img className="ui logo" src="./static/Micorsoft_logo_rgb_W-white_D-square.png" alt="Microsoft MakeCode Logo" />;
+
```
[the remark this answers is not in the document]
thsparks: Asset name is also misspelled, but while I was double checking this, I found that the asset I'm referencing is arcade-specific. There is a pxt-core one with the correct spelling under the /logo/ folder, so I'll change to that.

riknoll (approved):

PR #10398, opened 2025-02-26
Order Themes in Theme Picker

thsparks:
This allows us to specify a weight for each theme. Themes will be ordered with higher weights at the bottom. If a theme does not have a weight, it is put last.

image

I've also moved when themes are loaded to before the page is shown, so there is no time where the default theme will be visible (unless something goes wrong).

Fixes https://github.com/microsoft/pxt-arcade/issues/6708


srietkerk (approved):
Nice!

PR #10399, opened 2025-02-26
Use Default Theme in Multiplayer

thsparks:
Themes impact multiplayer in a few places (though not many, since it primarily uses Tailwind). Notably, the settings dropdown and a few buttons. Actually supporting themes would be a big undertaking, so for now, I've just set it up to always load the default theme.

With themes, the presence buttons were getting dark backgrounds because they were disabled. This looked odd, so I've added a tailwind class to force them back to their white backgrounds instead.

I also had to fix a bug in our multiplayer serve that assumed all webconfig values were strings (no longer the case with ocv).

Upload Target: https://arcade.makecode.com/app/bd228f8101401d913f282b0c3585b210d67a903b-d766c6703c--multiplayer

Fixes https://github.com/microsoft/pxt-arcade/issues/6722


multiplayer/src/index.tsx
```diff
@@ -49,7 +49,7 @@ window.addEventListener("DOMContentLoaded", () => {
         const wc = pxt.webConfig as any;
 
         for (const key of Object.keys(wc)) {
-            if (wc[key]?.startsWith("/") && wc[key]?.indexOf("worker") == -1) {
+            if (typeof wc[key] === "string" && wc[key]?.startsWith("/") && wc[key]?.indexOf("worker") == -1) {
```
srietkerk, line 52: Hmm, this makes me a bit worried that we have this constraint other places.. I'll have to take a look around to make sure I'm not breaking anything else. Thanks for fixing this here.

srietkerk (approved):
Looks good! One thing I saw on the upload target is that since the button colors are changed to the default, the underline color for Join and Host Game are now different from the button. Not blocking, but I wanted to bring it up in case it was something we wanted to change.

Live:
image

New:
image

thsparks:
> One thing I saw on the upload target is that since the button colors are changed to the default, the underline color for Join and Host Game are now different from the button. Not blocking, but I wanted to bring it up in case it was something we wanted to change.

Good catch. Turns out there's some primary/secondary/tertiary color usage that I didn't notice. I'll update those.

multiplayer/src/index.tsx
thsparks: No worries. TBH I don't think ocv is at fault here. Just an unsafe assumption in the code. ocv wasn't actually the first boolean value added to the config, just the first one that was actually set at this point.

PR #10405, opened 2025-03-05
Improve keyboard accessibility of the codecard component

microbit-robert:
This affects code cards on the experiments page. At the moment, it is not possible to trigger the feedback link of these cards via keyboard, this action simple toggles the experiment on and off. This PR addresses this and also improves the appearance of the focus-visible outline around the "Beta Editor" card.

@microbit-matt-hillsdon

### DIFF
--- a/webapp/src/codecard.tsx
+++ b/webapp/src/codecard.tsx
@@ -81,6 +81,13 @@ export class CodeCardView extends data.Component<CodeCardProps, CodeCardState> {
             card.onClick(e);
         } : undefined;
 
+        const keydownHandler = (e: React.KeyboardEvent) => {
+            const charCode = (typeof e.which == "number") ? e.which : e.keyCode;
+            if (charCode === /*enter*/13 || charCode === /*space*/32) {
+                clickHandler(e);
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


webapp/src/codecard.tsx
```diff
@@ -95,9 +102,18 @@ export class CodeCardView extends data.Component<CodeCardProps, CodeCardState> {
         const ariaLabel = card.ariaLabel || card.title || card.shortName || name;
 
         const style = card.style || "card"
-        const cardDiv = <div className={`ui ${style} ${color} ${card.onClick ? "link" : ''} ${className ? className : ''}`}
-            role={card.role} aria-selected={card.role === "option" ? "true" : undefined} aria-label={ariaLabel} title={card.title}
-            onClick={clickHandler} tabIndex={card.onClick ? card.tabIndex || 0 : null} onKeyDown={card.onClick ? fireClickOnEnter : null}>
```
microbit-robert, line 100: `fireClickOnEnter` does the wrong thing here when using "Enter" or "Space" while the "Feedback" anchor tag is focussed. It prevents the default behaviour, then triggers click on this element instead of the anchor tag from which the original event was triggered, causing the experiment to toggle on/off.

riknoll (approved):

PR #10414, opened 2025-03-07
setting size and flash top config

tballmsft:
this PR has two major changes to allow more efficient use of flash memory
1. the ability to configure the size of the libs/settings extension via Static TypeScript (using getConfig)
2. communication of the top of usable flash from the CODAL/C++ toolchain to MakeCode (as under different compile options to C++, we may get different top of flash)

the second part puts the top of flash (FLASH_TOP from compile) into the PXT shim table
```
#define PXT_SHIMS_BEGIN                                                                            \
    namespace pxt {                                                                                \
    const uintptr_t functionsAndBytecode[]                                                         \
        __attribute__((aligned(0x20))) = {0x08010801, 0x42424242, 0x08010801, 0x8de9d83e, FLASH_TOP,
```
This requires changes to hexfile.ts to extract FLASH_TOP from the hex file (or ELF) produced by the C++ compiler tool chain

Related PRs:
- https://github.com/microsoft/pxt-common-packages/pull/1502
- https://github.com/microsoft/pxt-microbit/pull/6110

TODO:
- make sure this works also for pxt-arcade


mmoskal (approved):
LGTM!

PR #10415, opened 2025-03-07
A couple of function dialog fixes

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/6764
fixes https://github.com/microsoft/pxt-arcade/issues/6770

a couple of fixes for the function dialog! 

first one is a bug with the initialization of the argument editors for custom non-primitive types. i ran into the same thing when doing https://github.com/microsoft/pxt/pull/10402; basically initSvg was getting caused out of order which led to weirdness. this change just switches them to be the same as all of the other types rather than using domToBlock

second one was just an oversight when i rewrote the functions implementation during the great blockly swap. i guess i never tested this!

### DIFF
--- a/pxtblocks/plugins/functions/fields/fieldArgumentEditor.ts
+++ b/pxtblocks/plugins/functions/fields/fieldArgumentEditor.ts
@@ -1,21 +1,18 @@
 import * as Blockly from "blockly";
 import { FunctionManager } from "../functionManager";
+import { ArgumentEditorBlock } from "../blocks/argumentEditorBlocks";
 
 export interface FieldArgumentEditorConfig extends Blockly.FieldTextInputConfig {
-    typeName?: string;
 }
 
 export class FieldArgumentEditor extends Blockly.FieldTextInput {
     static REMOVE_ARG_URI =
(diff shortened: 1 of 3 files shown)


pxtblocks/plugins/functions/fields/fieldArgumentEditor.ts
```diff
@@ -1,21 +1,18 @@
 import * as Blockly from "blockly";
 import { FunctionManager } from "../functionManager";
+import { ArgumentEditorBlock } from "../blocks/argumentEditorBlocks";
 
 export interface FieldArgumentEditorConfig extends Blockly.FieldTextInputConfig {
```
srietkerk, line 5: Do we still need this as a separate interface if it no longer adds something to the Blockly.FieldTextInputConfig?

eanders-ms (approved):
LGTM!

pxtblocks/plugins/functions/fields/fieldArgumentEditor.ts
riknoll: no, but maybe someday!

PR #10417, opened 2025-03-08
Theme Keyboard Nav & Melody Editor Fix

thsparks:
This adds keyboard nav support for the theme picker modal (just needed to add the cards into the tab ordering), and it fixes the bpm box in the melody editor (since the whole thing is hard-coded to block colors, we can just hardcode the input color as well).

Fixes https://github.com/microsoft/pxt-arcade/issues/6773
Fixes https://github.com/microsoft/pxt-arcade/issues/6758

### DIFF
--- a/react-common/components/theming/ThemeCard.tsx
+++ b/react-common/components/theming/ThemeCard.tsx
@@ -13,8 +13,10 @@ export const ThemeCard = (props: ThemeCardProps) => {
         <Card
             className="theme-card"
             role="listitem"
+            aria-label={theme.name}
             key={theme.id}
             onClick={() => onClick(theme)}
+            tabIndex={onClick && 0}
         >
             <div className="theme-info-box">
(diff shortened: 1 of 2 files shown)


react-common/components/theming/ThemeCard.tsx
```diff
@@ -13,8 +13,10 @@ export const ThemeCard = (props: ThemeCardProps) => {
         <Card
             className="theme-card"
             role="listitem"
+            aria-label={theme.name}
             key={theme.id}
             onClick={() => onClick(theme)}
+            tabIndex={onClick && 0}
```
srietkerk, line 19: Should `onClick` just be necessary?

thsparks: It could be, but I don't think the check hurts anything. I'm following a pattern that we already have in the extension cards.

srietkerk (approved):

PR #10418, opened 2025-03-10
High Contrast Fixes

thsparks:
This fixes a couple of high contrast bugs:

Fixes https://github.com/microsoft/pxt-arcade/issues/6780
Now that this are using react-common buttons, some of the default button css was conflicting with the programmatically-set colors for the color selection tool, and a few of the other buttons just looked odd. As far as I could find, you can't set `!important` programmatically, so I used a var to set the preview color and set important in css. I also changed the top, bottom, and side bars to all have black backgrounds, which is different from the released version but I think more in-line with high contrast in general.

Fixes https://github.com/microsoft/pxt-arcade/issues/6771
This fix was just updating some of the .hc specific css, but while I was there, I also moved it into the high-contrast theme's override css, since I'd like to start moving all that special-case stuff into one theme-supported location. Behavior now matches live, where the icon is colored-in when unselected and turns white when selected.

Upload Target: <s>https://arcade.makecode.com/app/82cd97d78761676ef6ad9029b9a877d2c77f3d88-474593e9d6</s> https://arcade.makecode.com/app/3d46dab3d87f5e72c678ea1dcc0e161e4df02769-de076282e5

### DIFF
--- a/react-common/styles/controls/Button.less
+++ b/react-common/styles/controls/Button.less
@@ -290,7 +290,7 @@ a.common-button.menu-button {
 .high-contrast, .hc {
     .common-button {
         color: @highContrastTextColor !important;
-        background: @highContrastBackgroundColor !important;
+        background-color: @highContrastBackgroundColor !important;
         border-color: @highContrastTextColor !important;
 
         &:hover, &:focus {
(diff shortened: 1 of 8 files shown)


srietkerk:
Not sure how I've missed this before and not sure if you want to look into it with this PR or a separate PR, but it looks like the multiplayer icons are askew in high contrast on the upload target. Checked the other themes and it seems like they're fine.

image

srietkerk:
Should these buttons/interaction areas have white borders around them? They have nice borders on hover, just wasn't sure if we wanted the white border by default as well.

image

riknoll (approved):

srietkerk (approved):
Nice!

PR #10420, opened 2025-03-10
allow debug to be toggled from any file open in monaco editor

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/6634

removes the restriction where the debug state can only be toggled from within main.py, main.ts, or main.blocks

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -602,6 +602,11 @@ export class ProjectView
             && this.editorFile && this.editorFile.name == pxt.ASSETS_FILE;
     }
 
+    isJavaScriptEditorActive() {
+        return !this.state.embedSimView && this.editor == this.textEditor
+            && this.editorFile && !this.isPythonActive();
+    }
+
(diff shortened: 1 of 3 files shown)


webapp/src/app.tsx
```diff
@@ -602,6 +602,11 @@ export class ProjectView
             && this.editorFile && this.editorFile.name == pxt.ASSETS_FILE;
     }
 
+    isJavaScriptEditorActive() {
```
srietkerk, line 605: I think we need to update the `isPythonActive` function as well. It looks like debug is disabled on custom `.py` files:
image

srietkerk: Is it possible to put python files in extensions? Should you be able to use those files in the debugger?

thsparks (approved):
Agree w/ Sarah that it'd be more intuitive if we had the same behavior in python (even if multiple file support is buggy), but if that's a pain to do, not blocking.

PR #10421, opened 2025-03-11
Skillmap bug fixes

srietkerk:
Fixes https://github.com/microsoft/pxt-arcade/issues/6311
- Moved the feedback to be under the settings menu as well in order to make it consistent with the non-skillmap editor experience

Fixes https://github.com/microsoft/pxt-arcade/issues/6229
- Makes the "kind" in share info "share" by default

Fixes https://github.com/microsoft/pxt-arcade/issues/3601
- With this, also updated the other misspellings of skillmap

Upload target: https://arcade.makecode.com/app/1ba78fc22d929a7477a9ea676cfaf41985914468-3784045f59--skillmap#mole


thsparks (approved):
Renaming the package makes me nervous, but I don't know of any concrete reason why it'd be an issue in this case. Seems good to me otherwise!

PR #10424, opened 2025-03-11
Fix Sim Button Toggle, Card Outlines, and Keymap Theming

thsparks:
Fixes https://github.com/microsoft/pxt-arcade/issues/6761
Mute and debug buttons now switch to a darker shade of the same color when toggled on, rather than a separate bright color. Requires changes in pxt-arcade also necessary to update secondary-accent colors.

Fixes https://github.com/microsoft/pxt-arcade/issues/6788
This just wasn't themed before. I've updated the css to use theme colors.

Fixes https://github.com/microsoft/pxt-arcade/issues/6789
Just needed to add a border in the semantic ui overrides.

Upload target: https://arcade.makecode.com/app/a2b90411277dc31bf416b4bc830035728af40d75-0cd92e0a68

### DIFF
--- a/theme/color-themes/overrides/high-contrast-overrides.css
+++ b/theme/color-themes/overrides/high-contrast-overrides.css
@@ -13,6 +13,15 @@
     filter: invert(1);
 }
 
+/* Sim toolbar */
+#simulator .editor-sidebar .simtoolbar .debug-button.active,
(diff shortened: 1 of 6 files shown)


eanders-ms (approved):
👍

PR #10429, opened 2025-03-12
improve music editor erase tool

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/5694

this PR does a few things to make the music editor erase tool more friendly:
* you can now start the drag for the erase tool in the area above the staff (the part where the measure numbers are). previously drags that started there were just ignored
* the erase cursor is now "wider", in other words we now check all notes that overlap with 1 tick before and after the drag coordinate
* the eraser now interpolates from the last coordinate in the drag which prevents us from skipping notes when the mouse is moving really fast

finally i also fixed an unrelated bug where dragging drum tracks on the bass clef was registering notes in the treble clef. i don't know if this is filed anywhere; i just noticed it while debugging.

### DIFF
--- a/webapp/src/components/musicEditor/MusicEditor.tsx
+++ b/webapp/src/components/musicEditor/MusicEditor.tsx
@@ -276,29 +276,39 @@ export const MusicEditor = (props: MusicEditorProps) => {
             }
         }
         else {
+            let editSong = currentSong;
+            let didEdit = false;
             for (let i = 0; i < currentSong.tracks.length; i++) {
                 if (hideTracksActive && i !== selectedTrack) continue;
 
-                const existingEvent = findPreviousNoteEvent(currentSong, i, coord.tick);
-                const closestToClickEvent = findPreviousNoteEvent(currentSong, i, coord.exactTick);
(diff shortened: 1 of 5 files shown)


webapp/src/components/musicEditor/MusicEditor.tsx
```diff
@@ -367,27 +377,63 @@ export const MusicEditor = (props: MusicEditorProps) => {
...
-                const existingEvent = findPreviousNoteEvent(currentSong, i, end.exactTick);
-
-                const isAtCoord = (note: pxt.assets.music.Note) => {
-                    const isBassClef = isBassClefNote(instrument.octave, note, isDrumTrack);
-
-                    if (isBassClef !== end.isBassClef) return false;
-
-                    return noteToRow(instrument.octave, note, isDrumTrack) === end.row;
-                }
+            if (end.row > 11 || end.row < 0) return;
```
thsparks, line 380: Not sure I'm quite following what the 11 & 12 magic numbers are coming from. Could these be constants with meaningful names?

thsparks (approved):

PR #10432, opened 2025-03-12
add a default option to the cannot pair dialog

srietkerk:
Fixes https://github.com/microsoft/pxt-arcade/issues/6655

The `getReasonUnavailable` function that the switch statement relies on for the Cannot Pair dialog content can return undefined. There was no case covering the undefined option, so I added a default case for defining the modal body content. It looks like it would be pretty hard to hit this case in micro:bit with the playing that I did, but in case any other target also hits the undefined scenario, I wanted to make sure that we had some more generic messaging in the content.

As mentioned in the issue, on live the "Connect Device" button only shows up if you have a device connected and chosen in the hardware dialog. Changing the flags around to avoid showing this option before hardware was selected didn't seem to do anything, so I figured the instructions could guide the user to what they need to do to successfully connect a device with WebUSB.

image

Upload target: https://arcade.makecode.com/app/06dba66f98e73697706c802bf16000f9899d64c4-cf4ba796f5#

### DIFF
--- a/pxtlib/browserutils.ts
+++ b/pxtlib/browserutils.ts
@@ -143,6 +143,10 @@ namespace pxt.BrowserUtils {
         return isPxtElectron() || isIpcRenderer();
     }
 
+    export function isArcade() {
+        return pxt.appTarget.id === "arcade";
+    }
+
     declare let Windows: any;
     export let isWinRT = () => typeof (Windows as any) !== "undefined";
 
--- a/pxtlib/browserutils.ts
+++ b/pxtlib/browserutils.ts
@@ -143,10 +143,6 @@ namespace pxt.BrowserUtils {
         return isPxtElectron() || isIpcRenderer();
     }
 
-    export function isArcade() {
-        return pxt.appTarget.id === "arcade";
-    }
-
     declare let Windows: any;
     export let isWinRT = () => typeof (Windows as any) !== "undefined";
 
--- a/webapp/src/editortoolbar.tsx
+++ b/webapp/src/editortoolbar.tsx
@@ -211,6 +211,12 @@ export class EditorToolbar extends data.Component<ISettingsProps, EditorToolbarS
     protected onCannotPairClick = async () => {
         pxt.tickEvent("editortools.pairunsupported", undefined, { interactiveConsent: true });
         const reasonUnsupported = await pxt.usb.getReasonUnavailable();
+
+        if (!reasonUnsupported && pxt.BrowserUtils.isArcade() && !pxt.getActiveHwVariant()) {
+            this.onHwItemClick()
+            return;
+        }
+
         let modalBody: string;
         switch (reasonUnsupported) {
             case "security":
@@ -226,12 +232,7 @@ export class EditorToolbar extends data.Component<ISettingsProps, EditorToolbarS
                 modalBody = lf("WebUSB is not supported by this browser; please check for updates.");
                 break;
             default:
-                if (pxt.BrowserUtils.isArcade()) {
-                    this.onHwItemClick();
-                    return;
-                } else {
-                    modalBody = lf("WebUSB device not detected. Make sure your device is connected and try again.");
-                }
+                modalBody = lf("WebUSB should be supported, but we're having trouble. Please try refreshing the page.");
                 break;
         }
 
--- a/webapp/src/editortoolbar.tsx
+++ b/webapp/src/editortoolbar.tsx
@@ -227,9 +227,10 @@ export class EditorToolbar extends data.Component<ISettingsProps, EditorToolbarS
                 break;
             default:
                 if (pxt.BrowserUtils.isArcade()) {
-                    modalBody = lf("WebUSB device not detected. Select your device in the 'Choose Hardware' menu option and try to connect again.");
+                    this.onHwItemClick();
+                    return;
                 } else {
-                    modalBody = lf("WebUSB device not detected. Make sure your device is connected to your computer and try again.");
+                    modalBody = lf("WebUSB device not detected. Make sure your device is connected and try again.");
                 }
                 break;
         }
--- a/webapp/src/editortoolbar.tsx
+++ b/webapp/src/editortoolbar.tsx
@@ -225,6 +225,13 @@ export class EditorToolbar extends data.Component<ISettingsProps, EditorToolbarS
             case "notimpl":
                 modalBody = lf("WebUSB is not supported by this browser; please check for updates.");
                 break;
+            default:
+                if (pxt.BrowserUtils.isArcade()) {
+                    modalBody = lf("WebUSB device not detected. Select your device in the 'Choose Hardware' menu option and try to connect again.");
+                } else {
+                    modalBody = lf("WebUSB device not detected. Make sure your device is connected to your computer and try again.");
+                }
+                break;
         }
 
         dialogAsync({
--- a/webapp/src/editortoolbar.tsx
+++ b/webapp/src/editortoolbar.tsx
@@ -227,7 +227,7 @@ export class EditorToolbar extends data.Component<ISettingsProps, EditorToolbarS
                 modalBody = lf("WebUSB is not supported by this browser; please check for updates.");
                 break;
             default:
-                modalBody = lf("WebUSB should be supported, but we're having trouble. Please try refreshing the page.");
+                modalBody = lf("Unable to connect to WebUSB. Please try refreshing the page.");
                 break;
         }
 
--- a/webapp/src/editortoolbar.tsx
+++ b/webapp/src/editortoolbar.tsx
@@ -212,11 +212,6 @@ export class EditorToolbar extends data.Component<ISettingsProps, EditorToolbarS
         pxt.tickEvent("editortools.pairunsupported", undefined, { interactiveConsent: true });
         const reasonUnsupported = await pxt.usb.getReasonUnavailable();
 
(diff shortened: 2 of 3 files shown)


webapp/src/editortoolbar.tsx
```diff
@@ -225,6 +225,13 @@ export class EditorToolbar extends data.Component<ISettingsProps, EditorToolbarS
             case "notimpl":
                 modalBody = lf("WebUSB is not supported by this browser; please check for updates.");
                 break;
+            default:
```
jwunderl, line 228: would

```typescript
        (this.props.parent as ProjectView).chooseHwDialog.show();
```

(similar to `onHwDownloadClick` above) work here?

webapp/src/editortoolbar.tsx
```diff
@@ -225,6 +225,13 @@ export class EditorToolbar extends data.Component<ISettingsProps, EditorToolbarS
...
+                if (pxt.BrowserUtils.isArcade()) {
+                    modalBody = lf("WebUSB device not detected. Select your device in the 'Choose Hardware' menu option and try to connect again.");
```
thsparks, line 230:
```suggestion
                    modalBody = lf("WebUSB device not detected. Select your device in the 'Choose Hardware' menu and try again.");
```

thsparks (commented):
Some wording nits, though I think Joey's recommendation makes sense so perhaps the wording changes are unnecessary :)

webapp/src/editortoolbar.tsx
srietkerk: I just played with this and it doesn't really give us the desired behavior. If I just show the hardware dialog and then return, avoiding the "Cannot Connect Device" dialog, when you select the hardware, it automatically downloads a file version of your program which takes a while instead of prompting the user for usb stuff. 

I can play with this option a bit to see what I can come up with.

srietkerk: Another thing I worry about is that showing the choose hardware option off the bat seems buggy and that choosing the different download options will just do the same thing.

thsparks: Isn't that basically what your instructions are telling them to do?

jwunderl: ah so arcade does nothave target flag set to have webusb as default experience, so it fallbacks to download as first experience, makes sense. Could add another flag by following path `skipDownload` flag goes starting in `app.tsx`, and have that initiate pair on completion instead of compile / download, though that may make this one take a bit longer since the path is a lil long

srietkerk: Oh, using `this.onHwItemClick()` looks like it works. The important part is that skipping the download is set to true, which it is in that function. With using the `{...}.hwDialog.show()` method, we use the default of `skipDownload` which is false, which does the compile and file download.

srietkerk: The one thing that is kinda a downside of using the `onHwItemClick` route is that you get a very different experience after you've connected and disconnected a device.
If y'all could play with it and see what you think, that would be super helpful: https://arcade.makecode.com/app/2cf11bbee78e6da5c33395047ede2b010faaccec-e3d341ab60#.

Steps:
1. Connect to a WebUsb device under the connect device option (get the hardware dialog)
2. Disconnect that device
3. Select connect device again

srietkerk: You can also compare the experience with this upload target with the one in the PR comment and see what's better.

thsparks: The experiences after disconnecting seemed the same to me. Maybe I'm missing something?

But I'd say I much prefer the experience where it automatically pops up the "choose device" dialog. Saves me 3 steps (reading the instructions, going to click the "choose hardware" manually, then going back and re-clicking the "connect device" again). Handling that automatically and just directly asking which device I have feels much more intuitive.

srietkerk: Yeah, I'm more thinking of the fact that when you click the connect device the first time, you get the dialog to select hardware. Once you disconnect and reconnect, you then only get the modal that tells you to connect via webusb.

I also had an experience yesterday that I was playing with that triggered the pair dialog after the hardware was chosen and I thought it might be a bit weird to have to choose hardware in one case and not the other. That experience was pretty buggy, though, and I think I'd want to leave it with just opening the hardware dialog if hardware hasn't been chosen yet.

webapp/src/editortoolbar.tsx
```diff
@@ -225,6 +225,14 @@ export class EditorToolbar extends data.Component<ISettingsProps, EditorToolbarS
             case "notimpl":
                 modalBody = lf("WebUSB is not supported by this browser; please check for updates.");
                 break;
+            default:
+                if (pxt.BrowserUtils.isArcade()) {
```
thsparks, line 229: Instead of this isArcade check, should we be adding a new reason that can be returned from `pxt.usb.getReasonUnavailable()`?

srietkerk: Sure!

srietkerk: Okay, I looked into this a bit more and a new reason should not be added to `getReasonUnavailable` in this case, unfortunately. 

`getReasonUnavailable` is specifically checking if a user's environment is capable of supporting webusb. It is really important that the function returns `undefined` if the user's environment is webUSB capable because there's a separate function that also calls `getReasonUnavailable` that says "if the unavailability is undefined, then webusb must be available". That setting webusb to available is crucial because it sets `pxt.usb.isEnabled` to true (in a bit of a roundabout way). 

When `pxt.usb.isEnabled` is true, we get the correct pairing experience to show up with the "Connect Device" menu option. Without hardware selected in arcade, `pxt.usb.isEnabled` is false, which is why we get the `onCannotPairClick`/switch statement experience. However, we are getting into this experience a bit presumptuously as "webusb is disabled" doesn't necessarily mean "webusb not supported". But we handle what modal experience the user gets in a kinda messy way. If you look at where `onCannotPairClick` you'll see what I mean. We can make another possible option in the drop down selection for when webusb hasn't been enabled yet--because we haven't chosen compatible hardware--and it's arcade, but I don't _really_ like the thought of adding something else to that list and I think `onCannotPairClick` can handle it fine.

All that being said, the arcade scenario where webusb can be used but you need to select hardware first should be handled _outside_ of the switch statement because the switch statement and the reasons returned by `getReasonUnavailable` handle the scenario where a user's environment physically cannot support webusb.

webapp/src/editortoolbar.tsx
```diff
@@ -225,6 +231,9 @@ export class EditorToolbar extends data.Component<ISettingsProps, EditorToolbarS
             case "notimpl":
                 modalBody = lf("WebUSB is not supported by this browser; please check for updates.");
                 break;
+            default:
+                modalBody = lf("WebUSB should be supported, but we're having trouble. Please try refreshing the page.");
```
thsparks, line 235:
```suggestion
                modalBody = lf("Unable to connect to WebUSB. Please try refreshing the page.");
```

thsparks: I worry to promise "should be supported" when we don't know all the details about the user's machine/setup.

thsparks (approved):
One nit on the default message, otherwise looks good!

webapp/src/editortoolbar.tsx
thsparks: Thanks for looking into it!

pxtlib/browserutils.ts
```diff
@@ -143,6 +143,10 @@ namespace pxt.BrowserUtils {
         return isPxtElectron() || isIpcRenderer();
     }
 
+    export function isArcade() {
```
jwunderl, line 146: typically we've avoided target specific checks in pxt core (e.g. there's one in `field_soundEffect.ts`, but with a fixme associated with it -- if possible, basing off the fact that there are hardware variants and one hasn't been selected yet, or adding a separate 'prompt for hardware selection' pxtarget flag worst case may be better.

jwunderl (approved):
one quibble but up to you

pxtlib/browserutils.ts
srietkerk: So you're saying that I don't need the extra arcade check? The "no hardware variant chosen" check should be sufficient enough?

jwunderl: ideally, if we cloned arcade repo & just changed id, behavior would be the same. Looking at it that seems safe with `!pxt.getActiveHwVariant`, though maybe need a check on `pxt.appTarget.variants` being defined, so that it doesn't occur if there *aren't* variants, if the current check doesn't account for that.

srietkerk: Ah, that's a good point, I should check for variants before checking for the active variant. That said, I don't know if just checking for a variants object is going to be enough to differentiate in this scenario. I don't think we'd want to pop open hardware dialog in microbit, but microbit also has a variants object for dal vs codal.

srietkerk: I'd rather not create another flag, but I think I found a flag that would actually work well in this situation. Microbit has a different connect device experience because we have the `downloadDialogTheme` defined there. If we have a download theme defined, then we wouldn't want to pop up the hardware choosing option, we would always want to show the defined experience. So, in addition to checking that variants exist, I will also check that a download theme doesn't exist before showing the hardware picker modal.

jwunderl: that sounds like would make sense yeah; this `isArcade` check is just a target flag that doesn't live in pxtarget, so any way you can differentiate is probably reasonable enough for this use case / can be adjusted in future if need be.

thsparks (approved):

PR #10437, opened 2025-03-15
fix for skillmap projects visible in main project list

eanders-ms:
Fixes https://github.com/microsoft/pxt-arcade/issues/6820


riknoll (approved):

srietkerk:
Ah, this came from me updating the flag to be skillmap. Should I look to see if we have any other regex that look like this one?

eanders-ms:
> Ah, this came from me updating the flag to be skillmap. Should I look to see if we have any other regex that look like this one?

I think this was the only one.

srietkerk:
Thank you for finding this!

PR #10441, opened 2025-03-18
on localhost, fallback to file download when api/deploy fails

eanders-ms:
(low priority, can wait for arcade release)
I wanted to download a built uf2 file when running on localhost, but the localhost environment installs a custom deploy path that publishes thru the cli. If that fails, it will now download to file.


eanders-ms:
Bumping, now that Arcade has been released.

jwunderl (approved):

PR #10444, opened 2025-03-18
Make Theme Names Localizable

thsparks:
This adds theme names to our target strings file for translation and adjusts our theme-picker modal to lookup the localized name, when available.

Fixes https://github.com/microsoft/pxt-arcade/issues/6806


riknoll (approved):

srietkerk (approved):
Nice! I haven't seen this use of localization before, so I learned a bit about how it fits together 😄!

PR #10446, opened 2025-03-19
Home screen keyboard focus

glastonbridge:
Addresses the lack of contrast on some of focus-visible highlights on the home screen, as well as making the carousel buttons a single tab stop, which brings it closer to [the WAI-ARIA recommendations for tabbed carousels](https://www.w3.org/WAI/ARIA/apg/patterns/carousel/#wai-ariaroles,states,andproperties).

[Demo](https://home-screen-keyboard-focus.review-pxt.pages.dev/)

### Before

Using tab to navigate

https://github.com/user-attachments/assets/9dc2b409-cd8e-4883-8b53-aecdc8de2fe5

### After

https://github.com/user-attachments/assets/b714ec3a-6aa6-4b9f-a3fd-9253a4d315b0

### High contrast theme

https://github.com/user-attachments/assets/2a6ee186-aabd-4f2f-9287-060e08666116

### DIFF
--- a/theme/home.less
+++ b/theme/home.less
@@ -90,16 +90,39 @@
 
         .action {
             margin: 1rem 0;
+            a:focus-visible, button:focus-visible {
+                outline-width: 4px !important;
+                outline-style: solid;
+                outline-color: var(--pxt-neutral-stencil3);
+            }
         }
 
-        .dots button {
-            transition: background-color 0.5s;
-            border: 2px solid @heroBannerDotOutlineColor;
-            margin-right: 1.25rem;
+        .dots {
+            align-self: flex-start;
+            border-radius: 50px;
+            line-height: 0;
+            margin-top: 8px;
 
-            &.active {
-                border-color: @heroBannerDotOutlineAtiveColor;
-                background-color: transparent;
+            &:focus-visible {
+                outline-width: 3px !important;
+                outline-style: solid;
+                outline-color: white;
+                outline-offset: 4px;
+            }
+
+            button {
+                transition: background-color 0.5s;
+                border: 2px solid @heroBannerDotOutlineColor;
+                margin-right: 1.25rem;
+
+                &.active {
+                    border-color: @heroBannerDotOutlineAtiveColor;
+                    background-color: transparent;
+                }
+
+                &:last-child {
+                    margin-right: 0px;
+                }
             }
         }
 
@@ -152,6 +175,12 @@
                     &:hover, &:focus {
                         border-color: var(--pxt-primary-background);
                     }
+                    &:focus-visible {
+                        outline-width: @homeCardBorderSize !important;
+                        outline-style: solid;
+                        outline-color: var(--pxt-primary-background);
+                        outline-offset: 2px;
+                    }
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


thsparks (approved):
Changes look great, thank you!

theme/home.less
```diff
@@ -173,6 +202,12 @@
     .import-dialog-btn {
         position: relative;
         z-index: 1; /* Move up so it's above the carousel container that has an offset margin */
+        &:focus-visible {
+            outline-width: @homeCardBorderSize !important;
+            outline-style: solid;
+            outline-color: var(--pxt-primary-background);
```
thsparks, line 208: For this one, I'm wondering if var(--pxt-focus-border) would be a more appropriate color. Would you be open to trying that and seeing if it still aligns to standards?

If we add more themes in the future, the primary color may vary, but the "focus border" should remain a fairly blue across all themes.

It makes sense to deviate in the other areas of this PR, since the content being outlined is also blue (or on a blue background), so a white outline makes more sense there. But in this particular case I _think_ the focus border may work well. But LMK if you disagree!

theme/home.less
```diff
@@ -449,7 +483,9 @@
     }
     .ui.card:focus {
         border: 2px solid transparent;
-        outline: @homeCardBorderSize solid var(--pxt-primary-background) !important;
+        outline-width: @homeCardBorderSize !important;
+        outline-style: solid;
+        outline-color: var(--pxt-primary-background);
```
thsparks, line 488: Same here, wondering if var(--pxt-focus-border) would work.

thsparks (commented):
Took a second look and a few comments came up. LMK what you think and we can get this checked in :)

glastonbridge:
Hi @thsparks !

Good catch! I had missed the designated focus border var I think.

### With --focus-border-color (proposed)

focus-color-blue

### With --pxt-primary-background (currently in this PR)

focus-color-purple

I'm not sure which is right in practice. From a pure accessibility standpoint I favour the purple as it is more visible, particularly against the New Project card, which otherwise eclipses the light blue of the focus color. This is a general feeling I have about the choice of color for `--focus-border-color` though, and the consistency of using `--focus-border-color` in the theme is going to make it more readily maintainable in future. If you have a strong leaning then I'm very much open to guidance! I have updated the PR with your suggested changes but happy to roll it back too.

thsparks:
> I'm not sure which is right in practice. From a pure accessibility standpoint I favour the purple as it is more visible, particularly against the New Project card, which otherwise eclipses the light blue of the focus color. This is a general feeling I have about the choice of color for `--focus-border-color` though, and the consistency of using `--focus-border-color` in the theme is going to make it more readily maintainable in future. If you have a strong leaning then I'm very much open to guidance! I have updated the PR with your suggested changes but happy to roll it back too.

Hmm, I see what you mean about the purple New Project card eclipsing the blue focus color. I don't think there's a perfect answer here. I see a few options:

1. Just leave it blue and see if we get any feedback about the lack of clarity
2. Special-case the new project card (already has a special css class, newprojectcard) and give it a different focus-border color (ties into your similar concerns with the Download button)
3. Change `--focus-border-color` to be purple (you can find it [here](https://github.com/microsoft/pxt-microbit/blob/93ab141855d1c08c603b583de8a1da48ad29d54a/theme/color-themes/microbit-light.json#L78)). This would require a change in the pxt-microbit repo, but it's definitely do-able.

I think any of these three would be reasonable, but if I had to pick one, I'd go with 3. There's so much blue in the microbit UI, I think we're likely to find a lot of scenarios where a blue focus border is not ideal. We may still have to special-case some buttons, but hopefully fewer of them...

All that said, any of those options are fine by me. I'll approve this PR today and merge it tomorrow as-is if I don't hear back. In the meantime, if you do want to make any further changes per the above options, feel free to do so (or let me know if you need more time and I can wait on the merge).

Thank you again for all these accessibility changes!

thsparks (approved):

PR #10449, opened 2025-03-20
fix name input in share info

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/6849

pretty self explanatory! not sure why this regressed though


thsparks (approved):
I'm guessing (not certain) that we'll need a pretty much identical change in a different component for https://github.com/microsoft/pxt-arcade/issues/6851 (@srietkerk FYI)

PR #10453, opened 2025-03-21
give preserve on blur a default

srietkerk:
Fixes https://github.com/microsoft/pxt-arcade/issues/6851

A lot of the inputs that are most commonly used don't use our react common input so it's a bit hard to catch, but the `preserveOnBlur` is falsey by default, so any use of the react common input that doesn't have `preserveOnBlur` defined would automatically lose the values on blur.

The other places I have seen this behavior in beta are: in the music editor name, the duration for the sound effect editor, the namer for the color picker, and maybe more.

 `preserveOnBlur` should default to true; a user shouldn't lose the text if they click off of the input.

### DIFF
--- a/react-common/components/controls/Input.tsx
+++ b/react-common/components/controls/Input.tsx
@@ -59,7 +59,7 @@ export const Input = (props: InputProps) => {
         onBlur,
         onOptionSelected,
         handleInputRef,
-        preserveValueOnBlur,
+        preserveValueOnBlur = true,
         options
     } = props;
 


thsparks (approved):
:shipit:

riknoll:
i think this is probably fine, but maybe we shouldn't apply it to all inputs just before releasing arcade?

jwunderl (approved):

abchatra:
Can we just fix the text box in question for this bug? Let's keep the surface area small for last minute bug fixes.

thsparks:
> i think this is probably fine, but maybe we shouldn't apply it to all inputs just before releasing arcade?

FWIW I looked at references and it doesn't seem to be used in that many places, and the one place I saw that was used and did not have this flag set also had the same bug (share link text box). So I think changing base is actually the right call.

srietkerk:
Double checked all of the instances in pxt and arcade that do not have `preserveOnBlur` defined. They are behaving as expected.

abchatra:
Awesome. Thanks @srietkerk

PR #10463, opened 2025-03-25
Clear system clipboard on block copy to prevent pasting files.

microbit-matt-hillsdon:
Fixes https://github.com/microsoft/pxt-microbit/issues/4944

### DIFF
--- a/pxtblocks/copyPaste.ts
+++ b/pxtblocks/copyPaste.ts
@@ -48,6 +48,8 @@ function registerCut() {
             return oldCut.preconditionFn(workspace);
         },
         callback(workspace, e, shortcut) {
+            clearSystemClipboard();
+
             const handler = getCopyPasteHandlers()?.cut;
 
             if (handler) {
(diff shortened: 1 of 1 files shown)


riknoll (changes requested):
I think the proper fix here would be to call preventDefault() on the event passed into the callback if there is a block copied rather than messing with the clipboard

microbit-matt-hillsdon:
> I think the proper fix here would be to call preventDefault() on the event passed into the callback if there is a block copied rather than messing with the clipboard

I worry this would disable file paste forever once you've copied a block (as the Blockly clipboard won't clear). I'm not sure that would be OK.

It's hard to do a perfect job here with a separate clipboard. With the clearing approach it's still not great in the copy a block then copy a file case as both events will still fire. But for typical user flows of copy and then paste it should be better. What do you think?

microbit-matt-hillsdon:
Investigated further and I think there's a simpler fix. The event sequence is:

1. keydown handled via Blockly onKeyDown (potentially pasting block)
2. paste event handled in setupDragAndDrop

Neither code is aware of the other.

However, calling preventDefault on the key event prevents the paste event and we can do this at the point we've decided to handle the paste. So that's a simple and logical fix. I'll close this and open a new PR for that.

Should ideally also happen in Blockly and in the keyboard nav plugin if it continues to replace paste actions. I'll follow up on those.

Closing.

PR #10474, opened 2025-03-31
unifying the messaging for type errors and adding type checks for builtin types

riknoll:
this pr surfaces some new functions in pxsim for reporting type errors like failed casts, accessing properties on null/undefined, etc. these new messages all actually go through localization and are generally more readable than what we have today.

i've also slightly tweaked our binary.js emitter to include the vtable information for failed casts so that we can include the expected type in the error message for that case.

finally, i went ahead and added type checking for all of our default sim implementations of the "native" types like strings, arrays, buffers, etc. i will have another pr in pxt-common-packages shortly that adds even more of these type checks.

### DIFF
--- a/pxtcompiler/emitter/backjs.ts
+++ b/pxtcompiler/emitter/backjs.ts
@@ -493,16 +493,19 @@ function ${id}(s) {
             }
         }
 
+        function vTableRef(info: ClassInfo) {
+            return `${info.id}_VT`
+        }
+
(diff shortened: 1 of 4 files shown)


thsparks (approved):
LGTM, though I was wondering: Is there any scenario where code that "worked" before would fail now because of the extra type checks (i.e. some implicit conversion which is now surfaced as an error)?

eanders-ms (approved):
❤️

PR #10475, opened 2025-03-31
pxt-core changes for bedrock target

eanders-ms:
This PR contains several changes to support the new pxt-bedrock target:
* `compile.unfetteredInitializers` -- Option in pxtarget.json to remove the restriction that enum initializers be of numeric value (string values are supported by the TS compiler, but we were blocking them).
* Support for shimmed property setters (shims were only supported on property getters previously, unless I just missed it which is entirely possible).
* Buffed up the APIs on core runtime types like `RefObject` and `RefCollection` with improved implementations of `toAny`, `fromAny`, and `toDebugString`.
* Minor tweaks to `sui.DropDownMenu` to better support injected menu items provided by the target.
* Added a few new target extension points:
    * `getDownloadMenuItems` -- Menu items to add to the gear menu and the "..." dropdown next to the big Download button.
    * `notifyProjectSaved` -- Notify the target when the project is saved.
    * `notifyProjectCompiled` -- Notify the target when the project is successfully compiled.
    * `onDownloadButtonClick` -- Let the target completely override handling of the download button.
    * `getDefaultProjectName` -- The target can provide a default name for new projects.
* Support for deploying a project to the host via electron.
* Tweaks to the sui.Dialog system to support multi-step dialogs (a.k.a. Wizards) driven from logic in the target.
* My experimental `FieldScopedValueSelector`, a field helper that can auto-assign a value in a drag/dropped block to something appropriate in the destination scope (wip).


webapp/src/compiler.ts
```diff
@@ -545,27 +545,31 @@ export function snippetAsync(qName: string, python?: boolean): Promise<string> {
...
                 refreshLanguageServiceApisInfo();
+                await ensureApisInfoAsync();
             }
-            return ensureApisInfoAsync();
-        })
-        .catch(catchUserErrorAndSetDiags(null))
+            return r;
+        } catch (e) {
+            catchUserErrorAndSetDiags(null);
+            return undefined;
```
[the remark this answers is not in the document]
eanders-ms: This matches the behavior of the original code.

webapp/src/container.tsx
```diff
@@ -295,6 +302,7 @@ export class SettingsMenu extends data.Component<SettingsMenuProps, SettingsMenu
             || this.state.accessibleBlocks != nextState.accessibleBlocks
             || this.state.showShare != nextState.showShare
             || nextProps.inBlocks !== this.props.inBlocks
+            || !!nextState.extDownloadMenuItems?.length
```
riknoll, line 305: should this be checking it against whatever is in the current state instead of always updating?

webapp/src/container.tsx
```diff
@@ -276,6 +277,12 @@ export class SettingsMenu extends data.Component<SettingsMenuProps, SettingsMenu
         this.props.parent.signOutGithub();
     }
 
+    onDropdownShow = () => {
+        // Ensure dropdown always has the editor's most up-to-date menu items
+        const extDownloadMenuItems = pxt.commands.getDownloadMenuItems?.() || [];
+        this.setState({ extDownloadMenuItems });
```
riknoll, line 283: instead of storing these in state, is it easier to just call the commands function again when rendering? caching like this seems like it might run into some bugs if things were to update...

webapp/src/editortoolbar.tsx
```diff
@@ -266,7 +265,7 @@ export class EditorToolbar extends data.Component<ISettingsProps, EditorToolbarS
         const targetTheme = pxt.appTarget.appTheme;
         const { compiling, isSaving } = this.props.parent.state;
         const { compileState } = this.state;
-        const compileTooltip = lf("Download your code to the {0}", targetTheme.boardName);
+        const compileTooltip = lf("Download your code to {0}", targetTheme.boardName);
```
riknoll, line 268: does this read worse for boards like the micro:bit? should we make this some sort of option?

webapp/src/editortoolbar.tsx
```diff
@@ -425,18 +430,12 @@ export class EditorToolbar extends data.Component<ISettingsProps, EditorToolbarS
...
 
         return <div id="editortools" className="ui" role="region" aria-label={lf("Editor toolbar")}>
-            <div id="downloadArea" role="menubar" className="ui column items">{headless &&
-                <div className="ui item">
-                    <div className="ui icon large buttons">
-                        {compileBtn && <EditorToolbarButton icon={downloadIcon} className={`primary large download-button mobile tablet hide ${downloadButtonClasses}`} title={compileTooltip} onButtonClick={this.compile} view='computer' />}
-                    </div>
-                </div>}
-                {/* TODO clean this; make it just getCompileButton, and set the buttons fontsize to 0 / the icon itself back to normal to just hide text */}
-                {!headless && !isTimeMachineEmbed && <div className="ui item portrait hide">
```
riknoll, line 435: why are you removing the check for headless and time machine?

webapp/src/sui.tsx
```diff
@@ -1272,13 +1277,15 @@ export class Modal extends data.Component<ModalProps, ModalState> {
...
             onClose, closeIcon, children, onKeyDown,
-            header, headerIcon, headerClass, headerActions, helpUrl, description,
+            header, headerFn, headerIcon, headerClass, headerActions, helpUrl, description,
             closeOnDimmerClick, closeOnDocumentClick, closeOnEscape,
             shouldCloseOnEsc, shouldCloseOnOverlayClick, shouldFocusAfterRender, overlayClassName, ...rest } = this.props;
         const { marginTop, scrolling, mountClasses } = this.state;
         const isFullscreen = size == 'fullscreen';
         const showBack = isFullscreen && !!closeIcon;
 
+        const resolvedHeader = header ? header : headerFn ? headerFn() : undefined;
```
riknoll, line 1287: this is the reverse of the button logic. for the buttonfn in coretsx, the function overrides the property not the other way around

pxtlib/cmds.ts
```diff
@@ -36,12 +36,18 @@ namespace pxt.commands {
...
+    export let electronFileDeployAsync: (deployRequest: pxt.electron.FileDeployRequest) => Promise<void> = undefined; // A pointer to the Electron file deploy function, so that targets can access it in their extension.ts
     export let webUsbPairDialogAsync: (pairAsync: () => Promise<boolean>, confirmAsync: (options: any) => Promise<WebUSBPairResult>, implicitlyCalled?: boolean) => Promise<WebUSBPairResult> = undefined;
     export let onTutorialCompleted: () => void = undefined;
     export let onPostHostMessage: (msg: any /*pxt.editor.EditorMessageRequest*/) => void;
     export let perfMeasurementThresholdMs: number = undefined;
     export let onPerfMilestone: (payload: { milestone: string, time: number, params?: Map<string> }) => void = undefined;
     export let onPerfMeasurement: (payload: { name: string, start: number, duration: number, params?: Map<string> }) => void = undefined;
     export let workspaceLoadedAsync: () => Promise<void> = undefined;
+    export let getDownloadMenuItems: () => any[] = undefined; // sui.ItemProps[]
```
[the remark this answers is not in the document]
eanders-ms: i will give that a try!

webapp/src/container.tsx
eanders-ms: will revisit. the intention is for it to get latest menu items each time the menu opens, because they are contextual.

webapp/src/editortoolbar.tsx
eanders-ms: Possibly bad merge. Will restore these flags.

webapp/src/sui.tsx
eanders-ms: Nice. will make this one behave the same.

pxtlib/cmds.ts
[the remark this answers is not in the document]
eanders-ms: We would need to reference React in pxtlib, which I don't think we do today.
image

webapp/src/editortoolbar.tsx
eanders-ms: Are strings in pxtarget.json localizable? If so, I could override the entire tooltip.

eanders-ms: reverting for now.

webapp/src/container.tsx
eanders-ms: Now grabs latest menu items when rendering, but for the items to refresh, I had to make it call forceUpdate when the dropdown opens.

webapp/src/editortoolbar.tsx
eanders-ms: Restored `isTimeMachineEmbed`. I couldn't simply re-add `headless` because bedrock is headless yet does want the download button to display. I think I was able to adjust the logic so it works for all targets. Tested in Arcade, Microbit, and Minecraft. All behaved like they do in production.

pxtlib/cmds.ts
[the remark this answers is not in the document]
eanders-ms: Oh, this appears in pxtlib. I don't think this is going to be possible without a lot of gymnastics.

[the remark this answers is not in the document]
riknoll: yeah no worries; that's not worth it

webapp/src/editortoolbar.tsx
riknoll: @eanders-ms did you have all the URL flags for minecraft though? i think the download button is hidden in minecraft...

webapp/src/container.tsx
eanders-ms: fixed

webapp/src/editortoolbar.tsx
eanders-ms: Here's what it looks like in EDU: 
image

riknoll (approved):

PR #10476, opened 2025-03-31
Fix the dialog for sharing vscode

abchatra:
Fixes https://github.com/microsoft/pxt-arcade/issues/6895

### DIFF
--- a/react-common/components/share/ShareInfo.tsx
+++ b/react-common/components/share/ShareInfo.tsx
@@ -409,7 +409,7 @@ export const ShareInfo = (props: ShareInfoProps) => {
                     {shareState === "publish-vscode" &&
                         <div className="project-share-vscode">
                             <div>
-                                {lf("🎉 All set! Your project is launching in VSCode. If it doesn't open soon, just click the button below!")}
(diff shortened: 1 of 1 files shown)


react-common/components/share/ShareInfo.tsx
```diff
@@ -409,7 +409,7 @@ export const ShareInfo = (props: ShareInfoProps) => {
                     {shareState === "publish-vscode" &&
                         <div className="project-share-vscode">
                             <div>
-                                {lf("Success! Your browser should automatically redirect youClick the button below to open your project in VS Code.")}
+                                {lf("🎉 All set! Your project is launching in VSCode. If it doesn't open soon, just click the button below!")}
```
thsparks, line 412:
```suggestion
                                {lf("🎉 All set! Your project is launching in VS Code. If it doesn't open soon, just click the button below!")}
```

thsparks (approved):
One suggested change to be consistent with VS Code docs.

react-common/components/share/ShareInfo.tsx
thsparks: VS Code docs use space, see https://code.visualstudio.com/:

image

thsparks (approved):

PR #10477, opened 2025-04-01
don't parse junk tilemaps literally

riknoll:
this fixes the issue encountered by @UnsignedArduino [here](https://github.com/microsoft/pxt-arcade/issues/6837#issuecomment-2770210833) but doesn't fix the underlying issue that got the project into that state in the first place.


eanders-ms (approved):

abchatra:
Port to stable.

PR #10479, opened 2025-04-02
add some missing localization strings

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/6891

looks like we were never getting the strings from react-common which is wild if true?

also, @thsparks, i edited your extra function types code slightly so that it now also includes the types without labels


jwunderl (approved):

thsparks (approved):
LGTM! (Not sure which part of the code you referred to as mine, but change seems fine regardless)

PR #10481, opened 2025-04-02
Teacher Tool Theme Fixes

thsparks:
This fixes some issues with the teacher tool UI after themes were introduced. While doing this, I decided to fully swap out the themepack with our new theme variables, just to keep things consistent across the apps. It should also make things "just work" if we introduce other themes for teacher tool (though I haven't done that in this change).

Since I didn't want to introduce new theme colors, I have had to make some minor adjustments to how things look (like the beta tag & toasts), but I tried to stay relatively true to the original designs.

Upload Target: https://makecode.microbit.org/app/2c6d961517b971a19375a8090903860676c8e0f7-fa4ad57a6d--eval

### DIFF
--- a/react-common/styles/profile/UserAvatarDropdown.less
+++ b/react-common/styles/profile/UserAvatarDropdown.less
@@ -21,5 +21,6 @@
     width: 2.5rem;
     height: 2.5rem;
 
-    background-color: var(--pxt-headerbar-accent);
+    background-color: var(--pxt-header-stencil);
+    color: var(--pxt-header-foreground);
 }
--- a/teachertool/src/components/CriteriaResultEntry.tsx
+++ b/teachertool/src/components/CriteriaResultEntry.tsx
@@ -105,6 +105,7 @@ const CriteriaResultToolbarTray: React.FC<{ criteriaId: string }> = ({ criteriaI
         }
     }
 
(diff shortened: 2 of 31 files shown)


eanders-ms:
I think we should introduce bespoke color vars for toast now, and plan to move the Toast component into react-common "soon" (perhaps as part of the current microbit work).

eanders-ms:
(minor) I think the sign in button lost its right margin: 
image

eanders-ms:
> I think we should introduce bespoke color vars for toast now, and plan to move the Toast component into react-common "soon" (perhaps as part of the current microbit work).

but it's something we could certainly do later!

eanders-ms (dismissed):

thsparks:
> I think we should introduce bespoke color vars for toast now, and plan to move the Toast component into react-common "soon" (perhaps as part of the current microbit work).

Had a quick chat offline: concern here was around differentiation between color vars (red/yellow/green) and status (error/warning/success). We should probably have different theme colors for both, even if they sometimes duplicate each other. For example, micro:bit green is quite dark. It'd be nice for the success color to be a bit brighter. But I will do that in a separate change.

eanders-ms (approved):

PR #10483, opened 2025-04-03
Allow keyboard to close the first Download dialog

microbit-matt-hillsdon:
At the moment the keyboard cannot close the first "Download" flow dialog with the connect animation.

We find folks using keyboard navigation can trigger "Download" by mistake and then find it frustrating to be forced to continue through to the next dialog.

We debated whether it still made sense to force the download when bailing out at this early stage and have no strong view (this PR does).

This seemed the easiest way to suggest the change but totally open to explanation as to why the flow is structured this way. But I think something is necessary to cater for keyboard users.


thsparks:
We removed the close button in https://github.com/microsoft/pxt/pull/9486 because we found users had a gut-reaction to immediately just close the popup, then they'd get confused as to why it didn't work.

How would you feel about adding a "cancel" button at the bottom, rather than the close button at the top? The thinking being, that may not trigger the same immediate "close" reaction we saw with some users. We have this in one or two other places, like our "Reset" dialog (though in this case, I'd put cancel on the left, not the right):
image

microbit-matt-hillsdon:
> How would you feel about adding a "cancel" button at the bottom, rather than the close button at the top? The thinking being, that may not trigger the same immediate "close" reaction we saw with some users. We have this in one or two other places, like our "Reset" dialog (though in this case, I'd put cancel on the left, not the right):

Yes, I think that's a good option.

microbit-matt-hillsdon:
Closing in favour of #10632

PR #10487, opened 2025-04-04
Add theme colors for success, warning, and error

thsparks:
I was originally thinking we could just use the colors-green, colors-yellow, and colors-red variables for this, but as was noted in https://github.com/microsoft/pxt/pull/10481, those may be a bit too generic to work well for these scenarios. In the teacher tool, for example, the microbit green color is a little too dark to work well as a "success" color. Separating them will lead to some duplication in certain themes, but it's nice to have the flexibility.

As is the case for all theme colors, any themes that do not specify a value in their json file will default to the value in base-theme.less.

For now, I've only used these new colors in the teacher tool. We can swap them out in other areas as needed.

With the new colors:
image

image

image

Fixes https://github.com/microsoft/pxt-microbit/issues/6137

### DIFF
--- a/react-common/styles/theming/base-theme.less
+++ b/react-common/styles/theming/base-theme.less
@@ -71,6 +71,21 @@
     --pxt-link-hover: #204467;
     --pxt-focus-border: #0078D4;
 
+    --pxt-success: #2ecc71;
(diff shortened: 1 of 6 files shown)


srietkerk (approved):
It looks really nice!!

PR #10492, opened 2025-04-04
fix argument order of setFieldValue

riknoll:
this was an issue reported on the forum [here](https://forum.makecode.com/t/code-deleting-randomly/35057/12?u=richard)

looks like i mixed up the arguments of setFieldValue. I checked out every other place we call setFieldValue and looks like this was the only one with this issue.


thsparks (approved):

abchatra:
Port to stable.

PR #10507, opened 2025-04-08
fix temporary asset ids when converting to named assets

riknoll:
fixing a bug i found while investigating asset errors. when converting temporary assets to named assets, the temporary asset id was not being updated correctly everywhere except in the blockly asset field editors. all other ways to edit assets (monaco field editors and the asset editor tab) were not updating the junk id that we use for temporary assets into a valid qualified name.


eanders-ms (approved):
LGTM

jwunderl (approved):

PR #10510, opened 2025-04-08
Add ability to create a PR when bumping package

eanders-ms:
This PR updates the `pxt bump` command to support bumping a protected branch by creating a PR rather than committing directly. It works in conjunction with the updated GitHub Action workflows to tag the commit upon merge, triggering a publish to npm.

When run, `pxt bump` queries the protection status of the current branch to determine if it should create a PR or commit directly. You can force the PR behavior with the `--pr` flag, e.g.: `pxt bump --pr`.

Additionally, I added a parameter that allows the user to specify which component of the version number to bump, or they can supply a complete version number to assign: `pxt bump --version [patch|minor|major|<version>]`. Default: patch.

#### bump master branch

1. User runs `pxt bump`.
2. Master is a protected branch, so a new working branch is created.
3. Package version is bumped locally.
4. This change is committed with the message "[pxt-cli] bumped version to vX.X.X".
5. The commit is not tagged at this point. That will be done later (see step 8).
6. Changes are pushed to remote.
7. A PR is opened with the title "[pxt-cli] bumped version to vX.X.X".
8. When the PR is merged, the workflow `pxt-buildpush` detects the bump commit, parses the version number from the commit message, and tags the commit sha.
9. The workflow branches to invoke `npm test` allowing publish to npm.

#### bump non-protected branch

1. User runs `pxt bump`.
2. Branch is not protected, so it works as it did before, bumping the version, tagging the commit, and pushing directly to remote.
3. On remote, the workflow `pxt-buildvtag` is triggered, which runs `npm test` allowing publish to npm.

### Workflow file changes

* **Deleted: pxt-buildpr.yml**
* **Renamed: pxt-buildmain.yml --> pxt-buildvtag.yml**
    * Only runs on new semver tags pushed from the client
    * Does not run on semver tags created by the pxt-buildpush workflow! Work performed in one GitHub action will not trigger another action in the same repo. (It can be done but requires gymnastics.)
* **Updated: pxt-buildpush.yml**
   * Changed trigger to run on push to any branch or tag
   * Will tag bump commits, only in a PR merge where the bump commit is the only included change.
   * For bump commits: Runs `npm test` allowing publish to npm
   * For other changes: Runs `npm test` disallowing publish to npm


eanders-ms:
Please review these changes carefully. If there are any build scenarios you'd like to see tested, let me know!

cli/cli.ts
```diff
@@ -31,6 +31,15 @@ import * as crowdin from './crowdin';
...
 import { SUB_WEBAPPS } from './subwebapp';
 
+const Reset = "\x1b[0m";
+const Bright = "\x1b[1m";
+const FgRed = "\x1b[31m";
+const FgGreen = "\x1b[32m";
+const FgYellow = "\x1b[33m";
+const FgBlue = "\x1b[34m";
+const FgMagenta = "\x1b[35m";
+const FgCyan = "\x1b[36m";
```
riknoll, line 41: any reason not to just use chalk? that's what we use in the pxt-mkc

cli/cli.ts
```diff
@@ -644,26 +638,93 @@ function tagReleaseAsync(parsed: commandParser.ParsedCommand) {
...
+
     if (fs.existsSync(pxt.CONFIG_NAME)) {
         if (upload) throw U.userError("upload only supported on targets");
 
-        return Promise.resolve()
-            .then(() => nodeutil.runGitAsync("pull"))
-            .then(() => justBumpPkgAsync())
-            .then(() => nodeutil.runGitAsync("push", "--tags"))
-            .then(() => nodeutil.runGitAsync("push"))
+        if (pr || branchProtected) {
```
riknoll, line 671: async/await?

cli/nodeutil.ts
```diff
@@ -174,6 +174,224 @@ export function needsGitCleanAsync() {
         })
 }
 
+export async function getDefaultBranchAsync(): Promise<string> {
+    const b = await gitInfoAsync(["symbolic-ref", "--short", "refs/remotes/origin/HEAD"], undefined, true);
+    if (!b)
+        throw Util.userError("no git remote branch found");
```
riknoll, line 180: iirc userError already throws

cli/nodeutil.ts
```diff
@@ -174,6 +174,224 @@ export function needsGitCleanAsync() {
...
+
+export async function getCurrentBranchNameAsync(): Promise<string> {
+    const b = await gitInfoAsync(["rev-parse", "--abbrev-ref", "HEAD"], undefined, true);
+    if (!b)
+        throw Util.userError("no git local branch found");
+    return b;
+}
+
+export async function getGitHubTokenAsync(): Promise<string> {
+    const input = Buffer.from("protocol=https\nhost=github.com\n\n", "utf8");
```
riknoll, line 192: why the buffer here?

cli/nodeutil.ts
```diff
@@ -174,6 +174,224 @@ export function needsGitCleanAsync() {
...
+}
+
+export async function getCurrentBranchNameAsync(): Promise<string> {
+    const b = await gitInfoAsync(["rev-parse", "--abbrev-ref", "HEAD"], undefined, true);
+    if (!b)
+        throw Util.userError("no git local branch found");
+    return b;
+}
+
+export async function getGitHubTokenAsync(): Promise<string> {
```
riknoll, line 191: is this tested on wsl and windows?

cli/nodeutil.ts
eanders-ms: It throws, but typescript typechecking doesn't know that and gets confused when the enclosing function should return a value. I will try updating userError to return `never`.

cli/nodeutil.ts
eanders-ms: copilot being dumb. will fix.

cli/nodeutil.ts
eanders-ms: Tested in bash and powershell. Each of us will need to test in our respective dev environments. I'm unable to test the complete matrix.

cli/cli.ts
```diff
@@ -7016,7 +7078,14 @@ ${pxt.crowdin.KEY_VARIABLE} - crowdin key
...
         flags: {
             update: { description: "(package only) Updates pxt-core reference to the latest release" },
-            upload: { description: "(package only) Upload after bumping" }
+            upload: { description: "(package only) Upload after bumping" },
+            pr: { description: "Create a pull request after bumping rather than push changes directly" },
+            version: {
+                description: "Which part of the version to bump, or a complete version number to assign. Defaults to 'patch'",
+                argument: "version",
+                type: "string",
+                possibleValues: ["patch", "minor", "major", /^\d+\.\d+\.\d+$/]
```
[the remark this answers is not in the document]
eanders-ms: These arguments are modeled after the `npm version` command, and are a strict subset. We invoke `npm patch` under the hood, so this is a clean implementation in the sense that we can just pass the arguments along to it.

cli/cli.ts
eanders-ms: I typically don't proactively convert to async/await and just do it when it simplifies the changes I'm making. The need didn't arise here. Happy to do it if you'd prefer that!

cli/cli.ts
eanders-ms: Makes sense. Will use chalk.

cli/cli.ts
eanders-ms: I also wanted to leave it in the promise format to clearly show that the non-pr codepath works the same as before.

cli/cli.ts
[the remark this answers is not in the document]
eanders-ms: Made v optional.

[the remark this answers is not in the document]
abchatra: Our electron builds and backend need v otherwise they don't trigger. This is going to be break things AFAIK

cli/cli.ts
eanders-ms: Downgraded chalk to v4 and still had to import using `require`. I don't see the utility of ESM modules.

.github/workflows/is-vtag.yml
```diff
@@ -0,0 +1,30 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      is_vtag: ${{ steps.set-output.outputs.is_vtag }}
+    steps:
+      - name: Check tag pattern
+        id: check-tag
+        run: |
+          if [[ "${GITHUB_REF_TYPE}" == "tag" && "${GITHUB_REF_NAME}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "Semver tag detected: ${GITHUB_REF_NAME}"
+            echo "true" > is_vtag.txt
```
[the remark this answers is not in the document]
eanders-ms: i asked copilot about this, and it seemed to think this was more correct in this particular instance. it worked, so i moved on to other things. will revisit.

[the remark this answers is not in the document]
thsparks: It's not really a big deal either way. This felt like adding opportunities for errors to creep in with weird disk read/write issues, but I don't feel strongly. If it works, it works!

cli/cli.ts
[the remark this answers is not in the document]
eanders-ms: The tag will always have a v.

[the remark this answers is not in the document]
eanders-ms: This argument is the more closely aligned to the version number as appears in package.json. A v present in package.json's version value would break things in other ways. The v is always included in the tag name, once it is eventually created.

.github/workflows/is-vtag.yml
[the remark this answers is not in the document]
eanders-ms: Totally agree. Fixed.

thsparks (approved):

abchatra (approved):

Issue #10464, opened 2025-03-25
Switch to latest Blockly keyboard navigation experiment


microbit-matt-hillsdon:
(The micro:bit team are expecting to do most of this work, this issue is for visibility and input from the MakeCode/Blockly teams)

The [new keyboard navigation plugin](https://github.com/google/blockly-keyboard-experimentation) is still a work-in-progress, but I wanted to raise this issue to start some tech-level discussion around switching over.

The most recent MakeCode demo as used in user testing is here: https://march-ut.review-pxt.pages.dev/

This version has the experiment hard-coded on for ease in user testing. We also merged early versions of various other in-flight PRs such as #10462.

For an initial plugin upgrade PR, I'm hoping to keep it as an experiment and then discuss defaulting it on at a later point.

## Blockers

Moved discussion of blockers to the PR.

PR #10512, opened 2025-04-10
Upgrade to latest Blockly keyboard navigation plugin

microbit-robert:
We are opening this PR in draft mainly for visibility at this point. This PR has now been rebased on master after the [upgrade to Blockly v12.0.0-beta.4](https://github.com/microsoft/pxt/pull/10521), on which the new keyboard navigation plugin depends.

[Link to demo](https://blockly-12-keyboard-exp.review-pxt.pages.dev/). Please note that 'accessible blocks' need to be set to 'on' in the settings menu.

A summary of changes and blockers to the plugin upgrade are listed below:

- Use [blockly/keyboard-experiment v0.0.7](https://www.npmjs.com/package/@blockly/keyboard-experiment/v/0.0.7)
- Remove toolbox tab stops and use arrow keys for toolbox nav
- Improve area for toolbox
- Add skip link for the workspace when keyboard nav is enabled
- Remove second step to enable keyboard nav after enabling the experiment, the objective is always enabled keyboard nav.

### Resolved blockers
- [x] MakeCode cut / copy / paste modifications are disabled to prevent conflict, next step needs discussion. The current implementation in MakeCode reuses Blockly code and hooks in localStorage to copy and paste between tabs - it is not clear if the same approach can be taken with the new keyboard plugin and needs investigation
    - Currently you need access to plugin internals to follow this route so exploring other options.
    - As a stopgap to enable moving forward with this PR, we could disable MakeCode's copy/paste overrides when the experiment was enabled.
- [x] We've sidestepped the caching MakeCode's flyout does (directly using CachingFlyout), to avoid breaking Blockly assumptions that the flyout / flyout workspace doesn't change over time. Caching is valuable though, need to explore reinstating perhaps at a different level.
    - Not sure that the same stopgap approach is so reasonable here. This is our next task but suggestions very welcome. Rachel suggested looking at the [continuous toolbox plugin](https://github.com/google/blockly-samples/blob/rc/v12.0.0/plugins/continuous-toolbox/src/RecyclableBlockFlyoutInflater.ts), which does some degree of caching/recycling (not at the whole-flyout level).
    - No longer a blocker thanks to https://github.com/microsoft/pxt/pull/10533

### Follow-up issues:
- [Help docs](https://github.com/microsoft/pxt-microbit/issues/6152)
- [Global shortcuts](https://github.com/microsoft/pxt-microbit/issues/6151)
- Black focus outline around the toolbox, flyout and workspace, respectively, when using keyboard navigation. The workspace currently shows this outline on click which is not desired behaviour. 
    - We think it's not so bad that this happens for a mouse click if you've enabled the keyboard nav experiment but it's not reasonable if we moved to enabling it by default. Propose tracking separately.
- [Avoid affecting theme](https://github.com/google/blockly-keyboard-experimentation/issues/147)
    - We're going to try to fix this in Blockly **but** I think they switched from yellow to orange for visibility reasons so it's worth considering making this change in MakeCode. We'll try to make the theme the source of truth for color for the selection color and the related cursor positions the plugin draws.
    - We discussed this on the last MakeCode call and agreed that the colour change was OK for users who opt-in to block keyboard navigation. Changing it globally is a discussion for the future.
- It's common to hit the issue addressed by this PR so it would be good to agree a way forward for that: https://github.com/microsoft/pxt/pull/10463

@riknoll @microbit-matt-hillsdon 

Closes https://github.com/microsoft/pxt/issues/10464


theme/blockly-core.less
```diff
@@ -373,3 +393,45 @@ text.blocklyCheckbox {
         font-size: 17pt !important;
     }
 }
+
+
+/*******************************
```
microbit-matt-hillsdon, line 398: We should review this CSS for unexpected impact when the experiment is not enabled.

webapp/src/blocks.tsx
```diff
@@ -532,31 +548,34 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
+            const injectionDiv = document.getElementById('blocksEditor');
+            injectionDiv.classList.add("accessibleBlocks");
+            const focusRingDiv = injectionDiv.appendChild(document.createElement("div"))
+            focusRingDiv.className = "blocklyWorkspaceFocusRingLayer";
+            this.editor.getSvgGroup().addEventListener("focus", () => {
+                focusRingDiv.dataset.focused = "true";
+            })
+            this.editor.getSvgGroup().addEventListener("blur", () => {
+                delete focusRingDiv.dataset.focused;
+            })
```
microbit-matt-hillsdon, line 565: This gives a more reasonable focus ring for the workspace, one that isn't occluded on the left hand side by the toolbox. We should discuss this with Blockly and potentially move the behaviour to the plugin (maybe with options for a better implementation).

microbit-matt-hillsdon:
Agreed to restructure this around a setting rather than an experiment. Setting can reload the page.

riknoll:
re: copy/paste, i assume this means the [blockly plugin](https://github.com/google/blockly-samples/tree/master/plugins/cross-tab-copy-paste) is also broken? seems like this should be something tackled by the blockly team. for now, your suggestion of disabling it when the experiment is active seems like an acceptable temporary solution

i'll look into a plan for the flyout

webapp/src/blocks.tsx
microbit-robert: We are continuing to have discussions regarding the focus indicators, but we propose moving forward with this change for this initial PR.

webapp/src/blocks.tsx
```diff
@@ -882,6 +915,9 @@ export class Editor extends toolboxeditor.ToolboxEditor {
             <div className="blocksAndErrorList">
                 <div className="blocksEditorOuter">
                     <div id="blocksEditor"></div>
+                    {/* This does not integrate well with MakeCode, but is useful for blockly
+                     keyboard navigation testing / evaluation purposes. */}
+                    <div id="shortcuts"></div>
```
microbit-robert, line 920: Remove this div as we don't intend to use its contents as populate by the keyboard plugin

webapp/src/draganddrop.ts
```diff
@@ -4,6 +4,10 @@ export function setupDragAndDrop(r: HTMLElement,
     draggedUri: (uri: string) => void
 ) {
     r.addEventListener('paste', function (e: ClipboardEvent) {
+        // FIXME: Sidestep conflicts with keyboard nav paste in the editor.
+        if (document.location.hash !== "") {
+            return;
+        }
```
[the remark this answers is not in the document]
microbit-robert: This has been removed from this PR [d0d6bb0](https://github.com/microsoft/pxt/pull/10512/commits/d0d6bb0c2dd1b8d8aaf73ce794f9e2d8ca54d03d).

webapp/src/blocks.tsx
microbit-robert: Removed here [d0d6bb0](https://github.com/microsoft/pxt/pull/10512/commits/d0d6bb0c2dd1b8d8aaf73ce794f9e2d8ca54d03d)

microbit-robert:
This PR is now in a state where it can be reviewed. Blockers have been addressed and the related follow-up issues will be tackled in follow-up PRs.

microbit-robert:
Just highlighting that the keyboard navigation plugin is at a somewhat arbitrary point in its development. Blocks inserted from the flyout are currently inserted in 'move mode', but this is not clear in this version. A move icon has been added since, but is not in this released version. You will need to hit "Enter" to confirm the block position before doing anything else.

I've added some information about the keyboard controls below, but expect this PR to be more about the integration, than the plugin behaviour itself as this will change with updates going forward.

The basic keyboard controls:
Screenshot 2025-05-01 at 17 37 30

Move mode controls:
Screenshot 2025-05-01 at 17 43 20

pxtblocks/plugins/duplicateOnDrag/dragStrategy.ts
```diff
@@ -5,6 +5,7 @@
  */
 
 import * as Blockly from "blockly";
+import { ConnectionPreviewer } from "../renderer";
```
riknoll, line 8: left behind from some refactor?

theme/blockly-core.less
```diff
@@ -373,3 +391,45 @@ text.blocklyCheckbox {
...
+    position: absolute;
+    top: 0;
+    left: 0;
+    bottom: 0;
+    right: 0;
+    pointer-events: none;
+    z-index: 99;
+}
+
+.accessibleBlocks .blocklyWorkspaceFocusRingLayer[data-focused = 'true'] {
```
riknoll, line 418: nitpick:
```suggestion
.accessibleBlocks .blocklyWorkspaceFocusRingLayer[data-focused="true"] {
```

webapp/src/blocks.tsx
```diff
@@ -482,7 +490,10 @@ export class Editor extends toolboxeditor.ToolboxEditor {
             pxtblockly.external.setPromptTranslateBlock(dialogs.promptTranslateBlock);
         }
 
-        pxtblockly.external.setCopyPaste(copy, cut, this.pasteCallback, this.copyPrecondition, this.pastePrecondition);
+        // Disable clipboard overwrite (initCopyPaste) when accessible blocks enabled.
+        if (!data.getData<boolean>(auth.ACCESSIBLE_BLOCKS)) {
```
riknoll, line 494: you should register a data subscriber to listen to updates to the `auth.ACCESSIBLE_BLOCKS` setting. check out the constructor of the `Editor` in `monaco.tsx` for an example

webapp/src/blocks.tsx
```diff
@@ -535,31 +557,35 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
-    //     } else {
-    //         this.navigationController.disable(this.editor);
-    //     }
-    // }
+    private initAccessibleBlocks() {
+        const enabled = data.getData<boolean>(auth.ACCESSIBLE_BLOCKS)
+        if (enabled && !this.keyboardNavigation) {
+            this.keyboardNavigation = new KeyboardNavigation(this.editor);
+
+            const injectionDiv = document.getElementById('blocksEditor');
```
riknoll, line 565:
```suggestion
            const injectionDiv = document.getElementById("blocksEditor");
```

webapp/src/blocks.tsx
```diff
@@ -927,17 +965,60 @@ export class Editor extends toolboxeditor.ToolboxEditor {
     }
 
     public moveFocusToFlyout() {
-        // if (this.navigationController) {
-        //     this.navigationController.navigation.focusFlyout(this.editor);
-        // }
+        if (this.keyboardNavigation) {
+            const flyout = this.editor.getFlyout();
+            const element = (flyout as any).svgGroup_ as SVGElement;
```
riknoll, line 970: instead of casting to any here, add a method to the Flyout in `pxtblocks/plugins/flyout/cachingFlyout.ts` to return this element and cast to that instead.

webapp/src/blocks.tsx
```diff
@@ -927,17 +965,60 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
+    // This modification is required to workaround the fact that cached blocks are not disposed in MakeCode.
+    private isFlyoutItemDisposed(node: Blockly.ASTNode) {
+        const sourceBlock = node.getSourceBlock();
+        if (sourceBlock?.disposed || sourceBlock?.hasDisabledReason(HIDDEN_CLASS_NAME)) {
+          return true;
+        }
+        const location = node.getLocation();
+        if (location instanceof Blockly.FlyoutButton) {
+          // No nice way to tell for a button. In Blockly v12 we could use getSvgGroup().
+          return (location as any).svgGroup.parentNode === null;
```
riknoll, line 987: you can probably add something for this in `pxtblocks/plugins/flyout/flyoutButton.ts`

webapp/src/blocks.tsx
```diff
@@ -927,17 +965,60 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
+            element?.focus();
+            this.defaultFlyoutCursorIfNeeded(flyout);
+        }
+    }
 
-        (this.editor.getInjectionDiv() as HTMLDivElement).focus();
+    // Modified from blockly-keyboard-experimentation plugin
+    // https://github.com/google/blockly-keyboard-experimentation/blob/main/src/navigation.ts
+    // This modification is required to workaround the fact that cached blocks are not disposed in MakeCode.
+    private isFlyoutItemDisposed(node: Blockly.ASTNode) {
```
riknoll, line 979: tabs/spaces seemed to be messed up in this function

webapp/src/toolbox.tsx
```diff
@@ -453,11 +448,81 @@ export class Toolbox extends data.Component<ToolboxProps, ToolboxState> {
...
+            // Escape tab and shift key
+        } else {
+            this.setSearch();
+        }
+    }
+
+    previousItem() {
+        const editorname = this.props.editorname;
+
+        pxt.tickEvent(`${editorname}.toolbox.keyboard.prev"`, undefined, { interactiveConsent: true });
```
riknoll, line 515:
```suggestion
        pxt.tickEvent(`${editorname}.toolbox.keyboard.prev`, undefined, { interactiveConsent: true });
```

webapp/src/toolbox.tsx
```diff
@@ -453,11 +448,81 @@ export class Toolbox extends data.Component<ToolboxProps, ToolboxState> {
...
+        this.setPreviousItem();
+    }
+
+    nextItem() {
+        const editorname = this.props.editorname;
+
+        pxt.tickEvent(`${editorname}.toolbox.keyboard.next"`, undefined, { interactiveConsent: true });
```
riknoll, line 522:
```suggestion
        pxt.tickEvent(`${editorname}.toolbox.keyboard.next`, undefined, { interactiveConsent: true });
```

riknoll (changes requested):
looks pretty good! just some minor comments

riknoll:
@microbit-matt-hillsdon here is an uploaded build with the changes from this PR:

https://makecode.microbit.org/app/5b64f555bd29c0c75350023549b02a28cb03c1b0-74c00c13f3

pxtblocks/plugins/duplicateOnDrag/dragStrategy.ts
microbit-robert: Left behind from a more complex way of doing something I think. Removed here [ae4f56b](https://github.com/microsoft/pxt/pull/10512/commits/ae4f56b41fce1373e0a454953143a3cc70dfe607)

webapp/src/toolbox.tsx
microbit-robert: Whoops [3312769](https://github.com/microsoft/pxt/pull/10512/commits/3312769d15e21c5d52ea46a46ff543ad44bfbafd)

webapp/src/toolbox.tsx
microbit-robert: Whoops again [3312769](https://github.com/microsoft/pxt/pull/10512/commits/3312769d15e21c5d52ea46a46ff543ad44bfbafd)

webapp/src/blocks.tsx
microbit-robert: Changed here [2d62a69](https://github.com/microsoft/pxt/pull/10512/commits/2d62a698b0fb3b989b3ce0af87243ca0cd670a7e)

webapp/src/blocks.tsx
microbit-robert: That makes sense. Done here [4b72884](https://github.com/microsoft/pxt/pull/10512/commits/4b72884fd0be029a4b0e09385faba345a616ddc4)

webapp/src/blocks.tsx
microbit-robert: Since we're using Blockly 12, there is a getSvgGroup method which I've used in a new `isDiposed()` method added to the CachingFlyout.

I caught a bug here where the flyout was re-rendering unnecessarily when the toolbox regained focus from the flyout. I've added a fix for that, which means the flyout cursor now remembers positions on buttons (and labels) in the flyout, not just on blocks.

Changes [a52729d](https://github.com/microsoft/pxt/pull/10512/commits/a52729d176ce96cb414281f6708cffc691d309c5)

webapp/src/blocks.tsx
microbit-robert: At the moment, we reload MakeCode when the keyboard navigation plugin is toggled on/off so I'm not sure this is needed, unless there is a way this setting can change other than the user interacting with the settings menu?

eanders-ms: Sign in and sign out both involve a page reload, and we only fetch user preferences from cloud on page load, so I think it's ok.

riknoll (approved):

PR #10516, opened 2025-04-11
Remove the loading state set when the iframe load event fires.

thsparks:
Fixes https://github.com/microsoft/pxt-arcade/issues/6893 (Hopefully. I have a repro, but it's not exact...)

With the "Read&Write for Google Chrome" extension enabled, the load event on the skillmap iframe gets delayed, and we were actually setting state to `loading` _after_ the `newproject` event fired, so when the user tried to start an activity, the app still thought it was loading and failed.

I looked into why we were setting the loading state here (according to [docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event), the load event fires when a page is _finished_ loading, and it felt a little strange to set our state to loading when the iframe finishes loading). Turns out, it was to fix a bug several years ago (https://github.com/microsoft/pxt/pull/8398). It was a niche live -> beta scenario, and I'm unable to repro it now, so I've just removed the code. The only way (that I know of) to trigger the load event on the iframe is to refresh the page, which also triggers recreation of the `makecodeFrame`, so the state was already set to `loading` in the constructor.

### DIFF
--- a/skillmap/src/components/makecodeFrame.tsx
+++ b/skillmap/src/components/makecodeFrame.tsx
@@ -154,7 +154,6 @@ class MakeCodeFrameImpl extends React.Component<MakeCodeFrameProps, MakeCodeFram
     protected handleFrameRef = (ref: HTMLIFrameElement) => {
         if (ref && ref.contentWindow) {
             window.addEventListener("message", this.onMessageReceived);
-            ref.addEventListener("load", this.handleFrameReload)
             this.ref = ref;
             // Hide Usabilla widget + footer when inside iframe view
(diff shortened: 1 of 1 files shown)


riknoll (approved):
i think this is probably fine, but if we end regressing something then we might consider just being smarter about clearing the loading state when we receive other messages from the iframe that would indicate that the load has happened. or maybe we could implement some sort of ping API to check periodically.

really we should move the skillmap to use the message channel implementation rather than post message at some point since i would expect that's a little more robust for this sort of thing.

PR #10521, opened 2025-04-16
Upgrade blockly to 12.0.0-beta.4

riknoll:
this pr upgrades us to the latest beta version of blockly 12.0.0. It also temporarily comments out the references to the keyboard navigation plugin, which is not yet compatible with the latest blockly version. 

here is a brief overview of changes but i'm probably forgetting some:

* fixed type signature of the flyout functions for the functions and variable flyouts
* added a temporary script to patch the field colour which has a typing incompatibility with the latest blockly. this will presumably be fixed by the time blockly 12.0.0 leaves beta
* replaced all the calls to deprecated variable functions on `Blockly.WorkspaceSVG` with `Blockly.WorkspaceSVG.getVariableMap()`
* replaced all the references to `VariableModel.name` with `VariableModel.getName()`
* removed the css classes we had that were referring to the old closure ui names (e.g. the ones that start with `goog-`) and replaced them with the new blockly ones
* converted all of our custom flyout rendering into implementations of the new flyout inflater extension points in blockly
* fixed the height of all of the fields that use the dropdown div. the biggest change here was the field image dropdown, which now calculates its dimensions differently
* replaced all of our references to `.blocklyToolboxDiv` with `.blocklyToolbox`

@microbit-robert @microbit-matt-hillsdon @microbit-grace FYI


pxtblocks/fields/field_gridpicker.ts
```diff
@@ -648,4 +649,130 @@ export class FieldGridPicker extends Blockly.FieldDropdown implements FieldCusto
...
+.blocklyGridPickerSelectedImage {
+    padding: 3px;
+    display: inline-block;
+    vertical-align: middle;
+}
+
+.ui.input input.blocklyGridPickerSearchBar {
+    background: none;
+    border: none;
+    color: white;
```
thsparks, line 756: Should we be using theme colors here? (though I'd have to check if the css vars are in scope with the blockly css stuff...)

pxtblocks/plugins/comments/blockComment.ts
```diff
@@ -225,6 +225,18 @@ export class CommentIcon extends Blockly.icons.Icon {
         this.setBubbleVisible(this.bubbleVisiblity);
     }
 
+    // TODO: switch our custom comment position serialization
```
thsparks, line 228: Was it intentional to leave these as to do's for the time being? Does this mean comments will be broken?

pxtblocks/fields/field_gridpicker.ts
thsparks: Oh, just saw that this is copied from fieldeditor.less. I think it's fine to leave it as-is for now, then.

thsparks (approved):
Nice!

jwunderl (approved):

pxtblocks/plugins/comments/blockComment.ts
riknoll: yes it's intentional and no they aren't broken. i already implemented a separate means of serializing the comment position in the data element for blocks and luckily that still works just fine; this todo is really just meant as a sort of code cleanliness thing since we might as well use the supported route.

main reason i didn't do it in this pr is because it will mean having to upgrade old projects that use my version of the serialization

PR #10528, opened 2025-04-23
Error Panel for Blocks

thsparks:
# The Change
This change adds the error panel for blocks and removes the experiment for it. It also refactors a lot of the error list code to simplify, improve readability, and separate concerns (block-related error code remains in blocks code, monaco-related error code in monaco code, etc...).

Practically, this boils down to:
1. Creating an abstract error display representation (`ErrorDisplayInfo`) used by the ErrorList for all types of errors. Conversions to this format exist in their relevant editor files, as there's very little overlap with block and text errors.
2. Pass errors directly into the ErrorList, instead of using a listener setup (I didn't see much benefit to the listener approach, but maybe I missed something here).

# Get Block Text
To create a friendlier stack trace for blocks, I've also added a `getBlockText` function. I recall we wanted similar functionality for some of the recent a11y work, so I've put it in the pxtblocks. Hopefully that will make it more convenient for reuse later (@srietkerk FYI).

We didn't have a great way to reliably get a human-readable text for all of our custom fields (songs, images, etc...), so I've added a new `getFieldDescription()` API that all of our custom fields can implement. In the base class, this just returns `getText()`, but fields can override it as necessary.

I also added a new field type, `FieldImageNoText` to represent things like our expand button, which appears on the block but shouldn't really be included in these strings.

# Out of Scope
In javascript, if you have compile errors, then by their very nature, you cannot have runtime errors (and vice versa). For blocks, some of the errors we display do not actually block compilation, so you can end up with both. In this iteration, a "runtime" error (complete with stack trace) will overwrite any of the "passive" errors. The passive errors will only be shown if there is no runtime error. This is something we may want to revisit, but it needs a bit of design work, and I don't want to block this change on it.

# Try It
Upload Target: https://makecode.microbit.org/app/153b7647e8a4a1d91c5f79754353224049d13793-dfd7366fcf (for testing purposes only, you can see block strings in the right click menu for a block)


webapp/src/errorList.tsx
```diff
@@ -106,132 +110,72 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
...
+                </div>
+                <div className="ui selection list">
+                    {(error.stackFrames).map((childErr, index) => {
+                        const errGrp = {error: childErr, count: 1, index: 0};
+                        return <ErrorListItem key={index} index={index} errorGroup={errGrp} className="stackframe"/>
+                    })}
+                </div>
+            </div>
+        ) : (
+            <div className={classList("item", className)} role="button"
```
eanders-ms, line 163: Should the assignment of `role="button"` be conditional here, like it is above?

riknoll: or how about just making this an actual button?

webapp/src/errorList.tsx
```diff
@@ -106,132 +110,72 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
...
+                onClick={error.onClick}
+                onKeyDown={fireClickOnEnter}
+                aria-label={lf("Go to {0}: {1}", error.stackFrames ? '' : 'error', message)}
+                tabIndex={0}>
+                {message} {(errorGroup.count <= 1) ? null : <div className="ui gray circular label countBubble">{errorGroup.count}</div>}
```
riknoll, line 168: could we wrap this message in a span or something? feels weird to have text content next to html content

webapp/src/errorList.tsx
```diff
@@ -106,132 +110,72 @@ export class ErrorList extends React.Component<ErrorListProps, ErrorListState> {
...
+                        const errGrp = {error: childErr, count: 1, index: 0};
+                        return <ErrorListItem key={index} index={index} errorGroup={errGrp} className="stackframe"/>
+                    })}
+                </div>
+            </div>
+        ) : (
+            <div className={classList("item", className)} role="button"
+                onClick={error.onClick}
+                onKeyDown={fireClickOnEnter}
+                aria-label={lf("Go to {0}: {1}", error.stackFrames ? '' : 'error', message)}
```
riknoll, line 166: if stack frames is defined, wont this end up having an extra space?

webapp/src/blocks.tsx
```diff
@@ -891,14 +892,77 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
+            onClick: () => {
+                // Block may already be highlighted, so we want to set highlighted directly to true
+                // as opposed to using `highlightBlock`, which toggles.
+                const block = this.editor.getBlockById(blockId);
+                if (!block) {
+                    return;
+                }
+
+                block.setHighlighted(true);
+                this.editor.centerOnBlock(blockId);
```
riknoll, line 916:
```suggestion
                this.editor.centerOnBlock(blockId, true);
```

you need to pass true here because without it the editor will center on the block and all blocks that are connected after it

webapp/src/blocks.tsx
```diff
@@ -891,14 +892,77 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
+            stackFrames
+        };
+    }
+
+    private getBlockText(block: Blockly.BlockSvg): string {
+        const fieldValues = [];
+        for (const input of block.inputList) {
+            if (input.fieldRow.length > 0) {
+                for (const field of input.fieldRow) {
+                    const text = field.getText();
```
riknoll, line 958: this isn't going to work well for all fields; i'm guessing that assets are just going to produce some un-pronounceable nonsense

riknoll (changes requested):
couple of comments

webapp/src/errorList.tsx
thsparks: We can't even reach this case if we have stackFrames. I think it's just a miss on my end while refactoring. Will remove the check.

pxtblocks/fields/field_utils.ts
```diff
@@ -9,6 +9,11 @@ export interface FieldCustom {
     isFieldCustom_: boolean;
     saveOptions?(): pxt.Map<string | number | boolean>;
     restoreOptions?(map: pxt.Map<string | number | boolean>): void;
+
+    /**
+     * Returns a human-readable description of the field's current value.
+     */
+    getFieldDescription(): string;
```
thsparks, line 16: Probably need to make this optional so I don't break anyone using this outside of pxt

pxtblocks/fields/field_ledmatrix.ts
```diff
@@ -435,6 +435,10 @@ export class FieldMatrix extends Blockly.Field implements FieldCustom {
         return `\`\n${FieldMatrix.TAB}${text}\n${FieldMatrix.TAB}\``;
     }
 
+    getFieldDescription(): string {
+        return lf("LED Grid");
```
riknoll, line 439: nitpick: maybe include the dimensions? e.g. "5x5 LED Grid". the microbit has a couple blocks in the images category with different grid sizes

pxtblocks/fields/field_tileset.ts
```diff
@@ -314,6 +314,10 @@ export class FieldTileset extends FieldImages implements FieldCustom {
         this.localTile = asset as pxt.Tile;
         super.loadState(pxt.getTSReferenceForAsset(asset));
     }
+
+    getFieldDescription(): string {
+        return lf("tileset");
```
riknoll, line 319: i'd make this one "tile" rather than "tileset", since it generally refers to a single tile

riknoll (approved):
LGTM!

PR #10533, opened 2025-04-24
move flyout caching functionality into a block inflater

riknoll:
moves our multi-flyout caching to a flyout block inflater. the motivation for this is better compatibility with the keyboard navigation plugin, which has a hard assumption that there is only one flyout at any time. when i was doing the great blockly upgrade, the flyout inflater APIs weren't yet available so this is a nice cleanup that better aligns with blockly.

this takes heavy inspiration from the [continuous toolbox plugin's inflater](https://github.com/google/blockly-samples/blob/rc/v12.0.0/plugins/continuous-toolbox/src/RecyclableBlockFlyoutInflater.ts), but with some key differences:
* recycled blocks are keyed based off of the XML defining the block rather than the block type. we have a lot of repeated instances of block types in our flyouts (variable blocks, function call blocks, math operations, customized blocks in tutorials, etc.) which made caching based off block type infeasible
* recycled blocks have a CSS class applied to them to hide them. not actually sure how the continuous flyout gets away without doing this (i didn't try it out)
* the flyout cache isn't cleared when the flyout is shown/hidden. instead it's cleared whenever the header is reloaded

from my quick testing it seems like perf is the same or slightly better than our current caching.

@microbit-robert @microbit-matt-hillsdon @microbit-grace FYI. i did no testing with the newest keyboard plugin since i haven't integrated your changes but i assume this refactor is more or less what @rachel-fenichel was suggesting when she mentioned the continuous toolbox plugin


eanders-ms (approved):

riknoll:
@microbit-robert sounds good, i'm going to merge this as-is and we can iterate on it. hopefully we can get some guidance from the blockly team.

and i don't think we'd ever have disabled blocks in the flyout; certainly not possible today

rachel-fenichel:
Yes, this matches the approach that I expected. @gonfunko to spot-check that you're using the inflators correctly/as expected.

Blockly allows disabled blocks in the flyout and manages state for them, fyi.

riknoll:
@rachel-fenichel oh yeah, i know blockly supports it. i was just mentioning that makecode doesn't generate any flyouts with disabled blocks in them

PR #10537, opened 2025-04-25
Fixing remaining accessibility bugs

srietkerk:
The first commit is just something that I fixed because it was something that I noticed, but if you start indexing the commits after the merge, the fixes per commit go as follows:

1. Fixes https://github.com/microsoft/pxt-microbit/issues/6085
2. Fixes https://github.com/microsoft/pxt-microbit/issues/6099
3. Fixes https://github.com/microsoft/pxt-microbit/issues/6095


riknoll (approved):

react-common/styles/controls/Modal.less
```diff
@@ -110,6 +118,10 @@
         }
     }
 
+    .common-modal-back:focus-within {
+        border: 1px, solid var(--pxt-primary-foreground);
```
thsparks, line 122: Would --pxt-focus-border be more appropriate here?

thsparks (approved):
One comment, okay to ignore if the normal focus color doesn't work for some reason.

react-common/styles/controls/Modal.less
```diff
@@ -54,9 +54,17 @@
...
+    .common-modal-close {
+        border: 1px solid transparent;
+    }
+
     .common-modal-close .common-button .fas {
         color: var(--pxt-neutral-foreground1);
     }
+
+    .common-modal-close:focus-within {
+        border: 1px solid var(--pxt-neutral-foreground1);
```
thsparks, line 66: Same here, would --pxt-focus-border be more appropriate?

react-common/styles/controls/Modal.less
srietkerk: Sure, I can see how this looks!

thsparks (approved):

PR #10551, opened 2025-04-30
Add global shortcuts to improve keyboard navigation

microbit-robert:
See https://github.com/microsoft/pxt-microbit/issues/6151. We've moved to an approach where Ctrl-U shows an overlay with navigable regions labelled by number in an effort to move forward with this after some recent discussion.

See [accompanying change](https://github.com/microsoft/pxt-microbit/pull/6169) in `pxt-microbit` which ensures that global shortcuts are also enabled when focus is inside the simulator iframe (e.g., simulator buttons, etc).

[Link to demo](https://global-shortcuts.review-pxt.pages.dev/). Please note that 'accessible blocks' need to be set to 'on' in the settings menu. Use Ctrl-U to show the navigation overlay, and then a number to move to that region. Ctrl-U is a shortcut to view the page source code which feels like a seldom used shortcut and one we can safely override.

Issues to fix:
- [ ] Prevent the navigation overlay from running on the home page - it currently ends badly
- [ ] Ctrl-U does not working the Monaco text editor due to existing editor shortcuts

@microbit-matt-hillsdon

### DIFF
--- a/webapp/src/blocks.tsx
+++ b/webapp/src/blocks.tsx
@@ -585,6 +585,35 @@ export class Editor extends toolboxeditor.ToolboxEditor {
                     return true
                 }
             });
+
+            const triggerEditorAction = (action: pxsim.SimulatorAction) => {
+                switch (action) {
+                    case "escape": {
+                        this.parent.setSimulatorFullScreen(false);
+                        return;
+                    }
+                    case "navigateregions" : {
+                        this.parent.showNavigateRegions();
+                        return
+                    }
+                }
+            }
+
+            const simulatorOrigins = [
+                window.location.origin,
+                // Simulator deployed origin.
+                "https://trg-microbit.userpxt.io"
+            ]
+            window.addEventListener("message", (e: MessageEvent) => {
+                // Listen to simulator iframe keydown post messages.
+                if (simulatorOrigins.includes(e.origin) && e.data.type === "pxtsim") {
+                    triggerEditorAction((e.data as pxsim.SimulatorActionMessage).action)
+                }
+            }, false)
+            document.addEventListener("keydown", (e: KeyboardEvent) => {
+                const action = pxsim.accessibility.getKeyboardShortcutEditorAction(e)
+                triggerEditorAction(action)
+            });
         }
     }
 
--- a/webapp/src/components/NavigateRegionsOverlay.tsx
+++ b/webapp/src/components/NavigateRegionsOverlay.tsx
@@ -1,0 +1,336 @@
+import { useCallback, useEffect, useRef, useState } from "react";
+import * as ReactDOM from "react-dom";
+import { Button, ButtonProps } from "../../../react-common/components/controls/Button";
+import { FocusTrap } from "../../../react-common/components/controls/FocusTrap";
+
+import IProjectView = pxt.editor.IProjectView;
+
+interface NavigateRegionsOverlayProps {
+    parent: IProjectView;
+}
+
+interface RectBounds {
+    top: number;
+    bottom: number;
+    left: number;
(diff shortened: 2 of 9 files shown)


kmcnaught:
> Should the navigation overlay work with the skip links at the start of the page or not?

I personally prefer what is currently implemented, i.e. region 1 starts with the editor toggle, not the skip links. I view the value of this global navigation to be quicker navigation throughout a coding session, and the skip links as mainly being valuable on first arrival. Most of the skip links are available within a couple of tabs anyway. The most valuable skip link for me is the "jump to Blocks workspace" when you have just arrived. If you've just triggered the global navigation overlay then you don't need it!

webapp/src/blocks.tsx
```diff
@@ -17,6 +17,7 @@ import * as dialogs from "./dialogs";
 import * as blocklyFieldView from "./blocklyFieldView";
 import { CreateFunctionDialog } from "./createFunction";
 import { initializeSnippetExtensions } from './snippetBuilder';
+import * as cmds from "./cmds"
```
microbit-matt-hillsdon, line 20: Import change here and below can be reverted

webapp/src/blocks.tsx
```diff
@@ -585,6 +587,35 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
+                    }
+                }
+            }
+
+            const simulatorOrigins = [
+                window.location.origin,
+                // Simulator deployed origin.
+                "https://trg-microbit.userpxt.io"
+            ]
+            window.addEventListener("message", (e: MessageEvent<pxsim.SimulatorActionMessage>) => {
```
microbit-matt-hillsdon, line 609: This feels like it should be more positively identifying the message before making assumptions about its structure/type - let's compare to other sim messages and their handling.

webapp/src/components/NavigateRegionsOverlay.tsx
```diff
@@ -0,0 +1,222 @@
...
+        editortools: document.querySelector("#editortools").getBoundingClientRect(),
+    })
+    const [regionRects, setRegionRects] = useState(getRects())
+
+    const focusRegion = useCallback((region: Region) => {
+        switch (region) {
+            case "mainmenu": {
+                (document.querySelector(".blocks-menuitem") as HTMLElement).focus();
+                parent.hideNavigateRegions();
+                return
```
microbit-matt-hillsdon, line 60: Maybe we break in the branches and do this at the end?
Requires corresponding change below to not call this with an undefined region if the `shortcutToRegion` lookup fails.

microbit-robert: Refactored here [8471d4f](https://github.com/microsoft/pxt/pull/10551/commits/8471d4fd946481fd450a89c71c3530ef28e471d8)

webapp/src/blocks.tsx
microbit-robert: Removed unused imports [8105a6e](https://github.com/microsoft/pxt/pull/10551/commits/8105a6e3b718b7a64e3077fbe0f2fc73824c8ec0)

webapp/src/blocks.tsx
microbit-robert: We now check the data type as `pxtsim` before trying to use the action [59d54e0](https://github.com/microsoft/pxt/pull/10551/commits/59d54e0c15e1dc158ab73634a35cb08c54f6e27b)

microbit-robert:
Closing in favour of alternative PR.

microbit-matt-hillsdon:
Replaced by https://github.com/microsoft/pxt/pull/10570

PR #10552, opened 2025-04-30
adding missing monaco snippets for the math category

riknoll:
this PR adds monaco toolbox snippets for all of the math functions that are in the `square root` and `round` dropdown blocks. i omitted integer divide and integer multiply, since those are makecode specific and not part of the core js lib. 

fixes https://github.com/microsoft/pxt-microbit/issues/5579


webapp/src/monacoSnippets.ts
```diff
@@ -253,14 +253,146 @@ function cachedBuiltinCategories(): pxt.Map<BuiltinCategoryDefinition> {
                     },
                     retType: "number"
                 },
+                {
+                    name: "Math.sqrt",
+                    snippetName: "sqrt",
+                    snippet: `Math.sqrt(4)`,
+                    pySnippet: `Math.sqrt(4)`,
```
srietkerk, line 260: Does capital M `Math` exist in python? Or is this something special that we do?

webapp/src/monacoSnippets.ts
```diff
@@ -253,14 +253,146 @@ function cachedBuiltinCategories(): pxt.Map<BuiltinCategoryDefinition> {
...
+                        blockId: "math_js_atan2",
+                        jsDoc: lf("Returns the angle in radians from the point (0, 0) to the point (x, y), where y is the first argument passed to the function and x is the second argument passed to the function")
+                    },
+                    retType: "number"
+                },
+                {
+                    name: "Math.round",
+                    snippetName: "round",
+                    snippet: `Math.round(1.2)`,
+                    pySnippet: `Math.round(1.2)`,
```
srietkerk, line 344: Does round exist in python? Or is this again something that we do?

ganicke (approved):

jwunderl (approved):

webapp/src/monacoSnippets.ts
jwunderl: it's a good point, in python proper it is lowercase math https://docs.python.org/3/library/math.html; ours is off the js namespace / capitalized (e.g. `math.cos(x)` is not valid in a makecode python program, where `Math.cos(x)` is)

srietkerk (approved):

PR #10553, opened 2025-05-01
don't reuse cell refs when inlining rt calls in inlined functions

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/6014

our compiler inlines functions with simple bodies as an optimization.

the above bug is actually pretty tricky to reproduce, you need to have a function that meets the following requirements:

1. the function body is simple enough to be inlined
2. the function body contains a runtime call that uses a parameter more than once in its arguments
3. the runtime call is not one of the ["nomask" thumb operations](https://github.com/microsoft/pxt/blob/3cd6be970f348fb0d80e895c079f680117c2acd9/pxtcompiler/emitter/emitter.ts#L3410)
4. the argument being passed into the function is a variable reference

with all of that satisfied, then both of the arguments will be replaced with the same cell reference, which ends up breaking the assertion we have for var uses in `ProctoAssembler.clearArgs`. fix is to simply have two different cell refs that point to the same variable.

### DIFF
--- a/pxtcompiler/emitter/ir.ts
+++ b/pxtcompiler/emitter/ir.ts
@@ -487,7 +487,7 @@ namespace ts.pxtc.ir {
         }
     }
 
-    function inlineSubst(e: Expr): Expr {
+    function inlineSubst(e: Expr, isRtCallArg?: boolean): Expr {
         e = Expr.clone(e)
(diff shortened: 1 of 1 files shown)


mmoskal (approved):
Looks good for what's it's worth.

PR #10554, opened 2025-05-01
Keyboard accessibility improvements to Input component's dropdown area.

glastonbridge:
Demo build: [https://dropdown-key-input-capture.review-pxt.pages.dev/](https://dropdown-key-input-capture.review-pxt.pages.dev/)

This PR attempts to make keyboard accessibility closer to the [WAI-ARIA standard for combo boxes](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/) when an input provides a dropdown. Specifically:

* Down arrow opens the dropdown, where it exists
* The dropdown is given focus and is navigable by arrow keys, space and enter to select, and escape to abort
* There is only one active tab stop, removing the need for a separate stop to display the menu as that is nonstandard

### How to test

1. Enable the Accessible Blocks plugin and add a Sound Effect block ([example project](https://makecode.microbit.org/_5UHgrUFrmfmg))
2. Open the sound effect block using the keyboard and navigate to "Duration (ms)"
3. Use down arrow to open the dropdown, space or enter to select, escape to leave the dropdown without selecting
4. Observe it also works with mouse, or a combination of keyboard and mouse
5. Observe that escape key and enter key have now got their usual functions in blockly (NB: escape will not close the Sound Effect editor while a HTML input, such as the collapsed "Duration (ms)" field, is selected. This is part of Blockly's implementation and is not solved here).
6. Observe that the behaviour is correct on multiple openings of the dropdown and/or editor
 
### Implementation concerns

Some of the accessible keys for a dropdown are intercepted by Blockly at a high level, specifically escape and enter, and these make accessible navigation problematic. I have implemented a lifecycle operation, to be enabled only for the lifetime of the visible dropdown, which blocks those keyboard shortcuts then reasserts them on close.

This is the best solution that I could find using the information I had, but it is unsatisfyingly complex and I welcome better ways to achieve this.

Whatever approach we use here will need to be extended to the Dropdown component too.

### DIFF
--- a/react-common/components/controls/Input.tsx
+++ b/react-common/components/controls/Input.tsx
@@ -1,5 +1,4 @@
 import * as React from "react";
-import * as Blockly from "blockly";
 import { classList, ControlProps } from "../util";
 
 import { Button } from "./Button";
--- a/react-common/components/controls/Input.tsx
+++ b/react-common/components/controls/Input.tsx
@@ -130,6 +130,8 @@ export const Input = (props: InputProps) => {
         (e.target as HTMLElement).blur();
         expandButtonClickHandler();
         document.getElementById(id)?.focus();
+        e.stopPropagation();
+        e.preventDefault();
     }
 
     const iconClickHandler = () => {
--- a/react-common/components/controls/Input.tsx
+++ b/react-common/components/controls/Input.tsx
@@ -33,45 +33,6 @@ export interface InputProps extends ControlProps {
     onOptionSelected?: (value: string) => void;
 }
 
-/**
- * Top-level Blockly handlers for escape and enter interfere with standard 
- * keyboard controls for navigating dropdowns. This temporarily suspends
- * them.
- * 
- * @returns a callback to restore previous handlers, as useEffect expects
- */
-const suspendBlocklyKeyHandlers = (onEscape: () => void) => {
-    const closeDropdownShortcut = "close_dropdown";
-    const selectDropdownShortcut = "select_dropdown";
-
-    const closeDropdownHandler: Blockly.ShortcutRegistry.KeyboardShortcut = {
-        name: closeDropdownShortcut,
-        allowCollision: true,
-        callback() {
-            onEscape();
-            return true;
-        },
-        keyCodes: [27 /* Escape */],
-    };
-
-    const selectDropdownHandler: Blockly.ShortcutRegistry.KeyboardShortcut = {
-        name: selectDropdownShortcut,
-        allowCollision: true,
-        callback() {
-            return true;
-        },
-        keyCodes: [13 /* Enter */],
-    };
-
-    Blockly.ShortcutRegistry.registry.register(closeDropdownHandler);
-    Blockly.ShortcutRegistry.registry.register(selectDropdownHandler);
-
-    return () => {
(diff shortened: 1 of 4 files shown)


react-common/components/controls/Input.tsx
```diff
@@ -32,6 +33,45 @@ export interface InputProps extends ControlProps {
...
+    const selectDropdownShortcut = "select_dropdown";
+
+    const closeDropdownHandler: Blockly.ShortcutRegistry.KeyboardShortcut = {
+        name: closeDropdownShortcut,
+        allowCollision: true,
+        callback() {
+            onEscape();
+            return true;
+        },
+        keyCodes: [27],
```
riknoll, line 54:
```suggestion
        keyCodes: [27 /* Escape */],
```

react-common/components/controls/Input.tsx
```diff
@@ -32,6 +33,45 @@ export interface InputProps extends ControlProps {
...
+    };
+
+    const selectDropdownHandler: Blockly.ShortcutRegistry.KeyboardShortcut = {
+        name: selectDropdownShortcut,
+        allowCollision: true,
+        callback() {
+            return true;
+        },
+        keyCodes: [13],
```
riknoll, line 63:
```suggestion
        keyCodes: [13 /* Enter */],
```

react-common/components/controls/Input.tsx
```diff
@@ -73,6 +113,17 @@ export const Input = (props: InputProps) => {
         setValue(initialValue || "");
     }, [initialValue]);
 
+    React.useEffect(() => {
+        if (expanded) {
+            return suspendBlocklyKeyHandlers(() => 
+                {
+                    setExpanded(false);
+                    document.getElementById(id)?.focus();
+                });
```
riknoll, line 122:
```suggestion
                }
            );
```

react-common/components/controls/Input.tsx
```diff
@@ -73,6 +113,17 @@ export const Input = (props: InputProps) => {
...
+        }
+        return undefined;
+    },[expanded]);
```
riknoll, line 125:
```suggestion
    }, [expanded]);
```

riknoll (changes requested):
currently, react-common has no dependency on blockly and i'd like to keep it that way. react-common is shared between several of makecode's various webapps many of which have nothing to do with blockly at all.

is there a way we can make this a little more generic? could we create a wrapper component in `webapp/` that handles the blockly side of things maybe?

glastonbridge:
> is there a way we can make this a little more generic? could we create a wrapper component in `webapp/` that handles the blockly side of things maybe?

Hi @riknoll, thanks for your feedback, I appreciate the benefits of good separation of concerns. I've made an attempt to implement a wrapper component as you suggested, have a look at the latest commit and see what you think.

react-common/components/controls/Input.tsx
```diff
@@ -1,4 +1,5 @@
 import * as React from "react";
+import * as Blockly from "blockly";
```
riknoll, line 2: remove this unused import

riknoll (changes requested):
LGTM! just remove the import

riknoll (approved):

riknoll:
@glastonbridge looks like a linting issue! you can run `gulp lint` locally to check it

PR #10556, opened 2025-05-02
Add keyboard navigation help documentation

microbit-robert:
This is intended as a follow-up PR to https://github.com/microsoft/pxt/pull/10512 and https://github.com/microsoft/pxt/pull/10570, and we are opening in draft for visibility.

The diff in this PR is not representative of the [changes proposed](https://github.com/microbit-matt-hillsdon/pxt/compare/global-shortcuts...microbit-matt-hillsdon:pxt:keyboard-help-docs).

See https://github.com/microsoft/pxt-microbit/issues/6152. There is an alternative suggestion on the linked issue that these docs could opened in a separate sidebar tab. The work we have done here does not include that suggestion as we're hoping for more feedback on the style, position and discoverability of the help documentation before implementing further changes. 

[Link to demo](https://keyboard-help-docs.review-pxt.pages.dev/). Please note that 'accessible blocks' needs to be set to 'on' in the settings menu.

@microbit-matt-hillsdon

### DIFF
--- a/common-docs/browsers.md
+++ b/common-docs/browsers.md
@@ -1,9 +1,18 @@
 # Supported Browsers
 
+You can experience MakeCode on a number of different operating system and browsers. In order to provide the best performance and support for features, the compatibility lists of browsers and operating system support for MakeCode is shown in the following sections.
+
 ## Windows
 
-You need one of these browsers running on Windows 7, Windows 8, Windows 8.1, or
-Windows 10:
+You need to have one of these **Windows** versions:
+
+* Windows 11
+* Windows 10 
+* Windows 8.1 
+* Windows 8
+* Windows 7
+
+The compatible **Windows** browsers are:
 
 * Microsoft Edge ([Download](https://www.microsoft.com/en-us/edge))
 * Google Chrome
@@ -11,8 +20,19 @@ Windows 10:
 
 ## Mac
 
-You need one of these browsers running on OS X 10.9 Mavericks, OS X 10.10
-Yosemite, OS X 10.11 El Capitan, or macOS 10.12 Sierra:
+You need to have one of these **macOS** versions:
+
+* Sequoia 15
+* Sonoma 14
+* Ventura 13
+* Monterey 12
+* Big Sur 11
+* Catalina 10.15
+* Mojave 10.14
+* High Sierra 10.13
+* Sierra 10.12
+
+The compatible **macOS** browsers are:
 
 * Safari
 * Microsoft Edge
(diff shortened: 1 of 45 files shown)


webapp/src/shortcut_formatting.ts
```diff
@@ -0,0 +1,116 @@
...
+ * This could be considerably simpler if we only bound shortcuts relevant to the
+ * current platform or tagged them with a platform.
+ * 
+ * Copied from blockly-keyboard-experiment.
+ *
+ * @param action The action name, e.g. "cut".
+ * @param modifierNames The names to use for the Meta/Control/Alt modifiers.
+ * @returns The formatted shortcuts.
+ */
+function getActionShortcutsAsKeys(
```
microbit-matt-hillsdon, line 68: This is all a bit gross but matches the keyboard nav plugin's code for its simpler keyboard shortcut dialog.

As per the comment above, it would all be much nicer if Blockly only bound shortcuts relevant to the current platform or at least tagged them with a platform.

microbit-matt-hillsdon:
Rebased after the global shortcut change went in and now ready for review @riknoll.

I can't move from draft because @microbit-robert created this PR, Rob will do that tomorrow if no-one else can/does.

webapp/src/components/KeyboardControlsHelp.tsx
```diff
@@ -0,0 +1,177 @@
...
+        case "⌘": {
+            aria = lf("Command");
+            break;
+        }
+        case "⌥": {
+            aria = lf("Option");
+            break;
+        }
+    }
+    return <span className="key" aria-label={aria}>{lf("{0}", value)}</span>
```
riknoll, line 174: no need to use `lf()` here since it's used above!

riknoll (changes requested):
one tiny comment, but otherwise LGTM!

webapp/src/components/KeyboardControlsHelp.tsx
microbit-matt-hillsdon: Thanks. This turned out to be more interesting than expected. There were a mix: shortcuts that were in some way hardcoded had already been through `lf` but ones from the registry where not otherwise going through `lf`. I suppose even though `lf` was being called via the line above for the registry ones, the strings would likely not be extracted correctly.

I have restructured to do the translation more directly from the shortcut registry format to translated text in [51b63a2](https://github.com/microsoft/pxt/pull/10556/commits/51b63a296af39827808830c11dcd48f09ab83956). I also managed to drop the hardcoding for the shortcuts where we document two alternatives. Follow up to fix `lf` quote style in 0183179156785a6db79d0c25846b6ca837dd61f5.

PR #10559, opened 2025-05-05
Correct some erroneous info in 'substr' ref page

ganicke:
Fix some text discussing the use of the `length` parameter.

Closes https://github.com/microsoft/pxt-microbit/issues/5436


ganicke:
@abchatra, @riknoll - I think we might have a bug here. There is no provision for leaving out the `length` parameter in blocks (the **(+)** control is needed to add the `length` parameter if desired. Also, stating that a `length` of `0` will copy to the end of the string is also incorrect (I'll need to revert that info in my changes to this page).

JavaScript -

```typescript
let ostr = "kljsdfkjlfdskljfdslk"
let sstr = ostr.substr(5)
let tstr = "This is my string.".substr(4)
basic.showString(sstr)
```

then Blocks -

image

then back to JavaScript -

```typescript
let ostr = "kljsdfkjlfdskljfdslk"
let sstr = ostr.substr(5, 0)
let tstr = "This is my string.".substr(4, 0)
basic.showString(sstr)
```

RE:

```typescript
//% blockId="string_substr" block="substring of %this=text|from %start|of length %length"
//% this.defl="this"
substr(start: number, length?: number): string;
```

riknoll:
hmmmm... yeah that looks like a bug. let's deprecate this block and make a new one where the parameter is optional; i'll open an issue

ganicke:
Depends on #10562.

riknoll (approved):

PR #10560, opened 2025-05-05
fix mutate callers/def when child is a handlerstatement block

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/6182

when right clicking on a function and doing "create 'call functionname'", the following events occur:
1. a new instance of the call block is created with the function name but without any of the arguments populated
2. the "on workspace changed" handler in the function definition block runs and detects that a call block for this function has been pasted on the workspace
3. that handler sees that the block is missing the right arguments, so it calls `mutateCallersAndDefinition`
4. `mutateCallersAndDefinition` mutates the caller block and populates it with the correct arguments

this is all fine and good, but the `mutateCallersAndDefinition` function was deleting all child argument reporters of the definition that didn't belong to the function because it assumed they were deleted parameters. this PR refactors the connection checking code so that it can be reused to see if an argument has a valid owner other than the function definition before we delete it.

funny coincidence that this is so closely related to https://github.com/microsoft/pxt/pull/10558 which i just merged!


srietkerk (approved):
Nice!

PR #10562, opened 2025-05-05
deprecate String.substr block in favor of one with optional length

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/6184

the length parameter of `String.substr` is optional, but it's not currently optional in blocks. this PR corrects that by deprecating the old block and creating a new one with an optional length parameter (with the usual + button to access optional parameters).

this change requires no upgrade rules. the old block definition has been moved on to `String.__substr` but will still convert to `String.substr` when switching to JavaScript because of the `blockAliasFor` annotation.

here is the old block:

image

and here is the new one:

image

the expanded text is exactly the same as the old one.

### DIFF
--- a/libs/pxt-common/pxt-core.d.ts
+++ b/libs/pxt-common/pxt-core.d.ts
@@ -312,6 +312,7 @@ declare interface String {
     //% this.shadow="text"
     //% this.defl="this"
     //% blockNamespace="text"
+    //% expandArgumentsInToolbox
     substr(start: number, length?: number): string;
 
     /**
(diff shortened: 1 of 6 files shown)


riknoll:
also note that this will invalidate the old translations. no getting around that unfortunately

Jaqster:
I don't understand what will happen if you don't give it a length.  Seems more confusing this way.  Could we have it expanded by default?

riknoll:
@Jaqster let me see about making it expanded by default. we don't support it today but should be easy to add.

also, if you don't specify the length then it's just the substring from the start to the end (which is what you want nine times out of ten anyhow)

riknoll:
now expanded in the toolbox by default: 
image

srietkerk (approved):
LGTM!

PR #10566, opened 2025-05-07
workaround blockly flyout variable clearing behavior

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/6181

for a detailed explanation of the behavior in blockly that necessitates this change, see https://github.com/google/blockly/issues/9007


pxtblocks/plugins/flyout/blockInflater.ts
```diff
@@ -74,6 +68,46 @@ export class MultiFlyoutRecyclableBlockInflater extends Blockly.BlockFlyoutInfla
...
+        // cached blocks as referencing that variable. we can remap any variable that exists
+        // as both a potential and non-potential variable to a new one with the same name to
+        // prevent its usages from being deleted
+        const potentialVariableMap = block.workspace.getPotentialVariableMap();
+
+        if (potentialVariableMap) {
+            const allBlocks = [block].concat(block.getChildren(false));
+
+            // loop over all child variable fields
+            for (const block of allBlocks) {
```
thsparks, line 84: nit: could we rename this to childBlock or something, so it doesn't conflict with the block passed into the `recycleBlock` function?

thsparks (approved):
One nit, okay to ignore if you want

PR #10567, opened 2025-05-07
add ability to localize callback parameters

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/1507

this PR adds the ability to localize callback parameters in blocks with draggable reporters. the json entries for callback parameters look like this:

```
// qname|handlerParam|paramName
"radio.onReceivedString|handlerParam|receivedString": "receivedString",
```

i went with this format over the usual `{id:whatever}` thing we use for most `rlf` strings because i wanted to match what we do for the other block strings, which i guess predate the nicer curly brace syntax. 

in order to actually display these strings, i store this localization `rlf` key as a mutation in the block when generating XML for the toolbox. that means this will not retroactively apply to old projects, which will not have the mutation present. also, this will still compile to the actual parameter name when switching to javascript rather than the localized one

### DIFF
--- a/pxtblocks/plugins/functions/blocks/argumentReporterBlocks.ts
+++ b/pxtblocks/plugins/functions/blocks/argumentReporterBlocks.ts
@@ -10,6 +10,8 @@ import { MsgKey } from "../msg";
 import { DUPLICATE_ON_DRAG_MUTATION_KEY, setDuplicateOnDragStrategy } from "../../duplicateOnDrag";
 import { PathObject } from "../../renderer/pathObject";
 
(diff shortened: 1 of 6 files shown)


riknoll:
i should mention that this isn't a perfect solution, in particular this won't work automatically for blocks in tutorials. however, i think it should work if the person authoring the tutorial simply writes a valid localized identifier in the javascript that we decompile into blocks.

i haven't thought up a good decompiler scenario for this yet... it's tough to associate arguments with their functions since they are separate blocks. however, i think this is a pretty niche scenario

jwunderl (approved):

thsparks (approved):

PR #10570, opened 2025-05-09
Add global shortcuts to improve keyboard navigation

microbit-matt-hillsdon:
Adds Ctrl/Cmd+B shortcut that shows a numbered navigation overlay. Integrates with a pxt-microbit change that propagates global-nav actions from the simulator (otherwise they don't take effect when the iframe has focus).

See [microsoft/pxt-microbit#6151](https://github.com/microsoft/pxt-microbit/issues/6151).

See [accompanying change](https://github.com/microsoft/pxt-microbit/pull/6169) in `pxt-microbit` which ensures that global shortcuts are also enabled when focus is inside the simulator iframe (e.g., simulator A/B buttons, etc).

[Link to demo](https://global-shortcuts.review-pxt.pages.dev/). Please note that 'accessible blocks' need to be set to 'on' in the settings menu. Use Ctrl+B to show the navigation overlay, and then a number to move to that region.

We felt it was important to keep keys consistent across different editor arrangements which leads to some compromises:
- When editors without a toolbox are displayed, there is no region 3 (e.g. project settings).
- When the mini sim is in use, it appears out of sequence (because it's visually quite different I don't feel this one is too bad)
- Tutorial is numbered 0 and appears out of sequence as it is present only in certain modes

An alternative might be to use Ctrl+B, Single letter. We think numbers are a bit more straightforward (easy to find on the keyboard). Letters could try to match region names, at least in English. It's not clear that there are good user facing names for the toolbars (maybe h/f for header/footer?) and tutorial and toolbox collide.


webapp/src/components/NavigateRegionsOverlay.tsx
```diff
@@ -0,0 +1,336 @@
...
+    {
+        id: "editor",
+        ariaLabel: lf("Editor"),
+        shortcutKey: "4",
+        getBounds(projectView: IProjectView) {
+            const editorSelectors = ["#pxtJsonEditor", "#githubEditor", "#blocksArea", "#serialEditor", "#assetEditor", "#monacoEditor"];
+            for (const selector of editorSelectors) {
+                const element = document.querySelector(selector) as HTMLElement | null;
+                if (element.offsetParent !== null) {
+                    console.log(element);
```
microbit-matt-hillsdon, line 123: Whoops.

```suggestion
```

webapp/src/app.tsx
```diff
@@ -5247,6 +5248,21 @@ export class ProjectView
...
+    ///////////////////////////////////////////////////////////
+    ////////////             Navigate regions     /////////////
+    ///////////////////////////////////////////////////////////
+
+    hideNavigateRegions() {
+        this.setState({ navigateRegions: false });
+    }
+
+    showNavigateRegions() {
+        const dialog = Array.from(document.querySelectorAll("[role=dialog]")).find(dialog => (dialog as any).checkVisibility());
```
riknoll, line 5260: checkVisibility isn't supported in all of the clients we target, but i wrote a polyfill for it a while back:

https://github.com/microsoft/pxt/pull/10241

probably need to move that polyfill up a bit so that it's applied before blockly is initialized, since this code might be called in a project with no blocks

webapp/src/blocks.tsx
```diff
@@ -585,6 +585,35 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
+                        this.parent.showNavigateRegions();
+                        return
+                    }
+                }
+            }
+
+            const simulatorOrigins = [
+                window.location.origin,
+                // Simulator deployed origin.
+                "https://trg-microbit.userpxt.io"
```
riknoll, line 605: pull this from `pxt.webConfig.simUrl`

webapp/src/blocks.tsx
```diff
@@ -585,6 +585,35 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
+            ]
+            window.addEventListener("message", (e: MessageEvent) => {
```
riknoll, line 607: i'd rather this code go directly in app.tsx in `handleMessage`, which already handles the other messages that are sent from the sim

webapp/src/components/NavigateRegionsOverlay.tsx
```diff
@@ -0,0 +1,347 @@
...
+                ? document.querySelector(".simPanel")
+                : projectView.state.collapseEditorTools ?
+                    document.querySelector("#computertogglesim")
+                    : document.querySelector("#editorSidebar")
+            const bounds = element?.getBoundingClientRect();
+            if (projectView.state.collapseEditorTools) {
+                // Custom presentation for the collapsed sim.
+                const isRtl = pxt.Util.isUserLanguageRtl();
+                const collapsedSimPadding = 30;
+                const copy = DOMRect.fromRect(bounds);
```
riknoll, line 72: bounds might be undefined here

webapp/src/components/NavigateRegionsOverlay.tsx
```diff
@@ -0,0 +1,347 @@
...
+        },
+        focus(projectView: IProjectView) {
+            // Note that pxtsim.driver.focus() isn't the same as tabbing to the sim.
+            if (isSimMini()) {
+                projectView.setSimulatorFullScreen(true);
+            } else {
+                if (projectView.state.collapseEditorTools) {
+                    projectView.toggleSimulatorCollapse();
+                }
+                (document.querySelector("#boardview") as HTMLElement).focus();
```
riknoll, line 91: probably best to guard this in case boardview isn't defined.

webapp/src/components/NavigateRegionsOverlay.tsx
```diff
@@ -0,0 +1,347 @@
...
+            }
+        },
+        getClassName(projectView: IProjectView) {
+            return `simulator-region ${projectView.state.collapseEditorTools ? "simulator-collapsed" : ""
```
riknoll, line 95: something weird going on with the string here

webapp/src/components/NavigateRegionsOverlay.tsx
```diff
@@ -0,0 +1,347 @@
...
+
+interface RegionButtonProps extends ButtonProps {
+    shortcutKey: string;
+    bounds: RectBounds;
+}
+
+const RegionButton = ({ shortcutKey, bounds, ...props }: RegionButtonProps) => {
+    const buttonRef = useRef<HTMLButtonElement>()
+
+    useEffect(() => {
```
riknoll, line 262: is there a reason to use useEffect here? i think you could just set it in the style object in the JSX and let the usual props updating logic work

riknoll (changes requested):

webapp/src/app.tsx
microbit-matt-hillsdon: Good spot. I've generally been using offsetParent === null but that won't work here because the modals use `position: fixed`.

webapp/src/blocks.tsx
microbit-matt-hillsdon: 52dc00c1334d5ab7070a2f45d0aa4de069437064

webapp/src/components/NavigateRegionsOverlay.tsx
microbit-matt-hillsdon: 52dc00c1334d5ab7070a2f45d0aa4de069437064 (and other similar case)

webapp/src/components/NavigateRegionsOverlay.tsx
microbit-matt-hillsdon: 52dc00c1334d5ab7070a2f45d0aa4de069437064

webapp/src/components/NavigateRegionsOverlay.tsx
microbit-matt-hillsdon: Oops, yeah it wrapped weirdly. On reflection I moved to managing class names with an array which felt clearer absent a utility for combining class names. 52dc00c1334d5ab7070a2f45d0aa4de069437064

webapp/src/components/NavigateRegionsOverlay.tsx
microbit-matt-hillsdon: Yep, thanks, changed as suggested in 52dc00c1334d5ab7070a2f45d0aa4de069437064

(Incidentally, I initially misinterpreted this and tried to remove the key handling part of the other useEffect, but that can't be done nicely as the FocusTrap div gets the focus but doesn't have an onKeyDown prop.)

webapp/src/blocks.tsx
microbit-matt-hillsdon: 2e84c145bc2f17c4a6353b54c2c4f6de8ec7761f

It made sense to move a bit more of the code into project view as the sim event and key event handling share code.

So far as I can tell that message handler doesn't check the origin of sim messages but I figure if that's going to change it should be a separate PR. Presumably it could do something like:

```typescript
const simulatorUrl = new URL(pxt.webConfig.simUrl, window.location.origin);
const simulatorOrigin = simulatorUrl.origin;
 window.addEventListener("message", ev => {
    if (ev.origin !== simulatorOrigin) {
        return;
    }
    // ...
```

?

webapp/src/app.tsx
microbit-matt-hillsdon: Moved alongside app.tsx. I've kept it separate from the pxtsim polyfill code but could combine them. I didn't think the sim embed would need checkVisible.

99c5974e34abdbb253076ab5952c251a5c83b4b9

Incidentally, there might be quite a few polyfills for IE and very old Safari that can be removed from injectPolyphils (!) now.

microbit-matt-hillsdon:
Thanks for the review @riknoll, I think I've addressed the feedback.

theme/navigateregions.less
```diff
@@ -0,0 +1,72 @@
...
+/* Reference import */
+@import (reference) "semantic.less";
+
+.navigate-regions-container {
+    position: fixed;
+    top: 0;
+    left: 0;
+    right: 0;
+    bottom: 0;
+    background-color: rgba(0,0,0,0);
```
thsparks, line 14: For the various transparency backgrounds, I'd recommend using the `pxt-neutral-alpha*` theme variables:
```
"pxt-neutral-alpha0": "rgba(0, 0, 0, 0)",
"pxt-neutral-alpha10": "rgba(0, 0, 0, 0.1)",
"pxt-neutral-alpha20": "rgba(0, 0, 0, 0.2)",
"pxt-neutral-alpha50": "rgba(0, 0, 0, 0.5)",
"pxt-neutral-alpha80": "rgba(0, 0, 0, 0.8)",
```

The main benefit is for dark themes, where these will fade white instead of black (which usually isn't very visible against a dark theme).

If 0% 10%, 20%, 50%, and 80% don't cover your cases, you can add more values to the base theme, but it'll probably be easier to just use whichever one is closest to your current value.

theme/navigateregions.less
```diff
@@ -0,0 +1,72 @@
...
+    width: 100%;
+    height: 100%;
+    z-index: @modalDimmerZIndex;
+
+    .region-button {
+        position: absolute;
+        background-color: rgba(0, 0, 0, .6);
+        border-radius: 0;
+        border: 3px solid white;
```
thsparks, line 23: I realize it'll be a little backwards (using foreground theme colors for a background and vice versa) but it's probably still slightly preferable to use neutral theme values in this file instead of `white` and `back`.  I'd recommend trying `pxt-neutral-background1` instead of white and `pxt-neutral-foreground1` instead of black.

theme/navigateregions.less
```diff
@@ -0,0 +1,75 @@
...
+    background-color: var(--pxt-neutral-alpha0);
+    width: 100%;
+    height: 100%;
+    z-index: @modalDimmerZIndex;
+
+    .region-button {
+        position: absolute;
+        background-color: var(--pxt-neutral-alpha50);
+        border-radius: 0;
+        border: 3px solid white;
```
thsparks, line 23:
```suggestion
        border: 3px solid var(--pxt-neutral-background1);
```

theme/navigateregions.less
microbit-matt-hillsdon: pxt-neutral-background1 (for light theme white) on a background of --pxt-neutral-alpha80 in high contrast seems to be black on black.

I've tweaked those cases to use --pxt-secondary-foreground for what was white. Hope that's OK.

Found an amazing consequence of the high contrast theme setting `z-index: 1` on button hover. Hovering over the editor region made the mini-sim vanish. Tweaked z-index as a result.

thsparks (approved):
One nitpick, but otherwise LGTM (will wait for Richard's sign-off too before merging).

riknoll (approved):

theme/navigateregions.less
thsparks: Thanks for double checking that. I did not think to check high-contrast. Your notes all seem good to me!

PR #10571, opened 2025-05-09
Add keyboard controls to imagedropdown and gridpicker field editors

microbit-robert:
The PR adds keyboard controls to the imagedropdown and gridpicker field editors.

Keyboard controls:
- Arrow keys: move the selection based on the grid. Left and Right overflow into the previous and next rows, respectively. Up and Down do not overflow.
- Home / End: move the selection to the start / end of the row
- Home / End with Ctrl or Cmd modifier: move the selection to the start / end of the grid

If the dropdown is opened with the keyboard, keyboard focus is shown, otherwise it is not. Mouse movements take priority, but pointermove events are listened to in order to prevent content moving under the pointer during scrolling from being interpreted as user triggered pointer movements. Keyboard control can be re-established at any point, even if the mouse has left an icon or the dropdown div completely, by using the keys noted above. 

We intend to move over to using [blockly-samples/field-grid-dropdown](https://github.com/google/blockly-samples/tree/master/plugins/field-grid-dropdown) once this has been updated to use Blockly v12.

See https://github.com/microsoft/pxt-microbit/issues/6077


pxtblocks/fields/field_gridpicker.ts
```diff
@@ -188,38 +285,46 @@ export class FieldGridPicker extends Blockly.FieldDropdown implements FieldCusto
             }
 
             if (this.shouldShowTooltips()) {
-                Blockly.browserEvents.conditionalBind(menuItem, 'click', this, this.buttonClickAndClose_);
+                Blockly.browserEvents.conditionalBind(menuItem, 'click', this, () => this.buttonClickAndClose_.bind(this)(value));
```
riknoll, line 288: shouldn't need to do both an arrow function and a call to `.bind(this)`. just the arrow function should do it. same below on 351+352

pxtblocks/fields/field_gridpicker.ts
```diff
@@ -348,7 +453,14 @@ export class FieldGridPicker extends Blockly.FieldDropdown implements FieldCusto
      * Create a dropdown menu under the text.
      * @private
      */
-    public showEditor_() {
+    public showEditor_(e?: MouseEvent) {
+        if (e) {
+            this.openingPointerCoords = {
```
riknoll, line 458: this should be disposed inside of `onClose_` right?

riknoll (changes requested):
seeing as how all three of these fields extend the same class; would be nice to see this duplicated code refactored into an abstract class that the fields can extend instead

pxtblocks/fields/field_gridpicker.ts
microbit-robert: Agreed, fixed here [1c0da5d](https://github.com/microsoft/pxt/pull/10571/commits/1c0da5daa210f9a870cb279da5b4fb9d30eb6808)

pxtblocks/fields/field_gridpicker.ts
microbit-robert: After refactoring, this is now being cleared inside `disposeGrid`, which is called in the same places as `disposeTooltip`, including `onClose_`.

microbit-robert:
> seeing as how all three of these fields extend the same class; would be nice to see this duplicated code refactored into an abstract class that the fields can extend instead

@riknoll Thanks for the review. Duplicated code and shared properties have now been moved into an abstract class `FieldDropdownGrid`, which these three fields now extend. I think all feedback has now been addressed.

PR #10576, opened 2025-05-12
remove the double quoting of commit messages in bump command

riknoll:
this is breaking the regex in the github action

the extra quotes are unnecessary because `child_process.spawn` already handles that for the args you pass i believe


riknoll:
confirmed, quotes are required on command prompt in windows. closing this; i'll open a new pr with a more complete fix

PR #10583, opened 2025-05-13
Color contrast: have skip links use a black color rather than navy

srietkerk:
Fixes https://github.com/microsoft/pxt-microbit/issues/6198

### DIFF
--- a/theme/accessibility.less
+++ b/theme/accessibility.less
@@ -11,7 +11,7 @@
 *******************************/
 
 @accessibleMenuBackground: rgba(255,255,255,.9);
-@accessibleMenuColor: var(--pxt-neutral-foreground1);
+@accessibleMenuColor: black;
 
 /* Main menu focus */
 #mainmenu .ui.item:focus {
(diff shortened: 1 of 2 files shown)


theme/accessibility.less
```diff
@@ -11,7 +11,7 @@
 *******************************/
 
 @accessibleMenuBackground: rgba(255,255,255,.9);
-@accessibleMenuColor: #387894;
+@accessibleMenuColor: var(--pxt-neutral-foreground1);
```
thsparks, line 14: Since the background is hard-coded, we should probably hard-code the foreground to black as well. (Otherwise, it'll probably be white on white in dark mode).

In an ideal world, we'd have both be theme colors instead of hard-coded, but I think a white fade makes sense here regardless of theme, so I guess it's fine. Otherwise we'd have to add theme vars for that 90% white color, which I don't think we really have at the moment (kind of the inverse of the `neutral-alpha` colors)...I wouldn't worry about it for now unless we find ourselves using this white fade in more places.

thsparks: To clarify, this is a long-winded way of me saying we should probably just hard-code the foreground to black so it's guaranteed to contrast with the hard-coded white fate.

thsparks (approved):

PR #10590, opened 2025-05-15
Error Help - Part 1

thsparks:
# Overview
This is the first set of front-end changes for the "Error Help" feature. It adds a "Help me understand" button to the error list, which sends a request to our backend for assistance on the current errors in the error list. In blocks, it then displays help in the form of an editor tour. In text, it simply adds a small chunk of explanatory text to the top of the error list.

Part of https://github.com/microsoft/pxt-microbit/issues/6138

# Breakdown
The key components of the change are:
1. The actual error-help logic (packaging up code, processing the response). This mostly happens in the `errorHelp.ts` file.
2. Some refactoring for the editor tour.
    - This was already pretty flexible, but I wanted to support different colors and some tour-level settings (like whether or not to show confetti or include the final step in the step count).
    - To make that change, I introduced a TourConfig with tour-level settings, but I kept colors at the bubble-level in case we ever want to have different colors within one tour.
    - It also seemed prudent to pull all of the "current step" vs "total steps" comparisons into the Tour instead of the Bubble, since tour-level settings affect it. This was the biggest component of the refactor.
4. Sign-in is required. If you're not signed-in, clicking the button will trigger the sign-in modal with some custom text.
5. Error handling & glue connecting various layers. I tried to minimize passing the "parent" around, and I tried to keep block-specific stuff inside `blocks.tsx` while text-specific stuff stayed in `monaco.tsx`...but I'll admit, it still feels a little scattered.

# What is NOT in this change
In the interest of keeping this change to a (somewhat) manageable size, I've decided to postpone the following items for future changes. Sub-issues filed here: https://github.com/microsoft/pxt-microbit/issues/6138
**Must Haves**
1. Permissions popup (to get user permission to send to AI)
2. Feedback (helpful / not helpful)
3. AI Disclaimer

**Nice to Haves**
1. Caching responses in case the user clicks it several times for the same error
2. Possibly "repairing" invalid block ids (i.e. checking if it's a variable/function id instead, and if we can map that to a block)
3. Premade responses for some error types, so we can skip the AI altogether
5. Trimming overly-repetitive stack traces (like stack overflows) to save on tokens
6. Mobile layout
7. Better icon. It's currently a question mark. Ideally, it'd be fun for this to be some "helper" themed to the target (like the duck), but I'm not sure what that would be for microbit. Would also be nice to include them in the error tour header. But could also just use a better icon, if people have ideas.
8. Nicer loading animation - I may need some help on this, can't think of how to make it look good. I've gone with "simple" for now.

Try it: https://microbit.staging.pxt.io/app/67201f689bf390a088508664dd715cfef4585b54-3712e8e736


webapp/src/components/onboarding/Tour.tsx
```diff
@@ -26,20 +26,22 @@ export const Tour = (props: TourProps) => {
...
         stepDuration: getStepDuration(),
         tourDuration: getTourDuration(),
         step: currentStep + 1,
-        totalSteps: tourSteps.length
+        totalSteps: steps.length
     });
 
     const onNext = () => {
         pxt.tickEvent("tour.next", data());
+        const nextStep = currentStep + 1;
```
eanders-ms, line 38: Are `nextStep` (and `prevStep` below) unused?

webapp/src/blocks.tsx
```diff
@@ -990,6 +1001,60 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
+            const helpResponse = await getErrorHelpAsTour(this.errors, "blocks", code);
+            const tour = this.createTourFromResponse(helpResponse);
+            this.parent.showTour(tour);
+        } catch (e) {
+            pxt.reportException(e);
+
+            if (e instanceof ErrorHelpException) {
+                core.errorNotification(e.getUserFacingMessage());
+            } else {
+                core.errorNotification(lf("Sorry, something went wrong. Please try again later."));
```
eanders-ms, line 1019: Suggest removing "Sorry".

```suggestion
                core.errorNotification(lf("Something went wrong. Please try again later."));
```

webapp/src/components/onboarding/Tour.tsx
thsparks: Yup, not even sure why I added them. Maybe from previous iteration I undid, maybe an overzealous ai autocomplete I didn't notice 🤔 Either way, I'll remove.

eanders-ms (approved):
Looks great. Added Copilot to see if I missed anything :).

riknoll (approved):

riknoll (approved):

PR #10596, opened 2025-05-21
Blockly 12 + keyboard nav plugin update

microbit-matt-hillsdon:
Blockly 12 and the plugin will definitely need a further upgrade to raise quality (there are useful fixes in core that aren't yet released, e.g. keyboard nav issues after delete, and the plugin is expected to get a few fixes too).

Demo: https://blockly-12-keyboard-exp.review-pxt.pages.dev/#editor

For comparison to see if identified issues are fixed in develop branch of Blockly / latest plugin you can check [this deployment](https://preview.review-pxt.pages.dev/#editor).

We'll follow up with a PR to use this plugin version to reinstate cross-tab copy/paste when accessible blocks is on.


package.json
```diff
@@ -160,10 +160,10 @@
       "source-map": "0.4.4"
     },
     "@blockly/field-colour": {
-      "blockly": "^12.0.0-beta.4"
+      "blockly": "^12.0.0"
```
microbit-matt-hillsdon, line 163: There are now released versions we can switch to.

pxtblocks/fields/field_textdropdown.ts
```diff
@@ -140,11 +140,11 @@ export class BaseFieldTextDropdown extends Blockly.FieldTextInput {
         for (let i = 0; i < options.length; i++) {
             const [label, value] = options[i];
             const content = (() => {
-                if (typeof label === 'object') {
+                if (isImageProperties(label)) {
```
microbit-matt-hillsdon, line 143: Align with Richard's approach. Worried this is stricter re alt.

pxtblocks/plugins/renderer/pathObject.ts
```diff
@@ -48,8 +48,41 @@ export class PathObject extends Blockly.zelos.PathObject {
         super.updateSelected(enable);
     }
 
-    override addConnectionHighlight(connection: Blockly.RenderedConnection, connectionPath: string, offset: Blockly.utils.Coordinate, rtl: boolean): void {
-        super.addConnectionHighlight(connection, connectionPath, offset, rtl);
+    override addConnectionHighlight(connection: Blockly.RenderedConnection, connectionPath: string, offset: Blockly.utils.Coordinate, rtl: boolean): SVGElement {
```
microbit-matt-hillsdon, line 51: This is a bit unfortunate. Matt/Rob to discuss and see if there's something we can feed back to Blockly here.

package.json
microbit-matt-hillsdon: cf5ffcb94c8b7583cebd5a7dca72f3a04ed8c6d8

pxtblocks/fields/field_textdropdown.ts
microbit-matt-hillsdon: Checked and alt is not optional. Shared the utility between the two scenarios which is as per the Blockly code.

pxtblocks/plugins/renderer/pathObject.ts
microbit-matt-hillsdon: We looked at this and don't have a straightforward suggestion here.

Blockly added a return type and moved to proactively calling `addConnectionHighlight`, controlling visibility with CSS `display` and only removing the highlight in `dispose`.

By default this meant our override always showed the custom part.

So we changed our override to return a group so that both parts are controlled by the same CSS display change Blockly makes. But that requires a bit of copy/paste and access to the connectionHighlights map in super.

Suggestions for now this could be improved very welcome.

theme/blockly-core.less
```diff
@@ -396,40 +397,72 @@ text.blocklyCheckbox {
         Focus styles
 *******************************/
 
-// Avoid focus outlines when not using the keyboard navigation plugin.
-div.blocklyTreeRoot > div > div[role="tree"]:focus-visible {
-    outline: none;
+
+html {
```
microbit-matt-hillsdon, line 401: CSS as per plugin test page. Probably not its ultimate home but a copy is needed here for now.

pxtblocks/plugins/comments/bubble.ts
```diff
@@ -7,7 +7,7 @@ import dom = Blockly.utils.dom;
  * bubble, where it has a "tail" that points to the block, and a "head" that
  * displays arbitrary svg elements.
  */
-export abstract class Bubble implements Blockly.IDeletable {
+export abstract class Bubble implements Blockly.IDeletable, Blockly.IBubble, Blockly.ISelectable {
```
microbit-matt-hillsdon, line 10: Where can we see this in action?

microbit-matt-hillsdon:
### Dragging block down in Music category doesn't close flyout

Instead it does nothing or can cause scroll behaviours. Doesn't affect e.g. Basic.

https://github.com/user-attachments/assets/0668ee64-56c7-4a43-8d04-48b5c6946067

microbit-matt-hillsdon:
### Clicking on the input fields for certain blocks cause them to be in drag mode.

**Repro steps**
1. Open https://blockly-12.review-pxt.pages.dev/#editor.
2. Create new project and drag in one of the following blocks (non-exhaustive of all that have issues):

Image

Image

Image

Image

Image

Image

Image

Image

Image

Image

3. Click on the text field to try and edit.
4. See the block go into drag mode. Repeated clicking does not stop the block from being dragged.

microbit-matt-hillsdon:
@microbit-robert spotted this logged error clicking on comments, likely due to the upgrade. Looking into it:

```
main.js:366 Uncaught TypeError: r.getBubble is not a function
    at Vd.lookUpFocusableNode (main.js:366:211539)
    at Js.findFocusableNodeFor (main.js:361:173463)
    at t (main.js:361:174171)
    at main.js:361:174523
```

microbit-matt-hillsdon:
I think this is in an OK place now @riknoll.

microbit-matt-hillsdon:
> Looking at the conflicts...

Fixed the conflicts. I've rebased, as it was easier for me to review the changes that way.

Blockly are intending to release an updated plugin later today that could be incorporated in this PR or as a follow-up depending on when it gets review.

microbit-matt-hillsdon:
> ### "show leds" mouse interaction glitch

This is now fixed so the PR is back to being reviewable/mergeable. We'll continue testing today too as this has certainly repaid testing to date with bugs!

pxtblocks/loader.ts
```diff
@@ -692,7 +692,7 @@ function initComments() {
 
 function initTooltip() {
     const renderTip = (el: any) => {
-        if (el.disabled)
+        if (el.hasDisabledReason?.("disabled"))
```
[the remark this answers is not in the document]
microbit-matt-hillsdon: Whoops, I fixed the others via conflicts but missed this one. 7159c20a8cdec30bed139ae7bf3c05360d428666

PR #10597, opened 2025-05-22
Use Up and down arrow keys in note field to select notes

glastonbridge:
Enable the user to select a note value via the keyboard using the up and down arrow keys.

https://github.com/user-attachments/assets/c2919ede-4f62-44c0-b617-6fded5545142

This is an unusual field for keyboard navigation because the Block's "frequency" input remains focused rather than moving focus to the piano view as an editor would. To make the piano view into a separate keyboard-navigable widget would be a larger piece of work. I suggest this as a pragmatic improvement that improves accessibility and makes the editor more engaging to keyboard users without otherwise changing the note field.

**Live demo:** https://note-field-keyboard-accessib.review-pxt.pages.dev/

### Testing

1. Enable keyboard navigation.
2. Add a "Play Tone" block from the "Music" flyout.
3. Navigate to the block using the keyboard and activate the block
4. You can type in a frequency value as before or interact with a mouse, but also press up and down to navigate keys within the piano view

### DIFF
--- a/pxtblocks/fields/field_note.ts
+++ b/pxtblocks/fields/field_note.ts
@@ -249,6 +249,11 @@ export class FieldNote extends Blockly.FieldNumber implements FieldCustom {
         this.refreshText()
     };
 
+    protected widgetDispose_(): void {
+        this.htmlInput_.addEventListener("keydown", this.keyHandler);
+        super.widgetDispose_();
+    }
+
     /**
      * Create a piano under the note field.
      */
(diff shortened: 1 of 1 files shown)


pxtblocks/fields/field_note.ts
```diff
@@ -249,6 +249,11 @@ export class FieldNote extends Blockly.FieldNumber implements FieldCustom {
         this.refreshText()
     };
 
+    protected widgetDispose_(): void {
+        this.htmlInput_.addEventListener("keydown", this.keyHandler);
```
riknoll, line 253: should this be removeEventListener?

riknoll (changes requested):

pxtblocks/fields/field_note.ts
glastonbridge: Oops, yes it definitely should! Fixed.

PR #10614, opened 2025-05-28
Add Evaluate Button to Share Page

thsparks:
This adds an "Evaluate" button to the share page, which will only appear if the `teachertool/showSharePageEvalButton` is set to true in targetconfig.

It also includes some fixes to the load-project-from-url behavior that already existed in the teacher tool. Specifically, it:
1. Ensures we preserve the hash and params on sign-in.
2. Updates the share link input.
3. Loads a new default checklist and goes to the results page automatically

The default checklist will only load if the teacher does not already have a checklist loaded. This way, if they're grading the same project for a bunch of students and have a customized rubric, they can still click the evaluate button here and it will load the same rubric they've been working with. It also prevents any accidental loss of work, which could happen if we overwrote the active rubric.

image


docfiles/script.html
```diff
@@ -140,6 +141,14 @@ <h2 class="ui item header">@title@</h2>
                                 { shareurl: "@id@" },
                             );
                         }
+
+                        var evalEnabled = targetConfig && targetConfig.teachertool.showSharePageEvalButton;
```
srietkerk, line 145: Is `var` used here to stay consistent with the rest of the file?

teachertool/src/components/ShareLinkInput.tsx
```diff
@@ -59,6 +68,7 @@ export const ShareLinkInput: React.FC<IProps> = () => {
                 preserveValueOnBlur={true}
                 autoComplete={false}
                 handleInputRef={setInputRef}
+                initialValue={text}
```
srietkerk, line 71: If we have this now, since we have text set to the empty string, will that mean that if users get to the teacher tool via something other than clicking the evaluate button, they'll see an empty text box in the project area or does the placeholder show?

srietkerk (commented):
Just a couple of questions

teachertool/src/transforms/handleProjectOnLoadAsync.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+export async function handleProjectOnLoadAsync(defaultChecklistUrl?: string | undefined): Promise<void> {
+    const { state } = stateAndDispatch();
+
+    const projectParam = window.location.href.match(/project=([^&]+)/)?.[1];
+    if (!!projectParam) {
+        const decoded = decodeURIComponent(projectParam);
+        const shareId = pxt.Cloud.parseScriptId(decoded);
+        if (!!shareId) {
+            pxt.tickEvent(Ticks.LoadProjectFromUrl);
```
eanders-ms, line 19: If would you prefer to remove the `project` param from the URL after processing it, this can be accomplished without a page reload using `window.history.replaceState`.

jwunderl (approved):

srietkerk (approved):
LGTM!

thsparks:
> I asked LLM itself on what are the good questions for middle school students project.
> 
> ✅ Basic Code Quality
> 
> * Is the code clear and easy to understand?
> * Are the blocks arranged in a logical order?
> * Do the names of variables make sense for what they do?
> * Are there any blocks that don’t seem necessary?
> 
> 🐞 Bug Checking
> 
> * Are there any bugs or mistakes in this program?
> * Does the program work the way it’s supposed to?
> * Is anything missing that would make the program work better or more completely?
> 
> 💡 Suggestions & Improvements
> 
> * What are some ways this program could be improved?
> * Can the code be made simpler or shorter without changing how it works?
> * What could the student add to make this project more interesting or fun?

Those seem to align fairly well with the questions included, which are:
1. Does this program demonstrate good software development practices? (Code Quality)
2. Are there any bugs in this program? (Bug checking)
3. How could this program be improved? (Suggestions & Improvements)

Jaqster:
Yeah, I'm not sure.  Just feels like evaluating the code is as important as opening the project in the editor (which it isn't).  We can keep and see what happens?  I think we'll just get a lot of confused people clicking on that button.  The other thing maybe we could do is only show it if you're signed in already to try and scope this feature to Teachers?

abchatra:
We could use different color for edit button as well, if that helps to prioritize. Sign-in for share page is lot of work and we shouldn't pick for this release.

PR #10620, opened 2025-05-30
Use widget div for ledmatrix field editor when editing via keyboard

microbit-robert:
Attempting to use Blockly's `focusManager` to take ephemeral focus on an element inside the workspace focus tree causes numerous issues. Considering the unusual case of the `showLeds` block, it's better to workaround this problem by making use of the widget div when the led grid is selected for editing via the keyboard. 

This PR uses the widget div outside of the workspace focus tree and makes use of existing Blockly focus management that this div provides.

No changes are intended for mouse users and no visual changes are intended for keyboard users.

### DIFF
--- a/pxtblocks/fields/field_ledmatrix.ts
+++ b/pxtblocks/fields/field_ledmatrix.ts
@@ -98,13 +98,14 @@ export class FieldLedMatrix extends FieldMatrix implements FieldCustom {
     showEditor_() {
         this.selected = [0, 0];
 
+        const matrixRect = this.matrixSvg.getBoundingClientRect();
+
         const widgetDiv = Blockly.WidgetDiv.getDiv();
         widgetDiv.append(this.matrixSvg);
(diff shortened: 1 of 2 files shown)


pxtblocks/fields/field_ledmatrix.ts
```diff
@@ -44,6 +44,7 @@ export class FieldLedMatrix extends FieldMatrix implements FieldCustom {
     private currentDragState_: boolean;
 
     protected clearSelectionOnBlur = true;
+    protected forceFocusVisible = true;
```
microbit-robert, line 47: Safari seems to have a problem in being able to determine `focus-visible` correctly. Notably after zooming the workspace. We can ignore what the browser thinks for this field editor as it is either in mouse mode or keyboard mode. The mouse cannot open the widget div for this field editor, and the use of the mouse on the dropdown instantly closes it.

riknoll (approved):

PR #10621, opened 2025-05-31
Add Support for Region-Based Feature Flags

thsparks:
This adds support for feature flags that can be enabled (or disabled) for specific regions, based on the backend changes here: https://github.com/microsoft/pxt-backend/pull/1117

I've created a more generic "enabledFeatures" section of the AppTheme in pxtarget in an attempt to manage the proliferation of top-level feature switches. If no `includeRegions` or `excludeRegions` are specified, then the feature is enabled globally. Otherwise, it will only be enabled if the user's region matches one of the `includeRegions` and does not match any of the `excludeRegions`.

For development purposes, I've added a `DEV_REGION` constant that we can change similar to how we manage `DEV_BACKEND`, which allows us to fake being in a different region on local builds. (This also simplified things, since our localhost cdn logic is a little weird and didn't actually go to staging.) In my own testing, however, I did ensure it works with an actual staging CDN URL, too.

Upload Target: https://microbit.staging.pxt.io/app/f213b4bfa13132146caf582b2110dae5f61052d3-b71ebd062f

Fixes https://github.com/microsoft/pxt-microbit/issues/6238

### DIFF
--- a/ThirdPartyNotice
+++ b/ThirdPartyNotice
@@ -881,7 +881,7 @@ General Public License.
 863. dashjs 4.4.0 (https://www.npmjs.com/package/dashjs/v/4.4.0)
 864. @fortawesome/fontawesome-free 5.15.4 (https://www.npmjs.com/package/@fortawesome/fontawesome-free/v/5.15.4)
 865. @blockly/plugin-workspace-search 4.0.10 (https://www.npmjs.com/package/@blockly/plugin-workspace-search/v/4.0.10)
-866. @blockly/keyboard-experiment 0.0.7 (https://www.npmjs.com/package/@blockly/keyboard-experiment/v/0.0.7)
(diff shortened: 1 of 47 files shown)


pxtlib/emitter/cloud.ts
```diff
@@ -310,6 +311,38 @@ namespace pxt.Cloud {
...
+        if (region !== undefined || !pxt.webConfig?.cdnUrl) {
+            return;
+        }
+
+        const url = new URL("geo", pxt.webConfig.cdnUrl).toString();
+        const options: Util.HttpRequestOptions = { url };
+
+        try {
+            const response = await Util.requestAsync(options);
+            if (response.statusCode !== 200) {
```
srietkerk, line 329: Is it possible that the status code that we get from the backend could be something other than 200 on something that's successful?

thsparks: This is something we control, and right now, 200 is the only "success" response we'd expect.

srietkerk (approved):
LGTM!

jwunderl (approved):

PR #10630, opened 2025-06-05
configurable max file size for cli

jwunderl:
should be all we need to relax file size limitation for specific target


eanders-ms (approved):
👍

riknoll (approved):

PR #10642, opened 2025-06-11
Improve search category selected state

microbit-robert:
Only select / highlight the search category when it is actually selected by the user. This can happen if the user hits "Enter", "Tab", or "ArrowDown" from the search input. The previous behaviour resulted in users thinking they could use "ArrowRight" from the search input to move into the flyout and this PR aims to address that.

This PR also:
-  fixes a toolbox focus bug resulting from clicking on a toolbox category, then clicking on the sim, then re-focusing the toolbox again via keyboard and doing this more than once. This resulted in no category being selected by default.
- clicking on Monaco flyout entries no longer results in a focus event and a click event being handled. Only focus events are now handled and roving tab index is used to reduce tab stops and ensure that clicking on a flyout entry also focuses it.
- makes general changes to the Monaco toolbox and flyout to ensure feature and behaviour parity between blocks and text-based modes.

### DIFF
--- a/webapp/src/monaco.tsx
+++ b/webapp/src/monaco.tsx
@@ -1925,7 +1925,7 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         const flyoutFocused = e.relatedTarget === this.flyout.refs.flyout || (this.flyout.refs.flyout as HTMLElement).contains(e.relatedTarget);
         if (((searchInputFocused && !hasSearch) || !searchInputFocused) && !flyoutFocused) {
             this.hideFlyout();
-        };
+        }
         if (!flyoutFocused) {
             this.toolbox.clear();
             this.toolbox.clearExpandedItem();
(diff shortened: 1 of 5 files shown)


webapp/src/monaco.tsx
```diff
@@ -1916,11 +1916,22 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
             }
         }
     }
 
+    public onToolboxBlur(e: React.FocusEvent, hasSearch: boolean): void {
+        const searchInputFocused = e.relatedTarget === (this.toolbox.refs.searchbox as toolbox.ToolboxSearch).refs.searchInput;
+        const flyoutFocused = e.relatedTarget === this.flyout.refs.flyout || (this.flyout.refs.flyout as HTMLElement).contains(e.relatedTarget);
+        if (((searchInputFocused && !hasSearch) || !searchInputFocused) && !flyoutFocused) {
+            this.hideFlyout();
+        };
```
riknoll:
```suggestion
        }
```

aznhassan (approved):
:shipit:

PR #10645, opened 2025-06-11
Rephrase Error Help Button to Indicate AI Usage

thsparks:
There was some concern over the fact that the "Help me understand" button did not clearly indicate that it would send your code to AI. To address this, I've changed the wording on the button. I've also changed the icon the sparkle, which seems to be becoming a standard AI-indicator.

Sparkle icon came from fluent ui: https://github.com/microsoft/fluentui-system-icons/blob/main/assets/Sparkle/SVG/ic_fluent_sparkle_48_filled.svg

I'm going to say this fixes https://github.com/microsoft/pxt-microbit/issues/6224 unless we hear otherwise during privacy review. Also fixes https://github.com/microsoft/pxt-microbit/issues/6228.

The button now looks like this:
image


srietkerk (approved):
Nice!

PR #10646, opened 2025-06-12
special case for second main simulator

thomasjball:
when we remove old jacdac simulator, the simdriver still gets "jacdac" messages. It's default behavior when there is no match is to start a second main simulator (such as mbit). But this only happens (it appears) when messageChannel has boolean value false (the case for radio).

### DIFF
--- a/pxtsim/simdriver.ts
+++ b/pxtsim/simdriver.ts
@@ -452,7 +452,9 @@ namespace pxsim {
                             isDeferrableBroadcastMessage = true;
                             // start secondary frame if needed
                             const mkcdFrames = frames.filter(frame => !frame.dataset[FRAME_DATA_MESSAGE_CHANNEL]);
-                            if (mkcdFrames.length == 0 || mkcdFrames.length == 1 && !this.singleSimulator) {
+                            if (<any>messageChannel === false && (mkcdFrames.length == 0 || mkcdFrames.length == 1 && !this.singleSimulator)) {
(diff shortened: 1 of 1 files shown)


thomasjball:
So the default behavior in simdriver is that if a message is received with type != "messagepacket" (like "radiopacket" for example), then a new main simulator will be started. in this case the code

```
    const messageChannel = msg.type === "messagepacket" && (msg as SimulatorControlMessage).channel;
```
yields messageChannel === false

pxtsim/simdriver.ts
```diff
@@ -452,7 +452,12 @@ namespace pxsim {
                             isDeferrableBroadcastMessage = true;
                             // start secondary frame if needed
                             const mkcdFrames = frames.filter(frame => !frame.dataset[FRAME_DATA_MESSAGE_CHANNEL]);
-                            if (mkcdFrames.length == 0 || mkcdFrames.length == 1 && !this.singleSimulator) {
+                            if (<any>messageChannel === false &&
```
eanders-ms: If it isn't too broad, just check if it's falsey.
```suggestion
                            if (!messageChannel &&
```

eanders-ms: Alternatively:
```ts
    if ((messageChannel as any) === false &&
```
My aim here is to move away from using the legacy `<any>` cast, as it is invalid in .tsx files.

thomasjball: Ah got it

eanders-ms (approved):

PR #10648, opened 2025-06-12
adding new checkbox style and converting menu items to checkboxes

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/6297

this pr does two things:

1. adds a new toggle checkbox style to the react-common component and replaces the toggles in the pxtjson editor with the new implementation
2. changes the keyboard controls and greenscreen menu items to have a checkbox instead of the confusing off/on text

the reason these are both in the same pr is that i originally planned to use the toggle in the menu items but it ended up looking worse than just a checkbox. still, the change lets use remove a semantic ui component so i think it's good to check in anyways :)

here's what the toggle looks like:

toggle-checkbox

(note that the border is a little subtle right now. this is something that needs to be fixed in the micro:bit theme)

and here are the new menu items:

image

@microbit-matt-hillsdon FYI


riknoll:
also those toggles should _probably_ be radio buttons. i'll open an issue about it

PR #10653, opened 2025-06-16
don't allow old jacdac sim if new one is present

thomasjball:
Context:

- We will have both old and new jacdac sims alive for a while; 
- We don't want old jacdac sim showing when new jacdac sim already is up; 
- also don't want to change the logic for dealing with old message sim, since they are going away. 

Solution:
- A temporary fix for the explicit case we want to rule out that will become dead code and can be eliminated once old jacdac sim is removed from pxtarget.json

### DIFF
--- a/pxtsim/simdriver.ts
+++ b/pxtsim/simdriver.ts
@@ -352,6 +352,9 @@ namespace pxsim {
             this.singleSimulator = true
         }
 
+        // BEGIN TEMPORARY: jacdac simulator
+        newJacdacSimulator: boolean = false;
+        // END TEMPORARY: jacdac simulator
         public postMessage(msg: pxsim.SimulatorMessage, source?: Window, frameID?: string) {
             if (this.hwdbg) {
                 this.hwdbg.postMessage(msg)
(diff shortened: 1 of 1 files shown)


pxtsim/simdriver.ts
```diff
@@ -352,6 +352,9 @@ namespace pxsim {
             this.singleSimulator = true
         }
 
+        // BEGIN TEMPORARY: jacdac simulator
+        newJacdacSimulator: boolean = false;
```
eanders-ms, line 356: Will this flag reset when a new project is loaded? I think it will, but wanted to verify.

pxtsim/simdriver.ts
```diff
@@ -438,11 +444,13 @@ namespace pxsim {
                             let messageFrame = frames.find(frame => frame.dataset[FRAME_DATA_MESSAGE_CHANNEL] === messageChannel);
                             // not found, spin a new one
                             if (!messageFrame) {
-                                const useLocalHost = U.isLocalHost() && /localhostmessagesims=1/i.test(window.location.href)
-                                const url = ((useLocalHost && messageSimulator.localHostUrl) || messageSimulator.url)
-                                    .replace("$PARENT_ORIGIN$", encodeURIComponent(this.options.parentOrigin || ""))
-                                    .replace("$LANGUAGE$", encodeURIComponent(this.options.userLanguage))
-                                startSimulatorExtension(url, messageSimulator.permanent, messageSimulator.aspectRatio);
+                                if (messageChannel !== "jacdac" || !this.newJacdacSimulator) { // TEMPORARY: jacdac simulator
```
eanders-ms, line 447: Is this a race condition? What happens when a "jacdac" message is received before a "jacdac/pxt-jacdac" message?

thomasjball: the case we are going after is when the new sim is loaded first, so I don't think so.

thomasjball: this will be the case when we load the new jacdac extension. No race in this case.  Why not? Because

- the new extension sends the message "jacdac/pxt-jacdac" to start the new simulator; this happens before it sends any "jacdac" message.  That is, the window messages are totally ordered and handled sequentially (AFAIK)

I was concerned that jacdac message from connected hardware would interfere and potentially race with the simulator start, but I've tested and that doesn't seem to be the case.  The sim code gets first chance.

pxtsim/simdriver.ts
thomasjball: yes, I believe we get a new simdriver in this case.

thomasjball: checking...

eanders-ms (approved):

PR #10656, opened 2025-06-17
Fix region navigation in debug mode

microbit-matt-hillsdon:
Push focusToolbox implementations down to blocks/monaco editors as toolbox editor is unaware of the debug toolbox.

Add a focus API to the debug toolbox and use it in debug mode.

Closes https://github.com/microsoft/pxt-microbit/issues/6326


webapp/src/monaco.tsx
```diff
@@ -1377,6 +1377,14 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         ReactDOM.render(debuggerToolbox, container);
     }
 
+    focusToolbox(itemToFocus?: string): void {
```
microbit-matt-hillsdon, line 1380: To avoid the duplication we could explore a refactor where `ToolboxEditor` became aware of the idea of debug mode, at least sufficiently to implement `handleDebugToolboxRef` and `isDebugging`. But for the moment debuggerToolbox is a subclass field so the duplication is needed. Seems hard to tell whether it's worth it without trying.

PR #10657, opened 2025-06-17
fix python conversion of int and trunc

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/3851

this fixes the type inference of prefix unary operators in the python converter


srietkerk (approved):
I don't fully understand how the unify function fixes this, but the changes look straightforward enough.

riknoll:
@srietkerk unify basically takes the return type of an abstract syntax tree node and asserts that it must be compatible with whatever type is passed in. for an example, let's take this bit of python code:

```py
-1.5
```

the number -1.5 is actually two AST nodes: an outer node which is the - operator (called a prefix unary expression) and an inner node which is a numeric literal (1.5).

the output of the `-` prefix unary expression is always a number, so we can unify the outer node's return type with number. likewise, the only valid argument for a `-` prefix unary expression is a number, so we can unify that type as well.

the reason this fixes the linked bug is because of how we emit calls to `int()` when converting to typescript; the `int()` function converts to one of two different functions depending on the type of the argument passed:
* if the argument is a string, we convert it to `parseInt()`
* if the argument is a number, we convert it to `Math.trunc()`
 
the python converter was looking at the expression being passed to `int()` and because it wasn't unified with the number type, it just assumed it was a string even though it clearly isn't.

srietkerk:
Oh, cool! That was very helpful, thank you!

PR #10659, opened 2025-06-17
Poll for error help response

thsparks:
This switches the frontend code to account for changes in https://github.com/microsoft/pxt-backend/pull/1129. Basically, the frontend now needs to poll the backend for a response to its error-help request. I introduced a new, simplistic exponential backoff utility function to help with this so we don't spam ourselves (and deep prompt) with too many status requests.

Separately, this also fixes https://github.com/microsoft/pxt-microbit/issues/6336 by setting an error region when we can't detect the actual region.


pxtlib/emitter/cloud.ts
```diff
@@ -332,6 +332,7 @@ namespace pxt.Cloud {
             region = response.text.trim();
         } catch (e) {
             handleNetworkError(options, e);
+            region = "error"; // Set to a default value to differentiate from uninitialized
```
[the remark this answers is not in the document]
thsparks: I don't think this will significantly enhance clarity

webapp/src/cloud.ts
```diff
@@ -696,19 +696,36 @@ export async function aiErrorExplainRequest(
...
-            request.err || `Unable to reach AI. Error: ${request.statusCode}.\n${request.err}`
-        );
+
+    // Start the request.
+    const queryStart = await auth.apiAsync(startUrl, data, "POST");
+    const resultId = queryStart.resp.resultId;
+
+    // Poll until the request is complete (or timeout).
+    const result = await pxt.Util.runWithBackoffAsync<pxt.auth.ApiResult<any>>(
+        async () => {
```
srietkerk, line 716: Is the async keyword needed here? Does this basically tell `runWithBackoffAsync` that the function will definitely return a promise?

thsparks: Ah, you're right. It's not needed. I originally had this awaiting on the apiAsync call, then realized that wasn't needed but forgot to remove the async. I'll remove it now.

webapp/src/cloud.ts
```diff
@@ -696,19 +696,36 @@ export async function aiErrorExplainRequest(
...
-    const request = await auth.apiAsync(url, data, "POST");
-    if (!request.success) {
-        throw new StatusCodeError(
-            request.statusCode,
-            request.err || `Unable to reach AI. Error: ${request.statusCode}.\n${request.err}`
-        );
+
+    // Start the request.
+    const queryStart = await auth.apiAsync(startUrl, data, "POST");
+    const resultId = queryStart.resp.resultId;
```
srietkerk, line 712: Should there still be a check that we have the correct data here? If the start function errors, would that mean that the call to `apiAsync` results in a rejected promise? If so, would we be able to get `resultId`?

pxtlib/util.ts
```diff
@@ -1088,6 +1088,40 @@ namespace ts.pxtc.Util {
...
+        maxDelayMs: number,
+        timeoutMs: number,
+        timeoutErrorMsg: string,
+        backoffFactor: number = 1.5
+    ): Promise<T> {
+        let delay = initialDelayMs;
+        const startTime = Date.now();
+
+        while (Date.now() - startTime < timeoutMs) {
+            const result = await action();
```
eanders-ms, line 1114: What is the expected behavior of `action()` throws?

thsparks: This `runWithBackoffAsync` function doesn't do any special handling, so it just propagates up like usual. For error help, we catch and handle exceptions in the `getHelpAsync` function here: https://github.com/microsoft/pxt/blob/master/webapp/src/errorHelp.ts#L144

webapp/src/cloud.ts
thsparks: Good catch, we should still handle errors here like we did before. It got lost in the refactor.

webapp/src/cloud.ts
thsparks: Had to add it back after fixing https://github.com/microsoft/pxt/pull/10659/files#r2153322629 :)

eanders-ms (approved):

srietkerk (approved):
LGTM!

PR #10660, opened 2025-06-17
fix regex in diff format function

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/6180

we should probably port this one to arcade! i messed up a regex when copying/pasting some code during my refactor of the diff algorithm for project history


abchatra:
Lets port.

jwunderl (approved):

srietkerk (approved):

PR #10661, opened 2025-06-17
cleanup treerow styles and fix image icons in toolbox

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/6318

fixes the missing jacdac icon, which was caused by an invalid CSS variable assignment in the toolbox react code (you can't include the `!important` in css values). while i was at it, i also moved a lot of the styling from the tsx into the less files and got rid of the need for the `!important` tag


srietkerk (approved):
Nice cleanup!

PR #10665, opened 2025-06-18
add targetconfig flag for hiding extensions from search results

riknoll:
this is the feature work for https://github.com/microsoft/pxt-microbit/issues/6011

i'll open a separate pr on pxt-microbit to actually hide the extension.

this PR adds a new flag to the targetconfig that allows extensions to be hidden from search results. originally i wanted to use the "hidden" flag that we have in pxt.json, but we don't actually have access to that when doing a search since it would require us to clone the repo


srietkerk (approved):
LGTM

PR #10669, opened 2025-06-18
Move "Debug My Code Button" to problems header & show in blocks

thsparks:
This moves the "Debug My Code" button into the problem header, and it adds the button for blocks as well as text.

This button does not make sense when you have a compile error, since you can't debug a program that can't run. As such, the existing behavior checks for a stack trace before showing the button. I've preserved this in the monaco scenario, but in blocks, it's rare to get a compile error that actually stops the code from running, so it simply always shows the debug button for blocks (some of our errors do lack stack traces, otherwise it'd be a non-issue).

Fixes https://github.com/microsoft/pxt-microbit/issues/6256

Upload target: https://microbit.staging.pxt.io/app/63a4ae6953d2f301b03ef86dc02649b4a1229f87-d5deaa850b

image

### DIFF
--- a/webapp/src/blocks.tsx
+++ b/webapp/src/blocks.tsx
@@ -516,7 +516,12 @@ export class Editor extends toolboxeditor.ToolboxEditor {
     private initBlocklyToolbox() {
         // Remove unwanted additional tab stop from the editor.
         // We add tabindex to the tree wrapping the toolbox categories (excluding search) instead.
-        this.getToolboxDiv().removeAttribute("tabindex");
+        const toolboxDiv = this.getToolboxDiv();
+
+        // toolbox div might not exist if this is a read-only editor
+        if (toolboxDiv) {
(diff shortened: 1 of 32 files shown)


webapp/src/blocks.tsx
```diff
@@ -1010,7 +1016,8 @@ export class Editor extends toolboxeditor.ToolboxEditor {
                     errors={this.errors}
                     onSizeChange={this.onErrorListResize}
                     getErrorHelp={this.getErrorHelp}
-                    showLoginDialog={this.parent.showLoginDialog} />}
+                    showLoginDialog={this.parent.showLoginDialog}
+                    startDebugger={this.startDebugger} />}
```
srietkerk: nit: move the `/>}` to a new line so if there's anything else that needs to be added to the ErrorList the old last line doesn't have to be affected.

srietkerk (approved):
One nit, not blocking

PR #10676, opened 2025-06-19
fix history file size issues

riknoll:
this fixes an issue that was reported on the forum where a project's history file was growing huge and making the editor super slow and preventing sharing from working. the underlying issue was that `_history` was being added to the `files` list in `pxt.json`. i'm still not sure how that happened exactly, but this PR fixes that and also adds some fixes to control the file size.

changes include:
* prevents `_history` from being added to `pxt.json` (or at least prevents it on the code path that i think is most likely the culprit)
* prevents `_history` from being included in the result of `filesToBePublishedAsync`, which decides which files are included in downloaded and share projects
* prunes existing history snapshots to remove `_history` if it's present in case any projects ran into this bug
* adds some logic to collapse the diff entries stored in `_history` once per day. all entries that are older than one day are squashed in an interval of 2 hours

### DIFF
--- a/pxteditor/history.ts
+++ b/pxteditor/history.ts
@@ -47,110 +47,103 @@ export interface FileEditedChange {
 }
 
 export interface CollapseHistoryOptions {
+    // the interval in milliseconds at which to collapse history entries
     interval: number;
+
+    // the minimum time of entries that are subject to collapsing (inclusive)
     minTime?: number;
+
+    // the maximum time of entries that are subject to collapsing (inclusive)
     maxTime?: number;
 }
 
 // 5 minutes. This is overridden in pxtarget.json
 const DEFAULT_DIFF_HISTORY_INTERVAL = 1000 * 60 * 5;
 
-// 15 minutes. This is overridden in pxtarget.json
-const DEFAULT_SNAPSHOT_HISTORY_INTERVAL = 1000 * 60 * 15;
+// 30 minutes. This is overridden in pxtarget.json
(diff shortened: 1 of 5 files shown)


jwunderl (approved):

pxteditor/history.ts
```diff
@@ -243,11 +245,23 @@ export function createSnapshot(text: ScriptText) {
             }
         }
 
+        if (result[pxt.HISTORY_FILE]) {
```
thsparks: nit: I think you could reduce some duplication if you just deleted lines 248 - 253 and let it continue on to the identical code after the try catch (259-264).

thsparks (approved):
One nit pick, not blocking.

pxteditor/history.ts
riknoll: oh yeah that was my intention actually, but i guess i forgot to delete this one

pxteditor/history.ts
```diff
@@ -246,8 +248,15 @@ export function createSnapshot(text: ScriptText) {
         return result;
```
thsparks, line 248: Should this return also be removed?

abchatra:
Do we need to hotfix this?

riknoll:
yep

Jaqster:
More people complaining - https://forum.makecode.com/t/critical-bugs-is-makecode-being-maintained-by-microsoft/37349/10  Can we cherry-pick this fix to prod in the next week or two?

abchatra:
@riknoll lets do hotfix.

PR #10677, opened 2025-06-20
fix tab navigation in side docs

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/6330

this pr fixes a few side docs tab navigation bugs
* the "go back" button wasn't tab navigable. it now has a tabindex and was converted to an `<a>` element
* the pop-out button was at the end of the tab order in the side docs. it's now the first element in the order
* the rendered SVG blocks still had the tabindex attribute on some elements


srietkerk:
What pop-out button are you referring to? Is it this one?
image
It's first in the tab order for me in beta and live when I open the help docs. Or did adding the `a` tag mess with things?

riknoll:
@srietkerk no, the one that appears in the top-right of the side docs and opens the page in a new tab

PR #10678, opened 2025-06-20
volume for sound expressions in simulator

srietkerk:
fixes https://github.com/microsoft/pxt-microbit/issues/5193

This pull request introduces support for specifying volume in sound expressions. The changes primarily focus on updating the `playSoundExpressionAsync` and related methods to include a `volume` parameter, and ensuring compatibility with existing components that utilize these methods.

### Enhancements to sound expression functionality:

* [`pxtsim/sound/soundexpression.ts`](diffhunk://#diff-5d9c9da0fbb445ba7f98e30bd27c0758fc5c33e6fc9c75767ff28bc8c07a9bb9R211): Added a `volume` property to the `PendingSound` interface and updated the `__playSoundExpression` and `playSoundExpressionAsync` methods to accept a `volume` parameter. Adjusted the internal logic to pass the `volume` value through the sound playback pipeline. [[1]](diffhunk://#diff-5d9c9da0fbb445ba7f98e30bd27c0758fc5c33e6fc9c75767ff28bc8c07a9bb9R211) [[2]](diffhunk://#diff-5d9c9da0fbb445ba7f98e30bd27c0758fc5c33e6fc9c75767ff28bc8c07a9bb9L226-R235) [[3]](diffhunk://#diff-5d9c9da0fbb445ba7f98e30bd27c0758fc5c33e6fc9c75767ff28bc8c07a9bb9L259-R261) [[4]](diffhunk://#diff-5d9c9da0fbb445ba7f98e30bd27c0758fc5c33e6fc9c75767ff28bc8c07a9bb9L287-R289) [[5]](diffhunk://#diff-5d9c9da0fbb445ba7f98e30bd27c0758fc5c33e6fc9c75767ff28bc8c07a9bb9L313-R315)

### Updates to components using sound expressions:

* [`webapp/src/components/soundEffectEditor/SoundEffectEditor.tsx`](diffhunk://#diff-ec29734ba8329f00f0204db07007473617ccc2581629a9c43711d6192cfdbce3L78-R78): Updated the call to `playSoundExpressionAsync` to include a default volume value of `127`.
* [`webapp/src/components/soundEffectEditor/SoundGallery.tsx`](diffhunk://#diff-a36f1c3c3b44e8c6d007f302f8399dfcc52670d0dc4496c533be1db6e367208bL156-R156): Modified the call to `playSoundExpressionAsync` to include a default volume value of `127`.


pxtsim/sound/soundexpression.ts
```diff
@@ -284,7 +286,7 @@ namespace pxsim.codal.music {
         };
     }
 
-    export function playSoundExpressionAsync(notes: string, isCancelled?: () => boolean, onPull?: (freq: number, volume: number) => void) {
+    export function playSoundExpressionAsync(notes: string, volume?: number, isCancelled?: () => boolean, onPull?: (freq: number, volume: number) => void) {
```
thsparks: Do you know if this function is used by 3rd parties and/or user code?

I'm thinking we should probably add volume to the _end_ of the parameter list with a default value (probably the max?), so we don't break anyone who is already using this function.

webapp/src/components/soundEffectEditor/SoundEffectEditor.tsx
```diff
@@ -75,7 +75,7 @@ export const SoundEffectEditor = (props: SoundEffectEditorProps) => {
             await pxsim.AudioContextManager.playInstructionsAsync(pxt.assets.soundToInstructionBuffer(toPlay, 20, 1), isCancelled, onPull);
         }
         else {
-            await pxsim.codal.music.playSoundExpressionAsync(soundToCodalSound(toPlay).src, isCancelled, onPull);
+            await pxsim.codal.music.playSoundExpressionAsync(soundToCodalSound(toPlay).src, 127, isCancelled, onPull);
```
riknoll: i think this should probably be a value between 0 and 1, right? does the sound effect editor sound alright with a value of 127?

pxtsim/sound/soundexpression.ts
srietkerk: That's a good point.. The reason why I didn't put it at the end to begin with was because I didn't want to put arguments in for the functions, but I don't want this to be a breaking change, so I'll move it to the end.

webapp/src/components/soundEffectEditor/SoundEffectEditor.tsx
srietkerk: Oh, yeah. I think I put it to 127 because I saw the default for the volume is 127, but it does cause some distortion. I'll make these the 0.03 that it was before.

pxtsim/sound/soundexpression.ts
```diff
@@ -223,14 +224,15 @@ namespace pxsim.codal.music {
         return playing;
     }
 
-    export function __playSoundExpression(notes: string, waitTillDone: boolean): void {
+    export function __playSoundExpression(notes: string, waitTillDone: boolean, volume: number): void {
```
thsparks: Can volume be optional here, like it is in `playSoundExpressionAsync`? This will break builds of any targets that use this function. I know you fixed the microbit side. Is it used anywhere else? Maybe fine if we're confident micro:bit is the only target using this function.

pxtsim/sound/soundexpression.ts
```diff
@@ -310,7 +312,7 @@ namespace pxsim.codal.music {
                     arr[i] = ((buff[i] - 512) / 512);
                 }
                 return arr;
-            }, synth.sampleRate, 0.03, () => cancelled || (isCancelled && isCancelled()))
+            }, synth.sampleRate, volume, () => cancelled || (isCancelled && isCancelled()))
```
thsparks, line 315: Should we pass in 0.03 if volume is undefined?

thsparks: If you do this, you could avoid the changes in SoundEffectEditor and SoundGallery since it'd just be the default, but it doesn't hurt to have those explicitly set the volume if you prefer it that way.

pxtsim/sound/soundexpression.ts
srietkerk: Yeah, that's a good point. Based on searches, microbit _should_ be the only thing that uses this, but I'd rather avoid causing problems. I'll change this to be optional.

srietkerk: Or rather, make it have a default.

thsparks (approved):

PR #10681, opened 2025-06-20
convert image icons into a masked image in toolbox

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/6357

it's not easy to set the color of an SVG that's used for a background image to a specific color. you can make some complicated SVG filters to make it happen, but it won't be pretty.

for that reason, i've converted the image icons in the toolbox to use a mask instead of a background image. the color is now set using the background color and the mask prevents all the background pixels from showing except for those which are overlapped with non-transparent pixels in the icon source image.
jacdac-icon

### DIFF
--- a/pxteditor/history.ts
+++ b/pxteditor/history.ts
@@ -244,8 +244,6 @@ export function createSnapshot(text: ScriptText) {
                 result[file] = text[file];
             }
         }
-
-        return result;
     }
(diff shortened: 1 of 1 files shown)


eanders-ms (approved):
Nice.

srietkerk (approved):

PR #10684, opened 2025-06-23
include pxt and target versions to our base telemetry

srietkerk:
Closes https://github.com/microsoft/pxt-microbit/issues/5165


abchatra:
Nice. Can we add this logic to tracking.html for the telemetry in the docfiles as well? 
https://github.com/microsoft/pxt/blob/0374bf72a2b9c4b0c9b8e1e2217428dcace9b561/docfiles/tracking.html#L39

thsparks (approved):

thsparks (approved):

PR #10688, opened 2025-06-24
Add `buildcoredts` cli command

eanders-ms:
Integrated Joey's cli command to generate .d.ts files from target libs. I modified the typescript compiler to include block comments in the emitted files, which turned out to be trickier than expected, but I did get to know the typescript compiler a bit better along the way.

Usage:
```
Usage:
    pxt buildcoredts [--shareId shareid] [--combined] [--index]

build d.ts files for core packages

Flags:
    --shareId -s <value> ... Share ID to build for
    --combined -c .......... Build combined d.ts file for all core packages
    --index -i ............. Build index.d.ts file for all core packages
```


eanders-ms:
Output goes to `temp/dts` in target folder, which is already .gitignored.

PR #10691, opened 2025-06-24
Reimplement FieldColourNumber based on grid dropdown plugin

microbit-matt-hillsdon:
This allows the value to be the actual value rather than the CSS value which fixes initial selection issues.

Closes https://github.com/microsoft/pxt-microbit/issues/6384

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pxt-core",
-  "version": "11.4.31",
+  "version": "11.4.32",
   "description": "Microsoft MakeCode provides Blocks / JavaScript / Python tools and editors",
   "keywords": [
     "TypeScript",
--- a/pxtblocks/layout.ts
+++ b/pxtblocks/layout.ts
@@ -346,6 +346,10 @@ export function cleanUpBlocklySvg(svg: SVGElement): SVGElement {
     pxt.U.toArray(svg.querySelectorAll('.blocklyBlockCanvas,.blocklyBubbleCanvas'))
         .forEach(el => el.removeAttribute('transform'));
 
+    svg.querySelectorAll("[tabindex]").forEach(el => {
+        el.removeAttribute("tabindex");
+    });
+
     // In order to get the Blockly comment's text area to serialize properly they have to have names
     const parser = new DOMParser();
     pxt.U.toArray(svg.querySelectorAll('.blocklyTextarea'))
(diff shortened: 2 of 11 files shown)


pxtblocks/fields/field_colour.ts
```diff
@@ -137,22 +175,183 @@ export class FieldColorNumber extends FieldColour implements FieldCustom {
...
+            return null;
+        }
+        const cssFormat = valueFormatToCSSFormat(newValue, this.allColoursCSSFormat_);
+        const normalizedValueFormat = cssFormatToValueFormat(
+            cssFormat,
+            this.valueMode_,
+            this.allColoursCSSFormat_
+        );
+        return normalizedValueFormat;
+    }
```
microbit-matt-hillsdon, line 278: I think this maintains the previous behaviour which means supporting all the formats that `valueFormatToCSSFormat` supports. But I'm not sure why these are supported. There's lots of opportunity for simplification if it's only necessary to support the formats that this field editor can produce now (e.g. this validation can be removed in favour of super).

I'd welcome test scenarios for other value modes. Are there known extensions that use these that I can try?

pxtblocks/fields/field_colour.ts
```diff
@@ -20,79 +29,99 @@ export interface FieldColourNumberOptions extends FieldCustomOptions {
...
+        let titles: string[] | undefined;
+        const valueMode = params.valueMode ?? "rgb";
+        // TOOD: replace with a logical value and render checkerboard-style.
+        const transparentPlaceholderValue = "#dedede";
+        if (params.colours) {
+            allColoursCSSFormat = JSON.parse(params.colours);
+            // Assume the first colour represents transparent if it's present elsewhere.
+            if (allColoursCSSFormat.lastIndexOf(allColoursCSSFormat[0]) > 0) {
+                allColoursCSSFormat[0] = transparentPlaceholderValue;
+            }
```
microbit-matt-hillsdon, line 57: I added this which means we get the grey instead of the black twice. A checkerboard would be a good follow-up.

I wasn't sure when I could assume 0 represented transparency (always when in index mode?) so I went for a safe scenario.

pxtblocks/fields/field_colour.ts
riknoll: not that i'm aware... i think we must have had a reason for this at some point but i can't remember what it was now for the life of me

pxtblocks/fields/field_colour.ts
```diff
@@ -1,16 +1,25 @@
-/// <reference path="../../built/pxtlib.d.ts" />
+/**
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
```
riknoll:
```suggestion

```

PR #10692, opened 2025-06-24
Add AI Error Explainer Experiment

thsparks:
This change adds an experiment users can turn on, which will enable AI Error Help regardless of their geo region. This flag has no effect if the feature is already enabled for their geo.

image

Fixes https://github.com/microsoft/pxt-microbit/issues/6381


eanders-ms (approved):

srietkerk (approved):

pxteditor/experiments.ts
```diff
@@ -171,6 +171,11 @@ export function all(): Experiment[] {
             name: lf("Time Machine"),
             description: lf("Save and restore past versions of a project")
         },
+        {
+            id: "forceEnableAiErrorHelp",
+            name: lf("AI Error Explainer"),
+            description: lf("Get AI's help explaining errors in your code"),
```
abchatra, line 177: Can you create an issue for the feedback and add the following:        

    feedbackUrl: "https://github.com/microsoft/pxt/issues/xxxx",

thsparks: Sure. Created https://github.com/microsoft/pxt/issues/10694 and will add it in a separate PR, since this one is already merged.

PR #10698, opened 2025-06-25
fix exception when accessing bubble position before rendered

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/6391

i made the setBubbleVisible method async a while ago to fix an issue with comment rendering. as a result, it's possible for the bubble to not have been created when getBubbleLocation is called even though bubbleIsVisible returns true.


srietkerk (approved):

thsparks (approved):
Will this also fix https://github.com/microsoft/pxt-microbit/issues/6375?

riknoll:
@thsparks yup!

PR #10700, opened 2025-06-25
make inner outline black for focus on lighter buttons

srietkerk:
Fixes https://github.com/microsoft/pxt-microbit/issues/6116

### DIFF
--- a/theme/themes/pxt/modules/modal.overrides
+++ b/theme/themes/pxt/modules/modal.overrides
@@ -152,7 +152,7 @@
 }
 
 .ui.modal .ui.button.neutral:focus {
-    outline: 1px solid @black;
+    outline: 1px solid var(--pxt-target-foreground1);
     outline-offset: -7px;
 }
 
--- a/theme/themes/pxt/modules/modal.overrides
+++ b/theme/themes/pxt/modules/modal.overrides
@@ -152,7 +152,7 @@
 }
 
 .ui.modal .ui.button.neutral:focus {
-    outline: 1px solid var(--pxt-target-foreground1);
+    outline: 1px solid var(--pxt-neutral-foreground2);
     outline-offset: -7px;
 }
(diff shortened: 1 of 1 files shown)


theme/themes/pxt/modules/modal.overrides
```diff
@@ -151,6 +151,11 @@
     outline-offset: -7px;
 }
 
+.ui.modal .ui.button.neutral:focus {
+    outline: 1px solid @black;
```
thsparks: Did you check how this look in all themes? I have a hunch we'll need to use a theme variable here rather than hardcoding black.

srietkerk: Good catch, it is very much non-existent in the dark theme.

theme/themes/pxt/modules/modal.overrides
```diff
@@ -151,6 +151,11 @@
     outline-offset: -7px;
 }
 
+.ui.modal .ui.button.neutral:focus {
+    outline: 1px solid var(--pxt-target-foreground1);
```
thsparks: Typically "neutral" buttons have their background set to `--pxt-neutral-background2` (set [here](https://github.com/microsoft/pxt/blob/06349b7938e17309b9a4dff6d6226e21ae9fb8c6/theme/semantic-ui-overrides.less#L307)), and there's no guarantee that target-foreground1 will contrast with neutral-background2 in all themes. The _safest_ option for this value would actually be `--pxt-neutral-foreground2`, matching the same foreground to the background. How does that look? You could also try `--pxt-neutral-stencil2`, though I suspect that will be too light for this.

srietkerk: Thanks. Yeah, neutral-foreground2 works well!

thsparks (approved):

PR #10701, opened 2025-06-26
Hide Error List in Version History & Debug

thsparks:
This change does a few things.

1. First and foremost, it hides the error list when in the version history and debug views (fixes https://github.com/microsoft/pxt-microbit/issues/6377)
2. It also hides the "Explain with AI" and "Launch in Debugger" buttons when in readonly mode - for example, the share page. Both of these buttons can have unexpected behavior in those scenarios (like asking for log-in where it isn't supported, or getting stuck in debug mode)
3. Fixes unrelated focus issues and null references in readonly mode, caused by trying to manipulate the toolbox when it wasn't available.
4. Adds a feedback URL for the "AI Error Explainer" experiment (as requested in https://github.com/microsoft/pxt/pull/10692#discussion_r2164986818)

Upload target: https://makecode.microbit.org/app/358fa96db053a5c91fd5827a0cc871fbaf1d6be4-bf5998c599

### DIFF
--- a/webapp/src/blocks.tsx
+++ b/webapp/src/blocks.tsx
@@ -514,6 +514,10 @@ export class Editor extends toolboxeditor.ToolboxEditor {
     }
 
     private initBlocklyToolbox() {
+        if (pxt.shell.isTimeMachineEmbed()) {
+            return;
+        }
+
         // Remove unwanted additional tab stop from the editor.
         // We add tabindex to the tree wrapping the toolbox categories (excluding search) instead.
         const toolboxDiv = this.getToolboxDiv();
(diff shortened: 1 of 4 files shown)


webapp/src/blocks.tsx
```diff
@@ -514,6 +514,10 @@ export class Editor extends toolboxeditor.ToolboxEditor {
     }
 
     private initBlocklyToolbox() {
+        if (pxt.shell.isTimeMachineEmbed()) {
```
srietkerk: Should this check just be for read only mode or would that cause problems?

webapp/src/blocks.tsx
```diff
@@ -1005,7 +1009,7 @@ export class Editor extends toolboxeditor.ToolboxEditor {
 
     display(): JSX.Element {
         let flyoutOnly = this.parent.state.editorState && this.parent.state.editorState.hasCategories === false;
-        let showErrorList = pxt.Util.isFeatureEnabled("blocksErrorList");
+        let showErrorList = pxt.Util.isFeatureEnabled("blocksErrorList") && !pxt.shell.isTimeMachineEmbed() && !this.parent.state.debugging;
```
srietkerk, line 1012: I'm wondering if this should also just be for `isReadOnly`..  Although, I think it might be helpful for the code eval tool. I'm just not sure if it would be nice to have in our other read only modes.

image

thsparks: I considered that, but I actually thought it was potentially helpful in both eval mode (as you pointed out) and in the share page. It can be useful for navigating around to different parts of the code that have errors, and I could see scenarios where you'd want to do that even in readonly mode, so I was inclined to leave it.

webapp/src/blocks.tsx
thsparks: Yeah, I think you're right. Given that we don't render the toolbox in readonly mode, we should probably just always skip this too: https://github.com/microsoft/pxt/blob/3249b5247e45b70d34a86e6df5eb5618877f54ee/webapp/src/blocks.tsx#L1250

I'll update it.

PR #10708, opened 2025-06-26
clean up accessibility tick events and add missing ticks

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/6396

this pr makes a few tweaks to our accessibility tick events:
* combines all of the toggleAccessibleBlocks ticks into a single tick. the source of the tick (settings, homesettings, or accmenu) is now a custom dimension.
* adds tick events for showing/hiding the navigation regions
* adds tick events for showing the keyboard control sidedocs
* adds a custom dimensions to our `app.editor` tick which indicates what file type is loaded for the editor and whether accessible blocks are enabled

@abchatra @microbit-matt-hillsdon @microbit-lucy are there any other scenarios you'd like to collect data on? any keyboard shortcuts you want me to instrument?

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -2853,7 +2853,11 @@ export class ProjectView
     }
 
     private editorLoaded() {
-        pxt.tickEvent('app.editor', { projectHeaderId: this.state.header?.id });
+        pxt.tickEvent('app.editor', {
+            projectHeaderId: this.state.header?.id,
+            fileType: this.editorFile?.getExtension(),
+            accessibleBlocks: this.getData<boolean>(auth.ACCESSIBLE_BLOCKS) ? "true" : "false"
+        });
     }
 
     unloadProjectAsync(home?: boolean) {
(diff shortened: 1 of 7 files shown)


webapp/src/app.tsx
```diff
@@ -2853,7 +2853,11 @@ export class ProjectView
     }
 
     private editorLoaded() {
-        pxt.tickEvent('app.editor', { projectHeaderId: this.state.header?.id });
+        pxt.tickEvent('app.editor', {
+            projectHeaderId: this.state.header?.id,
+            fileType: this.editorFile?.getExtension(),
+            accessibleBlocks: this.getData<boolean>(auth.ACCESSIBLE_BLOCKS) ? "true" : "false"
```
abchatra: Do we need to send accessible Blocks or not on every new project creation?

abchatra: Rest looks good.

riknoll: this is not just every project creation, but also every time a file is loaded in the editor (so this will also fire when switching between projects). the point is to count sessions where accessible blocks are enabled and a blocks project is opened.

i can add a new tick for this instead if that's preferable.

e.g. something that fires at the beginning of a session only. i guess that would just make the query slightly more annoying since you'll have to join correlate the session id and `app.editor` events

abchatra: I think new tick is preferable, it seems weird to add custom dimensions to other tick.

riknoll: @abchatra gotcha, i'm going to leave the file extension dimension here though since this seems like the best place for it

jwunderl (approved):

webapp/src/app.tsx
riknoll: @abchatra moved that custom dimension out of this tick and added a new `accessibilty.accessibleBlocksEnabledForSession` tick that fires on page load with a custom dimension indicating if the setting is true or not

PR #10715, opened 2025-06-27
Fix "on start" block in stack traces

thsparks:
Fixes https://github.com/microsoft/pxt-microbit/issues/6397

The JS code for anything in "on start" is just at the root level of the javascript, which leads to some confusion when we parse the locInfo associated with an "on start" stack frame.

In these cases, we can detect we're at the root level (therefore in "on start") by checking if the function name is the root function defined here: https://github.com/microsoft/pxt/blob/d15df7021d902c7e41f54a3e3d51c604b3097010/pxtcompiler/emitter/emitter.ts#L1056

# Before
image
image

# After
image
image

# Try it
Upload Target: https://makecode.microbit.org/app/9d047dac5da6b08a6c61a5d74a74ac4f59d78e98-b603a45996

### DIFF
--- a/webapp/src/blocks.tsx
+++ b/webapp/src/blocks.tsx
@@ -1089,7 +1089,10 @@ export class Editor extends toolboxeditor.ToolboxEditor {
                 return undefined;
             }
 
-            const blockId = this.compilationResult ? pxtblockly.findBlockIdByLine(this.compilationResult.sourceMap, { start: locInfo.line, length: locInfo.endLine - locInfo.line }) : undefined;
+            const blockId = locInfo.functionName === "<main>" /* Root JS function, therefore On Start block */ ?
+                this.editor.getBlocksByType(ts.pxtc.ON_START_TYPE)?.[0]?.id :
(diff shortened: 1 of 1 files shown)


webapp/src/blocks.tsx
```diff
@@ -1089,7 +1089,10 @@ export class Editor extends toolboxeditor.ToolboxEditor {
                 return undefined;
             }
 
-            const blockId = this.compilationResult ? pxtblockly.findBlockIdByLine(this.compilationResult.sourceMap, { start: locInfo.line, length: locInfo.endLine - locInfo.line }) : undefined;
+            const blockId = locInfo.functionName === "<main>" /* Root JS function, therefore On Start block */ ?
```
srietkerk, line 1092: Would it be possible for a user to name their function "main" and cause problems here?

thsparks: Thankfully this isn't an issue, since a function named "main" will just have "main" as its function name, not "\<main\>". And since `<` and `>` are not allowed in js function names, a function named "\<main\>" just becomes "main" in the loc info, so still no issue there.

Tested as well, to confirm:
image

PR #10719, opened 2025-06-30
High Contrast Fixes

thsparks:
This primarily updates areas in pxt where we were checking the old true/false high contrast setting (which is no longer used) instead of the using the new theme manager. It was just an oversight in the theming work. I've created a little static helper function in the theme manager to save some repetitive lines of code.

Separately, it fixes a few places where we were seeing superfluous yellow outlines in hc mode. We have some generic highlighting css logic which, in general, is convenient. But it does seem to be causing some issues too, highlighting things we don't want highlighted. I've special-cased those out for now, but if it becomes a more common occurrence, we may need to rework it more completely.

This fixes one of the issues brought up in https://github.com/microsoft/pxt-microbit/issues/6366 (inconsistent settings), but it does not remove the white background (which was by design). We can decide what we want to do for that separately and follow up in a different change, if needed.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -3904,7 +3904,7 @@ export class ProjectView
         const autoRun = this.state.autoRun && !clickTrigger; // if user pressed stop, don't restart
 
         // Only fire setState if something changed
-        const setStatePromise = this.state.simState !== SimState.Stopped || !!this.state.autoRun !== !!autoRun ? 
+        const setStatePromise = this.state.simState !== SimState.Stopped || !!this.state.autoRun !== !!autoRun ?
             this.setStateAsync({ simState: SimState.Stopped, autoRun: autoRun }) :
             Promise.resolve();
 
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -171,13 +171,6 @@ export class ProjectView
 
     private themeManager: ThemeManager;
 
-    private highContrastSubscriber: data.DataSubscriber = {
-        subscriptions: [],
-        onDataChanged: () => {
-            this.onHighContrastChanged();
-        }
-    };
-
     private cloudStatusSubscriber: data.DataSubscriber = {
         subscriptions: [],
         onDataChanged: (path) => this.onCloudStatusChanged(path)
@@ -197,9 +190,7 @@ export class ProjectView
         this.reload = false; //set to true in case of reset of the project where we are going to reload the page.
         this.settings = JSON.parse(pxt.storage.getLocal("editorSettings") || "{}")
         const shouldShowHomeScreen = this.shouldShowHomeScreen();
-        const isHighContrast = /hc=(\w+)/.test(window.location.href) || (window.matchMedia?.('(forced-colors: active)')?.matches);
         this.themeManager = ThemeManager.getInstance(document);
-        if (isHighContrast) core.setHighContrast(true);
 
         const simcfg = pxt.appTarget.simulator;
         this.state = {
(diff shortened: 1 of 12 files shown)


webapp/src/app.tsx
```diff
@@ -1275,8 +1270,8 @@ export class ProjectView
                 return previousEditor ? previousEditor.unloadFileAsync() : Promise.resolve();
             })
             .then(() => {
-                let hc = this.getData<boolean>(auth.HIGHCONTRAST)
-                return this.editor.loadFileAsync(this.editorFile, hc)
+                let hc = this.themeManager.isHighContrast(this.themeManager.getCurrentColorTheme()?.id);
```
srietkerk, line 1273: Why aren't these instances using `isCurrentThemeHighContrast`?

webapp/src/app.tsx
```diff
@@ -3892,7 +3887,7 @@ export class ProjectView
         return (!!debugging != simulator.driver.isDebug()) || (!!tracing != simulator.driver.isTracing())
     }
 
-    stopSimulator(unload?: boolean, opts?: pxt.editor.SimulatorStartOptions) {
+    stopSimulator(unload?: boolean, opts?: pxt.editor.SimulatorStartOptions): Promise<void> {
```
srietkerk, line 3890: If we change this to a promise, do we need to await this being called in other areas? I see you changed it in `onThemeChanged`, but are the other places it's called okay with not await-ing it?

webapp/src/container.tsx
```diff
@@ -322,7 +323,7 @@ export class SettingsMenu extends data.Component<SettingsMenuProps, SettingsMenu
 
     renderCore() {
         const hasIdentity = pxt.auth.hasIdentity();
-        const highContrast = this.getData<boolean>(auth.HIGHCONTRAST)
+        const highContrast = ThemeManager.isCurrentThemeHighContrast(document);
```
srietkerk: Why does the container need the document passed in while the other instances are okay with the default document?

webapp/src/toolbox.tsx
```diff
@@ -632,6 +632,7 @@ export class Toolbox extends data.Component<ToolboxProps, ToolboxState> {
                 }
                 <div className="blocklyTreeRoot">
                     <div
+                        className="blocklyTreeInner"
```
srietkerk, line 635: I see a lot of changes in here for the keyboard nav stuff, will adding this classname back in possibly mess with that?

webapp/src/app.tsx
thsparks: Just because we already had the theme manager in this file, this is slightly more efficient (if uglier)

webapp/src/app.tsx
thsparks: It should be okay in other areas. I only added the return because of the specific off-then-on again scenario, where I need to actually wait until it is *for-sure* done setting sim state to "off" before trying to turn it on again. TBH I don't know how it was working without this, but I was able to see the race fairly clearly in my debugging.

In other places, they weren't waiting for that state to finish updating before this change, and they still aren't waiting for it now, so it shouldn't result in any change in behavior (and I didn't want to expand the scope of the change by trying to update them)

webapp/src/toolbox.tsx
thsparks: Shouldn't cause any problems, since it's a new class and only being used in the one place I added it.

webapp/src/container.tsx
thsparks: Oh, it actually isn't necessary. This is the same as the default. I can remove it.

PR #10720, opened 2025-06-30
audio sim: lower the volume to be closer to that of the other audio blocks

srietkerk:
Fixes https://github.com/microsoft/pxt-microbit/issues/6408

To test: https://makecode.microbit.org/app/39186a02f452020872fe9c95c14f43a1cefb8b66-476243682f#

### DIFF
--- a/pxtsim/simlib.ts
+++ b/pxtsim/simlib.ts
@@ -933,6 +933,8 @@ namespace pxsim {
             channel.generator = distortion;
             channel.generator.connect(channel.gain);
             source.connect(distortion);
+            // scaling the volume to be a multiplier of 0.1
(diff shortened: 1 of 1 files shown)


pxtsim/simlib.ts
```diff
@@ -933,7 +933,7 @@ namespace pxsim {
             channel.generator = distortion;
             channel.generator.connect(channel.gain);
             source.connect(distortion);
-            channel.gain.gain.value = volume;
+            channel.gain.gain.value = volume * 0.1;
```
thsparks, line 938: Would recommend a comment here explaining why we're multiplying by 0.1.

thsparks (approved):
One nitpick, but LGTM otherwise!

thsparks (approved):

PR #10726, opened 2025-07-01
Refactor MenuDropdown and replace sui.DropdownMenu in settings and docs

riknoll:
this refactors the react-common MenuDropdown quite a bit and replaces the sui.DropdownMenu in the (?) and cog wheel icons in the top bar.

noticeable changes include:
* pressing tab while the dropdown pane is open now closes the pane and navigates to the next element in the page. previously tab could be used to move to the next menuitem in the dropdown pane in addition to the arrow keys, but the standard behavior is to only allow navigation between menuitems with the arrow keys
* all elements in the dropdown pane are now replaced with the semantically correct HTML elements:
  * the outer pane/focus trap is now a `<ul role="menu">`
  * each subgroup of items is a `<li role="none"><ul role="group">`
  * external link menuitems are now `<a role="menuitem">`
  * checkbox menuitems are `<li role="menuitemcheckbox">`
  * all other menuitems are `<li role="menuitem">`
* the settings/docs menus now scroll instead of overflowing the page
* focus styling for elements in the dropdown pane is now more consistent with our other elements (uses the standard 3 pixel focus border)

i referred heavily to the w3c [navigation menubar](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-navigation/) and [editor menubar](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-editor/) examples to make sure i was matching their behavior/structure

this change also involved a lot of tweaks to the `FocusTrap` component, some of which make the `FocusList` component obsolete. after this is merged i think i'll rename `FocusTrap` to `FocusManager` (or something like that), get rid of `FocusList` and clean up some of the options to make them easier to understand

since this PR is pretty big, we might not want to take it for this release because we're so late in the current release cycle. that being said, we might be able to hotfix it later

here is an upload target: https://makecode.microbit.org/app/b593582fb02e4fb5b20cb4afd83ec59549642d1b-9835ad0d78

@microbit-matt-hillsdon FYI

### DIFF
--- a/react-common/styles/controls/MenuDropdown.less
+++ b/react-common/styles/controls/MenuDropdown.less
@@ -11,7 +11,7 @@
     position: absolute;
     right: 0;
     padding: 0;
-    min-width: 12rem;
+    min-width: 8rem;
     list-style: none;
     margin: 0;
     background: var(--pxt-neutral-background1);


riknoll:
making draft until we have a stable branch in pxt to work off of

abchatra:
@riknoll is this ready for Minecraft ?

riknoll:
yes, once i have a sec to resolve merge conflicts

riknoll:
@microsoft/makecode-engineers ready for review!

eanders-ms (approved):
Looks good. container.tsx is much more readable now.

PR #10727, opened 2025-07-02
fix duplicate on drag with the start_move cut shortcuts

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/6414

this overrides the start_move and cut shortcuts so that they properly handle duplicate on drag blocks.

for the start_move shortcut, the block is cloned and the cloned block is moved instead of the original:

duplicate-on-keyboard

for the cut shortcut, it works exactly the same except the block isn't deleted, which is what you would expect for these blocks

@microbit-matt-hillsdon I would appreciate if you could review this!


jwunderl (approved):

microbit-matt-hillsdon:
@riknoll 

Problems:
1. The duplicate made for move mode, cut or even copy still has the duplicate on drag behaviour, so once moved you can inappropriately make more duplicates from the new block. This works correctly for mouse drag.
    - This is an issue on live too for the copy shortcut but a keyboard user is more likely to encounter this.
1. Undo grouping: If you move, confirm, undo you get an extra step where the new block was added. I think the move action will use an existing group if there is one so setting `Events.setGroup(true)` before calling the original callback in the move override will likely fix this.
1. There is a context menu entry for Move that needs the same change.

Potential improvements. I think these can be revisited in future and might not be viable to fix just in MakeCode:
1. Escape to cancel move mode would ideally remove the new block. The mover code in the plugin normally causes endDrag to clean it up (via wouldDeleteDraggable) as it knows whether you're inserting or moving. Behaviourally this is more like an insert than a move. I don't see how you can communicate this to the plugin though.
1. The first suggested move location isn't as good as it would be if you inserted an equivalently shaped value block, because the duplicating has already placed the block on the workspace whereas an insert would suggest a connection. I think we can live with this. Fix is likely related to the previous issue.
1. Delete shouldn't be allowed?
    - At least for radio block arguments, function arguments which can't meaningfully be replaced.
    - Loop variables is unclear to me, you can e.g. replace them as they're just variables, but codegen doesn't work in the interim state "Sorry, we cannot convert this program". Separate issue?

riknoll:
@microbit-matt-hillsdon updated pr

changes include:
* got rid of all the mutation stuff with duplicate on drag so that the duplicate on drag state no longer gets persisted when copy/pasted. this was all code from an earlier implementation anyhow, it wasn't actually being used
* patched the context menu entry for move
* set the event group as suggested
* all duplicate on drag blocks (including loop variables) are no longer deletable until they are dragged/moved out. i had to do a bit of a hack to get this behavior to work while still supporting cut/copy but it isn't that bad

here's an upload target for easy testing: https://makecode.microbit.org/app/994d2d3b4b6e47992c49e61d9ceaf05245673573-82d2cf2aea

microbit-matt-hillsdon:
LGTM, thanks to @microbit-grace too for testing.

PR #10729, opened 2025-07-02
More High Contrast Fixes

thsparks:
Fixes https://github.com/microsoft/pxt-microbit/issues/6420 (just did a targeted fix to reduce risk this close to release)
Fixes https://github.com/microsoft/pxt-microbit/issues/6426 by updating blue color for high contrast

No longer inverted colors in "move mode" section:
image

Better intellisense contrast:
image


srietkerk (approved):
I'm guessing it would be a decent amount of work, but the sidedocs change makes me wonder if we should pursue supporting themes in our docs eventually.

thsparks:
> I'm guessing it would be a decent amount of work, but the sidedocs change makes me wonder if we should pursue supporting themes in our docs eventually.

Yeah, might be something to consider for the future, but definitely a significant chunk of work.

jwunderl (approved):

PR #10752, opened 2025-07-10
Fix sidedocs + experiments banner layout

thsparks:
Fixes https://github.com/microsoft/pxt-microbit/issues/6450

The bug is caused by a few overlapping css rules, which only happen when the notification banner is visible and the sidedocs are present but collapsed. Specifically, the bug happens when:

1. Sidedocs `top` is set to some non-zero value
   - This is true when sidedocs are expanded *or* when the experiments banner (i.e. notification banner) is present
2. Sidedocs `height` is set to `100%`
   - This is true when sidedocs are collapsed (the rule gets overridden when expanded)
3. Sidedocs `position` set to `absolute` instead of `fixed`
   - This is true when sidedocs are collapsed (the rule gets overridden when expanded)

The above three conditions are only **all true** when the notification banner is present (making 1 true) and the docs are collapsed (making 2 and 3 true), and this is what triggers the bug. Basically: the sidedocs have `height: 100%` making them as tall as their containing parent, but they're also pushed down a few rem by the `top` setting, so they're actually extending below the containing parent, which causes them to overflow beyond the bottom of the page. When this happens *and* `position` is `absolute` instead of `fixed`, it causes a scrollbar to appear on the entire page (which auto-scrolls to the bottom), pushing the header bar and banner out of view. This is what we see in the bug (notice the scrollbar positioning on the right-hand side when docs are collapsed). You can't actually see the docs at this point because the element has `visibility: hidden`, but it still takes up space, causing the layout issue.

To fix it, I've removed the `height: %100` as I don't think we need it. I've also increased the specificity of the top-setting css, which ensures it only gets applied when the docs are expanded (and fixes an unrelated bug, where this was being overridden by the default top-setting code [here](https://github.com/microsoft/pxt/blob/master/theme/sidedoc.less#L50), causing it to be partially obscured by the header bar when the notification banner was visible)

Tangentially related: I've explicitly set the notification banner `width` to 100%. For reasons I don't fully understand, when this was unset, it would overflow _very slightly_ and the browser would try to show a horizontal scrollbar (which gets hidden by the editor tools). This threw off the layout and caused things to shift upwards about ~1rem when the notification banner was present. Setting the width prevents this.

Try it here: https://makecode.microbit.org/app/8cbb95a95f6790e0baf0c2a4c3eb8d788e5b3266-669c83a901


abchatra:
Please port to [stable12.0](https://github.com/microsoft/pxt/tree/stable12.0) after merge

jwunderl (approved):
lgtm

PR #10755, opened 2025-07-10
Area overlay fixes

microbit-matt-hillsdon:
- Ensure keydown listener cannot be registered multiple times in the sim. This fixes bad behaviour due to multiple listeners after sim re-run e.g. due to program changes, which combines unfortunately with the toggle behaviour for the area menu. 

- Correctly close the full screen sim before showing the area menu, switching to a simpler approach.

- Fix bail out logic as it logged a React warning. Perform the state change in an effect. So far as I know this can't be triggered as a user, but I tested it locally by removing "editor" from the area rects.

Fixes https://github.com/microsoft/pxt-microbit/issues/6462

### DIFF
--- a/pxtsim/accessibility.ts
+++ b/pxtsim/accessibility.ts
@@ -2,6 +2,7 @@
 // Helpers designed to help to make a simulator accessible.
 namespace pxsim.accessibility {
     let liveRegion: HTMLDivElement;
+    let keydownListenerAdded = false;
 
     export function makeFocusable(elem: SVGElement): void {
         elem.setAttribute("focusable", "true");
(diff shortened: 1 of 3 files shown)


microbit-matt-hillsdon:
Please can we get this one in the release?

We'll spend time today testing it to build more confidence before the start of the US day.

Deployment: https://area-overlay-fixes.review-pxt.pages.dev/#editor

microbit-robert:
I've tested this on the deployment linked above and can no longer repro the bug. The area overlay is now correctly shown from the simulator at all times, including when it is fullscreen.

microbit-matt-hillsdon:
@riknoll @abchatra tagging for visibility.

riknoll (approved):

PR #10775, opened 2025-07-17
Prevent the LED matrix from firing no-op change events

glastonbridge:
Fixes https://github.com/microsoft/pxt-microbit/issues/6248

The LED matrix was firing multiple redundant change events when the user interacted with it. Particularly bad was when the mousemove event fired multiple times over the same LED, triggering change events. This fix tests for an actual change in state and guards against firing additional events.

**Before**

https://github.com/user-attachments/assets/25310d7a-45b0-4444-b277-cec863df52ba

**After**

https://github.com/user-attachments/assets/580a9102-19ee-4295-95d3-78fdb91222df


microbit-matt-hillsdon:
This is just tidying up of a pre-existing bug in v7 that we spotted along the way. No strong view as to whether we need this in live too.

PR #10791, opened 2025-07-22
add dropdown options to block search fields

jwunderl:
image
When searching take into account dropdowns for search results (& also parameter help strings, that was implemented but just forgot return looks like)

left extracting the options to copilot & it did a decent job

### DIFF
--- a/pxtcompiler/emitter/service.ts
+++ b/pxtcompiler/emitter/service.ts
@@ -1133,9 +1133,51 @@ namespace ts.pxtc.service {
             const computeParameterString = (symbol: SymbolInfo): string => {
                 const paramHelp = symbol.attributes?.paramHelp;
                 if (paramHelp) {
-                    Object.keys(paramHelp).map(p => paramHelp[p]).join(" ");
+                    return Object.keys(paramHelp).map(p => paramHelp[p]).join(" ");
                 }
                 return "";
+            }
+
+            // Extract dropdown options from parameters for search indexing
+            const computeDropdownOptionsString = (symbol: SymbolInfo): string => {
+                let options: string[] = [];
+
+                if (symbol.parameters) {
+                    symbol.parameters.forEach(param => {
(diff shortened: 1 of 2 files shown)


pxtcompiler/emitter/service.ts
```diff
@@ -1133,11 +1133,53 @@ namespace ts.pxtc.service {
...
                 const paramHelp = symbol.attributes?.paramHelp;
                 if (paramHelp) {
-                    Object.keys(paramHelp).map(p => paramHelp[p]).join(" ");
+                    return Object.keys(paramHelp).map(p => paramHelp[p]).join(" ");
                 }
                 return "";
             }
 
+            // Extract dropdown options from parameters for search indexing
+            const computeDropdownOptionsString = (symbol: SymbolInfo): string => {
```
riknoll, line 1142: hmmm... this might work for enums but what about the other things that get turned into a dropdown like fixed instances or sprite properties?

the code for that is here btw:

https://github.com/microsoft/pxt/blob/a0a5464b07c5843c21be0c65385eaae3351d8c1f/pxtblocks/loader.ts#L431-L442

jwunderl: Think i've cover all these now, tested with image icons in microbit for alt text (I believe?), acceleration / vx, pins in microbit, etc. Last case i can think of right away would be kinds / spritekinds, but i would suspect that's a bit more complicated since IIRC it's handled in `field_kind.ts` / wouldn't be part of api info as it'd have to pull from projects variables (will check though)

jwunderl:
Also any gut feeling on changing result limit in search, it's just always been 7 arbitrarily but with fuzziness inherent to it that's a bit small

https://github.com/microsoft/pxt/blob/2165f7ed1301a2777bb3f0ec24eea131718fe2b2/pxtcompiler/emitter/service.ts#L1050

PR #10803, opened 2025-07-25
add the save only flag back to filedownload

srietkerk:
Fixes https://github.com/microsoft/pxt-microbit/issues/6545

The behavior was the same in arcade with or without the true flag. Wanted to clarify since the change was made (and then reverted incompletely) for arcade originally.


microbit-matt-hillsdon:
Thanks! I confirmed via controller.html that this reinstates the "save" message. I've not otherwise reviewed this.

abchatra (approved):

abchatra:
Port to stable12.0

PR #10804, opened 2025-07-25
add arm builds below existing builds

jwunderl:
image

Also threw in a windows arm build button but currently hidden -- i have a pr up to generate them but don't think we have a machine to test.

re: https://github.com/microsoft/pxt-microbit/issues/6543


jwunderl:
image
added a recommended label too

srietkerk (approved):
Nice!

abchatra:
Awesome! Lets put windows ARM too. I think we can test on a VM.

jwunderl:
since these are in docfiles i'll need to port to release branch, no worries on merging this now even when builds not prepped quite yet

PR #10805, opened 2025-07-25
PXT High Contrast Fixes

thsparks:
1. Fix the search toolbox category colors in high contrast (fixes https://github.com/microsoft/pxt-minecraft/issues/2760)
2. Fix the editor toggle buttons in high contrast (in conjunction with https://github.com/microsoft/pxt-minecraft/pull/2788)
3. Fix the "user provided content" header in high contrast (fixes https://github.com/microsoft/pxt-minecraft/issues/2762)

### DIFF
--- a/theme/toolbox.less
+++ b/theme/toolbox.less
@@ -304,7 +304,7 @@ div.blocklyTreeIcon span {
         color: var(--pxt-primary-foreground);
     }
 
-    div.blocklyTreeRow .blocklyTreeIcon {
+    div.blocklyTreeRow .blocklyTreeIcon:not(.blocklyTreeIconsearch) {
         color: var(--block-meta-color);
     }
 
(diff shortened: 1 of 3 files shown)


theme/toolbox.less
```diff
@@ -304,7 +304,7 @@ div.blocklyTreeIcon span {
         color: var(--pxt-primary-foreground);
     }
 
-    div.blocklyTreeRow .blocklyTreeIcon {
```
thsparks, line 307: Ah, need to find a different fix for this, as it breaks non-hc modes.

theme/toolbox.less
```diff
@@ -304,7 +304,7 @@ div.blocklyTreeIcon span {
...
+    div.blocklyTreeRow .blocklyTreeIcon:not(.blocklyTreeIconsearch) {
```
thsparks, line 308: Ah, need to find a different fix for this, as it breaks non-hc modes.

thsparks: Actually this is fine, it's inside the #root.hc, but there was a different bug for the search icon.

srietkerk (approved):

PR #10816, opened 2025-08-01
rephrase the reset dialog for minecraft, say delete instead of clear for other targets

srietkerk:
Fixes https://github.com/microsoft/pxt-minecraft/issues/2781

In Minecraft:
image

In microbit:
image

### DIFF
--- a/webapp/src/dialogs.tsx
+++ b/webapp/src/dialogs.tsx
@@ -717,9 +717,13 @@ export function showReportAbuseAsync(pubId?: string) {
 }
 
 export function showResetDialogAsync() {
+    const isMinecraft = pxt.appTarget.id === "minecraft";
+    const dialogBody = isMinecraft
+        ? lf("You are about to delete all MakeCode projects from Minecraft. Are you sure? This operation cannot be undone.")
+        : lf("You are about to delete all projects. Are you sure? This operation cannot be undone.");
     return core.confirmAsync({
         header: lf("Reset"),
-        body: lf("You are about to clear all projects. Are you sure? This operation cannot be undone."),
+        body: dialogBody,
         agreeLbl: lf("Reset"),
         agreeClass: "red",
         agreeIcon: "sign out",


thsparks (approved):
If we end up needing to customize this for even more targets, we might consider moving it to pxtarget or targetconfig or something. But I'd say this is fine for now.

Jaqster:
And this only pertains to local projects, right?  So, I don't know if we want to do it in this dialog, but maybe we should point to some documentation about what Reset actually does (i.e. it doesn't delete projects in the cloud)

abchatra:
I would suggest change this dialog to x at the top right, remove cancel and add a help button to point to documentation. 
Personally, I don't like target checks in PXT. These error messages need localization, so moving to targetconfig/pxttarget is hard. So maybe just say "You are about to delete all MakeCode projects. Are you sure? This operation can't be undone." This would serve all targets.

srietkerk:
> I would suggest change this dialog to x at the top right, remove cancel and add a help button to point to documentation. Personally, I don't like target checks in PXT. These error messages need localization, so moving to targetconfig/pxttarget is hard. So maybe just say "You are about to delete all MakeCode projects. Are you sure? This operation can't be undone." This would serve all targets.

Sure, I can change the modal to add a link to documentation, but I do think cancel is important for a modal like this. Cancel should be a very obvious action so we don't worry users. But I don't think adding "MakeCode" to the dialog without added context is a safe change. I don't know how many of our Minecraft users also use our other targets, but if I saw something that said "you are about to delete all MakeCode projects", I would be worried that it means that my arcade and microbit projects would also be deleted.

thsparks:
> > I would suggest change this dialog to x at the top right, remove cancel and add a help button to point to documentation. Personally, I don't like target checks in PXT. These error messages need localization, so moving to targetconfig/pxttarget is hard. So maybe just say "You are about to delete all MakeCode projects. Are you sure? This operation can't be undone." This would serve all targets.
> 
> Sure, I can change the modal to add a link to documentation, but I do think cancel is important for a modal like this. Cancel should be a very obvious action so we don't worry users. But I don't think adding "MakeCode" to the dialog without added context is a safe change. I don't know how many of our Minecraft users also use our other targets, but if I saw something that said "you are about to delete all MakeCode projects", I would be worried that it means that my arcade and microbit projects would also be deleted.

You could do something like "This will delete all MakeCode ${target} projects".

I agree I think cancel is good in this case. Nice to have a very clear "out" next to the big scary button.

PR #10825, opened 2025-08-04
add forceStatement annotation and deprecate handlerStatement

riknoll:
this PR adds a new comment annotation, `//% forceStatement`, which can be used to force a block to be a statement regardless of if it returns a value or has a function as an argument. `handlerStatement` should be considered deprecated in favor of `forceStatement`, which is the same except it also handles return values

i'm adding this mainly for minecraft where a lot of javascript APIs currently throw away return values so that the blocks can be statements. with this change we can have it both ways: return values in js and statements in blocks.


thsparks (approved):
Looks good, but what's the difference between forceStatement and handlerStatement?

riknoll:
@thsparks handlerStatement was only for events (handler is referring to the callback passed in) and is deprecated with this PR. the new name forceStatement is to make it clear it isn't just for events

PR #10834, opened 2025-08-13
don't cache used blocks when doing quick builds

riknoll:
this tweaks the `quick` cli option when building targets so that we skip the step where we cache used blocks for tutorials. notably, the quick option is passed to the target build when using the `--rebundle` flag on `pxt serve`.

in pxt-minecraft, we cache a bunch of hour of code tutorial blocks which takes a very long to cache when doing a serve:
* without this change, `pxt serve --rb` takes about 59 seconds on my machine to actually start the local dev server.
* with this change, `pxt serve --rb` takes 17 seconds

just know that if you're testing tutorial perf locally, make sure to not use rebundle (this is globally true of all perf investigations)

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -2563,7 +2563,7 @@ async function buildTargetCoreAsync(options: BuildTargetOptions = {}) {
     const hexCachePath = path.resolve(process.cwd(), "built", "hexcache");
     nodeutil.mkdirP(hexCachePath);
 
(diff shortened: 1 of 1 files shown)


srietkerk:
Would this mean that when testing hour of code tutorials locally, blocks will take longer to load?

riknoll:
@srietkerk yes, on local serve if you have the `--rebundle` flag set. not using that flag will cause everything to work normally.

PR #10836, opened 2025-08-15
Reset feature page

### LINKED ISSUE #10822: Documentation on Resetting the Editor
The reset functionality has been around for a while. With the addition of cloud sync and auth, it may be unclear what resetting actually does, especially when logged in. We should have some documentation about what resetting means when logged in and when not.

Once we have this, we should include a link to that doc in the reset modal:

Image

@Jaqster, just wanted to ping you in case there is anything else that should be noted here.

ganicke:
Add a destination help page for the "Reset" feature.

Kind of light on info in here so far. What else is affected by "Reset"?

Closes #10822

### DIFF
--- a/common-docs/SUMMARY.md
+++ b/common-docs/SUMMARY.md
@@ -80,6 +80,7 @@
     * [Sharing projects](/share)
     * [Offline support](/offline)
     * [Save](/save)
+    * [Reset](/reset)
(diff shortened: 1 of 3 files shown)


abchatra (approved):

abchatra:
This is good enough.

PR #10837, opened 2025-08-15
use anchor elements for social buttons

riknoll:
this is a redo of https://github.com/microsoft/pxt/pull/10277

replaces the button elements for the social buttons with anchor elements. also changes the minecraft in game behavior so that we use the default browser navigation instead of window.open (which is unsupported in their embedded webview).


thsparks (approved):
TIL `a` stands for anchor. I always wondered...

PR #10843, opened 2025-08-19
Support different tutorial instruction layouts per editor language

thsparks:
This change adds support for switching to the tutorial "sim sidebar" layout for specific editor languages (blocks, javascript, or python).

The name is confusing here. "Sim sidebar" means two things - the simulator is on the lefthand side *and* the instructions are at the top. It was made for Micro:bit, where it is used for all tutorials and having the sim on the side (vs minisim) is the more prominent difference. But in Minecraft (in-game), the simulator is hidden, so it just moves the instructions to the top, which is what we want. I considered renaming it to something like "top instruction layout", but that is overloaded with the layout for tablet-sized screens, so you could end up not enabling "top instructions" mode but still ending up with instructions at the top...which felt similarly confusing. So in the end, I just left it as-is.

I was a little worried about backwards compat when getting rid of the existing flag, but since pxtarget is versioned with the rest of the app, I think it's actually okay.

Once this is approved, I'll follow up with separate changes in pxt-minecraft, pxt-microbit, and pxt-calliope to add/update the flag.

Will fix https://github.com/microsoft/pxt-minecraft/issues/2743 once pxt-minecraft is also updated.

Try it: https://minecraft.makecode.com/app/0a4c40997c73033cbe3b20895572e677edcd6ae5-a9ffeb0a2d?gameVersion=1.21.91&inGame=1&ipc=1&lang=en-US&protocolVersion=17104896

### DIFF
--- a/localtypings/pxtarget.d.ts
+++ b/localtypings/pxtarget.d.ts
@@ -525,7 +525,8 @@ declare namespace pxt {
         tours?: {
             editor?: string // path to markdown file for the editor tour steps
         }
-        tutorialSimSidebarLayout?: boolean; // Enable tutorial layout with the sim in the sidebar (desktop only)
+        tutorialSimSidebarLayout?: boolean; // Enable tutorial layout with the sim in the sidebar (desktop only) for all languages.
+        tutorialSimSidebarLangs?: string[]; // List of languages that should use tutorialSimSidebarLayout.
         showOpenInVscode?: boolean; // show the open in VS Code button
         matchWebUSBDeviceInSim?: boolean; // if set, pass current device id as theme to sim when available.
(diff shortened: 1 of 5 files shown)


riknoll (approved):
can you test the area menu overlay? i recently made changes to that logic to fix the toolbox area's position depending on the tutorial layout https://github.com/microsoft/pxt/pull/10842/files

thsparks:
> can you test the area menu overlay? i recently made changes to that logic to fix the toolbox area's position depending on the tutorial layout https://github.com/microsoft/pxt/pull/10842/files

Checked blocks desktop, blocks tablet, and monaco (desktop & tablet). Seems okay to me:
image
image
image

PR #10850, opened 2025-08-25
remove "run simulator" context menu item in monaco and fix shortcut

riknoll:
https://github.com/microsoft/pxt-minecraft/issues/2840

this removes the "run simulator" context menu item from the monaco editor but leaves the keyboard shortcut behind (ctrl+enter). it also tweaks the keyboard shortcut so that it behaves the same as clicking the run button; in other words it now toggles the run state rather than doing nothing if already running.

the "clickTrigger" param here indicates user interaction and is used in the mc editor to indicate that we should hide the code window

### DIFF
--- a/webapp/src/monaco.tsx
+++ b/webapp/src/monaco.tsx
@@ -994,6 +994,15 @@ export class Editor extends toolboxeditor.ToolboxEditor {
                 run: () => Promise.resolve(this.parent.typecheckNow())
             });
 
+            this.editor.addAction({
+                id: "runSimulator",
+                label: lf("Run Simulator"),
+                keybindings: [monaco.KeyMod.CtrlCmd | monaco.KeyCode.Enter],
+                keybindingContext: "!editorReadonly",
+                precondition: "!editorReadonly",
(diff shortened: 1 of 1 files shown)


webapp/src/monaco.tsx
```diff
@@ -1000,9 +1000,7 @@ export class Editor extends toolboxeditor.ToolboxEditor {
                 keybindings: [monaco.KeyMod.CtrlCmd | monaco.KeyCode.Enter],
                 keybindingContext: "!editorReadonly",
                 precondition: "!editorReadonly",
-                contextMenuGroupId: "0_pxtnavigation",
```
srietkerk, line 1003: Is this something that should be removed for all editors? It seems like it works well for Arcade and micro:bit.

riknoll: it's not very useful, so we decided to remove it entirely

riknoll:
@srietkerk @jwunderl can i get a review on this?

jwunderl (approved):

PR #10855, opened 2025-09-01
Post users locale to simulator

Amerlander:
The jacdac simulator is introducing localization. The current draft is taking the language from browsers navigator: https://github.com/jacdac/jacdac-docs/pull/21

If pxt sent a message to the iframe, it would be possible to pick up the current language selection of the user inside the simulator and keep it in sync.

This pull request introduces a small enhancement to the simulator driver by ensuring the user's language preference is communicated to the simulator frame when initializing. This helps support localization and improves the user experience for non-English users.

Localization support:

* In `pxtsim/simdriver.ts`, the code now checks for a user language setting (`options.userLanguage`) and, if present, sends a `pxt-set-language` message to the simulator frame to set the appropriate language.

### DIFF
--- a/pxtsim/simdriver.ts
+++ b/pxtsim/simdriver.ts
@@ -851,6 +851,11 @@ namespace pxsim {
                 msg.breakOnStart = false;
             }
             this.postMessageCore(frame, msg);
+            const lang = this.options?.userLanguage;
+            if (lang) {
+                const langMsg: any = { type: 'pxt-set-language', lang: lang, source: MESSAGE_SOURCE };
+                this.postMessageCore(frame, langMsg);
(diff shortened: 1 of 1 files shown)


Amerlander:
@thomasjball here is the PR

@abchatra Could you take a look and assign the appropriate person for review?

thomasjball:
simdriver.ts already has this logic
```
if (this.options.userLanguage)
  url.searchParams.set("language", encodeURIComponent(this.options.userLanguage));
```
let me see if we can get jacdac simx working via this path first

thomasjball:
verified that simx logic already passed language=de as required when makecode language set to German

PR #10856, opened 2025-09-02
Check for secrets before logging strings and/or sending them to the backend

thsparks:
This adds a cleanData function (based off the [secrets checking in vscode](https://github.com/microsoft/vscode/blob/main/src/vs/platform/telemetry/common/telemetryUtils.ts#L328)), and calls it in a few places:
1. Tick event fields
2. Extension search (which gets sent directly to the backend)
3. AI questions in teacher tool (also get sent to backend)

I would've preferred to put the tick event checks all the way down in [cookieCompliance](https://github.com/microsoft/pxt/blob/master/docfiles/pxtweb/cookieCompliance.ts) but it seems a little messy to share code there, so I moved it up the stack a little.

Fixes https://github.com/microsoft/pxt-minecraft/issues/2815

### DIFF
--- a/pxtlib/util.ts
+++ b/pxtlib/util.ts
@@ -1990,7 +1990,7 @@ namespace ts.pxtc.Util {
             { label: 'Google API Key', regex: /AIza[A-Za-z0-9_\\\-]{35}/ },
             { label: 'Slack Token', regex: /xox[pbar]\-[A-Za-z0-9]/ },
             { label: 'GitHub Token', regex: /(gh[psuro]_[a-zA-Z0-9]{36}|github_pat_[a-zA-Z0-9]{22}_[a-zA-Z0-9]{59})/ },
-            { label: 'Generic Secret', regex: /\b(key|token|sig|secret|signature|password|passwd|pwd|android:value)[^a-zA-Z0-9]{0,200}/i },
+            { label: 'Generic Secret', regex: /\b(token|signature|password|passwd|pwd|android:value)[^a-zA-Z0-9]{0,200}/i },
             { label: 'CLI Credentials', regex: /((login|psexec|(certutil|psexec)\.exe).{1,50}(\s-u(ser(name)?)?\s+.{3,100})?\s-(admin|user|vm|root)?p(ass(word)?)?\s+["']?[^$\-\/\s]|(^|[\s\r\n\\])net(\.exe)?.{1,5}(user\s+|share\s+\/user:| user -? secrets ? set) \s + [^ $\s \/])/ },
             { label: 'JWT Token', regex: /\beyJ[A-Za-z0-9_\-]+?\.[A-Za-z0-9_\-]+?\.[A-Za-z0-9_\-]+?\b/ },
             { label: 'Email', regex: /\b[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,63}\b/i }
--- a/pxtlib/util.ts
+++ b/pxtlib/util.ts
@@ -1948,6 +1948,66 @@ namespace ts.pxtc.Util {
 
         return enabled;
     }
+
+    /**
+     * Remove potentially sensitive info from the given data to avoid logging it.
+     * Currently only supports string data.
+     */
+    export function cleanData(data: any, visited = new WeakSet(), depth = 0): any {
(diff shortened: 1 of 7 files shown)


pxtlib/util.ts
```diff
@@ -1948,6 +1948,62 @@ namespace ts.pxtc.Util {
...
+     * Taken from https://github.com/microsoft/vscode/blob/main/src/vs/platform/telemetry/common/telemetryUtils.ts
+     */
+    function removePropertiesWithPossibleUserInfo(property: string): string {
+        // If for some reason it is undefined we skip it (this shouldn't be possible);
+        if (!property) {
+            return property;
+        }
+
+        const userDataRegexes = [
+            { label: 'Google API Key', regex: /AIza[A-Za-z0-9_\\\-]{35}/ },
```
eanders-ms, line 1990: There was once an instance of logging a JWT token, so we might want to scan for those too: `{ label: 'JWT', regex: /\beyJ[A-Za-z0-9_\-]+?\.[A-Za-z0-9_\-]+?\.[A-Za-z0-9_\-]+?\b/ },`.

pxtlib/util.ts
```diff
@@ -1948,6 +1948,62 @@ namespace ts.pxtc.Util {
...
+            { label: 'Slack Token', regex: /xox[pbar]\-[A-Za-z0-9]/ },
+            { label: 'GitHub Token', regex: /(gh[psuro]_[a-zA-Z0-9]{36}|github_pat_[a-zA-Z0-9]{22}_[a-zA-Z0-9]{59})/ },
+            { label: 'Generic Secret', regex: /(key|token|sig|secret|signature|password|passwd|pwd|android:value)[^a-zA-Z0-9]/i },
+            { label: 'CLI Credentials', regex: /((login|psexec|(certutil|psexec)\.exe).{1,50}(\s-u(ser(name)?)?\s+.{3,100})?\s-(admin|user|vm|root)?p(ass(word)?)?\s+["']?[^$\-\/\s]|(^|[\s\r\n\\])net(\.exe)?.{1,5}(user\s+|share\s+\/user:| user -? secrets ? set) \s + [^ $\s \/])/ },
+            { label: 'Email', regex: /@[a-zA-Z0-9-]+\.[a-zA-Z0-9-]+/ } // Regex which matches @*.site
```
eanders-ms: This email regex seems overly broad. It will match strings like "@foo.anything". Consider using `/\b[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,63}\b/i`. It's narrower, will early out sooner, and has a max scanning length.

It also doesn't match the name part of the email at all. That's works ok here since you're nuking the whole value, but not a great if we ever decide to keep some of the surrounding text.

pxtlib/util.ts
```diff
@@ -1948,6 +1948,62 @@ namespace ts.pxtc.Util {
...
+        // If for some reason it is undefined we skip it (this shouldn't be possible);
+        if (!property) {
+            return property;
+        }
+
+        const userDataRegexes = [
+            { label: 'Google API Key', regex: /AIza[A-Za-z0-9_\\\-]{35}/ },
+            { label: 'Slack Token', regex: /xox[pbar]\-[A-Za-z0-9]/ },
+            { label: 'GitHub Token', regex: /(gh[psuro]_[a-zA-Z0-9]{36}|github_pat_[a-zA-Z0-9]{22}_[a-zA-Z0-9]{59})/ },
+            { label: 'Generic Secret', regex: /(key|token|sig|secret|signature|password|passwd|pwd|android:value)[^a-zA-Z0-9]/i },
```
eanders-ms: Suggest putting a few guardrails on this one:
- `\b` - don't match mid-word. You may get false positives matching "key" and "sig".
- `{0,200}` don't scan more than 200 chars.
```suggestion
            { label: 'Generic Secret', regex: /\b(key|token|sig|secret|signature|password|passwd|pwd|android:value)[^a-zA-Z0-9]{0,200})/i },
```

thsparks: Good points. I have actually removed a few entries from this list, as it's not hard to imagine legitimate extensions that might use "key" or "secret" or have words that start with "sig" (readable-signs-extension, or something).

PR #10870, opened 2025-09-16
Beta editor page

ganicke:
Add a page describing how to use the Beta version of the editor.

Closes https://github.com/microsoft/pxt-minecraft/issues/2879


common-docs/beta.md
```diff
@@ -0,0 +1,27 @@
...
+![Experiments button](/static/beta/about-experiments.jpg)
+
+In the "Experiments" window, find the **Beta** card and click on it. The Beta editor will now load.
+
+![Beta card](/static/beta/beta-card.jpg)
+
+Have fun with the Beta editor!
+
+### ~alert
+
```
thsparks, line 22: We probably need to add a caution note here about the possibility of any beta projects getting lost, becoming incompatible, and so on... Something like "Beta is for testing and experimentation **only**. We do not recommend using it for actual projects. It's under constant development, and any projects made in Beta may be lost or become nonfunctional in future updates."

thsparks (approved):

PR #10871, opened 2025-09-22
Fix Keyboard Nav with FlyoutOnly (aka UnifiedToolbox) Mode

thsparks:
This contains a number of fixes for keyboard nav, specifically around flyoutOnly mode (aka unifiedToolbox mode).

Firstly, it updates the logic for `getToolboxBounds` so that it returns the flyout bounds when in flyoutOnly mode. It also adjusts the focus logic to focus on the flyout when needed. This fixes https://github.com/microsoft/pxt-minecraft/issues/2860 (overlay now points to the correct place on the screen, and pressing the 3 button takes you to the flyout as expected).

Secondly, just as a simple cleanup task, I moved `registerKeyboardNavigationStyles` into `loadBlocklyAsync` which only runs once (unlike `prepareBlockly`), so we don't need a separate flag to track if it has run.

After doing that, I noticed a number of issues that also needed fixing, most of which stemmed from keyboardNavigation getting into a weird state whenever we had to call `refreshBlockly` (which we do when going between flyoutOnly and normal modes). To account for this, I made the following changes:
1. Added a `cleanupKeyboardNavigation` phase to refreshing blockly which resets keyboard nav and allows it to reinitialize properly alongside the rest of blockly (in `prepareBlockly`)
2. Adjust `initCopyPaste` to we can force re-initialization. This was needed because cleaning up keyboard nav would clear the copy, cut, and paste shortcuts, so they needed to be recreated, but oldCopy was still set so the init step was being skipped.
3. Patch Blockly's `addKeyMapping` so we cannot add duplicate mappings, which would not get cleaned up properly, which would in turn cause shortcut key collisions later on. This seemed to happen a lot in flyoutOnly mode. I struggled to pinpoint _why_ the duplicate registrations were coming in (if it was our code or blockly code), but I think it's safe to assume we don't want duplicate mapping entries regardless.

As a side note, a few of these issues have already been addressed in the keyboard accessibility extension code (here: https://github.com/google/blockly-keyboard-experimentation/pull/687), but we'd need to update blockly to pull that in, and I don't think we should do that so close to a release.

Upload Target:
https://minecraft.makecode.com/app/60a2e7a4b5181c32067ec6367be327c2ec377e15-5a9a04603d?inGame=1&ipc=1&lang=en-US&protocolVersion=17104896

For a flyoutOnly tutorial (note, you will need to turn on keyboard nav before going here):
https://minecraft.makecode.com/app/60a2e7a4b5181c32067ec6367be327c2ec377e15-5a9a04603d#tutorial:/hour-of-code/2024/tutorial?ingame=1&ipc=1&norunonx=1


webapp/src/blocks.tsx
```diff
@@ -1808,6 +1813,19 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         pxt.perf.measureEnd(Measurements.RefreshToolbox)
     }
 
+    cleanupKeyboardNavigation() {
+        if (this.keyboardNavigation) {
+            if (ShortcutRegistry.registry.getRegistry()["commitMove"]) {
```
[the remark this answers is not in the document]
thsparks: That API does not exist. The approach taken here is consistent with other places in the code file.

thsparks:
> LGTM!
> 
> Just curious, did you also check on microbit/arcade to make sure that keyboard nav, especially navigating to the toolbox via the shortcut, still works as expected?

Good call. Turns out there was an issue that surfaced in microbit. Fixed now.

pxtblocks/copyPaste.ts
```diff
@@ -9,18 +9,22 @@ let oldCut: Blockly.ShortcutRegistry.KeyboardShortcut;
...
-    oldPaste = { ...shortcuts[Blockly.ShortcutItems.names.PASTE] };
+    oldCopy = oldCopy || { ...shortcuts[Blockly.ShortcutItems.names.COPY] };
+    oldCut = oldCut || { ...shortcuts[Blockly.ShortcutItems.names.CUT] };
+    oldPaste = oldPaste || { ...shortcuts[Blockly.ShortcutItems.names.PASTE] };
 
     Blockly.ShortcutRegistry.registry.unregister(Blockly.ShortcutItems.names.COPY);
     Blockly.ShortcutRegistry.registry.unregister(Blockly.ShortcutItems.names.CUT);
     Blockly.ShortcutRegistry.registry.unregister(Blockly.ShortcutItems.names.PASTE);
 
+    Blockly.ShortcutRegistry.registry.unregister("keyboard_nav_copy");
```
thsparks: Noticed this seems to be causing issues with opening the context menu after switching between monaco and blocks. Guessing I need to be more careful about when we unregister/register these. Investigating...

thsparks: Fixed by putting initCopyPaste back to no-oping when oldCopy exists, but by adding a forceRefresh flag which we can use after keyboardNav cleanup to re-create the old shortcuts (which was the initial reason for removing the no-op behavior in the first place).

PR #10880, opened 2025-09-26
High contrast dropdown menu fixes

thsparks:
Fixes https://github.com/microsoft/pxt-minecraft/issues/2890

This change ensures items in the menu dropdown:
1. Don't change foreground color on hover (even when they're links)
2. Do have the expected border on hover in high contrast mode

As a part of this, I also removed some redundant CSS and moved a few rules into the high contrast overrides file, making things more consistent with the new theming.

Lastly, I turned off the automatic focus on the first element in the list. This was causing double highlighting (keyboard focus on one, mouse hover on the other) which seemed confusing. It's also more consistent with our other dropdowns. I confirmed you can still use arrow keys to navigate into the menu once it's open.

With auto-focus, you get double highlights:
image

Without, you can still navigate the menu with arrow keys, but you won't get the extra highlight until you start to do so, at which point it's much clearer what's happening:
image

Upload target: https://minecraft.makecode.com/app/d50b0bf765a03fb46e89b105df80bddc2f2cd59c-e7433d28d8

### DIFF
--- a/react-common/styles/controls/MenuDropdown.less
+++ b/react-common/styles/controls/MenuDropdown.less
@@ -35,7 +35,6 @@
         display: flex;
         align-items: center;
         text-align: left;
-        position: relative;
 
         cursor: pointer;
         user-select: none;
--- a/react-common/styles/controls/MenuDropdown.less
+++ b/react-common/styles/controls/MenuDropdown.less
@@ -16,7 +16,7 @@
     margin: 0;
     background: var(--pxt-neutral-background1);
     border: 1px solid var(--pxt-neutral-stencil1);
-    overflow: visible;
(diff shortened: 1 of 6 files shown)


react-common/styles/controls/MenuDropdown.less
```diff
@@ -16,6 +16,7 @@
     margin: 0;
     background: var(--pxt-neutral-background1);
     border: 1px solid var(--pxt-neutral-stencil1);
+    overflow: visible; /* Allow focus borders to extend outside the pane */
```
thsparks: Going to take a different approach here so we don't mess with scrollbar behavior. Will probably just pull the outline offset in so it's all within bounds of the element.

PR #10884, opened 2025-10-01
fix extension button filtering logic

riknoll:
my hotfix from yesterday caused the filtering logic for namespaces to start filtering out the "extensions" button outside of tutorials. this PR fixes it, and should be hotfixed ASAP


aznhassan (approved):
:shipit:

PR #10887, opened 2025-10-02
Jacdac cherry pick

thomasjball:
picking commits needed for moving to new jacdac


abchatra:
This is larger change than I anticipated and will require proper testing. I would like to wait on it till the team is back on Arcade to release.

PR #10901, opened 2025-10-10
Adding displayName support to extensions

abchatra:
Adds support for displayName in extensions which is other than the name.

### DIFF
--- a/webapp/src/blocks.tsx
+++ b/webapp/src/blocks.tsx
@@ -775,6 +775,9 @@ export class Editor extends toolboxeditor.ToolboxEditor {
 
     private markIncomplete = false;
     isIncomplete() {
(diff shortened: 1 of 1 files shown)


pxtlib/github.ts
```diff
@@ -1090,6 +1091,7 @@ namespace pxt.github {
                     fileName: rid.fileName,
                     slug: rid.slug,
                     name: rid.fileName ? `${meta.name}-${rid.fileName}` : meta.name,
+                    displayName: meta.displayName,
```
srietkerk, line 1094: Should this have a default if the displayName is not specified in the pxt.json?

abchatra: Name is default at the call site, I don't want to duplicate at this point as code paths are different for built in packages vs github pacakges.

PR #10908, opened 2025-10-21
create sprite kinds during field validation

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/6983

or at least, it should prevent that bug from breaking projects. this changes the kind field so that if it is passed a kind value that doesn't exist, it will create a new blockly variable for that kind rather than resetting its value to player.

i'd like to hotfix this since it's affecting a lot of users, is a bad bug, and this fix should be pretty safe

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pxt-core",
-  "version": "12.2.1",
+  "version": "12.2.2",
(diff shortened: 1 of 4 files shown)


jwunderl (approved):
10/22/2025 about 55 minutes uptime this occurred

PR #10913, opened 2025-10-22
workaround chrome bug that allows block text to be selected

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/6838

this "fixes" the bug where all the block text sometimes turns black. i put that in quotes because as far as i can tell, that bug is a browser bug where chrome won't respect the `user-select: none` css that is placed on all the blockly blocks. to workaround that behavior, i'm adding styling that makes it so that the selected text color is the same as the unselected text color so that you can't see the difference


jwunderl (approved):

Capybara-Are-Fun:
cool

PR #10921, opened 2025-10-23
add support for help pages for blockCombine blocks

riknoll:
this adds the documentation support required for https://github.com/microsoft/pxt-arcade/issues/5465

this adds three new block parameters:
```
blockCombineChangeHelp
blockCombineGetHelp
blockCombineSetHelp
```

as the names suggest, these let you specify the help urls for get/set/change blocks. you can put these annotations onto any one of the properties that are combined (doesn't matter if it's a getter or a setter). if you want to have the same page for all three, this PR also adds support for the usual `//% help` attribute which will apply to all three if the others aren't present.

@ganicke FYI


jwunderl (approved):

ganicke:
Wow, I've wanted this for a long time! Thanks, I'll give it a try.

PR #10944, opened 2025-11-11
fix GC-related crash in codegen

mmoskal:
This resolves issues encountered in the microcode app.

If we construct an array literal and drop the result immediately (with TD_ID) the code would be something like this:

```
    bl Array_::mk
    push {r0} // save pointer to array
   ...
    ldr r0, [sp, #4*0] // load pointer to array
   ....
    bl Array_::push // add value
   ...
    ldr r0, [sp, #4*0] // load pointer to array
   ....
    bl Array_::push // add value
   ...
    // last iteration
    pop {r0} // load pointer to array - it's no longer used so we pop it
   ....
    bl Array_::push // add value
```

If the last `Array_::push` needs to grow the array and the allocation triggers a GC, then during the GC the array is no longer on the TypeScript stack, so it's get GC-ed. Then we try to copy the data from deallocated array to the newly allocated space and crash.

A similar bug occurs for object literals. The fix is to keep the reference to the array/object until after the last push/mapSet. This is demonstrated by repro 1 and 3.

Additionally, when setting values in an object literal, we would not hold the reference to the value on the TypeScript stack while the mapSet is executing. Thus again, if the map needs to grow the value will get de-allocated and we'll store a bogus pointer in the map (repro 2). This was actually already handled correctly for array literals.

Finally, after fixing these bugs the layout of the generated code changed and I was running into the following on microcode app:

```
main.ts(1,1): error TS9200:   -> Line 43504 ('    bne .jmpz1827'), error: assembly error
   Maybe: bne   LABEL (argument out of range or mis-aligned at '.jmpz1827')
```

This is because we inject data for the "ldlit" pseudo instructions at opportune points in the instruction stream. In this case we would inject it around this `bne` instruction, which makes it target be out of range (+-256). I have modified the code to inject at most 160 bytes of ldlit so instructions don't move so much.

Repro 1 (crashes after a heartbeat or two) - this drops the map because it's not used (TD_ID):

```ts
//% shim=TD_ID
function ignore(v: any) {}

basic.forever(function () {
    basic.showIcon(IconNames.Heart)
    basic.showIcon(IconNames.SmallHeart)

    for (let i = 0; i < 10000; ++i) {
        ignore({
            "A": 1,
            "B": 2,
            "C": 3,
            "D": 4,
            "E": 5 
        })
    }
})
```

Repro 2 (also crashes after a heartbeat or two) - this drops the value added to the map when the map is extended:

```ts
function test(v: any) {
    control.assert(v["D"] == "d4")
    control.assert(v["E"] == "e5")
}

basic.forever(function () {
    basic.showIcon(IconNames.Heart)
    basic.showIcon(IconNames.SmallHeart)

    for (let i = 0; i < 10000; ++i) {
        test({
            "A": 1,
            "B": 2,
            "C": 3,
            "D": "d" + 4,
            "E": "e" + 5 
        })
    }
})
```

Repro 3 (same as 1 but for array literal not object literal)

```ts
//% shim=TD_ID
function ignore(v: any) { }

basic.forever(function () {
    basic.showIcon(IconNames.Heart)
    basic.showIcon(IconNames.SmallHeart)

    for (let i = 0; i < 10000; ++i) {
        ignore([1,2,3,4,5])
    }
})
```


mmoskal:

abchatra (approved):

PR #10945, opened 2025-11-11
fix typechecking for unions of array types

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/6113

when given a union type like `(number[] | string[])` our compiler was categorizing this as "not an array type" even though it's a union of two array types. as a result, the emitter would emit indexed access/assign operations using the generic map lookup functions rather than the array versions of those functions. our arrays are not addressable as maps, so you would end up getting an arcane runtime error when trying to index into an object of that type.

as i was adding a test for this, i also noticed that someone has added a test file for parseint but never actually added it to the pxt.json of our test project, so i added that as well.

### DIFF
--- a/pxtcompiler/emitter/emitter.ts
+++ b/pxtcompiler/emitter/emitter.ts
@@ -657,6 +657,14 @@ namespace ts.pxtc {
     }
 
     function isArrayType(t: Type) {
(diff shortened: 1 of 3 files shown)


srietkerk (approved):
LGTM

PR #10949, opened 2025-11-13
remove NPM_ACCESS_TOKEN secret for trusted deployment

riknoll:
replaces the NPM_ACCESS_TOKEN environment variable with one that just indicates that we want to publish. also scraps the .npmrc stuff because with trusted deployment we shouldn't need it.


abchatra:
I am assuming if this works, we will have to move to every stable branch (if we need to get a new build). Right?

PR #10955, opened 2025-11-19
Improve perf for field editor regexes and fix music editor bug

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/6501

the previous regexes for the image and music field editors favored compactness over performance; this didn't particularly matter for small images but could lead to hanging the page for large input strings (like a background image). this PR improves performance by splitting the regex into discrete regexes for each possible input case instead of chaining lots of `|` clauses that lead to endless backtracking.

also fixes an unreported bug that I noticed while testing: if a song asset was converted from a hex buffer to a named asset then the code output by the field editor would be invalid because it would strip the call to `music.createSong`. the fix here just looks for that case and outputs the call

### DIFF
--- a/pxteditor/monaco-fields/field_musiceditor.ts
+++ b/pxteditor/monaco-fields/field_musiceditor.ts
@@ -66,8 +66,16 @@ export class MonacoSongEditor extends MonacoReactFieldEditor<pxt.Song> {
             } else {
                 result = project.createNewSong(result.song, result.meta.displayName);
             }
-            this.isAsset = true;
-            return pxt.getTSReferenceForAsset(result, this.isPython);
+            let out = pxt.getTSReferenceForAsset(result, this.isPython);
+            if (!this.isAsset) {
+                if (this.isPython) {
+                    out = `music.create_song(${out})`;
+                }
+                else {
+                    out = `music.createSong(${out})`;
+                }
+            }
+            return out;
         }
 
         let hexString = pxt.assets.music.encodeSongToHex(result.song);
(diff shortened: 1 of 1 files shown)


pxteditor/monaco-fields/field_musiceditor.ts
```diff
@@ -101,25 +109,26 @@ function createFakeAsset(song: pxt.assets.music.Song): pxt.Song {
...
+    "music\\s*\\.\\s*createSong\\s*\\(\\s*hex`[a-fA-F0-9\\s\\n]*`\\s*\\)",
+    "assets\\s*\\.\\s*song\\s*`[a-zA-Z_\\s\\n]*`",
+
+    // python
+    'music\\s*\\.\\s*create_song\\s*\\(\\s*hex\\s*\\(\\s*"""[a-fA-F0-9\\s\\n]*"""\\s*\\)\\s*\\)',
+    'music\\s*\\.\\s*createSong\\s*\\(\\s*hex\\s*\\(\\s*"""[a-fA-F0-9\\s\\n]*"""\\s*\\)\\s*\\)',
+    'assets\\s*\\.\\s*song\\s*\\(\\s*"""[a-zA-Z_\\s\\n]*"""\\s*\\)'
+];
+
+const searchString = regexes.map(r => `(?:${r})`).join("|");
```
srietkerk, line 123: Why is this map needed? Could each of the regexes in the list just be wrapped with `(?:{content})`? Does the mapping before joining give us a perf bump?

riknoll: it just makes them easier to read to leave them out of the array

PR #10960, opened 2025-11-20
improve monaco experience for creating new assets and animations

riknoll:
this pr improves the experience of creating new assets in the monaco editor. now you can just type the reference with no id, for example:

```ts
assets.image``
assets.song``
assets.animation``
```

and clicking on the asset editor icon will create a new asset with an autogenerated name. also, this PR adds an animation editor for monaco. this new editor only works for named assets because regexing for arrays of images is a little too gross

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pxt-core",
-  "version": "12.2.8",
+  "version": "12.2.9",
   "description": "Microsoft MakeCode provides Blocks / JavaScript / Python tools and editors",
   "keywords": [
     "TypeScript",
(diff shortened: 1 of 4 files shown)


pxteditor/monaco-fields/field_animation.ts
```diff
@@ -0,0 +1,87 @@
...
+    protected editing: pxt.Asset;
+
+    protected textToValue(text: string): pxt.Animation {
+        this.isPython = text.indexOf("`") === -1
+
+        const match = pxt.parseAssetTSReference(text);
+        if (match) {
+            const { name: matchedName } = match;
+            const name = matchedName.trim();
+            const project = pxt.react.getTilemapProject();
```
srietkerk, line 17: Why does need to get the tilemap project?

srietkerk (approved):
Just a quick question for myself because of lack of context.

LGTM

pxteditor/monaco-fields/field_animation.ts
riknoll: the tilemap project holds all the assets for the project, not just tilemaps

PR #11006, opened 2025-12-10
Fix codeembed share option

srietkerk:
Fixes https://github.com/microsoft/pxt-arcade/issues/6924

The style-blockly.css file/reference no longer exists, so the look-up for that stylesheet caused failures in the codeembed option for the share. We now look for the correct blockly styling. I also verified that rtl languages display correctly on the code embed view.

Upload target: https://arcade.makecode.com/app/8b847c5d3864ce9d8fdee73355facc4f9ca5d0c1-fa94dfc65a---codeembed#pub:_8yXiH3dPdFxc

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -324,9 +324,11 @@ export class ProjectView
      * Run a global action based on shortcuts triggered in sim or main window.
      */
     private runGlobalAction(action: pxsim.GlobalAction) {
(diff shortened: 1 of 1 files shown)


jwunderl (approved):
lgtm, but would suggest clearing out / committing any local changes in your master branch to another branch(they'll disappear otherwise) then `git reset --hard origin/master`, or on master branch `git branch -m my-local-master-backup` then `git checkout master` to get back in sync with `microsoft/pxt`'s master branch

PR #11007, opened 2025-12-10
support using return blocks outside of function definitions

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/4485

adds support for using return statement blocks in event handlers in addition to function definitions. the rules for return statements are now as follows:

* return statements can always be used inside function definitions
* return statements can be used in event handlers iff they do _not_ have a return value
* return statements can always be attached to disabled blocks
* return statements will refuse to connect to blocks outside of a function handler (previously they would connect and then immediately disconnect)

if the project somehow breaks these rules, for example by attaching a return statement to a disabled block and then dragging that new block group into on start, the blockly compiler will give an error but will still attempt to compile the program.

as part of this, i also removed the old disconnecting logic for the return statement blocks in favor of the connection checker. now instead of immediately disconnecting when connecting to an invalid input, the connection is simply disallowed from the get-go.

also includes corresponding decompiler change and new test for both the decompiler and blockly compiler.

return_connection_fun

### DIFF
--- a/tests/blocklycompiler-test/baselines/return_statement_outside_function.ts
+++ b/tests/blocklycompiler-test/baselines/return_statement_outside_function.ts
@@ -1,4 +1,4 @@
-function doSomething() {
+function doSomething () {
     custom.handlerStatement(function () {
         return
     })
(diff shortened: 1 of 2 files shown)


jwunderl (approved):
LGTM pending test fix

PR #8290, opened 2021-07-13
Move skillmap projects to the editor workspace

riknoll:
Moving all of the skillmap projects from outside the editor iframe to inside! List of changes includes:

1. Skillmap headers are now saved along all of the other editor projects with a flag to filter them out of "my projects"
2. When we load the workspace, skillmap projects are automatically imported and soft-deleted from the skillmap workspace
3. All code carryover project merging now happens inside the iframe when the tutorial is loaded
4. Created some new tutorial events for reporting progress from the editor to the MakecodeFrame
5. The skillmap now has the webconfig embedded in the HTML file and points to the correct target version instead of the embedURL for the iframe
6. The "start fresh"/"keep code" dialog now shows *before* we load the editor instead of *after*

Here's a test upload:

* [editor](https://arcade.makecode.com/app/e0f5b5912347d2ab2265fcc9ae4b863f8c77451a-e68d4ba962)
* [skillmap](https://arcade.makecode.com/app/e0f5b5912347d2ab2265fcc9ae4b863f8c77451a-e68d4ba962---skillmap)

**Note**: If you open this editor/skillmap upload, all of your skillmap tutorial projects will suddenly show up in the beta/released editor. The code to filter skillmap projects out of "My Projects" is part of this PR

TODO:
1. Now that the dialog is before the editor load, we should start loading the editor in the background to improve perf
2. I need to fix the loading editor event so that you don't ever see "starting tutorial" when opening an activity

PR #11008, opened 2025-12-11
don't disable cloud sync for skillmap

riknoll:
this was regressed by https://github.com/microsoft/pxt/pull/9546. I believe that change went out in arcade in march, so it's been broken for most of this year.

fixes cloud syncing of skillmap projects.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -6260,7 +6260,7 @@ document.addEventListener("DOMContentLoaded", async () => {
     let theme = pxt.appTarget.appTheme;
     const isControllerIFrame = theme.allowParentController || pxt.shell.isControllerMode()
     // disable auth in iframe scenarios
-    if (isControllerIFrame)
+    if (isControllerIFrame && !pxt.BrowserUtils.isSkillmapEditor())
         pxt.auth.enableAuth(false);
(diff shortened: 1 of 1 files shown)


abchatra:
Now I question the popularity of skillmap

abchatra:
We should hotfix this right away.

jwunderl (approved):
it looks like that enableAuth(false) path was hit prior to that pr, that pr bumped it up a lil earlier but the original intention of the check was skipping for skillmap? https://github.com/microsoft/pxt/pull/7934/changes

or is there something else in between I'm missing there, like auth being enabled until workspace is set up then disabled after? either way change seems fine now just curious

riknoll:
@jwunderl that linked pr is super old. if you look at the file prior to the change i linked, the check for enabling auth was looking to see if we are using the iframe workspace. the skillmap uses the browser workspace (see the `ws=browser` part of the url [here](https://github.com/microsoft/pxt/blob/c095aef4e73bfa952f562d8c86b379b67a1188f1/skillmap/src/components/makecodeFrame.tsx#L139))

riknoll:
the workspace was changed here https://github.com/microsoft/pxt/pull/8290

PR #11010, opened 2025-12-12
[Stable11.3] copy over workflow files from master

riknoll:
this PR is copying over the workflow files from master, which is finally required for arcade stable because of our npm trusted deployment change.

@abchatra note that this will now require you to use a pxt cli built from master in order to bump stable11.3 because i am **not** cherry picking over the CLI changes (but the updates to the workflows will require eric's new bump-via-branch workflows)

the new process for bumping 11.3 will look like this:

```
git checkout master        # build pxt.js before switching to stable11.3 branch
gulp cli
git checkout stable11.3
git merge origin/stable11.3
pxt bump
```

and if you have a separate enlistment for stable 11.3, you can just invoke the cli from your other enlistment like so:

```
node ../path/to/master/enlistment/built/pxt.js bump
```

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -414,14 +414,10 @@ async function ciAsync() {
         buildInfo.branch = "local"
 
     const { tag, branch, pullRequest } = buildInfo
-    const atok = process.env.NPM_ACCESS_TOKEN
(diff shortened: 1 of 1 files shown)


riknoll:
actually, on second thought, i think i also need to cherry pick eric's CLI changes. closing now; will reopen

PR #11015, opened 2025-12-18
Gave blocks rendered in docs a readable aria label

srietkerk:
Fixes https://github.com/microsoft/pxt-arcade/issues/7247

I don't know how to actually access the given docs page, but for those that do, here's an upload target:
https://arcade.makecode.com/app/53337314028ae219be261b6605c60ef19992c2f4-d2ca462b7b

### DIFF
--- a/pxtrunner/renderer.ts
+++ b/pxtrunner/renderer.ts
@@ -823,7 +823,7 @@ function renderInlineBlocksAsync(options: BlocksRenderOptions): Promise<void> {
                         if (symbolInfo && symbolInfo.attributes.help) {
                             // Create accessible label for the link using a human-readable name
                             const readableName = symbolInfo.name || symbolInfo.qName;
-                            const ariaLabel = lf("documentation for {0} block", readableName);
+                            const ariaLabel = lf("Documentation for {0} block", readableName);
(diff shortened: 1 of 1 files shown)


pxtrunner/renderer.ts
```diff
@@ -821,7 +821,10 @@ function renderInlineBlocksAsync(options: BlocksRenderOptions): Promise<void> {
                     if (info && r.apiInfo) {
                         const symbolInfo = r.apiInfo.byQName[info.qName];
                         if (symbolInfo && symbolInfo.attributes.help) {
-                            $newel = $(`<a class="ui link"/>`).attr("href", `/reference/${symbolInfo.attributes.help}`).append($newel);
+                            // Create accessible label for the link using a human-readable name
+                            const readableName = symbolInfo.name || symbolInfo.qName;
+                            const ariaLabel = lf("documentation for {0} block", readableName);
```
jwunderl:
```suggestion
                            const ariaLabel = lf("Documentation for {0} block", readableName);
```

srietkerk:
> LGTM, can't test docs on upload targets. To validate can run locally (just do localhost:3232/{docs-path}

Thanks for the info! That's kinda what I thought but I wasn't completely sure. I did validate locally before putting up the PR.

PR #11030, opened 2026-01-05
Share button for most homepage content

jwunderl:
To make it more easy for teachers / etc to share assignments.

close https://github.com/microsoft/pxt-arcade/issues/7053
close https://github.com/microsoft/pxt-microbit/issues/4939

https://arcade.makecode.com/app/9809388749307f3f234144e23c3b66e3987ca4b8-df3987ed78

i used the same styling as the youtube 'Play Video Lesson' button (e.g. flashing heart on microbit has that); just opened up on my phone and noticed there's some overlap between that button & the tutorial buttons on mobile size, will fix that too before merging

### DIFF
--- a/teachertool/src/components/CriteriaInstanceDisplay.tsx
+++ b/teachertool/src/components/CriteriaInstanceDisplay.tsx
@@ -53,6 +53,11 @@ const InlineInputSegment: React.FC<InlineInputSegmentProps> = ({
             return;
         }
 
+        if (!initialValue) {
+            setErrorMessage(Strings.ValueRequired);
+            return;
+        }
+
         // We still allow some invalid values to be set on the parameter so the user can see what they typed
         // and the associated error.
         // Without this, we risk erroring too soon (i.e. typing in first digit of number with min > 10),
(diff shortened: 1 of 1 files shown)


riknoll (approved):
I do wonder if we should make the button a little less prominent, but we can certainly figure that out after merging. LGTM!

srietkerk:
I'm struggling to reach examples with the link that gets generated. It just hangs on loading.
https://arcade.makecode.com/app/9809388749307f3f234144e23c3b66e3987ca4b8-df3987ed78#example:/blocks-games/hot-air-balloon 
image


This is the only error that I'm seeing
image

jwunderl:
oh huh good catch sarah, name field is listed as optional on importExampleAsync but has crashed there for 5 years if not passed (which it isn't in the handleHash route) i'll fix that as well

jwunderl:
okay tried de-emphasizing it a tad, using same colors as buttons top half so ideally the orange is more attention grabby. as you mentioned easy to adjust post merge
image

also adjusted mobile display a tad, so it's not hidden behind the action buttons
image
but will probably want to fix a little more either way. current site mostly hides blocks on iphone se side so minor change (but it was getting fiddly and i didn't want to try to fiddle more in this pr and break unknown things)
image

another build https://arcade.makecode.com/app/79d241424adcd97da096a580e59e85bcf28ae929-7e7d548aa3, with working examples https://arcade.makecode.com/app/79d241424adcd97da096a580e59e85bcf28ae929-7e7d548aa3#example:/blocks-games/hot-air-balloon

Jaqster:
We could also use the Share icon we have in the editor for smaller screens...  
image

Jaqster:
Wow!!!  This is so great!!! :-)

PR #11035, opened 2026-01-06
fix filtering and snippets for blockAliasFor in monaco

riknoll:
fixes https://github.com/microsoft/pxt-minecraft/issues/3016

fix monaco toolbox filtering and snippets for APIs that use blockAliasFor.

our monaco toolbox filtering currently filters out all of these APIs because they generally start with an underscore, so this PR changes that check so that it doesn't apply if blockAliasFor is set.

also includes a fix that makes sure we follow the blockAliasFor link when generating snippets, which matches the blockly compiler behavior. currently we just generate snippets that include the non-alias API name, which is usually an underscore API that we don't actually want in the snippet.


abchatra:
We need to hotfix this for Minecraft.

PR #11037, opened 2026-01-08
Make link color meet contrast ratio requirements

srietkerk:
Fixes https://github.com/microsoft/pxt-arcade/issues/7231

Also addresses some other menuitem issues that we ran into in skillmap,

### DIFF
--- a/docs/extensions/naming-conventions.md
+++ b/docs/extensions/naming-conventions.md
@@ -18,6 +18,52 @@ Your extension name (defined in [pxt.json](./pxt-json.md)) must adhere to the fo
 
 In other words, it should match the following regex: `^[a-z][a-z0-9\-_]+`.
 
+### ~ reminder
+
+#### Headings in the Extensions Gallery info pages
+
+To display the information page for an extension (when you click `Learn more` on an extension card in the Extensions Gallery) the Editor will inject portions of the `README.md` file from the extension's repository. A new top-level heading (`h1`), in addition to the one from the `README.md`, is generated for this page. The new heading might contain something like `theowner/pxt-laser-ray 1.2.3`, depending on the repository, owner, and version.
+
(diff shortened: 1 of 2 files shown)


skillmap/src/components/HeaderBar.tsx
```diff
@@ -115,7 +115,7 @@ export class HeaderBarImpl extends React.Component<HeaderBarProps> {
 
     protected getTargetLogo(targetTheme: pxt.AppTheme) {
         const { activityOpen } = this.props;
-        return <div className={`ui item logo brand ${!activityOpen ? "noclick" : ""}`}>
+        return <div className={`ui item logo brand ${!activityOpen ? "noclick" : ""}`} aria-hidden="true">
```
jwunderl: i thought aria-hidden on this would be an accessibility issue, as this div wraps an interactive element with onclick? or am i misremembering / perhaps that doesn't cascade down as i would think.

srietkerk: We don't get flagged for it on the accessibility insights fast pass, but it is a good point. 

With some quick testing, it looks like the target logo is pretty inaccessible as it is. The span makes it so that you can't interact with the text at all with the Windows Narrator at least. When just using tabbing without a screen reader, you also can't interact with the logo text. I'll look into making this actually accessible.

skillmap/src/components/ActivityActions.tsx
```diff
@@ -118,17 +111,13 @@ export class ActivityActionsImpl extends React.Component<ActivityActionsProps> {
                 <Button
                     className="primary inverted"
                     tabIndex={-1}
-                    ariaPosInSet={2}
```
jwunderl, line 121: what's this getting removed for?

srietkerk: It's getting flagged in accessibility insights that buttons are not supposed to have the posInSet attribute.

image

PR #11041, opened 2026-01-09
Tilemap Gallery

jwunderl:
adds support for tilemap galleries re: https://github.com/microsoft/pxt-arcade/issues/6971

I'll leave that open for now instead of closing with this, because no built in tilemap galleries yet; will need to pick some out and plug em in

https://arcade.makecode.com/app/59b405c1b8e281cef0e7901b3588202b5ba29e57-4d8e7ba267

Also fix tooltips for tiles


pxtlib/tilemap.ts
```diff
@@ -1120,7 +1120,7 @@ namespace pxt {
                             meta: {
                                 // For tilemaps, use the id as the display name for backwards compat
                                 displayName: tm.displayName || tm.id,
-                                package: pack.id
+                                package: dep.id
```
jwunderl, line 1123: I don't fully recall this changing anything, as i made this change alongside others while debugging things; it mainly just stood out as it looked wrong, as it was tagging assets sourced from `dep` as being from `pack` in both branches here. if there's something i'm missing can try reverting and double checking.

pxtlib/tilemap.ts
```diff
@@ -1403,11 +1403,21 @@ namespace pxt {
...
-            let newName = defaultName;
-            let index = 0;
 
+            if (!this.isNameTaken(type, defaultName)) {
+                return defaultName;
+            }
+
+            // If ending in digits, continue incrementing that suffix (e.g. "level3" -> "level4")
+            // Otherwise append 1, 2, 3...
+            const matchEndingNumber = /^(.*?)(\d+)$/.exec(defaultName);
```
riknoll: i believe the ? is unnecessary here

webapp/src/components/ImageFieldEditor.tsx
```diff
@@ -246,9 +246,13 @@ export class ImageFieldEditor<U extends pxt.Asset> extends React.Component<Image
         if (options) {
             this.blocksInfo = options.blocksInfo;
 
-            if (options.filter) {
+            const filter = value.type === pxt.AssetType.Tilemap
+                ? options.tilemapFilter
+                : options.filter;
+
+            if (filter) {
```
srietkerk, line 253: I'm just wondering if we should also check that the gallery has content to filter? I know that we're planning on adding tilemaps to be able to filter through, but should the filter button show up at all if there is nothing to fliter?

riknoll (approved):

srietkerk (approved):

webapp/src/components/ImageFieldEditor.tsx
jwunderl: propagating the filter here just affects the editor state, filter button will still be there either way yeah. It's a lil messy without any content (e.g. should the gallery even show up, if there are no default contents, same as filter button), but ideally i'd lean towards "that's not a state we should be in outside of a few days in beta" / easier to keep in mind that it's missing if the button is there & just doing nothing till the content is implemented

PR #11042, opened 2026-01-12
top level 'ask ai' button in code eval tool

jwunderl:
close https://github.com/microsoft/pxt-arcade/issues/7283
close https://github.com/microsoft/pxt-arcade/issues/7290

https://github.com/microsoft/pxt-arcade/pull/7307 adds in categories so that it looks like this instead of just default 'Prompts' header

image

build (noting that the config change above isn't in yet so just the one category)

https://arcade.makecode.com/app/96d2a66e2ef052dcf52c5c824b88ac54fbebf17b-ed6ad58f1b--eval

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -1103,6 +1103,9 @@ function uploadCoreAsync(opts: UploadOptions) {
     let targetFieldEditorsJs = "";
     if (pxt.appTarget.appTheme?.extendFieldEditors)
         targetFieldEditorsJs = "@commitCdnUrl@fieldeditors.js";
+    let targetScriptPageJs = "";
+    if (pxt.appTarget.appTheme?.extendScriptPage)
+        targetScriptPageJs = "@commitCdnUrl@scriptPage.js";
 
     let replacements: Map<string> = {
         "/sim/simulator.html": "@simUrl@",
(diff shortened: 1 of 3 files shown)


teachertool/src/transforms/setAskAiOpen.ts
```diff
@@ -0,0 +1,18 @@
...
+
+    if (open) {
+        // Prevent overlapping overlays.
+        if (teacherTool.catalogOpen) {
+            setCatalogOpen(false);
+        }
+    }
+
+    if (teacherTool.askAiOpen !== open) {
+        dispatch(Actions.setAskAiOpen(open));
```
srietkerk, line 16: Should this dispatch happen even if the modal is open?

jwunderl: the intention here is "dispatch the action if the existing state (`teacherTool.askAiOpen`) is different from the requested state (`open`)"; so if modal is currently open this should only dispatch if request is to make it closed (`open == false`). mainly just mirrored `setCatalagOpen.ts` though

srietkerk: Ohh yeah, that makes sense. I just read this bad. Thanks for the clarification!

PR #11045, opened 2026-01-14
prevent docs HTML sanitizer from removing youtube embeds

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/7200

despite what the linked issue says, this was not mac specific. it also reproes on windows

we configure our docs renderer to take images where the href starts with `youtube:` and replace them with iframes that contain the youtube embedded player. the recent marked changes, however, were causing our dom sanitizer to then remove those iframe embeds from the generated html (which makes sense, it probably should do that). this PR fixes that by changing our markdown renderer to just leave a div in the rendered HTML, and then hydrating the youtube video embed after the fact. this also fixes it for docs pages, which currently don't handle the `youtube:` syntax correctly on live

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pxt-core",
-  "version": "12.2.16",
+  "version": "12.2.17",
   "description": "Microsoft MakeCode provides Blocks / JavaScript / Python tools and editors",
   "keywords": [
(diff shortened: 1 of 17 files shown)


srietkerk (approved):
Nice!

PR #11064, opened 2026-01-22
Add a note/hint to discuss H1 removal of extension title for extension info pages

ganicke:
Is this bug still a thing? If so, please check, correct my note here if I'm confused about why the heading is removed.

Thanks.

Closes https://github.com/microsoft/pxt-microbit/issues/6512

@ssande-fwd

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pxt-core",
-  "version": "12.2.19",
+  "version": "12.2.20",
   "description": "Microsoft MakeCode provides Blocks / JavaScript / Python tools and editors",
   "keywords": [
     "TypeScript",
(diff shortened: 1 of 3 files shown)


riknoll:
i think this is right!

ssande-fwd:
I don't think it's quite right. Here's what I would tell someone to avoid the confusion I had:
"If you don't want the extension information page to show 2 h1's (the autogenerated one and the one from your readme) then ensure that the h1 provided in your readme matches the extension's name in pxt.json."

So at the end when you say "If the names differ, you may see two top-level headings in the extension information page source." it sounds like you are talking about the repo and the pxt.json names, but it's actually the readme h1 and the pxt.json name.

PR #11070, opened 2026-01-28
propagate a query parameter for ai server changes

jwunderl:
To allow testing without full deployments (will be restricted / ignored if not valid)


pxtlib/browserutils.ts
```diff
@@ -1451,6 +1451,21 @@ namespace pxt.BrowserUtils {
...
+        if (typeof window === "undefined") return undefined;
+        const query = pxt.Util.parseQueryString(window.location.search || "");
+        const value = (query["useCopilotServer"] || "").toLowerCase();
+        return value === "ppe" || value === "prod" ? value : undefined;
+    }
+
+    export function appendCopilotServerQueryParam(url: string): string {
+        const value = getCopilotServerParam();
+        if (!value) return url;
+        const params = new URLSearchParams();
```
srietkerk, line 1464: I was wondering for extra security if we should check if the user is logged in at this point? Like if there's no login, don't append the query param?

jwunderl: any security for backend endpoints on the front end is fake; a malicious user will just curl or make the request themselves in the dev tools anyways. no harm in the query parameter existing as the back end is the one that determines whether it is a valid request or not.

(besides that all of our copilot integrations are gated on auth already)

srietkerk (approved):

PR #11071, opened 2026-01-29
Enforce min length of 10 for ask ai questions

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/7325

started roughly from parts of https://github.com/microsoft/pxt-arcade/issues/7325 and reworked till it felt right to me

when a question is in criteria and cut down to below 10 characters:
image
(hovering gives "question: Questions must be at least 10 characters")

when a prompt in custom is empty:
image
when it has some text, but not enough to count 
image


Jaqster:
Looks good.  And clicking on that error triangle also gives the "Questions must be at least 10 characters" message?

jwunderl:
Hovering over it gives that message yes

PR #11072, opened 2026-01-29
move mixer sound synthesis into an audioworklet and add spatial audio

riknoll:
this PR does a few things:
* refactors AudioContextManager into separate files and cleans up some of the audio channel stuff so that more code is shared (there's a new AudioSource base class)
* moves our sound instruction synthesis for arcade from using the hacked together web audio oscillators to generate tones to instead use a new audio worklet (more on this below)
* adds experimental spatial audio support

the new audio worklet allows us to generate PCM data using code closely adapted from [the C++](https://github.com/microsoft/pxt-common-packages/blob/master/libs/mixer/melody.cpp). audio worklet nodes are web audio nodes that spin up their own webworker-like thread that is able to generate sound on the fly, like an oscillator. because this code closely mirrors the C++, our sound is now much more accurate to what the hardware produces. this also fixes a number of edge-case bugs that we had with our old implementation such as the one reported [here](https://forum.makecode.com/t/extension-arcade-sound-instructions/39197/28) and the occasional "note playing backwards" bug you used to get when playing songs from the music editor.

as a bonus, the sound effect editor now displays an actual oscilloscope preview when sounds are playing rather than the fake one we used to show. i might do another PR later to update some of the button icons and block previews so that they more accurately reflect the actual waveform now that users can see it for themselves

spatial audio is also a part of this PR, but needs to be tested more thoroughly before it's surfaced to users. this is mainly intended as an accessibility feature for making games for people with low vision.

finally, there is no change to how sound expressions are played for the micro:bit. in the future, i might move that into the audioworklet as well since it will probably fix some of the choppy playback issues we have on firefox


jwunderl (approved):

docs/static/audioWorklet/audioWorkletProcessor.js
```diff
@@ -0,0 +1,305 @@
...
+            let v = fn(snd, (tonePosition >> 16) & 1023, (tonePosition >> 26) & 0x3f);
+            v = (v * (volume >> 16)) >> (10 + (16 - OUTPUT_BITS));
+            // if (v > MAXVAL)
+            //    target_panic(123);
+            for (const output of dst) {
+                for (const channel of output) {
+                    channel[j] = channel[j] + (v / 0xffff);
+                }
+            }
+            // console.log("v", v);
```
srietkerk, line 243: Will this console log be useful for future testing purposes or should it be removed?

docs/static/audioWorklet/audioWorkletProcessor.js
```diff
@@ -0,0 +1,305 @@
...
+            tonePosition += toneStep;
+            toneStep += toneDelta;
+            volume += volumeStep;
+            samplesLeft--;
+        }
+
+        if (snd.currInstr >= snd.instructions.length) {
+            // snd.sound.state = SoundState::Done;
```
srietkerk, line 251: Should this if-else just be an if of `snd.currInstr < snd.instructions.length` with the else logic? Or is this commented stuff going to be implemented?

pxtsim/sound/audioBufferStreamSource.ts
```diff
@@ -0,0 +1,80 @@
...
+                        buff.copyToChannel(data, 0);
+                    }
+                    else {
+                        const channelBuffer = buff.getChannelData(0);
+                        for (let i = 0; i < data.length; i++) {
+                            channelBuffer[i] = data[i];
+                        }
+                    }
+
+                    // Audio buffer source nodes are supposedly very cheap, so no need to reuse them
```
srietkerk, line 63: Did we use to reuse source nodes?

pxtsim/sound/audioBufferStreamSource.ts
```diff
@@ -0,0 +1,80 @@
...
+                    nodes.push(newNode);
+                    newNode.connect(this.vca);
+                    newNode.buffer = buff;
+                    newNode.addEventListener("ended", () => {
+                        nodes.shift().disconnect();
+                        playNext();
+                    });
+                    newNode.start(nextTime);
+                    nextTime += buff.duration;
+                }
```
srietkerk, line 74: Why does a BufferSource need special dispose logic but a BufferStreamSource not need dispose logic?

pxtsim/sound/audioContextManager.ts
```diff
@@ -0,0 +1,365 @@
...
+            }
+        }
+        return _context;
+    }
+
+    function freshContext(): AudioContext {
+        (<any>window).AudioContext = (<any>window).AudioContext || (<any>window).webkitAudioContext;
+        if ((<any>window).AudioContext) {
+            try {
+                // this call my crash.
```
srietkerk, line 48: Is this supposed to be "this call might crash"?

Also, why wouldn't we want to surface an error like this?

srietkerk: If audio resources are unavailable, does that mean that the user could do something to free up audio resources? Or does it mean that we haven't cleaned processes up properly?

docs/static/audioWorklet/audioWorkletProcessor.js
```diff
@@ -0,0 +1,305 @@
+/**
+ * This file is NOT used at runtime, it only exists for debug. The actual source
```
srietkerk, line 2: Can you elaborate on this more? I don't think I've seen this before/know how this helps with debug.

srietkerk (commented):
In addition to these comments, I was wondering if you could include comments for the different audio source types and what they're used for? Like I know that the audio element source is used for the audio extension on the microbit, but everything else is fuzzy. I feel like the added context would be helpful for debugging audio issues.

docs/static/audioWorklet/audioWorkletProcessor.js
riknoll: the commented stuff is in the C++, i'm leaving it to minimize differences from the C++ code

pxtsim/sound/audioBufferStreamSource.ts
riknoll: no this i just copy pasted

pxtsim/sound/audioBufferStreamSource.ts
riknoll: the created nodes are automatically disposed as they play

pxtsim/sound/audioContextManager.ts
riknoll: this means that the browser doesn't support web audio, which should never happen. also this code isn't changed it's just moved into a new file (spelling mistakes and all)

docs/static/audioWorklet/audioWorkletProcessor.js
riknoll: it's easier to debug when you have a javascript file to point to as opposed to a blob that gets evalled; the browser is able to pick up the source and display it as a normal file. plus you get syntax highlighting in a js file

THEb0nny:
@riknoll I can't upgrade pxt-core because of this error. How can I fix this? Please tell me. I think it has to do with this change.

PR #11078, opened 2026-02-02
adjust error message when too short

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/7364

add a specific error message beyond the red highlight on textbox when evaluate is pressed
image
https://arcade.makecode.com/app/30b54cc31a5894f47e7c87b4cea8a14cadeebd0e-d829a6082b--eval

the bug was asking to block adding on the ask ai modal side, but would rather not do that; it's not how the rest of the app behaves on error states, that just lets them be till evaluate & indicates how to fix them with red highlighting. Preventing adding puts you in a weird spot cause there are multiple things that can be checked; if they have a 5 character long question & 2 of the ones below checked, would have to:
* block adding till question is 10 characters long or unchecked (bad option)
* allow pressing add, but silently drop the short question (also bad, losing data / progress even if rather small)
* allow add but confirmation pop up with "xyz won't get added because it is short, are you sure" (complicated ux side for a problem that the warning already gives indication on resolution for)


teachertool/src/validatorPlanOverrides/runAiQuestionValidationPlanAsync.ts
```diff
@@ -7,14 +7,26 @@ export async function runAiQuestionValidationPlanAsync(
...
+        logDebug(`Skipping AI validation: question below min length (${trimmedQuestion.length}/${Constants.MinAIQuestionLength}).`);
+        return {
+            executionSuccess: false,
+            result: undefined,
+            notes: undefined,
+            executionErrorMsg: Strings.QuestionTooShort,
+        };
+    }
+
+    logDebug(`Asking question: '${trimmedQuestion}' on project with shareId: '${inputs.shareId}'`);
```
srietkerk, line 28: Is this something that we're allowed to log? Could it possibly have PII?

jwunderl: logdebug is console.log / only local

srietkerk (approved):

PR #11094, opened 2026-02-07
Export as PNG and Github

Kikketer:
A simple modification and cleanup to allow saving as PNG and Github at the same time.

More than once I'd had a situation where I keep an application in github for version control, publishing, etc. but then also need the ease of sharing the PNG file for others to import and learn from.

SCR-20260207-jmkp

SCR-20260207-iucc

**Note**: I don't see the whole picture, this modifies the root pxt framework so I would understand if this doesn't jive well with the other dependencies.

**Note Note**: I wasn't able to test this locally due to `localhost` and the github key handling where I'm not able to link github with localhost, I'm sure this path is possible to test but I'm not aware of how.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -220,6 +220,29 @@
 			"outFiles": []
 		},
 		{
+			"name": "pxt serve (core)",
+			"type": "node",
+			"request": "launch",
+			"program": "${workspaceRoot}/built/pxt.js",
+			"stopOnEntry": false,
+			"args": [
+				"serve",
+				"--rebundle",
+				"--noauth"
+			],
+			"cwd": "${workspaceRoot}",
+			"runtimeExecutable": null,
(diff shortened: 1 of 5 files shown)


abchatra:
Thanks @Kikketer  for contribution.

webapp/src/editortoolbar.tsx
```diff
@@ -421,16 +417,6 @@ export class EditorToolbar extends data.Component<ISettingsProps, EditorToolbarS
         const mobile = View.Mobile;
         const computer = View.Computer;
 
-        let downloadButtonClasses = "";
-        let saveButtonClasses = "";
-        if (isSaving) {
```
abchatra, line 426: What is the need for this section of code removal? If I remember correctly this so that kids don't smash bunch of buttons when saving is in progress.

PR #11099, opened 2026-02-10
use stable dist tag when publishing (#11012)

riknoll:
i can't bump arcade v3 because i guess this fix never went into master! i'll need to cherry pick this over to the arcade stable as well...


jwunderl (approved):

abchatra:
Hmm you don't need to bump v3 as it is mainly for docs. For code bumping stable3.0 should be sufficient.

PR #11105, opened 2026-02-12
fix render of youtube links and images in docs pages

riknoll:
fixing a couple of regressions caused by the recent marked/sanitize-html upgrades

we embed youtube videos for youtube links that are alone on a line in a markdown page. the marked upgrade broke this behavior because it was generating slightly different html that tripped up our big macro regex that replaces the generated html with the thumbnail embed

sanitize-html was also stripping all images from the generated html because img isn't in the default list of allowed tags

@ganicke FYI!

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -220,6 +220,29 @@
 			"outFiles": []
 		},
 		{
+			"name": "pxt serve (core)",
+			"type": "node",
+			"request": "launch",
+			"program": "${workspaceRoot}/built/pxt.js",
+			"stopOnEntry": false,
+			"args": [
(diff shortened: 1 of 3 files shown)


.vscode/launch.json
```diff
@@ -219,6 +219,29 @@
 			"sourceMaps": false,
 			"outFiles": []
 		},
+		{
+			"name": "pxt serve (core)",
```
srietkerk, line 223: Are these launch changes intentionally included?

riknoll: yup, we didn't have one for debugging core docs

PR #11119, opened 2026-02-17
Fix blockly dropdown scroll and media not being copied properly

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/7459
fixes https://github.com/microsoft/pxt-arcade/issues/7458

the scroll issue was fixed by editing the overflow css for the dropdown content. this change makes me a little nervous, but i tested all the fields i could think of and they all seem to be working properly

the media issue was caused by gulp trying to read all of the binary files as utf8 encoded. turns out all the sounds were broken too but nobody noticed :)

### DIFF
--- a/theme/debugger.less
+++ b/theme/debugger.less
@@ -12,7 +12,11 @@
 *******************************/
 
 /* Debugger toolbox */
-.debuggerToolbox.elements {
+#debuggerToolbox {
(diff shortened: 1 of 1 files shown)


riknoll:
nah, those are all svgs which are utf8 encoded

PR #11121, opened 2026-02-17
keep the error list collapsed if collapsed by user

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/7461

as the title says, keep track of if the error list was explicitly collapsed by the user and, if so, don't automatically re-expand it when the errors change


abchatra:
Lets hotfix this.

PR #11125, opened 2026-02-18
a few fixes for gallery tilemaps

riknoll:
three fixes:
* my [earlier change](https://github.com/microsoft/pxt/pull/11115) broke the tilemap gallery, so fixing that boolean expression
* fixed a random null pointer exception that i noticed sometimes occurs in new projects. it had to do with the tileset field editor not initializing its asset change listener before the constructor ran
* added some code to prevent gallery tiles from being added to your "my tiles" section when selecting a tilemap from the gallery

the third one is in response to multiple people on the forum thinking this was a bug instead of an intentional choice (and i agree with them)

### DIFF
--- a/webapp/src/components/ImageEditor/ImageEditor.tsx
+++ b/webapp/src/components/ImageEditor/ImageEditor.tsx
@@ -19,6 +19,7 @@ import { Unsubscribe, Action } from 'redux';
 import { createNewImageAsset, getNewInternalID } from '../../assets';
 import { AssetEditorCore } from '../ImageFieldEditor';
 import { classList } from '../../../../react-common/components/util';
+import { BUILTIN_CATEGORIES } from './tilemap/TilePalette';
 
 export const LIGHT_MODE_TRANSPARENT = "#dedede";
 
@@ -220,7 +221,7 @@ export class ImageEditor extends React.Component<ImageEditorProps, ImageEditorSt
         const remappedByBitmap: pxt.Map<pxt.Tile> = {};
 
         for (let i = 0; i < tiles.length; i++) {
-            const tile = tiles[i];
+            let tile = tiles[i];
             if (!tile) continue;
 
             const key = (tile as any).jresData || (tile.bitmap ? pxt.sprite.base64EncodeBitmap(tile.bitmap) : "");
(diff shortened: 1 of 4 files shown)


webapp/src/components/ImageEditor/ImageEditor.tsx
```diff
@@ -250,7 +251,16 @@ export class ImageEditor extends React.Component<ImageEditorProps, ImageEditorSt
                 continue;
             }
 
+            const galleryTile = project.lookupAsset(pxt.AssetType.Tile, tile.id);
+            if (galleryTile) {
+                tile = galleryTile;
+            }
+
             if (tile.bitmap) {
+                if (tile.meta.tags?.some(t => t.startsWith("category-") || BUILTIN_CATEGORIES.some(c => c.id === t))) {
```
srietkerk, line 260: I have reservations about this change. I do agree that there is a bug with duplicates.. If you select a tilemap with dungeon tiles multiple times, it will continue to add stuff to your "my tiles". I also understand that it would be frustrating to get random tiles in "my tiles" while just checking out the gallery for fun. In these ways I agree that there should be a fix. 

However, what if you're a somewhat new user and select a tilemap from the gallery to get started with and eventually want to add or make changes to that tilemap? There is a lot of context a user would have to gather to see the tiles that are getting used and then make the additions.

I'm fine with making this change for now, but I think this needs some rework. Maybe we need another tab or something that has "used tiles" for a tilemap that you're on currently? Or maybe in My Assets we include the tiles from the tilemap itself?

riknoll: used tiles isn't a bad idea, but i don't think you ever want gallery tiles duplicated in my tiles; especially since it essentially buries any tiles you've already added to your project. another fix could be to make the eye dropper tool more prominent (right now it's hidden behind holding down the alt key).

adding tiles to your my tiles section just seems wrong to me; it's like that time apple made everyone download that U2 album onto their ipods

riknoll: regardless, i think it's more annoying than helpful right now

PR #11139, opened 2026-02-23
fix memory leak of audio sources

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/7476

audio sources weren't being properly disposed of after use which caused a bunch of leaked web audio nodes that eventually led to the web audio context crashing


aznhassan (approved):
:shipit:

PR #11140, opened 2026-02-23
tweak focus on dispose logic to work for all blocks

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/7472

this swaps out my other fix for a new one that works for all top-level blocks, not just those with an output connection.

now i just search for whichever top-level block has the closest coordinate instead of using the connection db. this might still cause some scrolling in some instances but probably makes more sense to keyboard users since it will always be a top-level block that gets focused.


riknoll:
> LGTM. I'm curious how Blockly handled this before and why the change.

Blockly didn't have focus management prior to the keyboard control work (which is what introduced this bug)

PR #11141, opened 2026-02-23
remove file size filter for file drag/drop handler

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/6694

apparently this filter has been ignored since the beginning of time but i recently un-ignored it which led to issues on microbit. removing it entirely since it's clearly not needed.


abchatra:
Hotfix please

PR #11143, opened 2026-02-24
fix animation preview not stopping in non chromium browsers

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/7468

for some reason the leave event isn't firing correctly in firefox when you click on the field. this listener on the document acts as a failsafe in case the event is skipped


aznhassan (approved):
LGTM

PR #11152, opened 2026-02-26
Improve screen reader behaviour for area overlay

microbit-robert:
The current area overlay menu is keyboard accessible but the screen reader experience is not so good. This PR makes the following changes to improve the screen reader experience:

- Add roles of menu and menuitem
- Add aria-label for menu
- Add arrow navigation and focus the first menu item by default
- Align the area overlay menu item names with the accessibility documentation


webapp/src/components/AreaMenuOverlay.tsx
```diff
@@ -144,8 +144,8 @@ const areas: Area[] = [
         }
     },
     {
-        id: "editor",
-        ariaLabel: lf("Editor"),
+        id: "workspace",
+        ariaLabel: lf("Workspace"),
```
microbit-robert, line 148: Should this only be "Workspace" for the Blocks case, and "Editor" for the JavaScript/Python case?

riknoll: i think whatever it is should be consistent between experiences. i'm fine with workspace

riknoll (approved):
LGTM. do we have docs anywhere that need to be updated with the new terminology?

microbit-robert:
> LGTM. do we have docs anywhere that need to be updated with the new terminology?

I've aligned with [these MakeCode docs](https://makecode.com/accessibility#navigating-between-areas). I'm not aware of anything else that specifically points out the area menu in the same detailed way.

PR #11183, opened 2026-03-09
disable architecture detection for offline app page

jwunderl:
turning off the suggested platform because it is not working on arm device i have anymore / mac. Better to give no recommendation than guide to wrong one because browser is misleading.

for arcade hotfix put this up as https://github.com/microsoft/pxt/pull/11182


abchatra:
@jwunderl this API is better? https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData/getHighEntropyValues

PR #11189, opened 2026-03-13
Improve toolbox behaviour for keyboard and screen reader users

microbit-robert:
This PR:
- Uses a flat tree structure for the toolbox
    - This fixes the tree item count for screen readers (these tree items are now always in the DOM but visually hidden)
    - As there are no controls to expand or collapse the sub-categories (typically left/right arrow), we don't need levels of nesting or aria-levels. Categories are simply shown visually when navigated to
- Aligns toolbox behaviour with Blockly
    - The selected category is remembered when navigating back to the toolbox. The previous behaviour hindered users when they needed to add more than one block from the same category.
- Enter / Space on a category no longer toggles the flyout open / closed
    - This was observed to be a confusing behaviour for screen reader users
    - Enter / Space still opens the extensions overlay
    - Mouse clicks still toggle the flyout open / closed
- Ensures that using the down arrow from the search input always focuses the first item in the toolbox rather than re-focusing the previously selected category - the latter is what the Tab key does
- Renders the LED category in uppercase to improve screen reader output
    - When capitalised the screen reader outputs "L E D", otherwise "led"
    - LED is typically uppercase for translations, but this is not the case for the English text
- Adds an `id` to the toolbox to ensure that code related to Blockly's focus manager runs (see https://github.com/RaspberryPiFoundation/blockly/pull/9174)


webapp/src/toolbox.tsx
```diff
@@ -1016,7 +1030,7 @@ export class TreeRow extends data.Component<TreeRowProps, {}> {
             iconClass += ' image-icon';
             iconContent = undefined;
         }
-        const rowTitle = name ? name : Util.capitalize(subns || nameid);
+        const rowTitle = name ? name : nameid === 'led' ? nameid.toUpperCase() : Util.capitalize(subns || nameid);
```
riknoll: this seems a little too hacky for me... should we not just rename the category to LED?

webapp/src/toolbox.tsx
```diff
@@ -696,25 +713,24 @@ export class Toolbox extends data.Component<ToolboxProps, ToolboxState> {
...
-                                                ariaLevel={2}
-                                            />
-                                        )}
-                                    </div>
-                                )}
                             </CategoryItem>
+                            {treeRow.subcategories?.map(subTreeRow =>
+                                <CategoryItem
+                                    key={subTreeRow.nameid + subTreeRow.subns}
+                                    className={classList(expandedItem != treeRow.nameid && "sr-only")}
```
riknoll, line 719: where is this sr-only class defined?

webapp/src/toolbox.tsx
```diff
@@ -696,25 +713,24 @@ export class Toolbox extends data.Component<ToolboxProps, ToolboxState> {
...
+                                    index={index++}
+                                    selectedIndex={this.selectedIndex}
+                                    toolbox={this}
+                                    selected={selectedItem == (subTreeRow.nameid + subTreeRow.subns)}
+                                    treeRow={subTreeRow}
+                                    onCategoryClick={this.onCategoryClick}
+                                    ariaLevel={1}
```
riknoll: if everything is now at the same ariaLevel, can we just remove the attribute?

riknoll (changes requested):

webapp/src/toolbox.tsx
microbit-robert: Agreed. This is in a few places which I can remove in favour of name then nameid.

I guess renaming the namespace breaks a lot of things. Adding `block="LED"` to the namespace will do the right thing but break existing translations?

webapp/src/toolbox.tsx
microbit-robert: I had (incorrectly) assumed it was a semantic CSS built-in class. [Here is the definition](https://github.com/microsoft/pxt/blob/master/teachertool/src/components/styling/ActionAnnouncer.module.scss). Should it be moved/copied elsewhere?

webapp/src/toolbox.tsx
microbit-robert: In theory, yes. In practice, it depends on whether screen readers correctly infer the default value of 1. I'll test this.

microbit-robert: Yes, this attribute can be removed. Shall I leave it as an optional prop in the component chain or remove it entirely?

webapp/src/toolbox.tsx
riknoll: yes, let's move it into react-common's css

webapp/src/toolbox.tsx
riknoll: you can leave the optional prop!

webapp/src/toolbox.tsx
riknoll: correct! though i'm guessing most of those translations were just to "LED"

microbit-robert: led special casing removed by [789e46c](https://github.com/microsoft/pxt/pull/11189/commits/789e46c4bc87d3a127a42ebe9b18fa2926f975f7).

PR for adding block="LED" to the namespace https://github.com/microsoft/pxt-microbit/pull/6803

webapp/src/toolbox.tsx
microbit-robert: Moved into react-common css [df02153](https://github.com/microsoft/pxt/pull/11189/commits/df02153332e0b90d0fc6260e42b6c9d93b624e65)

Remove use of removed .module.scss file [3eaadba](https://github.com/microsoft/pxt/pull/11189/commits/3eaadba5ea7b43dcad809eecd2bf7b78fd1a9118)

webapp/src/toolbox.tsx
microbit-robert: [fb5a3a5](https://github.com/microsoft/pxt/pull/11189/commits/fb5a3a557b9871dda0d744055386f8209391b510)

riknoll (approved):

PR #11195, opened 2026-03-16
fixes for example: hash, share button

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/7527 in first commit, app.tsx
fix https://github.com/microsoft/pxt-arcade/issues/7526 by making it give toggle when there are multiple options. projects.tsx

build https://arcade.makecode.com/app/8f4927220a32ed9b2b76acaa7c66850a5ea9e55e-257e84a9fd


webapp/src/projects.tsx
```diff
@@ -943,10 +965,73 @@ export class ProjectsDetail extends data.Component<ProjectsDetailProps, Projects
...
     }
 
-    private getShareableEditors(): pxt.CodeCardEditorType[] {
+    private setShareDialogAction(actionKey: string) {
+        const shareActions = this.getShareableActions();
+        const selectedAction = shareActions.find(action => action.key === actionKey) || shareActions[0];
+        if (!selectedAction) return;
+
+        const editors = this.getShareableEditors(selectedAction.action).filter(e => !!e) as pxt.CodeCardEditorType[];
+        const previousEditor = this.state.shareDialogEditor;
```
srietkerk, line 1362: nit: variable name is confusing. I don't understand how the editor specified in the share dialog is previous.

webapp/src/projects.tsx
```diff
@@ -943,10 +965,73 @@ export class ProjectsDetail extends data.Component<ProjectsDetailProps, Projects
...
+    }
+
+    private buildSharedExampleLink(raw: string): string {
+        return this.getShareableLink({
+            url: raw,
+            cardType: "sharedExample"
+        });
+    }
+
+    private getForumEditorShareLink(): string {
```
srietkerk, line 1380: I don't understand this. How can we take a forum link and turn it into one of our share links? Maybe I'm not understanding something about what a forum editor share link is?

webapp/src/projects.tsx
jwunderl: this is the value for the editor was previously selected, before the action type was changed. If that editor type is not supported for this action type, we have to swap to default editor (editors[0]) for the type.

webapp/src/projects.tsx
jwunderl: this is an existing feature so i'm not certain i follow; the way we do it is in `ensureForumShareLink` below, using discourse api. The `Advanced Livestream` carousel on the home page is defined this way https://github.com/microsoft/pxt-arcade/blob/master/docs/advanced-stream.md, pxtlib/discourse.ts has the implementation

srietkerk (approved):

PR #11196, opened 2026-03-17
Fix race condition between simulator init and project loading

### LINKED ISSUE #6081: Timing-related page load issue that leaves UI in broken state (simulator driver init)
**Describe the bug**
We've spotted what seems to be a load/initialisation timing issue that can leave MakeCode UI in a broken state.

**To Reproduce**
1. Open a new incognito browser
2. Open dev tools and set the network tab data rate to "Slow 4G" (this significantly improves the chance of reproduction)
3. Load https://makecode.microbit.org/#editor (slowly!)
4. Most of the time you should see the UI in the bad state shown below with the following logged error:

```
main.js:1 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'preload')
    at me.internalLoadHeaderAsync (main.js:1:600490)
    at me.loadHeaderAsync (main.js:1:600276)
    at async me.createProjectAsync (main.js:1:625874)
```

[The error is here in the code](https://github.com/microsoft/pxt/blob/ab468923231a7dc03595c74e82d8ec3f29df4497/webapp/src/app.tsx#L1662).

If you need to try again to reproduce then close all incognito browser windows to start with no cache.

We originally saw this issue in micro:bit classroom in the embedded iframe scenario, where you get the same logged error but MakeCode recovers after showing this message:

Image

In the non-iframe scenario above MakeCode does not recover and the user has to reload the page and hope they avoid the issue next time.

**Expected behavior**
MakeCode loads as normal.

**Screenshots**

Image

N/A

 - OS: Mac
 - Browser Chrome
 - Version 132.0.6834.160 (Official Build) (arm64) 

CC @microbit-grace @microbit-robert

microbit-matt-hillsdon:
Gate internalLoadHeaderAsync on componentDidMount completion so the simulator driver is guaranteed to exist before project loading accesses it. Add a defensive null guard in stopSimulator.

Return the newProject() promise from the app init callback chain so errors propagate to the catch handler avoiding a blank screen if this step fails during init.

Fixes https://github.com/microsoft/pxt-microbit/issues/6081

**Note**: You can get a deterministic repro by adding e.g. await new Promise(resolve => setTimeout(resolve, 2_000)); to simulator's initAsync.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -168,6 +168,7 @@ export class ProjectView
     private preserveUndoStack: boolean;
     private rootClasses: string[];
     private pendingImport: pxt.Util.DeferredPromise<void>;
+    private editorMountComplete = pxt.Util.defer<void>();
     private shouldFocusToolbox: boolean;
 
     private themeManager: ThemeManager;
@@ -1202,6 +1203,7 @@ export class ProjectView
         // subscribe to user preference changes (for simulator or non-render subscriptions)
         data.subscribe(this.cloudStatusSubscriber, `${cloud.HEADER_CLOUDSTATE}:*`);
         data.subscribe(this.headerChangeSubscriber, "header:*");
+        this.editorMountComplete.resolve(undefined);
     }
 
     public componentWillUnmount() {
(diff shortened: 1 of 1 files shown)


riknoll (approved):
this stuff is always tricky, so we'll need to really hammer on this while testing, but LGTM

PR #11252, opened 2026-04-06
port over melody parsing from pxt-microbit

riknoll:
the other part of the fix for https://github.com/microsoft/pxt-microbit/issues/6664

copies over the melody parsing from https://github.com/microsoft/pxt-microbit/pull/6762 and uses it to parse the strings passed into the melody editor. should now allow anything that counts as a valid melody in the note range, e.g.

```
A4 B4 C5 D##:4
```

### DIFF
--- a/pxtblocks/fields/field_melodySandbox.ts
+++ b/pxtblocks/fields/field_melodySandbox.ts
@@ -297,14 +297,35 @@ export class FieldCustomMelody<U extends FieldCustomOptions> extends FieldMatrix
         let oldValue: string = value;
         try {
             value = value.slice(1, -1); // remove the boundary quotes
-            value = value.trim(); // remove boundary white space
+            const reader = new MelodyStringReader(value, true);
+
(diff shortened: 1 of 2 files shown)


pxtblocks/fields/field_melodySandbox.ts
```diff
@@ -944,3 +951,119 @@ function getPlaceholderColor(row: number): string {
...
+            const c = this.melody.charCodeAt(this.melodyStringIndex);
+            if (c < 48 || c > 57) break;
+            result = result * 10 + (c - 48);
+            this.melodyStringIndex++;
+        }
+        return result;
+    }
+
+    eatWhitespace() {
+        while (this.melodyStringIndex < this.melody.length && this.melody.charAt(this.melodyStringIndex) == " ") {
```
srietkerk, line 1060: Same question in the other PR about if this should be a `===` or not.

srietkerk (approved):

PR #11253, opened 2026-04-07
fix escaping of xml code snippets

riknoll:
fixes https://github.com/microsoft/pxt/issues/11243
fixes https://github.com/microsoft/pxt-arcade/issues/7549

fixes print. the root cause of this issue was that our markdown renderer wasn't properly escaping embedded xml snippets, which led to them being stripped from the generated XML when we ran the DOM sanitizer over the output


srietkerk (approved):

abchatra:
Lets hotfix this.

PR #11262, opened 2026-04-10
Homescreen search

jwunderl:
https://arcade.makecode.com/app/887b34eb813d38415c4c9fe2f80531d6b7142152-b79eca3302

not 100% sure where searchbox feels good so i put it in two places for demo sake; headerbar when on desktop, and on phone size it pops down to below hero banner. Happy to take any opinions / adjust as needed. Could also be hidden in a search / filter section overall i suppose


abchatra:
I feel like if it is in header it should search through all the projects as well? If not it should be below my projects to indicate it is searching only in the content and not custom projects.

Jaqster:
Yeah, it feels like it's a bit too far away from the content you're searching up at the very top.  If we can put it below the banner, I think that would feel more in-context.  Like next to the Import button?

jwunderl:
Believe all the changes we decided on are in, build for microbit: https://makecode.microbit.org/app/fee3618c1d10b2f220bf24e1fdcb57cda00173d2-36c6677407

search box -> next to import button 
image
pops into this view, focus on search 
image
detail view pops down same as normal
image
(reminds me i should turn on share button flag in microbit, but it's included)
projects included now too (if you set description in pxtjson it does show up on detail view)
image

Jaqster:
LOVE this!!!  Some things to consider for the next pass:
- Should we have some sort of "Search" title or label on the Search page so people know what view they're in?  Like we do for Extensions?
image
image

- Do we need the Import button on the Search page?
- Do we need the grey placeholder card?
image

jwunderl:
Okay should be last build for this pr https://makecode.microbit.org/app/78535ce8a7648a0fc5c1ed2d91b4a9c2dea3b0d9-a3a104ce92


* centered Search header
* hide import while in search
* ensure scroll bar so that we don't resize constantly as results change
* maintain footer so it doesn't feel as empty / scroll bar feels better

webapp/src/projects.tsx
```diff
@@ -63,6 +85,18 @@ export class Projects extends auth.Component<ISettingsProps, ProjectsState> {
         if (this.state.selectedCategory !== prevState.selectedCategory) {
             this.ensureSelectedItemVisible();
         }
+
+        // If search query exists but results were empty when galleries weren't loaded yet, retry once galleries are available
+        if (this.state.searchQuery && this.state.searchQuery === prevState.searchQuery) {
```
riknoll: this worries me... couldn't this end up infinite looping if the query legitimately has no results?

webapp/src/projects.tsx
```diff
@@ -73,6 +107,243 @@ export class Projects extends auth.Component<ISettingsProps, ProjectsState> {
...
+                : header.tutorialCompleted ? header.tutorialCompleted.steps
+                    : undefined;
+        const ghid = pxt.github.parseRepoId(header.githubId);
+
+        return {
+            projectHeader: header,
+            cardType: "file",
+            name: (ghid && pxt.github.join(ghid.project, ghid.fileName)) || header.name,
+            time: header.modificationTime,
+            url: header.pubId && header.pubCurrent ? "/" + header.pubId : "",
```
riknoll: what is this url used for? if it's a user project i assume we always want to open it in the editor?

webapp/src/projects.tsx
```diff
@@ -73,6 +107,243 @@ export class Projects extends auth.Component<ISettingsProps, ProjectsState> {
...
+        if (targetConfig?.searchGalleries)
+            pxt.Util.jsonCopyFrom(galleries, targetConfig.searchGalleries);
+        return galleries;
+    }
+
+    private collectGalleryCards(galleries: pxt.Map<string | pxt.GalleryProps>): SearchCard[] {
+        return this.collectGallerySearchEntries(galleries).cards;
+    }
+
+    private collectGallerySearchEntries(galleries: pxt.Map<string | pxt.GalleryProps>) {
```
riknoll, line 119: should this be cached? maybe in its own getData path? some of our targets do have a lot of cards

webapp/src/projects.tsx
```diff
@@ -73,6 +107,243 @@ export class Projects extends auth.Component<ISettingsProps, ProjectsState> {
...
+
+        compiler.homeSearchAsync({ term: normalized, entries })
+            .then(results => {
+                if (requestId !== this.searchRequestId) return;
+
+                const matches = results
+                    .map(result => cardMap[result.id])
+                    .filter(card => !!card);
+
+                this.setState({
```
riknoll, line 239: if the search takes a long time and they navigate back to the home screen because they're tired of waiting, is this going to snap them back to the search results when it completes?

webapp/src/projects.tsx
```diff
@@ -147,71 +422,159 @@ export class Projects extends auth.Component<ISettingsProps, ProjectsState> {
...
                     </div>
                 </div>
-                <div className="content">
+                {searchMode && <div className="content homescreen-search-box" role="search">
+                    <label className="accessible-hidden" htmlFor="homescreen-search">{lf("Search home content")}</label>
+                    <div className="ui fluid icon input">
+                        <input
+                            id="homescreen-search"
+                            type="text"
+                            placeholder={lf("Search tutorials, games, projects")}
```
riknoll: "games" might be a controversial term in some of our targets. perhaps "examples"?

webapp/src/projects.tsx
```diff
@@ -147,71 +422,159 @@ export class Projects extends auth.Component<ISettingsProps, ProjectsState> {
...
-                            <ProjectsCarousel ref={`${selectedCategory == galleryName ? 'activeCarousel' : ''}`}
-                                key={`${galleryName}_carousel`} parent={this.props.parent}
-                                name={galleryName}
-                                path={url}
-                                onClick={this.chgGallery} setSelected={this.setSelected}
-                                shuffle={shuffle}
-                                selectedIndex={selectedCategory == galleryName ? selectedIndex : undefined} />
+            {searchMode && <div key={`search_gallerysegment`} className="ui segment gallerysegment search-segment" role="region" aria-label={lf("Search results")}>
+                <div className="content search-results-grid" style={{ display: "grid", gridTemplateColumns: "repeat(auto-fill, minmax(220px, 1fr))", gap: "1rem", alignItems: "start", gridAutoFlow: "dense" }}>
+                    {!hasSearchQuery ? <p className="ui grey inverted segment search-empty-state">{lf("Start typing to search tutorials, games, and projects.")}</p>
```
riknoll: same comment on "games"

webapp/src/projects.tsx
```diff
@@ -147,71 +422,159 @@ export class Projects extends auth.Component<ISettingsProps, ProjectsState> {
...
+                                    time={scr.time}
+                                    projectId={scr.projectId}
+                                    tutorialStep={scr.tutorialStep}
+                                    tutorialLength={scr.tutorialLength}
+                                    selected={!scr.directOpen ? searchSelectedIndex === index : undefined}
+                                />
+                                {selectedSearchCard && searchSelectedIndex === index && <div ref="searchDetailView" className="detailview search-detailview" style={{ gridColumn: "1 / -1", minWidth: 0 }}>
+                                    <sui.CloseButton onClick={() => this.setSelected(SEARCH_CATEGORY, undefined)} />
+                                    <ProjectsDetail parent={this.props.parent}
+                                        name={selectedSearchCard.name}
```
riknoll, line 480: could maybe clean up a lot of these props with a `{ ...selectedSearchCard }`. same for the one directly above this

webapp/src/projects.tsx
```diff
@@ -147,71 +422,159 @@ export class Projects extends auth.Component<ISettingsProps, ProjectsState> {
...
+                                    scr={scr.projectHeader || scr} index={index}
+                                    onCardClick={this.handleSearchCardClick}
+                                    cardType={scr.cardType}
+                                    time={scr.time}
+                                    projectId={scr.projectId}
+                                    tutorialStep={scr.tutorialStep}
+                                    tutorialLength={scr.tutorialLength}
+                                    selected={!scr.directOpen ? searchSelectedIndex === index : undefined}
+                                />
+                                {selectedSearchCard && searchSelectedIndex === index && <div ref="searchDetailView" className="detailview search-detailview" style={{ gridColumn: "1 / -1", minWidth: 0 }}>
```
riknoll: move inline styles to the css class you're using there?

webapp/src/projects.tsx
```diff
@@ -147,71 +422,159 @@ export class Projects extends auth.Component<ISettingsProps, ProjectsState> {
...
+                            />}
+                        {canImport ?
                             <sui.Button key="import" icon="upload" className="import-dialog-btn neutral" textClass="landscape only" text={lf("Import")} title={lf("Import a project")} onClick={this.importProject} /> : undefined}
                     </div>
                 </div>
-                <div className="content">
+                {searchMode && <div className="content homescreen-search-box" role="search">
+                    <label className="accessible-hidden" htmlFor="homescreen-search">{lf("Search home content")}</label>
+                    <div className="ui fluid icon input">
+                        <input
```
riknoll: is there a reason not to use the react common input here?

webapp/src/projects.tsx
```diff
@@ -147,71 +422,159 @@ export class Projects extends auth.Component<ISettingsProps, ProjectsState> {
...
-                    >
-                        {scriptManager ? <h2 className="ui header myproject-header">
-                            {lf("My Projects")}
-                            <span className="view-all-button" tabIndex={0} title={lf("View all projects")} role="button">
-                                {lf("View All")}
-                            </span>
-                        </h2> : <h2 className="ui header">{lf("My Projects")}</h2>}
+                    <div className="column gallery-heading-column">
+                        {searchMode ?
+                            <sui.Button
```
riknoll: react common button?

riknoll (changes requested):

webapp/src/projects.tsx
jwunderl: had this one here mainly for when the search input was always there, since it's behind a button / things will already be in process of loading when search occurs, so just removed for now

webapp/src/projects.tsx
jwunderl: cut this one, had put it there to check while trying to debug something but realized it was a missed recompile and forgot to remove it

webapp/src/projects.tsx
jwunderl: It's ~instant ish in arcade for now even on phone so leaving this one, but will cache later if it we see it degraded at all

webapp/src/projects.tsx
jwunderl: fixed but since it was mildly funny; issue was that grid-column property wasn't working in less when i swapped it over and i was missing what was wrong, the way it's written in css `1 / -1` gets parsed in less to just `-1`. just swapped to listing -start and -end explicitly instead of semantic sugar version since I didn't immediately recall the syntax for skipping less pre-mathing it

webapp/src/projects.tsx
jwunderl: it'll bail out above on the id check, closing search increments that https://github.com/microsoft/pxt/blob/dev/jwunderl/homescreen-search/webapp/src/projects.tsx#L286 (added a comment to indicate that for future though)

riknoll (approved):

PR #11274, opened 2026-04-17
indicate which provider last successfully signed in

jwunderl:
We had a user that thought they lost all their projects, but they just signed in to microsoft with their gmail account instead of into gmail (or vice versa sounds like). Store a small bit so we can flag which one the user last used to nudge them in the right direction next time they try to sign in (not too attached to ui just mirrored from homescreen flags):

image

image


abchatra:
Nice!

srietkerk (approved):
Nice!

PR #11275, opened 2026-04-20
Improve the simulator live region

microbit-robert:
The live region is re-created if required. There is code in the sim that can clear the DOM.
The same message can be announced more than once with the addition (toggling) of a non-breaking space.


pxtsim/accessibility.ts
```diff
@@ -73,27 +73,27 @@ namespace pxsim.accessibility {
...
-            liveRegion.setAttribute("aria-hidden", "false");
             liveRegion.setAttribute("style", style);
             document.body.appendChild(liveRegion);
         }
 
-        if (liveRegion.textContent !== value) {
-            liveRegion.textContent = value;
+        if (value) {
+            /* eslint-disable @microsoft/sdl/no-inner-html */
+            liveRegion.innerHTML = `<p>${value}${addNonBreakingSpace ? '&nbsp;' : ''}</p>`;
```
riknoll: let's not use innerhtml here.. are we certain that everything being passed to this function is sanitized? if someone were to, say, pass a malicious string with html into our localization pipeline somehow would they be able to inject some code?

riknoll (changes requested):

pxtsim/accessibility.ts
microbit-robert: Makes sense. This uses textContent and unicode to do the same thing. [1cf885c](https://github.com/microsoft/pxt/pull/11275/commits/1cf885cdb7c688ef457ab689007a28229124e22d)

Tested on VoiceOver, NVDA, JAWS, Windows Narrator and ChromeVox.

riknoll (approved):

PR #11281, opened 2026-04-21
remove ancient patch floating blocks code

### LINKED ISSUE #6776: Unintentional migration of blocks into "on start"
Over the last six months the Foundation have been publishing more MakeCode projects with intentionally disabled blocks on the workspace that are intended to be assembled by the user ("Parson's problems"). We've recently noticed an issue where the disabled blocks are sometimes unexpectedly swept up into an "on start" block that wasn't in the program as originally written.

If you [open this project](https://makecode.microbit.org/#pub:94809-00860-64580-80106) you can see the disabled->enabled transition visually. If you examine the main.blocks XML [in the text part of the same shared project](https://makecode.microbit.org/api/94809-00860-64580-80106/text) you can see there is no "on start" block and blocks like "clear screen" are disabled and top-level (they have a disabled-reasons attribute).

I think this affects projects that:
- were created or edited in a version of pxt-microbit >= v8.0.0 but earlier than the current targetVersion
- that lack an "on start"
- that have disabled blocks on the workspace

This is true of many of our Parson's problem projects due to timing of their creation and because we don't include "on start" if it's not part of the puzzle.

As this affects recently produced content being used in live Foundation projects (6 projects & 17 lessons) it would be great if we can hotfix this.

## Direct repro

This reproduces what typically happens due to the passage of time / MakeCode releases.

1. Load https://makecode.microbit.org/v8.0.0
2. Create a new project
3. Delete "on start"
4. Drag a "show number" block on to the workspace so it's disabled
5. Download a hex
6. Load the hex in latest MakeCode and see "show number" unexpectedly in "on start"

## Analysis

Since last year's Blockly upgrade, Blockly serializes disabled blocks with a "disabled-reasons" attribute instead of "disabled".

There's [a migration in patchFloatingBlocks](https://github.com/microsoft/pxt/blob/66bf2eb9c017997386ec8b7ad6560720e300e7fd/pxtblocks/importer.ts#L205) that checks node.getAttribute("disabled"), but newer projects and those that have been re-saved will have blocks with disabled-reasons not disabled. So the check returns true and we migrate those blocks into a new "on start" block.

Options:
- add "disabled-reasons" check to the migration too
- gate the migrate on a version threshold (presumably this is targetting only very old projects)

It feels like either of these would be pretty simple/safe changes.

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/6776

removes the code for patching ancient scripts that existed prior to the existence of the on start block, back when we allowed you to put code anywhere on the workspace. this was also prior to us having the share feature so i think it's very unlikely that any of these scripts are floating around in the wild.

if someone had a hex file that was made before this time (10 years ago), then they could possibly run into issues importing it into makecode, but i think it's just as likely to fail for many other reasons given how different everything is now! they can always import it into one of the older versions of makecode like v0

### DIFF
--- a/pxtblocks/importer.ts
+++ b/pxtblocks/importer.ts
@@ -221,7 +221,7 @@ function patchFloatingBlocks(dom: Element, info: pxtc.BlocksInfo) {
         const nextNode = node.nextElementSibling;
         // does this block is disable or have s nested statement block?
         const nodeType = node.getAttribute("type");
-        if (!node.getAttribute("disabled") && !node.getElementsByTagName("statement").length
+        if (!(node.hasAttribute("disabled") || node.hasAttribute("disabled-reasons")) && !node.getElementsByTagName("statement").length
             && (buildinBlockStatements[nodeType] ||
                 (blocks[nodeType] && blocks[nodeType].retType == "void" && !hasArrowFunction(blocks[nodeType])))
         ) {
--- a/pxtblocks/importer.ts
+++ b/pxtblocks/importer.ts
@@ -202,60 +202,6 @@ export function loadWorkspaceXml(xml: string, skipReport = false, opts?: DomToWo
     }
 }
 
-function patchFloatingBlocks(dom: Element, info: pxtc.BlocksInfo) {
-    const onstarts = getBlocksWithType(dom, ts.pxtc.ON_START_TYPE);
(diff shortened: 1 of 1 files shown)


abchatra (approved):

abchatra:
Lets hotfix as well.

jwunderl (approved):

PR #11292, opened 2026-05-04
add a new color picker block

riknoll:
adds a new color picker block! the motivation here is that we've always had a lot of targets/extensions that define multiple blocks for getting colors in various color spaces. for example, the neopixel extension in micro:bit, the circuit playground editor, pxt-ev3 (for the home button color), some of the boards in pxt-maker, the color fading extension in arcade, etc.

as a result, we have a lot of duplicated color space conversion code in our many repos. given how universal this is to all of our targets, this PR aims to create one color picker block to rule them all!

the new block natively supports all of the usual color formats:
* RGB
* HSV
* HSL
* CMYK
* HEX strings

internally, the color is stored as HSV in a mutation on the block. the reason for this is that the HSV and HSL color spaces have a lot of points that map to the same color in the RGB derived color spaces, so if you compile down to RGB then you get a lot of jumping around for the various HSV and HSL channel values as the hue changes.

on hardware the colors are all actually stored as 24 bit RGB numbers which i believe is universally how we do it in all of our targets. there are new functions on the `colorHelpers` namespace for converting the various formats to 24 bit RGB.

the UI for the fields is a barebones HSV color picker:

color-picker2

like pauseuntil, this block is optional. unlike pauseuntil, we have a lot of targets where this block probably won't need to be in the default categories that come with the editor. for this reason, i had to a add a new scheme that allows extensions to contribute builtin blocks to the toolbox. in order to have this block appear in the toolbox, you simply need to define a function that has the `builtinBlockId="makecode_color_picker"` comment annotation like so:

```ts
    //% block
    //% builtinBlockId="makecode_color_picker"
    //% color=#0fbc11
    export function __colorPicker(value: number): number {
        return 0
    }
```

the color parameter is also necessary to make the block match the color of whatever category contains it. adding this will also add monaco toolbox entries for all the various colorHelper functions.

right now the `builtinBlockId` annotation only supports the color picker, but i plan to add support for pause until and other blocks in the future. the first place this block is going to be used is in the color fading extension in arcade (i have some devious plans to completely overhaul the APIs in that extension)


THEb0nny:
This is cool! I needed this earlier

THEb0nny:
I think the circle is enough to tell you which color I chose. I think the colored square on the left is unnecessary...

libs/pxt-common/pxt-helpers.ts
```diff
@@ -753,3 +753,170 @@ namespace __internal {
...
+        hue = Math.clamp(0, 360, hue) / 60;
+        saturation = Math.clamp(0, 100, saturation) / 100;
+        value = Math.clamp(0, 100, value) / 100;
+
+        const hi = Math.floor(hue) % 6;
+
+        const f = hue - Math.floor(hue);
+        const p = 255 * value * (1 - saturation);
+        const q = 255 * value * (1 - (saturation * f));
+        const t = 255 * value * (1 - (saturation * (1 - f)));
```
srietkerk, line 793: Do these letters mean something for hsv color specifications or are they just random letters for variable names?

pxtblocks/toolbox.ts
```diff
@@ -575,4 +579,62 @@ export function createFunctionsFlyoutCategory(workspace: Blockly.WorkspaceSvg) {
...
+        //         <shadow type="makecode_color_picker_number">
+        //             <field name="NUM">255</field>
+        //         </shadow>
+        //     </value>
+        //     <value name="INPUT2">
+        //         <shadow type="makecode_color_picker_number">
+        //             <field name="NUM">0</field>
+        //         </shadow>
+        //     </value>
+        // </block>
```
srietkerk, line 607: Is this comment here to demonstrate what the create elements below actually generates?

pxtblocks/plugins/colorpicker/colorPickerNumberField.ts
```diff
@@ -0,0 +1,152 @@
...
+
+        if (!colorPickerBlock) {
+            return;
+        }
+
+        if (!colorPickerBlock.colorHSVLoaded) {
+            colorPickerBlock.readColorFromInputs();
+        }
+
+        let coorHSV = [0, 0, 0];
```
srietkerk, line 100: Is `coor` short for coordinates or should this be `color`?

pxtblocks/plugins/colorpicker/colorPickerWidget.ts
```diff
@@ -0,0 +1,436 @@
...
+
+        this.canvasContainer.appendChild(this.saturationValueHandle);
+
+        container.style.width = "308px";
+        container.style.overflowX = "hidden";
+        container.style.overflowY = "hidden";
+
+        const focus = () => {
+            // In firefox, stealing focus from the range input interrupts
+            // the dragging of the slider
```
srietkerk, line 101: Does this mean Firefox handles focus changing by itself? Will there be problems with keyboard nav/screen readers in Firefox if we don't do this?

pxtblocks/plugins/colorpicker/colorPickerWidget.ts
```diff
@@ -0,0 +1,436 @@
...
+            if (!pxt.BrowserUtils.isFirefox()) {
+                if (!this.isKeyboardControlActive()) {
+                    this.focusHtmlInput();
+                }
+            }
+        };
+
+        // Configure event handler.
+        for (let i = 0; i < 3; i++) {
```
srietkerk, line 110: Is it possible to make this more adjustable? If there's a scenario where we want to add another slider to the color picker widget, we would have to change this and the slider list below.

libs/pxt-common/pxt-helpers.ts
riknoll: these are from the source code of the package i took the color conversions from, i left them unchanged. i have no idea if they refer to anything real or are just arbitrary

pxtblocks/toolbox.ts
riknoll: yep, we do that in most places where generate complex xml in the codebase

pxtblocks/plugins/colorpicker/colorPickerNumberField.ts
riknoll: should be color

pxtblocks/plugins/colorpicker/colorPickerWidget.ts
riknoll: this isn't new code; i copied it from the slider field which has already had a lot of accessibility work done on it. originally i just wanted to extend the slider field so that more of this code could be shared but it ended up being too cumbersome

pxtblocks/plugins/colorpicker/colorPickerWidget.ts
riknoll: i don't think that scenario exists; the HSV color space only has three values

riknoll:
@THEb0nny i'm just copying what most color pickers do for the preview! plus i think having a nice big square makes it way easier to see changes take effect

srietkerk (approved):

PR #11293, opened 2026-05-05
fix localization of toolbox subcategories and groups for staticpkg

riknoll:
fixes the issue reported [here](https://forum.makecode.com/t/group-strings-in-toolbox-not-applied-during-translation/43835) and [here](https://github.com/orgs/pxt-ev3-community/discussions/3#discussioncomment-16808881)

the issue was that we copy over the group/subcategory strings when we download the localizations for a package, but in staticpkg targets we download all of those localized strings from bundled-strings.json which doesn't go down the same code path. i've refactored the copying so that it is now invoked in both places.

@THEb0nny @PetteriAimonen can you try testing this out and let me know if it fixes the issue? i'm not really set up to test the EV3 staticpkg target locally

### DIFF
--- a/pxtlib/package.ts
+++ b/pxtlib/package.ts
@@ -1070,18 +1070,7 @@ namespace pxt {
                             })
                     })))
                 .then(() => {
-                    // Subcategories and groups are translated in their respective package, but are not really APIs so
-                    // there's no way for the translation to be saved with a block. To work around this, we copy the
-                    // translations to the editor translations.
-                    const strings = U.getLocalizedStrings();
-                    Object.keys(loc).forEach((l) => {
-                        if (U.startsWith(l, "{id:subcategory}") || U.startsWith(l, "{id:group}")) {
-                            if (!strings[l]) {
-                                strings[l] = loc[l];
-                            }
-                        }
-                    });
-                    U.setLocalizedStrings(strings);
+                    U.copySubcategoryStrings(loc);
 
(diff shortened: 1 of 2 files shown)


PetteriAimonen:
Thanks! The fix makes sense, now I'll just need to figure out how to build pxt-ev3 against the updated pxt. I'm very new to the inner workings of the pxt system :)

PetteriAimonen:
Sorry, I couldn't get `npm run build` for pxt to succeed tonight, presumably for reasons unrelated to the actual pull request. Just getting silly errors like `Module not found: Error: Can't resolve './Fonts.css' in '.../pxt/pxt/kiosk/src'`. I'll try to find time to try again tomorrow.

riknoll:
> Sorry, I couldn't get `npm run build` for pxt to succeed tonight, presumably for reasons unrelated to the actual pull request. Just getting silly errors like `Module not found: Error: Can't resolve './Fonts.css' in '.../pxt/pxt/kiosk/src'`. I'll try to find time to try again tomorrow.

@PetteriAimonen if `npm run build` gives you too much trouble, you can also give `gulp -n` a shot instead. that'll skip the building of all the subapps (like kiosk) and just build the core pxt stuff

jwunderl (approved):

PetteriAimonen:
Thanks, after a clean checkout I got it to build.

I confirm that the toolbox translation bug is fixed by this pull request.

PR #11296, opened 2026-05-07
extend matrix field to support editing sprites

riknoll:
for a long time we've been toying around with the idea of having an "arcade junior" for lower grades. one problem we've always run into, however, is our asset editors. they have a few features that make it harder for younger students to use them:
1. they're complicated with lots of features
2. they open in a modal, which breaks context while editing programs
3. the previews they show are tiny so they don't show up well in printed materials/screenshots
4. keyboard navigation is complicated

fast forward to last week, where someone sent me a link to the [diversibit](https://www.diversibit.org/en) which got me thinking about how hard it would be to take the matrix field editor that we use in micro:bit and make a version of it for a grid of neopixels, and i realized this would be a great way to kill two birds with one stone!

the matrix field pretty much solves all the arcade junior issues:

1. it's way simpler
2. it's inline in the code
3. the matrix is nice and big on the block
4. it already supports keyboard navigation

with the obvious downside that you can't resize the sprite, but that's probably a good restriction for arcade junior anyhow.

this PR takes our LED matrix field and adds optional support for color, plus some parameters that let you customize the appearance to make it more amenable to pixel grids (removing the corner radius, spacing between cells, etc.). the way the color switching works is through a second field which gets added before the matrix field on the block; this field just pops up a menu of color options that change the "on" color of the matrix. all the other keyboard/mouse interactions are the same: clicking a "lit" cell turns it off and clicking an "unlit" cell turns it on with the selected color.

as for the palette of colors, you can either take the project palette (for arcade) or specify the colors using a comment attribute (for diversibit/other neopixel grids). i briefly considered just using the project palette for the other case too, but realized that it might end up conflicting with the palette used by the display shield.

here's me messing around with keyboard controls:

inline-sprites

i'm probably going to have a follow up pr that adds support for tagged template literals with the field (right now it's always strings). i'm still trying to figure out how exactly to have the decompilation for that coexist with the regular image editor

### DIFF
--- a/pxtblocks/compiler/compiler.ts
+++ b/pxtblocks/compiler/compiler.ts
@@ -804,7 +804,7 @@ function compileImage(e: Environment, b: Blockly.Block, frames: number, columns:
         for (let j = 0; j < columns; ++j) {
             if (j > 0)
                 state += ' ';
-            state += (leds[(i * columns) + j] === '#') ? "#" : ".";
+            state += leds[(i * columns) + j];
         }
         state += '\n';
     }
--- a/pxtblocks/fields/field_ledmatrix.ts
+++ b/pxtblocks/fields/field_ledmatrix.ts
@@ -2,6 +2,7 @@
 /// <reference path="../../built/pxtsim.d.ts" />
 
 import * as Blockly from "blockly";
+import { DEFAULT_LED_COLORS } from "./field_ledmatrix_colorPicker";
 import { FieldMatrix } from "./field_matrix";
 import { FieldCustom } from "./field_utils";
 
@@ -27,9 +28,12 @@ export class FieldLedMatrix extends FieldMatrix implements FieldCustom {
     public SERIALIZABLE = true;
 
     private params: any;
-    private onColor = "#FFFFFF";
-    private offColor: string;
+
+    private palette: string[];
+
     private static DEFAULT_OFF_COLOR = "#000000";
+    private static DEFAULT_ON_COLOR = "#FFFFFF";
+    private offOpacity = 0.2;
 
     private scale = 1;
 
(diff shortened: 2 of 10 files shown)


pxtblocks/compiler/compiler.ts
```diff
@@ -804,7 +804,7 @@ function compileImage(e: Environment, b: Blockly.Block, frames: number, columns:
         for (let j = 0; j < columns; ++j) {
             if (j > 0)
                 state += ' ';
-            state += (leds[(i * columns) + j] === '#') ? "#" : ".";
```
srietkerk, line 812: Will getting rid of this cause problems?

pxtblocks/fields/field_ledmatrix.ts
```diff
@@ -372,12 +419,13 @@ export class FieldLedMatrix extends FieldMatrix implements FieldCustom {
 
     // Composes the state into a string an updates the field's state
     private updateValue() {
+        const chars = ".#23456789ABCDEFGHIJKLMNOPQRSTUVWXYZ";
```
srietkerk, line 422: Is there a reason there is no "1" in this string?

riknoll: traditionally, the led matrix field uses "." for 0 and "#" for 1

pxtblocks/compiler/compiler.ts
riknoll: nope, if you look at the code the only possible values in the past were "#" and ".", so it was essentially doing this:

```
if (char === "#") {
    return "#"
}
if (char === ".") {
    return "."
}
// never
```

which didn't make much sense

srietkerk (approved):

PR #11301, opened 2026-05-08
Add screen reader support for the melody field editor

microbit-robert:
This PR adds screen reader support for the melody field editor

General:
- Ensures that tab toggle (editor | gallery) is accessible
- Add variables for div ids to reduce repetition and fragility 
- Fix a bug where after selecting a melody from the gallery, focusing the grid would restore focus to the previously focused gridcell. This could be a note that is switched off in a column that has a different note switched on which leads to unpredictable behaviour.

Editor:
- Add an aria-label to the grid (field_matrix) explaining what it is
- Add note names to the grid cells, respect note on/off state and provide feedback via aria-live when they are toggled on and off
    - The changes here avoid using aria-labels inside of SVG elements with role="gridcell" as this causes screen readers to lag (e.g., as you are navigating, perhaps every other cell is output by the screen reader). This has required more change to the field_matrix than desired.
- Press 'p' to play the selected note without having to toggle it on/off
- The edges of the grid are hard stops - left / right no longer wrap
- Add a reactive aria-label to the play button

Gallery:
- Change the gallery markup so it is a grid. This ensures that the preview buttons are discoverable by screen reader users
- Fix the aria-activedescendant values which were wrong and causing the screen reader cursor to detach from the focus visible outlines causing confusion

### DIFF
--- a/pxtblocks/fields/field_matrix.ts
+++ b/pxtblocks/fields/field_matrix.ts
@@ -199,7 +199,7 @@ export abstract class FieldMatrix extends Blockly.Field {
         this.clearFocusIndicator();
         const focusVisible = this.matrixSvg.matches(":focus-visible");
         if (!focusVisible && !this.forceFocusVisible) return;
-        const cellRect = cell.querySelector('rect');
+        const cellRect = cell.children[1];
         const cellWidth = parseInt(cellRect.getAttribute("width"))
         const cornerRadius = parseInt(cellRect.getAttribute("rx"));
 
@@ -225,7 +225,7 @@ export abstract class FieldMatrix extends Blockly.Field {
                 "aria-hidden": "true"
             });
         }
-        const cellTextEl = cell.querySelector("text");
+        const cellTextEl = cell.firstElementChild;
         // Don't take effect for initial gridcell focus.
         // Only announce gridcells being toggled on/off, not navigated to.
         setTimeout(() => cellTextEl.setAttribute("aria-live", "polite"), 0);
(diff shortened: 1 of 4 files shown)


pxtblocks/fields/field_matrix.ts
```diff
@@ -187,40 +195,49 @@ export abstract class FieldMatrix extends Blockly.Field {
...
 
-    private setFocusIndicator(cell: SVGRectElement, useTwoToneFocusIndicator: boolean) {
+    private setFocusIndicator(cell: SVGElement, useTwoToneFocusIndicator: boolean) {
         this.clearFocusIndicator();
         const focusVisible = this.matrixSvg.matches(":focus-visible");
         if (!focusVisible && !this.forceFocusVisible) return;
-        const cellG = cell.parentNode as SVGRectElement;
-        const cellWidth = parseInt(cell.getAttribute("width"))
-        const cornerRadius = parseInt(cell.getAttribute("rx"));
+        const cellRect = cell.querySelector('rect');
```
riknoll: i think `cell.firstElementChild` should work here? best to avoid query selectors since they tend to be slower (though probably doesn't matter here)

pxtblocks/fields/field_melodySandbox.ts
```diff
@@ -487,14 +510,18 @@ export class FieldCustomMelody<U extends FieldCustomOptions> extends FieldMatrix
 
             for (let col = 0; col < this.numMatrixCols; col++) {
                 const cell = this.cells[col][row];
+                const cellRect = cell.querySelector('rect');
```
riknoll: again, prefer not to use query selector

riknoll (commented):
minor nitpicks, otherwise LGTM

pxtblocks/fields/field_matrix.ts
microbit-robert: `cell.firstElementChild` is the screen reader only text element so I've replaced querySelector with a mixture of `cell.firstElementChild` for the text element and `cell.children[1]` for the rect element [9a1af72](https://github.com/microsoft/pxt/pull/11301/commits/9a1af72b539ef49373bbeaa6b1d908fe5fad0216)

riknoll (approved):

PR #11321, opened 2026-05-19
Upgrade Blockly to v13.0.0-beta.5

### LINKED ISSUE #6870: Upgrade blockly to v13.0.0
From Blockly team: 

_We've just released another beta of v13 that is feature-complete, and we're working on ironing out the last few reported bugs. We expect to release the full version of v13.0.0 this week, but we do have some "fast follow" fixes planned for a v13.1.0 asap with additional features and bug fixes that aren't breaking changes. v13.0.0 will be ready to integrate with MakeCode when it's released, and v13.1.0 will include additional polish (such as improved screenreader announcements for some blocks and fixes for more obscure bugs that aren't show-stoppers) that will be easy to update to once your update to v13 is complete._

microbit-matt-hillsdon:
Demo deployment: https://blockly-v13-upgrade.review-pxt.pages.dev/#editor

Bump from 12.3.1 and drop the @blockly/keyboard-navigation plugin (now built into Blockly core).

User-facing changes:
- Keyboard navigation and screen-reader support are always enabled. Removed the accessible-blocks preference and its toggle UI. The heavier focus outlines for toolbox/workspace are enabled when using a meaningful keyboard navigation (previously matched setting).
- Aligned pxt's cut/copy/paste with copy/cut toasts and the cut audio cue.
- Variable creation focuses the new variable in the flyout (v13 behaviour); manual focus restore to the button removed.
- Audio cues when keyboard navigation reaches limits and when a block is dropped on the workspace from move mode or drag.
- Context menu changes were intentionally not ported from the keyboard navigation plugin. We could consider adding Cut alongside Copy/Paste, Copy/Paste are also not Blockly default.
- Mac uses Cmd not Option as the unconstrained movement modifier.
- Left and right arrow keys no longer wrap based on screen reader user feedback
- Added an "end" position at the end of statement inputs. This can be useful for block insertion but is mainly motivated by helping screen reader users understand hierarchy in a way that's closer to text-based coding and more usable than a full tree structure.

API:
- Custom drag strategy collapsed from a substantial copy of Blockly's internals thin subclass due as the base is now exported.
- Event handler shapes, navigator/focus manager flows, bubble interfaces and widget-div positioning all updated to match v13.
- Shortcut formatting is now a tracked port of Blockly's internal helper (not on the public API, previously this was copied from the nav plugin).

Workarounds for regressions in pxt customisations:
- Preserve duplicate-shadow-on-drag (e.g. melody). Patched at the gesture and drag-strategy level. Previously this only worked because of an odd behaviour in Blockly where the shadow was briefly focused (you can actually see this in live MakeCode when clicking the melody shadow)
- Snap dragged blocks back to the pointer at the snap-radius boundary (see https://github.com/RaspberryPiFoundation/blockly/issues/9898).
- Re-stack the connection-preview indicator over the highlight path, fix the indicators for keyboard nav.

CSS:
- Blockly v13 injects its CSS via adoptedStyleSheets, flipping the cascade. We've generally added body to selectors to address this.
    - The cascade change is currently under investigation to see if we can do better here in Blockly.
- Related, adoptedStyleSheets requires Safari 16, last year we agreed 15, Blockly are up for a fix that's backwards compatible (will help with some iPads that still get 15.x security fixes) and micro:bit are taking a look here.
    - I opened https://github.com/RaspberryPiFoundation/blockly/pull/9908 as a potential fix for both issues.

Translation:
- New aria labels, move-mode announcements, and screen-reader-mode toggle inherited from Blockly v13.
    - Note the "command" language will change to "statement" in the next Blockly release and we'll sync with that then.

Known regressions vs Blockly 12 + keyboard nav plugin:
- Glitch when dragging affecting mouse users - https://github.com/RaspberryPiFoundation/blockly/issues/9898 
     - Fixed in v13 branch
- Initial block placement isn't intelligent - https://github.com/RaspberryPiFoundation/blockly/issues/9878
     - Fixed in v13 branch
- Focused inputs not always respected for keyboard block insert - https://github.com/RaspberryPiFoundation/blockly/issues/9899 (Newly raised)

Follow ups will be needed for better screen reader integration and to update the keyboard controls help.

Closes https://github.com/microsoft/pxt-microbit/issues/6870


microbit-matt-hillsdon:
@riknoll FYI, we'll continue to test this and update it with Blockly beta releases.

I'm away next week so @microbit-robert will pick up as needed.

microbit-matt-hillsdon:
Discussed with @abchatra on today's dev sync. @riknoll, we think this is potentially mergeable (with future PRs upgrading to later betas/v13.0.0). But understandable to wait for the fixes to the regressions noted depending on your view of the impact. The next Blockly release will likely tweak strings too.

Either way @riknoll we'd really appreciate some testing and review of this PR.

pxtblocks/loader.ts
```diff
@@ -787,6 +784,132 @@ function initComments() {
     Blockly.Msg.WORKSPACE_COMMENT_DEFAULT_TEXT = '';
 }
 
+function initAccessibilityMessages() {
+    // Translatable overrides for Blockly's built-in keyboard-navigation strings.
+    // Excludes text used only in the shortcut dialog that we don't use.
+    Object.assign(Blockly.Msg, {
```
riknoll, line 790: this looks good, but i wonder if we should provide some context in some of these strings for the benefit of translators. a lot of these are probably safe, but several of these have ambiguous definitions that the translators might not pick up on (e.g. "Checked" could mean "Evaluated", "text" could mean a few things, etc.)

if you haven't seen it before, the way we do this is by putting something like `{id:checkbox}` at the start of the string. everything in curlies gets stripped out by `lf()`

webapp/src/app.tsx
```diff
@@ -6516,12 +6488,6 @@ document.addEventListener("DOMContentLoaded", async () => {
             initHashchange();
             socketbridge.tryInit();
             electron.initElectron(theEditor);
-            pxt.tickEvent(
-                "accessibilty.accessibleBlocksEnabledForSession",
```
riknoll, line 6520: @abchatra FYI on the removal of this tick

webapp/src/blocks.tsx
```diff
@@ -548,6 +535,17 @@ export class Editor extends toolboxeditor.ToolboxEditor {
          * it to select the first category and clear the selection.
          */
         const that = this;
+        Blockly.FlyoutNavigator.prototype.getOutNode = function(_node?: Blockly.IFocusableNode | null, _bypassAdjustments = false) {
```
riknoll, line 541: just occurred to me that we should really be adding asserts for all the functions we monkey patch to ensure they exist. we've had trouble in the past when we monkey patch something that gets renamed in blockly core.

i wish we could keep this in pxtblocks/monkeyPatches, but i don't think there's a clean way to do it in this case.

webapp/src/shortcut_formatting.ts
```diff
@@ -33,40 +33,165 @@ export enum ShortcutNames {
...
+ * User-facing name for a keycode. Mirrors Blockly's getKeyName but uses pxt's
+ * translation function for translatable strings.
+ */
+function getKeyName(keyCode: number): string {
+  if (keyCode >= 65 && keyCode <= 90) {
+    // letters a-z
+    return String.fromCharCode(keyCode);
+  }
+
+  const keyNames: Record<number, string> = {
```
riknoll, line 65: let's add `{keyboard}` or something like that for the strings here to give the translators some context

webapp/src/shortcut_formatting.ts
```diff
@@ -93,84 +216,28 @@ export function getActionShortcutsAsKeys(
...
   );
   const chosen = hasModifiers ? [currentPlatform[0]] : currentPlatform;
-  return chosen.map((shortcut) => {
-    return shortcut
-      .map((numericOrModifier) => shortcutRegistryKeyNames[numericOrModifier] ?? numericOrModifier)
-  });
+  return chosen.map((shortcut) =>
+    shortcut
+      .map((maybeNumeric) =>
+        Number.isFinite(+maybeNumeric)
```
riknoll, line 198: let's use `parseInt` here

webapp/src/shortcut_formatting.ts
microbit-matt-hillsdon: Reflecting on this, we should probably change these back to Blockly.Msg (aligning with Blockly internals this is copied from) and just address this in the initAccessibilityMessages above. Then the diff from Blockly's code stays small and we get the context in one place only.

webapp/src/shortcut_formatting.ts
microbit-matt-hillsdon: We could do this (I prefer it!), but this is a direct copy of Blockly internals (the change is us aligning it with the new Blockly core version rather than the keyboard nav version). As there's no behaviour change for the inputs used here I'm not sure a nit like this is worth the additional delta.

microbit-matt-hillsdon:
Spotted keyboard controls help is missing shortcuts, likely due to id changes. We should fix up those here but do a separate PR for new help changes.

pxtblocks/loader.ts
microbit-matt-hillsdon: I've tried this in 4231c8ce496a601b4b4267464ffd24f60db4b161. It feels quite a case-by-case judgement call so please do give feedback or just tweak them.

webapp/src/blocks.tsx
microbit-matt-hillsdon: Done in 75b37e11b7ddbfb3ca8c7c9641b76e927b5de206, but I've not extended this to other places in pxtblocks as that would be a large diff.

I started with a "patchMethod" function but the diff/churn was quite high and type safety could only be maintained with faffy generics so seems easier to just assert before a simple assignment. I suggest any further assertions are tackled separately.

webapp/src/shortcut_formatting.ts
microbit-matt-hillsdon: Done as part of I've tried this in https://github.com/microsoft/pxt/commit/4231c8ce496a601b4b4267464ffd24f60db4b161 which reverts these to Blockly.Msg and adds the context in initAccessibilityMessages.

microbit-matt-hillsdon:
> Spotted keyboard controls help is missing shortcuts, likely due to id changes. We should fix up those here but do a separate PR for new help changes.

Fixed in 69d3e46b030276fba33a792a0fe7cbea095ec706.

riknoll (approved):
let's merge and test!

PR #11333, opened 2026-05-27
Ipad toolbox bug fixes

### LINKED ISSUE #6802: [iPad] “…more” toolbox category does not respond to taps in Python and JavaScript editors
**Describe the bug**
When opening any project on iPad, the “…more” category in the Toolbox does not open the corresponding drawer when tapped in the Python or JavaScript editor.

**Steps to reproduce the behavior:**
1. Navigate to https://makecode.microbit.org/beta on **iPad**.
2. Create a new project or open an existing one.
3. Switch to the **Python** or **JavaScript** editor.
4. In the Toolbox, tap the “**…more**” entry under any category.
5. Observe the behavior.

**Actual behavior**
Tapping the "…more" category in the Toolbox produces no response in the Python or JavaScript editor, and the drawer does not open instead of opening the next category on iPad.
 
**Expected behavior**
Tapping the “…more” category should open the toolbox drawer and display its additional blocks in the Python or JavaScript editor, consistent with the behavior in the Blocks editor.

**Screen Recording**
https://github.com/user-attachments/assets/3ad2cb4c-7882-43d4-8b5c-ae5a6576ce9f

**Additional context**
1.OS: iPad Pro 11 26.4.2
2.Browser: Safari
3.Makecode.microbit.org version: 8.1.12
4.Microsoft MakeCode version: 12.3.7

srietkerk:
Fixes https://github.com/microsoft/pxt-microbit/issues/6802

This PR adds support for a pointer up event so that when we prevent default behavior onPointerDown, the pointer up event still gets handled, this fixed the bugs that were happening with the Monaco flyout.

Upload target: https://makecode.microbit.org/app/e532c5c7638ad17654a008d2ab97e2b8e2216b14-a0104c7626


riknoll (changes requested):
i don't think we should revert this focus change. instead, let's figure out why the check on line 500 isn't triggering correctly on ipad

riknoll (approved):

PR #11334, opened 2026-05-28
parse out param screen reader labels from block def for translation

### LINKED ISSUE #6804: User friendly and translatable parameter names to support Blockly screen reader work
**Is your feature request related to a problem? Please describe.**
The incoming Blockly screen reader work will enable screen reader output for individual inputs when they are focused as well as aria-live output when moving blocks to individual inputs. The default behaviour will be similar to this example, although tweaks are expected:

Image

This approach is fine for simple blocks, however, when considering a block with multiple inputs, it will be easy to lose context. This block, for example, will have three number inputs side by side from a keyboard navigation perspective, and only the input values to differentiate them:

Image

If we use the parameter names, as defined in the TypeScript block definitions (modified with this utility `camelCaseToLowercaseWithSpaces`), in the input labels and more importantly, for move mode, we can provide a lot more context for screen reader users. See the example below which does use parameter names instead of the default "number":

Image

For reference, [this is the proposed Blockly API to allow custom aria-labels for inputs](https://github.com/RaspberryPiFoundation/blockly/issues/9752). 

The problem is that the parameter names in the TypeScript block definitions are not always user friendly (e.g., `soundExpression`), and they are not currently translatable. They can also sometimes be unhelpful, or not match what sighted users observe. For example, in the output below, a sensible aria-label would be beat rather than fraction: 

Image

It is worth highlighting that this problem will also be in extensions, where single letter parameter names are often used. 

**Describe the solution you'd like**
We would like a way to pull out block parameter names so that they can be translatable, with the translations used in place of the English text where available. The design also needs to include a way to override the parameter names with user friendly alternatives if the default parameter names are not appropriate. 

For a prototype, we created an optional parameterName.label field / markup to the block definitions. E.g., 
```TypeScript
  //% blockGap=8
  //% fraction.label="beat"
  export function beat(fraction?: BeatFraction): number {
```

This allowed us to override "fraction" with "beat" for the screen reader output.

It is currently unclear how best to generate these strings for translation or how the English text will be substituted. These strings will also need appropriate context in Crowdin. One approach could be a script that pulls out the parameter names, passed them through the `camelCaseToLowercaseWithSpaces` function and adds them as parameter.label, but this feels heavy-handed.

The parameter names / labels will need manual review to decide whether the defaults need overriding as we have suggested doing for "fraction". 

**Describe alternatives you've considered**
There don't appear to be any good alternative behaviours, but the design of the feature is very much open.

**Additional context**
It is not clear at this point what exactly the final screen reader output will be. There is an argument that it should also include the underlying field type, e.g., "127, volume, number, button", but this might be too much information and the type of number may be easily inferred in this case.

The block itself will also have a shortcut for a verbose label output, which reads the block fluently but also injects the input type after the input value. There is concern here that the field type is useful, but the custom aria information we pass in is not. Consider this block label: "Current block: plot x, 0, x, y, 0, y, LED category, has inputs," vs "Current block: plot x, 0, number, y, 0, number, LED category, has inputs,". The latter is obviously better in this case and we will need to keep an eye on how this develops.

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/6804, extract param labels from block and stuffs them into translation files:

image
 
image

spot in loader.ts should be approximately where we do something with it I think, but I'm not up to date on the block screenreader support in general / current state of the blockly update / left that for the moment to check if that's something easy to call or if need to dig into blockly more


abchatra:
@microbit-robert

microbit-robert:
@jwunderl  We'd like to change [this line in loader.ts](https://github.com/microsoft/pxt/blob/master/pxtblocks/loader.ts#L606) to `input.setAlign(Blockly.inputs.Align.LEFT).setAriaLabelProvider(the-translated-param-name);`. But this needs to wait for another Blockly upgrade before this change can be made.

jwunderl:
> @jwunderl We'd like to change [this line in loader.ts](https://github.com/microsoft/pxt/blob/master/pxtblocks/loader.ts#L606) to `input.setAlign(Blockly.inputs.Align.LEFT).setAriaLabelProvider(the-translated-param-name);`. But this needs to wait for another Blockly upgrade before this change can be made.

perfect then, so this would be fine to merge as is for the extraction side (besides updating the comment) and then whenever the new blockly drop happens can update that in `applyInputLabel(..)` as that's exactly where it's called here

pxtblocks/loader.ts
```diff
@@ -1007,6 +1010,13 @@ function namedField(field: Blockly.Field, name: string): NamedField {
     return { field, name };
 }
 
+function applyInputLabel(input: Blockly.Input, parameter: pxt.blocks.BlockParameter) {
+    const label = (parameter.labelLocalizationKey && pxtc.getBlockTranslationsCacheKey(parameter.labelLocalizationKey)) || parameter.label;
+    if (!label) return;
+
+    // Do something to set label on input? check
```
jwunderl:
```suggestion
    // todo: uncomment following line after blockly update adds support
    // input.setAriaLabelProvider(label);
```

^ re: above i believe

microbit-robert:
> > @jwunderl We'd like to change [this line in loader.ts](https://github.com/microsoft/pxt/blob/master/pxtblocks/loader.ts#L606) to `input.setAlign(Blockly.inputs.Align.LEFT).setAriaLabelProvider(the-translated-param-name);`. But this needs to wait for another Blockly upgrade before this change can be made.
> 
> perfect then, so this would be fine to merge as is for the extraction side (besides updating the comment) and then whenever the new blockly drop happens can update that in `applyInputLabel(..)` as that's exactly where it's called here

Yes, this makes sense to me. I've run this locally with a later Blockly version and it appears to work as expected, thanks!

I'm not sure off the top of my head if the `setAriaLabelProvider` API is available in Blockly v13.0.0-beta.5 (current MakeCode beta version), but there are definitely issues with the the custom input labels being used incorrectly. So yes, if it's left as a commented out line, we can uncomment it when we do the next upgrade (and remove the outdated comment).

srietkerk (approved):
LGTM

jwunderl:
Ah something got merged in while this had auto merge on so it didn't go in, since i saw the blockly update went in i enabled arialabelprovider @microbit-robert

microbit-robert:
@jwunderl Thanks. Should be good!

PR #11337, opened 2026-05-29
Add thumb fast paths for switch equality and array length

humanapp:
Two more small fast-path optimizations that further decrease bytecode size and improve performance, this time in: 1) equality tests in `switch` statements, and  2) reading `array.length`. In both cases the compiler now handles the common case directly in a few inline instructions, and falls back to the C++ helper for anything unusual. Code gets smaller and faster, no change in behavior.

This optimization reduced the MicroCode hex file size by ~9k.

### What changes

These changes are scoped to the ARM/microbit code emitter. JS is not affected.
1. **Switch equality**: A `switch` statement compiles to a series of equality checks. The new fast path compares two tagged integers directly. If either operand isn't a tagged integer, it defers to the C++ helper exactly as it did before.
2. **Array.length**: `array.length` currently calls into C++ and then converts the result (from a tagged int to a machine int). The new path reads the length field directly from an array object and returns it ready-to-use, skipping the C++ call and the conversion. Anything that isn't a plain array object defers to C++.

### Risk assessment

I'd give this one a **low/low-moderate** risk. It carries a similar misclassification risk as the boolean fast-path optimizations, but is much narrower, as it only checks for tagged integers. A larger risk is future-proofing the `array.length` fast path. It no longer calls into the runtime for the value, it just reads the field directly straight out of a memory offset. `length` is at a known offset, but that could change if someone inserts a field ahead of it in `RefCollection` (the class that underpins arrays). To mitigate this I've added a `STATIC_ASSERT` to pxtbase.h in pxt-common-packages that will catch any change in the offset (will open a separate PR for that).

cc: @thomasjball


riknoll (approved):
this one seems pretty safe to me

jwunderl (approved):

PR #11342, opened 2026-06-01
Fixes for local github cache expiration, projects loading while offline

### LINKED ISSUE #6686: Cached extensions are unavailable after an hour, for importing hex files while offline.
**Describe the bug**

Arising from https://support.microbit.org/helpdesk/tickets/94068 (private)

Cached extensions are unavailable after an hour, for importing hex files while offline.

A year or so ago, the teacher was able to prepare their class set of Chromebooks, by leaving one file in the homepage, so that extensions would remain cached for importing other hex files while offline. 

"I work in many different schools across Hertfordshire and can't guarantee an internet connection. I therefore need the ability to import projects offline in order to teach my micro:bit workshops.".

They identified that it's currently not possible to import hex files while offline, from an hour after the computers were set up.

I am testing with Windows Chrome. I reset MakeCode before testing - not sure if that's necessary.

Having loaded two hex files using the BitBot extension, I can see cached pkgs in the github table. After an hour or so, when importing offline fails, the cached pkgs have gone.

A project that is left in the homepage can still be loaded. I have just noticed the extensions are also in the scripts table.

**To Reproduce**
1. Go to https://makecode.microbit.org/
2. Import both test hex files: Not-This-One and Test-1 [test-extension-cache.zip](https://github.com/user-attachments/files/24839515/test-extension-cache.zip)
3. Delete Test-1 but leave Not-This-One in the homepage.
4. Close MakeCode
5. Disconnect from internet
6. Wait at least an hour
7. Go to https://makecode.microbit.org/
8. Try to import Test-1
9. See error

**Expected behavior**

Extensions cached for a project in the homepage are available for importing other hex files while offline.

Related: https://github.com/microsoft/pxt-microbit/issues/5047

**Screenshots**

Image

Image

Image

Image

not hardware related

 - OS: [e.g. iOS] Chrome OS, Windows
 - Browser [e.g. chrome, safari] Chrome
 - Version [e.g. 22] 141-144

jwunderl:
Fix https://github.com/microsoft/pxt-microbit/issues/6686 in first commit
Should be a fix for https://github.com/microsoft/pxt-microbit/issues/5047 in follow up two commits, which are basically a refinement of my very old draft fix https://github.com/microsoft/pxt/pull/9529

in both cases i'm gonna need to test this a bit; i'll test the first issue before merging at least.

first commit: fix caching so that we do not purge everything > 60 minutes old on load from github db, and fall back to cache if expired but network unavailable.

follow up 2 commits: patch in catching info for extensions to generated hexfiles, so that they can be used as a backup if available when loading file with extensions while offline

build: https://makecode.microbit.org/app/97d79fec959f349ff930dc28daa1c7bde82c41a7-f2529e968c


pxtlib/github.ts
```diff
@@ -145,13 +145,22 @@ namespace pxt.github {
 
     export interface CachedPackage {
         files: Map<string>;
+        backupCopy?: boolean;
```
riknoll: what is the purpose of this property?

cli/cli.ts
```diff
@@ -276,8 +276,8 @@ class FileGithubDb implements pxt.github.IGithubDb {
         return this.loadAsync(repopath, tag, "pxt", (r, t) => this.db.loadConfigAsync(r, t));
     }
 
-    loadPackageAsync(repopath: string, tag: string): Promise<pxt.github.CachedPackage> {
-        return this.loadAsync(repopath, tag, "pkg", (r, t) => this.db.loadPackageAsync(r, t));
+    loadPackageAsync(repopath: string, tag: string, backupScriptText?: pxt.Map<string>): Promise<pxt.github.CachedPackage> {
```
riknoll: calling this `backupScriptText` is confusing to me... maybe `localCache`? `inMemoryCache`? or even `fallbackCache`? same applies to all the other places where backup is used

i typically read backup as a verb, so this kept reading to me as an optional boolean

pxtlib/github.ts
```diff
@@ -289,41 +298,42 @@ namespace pxt.github {
...
+                }
+                current.files[pxt.CONFIG_NAME] = pkg
+                const cfg: pxt.PackageConfig = JSON.parse(pkg)
+                await U.promiseMapAll(pxt.allPkgFiles(cfg).slice(1), async fn => {
+                    current.files[fn] = await downloadTextAsync(repopath, sha, fn);
+                });
+
+                // cache!
+                this.packages[key] = current;
+                return U.clone(current);
```
riknoll: i know this isn't changed code, but best not to use clone here. would be better to just use
```
return {
  files: {
    ...current.files
  }
}
```

iirc clone just calls JSON.stringify and JSON.parse, which is, uh, not the most memory efficient way of doing things especially when we already have a problem with too big scripts.

pxtlib/cpp.ts
```diff
@@ -1514,6 +1514,19 @@ namespace pxt.hexloader {
             })
     }
 
+    export function stringifyHexInfoForCache(hexInfo: pxtc.HexInfo) {
+        if (!hexInfo?.hex) return undefined;
+
+        const cachedMeta = U.clone(hexInfo);
```
riknoll: again, prefer to not use clone if possible

riknoll (dismissed):

webapp/src/workspace.ts
```diff
@@ -770,11 +770,79 @@ export async function installAsync(h0: InstallHeader, text: ScriptText, dontOver
...
+    if (!packagedExtensionFiles) return;
+
+    const config = await pxt.packagesConfigAsync();
+    const scriptCache = await getScriptCacheAsync();
+    const seen: pxt.Map<boolean> = {};
+
+    await Promise.all(Object.keys(packagedExtensionFiles).map(async key => {
+        if (pxt.github.isGithubId(key)) return;
+
+        const pubId = key.slice(0, 4) === "pub:" ? key.slice(4) : key;
```
srietkerk, line 801: Will these numbers always hold? Would it be helpful to split at least the 4 into a variable explaining why it's 4?

srietkerk (approved):
One small nit, but LGTM

webapp/src/workspace.ts
jwunderl: it'll be 4 always because it's trimming the 4 pub: characters at the start yeah

PR #11349, opened 2026-06-04
offline app webusb start

jwunderl:
I might adjust slightly, as i currently have it auto selecting on electron proper side, but mostly here


srietkerk (approved):
LGTM

PR #11351, opened 2026-06-04
make tabbing in the monaco editor navigate instead of inserting a character

srietkerk:
Addresses https://github.com/microsoft/pxt-microbit/issues/6859

This is a draft PR because this needs to be discussed within the team first. Right now in monaco, the toggle (Ctrl+M) to use tab to continue to navigate the page is available, but the user has to know about it. The default behavior is tabbing will insert a tab in the editor space. The proposition in this issue, and which has been implemented in this PR, is to make "tab navigation mode" be the default in the monaco editor. As in, if the user is focused on the workspace and tab, they will continue to navigate the page rather than insert a tab in the document. I think this is pretty big breaking behavior to have this on by default, but copilot cli did this well earlier, and I figured if we do want to ship this, I should have this available.

Upload target: https://makecode.microbit.org/app/68bc469e9693f2c198f3b1007f0bd65e82f9325f-9ec3d8863d#

### DIFF
--- a/docs/accessibility.md
+++ b/docs/accessibility.md
@@ -122,7 +122,7 @@ To close a dialog, press the `Escape` key or navigate to the close button (shown
 ### JavaScript and Python editor navigation
 
 These keyboard shortcuts are used in the JavaScript editor:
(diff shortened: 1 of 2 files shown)


srietkerk:
Not proceeding with this.

PR #11352, opened 2026-06-04
prevent infinite loop when rendering very short gate lengths

riknoll:
fixes three bugs

the first was reported on the forum here:

https://forum.makecode.com/t/new-song-format/17763/27

when rendering a sound for an instrument where the attack length is greater than the combined gate length and release time, you get an infinite loop. this was just an oversight in the logic

the second bug was an incongruity that could happen when the the gate length was less than the attack+decay time. previously the release would always count down from sustain; now it counts down from whatever value the envelope had at the end of the gate length. this should prevent some clipping for short gates

the third bug had to do with sounds that generated extremely short steps between time points when rendering. previously we just skipped over those points, but that changes the overall duration of the sound. now we keep the duration the same and just don't emit an extra instruction

i'll be opening a corresponding PR in common packages to fix this on the hardware side

### DIFF
--- a/pxtsim/sound/song.ts
+++ b/pxtsim/sound/song.ts
@@ -58,9 +58,13 @@ namespace pxsim.music {
                 timePoints.push(nextALTime);
                 nextALTime += ampLFOInterval;
             }
+            else if (time < gateLength) {
+                time = gateLength;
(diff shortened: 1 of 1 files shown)


srietkerk (approved):
LGTM

PR #11356, opened 2026-06-08
Add single letter shortcut to toggle simulator start / stop

microbit-robert:
This PR:
- adds a single letter shortcut to the blocks workspace to toggle the simulator running state (start / stop)
- adds this shortcut to the keyboard controls help documentation
- adds an aria-live region and util to use it
- uses the aria-live region to announce the simulator status when toggled
- prevents focus from moving to the simulator after playing or restarting the sim - this behaviour is unhelpful if you want to restart the sim more than once. If the user needs to focus the sim, they can tab backwards or use the area menu

This only targets the blocks workspace.

See https://github.com/microsoft/pxt-microbit/issues/6617 (this issue should remain open if  a similar shortcut should be added to the Monaco editor - not single letter for obvious reasons)

We based this work off of feedback from user testing. The preference of the users was to be able to start / stop the simulator rather than mute / unmute without leaving the workspace, and aria-live region feedback was requested. We completed user testing with this shortcut in place and screen reader users found it valuable.


webapp/src/blocks.tsx
```diff
@@ -688,6 +688,21 @@ export class Editor extends toolboxeditor.ToolboxEditor {
...
+            name: "toggle_simulator",
+            keyCodes: [Blockly.ShortcutRegistry.registry.createSerializedKey(Blockly.utils.KeyCodes.S, null)],
+            preconditionFn: (workspace, scope) => {
+                if (workspace.isFlyout || !scope.focusedNode) {
+                    return false
+                }
+                return true;
+            },
+            callback: () => {
+                this.parent.startStopSimulator({ clickTrigger: true });
```
riknoll, line 704: perhaps we should rename this clickTrigger option if it's also going to be used for keyboard shortcuts

riknoll (approved):

webapp/src/blocks.tsx
microbit-robert: `userTriggered`? I think following this change all the way through is a reasonable amount of diff, but happy to proceed if you're happy?

microbit-robert:
Merge conflicts addressed.

webapp/src/blocks.tsx
riknoll: userTriggered sounds good to me, but we can do it later. not a new issue!

PR #11362, opened 2026-06-08
Improve focus position after interaction with mutator buttons (keyboard only)

microbit-robert:
This PR improves the focus / selection position for keyboard users after interaction with mutator buttons. The current behaviour moves focus back to the source block which can be very frustrating. If possible, the same button now maintains focus, if this is not possible, focused button will switch from a plus to a minus button and vice versa based on visibility.

For `if` blocks, the nearest condition above is focused when a branch is removed. The first node on the newly created branch is focused when a branch is added.

These changes do not impact mouse users where the source block takes focus / selection.


pxtblocks/plugins/logic/ifElse.ts
```diff
@@ -82,6 +83,10 @@ const IF_ELSE_MIXIN = {
             this.elseCount_--;
         };
         this.update_(update);
+        // Focus the condition of the last elseif branch, fallback to the condition of the if.
+        const focusIndex = this.elseifCount_;
+        const inputName = focusIndex === 0 ? 'IF0' : 'IF' + focusIndex;
```
riknoll: unecessary ternary here? also below

pxtblocks/plugins/text/join.ts
```diff
@@ -97,7 +102,19 @@ const TEXT_JOIN_MUTATOR_MIXIN = {
             }, Blockly.config.bumpDelay);
         }
         if (block.rendered && block instanceof Blockly.BlockSvg) {
-            block.queueRender();
+            block.queueRender().then(() => {
```
riknoll, line 105: is this double render intentional? if so, can we get a comment why?

riknoll (changes requested):

pxtblocks/plugins/logic/ifElse.ts
microbit-robert: It sure is, cleaned up [d42f688](https://github.com/microsoft/pxt/pull/11362/commits/d42f6885a8facee81ee60e6e7d0e64ddbec3e37e)

pxtblocks/plugins/text/join.ts
microbit-robert: Unintentional, fixed [2ff56ab](https://github.com/microsoft/pxt/pull/11362/commits/2ff56abbce1dc3aecbd8ba92c7645146d045251a)

microbit-robert:
@riknoll merge conflicts addressed as well.

riknoll (approved):

PR #11368, opened 2026-06-08
Use dynamic aria-label for flyout list based on toolbox category

microbit-robert:
This improves the screen reader experience when navigating into the flyout and adds confirmation that you're in the correct category.

E.g.:
- list blocks -> list basic blocks
- list snippets -> list basic snippets


riknoll (approved):

microbit-robert:
@riknoll merge conflict now addressed

PR #11376, opened 2026-06-08
Hide chaff when a toolbox category is selected

microbit-matt-hillsdon:
Selecting a React toolbox category no longer dismissed open field editors, dropdowns, context menus etc. the way stock Blockly does.

This worked incidentally before the Blockly v12->v13 upgrade. In v12, Toolbox.onClick_ called hideChaff(true) unconditionally for any left-click that bubbled to the toolbox container, even when no Blockly ToolboxItem was resolved (our React rows never resolve one). v13 moved that hideChaff(true) call inside the `if (item?.isSelectable())` guard, so it stopped firing for our React toolbox clicks.

Restore it explicitly. Put it in showFlyoutInternal_, the funnel every flyout passes through (regular categories, search, top blocks, and the customClick Variables/Functions categories that bypass showFlyout), plus closeFlyout for the toggle-closed case.

Also stop field-editor open/close (field_editor_open) and field-value BlockChange events from hiding the flyout. Custom editors like the "play melody" note grid fire these asynchronously when dismissed; closing such an editor by clicking a category would otherwise fire after the new flyout is shown and immediately hide it. Neither event represents a workspace interaction that should close the toolbox flyout. TBH this filter might be less confusing if inverted as I'm not really sure what events should close the flyout but that seems a riskier change.


riknoll (approved):
i was just about to look into this

PR #11382, opened 2026-06-09
Scope toolboxes to editor to ensure unique ids

microbit-robert:
This is required to ensure that screen reader works with both the blocks and monaco toolboxes.

This also improves semantics for the extensions tree item which tells users that it will open a dialog.

### DIFF
--- a/webapp/src/toolbox.tsx
+++ b/webapp/src/toolbox.tsx
@@ -12,7 +12,6 @@ import Util = pxt.Util;
 import { fireClickOnEnter } from "./util"
 import { DeleteConfirmationModal } from "../../react-common/components/extensions/DeleteConfirmationModal"
 import { classList } from "../../react-common/components/util"
-import { Aria } from "react-modal"
 
 export const enum CategoryNameID {
     Loops = "loops",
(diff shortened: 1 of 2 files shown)


webapp/src/toolbox.tsx
```diff
@@ -12,6 +12,9 @@ import Util = pxt.Util;
 import { fireClickOnEnter } from "./util"
 import { DeleteConfirmationModal } from "../../react-common/components/extensions/DeleteConfirmationModal"
 import { classList } from "../../react-common/components/util"
+import { Aria } from "react-modal"
+
+type AriaHasPopup = boolean | "dialog" | "menu" | "false" | "listbox" | "grid" | "tree" | "true";
```
riknoll: we should move this to some central location and update the typing you added in sui.tsx to reference it as well

riknoll (changes requested):

webapp/src/toolbox.tsx
microbit-robert: Turns out there is a built-in we can use instead, so have removed the custom type and used this in toolbox and sui.tsx [9ffc483](https://github.com/microsoft/pxt/pull/11382/commits/9ffc4836f3439a0f122e427c1aa6453e4bd5e422)

riknoll (approved):

PR #11385, opened 2026-06-09
make sure config is up to date before calculating extinfo

### LINKED ISSUE #4378: Project with Bluetooth extension - security options are not taking effect
**Describe the bug**
It looks like the Bluetooth settings are not being set correctly for the hex compilation. For example, choosing passkey pairing does not result in a passkey being required.
 
Failure to require a passkey was not the original reported issue, but was one I could repeat. The original report was that a passkey was required even though just works was selected, and that micro:bit appeared not to be using a secure connection. Further details: https://github.com/lancaster-university/microbit-dal/issues/481

Everything seems to behave as expected using https://makecode.microbit.org/v2.

**To Reproduce**
1. Go to 'https://makecode.microbit.org/'
2. Click on 'New project' and give it a name
3. Add the Bluetooth extension
4. Enter code below
5. Go to Settings and choose "passkey pairing"
6. Download using a full flash by saving to the MICROBIT drive
7. Enter pairing mode and bond with nRF Connect (for example)
8. See error - no passkey required

```
bluetooth.onBluetoothConnected(function () {
    basic.showString("C")
})
bluetooth.onBluetoothDisconnected(function () {
    basic.showString("D")
    basic.pause(5000)
    basic.clearScreen()
})
bluetooth.startAccelerometerService()

```

**Expected behavior**
The Settings should affect the compiled hex.

**Screenshots**
Add screenshots to help explain your problem. You can copy paste the screenshot in the github report. The .gif screen recording is very useful as well. 

micro:bit V1.5. Repeated some aspects with a micro:bit V2.

 - OS: [e.g. iOS] Windows, iOS, Android
 - Browser [e.g. chrome, safari] Windows Chrome
 - Version [e.g. 22] NA

**Smartphone (please complete the following information):**
 - Device: [e.g. iPhone6] iPad Air 2, Samsung SMT-580
 - OS: [e.g. iOS8.1] iOS 14.8, Android 8.1
 - Browser [e.g. stock browser, safari] NA
 - Version [e.g. 22] NA

**Additional context**

### LINKED ISSUE #5416: bluetooth problems
Problem One:
I choose "No Pairing Required: Anyone can connect via Bluetooth" to make my microbit program.
The same program have different conditions on microbit v1 and v2.
https://drive.google.com/file/d/1InGlHbB7_ZxIO_FFYtJubk1rWTqvN6aw/view?usp=sharing
Microbit v1 asks password. Without password, I cannot pair microbit v1.
However, I don't set it.

Screenshot_20231128_200622_Settings

However, microbit v2 doesn't ask any password, so I can pairing microbit v2.

Screenshot_20231128_200801_Settings

Problem TWO:
Bluetooth option one and Bluetooth option two are upside down.
When I choose "No Pairing Required", I still need pair microbit.
When I choose "JustWorks pairing (default)", I don't need pair microbit.

image

### LINKED ISSUE #5996: Bluetooth device settings don't take effect when expected
**Describe the bug**
I am building a bluetooth project that needs to be 'open security' (no pairing required)
Image

There's a very strange inconsistency with when this setting actually gets applied. For example, if I build after changing the setting, initially the micro:bit behaviour doesn't change. After I do some set of things I can't identify the significance of, I get the new pairing settings
For example, in one case I had an extension that used bluetooth:
Image
In this configuration, the 'open' pairing was not taking effect
By adding the 'accelerometer service' (which had already been initialised in 'setup as input micro:bit', so this was a duplicate) and then downloading the code to my micro:bit the behaviour changed

**To Reproduce**
Here's a reproducer that works for me without my custom extension:

* create a new MakeCode project
* Add bluetooth extension (but leave all settings as normal)
* Use https://googlechrome.github.io/samples/web-bluetooth/device-info.html?namePrefix=BBC+micro%3Abit to verify that the device is visible in the Bluetooth dialog pairing mode, but not in 'normal' mode [this is expected]
* You can skip these steps, but they were my way of verifying that the right CODAL config where Bluetooth is enabled, not radio, had been applied
    * bond from phone, confirm this works
    * flash from phone - no need to enter pairing mode
        * IE: bluetooth extensions is working right 
* Change project settings in MakeCode to 'no pairing required', save
* Download code to micro:bit
* use https://googlechrome.github.io/samples/web-bluetooth/device-info.html?namePrefix=BBC+micro%3Abit to verify that the device is visible in the Bluetooth dialog pairing mode, but not in 'normal' mode [this is NOT expected]
   * micro:bit is NOT visible when not in pairing mode, but the open setting means it should be
* Add any block from the Bluetooth library, redownload
   * Still not visible. (this is weird because in my previous case with my extension, this change DID make the micro:bit visible)
* Add data logging extension, redownload: 
* Test in the same page, but the micro:bit is now visible even when not in pairing mode.

My expectation is that as soon as I toggle that setting, the next time I flash the micro:bit the setting will be applied
I also don't understand what is causing/triggering the rebuild for it to work

CC @martinwork perhaps you've seen this @microbit-carlos  has some theories!

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/5996
fixes https://github.com/microsoft/pxt-microbit/issues/5416
fixes https://github.com/microsoft/pxt-microbit/issues/4378

changes to the yotta config weren't being propagated into the extinfo we send to the compile worker. the root cause seemed to be that the config in memory on the mainpkg wasn't being re-parsed from the source. this fixes that by calling loadconfig explicitly in getExtInfo

upload target: https://makecode.microbit.org/app/a11e88665242c03f4c3fce840289aa8ee15b4fc5-a9b7f8066f

@microbit-carlos @martinwork can you try the above upload link and make sure it works with your test projects? my dev laptop is having webusb problems for some reason. i'm going to try another computer

### DIFF
--- a/pxtlib/cpp.ts
+++ b/pxtlib/cpp.ts
@@ -164,6 +164,8 @@ namespace pxt.cpp {
         let disabledDeps = ""
         let mainDeps: Package[] = []
 
+        mainPkg.loadConfig();
+
         // order shouldn't matter for c++ compilation,
         // so use a stable order to prevent order changes from fetching a new hex file
         const mainPkgDeps = mainPkg.sortedDeps(true)
--- a/pxtlib/cpp.ts
+++ b/pxtlib/cpp.ts
@@ -164,7 +164,7 @@ namespace pxt.cpp {
         let disabledDeps = ""
         let mainDeps: Package[] = []
 
-        mainPkg.loadConfig();
+        mainPkg.loadConfig(true);
 
         // order shouldn't matter for c++ compilation,
         // so use a stable order to prevent order changes from fetching a new hex file
--- a/pxtlib/package.ts
+++ b/pxtlib/package.ts
@@ -330,7 +330,7 @@ namespace pxt {
                 })
         }
 
-        loadConfig() {
+        loadConfig(skipSave = false) {
             if (this.level != 0 && this.invalid())
                 return; // don't try load invalid dependency
 
@@ -340,7 +340,9 @@ namespace pxt {
             this.parseConfig(confStr);
             if (this.level != 0)
                 this.installedVersion = this.version()
-            this.saveConfig()
+            if (!skipSave) {
+                this.saveConfig()
+            }
         }
 
         protected validateConfig() {


riknoll:
alright i was able to test this on my mac. not sure what's up with my dev laptop but it refuses to let me use webusb. maybe some weird policy

aznhassan (approved):
LGTM

PR #11386, opened 2026-06-09
warn on function ref without invokation as statement

### LINKED ISSUE #6912: Calling a function without () is allowed but does nothing
A simple mistake I have done a few times: Forgetting the paranthesis when calling a function. It does not give an error, stil does nothing. 

Could we be warned when we forget?

function doSomething () {
basic.showNumber(0)	
}
basic.forever(function () {
doSomething //forgot the ()
})

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/6912

not compile / simulator blocking, just renders as a warning

image

Gated the errors behind a compile flag that's only set on typecheck, since i know there are plenty of places like embed scenarios that check diags and may handle it as a proper error otherwise. Feels like there should be some other simple things we can use same flag for adding "probably wrong" warnings on. If it's anything mildly ambiguous would probably have to add an 'ignore' option with a button to fix (add tsignore like comment?) though

### DIFF
--- a/pxtcompiler/emitter/emitter.ts
+++ b/pxtcompiler/emitter/emitter.ts
@@ -4248,11 +4248,27 @@ ${lbl}: .short 0xffff
             if (!node || node.kind != SK.Identifier)
                 return
 
+            if (hasPrecedingTsIgnoreComment(node))
+                return
+
             const exprType = typeOf(node)
             const signatures = exprType && !(exprType.flags & TypeFlags.Any)
                 && checker.getSignaturesOfType(exprType, SignatureKind.Call)
             if (signatures && signatures.length)
                 warning(node, 9284, lf("Function reference used as a statement; did you mean to call it with '()'?"))
+        }
+
+        function hasPrecedingTsIgnoreComment(node: Node) {
+            const src = node.getSourceFile()
+            const line = ts.getLineAndCharacterOfPosition(src, node.getStart(src)).line
+            if (line <= 0)
+                return false
+
+            const lineStarts = src.getLineStarts()
+            const previousLineStart = lineStarts[line - 1]
+            const previousLineEnd = lineStarts[line]
+            const previousLine = src.text.slice(previousLineStart, previousLineEnd)
+            return /^\s*\/\/\s*@ts-ignore\b/.test(previousLine)
         }
 
         function emitExprAsStmt(node: Expression) {
--- a/pxtcompiler/emitter/emitter.ts
+++ b/pxtcompiler/emitter/emitter.ts
@@ -247,7 +247,7 @@ namespace ts.pxtc {
         pxt.log(stringKind(n))
     }
 
-    // next free error 9284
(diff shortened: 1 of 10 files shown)


pxtcompiler/emitter/emitter.ts
```diff
@@ -4241,8 +4241,25 @@ ${lbl}: .short 0xffff
             return v
         }
 
+        function maybeWarnOnBareFunctionReference(node: Expression) {
+            if (!opts.enhancedErrors)
+                return
+
+            if (!node || node.kind != SK.Identifier)
```
srietkerk: Should this be 
```suggestion
            if (!node || node.kind !== SK.Identifier)
```

webapp/src/monaco.tsx
```diff
@@ -1737,9 +1737,14 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         if (file && file.diagnostics) {
             const model = monaco.editor.getModel(monaco.Uri.parse(`pkg:${file.getName()}`))
             for (let d of file.diagnostics) {
+                const severity = d.category == ts.pxtc.DiagnosticCategory.Error
+                    ? monaco.MarkerSeverity.Error
+                    : d.category == ts.pxtc.DiagnosticCategory.Warning
```
srietkerk: Are the `==` here and on line 170 intentional or should they be `===`?

srietkerk (approved):
Just some questions about the type of equality checks being done. Otherwise looks good.

webapp/src/monaco.tsx
jwunderl:
```suggestion
                    : d.category === ts.pxtc.DiagnosticCategory.Warning
```

typescript 'solves' the typical common reason to use === with type coercion (e.g. 1 == "1" situations) b/c it enforces the types compile time so it doesn't make much of a difference (basically only reasonable situation it'll chagne things is == null also checking for undefined, but that's often intentional), but i'll swap em over just for the sake of it~

webapp/src/monaco.tsx
```diff
@@ -1737,9 +1737,14 @@ export class Editor extends toolboxeditor.ToolboxEditor {
         if (file && file.diagnostics) {
             const model = monaco.editor.getModel(monaco.Uri.parse(`pkg:${file.getName()}`))
             for (let d of file.diagnostics) {
+                const severity = d.category == ts.pxtc.DiagnosticCategory.Error
```
jwunderl:
```suggestion
                const severity = d.category === ts.pxtc.DiagnosticCategory.Error
```

jwunderl:
went ahead and did the 'ignore' button and ts-ignore handling
image

test build https://makecode.microbit.org/app/8d6bd6844490945b12c1e4784e9addeba1f374db-2711a50ff9#pub:_iKu4WcEYaazT

srietkerk (approved):
Nice addition!

PR #11391, opened 2026-06-10
Improve screen reader output for show leds block

microbit-robert:
This PR:
- Provides aria information for the show leds block
    - Adds a label that includes how many LEDs are on
    - Adds grid popup semantics (not exactly what a sighted user sees, but a consistent approach across field editors)
- Updates individual LED labels to indicate if they are on or off
- Addresses an id clash which is a result of using sourceBlock_.id
    - The same id is used for the flyout and workspace for the first inserted show leds block - this causes aria-activedescendant to fail causing screen reader problems when navigating inside the grid
-  Adds a `updateFocusIndicator` method to ensure that the two tone focus indicator is updated when the LEDs are toggled on and off (unintentionally broken by https://github.com/microsoft/pxt/pull/11301/changes#diff-84524797c860e262a8416d54b981fa43859564638b17676b040481db4ee05b7dR150-R152)

### DIFF
--- a/pxtblocks/fields/field_matrix.ts
+++ b/pxtblocks/fields/field_matrix.ts
@@ -304,9 +304,12 @@ export abstract class FieldMatrix extends Blockly.Field {
         }
     }
 
-    protected getCellId = (x: number, y: number) => `${this.sourceBlock_.id}:${x}-${y}`;
+    // sourceBlock._id can clash between flyout and main workspace
+    protected domId = Blockly.utils.idGenerator.getNextUniqueId();
 
-    protected getRowId = (index: number) => `${this.sourceBlock_.id}:row-${index}`;
+    protected getCellId = (x: number, y: number) => `${this.domId}:${x}-${y}`;
+
+    protected getRowId = (index: number) => `${this.domId}:row-${index}`;
 
     protected abstract attachPointerEventHandlersToCell(x: number, y: number, cellRect: SVGElement): void;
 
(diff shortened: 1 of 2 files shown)


pxtblocks/fields/field_matrix.ts
```diff
@@ -231,6 +232,28 @@ export abstract class FieldMatrix extends Blockly.Field {
...
+                width: cellWidth + 2,
+                height: cellWidth + 2,
+                rx: `${Math.max(2, cornerRadius)}px`,
+                stroke: "#000",
+                "stroke-width": 2,
+                fill: "none",
+                "aria-hidden": "true"
+            });
+        } else {
+            cell.children[3]?.remove();
```
riknoll: hmmm are we sure that subclasses of this matrix field will always have the same number of elements? might be better to tag this child with a data attribute and search for it instead of hardcoding an index

riknoll (changes requested):

pxtblocks/fields/field_matrix.ts
microbit-robert: We are not. This commit uses a data attribute to keep track of the two tone focus rect. It also fixes a bug where a two tone focus rect was removed when it was still wanted. [8aa194b](https://github.com/microsoft/pxt/pull/11391/commits/8aa194b6a3b6b976e28f4d6b9a20574e3a930db3)

riknoll (approved):

PR #11394, opened 2026-06-10
Update keyboard controls help

### LINKED ISSUE #6890: Keyboard navigation: toolbox shortcut is confusing
With keyboard navigation on by default, I found it really important open the documentation. I tried toggling "T" when I saw that it could take me to the toolbox and found it confusing that I did not get taken to the toolbox. After discussion with other team members, it was revealed that the "T" shortcut only works if you are already focused on the workspace. I find this slightly confusing. I thought I should be able to get to the toolbox with T from anywhere in the editor and found it frustrating that I had to do the extra focusing step first. 

If you have to focus on the workspace first to access the toolbox, I'm not sure the value of having a shortcut. But it's also possible that rephrasing the side docs could help this. All the other sections of the keyboard controls docs have titles describing what the controls apply to. If the first section had the title "Moving in the Workspace" or something like that, it could help clarify that the keyboard controls defined in the first section only happen in the workspace.

microbit-matt-hillsdon:
Group shortcuts by where they apply (Global, Workspace, Toolbox, Blocks, Move Mode) and add newer Blockly shortcuts: block stack navigation, toolbox headings, move-stack, block info, and screen reader mode.

Gate the download and simulator rows on target capability, adding pxt.canDownload() (shared with the editor toolbar's compile button). Please review this bit carefully, I don't have full context for other targets.

~~Draft for now as I want feedback from Foundation folks more involved in the user testing and keyboard controls resources (which has some similar presentations of keyboard shortcuts). Thoughts welcome.~~

Demo: https://scoped-keyboard-controls.review-pxt.pages.dev/#editor

Closes https://github.com/microsoft/pxt-microbit/issues/6890 (or at least helps, it's quite subjective!)

### DIFF
--- a/pxtlib/main.ts
+++ b/pxtlib/main.ts
@@ -321,6 +321,24 @@ namespace pxt {
         pxt.debug(`hwVariant: ${hwVariant} (${hwName})`)
     }
 
+    /**
+     * Whether the current target compiles user code to a file the user can
+     * download (hex, PNG or UF2).
+     */
+    function compilesToDownloadableFile(): boolean {
+        const compile = pxt.appTarget.compile;
+        return !!(compile?.hasHex || compile?.saveAsPNG || compile?.useUF2);
+    }
+
+    /**
+     * Whether a download action is available, either because the target
+     * produces a downloadable file or an extension overrides the download
+     * button.
+     */
+    export function canDownload(): boolean {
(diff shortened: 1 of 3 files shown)


microbit-matt-hillsdon:
It's going to be a while before we get internal feedback so thinking it's better to move ahead with this and then tweak as needed.

riknoll (approved):

PR #11396, opened 2026-06-10
fix --multi hash decoding and add saner options

### LINKED ISSUE #5397: ---multi# magic URL to open two shared projects
**Is your feature request related to a problem? Please describe.**
When using the radio function in a classroom setting with the simulator, it is intuitive to use the https://makecode.microbit.org/---multi# editors and, for example, have the sender program in the left editor window and the receiver program in the right.

**Describe the solution you'd like**
I'd like to be able to make a link that opens 2 shared projects, the first in the left window and the second in the right window, maybe something like this:
https://makecode.microbit.org/---multi#_YTbKYAAimAxz&_LxghRpHJaKq8

**Describe alternatives you've considered**
I have found no alternatives.

**Additional context**
The feature has been discussed on the forum:
https://forum.makecode.com/t/editor-magic-url-flags/10743/5?u=vegz78

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/5397

as noted in that issue, the way to use the --multi page was to include a hash like this:

```
makecode.microbit.org/--multi#lefthash:|:righthash
```

where `lefthash` will end up being applied to the left iframe and `righthash` to the right iframe. however, because the delimeter includes the character `|`, it was often getting encoded automatically as `%7c` which breaks the parsing. this fixes that bug by calling `decodeURIComponent` on the hash to make sure all the percent encoded characters are parsed.

still, this is imo a silly way to structure that URL, so i've also added a new format that is easier to use. you can now add arguments using query string syntax like so:

```
makecode.microbit.org/--multi#left=lefthash&right=righthash
```

or, if you're just referencing two scripts, you can do this:

```
makecode.microbit.org/--multi#left=leftscript=_xxxxxxxxx&rightscript=_xxxxxxxxx
```

which will automatically add the `pub:` prefix to the hash for the iframes

### DIFF
--- a/docs/multi.html
+++ b/docs/multi.html
@@ -146,10 +146,30 @@
                 ifrm.src = select.value + flags + h;
             }
             function handleHash() {
-                var route = (window.location.hash || "#");
-                var parts = route.replace(/^#/, '').split(':|:', 2);
-                updateSrc(left, selectLeft, parts[0]);
-                updateSrc(right, selectRight, parts[1]);
+                var route = decodeURIComponent(window.location.hash || "#");
+
+                if (route.indexOf(":|:") !== -1) {
+                    var parts = route.replace(/^#/, '').split(':|:', 2);
+                    updateSrc(left, selectLeft, parts[0]);
+                    updateSrc(right, selectRight, parts[1]);
+                }
+                else {
+                    route = route.replace(/^#/, '');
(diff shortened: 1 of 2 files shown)


srietkerk (approved):
LGTM

jwunderl (approved):

PR #11397, opened 2026-06-11
allow tutorials and examples to remove dependencies from blocksprj

### LINKED ISSUE #4058: An error about adding Bluetooth extension to tutorial project
**Describe the bug**
When you create a tutorial project and want to write a tutorial about how to use Bluetooth extension, make sure that "radio" has been removed from "pxt. json", report the following two errors:
* Oops, an error occured as we were loading the tutorial.
* conflict on yotta setting microbit-dal.bluetooth.enabled between extensions radio and bluetooth

**To Reproduce**
1. Go to ' https://makecode.microbit.org/ '
2. Create a GitHub project
3. Make sure "main.blocks" and "main.ts" and “test.ts” have been removed
4. Make sure that "radio" has been removed from "pxt. json"
5. Create a “tutorial.md”,the code such as ' https://github.com/Wind-stormger/bluetooth_tutorial_test/blob/master/tutorial.md '
6. Make sure to commit and push changes and create release
7. Enter tutorial project Preview,such as' https://makecode.microbit.org/#tutorial:github:wind-stormger/bluetooth_tutorial_test/tutorial '
8. See error:
* Oops, an error occured as we were loading the tutorial.
* conflict on yotta setting microbit-dal.bluetooth.enabled between extensions radio and bluetooth

**Expected behavior**
I expect to use Bluetooth extension correctly in tutorial project.

**Screenshots**
None

Specify it's not hardware related.

 - OS: [windows 10 20H2 19042.985 (64 bit)]
 - Browser [Microspft Edge 90.0.818.56 (official build) (64 bit)]

**Additional context**
None

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/4058

for tutorials and markdown examples, dependencies specified using the `package` snippet are merged into the dependencies of the default project. that's usually the behavior you want unless the package you're adding happens to conflict with a dependency of the default project.

really, this only comes up in pxt-microbit where the bluetooth extension is not compatible with the default radio extension. up until now, there hasn't been any way to write a tutorial for bluetooth.

this PR fixes that issue by allowing you to remove packages from the default config like so:

````markdown
```package
radio=remove
bluetooth
```
````

as part of this change, i also had to fix a bunch of places in the CLI where we were lazy and added `blocksprj` as a dependency to the config we use when compiling blocks snippets instead of doing the right thing and cloning its pxt.json.

i *think* i got all of the places where the `package` snippets are parsed in the codebase, but it's possible i could have missed one. it would be a good idea to test some docs pages just in case

### DIFF
--- a/cli/cli.ts
+++ b/cli/cli.ts
@@ -5839,7 +5839,9 @@ export async function buildCoreDeclarationFiles(parsed: commandParser.ParsedComm
     nodeutil.mkdirP(builtFolder);
     process.chdir(cwd);
 
-    const host = shareId ? new Host() : new SnippetHost("decl-build", { "main.ts" : "" }, { "blocksprj": "*" });
+    const blocksprjConfig = readBlocksprjConfig();
+
+    const host = shareId ? new Host() : new SnippetHost("decl-build", { "main.ts" : "" }, { ...blocksprjConfig.dependencies });
     const mainPkg = new pxt.MainPackage(host);
 
     if (shareId) {
@@ -6223,6 +6225,16 @@ function checkDocsAsync(parsed?: commandParser.ParsedCommand): Promise<void> {
     )
 }
 
+function readBlocksprjConfig(): pxt.PackageConfig {
+    const configPath = path.join("libs", pxt.BLOCKS_PROJECT_NAME, "pxt.json");
+
+    if (!nodeutil.fileExistsSync(configPath)) {
+        return undefined;
+    }
+    const config = nodeutil.readJson(configPath) as pxt.PackageConfig;
+    return config;
+}
+
 function checkFileSize(files: string[]): number {
     if (!pxt.appTarget.cloud)
         return 0;
@@ -6261,6 +6273,8 @@ function internalCheckDocsAsync(compileSnippets?: boolean, re?: string, fix?: bo
     // only check each snippet once.
     const existingSnippets: pxt.Map<boolean> = {};
     let snippets: CodeSnippet[] = [];
+
+    const blocksprjConfig = readBlocksprjConfig();
 
     const maxFileSize = checkFileSize(nodeutil.allFiles("docs", { maxDepth: 10, allowMissing: true, includeDirs: true, ignoredFileMarker: ".ignorelargefiles" }));
     if (!pxt.appTarget.ignoreDocsErrors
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: cli/cli.ts [hunk 1]
```diff
@@ -5839,7 +5839,9 @@ export async function buildCoreDeclarationFiles(parsed: commandParser.ParsedComm
     nodeutil.mkdirP(builtFolder);
     process.chdir(cwd);
 
-    const host = shareId ? new Host() : new SnippetHost("decl-build", { "main.ts" : "" }, { "blocksprj": "*" });
+    const blocksprjConfig = readBlocksprjConfig();
+
+    const host = shareId ? new Host() : new SnippetHost("decl-build", { "main.ts" : "" }, { ...blocksprjConfig.dependencies });
```
srietkerk, line 5844: I think we need some checking around this because if `readBlocksprjConfig` returns undefined, trying to expand `blocksprjConfig` will throw an error.

cli/cli.ts
```diff
@@ -6778,21 +6793,20 @@ export function getCodeSnippets(fileName: string, md: string): CodeSnippet[] {
...
-        "blocksprj": "*"
-    }
-    snippets.filter(snip => snip.type == "package")
+    const tutorialDeps = snippets.filter(snip => snip.type == "package")
         .map(snip => snip.code.split('\n'))
-        .forEach(lines => lines
-            .map(l => l.replace(/\s*$/, ''))
-            .filter(line => !!line)
-            .forEach(line => {
+        .reduce((acc, lines) => {
```
srietkerk, line 6798: Just curious why this was changed?

srietkerk (dismissed):

cli/cli.ts
jwunderl: throwing in cli on an invalid project build seems fine, though maybe an explicit exception would narrow in a bit quicker

jwunderl (approved):

cli/cli.ts
riknoll: if you run the cli from the wrong directory, i'm pretty sure you'll hit another error before this point

srietkerk: I'm not sure how that's relevant to my comment. I feel like that scenario would be us giving an error to inform whoever is running the cli. The error that we would get here would be a typescript error because you'd be trying to access a field from something undefined. We wouldn't want to get to that point in any scenario, right?

riknoll: what i'm saying is that every target is required to have a blocksprj directory, and if you were trying to run the docs commands from a directory that did not have a target then the CLI would have hit an exception before getting to this point, so it's moot.

cli/cli.ts
riknoll: so that i could use the pxt.tutorial.mergeTutorialDependencies function which has to the dependency removing functionality

PR #11401, opened 2026-06-11
Workaround for codegen issue on older Safari

microbit-matt-hillsdon:
In Safari < 18 `pxt staticpkg --minify` builds, I see getBoardView return a value but undefined is received by initAsync as the first parameter. This causes initAsync to bail, fail to init the driver and the initial header load to error on the undefined driver.

This doesn't happen with `pxt serve` so is likely triggered by the minification (which itself seems correct).

As an example of a `pxt staticpkg --minify` deployment with this issue, see https://blockly-v13-0-0-beta-9.review-pxt.pages.dev/. It doesn't seem to happen on the current beta, perhaps due to some difference in the build setup vs staticpkg (we also have npm linked pxt). A real puzzle.

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -1127,7 +1127,7 @@ export class ProjectView
 
     public async componentDidMount() {
         this.allEditors.forEach(e => e.prepare())
-        await simulator.initAsync(getBoardView(), {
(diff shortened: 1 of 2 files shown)


riknoll (approved):
huh, weird

PR #11406, opened 2026-06-12
extracting locales for default string arguments on blocks

### LINKED ISSUE #6687: Translations missing for some default values
**Describe the bug**
We have spotted two cases where the default values for blocks are not translated:

- The default value for the show string block is always "Hello!" - [see definition](https://github.com/microsoft/pxt-microbit/blob/master/libs/core/basic.cpp#L26)
-  The default value for the first parameter of the radio send value block is always "name" - [see definition](https://github.com/microsoft/pxt-common-packages/blob/master/libs/radio/radio.ts#L302)

**To Reproduce**
1. Go to https://makecode.microbit.org/#editor
2. Change the language to something other than English, e.g., Spanish
3. Open the basic category
4. Observed that the default value for "show string" remains "Hello!" instead of "Hola"

**Expected behavior**
These default values should be translated, as they are for the text category (though a different implementation).

**Screenshots**
Image

Not hardware related.

- macOS Sequoia 15.7.3
- Chrome Version 143.0.7499.111 (Official Build) (arm64)

**Additional context**
It is possible that there are other default values defined using "eg: ..." that are also untranslated.

@jaustin

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/6687

extracting out .defl= and e.g. defaults we use for parameter defaults (it looks like still some not checked in for e.g. unit conversions and color helpers from build), but these are the ones it'll add into microbit / will check in updated locales files after getting this in to update baseline:
image
image
image
image

### DIFF
--- a/pxtlib/blocks.ts
+++ b/pxtlib/blocks.ts
@@ -91,6 +91,28 @@ namespace pxt.blocks {
         localizationKey: string;
     }
 
+    export function parameterDefaultLocalizationKey(qName: string, actualName: string) {
+        return qName ? `${qName}|param|${actualName}|defl` : undefined;
+    }
+
+    export function parameterDefaultToLocalizationString(defaultValue: string, type?: string) {
+        if (!defaultValue) return undefined;
+        if (type === "string" && defaultValue.charAt(0) !== "\"") return defaultValue;
+        if (defaultValue.charAt(0) !== "\"") return undefined;
+
+        try {
+            const value = JSON.parse(defaultValue);
+            return typeof value === "string" ? value : undefined;
+        }
+        catch (e) {
+            return undefined;
+        }
+    }
+
+    export function localizationStringToParameterDefault(value: string) {
+        return JSON.stringify(value);
+    }
+
     // Information for blocks that compile to function calls but are defined by vanilla Blockly
     // and not dynamically by BlocklyLoader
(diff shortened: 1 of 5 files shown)


pxtlib/blocks.ts
```diff
@@ -91,6 +91,28 @@ namespace pxt.blocks {
         localizationKey: string;
     }
 
+    export function parameterDefaultLocalizationKey(qName: string, actualName: string) {
+        return qName ? `${qName}|param|${actualName}|defl` : undefined;
+    }
+
+    export function parameterDefaultToLocalizationString(defaultValue: string, type?: string) {
+        if (!defaultValue) return undefined;
+        if (type === "string" && defaultValue.charAt(0) !== "\"") return defaultValue;
```
srietkerk, line 100: Should this be 
```suggestion
        if (type === "string" && defaultValue.charAt(0) === "\"") return defaultValue;
```

jwunderl: the is the 'bail out if the default value is not explicitly a string' check, e.g. if this is set to a blockid for a block that returns a string

srietkerk: Ah, gotcha, thanks for the clarification!

srietkerk (approved):

PR #11408, opened 2026-06-12
make the variable validator wording closer to the function validator

srietkerk:
As discussed, I've changed the copy for the variable usage criterion. This aligns with what is said for the function validator, and is clearer about included variables.

Part of a fix for https://github.com/microsoft/pxt-microbit/issues/6917, but probably should keep the issue open if we are interested in making a separate "just looking for all project variables" validator.


Jaqster (approved):
Much clearer!  Thanks Sarah!

riknoll (approved):

PR #11411, opened 2026-06-15
unify extensions dialog filters into one tablist

riknoll:
contains a fix for https://github.com/microsoft/pxt-microbit/issues/6835

image

this pr simplifies the extension dialog UI by:
* unifying all of the various views into one tab list (previously it was a hodgepodge of multiple tab layers and hidden state)
* removing the header/breadcrumbs above the extension card list

as part of this change, the default view is now a tab called "Recommended" which is the same list of extensions as the old recommended view that showed when no filters were selected. having it be a tab means we can now properly implement the tablist role (and it's just more consistent)

here's an uploadtarget:

https://makecode.microbit.org/app/fdcb8f163db73e1f5177dba3443b81201594c036-7cf6be6cc7

### DIFF
--- a/react-common/components/controls/TabList.tsx
+++ b/react-common/components/controls/TabList.tsx
@@ -1,0 +1,88 @@
+import { classList } from "../util";
+import {  Button, ButtonViewProps } from "./Button";
+
+export interface TabListProps {
+    className?: string;
+    manualActivation?: boolean;
+    orientation: "horizontal" | "vertical";
+    selectedId: string;
+    onTabSelected: (id: string) => void;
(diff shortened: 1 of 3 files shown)


Jaqster (approved):
LGTM.  My only nit-pick is that blue button color.  Where does that come from?  It doesn't match the blue of the theme.  And it's not the purple accent color.

abchatra:
I love the simplification. Always wondered why we have bredcrumbs.

riknoll:
@Jaqster yeah, i might fix that. that color is what's in live, wasn't introduced by this PR

PR #11413, opened 2026-06-17
Keyboard help tweaks

microbit-matt-hillsdon:
- Unify the Control modifier label on "Ctrl" (was a mix of "Ctrl" and "Control") with a "Control" aria-label so screen readers don't mangle "Ctrl" (which we've chosen for space).

- Format the "Move anywhere" modifier as a key so it gets that label too.

- Avoid showing the bare menu key for the context menu shortcut on Windows: prefer the first modifier-bearing shortcut (e.g. Ctrl+Enter) because not every keyboard has a menu key. The menu shortcut was new in v13.

### DIFF
--- a/webapp/src/components/KeyboardControlsHelp.tsx
+++ b/webapp/src/components/KeyboardControlsHelp.tsx
@@ -180,7 +180,7 @@ const Key = ({ value }: { value: string }) => {
             break;
         }
         case CONTROL_KEY_SHORT: {
-            aria = lf("Control");
+            aria = Blockly.Msg['CONTROL_KEY'];
             break;
         }
     }
--- a/webapp/src/components/KeyboardControlsHelp.tsx
+++ b/webapp/src/components/KeyboardControlsHelp.tsx
@@ -23,9 +23,6 @@ const KeyboardControlsHelp = () => {
             <table>
                 <tbody>
                     <Row name={lf("Show/hide shortcut help")} shortcuts={[LIST_SHORTCUTS_SHORTCUT]} />
-                    <Row name={lf("Screen reader mode")} shortcuts={[names.TOGGLE_SCREENREADER]}>
-                        <p className="hint">{lf("Additional audio cues and navigation aids for screen reader users")}</p>
-                    </Row>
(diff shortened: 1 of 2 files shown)


webapp/src/components/KeyboardControlsHelp.tsx
```diff
@@ -177,6 +179,10 @@ const Key = ({ value }: { value: string }) => {
             aria = lf("Option");
             break;
         }
+        case CONTROL_KEY_SHORT: {
+            aria = lf("Control");
+            break;
+        }
```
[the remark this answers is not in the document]
microbit-matt-hillsdon: Fair enough, the one in initAccessibilityMessages has context so that might cause a duplicate.

webapp/src/components/KeyboardControlsHelp.tsx
```diff
@@ -10,20 +10,19 @@ const KeyboardControlsHelp = () => {
     React.useEffect(() => {
         ref.current?.focus()
     }, []);
-    const ctrl = lf("{id:keyboard symbol}Ctrl");
+    const ctrl = CONTROL_KEY_SHORT;
     const cmd = isMacPlatform ? "⌘" : ctrl;
     const orAsJoiner = lf("or")
     const enterOrSpace = { shortcuts: getShortcutKeysShortAll(names.PERFORM_ACTION), joiner: orAsJoiner}
+    // Split around {0} so the modifier renders as a <Key> (for its aria-label).
+    const moveAnywhere = lf("Hold {0} and press arrow keys").split("{0}");
```
riknoll: there's actually a function for this in react-common/components/util.tsx:

https://github.com/microsoft/pxt/blob/e9cdb397070bbad1d5c4523e4a81825d401a5a17/react-common/components/util.tsx#L15

here's an example of usage:

https://github.com/microsoft/pxt/blob/e9cdb397070bbad1d5c4523e4a81825d401a5a17/skillmap/src/components/AppModal.tsx#L584

riknoll (changes requested):
one tiny comment!

webapp/src/components/KeyboardControlsHelp.tsx
microbit-matt-hillsdon: Ah, thanks. Adapted in 4fbad89b86389722af9c12e31c6d0602961cf354.

riknoll (approved):

PR #11421, opened 2026-06-24
Detach recycled flyout blocks from the canvas

### LINKED ISSUE #6946: [Safari] Navigating up past extensions category with voice over enabled freezes page for up to a minute
In safari with voice over on:

1. tab to the blocks workspace
2. press T to focus the toolbox
3. use the down arrow key to navigate down the list into the advanced section
4. use the up arrow key to navigate up the list past the "Extensions" button
5. the page freezes when you press up on the "Extensions" button for up to a minute, then it unfreezes and rapidly goes through every key you hit in the meantime

safari version 26.5 (21624.2.5.11.4)
macOS 26.5.1

microbit-matt-hillsdon:
The flyout block recycler kept recycled blocks parented to the flyout canvas (hidden), so the canvas accumulated every block shown as the user moved between categories. That interacts poorly with the flyout's use of aria-owns on the canvas under VoiceOver/Safari: reopening a closed flyout gets progressively slower and can hang for tens of seconds once a screen reader has activated accessibility for the page.

Detach the block's SVG from the canvas while it sits in the recycle pool and re-attach it on reuse. The BlockSvg stays alive, so recycling still avoids re-rendering; only the parent changes.

@riknoll I think this will leave the benefit of the caching in place but appreciate a sanity check.

An alternative is to remove the use of aria-owns in Blockly. All typical flyout contents are in the right DOM hierarchy but Blockly doesn't require this so has used aria-owns just in case. See https://github.com/RaspberryPiFoundation/blockly/blob/a233ac19fd6b6471694fb41b056a3bf5a42f678a/packages/blockly/core/flyout_base.ts#L668-L679

Fixes microsoft/pxt-microbit#6946

Demo: https://fix-flyout-recycle-voiceover.review-pxt.pages.dev/

### DIFF
--- a/pxtblocks/plugins/flyout/blockInflater.ts
+++ b/pxtblocks/plugins/flyout/blockInflater.ts
@@ -37,6 +37,9 @@ export class MultiFlyoutRecyclableBlockInflater extends Blockly.BlockFlyoutInfla
         if (this.keyToBlock.has(key)) {
             block = this.keyToBlock.get(key);
             this.keyToBlock.delete(key);
+            // Re-attach the SVG we detached in recycleBlock(). The flyout positions
+            // the block during layout, after createBlock() returns.
+            workspace.getCanvas().appendChild(block.getSvgRoot());
         }
 
         block = block ?? super.createBlock(blockDefinition, workspace);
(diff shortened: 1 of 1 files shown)


riknoll (approved):
still seems snappy to me! will have to try out in other targets with heavier flyouts, but i think it'll be fine. at worst, we can limit it to safari only if it causes a significant degradation

PR #11425, opened 2026-06-24
remove hex info packaging

jwunderl:
ran into an issue where in some cases this failed to load back in hex file to app because the hex file writer was only writing bottom 16 bits instead of all 32; it was already a side feature so just removing, packaged extensions should still load just fine but won't be able to download until internet for full hex info. hex file was also noticeably increasing size of full hex, so just doesn't feel worth it -- it would have already been a bit of a stretch as would have required exact cpp match (e.g. if there was a 1 character change on live site vs latest offline app would have failed), so fine to cut, rather than dealing with weird stuff here.

The part that i committed & reverted does work, and should be fine, but is not necessary with the hexfile portion culled / don't see a reason to actually merge it and have to think about it. put it in here just in case so can have the exact change if it does need to be updated, for any reason

(this pr is purely deleting code added in https://github.com/microsoft/pxt/pull/11342)

### DIFF
--- a/pxtcompiler/emitter/hexfile.ts
+++ b/pxtcompiler/emitter/hexfile.ts
@@ -1102,9 +1102,8 @@ _stored_program: .hex ${res}
 
         res += U.uint8ArrayToString([
             metablob.length & 0xff, metablob.length >> 8,
-            binstring.length & 0xff, (binstring.length >> 8) & 0xff,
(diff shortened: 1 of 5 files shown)


aznhassan (approved):
:shipit:

abchatra:
Nice!

PR #11431, opened 2026-06-25
more github editor accessibility fixes

riknoll:
contains a fix for https://github.com/microsoft/pxt-microbit/issues/6851

adds a visual label to the commit message field and replaces it with a react-common input to match our other inputs
image

contains a fix for https://github.com/microsoft/pxt-microbit/issues/6845
contains a fix for https://github.com/microsoft/pxt-microbit/issues/6846

switched to react-common radio button group and made sure all labels are properly associated

contains a fix for https://github.com/microsoft/pxt-microbit/issues/6861

set the aria-label on the github button in addition to the title


react-common/components/controls/RadioButtonGroup.tsx
```diff
@@ -23,10 +24,11 @@ export const RadioButtonGroup = (props: RadioButtonGroupProps) => {
         className,
         ariaHidden,
         ariaLabel,
+        ariaLabelledby,
         role,
         choices,
         selectedId,
-        onChoiceSelected
+        onChoiceSelected,
```
srietkerk, line 31: Should this comma be here?

jwunderl: it's nice to leave trailing commas in lists / etc in ts when they're each on separate lines, because it makes future changes diff better (you don't get a line noted as changed that is just adding , to an existing line) / ts just treats as normal

jwunderl (approved):

PR #11442, opened 2026-06-30
installed ribbons on enabled extensions

### LINKED ISSUE #6974: Extension gallery behavior
Not a regression.
When you add a built-in extension to a project it gets removed from the Extension Gallery.
- try adding Servo to an open project
- now go back to Extension Gallery, it's gone

But when you add a 3rd party extension to a project, it's still in the Extension Gallery
- try adding maqueen (or any other 3rd party extension)
- now go back to Extension Gallery, it's still there

Image

Image

jwunderl:
superseding https://github.com/microsoft/pxt/pull/11437
need https://github.com/microsoft/pxt-microbit/pull/6984 so that it doesn't show those two on every project
fix https://github.com/microsoft/pxt-microbit/issues/6974

image

right now clicking them is basically a no op, can hook that up to a 'do you want to remove this extension' modal and update in future, but we didn't want to sneak that in so close to release / previously had integration issues there

### DIFF
--- a/pxtlib/service.ts
+++ b/pxtlib/service.ts
@@ -1839,6 +1839,7 @@ namespace ts.pxtc.service {
         imageUrl?: string;
         type?: ExtensionType;
         learnMoreUrl?: string;
+        installed?: boolean;
 
         pkgConfig?: pxt.PackageConfig; // Added if the type is Bundled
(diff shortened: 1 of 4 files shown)


aznhassan (approved):
:shipit:

srietkerk (approved):

PR #11449, opened 2026-07-02
add advanced blocks

otizemlanetzag:


abchatra:
Need description and feature details on why this change makes sense.

PR #11452, opened 2026-07-03
fix: work around blockly delete-area hit test at non-default zoom

### LINKED ISSUE #6997: Deleting blocks via dragging doesn't work reliably
Tracker for https://github.com/RaspberryPiFoundation/blockly/issues/10095 which has been fixed in Blockly post-13.1.0 by https://github.com/RaspberryPiFoundation/blockly/pull/10096

MakeCode has default non-1.0 zoom so this needs addressing for the release.

I've asked if Blockly will consider a patch release with their fix, failing that there might be monkey patch options.

Repro:
1. Drag the workspace background to move away from 0,0 (increases size of error giving an easier repro)
2. Drag blocks to the toolbox to delete, often they won't (note they might fall out of the viewport you can listen for the delete sound or zoom out to check)

Potential minimal monkey patch:
```javascript
const {svgMath, Coordinate} = Blockly.utils;
Blockly.WorkspaceSvg.prototype.getDragTarget = function (point) {
  const coordinate =
    point instanceof Coordinate
      ? svgMath.wsToScreenCoordinates(this, point.clone()) // .clone() is the fix
      : new Coordinate(point.clientX, point.clientY);
  for (const area of this.dragTargetAreas) {
    if (area.clientRect.contains(coordinate.x, coordinate.y)) return area.component;
  }
  return null;
};
```

This isn't the same as Blockly's upstream fix which is a bit harder to patch in, but passes their new tests when applied to Blockly and [fits my understanding of the cause](https://github.com/RaspberryPiFoundation/blockly/pull/10096#issuecomment-4875463937).

It seems to fix the issue in MakeCode and I'll open a PR if we can't get a quick release.

microbit-matt-hillsdon:
Monkey patch WorkspaceSvg.getDragTarget to clone the coordinate before the screen round-trip, so successive getDragTarget calls during a drag no longer double-scale a shared, in-place-mutated Coordinate and miss the delete area.

Fixed upstream in blockly 9f4f1c1 (not in the 13.1.0 we're using). https://github.com/RaspberryPiFoundation/blockly/issues/10095

Draft for the moment as I'm hoping we'll get a point release instead.

Closes https://github.com/microsoft/pxt-microbit/issues/6997

Demo: https://monkeypatch-coords-issue.review-pxt.pages.dev/#editor

### DIFF
--- a/pxtblocks/monkeyPatches/deleteArea.ts
+++ b/pxtblocks/monkeyPatches/deleteArea.ts
@@ -1,0 +1,46 @@
+import * as Blockly from "blockly";
+import { assertMethod } from "./util";
+
+/**
+ * Fixes delete-area hit testing at non-default zoom. Fixed upstream in
+ * commit 9f4f1c1, which is not in the blockly 13.1.0.
+ *   Issue: https://github.com/RaspberryPiFoundation/blockly/issues/10095
+ *   PR:    https://github.com/RaspberryPiFoundation/blockly/pull/10096
+ *
+ * Dragger.onDrag/onDragEnd compute the hit-test coordinate once and pass that
+ * same Coordinate object to several getDragTarget calls. getDragTarget runs it
+ * through wsToScreenCoordinates, whose `coordinate.scale(ws.scale)` mutates the
+ * coordinate in place. So the first call scales the shared coordinate by the
+ * zoom factor, the next call scales the already-scaled value again, and it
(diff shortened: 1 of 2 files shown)


microbit-matt-hillsdon:
Closing, there's a Blockly 13.1.1 with this fix in instead. I'll open a PR shortly.

PR #11485, opened 2026-07-15
Fix unreliable Duplicate button in My Assets

KohanMathers:
Fixes microsoft/pxt-arcade#7589

## Problem

The Duplicate button in the My Assets panel would silently do nothing for:
- Unnamed assets
- An asset that was just renamed (unnamed -> named) in the same session
- A brand-new asset with an auto-generated default name (e.g. a new tile named `myTile24`)
- A sprite embedded directly in a block

In all cases, reloading the project made duplication work again.

## Root causes

1. `isGalleryAsset()` could return `undefined` instead of `false` for any asset whose id doesn't start with `"sprites."` and has no `meta.package`  (tiles, temporary block-embedded assets, and others). `duplicateAssetHandler`/`copyAssetHandler` then called `.toString()` on that `undefined`, throwing synchronously before the duplicate/copy logic ever ran. This was the primary cause: a silent, uncaught exception in the click handler.
2. Stale/type-unsafe reselection after any asset-list refresh re-selected the current asset by `id` only, with no `type` check, and no fallback when a promoted (newly-named) asset's id changed.
3. `buildAssetsAsync()` cleared its rebuild flag before attempting work and had no `try/catch`, so any failure silently wedged asset syncing for the rest of the session with no user-visible error.
4. `pushUndo()` could delete the asset being duplicated via its unconditional `cleanupTemporaryAssets()` call, and `cloneAsset()` mutated its input object in place instead of producing an independent copy.

## Fix

- `isGalleryAsset()` now always returns a real boolean.
- Duplicate/edit/create now explicitly reselect the resulting asset instead of relying on implicit state carry-through.
- `buildAssetsAsync()` self-heals its rebuild flag on failure, and failures now surface as a visible error notification instead of failing silently.
- The in-flight selected asset is protected from `cleanupTemporaryAssets()`.
- `cloneAsset()` no longer mutates its input.

## API notes

Two `pxt.TilemapProject` methods gained a new optional trailing parameter, which is non-breaking for all existing callers (all current call sites across `pxtlib`/`webapp`/`pxtblocks`/`pxteditor` already pass zero args):
- `pushUndo()` -> `pushUndo(protectedAssetId?: string)`
- `cleanupTemporaryAssets()` (protected) -> `cleanupTemporaryAssets(protectedAssetId?: string)`

`cloneAsset()`'s signature is unchanged, but its behavior changed: it previously mutated the `meta` object on its input `asset` as a side effect before returning the clone; it no longer does. Any code relying on that mutation (none found in this repo) would need to be updated, but this was an unintentional side effect of what's documented/named as a pure clone function.

Fix was manually verified from a
```bash
cd ../pxt-arcade
pxt link ../pxt
pxt serve
```

### DIFF
--- a/pxtlib/tilemap.ts
+++ b/pxtlib/tilemap.ts
@@ -755,7 +755,7 @@ namespace pxt {
             }
         }
 
-        public pushUndo() {
+        public pushUndo(protectedAssetId?: string) {
             if (this.undoStack.length && this.committedState.revision === this.state.revision) return;
             this.redoStack = [];
             this.undoStack.push({
@@ -770,7 +770,7 @@ namespace pxt {
             }
 
             this.committedState = this.cloneState();
-            this.cleanupTemporaryAssets();
+            this.cleanupTemporaryAssets(protectedAssetId);
         }
 
         public revision() {
@@ -1477,9 +1477,10 @@ namespace pxt {
             return assets;
         }
 
-        protected cleanupTemporaryAssets() {
+        protected cleanupTemporaryAssets(protectedAssetId?: string) {
             const images = this.getAssetCollection(AssetType.Image);
-            const orphaned = images.getSnapshot(image => !image.meta.displayName && !(image.meta.blockIDs?.length));
+            const orphaned = images.getSnapshot(image =>
+                image.id !== protectedAssetId && !image.meta.displayName && !(image.meta.blockIDs?.length));
             for (const image of orphaned) {
                 images.removeByID(image.id);
             }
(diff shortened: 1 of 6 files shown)


KohanMathers:
@jwunderl would you mind taking a look when you get a chance? :)

KohanMathers:
@copilot's comments resolved and microsoft:master merged back in so the PR branch is up to date

KohanMathers:
Hey @riknoll - just waiting on that review :)

riknoll (approved):
lgtm

PR #11486, opened 2026-07-15
[stable13.0] Expose Screen Reader Mode over the iframe API (#11459)

microbit-matt-hillsdon:
Originally #11459, this just missed the branch point for the release but is needed for micro:bit classroom to expose the same setting.

Adds a "togglescreenreadermode" editor message that behaves like the settings menu toggle (including the "settings" telemetry event source), and reports the current state via a new screenReaderMode field on the "info" response.

As part of this change I realised that toggleScreenReaderMode was async in implementation but not interface so I've aligned both to be async.

CC @riknoll — could you include this in a pxt-microbit patch release in the next few days? I think this is the right branch to target but shout if I misunderstood.


riknoll:
will do!

riknoll (approved):

PR #11496, opened 2026-07-22
[stable13.0] Make custom field init() idempotent to fix stuck touch drags 

jwunderl:
port #11491 for microbit stable, have another bug fix for offline app to port / that fix looked good to me to port over as well?


aznhassan (approved):
:shipit:

PR #11512, opened 2026-08-12
fix embedded source for projects with 64 KiB+ of compressed source

### LINKED ISSUE #7062: CreateAI HEX over 64 KiB runs on micro:bit but cannot be imported into MakeCode
**Describe the bug**

A project trained in micro:bit CreateAI works correctly when its exported HEX file is flashed directly to a micro:bit V2. The trained model recognizes all seven movement classes.

However, importing the same HEX file back into MakeCode displays:

> Sorry, we couldn’t recognize this file.

The students followed the official CreateAI workflow correctly. The problem remains after importing the original sample JSON, retraining the model, exporting to MakeCode, and using “Download as File” again.

Technical inspection indicates that the MakeCode/PXT editable source embedded in Intel HEX record type `0x0E` exceeds the 64 KiB boundary.

For the affected file:

- Actual LZMA-compressed embedded source: `74,688 bytes`
- Length recorded in the embedded-source header: `9,152 bytes`
- Difference: exactly `65,536 bytes`
- Decompressed project size: `764,184 bytes`
- The type-`0x0E` address reaches `0xFFF0` and then wraps to `0x0000`

This appears to be a 16-bit length or address overflow:

`74,688 - 65,536 = 9,152`

The executable firmware remains valid, so the HEX can still be flashed and run. However, MakeCode cannot reconstruct the editable project because the embedded compressed source is incomplete or overwritten after the address wrap.

The embedded project contains:

- `autogenerated.ts`
- `dataset.json`
- `main.blocks`
- `main.ts`
- `ml-metadata.json`
- `pxt.json`
- `_packaged-extensions.json`

The largest embedded files are:

- `dataset.json`: 551,230 bytes
- `_packaged-extensions.json`: 161,986 bytes
- `autogenerated.ts`: 10,498 bytes

We confirmed the suspected cause by recovering the complete wrapped source and creating an editable copy in which only the embedded `dataset.json` was replaced with `{}`. The separate CreateAI sample JSON was retained for future retraining.

After this change:

- Compressed embedded source decreased to `33,535 bytes`
- No type-`0x0E` address wrap occurred
- The trained model in `autogenerated.ts` remained intact
- All seven ML event blocks remained intact
- The repaired HEX imported successfully into MakeCode

**To Reproduce**

1. Open micro:bit CreateAI.
2. Create seven movement classes.
3. Record approximately 20–30 movement samples for each class.
4. Train and test the model.
5. Export the trained project to MakeCode.
6. In MakeCode, use “Download as File” to save the project as a HEX file.
7. Flash the HEX directly to a micro:bit V2 and confirm that the trained model works.
8. Return to `https://makecode.microbit.org/`.
9. Select **Import → Import File**.
10. Select the same downloaded HEX file.
11. Click **Continue**.
12. MakeCode reports that it cannot recognize the file.

The issue appears when the LZMA-compressed PXT embedded-source data exceeds 65,535 bytes.

**Expected behavior**

A HEX file generated through the official CreateAI and MakeCode workflow should remain importable into MakeCode.

If embedded editable-source data cannot exceed 64 KiB, CreateAI or MakeCode should detect this before export and do one of the following:

- Store `dataset.json` separately instead of embedding it in the HEX
- Use an embedded-source format that supports data beyond the 64 KiB boundary
- Correctly encode and reconstruct multiple type-`0x0E` sections
- Warn the user before downloading a file that cannot be reopened
- Provide a separate editable project-backup format

Students should not receive an apparently valid project file that runs on the micro:bit but cannot be reopened in MakeCode.

**Screenshots**

Attached screenshots show:

1. MakeCode reporting that it cannot recognize the original CreateAI HEX file.
2. The repaired project successfully opened in MakeCode with all seven ML event blocks visible.

The original failing HEX and repaired HEX can also be attached for comparison.

- micro:bit V2
- The issue concerns CreateAI/MakeCode project export and import, rather than a hardware failure.
- The original HEX runs correctly when flashed to the micro:bit.

- OS: Windows
- Browser: Google Chrome
- Browser version: Current installed version
- MakeCode target: `9.0.8`
- PXT version: `13.0.5`
- Project editor: `blocksprj`
- Machine-learning extension: `pxt-microbit-ml v1.0.14`
- MakeCode URL: `https://makecode.microbit.org/`

**Smartphone (please complete the following information):**

- Device: Not applicable
- OS: Not applicable
- Browser: Not applicable
- Version: Not applicable

The problem was reproduced on a Windows desktop computer.

**Additional context**

This has caused repeated disruption in a classroom environment. Students successfully complete data collection and model training, download the project, and later discover that it cannot be reopened in MakeCode.

Because the same HEX still runs on the micro:bit, the downloaded file appears valid even though its editable project data cannot be recovered through the normal MakeCode interface.

The failure has occurred repeatedly and is not an isolated student error.

The original CreateAI project used seven movement classes with approximately 20–30 samples per class.

A separate sample JSON backup is available. Because it contains the students’ complete movement-training data, it is not being attached publicly, but it can be provided privately upon request.

Files available for reproduction:

- Original CreateAI HEX that runs on the micro:bit but cannot be imported into MakeCode
- Repaired MakeCode-importable HEX
- Original CreateAI sample JSON, available privately
- Screenshot of the MakeCode import error
- Screenshot of the repaired project opened successfully
- Technical comparison of the original and repaired type-`0x0E` records

[正手-刪減版-data-samples.json](https://github.com/user-attachments/files/30569770/-.-data-samples.json)

microbit-matt-hillsdon:
The binary header spec (docs/source-embedding.md) gives the embedded text length a 4-byte field, but packSource only ever wrote the low 2 bytes, leaving the high bytes zero. Projects whose compressed source reaches 64 KiB were silently written with a truncated length (true length mod 65536), so importing the hex read too little data and failed. The data itself was always fully present.

Fix the writer to store all 4 length bytes and the hex reader to combine the high bytes it already captured but ignored. The UF2/bin reader already read all 4 bytes and needs no change.

Add an embed-test suite covering the header layout and write/read round-trips either side of the 64 KiB boundary. packSource is now exported so the tests can drive the real writer.

I felt like I needed some tests in place here before making this change, but they're obviously a matter of taste and end up having implement toHexRecords to bridge the format independent write in packSource to the hex-specific read to get a round trip test. So we can drop or adapt if you prefer. A refactor might get you a shared readSourceHeader(bytes) or some other way to share the read part. But ideally we'd get this in live so I've tried to keep the production code change small.

You can repair the original hex ("microbit-正手抽球學生730 (1).hex" in the zip file) with this diff and it will then open successfully.

```diff
--- bug-original.hex	2026-07-30 11:22:18.000000000 +0000
+++ bug-repaired.hex	2026-08-12 16:43:34.324879572 +0000
@@ -17490,7 +17490,7 @@
 :1000000C42424242424242424242424242424242C4
 :1000000C42424242424242424242424242424242C4
 :0F00000C42424242424242424242424242424207
-:1000000E41140E2FB82FA2BBA7007D2200000000C6
+:1000000E41140E2FB82FA2BBA7007D2201000000C5
 :1000100E7B22636F6D7072657373696F6E223A2205
 :1000200E4C5A4D41222C2268656164657253697A7F
 :1000300E65223A3239332C227465787453697A65A5
```

Closes https://github.com/microsoft/pxt-microbit/issues/7062


jwunderl (approved):
LGTM, I had actually had similar in my branch when making the changes to the compressed source packing but didn't include it in final pr as wasn't necessary in the test files i had & didn't have context on if the empty bits would have any other implications on flashing / if the code was intentional there.

PR #11518, opened 2026-08-13
sim: pass the editor's origin as a simulator extension's parentOrigin

### LINKED ISSUE #11446: Simulator extension receives the embedder's origin as parentOrigin, not the editor's — breaks host→simx messaging when the editor is embedded cross-origin
**Describe the bug**

When the editor starts a **simulator extension** (simx) iframe, it passes the editor's *embedder* origin as the simx's `parentOrigin` (`options.parentOrigin`) instead of the editor's own origin. A simx that validates the origin of incoming `postMessage`s against that value (e.g. jacdac's `IFrameBridgeClient`) then rejects every message the editor forwards to it whenever the editor is embedded **cross-origin** in controller mode - because those messages arrive with the *editor's* origin, which never matches the embedder origin the simx was told to trust. Net effect: nothing the host sends reaches the simulator extension.

**To Reproduce**

Scenario: a host app embeds the editor cross-origin, owns the device connection (e.g. BLE/USB), and relays frames to/from the editor via `postMessage`.

1.  Embed the editor in an iframe on a **different origin** with `controller=2` and `parentOrigin=<host origin>` (so the editor whitelists the host).
2.  Run a project that activates a simulator extension whose iframe validates the origin of inbound messages against its `parentOrigin` (the Jacdac simx / `IFrameBridgeClient` does this).
3.  From the host, post `{ type: "messagepacket", channel: "jacdac", broadcast: true, data }` to the editor iframe.
4.  The simx never receives it; its origin check drops it. (It works when the editor is *not* embedded, because `parentOrigin` is then undefined and the simx falls back to accepting any origin, masking the bug.)

**Expected behavior**

The simulator extension receives the messages the editor forwards to it (so a host-bridged device shows up). Since the simx's parent window is the editor, it should be given the editor's own origin (`window.location.origin`) as `parentOrigin`.

**Screenshots**

N/A --- the symptom is the simx receiving no forwarded frames (its origin check rejects them).

**Desktop:**

-   OS: Windows 11
-   Browser: Chrome 149
-   Version: pxt-core 12.3.7 (behavior identical on `master`)

**Smartphone:** N/A

**Additional context**

In the `simulatorExtension` start path it does:
https://github.com/microsoft/pxt/blob/3aaeddbdb81a0029ae6798db3209e3cacc25146e/pxtsim/simdriver.ts#L429
i.e. the embedder (grandparent) origin, where the simx's actual parent is the editor.

One-line fix candidate: pass **`window.location.origin`** instead of `this.options.parentOrigin` (correct standalone and embedded). Happy to open a PR; open to a different approach if you'd prefer to solve it another way.

Amerlander:
Fixes #11446

### Problem

When the editor spawns a simulator extension (simx) iframe it sets the extension's `parentOrigin` query param. A simx that validates the origin of incoming `postMessage`s -- as [jacdac-docs does](https://github.com/jacdac/jacdac-docs/blob/69cf764a42b1c321d64eb6416a71acd98d4ba97b/src/jacdac/providerbus.ts#L85) -- uses that value to decide which sender to trust.

The editor was forwarding `options.parentOrigin`, but that is the *editor's* embedder origin, set only when the editor is itself hosted inside another page (controller mode). For the simx that origin is its **grandparent**, not its parent: the simx's parent window is the editor, and every simulator message the editor posts into the simx arrives with the editor's origin. So an origin-validating simx rejects all of them whenever the editor is embedded on an origin different from its own.

Repro: embed the micro:bit/Calliope editor cross-origin, use a jacdac block. The simx iframe loads but no message ever passes its origin check, so the jacdac dashboard stays dead.

### Fix

Pass `window.location.origin` -- the origin the simx actually receives messages from -- and set it unconditionally.

Also drops the `encodeURIComponent` calls: `URLSearchParams.set` percent-encodes internally, so the values were encoded twice and a simx reading `params.get("parentOrigin")` saw `https%3A%2F%2F...`, which never matches a real message origin. The same double encoding applied to `language` (harmless in practice, since language codes contain nothing that needs escaping). The `$PARENT_ORIGIN$` substitution on the legacy `messageSimulator` path just below builds its query string by hand, so its `encodeURIComponent` is correct and is left alone.

### Why this is safe

| Case | Before | After |
| --- | --- | --- |
| Editor embedded same-origin | works | unchanged (values are equal) |
| Editor not embedded | param unset, simx accepts any origin | param set to editor origin -- still validates, and tightens the check |
| Editor embedded cross-origin | **broken** | fixed |

### DIFF
--- a/pxtsim/simdriver.ts
+++ b/pxtsim/simdriver.ts
@@ -422,10 +422,16 @@ namespace pxsim {
                             // not found, spin a new one
                             if (!messageFrame) {
                                 const url = new URL(simulatorExtension.url);
-                                if (this.options.parentOrigin)
-                                    url.searchParams.set("parentOrigin", encodeURIComponent(this.options.parentOrigin));
+                                // The extension's parent window is this editor, so the origin it
+                                // should trust is the editor's own. options.parentOrigin is the
+                                // editor's *embedder* origin (controller mode) - the extension's
+                                // grandparent - so an extension that validates messages against
+                                // it rejects everything the editor sends when the editor is
+                                // embedded cross-origin.
+                                // searchParams.set() encodes, so don't encode the values here.
+                                url.searchParams.set("parentOrigin", window.location.origin);
                                 if (this.options.userLanguage)
-                                    url.searchParams.set("language", encodeURIComponent(this.options.userLanguage));
+                                    url.searchParams.set("language", this.options.userLanguage);
                                 startSimulatorExtension(url.toString(), simulatorExtension.permanent, simulatorExtension.aspectRatio);
                             }
                             // not running the current run, restart


Amerlander:
### Verified locally
Tested with the editor served by pxt serve at http://localhost:3232, embedded as a controller=2 iframe in a host page on a different origin (https://localhost:5273 + jacdac/pxt-jacdac, resolved via simxdev to its devUrl).
Reading the spawned simx iframe's query string (new URLSearchParams(...), one decode, no decodeURIComponent):
```
{
  "parentOrigin": "http://localhost:3232",
  "language": "en"
}
```

Before the patch the same iframe received parentOrigin = the embedder's origin (https://localhost:5273), additionally double-encoded (https%3A%2F%2F… after one decode), so the extension's msg.origin === parentOrigin check could never match.

image

abchatra (approved):

abchatra:
@humanapp fyi

jwunderl (approved):

PR #11540, opened 2026-08-25
show loading screen while replacing user code in skillmap tutorials

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/7643

i wasn't able to repro the linked issue, but i think the cause was likely the workspace being interacted with while the project was reloading. this PR adds a loading screen to prevent users from interacting with the workspace while reloading tutorial code.

as a bonus, it's also just a much nicer experience.


aznhassan (approved):
:shipit:

PR #11542, opened 2026-08-25
Fix reset button in skillmaps & hint layout, builtin extension issue

jwunderl:
fix https://github.com/microsoft/pxt-arcade/issues/7625, and just in general clean up tutorial hint layout since that kept jumping off screen for me while testing the bug
https://arcade.makecode.com/app/d300780148c18b0c1054a3498f5f05d5f6482e2d-80c429f213--skillmap#beginner

also fix bug richard saw during live stream; rehydrated api info was storing incorrect value for src (so e.g. libs/corgio instead of corgio). Poked around and should be safe to fix. Additionally, found another edge case, again, while testing (using controller extension, adding a rotary encoder block, then trying to remove) -- made it check all transitive dependencies that will be removed as well
https://arcade.makecode.com/app/d300780148c18b0c1054a3498f5f05d5f6482e2d-80c429f213


aznhassan (approved):
LGTM!

PR #11554, opened 2026-08-27
sim theming editor

jwunderl:
build https://arcade.makecode.com/app/fb7fdcd3b98aa85db6e19d1619281e1d62c3c53d-e3e237117d

(this was copilot written with just a few times sending it back, left it making it based a detailed-ish spec over lunch / team club based off jacqueline asking for it. reading through it in more detail now / only did a quick skim so far, but putting up in case


jwunderl:
https://arcade.makecode.com/app/2e7d98fa08c2a5e80b7660ca1edbfb77e4c685a6-1826fed55e

new build off latest i have for moment, still generated. It did find a few things (e.g. where it would double up syncing the newly set theme upwards in both outer and inner frame) / need to validate all of it, but functionally looks mostly there.

jwunderl:
okay new build https://arcade.makecode.com/app/0625f693773c72e92aa267d37bb248bf4a56c50e-fe41c6499c

Jaqster:
Nice!  Couple thoughts - 
- The downside of the Save button is that you're not able to preview the theme in the background like you were before.  So a feature request might be the ability to somehow preview your theme changes before clicking Save.
- The size of the dialog window changes when toggling between Editor and Sim
- The Save button seems too wide
- Love the color pickers!

jwunderl:
abhijiths tokyo nights them now automatically changing default sim to purple https://arcade.makecode.com/app/93f70f601ca47959d5825307abdc0846ccbd2d18-a83febae79

jwunderl:
https://arcade.makecode.com/app/18ee377efa9d7cc5ace869a7d6cf6a35320850a5-4941f84e82 okay one last build for the day; i resuscitated my old 'retro arcade' skin that was like the prerelease fantasy console, and made it so can pick the different skills in the theme picker itself

jwunderl:
and new build, save -> apply, retro skin, and layouts able to define which color options show up (e.g. retro shows 2 more for screen border colors, and junior shows 2 fewer for missing dpad)

https://arcade.makecode.com/app/aa94fa58ffef4048a1fa6882e902ca1e2aa0eb10-79992481aa

aznhassan:
If you select the `Tokyo Night` theme, it'll also previews the simulator theme as expected. However, if you apply a `Built-in theme` and then apply the `default` Built-in theme the sim theme preview for `Tokyo Night` no longer works. I'd expect having the simulator set to a default `Built-in theme` and default `Layout` should be the same as having never set your simulator preference to begin with. 

If you apply the `Tokyo Night` theme and refresh the page when it's in this state, the simulator theme will now be purple, but the page theme will be whatever it was before. Set a theme and refresh the page and refresh the page after a couple of seconds, often the theme preference isn't saved.

aznhassan (approved):
Looks good. Non-blocking, but take a look at the previous comment I left.

abchatra:
I still see Save in Editor theme and Apply in the Simulator. Can we make both Apply?

jwunderl:
https://arcade.makecode.com/app/28b34016011e26aa06bc9e89a964282c51526dd4-b0c8581be5 < build off last commit, theme page also apply & believe covered the cases for bugs you described (there was an immediate flag so that it wouldn't allow it to be deferred to later sync that i missed, with it as soon as i press apply refreshing works and persists properly)

PR #11575, opened 2026-09-03
fix named asset deletion problems and piano roll wrapping css

riknoll:
fixes https://github.com/microsoft/pxt-arcade/issues/7683
fixes https://github.com/microsoft/pxt-arcade/issues/7681
fixes https://github.com/microsoft/pxt-arcade/issues/7700

this pr makes 5 changes:
* rewrites the `isAssetUsed` function to always use regex scans for named assets and altered the regexes to be a little more permissive
* prevents the drum labels in the piano roll editor from wrapping
* fixes the flex wrap behavior of the music gallery so that it uses balance instead of wrapping
* renamed the "booming kick" to "boom kick" since it was really scraping right up against the width limit
* adds the piano roll as a monaco field editor

the `isAssetUsed` changes should fix all of the problems with not being able to delete assets that you aren't using. the problem was the `blockIds` metadata not being updated when blocks are created/disposed/etc. rather than try and deal with all of the edge cases, i decide to bypass it entirely.

### DIFF
--- a/pxteditor/index.ts
+++ b/pxteditor/index.ts
@@ -12,6 +12,7 @@ export * from "./monaco-fields/field_tilemap";
 export * from "./monaco-fields/field_musiceditor";
 export * from "./monaco-fields/field_soundEffect";
 export * from "./monaco-fields/field_sprite";
+export * from "./monaco-fields/field_pianoroll";
 export * from "./monaco-fields/field_animation";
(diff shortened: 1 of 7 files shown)


riknoll:
updated to also enable the piano roll as a monaco field editor

jwunderl (approved):

PR #11593, opened 2026-09-09
reset theme button, fix builtin extensions still being preferred, sim editor sizing

jwunderl:
* include the toggle for sharing with sim theme always / don't hide when on default
* fix builtins still getting added in as preferred, there was an old line that snuck through that added them separately
* button to reset sitewide theme
* some fixes to sizing so it handles small or wide screens without getting partially obscured / forced into widescreen mode

fix https://github.com/microsoft/pxt-arcade/issues/7735
fix https://github.com/microsoft/pxt-arcade/issues/7736

### DIFF
--- a/react-common/components/theming/useThemeReset.tsx
+++ b/react-common/components/theming/useThemeReset.tsx
@@ -39,7 +39,7 @@ export function useThemeReset(pickerId: string, onReset: (() => Promise<void>) |
     };
 
     const resetAction: ModalAction | undefined = onReset ? {
-        label: lf("Reset all themes"),
+        label: lf("Reset"),
         className: "primary inverted theme-reset-button",
         onClick: () => {
             setFailed(false);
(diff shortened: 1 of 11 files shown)


react-common/components/theming/useThemeReset.tsx
```diff
@@ -0,0 +1,67 @@
...
+            pxt.reportException(error);
+            setFailed(true);
+        } finally {
+            inFlight.current = false;
+            setResetting(false);
+        }
+    };
+
+    const resetAction: ModalAction | undefined = onReset ? {
+        label: lf("Reset all themes"),
```
abchatra: I would just call it Reset. It is obvious that it is resetting theme and we have a confirmation dialog

abchatra (approved):

react-common/components/theming/useThemeReset.tsx
jwunderl:
```suggestion
        label: lf("Reset"),
```

PR #11596, opened 2026-09-09
loop capture per iteration

### LINKED ISSUE #11563: [compiler] Closures capture let loop variables by final value, not per iteration
Closures capture `let` loop variables by final value, not per iteration. Standard TypeScript gives each `for (let i ...)` iteration a fresh binding; static TypeScript shares one binding, so every closure sees the loop's final value.

Repro: https://makecode.com/_1m7JkiYRK5Yw

```ts
const fns: (() => number)[] = []
for (let i = 0; i < 4; i++) {
    fns.push(() => i)
}
console.log(fns.map(f => f()).join(","))
// standard TypeScript: 0,1,2,3
// static TypeScript:   4,4,4,4
```

Confirmed on the simulator backend and on a micro:bit V2; both backends agree, so this is emitter-level scoping, not a backend bug. Workaround: capture through a function parameter (`function mk(v: number) { return () => v }`), which binds per call.

jwunderl:
close https://github.com/microsoft/pxt/issues/11563

build here https://arcade.makecode.com/app/48bf6ee7932fd01859f48cc3cadc2211a3ba2253-baa5c5f82b#pub:S11575-26979-25273-96005; should see rainbow diagonal of numbers, vs https://arcade.makecode.com/#pub:S11575-26979-25273-96005 being set at 14 / black so it shows nothing on screen

### DIFF
--- a/pxtcompiler/emitter/emitter.ts
+++ b/pxtcompiler/emitter/emitter.ts
@@ -3232,9 +3232,12 @@ ${lbl}: .short 0xffff
             if (info.capturedVars.length &&
                 info.usedBeforeDecl &&
                 node.kind == SK.FunctionDeclaration && !bin.finalPass) {
-                info.capturedVars.sort((a, b) => b.pos - a.pos)
+                // Synthetic captures (such as thisParameter) are available before
+                // source declarations, but may not have a source position.
+                const capturePos = (capture: VarOrParam): number => capture.pos == null ? -1 : capture.pos;
+                info.capturedVars.sort((a, b) => capturePos(b) - capturePos(a))
                 const scope = node.parent;
-                if (scope.kind === SK.Block && inLoop(scope) && info.capturedVars[0].pos < scope.pos) {
+                if (scope.kind === SK.Block && inLoop(scope) && info.capturedVars.every(capture => capturePos(capture) < scope.pos)) {
                     // All captures precede this loop block. Defining the function at
                     // the last captured variable would hoist it out of the iteration.
                     const id = getNodeId(scope);
--- a/pxtcompiler/emitter/emitter.ts
+++ b/pxtcompiler/emitter/emitter.ts
@@ -978,6 +978,7 @@ namespace ts.pxtc {
         let currUsingContext: PxtNode = null
         let needsUsingInfo = false
         let pendingFunctionDefinitions: FunctionDeclaration[] = []
+        const loopBlockFunctionDefinitions: pxt.Map<FunctionDeclaration[]> = {};
 
         currNodeWave++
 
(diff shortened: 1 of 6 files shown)


jwunderl:
(@humanapp in case you want to take a look, had some notable gpt assistance but took a stab at both this and https://github.com/microsoft/pxt/pull/11597 based off your last two issues, still validating / need to test hardware but so far think got it)

jwunderl:
image
tested on office gamego

pxtcompiler/emitter/emitter.ts
```diff
@@ -3232,10 +3233,19 @@ ${lbl}: .short 0xffff
                 info.usedBeforeDecl &&
                 node.kind == SK.FunctionDeclaration && !bin.finalPass) {
                 info.capturedVars.sort((a, b) => b.pos - a.pos)
-                const vinfo = getVarInfo(info.capturedVars[0])
-                if (!vinfo.functionsToDefine)
-                    vinfo.functionsToDefine = []
-                vinfo.functionsToDefine.push(node)
+                const scope = node.parent;
+                if (scope.kind === SK.Block && inLoop(scope) && info.capturedVars[0].pos < scope.pos) {
```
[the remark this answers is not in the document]
jwunderl: narrow but looks like real edge case / right, build after fix https://arcade.makecode.com/app/7537abbebd27997ef4e4312ef1c218991a9add97-28ed852db5

humanapp:
Small nit: This PR and https://github.com/microsoft/pxt/pull/11597 both add a test file with numeric prefix `58`. It doesn't break anything, but one of them should probably be `59`.

humanapp:
Fix looks good! The additional work of boxing/unboxing the captured values is unfortunate, but unavoidable and is the expected cost here.

pxtcompiler/emitter/emitter.ts
```diff
@@ -4663,7 +4717,10 @@ ${lbl}: .short 0xffff
         function emitLabeledStatement(node: LabeledStatement) {
             let l = getLabels(node.statement)
             emit(node.statement)
-            proc.emitLblDirect(l.brk)
```
humanapp, line 4666: This is interesting. Did you find a different compiler bug? Were labeled loops broken?

jwunderl: yes, this was a small side case of this emitting the same label a second time while going through it. Probably a case 0 people will ever hit but might as well fix, repro i came up with from description:

```
let count: number = 0;

outer: for (let i = 0; i < 3; i++) {
    count++;
    if (i === 1) break outer;
}

game.splash("Count: " + count);
```

on live hw compilation breaks, fixed in branch 

Image

(and re: test file, i'll swap 11597 to be 59 before merging yeah / after incorporating your suggested change on that one)

thomasjball (approved):

PR #11597, opened 2026-09-09
correct default values for dynamic dispatch

### LINKED ISSUE #11562: [compiler] Default parameter values are not applied on dynamic dispatch
Default parameter values are not applied when a method is called through dynamic dispatch. A call through an interface-typed or `any`-typed reference passes only the arguments written at the call site; the missing ones arrive as `undefined` and the parameter initializer never runs.

Repro: https://makecode.com/_EjWTcAMaqhyX

```ts
interface Shape {
    scaled(x: number, factor?: number): number;
}

class Thing implements Shape {
    scaled(x: number, factor = 5) {
        return x * factor;
    }
}

const t = new Thing();
const s: Shape = t;

console.log(t.scaled(2));   // 10  -- static call, default applied
console.log(s.scaled(2));   // NaN -- same object via interface: default not applied
```

Root cause is in the emitter, not a backend: defaults are filled in at the call site from the statically known signature (`addDefaultParametersAndTypeCheck`). A dynamically dispatched call site has no signature, so nothing fills the argument, and there is no callee-side default application to catch it -- the arity wrapper pads with `undefined`. A fix likely means emitting callee-side default application for methods reachable through dynamic dispatch.

It seems like a not-recent bug: reproduces on pxt v12.3.14 (the release prior to my recent optimizations), and the call-site filling mechanism dates to 2017.

A ready-made red test is checked in disabled: uncomment the REPRO block in `tests/compile-test/lang-test0/57defaultparamdispatch.ts` (added in https://github.com/microsoft/pxt/pull/11561) and `gulp testlang` fails with `qzdp:iface`. The file's active assertions pin the behavior that must hold regardless of the fix (concrete-path defaults, explicit-argument calls, and agreement between the interface and `any` paths).

Workaround: pass every argument explicitly at dynamic call sites, or have the callee test for `undefined` itself.

jwunderl:
fix https://github.com/microsoft/pxt/issues/11562
build and test https://arcade.makecode.com/app/42de182450c50f8d694f1eab5591cfb1fc05cb86-a44715369a#pub:S79673-77916-59274-53267 (same on live prints 10 NaN as expected)

### DIFF
--- a/pxtcompiler/emitter/emitter.ts
+++ b/pxtcompiler/emitter/emitter.ts
@@ -2364,10 +2364,14 @@ ${lbl}: .short 0xffff
                             }
                             args.push(irToNode(expr))
                         } else {
-                            if (!opts.unfetteredInitializers && !isNumericLiteral(prm.initializer)) {
-                                userError(9212, lf("only numbers, null, true and false supported as default arguments"))
+                            // Emitted functions apply their own defaults, including on
+                            // virtual/interface calls where the implementation may differ.
+                            // Shims and helpers may have no emitted parameter prologue.
+                            const needsCallerDefault = attrs.helper || attrs.shim;
+                            if (needsCallerDefault && !opts.unfetteredInitializers && !isNumericLiteral(prm.initializer)) {
+                                userError(9212, lf("only numbers, null, true and false supported as default arguments for shims and helpers"))
                             }
-                            args.push(prm.initializer)
+                            args.push(needsCallerDefault ? prm.initializer : irToNode(emitLit(undefined)))
                         }
                     } else {
                         userError(9213, lf("unsupported default argument (shouldn't happen)"))
@@ -3155,13 +3159,9 @@ ${lbl}: .short 0xffff
                 }
             }
 
-            const destructuredParameters: ParameterDeclaration[] = []
             const fieldAssignmentParameters: ParameterDeclaration[] = []
 
             proc.args = getParameters(node).map((p, i) => {
-                if (p.name.kind === SK.ObjectBindingPattern) {
-                    destructuredParameters.push(p)
-                }
                 if (node.kind == SK.Constructor && isCtorField(p)) {
                     fieldAssignmentParameters.push(p)
                 }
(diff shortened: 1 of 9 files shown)


jwunderl:
image
tested on office gamego

pxtcompiler/emitter/emitter.ts
```diff
@@ -3181,7 +3181,25 @@ ${lbl}: .short 0xffff
...
+            // Arity wrappers pad omitted arguments with undefined. Keep defaults in
+            // the shared body so direct, virtual, interface and action calls all run
+            // them, including native entries which can skip the arity wrapper.
+            const attrs = parseComments(node);
+            proc.args.forEach(l => {
+                const parameter = l.def as ParameterDeclaration;
+                // Simulator shim dummies must match the native caller-side behavior.
+                if (parameter.initializer && !attrs.shim && !attrs.helper) {
+                    const supplied = proc.mkLabel("defaultarg");
+                    proc.emitJmpZ(supplied, ir.rtcall("pxt::eqq_bool", [l.load(), emitLit(undefined)]));
```
humanapp:
```suggestion
                    if (isThumb()) {
                        // Tagged undefined is 0, so the guard is a bare
                        // compare against zero rather than a runtime call.
                        proc.emitJmp(supplied, l.load(), ir.JmpMode.IfNotZero);
                    } else {
                        // The simulator holds real JS values; a raw zero test
                        // would treat a supplied 0/false/null as omitted.
                        proc.emitJmpZ(supplied, ir.rtcall("pxt::eqq_bool", [l.load(), emitLit(undefined)]));
                    }
```

humanapp: Calling `pxt::eqq_bool` for every argument to answer the question "is this value undefined?" will work, but it is expensive at runtime. On thumb, the tagged `undefined` value is just `#0`, so a simple `cmp` should work. No C++ interop overhead.

tests/thumb-test/asmchecks.ts
```diff
@@ -221,6 +228,29 @@ export const asmChecks: pxt.Map<AsmCheck> = {
...
+    const procedure = (name: string): string => {
+        const match = new RegExp("^" + name + "__P\\d+_pre:([\\s\\S]*?); endfun", "m").exec(code);
+        chai.assert(!!match, "missing procedure " + name);
+        return match[1];
+    };
+    for (const name of ["DpProbe_padded", "DpProbe_exact", "DpProbe_capture", "DpCtor_constructor"]) {
+        const body = procedure(name);
+        chai.assert.equal(countMatches(body, /bl pxt::eqq_bool/g), 1, name + " strict undefined guard");
+        chai.assert(/_nochk:[\s\S]*?movs r1, #0\s+mov r7, sp\s+str r7, \[r6, #4\]\s+bl pxt::eqq_bool\s+cmp r0, #0\s+beq \.defaultarg_/.test(body),
+            name + " must check undefined (not truthiness) in the shared body");
```
humanapp:
```suggestion
        // Tagged undefined is 0, so the omitted-argument guard is an inline
        // compare against zero. A runtime call here is a regression.
        assertNoMatch(body, /bl pxt::eqq_bool/g, name + " runtime call in the undefined guard");
        chai.assert.equal(countMatches(body, /bne \.defaultarg_/g), 1, name + " inline undefined guard");
        chai.assert(/_nochk:[\s\S]*?cmp r0, #0\s+bne \.defaultarg_/.test(body),
            name + " must guard the default with an inline compare in the shared body");
```

pxtcompiler/emitter/emitter.ts
jwunderl: good call thank you, nice to avoid runtime cost here

thomasjball (approved):

PR #11609, opened 2026-09-14
side menu for sidedocs, planning whiteboard

jwunderl:
riffing off https://github.com/microsoft/pxt/pull/10888 and adding a saved whiteboard / notes area for planning purposes. 

i haven't looked at code at all for this (why it's a draft) since have a bit of a headache / basically just kicked off idea with a long rambling prompt and complained about things that felt bad / wanted fixed every few minutes when i took a look at it. could potentially see using the mobile behavior (a ... bubble that animates out the options) for all but not sure since it adds extra clicks / on desktop the size isn't that much of a concern. ~~on thing need to think of, right now i have clicking outside dismiss it, but that's not sidedocs behavior -- probably need to either have it as an option (like a pin button that makes it stay open) or just retain the behavior and force user to explicitly press the hide button if they want it gone.~~ implemented

implemented as an experiment and the bug where the experiment banner shifts things wrong and messes up website layout annoyed me so fixed that along the way

https://arcade.makecode.com/app/a838201445c9989555277f1e21c1128dbb184abc-aa661c99d9


Jaqster:
This is cool!  Definitely feels more polished than our side docs, but it does obscure more of the workspace and isn't as responsive.  But I like this idea of some sort of floating (or pinned) window for a variety of things - docs, saved snippets/assets (backpack), AI code explainer/helper?

jwunderl:
Yeah that was goal, having an extensible place to add 'workspace' level pieces like that to add more bubbles for those sorts of things as we make them. had it review existing sizing for side docs & make it so the horizontal bubbles last until the wide desktop breakpoint to try and reclaim some space. It had horizontal resizing but also added indicator / vertical resizing https://arcade.makecode.com/app/39e9673146d93e17e95597856fdbafe42facfc51-9f561c88ba

PR #11611, opened 2026-09-16
block snippet backpack

jwunderl:
i reallly, really need to figure out how to make this stop adding *so* many tests for every little tweak i make....

https://arcade.makecode.com/app/cb07a15575a7e967d3a7c4335496cc5df3db4ad6-98c5cdd261

right click on any block that can hold others, or drag to the backpack icon.

~~seing about implementing another route on backend for storing these individually rather than syncing all as part of account settings like it's currently written to (as that has 64kb cap).~~ implemented but that means new tests on this require staging, so check backend pr for test link.

does some other clean up as part of pr; that bug where the experiments / etc banner messes up css throughout editor is fixed, and copy / paste blocks should have the same checking for extensions / prompt to add if missing


pxtlib/auth.ts
```diff
@@ -511,7 +544,7 @@ namespace pxt.auth {
                 // Fetch latest prefs from remote
                 const getResult = await this.apiAsync<Partial<UserPreferences>>('/api/user/preferences');
                 if (!getResult.success) {
-                    pxt.reportError("identity", "failed to fetch preferences for patch", getResult as any);
+                    pxt.reportError("identity", "failed to fetch preferences for patch");
```
jwunderl:
```suggestion
                    pxt.reportError("identity", "failed to fetch preferences for patch", getResult as any);
```

pxtlib/auth.ts
```diff
@@ -532,7 +565,7 @@ namespace pxt.auth {
                     // Set user profile from returned value so we stay in sync
                     this.setUserPreferencesAsync(patchResult.resp);
                 } else {
-                    pxt.reportError("identity", "failed to patch preferences", patchResult as any);
+                    pxt.reportError("identity", "failed to patch preferences");
```
jwunderl:
```suggestion
                    pxt.reportError("identity", "failed to patch preferences", patchResult as any);
```

pxtblocks/backpack.ts
```diff
@@ -0,0 +1,599 @@
...
+            if (input.shadow) visitStates([input.shadow], visit);
+            if (input.block) visitStates([input.block], visit);
+        }
+        if (state.next?.shadow) visitStates([state.next.shadow], visit);
+        if (state.next?.block) visitStates([state.next.block], visit);
+    }
+}
+
+/** Parse untrusted storage before calling any Blockly loaders or field/mutation hooks. */
+export function parseBackpackCode(code: string): BackpackCode {
```
jwunderl, line 139: this feels a bit overcomplicating it, though reading it does seem to cover good cases. commenting to revisit / see how can narrow it down to be more like, feeling like something we'd actually write

jwunderl:
okay, tons of tests but if you just... ignore those... seems like reasonable spot to me so far. target branch is https://github.com/microsoft/pxt/pull/11609 (that's also where bulk of the lines removed are from since i did the test clean up all at once in here), but that's mainly just if you want extra context on the project tools portion

